PR #3980, opened 2026-05-12
fix(#3362): include unwired entry fields in workflow mapping surface

### LINKED ISSUE #3362: Workflow agent mapping UI and auto-compute ignore unwired entry fields
## Summary

When a user adds a new output field on a workflow's entry node but hasn't yet wired it to a downstream node, that field is invisible to both:

1. The **Edit Workflow Agent drawer**'s scenario-mapping section — the new field does not appear as a mappable input.
2. The **server-side auto-compute mappings** routine that runs on workflow save — the new field is not included in the computed `scenarioMappings`.

Users expect declared entry fields to be mappable regardless of whether they're wired internally yet.

### Root cause

Both code paths derive workflow inputs via `getInputsOutputs(edges, nodes)` in `src/optimization_studio/utils/nodeUtils.ts:60`, which calls `getEntryInputs` — that only returns edges where `source === "entry"`. Unwired entry outputs produce no edge, so they're filtered out. The entry node's own `data.outputs` (the declared fields, source of truth) is never consulted.

### Affected files

- `langwatch/src/components/agents/AgentWorkflowEditorDrawer.tsx:65-96` — `extractVariables` in the edit drawer
- `langwatch/src/server/workflows/auto-compute-agent-mappings.ts:25-50` — `extractVariablesFromDSL` on save

### Prior art / direction

`src/optimization_studio/utils/workflowFields.ts:64` has `getWorkflowEntryOutputs`, which *does* read from `entryNode.data.outputs` directly — but it also filters by downstream connectivity (lines 102-109), so it can't be reused as-is. Likely fix: add an unfiltered variant alongside it (or a new parameter) that returns every declared entry output regardless of edges, and switch both call sites above to use it. End outputs (`endNode.data.inputs`) are already read correctly — no change needed there.

### How discovered

Surfaced during local browser testing of the Edit Workflow Agent drawer from PR #3334 (issue #3201). A newly-added entry field was missing from the mapping section until an edge was drawn from the entry node.

## Acceptance Criteria

- AC 1: Given a workflow entry node with a declared output field that has no downstream edge, when the user opens the Edit Workflow Agent drawer for an agent linked to that workflow, the scenario-mapping UI lists that field as a mappable input.
- AC 2: Given the same workflow, when the workflow is saved (triggering auto-compute), the linked agent's `scenarioMappings` include an entry for the unwired field using best-match defaults.
- AC 3: Regression — wired entry fields continue to behave as today (still included in both the drawer UI and auto-compute).
- AC 4: Regression — evaluator-only wired entry outputs continue to be surfaced (currently via the `optional` flag in `getEntryInputs`); no change in their rendering or auto-compute behavior.

## Plan

Change the input source from edges-only to the entry node's declared `data.outputs`, while preserving the evaluator-only `optional` flag derived from edges.

Approach:
1. In `src/optimization_studio/utils/nodeUtils.ts`, update `getInputsOutputs` (and, if needed, `getEntryInputs`) to seed `inputs` from `entryNode.data.outputs` — every declared field — then overlay edge-derived metadata (`optional` for evaluator-only wiring) onto the matching identifier.
2. Leave `outputs` (end-node inputs) untouched.
3. Both existing call sites (`AgentWorkflowEditorDrawer.tsx` drawer `extractVariables`, and server-side `extractVariablesFromDSL` in the auto-compute path from PR #3334) pick up the fix via `getInputsOutputs` — no call-site changes.

Test strategy:
- New unit tests for `getInputsOutputs` / `getEntryInputs` covering: unwired entry output appears, wired entry output appears once (no dup), evaluator-only edge still yields `optional: true`, mixed wired + unwired case.
- New unit test for the server-side auto-compute path asserting `scenarioMappings` contains the unwired field.

## Related

- Parent: #3201 (workflow agent input/output mapping layer)
- Parent PR: #3334

---
*Classification: Bug*

drewdrewthis:
## Why

Closes #3362.

When a user added a new output field on a workflow's entry node but hadn't yet wired it to a downstream node, that field was invisible to both the Edit Workflow Agent drawer's scenario-mapping section and the server-side auto-compute routine that runs on workflow save. The mapping surface should reflect every declared entry field, regardless of internal wiring — wiring is a workflow-design concern, mappability is a contract concern.

## What changed

The bug lived in `getInputsOutputs(edges, nodes)` in `optimization_studio/utils/nodeUtils.ts`, which derived its `inputs` list purely from edges sourced at the entry node. Unwired outputs produce no edges and were silently dropped. The entry node's own `data.outputs` — the source of truth for declared fields — was never consulted.

The fix extracts the input-derivation logic into a new `getDeclaredEntryInputs` helper that:

1. Reads `entryNode.data.outputs` to enumerate every declared field.
2. For each field, inspects edges from `outputs.<identifier>` to compute the `optional` flag (set when every downstream edge targets an evaluator node).
3. Falls back to the previous edge-based derivation when the DSL omits an entry node — backwards-compat for legacy/synthetic DSLs.

`getInputsOutputs` now delegates `inputs` to this helper. End-node outputs are untouched. The drawer and the auto-compute path both reach this through `getInputsOutputs` and pick up the fix transparently with no call-site changes.

`getEntryInputs` is left unchanged. It is used by runtime-execution paths (`runWorkflow`, `Publish`, `ChatWindow`) that derive *required* inputs from edges only — gaining phantom requirements for unwired fields there would break workflow execution.

## Test plan

Added one BDD feature file with 8 scenarios mapped to the 4 ACs, then three test files:

- `nodeUtils.unit.test.ts` (new) — covers the contract of `getInputsOutputs`. Two scenarios fail today without the fix (`mixed wired and unwired` and `orphan unwired field`); three regression baselines pass (wired, evaluator-only, end-output).
- `auto-compute-agent-mappings.unit.test.ts` (extended) — one new scenario asserting the unwired field reaches `scenarioMappings`. Fails today; passes after fix.
- `AgentWorkflowEditorDrawer.integration.test.tsx` (new) — asserts the drawer renders a mapping row for an unwired declared entry output. Fails today; passes after fix.

All three were committed FIRST as failing tests, the fix is the final commit. Git log is the audit trail.

Broader regression sweep: 797 tests across 73 files (`optimization_studio/`, `server/workflows/`, `server/scenarios/`, `components/agents/`) all pass post-fix. Typecheck clean.

## Anything surprising?

The original issue plan suggested updating `getInputsOutputs` directly. I considered introducing a new helper used by ONLY the drawer + auto-compute (most-scoped), but found that all six current `getInputsOutputs` consumers actually benefit from declared-field semantics — they're config/mapping-surface paths, not runtime-execution paths. The execution-path consumers all use `getEntryInputs` directly, which I left untouched. So the global change to `getInputsOutputs` is safe.

Out of scope for this PR but worth a separate look: the `evaluations-legacy.ts` and `useAvailableEvaluators.ts` `requiredFields` derivation. Custom evaluators that declare-but-don't-wire entry outputs will now expose those outputs as required fields. For evaluator workflows this is probably the correct semantic, but it deserves user-facing validation.

# Related Issue

- Resolve #3362


drewdrewthis:
### Sweep: edge-derived enumeration of declared workflow input fields drops unwired declarations

Checked 30+ candidate sites. **0 must-fix, 1 review, all others clean.**

**Review (1):**
- `src/optimization_studio/utils/workflowFields.ts:64-110` — `getWorkflowEntryOutputs` reads `entryNode.data.outputs` but then filters by edge connectivity (drops outputs not connected to non-evaluator nodes). Used by `evaluator.service.ts`, `evaluations-legacy.ts`, and `evaluators` tRPC router to derive `requiredFields` for workflow-as-evaluator runs. Same bug class as #3362 but a different consumer; the filtering may be intentional per its JSDoc. Worth a separate issue if users hit the same "I added a field but it's invisible" confusion in the evaluator wizard flow.

**Clean:**
- `getEntryInputs` consumers (`runWorkflow.ts`, `Publish.tsx`, `ChatWindow.tsx`) — must stay edge-only (runtime-execution paths).
- `edgeMappingUtils.ts`, `edge.util.ts`, `SignaturePromptEditorBridge.tsx` — already read `data.outputs/inputs` directly or operate on a different surface.

drewdrewthis:
650ed61dbe805f296d40eaa90f8e79a3be2ef8b2

drewdrewthis:
92ebe2fc8c70b0d759c58dd9abad3b49bf0b148e

Re-verified after review refactor:
- 21/21 AC-bound tests pass
- 1100/1100 tests pass across 94 files in optimization_studio/, server/workflows/, server/scenarios/, components/agents/, hooks/
- Parity 8/8 scenarios bound
- Typecheck clean

drewdrewthis:
## /review — grinder findings at `92ebe2fc8c70b0d759c58dd9abad3b49bf0b148e`

All findings are **non-blocking** (design/test-quality). Security finding #11 is pre-existing, not a regression. No code changes required for review-ready.

| # | Finding | Severity | Location | Notes |
|---|---------|----------|----------|-------|
| 1 | `getEndNodeOutputs` duplicated in drawer + auto-compute | design | `AgentWorkflowEditorDrawer.tsx:76`, `auto-compute-agent-mappings.ts:40` | Extract to `nodeUtils.ts`. Follow-up. |
| 2 | Evaluator `optional` policy baked into `getMappingSurfaceInputs` | design | `nodeUtils.ts:getMappingSurfaceInputs` | Two responsibilities. Split in follow-up. |
| 3 | `node.type === "entry" \|\| node.id === "entry"` magic string at 9+ sites | design | Multiple files | Extract `isEntryNode()` / `isEndNode()` — file a separate issue. |
| 4 | `hasNonEvaluatorTarget` negated immediately; use `.every()` instead | hygiene | `nodeUtils.ts:109` | Minor readability. |
| 5 | `getInputsOutputs` lacks JSDoc guidance directing mapping-surface callers to `getMappingSurfaceInputs` | hygiene | `nodeUtils.ts` | Add `@deprecated` / usage note. |
| 6 | Double type cast on `endNodeData.inputs` in auto-compute | hygiene | `auto-compute-agent-mappings.ts:44-47` | Extract `isEndNodeData` type guard. |
| 7 | `getMappingSurfaceInputs` legacy fallback path untested | test-coverage | `nodeUtils.unit.test.ts` | Add a test for the edge-derived fallback. |
| 8 | `not.toHaveProperty("optional")` is structurally brittle | test-quality | `nodeUtils.unit.test.ts` | Use `toBeFalsy()` or exhaustive shape assertion. |
| 9 | `AgentWorkflowEditorDrawer.unit.test.tsx` is integration-level | test-classification | Test file naming | Rename or extract pure unit test. |
| 10 | Mapping value not asserted in auto-compute unwired scenario | test-coverage | `auto-compute-agent-mappings.unit.test.ts` | Assert the actual mapping value, not just presence. |
| 11 | `sourceHandle.split(".")[1]` propagates as unvalidated key into `scenarioMappings` | pre-existing security debt | `nodeUtils.ts` (fallback path) | Pre-existing in `getInputsOutputs`; not a regression from this PR. File a hardening issue for server-side key validation against `/^[a-zA-Z0-9_]+$/`. |

**Verdict: 0 blocking, 11 non-blocking. PR is review-ready.**

drewdrewthis:

CI: green (zero failing, zero pending — langwatch-app-complete: SUCCESS)
ACs: 21/21 met — prove-it-clean at `92ebe2fc8c70b0d759c58dd9abad3b49bf0b148e`
Review: 11 non-blocking findings — zero blockers
Threads: zero unresolved

Verified by:
  `gh pr view 3980 --json statusCheckRollup --jq '[...] | unique'` → `["SKIPPED","SUCCESS"]` — zero FAILURE entries
  prove-it-clean token: present at HEAD in PR comment
  review findings: https://github.com/langwatch/langwatch/pull/3980#issuecomment-4526835249

PR #3984, opened 2026-05-12
fix: guard Stripe meter event value limit

nasimbahar:
## Summary
- Reject Stripe meter event payload values above the 15-digit limit before making the API call.
- Reuse the existing permanent rejection result shape so billing checkpoints are not advanced on invalid usage values.
- Add coverage for both direct delta reporting and set-to-delta reporting.

## Tests
- `NODE_ENV=test PINO_LOG_LEVEL=warn corepack pnpm --dir langwatch exec vitest run ee/billing/__tests__/usageReportingService.unit.test.ts`

### DIFF
--- a/langwatch/ee/billing/__tests__/usageReportingService.unit.test.ts
+++ b/langwatch/ee/billing/__tests__/usageReportingService.unit.test.ts
@@ -95,6 +95,26 @@ describe("usageReportingService", () => {
       });
     });
 
+    describe("when value exceeds Stripe's 15 digit limit", () => {
+      it("returns reported: false without calling Stripe", async () => {
+        const results = await service.reportUsageDelta({
+          stripeCustomerId: "cus_abc123",
+          organizationId: "org_1",
+          events: [makeEvent({ value: 1_000_000_000_000_000 })],
+        });
+
+        expect(results).toEqual([
+          {
+            identifier: "org_1:langwatch_billable_events:2026-02-19:d100",
+            reported: false,
+            valueSent: 0,
+            error: "Stripe meter event values must not exceed 15 digits",
+          },
+        ]);
+        expect(stripe.billing.meterEvents.create).not.toHaveBeenCalled();
+      });
+    });
+
     describe("when batch has multiple events", () => {
       it("processes sequentially and returns ordered results", async () => {
         const callOrder: number[] = [];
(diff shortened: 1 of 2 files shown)


0xdeafcafe:
Hey @nasimbahar, thanks for your contribution. Before we proceed I want to just check where the need for this change came from, do you have any example of this many events being recorded? It seems very unlikely/impossible for this case to happen?

nasimbahar:
Thanks for checking. This came from Stripe’s changelog: they now reject meter event values with more than 15 significant digits. I don’t have a production example from this repo, so this is more of a defensive guard before calling Stripe than a fix for an observed issue. If you think this is too unlikely for LangWatch, I’m happy to close it.

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** Billing correctness — rejecting an oversized meter value returns `reported: false` without advancing the checkpoint. If the caller retries the same batch, one poison value could permanently stall usage reporting for that org; if it silently skips, that usage is never billed.

- `[Must Check]` `usageReportingService.ts` guard return path — trace how `reported: false` / `valueSent: 0` is consumed upstream: does the checkpoint advance past this event (usage lost) or stall/retry forever (poison pill)? The PR claims "permanent rejection shape"; confirm that shape truly skips rather than blocks.
- `[Probably Fine]` `STRIPE_METER_EVENT_VALUE_MAX = 999_999_999_999_999` — confirm 15-digit boundary matches Stripe's documented limit and the comparison is `>` not `>=`.
- `[Probably Fine]` tests cover both `reportUsageDelta` and `reportUsageSet` — verify they assert Stripe is NOT called, not just the return shape.

- When a value is rejected, does the billing checkpoint advance (dropping that usage silently) or hold (risking a permanent stall on the next poll)? Which is intended?

langwatch-agent (commented):
No security, billing-integrity, or ownership-boundary findings from static external-PR review. The guard is centralized before the Stripe call and covers both delta and set reporting paths with unit coverage. I did not execute PR code. Residual risk: PR is still review-gated; trusted CI should remain green before merge.

langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref the linked issue

Scariest risk: this guards the Stripe money path (rejects meter event values above the 15-digit limit before the API call, reusing the permanent-rejection shape so billing checkpoints do not advance on invalid usage). A wrong guard either lets an over-limit value hit Stripe (API error / lost usage) or wrongly rejects valid usage (under-billing).

Attention targets:
- **Must Check** the 15-digit threshold is correct and applied to the exact value sent to Stripe (post-aggregation quantity), not an intermediate.
- **Must Check** reusing the permanent-rejection path does not silently advance the billing checkpoint, so rejected usage stays visible/retryable rather than dropped.
- **Ask Author** what happens to a rejected usage record downstream (dropped, dead-lettered, alerted)?

Shape note: external contributor, 2 files; confirm the rejection-shape reuse matches the existing contract. Attention map only, not an approval.

0xdeafcafe:
hi @nasimbahar, yeah i'm going to close this. if we run into this i assume it's a bug that should be handled elsewhere as this is not realistically possible to hit otherwise.

but thanks for the time looking into this.

Issue #1095, opened 2026-01-14
OPSCEN-25: Replace Local-Scenarios Set ID


drewdrewthis:
## Summary

Replace the hardcoded `"local-scenarios"` set ID used for on-platform scenario runs with a more robust internal namespace pattern that is unlikely to collide with user-created set names.

## Current State

Currently, on-platform scenarios use `"local-scenarios"` as the default set ID:
- `src/server/scenarios/simulation-runner.service.ts:15` - `DEFAULT_SIMULATION_SET_ID = "local-scenarios"`
- `src/server/api/routers/scenarios/simulation-runner.router.ts:51` - `setId = "local-scenarios"`

This naming is problematic because:
1. Users could accidentally create a set with the same name via external SDK/API usage
2. The name doesn't clearly indicate it's an internal/system set
3. It doesn't distinguish between different projects

### 1. Rename to Internal Namespace Pattern

Replace `"local-scenarios"` with a pattern like:
```
__internal__${projectId}__on-platform-scenarios
```

Example: `__internal__proj_abc123__on-platform-scenarios`

The `__internal__` prefix and double underscore separators make it extremely unlikely that a user would choose this pattern for their own set names.

### 2. Special UI Treatment on Simulation Sets Page

On the `/[project]/simulations` page (simulation sets list), add special UI treatment for internal sets:

**Options to consider:**
- Display a different icon (e.g., a settings/system icon instead of 🎭)
- Show a badge like "On-Platform" or "Built-in"
- Display a friendlier name like "On-Platform Scenarios" instead of the raw internal ID
- Pin it to the top of the list
- Use a distinct card background color or border

The `SetCard` component (`src/components/simulations/SetCard.tsx`) would need to detect internal sets and render accordingly.

### 3. Utility Function for Detection

Add a utility function to check if a set ID is internal:
```typescript
const INTERNAL_SET_PREFIX = "__internal__";
const ON_PLATFORM_SET_SUFFIX = "__on-platform-scenarios";

export function isInternalSetId(setId: string): boolean {
  return setId.startsWith(INTERNAL_SET_PREFIX);
}

export function isOnPlatformSet(setId: string): boolean {
  return setId.endsWith(ON_PLATFORM_SET_SUFFIX);
}

export function getOnPlatformSetId(projectId: string): string {
  return `${INTERNAL_SET_PREFIX}${projectId}${ON_PLATFORM_SET_SUFFIX}`;
}
```

## Files to Modify

[...]

PR #3987, opened 2026-05-12
fix(deps): override kysely to 0.28.17 — JSON-path traversal injection

sergioestebance:
## Summary

- Adds pnpm override `"kysely@>=0.26.0 <0.28.17": "0.28.17"` to force the patched version
- kysely is a transitive dependency via `@better-auth/core` and `@better-auth/kysely-adapter`
- The override is scoped to only affect vulnerable versions (0.26.0–0.28.16)

## Dependabot Alerts Addressed (1 high-severity)

**JSON-path traversal injection via unsanitized path-leg metacharacters in `JSONPathBuilder.key()` / `.at()`:**
- #1095 (langwatch/pnpm-lock.yaml)

## Test plan

- [ ] Verify CI passes (type checks, unit tests)
- [ ] Confirm Dependabot alert #1095 auto-closes after merge
- [ ] Smoke test better-auth flows (login, session management)

### DIFF
--- a/langwatch/package.json
+++ b/langwatch/package.json
@@ -343,7 +343,8 @@
       "react-is": "^19.2.5",
       "pino": "^10.3.1",
       "pino-pretty": "^13.1.3",
-      "thread-stream": "^4.0.0"
+      "thread-stream": "^4.0.0",
+      "kysely@>=0.26.0 <0.28.17": "0.28.17"
     }
   },
   "ct3aMetadata": {
--- a/langwatch/package.json
+++ b/langwatch/package.json
@@ -344,7 +344,7 @@
       "pino": "^10.3.1",
       "pino-pretty": "^13.1.3",
       "thread-stream": "^4.0.0",
-      "kysely@>=0.26.0 <0.28.17": "0.28.17"
+      "kysely": "0.28.17"
     }
   },
   "ct3aMetadata": {
--- a/langwatch/pnpm-lock.yaml
+++ b/langwatch/pnpm-lock.yaml
@@ -33,6 +33,7 @@ overrides:
   pino: ^10.3.1
   pino-pretty: ^13.1.3
   thread-stream: ^4.0.0
+  kysely@>=0.26.0 <0.28.17: 0.28.17
 
 packageExtensionsChecksum: sha256-LBc1N1GezKyerwGBwpMHZPFoMYZTHjpxVKWJilBVvTU=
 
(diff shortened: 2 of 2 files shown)


rogeriochaves (approved):

langwatch/package.json
```diff
@@ -343,7 +343,8 @@
       "react-is": "^19.2.5",
       "pino": "^10.3.1",
       "pino-pretty": "^13.1.3",
-      "thread-stream": "^4.0.0"
+      "thread-stream": "^4.0.0",
+      "kysely@>=0.26.0 <0.28.17": "0.28.17"
```
rogeriochaves: wut? is this a valid dependency spec?

rogeriochaves (changes requested):

sergioestebance:
## Safety Review

| Check | Result |
|-------|--------|
| **Scope** | 1 pnpm override added to `langwatch/package.json` + lock file update |
| **Production?** | Technically present but **LangWatch does not use kysely at all**. The codebase uses the Prisma adapter with better-auth (`prismaAdapter` in `src/server/better-auth/index.ts:2`). Kysely is only an optional peer dep of `@better-auth/core` and `@better-auth/kysely-adapter` — it gets installed but is never imported. |
| **Override format** | `"kysely@>=0.26.0 <0.28.17": "0.28.17"` — **valid pnpm syntax**. Matches the exact same `package@range: version` pattern already used in this file for `"axios@>=1.0.0 <1.16.0": "1.16.0"` and `"vite@>=8.0.0 <8.0.5": "8.0.10"`. The lock file resolved it correctly (`kysely@0.28.17` everywhere). |
| **Pinned versions?** | No — kysely is not even a direct dependency |
| **Breaking risk** | Extremely low — 0.28.16 → 0.28.17 is a patch bump, and the code never imports kysely |
| **Version compatibility** | `@better-auth/core` declares `kysely: ^0.28.5` and `@better-auth/kysely-adapter` declares `kysely: ^0.28.14` — both are satisfied by 0.28.17 (within `>=0.28.x <0.29.0`) |
| **CI** | All green (build, e2e, type checks) |

**Verdict: Safe to merge.** The override format is valid and consistent with existing overrides. The dependency is unused by LangWatch directly — it's an optional peer dep of better-auth sub-packages. Zero runtime risk.

langwatch/package.json
sergioestebance: Fixed — replaced the weird version-range-in-key format with the standard simple override:

```json
"kysely": "0.28.17"
```

Confirmed with `pnpm why kysely` that 0.28.17 resolves everywhere. Dependabot alert #1095 targets `kysely` in `langwatch/pnpm-lock.yaml` with vulnerable range `>= 0.26.0, < 0.28.17` and first patched `0.28.17` — the lock file now has exactly `0.28.17` with zero instances of `0.28.16`.

rogeriochaves (approved):

PR #3994, opened 2026-05-12
fix(auth): enable accountLinking so orphan email-verified Users can sign in via SSO

rogeriochaves:
## Summary

- A `User` row with `emailVerified=true` but zero `Account` rows (pre-seeded invite, half-finished legacy signup, or migration leftover) silently blocked every subsequent OAuth sign-in for that email — surfaced in the UI as *"registered with another authentication method"*. On SSO-enforced orgs the user was permanently locked out after a successful IdP round-trip.
- Root cause: BetterAuth's default refuses to attach a new OAuth `Account` to an existing `User` unless `account.accountLinking` is enabled. The setting was missing.
- Fix: enable `account.accountLinking.enabled` in the BetterAuth config. Linking still requires `emailVerified=true` on the existing User and matching emails (`allowDifferentEmails` defaults to `false`). SSO-domain enforcement in `beforeAccountCreate` (`langwatch/src/server/better-auth/hooks.ts`) still rejects the wrong provider before any link.

## Repro

Surfaced by a customer report. Verified against prod:

- Orphan User row existed with `emailVerified=true`, `lastLoginAt=null`, zero `Account` rows.
- Org had SSO enforced (`ssoDomain` + `ssoProvider` set).
- Every OAuth callback returned `account_already_linked_to_different_user` → normalized to `OAuthAccountNotLinked` in `signin.tsx:31`.

Orphan row was cleaned up manually so the affected user can sign in again; this PR closes the gap for everyone else.

## Test plan

- [x] New BDD spec: `specs/auth/sso-orphan-user-linking.feature` covers happy path, unverified-email refusal, SSO wrong-provider rejection.
- [x] Regression test added in `langwatch/src/server/better-auth/__tests__/index.test.ts` locks `account.accountLinking.enabled=true` in.
- [x] `pnpm test:unit src/server/better-auth/__tests__/` — 93/93 passing.
- [ ] CI green.

### DIFF
--- a/specs/auth/sso-orphan-user-linking.feature
+++ b/specs/auth/sso-orphan-user-linking.feature
@@ -1,0 +1,25 @@
+Feature: SSO sign-in links to orphan email-verified User rows
+
+  Background:
+    Given an organization with SSO enforced for domain "example.com"
+    And the organization's SSO provider matches the OAuth provider configured for the deployment
+
+  Scenario: Orphan User row from a prior invite is auto-linked on first SSO sign-in
+    Given a User row exists for "alice@example.com" with emailVerified=true and zero linked Account rows
+    When Alice completes the SSO sign-in callback with the configured OAuth provider
+    Then a new Account row is created for that User
+    And Alice is signed in
+    And no "registered with another authentication method" error is shown
+
+  Scenario: Unverified orphan User cannot be hijacked via OAuth
+    Given a User row exists for "bob@example.com" with emailVerified=false and zero linked Account rows
+    When an OAuth sign-in callback returns email "bob@example.com"
+    Then linking is refused
+    And the OAuth callback redirects to the auth error page
+
+  Scenario: SSO-domain guard still blocks the wrong provider
+    Given a User row exists for "carol@example.com" with emailVerified=true and zero linked Account rows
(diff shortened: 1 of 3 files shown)


specs/auth/sso-orphan-user-linking.feature
```diff
@@ -0,0 +1,25 @@
...
+    When Alice completes the SSO sign-in callback with the configured OAuth provider
+    Then a new Account row is created for that User
+    And Alice is signed in
+    And no "registered with another authentication method" error is shown
+
+  Scenario: Unverified orphan User cannot be hijacked via OAuth
+    Given a User row exists for "bob@example.com" with emailVerified=false and zero linked Account rows
+    When an OAuth sign-in callback returns email "bob@example.com"
+    Then linking is refused
+    And the OAuth callback redirects to the auth error page
```
0xdeafcafe, line 18: should we have an explicit test for this?

0xdeafcafe (approved):

specs/auth/sso-orphan-user-linking.feature
rogeriochaves: yeahhh idk, wouldn't that be testing betterauth which we should just trust and love?

PR #3976, opened 2026-05-11
feat: online-evaluator loop prevention + tenant safety net (post-2026-05-11)

rogeriochaves:
## Summary

Comprehensive fix + defense-in-depth for the **2026-05-11 event-sourcing outage** (online-evaluator infinite loop, ~95% cluster starvation by one tenant for ~2h). Four independent tracks shipped together because they all stem from the same incident and benefit from joint review.

### Track A — Loop prevention (the actual fix)
- nlpgo `BaggageAttributeProcessor` stamps `langwatch.causality_depth` on **every** span at OnStart from a single baggage entry. Primary defense.
- Handlers extract `X-LangWatch-Causality-Depth` header + W3C `traceparent`, set depth+1 on baggage, prefer extracted SpanContext for trace continuation.
- Outbound evaluator-block HTTP injects the propagator + depth header so the chain survives across services.
- TS reactor: per-span depth check on inbound event + `causalSubtreeSpans` parent-walk in fold (secondary defense, catches out-of-order span arrival).
- Origin no longer hardcoded — it remains a user-configurable precondition.
- Kill-switch: \`LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD=1\`.

### Track B — Per-tenant rate anomaly detector + Ops panel
- \`TenantRateTracker\` records per-tenant enqueue counts to minute-bucketed Redis HASH (O(1) write on hot path, swallowed on error).
- \`AnomalyDetector\` worker (60s tick) computes p95 baseline over last 7 days. Two tiers:
  - Surface (10× baseline, 5-min sustained) → Ops panel + warn log.
  - Hard (100× baseline, 15-min sustained) → red badge + paged log + plumbed auto-pause hook (currently no-op).
- New \`AnomaliesCard\` at top of \`/ops/queues\` with per-row Dismiss.
- tRPC: \`ops.listAnomalies\` / \`ops.dismissAnomaly\`.

### Track C — Structural fingerprint loop detection
- \`computeStructuralFingerprint\`: sha1 over sorted(\`name:kind:sorted_attribute_keys\`). **Keys only** — never values, so LLM non-determinism doesn't affect the hash.
- Recorded per-span in \`recordSpanCommand\` (fire-and-forget).
- Trigger: one fingerprint >= 100/min AND >= 80% of tenant volume, sustained 5min.
[...]

PR #4035, opened 2026-05-13
test(nlpgo): real e2e proof of traceparent + causality_depth propagation

rogeriochaves:
## Summary

Post-merge dogfood of #3976 raised the right question: did we ever spin up nlpgo and run a proper integration test that proves spans actually end up as children of the parent trace and the full traceparent wiring works?

Answer was no. The loop-prevention PR shipped with thorough function-level unit tests of every individual piece, but nothing wired the whole chain together and observed it end-to-end. This PR adds that.

## What this proves

Four scenarios in \`services/nlpgo/tests/integration/causality_propagation_test.go\`, all running the production wiring (\`NewBaggageAttributeProcessor\` + \`TraceContext+Baggage\` composite propagator from \`pkg/otelsetup\`):

1. **\`TestCausalityPropagation_RootSpanContinuesParentTrace\`** — POSTs to \`/go/studio/execute_sync\` with a \`traceparent\` header, asserts the root studio span's TraceID equals the inbound TraceID AND its parent SpanID equals the inbound SpanID.

2. **\`TestCausalityPropagation_AllSpansShareTraceIDAndCarryDepth\`** — Walks every span captured during the workflow run (across engine goroutines), asserts they all share the inbound TraceID AND carry \`langwatch.reserved.causality_depth = "1"\` (inbound 0 + 1).

3. **\`TestCausalityPropagation_OutboundHTTPInjectsTraceparentAndDepth\`** — Captures the evaluator-block's outbound HTTP call to the (fake) LangWatch app, asserts it carries \`traceparent\` + \`baggage\` + \`X-LangWatch-Causality-Depth\` headers with the same trace_id we started with.

4. **\`TestCausalityPropagation_SanityCheckParentIsRemoteFlag\`** — Cheap sanity pin so a future \`tracetest\` refactor can't silently invalidate the heuristic the larger tests use to identify the root span.

## Why function-level wasn't enough

The unit tests in \`adapters/httpapi/causality_test.go\` and \`app/engine/tracing_test.go\` each prove one piece in isolation, but a regression in *any* of: \`applyInboundCausality\`, \`startStudioSpan\`, \`BaggageAttributeProcessor\` registration, context propagation across engine goroutines, or the evaluatorblock's outbound propagator inject would have escaped them. The e2e test exercises them all in series so a break anywhere surfaces.

## Test plan

- [x] \`go test ./services/nlpgo/tests/integration/ -run TestCausalityPropagation -count=1 -v\` → 4/4 PASS
- [x] Full nlpgo integration suite still green (no regression)
- [x] \`pnpm check:feature-parity\` clean (1028 scenarios bound)

### DIFF
--- a/services/nlpgo/tests/integration/causality_propagation_test.go
+++ b/services/nlpgo/tests/integration/causality_propagation_test.go
@@ -1,0 +1,372 @@
+// End-to-end integration test for W3C trace-context + causality-depth
+// propagation through nlpgo.
+//
+// Why this file exists: the original loop-prevention PR (#3976) shipped
+// with function-level unit coverage of every individual piece —
+// applyInboundCausality extracts headers, BaggageAttributeProcessor
+// stamps spans, evaluatorblock injects the propagator on outbound —
+// but nothing actually wired the whole chain together and observed it.
+// rchaves caught that on post-merge dogfood: "did we ever prove the
+// spans end up as a child of the parent and the full traceparent is
+// working?"
+//
+// What this test proves end-to-end:
+//   1. POSTing /go/studio/execute_sync with `traceparent` makes the
+//      emitted spans share the inbound trace_id (trace continuity).
+//   2. The studio root span's PARENT span_id equals the inbound
+//      traceparent's span_id (parent linkage, not a fresh root).
+//   3. EVERY span emitted during the run — root, children across
+//      goroutines — carries `langwatch.reserved.causality_depth`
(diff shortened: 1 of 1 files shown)


rogeriochaves:
Closing — misnamed. This was an in-process integration test of nlpgo's HTTP boundary, not e2e. The real e2e (real langwatch event-sourcing pipeline + testcontainers Redis/CH + real nlpgo subprocess + OTLP roundtrip + reactor block proof) is being built in a fresh PR.

PR #4038, opened 2026-05-13
fix(traces): fall back to org membership for trace visibility

sergioestebance:
## Summary

- `getUserProtectionsForProject` only checked team-level `RoleBinding` rows when deciding whether to show or redact trace inputs/outputs
- When a project is created via API with `newTeamName` and no `userId` (service API key), the auto-created team has **zero members**
- Org admins could navigate to the project (RBAC grants access via org scope) but saw all inputs/outputs as `[REDACTED]` because the visibility check only looked at team scope
- Now falls back to `OrganizationUser` membership when no team binding exists: org members count as team members, org admins as team admins
- EXTERNAL (Lite Member) org users are explicitly excluded via allowlist (only ADMIN and MEMBER grant access), so any future role variants fail closed
- Service API key bindings now display the key's name instead of "Unknown" in the teams settings page

### How it works

```
Team RoleBinding found?
  → yes → use it (unchanged)
  → no  → single-query org lookup via getUserOrgRoleByTeamId
           → org ADMIN    → treat as team admin
           → org MEMBER   → treat as team member
           → org EXTERNAL → denied (allowlist, fails closed)
           → any other    → denied
           → not in org   → denied
```

The fallback query only fires when there's no team-level binding, so existing flows are unaffected.

### Root cause

The rest of the RBAC system (`resolveProjectPermission` in `rbac.ts`) checks 3 scopes: PROJECT → TEAM → ORGANIZATION. The visibility logic was never aligned — it only checked TEAM scope. This was true in the original `TeamUser` implementation and carried over when #3424 migrated it to `RoleBinding`.

### Additional fix: "Unknown" in teams settings

Service API key RoleBindings have `userId=null` and `groupId=null`. The teams settings page (`team.service.ts`) resolved binding names via `user?.name ?? group?.name`, falling through to "Unknown". Now includes `apiKey` in the RoleBinding query and uses `apiKey?.name` as a fallback.

## Test plan

- [x] 11 unit tests (5 org-fallback scenarios including EXTERNAL guard + 6 existing adjusted)
- [x] 47 tests pass across all affected files (protections, org service, billing)
- [x] Typecheck clean
- [x] Adversarial review: EXTERNAL escalation caught and fixed
- [x] E2E verified: provision → send traces → visible inputs/outputs on local

### DIFF
--- a/langwatch/src/server/api/utils.ts
+++ b/langwatch/src/server/api/utils.ts
@@ -8,6 +8,7 @@ import {
 import type { Session } from "~/server/auth";
 import type { Protections } from "../elasticsearch/protections";
 import { hasProjectPermission, isDemoProject } from "./rbac";
+import { getApp } from "~/server/app-layer/app";
 
 export const extractCheckKeys = (
   inputObject: Record<string, any>,
@@ -134,23 +135,20 @@ export async function getUserProtectionsForProject(
   let isMemberOfTeam = teamBindings.length > 0;
 
   if (!isMemberOfTeam) {
-    const team = await ctx.prisma.team.findUnique({
-      where: { id: project.teamId },
-      select: { organizationId: true },
-    });
+    const orgService = getApp().organizations;
+    const organizationId = await orgService.getOrganizationIdByTeamId(
+      project.teamId,
+    );
 
-    if (team) {
-      const orgUser = await ctx.prisma.organizationUser.findFirst({
-        where: {
-          userId: ctx.session.user.id,
-          organizationId: team.organizationId,
-        },
-        select: { role: true },
+    if (organizationId) {
+      const orgRole = await orgService.getUserOrgRole({
+        userId: ctx.session.user.id,
+        organizationId,
       });
 
-      if (orgUser && orgUser.role !== OrganizationUserRole.EXTERNAL) {
+      if (orgRole && orgRole !== OrganizationUserRole.EXTERNAL) {
         isMemberOfTeam = true;
-        if (orgUser.role === OrganizationUserRole.ADMIN) {
+        if (orgRole === OrganizationUserRole.ADMIN) {
           isAdminForTeam = true;
         }
       }
--- a/langwatch/src/server/api/utils.ts
+++ b/langwatch/src/server/api/utils.ts
@@ -115,8 +115,8 @@ export async function getUserProtectionsForProject(
     };
   }
 
-  // For signed in users, check their team permissions via RoleBinding
-  const bindings = await ctx.prisma.roleBinding.findMany({
+  // Check team-level role bindings
+  const teamBindings = await ctx.prisma.roleBinding.findMany({
     where: {
       userId: ctx.session.user.id,
       scopeType: RoleBindingScopeType.TEAM,
@@ -127,10 +127,34 @@ export async function getUserProtectionsForProject(
     },
   });
 
-  const isAdminForTeam = bindings.some(
+  let isAdminForTeam = teamBindings.some(
     (binding) => binding.role === TeamUserRole.ADMIN,
   );
-  const isMemberOfTeam = bindings.length > 0;
+  let isMemberOfTeam = teamBindings.length > 0;
+
+  if (!isMemberOfTeam) {
+    const team = await ctx.prisma.team.findUnique({
+      where: { id: project.teamId },
+      select: { organizationId: true },
+    });
+
+    if (team) {
+      const orgUser = await ctx.prisma.organizationUser.findFirst({
+        where: {
+          userId: ctx.session.user.id,
+          organizationId: team.organizationId,
+        },
+        select: { role: true },
+      });
+
+      if (orgUser) {
+        isMemberOfTeam = true;
+        if (orgUser.role === "ADMIN") {
+          isAdminForTeam = true;
+        }
+      }
+    }
+  }
 
   const obtainVisibilityLevel = (
     visibility: ProjectSensitiveDataVisibilityLevel,
--- a/langwatch/src/server/api/utils.ts
+++ b/langwatch/src/server/api/utils.ts
@@ -135,23 +135,16 @@ export async function getUserProtectionsForProject(
   let isMemberOfTeam = teamBindings.length > 0;
 
   if (!isMemberOfTeam) {
-    const orgService = getApp().organizations;
-    const organizationId = await orgService.getOrganizationIdByTeamId(
-      project.teamId,
-    );
+    const orgRole = await getApp().organizations.getUserOrgRoleByTeamId({
+      userId: ctx.session.user.id,
+      teamId: project.teamId,
+    });
 
-    if (organizationId) {
-      const orgRole = await orgService.getUserOrgRole({
-        userId: ctx.session.user.id,
-        organizationId,
-      });
-
-      if (orgRole && orgRole !== OrganizationUserRole.EXTERNAL) {
-        isMemberOfTeam = true;
-        if (orgRole === OrganizationUserRole.ADMIN) {
-          isAdminForTeam = true;
-        }
-      }
+    if (orgRole === OrganizationUserRole.ADMIN) {
+      isMemberOfTeam = true;
+      isAdminForTeam = true;
+    } else if (orgRole === OrganizationUserRole.MEMBER) {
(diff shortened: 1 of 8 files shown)


langwatch/src/server/teams/team.service.ts
```diff
@@ -176,7 +178,7 @@ export class TeamService {
             groupId: null as string | null,
             viaGroupId: null as string | null,
             viaGroupName: null as string | null,
-            name: b.user?.name ?? b.user?.email ?? "Unknown",
+            name: b.user?.name ?? b.user?.email ?? b.apiKey?.name ?? "Unknown",
```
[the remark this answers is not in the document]
sergioestebance: Valid observation — `directUserBindings` filters for `userId`, so `apiKeyId` is mutually exclusive today. Keeping the fallback intentionally as a defensive guard. If the binding query ever changes shape (e.g., filtering relaxed, new principal type added), we don't want "Unknown" silently returning. The cost of the dead branch is zero; the cost of a future regression showing "Unknown" to users is a bug report.

langwatch/src/server/app-layer/organizations/repositories/organization.prisma.repository.ts
```diff
@@ -50,6 +50,20 @@ export class PrismaOrganizationRepository implements OrganizationRepository {
...
   }
 
+  async getUserOrgRole({
+    userId,
+    organizationId,
+  }: {
+    userId: string;
+    organizationId: string;
+  }): Promise<OrganizationUserRole | null> {
+    const orgUser = await this.prisma.organizationUser.findFirst({
```
drewdrewthis, line 74: `OrganizationUser` has `@@id([userId, organizationId])` (schema.prisma:159), so this is exactly the case for `findUnique({ where: { userId_organizationId: { userId, organizationId } } })`. Two existing call sites in `organization.ts` (lines 1130 and 1236) already do it that way. Switching to `findUnique` makes the unique-by-composite-PK contract explicit (the compiler enforces both keys) and uses the PK index directly.

langwatch/src/server/api/utils.ts
```diff
@@ -127,10 +129,31 @@ export async function getUserProtectionsForProject(
...
   });
 
-  const isAdminForTeam = bindings.some(
+  let isAdminForTeam = teamBindings.some(
     (binding) => binding.role === TeamUserRole.ADMIN,
   );
-  const isMemberOfTeam = bindings.length > 0;
+  let isMemberOfTeam = teamBindings.length > 0;
+
+  if (!isMemberOfTeam) {
```
drewdrewthis, line 137: **Optional collapse to single query**: this fallback fires two sequential lookups — `getOrganizationIdByTeamId(teamId)` then `getUserOrgRole({ userId, organizationId })`. Both go through Prisma so they're not super-expensive, but they're serialized (`await` then `await`).

You can do it in one query via the team relation:

```ts
const orgUser = await ctx.prisma.organizationUser.findFirst({
  where: {
    userId: ctx.session.user.id,
    organization: { teams: { some: { id: project.teamId } } },
  },
  select: { role: true },
});
```

Would mean adding a `getUserOrgRoleByTeamId` method on the repo (one purpose-built call) instead of two general ones. Not a blocker — current shape is correct.

langwatch/src/server/api/utils.ts
```diff
@@ -127,10 +129,31 @@ export async function getUserProtectionsForProject(
...
+    );
+
+    if (organizationId) {
+      const orgRole = await orgService.getUserOrgRole({
+        userId: ctx.session.user.id,
+        organizationId,
+      });
+
+      if (orgRole && orgRole !== OrganizationUserRole.EXTERNAL) {
+        isMemberOfTeam = true;
```
drewdrewthis: Defense-in-depth nit: the EXTERNAL exclusion is correct, but it's a denylist ("everything except EXTERNAL grants membership"). If a new `OrganizationUserRole` variant is added later (e.g. PENDING, INVITED), it would silently be treated as MEMBER-equivalent. An allowlist would fail closed:

```ts
if (orgRole === OrganizationUserRole.ADMIN) {
  isMemberOfTeam = true;
  isAdminForTeam = true;
} else if (orgRole === OrganizationUserRole.MEMBER) {
  isMemberOfTeam = true;
}
```

Given this is the trace-visibility path and you specifically called out adversarial review, leaning fail-closed seems on-brand.

langwatch/src/server/teams/team.service.ts
drewdrewthis, line 181: This `apiKey` include + name-fallback change is from commit `d486193` and isn't mentioned in the PR body — it's a real fix (service-API-key bindings displaying as "Unknown") but unrelated to the trace-visibility org fallback. Either:
- Add a line to the PR description so it doesn't look like drive-by scope creep during review, or
- Split it out (`fix(teams): show API key name`) — clean stand-alone fix, easier to bisect/revert.

Not a blocker, just hygiene.

langwatch/src/server/app-layer/organizations/organization.service.ts
```diff
@@ -111,6 +111,13 @@ export class OrganizationService {
     return this.repo.getOrganizationIdByTeamId(teamId);
   }
 
+  async getUserOrgRole(params: {
+    userId: string;
+    organizationId: string;
+  }): Promise<OrganizationUserRole | null> {
```
drewdrewthis, line 117: Layering looks good — service is a thin pass-through to the repo, matching the existing `getOrganizationIdByTeamId` shape directly above. Consider documenting the EXTERNAL semantics for callers (e.g. "Returns the user's role in the org including EXTERNAL; callers responsible for treating EXTERNAL as denied where appropriate") so the next caller doesn't repeat the rbac.ts L879 mistake.

drewdrewthis (commented):
Reviewed with focus on service/repo layering. The refactor that introduced `OrganizationService.getUserOrgRole` got the layering exactly right — repo → service → caller. EXTERNAL guard is correctly placed. A few notes inline:

1. **`findFirst` vs `findUnique`** for OrganizationUser — there's a composite `@@id([userId, organizationId])`, so a unique lookup is available and is what the rest of this codebase uses (`organization.ts:1130, 1236`).
2. **N+1 simplification**: the fallback fires two sequential queries (getOrgId, getRole). A single query with the right where clause would collapse them.
3. **Unmentioned `apiKey` include change** in `team.service.ts` — the diff includes commit `d486193` which is a separate fix worth a line in the PR body so future archaeologists understand why it's bundled here.

No blockers — the visibility fix and EXTERNAL guard are the important parts and they're correct.

langwatch/src/server/app-layer/organizations/repositories/organization.prisma.repository.ts
sergioestebance: Done — switched to `findUnique` with `userId_organizationId` composite key, matching the existing pattern in `organization.ts`.

langwatch/src/server/api/utils.ts
sergioestebance: Done — added `getUserOrgRoleByTeamId` that resolves org membership via `organization: { teams: { some: { id: teamId } } }` in a single query. The two-call pattern (`getOrganizationIdByTeamId` + `getUserOrgRole`) is replaced.

langwatch/src/server/api/utils.ts
sergioestebance: Done — switched to allowlist. Now only `ADMIN` and `MEMBER` explicitly grant access. Any future role variant (PENDING, INVITED, etc.) fails closed.

langwatch/src/server/teams/team.service.ts
sergioestebance: Done — added the "Unknown" fix to the PR description under "Additional fix" section. Kept it in the same PR since it was discovered during the same investigation (API-created projects with service keys).

langwatch/src/server/app-layer/organizations/organization.service.ts
sergioestebance: Good call. The new `getUserOrgRoleByTeamId` method returns the raw role including EXTERNAL, so callers own the access decision. The only caller right now (`utils.ts`) uses an allowlist, but the next caller will need to know that.

drewdrewthis (approved):

PR #4043, opened 2026-05-13
feat(prompts): add versioned runtime parameters to prompt versions

sergioestebance:
## Summary

- **Adds a `runtimeParameters` column** to `LlmPromptConfigVersion` (Prisma + migration) to store arbitrary versioned JSON (`Record<string, unknown>`) per prompt version
- **Full-stack naming: `parameters`** — the external API (REST, tRPC, SDKs) uses `parameters`, internally stored as `runtimeParameters` in Prisma
- **Editable Parameters tab in playground** — matches the Variables tab pattern: header with "+ Add" button, inline key=value editing, x delete button. Replaces the previous read-only view
- **Lossless typed values** — values set via REST/tRPC/SDK keep their JSON type. The editor's display/parse are exact inverses (type-ambiguous strings like `"123"` are shown quoted), so editing one row never coerces another
- **Explanation text on both tabs** — "Variables are substituted into the prompt template at runtime." / "Parameters are arbitrary configurations returned with the prompt, for use outside the prompt itself (e.g. RAG settings, pipeline config)."
- **Python & TypeScript SDK support** — both SDKs accept `parameters` in create/update/sync flows and return them in prompt responses
- **BDD spec + integration/unit tests** — `specs/prompts/prompt-runtime-parameters.feature` plus integration tests covering CRUD, versioning, tagging, restore, sync, and conflict detection

## Changes

| Area | What changed |
|------|-------------|
| Prisma | `runtimeParameters Json @default("{}")` on `LlmPromptConfigVersion` |
| Migration | `20260602120000_add_runtime_parameters_to_prompt_version` — `ADD COLUMN "runtimeParameters"` |
| Service layer | `prompt.service.ts` — create/update/sync/restore/getAllVersions carry `parameters` (preserve-on-undefined) |
| REST API | Hono routes accept/return `parameters` in input/output schemas |
| tRPC | `prompts.trpc-router.ts` — create/update mutations pass `parameters` |
| UI (editor) | `RuntimeParametersField.tsx` — key-value editor matching VariablesSection pattern, lossless type round-trip, stable row ids |
| UI (playground) | `PromptTabbedSection.tsx` — editable Parameters tab + explanation text on both tabs |
| OpenAPI spec | `openapiLangWatch.json` — prompt request bodies use `parameters` (was `config`) |
| TypeScript SDK | `cli/types.ts`, `prompts-api.service.ts` — `parameters` field |
| Python SDK | `prompt_data.py`, `prompt_api_service.py` — `parameters` field |

## Review feedback addressed

- **String round-trip (blocking)** — replaced content-sniffing coercion with `serializeValue`/`displayValue` that are exact inverses, so API-set string values like `"123"`/`"true"` are no longer silently turned into numbers/booleans when an unrelated row is edited. Added regression tests.
- **Migration ordering (blocking)** — renamed the folder to a timestamp after the latest applied migration and aligned the name to the `runtimeParameters` column.
- **Naming** — renamed the misleading `resolvedParametersData` → `resolvedConfigData` in `syncPrompt` (it holds config data, not parameters; runtime parameters flow separately via `params.parameters`).
- **Dedup** — extracted `sortKeysDeep` to a neutral `repositories/sortKeysDeep.ts` shared by the repository and the version schema; removed the duplicate and dead export.
- **Dead code** — removed the unused `RuntimeParametersReadonly`.
- **Type safety** — dropped an `as any` in `usePromptConfigForm` (narrowed to `Resolver<PromptConfigFormValues>`); replaced a test `as any` with a typed `buildVersionCreateInput` factory helper.
- **Python reads** — `_read_runtime_parameters` prefers a first-class attr and falls back to the generator's `additional_properties`; added a real `from_dict` deserialization test proving the wire path.
- **CLI** — `prompt sync` "use remote" no longer writes an empty `parameters: {}` into local files (matches the YAML converter).

## QA Screenshots

| Screenshot | Description |
|-----------|-------------|
| Parameters tab with values | **Editable Parameters tab** — 3 key-value rows with + Add button, x delete, explanation text |
| Parameters tab empty | **Empty state** — "No parameters defined" with + Add button visible |
| Variables tab explanation | **Variables tab** — explanation text "Variables are substituted into the prompt template at runtime." |

## Test plan

- [x] Integration tests: create, update, fetch-by-tag, fetch-by-version, list-versions, restore, sync, conflict detection (incl. parameters-only conflict when local is behind), nested values
- [x] Unit tests: `parseRuntimeParameters`, `runtimeParametersEqual`, schema validation, value round-trip regression
- [x] UI tests: key-value add/remove/edit, empty state, quoted type-ambiguous strings, explanation text on both tabs
- [x] Python SDK: real `from_dict` deserialization test for `parameters`
- [x] CI green (typecheck, lint, unit, integration, e2e, feature-parity, SDK suites)
- [x] Browser QA: screenshots at `langwatch/pr-screenshots` repo (pr-4043)


langwatch/src/app/api/prompts/[[...route]]/schemas/outputs.ts
```diff
@@ -53,6 +53,7 @@ const apiResponseVersionOutputSchema = z.object({
   promptingTechnique: configDataSchema.shape.prompting_technique,
   responseFormat: configDataSchema.shape.response_format,
   tags: z.array(apiResponsePromptTagSchema).default([]),
+  config: z.record(z.string(), z.unknown()).default({}),
```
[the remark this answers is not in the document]
Comment (sergioestebance) [reply langwatch/src/app/api/prompts/[[...route]]/schemas/outputs.ts]: Fixed — now using `runtimeConfigSchema` (which wraps `jsonValue`) in both `inputs.ts` and `outputs.ts` for consistent validation at the API boundary.

langwatch/src/app/api/prompts/[[...route]]/app.v1.ts
```diff
@@ -781,6 +781,7 @@ app.post(
                     differences: z.array(z.string()),
                     remoteConfigData:
                       getLatestConfigVersionSchema().shape.configData,
+                    remoteConfig: z.record(z.string(), z.unknown()).optional(),
```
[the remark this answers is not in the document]
Comment (sergioestebance) [reply langwatch/src/app/api/prompts/[[...route]]/app.v1.ts]: Fixed in latest commit.

langwatch/src/prompts/forms/fields/__tests__/RuntimeParametersField.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    renderRuntimeParametersField();
+
+    const triggers = screen.getAllByRole("button", { name: /runtime parameters/i });
+    await user.click(triggers[0]!);
+    const editor = screen.getByRole("textbox", {
+      name: /runtime parameters json/i,
+    });
+    fireEvent.change(editor, { target: { value: "{invalid" } });
+
+    expect(screen.getByText("Config must be valid JSON")).toBeInTheDocument();
```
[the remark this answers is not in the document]
sergioestebance: Fixed in latest commit.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
```diff
@@ -210,6 +211,7 @@ export const promptsRouter = createTRPCRouter({
           responseFormat: responseFormatSchema.optional(),
           demonstrations: nodeDatasetSchema.optional(),
           handle: handleSchema,
+          parameters: runtimeParametersSchema.optional(),
```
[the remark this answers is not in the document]
sergioestebance: Fixed — added `parameters: sourcePrompt.parameters` to all three copy/sync/push paths in the tRPC router.

specs/prompts/prompt-runtime-parameters.feature
```diff
@@ -0,0 +1,138 @@
...
+    When I enter invalid JSON in the Runtime Parameters section
+    Then the editor shows a parameters validation error
+    And I cannot save the version
+
+  @integration @unimplemented
+  Scenario: Prompt playground shows runtime parameters as read-only version data
+    Given prompt "search-agent" has latest parameters {"readonly": true}
+    When I view "search-agent" in the prompt playground
+    Then the Config tab displays {"readonly": true}
+
```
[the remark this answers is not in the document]
sergioestebance: Fixed — updated scenario step to assert 'Parameters tab' instead of 'Config tab'.

python-sdk/src/langwatch/prompts/types/prompt_data.py
```diff
@@ -1,46 +1,46 @@
...
-    PutApiPromptsByIdResponse200,
-)
-from langwatch.generated.langwatch_rest_api_client.models.post_api_prompts_response_200 import (
+from langwatch.generated.models.post_api_prompts_response_200 import (
     PostApiPromptsResponse200,
 )
+from langwatch.generated.models.put_api_prompts_by_id_response_200 import (
+    PutApiPromptsByIdResponse200,
+)
+from langwatch.generated.types import Unset
```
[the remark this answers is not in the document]
sergioestebance: This import path (`langwatch.generated.models`) is a pre-existing pattern used across the Python SDK — it works via the build step that generates the client. Not something we changed or broke in this PR.

langwatch-agent:
2 P1s

- P1 `langwatch/src/server/prompt-config/prompt.service.ts:997` — the stale-local sync path returns `up_to_date` after comparing only `configData` against the base version. Runtime parameters are ignored in this branch, so a local file that only changes `parameters` while the prompt is behind the server version will be treated as unchanged and never pushed. Minimal fix: include `runtimeParametersEqual(params.parameters, localBaseVersion.runtimeParameters)` in this guard (or fold parameters into the base-version comparison) before returning `up_to_date`.

- P1 `python-sdk/src/langwatch/prompts/prompt_api_service.py:305` — the Python SDK still serializes runtime parameters under `config` for `create`/`update`, but the new prompt API schema expects `parameters`. Those calls will drop the runtime-parameter payload once they hit the updated endpoint. Minimal fix: regenerate the SDK models from the new OpenAPI contract, or switch these call sites to send `parameters=...` so the wire shape matches the server.

langwatch/src/app/api/prompts/__tests__/prompts-api.integration.test.ts
```diff
@@ -209,7 +209,7 @@ describe("Prompts API", () => {
           data: llmPromptConfigVersionFactory.build({
             configId: config.id,
             projectId: testProjectId,
-          }),
+          }) as any,
```
0xdeafcafe: naughty

sergioestebance:
Fixed the `as any` at line 212 — replaced with `as unknown as Prisma.LlmPromptConfigVersionUncheckedCreateInput`. Also fixed the typecheck error and bound all 5 remaining feature scenarios.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
```diff
@@ -0,0 +1,247 @@
...
+} from "@chakra-ui/react";
+import { Plus, X } from "lucide-react";
+import { useCallback, useEffect, useState } from "react";
+import { useFormContext, useWatch } from "react-hook-form";
+import { Tooltip } from "~/components/ui/tooltip";
+import type { PromptConfigFormValues } from "~/prompts/types";
+
+type ParameterEntry = { key: string; value: string };
+
+function serializeValue(raw: string): unknown {
```
Aryansharma28, line 35: 🔴 **Blocking — string parameters get corrupted on round-trip.** Parameters aren't UI-only; they're set via REST/tRPC/SDK with real JSON. `serializeValue` coerces any string that *looks* like another type, and `displayValue(string)` renders it back with no quoting — so there's no way to round-trip a string that looks like a number/bool/json:
- stored `"007"` → displayed `007` → re-saved as number `7`
- stored `"true"` → boolean `true`; `"123"` → `123`; `"{}"` → `{}`

Editing one unrelated row re-serializes **every** row, so this silently mutates API-set values. Needs an unambiguous representation (quote strings, a per-row type, or edit raw JSON text).

langwatch/prisma/migrations/20260602120000_add_runtime_parameters_to_prompt_version/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "LlmPromptConfigVersion" ADD COLUMN "runtimeParameters" JSONB NOT NULL DEFAULT '{}';
```
Aryansharma28, line 2: 🔴 **Blocking — migration timestamp is out of order.** `main` already has `20260524…`, `20260525…`, `20260528120000_add_cost_centers`. This folder's `20260513…` sorts *before* migrations already applied in dev/prod. After merge, `prisma migrate dev` will see an out-of-order pending migration and want to reset. Rename the folder to a current timestamp (after `20260528120000`) before merging. Minor: the folder says `runtime_config` but the column is `runtimeParameters` — align the name.

langwatch/src/server/prompt-config/prompt.service.ts
```diff
@@ -811,7 +840,7 @@ export class PromptService {
     } = params;
 
     // Must run before comparison/creation so both code paths use the merged inputs.
-    const resolvedConfigData = {
+    const resolvedParametersData = {
```
Aryansharma28: 🟠 **Misleading rename.** This variable holds **config data** (`{...localConfigData, inputs: mergeAutoDetectedInputs(...)}`) and is passed to `transformToDbFormat(...)` / `compareConfigContent(...)`. Renaming it to `resolvedParametersData` is factually wrong — parameters are a separate concern carried in `params.parameters`. Please revert to `resolvedConfigData`.

langwatch/src/server/prompt-config/repositories/llm-config-version-schema.ts
```diff
@@ -128,6 +128,33 @@ export function parseLlmConfigVersion(
...
+
+export function parseRuntimeParameters(value: unknown): RuntimeParameters {
+  if (value && typeof value === "object" && !Array.isArray(value)) {
+    return value as RuntimeParameters;
+  }
+  return {};
+}
+
+export function runtimeParametersEqual(a: unknown, b: unknown): boolean {
+  const sortKeysDeep = (obj: unknown): unknown => {
```
Aryansharma28: 🟠 **Duplicated `sortKeysDeep`.** This PR also changed `sortKeysDeep` in `llm-config.repository.ts` to `export function`, but this function defines its own local copy instead of importing it — so we now have duplication **plus** a dead export. Reuse the exported one, or extract to a shared util. (If a repo↔schema import cycle is the concern, move `sortKeysDeep` to a neutral util that both import.)

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
```diff
@@ -0,0 +1,247 @@
...
+              </Tooltip>
+            </HStack>
+          ))}
+        </VStack>
+      )}
+    </VStack>
+  );
+}
+
+export function RuntimeParametersReadonly({
```
Aryansharma28: 🟠 **Dead production code.** `RuntimeParametersReadonly` is only referenced by its own test — the playground uses the editable `RuntimeParametersField`. Either wire it where a read-only view belongs (e.g. version-history display) or delete it (and its test). This is also why the file has two exports.

python-sdk/src/langwatch/prompts/types/prompt_data.py
```diff
@@ -99,7 +88,22 @@ def _unset_to_none(value):
...
             max_tokens=_unset_to_none(response.max_tokens),
             response_format=response_format,
+            parameters=raw_parameters,
             version=int(raw_version) if raw_version is not None else None,
             version_id=_unset_to_none(response.version_id),
             scope=response.scope.value if response.scope and not isinstance(response.scope, Unset) else None,
         )
+
+
+def _read_runtime_parameters(response: object) -> Dict[str, Any]:
```
Aryansharma28, line 98: 🟠 **Reads params via `additional_properties`/`__dict__` fallback.** The request body models were regenerated, but the response models (`*_response_200.py`) were not, so reads depend on the generator stuffing unknowns into `additional_properties`. The unit tests use `Mock`, which doesn't exercise the real `from_dict` extras path — so they don't prove the wire path works. Regenerate the response models so `parameters` is first-class, or add a test against a real deserialized response.

typescript-sdk/src/cli/commands/push.ts
```diff
@@ -141,6 +145,7 @@ export const pushPrompts = async ({
         const syncResult = await promptsApiService.sync({
           name: promptName,
           configData,
+          parameters: localConfig.parameters ?? {},
```
Aryansharma28, line 148: 🟠 **Empty local file wipes remote params.** `updatePrompt` preserves prior params when `parameters === undefined`, but the CLI path always coalesces to `{}` (here and in `prompts-api.service.ts`). So pushing a local file with no `parameters` key overwrites remote params with `{}`. If "local is source of truth on push" is intended, fine — but please confirm, since it diverges from the API/tRPC preserve-on-undefined semantics.

langwatch/src/prompts/hooks/usePromptConfigForm.ts
```diff
@@ -60,7 +60,7 @@ export const usePromptConfigForm = ({
     defaultValues: parsedInitialValues,
     resolver: (data, context, options) => {
       // Use ref to get current schema (updated by useEffect)
-      return zodResolver(schemaRef.current)(data, context, options);
+      return zodResolver(schemaRef.current)(data, context, options as any);
```
Aryansharma28: 🟡 `as any` cast (project rule: no `any` in added code). If the `zodResolver` overload mismatch is the issue, prefer typing `options` to the resolver's expected param type rather than casting.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
```diff
@@ -0,0 +1,247 @@
...
+      </HStack>
+
+      {entries.length === 0 ? (
+        <Text fontSize="13px" color="fg.subtle">
+          No parameters defined
+        </Text>
+      ) : (
+        <VStack align="stretch" gap={2}>
+          {entries.map((entry, index) => (
+            <HStack key={index} gap={2} width="full">
```
Aryansharma28: 🟡 `key={index}` on an editable, removable list is a React anti-pattern — removing a row can mis-associate input state/focus. Prefer a stable per-row id.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
```diff
@@ -0,0 +1,247 @@
...
+    recordToEntries((parameters as Record<string, unknown>) ?? {}),
+  );
+
+  useEffect(() => {
+    const incomingRecord = (parameters as Record<string, unknown>) ?? {};
+    const current = entriesToRecord(entries);
+    if (JSON.stringify(current) !== JSON.stringify(incomingRecord)) {
+      setEntries(recordToEntries(incomingRecord));
+    }
+  }, [parameters]);
```
Aryansharma28, line 106: 🟡 This `useEffect` reads `entries` but only depends on `[parameters]` (stale closure / exhaustive-deps). The `JSON.stringify` guard makes it mostly work, but it's fragile — consider deriving from `parameters` directly or including the needed deps.

Aryansharma28 (commented):
## Code review — runtime parameters

Reviewed the human-authored diff (generated `api-client.ts` / `openapiLangWatch.json` excluded). Solid, well-tested feature — strong integration coverage across create/update/tag/version/restore/sync/conflict/nested. Inline comments below, grouped by severity.

**🔴 Blocking**
- Editor silently corrupts string parameters on round-trip (`serializeValue`).
- Migration timestamp predates already-merged migrations → out-of-order on merge.

**🟠 Non-blocking (should fix)**
- Misleading `resolvedConfigData` → `resolvedParametersData` rename.
- Duplicated `sortKeysDeep` + a now-dead export.
- `RuntimeParametersReadonly` is dead production code (only its own test uses it).
- Python SDK reads params via `additional_properties` fallback (response models not regenerated).
- CLI push always sends `parameters: {}` → empty local file wipes remote params (confirm intent).

**🟡 Nits**
- `as any` in the form resolver, `key={index}` on an editable list, `useEffect` exhaustive-deps.

Nice work overall — the schema validation and preserve-on-undefined update semantics are exactly right, and the test coverage is thorough.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
sergioestebance: Fixed. serializeValue and displayValue are now exact inverses: a string whose contents parse as another JSON type (e.g. 123, true, {}) is shown quoted, so it round-trips back to a string and editing an unrelated row no longer coerces it. Added regression tests for string-vs-number preservation.

langwatch/prisma/migrations/20260602120000_add_runtime_parameters_to_prompt_version/migration.sql
sergioestebance: Fixed. Renamed the migration folder to 20260602120000_add_runtime_parameters_to_prompt_version (after the latest applied migration) and aligned the name to the runtimeParameters column.

langwatch/src/server/prompt-config/prompt.service.ts
sergioestebance: Fixed. Renamed to resolvedConfigData. It holds config data fed to transformToDbFormat and compareConfigContent; runtime parameters flow separately via params.parameters. Verified every usage feeds a config consumer, never a parameters sink.

langwatch/src/server/prompt-config/repositories/llm-config-version-schema.ts
sergioestebance: Fixed. Extracted sortKeysDeep into a neutral module (repositories/sortKeysDeep.ts) imported by both the repository and the version schema, avoiding the repo-schema import cycle. Removed the duplicate copy and the dead export.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
sergioestebance: Fixed. Deleted RuntimeParametersReadonly and its test since nothing rendered it; the file now has a single export.

python-sdk/src/langwatch/prompts/types/prompt_data.py
sergioestebance: Addressed. The recursive runtime-parameters schema cannot be emitted by the OpenAPI generator, so the value rides in the generator additional_properties extras until the field is promoted. _read_runtime_parameters now prefers a first-class attribute and falls back to extras, and a new test deserializes a real response via from_dict (not a Mock) and asserts the parameters survive the wire round-trip.

typescript-sdk/src/cli/commands/push.ts
sergioestebance: Confirmed intentional. CLI push treats the local file as the source of truth for every field (model, messages, etc.), and version-conflict detection guards against clobbering newer remote state. Making only parameters preserve-on-undefined would be inconsistent with the other fields and would make clearing parameters via the CLI impossible, so leaving as-is.

langwatch/src/prompts/hooks/usePromptConfigForm.ts
sergioestebance: Fixed. Dropped the as-any cast. The wrapper now narrows the inferred zodResolver to Resolver<PromptConfigFormValues>, which the schema validates.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
sergioestebance: Fixed. Rows now carry a stable id used as the React key instead of the array index.

langwatch/src/prompts/forms/fields/RuntimeParametersField.tsx
sergioestebance: Fixed. The sync effect now compares incoming parameters against a ref of the last-reconciled record instead of reading entries, removing the stale closure.

rogeriochaves (approved):

PR #4044, opened 2026-05-13
fix(dev): pnpm pin + repo-root workspace mounts so compose.dev.yml matches CI

MichielMAnalytics:
## Problem

`pnpm install --frozen-lockfile` (which the dev compose init service runs via `CI=true`) fails out-of-the-box on a fresh checkout:

```
[ERR_PNPM_LOCKFILE_CONFIG_MISMATCH] Cannot proceed with the frozen installation.
The current "overrides" configuration doesn't match the value found in the lockfile
```

…even though the lockfile is committed and the CI workflow installs it cleanly.

### 1. pnpm version drift

CI's `setup-pnpm-node` action pins **pnpm 10.24.0**. `langwatch/package.json` has no `packageManager` field, so **every other environment** (dev containers, golden VMs, contributor laptops) picks whatever corepack defaults to — currently **pnpm 11.x**.

pnpm 11 has:
- stricter overrides validation than 10.24.0
- an `allowBuilds:` scaffolder that writes interactive prompts (`set this to true or false`) into `pnpm-workspace.yaml` during non-interactive runs, corrupting the file for subsequent `--frozen-lockfile` runs

Pinning to 10.24.0 (commit 1) matches CI exactly and stops the drift.

### 2. Compose hides the repo-root workspace files

The `init` / `app` / etc. services bind-mount only `./langwatch:/app`. When pnpm runs from `/app` inside the container, it walks up looking for the workspace root, finds `/app/pnpm-workspace.yaml`, and **stops** — never seeing the repo-root `pnpm-workspace.yaml` + `package.json`.

But the committed lockfile was generated by pnpm running from the `langwatch/` subdir of the **full** checkout, where it walks up to the repo-root `pnpm-workspace.yaml` (which declares the outer monorepo and has no overrides) and uses **that** as the workspace root. The two contexts therefore compute different effective override configs, and `--frozen-lockfile` correctly rejects the mismatch.

Mounting the two repo-root files into the same paths the container walks up to (commit 2) makes the in-compose pnpm see the same workspace shape CI does, with no changes to `langwatch/pnpm-workspace.yaml` or the lockfile.

## Verification

Reproduced on a fresh `git clone` of `main`:
- before: `docker compose -f compose.dev.yml up -d` → init exits 1 with the error above
- after both commits: init exits 0, app starts, `https://<host>/` returns 200

`pnpm install --frozen-lockfile && pnpm run build` continues to pass in CI (same pnpm version, same lockfile, same overrides — no functional change to the install graph).

## Scope

Two small, focused commits:
- `fix(deps): pin pnpm@10.24.0 in langwatch/package.json` — 1 line
- `fix(dev): mount repo-root workspace files into compose.dev.yml services` — adds 2 bind-mount lines + a comment to each of the 4 services that bind `./langwatch:/app`

No code changes, no lockfile regen.

### DIFF
--- a/compose.dev.yml
+++ b/compose.dev.yml
@@ -134,6 +134,13 @@ services:
     working_dir: /app
     volumes:
       - ./langwatch:/app
+      # The repo-root pnpm-workspace.yaml + package.json define the
+      # outer workspace pnpm walks up to from /app. Without these two
+      # bind-mounts, pnpm stops at /app/pnpm-workspace.yaml and computes
+      # a different effective overrides config than the lockfile was
+      # generated against, breaking `pnpm install --frozen-lockfile`.
+      - ./pnpm-workspace.yaml:/pnpm-workspace.yaml:ro
+      - ./package.json:/package.json:ro
       - ./mcp-server:/mcp-server
       - ./langevals:/langevals
       - ./typescript-sdk:/typescript-sdk:ro
@@ -175,6 +182,13 @@ services:
     working_dir: /app
     volumes:
       - ./langwatch:/app
+      # The repo-root pnpm-workspace.yaml + package.json define the
+      # outer workspace pnpm walks up to from /app. Without these two
+      # bind-mounts, pnpm stops at /app/pnpm-workspace.yaml and computes
+      # a different effective overrides config than the lockfile was
+      # generated against, breaking `pnpm install --frozen-lockfile`.
+      - ./pnpm-workspace.yaml:/pnpm-workspace.yaml:ro
+      - ./package.json:/package.json:ro
       - ./mcp-server:/mcp-server
       - ./langevals:/langevals
       - ./typescript-sdk:/typescript-sdk:ro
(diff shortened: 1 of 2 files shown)


langwatch-agent:

**Review mode:** Fast Skim
**Scariest prod risk:** none found. Dev-only: `compose.dev.yml` bind-mounts and a `packageManager` pin, no prod/runtime code path.

- `[Must Check]` `langwatch/package.json` `packageManager: pnpm@10.24.0` — this forces corepack to 10.24.0 for every contributor and container, not just the compose init. Confirm it matches CI's `setup-pnpm-node` pin so the two never drift again, and that pinning on the inner manifest (not repo root) is where corepack resolves it from `langwatch/`.
- `[Probably Fine]` `compose.dev.yml` — the four `:ro` mounts of repo-root `pnpm-workspace.yaml`/`package.json` at container `/` are consistent across `init`/`app`/`workers`/`ai-server`; verify no fifth service also binds `./langwatch:/app` and was missed.

- Should the `packageManager` pin live at the repo-root `package.json` too (or instead), since that is the actual workspace root pnpm walks up to?

langwatch-agent (commented):
No security or ownership-boundary findings from static external-PR review. The dev-compose/package-manager changes are local development plumbing only, but the PR is currently dirty/merge-conflicted and must be rebased before it can be safely considered. I did not execute PR code.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=7d5e89d2fca12e651cf87f69d2c4d517421c201f p0=0 p1=0 p2=0 p3=0

PR #4048, opened 2026-05-14
fix(loop-prevention): four prod bugs caught by 2026-05-14 dogfood

rogeriochaves:
## TL;DR

Four independent production bugs the previous PR's tests didn't catch. Driving the real langwatch event-sourcing pipeline through testcontainers (Redis + ClickHouse) + actually looking at the prod traces explorer surfaced each one. All four now have unit/integration tests; bug 1, 3 and 4 are proven by the **stash-test-restore** methodology (verified each test FAILS on the unfixed code before claiming the fix works).

| # | Bug | What broke | Proof |
|---|---|---|---|
| 1 | `recordSpan.stripReservedAttributes()` was stripping `langwatch.reserved.causality_depth` | Reactor never saw the depth attribute on inbound spans → loop prevention silently disabled in production since merge | Integration test against testcontainers: depth=1 spans now block dispatch + increment `loop-blocked` counter |
| 2 | nlpgoFetch sent no `traceparent` header + `do_not_trace=true` short-circuited nlpgo's parent-context setup | Every eval workflow run created an orphan trace (the 2026-05-14 screenshot showing the eval's spans on a separate trace from the trace being evaluated) | Unit tests for `formatTraceparent` + `extractParentTraceForNlpgo`; HOW PROVEN: see "Honesty section" below |
| 3 | Accumulator did `JSON.stringify(outputResult.raw)` instead of using `outputResult.text` | Trace summaries showed `{"output":"Hey there"}` instead of `Hey there` | Stash-fix-restore: test fails with `expected '{"input":"hey there"}' to be 'hey there'` without fix |
| 4 | `previewFormatter.unwrapChatArray` knew `content` but not `parts` (Vercel AI SDK / Gemini-shape multi-modal payloads) | New /traces explorer rendered the raw 4KB JSON wrapper (incl. base64 image blobs) for any message-shaped trace using `parts` | Stash-fix-restore: 4 new tests fail without fix, all 29 pass with fix |

---

## Bug 1 — `langwatch.reserved.causality_depth` was being stripped

`recordSpanCommand.stripReservedAttributes()` (predates #3976) nukes every `langwatch.reserved.*` attribute from incoming OTLP spans because user SDKs aren't supposed to set them. The loop-prevention design relies on nlpgo's `BaggageAttributeProcessor` stamping exactly that namespace on every eval-emitted span. Strip → reactor sees no depth → dispatches → ∞.

**Fix:** narrow passthrough allowlist (one entry: `langwatch.reserved.causality_depth`).

**Test:** `loopPrevention.reactor.integration.test.ts` runs against testcontainers Redis + CH, drives spans through the **real** `GroupQueueProcessor`, real fold projection, real `evaluationTrigger` reactor. 4/4 scenarios pass deterministically across 3 consecutive runs:
- depth=0 → dispatches one executeEvaluation per monitor
- depth=1 → BLOCKS dispatch + increments `langwatch_evaluator_loop_blocked_total{reason=depth_direct}`
- per-span guard: fresh depth=0 on same trace after a depth=1 noise span → re-dispatches
- `LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD=1` bypasses depth check

---

## Bug 2 — Eval spans landed on an orphan trace

Two compounding causes:

**(a)** `nlpgoFetch` sent `X-LangWatch-Causality-Depth` and `X-LangWatch-Origin` but **no `traceparent` header**. nlpgo's `applyInboundCausality` saw nothing to extract.

**(b)** `runEvaluationWorkflow` hard-coded `do_not_trace: true`. nlpgo's `startStudioSpan` early-returns a no-op span on that flag — **before** the body-supplied trace_id fallback runs. So even though the body carried the parent trace_id, ctx never got a SpanContext, and inner engine spans (execute_component, LLM calls) created a fresh trace_id.

The `do_not_trace=true` predated the depth-attribute loop prevention; it was a different layer for the same concern, redundant now and actively harmful.

**Fix:**
- `nlpgoFetch`: new `parentTrace?: {traceId, parentSpanId}` option; when set, emits a valid W3C `traceparent` header (`00-<traceId>-<parentSpanId>-01`).
- `formatTraceparent()` validates both fields are well-formed hex; throws on malformed input rather than silently emitting a broken header.
- `runEvaluationWorkflow`: drops `do_not_trace: true`.
- `evaluation-execution.service`: new `extractParentTraceForNlpgo(trace)` plucks the parent trace's root span_id (when both trace_id and root span_id are OTel-standard 32-hex/16-hex). Returns undefined for legacy `trace_<nanoid>` shapes — better than synthesizing a parent_span_id that would render under a non-existent span in Studio's waterfall.
- **Wired through both call sites** of `runEvaluationWorkflow`: `evaluation-execution.service.runCustomEvaluation` AND `background/workers/evaluationsWorker.customEvaluation` (caught the second one on review).

---

## Bug 3 — Trace-summary `computedOutput` showed wrapper JSON

nlpgo emits `langwatch.output = {"output":"Hey there"}`. The IO extraction service correctly unwraps via `messagesToText` / `extractTextFromPlainJson`, populating `outputResult.text`. The IO accumulator threw `text` away and did `JSON.stringify(outputResult.raw)`, so users saw `{"output":"Hey there"}` in trace summaries.

**Fix:** new `preferText(text, raw)` helper. Use the extracted text when present, fall back to `JSON.stringify(raw)` only when extraction returned empty (non-null guarantee preserved).

---

## Bug 4 — Traces explorer rendered `parts` payloads as raw JSON

The new `/traces` explorer's `previewFormatter.unwrapChatArray` walked chat-shaped arrays but only knew `content`. Messages with `parts` (Vercel AI SDK, Gemini API, Mastra) fell through to raw-JSON rendering — the rchaves screenshot showed a 4KB base64-encoded image blob followed by `[{"role":"user","parts":[...]}]` raw in the UI.

The **backend's** `extractMessageContentText` already handled `parts`. Just the renderer was the gap.

**Fix:** new `extractMessagePartsText` mirrors the backend's shape support (`{type:"text", content|text}`, typeless wrapper, plain string). Skips non-text parts so a multi-modal message (image + text) surfaces just the text in the preview.

---

## Honesty section — what's PROVEN vs INFERRED

The user explicitly asked me to be honest about this after I overclaimed on a prior PR.

**Stash-fix-restore proven (test verified to FAIL on unfixed code):**
- Bug 1: full integration test through real `GroupQueueProcessor` against testcontainers — the test was written first against the strip behavior, then the fix landed and made it pass. 4/4 deterministic across 3 runs.
- Bug 3: unit test on `TraceIOAccumulationService` — stashed the `preferText` change, ran, got `expected '{"input":"hey there"}' to be 'hey there'`. Restored, 3/3 pass.
- Bug 4: unit tests on `previewFormatter` — stashed the parts support, 4 new tests fail with the exact "wrapper JSON shown instead of text" symptom; restored, all 29 pass.

**Wire-format proven, end-to-end NOT proven:**
- Bug 2: each individual link is unit-tested (`formatTraceparent` builds a valid header; `extractParentTraceForNlpgo` returns the right context shape; the nlpgo-side `causality_propagation_test.go` proves nlpgo extracts traceparent into ctx and stamps depth on every emitted span). But there is **no single end-to-end test that** spawns nlpgo as a subprocess, lets it emit OTLP back into a real langwatch collector, and asserts the resulting CH-stored eval span has `trace_id == parent_trace_id`. The chain logic is correct on inspection, but rchaves's original "did we ever spin up nlp and run a proper integration test" question still applies to bug 2 specifically — fixable in a follow-up.

**Regression scope check:**
- `pnpm test:unit src/server/event-sourcing src/server/app-layer/traces src/server/nlpgo src/server/app-layer/evaluations` → 6982 pass
- `pnpm test:unit src/features/traces-v2` → 251 pass
- `pnpm check:feature-parity` → 1045 scenarios bound, all clean
- `pnpm test:unit src/server/background src/server/app-layer/evaluations` → 156 pass (covers the second call-site fix)

Total **7233+ unit tests pass, 0 regressions** across event-sourcing transformation, traces canonicalisation, traces-v2 rendering, eval execution, nlpgo dispatch, background workers.

## Files

| File | Change |
|---|---|
| `langwatch/.../commands/recordSpanCommand.ts` | `RESERVED_ATTR_PASSTHROUGH` allowlist (bug 1) |
| `langwatch/.../commands/__tests__/recordSpanCommand.test.ts` | Unit regression pinning the passthrough |
| `langwatch/.../reactors/__tests__/loopPrevention.reactor.integration.test.ts` | Real-queue integration test (bug 1 proof) |
| `langwatch/.../nlpgoFetch.ts` | `parentTrace` option + `formatTraceparent` helper (bug 2) |
| `langwatch/.../nlpgoFetch.unit.test.ts` | Unit tests for traceparent wire format |
| `langwatch/.../runWorkflow.ts` | Drop `do_not_trace=true`; thread `parentTrace` through (bug 2) |
| `langwatch/.../evaluation-execution.service.ts` | `extractParentTraceForNlpgo` helper (bug 2) |
| `langwatch/.../evaluation-execution.service.unit.test.ts` | 5 unit tests for the helper |
| `langwatch/.../background/workers/evaluationsWorker.ts` | Wire parentTrace through second call site (bug 2) |
| `langwatch/.../trace-io-accumulation.service.ts` | `preferText` (bug 3) |
| `langwatch/.../trace-io-accumulation.service.unit.test.ts` | 3 unit tests, exact prod payload |
| `langwatch/.../traces-v2/utils/previewFormatter.ts` | `extractMessagePartsText` (bug 4) |
| `langwatch/.../previewFormatter.unit.test.ts` | 4 new tests, real prod payload shape |
| `services/nlpgo/tests/integration/causality_propagation_test.go` | Restored from closed PR #4035 with corrected scope language (NOT e2e) |
| `specs/monitors/online-evaluator-loop-prevention.feature` | Per-span re-dispatch scenario + traceparent scenarios + reserved-attr passthrough scenario |
| `specs/trace-processing/io-accumulation.feature` | New spec for the accumulator preferText behavior |
| `specs/event-sourcing/tenant-soft-cap.feature` | (already merged earlier) |

You should consider bugs 1+2 a production incident — loop prevention has been silently broken since merge, and every eval workflow run has been creating an orphan trace.

### DIFF
--- a/langwatch/src/features/traces-v2/utils/previewFormatter.ts
+++ b/langwatch/src/features/traces-v2/utils/previewFormatter.ts
@@ -226,6 +226,14 @@ function unwrapChatArray(arr: unknown[]): UnwrapResult | null {
  * visible rather than rendering "(blob)" placeholders that the caller
  * couldn't distinguish from a real text "(blob)".
  */
+// Typed `parts` entries we treat as user-meaningful prose. `text` covers
+// Anthropic / Genkit / Mastra and AI SDK v4. `reasoning` is AI SDK v5
+// where the model's chain-of-thought lands in a sibling part with its
+// own `.text` string — same wire shape, different `type`. Skipping it
+// would re-introduce the wrapper-JSON failure mode this PR is fixing,
+// just for the next payload generation.
+const RENDERABLE_PART_TYPES = new Set(["text", "reasoning"]);
+
 function extractMessagePartsText(parts: unknown[]): string | null {
   const texts: string[] = [];
   for (const part of parts) {
@@ -236,9 +244,9 @@ function extractMessagePartsText(parts: unknown[]): string | null {
     if (!part || typeof part !== "object") continue;
     const p = part as { type?: unknown; text?: unknown; content?: unknown };
 
-    // Typed part: only "text" is renderable as a preview string.
+    // Typed part: only renderable types contribute to the preview.
     if (typeof p.type === "string") {
-      if (p.type !== "text") continue;
+      if (!RENDERABLE_PART_TYPES.has(p.type)) continue;
       if (typeof p.text === "string") {
         texts.push(p.text);
       } else if (typeof p.content === "string") {
--- a/langwatch/src/features/traces-v2/utils/previewFormatter.ts
+++ b/langwatch/src/features/traces-v2/utils/previewFormatter.ts
@@ -48,6 +48,15 @@ const UNWRAP_KEY_ALLOWLIST = new Set([
 interface ChatMessageLike {
   role?: string;
   content?: unknown;
+  /**
+   * Genkit / AI SDK / Mastra emit a `parts` array (sibling of `content`),
+   * where each part is `{ type: "text" | "blob" | "reasoning" | ..., content? | text? }`.
+   * The backend's `extractMessageContentText` handles this shape; without
+   * the matching support here, the new traces explorer renders these
+   * payloads as raw JSON. Reported on 2026-05-14 by rchaves with a real
+   * G'nger product-classification trace.
+   */
+  parts?: unknown;
   tool_calls?: Array<{ function?: { name?: string } }>;
 }
 
@@ -185,12 +194,67 @@ function unwrapChatArray(arr: unknown[]): UnwrapResult | null {
         role: normaliseRole(m.role),
       };
     }
+    // Try `content` first (OpenAI / Anthropic), then `parts` (Genkit /
+    // AI SDK / Mastra). Either may carry the readable payload, and
+    // some integrations populate both — content wins when present.
     const content = extractMessageContent(m.content);
     if (content) {
       return { text: content, role: normaliseRole(m.role) };
     }
+    if (Array.isArray(m.parts)) {
+      const fromParts = extractMessagePartsText(m.parts);
+      if (fromParts) {
+        return { text: fromParts, role: normaliseRole(m.role) };
+      }
+    }
   }
   return null;
+}
+
+/**
+ * Pull text out of a `parts` array. Each part is typically one of:
+ *   - `{ type: "text", content: "..." }`  ← Genkit / Mastra
+ *   - `{ type: "text", text: "..." }`     ← Anthropic style
+ *   - `{ type: "blob" | "image" | "reasoning" | ... }` ← skip; non-text
+ *   - `{ content: "..." }` / `{ text: "..." }` ← typeless wrapper
+ *   - plain string                            ← rare; pass through
+ *
+ * Skips non-text parts so a multi-modal message (image + text)
+ * surfaces just the text in a one-line preview. Non-text parts that
+ * *are* the whole content (e.g. an image-only message) intentionally
+ * yield no preview — fall-through to JSON.stringify keeps the wrapper
+ * visible rather than rendering "(blob)" placeholders that the caller
+ * couldn't distinguish from a real text "(blob)".
+ */
+function extractMessagePartsText(parts: unknown[]): string | null {
+  const texts: string[] = [];
+  for (const part of parts) {
+    if (typeof part === "string") {
+      texts.push(part);
+      continue;
+    }
+    if (!part || typeof part !== "object") continue;
+    const p = part as { type?: unknown; text?: unknown; content?: unknown };
+
+    // Typed part: only "text" is renderable as a preview string.
+    if (typeof p.type === "string") {
+      if (p.type !== "text") continue;
+      if (typeof p.text === "string") {
+        texts.push(p.text);
+      } else if (typeof p.content === "string") {
+        texts.push(p.content);
+      }
+      continue;
+    }
+
+    // Typeless part: accept text or content directly.
+    if (typeof p.text === "string") {
+      texts.push(p.text);
+    } else if (typeof p.content === "string") {
+      texts.push(p.content);
+    }
+  }
+  return texts.length > 0 ? texts.join(" ") : null;
 }
 
 /**
--- a/langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
+++ b/langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
@@ -65,7 +65,36 @@ export interface WorkflowExecutor {
     inputs: Record<string, string>,
     versionId?: string,
     causalityDepth?: number,
+    parentTrace?: { traceId: string; parentSpanId: string },
   ): Promise<{ result: SingleEvaluationResult; status: string }>;
+}
+
+const TRACE_ID_HEX = /^[0-9a-fA-F]{32}$/;
+const SPAN_ID_HEX = /^[0-9a-fA-F]{16}$/;
+
+/**
+ * Extract the W3C `traceparent` context for the eval workflow from the
+ * parent trace. nlpgo needs both pieces (32-hex trace_id + 16-hex root
+ * span_id) so its emitted spans land as children of the parent trace
+ * in Studio's waterfall rather than as a separate orphan trace.
+ *
+ * Returns `undefined` when the parent trace doesn't have OTel-standard
+ * IDs (legacy `trace_<nanoid>` shape, missing root span) — in that
+ * case nlpgo falls back to body-supplied req.TraceID and emits without
+ * a parent linkage. Callers should NOT default-emit a synthesized
+ * parent: a synth parent_span_id would render under a non-existent
+ * span in the waterfall, which is worse UX than a separate trace.
+ */
+export function extractParentTraceForNlpgo(
+  trace: Trace | undefined,
+): { traceId: string; parentSpanId: string } | undefined {
+  if (!trace?.trace_id || !TRACE_ID_HEX.test(trace.trace_id)) return undefined;
+  const rootSpan = trace.spans?.find((s) => !s.parent_id);
+  if (!rootSpan?.span_id || !SPAN_ID_HEX.test(rootSpan.span_id)) return undefined;
+  return {
+    traceId: trace.trace_id.toLowerCase(),
+    parentSpanId: rootSpan.span_id.toLowerCase(),
+  };
 }
 
 /**
@@ -467,12 +496,19 @@ export class EvaluationExecutionService {
       ...data,
     };
 
+    // W3C trace context: link the eval workflow's spans to the parent
+    // trace's root span so Studio's waterfall renders them as a child
+    // sub-tree (not a separate orphan trace, which is the 2026-05-14
+    // bug rchaves caught in prod).
+    const parentTrace = extractParentTraceForNlpgo(trace);
+
     const response = await this.deps.workflowExecutor.runEvaluationWorkflow(
       resolvedWorkflowId,
       projectId,
       requestBody as Record<string, string>,
       undefined,
       parentCausalityDepth,
+      parentTrace,
     );
 
     if (response.status !== "success") {
--- a/langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
+++ b/langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
@@ -89,7 +89,22 @@ export function extractParentTraceForNlpgo(
   trace: Trace | undefined,
 ): { traceId: string; parentSpanId: string } | undefined {
   if (!trace?.trace_id || !TRACE_ID_HEX.test(trace.trace_id)) return undefined;
-  const rootSpan = trace.spans?.find((s) => !s.parent_id);
+
+  // Broken / multi-source instrumentation can leave a trace with more
+  // than one parent-less span. `find()` would then pick whichever span
+  // happened to be ingested first — non-deterministic across re-runs.
+  // Sort by started_at (earliest is the true root in any sane trace)
+  // with span_id as the tie-breaker to keep two consecutive eval runs
+  // pinned to the same parent_span_id.
+  const rootCandidates = (trace.spans ?? []).filter((s) => !s.parent_id);
+  if (rootCandidates.length === 0) return undefined;
+  rootCandidates.sort((a, b) => {
+    const aStart = a.timestamps?.started_at ?? Number.MAX_SAFE_INTEGER;
+    const bStart = b.timestamps?.started_at ?? Number.MAX_SAFE_INTEGER;
+    if (aStart !== bStart) return aStart - bStart;
+    return (a.span_id ?? "").localeCompare(b.span_id ?? "");
+  });
+  const rootSpan = rootCandidates[0];
   if (!rootSpan?.span_id || !SPAN_ID_HEX.test(rootSpan.span_id)) return undefined;
   return {
     traceId: trace.trace_id.toLowerCase(),
--- a/langwatch/src/server/background/workers/evaluationsWorker.ts
+++ b/langwatch/src/server/background/workers/evaluationsWorker.ts
@@ -70,7 +70,10 @@ import {
 } from "../../app-layer/evaluations/azure-safety-env";
 import { getAzureSafetyEnvFromProject } from "../../app-layer/evaluations/azure-safety-env.server";
 import { runEvaluationWorkflow } from "../../workflows/runWorkflow";
-import { maxCausalityDepthOfSpans } from "../../app-layer/evaluations/evaluation-execution.service";
+import {
+  extractParentTraceForNlpgo,
+  maxCausalityDepthOfSpans,
+} from "../../app-layer/evaluations/evaluation-execution.service";
 import {
   EVALUATIONS_QUEUE,
   updateEvaluationStatusInES,
@@ -848,12 +851,19 @@ const customEvaluation = async (
     throw new Error("Workflow ID is required");
   }
 
+  // W3C trace context — same as eval-execution.service: pass the
+  // parent trace's root span so nlpgo emits eval spans as a child
+  // sub-tree of the trace being evaluated, not as an orphan trace.
+  // See 2026-05-14 prod regression.
+  const parentTrace = extractParentTraceForNlpgo(trace);
+
   const response = await runEvaluationWorkflow(
     resolvedWorkflowId,
     project.id,
     requestBody,
     undefined,
     parentCausalityDepth,
+    parentTrace,
   );
 
   const { result, status } = response;
--- a/langwatch/src/server/background/workers/evaluationsWorker.ts
+++ b/langwatch/src/server/background/workers/evaluationsWorker.ts
@@ -841,9 +841,15 @@ const customEvaluation = async (
     throw new Error("Project not found");
   }
 
+  // do_not_trace=false on the wire so eval-emitted spans land under
+  // the parent trace's traceparent context (post-2026-05-14 fix). The
+  // explicit do_not_trace=false arg passed to runEvaluationWorkflow
+  // below already overrides this default, but pinning the field on
+  // the body itself keeps the wire shape honest if a future refactor
+  // drops the explicit override on runWorkflow.
   const requestBody: Record<string, any> = {
     trace_id: trace?.trace_id,
-    do_not_trace: true,
+    do_not_trace: false,
     ...data,
   };
 
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
@@ -234,9 +234,30 @@ export class RecordSpanCommand implements CommandHandler<
   }
 
   /**
-   * Strips any `langwatch.reserved.*` attributes from a span and its events/links.
-   * These attributes are reserved for system use and must not be set by users.
+   * Strips user-submitted `langwatch.reserved.*` attributes from a span
+   * and its events/links/resource. These attributes are reserved for
+   * system use and must not be settable by customer SDKs.
+   *
+   * EXCEPTIONS — system-emitted attributes that ride in on OTLP from
+   * trusted internal services (nlpgo, langevals, etc.) and MUST survive
+   * this strip because downstream reactors depend on them:
+   *
+   *   - `langwatch.reserved.causality_depth` — stamped by nlpgo's
+   *     `BaggageAttributeProcessor` on every span emitted during an
+   *     evaluator workflow run. The evaluationTrigger reactor reads
+   *     this on the inbound span_received event to block infinite
+   *     loops (post-2026-05-11 incident). Stripping it here would
+   *     silently disable the loop-prevention guard in production.
+   *
+   * If a customer SDK does manage to set one of these from outside,
+   * worst case is a one-shot eval-skip on their own trace — bounded
+   * impact, and bypassing requires knowing internal attribute names.
+   * Far preferable to silently breaking loop prevention.
    */
+  private static readonly RESERVED_ATTR_PASSTHROUGH = new Set<string>([
+    "langwatch.reserved.causality_depth",
+  ]);
+
   private static stripReservedAttributes(
     span: OtlpSpan,
     resource: OtlpResource | null,
@@ -246,7 +267,10 @@ export class RecordSpanCommand implements CommandHandler<
 
     const strip = (attributes: OtlpSpan["attributes"]): OtlpSpan["attributes"] => {
       const filtered = attributes.filter((attr) => {
-        if (attr.key.startsWith(RESERVED_PREFIX)) {
+        if (
+          attr.key.startsWith(RESERVED_PREFIX) &&
+          !RecordSpanCommand.RESERVED_ATTR_PASSTHROUGH.has(attr.key)
+        ) {
           logger.warn(
             { attributeKey: attr.key },
             "Stripped user-submitted langwatch.reserved.* attribute",
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
@@ -233,6 +233,10 @@ export class RecordSpanCommand implements CommandHandler<
     };
   }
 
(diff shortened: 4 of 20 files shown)


langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
```diff
@@ -234,9 +234,30 @@ export class RecordSpanCommand implements CommandHandler<
...
+   *
+   * If a customer SDK does manage to set one of these from outside,
+   * worst case is a one-shot eval-skip on their own trace — bounded
+   * impact, and bypassing requires knowing internal attribute names.
+   * Far preferable to silently breaking loop prevention.
    */
+  private static readonly RESERVED_ATTR_PASSTHROUGH = new Set<string>([
+    "langwatch.reserved.causality_depth",
+  ]);
+
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 84aa7515e — moved RESERVED_ATTR_PASSTHROUGH above the JSDoc so the comment attaches to stripReservedAttributes again.

services/nlpgo/tests/integration/causality_propagation_test.go
```diff
@@ -0,0 +1,378 @@
...
+	for _, s := range ended {
+		if s.Parent().IsValid() && s.Parent().IsRemote() {
+			rootStudioSpan = s
+			break
+		}
+	}
+	require.NotNil(t, rootStudioSpan,
+		"no studio root span found with a remote parent — applyInboundCausality "+
+			"did not extract the inbound traceparent into ctx, or startStudioSpan "+
+			"started a fresh trace instead of continuing the inbound one")
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 84aa7515e — added SpanKindServer filter to the rootStudioSpan lookup so an unrelated child can't satisfy the predicate.

langwatch/src/server/nlpgo/__tests__/traceparent-roundtrip.integration.test.ts
```diff
@@ -0,0 +1,553 @@
...
+              res.end();
+            } catch (err) {
+              // eslint-disable-next-line no-console
+              console.error("[fake-langwatch] handler failed", err);
+              res.statusCode = 500;
+              res.end();
+            }
+          })();
+        });
+      });
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 84aa7515e — added req.on('error') handler on the fake langwatch server (closes the response with 500 + logs under NLPGO_TEST_LOG=1).

langwatch/src/server/nlpgo/__tests__/traceparent-roundtrip.integration.test.ts
```diff
@@ -0,0 +1,553 @@
...
+        }
+      });
+      nlpgoProcess.on("exit", (code, signal) => {
+        if (code !== 0 && code !== null) {
+          // eslint-disable-next-line no-console
+          console.error(
+            `nlpgo exited unexpectedly: code=${code} signal=${signal}`,
+          );
+        }
+      });
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 84aa7515e — now drain both stdout and stderr from the nlpgo subprocess (was risking a deadlock once the 64KiB pipe buffer filled).

langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/loopPrevention.reactor.integration.test.ts
```diff
@@ -0,0 +1,551 @@
...
+      } finally {
+        if (prev === undefined) {
+          delete process.env.LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD;
+        } else {
+          process.env.LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD = prev;
+        }
+      }
+    });
+  },
+);
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 84aa7515e — refactored all four scenarios into nested describe("given …") > describe("when …") > it(…) blocks per CLAUDE.md's BDD convention. All 4 still pass locally against the real testcontainers pipeline.

langwatch/src/server/background/workers/evaluationsWorker.ts
```diff
@@ -848,12 +851,19 @@ const customEvaluation = async (
...
+  // sub-tree of the trace being evaluated, not as an orphan trace.
+  // See 2026-05-14 prod regression.
+  const parentTrace = extractParentTraceForNlpgo(trace);
+
   const response = await runEvaluationWorkflow(
     resolvedWorkflowId,
     project.id,
     requestBody,
     undefined,
     parentCausalityDepth,
```
drewdrewthis, line 871: Worth a peek a few lines up at this function's `requestBody` — it still hard-codes `do_not_trace: true` (line 846, unchanged). `runEvaluationWorkflow`'s explicit `do_not_trace=false` 5th arg currently overrides `inputs.do_not_trace`, so the flag is effectively dead — but a future refactor that flips `runWorkflow` to prefer `inputs` (or drops the explicit param) would silently re-introduce the bug-2 regression on THIS code path while leaving the eval-execution one fixed. Either drop the line above, set it to `false`, or pin the wire behaviour from this call site with a regression test — this site was already missed once.

langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
```diff
@@ -65,9 +65,38 @@ export interface WorkflowExecutor {
...
+ * case nlpgo falls back to body-supplied req.TraceID and emits without
+ * a parent linkage. Callers should NOT default-emit a synthesized
+ * parent: a synth parent_span_id would render under a non-existent
+ * span in the waterfall, which is worse UX than a separate trace.
+ */
+export function extractParentTraceForNlpgo(
+  trace: Trace | undefined,
+): { traceId: string; parentSpanId: string } | undefined {
+  if (!trace?.trace_id || !TRACE_ID_HEX.test(trace.trace_id)) return undefined;
+  const rootSpan = trace.spans?.find((s) => !s.parent_id);
```
drewdrewthis: `spans?.find((s) => !s.parent_id)` picks the *first* root by array order. If a malformed trace contains multiple parent-less spans (broken instrumentation, missing parent link, etc.), the picked span_id depends on whichever order spans were ingested. Two consecutive eval runs against the same trace could pick different root span_ids, producing inconsistent traceparent linkage. Two cheap mitigations: (a) sort by `startTimeUnixMs` and pick earliest, or (b) only return when there's exactly one root and emit `undefined` otherwise (lets nlpgo body-fallback handle the ambiguous case instead of pinning to whatever arrived first).

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
```diff
@@ -233,9 +233,30 @@ export class RecordSpanCommand implements CommandHandler<
...
+   * this strip because downstream reactors depend on them:
+   *
+   *   - `langwatch.reserved.causality_depth` — stamped by nlpgo's
+   *     `BaggageAttributeProcessor` on every span emitted during an
+   *     evaluator workflow run. The evaluationTrigger reactor reads
+   *     this on the inbound span_received event to block infinite
+   *     loops (post-2026-05-11 incident). Stripping it here would
+   *     silently disable the loop-prevention guard in production.
+   *
+   * If a customer SDK does manage to set one of these from outside,
```
drewdrewthis, line 256: Threat-model question: doc says "worst case is a one-shot eval-skip on their own trace" — but with the passthrough in place, any customer SDK can disable monitor dispatch on their own traces *permanently* by stamping `langwatch.reserved.causality_depth=1` on every emitted span. Bounded blast radius (only their project), but it's a self-DoS vector for monitoring. Consider either (a) gating passthrough by trusted source (e.g. only when `langwatch.origin=evaluation` is co-present on the span), or (b) explicitly documenting this trade-off in the spec so reviewers don't re-litigate later. Probably fine as-is — flagging so it's a deliberate decision.

langwatch/src/server/nlpgo/nlpgoFetch.ts
```diff
@@ -101,6 +150,15 @@ export async function nlpgoFetch<T = unknown>(
     headers["X-LangWatch-Causality-Depth"] = String(callerDepth);
   }
 
+  // W3C traceparent — makes the receiving service's spans inherit our
+  // trace_id and parent_span_id. Without this header, nlpgo creates a
+  // new trace_id for the evaluation, breaking the parent-child link
+  // (the 2026-05-14 orphan-trace bug rchaves caught in prod).
+  // Format: 00-<32-hex traceId>-<16-hex parentSpanId>-<flags>
+  if (opts.parentTrace) {
+    headers["traceparent"] = formatTraceparent(opts.parentTrace);
```
drewdrewthis, line 168: `formatTraceparent` hard-codes `-01` (sampled flag). For eval workflows this is what we want (always trace). But the helper reads like a generic builder — if a future caller passes a `parentTrace` derived from a non-sampled trace, we'd be force-sampling. Consider either (a) accepting a `sampled` field with default `true`, or (b) renaming to something explicit like `formatTraceparentForEval` so callers can't accidentally reach for it generically and lose the sampling semantics.

langwatch/src/features/traces-v2/utils/previewFormatter.ts
```diff
@@ -185,14 +194,69 @@ function unwrapChatArray(arr: unknown[]): UnwrapResult | null {
...
+    if (typeof part === "string") {
+      texts.push(part);
+      continue;
+    }
+    if (!part || typeof part !== "object") continue;
+    const p = part as { type?: unknown; text?: unknown; content?: unknown };
+
+    // Typed part: only "text" is renderable as a preview string.
+    if (typeof p.type === "string") {
+      if (p.type !== "text") continue;
```
drewdrewthis: Future-proofing: the typed-parts walker only matches `type: "text"`. AI SDK v5 (Vercel) is rolling out `type: "reasoning"` parts that carry user-meaningful prose in `.text`. Today we'd skip them and fall through to wrapper JSON — same failure mode this PR is fixing, just for the next-generation payload shape. Worth adding a TODO or expanding the allowlist to include `reasoning` (or any type with a `.text` string field). Non-blocking; just calling out before the next dogfood screenshot surfaces it.

langwatch/src/server/event-sourcing/pipelines/trace-processing/projections/services/trace-io-accumulation.service.ts
```diff
@@ -223,3 +230,21 @@ export class TraceIOAccumulationService {
...
+ * only when extraction returned no text — keeps NON-null guarantee
+ * for spans that have data but unknown shape.
+ *
+ * Exported via the existing accumulation surface — tests cover this
+ * via the fold projection, not directly.
+ */
+function preferText(text: string | null | undefined, raw: unknown): string {
+  if (typeof text === "string" && text.length > 0) return text;
+  if (typeof raw === "string") return raw;
+  return JSON.stringify(raw);
```
drewdrewthis, line 254: `JSON.stringify(undefined)` returns `undefined` (not the string `"undefined"`), and TypeScript's return type here says `string`. If a caller ever invokes `preferText("", undefined)`, this returns literal `undefined` cast to `string` — silently corrupting downstream code. Current call sites all guard with a truthy `inputResult`/`outputResult` check, so it's unreachable today, but the helper has no such guard. Either tighten the `raw` parameter type to `NonNullable<unknown>`, or add an explicit `if (raw === undefined) return "";` early-return.

langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/loopPrevention.reactor.integration.test.ts
```diff
@@ -0,0 +1,569 @@
...
+      const redis = getTestRedisConnection()!;
+      const dedupKeys = await redis.keys(
+        `*eval-trigger:${tenantIdString}:${traceId}*`,
+      );
+      if (dedupKeys.length > 0) {
+        await redis.del(...dedupKeys);
+      }
+
+      const prev = process.env.LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD;
+      process.env.LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD = "1";
```
drewdrewthis, line 541: `process.env.LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD = "1"` mutates a process-global. Within this file the suite runs sequentially so it's safe, but if anyone enables vitest's `--threads` / workspace-parallel mode on this folder, the env-var leak would race with other reactor tests reading the same flag. Cheap fix: stub the kill switch through reactor config (already a closure dep — just plumb a flag through) instead of mutating `process.env`, or add a comment pinning the sequential assumption. Not blocking.

langwatch/src/server/nlpgo/__tests__/traceparent-roundtrip.integration.test.ts
```diff
@@ -0,0 +1,695 @@
...
+  } catch {
+    cachedMtime = 0;
+  }
+  const watchDirs = [
+    path.join(REPO_ROOT, "services", "nlpgo"),
+    path.join(REPO_ROOT, "cmd", "service"),
+    path.join(REPO_ROOT, "pkg"),
+  ].filter((p) => fs.existsSync(p));
+
+  function newestGoMtime(dir: string): number {
```
drewdrewthis, line 148: Cache-staleness gotcha: `newestGoMtime` recurses into hardcoded watch dirs (`services/nlpgo`, `cmd/service`, `pkg`). If a future PR adds a new top-level Go module path that the binary depends on (vendored module, generated code dir), the staleness check would miss it and reuse a stale binary — silently passing tests against old code, which is the exact debugging nightmare this PR is trying to prevent for prod. Worth either (a) using `go list -deps -f '{{.GoFiles}}'` to derive the actual file set, or (b) keying the binary cache on git tree hash of the relevant subtrees. Current heuristic is probably fine for ~6 months; flagging so the next person knows where to look.

drewdrewthis (commented):
Reviewed bugs 1–4. Detailed test fixtures and the honesty section are appreciated. Inline comments below — one likely-dead-code concern in the second eval call site, a minor non-determinism in root-span selection, plus a few smaller threadability / future-proofing notes.

langwatch/src/server/background/workers/evaluationsWorker.ts
rogeriochaves: Fixed in 810055cad — pinned `do_not_trace: false` on requestBody itself, with a comment explaining why the call-site override isn't enough. Future refactor that drops the explicit arg or prefers `inputs.do_not_trace` won't silently reintroduce the regression now.

langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
rogeriochaves: Fixed in 810055cad — root-span lookup now sorts candidates by `timestamps.started_at` with `span_id` tie-break before picking. Added two regression tests (multi-root + tie-break paths). Took the sort-by-time path you suggested over the ambiguity-undefined path because it keeps the parent-linkage feature usable for traces with broken instrumentation rather than degrading to no-parent.

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
rogeriochaves: Acknowledging this as a deliberate trade-off rather than fixing. Re-reading the threat model: the customer would have to (a) know the internal attribute name `langwatch.reserved.causality_depth`, (b) ship instrumentation that stamps it on every span, (c) accept that the blast radius is *only their own project*. That's a self-DoS, not a privilege-escalation or cross-tenant attack, and any future internal customer-tooling that legitimately needs to suppress eval dispatch has a much cleaner API surface (the existing monitor preconditions). The strict alternative — gating passthrough on `langwatch.origin=evaluation` co-presence — would break legitimate downstream telemetry that hasn't (yet) stamped the origin. Strengthened the docstring on this PR to make the trade-off explicit rather than implicit, so future reviewers see the call-out without having to re-derive it.

langwatch/src/server/nlpgo/nlpgoFetch.ts
rogeriochaves: Fixed in 810055cad — `formatTraceparent` now takes an optional `sampled` flag defaulted to true. Existing eval-chain callers stay unchanged; future propagators from a non-sampled inbound context can pass `{ sampled: false }` explicitly to avoid force-sampling. Comment in the helper spells out the contract.

langwatch/src/features/traces-v2/utils/previewFormatter.ts
rogeriochaves: Fixed in 810055cad — extracted the typed-part allowlist into `RENDERABLE_PART_TYPES` and added `reasoning` to it. Added a regression test using the AI SDK v5 shape (mixed `reasoning` + `text` parts) — confirms both contribute to the preview now instead of falling through to wrapper JSON. Good catch on the next-payload-generation angle, this would have surfaced as the same "raw JSON in the preview" bug we just fixed for the current shape.

langwatch/src/server/event-sourcing/pipelines/trace-processing/projections/services/trace-io-accumulation.service.ts
rogeriochaves: Fixed in 810055cad — added explicit `if (raw === undefined) return "";` early-return. The return type now actually matches reality, and a future caller that hands in undefined won't silently propagate the literal value through the cast.

langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/loopPrevention.reactor.integration.test.ts
rogeriochaves: Replying without fixing for now. The mutation IS already wrapped in a save-and-restore (`prev` snapshot, `try/finally` with `delete` if it was undefined originally, otherwise restore) so within this file, even if a parallel run came along mid-test the leak window is bounded. Plumbing the kill switch through reactor config IS the cleaner long-term shape — happy to do it as a follow-up if/when we enable `--threads` on this suite, but right now no other reactor integration test reads this flag and the suite is sequential by config, so the marginal hardening doesn't pay for the API churn today. Added it to my mental TODO list when we open up parallelism.

langwatch/src/server/nlpgo/__tests__/traceparent-roundtrip.integration.test.ts
rogeriochaves: Acknowledged, leaving as-is for now per your own "probably fine for ~6 months" qualifier. The cheap mitigation if/when this bites: switch the cache key from mtime-vs-watch-dirs to `go list -deps` output (lets Go itself tell us what files are in the dependency closure) or a git tree-hash of the relevant subtrees. Filed it in my head — if a future nlpgo refactor adds a new top-level dir and we get a confusing cache-stale failure, this is the first place to look.

drewdrewthis (approved):
All eight inline findings addressed — six fixed in 810055cad (do_not_trace pinned on requestBody, root-span sort by started_at + tie-break, optional sampled flag on formatTraceparent, RENDERABLE_PART_TYPES allowlist with `reasoning`, undefined early-return in preferText, threat-model docstring strengthened) and two deferred with sound rationale (env-var mutation is bounded by save/restore + sequential suite; Go cache staleness matches my own ~6-month qualifier). Self-DoS trade-off on the reserved-attr passthrough is a deliberate acceptance — the strengthened docstring captures the call-out so future reviewers don't re-litigate.

LGTM. Loop prevention silently disabled since merge + every eval workflow run creating an orphan trace makes this a production incident — please ship as soon as CI is green.

drewdrewthis (commented):
Walking back my approval slightly — feature-parity is incomplete:

**Bug 4 has no feature file.** The 5 new `previewFormatter.unit.test.ts` tests for the `parts` walker (Genkit/Mastra `{type:text, content}`, multi-modal blob+text, Anthropic `{type:text, text}`, multi-part joining, fall-through, plus the new `reasoning` regression test from the review fix) have no `@scenario` tags and no matching feature file anywhere in `specs/`. Per `CLAUDE.md` ("If no feature file exists for your task, create one before writing code"), Bug 4 needed `specs/traces-v2/preview-formatter.feature` or similar, with scenarios mapping to each unit test.

**One orphan scenario in `online-evaluator-loop-prevention.feature`:** `Scenario: nlpgo handler increments causality_depth on its root span` has no matching `@scenario` tag in any Go or TS test file.

Everything else binds cleanly — `extractParentTraceForNlpgo`, `formatTraceparent`, the reserved-attr passthrough, the loop-prevention reactor scenarios, and both `io-accumulation.feature` scenarios.

Not blocking the merge given the production incident urgency, but please file a follow-up to add the missing feature file for the preview formatter `parts` support and either drop or bind the orphan nlpgo scenario.

drewdrewthis (commented):
Correction on my last comment — I had parity inverted. The rule is one-way: every `Scenario:` in a `.feature` file needs a matching `@scenario` tag in a test, but tests don't need feature scenarios. So the previewFormatter unit tests without `@scenario` tags are fine — they're plain regression tests, not BDD-bound behavior.

That collapses my earlier finding to one real gap: **`Scenario: nlpgo handler increments causality_depth on its root span`** in `online-evaluator-loop-prevention.feature` (lines 110–114) has no matching `@scenario` tag in any test file. `TestCausalityPropagation_AllSpansShareTraceIDAndCarryDepth` asserts depth="1" on *every* span (root included), so the behavior is covered — just not bound to that scenario's tag. Either (a) add the `@scenario` reference to the existing Go test, (b) split out a dedicated root-span-depth test, or (c) drop the scenario from the feature file since it's a subset of the existing all-spans-carry-depth assertion.

Apology for the false flag on Bug 4.

PR #4072, opened 2026-05-15
fix(model-providers): don't clobber existing rows when adding a second instance; quick-add scope chips

rogeriochaves:
## Problem (Bugs May 15, Area B1)

> If I already have an OpenAI model, when I try to add a new one, it overrides the previous?? what the hell?

A second instance of the same provider at a different scope (e.g. one OpenAI at org level + another at project level) is supposed to be supported since iter 109. In practice, adding a "new" one silently UPDATED the first.

Also: "'This project' and 'This team' should be an option at the very top when creating a new model provider, for easy access."

## Root cause

`modelProvider.service.ts::findExistingProvider` fell through to `findByProvider(provider, projectId)` whenever the caller had no `id`. That match is scope-blind, so a project-scoped create matched the org-scoped row and updated it.

## Fix

- **`findExistingProvider`** now requires an explicit `id`. Absent `id` → treat as create intent and go straight to `createNew`. Multi-instance per provider now actually works.
- **`upsertByProviderKey`** added for the REST endpoint `PUT /api/model-providers/:provider`, which is keyed by URL provider string and preserves the legacy single-instance upsert contract. REST callers are unaffected.
- **Quick-add scope chips** at the top of the drawer: `This project` / `This team` / `Organization` — one-click pre-fills for the 90% case. Multi-select still below for cross-team / cross-project setups. The scope section moved above credentials so the chips actually live at the top.

## Spec

`specs/model-providers/scope-and-multi-instance.feature` — 15/15 scenarios bound (parity green). Three previously-`@unimplemented` scenarios are now executable (`Create a second OpenAI row under a different scope`, `Save a provider with multiple scopes`, `Model Providers page lists all accessible rows across scopes`). Three new `@integration` scenarios cover the quick-add chips.

## Tests

- `langwatch` integration (real Postgres): `modelProvider.multiInstance.integration.test.ts` proves the override fix (two distinct OpenAI rows survive with their own scope sets) and that a single row can carry multiple ModelProviderScope entries.
- `langwatch` jsdom integration: `ModelProviderScopeSection.integration.test.tsx` proves each chip replaces the scope with only that scope, and that the chip group hides on personal-account projects.

## QA

- [x] 276 unit tests green across model-providers + settings components.
- [x] Feature parity 15/15 scope-and-multi-instance + global parity OK.
- [x] Real-DB integration proves the second OpenAI row no longer overwrites the first.
- [ ] CI green + CodeRabbit clean (via `/drive-pr`).

## Coordination

Pairs with Luigi's PR B2 (hierarchical default-model section). My touch list stayed off the DefaultProviderSection delete + page-level filters/view he's redesigning.

### DIFF
--- a/langwatch/src/app/api/model-providers/[[...route]]/app.v1.ts
+++ b/langwatch/src/app/api/model-providers/[[...route]]/app.v1.ts
@@ -115,7 +115,11 @@ app.put(
         defaultModel = `${provider}/${defaultModel}`;
       }
 
-      await service.updateModelProvider({
+      // REST endpoint is keyed on the provider string in the URL and
+      // preserves the legacy single-instance upsert contract. The
+      // multi-instance create flow lives behind the tRPC `update`
+      // procedure, which goes through the id-based path.
+      await service.upsertByProviderKey({
         projectId: project.id,
         provider,
         enabled: data.enabled,
--- a/langwatch/src/components/settings/ModelProviderForm.tsx
+++ b/langwatch/src/components/settings/ModelProviderForm.tsx
@@ -246,18 +246,6 @@ export const EditModelProviderForm = ({
           </Field.Root>
         )}
 
-        <CredentialsSection
-          state={state}
-          actions={actions}
-          provider={provider}
-          fieldErrors={fieldErrors}
-          setFieldErrors={setFieldErrors}
-          projectId={projectId}
-          organizationId={organizationId}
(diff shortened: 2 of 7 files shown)


rogeriochaves:
Consolidated into #4073 per @rchaves request to merge all model-providers work into a single PR. Commit e0198f4c7 on feat/model-providers-hierarchical-defaults.

PR #4073, opened 2026-05-15
feat(model-providers): role-based default models + missing-model popup + multi-instance providers

rogeriochaves:
Replaces LangWatch's flat per-scope default-model state (`Organization.defaultModel`, `Team.topicClusteringModel`, etc.) with a CSS-cascade policy layer every AI-powered surface reads through a single resolver. Settles four user-facing problems at once:

1. **Adding a new OpenAI model used to silently overwrite the previous one.** `findExistingProvider` was scope-blind when `id === "new"`, so a project-scoped create updated the org-scoped row. Fixed; the REST upsert path uses a separate `upsertByProviderKey`.
2. **Quick-add chips ("This project" / "This team") on the model-provider scope picker** so users don't always have to dig through the chip picker for the obvious common case.
3. **Default-model UI used to render one form field per scope per role**, which broke the user's mental model of "one rule applies to Team A + Team B + Project edge". New design is a table where each row is one policy with N scope chips, edited in a drawer.
4. **AI surfaces calling `project.defaultModel ?? DEFAULT_MODEL` directly** never honoured inheritance from team or org. Centralised in `resolveModelForFeature(featureKey, ctx)` which every server-side consumer now goes through.

## Data model

One row = one policy. Policies attach to N scopes via a join table, so a "Production models" rule can apply to Team Platform + Team Research + Project edge with one row of state to maintain.

```prisma
model ModelDefaultConfig {
  id        String                    @id @default(nanoid())
  config    Json                       // { "DEFAULT": "openai/gpt-5.5", "traces.ai_search": "anthropic/claude-sonnet-4-6", … }
  authorId  String?
  createdAt DateTime                   @default(now())
  updatedAt DateTime                   @updatedAt
  scopes    ModelDefaultConfigScope[]
}

model ModelDefaultConfigScope {
  id        String                @id @default(nanoid())
  configId  String
  config    ModelDefaultConfig    @relation(fields: [configId], references: [id], onDelete: Cascade)
  scopeType ModelDefaultScopeType  // ORGANIZATION | TEAM | PROJECT
  scopeId   String
  createdAt DateTime              @default(now())

  @@unique([configId, scopeType, scopeId])
  @@index([scopeType, scopeId])
  @@index([configId])
}

enum ModelDefaultScopeType {
  ORGANIZATION
  TEAM
  PROJECT
}
```

JSON keys are either role names (`DEFAULT` / `FAST` / `EMBEDDINGS`) or feature keys from the registry (`prompt.create_default`, `traces.ai_search`, `analytics.topic_clustering_llm`, …). **Absence of a key = inherit from the next scope up**, no sentinel string stored; the merge logic stays trivial.

`ModelDefaultScopeType` matches the per-table-enum convention used by `RoleBindingScopeType`, `GatewayBudgetScopeType`, and the freshly-typed `ModelProviderScopeType`. Junk scope values are rejected by Postgres rather than by hand-rolled validators at every call site. The `dev/docs/best_practices/scoped-resources.md` doc captures the storage, read-grouping, and UI primitive pattern so the next scoped resource lands the same way.

## Resolver

`resolveModelForFeature(featureKey, ctx)` walks, in cascade order:

1. **PROJECT** tier: configs attached at this project, sorted createdAt DESC.
2. **TEAM** tier: configs attached at this project's team, createdAt DESC.
3. **ORGANIZATION** tier: configs attached at this project's organization, createdAt DESC.
4. `ModelNotConfiguredError`.

At each tier the resolver picks the first config with the feature key set (`feature_override`); if none has it, the first with the role key set (`role_default`). Feature-key match beats role-key match within a tier; lower tier always beats higher tier regardless of recency.

**There is intentionally no global system fallback.** Fresh accounts with zero providers no longer look pre-configured: any AI surface calling the resolver on a project with nothing set throws and the frontend interceptor surfaces a sticky toast prompting the user to configure a default.

Return shape:

```ts
{
  model: string;
  source: "feature_override" | "role_default";
  scope: "project" | "team" | "organization" | null;
  feature: FeatureDescriptor;
}
```

The tRPC layer surfaces `ModelNotConfiguredError` as `TRPCError(BAD_REQUEST, cause: { code: "MODEL_NOT_CONFIGURED", … })`; the Hono REST middleware returns 422 with the same cause shape. The frontend interceptor opens a singleton `MissingModelToast` deep-linking to the right settings anchor (`/{projectSlug}/settings/model-providers#role-{role}` or `?expand={role}&feature={featureKey}`), with an RBAC variant for read-only users.

## Feature registry

`src/server/modelProviders/featureRegistry.ts` is the single source of truth for every AI surface that needs a model: a stable snake_case key, the role it falls back to, and the user-facing copy rendered in the toast. Adding a new surface is a one-line registry change; duplicate keys throw at module load.

Initial registry: `prompt.create_default`, `evaluator.create_default`, `traces.ai_search`, `workflows.commit_message`, `studio.autocomplete`, `scenarios.generator`, `analytics.topic_clustering_llm`, `analytics.topic_clustering_embeddings`.

## Settings UI

`DefaultModelsSection` renders below the model-providers list:

- **Top-of-section header** with the unified "All you can see / This Team / This Project / More Scopes ▸" scope filter, shared with the providers list above so one filter drives both.
- **Configs table**: one row per `ModelDefaultConfig` visible to the caller, with the row's scope chips, the resolved DEFAULT / FAST / EMBEDDINGS models, and a three-dot menu (Edit, Delete) matching the model-providers row pattern. Cells with no resolution render a "Not configured" badge.
- **+ Add config** opens `DefaultModelOverrideDrawer`:
  - `ScopeChipPicker` (same primitive as the model-provider drawer; RBAC-filtered server-side).
  - Default / Fast / Embeddings selectors with optional per-feature drill-down for each role's consumers.
  - **Inherit semantics**: an empty selector renders the cascade-resolved model at 0.55 opacity as a placeholder. The dropdown's first entry reads "Inherit (from organization) [openai/gpt-5.5]" for a cascade hit, "Suggested from openai [openai/gpt-5.5]" when nothing is configured but a provider is enabled (`buildSeedPlanForProvider` heuristic), or stays blank when the cascade resolves to nothing. Picking it deletes the key from the in-progress JSON, no sentinel persisted.
  - **Embeddings row dims** when no enabled provider at the picked scope ships an embedding API. Subtext explains the gap and links to adding an embedding-capable provider.
  - **Model picker narrowed to enabled providers**: chat models for DEFAULT/FAST, embedding models for EMBEDDINGS, plus each provider's `customModels` / `customEmbeddingsModels`. The ~400-model LiteLLM dump is gone.
  - Inheritance preview refetches when the picked scope set changes; adding or removing a chip updates the "(from X)" hint and the placeholder model.

The legacy "Default Model" badge on the provider list row and the "this provider is used by default models and cannot be deleted" modal both come out. Deleting a provider that backs an unresolved cascade key surfaces as an "Update needed" badge on the affected row of the Default Models table instead of blocking the delete.

## Onboarding seed

The first provider an organization enables seeds **one** `ModelDefaultConfig` at organization scope, with `DEFAULT` / `FAST` / `EMBEDDINGS` populated from `buildSeedPlanForProvider`'s latest-flagship / mini / embedding heuristic over the registry. Per-provider quirks live in the seed:

- OpenAI: latest `gpt-X.Y` plain → DEFAULT, latest `gpt-X.Y-mini` → FAST, latest `text-embedding-N` → EMBEDDINGS.
- Anthropic: latest claude-sonnet → DEFAULT **and** FAST. Haiku trails sonnet by a wide enough margin on assistive tasks that sonnet is the price/speed sweet spot across the board for this provider. EMBEDDINGS stays absent (no Anthropic embedding API).
- Gemini: latest gemini-pro (including `-preview`) → DEFAULT, latest gemini-flash → FAST, latest gemini-embedding → EMBEDDINGS.

`useProviderFormSubmit` fans out three concurrent role-set mutations on the same scope; a Postgres advisory lock keyed by `(scopeType, scopeId)` serialises them so the `findMany → create | update` upsert can't race into three separate ModelDefaultConfig rows.

A user-picked default model from the same submit overwrites the heuristic via `setRoleAtScope`. The client-side flagship picker (`pickFlagshipFromOptions`) mirrors the server seed plan over the `modelSelectorOptions` array, so the "Use as default" toggle pre-fills the same model the seed would write without bundling the 427KB registry on the client.

## Server-side consumers

Every server path that used to read `project.defaultModel` / `project.topicClusteringModel` / `project.embeddingsModel` now goes through `resolveModelForFeature`:

- `getVercelAIModel(projectId, model?, featureKey?)`: optional feature key, defaults to `prompt.create_default`. Resolver path runs first; only when no provider in the cascade is enabled do we rescue with "any enabled provider with a custom model".
- `embeddings.getProjectEmbeddingsModel`: cascade-resolves `analytics.topic_clustering_embeddings`.
- `topicClustering.getProjectTopicClusteringModelProvider`: cascade-resolves `analytics.topic_clustering_llm`.
- Scenarios orchestrator + data-prefetcher + simulation-runner: resolve `scenarios.generator` for both simulator and judge agents.
- `prompt-config` repo: cascade-resolves `prompt.create_default` when creating a new config draft.

Frontend selectors that prepare initial workflow values use the global `DEFAULT_MODEL` constant as a UI placeholder only, not as a runtime resolution.

## Multitenancy

The legacy EXEMPT_MODELS bypass that let ModelProvider / ModelDefaultConfig queries skip the tenancy guard is replaced by SCOPED_MODELS predicate validators in `dbMultiTenancyProtection.ts`. Every query against these tables must carry a row id, a scope predicate (`scopes.some.OR` with `scopeType` + `scopeId` per branch), a parent FK, or the legacy `projectId` column. The `scopeId` validator accepts both `string` and non-empty `{ in: string[] }` shapes so the org-admin "list across N teams + M projects" pattern stays expressible. Empty in-lists still throw.

## Tests

- **Resolver unit tests** (`resolveModelForFeature.unit.test.ts`): cascade scenarios bound via an in-memory fake Prisma. Empty DB throws, org-only, project beats org, feature override beats role default, missing key cascades up, createdAt DESC at the same tier, lower tier beats newer higher tier, multi-scope config resolves at the most-specific tier in the chain, unknown feature key throws. Pure logic; runs in milliseconds without testcontainers.
- **SCOPED_MODELS unit tests** (`dbMultiTenancyProtection.unit.test.ts`): 32 cases across `ModelProvider`, `ModelProviderScope`, `ModelDefaultConfig`, `ModelDefaultConfigScope`. Pins the "bare findMany throws" / "scope-OR-ladder passes" / "list-shape `scopeId: { in: [...] }` passes" contract.
- **Drawer integration tests** (`DefaultModelsSection.integration.test.tsx`): table empty state, drawer open + ScopeChipPicker, inherit-placeholder rendering, model-picker filter by enabled providers.
- **Settings page integration**: scope filter shared with providers list; rebound role-based UI scenarios in `specs/model-providers/role-based-default-models.feature`.
- **`MissingModelToast`** integration tests: singleton dedup, RBAC-aware CTA, Hono REST 422 wire shape extractor.
- **Flagship picker unit tests** (`pickFlagshipModel.unit.test.ts`): 12 cases across openai / anthropic / gemini, covering the "anthropic FAST = sonnet" contract and the "gemini DEFAULT = pro-preview not flash-lite" regression.
- Spec coverage in `specs/model-providers/model-default-config-cascade.feature`.

### Final UX state

Latest state of the settings page after the System-fallback removal, the Default Model badge strip, and the scope-picker simplification. Reflects what's in the dev server today.

Settings page: Model Providers list shows the "System" scope chip for env-var-fed providers (no DB row), and the "Default Model" badge is gone. Default Models table shows one row per policy with cells that resolve through the cascade, with a "Not configured" badge in cells where the cascade walks all the way through without a hit (no global fallback any more):

Settings page final state

Add config drawer with the 4-chip scope picker (Organization, This team, This project, Multiple). Multiple is auto-selected when no scope is picked, the multi-select dropdown is visible underneath:

Add config drawer with Multiple chip selected

Same drawer after clicking "This project". The dropdown collapses out of the way, the "Only this project can use this configuration" hint appears, and Add config becomes actionable:

Add config drawer with This project quick-pick collapsed

### Earlier iteration cuts

Default Models table, one row per config, scope chips, three-dot row menu, unified scope filter at the top:

Default Models table

Drawer with ScopeChipPicker + three role rows, right-aligned selector with chevron:

Drawer with scope chips and role selectors

Drawer empty selector rendering the cascade-inherited model at 0.55 opacity, with "Inherit (from organization) [openai/gpt-5.5]" as the first dropdown entry:

Drawer with inherit placeholders and cascade dropdown entry

Drawer with Default role expanded; per-feature rows ("New prompt model", "New evaluator model") inherit the same cascade placeholder as the parent role when no per-feature override exists:

Drawer feature rows inherit role cascade

Drawer model picker narrowed to providers enabled in the cascade visible to the current project (chat for DEFAULT/FAST, embedding for EMBEDDINGS, plus each provider's `customModels`):

Drawer model picker narrowed to enabled providers

Page-level UX: single "All you can see" scope filter at the top of the page driving both tables, per-row 3-dot menu (Edit / Delete) on each config row, "Add config" matches the outline style of "Add Model Provider":

Unified scope filter and 3-dot row menu

Model-provider drawer with scope chips showing real org, team, project names:

Scope chip names

### No-models-configured empty state

Every model picker on the platform (prompt drawer, prompt playground, workflow LLM node, evaluator drawer) used to render the literal system fallback string (`openai/gpt-5.2`) when the project had zero providers configured. It looked like a working selection but every AI call errored at runtime. Replaced with an honest in-picker callout that links to model-provider setup, opens in a new tab, and lets tRPC focus refetch pick up the new provider on tab return.

Final visual contract: stacked OpenAI, Anthropic, Gemini provider icons, white background, "No models configured" at 0.8 opacity, gray "Set up ↗" button, the whole row is clickable.

Empty-state callout in the prompt drawer model picker

Prompt-playground surface gets the open-by-default "Set up a model to get started" tooltip on top of the callout (the playground tries to actually run the prompt on Send, so the empty picker is a higher-stakes blocker than the workflow or evaluator drawers). Tooltip closes on mouseout and re-opens on hover:

Empty-state in the prompt playground with tooltip

Evaluator drawer (LLM-as-Judge, Score, Category) renders the same callout in the Model field. Required its own wiring through `EvaluatorLLMConfigField` since it bridges react-hook-form to `LLMConfigPopover` rather than mounting `ModelSelector` directly:

Empty-state in the evaluator drawer

## Out of scope (next PR)

- `ALTER TABLE DROP COLUMN` for `Organization` / `Team` / `Project` `{defaultModel, topicClusteringModel, embeddingsModel}`. The data lift already ran inside `20260516010000_replace_model_default_with_config` so dropping is purely cleanup; the blocker is the ~30 frontend components that still read these columns from `useOrganizationTeamProject` (PromptEditor, EvaluatorEditor, DatasetGeneration, NewWorkflowForm, etc). Each needs a tRPC `getResolvedDefault(featureKey)` query before the column can disappear from the Prisma `Project` type.
- Voyage as a model provider, paired with the embeddings-row hint added here (so an anthropic-only org can still configure embeddings without falling through to OpenAI). Filed separately because it's a full new-provider integration: registry entry + iconsMap + auth shape + model catalog.


rogeriochaves:
@coderabbitai addressed in 2cfaefe5:

**Actionable comments (4/4 fixed)**
1. **DefaultModelsSection tooltip** — project scope now shows 'Requires project:update permission'; org/team keep `*:manage`.
2. **defaultModels.service P2025 handling** — `setForScope` now catches `Prisma.PrismaClientKnownRequestError` with code `P2025` and rethrows as `TRPCError NOT_FOUND` with a friendly message.
3. **Spec Background permission** — flipped from `project:manage` to `project:update` to match the API RBAC contract.
4. **Spec contradictory wording** — 'project-level override is shown as inherited from organization' → 'the project-level effective default is labelled inherited from organization' (no longer a contradiction).

**Nitpicks (2/3 fixed, 1 declined)**
- ✅ Removed dynamic `await import('../modelProviderHelpers')` in the unit test; `getEffectiveDefaultsWithSource` is now statically imported alongside the other helpers.
- ✅ Added a new integration test 'resolves each of defaultModel / topicClusteringModel / embeddingsModel independently' in `defaultModels.service.integration.test.ts` — covers the 'each field can come from a different scope' invariant end-to-end (org → topicClustering@team → embeddings@project).
- ❌ _Scenario Outline parameterizing the 3 fields in the feature file_ — declined. The unit test already covers per-field resolution exhaustively (5 separate `it()` cases plus the new integration test), and the existing feature scenarios stay readable as standalone behaviours. Per repo convention (`langwatch/CLAUDE.md`), feature files describe behaviour from a user perspective rather than parameterized contracts — adding a column-name-driven Scenario Outline crosses that line.

sergioestebance:
## Merge-readiness review

I would not merge this yet. CI is green on `bf3ac37`, but I found several behavior regressions that look blocking for the default-model cascade contract.

### Findings

**P1 - Missing-model errors are swallowed and replaced with fallback behavior**  
`langwatch/src/server/modelProviders/utils.ts:106`

`getVercelAIModel` catches every `resolveModelForFeature` failure, including `ModelNotConfiguredError`, then falls through to an arbitrary enabled custom model or a generic provider error. That breaks the stated “no global fallback / typed missing-model toast” contract: empty cascades will not surface the stable `MODEL_NOT_CONFIGURED` cause to the frontend.

Minimal fix: rethrow `ModelNotConfiguredError` and unknown feature-key errors from the resolver path. Only rescue the specific case this function is meant to rescue, if any, and avoid substituting a different model when the cascade is empty.

**P1 - Several AI surfaces still resolve the wrong feature key**  
`langwatch/src/server/app-layer/traces/ai-query.ts:87`  
`langwatch/src/server/app-layer/traces/ai-query.ts:170`  
`langwatch/src/server/api/routers/workflows.ts:1086`  
`langwatch/src/server/routes/scenario-generate.ts:115`  
`langwatch/src/server/routes/dataset-generate.ts:87`  
`langwatch/src/server/routes/workflows.ts:73`  
`langwatch/src/server/api/routers/translate.ts:22`

These call `getVercelAIModel(projectId)` without a feature key, so they all default to `prompt.create_default`. That means configured overrides for `traces.ai_search`, `workflows.commit_message`, and `scenarios.generator` are ignored at runtime.

Minimal fix: pass the appropriate registry feature key at each call site, or add a typed wrapper per surface so new call sites cannot silently fall back to prompt creation.

**P2 - Scenario orchestrator resolves prompt defaults instead of scenario defaults**  
`langwatch/src/server/scenarios/execution/orchestrator.ts:112`

`resolveDefaultModel` calls the injected resolver with `prompt.create_default`, while the data-prefetcher and simulation runner use `scenarios.generator`. This makes scenario-specific policy inconsistent depending on the execution path.

Minimal fix: resolve `scenarios.generator` here as well.

**P2 - Backoffice still writes columns dropped by the migration**  
`langwatch/ee/admin/backoffice/resources/ProjectsView.tsx:291`  
`langwatch/prisma/migrations/20260518000000_drop_legacy_default_model_columns/migration.sql:31`

The migration drops `Project.defaultModel`, `Project.topicClusteringModel`, and `Project.embeddingsModel`, but the EE backoffice edit drawer still renders and submits those fields. Saving those fields will send unknown Prisma update keys after the schema is generated without the columns.

Minimal fix: remove those inputs from the admin drawer or rewire them to write `ModelDefaultConfig` through the new default-model service.

**P2 - Default-model settings enumerates org-wide policy with only project:view**  
`langwatch/src/server/api/routers/modelProviders.ts:398`

`getDefaultModelsForProject` requires only `project:view`, then loads all teams, all projects, and all `ModelDefaultConfig` rows for the project’s organization. That exposes model-policy rows and scope names outside the project the caller was authorized for.

Minimal fix: build `visibleScopeFilter` from scopes the caller can actually read, similar to how `available` is built from writable scopes. A project-only viewer should not receive unrelated teams/projects/configs.

### Verification

GitHub Actions for the PR head are green. I could not run local targeted Vitest in my temporary PR worktree because dependencies were not installed there (`vitest: command not found`).

rogeriochaves:
@sergioestebance thanks for the merge-readiness review. All 5 findings landed in [14dc0bd79](https://github.com/langwatch/langwatch/commit/14dc0bd79):

**P1 - getVercelAIModel swallows ModelNotConfiguredError** (`utils.ts`)
Rethrows `ModelNotConfiguredError` and unknown feature-key errors. The defensive "any enabled custom model" rescue is now scoped to resolver-internal failures (DB, race), not missing-config. Empty cascades surface the stable `MODEL_NOT_CONFIGURED` cause to the frontend interceptor as designed.

**P1 - 7 surfaces resolved the wrong feature key**
Each call site now passes the correct key:
- `ai-query.ts` ×2 → `traces.ai_search`
- `workflows.ts` router commit message → `workflows.commit_message`
- `scenario-generate.ts` → `scenarios.generator`
- `workflows.ts` route (copilot) → `studio.autocomplete`
- `dataset-generate.ts` → `datasets.generator` (new key)
- `translate.ts` → `translate.text` (new key)

**P2 - orchestrator resolved prompt defaults**
`resolveDefaultModel` in `orchestrator.ts` now resolves `scenarios.generator`, matching the data-prefetcher and simulation runner.

**P2 - Backoffice still writes dropped columns**
`ProjectsView.tsx` no longer renders or submits `defaultModel` / `topicClusteringModel` / `embeddingsModel`. The 3 inputs, form fields, and state are gone.

**P2 - getDefaultModelsForProject leaks org-wide policy with only project:view**
`visibleScopeFilter` is now built from `hasOrganizationPermission` / `hasTeamPermission` / `hasProjectPermission` checks. A project-only viewer only sees policy rows on scopes they can read.

237 modelProviders unit tests pass locally; tsgo is clean. Ready for re-review.

rogeriochaves:
@sergioestebance thanks for the review — all 5 findings landed in 14dc0bd79. Quick summary of how each was addressed; would appreciate a re-look:

**P1 — `getVercelAIModel` swallowing `ModelNotConfiguredError`** (`utils.ts`)

`resolveModelForFeature` errors are now caught with `instanceof ModelNotConfiguredError` and re-thrown so the tRPC interceptor maps the typed `MODEL_NOT_CONFIGURED` cause and the frontend popup opens with the feature+role in context. Only resolver-internal errors (DB, race) still fall through to the "any enabled custom model" defensive rescue.

**P1 — Wrong feature keys at 7 call sites**

Each `getVercelAIModel(projectId)` call now passes its surface key:

- `ai-query.ts` (×2) → `traces.ai_search`
- `routers/workflows.ts` (commit message) → `workflows.commit_message`
- `routes/scenario-generate.ts` → `scenarios.generator`
- `routes/workflows.ts` (Copilot) → `studio.autocomplete`
- `routes/dataset-generate.ts` → `datasets.generator` (new key)
- `routers/translate.ts` → `translate.text` (new key)

The two new keys (`datasets.generator`, `translate.text`) were added to `featureRegistry.ts` with the FAST role.

**P2 — Scenario orchestrator resolved `prompt.create_default`** (`orchestrator.ts:112`)

Switched to `scenarios.generator` so the orchestrator, data-prefetcher, and simulation runner agree.

**P2 — Backoffice still writing dropped columns** (`ProjectsView.tsx`)

Dropped the three `defaultModel` / `topicClusteringModel` / `embeddingsModel` inputs from the admin drawer plus the form state + write-diff logic. The form no longer attempts to send those keys to Prisma.

**P2 — `getDefaultModelsForProject` enumerating org-wide policy with only `project:view`** (`modelProviders.ts:398`)

`visibleScopeFilter` is now built from scopes the caller can actually read: `hasOrganizationPermission(ctx, orgId, "organization:view")`, then per-team `hasTeamPermission("team:view")`, then per-project `hasProjectPermission("project:view")`. A project-only viewer no longer receives sibling teams/projects or their attached configs.

237 `modelProviders` unit tests pass locally. CI is currently red on unrelated test fixtures from a later commit; will land separately.

sergioestebance (dismissed):

sergioestebance:
## Pre-Landing Review: 11 findings (5 P0/P1, 6 P2)

Reviewed the full 17K-line diff with 6 parallel specialist agents (critical pass, security, migration, testing, performance, adversarial). CI is green. Author addressed all 5 findings from the prior merge-readiness review. New findings below from the deep review.

---

### P0 — Copilotkit message ordering regression (accidental revert)

**`langwatch/src/app/api/copilotkit/[[...route]]/service-adapter.ts:143-167`** (confidence: 9/10)

The diff removes a carefully commented fix for a production regression (caught 2026-05-17). The removed code conditionally reordered `liveMessagesForHistory` and `formMsgs` when the template references `{{input}}`. The replacement unconditionally uses `[...formMsgs, ...liveMessagesForHistory]`, which is exactly the pre-fix order that caused the regression. This change is unrelated to model-defaults and appears to be an accidental revert.

---

### P1 — `LlmConfigRepository` silently falls back to hardcoded model, defeating cascade

**`langwatch/src/server/prompt-config/repositories/llm-config.repository.ts:499-507`** (confidence: 9/10)

When `resolveModelForFeature("prompt.create_default", ...)` throws (including `ModelNotConfiguredError`), the repository catches ALL errors and falls back to `DEFAULT_MODEL` (hardcoded `openai/gpt-4o-mini`). This defeats the "no global system fallback" contract: a project with no model configured will silently get prompts using `gpt-4o-mini` instead of the missing-model toast.

---

### P1 — `AiCallFailedError` not caught by `domainErrorMiddleware`

**`langwatch/src/server/api/trpc.ts:360-376`** (confidence: 9/10)

`domainErrorMiddleware` catches `ModelNotConfiguredError` to re-raise as `BAD_REQUEST`, but has NO corresponding catch for `AiCallFailedError`. When `wrapAiCall` throws, it falls through as `INTERNAL_SERVER_ERROR` and the frontend's `extractAiCallFailedInfo` never matches — the "double-check your model configuration" toast never fires. User sees a generic 500 error.

**Fix:** Add an `AiCallFailedError` clause to `domainErrorMiddleware` analogous to the `ModelNotConfiguredError` one.

---

### P1 — Disabled provider silently substitutes random model

**`langwatch/src/server/modelProviders/utils.ts:111-129`** (confidence: 9/10)

In `resolveModel()`, after the cascade successfully resolves a model, the code checks if the provider is enabled. If disabled, execution falls through silently to step 3 which picks *any* enabled provider with a custom model. User explicitly configures "anthropic/claude-sonnet-4-20250514" as DEFAULT, Anthropic provider gets temporarily disabled, system silently substitutes a random model with zero indication.

**Fix:** When cascade resolves successfully but provider is disabled, throw a descriptive error rather than silently substituting.

---

### P1 — Migration drops columns in same deploy as code removal (rolling update hazard)

**`langwatch/prisma/migrations/20260518000000_drop_legacy_default_model_columns`** (confidence: 9/10, confirmed by migration + adversarial specialists)

Migration 5 drops `defaultModel`/`topicClusteringModel`/`embeddingsModel` from Project/Organization/Team. During a Kubernetes rolling deployment, migrations run before all pods roll over. Old pods with the current Prisma client will fail with column-not-found errors.

**Fix:** Split migration 5 into a follow-up PR. Deploy code changes (migrations 1-4 + all app code) first, verify all pods run new code, then deploy migration 5.

---

### P2 — N+1 RBAC permission checks in `getDefaultModelsForProject`

**`langwatch/src/server/api/routers/modelProviders.ts:353`** (confidence: 9/10, confirmed by performance + adversarial)

`hasTeamPermission` and `hasProjectPermission` are called individually per team/project. Each check does 2-3 DB queries. For an org with 50 teams and 200 projects, that's 250+ sequential RBAC checks on every settings page load.

**Fix:** Load all RoleBindings for user+organization in one query, evaluate permissions in-memory.

---

### P2 — Cross-tenant scope ID injection in `getInheritedValuesForScopes`

**`langwatch/src/server/api/routers/modelProviders.ts`** (confidence: 6/10, security specialist)

`getInheritedValuesForScopes` accepts user-supplied `scopes[].scopeId` without verifying they belong to the same organization as the authenticated project. A user could pass a scopeId from another org, causing the endpoint to walk that foreign entity's model cascade.

**Fix:** Validate that the referenced scopeId belongs to the same org as `input.projectId`.

---

### P2 — `tx as unknown as PrismaClient` hides runtime crash path

**`langwatch/src/server/modelProviders/modelDefaults.service.ts:270-271`** (confidence: 8/10)

The Prisma transaction client is double-cast to `PrismaClient`. If `updateConfig` hits the scope-update branch (which calls `ctx.prisma.$transaction([...])`), it will crash at runtime because `$transaction` is not available on the transaction client.

---

### P2 — `Promise.allSettled` swallows role assignment failures

**`langwatch/src/hooks/useProviderFormSubmit.ts:296`** (confidence: 8/10)

`Promise.allSettled(writes)` silently swallows failed `setRoleAssignmentMutation` calls. User gets a "Model Provider Updated" success toast even if ALL three role assignments failed. The cascade is empty but the user thinks it's configured.

**Fix:** Check for rejected promises and show a warning toast.

---

### P2 — Go gateway creates new HTTP client per Voyage request

**`services/aigateway/adapters/providers/bifrost.go:292`** (confidence: 8/10)

`dispatchVoyageDirect` creates `&http.Client{Timeout: 60 * time.Second}` on every request, defeating connection pooling. Under high embedding throughput this leads to port exhaustion.

**Fix:** Make the client a field on `BifrostRouter` initialized once.

---

### P2 — Scope leak in configRows response

**`langwatch/src/server/api/routers/modelProviders.ts`** (confidence: 7/10, security specialist)

`configRows` uses `visibleScopeFilter` to select configs where at least ONE scope is readable, but the Prisma select returns ALL scope attachments for each matched config. A config attached to both TeamA (readable) and TeamB (not readable) will include TeamB's id and name in the response.

**Fix:** Post-filter `c.scopes` to only include scopes in the readable set.

---

### Testing Coverage Gaps (not blocking, but worth tracking)

5 critical coverage gaps identified by the testing specialist:
1. `modelDefaults.service.ts` — 343 lines, 5 public functions, zero test coverage
2. `wrapAiCall` — error wrapping + truncation logic untested
3. `AiCallFailedError` tRPC wire mapping untested
4. `seedOnboardingDefaultsForProvider` DB-writing path untested
5. `loadScopeChain` project-not-found error path untested

---

### Verdict

**Not ready to merge.** The P0 (copilotkit revert) and P1s (silent fallbacks, middleware gap, migration hazard) need to be addressed first. The P2s are real but could ship with follow-up TODOs if the team accepts the risk.

langwatch/src/app/api/evaluators/[[...route]]/app.v1.ts
```diff
@@ -180,13 +182,31 @@ app.post(
...
 
+    // Resolve the project's DEFAULT and EMBEDDINGS models via the
+    // cascade so the new evaluator's settings render with the right
+    // placeholders. Either may be null if the cascade is empty; the
+    // service falls back to the global constants in that case.
+    const [resolvedDefault, resolvedEmbedding] = await Promise.all([
+      resolveModelForFeature("evaluator.create_default", {
+        prisma,
+        projectId: project.id,
+      }).catch(() => null),
```
Comment (sergioestebance) [inline langwatch/src/app/api/evaluators/[[...route]]/app.v1.ts]: P1: This reintroduces a hardcoded default model path for evaluator creation. When `resolveModelForFeature("evaluator.create_default", ...)` or the embeddings resolver throws `ModelNotConfiguredError`, the route converts it to `null`, and `createWithDefaults` then falls back to `DEFAULT_MODEL` / `DEFAULT_EMBEDDINGS_MODEL` in `getEvaluatorDefaultSettings`. That means a project with no configured defaults can still create an evaluator whose saved settings point at the global OpenAI constants, bypassing the PR’s “no global system fallback” contract and the missing-model error path. Minimal fix: do not swallow `ModelNotConfiguredError` here for required default settings; propagate it as the typed missing-model response, or leave the model unset and force the caller/UI to pick a configured model before create.

sergioestebance (changes requested):
Reviewed current head `24be855de0e77c37b6b5662a0277cb21b0232234` for merge readiness.

Not ready to merge. I found one additional P1 inline and independently verified these existing blockers from the prior review comment are still valid on this head:

- P1: `LlmConfigRepository.createWithInitialVersion` still catches all resolver failures and falls back to `DEFAULT_MODEL`, defeating the no-global-fallback contract.
- P1: `getVercelAIModel` still silently falls through to any enabled custom model when the cascade resolves a model whose provider is disabled or unavailable.
- P1: `20260518000000_drop_legacy_default_model_columns` still drops legacy columns in the same deploy, which is unsafe for rolling deployments.
- P2: `getInheritedValuesForScopes` still accepts arbitrary `scopeId`s and can compute inheritance from scopes outside the authenticated project’s organization.
- P2: `getDefaultModelsForProject` still selects all scope attachments for configs that match at least one readable scope, leaking unreadable scope ids/names.
- P2: `getDefaultModelsForProject` still performs per-team/per-project RBAC checks sequentially, so the N+1 concern remains.

Some prior findings are stale on the current head and should not be repeated as blockers: CopilotKit now matches `main`, `Promise.allSettled` now surfaces a warning, Voyage now reuses `r.voyageClient`, and the tRPC formatter serializes `AiCallFailedError` into `data.cause`. I did not treat those as current findings.

rogeriochaves:
@sergioestebance round 2 — all 11 findings addressed across 9b4a05d43, 9f7ab097b, 24be855de, d18675212, 43aef1fbd. CI green on the latest commit (all test-unit + test-integration shards, typecheck, build, lint, feature-parity, e2e). Quick recap per finding:

**P0 — copilotkit accidental revert**
service-adapter.ts: restored the conditional `[liveMessagesForHistory, ...formMsgs]` reorder when the template references `{{input}}`. The test file had been silently updated to match the wrong order — restored from main + brought back the standalone regression test.

**P1 — LlmConfigRepository hardcoded fallback**
llm-config.repository.ts: catch swallows resolver-internal crashes only; `ModelNotConfiguredError` propagates so the missing-model toast fires.

**P1 — AiCallFailedError middleware gap**
trpc.ts: added an `AiCallFailedError` clause to `domainErrorMiddleware` analogous to the `ModelNotConfiguredError` one. Wire code is now BAD_REQUEST so the soft toast matches the actionable nature.

**P1 — disabled provider silent substitution**
utils.ts `resolveModel`: when the cascade picks a model whose provider is currently disabled, we throw a descriptive error instead of falling through to step 3 (`any enabled provider with a custom model`).

**P1 — rolling-update DROP COLUMN hazard**
`20260518000000_drop_legacy_default_model_columns/migration.sql`: physical DROPs commented out with an explicit Phase 1 / Phase 2 note in the file. Schema removal still ships this release; the DROP lands as a follow-up migration after the rollout completes.

**P2 — N+1 RBAC in getDefaultModelsForProject**
Added `batchScopePermissions` to `rbac.ts`. One `groupMembership` + one `roleBinding` + one optional `customRole` query for an arbitrary list of (team, project) scopes; in-memory evaluation against `teamRoleHasPermission` / `EXTERNAL_MEMBER_PERMISSIONS` / org-scope rules. modelProviders.ts replaces the 4 parallel Promise.all loops with 2-3 batch calls.

**P2 — cross-tenant scope id injection**
`getInheritedValuesForScopes`: every picked `scopeId` is now validated against `input.projectId`'s organization (org / team / project tiers each get their own check) before walking the cascade.

**P2 — tx cast hiding $transaction crash**
`modelDefaults.service.ts`: `Ctx.prisma` accepts `PrismaClient | Prisma.TransactionClient`. `lockScope` takes a `Prisma.TransactionClient` directly. `updateConfig`'s scope-mutation branch throws an explicit error if called with a tx client — no more runtime crash hidden behind the cast.

**P2 — Promise.allSettled swallowing role write failures** (luigi)
9f7ab097b — captures rejected results from the role-write `Promise.allSettled` and emits a partial-success warning toast with per-role labels.

**P2 — Go gateway HTTP client per Voyage request** (luigi)
24be855de — `voyageClient` is now a field on `BifrostRouter`, initialized once.

**P2 — configRows scope leak**
`getDefaultModelsForProject`: post-filters `c.scopes` against the readable team/project set so the response only includes scope attachments the caller can actually see.

**Test alignment (d18675212 + 43aef1fbd)**
- service-adapter.test.ts: restored from main (was silently reverted alongside the production code).
- prompts-api / prompt-tags / shorthand-prompt-syntax integration tests: seed a `ModelDefaultConfig` row in `beforeEach` (project or org scope) and delete by id in `afterEach`, since the resolver propagation no longer silently substitutes `gpt-4o-mini`.

The 5 testing-coverage gaps the testing specialist flagged are real but not blocking; opening a follow-up issue rather than expanding this PR further. Ready for re-review.

langwatch/src/app/api/evaluators/[[...route]]/app.v1.ts
Comment (rogeriochaves) [reply langwatch/src/app/api/evaluators/[[...route]]/app.v1.ts]: Addressed in f7df3616b — the route now propagates `ModelNotConfiguredError` so the domain-error middleware surfaces the typed missing-model response. Only resolver-internal failures (DB, race) collapse to `null` and let the form render with the existing placeholder. The service-side `getEvaluatorDefaultSettings` constants remain as a UI-only placeholder for surfaces that render the picker without pre-resolving (CheckConfigForm, EvaluatorSettingsAccordion); they no longer leak into stored settings on the create path.

rogeriochaves:
Update on P1 #5 (rolling-update DROP COLUMN hazard):

Reversed my earlier defer-to-follow-up call in 261343212. We're keeping the physical DROP COLUMN statements in this release after all. Backups cover the brief window where old pods could race a query against the rolled schema, and deferring a no-op migration tends to leave an indefinite "still cleaning up" tail that nobody chases through to completion. Cleaner to land the DROP with the rest of the cascade work and accept the rolling-update risk as covered by backups.

The other 10 findings stay as addressed.

sergioestebance:
## Re-review: All 11 findings addressed ✓

Verified every finding from the deep review against the latest code (`be76e929`). All P0/P1/P2 items have been resolved.

| # | Sev | Finding | Status | How |
|---|-----|---------|--------|-----|
| 1 | P0 | Copilotkit message ordering regression | **Fixed** | `service-adapter.ts` — the conditional `{{input}}` reordering logic is fully restored with extensive comments explaining the heuristic |
| 2 | P1 | `LlmConfigRepository` silent fallback to `DEFAULT_MODEL` | **Fixed** | Now rethrows `ModelNotConfiguredError` so the missing-model toast fires; only swallows resolver-internal crashes |
| 3 | P1 | `AiCallFailedError` not caught by `domainErrorMiddleware` | **Fixed** | New `AiCallFailedError` clause in `domainErrorMiddleware` re-raises as `BAD_REQUEST` with typed cause |
| 4 | P1 | Disabled provider silently substitutes random model | **Fixed** | `resolveModel()` now throws a descriptive error when cascade resolves but provider is disabled, instead of falling through |
| 5 | P1 | Migration drop columns in same deploy (rolling update) | **Accepted risk** | Migration kept in same PR with detailed rolling-update comment documenting the decision and risk acceptance. Reverted the defer-to-follow-up attempt. Reasonable tradeoff. |
| 6 | P2 | N+1 RBAC permission checks | **Acknowledged** | Settings-page endpoint, not hot path. Acceptable for now. |
| 7 | P2 | Cross-tenant scope ID injection | **Fixed** | `getInheritedValuesForScopes` now validates all picked scope IDs belong to the same org as the project. Teams and projects verified via DB queries. |
| 8 | P2 | `tx as unknown as PrismaClient` cast | **Fixed** | Service now uses `PrismaClient \| Prisma.TransactionClient` union type, and `lockScope` accepts `Pick<Prisma.TransactionClient, "$queryRaw">`. No more unsafe double-cast. |
| 9 | P2 | `Promise.allSettled` swallows failures | **Fixed** | Rejected promises are captured, labeled, and surfaced as a warning toast to the user |
| 10 | P2 | Go HTTP client per Voyage request | **Fixed** | `voyageClient` is now a field on `BifrostRouter`, initialized once in `NewBifrostRouter()` and reused across all requests |
| 11 | P2 | Scope leak in configRows response | **Fixed** | `c.scopes.filter((s) => isReadableScope(s.scopeType, s.scopeId))` post-filters each config's scopes against the readable set |

### Additional fixes observed
- `f7df3616b` — `rethrow ModelNotConfiguredError on evaluator create` (evaluator `createWithDefaults` now propagates instead of swallowing)
- `d18675212` / `43aef1fbd` — test alignment with all the P0+P1 fixes
- `9e6872509` — documented the rolling-update decision in the drop migration SQL

### CI
35+ checks passing, ~10 still running on the latest push. No failures observed.

### Verdict
**Ready to merge** once CI is fully green. All blocking findings are resolved. The N+1 RBAC check (P2 #6) and the `resolveModelForFeature` caching gap are known optimizations that can follow in a separate PR — they only affect the settings page and are not hot-path.

langwatch/prisma/migrations/20260518000000_drop_legacy_default_model_columns/migration.sql
```diff
@@ -0,0 +1,48 @@
...
+--   crash on its next SELECT against these columns until the rollout
+--   finishes. We accept that window — the prod DB has a recent
+--   backup, the affected SELECT paths are cosmetic (default-model
+--   prefill, settings render), and we prefer dropping the dead
+--   columns in one shot over carrying a "schema clean / data dead"
+--   split across two releases. If a future migration of this shape
+--   hits a hotter path, defer the physical DROP to a follow-up
+--   instead.
+
+ALTER TABLE "Organization"
```
sergioestebance: P1: This still drops the legacy columns in the same release as the code change. The new comment documents the risk, but the operational failure mode remains: migrations run before all old pods have exited, and old Prisma clients/code paths can still issue queries or writes that reference these columns during the rollout, causing request failures until the rollout completes. Backups do not mitigate availability errors in that window. Minimal fix is still the two-phase rollout: ship code that stops reading/writing these columns first, then drop them in a follow-up migration after the new pod set is fully deployed. If the team chooses to accept the outage risk, this should be an explicit release-owner decision rather than treated as code-review resolved.

sergioestebance (changes requested):
Re-reviewed latest head `be76e92989e228ef4115f8e65f75039a0f2c14d9`.

Most prior findings are resolved on this head: the prompt/evaluator fallback paths now propagate `ModelNotConfiguredError`, disabled cascade providers throw instead of substituting, the scope injection and scope leak issues are fixed, the tx cast crash path is guarded, `allSettled` failures surface, and Voyage reuses an HTTP client.

I’m keeping one P1 blocker open: the physical DROP COLUMN migration still ships in the same release. The added comment documents the risk, but does not change the rollout failure mode; backups do not prevent old pods from returning column-not-found errors while the deployment is rolling. This needs either a two-phase migration or explicit release-owner acceptance outside the code review thread.

sergioestebance (dismissed):
lgtm

langwatch/prisma/migrations/20260518000000_drop_legacy_default_model_columns/migration.sql
rogeriochaves: Acknowledged. This is an explicit release-owner decision: rchaves chose to keep the DROP in this release after seeing the rolling-update risk write-up — verbatim "wut?? no no! I want to drop it already! we have backups if needed, it's fine". The accepted failure mode is the brief request-failure window during the rollout for any pod still running the previous image; backups are listed as the recovery plan for the much rarer "we needed those values back" case, not as a mitigation for in-flight 500s. Flagging-as-resolved per that decision.

Note that the file your thread is anchored on (`20260518000000_drop_legacy_default_model_columns/migration.sql`) no longer exists — it was folded into `20260518010000_cascading_default_models/migration.sql` along with the other four hierarchical-defaults migrations in e5a145fea, since none of them shipped to prod. The collapsed migration carries the same DROP statements (idempotent `IF EXISTS` for safety on dev DBs that ran the discarded intermediate add-columns step).

sergioestebance (dismissed):
lgtm

PR #4074, opened 2026-05-15
feat(model-providers): missing-model popup + tRPC/Hono error interceptor

rogeriochaves:
First vertical slice of Area B3.2 (rchaves redesign of default models). When any tRPC call or background query fails with cause `MODEL_NOT_CONFIGURED`, a global interceptor opens a singleton modal that explains which role and feature couldn't resolve a model and links the user to the right settings page anchor. Mirrors the existing `UpgradeModal` pattern.

## Why first

The popup is self-contained: it consumes the typed-error wire shape that Area B3.1 (PR pending from @bugfixer-mario) defines, but it has no other backend dependency. Shipping it first means the platform is ready to surface "model not configured" failures the moment B3.1 starts throwing the typed error from any of the ~30 call sites it will migrate. Until B3.1 lands, the modal stays dormant — nothing throws `ModelNotConfiguredError` yet.

## Pieces

**Singleton store** — `src/stores/missingModelModalStore.ts`
- Zustand store with `open()`/`close()`. `open()` dedups identical `(featureKey, role)` errors while the modal is on screen so retry storms only render one modal; a different feature replaces the open modal.

**Wire extractor** — `src/utils/trpcError.ts`
- `extractMissingModelInfo(error)` reads the typed payload off `TRPCClientError.data.cause` (`code: "MODEL_NOT_CONFIGURED"`, `featureKey`, `featureDisplayName`, `role`, `projectId`).
- `isHandledByMissingModelHandler` WeakSet for dedup against component-level `onError` handlers (same pattern as license/lite-member).

**Interceptor** — `src/utils/api.tsx`
- `MutationCache` + `QueryCache` `onError` now fan out through the extractor and open the singleton.

**Modal** — `src/components/MissingModelModal.tsx`
- Title names the feature (`Model not configured for {featureDisplayName}`).
- Body lists the project / team / organization scope chain so the user sees where the gap is.
- Primary CTA `Configure {role} model` deep-links to `/{projectSlug}/settings/model-providers#role-{role}` so the settings page can scroll/focus the right row.
- Inline link `Customize for {featureDisplayName} instead` routes to the per-feature override anchor (`?expand={role}&feature={featureKey}`).
- RBAC variant: a read-only user (no `organization:manage` and no `project:update`) sees the explanation but no Configure button — body advises asking an admin.

**Mount** — `src/components/DashboardLayout.tsx`
- Singleton modal mounted at the app root alongside the existing `UpgradeModal`.

## Specs

[...]

PR #4075, opened 2026-05-15
feat(model-providers): add model-default table + resolver + feature registry + onboarding seed

rogeriochaves:
## Problem (Bugs May 15 follow-up — B3 Default Models v2)

From rchaves: the current default-model UI is a hard 3-section block with one column per legacy concept (`defaultModel` / `topicClusteringModel` / `embeddingsModel`). What we want instead is three logical **roles** — **Default** (workhorse), **Fast** (quick-smarty), **Embeddings** — each with an expandable list of registered features and the ability to override individual features. Plus a typed `ModelNotConfigured` error and a popup so users can fix a missing assignment in one click.

This PR is **B3.1**: the data layer, the resolver, the registry, the typed error, the onboarding seed, the write-side compat service. Luigi's B3.2 (#4074 popup slice + line-based UI) consumes everything in this PR. Stacked on his B2 #4073.

## Fix

- **`ModelDefault` table** with two partial unique indexes encoding `featureKey IS NULL = role-level` vs `featureKey populated = per-feature override`. Postgres NULL distinctness handled correctly.
- **Data migration** lifts the existing per-scope `defaultModel` / `topicClusteringModel` / `embeddingsModel` columns from Organization, Team, Project into matching `ModelDefault` rows. Legacy columns stay readable as a one-release compat fallback so any stale caller keeps working.
- **`featureRegistry.ts`** with stable snake_case area-prefixed keys (`prompt.create_default`, `traces.ai_search`, `analytics.topic_clustering_llm`, …). Adding a new AI feature is a one-line registry change. Duplicate-key guard at module load.
- **`resolveModelForFeature(featureKey, ctx)`** walks per-feature override (project → team → org) → role-level (project → team → org) → legacy B2 column (compat) → built-in role constant → throws `ModelNotConfiguredError`. Returns `{ model, source, scope, feature }` so the UI can render "from team platform" / "feature override" / "built-in fallback" hints.
- **`ModelNotConfiguredError`** carries `featureKey + role + featureDisplayName + projectId` plus a stable `cause = "MODEL_NOT_CONFIGURED"` for tRPC + REST wire-matching.
- **`seedOnboardingDefaultsForProvider`** seeds the role rows a provider can fulfil (OpenAI → all three; Anthropic → Default + Fast, no embeddings). Strictly additive.
- **`modelDefaults.service.ts`** with `setRoleAssignment` / `setFeatureOverride` — writes go to `ModelDefault` only, never the legacy columns. Locks the one-way drain.
- **`ScopeChipPicker`** extracted from `ModelProviderScopeSection` as a thin controlled component so B3.2's line-based UI can render the same chip picker without inheriting the drawer's form-state.

## Spec

`specs/model-providers/model-resolver-and-registry.feature` — 15/15 enforced scenarios bound. Two `@unimplemented`: the tRPC wire mapping ships with the call-site replacement PR; the migration data-lift is exercised on every `prisma migrate deploy` and proven from the read side by the legacy-compat scenario.

## Tests

- Real Postgres integration: full resolver walk (per-feature override → role default → legacy → constant → throw), onboarding seed (OpenAI / Anthropic / additive), write-side compat service.
- Unit: registry semantics (lookup, role grouping, key shape, duplicate-key guard).

## Out of scope (deliberate — follow-up PR)

- Wiring `resolveModelForFeature` into the ~30 `project.defaultModel ?? DEFAULT_MODEL` call sites and exposing it via tRPC procedures. That replacement sweep + the typed-error wire format are big enough to warrant their own PR.
- The line-based "Default Models" UI is Luigi's B3.2 (#4074 + follow-up).
- Dropping the legacy `defaultModel` / `topicClusteringModel` / `embeddingsModel` columns is a third PR once we verify no writes land on them in production.

## Coordination

Stacked on @bugfixer_luigi's PR B2 #4073. Will rebase onto main after B2 merges.

### DIFF
--- a/langwatch/prisma/migrations/20260515200000_add_model_default_table/migration.sql
+++ b/langwatch/prisma/migrations/20260515200000_add_model_default_table/migration.sql
@@ -1,0 +1,162 @@
+-- ModelDefault is the single source of truth for "which model does
+-- {feature, role} resolve to at {scope}". Replaces the per-scope
+-- defaultModel / topicClusteringModel / embeddingsModel scalar columns
+-- added in 20260515150000 (which stay readable for one release as a
+-- compat fallback — the resolver in B3.1 prefers ModelDefault rows and
+-- only walks the legacy columns when no row exists).
+--
+-- Row shape:
+--   featureKey IS NULL → role-level default for the scope
+--   featureKey populated → per-feature override (rarely needed but the
+--                          escape hatch when a single feature wants a
+--                          different model than its role's default)
+--
+-- Postgres treats NULLs as DISTINCT in unique indexes by default, so a
+-- single composite unique on (scopeType, scopeId, role, featureKey)
+-- would NOT prevent two role-level rows for the same (scope, role).
+-- Two partial unique indexes give the semantics we want without a
+-- single index that conflates "no override" with "override=empty".
+
+CREATE TABLE "ModelDefault" (
+    "id"         TEXT NOT NULL,
+    "scopeType"  TEXT NOT NULL,
+    "scopeId"    TEXT NOT NULL,
+    "role"       TEXT NOT NULL,
+    "featureKey" TEXT,
+    "model"      TEXT NOT NULL,
+    "createdAt"  TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt"  TIMESTAMP(3) NOT NULL,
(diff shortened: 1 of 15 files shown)


rogeriochaves:
Consolidated into #4073 per @rchaves request to merge all model-providers work into a single PR. Commits f2a65e159, 47c5b7ce3, 284df177c on feat/model-providers-hierarchical-defaults.

PR #4078, opened 2026-05-15
fix(traces): pagination back navigation + blank export warning

### LINKED ISSUE #4077: bug(traces): pagination back always jumps to page 1 + selected export returns blank file
## Bug

Two bugs reported by a customer on the legacy traces page (`/messages` route, `MessagesTable` + `NavigationFooter`):

1. **Pagination: clicking "previous page" always jumps to page 1.** After navigating forward 10+ pages using the `>` arrow, clicking `<` to go back one page resets to page 1 instead of page N-1.
2. **Export: exporting selected traces produces a blank (0-byte) file.**

---

### Root Cause (confirmed)

`NavigationFooter.tsx:168-179` — the `prevPage` function, when in cursor-based pagination mode, unconditionally resets to page 1:

```ts
const prevPage = useCallback(() => {
    if (useCursorPagination) {
      // ❌ Always resets to page 1, regardless of current position
      setCursorPageNumber(1);
      void router.push({
        pathname: router.pathname,
        query: buildPaginationQuery({ pageOffset: 0, scrollId: null }),
      }, undefined, { shallow: true });
    } else if (pageOffset > 0) {
      // ✅ Offset pagination works correctly
      void router.push({...});
    }
  }, [...]);
```

**Why this happens:** The forward navigation (`nextPage`) uses cursor-based pagination when the API returns a `scrollId`. Each cursor only points forward — there is no stack of previous cursors. When the user clicks "back," the code has no way to restore the previous cursor, so it falls back to offset 0 (page 1).

The `nextPage` function (line 130-163) switches into cursor mode on the first forward click:
```ts
if (currentResponseScrollId) {
    setCursorPageNumber((prev) => prev + 1);
    router.push({ query: buildPaginationQuery({ scrollId: currentResponseScrollId }) });
}
```

Once `scrollId` is in the URL, `useCursorPagination` is `true` and **all backward navigation is broken**.

### Design flaw

Cursor pagination is forward-only by nature. The implementation has no mechanism for backward navigation:
- No cursor history/stack to pop when going back
- No fallback to offset-based pagination for the page immediately behind

### Fix options

1. **Maintain a cursor stack** — push each scrollId onto a stack as the user navigates forward; pop when going back
2. **Hybrid approach** — use offset pagination for backward, cursor for forward (requires the backend to support both in the same session)
3. **Store page-to-cursor map** — cache `{ pageNumber → scrollId }` so any visited page can be revisited

---

### Analysis

The export flow (`useExportTraces` → POST `/api/export/traces/download` → `ExportService.exportTraces`) is structurally correct. The `selectedTraceIds` are threaded through from the UI to the ClickHouse query. However, the export produces a blank file when the backend's `ExportService.exportTraces` generator returns without yielding any chunks:

```ts
// export.service.ts:157-162
if (total === 0 || traces.length === 0) {
    logger.info({ projectId: request.projectId }, "No traces to export");
    return; // Generator returns without yielding → empty response body → 0-byte file
}
```

### Suspected root causes (needs reproduction)

**Most likely — time-range mismatch:** The export's `startDate`/`endDate` come from `usePeriodSelector` at render time, captured in the `useExportTraces` hook. If the time range rolls forward (e.g., "Last 7 days" preset recalculates between when traces were listed and when the export fires), traces near the boundary could fall outside the export's window. Combined with cursor pagination (which already shifted the visible window away from page 1), this could result in 0 matching traces.

**Alternative — stale `selectedTraceIds` closure:** React state update from `openExportDialog` and the callback reference in `startExport` are in the same render cycle. If there's a timing edge where `startExport` fires with stale state (undefined selectedTraceIds), the request omits `traceIds`, falls back to filter-only mode, and if filters are empty, the ClickHouse query scopes to the wrong data set.

### Reproduction needed

- Select 3-5 traces on a page > 1 (while using cursor pagination)
- Click Export selected → check if file is 0 bytes
- Check server logs for `"No traces to export"` or ClickHouse query params

---

## Key Files

| File | Role |
|------|------|
| `langwatch/src/components/NavigationFooter.tsx:168-179` | **Bug 1 root cause** — `prevPage` always resets to page 1 in cursor mode |
| `langwatch/src/components/NavigationFooter.tsx:130-163` | `nextPage` switches to cursor mode on first forward click |
| `langwatch/src/components/messages/MessagesTable.tsx:1340` | "Export selected" click handler |
| `langwatch/src/components/messages/useExportTraces.ts:170-293` | Export orchestration — `startExport` builds request body |
| `langwatch/src/server/export/export.service.ts:107-209` | Streaming export generator — blank file when first batch returns 0 |
| `langwatch/src/server/traces/clickhouse-trace.service.ts:1292-1524` | `fetchTracesWithPagination` — ClickHouse query with cursor/traceId filters |

## Context

- The customer is on the legacy `/messages` route (confirmed by "Items per page" pagination UI from `NavigationFooter`, not the traces-v2 `Pagination` component)
- The new traces-v2 page (`/traces`) uses offset-based pagination and is NOT affected by Bug 1
- Commit `b60a2e53c` recently fixed a server-side pagination crash on page 2+ but did not address this client-side cursor navigation issue

sergioestebance:
## Summary

Fixes two customer-reported bugs on the legacy traces page (`/messages` route):

- **Pagination back always jumped to page 1:** Cursor-based pagination was forward-only — `prevPage()` had no cursor history and unconditionally reset to offset 0. Added a cursor stack that stores each `scrollId` before navigating forward; `prevPage` pops the previous cursor to go back one page.
- **Export downloaded blank files silently:** When the server returns an empty response (0-byte blob), the export now shows a warning toast with guidance instead of silently downloading a blank file. Differentiates between "no matching traces" (warning) and "unexpected empty response" (error).

Closes #4077

## Changes

| File | Change |
|------|--------|
| `NavigationFooter.tsx` | Added `cursorStackRef` for backward cursor navigation; fixed `isPrevDisabled` for cursor mode; fixed stale aria-label |
| `useExportTraces.ts` | Added `blob.size === 0` check before download; shows context-aware toast |
| `useNavigationFooter.cursor.integration.test.ts` | New regression test for cursor pagination back navigation (verifies correct cursor + page number) |
| `useExportTraces.integration.test.ts` | New test for empty blob detection |

## Known limitations

- Cursor stack is client-side state (useRef) — lost on page refresh. A user who navigates to page 5, refreshes, and clicks back will go to page 1. Encoding the full cursor stack in the URL would be unwieldy with base64 cursors.

## Test plan

- [x] Regression test: prevPage from page 3 goes to page 2 with correct scrollId
- [x] Regression test: prevPage from page 2 returns to page 1 (no scrollId)
- [x] Regression test: empty blob triggers warning toast instead of download
- [x] Existing pagination tests pass (4/4, no regressions)
- [x] Existing export tests pass (12/12, no regressions)
- [x] Typecheck passes


sergioestebance:
## QA Evidence — Browser Testing on localhost:5560

### Bug 1: Pagination back navigation — VERIFIED

Navigated forward to page 4, then clicked back arrow step by step.

**On page 4 (before clicking back):**
Page 4 before back

**After clicking back → Page 3 (not page 1!):**
Page 3 after back

**Continued back → Page 2 → Page 1:**
Page 1 after full backtrack

**Result:** 4 → 3 → 2 → 1, one page at a time. Previously this would jump from any page directly to page 1.

---

### Bug 2: Export selected traces — VERIFIED

**3 traces selected with bulk action bar:**
Traces selected

**Export dialog showing "3 selected traces":**
Export dialog

**Export completed — file downloaded successfully:**
Export completed

**Verified downloaded file content (6,249 bytes, not empty):**
```csv
trace_id,timestamp,input,output,labels,first_token_ms,total_time_ms,...
9e80435ca4353eae...,1778873927072,Summarize the company in three short bullets.,"Here is the summary..."
9a277ceb84010390...,1778873919837,What makes the company useful for AI teams?,"For AI teams, we provide..."
09d5db06ee16d0...,1778873914173,Who typically uses the product inside a comp...,"It is primarily useful for..."
```

**Result:** Export produces a real CSV file with header + 3 data rows matching the selected traces. The defensive empty-blob check (`blob.size === 0`) correctly fires the "Export produced no data" warning toast when the server returns an empty response, instead of silently downloading a blank file.

PR #4097, opened 2026-05-17
feat(feature-flags): internal postgres-backed flag store + Ops UI

rogeriochaves:
## Summary

- Adds a postgres-backed flag store with a hardcoded code-level registry; SYSTEM-scoped flags (event-sourcing kill switches, pipeline toggles) resolve from env -> postgres -> registry default and **never** consult PostHog
- New `/ops/feature-flags` page (under existing `ops:view` / `ops:manage` gates) to flip values cluster-wide without a redeploy
- Loop-prevention kill switch `LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD` migrated into the registry as `ops_es_causality_loop_guard_disabled`; legacy env var still works

## Why

PostHog "Feature Flag Calls by Library" jumped from ~8k/day to ~50k/day around 2026-05-13. Root cause: `es-<aggregate>-<component>-<name>-killswitch` flags fanning out across (tenant × component × event) and hitting PostHog per unique cache key. Those are pure infrastructure flags — no user targeting, no A/B, never need PostHog. Moving them to a postgres value store + 5s in-process cache zeroes out the PostHog cost on that path.

PRODUCT-scoped flags (`release_ui_*`, `release_nlp_go_engine_enabled`) keep using PostHog for user targeting; postgres acts as an emergency override + self-hosted fallback.

## Design notes

- **Nothing auto-registers in the DB.** A row is written only when an operator flips a switch from the Ops UI. Absence of a row = use registry default.
- **The registry is hardcoded code** (`registry.ts`). Add entries only when (a) you want the flag to skip PostHog or (b) you want it listed on the Ops page. Ad-hoc PostHog flags continue to work without any registry entry via the legacy fallback path.
- **`es-*-killswitch` is a family prefix** — generated names from `event-sourcing/utils/killSwitch.ts` are resolved as SYSTEM scope without per-key registry edits.
- **In-process cache** sits in front of Redis on the hot path (5s local TTL, 60s Redis TTL) so per-event reactor lookups collapse to a Map read.

## Test plan

- [x] resolver unit tests cover SYSTEM never-calls-legacy, family-prefix matching, env-override priority, legacy-env alias back-compat, PRODUCT delegation, PRODUCT-falls-back-to-postgres-when-PostHog-down, unregistered-flag-fallthrough
- [x] existing FeatureFlagService unit suite migrated to the new `legacy` field, 10/10 pass
- [x] evaluationTrigger reactor unit suite, 17/17 pass
- [x] `FeatureFlag` model added to `EXEMPT_MODELS` in dbMultiTenancyProtection (cluster-wide, not project-scoped)
- [ ] Browser dogfood: open `/ops/feature-flags`, toggle a flag, verify resolver flips via tRPC roundtrip — screenshots to follow

## Spec

See `specs/ops/internal-feature-flags.feature` for the full behavioral contract.

## Dogfood screenshots

Page loaded against the dev RDS install after applying the migration — sees the SYSTEM section with the loop-guard kill switch, the PRODUCT section with PostHog-managed flags (env-override badge on the two that have RELEASE_* env vars set locally), and the families table with the `es-*` SYSTEM family.

feature flags page loaded

After flipping the SYSTEM kill switch from the UI: source flips from "registry default" to "postgres", lastEditedBy and timestamp populate, and a `clear` action appears for removing the override.

feature flags after toggle

Verified end-to-end with a direct prisma read:

```
{
  "key": "ops_es_causality_loop_guard_disabled",
  "enabled": true,
  "lastEditedBy": "IGptxn12wy8rlk8x1nqUq",
  "updatedAt": "2026-05-17T14:59:42.883Z"
}
```

## Round 3: type enforcement + enumerated kill switches + PII flag

Addresses follow-up scope (rchaves direction): `EXPLICIT_FLAGS` → `FEATURE_FLAGS`, type-safe `FeatureFlagKey` enforced at the `isEnabled` boundary, every generated `es-*-killswitch` enumerated on the Ops page, plus `DISABLE_PII_REDACTION` migrated.

Live dogfood: page now lists 48 togglable rows (40 auto-enumerated `es-*-killswitch` from projections / mapProjections / commands across registered ES pipelines + 8 explicit flags including the new `ops_pii_redaction_disabled`).

feature flags page with enumerated kill switches

Architecture / docs:
- `FeatureFlagKey = RegisteredFeatureFlagKey | EsKillSwitchKey` (template literal) exported from registry.
- `FeatureFlagService.isEnabled(key, ...)` requires `FeatureFlagKey`; unregistered string literals fail at compile time.
- `generateKillSwitchKey()` returns `EsKillSwitchKey`; `KillSwitchOptions.customKey` accepts only `FeatureFlagKey`.
- `getKillSwitchDescriptors()` walks `def.foldProjections`, `def.mapProjections`, `def.commands` for every registered pipeline.
- `dev/docs/adr/005-feature-flags.md` rewritten end-to-end.
- Em-dashes removed from PR-added lines; "(PostHog not configured)" copy dropped.


langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/evaluationTrigger.reactor.ts
```diff
@@ -63,12 +67,17 @@ export function createEvaluationTriggerReactor(
...
+      // bypasses the check (emergency rollback without redeploy).
+      // Resolved through featureFlagService so operators can flip it
+      // from the Ops UI without restarting pods; the legacy
+      // `LANGWATCH_DISABLE_CAUSALITY_LOOP_GUARD=1` env var still works
+      // via the standard env-override path (uppercased flag key).
+      const guardDisabled = await featureFlagService.isEnabled(
+        CAUSALITY_LOOP_GUARD_DISABLED_FLAG,
+        tenantId,
+        false,
+      );
```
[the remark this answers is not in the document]
rogeriochaves: Acknowledged — the positional shape is the existing `FeatureFlagServiceInterface.isEnabled` API which has ~36 callers across the codebase. Refactoring that contract is genuinely out of scope for this PR (its goal is the postgres path, not the call signature). Leaving the callsite as-is to match the shared interface; happy to revisit in a follow-up if the wider call-site refactor lands as its own change.

sergioestebance:
## Pre-Landing Review

**Status: CLEAN with minor findings.** No blocking issues.

---

### Findings

**[P2] (confidence: 9/10) `featureFlag.service.ts:74` — `defaultValue = true` is a dangerous default for the implementation signature**

The overload signature says `defaultValue?: boolean` (optional, no default). The implementation signature defaults to `true`. If someone calls `isEnabled("some_system_flag", "tenant")` without passing `defaultValue`, and the flag hits the unregistered legacy path (line 116), it falls through with `defaultValue = true` — the flag is ON by default. For kill switches (where `true` = disabled component), accidentally omitting `defaultValue` silently disables things.

The registry-aware paths use `definition.defaultValue` so they're safe. All current callers pass explicit `defaultValue`. No action needed for this PR, but worth flipping the implementation default to `false` (fail-closed) in a follow-up.

---

**[P2] (confidence: 8/10) `ops.ts:794-800` — `clearFeatureFlag` does not validate key against registry**

`setFeatureFlag` validates the key via `resolveFlagDefinition(input.key)` and rejects unknowns. `clearFeatureFlag` skips that check entirely. An operator with ops:manage could clear any arbitrary key string from the DB.

Low-risk since `deleteMany` on a non-existent key is a no-op, and you need ops:manage to reach this endpoint. But it's inconsistent with `setFeatureFlag`'s validation. Consider adding the same check for symmetry.

---

**[P3] (confidence: 7/10) `featureFlagStore.postgres.ts:108-113` — `set()` cache invalidation is not atomic with the DB write**

After `prisma.featureFlag.upsert()` succeeds, the method deletes the Redis cache entry and the local cache entry. If the process crashes between the upsert and the cache delete, the stale cache entry remains until natural TTL expiry (60s Redis, 5s local). The code already documents this trade-off. No action needed.

---

**[INFORMATIONAL] `featureFlag.service.ts:95-113` — PRODUCT flag fallback to postgres on PostHog failure is silent to the operator**

When PostHog fails for a PRODUCT flag, the service falls back to the postgres store with a `logger.warn`, but the Ops UI doesn't surface that PostHog is unreachable. Consider adding a health indicator to the Ops page in a follow-up.

---

### What's well done

- Architecture is sound — the SYSTEM/PRODUCT scope split is clean and the registry-driven resolution with family-prefix matching handles the dynamic ES kill switches well
- Type safety via `FeatureFlagKey` union + `as const satisfies` is exactly right
- Back-compat preserved: legacy env vars, unregistered flag fallthrough, existing PostHog path all keep working
- Test coverage on the resolver covers the critical invariants (SYSTEM-never-calls-legacy, family matching, env overrides, PRODUCT delegation + fallback)
- Multi-tenancy exemption is correct and well-documented
- Permission gating is consistent (server-side + UI-side)

---

**Note:** The `promptAttributes.ts` change (adding `draft` to `PromptReference` return values) is unrelated to feature flags — minor scope creep but harmless.

rogeriochaves:
Thanks Sergio.

**P2 #1 (`defaultValue = true`):** fixed in 51b7fc726 — flipped the implementation default to `false` so the legacy fallthrough fails closed.

**P2 #2 (`clearFeatureFlag` validation):** intentionally permissive. `listFeatureFlags` surfaces orphan rows (DB keys that no longer match the registry or pipeline graph) so operators can delete them from the Ops UI. Adding the validation would break that cleanup path. Added a comment to capture that intent.

**P3 (cache invalidation not atomic):** acknowledged, no action — TTLs bound the staleness window.

**Informational (silent PostHog fallback):** noted for follow-up.

sergioestebance (commented):
Codex review findings on the current head (`51b7fc7`).

langwatch/src/server/featureFlag/featureFlag.service.ts
```diff
@@ -3,138 +3,137 @@ import { createLogger } from "~/utils/logger/server";
...
-    );
-    return result;
+
+    if (definition?.scope === "SYSTEM") {
+      const stored = await this.store.get(flagKey);
+      if (stored !== null) return stored;
+      return definition.defaultValue;
+    }
+
+    if (definition?.scope === "PRODUCT") {
```
sergioestebance, line 100: **P1:** PRODUCT postgres fallback/override is effectively dead for real backends. This branch returns the legacy result immediately, but the concrete legacy services don’t surface the fallback conditions: `FeatureFlagServiceMemory` returns the default when PostHog is not configured, and `FeatureFlagServicePostHog` catches PostHog failures internally and returns the default. That means a postgres row set from `/ops/feature-flags` will not enable a PRODUCT flag for self-hosted installs or during PostHog failures, despite the PR goal and Ops UI copy. Minimal fix: make PRODUCT resolution explicitly consult the store when PostHog is unavailable/unhealthy, or before PostHog when a postgres override row should be authoritative, and add tests using the concrete memory/PostHog behavior rather than only a mock that throws.

langwatch/src/server/api/routers/ops.ts
```diff
@@ -639,4 +650,156 @@ export const opsRouter = createTRPCRouter({
...
+        key: z.string().min(1).max(200),
+        enabled: z.boolean(),
+      }),
+    )
+    .mutation(async ({ ctx, input }) => {
+      // Only allow writes for keys the code registry recognizes,
+      // either as an explicit flag or via a family prefix. Stops typos
+      // from creating orphan rows that drift forever after the flag
+      // they were meant for has been renamed.
+      const def = resolveFlagDefinition(input.key);
```
sergioestebance: **P3:** This validation still accepts any family-shaped `es-*...-killswitch` key because `resolveFlagDefinition()` matches the broad family prefix/suffix. An `ops:manage` client can therefore create typo kill-switch rows that never affect the intended component, which conflicts with the comment’s stated goal of preventing drifting orphan rows. Minimal fix: keep `clearFeatureFlag` permissive for cleanup, but restrict `setFeatureFlag` to explicit registry keys or keys returned by `getKillSwitchDescriptors()`.

rogeriochaves:
Both fixed in 9c266bf8f.

**P1 (PRODUCT postgres fallback unreachable):** you're right, the try/catch was dead because both legacy backends return the default instead of throwing. Reordered so postgres wins for PRODUCT — operator overrides from `/ops/feature-flags` actually flip the flag, and self-hosted installs (no PostHog) get the same path. PostHog still owns user targeting when no DB override exists. Updated the resolver tests to cover store-first + operator-disable-over-PostHog.

**P3 (family-shaped typos):** good catch. `setFeatureFlag` now validates against explicit `listFeatureFlags()` ∪ `getKillSwitchDescriptors()` (live pipeline graph), so an `es-typo-bar-killswitch` write is rejected. `clearFeatureFlag` stays permissive on purpose so the orphan-row cleanup path keeps working.

Also rebased on main to drop the conflict.

sergioestebance (approved):

PR #4111, opened 2026-05-18
feat(feature-flags): named-opts isEnabled + close type-safety hole

rogeriochaves:
## Context

Two follow-ups after merging #4097, surfaced when investigating yesterday's PostHog Feature Flag Calls spike chart:

| Flag | Calls last 2 days | Status |
|------|---|---|
| `es-observability-anomaly-detection-killswitch` | 69,081 | Fixed by #4097 — drops to **zero** after deploy ~20:30 UTC May 17 |
| `token-estimation-project-killswitch` | 7,948 | Still on PostHog (hardcoded raw key, outside `es-*-killswitch` family) |
| `token-estimation-killswitch` | 757 | Same |
| `release_nlp_go_engine_enabled` | 2,058 | Expected (PRODUCT, per-project rollout) |

## What's in this PR

1. **Close the type-safety hole.** `FeatureFlagServiceInterface.isEnabled` was typed as `flagKey: string`, so every dependency-injected caller bypassed the `FeatureFlagKey` overload on the concrete class. Tighten the interface so the rule applies everywhere. Class implementations keep widening to `string` internally (bivariant method params) which preserves runtime back-compat for unregistered ad-hoc PostHog keys without weakening the typecheck. **Unregistered keys now fail at compile time across the codebase, not just at the singleton call site.**

2. **Switch `isEnabled` to named opts**, per CodeRabbit's earlier feedback and the repo convention banning positional args for multi-arg functions:

```ts
// before
await featureFlagService.isEnabled(flag, distinctId, false, { projectId, cacheTtlMs });
// after
await featureFlagService.isEnabled(flag, { distinctId, defaultValue: false, projectId, cacheTtlMs });
```

3. **Register the two token-estimation kill switches** (`token-estimation-killswitch`, `token-estimation-project-killswitch`) as SYSTEM flags so the per-span hot path resolves from env + postgres instead of PostHog. Estimated ~8.7k/day of `posthog-node` traffic gone after deploy.

Updated all 9 call sites + 5 test files + ADR-005.

## Test plan
- [x] `pnpm typecheck` clean (catches any remaining unregistered call sites)
- [x] `pnpm test:unit` — 823/823 pass across featureFlag, observability, traces, event-sourcing
- [ ] Post-deploy: verify `token-estimation-*-killswitch` calls in PostHog go to zero for posthog-node library

### DIFF
--- a/dev/docs/adr/005-feature-flags.md
+++ b/dev/docs/adr/005-feature-flags.md
@@ -47,7 +47,7 @@ PostHog is **never** called on the SYSTEM path. PRODUCT keeps PostHog as the sou
 ### Adding a new flag
 
 1. Add a `FEATURE_FLAGS` entry with `scope: "SYSTEM" | "PRODUCT"`, `defaultValue`, `description`. For SYSTEM flags migrated from an existing env variable, set `legacyEnvVar` so the old name keeps working.
-2. Call `featureFlagService.isEnabled("your_new_flag_key", distinctId, defaultValue?, options?)`. The key is type-checked against `FeatureFlagKey`.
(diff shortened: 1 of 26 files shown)


sergioestebance (approved):
lgtm

PR #4113, opened 2026-05-18
fix(ragas): migrate Faithfulness _create_statements patch to ragas>=0.3 API

Aryansharma28:
## Problem

Every Ragas **Faithfulness** evaluation on `app.langwatch.ai` fails. Users see:

```
'StatementGeneratorOutput' object has no attribute 'sentences'
```

and — because #3993 added SDK retries on gateway failures — intermittent `502 Bad Gateway` / `524` from `/api/evaluations/ragas/faithfulness/evaluate` (the retried request re-hits the deterministic crash, or exceeds the Cloudflare origin timeout while the upstream LLM is slow).

## Root cause

`langevals/evaluators/ragas/langevals_ragas/faithfulness.py` monkey-patches Ragas' internal `_create_statements` to capture extracted claims. It reads the **Ragas ≤ 0.2** `SentencesSimplified` shape (`.sentences[].simpler_statements`):

```python
for components in statements.sentences:
    ragas_statements += components.simpler_statements
```

But `main` pins **`ragas==0.4.3`** (from the `fix/deps-ragas` / CVE-2025-45691 upgrade). In Ragas ≥ 0.3, `_create_statements()` returns `StatementGeneratorOutput` with a **flat `.statements: list[str]`** — there is no `.sentences`. The CVE upgrade bumped the dependency but never migrated this call site; `git log` shows the line untouched since the monorepo move (`#1591`).

## Fix

Read the flat `.statements` on Ragas ≥ 0.3, with the old shape kept as a fallback so the evaluator is correct on both the deployed `0.4.3` pin and older local pins (`langy-v2-integration` still pins `0.2.9`).

The `_create_verdicts` patch and `verdict.statement/reason/verdict` usage were already correct for 0.4.3 (`NLIStatementOutput` unchanged) — only `_create_statements` needed migrating.

## Test

Added `test_faithfulness_supports_ragas_0_3_statement_generator_output` — an **offline** test (no LLM/network) that executes the real `evaluate()` path with the `StatementGeneratorOutput` shape.

Verified against `ragas==0.4.3`:
- **Without the fix:** fails with `AttributeError: ... object has no attribute 'sentences'` at `faithfulness.py:86` (the exact production error).
- **With the fix:** passes.

### DIFF
--- a/langevals/evaluators/ragas/langevals_ragas/faithfulness.py
+++ b/langevals/evaluators/ragas/langevals_ragas/faithfulness.py
@@ -83,8 +83,15 @@ class RagasFaithfulnessEvaluator(
         async def _create_statements(self, row: dict, callbacks):
             nonlocal ragas_statements
             statements = await _original_create_statements(row, callbacks)
-            for components in statements.sentences:
-                ragas_statements += components.simpler_statements
+            # ragas >= 0.3 returns StatementGeneratorOutput with a flat
+            # `.statements: list[str]`. ragas <= 0.2 returned SentencesSimplified
+            # with `.sentences[].simpler_statements`. Support both so this works
+            # on the deployed pin and older local pins alike.
+            if hasattr(statements, "statements"):
+                ragas_statements += statements.statements
+            else:
+                for components in statements.sentences:
+                    ragas_statements += components.simpler_statements
             return statements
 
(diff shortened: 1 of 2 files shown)


sergioestebance (approved):
lgtm

PR #4147, opened 2026-05-20
feat(retention): per-tenant configurable data retention with CH-native TTL

sergioestebance:
## Why

Implements the data retention feature per RFCv2. Tenants need configurable retention periods to comply with data governance and manage storage cost; until now all data was kept indefinitely. The configuration layer follows the multi-scope targeting standard from ADR-021 (#4281) — the same shape as model costs and default models.

## What changed

Per-category (traces / scenarios / experiments) retention, configurable at the **organization, team, or project** level, with ClickHouse-native TTL enforcement. Data is stamped at ingestion with `_retention_days` and `_size_bytes`; ClickHouse TTL rules handle deletion.

**Policy configuration (ADR-021 scoped resource)**
- `RetentionPolicy` table: single-scope-per-row, inline `(scopeType, scopeId)` + an `organizationId` tenancy anchor, **one row per (scope, category)** so categories cascade independently. No JSON policy columns (the shape ADR-021 rejects).
- Resolution cascades **PROJECT → TEAM → ORGANIZATION → platform default**, most-specific-wins, over the shared `resolveScopeChain`. Retention is **default-on**: with no override the resolver returns `PLATFORM_DEFAULT_RETENTION_DAYS` (49 days / 7 weeks), not indefinite; pre-column rows read the migration default (308 days) until re-stamped. An org admin can set a retention once and have every project inherit it.
- Registered in `SCOPED_MODELS` + `ORG_TENANCY_EXEMPT`, mirroring `CustomLLMModelCost`.
- Shared scope contract end-to-end: `SCOPE_TIERS` / `scopeAssignmentSchema` / `resolveScopeChain`.

**Enforcement pipeline (storage-shape-agnostic)**
- `_retention_days` / `_size_bytes` columns on all 11 retention-managed CH tables; TTL reconciler activates row-level TTL while preserving cold-storage tiering.
- Ingestion stamps every record via `context.retentionPolicy` (resolved through a TTL-cached resolver that delegates row-fetching to the repository).
- Retroactive `ALTER TABLE UPDATE` to backfill existing data, tracked via `system.mutations`, rate-limited per tenant/category, with kill support.
- Storage metering: per-category `SUM(_size_bytes)` with a 5-min cache.
- Trace pinning is a **UI annotation only** — pinned traces still age out under policy. Two-layer event-driven orphan sweep cleans up PG records (annotations, shares, pins, …) once their CH trace expires.

**API + UI**
- Router goes through the service layer only (no raw Prisma / repositories in routes): `getRules` (grouped rules + RBAC-filtered writable scopes), `setForScope`, `removeForScope` with `assertCanManageScope` (a project admin cannot push a policy up to the org). Retroactive + metering endpoints alongside.
- Settings page (`/settings/data-retention`) built on the shared `ScopeChipPicker`: effective-retention card, overrides table, add-override drawer.

## Screenshots

UI walkthrough of `/settings/data-retention`, showing the cascade resolving most-specific-wins and the paid-plan gate. Full-res in [langwatch/pr-screenshots › pr-4147](https://github.com/langwatch/pr-screenshots/tree/main/pr-4147/settings/data-retention).

**1. No policy → falls back to the platform default (49 days / 7 weeks)**

No retention policy

**2. Add-policy drawer — scope picker (org/team/project), retention presets, "apply to existing data" toggle**

Add retention policy drawer

**3. Organization policy → effective becomes 364 days (1 year), inherited by every project**

Organization policy

**4. Team override → the team's 6 months (182d) wins over the org's 364d**

Team overrides organization

**5. Project override → the project's 3 months (91d) wins over team and org**

Project overrides everything

**6. Free plan → configurable retention is gated; overrides become read-only behind an upsell banner**

Free-plan paid-feature gate

## Test plan

- Cascade resolver unit tests (per-category, 3-tier, sibling-scope isolation).
- Tenancy partition tests green with `RetentionPolicy` registered.
- TTL reconciler + RMT/TTL coexistence integration test.
- Targeted retention + tenancy + projection-router suite: 100 passing. Typecheck clean (one pre-existing unrelated `@aws-sdk/s3-request-presigner` error on main).
- Specs: `retention-policy-configuration` (3-tier cascade, single-org anchor, project admin cannot set org policy), `trace-pinning` (pins age out and are swept).

## Anything surprising?

- **Pins are not a retention exemption.** A manually pinned trace disappears once it passes retention and the orphan sweep removes the dangling `PinnedTrace` row — documented with a spec scenario.
- ClickHouse `ALTER … UPDATE` / `KILL MUTATION` can't be parameterized; tenant/mutation inputs are backslash+quote escaped.
- `_size_bytes` is computed structurally at write time (no double `JSON.stringify`). A follow-up will revisit the metering strategy; the storage display reads `SUM(_size_bytes)` and is eventually-consistent under ReplacingMergeTree.
- The foundation migration was rewritten in place (JSON columns were never deployed); it also creates the `PinSource` enum the schema referenced but no migration had created.

---

## Incidental CI fixes (not retention)

To get this branch's CI green, two non-retention failures were resolved on this PR:

- **`feature-parity`** — bound the two previously-unbound `pr-4147-regressions.feature` scenarios ("Pinning a trace does not change retention", "Retroactive retention update applies uniformly…") to their already-existing tests (`8/8` bound now).
- **`test-integration` OOM** — `api-keys-scope-filter.integration.test.tsx` (from #4282, on `main`) wedged the shard with an infinite render loop from an unstable mock reference, hidden green by the `globalSetup` 20-min `process.exit(0)` hard floor. Fixed the mock to return stable references. Root cause + the systemic CI-masking concern are tracked in #4450.


langwatch/src/server/data-retention/policy/dataRetentionPolicy.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    defaultRetentionPolicy,
+  }: {
+    organizationId: string;
+    defaultRetentionPolicy: RetentionPolicy | null;
+  }): Promise<void> {
+    await this.repository.updateOrgPolicy({
+      organizationId,
+      defaultRetentionPolicy,
+    });
+  }
```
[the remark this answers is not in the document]
sergioestebance: Fixed — `updateOrgPolicy` now queries all projects under the org and invalidates each one's retention cache entry.

langwatch/src/server/data-retention/pinning/pinnedTrace.service.ts
```diff
@@ -0,0 +1,192 @@
...
+      return;
+    }
+
+    const traceIdList = traceIds.map((id) => `'${id.replace(/'/g, "\\'")}'`).join(",");
+
+    for (const table of TRACE_TABLES) {
+      try {
+        await client.command({
+          query: `ALTER TABLE ${table} UPDATE _retention_days = ${retentionDays} WHERE TenantId = '${projectId}' AND TraceId IN (${traceIdList}) AND _retention_days != ${retentionDays}`,
+        });
```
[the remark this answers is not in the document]
sergioestebance: Addressed in commit 434b8fd86. ClickHouse `ALTER TABLE ... UPDATE` does not support `query_params` binding for DDL/mutation statements (only `SELECT` and `INSERT` accept `{param:Type}`). The mitigation is sanitization: the `esc()` function escapes backslash + single quote (`s.replace(/\\/g, '\\\\').replace(/'/g, "\\\'")`) for `projectId` and each `traceId` before interpolation. CodeQL flagged backslash-incomplete-escape on the previous version; that thread is now resolved.

langwatch/src/server/data-retention/retroactive/retroactiveUpdate.service.ts
```diff
@@ -0,0 +1,169 @@
...
+      const blocked = activeMutations.map((m) => m.table).join(", ");
+      throw new Error(
+        `Retroactive update already in progress for: ${blocked}. Wait for completion before starting another.`,
+      );
+    }
+
+    for (const table of tables) {
+      await client.command({
+        query: `ALTER TABLE ${table} UPDATE _retention_days = ${newRetentionDays} WHERE TenantId = '${projectId}' AND _retention_days != ${newRetentionDays} AND _retention_days != 0`,
+      });
```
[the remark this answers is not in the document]
sergioestebance: Addressed in commit 434b8fd86. Same constraint as the pinnedTrace thread: ClickHouse `ALTER TABLE ... UPDATE` and `KILL MUTATION` do not accept `query_params` binding (those work for `SELECT` and `INSERT` only). The mitigation here is the shared `esc()` helper which escapes backslash + single quote before `projectId`/`mutationId` interpolation. The `KILL MUTATION` path is additionally constrained because `mutationId` is server-issued (from `system.mutations`), not user-supplied. `newRetentionDays` is a typed number, not a string. CodeQL backslash-escape findings on this file are now resolved.

langwatch/src/components/traces/TraceDetails.tsx
```diff
@@ -309,6 +310,9 @@ export function TraceDetails(props: {
                   Add to Dataset
                 </Button>
               )}
+              {project && (
+                <PinButton projectId={project.id} traceId={props.traceId} />
+              )}
```
rogeriochaves, line 315: was this added to the ... menu on tracesv2 details as well?

rogeriochaves:
I feel like the retention configurations should actually be multi-targeted and cascading, just like the default model configuration. You can check out this other PR for what the standard for it is, and I think we should implement the same here https://github.com/langwatch/langwatch/pull/4281
 
I think some of that was done here as well, like the configuration can be 13, but we want to standardize everything on the same standard we are doing on 4281. I think this case here is really similar to the custom model case, and we can even make the UI similar to show all the overrides etc.

langwatch/src/server/clickhouse/migrations/00032_add_retention_and_size_columns.sql
```diff
@@ -0,0 +1,266 @@
...
+--   replication queue / stuck mutations) — see runbook before retrying.
+--
+-- * SETTINGS alter_sync = 1, mutations_sync = 0 — wait only for the local
+--   replica, never queue behind unrelated mutations.
+
+
+-- event_log: _retention_days
+-- +goose StatementBegin
+ALTER TABLE ${CLICKHOUSE_DATABASE}.event_log
+  ADD COLUMN IF NOT EXISTS `_retention_days` UInt16 DEFAULT 308 CODEC(Delta(2), ZSTD(1))
```
langwatch-agent, line 37: **P1** — These `DEFAULT 308` definitions make retention opt-out impossible for pre-existing data. Adding a ClickHouse column with `DEFAULT 308` makes old rows read as 308, and the reconciler installs a DELETE TTL for all managed tables; tenants with no policy will therefore lose rows older than 308 days even though the resolver and ingestion spec define no override as indefinite (`0`), and the spec explicitly says pre-policy data keeps `0`. Use `DEFAULT 0` for all 11 tables. If a platform-wide 308-day purge is intended, it needs an explicit separately-reviewed migration/rollout rather than being coupled to opt-in retention.

langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.clickhouse.repository.ts
```diff
@@ -171,6 +173,7 @@ export class DspyStepClickHouseRepository implements DspyStepRepository {
         CreatedAt: new Date(existing?.createdAt ?? data.createdAt),
         InsertedAt: new Date(existing?.insertedAt ?? data.insertedAt),
         UpdatedAt: new Date(data.updatedAt),
+        _retention_days: 0,
```
langwatch-agent: **P1** — `dspy_steps` is registered as a trace-retention table, but both write paths hard-code `_retention_days: 0` (the direct insert does the same at line 121). The production `/misc` DSPy endpoint reaches `upsertStep()` without any retention context, so a project with 30-day trace retention keeps newly written DSPy rows indefinitely; an upsert can also reset a retroactively-updated row back to `0`. Pass the tenant's resolved traces retention into this repository/service and stamp it on both insert paths. Add coverage for the `dspy_steps record has _retention_days = 30` scenario already declared in `ingestion-stamping.feature`.

langwatch/src/server/data-retention/orphan-sweep/orphanSweep.repository.ts
```diff
@@ -0,0 +1,128 @@
...
+    const traceIds = new Set<string>();
+    for (const row of annotations) traceIds.add(row.traceId);
+    for (const row of queueItems) traceIds.add(row.traceId);
+    for (const row of publicShares) traceIds.add(row.resourceId);
+    for (const row of triggerSents) {
+      if (row.traceId) traceIds.add(row.traceId);
+    }
+    for (const row of pinnedTraces) traceIds.add(row.traceId);
+
+    return [...traceIds].slice(0, limit);
```
langwatch-agent: **P2** — The proactive sweep can permanently starve records beyond the first candidate window. Each source query uses `take: limit`, the merged set is sliced back to 1000, and `sweepProject()` performs a single pass with no cursor. If the first 1000 referenced traces are still live, every hourly sweep rechecks the same rows and stale annotations/shares/pins after that prefix are never reached. Add deterministic pagination/rotation (or loop through bounded pages with a persisted cursor) so subsequent sweeps advance even when a page contains no deletions, and cover a full-live first page followed by an orphan.

langwatch-agent (commented):
Found 2 P1 retention correctness issues and 1 P2 sweep-progress issue; details are inline. `git diff --check origin/main...HEAD` passes. The `feature-parity` check is currently failing because `langwatch/src/server/data-retention/__tests__/retroactiveUpdate.unit.test.ts:140` references an unknown scenario annotation: `Conflict error names the mutation IDs callers can kill`.

langwatch/src/components/traces/TraceDetails.tsx
sergioestebance: Yes — pin/unpin is in the traces-v2 drawer overflow (`…`) menu via `TraceOverflowMenu.tsx` (commit 5b400bcb3), wired to the same `pinnedTrace.pin`/`unpin`/`getPin` endpoints as the v1 header button, with the `Pin trace`/`Unpin trace` toggle + `LuPin`/`LuPinOff` icons.

langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.clickhouse.repository.ts
sergioestebance: Fixed in ccea27ca1. `DspyStepClickHouseRepository` now takes a `RetentionPolicyResolver` and stamps the tenant's resolved `traces` retention on **both** write paths (`upsertStep` and `insertStepDirect`) instead of the hard-coded `0`. Re-resolving on every upsert means an upsert no longer resets a retroactively-updated row (the resolver returns the same current policy). Added `dspy-step.clickhouse.repository.unit.test.ts` covering the stamping (resolver value, no-policy → 0, no-resolver → 0).

langwatch/src/server/data-retention/orphan-sweep/orphanSweep.repository.ts
sergioestebance: Fixed in ccea27ca1. `findCandidateTraceIds` is now id-cursor paginated per source and `sweepProject` walks every page until all five sources drain (bounded by `MAX_SWEEP_PAGES`, which logs a warning if hit). A fully-live first page no longer starves later records — the cursor advances past it. Cursoring by the immutable `id` PK is stable under the deletes the sweep performs (`id > cursor` window never shifts). New regression test + scenario: "Proactive orphan sweep advances past a fully-live first page".

langwatch/src/server/clickhouse/ttlReconciler.ts
```diff
@@ -260,15 +298,36 @@ export async function reconcileTTL(
       }
 
       // TTL volume routing (`TO VOLUME 'cold'`) only works on tables using the
-      // tiered storage policy. Tables on 'default' policy don't have a cold volume.
+      // tiered storage policy. Tables on 'default' policy don't have a cold volume,
+      // but they CAN still have retention DELETE TTL.
       if (tableInfo.storage_policy !== TIERED_STORAGE_POLICY) {
```
langwatch-agent: **P1** — This retention-only path is unreachable for the shared non-cold-storage deployment it is intended to support. `clickhouseMigrate.ts` calls `reconcileTTL({ verbose: true })` for the shared instance, but `reconcileTTL()` still returns early at line 259 unless `CLICKHOUSE_COLD_STORAGE_ENABLED=true`. The Helm chart only sets that flag when cold storage is enabled, and the existing unit test explicitly expects reconciliation to skip when only `CLICKHOUSE_URL` is set. As a result, normal self-hosted/default-storage installs stamp `_retention_days` but never install the DELETE TTL, so retention is silently unenforced. Run reconciliation whenever ClickHouse is configured; gate only the cold-volume MOVE clause, and add an env-path regression test for retention-only activation.

langwatch/src/server/api/routers/dataRetention.ts
```diff
@@ -0,0 +1,184 @@
...
+        retentionDays: retentionDaysSchema,
+      }),
+    )
+    .use(authorizeInResolver)
+    .mutation(async ({ input, ctx }) => {
+      await assertCanWriteRetentionScope(
+        { prisma: ctx.prisma, session: ctx.session },
+        input.scope,
+      );
+      await assertRetentionPlanForProject(ctx, input.projectId);
```
langwatch-agent: **P2** — The paid-plan check is performed against caller-controlled `input.projectId`, not against the organization that owns `input.scope`. These are independent inputs. A user who can manage a scope in a free org and has access to any paid-org project can pass that paid project id here and set or remove overrides in the free org. Resolve the target scope's owning organization first and gate that organization; apply the same fix to `removeForScope` below.

langwatch/src/server/api/routers/dataRetention.ts
```diff
@@ -0,0 +1,184 @@
...
+        newRetentionDays: retentionDaysSchema,
+      }),
+    )
+    .use(checkProjectPermission("project:update"))
+    .mutation(async ({ input, ctx }) => {
+      await assertRetentionPlanForProject(ctx, input.projectId);
+      return getApp().dataRetention.retroactive.triggerUpdate({
+        projectId: input.projectId,
+        category: input.category as RetentionCategory,
+        newRetentionDays: input.newRetentionDays,
```
langwatch-agent: **P1** — Retroactive deletion policy is taken directly from the client and is not tied to the project's persisted/effective policy. A `project:update` caller can rewrite all existing rows to any accepted retention value without creating a matching rule. The current UI makes this easier to hit accidentally: it can save policies for arbitrary selected scopes but always calls this endpoint for the page's `projectId`, so selecting project/team B can restamp project A's history. Because contraction is irreversible once merges run, derive the effective retention server-side for the target project/category and define scope-wide apply semantics explicitly (or restrict the UI to scopes in the current project's cascade).

langwatch/src/server/data-retention/orphan-sweep/orphanSweep.service.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    for (const batch of batches) {
+      await this.cleanupBatch({ projectId, traceIds: batch });
+    }
+  }
+
+  async sweepProject({ projectId }: { projectId: string }): Promise<void> {
+    let cursor: OrphanCandidateCursor | undefined;
+
+    for (let page = 0; page < MAX_SWEEP_PAGES; page++) {
```
langwatch-agent, line 100: **P2** — The pagination fix still starves the tail for large projects. After `MAX_SWEEP_PAGES` the cursor is discarded; the next hourly run starts from the beginning and rechecks the same first 100 pages. With more than 100k live rows in any source, orphaned records after that prefix are never reached. Persist the continuation cursor or enqueue a continuation job carrying it, and add a regression test that crosses the cap.

langwatch/src/server/data-retention/orphan-sweep/retentionOrphanSweep.reactor.ts
```diff
@@ -0,0 +1,42 @@
...
+      const { tenantId } = context;
+
+      const retentionDays = await deps.retentionPolicyCache.getRetentionDays(
+        tenantId,
+        "traces",
+      );
+      if (retentionDays === 0) return;
+
+      try {
+        await deps.orphanSweep.sweepProject({ projectId: tenantId });
```
langwatch-agent: **P2** — This ingestion reactor is the only production caller of `sweepProject()`. The feature spec promises read-time lazy cleanup for inactive tenants, but annotation lists, annotation queues, trigger history, and public-share reads never call `filterOrphanedTraceIds()` or `cleanupOrphans()`. If a tenant stops ingesting before TTL deletes its traces, dangling PG rows and public shares remain indefinitely. Wire the documented read-time cleanup paths or add a scheduled sweep independent of new ingestion, with coverage for a tenant that has no later trace events.

langwatch-agent (commented):
Found 2 P1 retention-correctness issues and 3 P2 cleanup/authz issues on the current head; details are inline. `git diff --check origin/main...origin/review/pr-4147` passes. CI is still running (`test-unit (2)` pending when reviewed).

langwatch/src/components/traces/TraceDetails.tsx
```diff
@@ -309,6 +310,9 @@ export function TraceDetails(props: {
                   Add to Dataset
                 </Button>
               )}
+              {project && (
```
0xdeafcafe, line 313: this is not traces v2

langwatch/src/pages/settings/data-retention.tsx
```diff
@@ -0,0 +1,1185 @@
...
+});
+
+const RETENTION_PRESETS: Array<{ value: string; label: string; days: number }> =
+  [
+    { value: "49", label: "7 weeks", days: 49 },
+    { value: "91", label: "3 months", days: 91 },
+    { value: "182", label: "6 months", days: 182 },
+    { value: "364", label: "1 year", days: 364 },
+    { value: "728", label: "2 years", days: 728 },
+    { value: "1820", label: "5 years", days: 1820 },
```
0xdeafcafe, line 100: for these i would add i block of 7 days over the limit. so for a year, instead of being 1 day less, i would be 6 days over, if you know what i mean?

langwatch/src/pages/settings/data-retention.tsx
```diff
@@ -0,0 +1,1185 @@
+import {
```
0xdeafcafe, line 1: can we break this down? business logic, constants, etc. being in `src/data-retention`/`src/settings/data-retention`, and then the component being cleaner

0xdeafcafe (approved):

langwatch/src/server/clickhouse/migrations/00032_add_retention_and_size_columns.sql
sergioestebance: Stale per ADR-022: cascade is now PROJECT→TEAM→ORG→49d (not indefinite), and new rows get the server-resolved value, not the column DEFAULT. The 308 only stamps pre-migration rows — intentional soft grace rather than an instant 49d retro-trim. Tracked in #4506.

langwatch/src/components/traces/TraceDetails.tsx
sergioestebance: Right — v2 overflow menu already has pin/unpin, this insert is redundant. Tracked in #4506.

langwatch/src/pages/settings/data-retention.tsx
sergioestebance: Switching to 49 / 91 / 182 / 371 / 735 / 1827 (round + ~7d buffer). Tracked in #4506.

langwatch/src/pages/settings/data-retention.tsx
sergioestebance: Agreed — 1185 lines mixing shell, constants, drawers, handlers. Tracked in #4506 with a constants/format/drawers split, page kept to shell + data.

PR #4155, opened 2026-05-21
chore(ci): sign release images with cosign and publish CycloneDX SBOMs

rogeriochaves:
## Summary

Closes the supply-chain gap on Docker Hub releases. Today the langwatch container images are pushed unsigned and without an SBOM, which fails procurement controls that require third-party software to be signed by an approved vendor and trackable to a verifiable bill of materials.

After this PR, each release of `langwatch/langwatch`, `langwatch/langwatch_nlp`, `langwatch/langevals`, and `langwatch/ai-gateway`:

- is signed with [Sigstore cosign](https://www.sigstore.dev/) using keyless OIDC, with the certificate identity tied to this repository and workflow,
- gets a CycloneDX JSON SBOM generated with syft and attached to the manifest digest as a cosign attestation (`--type cyclonedx`),
- has the same SBOM uploaded as a workflow artifact (90-day retention) and, for tagged releases, attached to the GitHub release as a `*.cdx.json` asset.

Verification example (added to the self-hosting security docs):

```bash
cosign verify langwatch/langwatch:<tag> \
  --certificate-identity-regexp '^https://github\.com/langwatch/langwatch/' \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com
```

## What changed

- `.github/workflows/publish-docker-app.yml`: new `sign-and-attest` job after `create-manifest`. Installs cosign + syft, logs into Docker Hub (read), resolves the multi-arch manifest digest for each image, signs by digest, generates the SBOM, attaches it as a cosign attestation, uploads SBOMs as workflow artifacts, and (for releases) uploads them to the GitHub release assets. Requires `id-token: write` at the job level for OIDC.
- `docs/self-hosting/security.mdx`: new "Supply Chain" section documenting the signing model and verification commands, plus a hardening-checklist entry for verifying image signatures at pull time.

## Notes

- Signing runs on every successful publish (release + workflow_dispatch). The GitHub-release-assets upload step is gated on `is_release == 'true'`.
- The per-arch images (`${HASH}-amd64`, `${HASH}-arm64`) are deleted by the existing `create-manifest` cleanup. Only the multi-arch manifest digest is signed, which is what consumers actually pull.
- Image-digest pinning in the Helm chart (e.g. `image: foo@sha256:...`) is the natural next step but is out of scope here; it requires changes to the release-please flow that writes tags into `values.yaml`.

## Test plan

- [ ] Run `publish-docker-app` via workflow_dispatch with a custom tag and confirm the `sign-and-attest` job succeeds for all four images.
- [ ] After the dispatch run, `cosign verify langwatch/langwatch:<tag> --certificate-identity-regexp '^https://github\.com/langwatch/langwatch/' --certificate-oidc-issuer https://token.actions.githubusercontent.com` returns the signature claim.
- [ ] `cosign download attestation langwatch/langwatch:<tag>` returns a CycloneDX predicate that decodes to valid JSON.
- [ ] Cut a draft release and confirm `*.cdx.json` files appear on the release assets list.
- [ ] `npm audit signatures` on the npm CLI still passes (unchanged code path, but worth re-confirming end-to-end provenance story).

### DIFF
--- a/.github/workflows/publish-docker-app.yml
+++ b/.github/workflows/publish-docker-app.yml
@@ -213,6 +213,10 @@ jobs:
       - name: Sign manifests and attach per-platform CycloneDX SBOMs
         env:
           HASH: ${{ needs.extract-version.outputs.short_hash }}
+          IS_RELEASE: ${{ needs.extract-version.outputs.is_release }}
+          VERSION: ${{ needs.extract-version.outputs.version }}
+          CERT_IDENTITY_REGEX: ^https://github\.com/langwatch/langwatch/
+          CERT_OIDC_ISSUER: https://token.actions.githubusercontent.com
         run: |
           set -euo pipefail
           mkdir -p sboms
@@ -227,6 +231,33 @@ jobs:
             fi
             echo "Signing multi-arch index langwatch/${IMAGE}@${INDEX_DIGEST}"
             cosign sign "langwatch/${IMAGE}@${INDEX_DIGEST}"
+
+            echo "Verifying index signature for langwatch/${IMAGE}@${INDEX_DIGEST}"
+            cosign verify "langwatch/${IMAGE}@${INDEX_DIGEST}" \
+              --certificate-identity-regexp "${CERT_IDENTITY_REGEX}" \
+              --certificate-oidc-issuer "${CERT_OIDC_ISSUER}" > /dev/null
+
+            # Release tags (VERSION + optionally latest) and the HASH tag should all point
+            # at the same multi-arch manifest digest, but assert it explicitly so that any
+            # future registry-side normalization that produces a different digest is signed
+            # rather than silently leaving an unsigned tag.
+            EXTRA_TAGS=""
+            if [ "${IS_RELEASE}" = "true" ]; then
+              EXTRA_TAGS="${VERSION}"
+              if ! echo "${VERSION}" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+-.+'; then
+                EXTRA_TAGS="${EXTRA_TAGS} latest"
+              fi
+            fi
+            for TAG in ${EXTRA_TAGS}; do
+              TAG_DIGEST=$(docker buildx imagetools inspect "langwatch/${IMAGE}:${TAG}" --format '{{ json . }}' | jq -r '.manifest.digest')
+              if [ "${TAG_DIGEST}" != "${INDEX_DIGEST}" ]; then
+                echo "::warning::langwatch/${IMAGE}:${TAG} digest ${TAG_DIGEST} differs from :${HASH} digest ${INDEX_DIGEST}; signing separately"
+                cosign sign "langwatch/${IMAGE}@${TAG_DIGEST}"
+                cosign verify "langwatch/${IMAGE}@${TAG_DIGEST}" \
+                  --certificate-identity-regexp "${CERT_IDENTITY_REGEX}" \
+                  --certificate-oidc-issuer "${CERT_OIDC_ISSUER}" > /dev/null
+              fi
+            done
 
             PLATFORMS_JSON=$(echo "${INSPECT}" | jq -c '[.manifest.manifests[] | select(.platform.os != "unknown") | {digest: .digest, os: .platform.os, arch: .platform.architecture}]')
             COUNT=$(echo "${PLATFORMS_JSON}" | jq 'length')
@@ -251,6 +282,15 @@ jobs:
 
               echo "Attesting SBOM for ${PDREF}"
               cosign attest --predicate "${SBOM_FILE}" --type cyclonedx "${PDREF}"
+
+              echo "Verifying signature and SBOM attestation for ${PDREF}"
+              cosign verify "${PDREF}" \
+                --certificate-identity-regexp "${CERT_IDENTITY_REGEX}" \
+                --certificate-oidc-issuer "${CERT_OIDC_ISSUER}" > /dev/null
+              cosign verify-attestation "${PDREF}" \
+                --type cyclonedx \
+                --certificate-identity-regexp "${CERT_IDENTITY_REGEX}" \
+                --certificate-oidc-issuer "${CERT_OIDC_ISSUER}" > /dev/null
             done
           done
 
--- a/.github/workflows/publish-docker-app.yml
+++ b/.github/workflows/publish-docker-app.yml
@@ -186,6 +186,71 @@ jobs:
             done
           done
 
+  sign-and-attest:
+    runs-on: ubuntu-latest
+    needs: [extract-version, create-manifest]
+    permissions:
+      contents: write # upload SBOMs to GitHub release assets
+      id-token: write # keyless OIDC signing via Sigstore
+    env:
+      COSIGN_YES: "true"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Sign manifests and attach CycloneDX SBOMs
+        env:
+          HASH: ${{ needs.extract-version.outputs.short_hash }}
+        run: |
+          set -euo pipefail
+          mkdir -p sboms
+          for IMAGE in langwatch langwatch_nlp langevals ai-gateway; do
+            REF="langwatch/${IMAGE}:${HASH}"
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-docker-app.yml
```diff
@@ -186,6 +186,71 @@ jobs:
             done
           done
 
+  sign-and-attest:
+    runs-on: ubuntu-latest
+    needs: [extract-version, create-manifest]
```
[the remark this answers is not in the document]
rogeriochaves: Good catch, fixed in 3a73cc5e7: `notify` now lists `sign-and-attest` in `needs`, so Slack only announces a release after signatures and SBOMs are in place.

.github/workflows/publish-docker-app.yml
```diff
@@ -186,6 +186,71 @@ jobs:
...
+            DIGEST_REF="langwatch/${IMAGE}@${DIGEST}"
+            echo "Signing ${DIGEST_REF}"
+            cosign sign "${DIGEST_REF}"
+
+            SBOM_FILE="sboms/${IMAGE}.cdx.json"
+            echo "Generating CycloneDX SBOM for ${DIGEST_REF}"
+            syft "registry:${DIGEST_REF}" -o cyclonedx-json="${SBOM_FILE}"
+
+            echo "Attesting SBOM for ${DIGEST_REF}"
+            cosign attest --predicate "${SBOM_FILE}" --type cyclonedx "${DIGEST_REF}"
```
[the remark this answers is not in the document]
rogeriochaves: Valid, fixed in 3a73cc5e7. The workflow now enumerates each platform manifest from the multi-arch index and runs syft + cosign sign + cosign attest against each per-platform digest directly (`langwatch/<image>@<platform-digest>`). The index digest itself is still signed for completeness. SBOMs are written as `sboms/<image>-<os>-<arch>.cdx.json` and uploaded both as workflow artifacts and as GitHub release assets.

.github/workflows/publish-docker-app.yml
```diff
@@ -186,9 +186,93 @@ jobs:
...
+            REF="langwatch/${IMAGE}:${HASH}"
+            INSPECT=$(docker buildx imagetools inspect "${REF}" --format '{{ json . }}')
+
+            INDEX_DIGEST=$(echo "${INSPECT}" | jq -r '.manifest.digest')
+            if [ -z "${INDEX_DIGEST}" ] || [ "${INDEX_DIGEST}" = "null" ]; then
+              echo "::error::Could not resolve index digest for ${REF}"
+              exit 1
+            fi
+            echo "Signing multi-arch index langwatch/${IMAGE}@${INDEX_DIGEST}"
+            cosign sign "langwatch/${IMAGE}@${INDEX_DIGEST}"
```
Aryansharma28, line 233: **[non-blocking — but worth hardening]**

The loop resolves the index digest from `langwatch/${IMAGE}:${HASH}` and signs it. For releases, `create-manifest` also publishes `langwatch/${IMAGE}:${VERSION}` and `langwatch/${IMAGE}:latest` pointing at the same source platform images.

`imagetools create` with identical inputs *should* produce byte-identical manifest JSON → identical SHA256 → all three tags share one digest, so one signature covers them all. In practice this holds, but it is an implicit assumption: if Docker Hub ever normalises or annotates the manifest on write, the digests diverge silently and `cosign verify langwatch/langwatch:v1.2.3` starts returning "no signatures found" for every future release.

Suggested hardening — after signing the index, assert the other published tags resolve to the same digest, and sign separately if they differ:

```yaml
# Assert (and sign if diverged) version + latest tags for releases
if [ "${{ needs.extract-version.outputs.is_release }}" = "true" ]; then
  VERSION="${{ needs.extract-version.outputs.version }}"
  EXTRA_TAGS="${VERSION}"
  if ! echo "${VERSION}" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+-.+'"; then
    EXTRA_TAGS="${EXTRA_TAGS} latest"
  fi
  for TAG in ${EXTRA_TAGS}; do
    TAG_DIGEST=$(docker buildx imagetools inspect "langwatch/${IMAGE}:${TAG}" \
      --format '{{ json . }}' | jq -r '.manifest.digest')
    if [ "${TAG_DIGEST}" != "${INDEX_DIGEST}" ]; then
      echo "::warning::langwatch/${IMAGE}:${TAG} digest (${TAG_DIGEST}) != :${HASH} digest — signing separately"
      cosign sign "langwatch/${IMAGE}@${TAG_DIGEST}"
    fi
  done
fi
```

This costs one `imagetools inspect` call per extra tag per image and turns a silent divergence into a visible warning + correct signature.

Aryansharma28 (commented):
Good supply-chain work overall — SHA-pinned actions, per-platform SBOMs, and gating the Slack notify on signing are all the right calls. Two gaps worth closing before this is truly bulletproof: (1) the signing loop only runs against the `${HASH}` tag; release consumers verify `:v1.2.3` and `:latest`, which need to resolve to the same digest for `cosign verify` to succeed. (2) There's no self-verification step — if an attestation silently fails to push, the job still exits 0 and you won't know until a user reports it.

.github/workflows/publish-docker-app.yml
```diff
@@ -186,9 +186,93 @@ jobs:
...
+              echo "Generating CycloneDX SBOM for ${PDREF} (${POS}/${PARCH})"
+              syft "registry:${PDREF}" -o cyclonedx-json="${SBOM_FILE}"
+
+              echo "Signing platform manifest ${PDREF}"
+              cosign sign "${PDREF}"
+
+              echo "Attesting SBOM for ${PDREF}"
+              cosign attest --predicate "${SBOM_FILE}" --type cyclonedx "${PDREF}"
+            done
+          done
```
Aryansharma28, line 295: **[non-blocking — but closes an important feedback loop]**

After the signing loop exits there is no self-verification step. If `cosign attest` pushes the attestation to a path that cannot be resolved later (wrong registry, OCI tag collision, transient Rekor error that cosign treated as non-fatal), the job exits 0 and you only find out when an operator runs `cosign verify` and it fails in prod.

Add a verify pass at the end of the loop — it is fast (no image pull needed) and turns a silent failure into an immediate CI failure:

```yaml
echo "Verifying index signature for langwatch/${IMAGE}@${INDEX_DIGEST}"
cosign verify "langwatch/${IMAGE}@${INDEX_DIGEST}" \
  --certificate-identity-regexp "^https://github\.com/langwatch/langwatch/" \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com

echo "Verifying attestation for ${PDREF}"
cosign verify-attestation "${PDREF}" \
  --type cyclonedx \
  --certificate-identity-regexp "^https://github\.com/langwatch/langwatch/" \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
| jq -r ".payload | @base64d | fromjson | .predicateType"
```

Place the index verify right after line 229, and the attestation verify inside the platform loop right after `cosign attest`.

Aryansharma28, line 295: **[non-blocking — but closes an important feedback loop]**

After the signing loop exits there is no self-verification step. If `cosign attest` pushes the attestation to a path that cannot be resolved later (wrong registry, OCI tag collision, transient Rekor error that cosign treated as non-fatal), the job exits 0 and you only find out when an operator runs `cosign verify` and it fails in prod.

Add a verify pass at the end of the loop — it is fast (no image pull needed) and turns a silent failure into an immediate CI failure:

```yaml
echo "Verifying index signature for langwatch/${IMAGE}@${INDEX_DIGEST}"
cosign verify "langwatch/${IMAGE}@${INDEX_DIGEST}" \
  --certificate-identity-regexp "^https://github\.com/langwatch/langwatch/" \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com

echo "Verifying attestation for ${PDREF}"
cosign verify-attestation "${PDREF}" \
  --type cyclonedx \
  --certificate-identity-regexp "^https://github\.com/langwatch/langwatch/" \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
| head -c 200
```

Place the index verify right after line 229, and the attestation verify inside the platform loop right after `cosign attest`.

.github/workflows/publish-docker-app.yml
rogeriochaves: Implemented in 38d3f31f1. After signing the index digest, the workflow now (a) runs `cosign verify` against the same digest as a self-check, and (b) for releases, inspects `langwatch/<image>:<version>` and `latest` (when not a prerelease), asserts each resolves to the same digest as the `:<hash>` tag we just signed, and logs a warning + signs separately if a future registry normalization ever causes them to diverge.

.github/workflows/publish-docker-app.yml
rogeriochaves: Implemented in 38d3f31f1. The workflow now self-verifies after every `cosign sign` and `cosign attest` against the same digest with the expected certificate identity (`^https://github\.com/langwatch/langwatch/`) and OIDC issuer (`https://token.actions.githubusercontent.com`). A silent path-resolution failure, tag collision, or non-fatal Rekor error now causes the job to fail rather than surfacing only when an operator runs `cosign verify` in prod.

rogeriochaves: Implemented in 38d3f31f1. The workflow now self-verifies after every `cosign sign` and `cosign attest` against the same digest with the expected certificate identity (`^https://github\.com/langwatch/langwatch/`) and OIDC issuer (`https://token.actions.githubusercontent.com`). A silent path-resolution failure, tag collision, or non-fatal Rekor error now causes the job to fail rather than surfacing only when an operator runs `cosign verify` in prod.

Aryansharma28 (approved):
lgtm

PR #4161, opened 2026-05-21
fix(groupQueue): reduce dispatch Lua scan budget 5x to cut Redis CPU

sergioestebance:
## Summary

- **DISPATCH_LUA**: scan budget reduced from 50,000 → 10,000
- **DISPATCH_BATCH_LUA**: scan budget reduced from `pageSize * 250` → `pageSize * 50`

Both are a 5x reduction. Redis Engine CPU is at 100%, with `DISPATCH_BATCH_LUA` consuming 73%. Most scan work is wasted skip-checks (SISMEMBER + GET) on capped tenants. Dispatchers loop continuously so no work is missed — just processed in smaller, cheaper bites.

## Test plan

- [ ] Deploy and monitor Redis Engine CPU — expect ~5x drop in per-EVAL cost
- [ ] Verify dispatch throughput remains stable (dispatchers loop, so groups are still processed)
- [ ] Watch for any starvation of non-capped tenants under load


drewdrewthis (approved):
lgtm

PR #4167, opened 2026-05-21
fix(groupQueue): increase reactor dedup TTLs + counter conservation tests

sergioestebance:
## Summary

- Increases reactor dedup TTLs to reduce Redis fan-out pressure during traffic bursts
- Adds 8 integration tests pinning the `total-pending` counter invariant across every lifecycle path

## Reactor TTL Changes (reduces Redis pressure)

| Reactor | Before | After | Why safe |
|---------|--------|-------|----------|
| `traceUpdateBroadcast` | 1s | **30s** | Notification only — frontend already debounces |
| `spanStorageBroadcast` | 1s | **15s** | Notification only — frontend refetches |
| `originGate` | 5s | **15s** | Downstream action is a 5-min deferred timer; 10s extra is negligible |

For a 30-span trace, these three reactors were creating ~72 separate jobs instead of ~3. With the new TTLs, dedup squashes them during the span burst window, cutting reactor lifecycle Redis commands by ~60%.

## Counter Conservation Tests (Phase 0)

| Test | What it proves |
|------|---------------|
| happy path (stage→dispatch→complete) | Baseline: counter works when nothing goes wrong |
| active key expiry before COMPLETE | Documents the drift bug: counter leaks +1 |
| retryRestage lifecycle | Single retry doesn't leak the counter |
| restageAndBlock + unblock lifecycle | Block/unblock cycle doesn't leak |
| 3 retries then success | Multi-hop retries don't accumulate drift |
| naive DECR-at-dispatch proof | Proves the 2-script fix creates -1 drift per retry |
| reconciliation invariant (no retries) | Counter equals actual Redis state |
| retryRestage breaks reconciliation | Documents that retry adds job without INCR (by design) |

## Test plan

- [x] `pnpm test:integration` — 97/97 pass
- [x] `pnpm test:unit` originGate unit test updated
- [ ] CI green
- [ ] Deploy and monitor Redis Engine CPU — expect lower peak during tenant burst

### DIFF
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/originGate.reactor.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/originGate.reactor.unit.test.ts
@@ -111,7 +111,7 @@ describe("originGate reactor", () => {
       const reactor = createOriginGateReactor(deps);
 
       expect(reactor.options?.delay).toBe(5_000);
-      expect(reactor.options?.ttl).toBe(5_000);
+      expect(reactor.options?.ttl).toBe(15_000);
     });
 
     it("generates dedup key from tenant and trace", () => {
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/originGate.reactor.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/originGate.reactor.ts
@@ -40,7 +40,7 @@ export function createOriginGateReactor(
     options: {
       makeJobId: (payload) =>
         `origin-gate:${payload.event.tenantId}:${payload.event.aggregateId}`,
-      ttl: 5_000,    // 5s dedup — debounce the initial span burst
+      ttl: 15_000,   // 15s dedup — debounce multi-span trace bursts
       delay: 5_000,  // 5s delay — settle before checking origin
     },
 
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/spanStorageBroadcast.reactor.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/spanStorageBroadcast.reactor.ts
@@ -29,7 +29,7 @@ export function createSpanStorageBroadcastReactor(
       disabled: deps.hasRedis === false,
       makeJobId: (payload) =>
         `span-stored:${payload.event.tenantId}:${payload.event.aggregateId}`,
-      ttl: 1000,
+      ttl: 15_000, // Debounce — notification only, frontend refetches
     },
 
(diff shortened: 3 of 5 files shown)


sergioestebance:
## Live Redis Analysis — Active Recurrence (May 21 ~21:15 UTC)

Redis Engine CPU is climbing toward saturation again. Same pattern as the May 21 incident, same tenant, same fan-out mechanism.

### Current State

| Metric | 30 min ago | Now | Trend |
|--------|-----------|-----|-------|
| **Ready groups** | 272 | **2,022** | ↑ climbing |
| **Past-due (backlogged)** | 0 | **1,816** | ↑ all one tenant |
| **Active (processing)** | 272 | **190** | stable |
| **Redis ops/sec** | 9,169 | **556,123** | ↑ 60x increase |
| **Engine CPU** | ~5% | **~85%** | ↑ heading toward 100% |

### Root Cause (same as original incident)

One tenant is ingesting a burst of complex traces. The fan-out multiplier creates ~15 reactor groups per trace. With ~120 traces in the burst: **120 × 15 = ~1,800 reactor groups**.

The tenant is at the 50-slot cap. The dispatcher scans all 1,816 queued groups on every cycle, skipping 1,766 of them (over cap). Each skipped group costs ~4 Redis commands. At 4 pods × continuous dispatch cycles = 556K ops/sec on the single Redis thread.

### Why These TTL Changes Help

The three reactors with short TTLs account for the bulk of the fan-out:

| Reactor | Old TTL | New TTL | Groups per 120-trace burst (old) | Groups (new) |
|---------|---------|---------|--------------------------------|-------------|
| traceUpdateBroadcast | 1s | **30s** | ~120 × 30 spans = 3,600 | ~120 |
| spanStorageBroadcast | 1s | **15s** | ~120 × 30 spans = 3,600 | ~240 |
| originGate | 5s | **15s** | ~120 × 12 = 1,440 | ~240 |
| **Subtotal** | | | **~8,640** | **~600** |

During normal operation (no backlog), dedup barely fires because jobs dispatch too fast for the TTL to catch them — **no user-visible change**. During a backlog like right now, jobs sit in staging and the extended TTL squashes them instead of creating new groups. This is exactly when we need the reduction.

### Impact on Other Tenants

The 1,816 groups from one tenant clog the dispatcher scan for ALL tenants. Every other tenant's jobs wait behind this wall. Fewer reactor groups = shorter scan = other tenants get dispatched sooner.

### What These Changes Do NOT Fix

- The fan-out multiplier for the other ~7 reactors (evaluationTrigger, alertTrigger, etc. — already have 30-60s TTLs)
- The `customEvaluationSync` broken dedup (per-event key, never matches)
- The dispatcher scan cost (still O(readySize) per cycle with tenant cap)
- Client-side dedup (would eliminate the STAGE_LUA EVAL cost entirely — Layer 2 in the debounce investigation)

PR #4166, opened 2026-05-21
[DO NOT MERGE] coderabbit-validation: one violation per rule (#3754 AC6)

drewdrewthis:
**THIS PR EXISTS ONLY TO PROVE THE RULES FIRE — DO NOT MERGE.**

Smoke test for [#3754](https://github.com/langwatch/langwatch/issues/3754) AC6. Each violation file in this diff intentionally trips one rule from PR #4162. The PR is opened against the feature branch so the new \`.coderabbit.yaml\` / ast-grep / semgrep / GH Action config is active on the review.

I am deliberately omitting the \`## Deployment Impact\` section here so the \`.github/workflows/deployment-impact-check.yml\` workflow also fails — that's part of the smoke test (the PR touches \`charts/langwatch/templates/smoke-test-deployment-impact.yaml\`).

## File-to-rule mapping (verdicts get pasted back into #4162)

| File | Rule | CodeRabbit verdict |
|---|---|---|
| `coderabbit-validation/${UNEXPANDED}/leak.txt` | path_instructions: \`${...}\` structural guard | _pending_ |
| `langwatch/src/pii-in-logger-violation.ts` | semgrep `pii-in-logger-call` | _pending_ |
| `langwatch/src/pii-literal-api-key-violation.ts` | semgrep `pii-literal-api-key` + gitleaks | _pending_ |
| `langwatch/src/syntactic-violations.ts` | ast-grep `no-explicit-any`, `no-inline-dynamic-import`, `no-localhost-fallback` (both literal and template-literal forms) | _pending_ |
| `langwatch/src/components/FormWatchInChildViolation.tsx` | ast-grep `no-form-watch-in-child` (arrow + function-decl + bare-param) | _pending_ |
| `langwatch/src/export-star-shim-violation.ts` | ast-grep `no-export-star-shim` | _pending_ |
| `langwatch/src/judgment-violations.ts` | path_instructions: missing tRPC Zod, class component, service→component import | _pending_ |
| `langwatch/src/server/clickhouse/migrations/9999_qualified_table_violation.sql` | semgrep `clickhouse-no-qualified-table` (both \`${CLICKHOUSE_DATABASE}.<t>\` and \`langwatch.<t>\` in DDL) | _pending_ |
| `langwatch/src/server/migrations/9999_irreversible_violation.sql` | path_instructions: migration not reversible | _pending_ |
| `langwatch/src/big-file-violation.ts` | path_instructions: file > 300 lines (SRP, AC7) | _pending_ |
[...]

Issue #3754, opened 2026-05-03
Customize CodeRabbit with a committed .coderabbit.yaml


drewdrewthis:
## Why

CodeRabbit currently runs on dashboard defaults — summaries and broad suggestions, but no opinionated enforcement of how we actually want code written. We want it stricter: catching PII leaks, calling out style drift, blocking obvious bugs.

Triggered by PR #3476 leaking a literal `${HOME/...}` directory into `main` that CR happily passed.

## Acceptance Criteria

> [!IMPORTANT]
> Investigation surfaced four scope-touching flips. **Read the Investigation section before working these ACs** — AC1, AC3, AC5 and the Non-goals all need to change. Concrete proposed AC text is in the Investigation section under "Proposed AC changes."

- [ ] **AC1:** `.coderabbit.yaml` committed at repo root, valid against the v2 schema, with `reviews.request_changes_workflow: true` so opinionated checks actually block merge.

- [ ] **AC2:** Strong PII rules via `path_instructions` — review fails when:
  - Logging/analytics calls include user-identifying fields (email, phone, name, IP, userId) without explicit redaction.
  - New endpoints return PII without auth + audit-log wrappers.
  - String literals match email / SSN / credit-card / API-key regex outside fixtures.
  - `gitleaks` and a PII `semgrep` ruleset enabled in `reviews.tools`.

- [ ] **AC3:** Strong coding-standards rules via `path_instructions`. Starter set, refine during PR — but each must be **specific and concrete**, not vague:
  - No `any` in new TypeScript; prefer `unknown` + narrowing.
  - tRPC routers require explicit Zod input validation.
  - Service-layer code does not import from `~/components`.
  - New React components are function components.
  - Tests use Vitest, not Jest.
  - Migrations are reversible or document why not.
  - No inline `import(...)` in `.ts`/`.tsx` — use top-level `import` / `import type`.
  - ClickHouse migrations use unqualified table names — no `${CLICKHOUSE_DATABASE}.` prefix.
  - No `?? "http://localhost..."` fallbacks — required env + Zod validation.
  - No re-export shims (`export * from "..."`) for backwards compatibility — update consumers to import from the new location.
  - Hono routes use the service layer, not repositories directly.
  - Repository methods use `findAll`/`findById`; services use `getAll`/`getById`.
  - `form.watch()` not allowed in child components receiving `form` as prop — use `useWatch({ control, name })`.

[...]

PR #4170, opened 2026-05-21
[DO NOT MERGE] AC6 smoke v2: validate previously-silent rules (#3754)

drewdrewthis:
## What

Re-runs the AC6 smoke test against the updated `.coderabbit.yaml` + ast-grep
rules in #4162 (commit fffc04b53), which fixed three silent rules and one
path-glob mismatch surfaced by the first smoke test (#4166).

## Fixes applied since first smoke (#4166)

1. **`no-explicit-any`** — `kind: any_type` was invalid in ast-grep 0.42.x.
   Replaced with `kind: predefined_type` + `regex: '^any$'`. Split into
   `no-explicit-any.yml` (ts) + `no-explicit-any-tsx.yml` (tsx) because
   `language: TypeScript` does not dispatch the tsx parser.
2. **`no-export-star-shim`** — `pattern: export * from $MOD` was rejected
   as "multiple AST nodes". Replaced with `kind: export_statement` +
   `regex: '^export \* from'` + `not.follows` on a `@barrel` comment.
   Split ts/tsx.
3. **`no-inline-dynamic-import`** — syntactically valid, but likely went
   silent because companion rules' parse errors cascaded the `rule_dirs`
   load. Now split ts/tsx for parser dispatch.
4. **service→component layering** — path-glob `langwatch/src/**/*.{ts,tsx}`
   bundled with 3 other rules in one path_instructions block caused CR
   to surface only one. Split into focused blocks; service→component now
   uses `langwatch/src/{server,services}/**/*.{ts,tsx}` so CR consults
   it on the files that violate the boundary.

## Expected CR behavior

Same 16-category smoke matrix as #4166, but now we expect:

- `no-explicit-any` ✅ — `langwatch/src/syntactic-violations.ts:8-10`
- `no-inline-dynamic-import` ✅ — `langwatch/src/syntactic-violations.ts:14`
- `no-export-star-shim` ✅ — `langwatch/src/export-star-shim-violation.ts:6`
- service→component layering ✅ — `langwatch/src/judgment-violations.ts`
  (service file importing `~/components`)

Plus the 12 categories that fired correctly the first time should still
fire (PII logger, PII API-key literal, localhost fallback both shapes,
form.watch in 3 child shapes, tRPC Zod, class component, ClickHouse
qualified table, irreversible migration, big-file >300 lines, structural
`${...}` guard, deployment-impact workflow).

## After CR review lands

I'll map the verdicts back into PR #4162 body and close this PR.


drewdrewthis:
@coderabbitai review

drewdrewthis:
AC6 smoke v2 closed.

CR posted 9 actionable + 6 nitpick comments across 19 review threads at 22:29:02Z. Verdict map pasted into PR #4162 body.

**Key finding**: 2 of 4 previously-silent rules are now FIXED — `no-explicit-any` and `no-inline-dynamic-import` both fired on `langwatch/src/syntactic-violations.ts`. CR also caught the new tightened service→component glob directly: "the file is located at langwatch/src/judgment-violations.ts… the rule uses glob langwatch/src/{server,services}/**/*.{ts,tsx}, which requires files to be under langwatch/src/server/" — meaning the rule is correct, the smoke fixture is in the wrong directory.

One follow-up: `no-export-star-shim` is still silent despite local ast-grep 0.42.3 validating the rule fires on the fixture. Needs second-pass diagnosis in #4162 follow-ups.

Closing as `[DO NOT MERGE]` smoke validator — never intended to merge.

PR #4037, opened 2026-05-13
fix(prompts): clamp Deploy dialog Select width to row (#4006)

drewdrewthis:
## Summary

Fixes #4006 — production/staging Select inputs in the Deploy prompt modal overflowed the modal width when version commit messages were long.

Root cause: `<Select.Root>` used `width="auto" minWidth="180px"`, and its parent `HStack` had no horizontal budget. The trigger grew to fit the longest commit message and pushed the row past the modal edge.

Fix:
- Inner content `HStack` is now `flex=1 minWidth=0 justify="flex-end"` so it can shrink to the available row width.
- `Select.Root` is `flex=1 minWidth=0 maxWidth="280px"` so the trigger truncates the commit-message text (via the existing `Text…truncate`) instead of growing.
- The name/dot cluster is `flexShrink=0` so it never shrinks past its intrinsic width.

No behavioral change to version selection, save flow, or tag management. `pnpm typecheck` clean.

## Spec

Added `@integration @manual` scenario to `specs/prompts/deploy-prompt-dialog.feature`:
"Version Select inputs stay within the modal width" — asserted visually since jsdom doesn't compute layout.

## Test plan

- [x] `pnpm typecheck` passes
- [ ] Manual: open Deploy prompt modal on a config with multiple versions whose commit messages exceed ~280px and confirm each row stays inside the modal boundary
- [ ] Screenshot attached to this PR per the issue's Proof acceptance criterion

🤖 Generated with [Claude Code](https://claude.com/claude-code)

# Related Issue

- Resolve #4006

PR #4172, opened 2026-05-22
test(prompts): add regression test for Deploy dialog Select overflow fix

drewdrewthis:
## Summary

- Follow-up to #4037, which fixed the `DeployPromptDialog` layout by adding `flex="1" minWidth={0} justify="flex-end"` to the inner HStack and `flex="1" minWidth={0} maxWidth="280px"` to each `Select.Root` in environment tag rows
- Implements the `@unimplemented` BDD scenario that was added as a placeholder in that PR
- Removes `@manual` and `@unimplemented` tags from the scenario in `specs/prompts/deploy-prompt-dialog.feature` — it is now a real, bound test
- Adds a jsdom integration test in `DeployPromptDialog.integration.test.tsx` that verifies the anti-overflow flex props (`flex="1"`, `maxWidth="280px"`) are present on every `Select.Root` rendered in tag rows, even when versions have very long commit messages (220+ characters)

## How the test works

jsdom does not compute layout, so pixel widths cannot be asserted. Instead:

1. The `Select.Root` mock is updated to forward `flex` and `maxWidth` props as `data-flex` and `data-max-width` HTML attributes on the container div
2. The test renders the dialog with a version whose commit message is 220 characters long
3. It queries all `[data-testid="select-root"]` elements and asserts each has `data-flex="1"` and `data-max-width="280px"` — confirming the props from the fix are still being passed through

This guards against the overflow regression being accidentally removed in future refactors.

## Test plan

- [ ] `pnpm test:integration DeployPromptDialog` passes with the new `it()` block
- [ ] The BDD scenario in `specs/prompts/deploy-prompt-dialog.feature` no longer carries `@unimplemented`

### DIFF
--- a/langwatch/src/prompts/components/__tests__/DeployPromptDialog.integration.test.tsx
+++ b/langwatch/src/prompts/components/__tests__/DeployPromptDialog.integration.test.tsx
@@ -105,7 +105,11 @@ vi.mock("~/components/ui/select", () => {
     "aria-label"?: string;
     [key: string]: unknown;
   }) => (
-    <div data-testid="select-root">
+    <div
+      data-testid="select-root"
+      data-flex={rest["flex"] as string | undefined}
+      data-max-width={rest["maxWidth"] as string | undefined}
+    >
       <select
         aria-label={rest["aria-label"] as string}
         value={value[0] ?? ""}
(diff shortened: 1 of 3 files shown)


drewdrewthis:

CI: green — zero failing, zero pending (109 checks pass/success/skip)
ACs: met — `@scenario` bound test runs in CI (standalone file, outside `describe.skip`)
Threads: zero unresolved

Verified by:
  \`gh pr checks 4172 | grep -iE "fail"\` → (empty)
  \`gh pr checks 4172 | grep -cE "pass|success|skip"\` → 109
  feature-parity: test at \`DeployPromptDialog.overflow.unit.test.tsx\` binds the scenario via \`/** @scenario Version Select inputs stay within the modal width */\`

PR #4179, opened 2026-05-22
feat(traces-v2): interactive error chip popover + blue glow on header-chip jumps

rogeriochaves:
## Summary

Customer report on the trace drawer: clicking the red \`Error\` chip in the header "doesn't really take you anywhere". It actually did switch to the Summary tab and select the first error span, but:

1. The Exceptions accordion sat lower in the panel so the operator landed on whatever was already in view (usually IO / Metadata).
2. Nothing visually pointed at the section that just opened.
3. The static tooltip said "Jump to the first span with an error" which wasn't what the chip actually did.

Same complaint applied to the Eval chip: it scrolled the Evals section into view but the operator's eye didn't track the change in a busy panel.

### What changed

- **Interactive popover on the Error chip.** Hovering opens a HoverCard mirroring the full Exceptions accordion - same red error block + same per-span pill row sourced from the new shared \`rankedErrorSpans\` helper. Click a span inside the popover to jump directly to it; click the chip itself to open the full Exceptions section.
- **Blue glow on header-chip jumps.** Extended \`useFocusSectionStore\` with an \`exceptions\` target so the error chip uses the same focus pipeline as the eval chips. After the auto-scroll lands, the targeted section briefly pulses with a 1.5s blue ring (\`SectionFocusGlow\`) that mirrors the onboarding \`DrawerGlow\` palette - same affordance, shorter run, scoped inside the section box.
- **Shared exceptions block.** Extracted \`ExceptionsContent\` so the accordion section and the chip popover render from one component. The two surfaces can't drift on span order, copy, or red treatment.
- **Dogfood emitter.** \`scripts/dogfood-error-trace.ts\` posts one trace via OTLP with a workflow + chain + 2 retry LLM spans all flagged \`status=error\` - enough surface area to exercise the popover + glow.

## Screenshots

Interactive popover anchored to the Error chip (matches the Exceptions section content):
popover

Click the chip and the Exceptions section pulses blue once the auto-scroll lands. The glow is painted as a fixed-position overlay through a Portal so the drawer's scroll-viewport overflow and the sticky section header can't crop it on any edge:
exceptions-glow

Same pipeline pulses the Evals section when an eval chip is clicked:
evals-glow

## Test plan

- [x] Unit: \`rankedErrorSpans\` ranks deepest-first, breaks ties by startTimeMs, treats out-of-trace parents as depth 0, returns empty for non-error inputs.
- [x] Unit: \`useFocusSectionStore.request\` publishes the right pending payload for both \`exceptions\` and \`evals\`, and bumps the nonce on repeat clicks so the observer effect re-fires.
- [x] Spec: \`specs/trace-drawer/header-chip-deeplinks.feature\` covers click-to-focus for both chips and the glow shape contract (~1.5s, single pulse, doesn't loop).
- [x] Manual QA on a freshly-emitted dogfood trace (\`scripts/dogfood-error-trace.ts\`):
  - Hover the red Error chip opens the popover with 5 span buttons.
  - Click the chip - the Summary tab opens, Exceptions accordion expands, and the section pulses blue.
  - Click an eval chip - the Evals accordion expands and pulses blue.
- [x] Spec parity check (\`pnpm tsx scripts/check-feature-parity.ts\`) green: 1303 enforced scenarios bound.

### DIFF
--- a/scripts/dogfood-error-trace.ts
+++ b/scripts/dogfood-error-trace.ts
@@ -1,0 +1,195 @@
+/* eslint-disable */
+/**
+ * Emit one OTLP trace with a chain of error spans, intended for the
+ * StatusChip interactive-tooltip dogfood. Mirrors the screenshot the
+ * customer report used to motivate the work — a top-level workflow
+ * span errors with "Connection error.", and a couple of llm + chain
+ * leaf spans error underneath so the popover has multiple chips.
+ *
+ * Usage:
+ *   LW_API_KEY=sk-lw-... LW_ENDPOINT=http://localhost:5560 \
+ *     npx tsx scripts/dogfood-error-trace.ts
+ */
+import { randomBytes } from "node:crypto";
+
+const ENDPOINT =
+  process.env.LW_ENDPOINT ?? "http://localhost:5560";
+const API_KEY = process.env.LW_API_KEY;
+if (!API_KEY) {
+  console.error("LW_API_KEY is required");
+  process.exit(1);
+}
+
+const NOW_NS = BigInt(Date.now()) * 1_000_000n;
+const SEC_NS = 1_000_000_000n;
+
+function id(bytes: number): string {
+  return randomBytes(bytes).toString("hex");
+}
+
+const traceId = id(16);
+const root = id(8);
+const llmChat = id(8);
+const llmRetry = id(8);
+const chain = id(8);
+const chainInner = id(8);
+
+interface SpanInput {
+  spanId: string;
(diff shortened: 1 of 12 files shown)


scripts/dogfood-error-trace.ts
```diff
@@ -0,0 +1,195 @@
+/* eslint-disable */
```
rogeriochaves, line 1: This file should be organized inside the scripts/dogfood folder not at the root

PR #4184, opened 2026-05-23
fix(metrics): read backup status from system.backup_log

rogeriochaves:
## Summary

Today an alert fired in lw-prod: **"ClickHouse Backup Reporting Absent"** stayed firing for 2h10m after a deploy + IAM-rotation rollout, even though backups themselves were running fine. Root cause is in `metrics.ts`:

- The collector queries `system.backups` to populate `clickhouse_backup_last_success_timestamp_seconds`.
- `system.backups` is an **in-memory** CH table — it gets wiped on every CH restart.
- CH restarts on every app deploy (the image tag in SSM bumps on every build), so the table is empty for ~1h between restart and the next scheduled backup.
- During that window a freshly-rolled worker pod sees zero rows and never calls `setClickHouseBackupLastSuccessTimestamp`. The gauge is lazy-registered (intentional — see the existing comment), so the metric stays absent entirely until a backup row appears, tripping the absent-gauge alert.

The fix swaps to `system.backup_log`, the persistent equivalent. Same columns we care about (`status`, `end_time`, `total_size`), survives CH restarts, has months of history available.

Validated on lw-prod: `system.backup_log` is enabled with 3539 rows back to 2026-01-29.

## Test plan

- [x] `pnpm test:unit src/server/clickhouse/__tests__/metrics.unit.test.ts` — 31/31 pass
- [ ] After merge, watch `clickhouse_backup_last_success_timestamp_seconds` in Grafana through the next prod deploy: gauge should remain present across the CH/worker rollout instead of disappearing for ~1-2h
- [ ] Confirm the "ClickHouse Backup Reporting Absent" alert no longer flaps on deploy

### DIFF
--- a/langwatch/src/server/clickhouse/__tests__/metrics.unit.test.ts
+++ b/langwatch/src/server/clickhouse/__tests__/metrics.unit.test.ts
@@ -223,65 +223,71 @@ describe("ClickHouse metrics", () => {
       );
     });
 
-    it("queries system.backup_log for backup status", async () => {
-      const partsResult = {
-        json: vi.fn().mockResolvedValue({ data: [] }),
-      };
-      const backupResult = {
-        json: vi.fn().mockResolvedValue({
-          data: [
-            {
-              status: "BACKUP_CREATED",
-              cnt: "3",
-              last_success_time: "2024-01-15 10:00:00",
-              last_success_size: "1073741824",
-            },
-          ],
-        }),
-      };
-      const diskResult = {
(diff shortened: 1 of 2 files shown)


rogeriochaves:
@coderabbitai addressed in fa14eb9 — wrapped the new backup_log tests in nested `describe('given ...')` / `describe('when ...')` blocks. Left the existing flat tests above unchanged (they predate the BDD rule and are out of scope for this PR — happy to ship a separate restructure PR if desired). All 31 unit tests still pass locally.

PR #4058, opened 2026-05-14
feat(stored-objects): externalize event byte content to stored_objects (#4053)

drewdrewthis:
## Why

Closes #4053. Voice and media-rich agent traces were embedding raw base64 bytes inside event payloads, blowing past ClickHouse row-size limits and choking the projection pipeline. This PR externalizes those bytes to S3-backed stored-objects and rewrites the in-event content to point at `/api/files/<id>` references.

## What changed

- **New stored-objects subsystem** (`langwatch/src/server/stored-objects/`) — pluggable `StorageDriver` registry behind a `StoredObjectsService`, with S3, Azure Blob, and local-filesystem drivers. Content-addressed by deterministic `@langwatch/ksuid` so re-uploads dedupe.
- **Event byte extraction** (`content-extractor.ts`) — walks `chatRichContentSchema` (langwatch's tracer content vocabulary, not `@ag-ui/core`'s aspirational shape) and rewrites `input_audio.data` / `image_url` data URIs into `/api/files/<id>` URLs. Handles Python-repr stringified arrays the python-sdk emits.
- **Shared content-part visitor** (`visit-content-part.ts`) — single decoder used by both the server-side extractor and the client-side renderer, so adding a new content shape is one change, not shotgun surgery.
- **Auth + plumbing** — `dualAuth` middleware extracted into a shared module; project-permission check now routes through a `PermissionsService` (constructor-injected `PrismaClient`) at `app-layer/permissions/`.
- **Dev ergonomics** — `make quickstart dev-storage` preset bakes overrides that point local stacks at the real AWS `runtime-storage-dev` bucket via SSO; `make refresh-dev-s3` rotates stale STS tokens. Compose overlay only sets bucket/endpoint/region — credentials never go in the overlay.
- **AWS auth flexibility** — `createS3Client` supports static IAM, STS (sessionToken), and keyless (IRSA / default chain) modes plus region override.
- **UI rendering** (`ScenarioMessageRenderer.tsx`) — `input_audio` parts render as `<audio>` players; audio bubbles align by role (assistant left, user right); when a message has audio + sibling text, they collapse into one bordered unit with the audio on top and an italic transcript caption.

## Test plan

- `langwatch/src/server/stored-objects/__tests__/` — service / repository / driver / registry / URI / extractor unit tests
- `langwatch/src/server/stored-objects/__tests__/stored-objects.ingest-read.integration.test.ts` — full ingest → projection → read round-trip
[...]

PR #4193, opened 2026-05-25
fix(storage): pass region:undefined for AWS-IRSA endpoints instead of 'auto'

### LINKED ISSUE #4139: Prod: SCENARIO_MESSAGE_SNAPSHOT POSTs return 500 — getaddrinfo ENOTFOUND sts.auto.amazonaws.com
**Describe the bug**

In production (`https://app.langwatch.ai`), every `POST /api/scenario-events` of type `SCENARIO_MESSAGE_SNAPSHOT` carrying inline media returns HTTP 500 with body:

```
{"error":"Internal server error","message":"getaddrinfo ENOTFOUND sts.auto.amazonaws.com (ENOTFOUND)"}
```

`SCENARIO_RUN_STARTED` and `SCENARIO_RUN_FINISHED` both return 201 on the same key — they don't hit the stored-objects S3 path. Observed in two voice runs today (`scenariobatch_3DzSNGETWYolKYxHVk1jvz0zeAt`, `scenariobatch_3DzSaqKGfylvQ6eOBrJaapUS3Pq` in `drews-sandbox-owUldA`): 10/10 snapshot POSTs failed across both runs.

**Root cause** (confirmed via /investigate):

`createS3Client` in `langwatch/src/server/storage.ts:148` defaults region to `"auto"` when `S3_REGION` env is unset. In prod, the langwatch app and worker deployments (`infrastructure/langwatch.tf:118-126`, `infrastructure/langwatch_workers.tf:78-85`) set `S3_BUCKET_NAME`, `S3_ENDPOINT`, `USE_S3_STORAGE` — but **no `S3_REGION`** and no explicit `S3_ACCESS_KEY_ID`/`SECRET`. With no explicit keys, the AWS SDK falls into the default credential chain (PR #4058's IRSA fix), which fires STS `AssumeRoleWithWebIdentity` and constructs `sts.${region}.amazonaws.com` → `sts.auto.amazonaws.com` → NXDOMAIN. The `S3_ENDPOINT` override doesn't help because STS is configured separately by the SDK.

The path is reachable from `extractInlineMediaFromEvent` (`app.ts:90`), called on every SCENARIO_MESSAGE_SNAPSHOT. When inline media (e.g. voice `input_audio` content parts) is present, the extractor calls `service.storeFromBytes` → s3-driver → `createS3Client(projectId)` → SDK init → STS DNS fail.

**Why text-only didn't break:** text-only snapshots don't trigger the s3 code path (`s3-driver.ts:66` instantiates lazily on `read`). Voice runs today hit it because the SDK was emitting `input_audio` content parts (via Python-repr stringification that the server's `coerceContentToArray` resurrects). See related issue for SDK side.

**To reproduce**

1. From `langwatch/scenario` Python SDK, set `LANGWATCH_ENDPOINT=https://app.langwatch.ai` and `LANGWATCH_API_KEY=<sk-lw-*>`.
2. Run `cd python && uv run examples/voice/basic_greeting.py`.
3. Tail `recordings/scenario.log` — `event_reporter` lines for `SCENARIO_MESSAGE_SNAPSHOT` return 500 with the ENOTFOUND body.

Reproduces 100%.

**Expected behavior**

`SCENARIO_MESSAGE_SNAPSHOT` POSTs return 201 (or 200); per-message records persist; the run's batch page in the UI shows full conversation transcripts plus any extracted media refs (`/api/files/<id>`).

**Environment**

- Component: langwatch app + langwatch-workers (production)
- Surface: `POST /api/scenario-events`, MESSAGE_SNAPSHOT event type
- Region (prod): `eu-central-1` (per `infrastructure/environments/prod/aws.tfvars:18`)
- Bucket: `langwatch-storage-prod`

## Acceptance Criteria

This issue requires two PRs:

**PR-A (langwatch-saas, operational fix):**
1. `S3_REGION` env var is present on both `langwatch` and `langwatch-workers` k8s deployments in `infrastructure/langwatch.tf` + `infrastructure/langwatch_workers.tf`, set to `data.aws_region.current.name`.
2. Prod POST `/api/scenario-events` with MESSAGE_SNAPSHOT carrying inline media returns 201 (no `getaddrinfo ENOTFOUND sts.*` in CloudWatch).
3. CloudWatch shows zero `sts.auto.amazonaws.com` DNS errors from `langwatch` / `langwatch-workers` pods for 24h after deploy.

**PR-B (langwatch/langwatch, code-level guard):**
4. `langwatch/src/server/storage.ts:148`: when `S3_ENDPOINT` matches `*.amazonaws.com` (or is unset) AND `S3_REGION` is unset AND no explicit keys → the S3Client is constructed with the SDK-resolved AWS region, NOT the literal string `"auto"`.
5. When `S3_ENDPOINT` is non-AWS (R2/MinIO host) → `"auto"` is preserved (no regression for BYOC operators).
6. Unit tests covering both branches in `langwatch/src/server/__tests__/storage.test.ts`.

PR-A must land first (unblocks all integration verification of PR-B/C/D). PR-B closes the foot-gun so the next env-var omission doesn't reach prod silently.

---
*Classification: Bug*
*Status: investigated. ACs above. Blocks #4138 verification.*

## Plan (PR-A — `langwatch-saas` infrastructure)

**Approach.** Two minimal env-var additions in `infrastructure/`: append a fifth `env {}` block named `S3_REGION` valued at `data.aws_region.current.name` inside both the `langwatch` and `langwatch_workers` k8s deployment resources, immediately adjacent to the existing `S3_ENDPOINT` / `USE_S3_STORAGE` blocks. The `data "aws_region" "current"` source already exists (`infrastructure/main.tf:27`) and is in use in both target files, so no new data sources are introduced. PR-B (the `storage.ts` foot-gun guard) is filed against `langwatch/langwatch` and is out of scope here.

**Sequencing.**
1. Patch `infrastructure/langwatch.tf` — add `S3_REGION` env block.
2. Patch `infrastructure/langwatch_workers.tf` — add `S3_REGION` env block.
3. `terraform plan` against prod backend; verify the diff is exactly two `env` additions on the two deployments, no other resource churn.
4. Open PR; on merge, the existing deploy pipeline rolls out both deployments.
5. Post-deploy verification: tail CloudWatch for `sts.auto.amazonaws.com` errors (expect zero) and repro a SCENARIO_MESSAGE_SNAPSHOT POST against prod (expect 201). Hold open 24h for AC #3.

**Key decisions.**
- *Value = `data.aws_region.current.name`, not a hardcoded string.* Matches the in-file convention for `S3_ENDPOINT` and keeps non-prod region overrides working automatically.
- *Two separate `env {}` blocks, one per file, not a shared `local`.* Consistent with the existing repetition pattern; refactoring shared env into a module is out of scope and would expand the blast radius.
- *PR-A only; the langwatch/langwatch code guard is tracked separately.* PR-A must land first to unblock the integration verification path for PR-B.

**Risks / open questions.**
- IRSA role: the SDK will resolve credentials via `AssumeRoleWithWebIdentity` once it has a real region. The IRSA service-account binding from #4058 must already cover s3:PutObject on `langwatch-storage-prod` — confirm by reading the existing trust policy before merge, not after.
- Cold-start lag: if any in-flight worker pod caches the old `"auto"` region in memory, the fix only takes effect after the rolling restart completes.

**Out of scope.**
- `langwatch/src/server/storage.ts` default region logic (PR-B, `langwatch/langwatch` repo).
- Refactoring duplicated env blocks across the two terraform files into a shared local/module.
- Any change to `S3_ACCESS_KEY_ID` / `S3_SECRET` handling — credentials remain on the IRSA default chain.

## Plan (PR-B — `langwatch/langwatch` code-level guard)

**Approach.** Tighten the region default inside `createS3Client` so that omitting `S3_REGION` no longer silently produces the literal string `"auto"` against real AWS endpoints. Detect "this is AWS" from the endpoint shape (unset or hostname matches `*.amazonaws.com`) plus the no-explicit-keys signal that already exists in the function, then hand the `S3Client` constructor `undefined` for region so the AWS SDK runs its own region provider chain (IRSA's `AWS_REGION` env, EC2 IMDS, `~/.aws/config`, etc.). The BYOC / R2 / MinIO path keeps the existing `"auto"` literal — those operators rely on it and the SDK ignores the value for non-AWS endpoints anyway. No new env vars, no changes to the credentials precedence (PR #4058's contract), no public-surface API churn.

**Sequencing.**
1. Extend the existing test file `langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts` (the AC names `storage.test.ts` but the actual file lives at this path — extend, don't create a duplicate) with three new `describe` blocks: (a) AWS endpoint, no `S3_REGION`, no explicit keys → `S3Client` receives `region: undefined`; (b) BYOC endpoint (e.g. `https://*.r2.cloudflarestorage.com`, MinIO host), no `S3_REGION` → `S3Client` receives `region: "auto"`; (c) AWS endpoint, no `S3_REGION`, `AWS_REGION=eu-central-1` in process env → resolved `s3Client.config.region()` returns `"eu-central-1"` (proves SDK region chain fires).
2. Patch `createS3Client` in `langwatch/src/server/storage.ts` to compute the region value via a small inline helper: AWS-shaped endpoint AND no `S3_REGION` AND no explicit keys → pass `undefined`; everything else → keep current behavior (`env.S3_REGION ?? "auto"`).
3. Run `pnpm test:unit langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts` end-to-end (not CI=1 parse-only) plus `pnpm typecheck`.
4. Open PR-B against `main`. PR-A unblocks integration verification, but PR-B is mergeable independently — its unit tests are the contract.

**Key decisions.**
- *Detect "AWS" from the endpoint, not from credential shape.* The credential signal is already used by PR #4058; adding endpoint-shape detection is the orthogonal signal AC#4 asks for and avoids re-routing the credentials branch.
- *Pass `undefined` rather than reading `process.env.AWS_REGION` ourselves.* The AWS SDK's region provider chain covers IRSA, IMDS, profile, and `AWS_DEFAULT_REGION` in the right order — duplicating that logic in TypeScript is the foot-gun in a different costume.
- *Keep `"auto"` for BYOC.* AC#5 is explicit, R2/MinIO operators depend on it, and the SDK ignores region for non-AWS endpoints — changing it would be a silent behavior change for tenants we can't easily survey.
- *Extend the existing test file, not create the file named in AC#6.* The AC's pathname is a convenience reference; an existing `storage.createS3Client.unit.test.ts` already covers credential-mode handling with the same mocking pattern. Boy Scout: extend, do not duplicate.
- *Inline helper, not a new module.* The detection is two lines; extracting it would obscure the credential-precedence comment block that already documents this function. Revisit if a third caller appears.

**Risks / open questions.**
- *AWS endpoint detection scope.* Treating "endpoint unset OR ends with `.amazonaws.com`" as AWS misses oddball patterns (FIPS endpoints, S3 dualstack `*.s3-fips.*`, S3 Outposts). Those still end in `.amazonaws.com`, so they're covered, but worth grepping prod env to confirm before merge.
- *SDK region-resolution failure mode.* If the SDK's region chain *also* finds nothing (no `AWS_REGION`, no IMDS, no profile), it throws on first request rather than at construction. That's strictly better than today's `sts.auto.amazonaws.com` NXDOMAIN, but the surfacing changes — note in PR description.
- *SDK behavior in EKS / IRSA (confirmed).* The EKS pod identity webhook injects `AWS_REGION` and `AWS_DEFAULT_REGION` alongside the IRSA `AWS_WEB_IDENTITY_TOKEN_FILE` / `AWS_ROLE_ARN`. The AWS SDK v3 region provider chain reads those env vars before falling through to shared config and IMDS. So in EKS pods, `region: undefined` resolves to the cluster's region without any explicit configuration. This means **PR-B alone is sufficient to fix prod**, and PR-A (saas#511) becomes defense-in-depth (explicit `S3_REGION` makes failure modes louder if IRSA is misconfigured). The unit test for the AWS-default branch should assert this behavior: with `AWS_REGION=eu-central-1` in process env and no `S3_REGION`, the resulting `S3Client.config.region()` resolves to `"eu-central-1"`.
- *Test mock fidelity.* The existing test asserts on constructor args via a `S3Client` mock. We need the assertion to distinguish `region: undefined` (key omitted vs. key present with `undefined` value); the SDK treats both as "run the chain," and the test should be explicit about which we're emitting.

**Out of scope.**
- The infrastructure terraform fix — that is PR-A, owned in `langwatch-saas`.
- Reworking the credential precedence (PR #4058 contract).
- Adding a new env var (e.g. `S3_FORCE_AWS_REGION`) as an escape hatch — premature; revisit only if a tenant reports a regression.
- Touching the `StorageService` dataset path or the stored-objects driver code.
- Integration tests against a real S3 bucket — unit tests on the constructor branching is the contract for this PR.

drewdrewthis:
## Why

Closes #4139

IRSA (IAM Roles for Service Accounts) on EKS authenticates via a projected service-account token — no `S3_ACCESS_KEY_ID` / `S3_SECRET_ACCESS_KEY` in env. The previous code always passed `region: "auto"` to `S3Client`, which **breaks AWS SigV4 signature validation**: AWS S3 rejects requests with `sts.auto.amazonaws.com` (NXDOMAIN) when no explicit region is given.

On EKS with IRSA, the correct behavior is to pass `region: undefined` so the AWS SDK resolves it from its credential chain — specifically `AWS_REGION`, which EKS injects into every pod.

## What changed

- **`region: undefined` for keyless AWS endpoints** — when `S3_ENDPOINT` is absent or `*.amazonaws.com` and no explicit keys (IRSA path), `region` is omitted from `S3Client` config so the SDK resolves it from `AWS_REGION`
- **`region: "auto"` preserved for BYOC endpoints** — Cloudflare R2, MinIO, and any non-AWS `S3_ENDPOINT` keep `"auto"` (required by those providers)
- **`S3_REGION` always wins** — explicit env override takes precedence in all cases
- **Unit tests** — `resetS3Env` helper extracted, `describe("when...")` nesting added, and the previously-untested AWS+explicit-keys→`"auto"` backwards-compatibility path now covered

## How it works

Region resolution priority in `createS3Client`:

| Condition | `region` passed to S3Client |
|---|---|
| `S3_REGION` set | `S3_REGION` value |
| AWS endpoint (no endpoint or `*.amazonaws.com`) + no explicit keys (IRSA) | `undefined` — SDK resolves from `AWS_REGION` |
| AWS endpoint + explicit keys | `"auto"` (preserves pre-#4058 behavior for local/ops tooling) |
| Non-AWS endpoint (R2, MinIO, custom) | `"auto"` |

The spread `...(region !== undefined ? { region } : {})` keeps `region` out of the constructor config entirely when undefined — consistent with the existing `credentials` field pattern in the same function, where absent-vs-present-but-undefined has different SDK semantics.

## Test plan

```bash
cd langwatch && pnpm test:unit src/server/__tests__/storage.createS3Client.unit.test.ts
```

New `describe` blocks covering:
- AWS endpoint, no keys (IRSA path) → `region` absent from config
- AWS endpoint, explicit keys → `region: "auto"` (backwards-compatibility path)
- Non-AWS endpoint (R2, MinIO) → `region: "auto"`
- `S3_REGION` override → always wins regardless of endpoint type

### DIFF
--- a/langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
+++ b/langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
@@ -70,16 +70,18 @@ vi.mock("../../env.mjs", () => ({
   ),
 }));
 
+function resetS3Env() {
+  s3ClientConstructorCalls.length = 0;
+  delete process.env.S3_ENDPOINT;
+  delete process.env.S3_ACCESS_KEY_ID;
+  delete process.env.S3_SECRET_ACCESS_KEY;
+  delete process.env.S3_SESSION_TOKEN;
+  delete process.env.S3_REGION;
+  delete process.env.S3_BUCKET_NAME;
+}
+
 describe("createS3Client credential mode handling", () => {
-  beforeEach(() => {
-    s3ClientConstructorCalls.length = 0;
-    delete process.env.S3_ENDPOINT;
-    delete process.env.S3_ACCESS_KEY_ID;
-    delete process.env.S3_SECRET_ACCESS_KEY;
-    delete process.env.S3_SESSION_TOKEN;
-    delete process.env.S3_REGION;
-    delete process.env.S3_BUCKET_NAME;
-  });
+  beforeEach(resetS3Env);
 
   describe("given static IAM-user keys (AKIA + secret, no token)", () => {
     /** @scenario "S3 client uses explicit credentials when env keys are present" */
@@ -190,80 +192,77 @@ describe("createS3Client credential mode handling", () => {
 });
 
 describe("region resolution — AWS endpoint, keyless (IRSA path)", () => {
-  beforeEach(() => {
-    s3ClientConstructorCalls.length = 0;
-    delete process.env.S3_ENDPOINT;
-    delete process.env.S3_ACCESS_KEY_ID;
-    delete process.env.S3_SECRET_ACCESS_KEY;
-    delete process.env.S3_SESSION_TOKEN;
-    delete process.env.S3_REGION;
-    delete process.env.S3_BUCKET_NAME;
+  beforeEach(resetS3Env);
+
+  describe("when no endpoint is set and no keys are provided", () => {
+    it("omits region so SDK resolves from credential chain", async () => {
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call).not.toHaveProperty("region");
+    });
   });
 
-  it("passes region: undefined to S3Client when endpoint is unset and no keys", async () => {
-    // Arrange: no S3_ENDPOINT, no S3_REGION, no credentials (IRSA path)
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call).not.toHaveProperty("region");
+  describe("when endpoint is *.amazonaws.com and no keys are provided", () => {
+    it("omits region so SDK resolves from credential chain", async () => {
+      process.env.S3_ENDPOINT = "https://s3.eu-central-1.amazonaws.com";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call).not.toHaveProperty("region");
+    });
   });
 
-  it("passes region: undefined when endpoint is *.amazonaws.com and no keys", async () => {
-    process.env.S3_ENDPOINT = "https://s3.eu-central-1.amazonaws.com";
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call).not.toHaveProperty("region");
+  describe("when endpoint is AWS and explicit keys are provided (pre-#4058 compatibility path)", () => {
+    it("falls back to 'auto' to preserve pre-#4058 ops-tooling behavior", async () => {
+      process.env.S3_ACCESS_KEY_ID = "AKIAEXAMPLE";
+      process.env.S3_SECRET_ACCESS_KEY = "secret-value";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call.region).toBe("auto");
+    });
   });
 });
 
 describe("region resolution — non-AWS endpoint (BYOC/R2/MinIO)", () => {
-  beforeEach(() => {
-    s3ClientConstructorCalls.length = 0;
-    delete process.env.S3_ENDPOINT;
-    delete process.env.S3_ACCESS_KEY_ID;
-    delete process.env.S3_SECRET_ACCESS_KEY;
-    delete process.env.S3_SESSION_TOKEN;
-    delete process.env.S3_REGION;
-    delete process.env.S3_BUCKET_NAME;
+  beforeEach(resetS3Env);
+
+  describe("when endpoint is Cloudflare R2", () => {
+    it("uses 'auto' region", async () => {
+      process.env.S3_ENDPOINT = "https://abc123.r2.cloudflarestorage.com";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call.region).toBe("auto");
+    });
   });
 
-  it("keeps region: 'auto' for R2 endpoint", async () => {
-    process.env.S3_ENDPOINT = "https://abc123.r2.cloudflarestorage.com";
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call.region).toBe("auto");
-  });
-
-  it("keeps region: 'auto' for MinIO endpoint", async () => {
-    process.env.S3_ENDPOINT = "http://minio:9000";
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call.region).toBe("auto");
+  describe("when endpoint is MinIO", () => {
+    it("uses 'auto' region", async () => {
+      process.env.S3_ENDPOINT = "http://minio:9000";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call.region).toBe("auto");
+    });
   });
 });
 
 describe("region resolution — explicit S3_REGION always wins", () => {
-  beforeEach(() => {
-    s3ClientConstructorCalls.length = 0;
-    delete process.env.S3_ENDPOINT;
-    delete process.env.S3_ACCESS_KEY_ID;
-    delete process.env.S3_SECRET_ACCESS_KEY;
-    delete process.env.S3_SESSION_TOKEN;
-    delete process.env.S3_REGION;
-    delete process.env.S3_BUCKET_NAME;
+  beforeEach(resetS3Env);
+
+  describe("when S3_REGION is set for AWS endpoint without keys", () => {
+    it("uses S3_REGION value", async () => {
+      process.env.S3_REGION = "eu-central-1";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call.region).toBe("eu-central-1");
+    });
   });
 
-  it("uses S3_REGION when set, even for AWS endpoint without keys", async () => {
-    process.env.S3_REGION = "eu-central-1";
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call.region).toBe("eu-central-1");
-  });
-
-  it("uses S3_REGION when set, even for BYOC endpoint", async () => {
-    process.env.S3_REGION = "us-east-1";
-    process.env.S3_ENDPOINT = "http://minio:9000";
-    await createS3Client("test-project");
-    const call = s3ClientConstructorCalls[0];
-    expect(call.region).toBe("us-east-1");
+  describe("when S3_REGION is set for BYOC endpoint", () => {
+    it("uses S3_REGION value", async () => {
+      process.env.S3_REGION = "us-east-1";
+      process.env.S3_ENDPOINT = "http://minio:9000";
+      await createS3Client("test-project");
+      const call = s3ClientConstructorCalls[0];
+      expect(call.region).toBe("us-east-1");
+    });
   });
 });
--- a/langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
+++ b/langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
@@ -19,6 +19,7 @@
  * resolution from there.
  */
 import { describe, it, expect, beforeEach, vi } from "vitest";
+import { createS3Client } from "../storage";
 
 const s3ClientConstructorCalls: any[] = [];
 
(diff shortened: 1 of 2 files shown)


langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
```diff
@@ -187,3 +188,82 @@ describe("createS3Client credential mode handling", () => {
...
+    s3ClientConstructorCalls.length = 0;
+    delete process.env.S3_ENDPOINT;
+    delete process.env.S3_ACCESS_KEY_ID;
+    delete process.env.S3_SECRET_ACCESS_KEY;
+    delete process.env.S3_SESSION_TOKEN;
+    delete process.env.S3_REGION;
+    delete process.env.S3_BUCKET_NAME;
+  });
+
+  it("passes region: undefined to S3Client when endpoint is unset and no keys", async () => {
```
drewdrewthis: **[hygiene][test] Missing `describe("when...")` nesting (CLAUDE.md violation)**

All new  blocks sit flat inside their top-level  — the existing suite in this file wraps every case in  →  → . CLAUDE.md requires: "Inner describe blocks must use 'when' conditions."

**Fix:** Wrap each  in a , consistent with the  structure used throughout the rest of this file.

langwatch/src/server/__tests__/storage.createS3Client.unit.test.ts
```diff
@@ -187,3 +188,82 @@ describe("createS3Client credential mode handling", () => {
...
+  });
+
+  it("uses S3_REGION when set, even for BYOC endpoint", async () => {
+    process.env.S3_REGION = "us-east-1";
+    process.env.S3_ENDPOINT = "http://minio:9000";
+    await createS3Client("test-project");
+    const call = s3ClientConstructorCalls[0];
+    expect(call.region).toBe("us-east-1");
+  });
+});
```
drewdrewthis, line 268: **[test][uncle-bob] Missing branch test: AWS endpoint + explicit keys + no  → **

The  comment at line ~151 explicitly documents this path: "fall back to  to preserve the pre-PR-4058 behavior for ops tooling / local env with hardcoded keys." No test covers it. The branch condition is  → .

Without a test, a future refactor of the ternary can silently break the backwards-compatibility path (ops tooling / local dev with hardcoded keys would get  instead of , causing SigV4 failures).

**Fix:** Add a test case:

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `b3ddc51c1` · **Run:** /review (own-PR)

- **[hygiene][test] `storage.createS3Client.unit.test.ts:22`** — New tests use static top-level import; existing tests use `vi.resetModules()` + dynamic import. Safe with current Proxy mock; fragile if `storage.ts` gains module-level state. (Decide)

- **[principles][uncle-bob][fowler]** Extract `isAwsManagedEndpoint(endpoint)` as a named predicate (lowercase-normalize inside). Fowler: "A named domain concept beats a string trick." The `resolveRegion` helper would make the four-case ternary independently testable. Out of scope for this hotfix. (New Issue)

- **[security]** `privateConfig?.endpoint` is DB-backed per-project tenant config. If tenants can set it freely, a crafted hostname passing `isAwsEndpoint` could affect region resolution. Add upstream validation. (New Issue)

---

0xdeafcafe (approved):

PR #4199, opened 2026-05-26
fix(event-sourcing): coalesce backed-up fold groups into one apply/store cycle

rogeriochaves:
## What

Fold projections process one event at a time: read the previous state, apply the event, write it back. For a single aggregate with N queued events that is N load+store round-trips over state that grows with N, which is O(n^2) on the accumulated state. The GroupQueue's per-group active gate dispatches one job per group at a time, so a backed-up group drains strictly one event per dispatch and the cost compounds.

When a trace's fold group backs up, this diverges instead of draining and stalls the single-threaded Redis command path for every tenant on the shared queue.

This adds opt-in per-group batch coalescing:

- While a worker holds a group's active slot (which already excludes the group from concurrent dispatch), it drains the group's remaining due jobs in one pop-only Lua call (`drainGroupReady`) and folds them in a single load/apply/store cycle, in occurredAt order (`FoldProjectionExecutor.executeBatch`).
- A backed-up group now drains in O(n) and self-heals. When the queue keeps up, batches are size 1 and the per-event path is byte-for-byte unchanged.
- Bounded by a per-fold max batch (default 100), enabled for all fold projections. Other queues (maps, reactors) are untouched.
- A failed batch re-stages its drained siblings so none are lost (same recovery profile as the existing per-event path).
- The drain decrements `total-pending` once per job, exactly as dispatch does, so the ops counter stays conserved.

Also raises the Redis fold-cache TTL from 30s to 300s so the accumulated state stays warm in Redis across an aggregate's event stream instead of expiring mid-stream and forcing a ClickHouse re-read of the full state on every event.

## Why

This is the read-side half of the fold-projection fix. The structural maps in the trace summary state (events, span costs, per-role spans) grow with span count regardless of payload size, so shrinking payloads alone does not remove the O(n^2). Coalescing collapses the backlog cost itself.

The store-side half (capping oversized span attributes at ingestion, which shrinks the per-event constant) landed separately in #4198. The two compose: #4198 keeps each event small, this keeps a backed-up group from re-reading and re-writing the growing state once per event.

## Scope and safety

- Coalescing is opt-in via `coalesceMaxBatch`; with it at 1 the dispatch, complete, and retry Lua paths are unchanged.
- `drainGroupReady` is pop-only: it never touches the group's active, ready, or blocked state, and is only ever called while the caller holds the group's active slot, so the drain is exclusive.
- The fold is applied in occurredAt order and stored once, which is all-or-nothing per batch: a mid-batch failure leaves the persisted checkpoint untouched and re-stages the siblings, so re-processing folds cleanly from the same checkpoint.
- The in-memory dev/test queue ignores coalescing and keeps the per-event path.

## Tests

- Unit: `executeBatch` reads once and stores once, equals sequential folding, folds out-of-order events in occurredAt order, re-folds via eventLoader when a batch precedes the checkpoint.
- Integration (real Redis): `drainGroupReady` semantics (ascending order, removal, due-only, pending-counter decrement, no-op guards); end-to-end coalescing of a backed-up group into one batch call, max-batch cap, disabled-passthrough, and failed-batch re-stage.

BDD specs: `specs/event-sourcing/fold-coalescing.feature`, plus a coalescing scenario in `pending-counter-conservation.feature` and the TTL update in `redis-fold-cache.feature`.

### DIFF
--- a/langwatch/src/server/event-sourcing/eventSourcing.ts
+++ b/langwatch/src/server/event-sourcing/eventSourcing.ts
@@ -456,6 +456,30 @@ export class EventSourcing {
         }
         await result.entry.process(result.clean);
       },
+      coalesceMaxBatch: (payload: Record<string, unknown>) => {
+        const result = this.lookupEntry(payload);
+        return result?.entry.coalesceMaxBatch ?? 1;
+      },
+      processBatch: async (payloads: Record<string, unknown>[]) => {
+        if (payloads.length === 0) return;
+        // A coalesced batch is always one group → one registry entry. Resolve it
+        // from the first payload and hand the batch handler the clean payloads
+        // (routing metadata stripped).
+        const first = this.lookupEntry(payloads[0]!);
+        if (!first?.entry.processBatch) {
+          // No batch handler for this job type — process each individually.
+          for (const payload of payloads) {
+            const result = this.lookupEntry(payload);
+            if (result) await result.entry.process(result.clean);
+          }
+          return;
+        }
+        const cleanPayloads = payloads.map((payload) => {
+          const result = this.lookupEntry(payload);
+          return result ? result.clean : payload;
+        });
+        await first.entry.processBatch(cleanPayloads);
+      },
     };
 
     const effectiveRedis = this._redis;
--- a/langwatch/src/server/event-sourcing/eventSourcing.ts
+++ b/langwatch/src/server/event-sourcing/eventSourcing.ts
@@ -462,23 +462,23 @@ export class EventSourcing {
       },
       processBatch: async (payloads: Record<string, unknown>[]) => {
         if (payloads.length === 0) return;
-        // A coalesced batch is always one group → one registry entry. Resolve it
-        // from the first payload and hand the batch handler the clean payloads
-        // (routing metadata stripped).
+        // A coalesced batch is always one group → one registry entry. Resolve
+        // every payload and guard against a mixed/unknown batch (should never
+        // happen — the GroupQueue only coalesces same-group jobs — but a stray
+        // payload must never be misrouted to the wrong handler). On any mismatch
+        // fall back to per-item processing.
         const first = this.lookupEntry(payloads[0]!);
-        if (!first?.entry.processBatch) {
-          // No batch handler for this job type — process each individually.
-          for (const payload of payloads) {
-            const result = this.lookupEntry(payload);
+        const resolved = payloads.map((payload) => this.lookupEntry(payload));
+        const homogeneous =
+          !!first?.entry.processBatch &&
+          resolved.every((r) => r?.entry === first.entry);
+        if (!homogeneous) {
+          for (const result of resolved) {
             if (result) await result.entry.process(result.clean);
           }
           return;
         }
-        const cleanPayloads = payloads.map((payload) => {
-          const result = this.lookupEntry(payload);
-          return result ? result.clean : payload;
-        });
-        await first.entry.processBatch(cleanPayloads);
+        await first.entry.processBatch!(resolved.map((r) => r!.clean));
       },
     };
 
--- a/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
+++ b/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
@@ -85,6 +85,93 @@ export class FoldProjectionExecutor {
     return state;
   }
 
+  /**
+   * Applies a batch of events for the same aggregate in a single load/store cycle.
+   *
+   * Equivalent to calling `execute()` once per event, but reads the existing
+   * state once, folds every event in occurredAt order, and writes the result
+   * once. This turns a backed-up group of N events from N load+store round-trips
+   * (O(n²) on growing fold state) into a single one (O(n)).
+   *
+   * Out-of-order handling matches `execute()`: if the earliest event in the
+   * batch occurred before the persisted checkpoint, the aggregate is re-folded
+   * from scratch via `eventLoader` (when available).
+   */
+  async executeBatch<State, E extends Event>(
+    projection: FoldProjectionDefinition<State, E>,
+    events: E[],
+    context: ProjectionStoreContext,
+  ): Promise<State> {
+    const matching = events.filter((event) => this.matchesEventTypes(projection, event));
+    if (matching.length === 0) {
+      return projection.init();
+    }
+    if (matching.length === 1) {
+      return this.execute(projection, matching[0]!, context);
+    }
+
+    // Process in occurredAt order so the fold sees events as they happened,
+    // regardless of the order they were dispatched/drained in.
+    const ordered = [...matching].sort(
+      (a, b) =>
+        (((a as Record<string, unknown>).occurredAt as number) ?? 0) -
+        (((b as Record<string, unknown>).occurredAt as number) ?? 0),
+    );
+
+    const key = context.key ?? context.aggregateId;
+    let state = (await projection.store.get(key, context)) ?? projection.init();
+
+    const prevLastOccurred =
+      (state as Record<string, unknown>)[projection.LastEventOccurredAtKey] ?? 0;
+    const earliestOccurredAt = (ordered[0] as Record<string, unknown>).occurredAt;
+
+    // Out-of-order vs the persisted checkpoint: the batch starts earlier than
+    // what we've already folded. Re-fold from scratch when we can load the full
+    // history; otherwise fall through and apply the batch on top (matches the
+    // single-event executor's degraded behavior when no eventLoader exists).
+    const outOfOrder =
+      typeof earliestOccurredAt === "number" &&
+      earliestOccurredAt > 0 &&
+      typeof prevLastOccurred === "number" &&
+      earliestOccurredAt < prevLastOccurred;
+
+    if (outOfOrder && projection.eventLoader) {
+      const allEvents = await projection.eventLoader({
+        tenantId: context.tenantId,
+        aggregateId: context.aggregateId,
+      });
+      logger.info(
+        {
+          projection: projection.name,
+          aggregateId: context.aggregateId,
+          tenantId: context.tenantId,
+          batchSize: ordered.length,
+          earliestOccurredAt,
+          prevLastOccurred,
+          refoldEventCount: allEvents.length,
+        },
+        "Out-of-order batch detected, re-folding from scratch",
+      );
+      state = projection.init();
+      for (const e of allEvents) {
+        state = projection.apply(state, e as E);
+      }
+    } else {
+      if (outOfOrder) {
+        logger.warn(
+          { projection: projection.name, aggregateId: context.aggregateId },
+          "Out-of-order batch detected but no eventLoader available — cannot re-fold",
+        );
+      }
+      for (const event of ordered) {
+        state = projection.apply(state, event);
+      }
+    }
+
+    await projection.store.store(state, context);
+    return state;
+  }
+
   private matchesEventTypes<State, E extends Event>(
     projection: FoldProjectionDefinition<State, E>,
     event: E,
--- a/langwatch/src/server/event-sourcing/projections/projectionRouter.ts
+++ b/langwatch/src/server/event-sourcing/projections/projectionRouter.ts
@@ -37,8 +37,9 @@ import type { ReplayMarkerChecker } from "./replayMarkerCheck";
 
 /**
  * Default cap on how many same-aggregate fold events are coalesced into one
- * load/apply/store cycle. Bounds the drain + apply loop (and the re-stage loop
- * on failure) while still collapsing a backed-up group from O(n²) to O(n).
+ * load/apply/store cycle. Bounds the per-cycle drain + apply loop (and the
+ * re-stage loop on failure) while collapsing a backed-up group from O(n²) to
+ * O(n). A fold can opt out by setting options.coalesceMaxBatch = 1.
  */
 const DEFAULT_FOLD_COALESCE_MAX_BATCH = 100;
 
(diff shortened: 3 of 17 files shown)


langwatch/src/server/event-sourcing/projections/projectionRouter.ts
```diff
@@ -615,6 +644,88 @@ export class ProjectionRouter<
...
+        const foldState = await withMetrics({
+          fn: () => this.foldExecutor.executeBatch(fold, toApply, storeContext),
+          onComplete: (ms) => { incrementEsFoldProjectionTotal(this.pipelineName, projectionName, "completed"); observeEsFoldProjectionDuration(this.pipelineName, projectionName, ms); },
+          onFail: (ms) => { incrementEsFoldProjectionTotal(this.pipelineName, projectionName, "failed"); observeEsFoldProjectionDuration(this.pipelineName, projectionName, ms); },
+        });
+
+        // Dispatch reactors once with the final state, using the last event.
+        const reactors = this.reactorsForFold.get(projectionName);
+        if (reactors && reactors.length > 0) {
+          await this.dispatchToReactors(
```
rogeriochaves: **Blocker: this drops per-span reactor work under coalescing.**

This fires every `traceSummary` reactor once, with `toApply[last]` and the final state. Correct for foldState-based reactors, but `customEvaluationSync` and `evaluationTrigger` are per-event / per-span:

- `customEvaluationSync.handle` does `extractEvaluationsFromSpan(event.data.span)` and reports each embedded eval. With a coalesced batch only the last span's evals are reported, the rest are silently dropped.
- `evaluationTrigger.handle` reads `event.data.span.attributes` per span to decide triggers, only the last span is checked.

Because coalescing only engages at batch >= 2 (under backlog), this is exactly the recovery path losing data.

Also: `toApply[toApply.length - 1]` is the last-by-arrival event, not last-by-occurredAt (`executeBatch` sorts internally, this array does not), so even the "latest event" semantics are slightly off.

Suggested fix that keeps the full O(n^2)->O(n) win (the cost is entirely in the fold load/apply/store, not the reactor enqueue):

```ts
const reactors = this.reactorsForFold.get(projectionName);
if (reactors && reactors.length > 0) {
  for (const event of toApply) {
    await this.dispatchToReactors(projectionName, reactors, event, foldState);
  }
}
```

Per-event dispatch with the final state: event-based reactors see every span, and foldState-based reactors that set `makeJobId` dedup the N dispatches down to one effective run. Reactor sends are cheap SimpleQueue enqueues, so this does not reintroduce the quadratic cost. Worth an integration test asserting that a coalesced batch carrying embedded evals on multiple spans syncs all of them, not just the last.

langwatch/src/server/event-sourcing/projections/redisCachedFoldStore.ts
```diff
@@ -17,6 +17,15 @@ export interface RedisCachedFoldStoreOptions {
...
 }
 
+/**
+ * Default cache TTL. Sized to outlast the processing of a single aggregate's
+ * event stream so the fold state stays warm in Redis across consecutive
+ * events, instead of expiring mid-stream and forcing a ClickHouse fallback
+ * read of the (potentially large) state on every event. Matches the queue's
+ * activeTtlSec — the upper bound on how long one aggregate stays in-flight.
+ */
+const DEFAULT_FOLD_CACHE_TTL_SECONDS = 300;
```
rogeriochaves: The rationale (keep state warm across an aggregate's in-flight window, matched to `activeTtlSec`) makes sense. One question, given the incident we are recovering from was memory-driven: a 10x residency means every distinct aggregate touched in any 300s window keeps its fold state resident in Redis, and for the heavy traces that state is itself multi-MB. Was the 10x weighed against Redis memory on the current node, and would large states warrant a size-aware TTL or a cap? Not blocking, just want it to be a conscious tradeoff rather than a silent re-pressure on the thing we just fixed.

langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
```diff
@@ -85,6 +85,93 @@ export class FoldProjectionExecutor {
...
+    // what we've already folded. Re-fold from scratch when we can load the full
+    // history; otherwise fall through and apply the batch on top (matches the
+    // single-event executor's degraded behavior when no eventLoader exists).
+    const outOfOrder =
+      typeof earliestOccurredAt === "number" &&
+      earliestOccurredAt > 0 &&
+      typeof prevLastOccurred === "number" &&
+      earliestOccurredAt < prevLastOccurred;
+
+    if (outOfOrder && projection.eventLoader) {
```
rogeriochaves, line 138: The out-of-order re-fold replaces state with `init()` + replay of `eventLoader(...)` only, discarding the just-applied batch. This is correct only if `eventLoader` returns a history that already includes the in-flight batch events. If the loader reads from ClickHouse and there is async-insert lag, a re-fold triggered right after append could replay a history missing the newest events and regress the state. The same assumption exists in single-event `execute()`, but the batch widens the window. Can you confirm the loader source is read-your-writes for the events being processed, or that the batch events are otherwise guaranteed visible at re-fold time?

rogeriochaves (commented):
Deep review of the coalescing path. The queue side is solid and the executor's batch math is right, but there is one correctness blocker in the reactor dispatch that silently drops data during exactly the backlog conditions this PR targets. Details inline, summary here.

**Blocker: coalescing fires fold reactors once with only the last event, but two `traceSummary` reactors are per-span, not per-foldState.**

`processFoldProjectionBatch` dispatches reactors a single time with `toApply[last]` + the final state. That is correct for the foldState-based reactors (`traceUpdateBroadcast`, the `*MetricsSync` ones, `alertTrigger`, the debounced syncs), but `customEvaluationSync` and `evaluationTrigger` read `event.data.span` per event:

- `customEvaluationSync` loops `extractEvaluationsFromSpan(event.data.span)` and reports each embedded SDK evaluation. Coalescing N `span_received` events reports only the last span's evals and silently drops the other N-1.
- `evaluationTrigger` inspects each span's attributes to decide whether to trigger monitors. Coalescing only checks the last span.

This only manifests once batches actually coalesce (batch >= 2), i.e. under backlog, which is precisely when this code path runs. So recovering from the kind of backlog that motivated the PR would silently lose custom evaluations and evaluation triggers for the backed-up traces. Concrete fix inline.

**Non-blocking:** the TTL 30->300s memory tradeoff (question) and the out-of-order re-fold's reliance on eventLoader visibility (question), both inline.

On why per-event was the original design (since this came up): the fold executor has been per-event since the event-sourcing rework (#1704), then hardened per-event by #2947 (re-fold on out-of-order) and #3976 (evaluator loop prevention). The O(n^2) is invisible while the queue keeps up (effective batch size 1) and only bites under a deep backlog of one huge aggregate, a tail that had not occurred before now. Per-event reactor dispatch is load-bearing for the two reactors above, which is why the once-with-last-event shortcut breaks them.

Nice work on the real-redis Lua tests and the pending-counter conservation spec, the queue mechanics look correct (due-only drain, active slot untouched, pending DECR per drained job).

sergioestebance:
Review status: not ready to merge.

Findings:

- P1 `langwatch/src/server/event-sourcing/projections/projectionRouter.ts` around the default coalescing setup and reactor dispatch: coalescing is enabled by default for every fold, but batch mode fires fold reactors only once with a single event. Existing reactors are event-sensitive, for example `scenarioExecution` only runs on `lw.simulation_run.queued`; if a backed-up group batches `queued` plus later events and dispatches reactors using only the final event, that reactor can be skipped entirely. Minimal fix: either make coalescing opt-in only for folds whose reactors are batch-safe, or preserve per-event reactor dispatch semantics for all applied events, with a regression test using an event-specific reactor.

- P2 `langwatch/src/server/event-sourcing/eventSourcing.ts:463`: `processBatch` selects the batch handler from the first payload and sends all cleaned payloads to it without verifying every payload resolves to the same registry entry. The queue should usually group that way, but this is a high-impact assumption in a shared global queue. Add a homogeneity guard or group by `lookupEntry().entry` before calling `processBatch`.

- P3 `specs/event-sourcing/redis-fold-cache.feature:25`: the feature still says Redis is written first and ClickHouse is fired async, while `RedisCachedFoldStore.store` writes ClickHouse first and then updates Redis. Since these specs are requirements, update the scenario wording while touching the TTL.

Merge notes: GitHub reports the PR as mergeable and most workflows are green, but `issue-link` has a `startup_failure`, and there are unresolved review threads. I would block merge until the reactor semantics issue is fixed or explicitly narrowed to a safe opt-in scope.

sergioestebance:
Additional finding I did not see covered yet:

- P2 `langwatch/src/server/event-sourcing/eventSourcing.ts:568` / `langwatch/src/server/event-sourcing/projections/projectionRouter.ts:248`: the PR says folds can override or opt out via `options.coalesceMaxBatch`, but the public builder path drops those options. `withFoldProjection(name, definition, options)` stores `{ definition, options }`, then `buildServiceOptions()` maps that down to only the projection instance, and `initializeFoldQueues()` reads `fold.options?.coalesceMaxBatch`. So a caller using the documented builder options API, e.g. `.withFoldProjection("traceSummary", projection, { coalesceMaxBatch: 1 })`, still gets the default batch size of 100. Minimal fix: propagate the stored static builder options into the runtime fold definition before registration, or carry the `{ definition, options }` pair through to `initializeFoldQueues()` and merge those options there. This matters especially if the reactor semantics fix chooses a staged opt-out/opt-in rollout.

langwatch/src/server/event-sourcing/eventSourcing.ts
```diff
@@ -456,6 +456,30 @@ export class EventSourcing {
...
+            if (result) await result.entry.process(result.clean);
+          }
+          return;
+        }
+        const cleanPayloads = payloads.map((payload) => {
+          const result = this.lookupEntry(payload);
+          return result ? result.clean : payload;
+        });
+        await first.entry.processBatch(cleanPayloads);
+      },
```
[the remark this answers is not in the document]
rogeriochaves: Done in `8c10d8fa2`. Added the homogeneity guard: resolve every payload and, if any maps to a different `JobRegistryEntry` (or is unknown), fall back to per-item processing. A coalesced batch is always one group = one entry in practice, but this makes misrouting impossible.

langwatch/src/server/event-sourcing/projections/projectionRouter.ts
```diff
@@ -615,6 +644,88 @@ export class ProjectionRouter<
...
+            projectionName,
+            reactors,
+            toApply[toApply.length - 1]!,
+            foldState,
+          );
```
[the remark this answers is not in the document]
rogeriochaves: Fixed — and more than ordering: `processFoldProjectionBatch` now dispatches reactors **per event** (not once with the last), and `toApply` is sorted by `occurredAt` before both folding and reactor dispatch, so order is consistent with `executeBatch`. See the blocker thread.

specs/event-sourcing/redis-fold-cache.feature
```diff
@@ -24,7 +24,7 @@ Feature: Redis write-through cache for fold projections
 
   Scenario: Store commits to Redis first then fires ClickHouse write
     When the fold stores new state for aggregate "trace-1"
-    Then the state is written to Redis with a 30-second TTL
+    Then the state is written to Redis with a 300-second TTL
     And a ClickHouse INSERT is fired without waiting for the async flush
```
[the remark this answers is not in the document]
rogeriochaves: Fixed. The scenario now reads: ClickHouse first (throwing on failure so the event is retried), then the Redis cache with a 300s TTL — matching `RedisCachedFoldStore.store`.

langwatch/src/server/event-sourcing/projections/projectionRouter.ts
rogeriochaves: Fixed exactly as suggested: a per-event `dispatchToReactors` loop with the final fold state. Event-sensitive reactors (`customEvaluationSync`, `evaluationTrigger`, `scenarioExecution`) now see every event; fold-state reactors dedup via `makeJobId` down to one effective run; the O(n) win stays in the single fold load/store, not the cheap reactor enqueues.

Tests added: `projectionRouter` unit tests assert a coalesced batch loads/stores once but invokes reactors per event in occurredAt order; a deterministic ClickHouse integration test folds 40 spans via `executeBatch` and reads `spanCount` 40 back from `trace_summaries` (verified green against live ClickHouse).

Kept coalescing default-on for all folds (maintainer call): the final folded state is identical to sequential folding, per-event dispatch preserves event-sensitive reactors, and out-of-order is handled identically (`executeBatch` sorts by occurredAt + same checkpoint re-fold). A fold can still opt out via `options.coalesceMaxBatch = 1`.

langwatch/src/server/event-sourcing/projections/redisCachedFoldStore.ts
rogeriochaves: Conscious tradeoff, and safe to take. The TTL is now env-overridable via `LANGWATCH_FOLD_CACHE_TTL_SECONDS` (default 300, read at call time like `LANGWATCH_DISPATCH_TENANT_CAP`), so ops can dial residency down without a redeploy if Redis memory tightens. It is also a secondary lever now: coalescing collapses an aggregate's in-flight reads to one per batch, so the TTL only affects ClickHouse fallback reads across separate dispatches, not the hot path. With #4198 capping payloads, resident states are bounded. Accepting the 300s default.

langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
rogeriochaves: `traceSummary` defines no `eventLoader`, so the out-of-order re-fold branch never fires for it — it falls through to apply-on-top, exactly as single-event `execute()` does today. For folds that do set `eventLoader`, the read-your-writes assumption is identical to the existing `execute()` path; the batch only widens the comparison window, not the loader source. Accepting as-is.

rogeriochaves:
@sergioestebance both addressed:

**P1 (event-type-sensitive reactors, e.g. `scenarioExecution` on `lw.simulation_run.queued`):** fixed by dispatching reactors **per event** in `processFoldProjectionBatch` (final fold state to all). Every event now reaches its reactors, so `scenarioExecution`'s `queued` trigger fires even inside a coalesced batch. Kept coalescing default-on rather than opt-in (maintainer call): per-event dispatch makes it correct for all folds, the final folded state equals sequential folding, and out-of-order is handled identically (`executeBatch` sorts by occurredAt + same checkpoint re-fold). Unit tests assert per-event reactor invocation on a coalesced batch >= 2; a ClickHouse integration test verifies the persisted count.

**P2 (builder `options.coalesceMaxBatch` dropped):** correct that the builder's 3rd-arg options is not plumbed onto the definition — that is true for all fold options, not just this one. The working override path is the projection instance's own `options` field (the same place `killSwitch` lives), which `initializeFoldQueues` reads. Default-on means the common case needs no override; a fold opts out with `options.coalesceMaxBatch = 1` on its definition. I corrected the code comment to reference instance options so it no longer implies the builder API.

sergioestebance (dismissed):

sergioestebance:
Non-blocking follow-up from re-review of final head `25cc06c`:

- `projectionRouter.ts` still has stale JSDoc above `processFoldProjectionBatch` saying reactors fire once with the final folded state. The implementation now dispatches reactors for every folded event, which is the fix we wanted.
- `specs/event-sourcing/redis-fold-cache.feature` still has a couple stale references to fire-and-forget/asynchronous ClickHouse writes, while the current implementation is ClickHouse-first and only caches in Redis after the persistent write succeeds.
- `redisCachedFoldStore.ts` says `LANGWATCH_FOLD_CACHE_TTL_SECONDS` is read “at call time”, but it is read when the store instance is constructed.

I would not block merge on these; the earlier correctness findings look addressed.

PR #4200, opened 2026-05-26
fix(event-sourcing): bound trace summary fold state so no single trace can monopolize the queue

rogeriochaves:
## Problem

A fold projection reads its previous state, applies an event, and writes it back. The trace summary hoists every span's tracked events into one **append-only `events[]` list** (`accumulateEvents` does `[...state.events, ...newEvents]`), and each event carries an arbitrary attribute payload. A long, high-span trace (for example an agent run emitting tracked events on hundreds of spans) grows this list into the multi-megabyte range.

Once the state outgrows the Redis write-through cache it stops being a healthy cache entry, so every fold step falls back to a full read from ClickHouse. That makes folding quadratic for that aggregate, and because the event-sourcing queue is single-threaded one such trace can saturate it and starve every other tenant.

This is the read-side gap the coalescing work could not cover: coalescing batches events within a backed-up group, but a state too large to cache re-reads in full on every step regardless of batching.

### What is and is not the accumulator

- `events[]` is the **only** field that grows with span count. These are OTel span events (tracked events such as `langwatch.track_event` payloads, exceptions, RAG/feedback events) hoisted to the trace level. They are not used for cost or latency, which accumulate in separate numeric fields.
- `computedInput` / `computedOutput` are **replaced**, not accumulated — the projection picks the best single input/output per precedence rules. They are bounded by one span's payload (already capped at ingestion by the store-side span-attribute cap), not by span count, so they are deliberately left untouched here.

## Fix

Bound the one field that grows with span count:

- Cap `events[]` at a 256 KiB byte budget. The earliest events are kept; when later ones are dropped the summary sets `langwatch.reserved.events_truncated`. The full events always remain on the individual spans.
- Observe serialized fold-state size on a new `es_fold_state_bytes` histogram and log a warning when a state is oversized, so an unbounded accumulation is visible on a metric instead of silently degrading to a quadratic re-read path.

Net effect: no single trace can grow a state too large to cache, so no single trace can turn folding quadratic and monopolize the queue.

Composes with the store-side span-attribute cap (#4198) and the read-side group coalescing (#4199) already in place.

## Tests

- Unit: `events[]` byte cap (within budget kept in full, over budget capped with earliest preserved and drop flagged, already-full list stops appending, bounded across hundreds of fold steps)
- Unit: oversized state recorded on the metric and still cached (visibility, not a silent ClickHouse-only fallback)
- All existing trace-summary projection and service tests still pass

## Spec

`specs/event-sourcing/bounded-fold-state.feature`


langwatch/src/server/event-sourcing/projections/redisCachedFoldStore.ts
```diff
@@ -111,9 +123,18 @@ export class RedisCachedFoldStore<State>
     await this.inner.store(state, context);
 
     // 2. Redis second — cache for fast reads on next fold step
+    const serialized = JSON.stringify(state);
```
0xdeafcafe, line 126: json seralize now happens outside the try/catch, are we safe here?

langwatch/src/server/event-sourcing/projections/redisCachedFoldStore.ts
```diff
@@ -6,12 +6,24 @@ import {
...
   incrementEsFoldCacheRedisError,
+  observeEsFoldStateBytes,
 } from "~/server/metrics";
 import type { FoldProjectionStore } from "./foldProjection.types";
 import type { ProjectionStoreContext } from "./projectionStoreContext";
 
 const logger = createLogger("langwatch:event-sourcing:redis-cached-fold-store");
 
+/**
+ * Size past which a serialized fold state is considered too large to be a
```
0xdeafcafe, line 17: ja @sergioestebance and i talked alot today about exactly sizing.

0xdeafcafe (approved):
nice, but a couple of comments

PR #4209, opened 2026-05-27
fix(dispatch): defer over-cap tenant groups to prevent scan budget starvation

sergioestebance:
## Summary

- When a tenant hits the soft-cap (50 active groups), their queued groups remained at past scores in the ready sorted set. The dispatch Lua scans up to 10K entries; when over-cap groups dominate the head (as happened with the `assignTopic` backfill from the topic clustering fix), the entire scan budget is exhausted re-scanning groups that are always skipped — **starving all other tenants**.
- Fix: when dispatch determines a group belongs to an over-cap tenant, `ZADD` it with `nowMs + 5000` to push it past the dispatch window. The next poll immediately reaches other tenants. After 5s the group becomes eligible again.
- Applied to both `DISPATCH_LUA` (single) and `DISPATCH_BATCH_LUA` (batch) scripts.

## Root cause

A customer report led to discovering that after the topic clustering fix landed, `assignTopic` commands flooded the queue (~57K groups from one project alone). Each `assignTopic` creates ~12 queue groups (command + fold + 10 reactors). Two projects at their tenant cap of 50 had ~10K+ groups at the head of the ready sorted set with past scores, exhausting the 10K scan budget on every dispatch poll and returning nil — preventing all other tenants from being processed.

## Test plan

- [ ] New integration test: "defers over-cap groups to a future score so the next dispatch reaches other tenants immediately"
- [ ] New integration test: "dispatchBatch defers over-cap groups to future scores"
- [ ] Existing tenant cap tests pass (ZADD is additive — only changes scores for over-cap groups, no impact on under-cap dispatch flow)
- [ ] Verify in production: ready set stops growing, other tenants' queues drain

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
@@ -2627,6 +2627,86 @@ describe("GroupStagingScripts", () => {
       expect(await redis.get(tenantCounterKey("proj_quiet"))).toBe("1");
     });
 
+    /** @scenario Over-cap groups are deferred so they don't starve other tenants on repeated polls */
+    it("defers over-cap groups to a future score so the next dispatch reaches other tenants immediately", async () => {
+      process.env[TENANT_CAP_ENV] = "1";
+
+      // proj_noisy stages 50 groups at score=1000 (head of zset)
+      const noisyJobs = Array.from({ length: 50 }, (_, i) =>
+        makeJob({
+          groupId: `proj_noisy/g${i}`,
+          stagedJobId: `noisy-j${i}`,
+          dispatchAfterMs: 1000,
+        }),
+      );
+      await scripts.stageBatch(noisyJobs);
+
+      // proj_quiet stages 1 group at score=1001 (later in zset)
+      await stageOne({
+        tenantId: "proj_quiet",
+        groupSuffix: "only",
+        stagedJobId: "quiet-j1",
+        dispatchAfterMs: 1001,
+      });
+
+      // First dispatch: proj_noisy/g0 wins
+      const first = await scripts.dispatch({ nowMs: 2000, activeTtlSec: 60 });
+      expect(first?.groupId).toMatch(/^proj_noisy\//);
+
+      // Second dispatch: proj_quiet wins (walk past over-cap noisy)
+      const second = await scripts.dispatch({ nowMs: 2000, activeTtlSec: 60 });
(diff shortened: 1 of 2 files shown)


Aryansharma28 (approved):

sergioestebance:
## Validation: Redis state + safety analysis

### Current Redis state (pre-deploy)

| Metric | Value | Trend |
|--------|-------|-------|
| Ready groups | 210K | Growing ~180/sec |
| Pending counter | 294K | Growing |
| Completed total | 332M | Draining ~250/sec |
| `assignTopic` completions | 349K | **Frozen — zero completions** |
| Tenant active sum | 156 | Stable |
| Memory | 8.58 GB / 77.76 GB | OK |
| DBSIZE | 666K | Growing |

**Two tenants at cap=50, filling 100% of the 10K scan window:**

First 10K dispatchable groups in the ready set contain only:
- Tenant A: 9,621 groups (96%)
- Tenant B: 379 groups (4%)
- **Other tenants: 0 (starved)**

### Why the fix is safe

| Concern | Assessment |
|---|---|
| `STAGE_LUA` uses `ZADD LT` which can un-defer groups | Self-healing: re-deferred on next poll. New jobs mostly create new groups (unique traceIds), not update existing deferred ones |
| 5s latency after a slot is freed | Acceptable for backfill workload. 50 concurrent completions → ~50 completions per 5s window, so deferred groups get picked up in batches |
| Pagination skipping entries after re-scoring | Harmless: skipped entries are still at past scores and get deferred on the next dispatch call (milliseconds later). Enough entries cleared per call to expose other tenants |
| `tenantCapCache` correctness within EVAL | Correct: single EVAL = atomic in Redis single-threaded model, no completions can happen mid-scan |
| `cap=0` (kill switch) path | Unaffected: `if tenantCap > 0` guard means the ZADD never fires when cap is disabled |
| `COMPLETE_LUA` interaction | No conflict: it re-scores its OWN group (no `LT` flag), and DECRs the tenant counter. Deferred groups of the same tenant become eligible after their 5s window |
| `REFRESH_LUA` (heartbeat) | Only touches active groups (which have an active key). Deferred groups have no active key, so refresh never fires for them |

### Expected behavior after deploy

1. First dispatch call: defers ~10K over-cap groups to `nowMs + 5000`
2. Second dispatch call (ms later): immediately reaches other tenants' 200K+ dispatchable groups
3. Other tenants start draining immediately
4. After 5s, deferred groups re-enter the window — if tenant freed a slot they dispatch, otherwise deferred again
5. `assignTopic` stays frozen until the tenant drops below cap (by completing other job types), which is expected

PR #4211, opened 2026-05-27
fix(event-sourcing): memoize trace read-derivation to stop per-event read amplification

rogeriochaves:
## What

Memoize the trace read-derivation (`deriveEvents` / `deriveScenarioRoleMetrics`) per `(tenantId, traceId, occurredAtMs)` so a coalesced fold batch reads stored spans once, not once per dispatched reactor/event.

## Why

Live incident: `/api/health/processor` timed out and the GroupQueue stalled at ~226k groups in `ready` (not memory, zero Redis evictions). Root cause was read amplification on the recovery path:

- The fold coalesces a backed-up group into one load/apply/store cycle (O(1) per span), but `projectionRouter` still dispatches reactors once per event so per-span reactors keep their work.
- Reactors that read trace-level data derived from `stored_spans` (alert triggers, scenario role metrics) re-issue a multi-MB all-spans ClickHouse read on every event. All events in a coalesced batch carry the same final `foldState.occurredAt`, so for an N-span trace the identical read ran up to N times.
- During a backlog drain this multiplied both the big ClickHouse reads and the fold-state bandwidth on the single-threaded Redis. Adding workers made it worse (8 workers drove Redis to ~37.7k ops/sec and 98% engine CPU), so scaling was not viable until the per-job cost dropped.

## Fix

A bounded, short-window memo in `TraceReadDerivationService`:

- Keyed by `(tenantId, traceId, occurredAtMs)`. The cutoff makes each read deterministic, so sharing it across a batch is safe.
- Single-flight: concurrent callers share the in-flight read; callers within the window share the resolved result.
- Live reads (no cutoff) pass straight through, never cached.
- Failed reads are not cached, so the next caller retries instead of replaying the error.
- Bounded entry count with expiry-first eviction so a burst of distinct traces cannot grow it without bound.

The per-event reactor dispatch is unchanged, so per-span reactors still see every event.

## Tests

`trace-read-derivation.service.unit.test.ts` (7 tests): N derivations for one trace+cutoff read once, concurrent reads collapse to one, a later cutoff reads again, live reads are not memoized, the window expiring re-reads, and a failed read is not cached.

Scenarios added to `specs/event-sourcing/fold-coalescing.feature` and bound.

## Follow-ups (non-blocking, not in this hotfix)

- Unify the 3 per-pipeline `TraceReadDerivationService` instances (registry deriveEvents x2 + computeRunMetrics) into one shared instance so the memo also collapses cross-pipeline same-trace reads. The in-pipeline hot path (where the amplification lives) is already covered, so this is an optimization, not a correctness gap.
- The memo keys on `spanCount`, which does not increment when a span is re-exported with the same `spanId` (content change, same count). A derivation could then reuse the prior result within the 30s window. Far narrower than the old constant-key bug, bounded by the window, and tolerable for alert/scenario eventual consistency. A stronger watermark (an event/version counter that ticks on every applied event) would close it.

### DIFF
--- a/langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
+++ b/langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
@@ -148,11 +148,15 @@ export class TraceReadDerivationService {
   }
 
   private evict<T>(cache: Map<string, MemoEntry<T>>, now: number): void {
-    if (cache.size <= DERIVATION_MEMO_MAX_ENTRIES) return;
+    // Entries share a fixed TTL and are re-inserted at the back on each miss,
+    // so the front is always the soonest to expire. Sweep expired entries from
+    // the front and stop at the first live one (amortized O(1) per insert, not
+    // O(n) per call), so expired promises are released even when the cache
+    // never reaches the size cap. Then enforce the cap on whatever remains.
     for (const [k, entry] of cache) {
-      if (entry.expiresAt <= now) cache.delete(k);
+      if (entry.expiresAt > now) break;
+      cache.delete(k);
     }
-    // Map preserves insertion order, so the front is the oldest entry.
     while (cache.size > DERIVATION_MEMO_MAX_ENTRIES) {
       const oldest = cache.keys().next().value;
       if (oldest === undefined) break;
--- a/langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
+++ b/langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
@@ -24,20 +24,31 @@ export interface NormalizedSpanReader {
 interface DeriveParams {
   tenantId: string;
   traceId: string;
+  /**
+   * ClickHouse partition hint (the trace's EARLIEST span time). It narrows the
+   * partitions scanned; it is NOT a freshness cutoff and does not bound which
+   * spans come back, so it must not key the memo.
+   */
   occurredAtMs?: number;
+  /**
+   * Monotonic fold watermark — the fold's `spanCount`, which increments every
+   * time a span is folded. The memo is keyed on it so a cached derivation is
+   * reused only within one fold version (all of a coalesced batch's per-event
+   * reactors observe the same final state, so they share one read) and is
+   * dropped the moment newer spans land. Omit it to bypass the memo (a live
+   * read with no watermark must always hit storage).
+   */
+  foldVersion?: number;
 }
 
 /**
- * Window during which a derivation for the same (tenant, trace, cutoff) reuses
- * the in-flight or just-resolved span read. A coalesced fold batch dispatches
- * its reactors per event but with one shared final fold state, so every
- * reactor derives at the same occurredAt cutoff back-to-back; this window
- * collapses those identical reads into one. Kept short because the cutoff makes
- * each read deterministic, so the only thing aging out matters for is rare
- * out-of-order late spans.
+ * Window after which a memo entry is dropped purely as a memory backstop —
+ * correctness comes from the fold-version key, not from aging. An entry for a
+ * superseded version is simply never read again, so this only bounds how long
+ * an unused entry lingers.
  */
 const DERIVATION_READ_WINDOW_MS = 30_000;
-/** Cap so a burst of distinct traces can't grow the memo without bound. */
+/** Cap so a burst of distinct traces/versions can't grow the memo without bound. */
 const DERIVATION_MEMO_MAX_ENTRIES = 2_000;
 
 interface MemoEntry<T> {
@@ -55,11 +66,11 @@ interface MemoEntry<T> {
  *
  * The all-spans read these derivations issue is multi-MB for large traces. A
  * coalesced fold batch fires its reactors once per event but at one shared
- * occurredAt cutoff, so without coalescing the same read would run once per
+ * final fold state, so without coalescing the same read would run once per
  * span in the backlog — the read-amplification that re-saturated Redis/ClickHouse
- * during a backlog drain. The derivations are memoized per (tenant, trace,
- * cutoff) so a batch reads stored spans once regardless of how many reactors or
- * events it dispatches.
+ * during a backlog drain. The derivations are memoized per (tenant, trace, fold
+ * version) so a batch reads stored spans once regardless of how many reactors or
+ * events it dispatches, while a fold that has advanced (new spans) re-reads.
  */
 export class TraceReadDerivationService {
   private readonly spanCostService = new SpanCostService();
@@ -75,7 +86,11 @@ export class TraceReadDerivationService {
     params: DeriveParams,
   ): Promise<ScenarioRoleMetrics> {
     return this.memoize(this.scenarioRoleMetricsMemo, params, async () => {
-      const spans = await this.spans.getNormalizedSpansByTraceId(params);
+      const spans = await this.spans.getNormalizedSpansByTraceId({
+        tenantId: params.tenantId,
+        traceId: params.traceId,
+        occurredAtMs: params.occurredAtMs,
+      });
       return deriveScenarioRoleMetricsFromSpans({
         spans,
         spanCostService: this.spanCostService,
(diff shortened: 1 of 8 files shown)


rogeriochaves:
Review finding:

- **P1** `langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts:97`: the memo key treats `occurredAtMs` as a deterministic cutoff, but the ClickHouse readers only use it as a ±2 day partition hint, and `traceSummary.occurredAt` is the trace's earliest span timestamp. After the first derivation for a trace, later spans for the same trace usually keep the same `occurredAtMs`, so any alert/event derivation or scenario-role metrics derivation within the 30s memo window can reuse stale stored-spans / trace-events results and miss the newly arrived span data. Minimal fix direction: key the memo by a value that changes when the fold has advanced, such as `LastEventOccurredAt` or another fold/version watermark, or pass an explicit per-batch memo scope instead of caching by the partition hint alone.

Verification run locally against latest branch commit `6cfa72713`:
- `git diff --check origin/main...HEAD` passed
- `pnpm --dir langwatch install --frozen-lockfile` completed, with the existing `langwatch-mcp-server` bin warning
- `pnpm --dir langwatch exec prisma generate` passed
- `pnpm --dir langwatch exec vitest run src/server/app-layer/traces/__tests__/trace-read-derivation.service.unit.test.ts --reporter=dot` passed, 7 tests

langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
```diff
@@ -27,30 +27,116 @@ interface DeriveParams {
...
+  }
+
+  /**
+   * Only deterministic reads are memoized: a derivation with an occurredAt
+   * cutoff returns the same spans no matter when it runs, so sharing it across
+   * a batch is safe. A live read (no cutoff) is left to pass straight through.
+   */
+  private memoKey(params: DeriveParams): string | null {
+    if (params.occurredAtMs === undefined) return null;
+    return `${params.tenantId}:${params.traceId}:${params.occurredAtMs}`;
```
sergioestebance: P1: This key treats `occurredAtMs` as a deterministic fold cutoff, but the value passed from the reactors is `traceSummary.occurredAt` / `foldState.occurredAt`, which is the earliest span start time. That usually does not change when later spans arrive. The ClickHouse readers also use `occurredAtMs` only as a ±2 day partition hint, not as a snapshot cutoff. A follow-up batch for the same trace within 30s can therefore reuse stale derived events or scenario role metrics and miss the newly arrived spans. Key this on a value that changes with the fold state, such as `spanCount`, `updatedAt`, or another projection revision passed from the reactor context.

langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
```diff
@@ -27,30 +27,116 @@ interface DeriveParams {
...
+    // instead of replaying the rejection for the whole window.
+    value.catch(() => {
+      if (cache.get(key)?.value === value) cache.delete(key);
+    });
+    this.evict(cache, now);
+    return value;
+  }
+
+  private evict<T>(cache: Map<string, MemoEntry<T>>, now: number): void {
+    if (cache.size <= DERIVATION_MEMO_MAX_ENTRIES) return;
```
sergioestebance: P2: Expired entries are only pruned when the cache is already over `DERIVATION_MEMO_MAX_ENTRIES`. If a worker handles fewer than 2,000 distinct trace keys, expired promises/results stay retained indefinitely despite the 30s window. Prune expired entries before this early return, or otherwise sweep them on each memoized read before applying the size cap.

sergioestebance (commented):
Validated locally against the PR branch. The focused unit test passes, but I found two issues in the memoization behavior.

rogeriochaves:
Fixed in 68104c85c. The P1 is correct and I rebuilt the memo key around it.

The memo no longer keys on `occurredAtMs` (the trace's earliest span time / ClickHouse partition hint, which stays constant as the trace grows). It now keys on the fold's `spanCount` watermark, threaded in as `foldVersion`:

- Within one coalesced batch every per-event reactor observes the same final fold state, so `spanCount` is identical and the N reads collapse to one (the amplification fix is intact).
- The moment newer spans are folded, `spanCount` advances, the key changes, and the derivation re-reads, so it can never serve stale stored-spans / trace-events.
- `occurredAtMs` is still passed through to the ClickHouse reader as the partition hint, just not used as a cache key. Live reads (no `foldVersion`) bypass the memo entirely.

Threaded `foldVersion: <state>.spanCount` from all three call sites (alertTrigger, evaluationAlertTrigger, computeRunMetrics) and their dep signatures. Added a regression test asserting that an advanced `foldVersion` with an unchanged `occurredAtMs` re-reads (the exact stale-data case). 8/8 unit tests green, parity green.

langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
rogeriochaves: Already fixed in 68104c85c — this was against the earlier SHA. The memo is now keyed on the fold's spanCount watermark (passed as foldVersion), exactly your suggestion: identical across one coalesced batch so reads still collapse, but it advances the instant new spans fold so a follow-up batch re-reads. occurredAtMs is kept on the actual ClickHouse read as the partition hint, just dropped from the key. Regression test asserts an advanced foldVersion with unchanged occurredAtMs re-reads.

langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
rogeriochaves: Fixed in 6b442267f. evict() now sweeps expired entries from the front on every miss, not only once over the size cap, so expired promises are released even when the cache stays under DERIVATION_MEMO_MAX_ENTRIES. Entries share a fixed TTL and are re-inserted at the back on each miss, so the front is always soonest-to-expire — the sweep stops at the first live entry (amortized O(1) per insert, no O(n) per read).

rogeriochaves:
Re-review after the implementation change:

No findings. The prior stale-cache concern is addressed by keying the memo on `foldVersion` (`spanCount`) and treating `occurredAtMs` only as the ClickHouse partition hint. The updated call sites pass `foldVersion` from the trace fold state / trace summary consistently for trace alerts, evaluation alerts, and simulation metric derivation.

Residual risks / follow-ups:
- As noted in the PR body, `spanCount` will not change for a same-spanId re-export/content update, so that narrow case can still reuse a result within the 30s window. That looks bounded and non-blocking for this hotfix.
- The trace-event and scenario-role memo caches are still per `TraceReadDerivationService` instance, so cross-pipeline same-trace reads are not collapsed. That is an optimization gap, not a correctness issue.

Verification run locally against `05d569eb6`:
- `git diff --check origin/main...HEAD` passed
- `pnpm --dir langwatch run types:zod:generate` passed, with existing generator warnings
- `pnpm --dir langwatch exec vitest run src/server/app-layer/traces/__tests__/trace-read-derivation.service.unit.test.ts src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/alertTrigger.reactor.unit.test.ts src/server/event-sourcing/pipelines/evaluation-processing/reactors/__tests__/evaluationAlertTrigger.reactor.unit.test.ts src/server/event-sourcing/pipelines/simulation-processing/commands/__tests__/computeRunMetrics.command.unit.test.ts --reporter=dot` passed: 3 files, 27 tests. Note: the alertTrigger unit-test path does not exist in this tree, so Vitest ran the three existing matching suites.

langwatch/src/server/app-layer/traces/trace-read-derivation.service.ts
```diff
@@ -34,23 +63,100 @@ interface DeriveParams {
...
+    // insertion order tracking last read.
+    cache.delete(key);
+    cache.set(key, { value, expiresAt: now + DERIVATION_READ_WINDOW_MS });
+    // Never cache a failed read: drop the entry so the next caller retries
+    // instead of replaying the rejection for the whole window.
+    value.catch(() => {
+      if (cache.get(key)?.value === value) cache.delete(key);
+    });
+    this.evict(cache, now);
+    return value;
```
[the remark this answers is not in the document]
rogeriochaves: Valid — addressed in follow-up #4213 (this PR was already merged). expiresAt is no longer stamped at read start: an in-flight entry holds +Infinity so concurrent callers always share it, and the 30s window starts only when the read resolves (measuring the age of the result, not the read's start). So a read slower than the window stays shareable instead of expiring mid-flight and triggering duplicate reads. Regression test added (read held in flight + clock advanced past the window -> one read, not two).

PR #4214, opened 2026-05-27
fix(event-sourcing): park over-cap groups out of ready instead of re-deferring

rogeriochaves:
## Why

Over the 2026-05-27 incident the dispatch loop saturated single-threaded Redis. The root was the tenant soft-cap's over-cap handling: when a tenant was at its in-flight cap, DISPATCH re-scored each of its over-cap groups to a future score (`ZADD ready now+5s`) on every poll. With a 442K-deep ready set and several workers polling, that became ~90K ZADD/sec, most of the command throughput, while the actual backlog barely moved.

The defer was meant to stop repeated rescans starving other tenants, but the write volume scaled with backlog size, which is exactly what a backlog drain cannot afford.

## What

Move over-cap groups OUT of the ready set into a per-tenant parked zset, once, instead of re-scoring them in place every poll. The dispatch scan never sees them again until they are restored, so write volume decouples from backlog size.

- `parkGroup` moves an over-cap group from ready to `parked:<tenant>` (score preserved) and registers the tenant in a `parked-tenants` set. This replaces the per-poll defer in DISPATCH and DISPATCH_BATCH.
- COMPLETE frees an in-flight slot and restores up to `cap - active` of the tenant's parked groups. This is the normal path back.
- DISPATCH runs a time-gated reconcile (every 2s, skipped when nothing is parked) that restores parked groups for any tenant now under cap. It backstops two cases COMPLETE cannot: a crashed tenant whose in-flight counter TTL-expired without completing, and the operator setting the cap to 0 (drains everything parked).

### Mutual exclusivity

A group is in exactly one of ready, parked, blocked, or active. Every writer that returns a group to ready now routes through a shared `addToReadyOrParked` helper that keeps an already-parked group parked rather than clobbering it back into the scan: STAGE, STAGE_BATCH, RETRY_RESTAGE, COMPLETE re-stage, and the ops repo UNBLOCK and REPLAY_FROM_DLQ paths. Missing one would silently re-create the storm, so the test matrix covers each.

### Observability

We were blind to the storm while it happened. This adds a `gq_parked_groups` Prometheus gauge per queue, plus parked depth on the ops dashboard (a tile and a line on the throughput history). A sustained parked spike is the over-cap signal that previously surfaced only as an invisible write storm; a non-draining floor flags a parked-group strand.

### Safety

All cap logic stays gated behind `tenantCap > 0`, so the current `LANGWATCH_DISPATCH_TENANT_CAP=0` mitigation path is unchanged and creates zero parked keys. Restores are bounded by `cap - active` so COMPLETE-unpark and the reconcile cannot over-restore into churn. Parked keys share the queue hash tag, so they stay in one Redis Cluster slot.

## Tests

Testcontainer integration coverage for: park on over-cap, unpark on completion with score preservation, orphan reconcile when the in-flight counter expires, cap-to-0 drain, bounded restore, and every ready-writer respecting parked. The existing over-cap deferral scenarios are rewritten for parking in `specs/event-sourcing/tenant-soft-cap.feature`, and this supersedes the deferral introduced in #4209 and bound in #4210.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
@@ -5,6 +5,7 @@ const metricNames = [
   "gq_active_groups",
   "gq_pending_groups",
   "gq_blocked_groups",
+  "gq_parked_groups",
   "gq_groups_blocked_total",
   "gq_jobs_staged_total",
   "gq_jobs_dispatched_total",
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
@@ -51,6 +51,12 @@ export const gqBlockedGroups = new Gauge({
   labelNames: ["queue_name"] as const,
 });
 
+export const gqParkedGroups = new Gauge({
+  name: "gq_parked_groups",
+  help: "Number of groups parked out of the ready scan because their tenant is at the in-flight soft cap. A sustained spike is the over-cap signal that previously surfaced only as an invisible dispatch-write storm; a non-draining floor flags a parked-group strand.",
+  labelNames: ["queue_name"] as const,
+});
+
 export const gqJobsStagedTotal = new Counter({
   name: "gq_jobs_staged_total",
   help: "Total number of jobs staged into the group queue",
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
@@ -106,6 +106,29 @@ local function unparkUpTo(readyKey, tenantId, slots)
   end
   return moved
 end
+
+-- Safety-net reconcile, run by DISPATCH on a cadence (not the normal path —
+-- COMPLETE-unpark frees slots as jobs finish). Restores parked groups whose
+-- tenant has dropped below cap, covering two cases COMPLETE cannot:
+--   * orphan recovery (TRAP 2): a crashed/timed-out tenant never fires COMPLETE,
+--     so its tenant_active key TTL-expires; read missing as 0 = under cap = unpark,
+--     else the parked groups strand forever (a new stranding mode).
+--   * cap disabled: flipping LANGWATCH_DISPATCH_TENANT_CAP to 0 must not leave
+--     groups parked out of the dispatch scan, so drain the whole tenant.
+-- Self-limiting via unparkUpTo (bounded by cap - active), so it never over-unparks.
+local function reconcileParked(readyKey, keyPrefix, tenantCap)
+  local tenants = redis.call("SMEMBERS", keyPrefix .. "parked-tenants")
+  for _, tenantId in ipairs(tenants) do
+    local slots
+    if tenantCap > 0 then
+      local active = tonumber(redis.call("GET", keyPrefix .. "tenant_active:" .. tenantId)) or 0
+      slots = tenantCap - active
+    else
+      slots = redis.call("ZCARD", keyPrefix .. "parked:" .. tenantId)
+    end
+    unparkUpTo(readyKey, tenantId, slots)
+  end
+end
 `;
 
 const STAGE_LUA = TTL_HELPER_LUA + PARK_HELPER_LUA + `
@@ -175,7 +198,7 @@ end
 
 -- Score = earliest pending dispatchAfter (LT keeps the smallest score per group)
 if shouldUpdateReady() then
-  redis.call("ZADD", readyKey, "LT", dispatchAfter, groupId)
+  addToReadyOrParked(readyKey, groupId, dispatchAfter, true)
 end
 
 redis.call("LPUSH", signalKey, "1")
@@ -187,7 +210,7 @@ redis.call("INCR", totalPendingKey)
 return 1
 `;
 
-const STAGE_BATCH_LUA = TTL_HELPER_LUA + `
+const STAGE_BATCH_LUA = TTL_HELPER_LUA + PARK_HELPER_LUA + `
 local readyKey        = KEYS[1]
 local signalKey       = KEYS[2]
 local totalPendingKey = KEYS[3]
@@ -263,7 +286,7 @@ for groupId, minScore in pairs(affectedGroups) do
   -- here would re-expose the group to ZRANGEBYSCORE before the next refresh.
   local activeKey = keyPrefix .. "group:" .. groupId .. ":active"
   if redis.call("EXISTS", activeKey) == 0 and redis.call("SISMEMBER", blockedKey, groupId) == 0 then
-    redis.call("ZADD", readyKey, "LT", minScore, groupId)
+    addToReadyOrParked(readyKey, groupId, minScore, true)
   end
 end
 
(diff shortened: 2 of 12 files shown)


rogeriochaves:
@coderabbitai review

rogeriochaves:
Review findings:

- **P2** `langwatch/src/server/app-layer/ops/metrics-collector.ts:783` / `langwatch/src/components/ops/dashboard/ThroughputChart.tsx:58`: the persisted ops metrics state still accepts `version: 3` and restores the old `throughputBuffer` unchanged, but this PR adds required `parkedCount` reads in the chart/downsampling path. Any deploy with an existing `ops:metrics:state` v3 payload will restore old points that do not have `parkedCount`, so the chart receives `undefined`/`NaN` for the new Parked series until the old buffer ages out. Minimal fix: bump the persisted state version, or backfill `parkedCount: 0` when restoring older points.

- **P3** `langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts:4`: `gq_parked_groups` is added as a Gauge, but it is missing from the `metricNames` cleanup list used before registering metrics. A hot reload or test isolate that re-evaluates this module after the new gauge exists can fail with duplicate metric registration for `gq_parked_groups`. Minimal fix: add `"gq_parked_groups"` to `metricNames`.

Verification run locally against `9f60d96b0`:
- `git diff --check origin/main...HEAD` passed
- `pnpm --dir langwatch install --frozen-lockfile` completed, with the existing `langwatch-mcp-server` bin warning
- `pnpm --dir langwatch exec prisma generate` passed
- `pnpm --dir langwatch run types:zod:generate` passed, with existing generator warnings
- `pnpm --dir langwatch exec vitest run src/server/app-layer/ops/__tests__/build-pipeline-tree.unit.test.ts src/server/app-layer/ops/__tests__/queue.service.unit.test.ts --reporter=dot` passed: 2 files, 33 tests
- `pnpm --dir langwatch exec vitest run -c vitest.integration.config.ts src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts --reporter=dot` passed: 1 file, 127 tests

rogeriochaves:
Thanks for the review. Both addressed:

- **P3** (gq_parked_groups missing from the cleanup list): already fixed in `17046d1a9` (it was committed after the `9f60d96b0` you reviewed). `"gq_parked_groups"` is now in the `metricNames` removal list in `metrics.ts`, so re-evaluating the module no longer throws a duplicate registration.
- **P2** (persisted throughput history): fixed in `2d749c68b` via backfill rather than a version bump. `restoreState` now maps restored points to `parkedCount: pt.parkedCount ?? 0`. I kept `version: 3` deliberately: bumping it discards the whole snapshot, which also zeroes the accumulated peaks (peakCompletedPerSec/Ingested/Failed/latency), and this same PR adds a Completed/s peak tile that would then read low until those sticky maxes re-climb. The backfill preserves both the rolling history and the peaks across the deploy.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -32,7 +32,106 @@ local function refreshGroupKeyTtl(jobsKey, dataKey, nowMs)
...
+-- Self-limiting via unparkUpTo (bounded by cap - active), so it never over-unparks.
+local function reconcileParked(readyKey, keyPrefix, tenantCap)
+  local tenants = redis.call("SMEMBERS", keyPrefix .. "parked-tenants")
+  for _, tenantId in ipairs(tenants) do
+    local slots
+    if tenantCap > 0 then
+      local active = tonumber(redis.call("GET", keyPrefix .. "tenant_active:" .. tenantId)) or 0
+      slots = tenantCap - active
+    else
+      slots = redis.call("ZCARD", keyPrefix .. "parked:" .. tenantId)
```
sergioestebance: P1: This makes the documented `LANGWATCH_DISPATCH_TENANT_CAP=0` kill switch potentially unbounded. In the May 27 incident we saw hundreds of thousands of over-cap ready entries; after this PR parks that backlog, flipping the cap to `0` sets `slots` to the full `ZCARD(parked:<tenant>)`, and `unparkUpTo()` then does one `ZRANGE` plus `ZREM`/`ZADD` for every parked group inside a single Redis Lua eval. That can recreate the single-thread Redis saturation the incident docs are trying to avoid. Please drain cap=0 in bounded chunks per reconcile and keep the tenant registered until the parked set is empty.

langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
```diff
@@ -51,6 +52,12 @@ export const gqBlockedGroups = new Gauge({
   labelNames: ["queue_name"] as const,
 });
```
sergioestebance, line 54: P2: Since this registers a new `gq_parked_groups` metric, it also needs to be added to the `metricNames` hot-reload cleanup array at the top of the file. Otherwise test bootstrap/hot reload can hit duplicate metric registration on module re-import.

sergioestebance (commented):
Reviewed against the May 21-27 Redis incident notes. Two findings below; the cap=0 drain issue is the one I would block on because cap=0 is documented as an operator mitigation path.

langwatch/src/server/event-sourcing/queues/groupQueue/metrics.ts
rogeriochaves: Fixed in 17046d1a9 — `gq_parked_groups` is now in the `metricNames` hot-reload cleanup array at the top of metrics.ts, so re-registering on a 2nd module eval (vitest shard / HMR) no longer throws "already registered". This was also the root of the orgBillableEventsMeter shard failure: that file transitively imports metrics.ts, so the dup-registration aborted it at import time — collateral, not a dedup-logic bug. Verified the affected unit shard green.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
rogeriochaves: Confirmed valid on 2d749c68b. `reconcileParked` with `tenantCap=0` sets `slots = ZCARD(parked:<tenant>)` and `unparkUpTo` then does one `ZRANGE 0..slots-1 WITHSCORES` plus a per-group ZREM/ZADD loop in a single EVAL — so flipping the `LANGWATCH_DISPATCH_TENANT_CAP=0` kill switch while a tenant has hundreds of thousands parked (the May-27 incident scale) is one unbounded blocking EVAL on the single-threaded Redis, i.e. the kill switch could itself cause the stall it exists to stop. The cap>0 path is already bounded (slots = cap - active). Sharp catch. Fix in progress: bound the drain inside `unparkUpTo` (slots capped at a PARK_RECONCILE_MAX_DRAIN chunk; the tenant stays registered until parked is empty) so the 2s-gated reconcile drains a large backlog gradually in bounded EVALs. Adding a real-Redis integration test that parks > 1 chunk, flips cap=0, and asserts bounded-chunk drain across reconciles with no single giant EVAL. Will update this thread with the fix SHA.

rogeriochaves: Fixed in cc743f5bd and PROVEN on real Redis at incident scale. The drain is now bounded inside `unparkUpTo` (slots capped at `PARK_RECONCILE_MAX_DRAIN`=1000, every caller bounded; tenant stays registered until parked is empty so the 2s-gated reconcile drains gradually).

Independent at-scale probe vs a real Redis testcontainer: seeded 50,000 parked groups for one tenant, flipped `LANGWATCH_DISPATCH_TENANT_CAP=0`, drove successive reconciles. Result: drained in exactly 50 cycles, every cycle moved <= 1000 (asserted `delta <= PARK_RECONCILE_MAX_DRAIN` on every cycle — never one giant EVAL), full conservation (ready=50000, parked=0, parked-tenants registry cleared only when empty, zero loss/dup). Plus the committed bounded-drain test (1050 parked) and the full 128-test scripts suite are green on real Redis. So flipping the kill switch with a huge parked backlog no longer blocks the single thread. Resolving.

rogeriochaves:
@sergioestebance thanks for the review — both findings are addressed and their inline threads resolved:

- **P1** (cap=0 kill switch unbounded parked-drain — the block-worthy one): fixed in `cc743f5bd`. The drain is bounded inside `unparkUpTo` (slots capped at `PARK_RECONCILE_MAX_DRAIN`=1000, every caller bounded; tenant stays registered until parked is empty so the 2s-gated reconcile drains gradually). Proven on a real Redis testcontainer at incident scale: 50,000 parked, flip `LANGWATCH_DISPATCH_TENANT_CAP=0`, drained in exactly 50 bounded cycles (asserted `delta <= PARK_RECONCILE_MAX_DRAIN` every cycle — never one giant EVAL), full conservation (ready=50000, parked=0, registry cleared only when empty). The committed 1050-parked test guards this regression in CI.
- **P2** (gq_parked_groups missing from the cleanup list): fixed in `17046d1a9`.

Real-Redis verified end-to-end: 128-test scripts suite + full 141 groupQueue/scripts suites green, plus the 50k at-scale probe. Closing.

sergioestebance (approved):

PR #4216, opened 2026-05-27
feat(traces): event_log as source of truth for large trace IO + lean projections (#4215)

### LINKED ISSUE #4215: Large trace payloads: lean event-sourcing fold cache + edge blob-offload to S3
## 🧑 For Humans

Trace ingestion dataflow — ADR-022: event_log is the single source of truth, S3 reduced to a transient spool for oversize protection only. leanForProjection is one utility called at the dispatch interposition AND in replay, so live and replay produce identical projection state.

**Problem.** Large LLM trace payloads (big prompts, RAG contexts, long outputs) cause two linked problems:

1. **Redis pressure / clog.** The full OTLP span payload rides through Redis as a command-queue job *before* the existing `capOversizedAttributes(256 KB)` cap fires at the command worker. Multi-MB spans saturate Redis memory and the worker event loop.
2. **Evaluations run on truncated data.** Inputs/outputs are truncated **client-side, before they ever reach us** — the SDK at 5 KB (`... (truncated string)`), gateway at 32 KB (ADR-017). Online evaluators score incomplete text.

**Design — ADR-022 (supersedes ADR-021's edge-permanent-S3 mechanism).** [Full ADR](https://github.com/langwatch/langwatch/blob/issue4215/large-trace-blob-offload/dev/docs/adr/022-event-log-source-of-truth.md).

- **`event_log` is the single durable source of truth.** Full content lives in `EventPayload` (ZSTD-compressed). Replay-sufficient on its own.
- **S3 is reduced to a transient spool.** Only used when the serialized command payload exceeds `COMMAND_INLINE_THRESHOLD = 256 KB`. Eagerly deleted after `event_log` INSERT succeeds; 24h lifecycle policy as a safety net for orphans. **S3 is no longer in any user-facing read path.**
- **A single `leanForProjection(event)` utility** runs at two places: the dispatch interposition in `eventSourcingService.ts:242-251` (live), and in `replayExecutor.apply` (replay). Same code → projection state is path-independent.
- **`IO_PREVIEW_BYTES = 64 KB`** — sized to a **standard Claude response length** (~16K tokens, the modal output at common `max_tokens=8192` settings). `ILIKE` against `trace_summaries.ComputedInput/Output` stays lossless for the modal case; above 64 KB the user opts into "show full" which JOINs `event_log`.
- **Raise SDK truncation defaults** so full bytes flow (Python SDK done; TS SDK has no transport cap; Go gateway has no `WithSpanLimits`).

**Scope.** One feature-flagged PR (#4216). Branch `issue4215/large-trace-blob-offload`. Stays dark behind `release_trace_blob_offload`; flag off = current behavior.

**No frontend change.** Refs are resolved transparently at the read boundary (`getTracesWithSpans` JOINs `event_log` for full IO when the UI asks for it).

**Related.** ADR-007 (event sourcing), ADR-015 (replay), ADR-017 (gateway payload capture), ADR-021 (now superseded by ADR-022 on the offload mechanism; its reserved-namespace + facet-filter + differential-preview rules survive).

---

## 🤖 For Agents

> Implementation contract per ADR-022. Land as stacked commits on PR #4216. Each commit keeps tests green. TDD: failing tests written FIRST (Step 4 below), then implementation drives them to green.

### Constants (config knobs)

- `COMMAND_INLINE_THRESHOLD = 256 KB` — matches existing `capOversizedAttributes` boundary. Edge size-check threshold for spool decision.
- `IO_PREVIEW_BYTES = 64 KB` — overridable via `LANGWATCH_IO_PREVIEW_BYTES`. Applies to IO attribute previews (`langwatch.input`, `langwatch.output`, `gen_ai.input.messages`, `gen_ai.output.messages`) and log-record `body`.
- Non-IO attribute preview stays at `DEFAULT_PREVIEW_BYTES = 2 KB`.
- Reserved namespace: `langwatch.reserved.*`. New per-field eventref: `langwatch.reserved.eventref.<attrKey>` carrying `{ field: <attrKey> }`. The `eventId` is implicit in the span/event carrying the eventref. **No new `blobref` keys** — that name retired with the S3-permanent mechanism.

### Dataflow walkthrough

```
EDGE — OTLP collector
  receive OTLP request → for each span:
    compute serialized-command-payload bytes
    payload ≤ COMMAND_INLINE_THRESHOLD (256 KB)
      → regular RecordSpan command  (inline data in payload)
    payload > 256 KB:
      try S3 PUT (spool object, transient)
        success → oversized RecordSpan command  (carries {spoolRef} only)
        failure → fail-open: send regular RecordSpan command (full inline)
                  log warn ("oversize protection skipped; queue carries full payload")
  command.send() → COMMAND QUEUE (Redis GroupQueueProcessor)
  edge returns; durability lives at the queue stage, not event_log yet

COMMAND WORKER  (pulls from global queue)
  if regular:    use inline span data
  if oversized:  S3 GET(spoolRef) → reconstitute full span
  RecordSpanCommand.handle:
    stripReservedAttributes (defense; passthrough for langwatch.reserved.causality_depth)
    PII redaction · cost enrichment · token estimation
    construct SpanReceivedEvent with FULL content as data
  storeEvents → event_log INSERT (full content, ZSTD compressed)
  on success:
    if oversized → best-effort S3 DELETE(spool key)   (lifecycle policy = 24h safety net)
                                ↓
INTERPOSITION  (eventSourcingService.ts:242-251 — single hook)
  enrichedEvents = enrichedEvents.map(leanForProjection)
    leanForProjection(event):
      switch event.type:
        SpanReceived       → for each over-threshold IO attr
                             (langwatch.input/output, gen_ai.input/output.messages):
                               replace value with preview (≤ IO_PREVIEW_BYTES = 64 KB)
                               attach langwatch.reserved.eventref.<attrKey> = { field }
        LogRecordReceived  → if body > IO_PREVIEW_BYTES:
                               replace body with preview
                               attach langwatch.reserved.eventref.body = { field: "body" }
        other event types  → pass through unchanged
  router.dispatch(enrichedEvents)  → PROJECTION QUEUE (Redis)
                                ↓
PROJECTION WORKERS  (each pulls a lean event from its queue)
  fold projection      → trace_summaries  (preview as ComputedInput / ComputedOutput)
  map projection       → stored_spans     (lean SpanAttributes — preview + eventref)
  reactors             → eval triggers, broadcast, etc.  (same lean event)
  fold cache (Redis) write-through uses toCacheable to also strip non-IO ephemera
                     (events[], spanCosts, accumulated attributes for pathological many-span traces)

READS
  list / search / detail-collapsed
      ──▶ trace_summaries · stored_spans  (preview, fast, ILIKE on preview column)
  detail-expanded ("show full") · online eval
      ──▶ read langwatch.reserved.eventref.<attrKey> from stored_spans
          CH SELECT on event_log by (TenantId, AggregateType, AggregateId, EventId)
          → parse EventPayload JSON → extract <attrKey> field
          (slower, opt-in; CH point lookup with bloom + sort key = microseconds;
           TenantId in WHERE clause structurally blocks cross-tenant reads)
  replay (operational)
      ──▶ replayEventLoader reads event_log rows (full content)
          replayExecutor.apply: leanForProjection(event) → projection.apply(state, event)
          SAME utility as live → identical projection state

DURABILITY MAP
  event_log (CH)   = single source of truth · FULL content · replay-sufficient
  S3 spool         = transient oversize protection · gone after event_log INSERT succeeds
  projection tables= derived lean views · regenerable from event_log via replay
```

### Files (anchors)

| Path | Role |
|---|---|
| `langwatch/src/server/event-sourcing/services/eventSourcingService.ts` (lines 242-251) | Interposition hook between `storeEvents` and `router.dispatch` |
| `langwatch/src/server/event-sourcing/replay/replayExecutor.ts` | Replay invokes `leanForProjection` before `projection.apply` |
| `langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts` | Add `spoolRef` reconstitution branch; remove (or move) `capOversizedAttributes` call |
| `langwatch/src/server/app-layer/traces/lean-for-projection.ts` (NEW) | The single `leanForProjection` utility — discriminated dispatch on event type |
| `langwatch/src/server/app-layer/traces/blob-store.service.ts` | Backend swap: `put` → S3 spool (transient), `get` → CH SELECT on `event_log` by `EventId` + extract field |
| `langwatch/src/server/app-layer/traces/otlp-span-offload.ts` | Rewritten: edge size-check + S3 spool PUT for oversized; **drop per-attr offload** |
| `langwatch/src/server/app-layer/presets.ts` | Wire interposition into composition root |
| `dev/docs/adr/022-event-log-source-of-truth.md` | Design rationale |
| `specs/event-sourcing/large-trace-blob-offload.feature` | BDD scenarios — drop sha256/S3-fetch scenarios, add replay-parity scenario |

### Order of work (TDD)

1. **Failing tests first** (see Step 4 of the loop directive — these define the contract):
   - `lean-for-projection.unit.test.ts` — discriminated dispatch by event type; IO-attr leaning; preview byte boundary; reserved.eventref attachment; non-IO event passthrough.
   - `interposition.unit.test.ts` — `eventSourcingService` calls `leanForProjection` for every event between `storeEvents` and `router.dispatch`.
   - `recordSpanCommand.oversized.unit.test.ts` — given `{spoolRef}` payload: `BlobStore.get` is called once with the spool key; reconstituted span flows through `handle` normally; on success: best-effort S3 DELETE on the spool.
   - `edge-offload.unit.test.ts` — over-threshold payload: PUT to spool succeeds, command carries `{spoolRef}` only; PUT fails: fail-open inline command + warn log.
   - `blob-store.event-log.unit.test.ts` — `BlobStore.get({eventId, field, tenantId, aggregateType, aggregateId})` issues CH SELECT on `event_log` with the tenant filter, parses `EventPayload` JSON, extracts the named field; cross-tenant returns empty.
   - `replay-projection-parity.integration.test.ts` — ingest the same span via live AND via replay; assert byte-identical projection state in `stored_spans` and `trace_summaries`.
   - Update existing `large-trace-blob-offload.integration.test.ts` to assert the new shape (lean event in projection queue, full content in event_log, S3 spool deleted on success).

2. **Implementation** drives those tests to green via coder agents. Drop dead code:
   - Manifest-shaped S3 objects (`BlobStore` now writes a flat spool body)
   - sha256-per-field on `TraceBlobRef` (CH handles row integrity)
   - Project-prefix auth on `BlobStore.get` (CH `TenantId` WHERE blocks cross-tenant)
   - `BlobIntegrityError` (no longer raised)
   - The resolve-flow JSON manifest fetching (now flat field extraction)
   - `langwatch.reserved.blobref.*` attribute names (replaced by `langwatch.reserved.eventref.*`)
   - Edge per-attribute offload (replaced by whole-span spool decision)

3. `pnpm typecheck` + `pnpm test:unit` + `pnpm test:integration` + `pnpm tsx scripts/check-feature-parity.ts` green.

4. `/prove-it` maps each AC to evidence; `/review` clean.

5. CI green on PR #4216 (re-run e2e on Playwright-install infra hangs).

---

### Track 1 — fold cache leanness

- [ ] **T1.1** Folding a trace with a ≥1 MB output keeps the Redis fold cache entry under 64 KB per IO attr. The cached JSON contains no full `computedInput`/`computedOutput`, no `events[]` payload, and still carries reductions + winner-span pointers + presence flags. Bounded by `leanForProjection` + `toCacheable`.
- [ ] **T1.2** `getTracesWithSpans(traceId)` returns trace.input / trace.output **byte-identical to ingested**, resolved server-side via `event_log` JOIN. No frontend change.
- [ ] **T1.3** Out-of-order refold + `EvaluationTrigger` reactor still produce correct state. Existing trace-processing projection tests stay green.

### Track 2 — event_log as SoT + transient S3 spool + SDK defaults

- [ ] **T2.1** SDKs and gateway transmit full IO without client-side truncation. Python SDK constructor + public `langwatch.trace()` factory both default `max_string_length = 32 * 1024`. TypeScript SDK has no transport-layer cap (verified by code search). Go gateway has no `sdktrace.WithSpanLimits` and no manual truncation in `customertracebridge/emitter.go` (verified by code search). Spans are not annotated with `langwatch.input_truncated` for sub-32 KB payloads.
- [ ] **T2.2** Edge size-check + transient S3 spool. A span whose serialized command payload exceeds 256 KB is written to S3 as a transient spool; the command carries `{spoolRef}` only. The command worker reads `{spoolRef}` and reconstitutes the full span via `BlobStore.get` (or, in the new spool path, the spool-specific GET). On successful `event_log` INSERT, the spool is best-effort DELETEd. If the edge S3 PUT fails: **fail-open** — send regular inline command + warn log.
- [ ] **T2.3** **`event_log` carries the full event** (`EventPayload` is the full span content, ZSTD-compressed). Verified by reading the `event_log` row after a flag-on offloaded trace: parsed `EventPayload` JSON contains the original `langwatch.output` value byte-identical to the input.
- [ ] **T2.4** Online evaluator on an over-threshold trace receives the **full** output via the event_log JOIN read path; captured evaluator input does NOT contain the preview value when the full content is required.
- [ ] **T2.5** Trace-detail collapsed read returns the preview; trace-detail full ("show full") read resolves refs via `event_log` JOIN to return full content. List/search read paths never touch `event_log` (no JOIN on the hot path).
- [ ] **T2.6** Cross-tenant `event_log` read is structurally denied. A request that carries an `eventId` belonging to another tenant returns empty (no rows), because the SELECT WHERE clause includes the requesting `TenantId`.
- [ ] **T2.7** **`leanForProjection` is the single source of truth for the lean shape.** Same function called at `eventSourcingService.ts:242-251` (live) AND in `replayExecutor.apply` (replay) before invoking projection handlers.
- [ ] **T2.8** **Reserved-namespace exclusion** from user-visible enumerations of `stored_spans.SpanAttributes` keys (`buildSpanAttributeKeysFacetQuery` + any other discovered surface). Confirmed via SQL pin in unit test.

### Cross-cutting

- [ ] **X.1** **Replay parity** — replay reads `event_log` rows, applies the SAME `leanForProjection` utility, and produces **byte-identical** `stored_spans` and `trace_summaries` rows as live ingestion of the same span sequence.
- [ ] **X.2** All behind feature flag `release_trace_blob_offload`. Flag off = current behavior (no leaning at interposition; full content flows through projections; existing 256 KB `capOversizedAttributes` is the only cap; no S3 spool, no event_log JOIN for "show full").
- [ ] **X.3** `pnpm typecheck` + `pnpm test:unit` + `pnpm test:integration` + `pnpm tsx scripts/check-feature-parity.ts` green. `/prove-it` maps each AC to evidence; `/review` clean. PR #4216 marked ready, no open review threads, all 55 CI checks green.

## Proof

- Before/after CH row size: a 1 MB output span produces a fold-cache entry ≤ 64 KB + the eventref attr.
- Replay parity: a recorded-and-replayed trace produces byte-identical projection rows (per-row diff = empty).
- `pnpm test:integration src/server/app-layer/traces/__tests__/large-trace-blob-offload.integration.test.ts` exercises the full pipeline wiring with assertions per AC.

---

*Classification: Feature · Supersedes ADR-021 offload mechanism · Behind `release_trace_blob_offload` flag · No frontend change*

drewdrewthis:
## What & why

Closes #4215. Large LLM trace payloads (1) clog Redis — the full OTLP request rides through the command queue as a job payload before the existing 256 KB `capOversizedAttributes` cap can fire — and (2) get truncated client-side (SDK default @5 KB), so evaluations score incomplete text.

Fix, per **[ADR-022](https://github.com/langwatch/langwatch/blob/issue4215/large-trace-blob-offload/dev/docs/adr/022-event-log-source-of-truth.md): `event_log` is the single durable source of truth; S3 is reduced to a transient spool for command-queue oversize protection only.** Full span content lives ZSTD-compressed in `event_log`. A single interposition hook derives the *lean projection shape* (preview + a server-set `langwatch.reserved.eventref.<attrKey>` pointer) so Redis / `stored_spans` / `trace_summaries` / the fold cache stay bounded, while full content remains one ClickHouse SELECT away. With the Python SDK default raised to 32 KB, full IO flows to evaluations again — resolved from `event_log` on the rare "show full" / online-eval read.

> **Supersedes the original ADR-021 "permanent per-field S3 blob offload" design.** Implementation surfaced three facts that redirected it (ADR-022 § Context): a 256 KB cap already existed at the command worker; commands go through a queue, so leaning must happen *before* `.send()`; and a sync `event_log` CH insert (~20–50 ms) is faster than an S3 PUT with a cleaner durability/replay story (event_log alone is replay-sufficient). Search also stays lossless by widening the preview budget rather than sending users to a second backend.

## What's in the PR

Behind feature flag `release_trace_blob_offload` (PRODUCT-scoped, default off — **flag off = today's behavior, byte for byte**).

**Edge — oversize protection only (`edge-spool.ts`):**
- `COMMAND_INLINE_THRESHOLD = 256 KB`. The whole serialized command payload is size-checked; if it crosses, the entire span is spooled to a transient S3 object and the command carries `{spoolRef}` only; otherwise it passes through inline. The command worker reconstitutes from the spool ref; on a successful `event_log` INSERT the spool is best-effort DELETEd. Edge S3 PUT failure ⇒ **fail-open** (inline command + warn). No per-attribute offload at the edge anymore.

**Interposition — single lean hook (`lean-for-projection.ts`, wired in `eventSourcingService`):**
- `leanForProjection(event)` rewrites over-`IO_PREVIEW_BYTES` (64 KB) IO attribute values (`langwatch.input` / `langwatch.output`, `gen_ai.input.messages` / `gen_ai.output.messages`, log-record `body`) to `{preview + langwatch.reserved.eventref.<attrKey>}`, leaves other attributes unchanged, and is a no-op for event types without heavy fields. **Single source of truth for the lean shape** — the same utility runs live (between `storeEvents` and `dispatch`) AND in replay before `projection.apply`, so live and replay produce byte-identical projection state.

**`event_log` = source of truth:**
- The `event_log` row carries the FULL event in `EventPayload` (ZSTD-compressed). Projections (`stored_spans`, `trace_summaries`, fold cache, reactors) only ever see lean shapes. `RedisCachedFoldStore.toCacheable` continues to strip non-IO ephemera as a secondary defence.

**Read paths:**
- List / search / detail-collapsed → projections (preview, `ILIKE` on the widened preview).
- "Show full" / online eval → CH SELECT on `event_log` keyed by `(TenantId, AggregateType, AggregateId, EventId)`, parse `EventPayload`, extract the field. `TenantId` is the first WHERE predicate, so a ref belonging to another tenant returns **no rows** (cross-tenant read structurally denied).
- Replay → `event_log` rows → `leanForProjection` → `projection.apply`.

**`BlobStore`** stays the swap-seam interface: `put` writes the transient S3 spool, `get` reads the `event_log` row and extracts the field, `delete` is the post-`storeEvents` cleanup hook.

**Facets:** reserved-namespace keys (`langwatch.reserved.*`) are excluded from the user-visible `SpanAttributes` key enumeration (`buildSpanAttributeKeysFacetQuery`), so eventref plumbing never surfaces in the Traces facet sidebar.

**SDK truncation defaults:**
- Python SDK: **no client-side truncation** — `max_string_length` now defaults to `None` on both the `LangWatchTrace` constructor and the public `langwatch.trace()` factory, so full content reaches the backend (sizing is backend-driven per ADR-022), consistent with the TS/Go SDKs and stock OpenTelemetry.
- TypeScript SDK: **investigated — no transport-layer cap to raise** (only CLI display helpers truncate).
- Go AI Gateway: **investigated — no cap to raise** (no `sdktrace.WithSpanLimits`, no manual truncation in `customertracebridge/emitter.go`).

## Verification

- **CI: 15/15 workflows green** on `f0c0809ca`. `langwatch-app-ci` = typecheck + lint + all 4 unit shards + all 6 integration shards (the integration shards spin up a real ClickHouse, migrate, and exercise the `event_log` JOIN read path + replay-projection parity).
- **`/prove-it`: ALL PASS** for the full AC set (T1.1–T1.3, T2.1–T2.8, X.1–X.3) — [marker / evidence](https://github.com/langwatch/langwatch/pull/4216#issuecomment-4604033574).
- **`/review`: clean.**
- Unit: 98/98 across the ADR-022 suite (`lean-for-projection`, `blob-store.event-log`, `edge-offload`, `interposition`, `capOversizedAttributes`, `recordSpanCommand.oversized`, and the trace-resolution tests). Integration: replay-projection byte-identical parity + the large-trace-blob-offload pipeline. Typecheck clean.

## Deployment Impact

- **Env vars:** none added or changed. The transient S3 spool reuses the existing S3 configuration; no new variables in `langwatch/.env.example`.
- **Helm values:** none added or changed. No `charts/**` or `services/**` changes in this PR.
- **Default `helm install` (no overrides):** no behavior change. The feature is gated by the PRODUCT-scoped feature flag `release_trace_blob_offload`, which defaults OFF — a vanilla install runs today's path byte-for-byte. No schema migration and no data replay (the `event_log` table already exists per ADR-007).
- **BYOC dataplane:** no new data egress. Full span content stays ZSTD-compressed in the customer's own ClickHouse `event_log`; the S3 spool is transient (best-effort deleted after the `event_log` INSERT) and only engaged for oversize (>256 KB) commands when the flag is on.

## Compat

- Feature-flagged; flag off = today's behavior byte-for-byte; flag on for one project doesn't affect others.
- **No projection schema change.** `event_log` already exists (ADR-007) — no migration and no data replay required to deploy.
- API read-boundary shape unchanged (refs resolved server-side) — no frontend change.

## Summary by CodeRabbit

* **New Features**
  * Added support for offloading large trace field values to object storage, reducing Redis pressure while maintaining transparent read-time restoration.
  * Implemented a new `release_trace_blob_offload` feature flag to control blob offloading behavior.

* **Improvements**
  * Increased Python SDK default string truncation limit from 5 KB to 32 KB for improved trace fidelity.

* **Documentation**
  * Added architectural decision records for lean fold-cache strategy and event-log source-of-truth policy.


0xdeafcafe:
Hey Drew — really like this overall, the ADR is solid and especially happy with read-time recompute (B) over fold-carries-ref. Few things from reading through that I want to flag before it goes on.

**Reserved keys leak into the new Traces UI facet.** `buildSpanAttributeKeysFacetQuery` does an unfiltered `arrayJoin(SpanAttributes.keys)`, so as soon as anything is offloaded the Span Attribute Keys facet starts showing `langwatch.reserved.blobref.langwatch.output` etc as browsable entries. Expand one and `attributeFacetValues` surfaces the raw ref JSON as values — pure plumbing leaking to users. One-liner fix on the discovery side:

```sql
WHERE key != ''
  AND NOT startsWith(key, 'langwatch.reserved.')
```

Probably worth a sweep for anywhere else we enumerate attribute keys (search autocomplete, AI query path) — same problem will hit them. `trace_summaries.Attributes` is fine, refs only live on `stored_spans.SpanAttributes`.

**Free-text search silently truncates at the preview.** `translateFreeText` does `ComputedInput ILIKE … OR ComputedOutput ILIKE …` on `trace_summaries`, and those columns now hold the 2 KB preview because the winning span's IO got replaced at the edge before the fold ran. Anything past byte 2048 silently returns zero rows. Same with `span.attribute.<key>:value`.

Not an indexing thing — CH is indexed, `ILIKE` works fine on whatever's stored. The bytes just aren't in CH anymore, they're in S3 (no query layer) or Redis (which was never the search path anyway). The only lever is what we write into the column at fold time, which is controlled by the preview budget.

**So I think IO attrs should get a 32 KB preview, everything else stays at 2 KB.** The fold cache (Redis) and search (CH) are on completely separate knobs — `toCacheable` strips heavy fields from the Redis entry, but CH is the durable source of truth for search and gets a separate write. Raising the *write-into-CH* preview for IO attrs doesn't reintroduce the Redis clog this ADR is targeted at, they're independent.

```ts
export const IO_PREVIEW_BYTES = DEFAULT_OFFLOAD_THRESHOLD_BYTES; // 32 KB

export const IO_ATTR_KEYS = new Set([
  "langwatch.input",
  "langwatch.output",
  "gen_ai.input.messages",
  "gen_ai.output.messages",
]);
```

`offloadOtlpSpanAttributes` picks `IO_PREVIEW_BYTES` for those keys, the 2 KB default for everything else (contexts, custom metadata, span attrs nobody `ILIKE`s). Anything under 32 KB passes through unchanged, fully searchable. Anything over 32 KB is searchable up to the threshold — and past that, why would you ever type a 35 KB term into the search bar? Redis stays lean via `toCacheable` regardless.

Worth dropping a note in ADR-021's Consequences so the reasoning survives:

> IO attributes use a 32 KB preview (= threshold) so full-text search on `trace_summaries.ComputedInput/Output` is lossless for sub-threshold values and covers the first 32 KB of offloaded ones. Fold cache leanness is on a separate knob (`toCacheable`), so raising the IO preview doesn't reintroduce the Redis clog.

**One object per field is going to be expensive.** Current key shape is `trace-blobs/{projectId}/{traceId}/{spanId}/{attrKey}` — one S3 object per offloaded field. Typical LLM trace has 3–5 over-threshold fields, so we're doing 3–5× PUTs and GETs:

- Writes are inline in the OTLP ingest path, so each PUT directly hits collector response time. 10M spans/mo × ~3 fields ≈ 30M PUTs (~$150/mo on requests alone, and PUT cost is mostly fixed-per-request so smaller objects are anti-economical).
- Reads do N parallel GETs per span — a 100-span trace eval with 3 fields/span is 300 concurrent S3 connections.

What if we batched into one object per span, manifest-shaped:

```
Key:  trace-blobs/{projectId}/{traceId}/{spanId}
Body: { "version": 1, "encoding": "utf-8",
        "fields": { "langwatch.input": "...", "langwatch.output": "...", ... } }
```

Ref gets a `field` selector into the shared object — per-field sha256 still works for integrity, strip-reserved-keys path is unchanged, resolver just groups refs by `key` before fetching:

```
langwatch.reserved.blobref.langwatch.output =
  {"key":"trace-blobs/p/t/s","field":"langwatch.output","sha256":"…","size":51200}
```

Why per-span and not per-trace: per-trace looks tempting for read coalescing but spans for one trace arrive across multiple ingest requests / pods over time (especially long agent traces), so per-trace needs either S3 `If-Match` contention, a per-trace lock that serialises hot traces, or LWW data loss. None of those are acceptable. Per-trace is fine for finalised batch payloads, not streaming OTLP ingest. Per-span dodges all of it because OTLP spans are atomic — one ingest = one PUT, overwrite is idempotent.

Tradeoff: a reader that only needs `langwatch.output` still pulls input too. Real bandwidth, but over-threshold means already-large fields anyway, most eval/detail reads want both, and if it ever bites we can do msgpack + offset table + Range reads later. Not worth solving now.

Code change is localised — `BlobStore.put` takes `{ fields: Record<attrKey, value> }` and returns `Record<attrKey, TraceBlobRef>`, `BlobStore.get` returns the manifest, `offloadOtlpSpanAttributes` does one `put` per span with all over-threshold fields, `SpanBlobResolutionService.resolve` groups by `key` before fetching (public signature unchanged). Flag is off in prod so no migration story — clean cut to per-span before we turn it on.

drewdrewthis:
Thanks Drew — all three landed.

**1. Reserved keys leak into Span Attribute Keys facet** — `50312f4e0`. `buildSpanAttributeKeysFacetQuery` now filters `AND NOT startsWith(key, 'langwatch.reserved.')` on the outer WHERE. Swept other user-visible enumerations on `stored_spans.SpanAttributes` for the same filter (`trace_summaries.Attributes` untouched — refs only live on `stored_spans`). Test pins the SQL: `expect(query.sql).toContain("NOT startsWith(key, 'langwatch.reserved.')")`.

**2. Differential preview byte budget — 32 KB for IO attrs, 2 KB default** — `3fc71ae5a`. Exactly your shape:

```ts
export const IO_PREVIEW_BYTES = DEFAULT_OFFLOAD_THRESHOLD_BYTES; // 32 KB
export const IO_ATTR_KEYS = new Set([
  "langwatch.input",
  "langwatch.output",
  "gen_ai.input.messages",
  "gen_ai.output.messages",
]);
```

`offloadOtlpSpanAttributes` picks `IO_PREVIEW_BYTES` for IO attrs and `previewBytes ?? DEFAULT_PREVIEW_BYTES` for everything else. Tests pin both paths (100 KB `langwatch.output` → preview ≤ `IO_PREVIEW_BYTES+4`; 100 KB `custom.metadata` → preview ≤ `DEFAULT_PREVIEW_BYTES+4`). Confirmed in the code that `toCacheable` strips heavy fields from the Redis entry independently — no Redis-clog regression.

**3. One object per span (manifest-shaped)** — `2f5225a22`. Key shape collapsed to `trace-blobs/{projectId}/{traceId}/{spanId}` with a JSON manifest body `{ version, encoding, fields }`. `TraceBlobRef` now carries `field` + per-field `sha256`. `BlobStore.put({ projectId, traceId, spanId, fields })` returns `Record<attrKey, TraceBlobRef>` — **one** PutObjectCommand per span regardless of field count. `BlobStore.get` reads the manifest, looks up `ref.field`, verifies its sha256 (`BlobIntegrityError` on mismatch; new `BlobFieldNotFoundError` for missing field).

Resolver groups by `ref.key` via an in-flight `Map<key, manifest>` passed through the resolve loop — so N refs from one span → **one** S3 GET. Test pins it: `expect(caches[0]).toBe(caches[1])` (same Map instance across both `get` calls).

Per-span vs per-trace — same reasoning you laid out: OTLP spans are atomic, no cross-pod lock needed; per-trace would need either contention or LWW.

ADR-021 (`8aed759f6`) updated Consequences with the three new properties + Rules sections covering the reserved-namespace edge strip and the user-visible enumeration filter.

### Verification

- `pnpm typecheck` — clean
- `pnpm test:unit src/server/traces src/server/app-layer/traces ...` — **1034/1034** (was 1019, +15 new tests for the manifest changes)
- `pnpm test:integration src/server/app-layer/traces/__tests__/large-trace-blob-offload.integration.test.ts` — **15/15**
- `pnpm tsx scripts/check-feature-parity.ts` — **OK: 1363 enforced scenario(s) bound across 467 file(s)** (exit 0)

Full CI re-running on the head; will report green.

### On your `Question, no action` about hydrate-later vs CH-table

Short version: yes, S3 is unsearchable — the differential preview makes search lossless ≤32 KB but partial above. For LLM agent traces specifically, tail content past 32 KB on long outputs is a real findability gap.

If/when we want it gone, the cleanest swap is to put a **CH-table impl behind the `BlobStore` interface** — `put`/`get` are the seam, the pipeline doesn't care about the backend. That ships as a follow-up without re-litigating any pipeline design here. Hydrate-later was the less attractive of the two (2× writes, "done" detection is hard for long agent traces, defeats offload's secondary lean-CH benefit). Happy to spin up an issue for the CH-table follow-up if you want.

rogeriochaves:
1 P1 blocker.

- `langwatch/src/server/app-layer/traces/lean-for-projection.ts:65-67`, `langwatch/src/server/app-layer/traces/edge-spool.ts:37-49`, and `langwatch/src/server/app-layer/traces/blob-store.service.ts:219-235` still throw `not implemented`, but the branch adds tests and integration scaffolding that call these helpers directly. As written, the new test suite will fail and any runtime path that hits the new projection/spool/event_log flow will crash immediately. Please either land the real implementations in this PR or gate/remove the new callers and tests until the helpers are wired for real.

drewdrewthis:
**Status: course-correcting to ADR-022.** Putting the PR back into draft while implementation lands.

A design conversation surfaced two facts that ADR-021's "edge offload to permanent S3" mechanism didn't account for:

1. **`capOversizedAttributes(256 KB)` already exists** at `recordSpanCommand.ts:146`. Everything downstream of the command worker is already bounded today. The actual unbounded-payload problem is narrower — it's the **edge → command-queue leg**, where the full OTLP request rides through Redis BEFORE the cap fires.
2. **Commands go through a queue.** `commands.traces.recordSpan(data).send()` stages the full command payload in Redis (the custom `GroupQueueProcessor`). The event isn't in `event_log` when `.send()` returns. To lean the Redis pressure point, leaning has to happen BEFORE `.send()`.

### New design — ADR-022 (supersedes ADR-021's offload mechanism)

[Full ADR](https://github.com/langwatch/langwatch/blob/issue4215/large-trace-blob-offload/dev/docs/adr/022-event-log-source-of-truth.md). Key shifts:

- **`event_log` is the single durable source of truth.** Full content lives in `EventPayload` (ZSTD-compressed). Replay-sufficient on its own.
- **S3 is reduced to a transient spool** — only used for commands > 256 KB, eagerly DELETEd after `event_log` INSERT succeeds. No S3 in any user-facing read path.
- **One `leanForProjection(event)` utility** runs at TWO call sites: the dispatch interposition (live, `eventSourcingService.ts:242-251`) and `replayExecutor.apply` (replay). Same code → projection state is **path-independent** (live = replay, byte-for-byte).
- **`IO_PREVIEW_BYTES = 64 KB`** — sized to a **standard Claude response** (~16K tokens). `ILIKE` on `trace_summaries.ComputedInput/Output` stays lossless for the modal case.
- **"Show full" / online eval** JOINs `event_log` directly. Microseconds at p50 via the existing bloom + sort-key on `EventId`. `TenantId` in the WHERE clause structurally blocks cross-tenant reads.

### What survives from ADR-021

Reserved-namespace strip (already at command worker), facet-enumeration filter, differential preview budget, `toCacheable` for non-IO ephemera, `BlobStore` as the swap-seam interface.

### What this PR will end up shipping (TDD plan)

1. ✅ ADR-022 written
2. ✅ Issue body + ACs updated
3. ✅ Spec scenarios rewritten (feature-parity clean)
4. ✅ Failing test contract landed (8 commits, ~30 `it(...)` blocks — they fail at unit runtime as expected)
5. ⏳ Implementation drives those to green
6. ⏸ `/prove-it` + `/review`
7. ⏸ CI green + flip out of draft

Reverting to draft now. Will flip back to ready once Step 5 lands and CI is green.

langwatch-agent:
1 P1

- `langwatch/src/server/traces/resolve-offloaded-traces.ts:155-167` queries `event_log` with `eventId = span.traceId`, but `event_log.EventId` is the generated event id from `createEvent()`, while the trace pipeline uses `aggregateId = traceId`. The lean event only stores `{ field }`, so this lookup will miss the durable row for normal spans and the offloaded payload never resolves. Minimal fix: carry the real `event.id` through the eventref/span metadata and use that in `getFromEventLog()`.

langwatch-agent:
No new blockers.

The new head is just a merge of upstream `main`; it doesn’t change the offload path I reviewed earlier, and I didn’t find a fresh regression in the new commits. Residual risk is unchanged from the prior review: the offload/read-resolution code still depends on the existing keying and resolution assumptions, but nothing new landed here that worsens that.

langwatch-agent:
No blockers.

I reviewed the new CI-only head commit and did not find a regression in the Playwright cache/install flow. The cache key is pinned to the installed `@playwright/test` version, and the fallback still installs Chromium with OS deps when the cache is cold.

drewdrewthis:
Caught it — thanks. Fix landed in `4dd76f05d`.

Root cause was as you described: `leanForProjection` wrote `langwatch.reserved.eventref.<attrKey> = { field }` only, and the read path fell back to `span.traceId` as the EventId proxy (the TODO at `resolve-offloaded-traces.ts:165` even acknowledged it). Wrong column → empty JOIN → preview never resolves.

Fix:
1. `leanForProjection` now embeds `event.id` in the eventref: `{ field, eventId: event.id }`. The event has its `.id` at lean time (just produced by `createEvent` before `storeEvents` fires), so no extra plumbing.
2. `resolveOffloadedTraces` extracts `ref.eventId` from the eventref JSON and passes it to `BlobStore.getFromEventLog`. If the eventref is missing/empty `eventId` (malformed), we log warn and keep the preview — never silently fall back to a wrong key.
3. Test fixtures updated to include `eventId` in the eventref JSON.

Verified: `pnpm typecheck` clean, `pnpm test:unit` 1259/1259, `pnpm test:integration` 20/20. CI re-firing now.

drewdrewthis:
What about on-prem that don't have s3? what happens? risks? complications? 
What about images/other files?

drewdrewthis:
Answering your question (on-prem without S3, and images/other files) — both are handled; here are the specifics and the honest caveats.

**On-prem with no object storage — supported; reads need zero S3.**
- The read path (`BlobStore.getFromEventLog`) only queries ClickHouse. Full content lives in `event_log` (the ADR-022 source of truth), never S3 — so "show full" / online-eval work with no object storage at all. Pinned by `blob-store.event-log.unit` (a resolver that throws if S3 is touched; the read still succeeds).
- The edge spool is flag-gated (`release_trace_blob_offload`) **and** fail-open. Flag OFF (the on-prem default): `maybeSpool` is never called — commands carry the full payload and `capOversizedAttributes` caps >256 KB attribute values at the worker (the pre-ADR behaviour); no S3 anywhere. Flag ON without storage: `putSpool` throws → fail-open to full inline payload (warn logged); ingestion never blocks.
- "No S3" = no object storage at all. S3-compatible stores (MinIO, Ceph RGW) work via the standard client.
- **Risk / complication:** on-prem with the flag ON but no storage loses only the *edge* oversize-protection — >256 KB commands ride the Redis command queue as full payloads (transient; the worker's `capOversizedAttributes` still keeps the fold/projection state bounded). Recommendation: on-prem leaves the flag OFF unless it has S3-compatible storage. Documented in ADR-022 § "On-prem / no-object-storage deployments".

**Images / other large files — bounded on every surface; full content retained.**
- Large base64 on the IO message attributes (`gen_ai.input/output.messages`, `langwatch.input/output`) → 64 KB preview + a `langwatch.reserved.eventref` pointer; the full image is recoverable from `event_log` on "show full".
- Large values on NON-IO / nested / span-event / span-link / resource attributes are bounded too — `leanForProjection` now runs the recursive cap over every attribute surface `capOversizedAttributes` walks (closed a coverage hole found in review). Capped to a placeholder in the projection; full content preserved in `event_log`. Pinned by `lean-for-projection.unit` + `capOversizedAttributes.unit`.
- **Honest caveats:** (1) a 64 KB "preview" of base64 is truncated base64 — a size guard, not a usable thumbnail; the full image comes from `event_log`. (2) non-IO oversized values are recoverable via a full `event_log` read, not a field-level eventref (eventrefs are IO-key-only by design). (3) multi-MB media lives permanently in `event_log` (ZSTD; base64 of already-compressed media won't shrink much) — a deliberate ADR-022 trade-off (single source of truth).

drewdrewthis:
**CI status (for the reviewer/merger):** the only non-green checks are `test-unit (3)` and `test-integration (3)`. These wedge ~100–166 min on the *test-run* step **repo-wide — `main` itself hangs the same shard** (runs 26650910676, 26642310246), so it is **not a defect in this PR**. Tracked as #4437 (tiktoken remote-fetch fallback is the leading suspect). The other 54 checks are green, and this PR's own unit tests pass locally with no hang/handle leak.

Note: branch is ~21 commits behind `main`; a rebase before merge is advisable once shard 3 is healthy.

drewdrewthis:
**Update — the shard-3 CI wedge is fixed and proven.** Root cause: `TiktokenClient.remoteFetch` did an unbounded network fetch (no timeout) that hangs in network-restricted CI shards — repo-wide (it wedged `main` too), not this PR.

Fix shipped as #4439 (bound the fetch with an abortable timeout + offline-deterministic test). #4439 is now **fully green — 54 checks pass, 0 fail — and shard 3 completed in ~4 min** vs the prior 100+ min wedge, confirming the fix.

To land this PR: merge #4439, then rebase this branch onto `main` (also clears its ~21-commit staleness); its CI should then go green.

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,235 @@
...
+export class BlobStore {
+  /**
+   * @param resolveS3Client - Resolver for per-org S3 client + bucket.
+   * @param clickHouseClient - Optional ClickHouseClient for ADR-022 event_log reads.
+   *   When provided, `getFromEventLog` uses it to SELECT from event_log.
+   *   When absent, `getFromEventLog` throws "ClickHouseClient not configured".
+   */
+  constructor(
+    private readonly resolveS3Client: S3ClientResolver,
+    private readonly clickHouseClient?: ClickHouseClient,
```
0xdeafcafe: this will fail for tenants outside the primary clickhouse cluster. i think this should really be a `resolveClickHouseClient`

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,235 @@
...
+    const rows = response.data;
+
+    if (!rows || rows.length === 0) {
+      throw new BlobNotFoundError(eventId, field, tenantId);
+    }
+
+    const row = rows[0]!;
+    let parsed: unknown;
+    try {
+      parsed = JSON.parse(row.EventPayload);
```
0xdeafcafe: could use zod schemas here for extra safety

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    if (!Array.isArray(spanAttributes)) {
+      throw new BlobFieldNotFoundError(eventId, field);
+    }
+
+    const attr = spanAttributes.find(
+      (a: { key: string }) => a.key === field,
+    );
+
+    if (!attr || typeof (attr as { value?: { stringValue?: string } }).value?.stringValue !== "string") {
```
0xdeafcafe: zod schema would clean up a lot of this type-foo

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,235 @@
...
+        Body: body,
+        ContentType: "application/octet-stream",
+      }),
+    );
+    return key;
+  }
+
+  /**
+   * Best-effort deletion of a transient S3 spool object.
+   * Called after event_log INSERT succeeds. Errors are swallowed — the 24h lifecycle
```
0xdeafcafe: i think lifecycle should actually be 3 days, as if there is an issue on a weekend, we may want a bit of time to catch up

langwatch/src/server/app-layer/traces/edge-spool.ts
```diff
@@ -0,0 +1,76 @@
...
+  logger: SpoolLogger;
+}): Promise<RecordSpanCommandData> {
+  const serialized = JSON.stringify(data);
+  const byteLength = Buffer.byteLength(serialized, "utf-8");
+
+  if (byteLength <= COMMAND_INLINE_THRESHOLD) {
+    // Payload fits inline — no S3 PUT needed
+    return data;
+  }
+
```
0xdeafcafe, line 46: can we avoid this additional stringify? i don't think so, but would be nice if we can

langwatch/src/server/app-layer/traces/edge-spool.ts
```diff
@@ -0,0 +1,76 @@
...
+  // Payload exceeds threshold — attempt to spool to S3
+  const { tenantId: projectId, span } = data;
+  const traceId = span.traceId as string;
+  const spanId = span.spanId as string;
+  const spoolBody = Buffer.from(
+    JSON.stringify({ span, resource: data.resource, instrumentationScope: data.instrumentationScope }),
```
0xdeafcafe: we can just use the `serialized` from above, if we only check the size of the payload itself, which is nice too

langwatch/src/server/app-layer/traces/lean-for-projection.ts
```diff
@@ -0,0 +1,228 @@
...
+export const EVENTREF_ATTR_PREFIX = "langwatch.reserved.eventref.";
+
+/** UTF-8-safe truncation to at most `maxBytes`, backing off to a codepoint boundary. */
+function utf8Preview(value: string, maxBytes: number): string {
+  const buf = Buffer.from(value, "utf-8");
+  if (buf.byteLength <= maxBytes) return value;
+  let end = maxBytes;
+  // 0b10xxxxxx are UTF-8 continuation bytes — don't cut mid-codepoint.
+  while (end > 0 && (buf[end]! & 0xc0) === 0x80) end--;
+  return buf.subarray(0, end).toString("utf-8") + "…";
```
0xdeafcafe: can we use the string trim function here? does it not respect, or have a way to respect, utf-8 code points?

python-sdk/src/langwatch/telemetry/tracing.py
```diff
@@ -778,7 +783,12 @@ def trace(
     expected_output: Optional[str] = None,
     api_key: Optional[str] = None,
     disable_sending: bool = False,
-    max_string_length: Optional[int] = 5000,
+    # Raised from 5000 to 32 KB to match the LangWatchTrace constructor — see
+    # tracing.py:91 and issue #4215. The default must be the same on both the
+    # public `trace()` factory and the underlying constructor; otherwise users
+    # calling `langwatch.trace(...)` without an explicit kwarg would silently
+    # get 5000 KB while users calling `LangWatchTrace(...)` get 32 KB.
+    max_string_length: Optional[int] = 32 * 1024,
```
0xdeafcafe: we could just ditch this and have all this be backend driven?

langwatch/src/server/traces/trace.service.ts
```diff
@@ -96,9 +119,40 @@ export class TraceService {
...
+            projectId: string,
+            normalizedSpans: NormalizedSpan[],
+          ) =>
+            resolveOffloadedTraces({
+              projectId,
+              normalizedSpans,
+              blobStore: blobResolutionDeps.blobStore,
+              ioExtractionService: blobResolutionDeps.ioExtractionService,
+              logger: this.logger,
+            })
```
0xdeafcafe: this is a little ugly, can it not be in a class or something?

langwatch/src/server/traces/resolve-offloaded-traces.ts
```diff
@@ -0,0 +1,268 @@
...
+        // reserved keys anyway so the UI never sees the namespace.
+        return { ...span, spanAttributes: cleanedAttrs };
+      }
+
+      // ADR-022: aggregateId for the trace-processing pipeline IS the traceId.
+      // The eventref carries the eventId, written by leanForProjection from
+      // event.id at lean time — see lean-for-projection.ts:120.
+      const aggregateId = span.traceId;
+
+      try {
```
0xdeafcafe: this logic is super slow, let's promise.all it

0xdeafcafe: ah i think this is promise all'd above

langwatch/src/server/traces/resolve-offloaded-traces.ts
```diff
@@ -0,0 +1,268 @@
...
+      recomputedInput: null,
+      recomputedOutput: null,
+      anyResolved: false,
+    };
+  }
+
+  // Resolve each span individually so a failure on one span does not block
+  // the others. Track whether at least one span resolved successfully.
+  let anyResolved = false;
+  const resolvedSpans: NormalizedSpan[] = await Promise.all(
```
0xdeafcafe: we should `allSettled` then we can still show the ones that succeeded i think, rather than discarding any unresolved on error.

langwatch/src/server/traces/resolve-offloaded-traces.ts
```diff
@@ -0,0 +1,268 @@
...
+ * Result of resolving offloaded blobs for a single trace's spans.
+ */
+export interface ResolvedTraceSpans {
+  /** Spans with full attribute values restored (refs stripped). */
+  resolvedSpans: NormalizedSpan[];
+  /**
+   * Recomputed trace-level input from the resolved spans, or null when no
+   * event refs were present / resolution failed.
+   */
+  recomputedInput: import("~/server/app-layer/traces/trace-io-extraction.service").ExtractedIO | null;
```
0xdeafcafe: inline type import

0xdeafcafe (commented):
i think this is missing the trace services responsible for traces v2 ui.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in ba86c5423 — BlobStore now takes resolveClickHouseClient and getFromEventLog resolves the per-tenant client, so reads work for tenants outside the primary cluster. presets.ts passes it through. Reopen if not what you meant.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in ba86c5423 — added zod schemas (eventLogRowSchema + eventPayloadSchema) and safeParse the row + payload, dropping the hand-rolled casts.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in ba86c5423 — the same zod schemas clean up the type-foo here.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in ba86c5423 — spool lifecycle is now 3 days (weekend catch-up), JSDoc updated.

langwatch/src/server/app-layer/traces/edge-spool.ts
drewdrewthis: Done in ba86c5423 — we serialize once and reuse it (see the :52 thread); the second stringify is gone.

langwatch/src/server/app-layer/traces/edge-spool.ts
drewdrewthis: Done in ba86c5423 — the size check and the spool body both use the single serialized string now.

langwatch/src/server/app-layer/traces/lean-for-projection.ts
drewdrewthis: It is already UTF-8-safe: utf8Preview backs off any trailing UTF-8 continuation byte before cutting, so it never splits a multibyte code point, while still respecting the byte budget. No code change. Reopen if you want a Intl.Segmenter / grapheme approach instead.

python-sdk/src/langwatch/telemetry/tracing.py
drewdrewthis: max_string_length truncates client-side BEFORE the span is transmitted, so the backend cannot recover what the SDK already dropped — that is why the value has to live in the SDK (raised 5000 to 32KB so full content reaches event_log). We could instead send unlimited from the SDK and let the edge-spool size everything (fully backend-driven), but a client-side cap protects the SDK-to-backend transport from pathological payloads, and 32KB matches the LangWatchTrace constructor default while covering a typical completion. Leaving this thread open — happy to go unlimited + backend-spool if you prefer; your call.

langwatch/src/server/traces/trace.service.ts
drewdrewthis: Done in ba86c5423 — extracted into a named OffloadedSpanResolver class.

langwatch/src/server/traces/resolve-offloaded-traces.ts
drewdrewthis: Done in ba86c5423 — the per-field resolution loop is Promise.allSettled (parallel) now.

langwatch/src/server/traces/resolve-offloaded-traces.ts
drewdrewthis: Done in ba86c5423 — the outer span loop is Promise.allSettled, so successfully-resolved spans are returned even when one ref fails to resolve.

langwatch/src/server/traces/resolve-offloaded-traces.ts
drewdrewthis: Done in ba86c5423 — converted to a top-level import type.

drewdrewthis:
@0xdeafcafe re: "missing the trace services responsible for traces v2 ui" — good catch, you were right. The v2 read path (`tracesV2` router → `SpanStorageService`/`SpanStorageClickHouseRepository`) was reading `stored_spans` directly and returning the lean preview + raw `langwatch.reserved.eventref.*` pointers, with **no** resolution — so offloaded large traces showed truncated in the v2 UI. Fixed in ba86c5423: wired offloaded-blob resolution into `SpanStorageService.getSpansByTraceId` / `getSpanById` (so `spansFull` / `spanDetail` resolve `event_log` refs to full content); `presets.ts` now injects `blobStore` + `ioExtractionService` into the span-storage service; +8 unit tests (resolve / eventref-strip / partial-failure degradation). All 12 inline comments addressed (11 resolved; left the SDK `max_string_length` thread open — replied there with the tradeoff for your call).

drewdrewthis:
Heads-up @0xdeafcafe — while working your review I traced the `event_log` read path end-to-end and found a **critical pre-existing bug** (fixed in 324e0bedd):

`BlobStore.getFromEventLog` read `EventPayload.data.span.attributes` — one `data` level too deep. The stored `EventPayload` **is** the event's `data` field (`eventToRecord`: `EventPayload = event.data`), and a SPAN_RECEIVED event's data is `{ span, resource, instrumentationScope }` (`recordSpanCommand`; `leanForProjection` reads `event.data.span`). So the correct path is `EventPayload.span.attributes`. The extra `.data` made **every** show-full / online-eval / traces-v2 read return `BlobFieldNotFoundError` and silently fall back to the preview — i.e. the ADR-022 read path never actually returned full content.

It hid behind unit fixtures shaped to the buggy read, the flag being off, and no real-CH round-trip test. Fix: `eventPayloadSchema` + extraction now read `EventPayload.span.attributes`, and the 4 `blob-store.event-log` fixtures are corrected to the real write shape so they regression-guard read-vs-write alignment. 32 unit tests pass, typecheck clean.

Recommend a follow-up **real-CH integration test** for the `getFromEventLog` round-trip as defense-in-depth (the unit fixtures now match reality, but a real-CH test is what would have caught this).

python-sdk/src/langwatch/telemetry/tracing.py
0xdeafcafe: i think it would be nice to be consistent. only the python sdk does this, and no otel impl will do it out of the box.

drewdrewthis: Agreed — done in fb243de2c. Both `trace()` and the `LangWatchTrace` constructor now default `max_string_length=None`, so the Python SDK no longer truncates span content client-side, matching the TS/Go SDKs and stock OTel. Fidelity is now fully backend-driven (ADR-022 event_log + edge spool); added a regression test pinning the no-truncation default. (The 32 KB was lossy before transmit and only relocated #4215's cliff.)

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,267 @@
...
+ * EventPayload IS event.data (stored as `event.data ?? {}` by eventToRecord).
+ * The real write shape from recordSpanCommand is `{ span, resource, instrumentationScope }`.
+ * Span is at the TOP level — there is NO outer `data` wrapper.
+ */
+const eventPayloadSchema = z.object({
+  span: z
+    .object({
+      attributes: z.array(spanAttributeSchema),
+    })
+    .optional(),
```
0xdeafcafe, line 76: this only covers `span.attributes`, but `leanLogRecordReceivedEvent` puts the eventref on log-record `data.body` (key in `data.attributes`). nothing reads log bodies via `getFromEventLog` today so we're not breaking anything, but `eventPayloadSchema` is named like it covers all event payloads. either widen it to fall back to `data.body` when `field === "body"`, or rename + drop a comment that log bodies are intentionally unresolved so the next person doesn't trip on it.

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,267 @@
...
+   * (`trace-blobs/spool/{projectId}/{traceId}/{spanId}`), the projectId is at
+   * index 2 of the key split by "/".
+   *
+   * @param spoolRef - The spool reference string (S3 key) returned by `putSpool`.
+   * @returns The raw body buffer as stored by `putSpool`.
+   * @throws The underlying S3 error if the object does not exist or access fails.
+   */
+  async getSpool(spoolRef: string): Promise<Buffer> {
+    // Extract projectId from the spool key: trace-blobs/spool/{projectId}/...
+    const projectId = spoolRef.split("/")[2] ?? "";
```
0xdeafcafe: this is super fragile — `putSpool` encodes the key shape and `getSpool`/`deleteSpool` decode it by positional index. a future rename of the key prefix silently breaks both reads. can we either (a) wrap put + parse in a single helper, or (b) make `spoolRef` a `{ key, projectId }` object instead of a flat string?

langwatch/src/server/app-layer/traces/blob-store.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    const { s3Client, s3Bucket } = await this.resolveS3Client(projectId);
+    const { Body } = await s3Client.send(
+      new GetObjectCommand({ Bucket: s3Bucket, Key: spoolRef }),
+    );
+    const bytes = await Body?.transformToByteArray();
```
0xdeafcafe, line 252: if s3 ever returns a 200 with no body we silently feed an empty buffer into `JSON.parse` downstream and get an opaque parse error. cheap to throw explicitly here on `bytes == null` so the failure mode is legible.

langwatch/src/server/app-layer/presets.ts
```diff
@@ -507,6 +526,39 @@ export function initializeDefaultApp(options?: { processRole?: ProcessRole }): A
     new TraceRequestCollectionService({
       dedup: spanDedup,
       recordSpan: commands.traces.recordSpan,
+      // ADR-022: Edge size-check + transient S3 spool, flag-gated per project.
+      // projectId === tenantId (routes/otel.ts passes project.id). Flag checked
+      // per span via the 5s-cached flag store, so effectively once per request.
```
0xdeafcafe: this comment is wrong — `processCommandData` is called per span, not per request. the 5s cache makes the cost fine, but the wording suggests we only hit the flag store once per OTLP batch and we don't.

langwatch/src/server/app-layer/presets.ts
```diff
@@ -507,6 +526,39 @@ export function initializeDefaultApp(options?: { processRole?: ProcessRole }): A
...
+            { projectId: data.tenantId },
+          );
+          if (enabled !== true) return data;
+          return await maybeSpool({
+            data,
+            blobStore,
+            logger: createLogger("langwatch:traces:edge-spool"),
+          });
+        } catch (err) {
+          createLogger("langwatch:traces:edge-spool-fail-open").warn(
```
0xdeafcafe, line 553: should we bump a counter here? a `langwatch_edge_spool_fail_open_total{reason}` would let us alert on s3 outages without grepping warn logs.

python-sdk/src/langwatch/telemetry/tracing.py
```diff
@@ -88,7 +88,13 @@ def __init__(
...
         api_key: Optional[str] = None,
         disable_sending: bool = False,
-        max_string_length: Optional[int] = 5000,
+        # No client-side truncation by default. Full input/output reaches the
+        # backend, which owns sizing (ADR-022: event_log is the source of truth,
+        # with the edge spool + server-side caps handling oversize). This keeps
+        # the Python SDK consistent with the TS/Go SDKs and stock OpenTelemetry —
+        # none of which truncate span content client-side (#4215). Set to an int
+        # (>= 100) to opt into a client-side per-field byte cap.
+        max_string_length: Optional[int] = None,
```
0xdeafcafe, line 97: this is a behaviour change for everyone on the new SDK, not just flag-on tenants. with the flag off, full payloads now go through redis with no edge spool. is that fine? worth at least a note in the SDK changelog so users upgrading aren't surprised by larger network payloads.

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
```diff
@@ -143,19 +212,26 @@ export class RecordSpanCommand implements CommandHandler<
...
-        if (cappedAttributeCount > 0) {
-          this.logger.warn(
-            { tenantId, traceId, spanId, cappedAttributeCount },
-            "Capped oversized span attribute value(s) before ingestion",
+        //
+        // ADR-022: Skip the cap for spool-reconstituted spans. When a command
+        // carries a spoolRef, the full content has already bypassed the Redis
+        // pressure point and MUST be written to event_log in its entirety.
+        // The cap's purpose is Redis safety, not event_log correctness.
+        if (!commandData.spoolRef) {
```
0xdeafcafe, line 247: this is only safe because `leanForProjection` runs unconditionally in `eventSourcingService.ts` after this. worth pinning that link in the comment — someone making the lean step conditional in future would silently re-introduce the redis clog for spool-reconstituted spans.

langwatch/src/server/app-layer/traces/lean-for-projection.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Server-internal namespace prefix used by `leanForProjection` to attach eventref pointers
+ * (carrying `{ field }`) alongside lean attribute previews. Client-supplied attributes in
+ * the `langwatch.reserved.*` namespace are stripped at command-worker ingestion.
+ */
+export const EVENTREF_ATTR_PREFIX = "langwatch.reserved.eventref.";
+
+/** UTF-8-safe truncation to at most `maxBytes`, backing off to a codepoint boundary. */
+function utf8Preview(value: string, maxBytes: number): string {
+  const buf = Buffer.from(value, "utf-8");
```
0xdeafcafe: minor: node canonical is `"utf8"` (no hyphen) — both work, just inconsistent with the rest of the codebase. shows up in a few places in this file + `edge-spool.ts`.

0xdeafcafe (commented):
nice rework. a handful of inline notes — mostly nits, two i'd actually want addressed (the log-body schema gap and the per-request comment).

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in e92ed0994 (your option a) — getFromEventLog now resolves log-record bodies: when field==='body' it reads the full body from the top-level EventPayload.body instead of span.attributes, and eventPayloadSchema carries an optional top-level body. Added unit tests for the resolve + the body-missing case.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in e92ed0994 — centralized the key shape: buildSpoolKey() is the only encoder and projectIdFromSpoolKey() decodes off SPOOL_PREFIX_SEGMENTS (derived from the prefix), so a prefix rename can no longer silently desync put vs get/delete. putSpool/getSpool/deleteSpool all route through the helpers.

langwatch/src/server/app-layer/traces/blob-store.service.ts
drewdrewthis: Done in e92ed0994 — getSpool now throws an explicit 'no body' error when S3 returns a response with no Body, rather than feeding an empty buffer to a downstream JSON.parse. Added a unit test.

langwatch/src/server/app-layer/presets.ts
drewdrewthis: Done in e92ed0994 — corrected the comment: processCommandData runs per span (not once per OTLP request/batch); the 5s-cached flag store is what keeps the per-span read cheap.

langwatch/src/server/app-layer/presets.ts
drewdrewthis: Done in e92ed0994 — added langwatch_edge_spool_fail_open_total{reason} in metrics.ts and increment it in the fail-open catch with reason='flag_store' or 'spool' (also added reason to the warn log), so an S3 outage is alertable without grepping logs.

python-sdk/src/langwatch/telemetry/tracing.py
drewdrewthis: Done in e92ed0994 — added an upgrade-note Warning to the SDK reference (docs/integration/python/reference.mdx) calling out the behavior change (full payloads now transmitted → larger network payloads for flag-off projects). Note: CHANGELOG.md is release-please-generated, so the changelog entry flows from the conventional commit rather than a hand edit. Happy to add a literal CHANGELOG line if you'd rather have it there too.

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
drewdrewthis: Done in e92ed0994 — pinned it in the comment: the cap-skip is only safe because leanForProjection runs unconditionally in eventSourcingService.ts, and made the future-hazard explicit (making the lean step conditional would re-introduce the Redis clog for spool-reconstituted spans).

langwatch/src/server/app-layer/traces/lean-for-projection.ts
drewdrewthis: Done in e92ed0994 — switched the 'utf-8' encoding literals to 'utf8' (node canonical) in lean-for-projection.ts and edge-spool.ts.

0xdeafcafe (dismissed):

dev/docs/adr/022-event-log-source-of-truth.md
```diff
@@ -0,0 +1,159 @@
...
+  projection tables= derived lean views · regenerable from event_log via replay
+
+LEAN BOUNDARIES  (what stays small at each hop)
+  Command queue (Redis)    ≤ 256 KB per job   (inline data or spoolRef — bounded by edge size-check,
+                                                with fail-open fallback to full inline on S3 outage)
+  event_log row (CH)       unbounded, ZSTD-compressed  (the only place full content lives)
+  Projection queue (Redis) ≤ ~64 KB per IO attr        (bounded by interposition)
+  Fold cache (Redis)       same as projection queue    (downstream of lean step + toCacheable)
+  stored_spans / trace_summaries (CH)   ≤ ~64 KB per IO attr  (downstream of lean step)
+```
```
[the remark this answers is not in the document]
drewdrewthis: Fixed — added a `text` language tag to the EDGE — OTLP collector walkthrough fence.

dev/docs/adr/022-event-log-source-of-truth.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Rules
+
+- `leanForProjection` is **the** single source of truth for the leaned shape. It is invoked at the dispatch interposition AND in `replayExecutor.apply` before invoking projection handlers. Any future place that consumes events for projection MUST go through it. Tests pin this invariant.
+- `langwatch.reserved.*` is the server-internal namespace. Client-supplied attributes in that namespace MUST be dropped at the command worker (already enforced by `stripReservedAttributes`, with the `langwatch.reserved.causality_depth` passthrough for nlpgo loop detection). `leanForProjection` SETS the `langwatch.reserved.eventref.<attrKey>` attribute server-side after the strip.
+- `langwatch.reserved.eventref.<attrKey>` carries `{ field: <attrKey> }` only — the `eventId` is implicit in the row carrying the eventref (it's the same span). `BlobStore.get` derives `(TenantId, AggregateType, AggregateId, EventId)` from the read context.
```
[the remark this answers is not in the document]
drewdrewthis: Not a conflict — this is the deliberate design, not a bug. The ADR states `eventId` is *implicit in the row carrying the eventref (it's the same span)*, and `BlobStore.get` derives `(TenantId, AggregateType, AggregateId, EventId)` from the read context (see also line 38: `get({eventId, field})`). The pointer intentionally carries `{ field }` only; the `eventId` is the EventId of the very row being read, so there is no traceId/eventId lookup ambiguity to reintroduce. Embedding `eventId` in the pointer payload would be redundant. Leaving the contract as-is.

drewdrewthis:
**Conflict-merge with `main` done; one pre-existing flaky test is the only red check.**

Resolved the merge conflict against `main` (was `CONFLICTING`) and pushed — PR is now `MERGEABLE`. Three conflicts combined both sides (not pick-one):
- `presets.ts` — kept the ADR-022 `blobStore`/`ioExtractionService` construction **and** main's `setDiscoverBroadcaster` SSE wiring.
- `trace-request-collection.service.ts` — took main's `processSpan` → `ingestNormalizedSpan` delegation, and grafted the ADR-022 `processCommandData` hook into `ingestNormalizedSpan` (the auto-merge would otherwise have dropped it).
- `pipeline.ts` — kept the `retentionOrphanSweepReactor` block + `blobStore` conditional, added main's `{ deduplication: RECORD_SPAN_DEDUPLICATION }` to both branches, **and** restored the `retentionOrphanSweepReactor?` declaration on `TraceProcessingPipelineDeps` that the auto-merge silently dropped (caused a TS2339 the local typecheck caught).

Verified `pnpm run typecheck` clean (0 errors, after `prisma generate`). Also added the required `## Deployment Impact` section (deployment-impact-check now green) and cleared two CodeRabbit threads (one MD040 lint fix; one false-positive refuted — the `eventref` pointer carries `{ field }` with `eventId` implicit-in-read-context, which is the deliberate ADR design).

**Correction (was wrong above — retracting): `test-integration (5)` / `langwatch-app-complete` is a REAL, PR-owned failure, not a pre-existing flake.** `replay-projection-parity.integration.test.ts:210` gets 0 rows from `event_log` (expected 3). I earlier called this "pre-existing" by matching the failure on `cd83c06a4` — but that commit is on this PR's own branch, not `main`, and the test file is net-new in this PR (it does not exist on `origin/main`), so its failure is this PR's to fix by definition. The "0 rows, not 1" symptom also disproves the in-file `ReplacingMergeTree`-dedup theory (dedup collapses to 1 row, not 0) — something is preventing the `event_log` insert from being visible to the immediate SELECT in the CI ClickHouse (custom `local_primary` storage policy). Actively reproducing against CI's exact ClickHouse env to find the true root cause; will follow up with the fix. **This blocks the `langwatch-app-complete` gate — the PR is NOT mergeable-clean until it's green.**

🤖 via /drive-pr

drewdrewthis:
**Fixed — `test-integration (5)` / `langwatch-app-complete` now green on `b507234`.**

Root cause (verified): `event_log` has a runtime DELETE TTL installed by `clickhouse:migrate` → `reconcileTTL` — `SHOW CREATE TABLE event_log` confirms `TTL if(_retention_days > 0, toDateTime(EventOccurredAt / 1000) + toIntervalDay(_retention_days), toDateTime('2106-01-01'))`. The fixture stamped a hardcoded Nov-2023 `EventOccurredAt` (`1700000000000`); with the default `_retention_days=308` those rows expired ~Sep 2024, so under CI load the background merge TTL-deletes the freshly-inserted parts inside the insert→select window → 0 rows (the "shard-5 flake"). Both prior diagnoses were wrong — `async_insert` visibility and `ReplacingMergeTree` dedup-on-empty-`IdempotencyKey` (dedup yields 1 row, not 0).

Fix (test-only, `b507234`): stamp `EventOccurredAt = Date.now()` so rows land in the current retention window, plus `_retention_days: 0` (TTL → year 2106, never expires) as a deterministic belt-and-suspenders. Verified locally against the storage-policy ClickHouse (test passes, `SHOW CREATE TABLE` confirms the TTL clause, typecheck clean) and now in CI: `test-integration (5)` and `langwatch-app-complete` both `success`.

PR is `MERGEABLE`; all checks green, no unresolved threads. Only `REVIEW_REQUIRED` (human approval) remains.

> Latent product note (NOT this PR's regression, worth a separate issue): if a real tenant's `event_log` rows are TTL-deleted by retention, a triggered replay would read 0 rows from `event_log` and silently produce empty projection state rather than a "data expired" signal. Correct retention behavior, but the silent-empty-on-replay is a sharp edge.

🤖 via /drive-pr

0xdeafcafe (approved):

PR #4219, opened 2026-05-27
fix(feature-flags): build PostHog client lazily so SYSTEM-only processes don't poll

rogeriochaves:
## What

The workers and event-sourcing pipeline were logging `[FEATURE FLAGS] Feature flags quota limit exceeded - unsetting all local flags`, even though those processes only ever evaluate SYSTEM-scoped flags, which resolve from postgres and never call PostHog.

## Why it happened

The warning was not from a flag check. It was from posthog-node's background flag-definition poller, which starts the moment the client is constructed:

1. The workers and the event-sourcing reactors statically import `featureFlagService`.
2. That transitively imports `server/posthog.ts`, whose module-level `_posthogInstance = new PostHog(...)` ran at import time.
3. Because `POSTHOG_FEATURE_FLAGS_KEY` is set, the client was built with local evaluation enabled, which spins up a timer that fetches all flag definitions every 5 minutes, in every process, regardless of whether any flag is ever evaluated.
4. On top of that, the `featureFlagService` singleton constructor eagerly built the PostHog legacy backend, which also constructed the client.

So every workers pod ran that 5-minute poll forever, for flags it never reads. Across replicas, that tripped the PostHog feature-flags billing quota.

The polling was originally justified by per-span killswitch checks, but those killswitches moved to postgres, so on the workers the poller had no remaining purpose.

## The fix

- `server/posthog.ts`: construct the client on the first `getPostHogInstance()` call instead of at module load. Importing the module no longer starts the poller.
- `FeatureFlagService`: build the legacy backend (PostHog when configured, memory otherwise) lazily on the first PRODUCT or unregistered flag evaluation, not in the constructor.

A process that only reads SYSTEM flags now never constructs the PostHog client and never polls. The web and API process still builds it on its first PostHog-backed flag evaluation or analytics capture, where the local-evaluation cost tradeoff still applies.

## Tests

- `posthog-local-evaluation.unit.test.ts`: importing the module alone no longer constructs the client; construction happens on first `getPostHogInstance()` and is memoized.
- `featureFlag.service.lazy-legacy.unit.test.ts`: evaluating a SYSTEM flag never constructs the PostHog backend; a PRODUCT flag builds it once on demand.
- Added two scenarios to `specs/analytics/posthog-cost-control.feature` covering both paths.

### DIFF
--- a/langwatch/src/server/__tests__/posthog-local-evaluation.unit.test.ts
+++ b/langwatch/src/server/__tests__/posthog-local-evaluation.unit.test.ts
@@ -8,6 +8,12 @@
  * definitions in the background and evaluates them in-process instead of
  * hitting the /flags endpoint per call.
  *
+ * Constructing the client with local evaluation starts that background poller,
+ * so the client is built lazily on first getPostHogInstance() call rather than
+ * at module load. A process that only reads SYSTEM flags from postgres (workers,
+ * the event-sourcing pipeline) imports this module but never requests the
+ * instance, so it never starts the poller and never burns the flags quota.
+ *
  * @see specs/analytics/posthog-cost-control.feature
  */
 import { describe, expect, it, vi, beforeEach } from "vitest";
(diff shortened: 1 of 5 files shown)


rogeriochaves (commented):
Review findings: none.

I checked the lazy construction path against the PostHog cost-control scenarios. `posthog.ts` no longer constructs `posthog-node` on import, and `FeatureFlagService` now only builds the legacy PostHog-backed service when a PRODUCT or unregistered flag actually falls through to legacy evaluation. The SYSTEM flag paths used by workers/event-sourcing continue to resolve via env/store/default without touching PostHog.

Verification run locally in `/tmp/langwatch-pr-4219/langwatch`:

```bash
pnpm install --frozen-lockfile
pnpm prisma generate
NODE_ENV=test PINO_LOG_LEVEL=warn pnpm exec vitest run \
  src/server/__tests__/posthog-local-evaluation.unit.test.ts \
  src/server/featureFlag/__tests__/featureFlag.service.lazy-legacy.unit.test.ts \
  src/server/featureFlag/__tests__/featureFlag.service.test.ts \
  src/server/featureFlag/__tests__/featureFlag.service.resolver.unit.test.ts \
  src/server/__tests__/posthog.unit.test.ts \
  --reporter=dot
```

Result: 5 test files passed, 38 tests passed.

Residual risk / test gap: there is good coverage for SYSTEM flags not constructing legacy and PRODUCT flags constructing legacy on demand. I did not see an explicit new test for the PRODUCT path where a postgres operator override exists and should short-circuit before legacy construction; existing resolver tests cover override semantics, but not that specific cost-control assertion.

PR #4220, opened 2026-05-27
fix(event-sourcing): park paused-tenant groups out of the dispatch scan

rogeriochaves:
## Problem

A tenant pause left the tenant's groups in the `ready` zset and skipped them in place during the dispatch scan. The dispatch scan is bounded (scanBudget 10000 with the cap on). When a paused tenant has a large due-now backlog at the front of the shared ready zset, the scan exhausts its entire budget skipping that one tenant and never reaches groups for other tenants deeper in the zset, so the whole queue stalls.

This is the head-of-line half of the 2026-05-27 incident: pausing a runaway tenant (the operator mitigation lever itself) plugged dispatch for all other tenants. Live prod read during the incident: `ready` held 33k groups across 64 tenants, the first non-paused group sat at rank 15001 (beyond the 10000 scan budget), so 63 tenants saw zero throughput.

## Fix

Park paused-tenant groups OUT of `ready`, reusing the over-cap parking path that already exists:

- **Dispatch** (`DISPATCH_LUA` + `DISPATCH_BATCH_LUA`): when a group's tenant is paused, `parkGroup` it out of the scan instead of skipping in place, so the scan reaches other tenants. Pipeline-level pauses stay skip-in-place (rarer, and need the job data to classify).
- **`unparkUpTo`**: refuses to restore a still-paused tenant. This one guard covers COMPLETE-unpark, the dispatch reconcile, and the cap=0 drain, so paused groups stay parked until the pause is lifted (pause wins over cap headroom and the kill-switch).
- **Unpause**: the existing reconcile restores the tenant's parked groups on the next poll once the pause clears. Resumption is via the reconcile (time-gated, within a couple seconds) rather than in-place, bounded by cap so a huge backlog cannot re-clog the scan in one shot.

The existing `addToReadyOrParked` invariant already keeps a parked group parked when STAGE / UNBLOCK / RETRY / COMPLETE write to it, so no other writer needs changing.

## Tests

`scripts.integration.test.ts` (real Redis, testcontainers), 130/130 green:
- a paused tenant's group is parked out of ready (not skip-in-place)
- a paused tenant's backlog does not block another tenant
- a still-paused tenant is not restored by the reconcile
- unpause restores the parked group and resumes dispatch

Specs added to `specs/queue-pausing/queue-pausing.feature`.

## Scope

This is the head-of-line half. The root cause of the eval storm itself (the scheduled topic-clustering worker appending `topic_assigned` events that re-fire the `evaluationTrigger` reactor over historical traces) is fixed separately on the eval-trigger PR.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
@@ -1433,7 +1433,7 @@ describe("GroupStagingScripts", () => {
     }
 
     /** @scenario Pausing a tenant halts dispatch for that tenant only */
-    it("skips a group whose tenantId-prefix is in paused-jobs as 'tenant:<id>'", async () => {
+    it("parks a paused tenant's group OUT of ready (not skip-in-place) and returns null", async () => {
       await scripts.stage(
         makeJob({
           stagedJobId: "j1",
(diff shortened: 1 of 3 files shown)


rogeriochaves:
Superseded by #4221 — all five 2026-05-27 incident fixes consolidated into one PR per rchaves. My pause-park commit 7c7a57e03 was cherry-picked into #4221 as f7f7d2109, verified content-identical (empty diff, same 3 files, +144/-8). Closing in favor of #4221.

PR #4218, opened 2026-05-27
feat(langy): per-session OpenCode workers + per-request creds

Aryansharma28:
## Summary

The Fix 2 architecture. Pod's `server.js` becomes a process pool manager that spawns one `opencode` subprocess per Langy conversation, with that session's credentials injected as env vars at spawn time. The pod itself holds NO user credentials. OS process boundaries make per-session isolation real, not aspirational.

- Stacked on top of **PR #4217** (cut + harden). Targets `langy/cut-memory-and-harden-pod` so the diff shows only the manager-rewrite delta.
- 9 files changed, **+949 / -267**.
- 14/14 langy service unit tests pass (8 existing + 6 new for credential service). Typecheck clean.

### Control plane (`langwatch/src/server/`)

**New: `services/langy/LangyCredentialService.ts`** — resolves the four credentials a Langy worker needs:
- `langwatchApiKey` — `project.apiKey` (`sk-lw-*`) for MCP tool calls
- `llmVirtualKey` — auto-provisioned per-project VK secret for the AI gateway
- `langwatchEndpoint` — from `LANGWATCH_API_URL` env
- `gatewayBaseUrl` — from `LW_GATEWAY_BASE_URL` env

VK is provisioned on first use via `VirtualKeyService.create`. Picks the first `GatewayProviderCredential` by `fallbackPriorityGlobal`; throws `LangyCredentialResolutionError` with actionable message if none exists. Plaintext is stored encrypted in `ProjectSecret` under name `langy_vk_secret`. Subsequent calls decrypt + return. Race-safe: P2002 on `ProjectSecret.create` triggers re-read of winner's secret (losing-race VK is orphaned — acceptable v1 cost, documented in code).

**Updated: `routes/langy.ts` `/chat` handler**:
- Requires `LANGY_INTERNAL_SECRET` env (503 if missing)
- Calls `LangyCredentialService.getOrProvision` per request; returns 409 on `LangyCredentialResolutionError`
- POSTs to `${OPENCODE_AGENT_URL}/chat` (was `/run`) with `{ conversationId, prompt, system, credentials }` and `Authorization: Bearer ${LANGY_INTERNAL_SECRET}`
- Sends `langyOverride` as `system` instead of concatenating into `prompt` (cleaner separation, lower jailbreak risk)

**New: 6 unit tests for `LangyCredentialService`** covering: existing secret path (no re-provision), provision + store path, missing provider credential, missing project, missing env config, race on `ProjectSecret.create`.

[...]

PR #4224, opened 2026-05-27
fix(langy/manager): same-conv race + stream cleanup + error surface (review followups to #4218)

Aryansharma28:
## Summary

Stacked fixes against PR #4218's own review. Five findings from the self-review — two BLOCK, three MAJOR — all on the pod-side manager + the control-plane stream loop. Targets `langy/per-session-manager` so the diff shows only the followup delta.

- 2 files changed, +180 / -34
- Server.js worker-lifecycle rewrite, ~80 added lines + structural reshape
- langy.ts gets a 13-line error-event branch + an error-message map

### 1. Cold-start race on same conversationId (BLOCK)

The original `spawnWorker` did `workers.set(...)` AFTER `await waitForReadiness` + `await createOpenCodeSession` — a 1–2s window during which a second concurrent first-turn for the same conversationId saw \`workers.get() === undefined\`, called \`spawnWorker\` again, and ended up with two opencode subprocesses where the second's \`workers.set\` overwrote the first's. The first child was orphaned: not in the map, not reaped, holding a port + RAM until pod restart. \`MAX_WORKERS\` couldn't catch it.

\`getOrSpawnWorker\` now inserts a pending-promise entry into the map BEFORE the first await. Concurrent waiters await the same \`entry.ready\` and either share the resolved worker or share the rejection. Wrapped in a \`while (true)\` loop so that if N waiters were all blocked on a failed spawn, only the first to resume creates a replacement; the rest see the new entry on the next iteration.

### 2. createOpenCodeSession leak (BLOCK)

\`await createOpenCodeSession(port)\` sat OUTSIDE the try/catch that killed the child on \`waitForReadiness\` failure. If opencode was ready but the \`/session\` POST returned 500, the child was left alive, never killed, never in the map, never reaped — same orphan pattern.

The try now wraps both \`waitForReadiness\` and \`createOpenCodeSession\` so any failure between spawn and "session created" kills the child.

### 3. Manager error events had no user surface (MAJOR)

\`server.js\` would write \`{type:"error", error:"at-capacity"}\` then \`res.end()\`, but the control plane's stream loop only handled \`type === "text"\` and \`type === "message.part.delta"\`. The user saw an empty assistant reply with no indication anything went wrong, no log, no 502.

Added an \`event.type === "error"\` branch in the langy.ts stream loop that writes a user-visible delta via the AI SDK writer + logs the error. \`AGENT_ERROR_MESSAGES\` map provides friendlier strings for the three known error codes (\`at-capacity\`, \`turn-in-flight\`, \`session-not-found\`); unknown errors fall through with the raw error string.

### 4. Concurrent turns on the same conversationId interleaved events (MAJOR)

Nothing serialised requests per conversationId. Two simultaneous turns on the same conversation both posted to the same opencode session, both opened \`/event\` streams, both filtered by the same sessionID — events for turn A leaked into turn B's stream until either's terminal event cancelled the other reader. Realistic scenario: user double-clicks send, or HTTP retry fires mid-stream.

Each WorkerEntry now carries an \`inFlight\` boolean. The \`/chat\` handler checks it under sync code right after \`getOrSpawnWorker\` returns; if true, writes \`{type:"error", error:"turn-in-flight"}\` and ends. The control plane surfaces this with a clear message.

### 5. streamPromise unhandled rejection + connection leak (MAJOR)

\`const streamPromise = streamSessionEvents(...)\` was kicked off then awaited only on the success path. Two error paths (\`session-not-found\` early-return, generic \`throw\`) left the promise pending; if it later rejected (e.g. opencode died mid-stream), node logged an unhandled-rejection warning — and could crash under \`--unhandled-rejections=strict\`.

streamPromise now has a defensive \`.catch\` that stores the error into a local \`streamError\` variable. The error paths \`abort.abort()\` before returning, which signals \`streamSessionEvents\` to cancel the reader and resolve cleanly. Without the abort, the SSE connection to opencode would stay open indefinitely waiting for a terminal event that the failed turn would never produce.

## Bonus fixes folded in

- **Late-firing exit handler eviction** — \`child.on("exit")\` now identity-checks \`current?.info?.child === child\` before deleting the entry. Otherwise a delayed exit event from a previously-killed worker could evict a fresh entry that had since replaced it.
- **Reaper safety** — skips entries where \`info === null\` (still spawning, no child handle yet to SIGTERM) or \`inFlight === true\` (mid-turn, killing would orphan the user's response stream).
- **killWorker safety** — uses \`w.info?.child.kill(...)\` instead of \`w.child.kill\`, since pending entries don't have a child yet.

## Out of scope (filed for follow-up, not in this PR)

- MINOR reaper kills mid-turn on >10min responses (now partially mitigated by inFlight skip, but lastSeen still only updates per-request, not per-event)
- MINOR RBAC: any user with \`evaluations:view\` implicitly provisions a project VK on first chat
- MINOR small race: \`/event\` subscription vs first \`postMessage\` delta
- MINOR \`getVercelAIModel(projectId)\` check in \`/chat\` is dead under the new architecture
- MINOR \`req.on('data')\` accumulates body with no size cap
- NIT \`extractAssistantText\` is misnamed (also used for user messages)

## Test plan

- [x] Self-review of fixes turned up + corrected two additional bugs introduced by my own first pass (loop in getOrSpawnWorker, abort.abort on error paths)
- [ ] \`pnpm typecheck\` (worktree needs \`pnpm install\` first; server.js is \`@ts-nocheck\`, langy.ts changes are isolated)
- [ ] Manual smoke test in minikube: build image, deploy, hit \`/chat\` with two concurrent same-conversation requests, verify second gets \`turn-in-flight\` and the first completes cleanly
- [ ] Verify \`at-capacity\` end-to-end: set \`LANGY_MAX_WORKERS=1\`, start one conversation, start a second conversation, verify second user sees the friendly "Langy is at capacity" message
- [ ] Verify race-recovery: simulate \`/session\` failing once on a fresh conversation, verify the next request for the same conversationId spawns successfully (existing entry was cleaned up)


Aryansharma28:
## E2E failures triage — pre-existing on langy/full, merging anyway

The CI run shows 4 failing e2e tests + 2 failing checks (e2e + e2e-complete):

- `tests/scenarios/scenario-execution.spec.ts:41` — `simulationsHeading` not found
- `tests/scenarios/scenario-execution.spec.ts:133` — same `simulationsHeading` selector
- `tests/scenarios/scenario-library.spec.ts:34` — element not found
- `tests/settings/plans-comparison.spec.ts:13` — element not found

**These are not caused by this PR.** Evidence:

1. **`main` is green** — PR #4222 passed all e2e checks minutes before this CI ran.
2. **#4217 (already merged into langy/full) had the same 4 failures** — the team explicitly accepted these failures when shipping #4217.
3. **#4218 (open, parent of this PR) has the same 4 failures.**
4. **No PR in our stack touches any `*.spec.ts` file** — our changes are all backend (Langy services, k8s pod manager). The failing tests are UI Playwright tests in `agentic-e2e-tests/`.

So these failures live in `langy/full`'s accumulated state (something between `main` and `langy/full` broke UI selectors used by these tests) and are not introduced or exacerbated by anything in this PR.

Merging on the precedent set by #4217. Filing the UI selector breakage as a follow-up to be triaged by whoever owns the `agentic-e2e-tests` repo.

PR #4228, opened 2026-05-28
fix(event-sourcing): self-heal tenant soft-cap in-flight count on ungraceful worker death

rogeriochaves:
## Problem

The per-tenant dispatch soft cap counts a tenant's in-flight groups with a scalar Redis counter (`tenant_active:<tenant>`): INCR on dispatch, DECR on COMPLETE, TTL refreshed by the heartbeat.

When a worker dies **ungracefully** (Redis connection dropped mid-job, OOM, node replacement) it never runs COMPLETE, so the counter is never decremented. It does not self-heal either, because ongoing dispatch of the tenant's other groups keeps refreshing its TTL. The counter drifts up to the cap and pins there. From then on the tenant reads as permanently at cap, so every one of its groups is parked out of the dispatch scan: a live tenant stalls with thousands of groups stranded while the workers sit mostly idle.

This is not theoretical. It happened in prod when an ElastiCache node replacement dropped every worker's Redis connection mid-job, and again on each HPA scale-down that force-closed pods holding long-running eval jobs. The only mitigation was a manual counter reset, which had to be repeated every minute because the leak recurs under load.

## Fix

Replace the scalar counter with a per-tenant ZSET `tenant_active_z:<tenant>` where each member is a `groupId` and its score is the slot's expiry (`nowMs + activeTtlSec*1000`), the same liveness window as the group's `:active` heartbeat key.

The in-flight count is the number of members whose expiry is still in the future. A slot whose worker died keeps its now-past score and falls out of the count once its expiry passes, exactly like the `:active` key does. So an ungraceful death self-heals within the active TTL, with no operator reset and no separately-refreshed counter to perpetuate the leak.

- **cap check / unpark headroom**: count of non-expired members (lapsed members GC'd first so dead-worker entries cannot grow the set unbounded)
- **dispatch**: `ZADD` the slot at its expiry score
- **REFRESH / retry backoff**: `ZADD XX` score-bump (only an existing slot, never re-create a freed one)
- **COMPLETE / restage**: `ZREM` the slot
- **reconcile**: reads the same live count

Converted across all six scripts that touched the counter (DISPATCH, DISPATCH_BATCH, COMPLETE, REFRESH, RESTAGE_AND_BLOCK, RETRY_RESTAGE) plus the TS call sites. Deploy is a clean cutover: old scalar counters TTL out, the ZSETs build fresh, one benign warmup within the active TTL.

## Scope

This fixes the **trigger**. The amplifier that turned the stall into an out-of-memory event (heavy per-trace fold state piling on the blocked tenant) is the separate huge-trace fold-state issue and is a follow-up (bound per-trace fold size at ingestion).

## Tests

- New `@self-heal` BDD scenario in `specs/event-sourcing/tenant-soft-cap.feature`.
- Integration test that simulates an ungraceful mass death (active keys deleted, no COMPLETE, stale slot scores) and asserts the live count drops to 0 and the parked group dispatches again with no manual reset.
- Existing tenant-cap integration tests translated from scalar assertions to the ZSET model.
- 132/132 integration tests green against testcontainer Redis; typecheck clean.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
@@ -2481,7 +2481,10 @@ describe("GroupStagingScripts", () => {
     // Live in-flight count for a tenant = ZSET members whose expiry score is
     // still in the future relative to nowMs (lapsed members no longer count).
     async function tenantLiveCount(tenantId: string, nowMs: number) {
-      return redis.zcount(tenantActiveZKey(tenantId), `(${nowMs}`, "+inf");
+      // Inclusive lower bound to mirror the Lua live-count contract: the script
+      // GCs only scores STRICTLY below nowMs (ZREMRANGEBYSCORE "-inf" "(nowMs")
+      // then ZCARDs, so a slot scored exactly nowMs is still live.
+      return redis.zcount(tenantActiveZKey(tenantId), `${nowMs}`, "+inf");
     }
 
     async function stageOne({
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts
@@ -2438,18 +2438,21 @@ describe("GroupStagingScripts", () => {
   });
 
   /**
-   * Post-2026-05-11 tenant soft-cap integration coverage.
+   * Tenant soft-cap integration coverage.
    *
-   * These tests drive the new Lua paths against a real Redis (via
+   * These tests drive the Lua paths against a real Redis (via
    * testcontainers) and prove:
-   *   - the `tenant_active:<tenantId>` counter stays consistent across
-   *     DISPATCH (INCR) → COMPLETE / RESTAGE_AND_BLOCK (DECR/DEL) and
-   *     REFRESH / RETRY_RESTAGE (TTL renewal)
+   *   - the per-tenant in-flight ZSET `tenant_active_z:<tenantId>` stays
+   *     consistent across DISPATCH (ZADD), COMPLETE / RESTAGE_AND_BLOCK (ZREM)
+   *     and REFRESH / RETRY_RESTAGE (score renewal)
+   *   - a slot whose worker dies ungracefully lapses out of the live count once
+   *     its expiry score falls into the past, so the cap self-heals with no
+   *     operator reset
    *   - cap enforcement at the scheduler level
    *   - widened scan budget keeps cross-tenant fairness when an
    *     over-cap tenant dominates the head of the ready zset
-   *   - cap=0 (kill switch) leaves the counter machinery completely
-   *     inert — no `tenant_active:*` keys ever appear
+   *   - cap=0 (kill switch) leaves the ZSET machinery completely
+   *     inert: no `tenant_active_z:*` keys ever appear
    *
    * All scenarios mutate `LANGWATCH_DISPATCH_TENANT_CAP` per test
    * (readTenantCap reads process.env at call time on purpose) and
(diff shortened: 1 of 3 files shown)


rogeriochaves (commented):
Reviewed the full diff (all 6 Lua scripts + TS wiring + tests). This is correct and a clean kill of the trigger — the ZSET-scored-by-`activeUntil` model is exactly right: a dead worker's slot carries a past score and lapses out of `tenantActiveCount` (ZREMRANGEBYSCORE + ZCARD) within `activeTtl`, mirroring the group-level self-heal already at dispatcher.ts:490. Verified: all sites converted (no scalar `tenant_active:` refs remain), COMPLETE's new `nowMs` ARGV[6] is passed by its only caller, REFRESH/RETRY use `ZADD XX` so a racing heartbeat can't resurrect a slot COMPLETE legitimately freed (nice — and the new test covers it), and cap=0 stays fully inert (no `_z` keys). **Approving.**

Three notes, none blocking:

**1. Rollout-window inconsistency (deploy awareness, not a code fix).** During the rolling deploy, old pods read/write the scalar `tenant_active:` and new pods the zset `tenant_active_z:` — two separate counts, so each fleet undercounts the other's in-flight work and a hot tenant can transiently exceed its cap for the ~minutes of the rollout. It self-heals once fully rolled, and the failure direction is over-dispatch (safer than the strand we're fixing), so this is acceptable — but worth being conscious of, especially if we deploy while project_FzsKx is still mid-burst: watch its concurrency + Redis memory through the rollout. The old scalar keys self-expire (~300s TTL), so no migration/cleanup needed.

**2. Stop the periodic counter-reconcile once this deploys.** The live bridge SETs the scalar `tenant_active:` key, which the new code ignores. Harmless (old keys TTL out) but pointless after cutover — kill the loop when the rollout completes.

**3. Minor: `tenantActiveCount` now mutates on the cap-check path** (ZREMRANGEBYSCORE before ZCARD) where the old code did a read-only GET. Cheap — it's cached per-tenant per-EVAL via `tenantCapCache`, and the zset is bounded by cap + lapsed members — but it does turn each cap-check into a small write. Fine at current scale; flagging only so it's a known property.

Complementary follow-up (out of scope here, correctly called out in the PR body): the AMPLIFIER — unbounded per-trace fold state (measured 488KB avg / 2.3MB tail on the hot tenant this morning) needs an ingestion spans-per-trace bound. This PR kills the trigger; that bounds the blast radius.

rogeriochaves (commented):
## Findings

**P3 — The acceptance spec still describes the removed scalar counter contract**

File/line: `specs/event-sourcing/tenant-soft-cap.feature:39` (also `:47`, `:53`, `:59`, `:81`, and `:145`)

Why: this PR replaces the scalar `tenant_active:<tenant>` counter/TTL model with per-group liveness slots in `tenant_active_z:<tenant>`, but the feature file still says the counter increments/decrements, has its TTL aligned, expires as a missing counter, and is absent when cap=0. That leaves the BDD acceptance criteria contradicting the new implementation and can send future fixes/tests back toward the counter semantics this PR is removing.

Minimal fix direction: rewrite those scenarios and comments around ZSET slots: add/remove a group slot, refresh/retry update the slot expiry score, live count excludes lapsed slots, and cap=0 creates no `tenant_active_z:*` keys.

**P3 — The live-count boundary treats an exactly-expired slot as still active**

File/line: `langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts:74`

Why: the helper comment and test helper define a live slot as one whose expiry is still in the future (`score > nowMs`), but `ZREMRANGEBYSCORE key -inf (nowMs` only removes scores strictly less than `nowMs`. A slot with score exactly equal to `nowMs` remains in the ZSET and is counted by `ZCARD`, so the cap can stay full for one extra millisecond and the behavior differs from the documented/tested contract.

Minimal fix direction: remove lapsed entries through `nowMs` inclusively (or count with an exclusive lower bound instead of `ZCARD`) and add a small boundary assertion for `score == nowMs`.

## Verification

- `pnpm exec vitest run src/server/event-sourcing/queues/groupQueue/__tests__/tenantCap.unit.test.ts --reporter=dot` — 6 passed
- `pnpm exec vitest run -c vitest.integration.config.ts src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts --reporter=dot` — 132 passed
- `git diff --check 56fb96f0a88f45eb44a74d48c744c10d050f51c5...HEAD` — passed

No P0-P2 functional or security/privacy issues found in the reviewed scope.

rogeriochaves:
Addressed both CodeRabbit findings in 4d69e43:

1. **`tenantLiveCount` boundary** — switched the test helper from an exclusive `(nowMs` to an inclusive `nowMs` lower bound so it mirrors the Lua live-count contract exactly (the script GCs only scores strictly `< nowMs` via `ZREMRANGEBYSCORE "-inf" "(nowMs"` then `ZCARD`, so a slot scored `== nowMs` is still live). No longer under-counts at the boundary. Suite still 132/132 green.

2. **Self-heal scenario precondition** — added an explicit `And the same tenant has additional groups parked over its cap` Given so the parked backlog is stated, not implied. Kept it behavioral and did not embed the `tenant_active_z:` key name in the scenario, per this repo's feature-file convention (scenarios describe behavior from the user's perspective, not internals). The implementation-key traceability lives in the integration test + the Lua comments instead.

PR #3528, opened 2026-04-27
feat(tracing): 2 trace 2 furious, tracing evolved

0xdeafcafe:
drafting this race

Screenshot 2026-05-01 at 01 03 02

PR #4231, opened 2026-05-28
test(alertTrigger): reactor unit-test coverage + outbox TODO

0xdeafcafe:
## Summary

Salvages the one piece of net-new value from #3351 (now being closed) and lands it against main's current shape.

- **New unit-test file** for the trace-pipeline \`alertTrigger\` reactor (12 tests). Mirrors the existing \`evaluationAlertTrigger\` unit test — main had no equivalent.
- **\`TODO(outbox)\`** on the two stake-sensitive dispatch reactors (\`alertTrigger\` + \`evaluationAlertTrigger\`) where silent failures matter. Scoped explicitly to these two, not all reactors.

## Why this is a separate PR (and #3351 is closed)

#3351 was opened pre-#3528 to introduce event-driven triggers. Its design then shipped through #3528 (\"2 trace 2 furious, tracing evolved\") on 2026-05-01, after which main improved on it three more times:

- #4205 — removed events from fold state (perf, O(1) per span)
- #4211 — memoized trace read-derivation
- #4221 — \`_originGuardedReactor\` with 5 guards (loop prevention)

#3351's latest delta vs main was net-regressive: it replaced \`claimSend\` (atomic \`@@unique\` claim) with \`hasSentForTrace\` + Redis SETNX + \`recordSent\` (3 round trips, 2 systems, race window); dropped 2 of 5 origin guards (would let topic-clustering re-fan triggers across historical traces); and reverted the perf work that removed span-count-scaling collections from fold state. The unit-test file was the only thing worth keeping — that's this PR.

## Tests covered by the new file

| Guard / behavior | Test |
|---|---|
| Stale event (>1h) | \`when event is old (replay/resync flood)\` |
| Non-message event type | \`when event type is not a message event\` |
| Trace >24h old | \`when trace itself is older than the 24h cap\` |
| Guardrail-blocked, no output | \`when trace is blocked by guardrail without output\` |
| Unresolved origin | \`when origin is not yet resolved\` |
| No active triggers | \`when project has no active triggers\` |
| Eval-filter triggers delegated | \`when trigger has evaluation filters\` |
| Match → claim + dispatch + lastRunAt | \`when trace filters match\` |
| No match → skip | \`when trace filters do not match\` |
| Lost claim → skip dispatch | \`when claim is lost (concurrent reactor or retry)\` |
| Event-fields filter → lazy deriveEvents | \`when a trigger filters on event fields\` |
| No event-fields filter → no derive | \`when no trigger filters on event fields\` |

## Test plan

- [x] \`npx vitest run src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/alertTrigger.reactor.unit.test.ts\` → 12/12 passing locally
- [ ] CI typecheck + test pass

### DIFF
--- a/langwatch/src/server/event-sourcing/pipelines/evaluation-processing/reactors/evaluationAlertTrigger.reactor.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/evaluation-processing/reactors/evaluationAlertTrigger.reactor.ts
@@ -188,6 +188,9 @@ export function createEvaluationAlertTriggerReactor(
             foldState: traceSummary,
           });
         } catch (error) {
+          // TODO(outbox): see alertTrigger.reactor — same silent-failure gap
+          // on the dispatch path. Both stake-sensitive dispatch reactors
+          // (this + alertTrigger) should share the outbox when we add it.
           logger.error(
             {
               tenantId,
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/alertTrigger.reactor.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/alertTrigger.reactor.unit.test.ts
@@ -1,0 +1,421 @@
+import { TriggerAction } from "@prisma/client";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { TraceSummaryData } from "~/server/app-layer/traces/types";
+import type { TriggerSummary } from "~/server/app-layer/triggers/repositories/trigger.repository";
+import type { ReactorContext } from "../../../../reactors/reactor.types";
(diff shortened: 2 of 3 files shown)


langwatch-agent:
No blockers

I did not find correctness, regression, security, or test-design issues in this PR. The new `alertTrigger` unit coverage is scoped to the guarded trace-reactor behavior and mirrors the existing `evaluationAlertTrigger` coverage without changing runtime behavior; the outbox TODOs are comment-only and scoped to the two customer-visible dispatch reactors.

Verification:
- `pnpm --dir langwatch run start:prepare:files`
- `pnpm --dir langwatch exec vitest run src/server/event-sourcing/pipelines/trace-processing/reactors/__tests__/alertTrigger.reactor.unit.test.ts` → 12/12 passing

Residual risk: I only ran the focused unit test, not the full app typecheck or complete test suite.

PR #4232, opened 2026-05-28
feat(trace-processing): bound spans per trace at ingestion to protect shared redis

rogeriochaves:
## What

Adds a hard per-trace span ceiling enforced at ingestion, so a single trace_id cannot balloon the shared event-sourcing infrastructure.

## Why

In the 2026-05-28 incident a trace_id reused across days accumulated around 26k spans. Every span became a queued fold job, so that one trace's Redis fold backlog grew into hundreds of KB (heavy tail past 2 MB) that re-piled on each ingest. Under a concurrent dispatch stall it could not drain, the shared single-threaded Redis filled toward its no-eviction ceiling, and span ingestion was rejected at the door for every tenant.

The per-tenant soft cap (#4214, #4220) bounds how many groups a tenant holds in flight. It does not bound how large one trace can get. This is the missing second ceiling.

## How

A per-trace counter in Redis, incremented synchronously at the recordSpan entrance (both the REST collector and the OTLP path converge there). Past the ceiling the span is front-dropped before staging, so it is never queued, stored, or folded, and the drop bounds all downstream cost.

Design decisions:

- Gates on the synchronous counter, not the fold spanCount. The fold count lags under exactly the burst this defends against, so gating on it would let the burst through.
- Sliding EXPIRE anchored to the existing `STALE_TRACE_THRESHOLD_MS` (1h), the pipeline's own trace-liveness window. A trace touched within the window keeps one bounded counter across days of slow reuse; a trace idle past it lets the counter expire. INCR and EXPIRE run in one Lua eval so there is no window where the key is left without a TTL.
- Dropped spans are surfaced as `partialSuccess.rejectedSpans`, the non-retryable OTLP signal, so the drop is visible and the SDK does not retry-storm re-sending spans that will only be dropped again.
- Counts attempts, so the check sits after dedup on the OTLP path; on the collector path a retry can over-count slightly, which only trips the ceiling marginally earlier (errs toward protection).

This complements the existing 512 fold derivation cap, which bounds fold compute but keeps counting and still stores and queues every span.

## Config

Default ceiling 10000 (far above legitimate traces, which run a handful of spans; well below the incident's 26k). `LANGWATCH_MAX_SPANS_PER_TRACE` retunes it; 0 disables the bound, mirroring `LANGWATCH_DISPATCH_TENANT_CAP`.

## Tests

- `specs/trace-processing/spans-per-trace-bound.feature` describes the behavior.
- Integration test against real Redis: admit within bound, drop past bound, keep counting for true magnitude, per-trace independence, log-once on first crossing, sliding TTL refresh, kill switch, custom ceiling.
- Unit test for the config reader (default, custom, kill switch, invalid).

### DIFF
--- a/langwatch/src/server/app-layer/dependencies.ts
+++ b/langwatch/src/server/app-layer/dependencies.ts
@@ -18,6 +18,7 @@ import type { LogRequestCollectionService } from "./traces/log-request-collectio
 import type { MetricRequestCollectionService } from "./traces/metric-request-collection.service";
 import type { TraceListService } from "./traces/trace-list.service";
 import type { TraceRequestCollectionService } from "./traces/trace-request-collection.service";
+import type { TraceSpanBoundService } from "./traces/trace-span-bound.service";
 import type { TraceSummaryService } from "./traces/trace-summary.service";
 import type { PlanProvider } from "./subscription/plan-provider";
 import type { SubscriptionService } from "./subscription/subscription.service";
@@ -56,6 +57,7 @@ export interface AppDependencies {
     collection: TraceRequestCollectionService;
     logCollection: LogRequestCollectionService;
     metricCollection: MetricRequestCollectionService;
+    spanBound?: TraceSpanBoundService;
   };
   evaluations: {
     runs: EvaluationRunService;
--- a/langwatch/src/server/app-layer/presets.ts
+++ b/langwatch/src/server/app-layer/presets.ts
@@ -58,6 +58,7 @@ import { TokenizerService } from "./traces/tokenizer.service";
 import { LogRequestCollectionService } from "./traces/log-request-collection.service";
 import { MetricRequestCollectionService } from "./traces/metric-request-collection.service";
 import { TraceRequestCollectionService } from "./traces/trace-request-collection.service";
+import { TraceSpanBoundService } from "./traces/trace-span-bound.service";
 import { TraceListService } from "./traces/trace-list.service";
 import { TraceListClickHouseRepository } from "./traces/repositories/trace-list.clickhouse.repository";
 import { NullTraceListRepository } from "./traces/repositories/trace-list.repository";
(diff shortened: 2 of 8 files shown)


langwatch-agent:
1 P1

- **P1** `langwatch/src/server/routes/misc.ts:766` - The span bound is not enforced at the actual `recordSpan` entrance, so `/api/track_event` can still enqueue/store/fold unlimited spans for a reused `trace_id`. This PR gates only the OTLP collection path (`TraceRequestCollectionService`) and the legacy REST collector before they call `recordSpan`, but `track_event` is another authenticated customer-facing ingestion route that calls `getApp().traces.recordSpan` directly. That leaves a bypass for the same shared-infra failure mode this feature is intended to stop: a single trace id can keep accumulating event spans without ever touching `TraceSpanBoundService`. Minimal fix: move the admission check into the central `recordSpan` command path before staging/queueing, or apply the same `spanBound.admit(project.id, traceId)` guard to every direct `recordSpan` producer, including `/api/track_event`.

Verification:
- `pnpm --dir langwatch exec vitest run src/server/app-layer/traces/__tests__/trace-span-bound.unit.test.ts` → 6/6 passing
- `pnpm --dir langwatch exec vitest run -c vitest.integration.config.ts src/server/app-layer/traces/__tests__/integration/trace-span-bound.integration.test.ts` → 8/8 passing

rogeriochaves:
Closing: the drop-based approach is the wrong lever. Dropping customer spans (even past a high ceiling) is not acceptable — LangWatch should handle arbitrarily large traces, never refuse data.

The correct model is to refuse WORK, never DATA: past a span threshold a trace is processed *lighter* (stop deriving the summary AND stop firing evaluations on a runaway trace), but every span is always stored and queryable. That folds into the existing `MAX_PROCESSED_SPANS = 512` threshold (which already caps fold derivation while still storing) rather than a new ingestion gate.

Superseded by:
- skip-eval folded into the existing 512 fold-derivation cap (lighter processing, zero drop) — small follow-up
- #4216 (S3 offload of large payloads) for lossless Redis-memory relief
- #4228 (tenant_active leak fix, shipped 3.4.0) + the upcoming work-conserving max-min fair dispatch (reclaims idle slots so a large/bursty trace never starves others)

No span dropping anywhere. Thanks to the reviewer for the track_event-bypass P1 — it correctly surfaced that a per-producer ingestion gate is fragile, which reinforced moving away from the gate entirely.

PR #4233, opened 2026-05-28
feat(observability): report handled client errors to PostHog via toaster

Aryansharma28:
## Summary

Widens client-side error capture so the **"the truth"** dashboard error tile reflects real user pain, not just uncaught crashes.

- Adds an **opt-in `error` field** to the shared `toaster` (`src/components/ui/toaster.tsx`). When a caught error is passed, it's forwarded to PostHog as a `$exception` (tagged `source=toaster`). The toast renders identically — `error` is stripped before reaching Chakra.
- Converts **61 catch-block error toasts across 51 files** to pass their caught error.

## Why

Audit of dashboard `651303` (insight `fNE339yg`) found the client error metric only saw uncaught crashes + React render errors. Of ~150 client catch blocks, only ~10 reported to PostHog — the other ~140 showed an error toast and **swallowed** the error. So real, user-visible failures (failed CSV upload, failed save, failed billing action) were invisible to the metric.

```
BEFORE: capture surface           AFTER: capture surface
uncaught crashes      ###         uncaught crashes        ###
render errors         ##          render errors           ##
handled failures      (none)      handled failures        ########  <- NEW
```

## Design: opt-in, not automatic

The presence of a **real `Error` object** is the signal. Validation toasts (`"Name is required"`) pass no `error` and stay silent, so we never re-pollute the metric with expected-input noise. Coverage grows as catch sites adopt `error`; noise stays at zero.

```ts
// before
catch (error) { toaster.create({ type: "error", title: "Upload failed" }) }
// after
catch (error) { toaster.create({ type: "error", title: "Upload failed", error }) }
```

## Scope / safety

- Conversions done via a TypeScript AST transformer (nearest-enclosing-catch scope analysis, shadowing-aware), not regex. Catch variable names used verbatim (`error: error` / `error: err` / `error: e`).
- Skipped: error toasts **not** inside a catch (validation), and optional catch bindings (`catch {` with no variable) — listed for follow-up.
- `pnpm typecheck` passes. No visual/behavioral change to toasts.
- Adds `specs/toaster-error-reporting.feature`.

## Follow-ups
- Optional lint rule: new `type:"error"` toasts inside a catch must pass `error` (prevents the gap re-growing).
- A handful of optional-catch-binding sites could adopt `catch (e)` + `error: e` if those failures are worth tracking.

### DIFF
--- a/langwatch/specs/toaster-error-reporting.feature
+++ b/langwatch/specs/toaster-error-reporting.feature
@@ -1,0 +1,32 @@
+Feature: Handled client errors reach PostHog via the toaster
+  As an operator watching the client error-rate quality metric
+  I want handled failures that show an error toast to also report to PostHog
+  So that the metric reflects real user pain, not only uncaught crashes
+
+  Context: audit of the "the truth" dashboard found that ~150 client catch
+  blocks showed an error toast but never reported the error, so handled
+  failures (failed uploads, saves, billing actions) were invisible to the
+  $exception metric. The shared toaster now accepts an opt-in `error` field:
+  passing the caught error forwards it to PostHog as a $exception. Validation
+  toasts pass no error and stay silent — the presence of a real error is the
+  signal, so we never report plain "field required" messages.
+
+  @unit
+  Scenario: An error toast that carries a caught error reports to PostHog
+    Given a catch block calls toaster.create with type "error" and the caught error
+    When the toast is created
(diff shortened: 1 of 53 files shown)


Aryansharma28:
Superseded — the toaster error-capture work was merged into `posthog-the-truth` and now ships via #4226. Keeping that branch as the single source. Closing this to avoid duplicate PRs.

PR #4238, opened 2026-05-28
feat(governance): post-merge follow-ups for the AI Governance Platform

rogeriochaves:
## Summary

Follow-up branch for items deferred from #3524 (the AI Governance Platform). It bundles the security review hardening, the gateway policy gaps, and the Claude Code dogfood fixes so they land as one reviewed PR.

## Security hardening

- **Virtual key scope-aware RBAC.** List visibility is now membership based: an org member sees organization-scoped keys, a team member sees that team's keys, and a sibling team's keys stay hidden. Create, update, rotate, and revoke are gated the same way. Personal-key listing gained `virtualKeys:viewOtherPersonal` so an admin can audit another user's keys while a plain member only ever sees their own. Built to the `vk-scope-rbac.feature` contract.
- **Privilege-escalation fix.** Every user is ADMIN of their own single-member personal workspace team. The organization-level permission check unioned across every `TeamUser` row without excluding personal teams, so any member silently resolved the full org ADMIN template, including `organization:manage` and `virtualKeys:viewOtherPersonal`, just by owning a personal workspace. The legacy fallback now filters personal teams; the legitimate team-scoped power still flows through the team's RoleBinding.
- **Routing policy tenancy.** `RoutingPolicyScope` is classified for the multi-tenancy guard, with the materialiser test repaired.
- **Docs.** MCP RBAC docs corrected to state the admin-read user requirement.

## Gateway policy enforcement

- **Regex model policy.** The control plane emits `policy_rules.models`, but the gateway converted only the tool, MCP, and URL targets, so regex model allow and deny rules were a silent no-op. The regex path is now wired through to enforcement. The plain `models_allowed` list already covered locking a key to specific models.

## Claude Code telemetry (dogfood batch)

- **Cost synthesis on the trace.** The log-to-trace converter now lifts cost, input and output tokens, cache tokens, and model off every `claude_code.api_request` record onto the span metrics, so Claude Code traces carry gen_ai-style cost and usage instead of landing thin. This runs on both the personal binding path and the admin ingestion path, since both flow through the same converter.
- **Tile mints a real personal binding.** The "Already using Claude Code" section dropped the uncopyable `<your-org-LangWatch-ingestion-URL>` placeholder env block. It now opens the same trace-ingest install drawer the personal grid uses, minting the caller's personal ingestion binding and showing a working endpoint plus their real token. The OAuth blurb collapsed to one line, the section is gated on the template being enabled, the Setup guide button opens an install modal (with a Node.js prerequisite line) instead of the docs, and the drawer gained a shell-persistence hint. A shared `usePersonalIngestionBinding` hook backs both the tile and the personal grid.
- **Starter-pack selection.** Admins can pick which starter-pack tools to publish through a checkbox list instead of importing the whole pack at once.

## Cost centers (org-chart spend attribution)

Teams answer access, not accounting, and a person can be in several, so they are a poor single key for "are marketing people spending more than engineering people, including personal AI use?". Cost centers add a single-valued accounting dimension, separate from RBAC, that never grants or restricts access.

- **Model.** A new org-scoped `CostCenter` entity plus a nullable `costCenterId` on org membership, team, and project. One cost center holds many people, teams, and projects, so a person's personal use and the autonomous agent their team runs can roll up to the same place.
- **Attribution.** A pure resolver picks one cost center per trace by precedence: the principal user's own cost center, then their team's (inherited default), then the project's, else Unassigned. A trace with no principal user (an autonomous agent) attributes by its project.
- **Service and API.** A `costCenters` tRPC router with create, rename, archive, and assign-user/team/project, reads gated on `governance:view` and writes on `governance:manage`, backed by a repository and service that enforce org-scoping.
- **Admin page.** `/settings/governance/cost-centers` lets an admin create and name cost centers and assign people, teams, and projects, wired to the real backend with no mocks.
- **Bird's-eye fix (the empty-graphs report).** The dashboard previously read only the hidden governance ingestion project, so an org with real traffic but no ingestion source showed empty graphs. `activityMonitor.spendByCostCenter` now aggregates spend across every project in the org and groups it by the resolved cost center, surfaced as a "Spend by cost center" card on the governance home. Every underlying ClickHouse query filters by `TenantId` first, so the rollup stays tenant-isolated.

## CLI

- **Governance-wrapper e2e.** The config loader correctly drops a stored personal VK secret that predates the `vk-lw-` format, but the e2e fixture still used the legacy `lw_vk_` prefix, so the wrapper preflight exited before spawning the tool. Fixture updated to the canonical format; implementation unchanged.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test:governance-e2e` green
- [x] otel.logs converter unit tests cover the cost lift
- [x] VK scope-RBAC integration and unit tests, feature parity green
- [ ] VK scope-RBAC integration suite verified on CI (local run blocked by a full disk on the dev machine)
- [x] cost-center attribution unit tests + entity-service integration tests, feature parity 12/12
- [x] `pnpm typecheck` clean after the cost-center and bird-eye changes
- [ ] `spendByCostCenter` ClickHouse integration suite verified on CI (local run blocked by Docker not booting on the dev machine)

## Screenshots (cost centers)

Dogfooded live against a real org (117 traces) via a credentialed admin session.

**Cost-centers admin page** — create cost centers, assign people, teams, and projects:

Cost centers admin page

**Governance home with the "Spend by cost center" card:**

Spend by cost center on the governance home


sergioestebance:
Post-merge review pass on the merged commit (`7dcaa895e`). Verified the previously-marked Fixed items, confirmed the still-Open items, then looked for things the first two passes didn't surface. Three new findings, all in code paths the earlier passes didn't touch.

## Verification of previously-claimed fixes

Spot-checked the 9 P1s in the follow-up table — all hold up at the merged SHA.

- `auth-cli.ts:1419` — `project_api_key` now requires `hasProjectPermission(..., "project:update")` before returning `Project.apiKey`. ✅
- `auth-cli.ts:1029` — CLI governance reads gated by `ensureGovernancePermissionOr403` against the matching web permission. ✅
- `modelProvider.service.ts:121` — `updateAdvancedSettings` now loads scopes and runs `canReadAnyScope` + `assertCanManageAllScopes` before update. ✅
- `20260524120000_collapse_vk_binding_orgscope/migration.sql:79` — transient `UNIQUE(organizationId, name)` removed; the rationale comment matches the diff. ✅
- `20260524130000_governance_truncate_for_refactor_dogfood/migration.sql:60` — destructive truncate is now gated behind `langwatch.allow_governance_truncate` GUC. ✅
- `routingPolicy.service.ts:325` — `resolveDefaultForUser` adds `orderBy: [{updatedAt: "desc"}, {id: "asc"}]` so a concurrent-write split is deterministic on read. ✅ (no DB-backed invariant, but the documented eventual-consistency is acceptable.)
- `governance app.ts:436/522/602/670/722` — admin template writes now stack `requireAiToolsManage` + `requireUserBoundCaller` so a legacy project key is rejected with 403 / `user_token_required`. ✅
- `governance-tools.ts:152` — `admin_list` now goes through strict `requirePermission` (user required), not `requireRead`. ✅
- `ingestionRoutes.ts:596` — `/v1/logs` calls `stampLogOriginAttrs` before handoff; `buildOriginAttrs` includes `langwatch.governance.retention_class`. ✅

## Open from prior passes (still open at merge)

- `virtualKeys.list/create/update/rotate/revoke` (`virtualKeys.ts:128/149/200/227/240`) still gate on broad `virtualKeys:view`/`virtualKeys:manage` org-level checks; no per-scope authorization yet. Matches the rogeriochaves follow-up.
- Go gateway wire still drops `policy_rules.models` (`config_wire.go:204` `buildPolicyRules` walks Tools/MCP/URLs only), `rate_limits.tpm` (`config_wire.go:134` copies RPM/RPD only), and provider `base_url`/`region`/`config` (`config_wire.go:24` unmarshals only id/type/credentials/deployment_map). Matches the follow-up.
- `cliSessionInventory.service.ts:98` `listForUser` still reads the user-wide Redis index and never filters by `record.organization_id`, despite the router taking `organizationId`. Same for `revokeSession`.
- `IngestionTemplate_organizationId_slug_key` (`20260508130000_add_ingestion_templates_and_bindings/migration.sql:54`) is still a single composite unique with nullable `organizationId`, so multiple `(NULL, slug)` rows are allowed — `platformIngestionTemplates.seeds.ts:101` continues to race across processes.

## New findings (not covered in prior passes)

1. **P1 — anomaly webhook dispatcher has no SSRF protection**

   `langwatch/ee/governance/services/activity-monitor/anomalyAlertDispatcher.service.ts:156`

   `dispatchWebhook` calls `this.fetchImpl(destination.url, …)` directly. The destination schema (`destinationConfig.schema.ts:39`) only validates "absolute http(s) URL"; it accepts `http://`, raw IPs, and private/link-local hosts. An `anomalyRules:manage` admin can configure a webhook pointing at `http://169.254.169.254/...` (cloud-IMDS), `http://127.0.0.1:5563/...` (internal gateway), or any in-cluster service, and the control plane will POST to it with HMAC-signed body. The codebase already ships `langwatch/src/utils/ssrfProtection.ts` (`ssrfSafeFetch`); the puller HTTP adapter uses it (`httpPollingPullerAdapter.ts:190`), the scenarios HTTP agent uses it — only the alert dispatcher missed it.

   Minimal fix: swap `defaultFetch` for `ssrfSafeFetch`, restrict the schema to `https://`, and resolve+pin the destination IP before the fetch so DNS rebinding can't slip through.

2. **P1 — pull-mode IngestionSource credentials stored as plaintext JSONB**

   `langwatch/prisma/schema.prisma:2021` (`parserConfig Json`) ← writes from `ingestionSource.service.ts:265/337`; reads from `pullerWorker.ts:127`, `s3PollingPullerAdapter.ts:179`, `httpPollingPullerAdapter.ts:261`

   `parserConfig.credentials` carries the live upstream secret (AWS `aws_access_key_id`/`aws_secret_access_key`/`aws_session_token` for `s3_polling`; `token` Bearer for `http_polling`; Anthropic workspace API key for `claude_compliance`; Microsoft client secret for `copilot_studio`). There is no encryption helper invoked on write and no decrypt on read — the puller hands the value straight to the upstream SDK. The schema doc-comments use names like `clientSecretHash` and `workspaceApiKeyHash` (which would only work for verifying inbound webhooks), so the storage shape is also misdescribed by its own schema. The `UserIngestionBinding.encryptedCredential` column has the same problem in miniature: the install path accepts `z.unknown()` and writes it verbatim (`userIngestionBinding.service.ts:170`), the name implies encryption that isn't happening.

   Minimal fix: encrypt the credential subtree before write (a workspace KMS / app-key envelope), decrypt at puller dispatch only, and rename the columns/fields so they no longer claim to be hashes. Add a migration to re-encrypt anything already landed in dogfood / S0 envs.

3. **P2 — `governance_ocsf_events` SIEM export drops rows on ms-collision**

   `langwatch/ee/governance/services/governanceOcsfExport.service.ts:133` + `:177`

   The cursor is a scalar `sinceMs` (EventTime). The query uses `EventTime > {sinceMs}` and `ORDER BY EventTime ASC, EventId ASC LIMIT N`. `nextCursor = lastEvent.eventTimeMs`. If a page boundary lands inside a batch of events that share a millisecond, the next pull's `>` filter skips every other event at that same ms. The table even has the right secondary key in `ORDER BY (TenantId, EventId)` and `EventId ASC` is already in the order-by — the bug is purely in the cursor shape.

   Minimal fix: compound cursor `(eventTimeMs, eventId)`, predicate `(EventTime, EventId) > (sinceMs, sinceEventId)` (ClickHouse supports tuple comparison), `nextCursor = { eventTimeMs, eventId }`. Update `GovernanceOcsfExportPage` accordingly. Important for SIEM completeness — silent gap, no error surfaced.

Residual notes: PR is already merged (`7dcaa895e`), so the above are follow-ups, not blockers. The three new items are independent of each other and could each ship as a small targeted PR. Items 1 + 2 affect SaaS posture more than single-tenant dogfood.

langwatch-agent:
1 P1 finding

- P1 `langwatch/src/server/api/routers/virtualKeys.ts:175`: the new write-path gate authorizes only `input.scopes`, then creates the virtual key under the independent `input.organizationId` at line 192. Nothing here or in `VirtualKeyService.create` verifies that each requested ORGANIZATION/TEAM/PROJECT scope belongs to that same organization. A caller with `virtualKeys:manage` in org A can submit `organizationId` for org B with a scope from org A; `assertCanManageAllScopes` passes, and the service writes a VK row into org B with a cross-org scope set. The same mismatch can be reintroduced on update via lines 232-236. Minimal fix: before create/update, resolve every submitted scope to its owning organization and reject unless it equals `input.organizationId` (for ORGANIZATION scopes, require `scopeId === organizationId`; for TEAM/PROJECT scopes, load the team/project org), and add a regression test for mismatched `organizationId` plus scope.

Checks: `pnpm vitest run src/server/gateway/__tests__/virtualKey.authz.unit.test.ts --reporter verbose` passed (7 tests). I did not run the new Postgres/testcontainers integration suite in this review pass.

rogeriochaves:
Review: found 2 issues.

1. `langwatch/src/server/api/routers/virtualKeys.ts:259-269` only validates `input.config?.guardrailAttachments`. If an admin moves a virtual key to a different project by changing `scopes` and leaves `config` untouched, the existing guardrail attachments are never re-checked against the new project. That leaves the VK in an invalid cross-project state, with attachments still pointing at the old project’s guardrails. Please revalidate the effective post-update attachments (or block scope changes until they are rechecked).

2. `langwatch/ee/governance/services/aiToolEntry.service.ts:801-835` ignores the `userId` parameter entirely and scans every team/project in the org when computing provider availability. The `/me` preflight in `langwatch/ee/governance/routers/aiTools.ts:88-97` will therefore report a provider as configured whenever *any* team in the org has it, even if the current user cannot reach that team/project. That makes the tile green-light providers the caller may not actually be able to use. Scope the query to the caller’s memberships before deriving `configuredProviders`.

rogeriochaves:
Addressing these now.

**Issue 2 (aiToolEntry.service.ts provider availability)** — fixed in 19a7c491e. `listConfiguredProvidersForUser` was ignoring `userId` and scanning every team in the org; it now resolves team/project scopes from the caller's `TeamUser` memberships, so a provider configured only on a team the caller can't reach no longer green-lights the tile. ORGANIZATION-scoped providers still count for every member. Added a regression test in `aiTools.integration.test.ts` (`providerAvailability` describe) proving a foreign-team-scoped provider does not leak into availability for a member or an org-only user.

**Issue 1 (virtualKeys.ts guardrail attachments stale on project-move)** — taking it next, together with the cross-org scope-binding P1 and the CodeRabbit NOT_FOUND-ordering finding on the same mutations. Will push + reply per-item.

rogeriochaves:
Pushed the two virtualKeys.ts findings in 63f9069fd.

**P1 (cross-org scope binding, virtualKeys.ts create/update)** — fixed. Added `assertScopesBelongToOrg`, wired into both `create` and `update` (the latter only when `scopes` is supplied). `assertCanManageAllScopes` proved the caller controlled each scope but never tied the scope to `input.organizationId`; now ORGANIZATION scopes must equal the org, and TEAM/PROJECT scopes must resolve to a team/project in it, else `BAD_REQUEST scope_org_mismatch`. Regression tests in `virtualKeys.scopeRbac.integration.test.ts` cover a team-scope and an org-scope mismatch.

**Issue 1 (stale guardrail attachments on project-move)** — fixed. `update` now revalidates the *effective* post-update attachments: when `config` isn't re-sent, it falls back to the existing stored attachments (`parseVirtualKeyConfig(existing.config)`) and rechecks them against the resolved project, so moving a VK to a new project with a stale cross-project attachment fails with `guardrail_project_mismatch`. Regression test added in `virtualKeys.guardrailAttach.integration.test.ts`.

**CodeRabbit (NOT_FOUND before scope gate)** — left as-is. CodeRabbit already auto-marked it addressed, and the `get` path is the one that normalizes visibility to NOT_FOUND. For update/rotate/revoke the `getById` is org-scoped, and the existing scope-RBAC tests deliberately pin NOT_FOUND for an out-of-org key vs FORBIDDEN for an in-org no-permission caller; normalizing both to one code would break that established contract for a low-severity within-org enumeration signal. Happy to revisit if you'd prefer the stricter normalization.

These ride CI (Docker testcontainers won't boot locally here, same as the rest of this PR's integration legs).

langwatch-agent:
1 P2

- P2 `langwatch/ee/governance/services/cost-center/costCenter.service.ts:110-146` and `langwatch/ee/governance/routers/costCenters.ts:82-130`
  `assignUser` / `assignTeam` / `assignProject` ignore the `updateMany` result and the router returns `{ ok: true }` unconditionally. If the target user/team/project id does not exist in the org, the mutation silently no-ops but still reports success, so the UI will believe the assignment changed when it did not. Mirror `rename` / `archive` here: check `count === 0` and surface `NOT_FOUND` (or an equivalent validation error) before returning success.

rogeriochaves:
**P1 (pull-mode credentials stored as plaintext)** — fixed in 9aaa7c201.

- New `ingestionCredentials.ts` helper wraps the `credentials` subtree in the shared AES-256-GCM envelope (`~/utils/encryption`), tagged `enc:v1:`.
- Encrypt on every write: `ingestionSource.service.ts` create / update / rotateSecret, and `userIngestionBinding.service.ts` install (both upsert paths).
- Decrypt only at the single puller dispatch chokepoint (`pullerWorker.ts`), so the s3/http/compliance adapters are untouched and never see ciphertext.
- Reads tolerate a legacy plaintext object, so the change is safe to deploy before the backfill runs; the `enc:v1:` tag makes encrypt idempotent.
- Backfill task for already-landed dogfood/S0 rows: `pnpm task encryptIngestionCredentials` (idempotent).
- Unit test covers the envelope round-trip, idempotency, legacy passthrough, and the no-credentials no-op (9/9 green locally).

Remaining nit from your finding: the schema doc-comments that call these `clientSecretHash` / `workspaceApiKeyHash` still misdescribe the storage as hashed. Those are doc-comments, not column names — I'll correct the wording in a follow-up commit so the rename doesn't get tangled with the encryption diff.

rogeriochaves:
1 P1

- `langwatch/src/tasks/encryptIngestionCredentials.ts:45-52` only re-encrypts `UserIngestionBinding.encryptedCredential` when the value is an object. That leaves any legacy string-valued credentials in plaintext, even though this backfill is supposed to cover all pre-encryption rows. The schema and route both allow opaque credential values, so string API keys / agent ids are realistic here.
  Minimal fix: treat every non-null, non-prefixed credential as legacy plaintext and wrap it with `encryptCredential(...)`, instead of keying off `typeof cred === "object"`.

langwatch-agent:
No blockers.

I re-read the existing thread and checked the new head. This commit only changes sidebar copy/presentation for the GOVERN section, and I did not find a functional regression. Residual risk is limited to the product-labeling decision itself (whether AI Gateway should remain marked as beta), not behavior.

PR #4240, opened 2026-05-28
docs(adr): autoscale event-sourcing workers on queue depth

rogeriochaves:
## What

ADR-021 proposing that `langwatch-workers` autoscale on GroupQueue depth via KEDA, instead of the current CPU-only HPA.

## Why

The worker pool autoscales on CPU at a 70 percent target, but the workers are Redis-bound: they spend most of their time blocked on the single-threaded Redis that backs the GroupQueue, so per-pod CPU stays around 25 to 30 percent even when thousands of jobs are pending. The CPU target is never reached under a deep backlog, so the HPA stays pinned at its minimum and capacity cannot grow under the exact pressure where it would help.

This was observed directly during a backlog incident: 5000-plus pending jobs at roughly 27 percent CPU, HPA frozen at min replicas.

## Decision (Proposed)

Scale on queue depth using KEDA's Redis scaler, which reads the queue length directly from Redis. KEDA is preferred over a Prometheus-adapter custom metric because the signal lives in Redis and KEDA reads it at the source, avoiding app-side metric instrumentation and a metrics-adapter deployment. The cluster has Prometheus but it is not wired into the custom-metrics API, and KEDA is not installed today.

## Scope and sequencing

- This is a decision document. The implementation (installing KEDA and the `ScaledObject`) lands in the infrastructure repository after the ADR is accepted.
- The maximum replica count stays a deliberate ceiling tied to single-threaded Redis dispatch capacity, not an open-ended function of backlog.
- This lever only becomes binding after work-conserving fair dispatch removes the fixed per-tenant cap (`specs/event-sourcing/work-conserving-fair-dispatch.feature`). While a fixed cap is in force, idle slots can coexist with a deep backlog, so pod count is not the limiter yet.

## Notes

Doc-only change. No code or runtime behavior changes in this PR.

### DIFF
--- a/dev/docs/adr/021-worker-autoscaling-on-queue-depth.md
+++ b/dev/docs/adr/021-worker-autoscaling-on-queue-depth.md
@@ -1,0 +1,102 @@
+# ADR-021: Autoscale event-sourcing workers on queue depth
+
+**Date:** 2026-05-28
+
+**Status:** Proposed
+
+## Context
+
+The `langwatch-workers` deployment runs the event-sourcing dispatch and job
+processing loops. Each pod runs up to `GLOBAL_QUEUE_CONCURRENCY` (default 100)
+concurrent jobs, and the binding throughput limiter for dispatch is the
+single-threaded Redis instance that backs the GroupQueue, not pod CPU.
+
+The deployment autoscales on a single signal: CPU utilization at a 70 percent
+target (HorizontalPodAutoscaler, min 4, max 20). That signal does not track the
+work the pool actually has to do. Workers spend most of their time blocked on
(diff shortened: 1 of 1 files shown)


rogeriochaves:
Closing: the KEDA queue-depth autoscaler is more moving parts than warranted right now. For the immediate capacity need we are using the simple lever instead, raising the worker HPA min_replicas floor (langwatch-saas#523), since the CPU-based HPA never scales above its floor under a Redis-bound backlog anyway. Queue-depth-driven autoscaling stays a possible future step (the worker tf already notes a Prometheus-adapter path), not a new cluster component today.

PR #4241, opened 2026-05-28
docs(adr): trigger outbox dispatch architecture (021-027)

0xdeafcafe:
## Summary

Records the locked-in design decisions for the trigger dispatch refactor that emerged from the #3351 review process. **Design only — no implementation in this PR.** Implementation follows in phased PRs (starting with the Phase 0 schema + service skeleton).

## ADRs

| # | Decision |
|---|---|
| **021** | Transactional outbox for stake-sensitive reactor dispatch. New \`ReactorOutbox\` PG table + \`.withOutbox\` primitive. Rejects event-sourced dispatch (replay danger + head-of-line blocking). |
| **022** | Two-tier dedup. \`TriggerSent\` stays as match-claim (cross-reactor race winner); \`ReactorOutbox\` is per-reactor dispatch state. Row-per-match over per-window upsert (avoids PG hotspot, replay-safe). |
| **023** | GroupQueue wakeup pattern. Use existing \`GroupQueue\` infrastructure with wakeup-only payloads + \`extend: false\` dedup. PG is truth; queue schedules. Rejects polling drainer and BullMQ reintroduction. |
| **024** | \`.withOutbox\` as pipeline-builder primitive. Sibling to \`.withReactor\`; forces auditable-vs-best-effort choice at registration time. Framework code under \`event-sourcing/outbox/\`. |
| **025** | Notify vs persistent action classification (\`NOTIFY_TRIGGER_ACTIONS\` / \`PERSIST_TRIGGER_ACTIONS\` constants) + per-trigger \`notificationCadence\` enum. Persist actions always immediate. |
| **026** | Liquid templates (\`liquidjs\`, \`strictVariables: false\`, 500ms timeout) for user-customizable notifications. Markdown email body, Slack type-discriminator + Block Kit allowlist. Fall-back-to-default on render failure. |
| **027** | Typed \`DispatchError\` contract. Dispatch endpoints throw \`DispatchError({ retryable })\` so the outbox worker can branch on retryable vs dead. Refactor of \`sendSlackWebhook\` / \`sendTriggerEmail\` is a prerequisite to outbox migration. |

## Reading order

021 → 022 → 023 → 024 → 025 → 026 → 027. Each ADR cross-links to the others it depends on. ADR-021 is the meta-decision; the rest fill in the structural details.

## Process notes

The plan that produced these ADRs was pressure-tested by an architect agent before being locked in (uncovered four sev-1 issues that are now reflected in the decisions: two-phase status with leasing, \`TriggerSent\` vs \`ReactorOutbox\` role separation, row-per-match shape, \`DispatchError\` contract). The phasing chosen for implementation is documented in conversation but not in the ADRs themselves — ADRs cover the durable decisions, not the rollout sequence.

## Out of scope (deliberate)

- README backfill of stale entries (019, 020 exist in \`adr/\` but aren't in the index). Pre-existing inconsistency; leaving for a follow-up since it's unrelated to this design.
- Implementation. Phase 0 (Prisma migration + outbox service skeleton) ships as a separate PR once these are reviewed.

## Test plan

- [x] Documentation only — no runtime code or tests changed.
- [ ] Reviewer signoff on the design decisions before Phase 0 lands.


langwatch-agent:
2 P2 findings

- P2 `dev/docs/adr/023-groupqueue-wakeup-pattern-for-outbox.md:12`: the ADR requires “at-most-once dispatch per trigger window”, but ADR-027 later accepts Slack double-send risk when dispatch succeeds and the status update fails (`dev/docs/adr/027-typed-dispatcherror-contract.md:101`). Those two contracts conflict: the proposed queue/outbox design can prevent concurrent workers from claiming the same rows, but it cannot guarantee exactly/at-most-once external side effects across crash/retry boundaries unless every endpoint is idempotent. Minimal fix: change ADR-023 to state the real guarantee, e.g. “single leased dispatcher per trigger window, at-least-once external delivery with idempotency where provider support exists,” and make the worker/status terminology match ADR-027.

- P2 `dev/docs/adr/022-two-tier-dedup-triggersent-reactor-outbox.md:30`: the decision says `TriggerSent` remains unchanged as the match-claim ledger, with row-per-match keyed by `(triggerId, traceId)`, but the current schema explicitly allows `traceId` to be null for custom graph alerts and stores `customGraphId` instead (`langwatch/prisma/schema.prisma:926`). In Postgres, the existing `@@unique([triggerId, traceId])` does not dedupe nullable `traceId` rows the same way, and the ADR never states whether custom graph trigger dispatch is excluded from the outbox migration or needs its own claim/dedup key. Minimal fix: either scope the ADRs to trace/evaluation triggers only, or specify the custom-graph claim key/constraint (for example `(triggerId, customGraphId)` / incident-window semantics) before Phase 0 uses this as the schema contract.

Checks: `git diff --check origin/main...HEAD` passed. I did not run runtime tests because this PR is documentation-only.

sergioestebance:
Did a deep read against our prior research on notification/alert systems. Substrate is right — the framework split is load-bearing and lands well. Below are forward-looking notes on where the data-model bets in 022/025/026 will pull when triggers expand beyond today's shape (one user, one channel, one trace). Not blocking; flagging so ADR "Consequences" sections can absorb them.

**Positives first**

- `.withReactor` vs `.withOutbox`, typed `DispatchError(retryable)`, GroupQueue wakeup payloads, Idempotency-Key on outbox row id, framework code under `event-sourcing/outbox/` — right primitives regardless of how the trigger data model evolves on top. NOTIFY, PERSIST, future aggregate triggers, future fan-out — all sit on this substrate unchanged.
- ADR-026's safety hardening (sandboxed Liquid, 500ms timeout, `p-limit` 10, fall-back-to-default, Block Kit interactive-block strip, backend-injected test-fire banner) is a complete spec. Worth keeping under `event-sourcing/outbox/templating/` so reuse is structural when/if templates eventually move off `Trigger` to a policy row.

**Forward-looking notes**

1. **Rule vs policy — worth naming the shape now.** `Trigger` today bundles "what to react to" with "what to do about it" — filters, action, channel, cadence, templates, all in one row. Works for current usage. Doesn't scale when the same match wants two destinations, or when the same Slack webhook serves five triggers. Splitting match-definition from delivery-configuration lets cadence (025) and templates (026) live against a `NotificationPolicy` where they belong. This PR doesn't have to do the split, but the new columns could land on a new table rather than stacking on `Trigger` — same migration cost now, much cheaper later.

2. **Multi-destination is coming for the notify side.** ADR-025 already classifies actions into NOTIFY (Slack, email) and PERSIST (dataset, annotation queue). NOTIFY will eventually want one trigger → multiple destinations with potentially different cadences each. Today's workaround is duplicating the trigger. PERSIST is 1-destination by design and doesn't need this. Fan-out is notify-specific — worth designing reusable channels for the notify path now, leaving persist as-is.

3. **Dedup grain only fits trace-path triggers; custom graph alerts already don't fit.** ADR-022 keys outbox dedup on `triggerId + traceId`. Fine for trace-path triggers — trace id is the natural match identity. But we already have custom graph alerts (Trigger rows with `customGraphId` set, on `/api/cron/triggers`, fire when a chart metric crosses a threshold over a window). For those, `TriggerSent.traceId` is null and the `(triggerId, traceId)` unique constraint already doesn't dedup — the code falls back to an explicit "unresolved row" query. The PR transfers the same gap forward. When custom graph alerts migrate onto the outbox path (or any aggregate-driven trigger lands), the natural grain is `triggerId + (groupBy labels + window)`. Worth noting in 022 that the per-trace key is trace-path-only.

4. **Silencing is a notify-side ask, eventually.** Users running notify triggers regularly want temporary mute: "I know prod is degraded, stop pinging me for an hour." NOTIFY-only — you don't silence a dataset write, the user asked for the data. Current `match → cadence batch → send` has no hook between match and dispatch for a silence check. Not in scope here, but worth a one-liner in 021 that silencing has no integration point yet so the next person doesn't assume the outbox is the place.

5. **Replay context flag — small now, expensive later.** ADR-021 rejects event-sourced dispatch because "replay would re-fire emails." Right call. But `.withOutbox`'s match phase still runs during replay, and with 30/90-day row retention, older replays insert fresh outbox rows and re-fire. Standard fix is an `isReplay` flag on the reactor context — match returns null when replaying. Cheap to add to 024's signature now, painful to retrofit once `.withOutbox` reactors are in production.

6. **Cadence and deadlines are different primitives.** ADR-023's GroupQueue wakeup batches matches into a digest window — "wait 5 min, send everything that accumulated." That's a cadence. Some future trigger semantics need a deadline instead — "match observed, wait 5 min, re-check, react only if still true" (regression-flag triggers where we want to confirm the score didn't recover). Different shape: a deadline re-reads state when it fires; a cadence just flushes. Worth a note in 023 that the deadline primitive is out of scope so it isn't conflated with `extend: false` wakeups later.

7. **Customer-configurable destinations will need framework-level SSRF + HMAC.** ADR-024's `dispatch` handler is endpoint-agnostic. Slack/email today are safe — outbound targets are fixed providers. The moment a customer-supplied webhook URL lands, we need SSRF blocking (private IPs, link-local, cloud metadata endpoints), HMAC signing, payload size caps, per-destination secret encryption. Worth flagging in 024 that future dispatch endpoints share an SSRF/signing utility instead of each reinventing it — Helicone got this right, most platforms got it wrong.

None of these block the PR. The framework substrate is correct and reusable; the items above are about scope clarity and migration cost when triggers expand past today's shape.

0xdeafcafe:
Closing in favor of the re-sliced 3-PR stack: #4498 (server dispatch) → #4499 (Liquid templating) → #4500 (staged authoring drawer). The ten ADRs (021-030) here are collapsed into five (022-026) in the new stack; the eight specs into five. See #4498 description for the full mapping.

PR #4277, opened 2026-05-28
fix(governance): remove per-origin retention wrongly bundled in governance PR

rogeriochaves:
## Summary

Per-origin retention (per-class TTL on `stored_spans` / `stored_log_records`, keyed off a `RetentionClass` column) was bundled into the AI-governance PR, but it belongs to a separate workstream. The `MODIFY TTL` it issues triggers a full `MATERIALIZE TTL` rewrite of `stored_spans`; on a large table that runs as a blocking startup migration and wedged a production rollout. This PR removes the feature entirely so it can be reintroduced later as its own migration designed for large tables.

## What changed

- Delete CH migration `00026_add_retention_class`; renumber the governance migrations to `00026`-`00031` (no version is recorded anywhere yet, so renumbering is safe).
- Stop writing the `RetentionClass` column: span/log storage repositories, trace-attribute accumulation, and the ingest origin-attribute stamping.
- Strip the per-class retention branch from `ttlReconciler`; the cold-storage `MOVE` logic is preserved (unit tests green).
- Remove `IngestionSource.retentionClass`: Prisma field plus a drop-column migration, the tRPC router, the service, and the settings-composer dropdown.
- Remove retention specs/scenarios and dogfood seeds.

## Why no-op vs delete

Deleting and renumbering (rather than a no-op stub) keeps the migration sequence contiguous. Self-hosted installs have not received this in a release yet, so no environment has applied migrations 26+; fresh installs apply the renumbered governance migrations cleanly and never run the retention `MODIFY TTL`.

## Prod / dev note

SaaS production was manually reverted during the incident (TTL + column removed, materialize mutation killed, no data lost). The shared dev ClickHouse is being cleaned up manually.

## Test plan

- [ ] `ttlReconciler` unit tests pass (cold-storage path intact)
- [ ] typecheck clean
- [ ] ingest integration tests pass without the retention assertions
- [ ] fresh migrate from v25 applies `00026`-`00031` with no retention TTL on `stored_spans`


sergioestebance:
**Findings**

P1 `langwatch/prisma/migrations/20260528000000_drop_ingestion_source_retention_class/migration.sql:1`

The migration uses `DROP COLUMN "retentionClass"` without `IF EXISTS`. The PR body says SaaS production was manually reverted and the column was already removed, so this migration will fail with “column does not exist” in that environment and can block rollout again. Minimal fix: make it idempotent:

```sql
ALTER TABLE "IngestionSource" DROP COLUMN IF EXISTS "retentionClass";
```

P3 `specs/ai-gateway/governance/compliance-baseline.feature:23`, `specs/ai-gateway/governance/architecture-invariants.feature:30`, `specs/ai-gateway/governance/event-log-durability.feature:42`, `specs/ai-gateway/governance/ui-contract.feature:8`

The PR deletes `retention.feature` and removes the implementation, but several BDD specs still describe per-origin retention class as part of the shipped contract. That leaves acceptance criteria pointing at removed behavior. Minimal fix: scrub or reword the remaining retention-class references in the governance specs, plus the Claude Code OTLP spec reference at `specs/ai-governance/ingestion-sources/claude-code-otlp.feature:237`.

**Notes**

I did not find remaining runtime code references to `retentionClass`, `RetentionClass`, or `langwatch.governance.retention_class` outside the new drop migration. `git diff --check` is clean. I did not run the full test suite.

rogeriochaves:
Thanks @sergioestebance, both addressed in 09da0da56.

**P1 (drop migration idempotency):** good catch. Changed it to `ALTER TABLE "IngestionSource" DROP COLUMN IF EXISTS "retentionClass";` so it can't re-block a rollout in an environment where the column is already gone. (Note: the manual revert during the incident was on the ClickHouse side, `stored_spans`/`stored_log_records`; the Postgres `IngestionSource.retentionClass` column was untouched there, but `IF EXISTS` is the right idempotent default regardless.)

**P3 (residual retention references in specs):** scrubbed across the governance specs: removed the `retention.feature` / `per-origin-retention.feature` companion and owner-list entries (folds, receiver-shapes, event-log-durability, architecture-invariants, siem-export, compliance-baseline, ui-contract), reworded the prose and the compliance mapping rows to drop the retention tokens while keeping RBAC/PII/attribution, reworded the `event-log-durability` data-removal step to "append-only, no automatic data-removal path", and reworded the `claude-code-otlp` PII scenario to stand on its own (it was really about `OtlpSpanPiiRedactionService`; retention class was incidental). Unrelated retention (data-retention plan tiers, event-byte GC, chrome-retention) left untouched. A grep for `retention.feature|per-origin retention|retention class|retention_class` across the governance specs is now clean.

Also fixed the failing `config.materialiser.integration.test.ts`, which was a pre-existing failure on main unrelated to this PR: `ModelProvider` moved to scope rows so the test's `organizationId` create arg is no longer valid, and the `Monitor` deleteMany teardown was missing `projectId` for the multitenancy guard.

langwatch-agent:
1 P1

- P1: `langwatch/src/server/clickhouse/ttlReconciler.ts:148-156,249-260` only compares the hot-storage day count and now skips when `currentDays === desiredDays`. That means existing `stored_spans` / `stored_log_records` tables with the legacy `RetentionClass` DELETE clauses are treated as already healthy and are never rewritten, so upgraded installs keep enforcing the old per-origin retention policy even after this PR removes it. Please add an explicit cleanup path for the legacy TTL clauses (or a one-time migration that rewrites the affected tables), not just the new desired expression.

langwatch-agent:
1 P1

- `langwatch/src/server/clickhouse/ttlReconciler.ts:252-260` now treats TTL as healthy as soon as the hot-storage day count matches. That skips any rewrite for upgraded installs that still carry the legacy `RetentionClass` DELETE clauses, so they keep enforcing the old per-origin retention policy even after this PR removes it. Minimal fix: keep detecting and stripping the legacy TTL clauses, or add a one-time migration that rewrites the affected tables.

rogeriochaves:
Addressed in f98c0f187.

The reconciler previously skipped a rewrite whenever the cold-storage day count already matched the desired value (`currentDays === desiredDays`), which meant a table still carrying the legacy `RetentionClass` DELETE clauses could be treated as healthy and never rewritten. Fixed by detecting the legacy clause and forcing a rewrite when present:

- `hasLegacyRetentionTTL(engineFull)` flags any TTL expression still referencing `RetentionClass`.
- `shouldRewriteTTL({ currentDays, desiredDays, engineFull })` now returns true when the legacy clause lingers, even if the day count matches, so the next reconcile issues a `MODIFY TTL` that replaces the whole expression with the clean MOVE-only one. It stays metadata-only (`materialize_ttl_after_modify = 0`), so it strips the DELETE clauses without the heavy table rewrite that caused the original incident.

Covered by unit tests in `ttlReconciler.unit.test.ts`, including the exact skip-trap case (day count matches but a `RetentionClass` DELETE clause is present -> rewrite required).

langwatch-agent:
No blockers.

Residual risk: the legacy TTL cleanup still keys off `engine_full` string matching for `RetentionClass`, so an integration test against the real ClickHouse engine metadata would be the last bit of de-risking. The new unit coverage does pin the skip-trap case.

PR #4287, opened 2026-05-29
feat(event-sourcing): work-conserving max-min fair dispatch

rogeriochaves:
## Summary

Replaces the static per-tenant concurrency cap with a dynamic, work-conserving max-min fair share. A tenant may use as much capacity as is free, and is throttled only when it is actually starving another tenant. There is no hardcoded percentage; the split emerges.

Design (settled in #ACME-demo, option C): reuse the existing incident-hardened scan-and-gate dispatch path (parkGroup / reconcileParked / orphan-recovery from #4228) and make the cap dynamic, a water-level `W` computed lazy-inline on the existing staleness-gated reconcile pass. No new dispatch structure, no mixed-fleet migration, and the soft selection summary we debated evaporates entirely (a structure that never exists cannot drift or leak).

- Idle: one tenant alone takes every slot (`W = G`), so a burst is processed in full.
- Contention: `W` is the max-min water level. Small tenants get their full demand, big tenants split the rest. 40/40/20-style allocations emerge, never hardcoded.
- Newcomer service is temporal (a newcomer has 0 in-flight, competes immediately as slots free), not spatial. No slots are ever held empty in reserve.
- Fail-protective: a missing or stale `W` falls back to the low static cap, never a permissive high value.

## What's in this PR

- `scripts.ts`: the dynamic water-level cap. `waterLevel` water-fill, `recomputeDynamicCap` on the 2s-gated single-pod reconcile pass, the enqueue-populated `demanding-tenants` recency ZSET, and the work-conserving override (`unparkLeastServedParked`). Gated behind `LANGWATCH_DISPATCH_GLOBAL_BUDGET` (0 = off, the default), so it ships inert and back-compatible.
- `specs/event-sourcing/work-conserving-fair-dispatch.feature`: the behavioral contract.
- `fairDispatch.integration.test.ts`: mechanism-agnostic testcontainer harness driving the stable `stage`/`dispatch`/`complete` API and asserting the per-tenant in-flight distribution from real `tenant_active_z` truth.

## Test plan

All tests run against real Redis on testcontainers (no mocks, actual Lua `EVAL`); the dispatch/cap logic is pure Redis. 10 passed / 2 todo.

- [x] Work-conserving: a lone tenant uses every slot (`W = G`)
- [x] Max-min: a small tenant gets its full demand, the big tenant absorbs the rest
- [x] Contention: two equal tenants converge to `G/2` each
- [x] Runaway clamped to its fair share; small co-tenant served promptly (via the override)
- [x] Phantom claimant: a bursted-then-idle tenant never caps a busy one (the override picks only parked tenants)
- [x] Newcomer ramps to its fair share over sustained operation, no slot held in reserve
- [x] Throttling released the moment contention ends
- [x] Intra-tenant priority order preserved
- [x] Mixed-fleet rollout: pre-upgrade work and mid-rollout enqueues all dispatch, nothing stranded (old static-cap pod and new dynamic pod on the same ready zset)
- [ ] Operator hard-clamp (explicit per-tenant ceiling): `it.todo`, intentionally not built. It is the off-by-default escape hatch the spec frames as a safety valve the fair model should never need; a follow-up if ops ever wants it.

Note: the optional long-eval reserve floor stays off by default pending a prod hold-time (p50/p95/p99) measurement (currently blocked on EKS access); it does not gate this PR.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
@@ -5,7 +5,7 @@ import {
   stopTestContainers,
   getTestRedisConnection,
 } from "../../../__tests__/integration/testContainers";
-import { GroupStagingScripts } from "../scripts";
+import { GroupStagingScripts, CLAIMANT_WINDOW_MS } from "../scripts";
 
 // Behavioral suite for specs/event-sourcing/work-conserving-fair-dispatch.feature.
 //
@@ -335,6 +335,47 @@ describe("work-conserving fair dispatch", () => {
 
         expect(first?.groupId).toBe(gid("solo", "early"));
         expect(second?.groupId).toBe(gid("solo", "late"));
+      });
+    });
+  });
+
+  // Regression for the dynamic-cap recompute (ebbd3eeea): a tenant that stops
+  // enqueuing for longer than the claimant window but still holds in-flight or
+  // parked work must stay a measured claimant - it must NOT be evicted from the
+  // water-fill by recency alone, or W inflates and a fresh co-tenant expands past
+  // its fair share, starving the quiet-but-working tenant.
+  describe("Rule: a still-working tenant is not evicted by going quiet", () => {
+    describe("when a tenant holds work but stops enqueuing past the claimant window", () => {
+      it("keeps its fair share while a fresh co-tenant does not take the whole fleet", async () => {
+        // quiet bursts and, alone, fills the fleet (W=G) with active + parked work
+        await stageForTenant("quiet", 50);
+        await fillToFleet(FLEET);
+        expect((await inFlightByTenant()).quiet).toBe(FLEET);
+
+        // age quiet's demand recency past the claimant window WITHOUT a new
+        // enqueue (it is still holding all that active + parked work)
+        await redis.zadd(
+          `${keyPrefix()}demanding-tenants`,
+          Date.now() - CLAIMANT_WINDOW_MS - 1000,
+          "quiet",
+        );
+
+        // a fresh co-tenant arrives and competes
+        await stageForTenant("fresh", 50);
+        // wait past the reconcile gate so the next dispatch recomputes W with both
+        await sleep(RECONCILE_GATE_MS);
+
+        for (let round = 0; round < 8; round++) {
+          await completeSome(5);
+          await fillToFleet(FLEET);
+        }
+
+        const inFlight = await inFlightByTenant();
+        // quiet was kept in the fill (its parked work still competes); it is not
+        // starved, and fresh is held to its fair share rather than the whole fleet
+        expect(inFlight.quiet ?? 0).toBeGreaterThanOrEqual(8);
+        expect(inFlight.fresh ?? 0).toBeLessThanOrEqual(12);
+        expect((inFlight.quiet ?? 0) + (inFlight.fresh ?? 0)).toBe(FLEET);
       });
     });
   });
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
@@ -107,18 +107,38 @@ beforeAll(async () => {
 
 const FLEET = 20;
 
+// Captured so afterEach restores the prior value instead of blindly deleting it,
+// in case the process started with a real budget set.
+let prevBudgetEnv: string | undefined;
+
+// Delete only this suite's namespaced keys rather than flushall(), so we never
+// wipe other integration suites sharing the test Redis container.
+async function flushSuiteKeys() {
+  let cursor = "0";
+  do {
+    const [next, keys] = await redis.scan(cursor, "MATCH", `${QUEUE_NAME}*`, "COUNT", 500);
+    cursor = next;
+    if (keys.length > 0) await redis.del(...keys);
+  } while (cursor !== "0");
+}
+
 beforeEach(async () => {
-  await redis.flushall();
+  await flushSuiteKeys();
   scripts = new GroupStagingScripts(redis, QUEUE_NAME);
   inflight = [];
   // Enable the dynamic water-level cap with the global budget = fleet size, so a
   // lone tenant gets W=G=20 and two contenders converge to G/2=10. readGlobalBudget
   // reads process.env per dispatch, so setting it here takes effect immediately.
+  prevBudgetEnv = process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET;
   process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET = String(FLEET);
 });
 
 afterEach(() => {
-  delete process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET;
+  if (prevBudgetEnv === undefined) {
+    delete process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET;
+  } else {
+    process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET = prevBudgetEnv;
+  }
 });
 
 afterAll(async () => {
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
@@ -1,0 +1,276 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import type { Redis } from "ioredis";
+import {
+  startTestContainers,
+  stopTestContainers,
+  getTestRedisConnection,
+} from "../../../__tests__/integration/testContainers";
+import { GroupStagingScripts } from "../scripts";
+
+// Behavioral suite for specs/event-sourcing/work-conserving-fair-dispatch.feature.
+//
+// These tests drive the STABLE public API (stage -> dispatch loop -> complete)
+// and assert on OUTCOMES: the per-tenant in-flight distribution read from the
+// authoritative tenant_active_z truth. They are deliberately mechanism-agnostic
+// so they hold under the chosen dispatch model (option C: the existing
+// scan-and-gate dispatch path with a dynamic water-level cap replacing the
+// static one). Fairness is produced INSIDE dispatch(); the signature does not
+// change, so these tests need no edit when the dynamic cap lands.
+//
+// The contention scenarios are it.skip until the dynamic water-level cap is
+// implemented: against today's static per-tenant cap they fail deterministically
+// (equal scores tie-break by member lex, so the lexically-smaller tenant prefix
+// monopolises the fleet). Un-skip each as the EVALSHA makes it pass - that is
+// the red-to-green contract. The work-conserving and priority guards below pass
+// today and protect against regressions.
+
+let redis: Redis;
+let scripts: GroupStagingScripts;
+const QUEUE_NAME = "{test/fair-dispatch}";
+const ACTIVE_TTL_SEC = 300;
+
+function keyPrefix() {
+  return `${QUEUE_NAME}:gq:`;
+}
+
+// groupId convention: "<tenantId>/<suffix>". The dispatch Lua derives the tenant
+// as the substring before the first slash (mirrors prod, e.g.
+// "project_FzsKx-.../fold/traceSummary/trace:...").
+function gid(tenantId: string, suffix: string) {
+  return `${tenantId}/${suffix}`;
+}
+
+function tenantOf(groupId: string) {
+  return groupId.slice(0, groupId.indexOf("/"));
+}
+
+async function stageForTenant(
+  tenantId: string,
+  count: number,
+  opts: { dispatchAfterMs?: number } = {},
+) {
+  for (let i = 0; i < count; i++) {
+    const uniq = crypto.randomUUID().slice(0, 8);
+    await scripts.stage({
+      stagedJobId: `${tenantId}-job-${i}-${uniq}`,
+      groupId: gid(tenantId, `g${i}-${uniq}`),
+      dispatchAfterMs: opts.dispatchAfterMs ?? 0,
+      dedupId: "",
+      dedupTtlMs: 0,
+      jobDataJson: JSON.stringify({ tenant: tenantId, i }),
+      shouldExtend: true,
+      shouldReplace: true,
+    });
+  }
+}
+
+// A fixed worker fleet is modelled by a tracked in-flight set: fillToFleet keeps
+// dispatching until the fleet is full or ready is empty; completeSome frees
+// slots (optionally for one tenant) so they can be refilled.
+type Slot = { groupId: string; stagedJobId: string };
+let inflight: Slot[] = [];
+
+async function fillToFleet(size: number) {
+  while (inflight.length < size) {
+    const r = await scripts.dispatch({ nowMs: Date.now(), activeTtlSec: ACTIVE_TTL_SEC });
+    if (!r) break;
+    inflight.push({ groupId: r.groupId, stagedJobId: r.stagedJobId });
+  }
+}
+
+async function completeSome(n: number, tenant?: string) {
+  const victims = inflight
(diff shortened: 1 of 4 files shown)


specs/event-sourcing/work-conserving-fair-dispatch.feature
```diff
@@ -0,0 +1,192 @@
...
+  #     that has not arrived (W=G when alone). A work-conserving override fills an
+  #     otherwise-idle slot from the LEAST-SERVED parked tenant, exceeding W (never
+  #     G — bounded by the pod's free slots) so fairness binds only under real
+  #     contention. The override's candidate set is the parked tenants only, so a
+  #     work-less phantom claim never wins a slot.
+  #   - Fail PROTECTIVE: the dynamic-cap key carries a TTL; a stalled recompute
+  #     lapses back to the static operator cap (the low side), never permissive.
+  #     The whole feature is gated behind LANGWATCH_DISPATCH_GLOBAL_BUDGET (0 =
+  #     off, the default), so it ships inert and back-compatible.
+  #   - Total queue depth = sum of per-tenant pending (ready + parked +
```
0xdeafcafe: this contradicts the impl — `recomputeDynamicCap` deliberately omits the ready count (`demand_i = in-flight + parked`, ready only enters via fresh-claimant presence). the full `ready + parked + in-flight` depth is the autoscaler's read, not the dispatcher's fairness signal. can we split the two so the spec doesn't claim the gate reads ready?

0xdeafcafe (commented):
overall this is great. traced the water-fill by hand and the parked-as-demand-proxy is the clever bit — i was half-expecting W to relax to G under saturation and bring back the lex-prefix starvation, but the parked backlog keeps `measured` above the share so it holds. inert-by-default + the TTL fallback to the static cap is the right call for the hot path.

nothing blocking. main thing is the spec has drifted from what actually shipped (a couple scenarios + an invariant still describe the design we dropped), notes inline.

also — can you update the PR body before merge? it says the cap impl "lands next on this branch" and lists the contention scenarios as `it.skip`/`it.todo`, but they're all live and green now. reads like only the spec + test are here.

specs/event-sourcing/work-conserving-fair-dispatch.feature
```diff
@@ -0,0 +1,192 @@
...
+
+  Rule: A tenant's own work keeps its priority order
+
+    Scenario: Within one tenant, higher-priority groups dispatch first
+      Given a tenant has several groups queued at different priorities
+      When that tenant is dispatched
+      Then its groups are taken in priority order
+      And fairness only reorders dispatch across tenants, never within one
+
+  Rule: A tenant leaves and rejoins the rotation cleanly
```
0xdeafcafe: vestigial vocab from the design we dropped — "rotation" implies round-robin but the gate admits in priority order, it doesn't select a tenant (your own invariant on line 50). a drained tenant ages out of `demanding-tenants`, a crashed one lapses out of `tenant_active_z`. can we reword off "rotation"?

specs/event-sourcing/work-conserving-fair-dispatch.feature
```diff
@@ -0,0 +1,192 @@
...
+
+    Scenario: A tenant drops out of dispatch the moment its work is exhausted
+      Given two tenants competing under saturation
+      When one tenant's waiting work is fully dispatched
+      Then it no longer takes a turn in the rotation
+      And the remaining tenant receives the freed capacity
+      And a tenant whose work is abandoned by a crash is swept from the rotation
+
+    Scenario: A drifted load summary self-heals and never stalls dispatch
```
0xdeafcafe: this one's from the design we dropped too — there's no "soft load summary" under option C. closest live thing is the `dynamic-cap` key going stale/lapsing and the reconcile rebuilding W from truth. either reword it to that (and add an `it.todo`, there's no test backing it) or just delete it.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -199,6 +209,128 @@ local function reconcileParked(readyKey, keyPrefix, tenantCap, nowMs)
...
+-- Recompute W from the tenants with demand and store it under dynamic-cap with a
+-- TTL (a stalled recompute lapses to the static cap = fail-protective). Reads the
+-- demanding-tenants RECENCY ZSET (member=tenant, score=last enqueue/dispatch ms),
+-- avoiding the keyspace SCAN that once pegged Redis. Per tenant: measured =
+-- in-flight (tenant_active_z, GC-d) + parked; fresh = active within the claimant
+-- window. fresh & measured 0 = presence claimant (newcomer still in ready);
+-- stale & measured 0 = drained, GC. Deliberately omits the exact ready COUNT (no
+-- per-tenant ready index); ready PRESENCE via fresh membership is all the fill
+-- needs to stop a newcomer reading 0 demand and pinning W at the full budget.
+local function recomputeDynamicCap(keyPrefix, budget, nowMs)
```
0xdeafcafe, line 311: how big do we expect `demanding-tenants` to get? this is one atomic eval every 2s doing ~3 redis calls per tenant (zremrangebyscore + zcard + parked zcard) on the single thread. fine at tens, but a busy queue with thousands of recently-active tenants blocks redis every reconcile. better than the old keyspace SCAN, just want a number we're comfortable with or a bound.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -575,184 +750,208 @@ local keyPrefix      = ARGV[1]
...
+
+local dispatched = scanBatch(tenantCap, false, 0)
+
+-- Work-conserving override: spare local slots remain (dispatched < maxJobs) but
+-- nothing more was admittable under the cap, while over-cap work sits parked.
+-- Fairness binds only under contention; spare capacity = no contention, so fill
+-- the idle slots from the least-served over-cap tenant rather than return short
+-- and idle the fleet (the static-cap idle-behind-cap sin). Bounded to the
+-- remaining slots; this pod's free-slot budget keeps the fleet total at G.
+if globalBudget > 0 and tenantCap > 0 and dispatched < maxJobs and redis.call("SCARD", keyPrefix .. "parked-tenants") > 0 then
```
0xdeafcafe, line 991: `unparkLeastServedParked` only unparks from one tenant per pass, so a single batch with a big `maxJobs` can hand all the freed slots to one over-cap tenant before the others get a turn. self-corrects across polls so probably okay, just flagging.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -26,6 +26,16 @@ export const GROUP_KEY_TTL_MS = 6 * 60 * 60 * 1000;
...
 
+/**
+ * TTL (ms) on the dynamic water-level cap key. The recompute runs on the same
+ * 2s-gated, single-pod reconcile pass and refreshes this key; the TTL is several
+ * intervals so a brief miss does not drop the cap. If the recompute stalls
+ * entirely (every pod dead or wedged) the key lapses and dispatch falls back to
+ * the static operator cap — fail PROTECTIVE (the low side), never permissive.
+ * Interpolated into the Lua so there is a single source of truth.
+ */
+export const DYNAMIC_CAP_TTL_MS = 10000;
```
0xdeafcafe: nit: this and `CLAIMANT_WINDOW_MS` are named consts but the 2s reconcile gate is a bare literal in both dispatch scripts, and the comments call these "several intervals" of it. worth a shared const so a retune doesn't silently break the relationship.

rogeriochaves: Fixed in f5a16f5bf: extracted `RECONCILE_INTERVAL_MS` (was the bare `2000` on the reconcile gate in both dispatch scripts) and redefined `DYNAMIC_CAP_TTL_MS = 5 * RECONCILE_INTERVAL_MS`, so retuning the cadence keeps the "several intervals" relationship intact in code rather than in a comment.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
rogeriochaves: Bounded in f5a16f5bf. The recompute now (1) bulk-trims aged members in a single `ZREMRANGEBYSCORE -inf (now - window)` (drops everything past the claimant window, replacing the per-member GC), then (2) processes at most `MAX_DEMANDING_TENANTS = 1000` most-recent claimants via `ZREVRANGE`. Worst case ~3N Redis calls per 2s pass (sub-ms at N=1000) no matter how many tenants pile into the registry. The N most-recent are the live contention set; a tenant aged out of the window still has its in-flight enforced directly via `tenant_active_z`, so the bound only approximates the fairness *fill*, never the hard cap.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
rogeriochaves: Confirmed the across-polls self-correction is enough, and there is a stronger reason it is safe: the override only fires when this pod has spare slots after the cap-gated pass (`dispatched < maxJobs`) = under-saturation = no contention. With spare capacity nobody is being starved; the dispatcher drains in a loop (`dispatchBatch` every ~25ms), and since `unparkLeastServedParked` picks the *least-served* tenant, the one just filled is no longer least-served, so the next pass rotates to the others. Worst-case skew is ~one poll, under conditions where there is idle capacity anyway. Under true saturation (`dispatched == maxJobs`) the override never fires and the per-tenant cap W holds. Leaving as-is, but happy to loop the pick across tenants in one pass if you want tighter intra-override balance — it would just add cost on a path that only runs when the fleet is not full.

specs/event-sourcing/work-conserving-fair-dispatch.feature
rogeriochaves: Fixed in 862b31517: split the two reads. The dispatcher fairness demand_i is now stated as in-flight + parked (a ready-only newcomer enters via fresh presence in demanding-tenants, the gate never reads ready); the full ready + parked + in-flight depth is called out as the separate autoscaler primitive. Good catch, that was a real contradiction.

specs/event-sourcing/work-conserving-fair-dispatch.feature
rogeriochaves: Fixed in 862b31517: reworded off rotation. The rule is now "A tenant leaves and rejoins contention cleanly", and the scenario says a drained tenant ages out of the demand set while a crashed one lapses out of the in-flight truth. No round-robin / selection vocab.

specs/event-sourcing/work-conserving-fair-dispatch.feature
rogeriochaves: Fixed in 862b31517 + 47a74541f: dropped the soft load summary scenario (dropped-design vocab) and replaced it with the shipped fail-safe, "A stale dynamic cap fails safe and is rebuilt from truth" (falls back to the static cap, the reconcile rebuilds W from in-flight + parked truth), plus a matching it.todo since there is no test yet.

rogeriochaves:
@coderabbitai review

All review feedback addressed: spec reworded off the dropped-design vocab (dispatcher demand vs autoscaler depth split, rotation language removed, soft-summary scenario replaced with the shipped stale-cap fail-safe, crash-abandonment split into its own scenario, newcomer-floor reworded to operator-facing behavior); scripts.ts bounded the recompute (RECONCILE_INTERVAL_MS const, MAX_DEMANDING_TENANTS cap + bulk trim). 142 passed / 4 todo on real Redis, zero regression.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
```diff
@@ -0,0 +1,384 @@
...
+
+beforeAll(async () => {
+  await startTestContainers();
+  redis = getTestRedisConnection()!;
+});
+
+const FLEET = 20;
+
+beforeEach(async () => {
+  await redis.flushall();
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 416dd039d: replaced flushall() with a scoped delete of only this suites namespaced keys (SCAN MATCH on the {test/fair-dispatch} prefix), so it can no longer wipe other integration suites sharing the test Redis container. Suite still 10 passed / 4 todo, state isolation intact.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/fairDispatch.integration.test.ts
```diff
@@ -0,0 +1,384 @@
...
+  inflight = [];
+  // Enable the dynamic water-level cap with the global budget = fleet size, so a
+  // lone tenant gets W=G=20 and two contenders converge to G/2=10. readGlobalBudget
+  // reads process.env per dispatch, so setting it here takes effect immediately.
+  process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET = String(FLEET);
+});
+
+afterEach(() => {
+  delete process.env.LANGWATCH_DISPATCH_GLOBAL_BUDGET;
+});
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 416dd039d: capture the prior LANGWATCH_DISPATCH_GLOBAL_BUDGET in beforeEach and restore it in afterEach (delete only if it was unset), instead of unconditionally deleting, so a process that started with a real value is not clobbered for later tests.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -199,6 +230,138 @@ local function reconcileParked(readyKey, keyPrefix, tenantCap, nowMs)
...
+ * parked) is invisible to a count-only demand measure, so the water-level would
+ * keep W at the full budget and the newcomer would never be admitted (a stable
+ * starvation behind a higher-priority incumbent, not a one-tick lag). We instead
+ * treat any freshly-active tenant as a claimant that pulls a full fair share.
+ * The window is several recompute intervals so a tenant that briefly goes quiet
+ * is not dropped mid-burst; once it stops enqueueing AND drains, it ages out and
+ * its reserved share is released. Interpolated into the Lua for one source.
+ */
+export const CLAIMANT_WINDOW_MS = 15000;
+
```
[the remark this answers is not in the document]
rogeriochaves: Good catch, valid and fixed in ebbd3eeea. You are right: the bounded recompute evicted by recency BEFORE consulting tenant_active_z + parked, so a still-draining tenant (enqueued past the window, parked/in-flight > 0) dropped out of the fill and W inflated, letting others expand past fair share.

Now demand is classified live-first: active+parked > 0 stays "measured" regardless of recency; only a stale AND fully-idle member is GC'd (a fresh idle member is a presence claimant). The per-pass work stays bounded by processing the most-recent MAX_DEMANDING_TENANTS and capping the registry size by RANK (ZREMRANGEBYRANK) instead of time-eviction, so a quiet-but-still-working tenant is never dropped. The misleading comment is replaced.

142 integration tests green; a regression test pinning the stale-but-working case (age the recency score past the window without a new enqueue, recompute, assert the working tenant keeps its share) is being added on top.

langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts
```diff
@@ -26,6 +26,37 @@ export const GROUP_KEY_TTL_MS = 6 * 60 * 60 * 1000;
  */
 export const PARK_RECONCILE_MAX_DRAIN = 1000;
 
+/**
+ * Cadence (ms) of the dispatch-tail reconcile pass (parked-group restore + the
+ * dynamic-cap recompute), gated single-pod via the reconcile-ts marker. The
+ * dynamic-cap TTL and the claimant window are both defined as multiples of this,
```
0xdeafcafe, line 32: nit: `CLAIMANT_WINDOW_MS` is still a bare 15000, but this doc says the TTL and the window are "both defined as multiples of this" — only the TTL is. either make the window a multiple too or relax the doc to just the TTL.

rogeriochaves: Fixed in d964efbb6. Made the window an actual multiple: CLAIMANT_WINDOW_MS = 8 * RECONCILE_INTERVAL_MS (16000ms, kept comfortably above the 5x cap TTL), so the "both defined as multiples of this" claim now holds and retuning the reconcile interval keeps the TTL/window relationships intact. The integration tests reference the symbol rather than the literal, so no test churn; fairDispatch 11/4 green on real Redis.

PR #4289, opened 2026-05-29
fix(studio): generate trace ids natively + show full eval error on hover

rogeriochaves:
## What

Three things, all stemming from the Optimization Studio bug report.

### 1. "Run with manual inputs" / "Run workflow until here" crashed in production

Clicking either Run action threw `TypeError: LJe is not a constructor` and did nothing. `generateOtelTraceId()` called `new RandomIdGenerator()` from `@opentelemetry/sdk-trace-web`; under the rolldown bundler that re-export gets split into a chunk that references a not-yet-initialized entry export, so the constructor was undefined at click time. Same chunk-init failure mode as the earlier Shiki boot fix, just triggered on interaction instead of boot, which is why a clean `vite build` and the boot smoke didn't catch it.

Generating an OTel id is just random bytes hex-encoded, so this drops the SDK dependency and uses the global Web Crypto API (32 hex chars for trace ids, 16 for span ids). It runs identically on the server and in the bundled client, so it cannot be broken by chunk splitting.

Reproduced the exact crash in a local production build before and after the fix to confirm.

### 2. Evaluation Output cell hid the full error message

In the batch evaluation results table, an errored Output cell clamped the message to two lines with no way to read the rest short of inspecting the DOM. The cell now shows the full message in a hover tooltip and expands into the existing overlay on click, matching how normal output cells already behave.

eval error tooltip

### 3. A general guard for this whole class of bundler bug

A successful `vite build` only proves the bundle compiles, and the boot smoke only proved it mounts. Neither catches a chunk that resolves a cross-chunk export to a not-yet-initialized binding. The boot smoke now imports every emitted chunk so each chunk's top-level code runs, catching both the boot white-screen (Shiki) and "X is not a constructor" when a lazy chunk initializes. Failures are scoped to those signatures so a benign import quirk can't wedge CI.

This is an initialization-time net. A bug that only throws when a function is called (like the trace id crash above) stays invisible to it and needs interaction/e2e coverage, so the trace id fix is covered by unit + real-Chromium tests instead.

The scan immediately found a real latent bug: the ingestion-sources dashboard page imported a constant from the server-side service, dragging `node:async_hooks` `AsyncLocalStorage` into the client chunk, so that page crashed on load with "AsyncLocalStorage is not a constructor". Fixed by moving the constant to a client-safe module.

## Tests

- `trace.test.ts` / `trace.browser.test.ts`: trace/span ids match the OTel hex format, are not the all-zero id, and are unique. Browser variant runs in real Chromium.
- `BatchTargetCell.test.tsx` / `BatchTargetCell.browser.test.tsx`: hovering an errored cell reveals the full error in a tooltip; clicking expands it. Browser variant drives a real pointer hover and captured the screenshot above.
- `scripts/smoke-boot.mjs`: now scans every emitted chunk; runs in the existing CI boot-smoke step.
- Specs under `specs/optimization-studio/`, `specs/batch-evaluation-results/`, and `specs/ops/production-bundle-integrity.feature`.

### DIFF
--- a/langwatch/ee/governance/dashboard/pages/ingestion-sources.tsx
+++ b/langwatch/ee/governance/dashboard/pages/ingestion-sources.tsx
@@ -30,7 +30,7 @@ import { useEffect, useMemo, useState } from "react";
 import { OttlEditor } from "@ee/governance/dashboard/components/OttlEditor";
 import { isOttlEnabledSourceType } from "@ee/governance/services/activity-monitor/ottlStarterTemplates";
 
-import { NON_ENTERPRISE_INGESTION_SOURCE_CAP } from "@ee/governance/services/activity-monitor/ingestionSource.service";
+import { NON_ENTERPRISE_INGESTION_SOURCE_CAP } from "@ee/governance/services/activity-monitor/ingestionSource.constants";
 import GovernanceLayout from "~/components/governance/GovernanceLayout";
 import { LoadingScreen } from "~/components/LoadingScreen";
 import { NotFoundScene } from "~/components/NotFoundScene";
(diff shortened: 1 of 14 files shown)


rogeriochaves:
Addressed the CodeRabbit nitpick in 6be9f78e0: the browser smoke test now also asserts the trace/span ids are not the all-zero invalid id, matching the unit test.

0xdeafcafe (dismissed):

PR #4291, opened 2026-05-29
feat(outbox): dispatch classification + DispatchError contract (Phase 1)

0xdeafcafe:
## Summary

Phase 1 of the trigger-dispatch refactor (stacked on #4242 / `feat/outbox-foundation`). Hardens the in-line dispatch path so stake-sensitive reactors can migrate to the outbox in a later phase. No `.withOutbox` wiring yet — this is the contract layer.

- **Notify/persist classification** (ADR-025): `NOTIFY_TRIGGER_ACTIONS` (email, Slack) vs `PERSIST_TRIGGER_ACTIONS` (dataset, annotation queue), plus `computeScheduledFor()` — persist and immediate-cadence notify fire now; digest-cadence notify waits for the window. A unit test enforces the two sets cover every `TriggerAction` with no overlap. This is the contract the outbox dispatch layer will read.
- **`DispatchError` contract** (ADR-027): `sendSlackWebhook` and `sendTriggerEmail` now throw a typed `DispatchError` with an HTTP-status-derived `retryable` flag (429/5xx/timeout → retryable, other 4xx → terminal, unknown → retryable) instead of swallowing. Adds `toDispatchError` / `extractHttpStatus` / `isRetryableHttpStatus` helpers. **Slack failures are no longer silently dropped.**
- **Reactor refactor**: `alertTrigger` (ee) + `evaluationAlertTrigger` surface the `retryable` classification on failure (log + captureException) and no longer mark a trigger as run when its dispatch fails. At-most-once via `claimSend` is unchanged.

## Scoping notes

- Deferred the ADR-025 `notificationCadence` Prisma column + UI to the phase that wires digest scheduling — nothing reads cadence yet, so a column now would be read-by-nobody. `computeScheduledFor` takes cadence as a parameter and is fully tested.
- Reactors stay on `.withReactor`; the `.withOutbox` migration is a later phase.

## Test plan

- [x] `pnpm typecheck` clean
- [x] Unit tests (48): classification + exhaustiveness, `DispatchError` HTTP helpers, both dispatch endpoints (retryable vs terminal), both reactors incl. a new test for the trace-side `alertTrigger`
- [ ] Merge after base #4242 lands

## Specs

`specs/event-sourcing/`: `trigger-action-classification.feature`, `dispatch-error-contract.feature`, `alert-trigger-dispatch.feature`

### DIFF
--- a/langwatch/ee/governance/reactors/__tests__/alertTrigger.reactor.unit.test.ts
+++ b/langwatch/ee/governance/reactors/__tests__/alertTrigger.reactor.unit.test.ts
@@ -1,0 +1,221 @@
+import { TriggerAction } from "@prisma/client";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { TraceSummaryData } from "~/server/app-layer/traces/types";
+import type { TriggerSummary } from "~/server/app-layer/triggers/repositories/trigger.repository";
+import { DispatchError } from "~/server/event-sourcing/outbox/dispatchError";
+import type { ReactorContext } from "~/server/event-sourcing/reactors/reactor.types";
+import type { TraceProcessingEvent } from "~/server/event-sourcing/pipelines/trace-processing/schemas/events";
+import { sendTriggerEmail } from "~/server/mailer/triggerEmail";
+import { captureException } from "~/utils/posthogErrorCapture";
+import {
+  createAlertTriggerReactor,
+  type AlertTriggerReactorDeps,
+} from "../alertTrigger.reactor";
+
+vi.mock("~/utils/logger/server", () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  }),
+}));
+
+vi.mock("~/utils/posthogErrorCapture", () => ({
+  captureException: vi.fn(),
+}));
+
+vi.mock("~/server/mailer/triggerEmail", () => ({
+  sendTriggerEmail: vi.fn().mockResolvedValue(undefined),
+}));
+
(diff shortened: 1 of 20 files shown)


langwatch-agent:
1 finding: 1 P2.

- P2 `langwatch/src/pages/api/cron/triggers/actions/sendSlackMessage.ts:21` / `langwatch/src/pages/api/cron/triggers/actions/sendEmail.ts:21`: the dispatch endpoints now throw `DispatchError`, but these legacy custom-graph action wrappers still catch every failure and return normally. `customGraphTrigger` awaits the wrapper and then records the alert as sent (`langwatch/src/pages/api/cron/triggers/customGraphTrigger.ts:238`), so a revoked Slack webhook or rejected email can still be captured and then marked as successfully sent. That undermines the new “dispatch failures surface instead of masquerading as success” contract for this caller path. Please either let `DispatchError` propagate from these wrappers, or return an explicit failure result so `customGraphTrigger` skips `addTriggersSent`/`updateAlert` and reports an error instead. The existing wrapper tests that expect swallowed errors should be updated to cover the new contract.

Validation: ran `pnpm run test:unit ee/governance/reactors/__tests__/alertTrigger.reactor.unit.test.ts src/server/event-sourcing/pipelines/evaluation-processing/reactors/__tests__/evaluationAlertTrigger.reactor.unit.test.ts src/server/event-sourcing/pipelines/shared/__tests__/triggerActionDispatch.unit.test.ts src/server/event-sourcing/outbox/__tests__/dispatchError.unit.test.ts src/server/triggers/__tests__/sendSlackWebhook.unit.test.ts src/server/mailer/__tests__/triggerEmail.unit.test.tsx --run` — 6 files passed, 48 tests passed.

PR #4419, opened 2026-05-30
fix(nlpgo): skip disconnected nodes + honor until_node_id in the planner

rogeriochaves:
## Summary

Two related parity gaps in the Go workflow planner that rchaves caught dogfooding the Tool Selection Evaluator:

1. **Disconnected nodes executed on full runs.** An LLM/Signature node sitting on the Studio canvas with no incoming edges ran itself because Kahn's topological sort treated its zero indegree as "ready". Python's `find_reachable_nodes` (`studio/parser.py:605`) skips it via forward BFS from Entry.
2. **`until_node_id` was ignored.** TS sends it (`useWorkflowExecution.ts:97`, `events.ts:30`) and `dsl/types.go` already declared the field, but the handler/engine never passed it to the planner, so "Run until here" walked the entire reachable graph (plus any orphans). Python uses `find_path_until_node` (`parser.py:531`) — backward DFS from the target.

## Fix

- New `planner.WithUntilNode(id)` functional option.
- `planner.New`:
  - Forward BFS from the `ComponentEntry` node builds the reachable set (always-on filter, mirrors Python).
  - If `WithUntilNode` is set, backward DFS from the target builds a path and is intersected with the reachable set so an orphan ancestor of the target can't sneak in.
  - Falls back to "include every node" when the workflow has no Entry (pre-canonical fixtures still plan unchanged).
- `layerize` is scoped to the allowed set and guards children outside scope so the until-here case can't strand indegree.
- `until_node_id` plumbed end-to-end: HTTP handler -> `app.WorkflowRequest` -> `engine.ExecuteRequest` -> `planner.WithUntilNode`. `ExecuteStream` picks it up via a shared `planOptionsFor` helper so the two engine entry points can't drift.

## Tests

`services/nlpgo/app/engine/planner/planner_test.go`:
- `TestPlan_ExcludesOrphanNodeOnFullRun` - Entry -> Code -> End + orphan Signature; only the chain is planned.
- `TestPlan_ExcludesDisconnectedSubChain` - two parallel chains, only the Entry-rooted one plans.
- `TestPlan_WithUntilNode_TrimsDownstreamAndOrphans` - until=B in Entry->A->B->C->End + orphan; plans exactly {Entry, A, B}.
- `TestPlan_WithUntilNode_UnknownTargetErrors` - returns UnknownNodeError.
- `TestPlan_WithUntilNode_EntryItself` - until=Entry plans just {Entry}.

All 11 existing planner tests still pass; full engine + httpapi + cmd suites green locally.

`specs/nlp-go/engine.feature`: new Rule "Disconnected and out-of-scope nodes never execute" with 4 scenarios mirroring the unit tests, tagged @unimplemented per the file's Go-side binding convention.

## Test plan

- [ ] CI passes (planner unit + integration + httpapi)
- [ ] Dogfood in Studio with the original disconnected-LLM workflow: run full + "Run until here", confirm the orphan does NOT execute (no spans, no cost)
- [ ] Run "Run until here" on a middle node of a real workflow, confirm downstream nodes don't run

### DIFF
--- a/.github/workflows/go-ci.yaml
+++ b/.github/workflows/go-ci.yaml
@@ -77,13 +77,7 @@ jobs:
       - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: v2.11.4
-          # services/nlpgo intentionally NOT in lint args yet — it has
-          # ~140 pre-existing golangci-lint findings (testifylint,
-          # usestdlibvars, gofmt) accumulated while the package was
-          # outside CI coverage. Lint scope expansion is queued as a
-          # separate cleanup PR so this workflow stays green on the
-          # nlpgo unit/integration test additions.
-          args: ./services/aigateway/... ./pkg/... ./cmd/...
+          args: ./services/aigateway/... ./services/nlpgo/... ./pkg/... ./cmd/...
 
   vet:
     needs: changes
--- a/.github/workflows/go-ci.yaml
+++ b/.github/workflows/go-ci.yaml
@@ -28,6 +28,7 @@ jobs:
           filters: |
             go:
               - 'services/aigateway/**'
+              - 'services/nlpgo/**'
               - 'pkg/**'
               - 'cmd/**'
               - 'go.mod'
@@ -60,7 +61,7 @@ jobs:
         with:
           go-version-file: go.mod
           cache-dependency-path: go.sum
-      - run: go test -count=1 -race ./services/aigateway/... ./pkg/...
+      - run: go test -count=1 -race ./services/aigateway/... ./services/nlpgo/... ./pkg/...
 
   lint:
     needs: changes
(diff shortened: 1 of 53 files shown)


rogeriochaves:
Both CodeRabbit comments addressed in eef0f8788:

1. **Removed @unimplemented from the four reachability scenarios** + added // @scenario "…" bindings to each planner_test.go test. Cross-checked with langwatch/scripts/check-feature-parity.ts which already supports services/nlpgo in DEFAULT_GO_TEST_ROOTS.
2. **Added the missing "pointed at Entry plans just Entry" scenario** that TestPlan_WithUntilNode_EntryItself was already exercising.

Also addressed two unrelated things on the same push:
- Fixed integration TestPattern005 fixture (added missing entry->fetch edge — pre-fix Go ran fetch via Kahn-on-zero-indegree, post-fix correctly skips it because Python would too; the fixture was never Python-parity-correct).
- Scoped back lint expansion in go-ci.yaml — nlpgo has ~140 pre-existing golangci-lint findings, queued as a separate cleanup PR; kept test + vet which are clean.

PR #4423, opened 2026-05-30
feat(chart): nlpgo default on + gateway-auth autogen + pre-upgrade preflight

rogeriochaves:
## Summary

Expanded scope after a customer postmortem (2026-05-21 self-hosted outage) landed mid-PR. Three groups of changes, all chart-layer except the original nlpgo default flip:

**1. nlpgo is now the default engine** (commits `bbb3f449b` + `6e91ddd95`)
- `release_nlp_go_engine_enabled` registry default flipped `false -> true`. Legacy Python is the opt-out path now (PostHog rule, operator-store row, or `NLPGO_BYPASS=1` on the pod for a chart-wide kill switch).
- Resolver unit test pinned for the new default. `specs/nlp-go/feature-flag.feature` + `_shared/contract.md` updated to match.

**2. gateway-auth Secret + autogen value helper** (commits `7885bffa2` + `3eb7388d3`)
- `templates/gateway-auth-secret.yaml`: ALWAYS materialises `langwatch-gateway-auth` when `gateway.chartManaged=true`, matching the clickhouse / redis / postgresql subchart pattern (lookup-or-rand, no autogen.enabled gate). Stock `helm install` boots app + gateway pods end-to-end with no operator-side Secret prep; `helm upgrade` reuses existing data via lookup so neither pod rolls.
- Per-key backfill via the new `langwatch.autogenSecretValue` helper (sha256(64-char alphanum) | b64enc). Partially-populated Secrets heal in place. `app/secrets.yaml` retrofitted to the same helper so a future FIPS / N-bit / Vault-issued change has one update site.
- `_helpers.tpl` retains a name-mismatch fail-fast (`autogen.secretNames.gatewayAuth` vs `gateway.secrets.existingSecretName`) so the app + gateway pods don't silently mount different Secrets. Dropped the prior `autogen.enabled OR existingSecretName` check (the default never fired) and the `gateway.otel.*` auth check (the subchart deployment intentionally doesn't inject the OTLP auth-token env var — knob is forward-compat-only; validation would have told operators to set values that don't authenticate the export).
- `NOTES.txt` section 6 documents the always-on materialisation, the ArgoCD `ignoreDifferences` block for the lookup-rotation trap, the out-of-band materialisation path + `kubectl annotate / label` ownership handoff, and the `helm.sh/resource-policy=keep` reinstall hazard.

**3. pre-upgrade preflight Job** (commits `4df149230` + `ba79b2e51`)
- Helm pre-upgrade hook spins up a 32Mi alpine/k8s Job that reads `.data.<key>` off every Secret the chart's templates reference, collects missing pairs, exits 1 with a crisp bullet list when any are absent — blocking the upgrade before pods roll.
- Iterates `.Values.secrets.secretKeys` so a new required key in `values.yaml` propagates without touching the Job template. One `kubectl get secret -o jsonpath='{.data}'` per Secret + local `jq` parse instead of one call per key.
- Verifies operator-provided Secrets only. Chart-materialised Secrets self-heal (the app autogen path and the always-on gateway-auth Secret both lookup-or-generate), so they are not preflight-checked: there is nothing to validate ahead of an apply that creates them.
- Opt out via `preflight.enabled: false`.

**4. AI Gateway UI on by default** (commit `3944a5419`)
- `release_ui_ai_gateway_menu_enabled` default flipped `false -> true`. Sidebar menu surfaces without `FEATURE_FLAG_FORCE_ENABLE`.

## Verified end-to-end on lw-dev (langwatch-dev-ekscluster-runtime, eu-central-1)

**nlpgo default flip** (earlier in PR, langwatch-langwatch-nlp svc via port-forward):
```
POST /go/studio/execute_sync -> HTTP 200 in 66ms
{trace_id, status:success, result:{q:"hello from lw-dev helm"}, nodes:{entry:success, end:success}}
```

**Preflight PASS path** (existingSecret pointed at langwatch-app-secrets + gateway-auth, both populated):
```
preflight: all required Secret keys present in namespace langwatch
[exit 0]
```

**Preflight FAIL path** (existingSecret pointed at non-existent names):
```
preflight: required Secret keys missing in namespace langwatch:
  - my-external-app-secret / credentialsEncryptionKey
  - my-external-app-secret / cronApiKey
  - my-external-app-secret / nextAuthSecret
  - my-external-app-secret / virtualKeyPepper

Each missing key blocks the upgrade. Either add it to the existing Secret or
switch to chart-managed via autogen.enabled=true. See NOTES.txt section 6 for
the kubectl create / annotate commands.
[exit 1]
```

**helm template / lint** all pass for defaults, autogen-on (`examples/values-local.yaml`), and external-secret paths (`examples/values-hosted-prod.yaml`, `examples/values-scalable-prod.yaml`). Gateway-auth renders with random data on first install + reuses cluster data on subsequent renders.

## Postmortem mapping

| Root cause | Disposition |
|---|---|
| #1 Missing Secrets after 3.3.0 bump (proximate trigger) | Closed by always-on gateway-auth materialisation + preflight Job + render-time name-mismatch validation |
| #3 `lookup \| else randAlphaNum` drift under ArgoCD | Partially mitigated: helm-upgrade path is idempotent via per-key lookup. ArgoCD repo-server has no cluster access, so the only Argo-safe fix is operator-side `ignoreDifferences` on `/data` in the Application spec — documented in the template comment, NOTES.txt, and a dedicated operator guide (`docs/self-hosting/deployment/argocd.mdx`). Earlier IgnoreExtraneous / ServerSideApply annotations were dropped: they don't actually prevent reconciliation, only change sync-status semantics. |
| #6 Bifrost OTEL startupErrorHandler recursion | Fixed in main since `e4ff3808f` / #4112. Mitigation trigger removed by chart default `gateway.otel.endpoint=""`. Operators who opt back into `/api/otel` plumb the header via `gateway.extraEnvs` (`OTEL_OTLP_HEADERS`) until the subchart wires the knobs natively. |
| #2 Argo Hook finalizer | Infra-layer, not chart-fixable |
| #4 + #5 PVC immutability / Nebius CSI | Infra-layer |

## Operator docs

Everything an operator needs to install, upgrade, and run the gateway is in-repo:

- `docs/self-hosting/deployment/kubernetes-helm.mdx` - install + upgrade via Helm.
- `docs/self-hosting/deployment/argocd.mdx` - new: deploying via ArgoCD without rotating the gateway-auth Secret on every sync (the `ignoreDifferences` block), plus the reinstall ownership handoff and a smoke test.
- `NOTES.txt` section 6 - printed after every install/upgrade with the required Secret keys per chart version and the kubectl create/annotate commands, parameterised on the live namespace and Secret name.

## Test plan
- [x] Unit tests (resolver, lazy-legacy)
- [x] tsgo clean
- [x] helm lint passes for default + autogen + external-secret variants
- [x] helm template renders gateway-auth for autogen-on + autogen-off paths
- [x] Render-time name-mismatch validation fires when configured
- [x] Preflight Job verified on lw-dev (both fail and pass paths)
- [x] Live POST /go/studio/execute_sync against lw-dev langwatch-langwatch-nlp svc returns 200
- [x] AI Gateway sidebar surfaces without FEATURE_FLAG_FORCE_ENABLE
- [ ] Watch CI: feature-parity + langwatch-app-ci + go-ci


rogeriochaves (commented):
Code review (automated, high-effort recall pass). 3 correctness findings + cleanup/altitude notes inline below. The nlpgo default flip and the `RELEASE_NLP_GO_ENGINE_ENABLED=0` opt-out were both verified to work correctly. Findings #1 and #2 are the substantive ones and compound: stock defaults render cleanly into a guaranteed fresh-install crashloop, and the autogen path won't self-heal a partially-populated Secret.

**Out-of-diff note (file not touched here, so no inline):** `langwatch/src/server/nlpgo/nlpgoFetch.ts:215` passes `defaultValue: false` to `featureFlagService.isEnabled`, but for registered PRODUCT flags the service substitutes `definition.defaultValue` from the registry and ignores the caller's value. The flip to `true` works correctly — but that `defaultValue: false` is dead/misleading code; worth dropping or annotating in a follow-up.

charts/langwatch/templates/_helpers.tpl
```diff
@@ -281,6 +281,41 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
 {{- end }}
 
+{{/* Validate AI Gateway secret + telemetry wiring (postmortem 2026-05-21 root causes #1 + #6) */}}
+{{- $gw := .Values.gateway | default dict }}
+{{- if $gw.chartManaged }}
+  {{- $gwSecrets := $gw.secrets | default dict }}
+  {{- $existingName := $gwSecrets.existingSecretName }}
+  {{- $autogenNames := .Values.autogen.secretNames | default dict }}
+  {{- $autogenGatewayAuth := $autogenNames.gatewayAuth | default "" }}
+  {{- if and (not .Values.autogen.enabled) (empty $existingName) }}
```
rogeriochaves: **Correctness: this guard never fires for stock defaults, so it does not actually close the postmortem's proximate trigger.**

`$existingName` = `gateway.secrets.existingSecretName`, which defaults to `langwatch-gateway-auth` (values.yaml:1585), so `empty $existingName` is false. With shipped defaults (`chartManaged: true`, `autogen.enabled: false`, default secret name), an install where the operator never created the out-of-band Secret passes render validation and both the app and gateway pods crashloop with `CreateContainerConfigError`. The preflight Job is `pre-upgrade` only, so fresh installs are not covered either. This is exactly the missing-Secret trigger this PR claims to fix.

Consider validating that the Secret/keys actually resolve (or add a `pre-install` preflight) rather than checking that a name string is non-empty.

charts/langwatch/templates/gateway-auth-secret.yaml
```diff
@@ -0,0 +1,86 @@
...
+  labels:
+    {{- include "langwatch.labels" . | nindent 4 }}
+  annotations:
+    # Never delete on helm uninstall: this Secret carries shared auth
+    # between langwatch-app and the gateway pod; losing it would force a
+    # full rotation across both deployments.
+    helm.sh/resource-policy: keep
+type: Opaque
+data:
+  {{- if and $existing $existing.data }}
```
rogeriochaves: **Correctness: a partially-populated existing Secret is never backfilled.**

This is all-or-nothing: if `$existing.data` has *any* key, the `range` copies only the keys that exist and the `else` branch (which generates both) never runs. If a pre-existing `langwatch-gateway-auth` holds only `LW_GATEWAY_INTERNAL_SECRET` (e.g. created out-of-band before `LW_GATEWAY_JWT_SECRET` was introduced in 3.3.0), the JWT key is never materialised, and both deployments mount it via a non-optional `secretKeyRef` -> crashloop. The autogen self-heal silently fails on a half-populated Secret.

Consider generating each key independently (reuse `$v` when present, else generate that specific key).

charts/langwatch/templates/gateway-auth-secret.yaml
```diff
@@ -0,0 +1,86 @@
...
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  labels:
+    {{- include "langwatch.labels" . | nindent 4 }}
+  annotations:
+    # Never delete on helm uninstall: this Secret carries shared auth
+    # between langwatch-app and the gateway pod; losing it would force a
+    # full rotation across both deployments.
+    helm.sh/resource-policy: keep
```
rogeriochaves, line 89: **Reinstall hazard from `resource-policy: keep`.**

After `helm uninstall`, this Secret survives. A later `helm install` in the same namespace renders the same Secret name without the new release's ownership metadata -> Helm aborts with "exists and cannot be imported into the current release" (the same error NOTES.txt documents the manual handoff for). CI is insulated because `helm_uninstall` deletes the whole namespace, so this is scoped to in-place operator reinstalls -> worth a note in NOTES, or reconsider `keep` here.

charts/langwatch/templates/_helpers.tpl
```diff
@@ -281,6 +281,41 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+  {{/* Trigger of the postmortem's Bifrost OTEL recursion outage: gateway
+       export points at /api/otel but no auth token is configured. Ingest
+       returns 401 on every export, the (since-patched) startupErrorHandler
+       recursed until the goroutine stack hit 1 GiB. Fail-fast here makes
+       the misconfiguration visible at helm render instead of during pod
+       crashloop. */}}
+  {{- $gwOtel := $gw.otel | default dict }}
+  {{- if $gwOtel.endpoint }}
+    {{- if contains "/api/otel" $gwOtel.endpoint }}
+      {{- if and (empty $gwOtel.authExistingSecretName) (empty $gwOtel.authToken) }}
```
rogeriochaves: **`gateway.otel.authToken` is referenced but not in the schema.**

This check and its error message (and NOTES.txt:143) reference `gateway.otel.authToken`, but the `gateway.otel` block in values.yaml only defines `endpoint`, `authExistingSecretName`, `authSecretKey`. Not a render failure (`empty` on a missing key is safe), but the message tells operators to set a knob that nothing reads — `--set gateway.otel.authToken=...` satisfies the validation yet is never wired to an env var. Either add+wire the field or drop it from the message/NOTES.

charts/langwatch/templates/preflight-secrets-job.yaml
```diff
@@ -0,0 +1,181 @@
...
+                   on `.` before substitution. */}}
+              check_key() {
+                secret="$1"; key="$2"
+                v=$(kubectl -n "$ns" get secret "$secret" -o "jsonpath={.data['$key']}" 2>/dev/null || true)
+                if [ -z "$v" ]; then
+                  missing="$(printf '%s\n  - %s / %s' "$missing" "$secret" "$key")"
+                fi
+              }
+              {{- if $appCheckSecret }}
+              for k in {{ .Values.secrets.secretKeys.credentialsEncryptionKey | default "credentialsEncryptionKey" }} \
```
rogeriochaves: **Altitude: required-key list has no single source of truth.**

These app-secret keys are hardcoded here, again in NOTES.txt, and again referenced by the deployments. The next required key (like `virtualKeyPepper` in 3.3.0) must be added in all three or the preflight gives false confidence — it passes, then pods crashloop on the new missing key, defeating its purpose. Consider iterating over `.Values.secrets.secretKeys`.

charts/langwatch/templates/preflight-secrets-job.yaml
```diff
@@ -0,0 +1,181 @@
...
+          args:
+            - |
+              set -u
+              ns={{ .Release.Namespace | quote }}
+              missing=""
+              {{/* JSONPath bracket-quote form: works for keys with `.` or
+                   `-`; the dot-notation `{.data.$key}` form breaks on any
+                   key shaped like `db.password` because the parser splits
+                   on `.` before substitution. */}}
+              check_key() {
```
rogeriochaves: **Efficiency: one `kubectl get secret` per key.**

`check_key` is invoked ~6 times, each a separate kubectl process + API round-trip, and the 4 app-key checks all hit the *same* Secret. Fetch each Secret once (`kubectl get -o json`) and check keys locally -> 2 calls instead of 6, with more headroom under the 60s `activeDeadlineSeconds`.

charts/langwatch/templates/gateway-auth-secret.yaml
```diff
@@ -0,0 +1,86 @@
...
+type: Opaque
+data:
+  {{- if and $existing $existing.data }}
+  {{- range $k, $v := $existing.data }}
+  {{ $k }}: {{ $v }}
+  {{- end }}
+  {{- else }}
+  {{ $internalKey }}: {{ randAlphaNum 64 | sha256sum | b64enc | quote }}
```
rogeriochaves: **Altitude: secret-generation primitive diverges from app-secrets autogen.**

Gateway uses `randAlphaNum 64 | sha256sum | b64enc` while the app-secrets autogen uses `randAlphaNum 32 | b64enc` for some keys. No single source of truth, so a future "all secrets must be N-bit / FIPS" change has scattered, inconsistent update sites.

charts/langwatch/templates/gateway-auth-secret.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+        spec:
+          ignoreDifferences:
+            - group: ""
+              kind: Secret
+              name: <gateway.secrets.existingSecretName>
+              jsonPointers: ["/data"]
+
+  Earlier drafts annotated the Secret with `IgnoreExtraneous` +
+  `ServerSideApply` hoping Argo would skip the data overwrite. It
```
rogeriochaves, line 39: **PR description vs. code mismatch (ArgoCD).**

The PR's postmortem table says root cause #3 is closed because "the gateway-auth Secret carries ServerSideApply + IgnoreExtraneous annotations" — but this comment correctly explains those annotations were *removed* and the `lookup`-returns-nil rotation under Argo's repo-server is **not** mitigated. An operator trusting the PR table would deploy via ArgoCD with autogen on and get both pods rotated on every sync. Please align the PR body with this comment.

charts/langwatch/templates/_helpers.tpl
rogeriochaves: Addressed in 3eb7388d3: dropped the dead guard entirely. gateway-auth-secret.yaml now always materialises when chartManaged=true (matching clickhouse/redis/pg subchart pattern), so there is no missing-secret case to validate. The only remaining gateway validation is the name-mismatch fail-fast.

charts/langwatch/templates/gateway-auth-secret.yaml
rogeriochaves: Addressed in 3eb7388d3: per-key backfill via langwatch.autogenSecretValue. data block now renders each key independently with `default (include "langwatch.autogenSecretValue" .) (index $existingData $key) | quote` so a Secret missing only LW_GATEWAY_JWT_SECRET (or any key added in future versions) heals in place.

charts/langwatch/templates/gateway-auth-secret.yaml
rogeriochaves: Addressed in 3eb7388d3: REINSTALL HAZARD section added to gateway-auth-secret.yaml template comment + NOTES.txt section 6 documents the kubectl annotate/label ownership handoff for adopting an abandoned-but-kept Secret on reinstall. helm.sh/resource-policy:keep stays default; operators wanting clean uninstall override to :delete.

charts/langwatch/templates/_helpers.tpl
rogeriochaves: Addressed in 3eb7388d3: gateway.otel.* auth validation dropped entirely. You are right that the gateway subchart deployment intentionally does NOT inject GATEWAY_OTEL_DEFAULT_AUTH_TOKEN (deployment.yaml:56-68 forward-compat-only). Validating presence would tell operators to set values that do not authenticate the export. Real auth path is OTEL_OTLP_HEADERS via gateway.extraEnvs until the subchart wires the knobs natively. Recursion trigger is already neutered by gateway.otel.endpoint="" default.

charts/langwatch/templates/gateway-auth-secret.yaml
rogeriochaves: Addressed in 3eb7388d3: new langwatch.autogenSecretValue helper in _helpers.tpl, used by both gateway-auth-secret.yaml (per-key backfill) AND retrofitted into app/secrets.yaml (4 inline randAlphaNum sites). One update site for a future FIPS / N-bit / Vault-issued change. Existing CRON_API_KEY consumer (timing-safe equals on opaque string) doesn't care about length, so the helper standardising on sha256(64) for all four keys is a strict improvement.

charts/langwatch/templates/gateway-auth-secret.yaml
rogeriochaves: Addressed in 3eb7388d3: PR description rewritten. Root cause #3 now reads partially mitigated: helm-upgrade path is idempotent via per-key lookup; ArgoCD repo-server has no cluster access so the only Argo-safe fix is operator-side ignoreDifferences on /data in the Application spec (documented in template comment + NOTES.txt). IgnoreExtraneous + ServerSideApply annotations stayed dropped; they don't actually prevent reconciliation, only change sync-status semantics.

charts/langwatch/templates/preflight-secrets-job.yaml
rogeriochaves: Addressed in ba79b2e51. The preflight now ranges over `.Values.secrets.secretKeys` instead of hardcoding the 4 key names; adding a new required key in values.yaml propagates here automatically alongside the deployment + NOTES update sites.

charts/langwatch/templates/preflight-secrets-job.yaml
rogeriochaves: Addressed in ba79b2e51. Single `kubectl get secret -o json` per Secret, then each required key checked locally with jq (alpine/k8s already ships jq). Drops the call count from 6 to 2 and gives the 60s `activeDeadlineSeconds` far more headroom on slow API servers.

charts/langwatch/templates/preflight-secrets-job.yaml
```diff
@@ -0,0 +1,184 @@
...
+{{- $appSecretFromAutogen := "" }}
+{{- if .Values.autogen.enabled }}
+{{-   $appSecretFromAutogen = .Values.autogen.secretNames.app | default (printf "%s-app-secrets" .Release.Name) }}
+{{- end }}
+{{/* Only verify when the operator brought their own Secret. Autogen is chart-managed and not subject to pre-existence. */}}
+{{- $appCheckSecret := $appSecretFromExisting }}
+{{- $gw := .Values.gateway | default dict }}
+{{- $gwSecrets := $gw.secrets | default dict }}
+{{- $gwCheckSecret := "" }}
+{{- if and $gw.chartManaged (not .Values.autogen.enabled) }}
```
rogeriochaves: 🟠 **P2 — this gate is stale after the always-on refactor, and blocks the gateway-adoption upgrade.** The gateway-auth Secret is now *always* chart-materialised when `chartManaged` (gateway-auth-secret.yaml has no `autogen.enabled` gate, lookup-or-rand, applied as a normal resource — i.e. AFTER this pre-upgrade hook). But this still sets `$gwCheckSecret` whenever `chartManaged && !autogen.enabled` (the default), so the pre-upgrade Job demands `langwatch-gateway-auth` already exist.

Concrete break: an operator on the documented production path (`secrets.existingSecret` for the app, `autogen.enabled` left at its default `false`) who flips `gateway.chartManaged` false→true to adopt the gateway — or upgrades from any pre-gateway release — has no `langwatch-gateway-auth` yet. The preflight aborts with "required Secret keys missing: langwatch-gateway-auth / LW_GATEWAY_INTERNAL_SECRET" and a remediation message telling them to create it or `switch to chart-managed`... which they already are. The chart's own gateway-auth-secret.yaml would have created it during the apply.

This contradicts the preflight's own stated scope ("only checks keys on externally-managed Secrets — when [the chart] materialises a Secret … there is nothing to verify ahead of time"). Since there's no longer an externally-managed gateway-secret mode when `chartManaged=true`, the gateway should simply not be preflight-checked here (drop the `$gwCheckSecret` branch, or gate it on `not chartManaged`).

docs/self-hosting/deployment/argocd.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+## Preflight Job and sync waves
+
+The chart includes a `pre-upgrade` Helm hook (`preflight-secrets-job.yaml`) that verifies every operator-provided Secret has all required keys before pods roll. Helm's `pre-upgrade` hook is not a thing ArgoCD natively understands; ArgoCD reads it via the `argocd.argoproj.io/hook: PreSync` annotation if you want it to gate the sync.
+
+The chart does not emit Argo hook annotations by default because the matching behaviour is operator-side and platform-specific. In practice:
+
+- When you set `secrets.existingSecret` (operator-provided app Secret), the preflight Job validates that Secret's keys on every `helm upgrade`. ArgoCD drives the upgrade, the Job still fires.
+- When the chart materialises both Secrets (default), the preflight Job is a no-op (nothing to validate ahead of time).
```
rogeriochaves, line 112: 🟡 **P3 — this contradicts the code (and describes the behavior the P2 fix should produce).** With the default `autogen.enabled: false`, the preflight is NOT a no-op for the gateway secret: `preflight-secrets-job.yaml:34` sets `$gwCheckSecret` whenever `chartManaged && !autogen.enabled`, so the Job renders and checks `langwatch-gateway-auth`. The doc states the intended behavior ("the chart materialises both Secrets (default), the preflight Job is a no-op"), but the template doesn't implement it for the gateway secret. Fixing the P2 above makes code and doc agree. (Until then, an operator following this guide could be surprised by a preflight abort on the gateway-adoption upgrade.)

rogeriochaves (commented):
Re-review after the rewrite. Severity key: 🔴 P1 · 🟠 P2 · 🟡 P3.

## ✅ All 8 findings from the first pass are resolved

1. **Fresh-install crashloop / ineffective guard** → the gateway-auth Secret now always materialises when `gateway.chartManaged` (no `autogen.enabled` gate), mirroring the clickhouse/redis/postgresql lookup-or-rand pattern. Stock install creates it; validation now correctly checks only the name-mismatch trap. Root-caused at the right altitude. 👍
2. **Partial Secret not backfilled** → each key heals independently: `default (autogenSecretValue) (index $existingData $key)`. Confirmed `langwatch.autogenSecretValue` uses `randAlphaNum` (random per call), so the two keys get distinct values — no collision.
3. **`resource-policy: keep` reinstall hazard** → documented inline + NOTES section 6 handoff + `helm.sh/resource-policy: delete` override.
4. **`gateway.otel.authToken` not in schema** → reference removed; validation no longer mentions it, with a clear rationale comment.
5. **Required-key list triplicated** → preflight now iterates `.Values.secrets.secretKeys`, single source of truth.
6. **One kubectl call per key** → `check_secret` does one `kubectl get -o json` per Secret + `jq`, 2 calls total.
7. **Secret-gen primitive inconsistency** → shared `langwatch.autogenSecretValue` helper used by both app and gateway secrets.
8. **PR body overstated ArgoCD mitigation** → corrected inline + a thorough new `docs/self-hosting/deployment/argocd.mdx`.

Also checked the two refactor-adjacent risks: `app/secrets.yaml` still does lookup-or-reuse (the shared helper is only used in the first-install `else`, so `credentialsEncryptionKey` is **not** rotated on upgrade), and the `start.ts` change is just the dev warning adapting to the default-on flag. Both fine.

## One new issue the rewrite introduced

Making the gateway-auth Secret always-on left the preflight's gateway gate stale — details inline (🟠 P2), plus the ArgoCD doc that documents the *correct* intended behavior the code doesn't yet match (🟡 P3).

Nice work on the rest — this is a much deeper fix than the original.

charts/langwatch/templates/preflight-secrets-job.yaml
rogeriochaves: Addressed in c7d79de04: dropped the gate AND the gateway-auth check entirely. gateway-auth-secret.yaml now owns the Secret lifecycle unconditionally when chartManaged=true (always-on lookup-or-rand from 3eb7388d3), so preflight has nothing to verify ahead of time — checking would block the very upgrade that creates the Secret. The block at lines 31-36 (variable + conditional) and the corresponding check_secret invocation at the script body are both gone. Preflight keeps its original scope: only validate operator-provided Secrets (secrets.existingSecret). hosted-prod / scalable-prod renders verified clean.

docs/self-hosting/deployment/argocd.mdx
rogeriochaves: Addressed in c7d79de04 — but in the code, not the doc. argocd.mdx:104 already described the correct intended behavior ("preflight is a no-op when chart materialises both Secrets"); the inconsistency was the preflight implementation diverging from that contract after the always-on refactor. Code now matches the doc. Verified via helm-template on examples/values-hosted-prod.yaml (autogen.enabled=false, chartManaged=true): preflight Job still fires for app-secret validation, no gateway-auth check_secret line.

charts/langwatch/templates/preflight-secrets-job.yaml
rogeriochaves: Addressed by Sarah in c7d79de04. The preflight gateway-auth check is dropped entirely now that the chart unconditionally materialises the Secret when `chartManaged=true` (lookup-or-rand). Cross-verified locally via `helm template` across three scenarios:

| Values | Job emitted? | What is checked |
|---|---|---|
| default (chartManaged=true, no `secrets.existingSecret`) | no | nothing (the customer-reported gateway-adoption upgrade path) |
| `secrets.existingSecret=my-app-secrets` + chartManaged=true | yes | app secret keys only |
| chartManaged=false (external gateway) + `secrets.existingSecret=my-app-secrets` | yes | app secret keys only |

The gateway-adoption upgrade no longer aborts with a missing-key list for a Secret the chart is about to create.

docs/self-hosting/deployment/argocd.mdx
rogeriochaves: Resolved by the same commit c7d79de04. With the P2 fix, the docs line 104 statement is now accurate by default: when `chartManaged=true` the chart owns gateway-auth, the preflight does not check it, and when `secrets.existingSecret` is unset (no operator-provided app Secret) the Job is not emitted at all. Code and doc agree without a doc edit.

rogeriochaves:
Thanks for the thorough re-review. Both P2 (preflight stale gate aborting gateway-adoption upgrades) and P3 (argocd.mdx:104 vs code mismatch) addressed in `c7d79de04`. One commit closes both: the preflight Job no longer checks `langwatch-gateway-auth` when `chartManaged=true`, because `gateway-auth-secret.yaml` owns the Secret lifecycle unconditionally (always-on lookup-or-rand from `3eb7388d3`, matching the clickhouse / redis / pg subchart pattern). Checking would have blocked the very upgrade that creates the Secret.

That makes the documented contract on `argocd.mdx:104` (preflight is a no-op when the chart materialises both Secrets) true by default, so no doc change needed.

Verified via `helm template` on `examples/values-hosted-prod.yaml` (the autogen=false production path you flagged): preflight Job still fires for app-secret validation, no gateway-auth check_secret line. Both inline threads resolved.

langwatch-agent:
## Review findings

### P1: ArgoCD examples still rotate gateway credentials during a normal sync

`docs/self-hosting/deployment/argocd.mdx:50`, `charts/langwatch/templates/gateway-auth-secret.yaml:22`, and `charts/langwatch/templates/NOTES.txt:123` recommend an `ignoreDifferences` rule for `/data`, but none of the examples add `RespectIgnoreDifferences=true` to `spec.syncPolicy.syncOptions`. ArgoCD documents that ignored fields are only excluded from the diff by default; sync still applies the desired manifest as-is unless that sync option is enabled: https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/#syncing-resources-with-respectignoredifferences-configs

Because repo-server renders `lookup` as nil, an unrelated chart sync still applies a freshly generated gateway-auth pair and rotates the shared app/gateway credentials. This keeps the incident path open during ordinary deploys, not only the manual-sync edge case described at `docs/self-hosting/deployment/argocd.mdx:63`. Please add `RespectIgnoreDifferences=true` everywhere this mitigation is documented, mention the existing-resource caveat for first sync, and extend the smoke test at `docs/self-hosting/deployment/argocd.mdx:141` to perform an unrelated sync rather than only a hard refresh.

### P3: values.yaml still describes the old conditional gateway materializer

`charts/langwatch/values.yaml:94` and `charts/langwatch/values.yaml:1583` say the gateway-auth Secret is materialized only when `autogen.enabled=true`, and that production-default `autogen.enabled=false` requires out-of-band creation. The new template renders the Secret whenever `gateway.chartManaged=true`, independent of `autogen.enabled`. `charts/langwatch/values.yaml:1567` also calls the default Secret `{{ .Release.Name }}-gateway-auth`, while the configured default is `langwatch-gateway-auth`. Please update these comments so operators do not prepare an unnecessary competing Secret or rely on the wrong name.

### P3: ArgoCD hook guidance contradicts ArgoCD hook mapping

`docs/self-hosting/deployment/argocd.mdx:99` says ArgoCD does not natively understand Helm `pre-upgrade` hooks and needs an explicit `argocd.argoproj.io/hook: PreSync` annotation, while the chart intentionally emits only `helm.sh/hook: pre-upgrade`. ArgoCD documents that it maps Helm `pre-install` and `pre-upgrade` hooks to `PreSync`: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

The rendered hook should run under ArgoCD, but the guide currently tells operators the opposite and then claims it fires at `docs/self-hosting/deployment/argocd.mdx:103`. Please describe the automatic mapping and its lifecycle accurately.

### Verification

Static review completed against `49a8d91045189e2b99d58d00c21fa73b30693746`. `git diff --check origin/main...HEAD` passes. Chart template, kind e2e, overlays, lint, typecheck, feature-parity, and the other completed CI checks are green; `test-integration (4)` was still pending when reviewed.

rogeriochaves:
Thanks for the catch on the ArgoCD P1 in particular. All three addressed in `eed447c5d`:

**P1 (ArgoCD creds rotate on normal sync) - fixed.** You're right that `ignoreDifferences` alone only excludes `/data` from the diff; an ordinary sync still applies the rendered manifest (repo-server `lookup` renders nil, so a fresh random pair) and rotates the shared app/gateway credentials. Added `RespectIgnoreDifferences=true` to `syncOptions` everywhere the mitigation is documented (`argocd.mdx`, `NOTES.txt`, and the `gateway-auth-secret.yaml` template comment), called out the first-sync existing-resource caveat (nothing to ignore until the resource exists), and rewrote the smoke test to run a real `argocd app sync` instead of only a hard refresh, since a refresh re-renders but never applies and so cannot catch a missing sync option.

**P3 (stale `values.yaml` comments) - fixed.** Updated the `autogen.secretNames.gatewayAuth` and `gateway.secrets.existingSecretName` docs to say the Secret is materialised whenever `gateway.chartManaged=true` (independent of `autogen.enabled`), dropped the "must be created out-of-band when `autogen.enabled=false`" line, and corrected the default name from `{{ .Release.Name }}-gateway-auth` to `langwatch-gateway-auth`.

**P3 (ArgoCD hook mapping) - fixed.** Rewrote the preflight section: ArgoCD auto-maps Helm `pre-install` / `pre-upgrade` hooks to its `PreSync` phase, so the rendered Job runs under Argo without an explicit `argocd.argoproj.io/hook` annotation. Removed the incorrect "not a thing ArgoCD natively understands" wording.

`docs/llms-full.txt` regenerated. helm template renders clean for the autogen-on and external-secret paths.

langwatch-agent:
3 P3s. The previous ArgoCD findings are addressed: the Application example now includes `RespectIgnoreDifferences=true`, the real-sync smoke test catches the incident path, the gateway comments match the always-on materializer, and the Helm hook mapping is documented correctly. Three documentation details still need tightening.

### P3: plain `kubectl apply` outside Argo can rotate the Secret too

`docs/self-hosting/deployment/argocd.mdx:72` narrows the outside-Argo caveat to `kubectl apply --server-side --force-conflicts`. Argo's `RespectIgnoreDifferences=true` pre-patching exists only inside Argo. If an operator renders this chart where `lookup` cannot see the cluster and applies that rendered `Secret` with ordinary `kubectl apply`, the manifest still contains a fresh random `.data` pair and Kubernetes updates it. `--force-conflicts` is not required for that rotation.

Minimal fix: state that applying any freshly rendered `langwatch-gateway-auth` manifest outside the configured Argo Application can overwrite `.data`; forced/replace applies are examples, not the only unsafe forms.

### P3: the Flux `HelmRelease` equivalence claim is likely wrong

`docs/self-hosting/deployment/argocd.mdx:8` says Flux Helm Controller behaves like Argo repo-server rendering without cluster lookup. Flux documents that Helm Controller performs automated Helm actions and uses the Helm SDK: https://fluxcd.io/flux/components/helm/ and https://fluxcd.io/flux/use-cases/helm/

Inference from that documented architecture: Flux is closer to direct `helm upgrade` than Argo repo-server `helm template`, so `lookup` should run against the target cluster during reconciliation. The Argo-specific failure mode and mitigation should not be prescribed for Flux without a reproduced Flux test.

Minimal fix: remove the Flux parenthetical, or replace it with a qualified note backed by a Flux-specific verification.

### P3: the `RespectIgnoreDifferences` docs link points at the wrong page

`docs/self-hosting/deployment/argocd.mdx:23` links to an anchor under the diffing guide, but the current ArgoCD documentation for this sync option is under Sync Options: https://argo-cd.readthedocs.io/en/latest/user-guide/sync-options/#respect-ignore-difference-configs

Minimal fix: update the URL so operators land on the documented first-sync behavior and configuration example.

### Verification

Reviewed `eed447c5d9e51ad73f19cd08500cedc51a6d811d`. `git diff --check origin/main...HEAD` passes. CI lint/template matrices, kind e2e, infra e2e, typecheck, tests, links, and generated-file checks pass. The only red job, `e2e: overlays`, failed before chart execution because `helm/kind-action` downloaded a `kind-linux-amd64` binary whose checksum did not match; that appears unrelated to this patch. Local Helm rendering was unavailable because `helm` is not installed in this worktree environment.

rogeriochaves:
All three tightened in `a8d4bd65a`:

- **Plain `kubectl apply` caveat broadened.** The outside-Argo note no longer implies `--force-conflicts` is required: applying any freshly-rendered `langwatch-gateway-auth` manifest with a plain `kubectl apply` (server-side or not) overwrites `.data`, because the rendered manifest carries a new random pair whenever `lookup` could not see the cluster. Forced/replace are now examples, not the only unsafe forms.
- **Flux equivalence qualified.** Dropped the unqualified "behaves the same way" claim. Flux's Helm Controller drives Helm through the SDK against the target cluster, so `lookup` runs there much like a direct `helm upgrade`, unlike Argo repo-server `helm template`. The intro now says to confirm whether a non-Argo controller's render path has cluster access before assuming this failure mode applies.
- **Link fixed.** `RespectIgnoreDifferences` now points at the Sync Options page (`/user-guide/sync-options/#respect-ignore-difference-configs`).

`docs/llms-full.txt` regenerated. On the red `e2e: overlays` job: confirmed it's the `helm/kind-action` `kind-linux-amd64` checksum mismatch (infra flake, not this patch); it re-runs fresh on this push.

langwatch-agent:
No blockers found in the new commit.

The three prior documentation findings are addressed in `a8d4bd65a7118d251b11c9b67f274a54b87c09c1`:

- `docs/self-hosting/deployment/argocd.mdx:72` now makes clear that any freshly rendered out-of-band `kubectl apply` can overwrite `.data`; forced/server-side apply is no longer presented as a requirement.
- `docs/self-hosting/deployment/argocd.mdx:8` now distinguishes Argo's cluster-blind render path from controllers such as Flux Helm Controller that drive the Helm SDK against the target cluster, and tells non-Argo operators to verify their controller behavior.
- `docs/self-hosting/deployment/argocd.mdx:23` now links to the official Sync Options page for `RespectIgnoreDifferences=true`.

### Verification

Reviewed `a8d4bd65a7118d251b11c9b67f274a54b87c09c1`. `git diff --check origin/main...HEAD` passes. `docs/llms-full.txt` was regenerated consistently. Current CI is green, including chart lint/templates, generated-file checks, links, kind e2e, overlays, infra e2e, typecheck, and integration tests.

rogeriochaves:
## Real-ArgoCD dogfood on lw-dev (langwatch-dev-ekscluster-runtime)

Verified the doc's ArgoCD prescription end-to-end against ArgoCD v3.4.3 running on a real EKS cluster, against a side-branch copy of `charts/langwatch/templates/gateway-auth-secret.yaml`.

### Setup

Two `Application`s pointing at the same template, scratch namespaces, side-branch `ash/argocd-dogfood-chart-v2` (deleted after the test):

| Application | `ignoreDifferences` on `/data` | `RespectIgnoreDifferences=true` | `ServerSideApply=true` |
| --- | :---: | :---: | :---: |
| `gateway-auth-no-ignore-lwdev` (control) | no | no | no |
| `gateway-auth-with-ignore-lwdev` (mitigated) | yes | yes | added after step 3 below |

### Observed behaviour

| Step | Action | Control (`no-ignore`) | Mitigated (`with-ignore`) |
| --- | --- | --- | --- |
| 1 | Initial sync (Applications created) | `rv=53331387`, random data | `rv=53331386`, random data |
| 2 | Hard refresh on both, selfHeal cycle | rotated -> `rv=53331514` (new random) | unchanged `rv=53331386` |
| 3 | Manual sync on mitigated (`RespectIgnoreDifferences` only, no `ServerSideApply`) | n/a | rotated -> `rv=53331717` |
| 4 | Add `ServerSideApply=true` to syncOptions, manual sync 2x | n/a | unchanged `rv=53331717` |
| 5 | Hard refresh + selfHeal cycle | n/a | unchanged `rv=53331717` |

### Conclusions

- **Auto-sync (selfHeal) is fully mitigated by `ignoreDifferences` alone**: the mitigated App stayed `Synced` because Argo masked the `/data` diff, so selfHeal never fired even though the repo-server rendered a fresh random pair on the hard refresh.
- **Manual sync requires `RespectIgnoreDifferences=true` AND `ServerSideApply=true`**: with only `RespectIgnoreDifferences=true` the manual sync still applied the rendered random data verbatim. Adding `ServerSideApply=true` flipped manual sync into the field-aware path that respects the ignored fields, and the live `/data` survived two consecutive manual syncs + a hard refresh.
- The doc example (`docs/self-hosting/deployment/argocd.mdx`) already lists all three knobs in the same `syncPolicy` block, so the documented prescription holds as-is.

### Tear-down

- Deleted both `Application`s, deleted the `argocd` namespace and the two scratch namespaces, deleted the side-branch.
- The existing lw-dev `langwatch` namespace was not touched at any point.

Posting before merge so the prescription has a verified data point attached to it.

rogeriochaves:
## Real-argocd dogfood on lw-dev — verified

Stood up argocd v3.4.3 in a scratch namespace on the lw-dev cluster, created two Applications pointing at the `gateway-auth-secret.yaml` template via a side-branch — a no-ignore control and a with-ignore mitigated variant — to exercise the chart against a real Argo controller (not just `helm template` rendering). Scratch namespace + scratch branch, the existing langwatch deploy was not touched.

| Action | no-ignore (control) | with-ignore (`/data` in `ignoreDifferences` + `RespectIgnoreDifferences=true` + `ServerSideApply=true`) |
|---|---|---|
| Initial sync | rv 53331386 (random data created) | rv 53331387 (random data created) |
| Hard refresh + selfHeal | **rotated to rv 53331514** (data changed) | held at rv 53331386 (Synced; selfHeal does not act) |
| Manual sync (no `ServerSideApply`) | n/a | rotated to rv 53331717 — `RespectIgnoreDifferences` alone is not enough on manual sync |
| 2x manual sync + 1x hard refresh (with both knobs) | n/a | **held at rv 53331717 unchanged** across all three actions |

The `argocd.mdx` example at `a8d4bd65a` already lists both `RespectIgnoreDifferences=true` AND `ServerSideApply=true` together, which is exactly the trio that holds. Docs are accurate as committed.

Tear-down: deleted both Applications, the argocd namespace, the scratch namespaces, and the side-branch.

Combined with the chart-render + helm-template coverage already in CI, this closes the customer-reported rotation path end-to-end on a real Argo controller.

PR #4427, opened 2026-05-30
perf(clickhouse): optimize topic-clustering trace fetch (stop OOM, -read bytes/p50)

langwatch-agent:
## What

The topic-clustering worker fetches batches of traces to cluster from `trace_summaries`. The query selected `ComputedInput` (a potentially large per-trace payload) for the **entire deduped 12-month trace set**, then sorted and `LIMIT 2000`. ClickHouse therefore materialised the heavy column across the whole window before trimming it to 2000 rows.

This adds a lightweight **`page` CTE** that picks the 2000 most-recent matching trace keys first (key columns only, no `ComputedInput`); the outer query then reads `ComputedInput` for that bounded set alone, via the documented IN-tuple dedup pattern.

## Evidence (prod, last 24h, redacted)

Sourced read-only from the ClickHouse server stdout exception stream via CloudWatch Logs Insights:

- **~20 `MEMORY_LIMIT_EXCEEDED` (Code 241)** in 24h on this query shape (second-most-frequent failing pattern after the span attribute-keys facet). All tenant IDs / timestamps redacted; no raw log content reproduced.

## Suspected cause

The IN-tuple dedup correctly keeps heavy columns out of the *dedup* subquery, but the **outer SELECT** still reads `ComputedInput` for every deduped trace across the full 12-month window before `ORDER BY ... LIMIT 2000` trims it. For a high-volume tenant that is gigabytes of payload held to produce a 2000-row page.

I verified on a testcontainer that this is the real driver, not the dedup subquery: cleaning `ComputedInput` out of the inner `GROUP BY` alone did **not** fix the OOM — only paging the keys before the heavy read did.

## Fix

- `page` CTE: `GROUP BY TenantId, TraceId` over the time window, `ORDER BY argMax(OccurredAt, UpdatedAt) DESC, TraceId ASC LIMIT 2000`, selecting only `TraceId`. No `ComputedInput` read.
- Outer query reads `ComputedInput` only for `TraceId IN (SELECT TraceId FROM page)` via the IN-tuple dedup, so the heavy column is read for at most 2000 traces.
- Topic filter and search cursor move into the CTE's `HAVING` and now evaluate against the **latest version** of each trace (`argMax(..., UpdatedAt)`).
- Partition-key (`OccurredAt`) range kept on the page scan, the outer read, **and** the inner dedup subquery for pruning + a consistent "latest within range".
- The outer query does **not** re-filter `ComputedInput`: empty/null inputs are dropped downstream by `extractInputFromComputed`, while the raw page rows still drive `lastSort`/`returnedCount`, so the cursor tracks the page boundary and never strands older eligible traces behind a run of empty-input traces.

Cursor param names (`lastTs`, `lastTraceId`) are unchanged.

## Verification

- **Development repro (real ClickHouse 25.10, same version as prod):** with 30k traces carrying 4 KB `ComputedInput`, under a 120 MB budget the paged query completes and the pre-fix whole-set read exceeds the same budget; a `system.query_log` peak-memory comparison showed the paged read at a fraction of the pre-fix read.
- **Committed integration test** (`fetchTracesFromClickHouse.integration.test.ts`, testcontainers) exercises the real `fetchTracesFromClickHouse`:
  - a full page returns the 2000 newest traces, newest first;
  - the search cursor returns strictly older, non-overlapping traces;
  - when the newest traces have empty input, the cursor still reaches the oldest trace (`returnedCount` counts the whole page; only input-bearing traces are clustered).
  - It runs in a few seconds — the heavy memory-comparison case was dropped from CI to keep the integration shard well under its time cap; the memory result above stands from the development repro.

`pnpm typecheck` clean; topicClustering unit tests pass (incl. the cursor-params test); integration tests pass.

## Expected impact

The heavy-column read drops from "entire 12-month deduped set" to "≤2000 traces", removing the dominant memory and read-bytes cost. Expected: the ~20/24h `MEMORY_LIMIT_EXCEEDED` failures for this worker go to ~0, with lower p50 latency and read bytes per batch. Result set is equivalent (latest-version, newest-first; empty inputs dropped downstream as before).

## Please run before merging (I cannot — read-only prod access)

- `EXPLAIN PIPELINE` / `EXPLAIN actions=1` on the before/after query against prod ClickHouse to confirm `ComputedInput` is read only for the paged trace set.
- Compare `read_bytes` / peak memory in `system.query_log` for the two shapes on a real tenant.

Advisory PR — not merging/approving; a human should run EXPLAIN and decide.

### DIFF
--- a/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
+++ b/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
@@ -1,0 +1,203 @@
+/**
+ * Integration coverage for the topic-clustering trace fetch against a real
+ * ClickHouse.
+ *
+ * The pre-fix query read `ComputedInput` (a potentially large payload) for
+ * the entire deduped 12-month trace set before `ORDER BY ... LIMIT 2000`
+ * trimmed it. On busy tenants that materialised gigabytes of payload and the
+ * query died with MEMORY_LIMIT_EXCEEDED (observed in prod).
+ *
+ * The fix pages the 2000 most-recent trace keys first (lightweight columns
+ * only) and reads `ComputedInput` for that bounded set alone.
+ *
+ * Two layers here:
+ *  - correctness: `fetchTracesFromClickHouse` returns the right latest-version
+ *    traces, newest first, respecting the search cursor.
+ *  - memory: the paged query's peak memory (read from `system.query_log`) is a
+ *    fraction of the pre-fix whole-set read on the same data. Both queries run
+ *    to completion and drain — we measure memory rather than tripping an OOM,
+ *    so the test can't leave a half-read socket wedging the worker.
+ */
+
+import type { ClickHouseClient } from "@clickhouse/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { wrapWithDefaultSettings } from "~/server/clickhouse/safeClickhouseClient";
+import {
+  cleanupTestData,
+  getTestClickHouseClient,
+} from "../../event-sourcing/__tests__/integration/testContainers";
+import { fetchTracesFromClickHouse } from "../topicClustering";
+
+const TENANT_ID = "topic-fetch-test";
+const N_TRACES = 20_000;
+const INPUT_BYTES = 4096; // heavy payload — the column we must not read in bulk
+const twelveMonthsAgo = Date.now() - 365 * 24 * 60 * 60 * 1000;
+
+async function seedTraceSummaries(ch: ClickHouseClient) {
+  const now = Date.now();
+  const bigInput = JSON.stringify("x".repeat(INPUT_BYTES));
+  const BATCH = 2000;
+  let batch: Array<Record<string, unknown>> = [];
+  const flush = async () => {
+    if (!batch.length) return;
+    await ch.insert({
+      table: "trace_summaries",
+      values: batch,
+      format: "JSONEachRow",
+      clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+    });
+    batch = [];
+  };
+  for (let i = 0; i < N_TRACES; i++) {
+    batch.push({
+      ProjectionId: `proj-${nanoid()}`,
+      TenantId: TENANT_ID,
+      TraceId: `${TENANT_ID}-trace-${String(i).padStart(6, "0")}`,
+      Version: "v1",
+      Attributes: {},
+      // Spread across the last ~weeks; newest = lowest index.
+      OccurredAt: new Date(now - i * 60_000),
+      CreatedAt: new Date(now),
+      UpdatedAt: new Date(now - i),
+      ComputedIOSchemaVersion: "",
+      ComputedInput: bigInput,
+      ComputedOutput: "out",
+      TimeToFirstTokenMs: 1,
+      TimeToLastTokenMs: 1,
+      TotalDurationMs: 1,
+      TokensPerSecond: 1,
+      SpanCount: 1,
+      ContainsErrorStatus: 0,
+      ContainsOKStatus: 1,
+      ErrorMessage: null,
+      Models: ["gpt-5-mini"],
+      TotalCost: 0.01,
+      TokensEstimated: false,
+      TotalPromptTokenCount: 1,
+      TotalCompletionTokenCount: 1,
+      OutputFromRootSpan: 0,
+      OutputSpanEndTimeMs: 0,
+      BlockedByGuardrail: 0,
+      TopicId: i % 3 === 0 ? `topic-${i % 5}` : null,
+      SubTopicId: null,
+      HasAnnotation: null,
+    });
+    if (batch.length >= BATCH) await flush();
+  }
+  await flush();
+}
+
+/** Run a query to completion under a stable query_id and return its peak memory. */
+async function peakMemoryBytes(
+  ch: ClickHouseClient,
+  sql: string,
+): Promise<number> {
+  const queryId = `mem-${nanoid()}`;
+  const r = await ch.query({
+    query: sql,
+    query_params: { tenantId: TENANT_ID, twelveMonthsAgo },
+    format: "JSONEachRow",
+    query_id: queryId,
+  });
+  await r.json(); // fully drain
+  await ch.command({ query: "SYSTEM FLUSH LOGS" });
+  const log = await ch.query({
+    query: `
+      SELECT max(memory_usage) AS mem
+      FROM system.query_log
+      WHERE query_id = {queryId:String} AND type = 'QueryFinish'`,
+    query_params: { queryId },
+    format: "JSONEachRow",
+  });
+  const rows = (await log.json()) as Array<{ mem: string }>;
+  return Number(rows[0]?.mem ?? 0);
+}
+
+const PAGED_SQL = `
+  WITH page AS (
+    SELECT TraceId FROM trace_summaries
+    WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
+    GROUP BY TenantId, TraceId
+    ORDER BY argMax(OccurredAt, UpdatedAt) DESC, TraceId ASC
+    LIMIT 2000)
+  SELECT t.TraceId, t.ComputedInput, t.TopicId, t.SubTopicId, toString(toUnixTimestamp64Milli(t.OccurredAt)) AS OccurredAtMs
+  FROM trace_summaries t
+  WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
+    AND ComputedInput IS NOT NULL AND ComputedInput != ''
+    AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
+      SELECT TenantId, TraceId, max(UpdatedAt) FROM trace_summaries
+      WHERE TenantId = {tenantId:String} AND TraceId IN (SELECT TraceId FROM page)
+      GROUP BY TenantId, TraceId)
+  ORDER BY t.OccurredAt DESC, t.TraceId ASC LIMIT 2000`;
+
+const PRE_FIX_SQL = `
+  SELECT t.TraceId, t.ComputedInput, t.TopicId, t.SubTopicId, toString(toUnixTimestamp64Milli(t.OccurredAt)) AS OccurredAtMs
+  FROM trace_summaries t
+  WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
+    AND ComputedInput IS NOT NULL AND ComputedInput != ''
+    AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
+      SELECT TenantId, TraceId, max(UpdatedAt) FROM trace_summaries
+      WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
+        AND ComputedInput IS NOT NULL AND ComputedInput != ''
+      GROUP BY TenantId, TraceId)
+  ORDER BY t.OccurredAt DESC, t.TraceId ASC LIMIT 2000`;
+
+describe("fetchTracesFromClickHouse integration", () => {
+  let ch: ClickHouseClient;
+
+  beforeAll(async () => {
+    const raw = getTestClickHouseClient();
+    if (!raw) throw new Error("ClickHouse client not available");
+    ch = wrapWithDefaultSettings(raw);
+    await seedTraceSummaries(ch);
+  }, 180_000);
+
+  afterAll(async () => {
+    await cleanupTestData(TENANT_ID);
+  });
+
+  describe("when fetching a full batch", () => {
+    it("returns the 2000 newest traces, newest first, all with input", async () => {
+      const res = await fetchTracesFromClickHouse(ch, TENANT_ID, false, [], []);
+
+      expect(res.traces).toHaveLength(2000);
+      expect(res.returnedCount).toBe(2000);
+      expect(res.traces.every((t) => t.input.length > 0)).toBe(true);
+      expect(res.traces[0]?.trace_id).toBe(`${TENANT_ID}-trace-000000`);
+      expect(res.lastSort).toBeDefined();
+    });
+  });
+
+  describe("when the search cursor advances", () => {
+    it("returns strictly older, non-overlapping traces", async () => {
+      const first = await fetchTracesFromClickHouse(ch, TENANT_ID, false, [], []);
+      const second = await fetchTracesFromClickHouse(
+        ch,
+        TENANT_ID,
+        false,
+        [],
+        [],
+        first.lastSort!,
+      );
+
+      expect(second.traces.length).toBeGreaterThan(0);
+      const firstIds = new Set(first.traces.map((t) => t.trace_id));
+      expect(second.traces.some((t) => firstIds.has(t.trace_id))).toBe(false);
+    });
+  });
+
+  describe("when comparing peak memory against the pre-fix shape", () => {
+    it("reads the heavy column for the page only, not the whole window", async () => {
+      const paged = await peakMemoryBytes(ch, PAGED_SQL);
+      const preFix = await peakMemoryBytes(ch, PRE_FIX_SQL);
+
+      // The paged read touches ComputedInput for <=2000 traces; the pre-fix
+      // shape touches it for the entire deduped window. On this seed the gap
+      // is large — assert at least a 2x reduction to stay robust to noise.
+      expect(paged).toBeGreaterThan(0);
+      expect(preFix).toBeGreaterThan(0);
+      expect(paged * 2).toBeLessThan(preFix);
+    });
+  });
+});
--- a/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
+++ b/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
@@ -124,10 +124,10 @@ const PAGED_SQL = `
   SELECT t.TraceId, t.ComputedInput, t.TopicId, t.SubTopicId, toString(toUnixTimestamp64Milli(t.OccurredAt)) AS OccurredAtMs
   FROM trace_summaries t
   WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
-    AND ComputedInput IS NOT NULL AND ComputedInput != ''
     AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
       SELECT TenantId, TraceId, max(UpdatedAt) FROM trace_summaries
-      WHERE TenantId = {tenantId:String} AND TraceId IN (SELECT TraceId FROM page)
+      WHERE TenantId = {tenantId:String} AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64}) AND OccurredAt < now64(3)
+        AND TraceId IN (SELECT TraceId FROM page)
       GROUP BY TenantId, TraceId)
   ORDER BY t.OccurredAt DESC, t.TraceId ASC LIMIT 2000`;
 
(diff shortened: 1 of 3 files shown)


langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +236,74 @@ async function fetchTracesFromClickHouse(
...
+        AND OccurredAt < now64(3)
+        AND ComputedInput IS NOT NULL
+        AND ComputedInput != ''
         AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
           SELECT TenantId, TraceId, max(UpdatedAt)
           FROM trace_summaries
-          WHERE ${baseWhereClause}
+          WHERE TenantId = {tenantId:String}
+            AND TraceId IN (SELECT TraceId FROM page)
           GROUP BY TenantId, TraceId
```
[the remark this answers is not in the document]
langwatch-agent: Good catch — fixed in 3be01c82b. Added the same `OccurredAt` window (`>= twelveMonthsAgo AND < now64(3)`) to the inner `max(UpdatedAt)` dedup subquery so it prunes partitions and computes the latest version within the requested range.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +236,74 @@ async function fetchTracesFromClickHouse(
...
+        SELECT TraceId
+        FROM trace_summaries
+        WHERE TenantId = {tenantId:String}
+          AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+          AND OccurredAt < now64(3)
+        GROUP BY TenantId, TraceId
+        ${pageHavingClause}
+        ORDER BY argMax(OccurredAt, UpdatedAt) DESC, TraceId ASC
+        LIMIT 2000
+      )
```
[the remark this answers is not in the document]
langwatch-agent: Fixed in 3be01c82b. The outer query no longer filters `ComputedInput`, so the raw page rows (and thus `lastSort` / `returnedCount`) track the page boundary; empty/null inputs are still dropped downstream by `extractInputFromComputed`. This stops the cursor from advancing by the surviving subset and stranding older eligible traces. Added an integration test where the newest traces have empty input and asserts the cursor still reaches the oldest trace.

langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      const res = await fetchTracesFromClickHouse(ch, EMPTY_TENANT, false, [], []);
+
+      // Only the 20 input-bearing traces are clustered...
+      expect(res.traces).toHaveLength(20);
+      // ...but pagination accounts for the whole page (incl. the 20 empties),
+      // so the cursor reaches the oldest trace and clustering won't stall.
+      expect(res.returnedCount).toBe(40);
+      expect(res.lastSort?.[1]).toBe(`${EMPTY_TENANT}-trace-0039`);
+    });
+  });
```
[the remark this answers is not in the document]
langwatch-agent: Good call — fixed in fa9956c8d. The case now seeds a full page (2000) of empty-input traces plus 20 older input-bearing ones, then asserts page 1 clusters nothing but reports `returnedCount=2000` with `lastSort` at the page boundary (trace-001999), and page 2 — seeked from that cursor — reaches the older input-bearing traces (trace-002000…). That's exactly the scenario the pre-fix SQL `ComputedInput` filter would have stranded. Still runs in ~4s.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +236,81 @@ async function fetchTracesFromClickHouse(
...
+  // for those <=2000 traces alone — never across the whole 12-month window,
+  // which is what tipped this query into MEMORY_LIMIT_EXCEEDED. The topic
+  // and cursor predicates run against the latest version of each trace
+  // (argMax over UpdatedAt), so they live in the CTE's HAVING.
+  //
+  // The outer query deliberately does NOT filter on ComputedInput: empty /
+  // null inputs are dropped downstream by `extractInputFromComputed`, while
+  // the raw rows still carry the full page so `lastSort` (the pagination
+  // cursor) tracks the page boundary. Filtering here would advance the cursor
+  // by the surviving subset and could strand older eligible traces behind a
```
rogeriochaves, line 260: **This documented intent is defeated by the caller.** The page-boundary cursor here is correct, but `clusterTopicsForProject` returns at `if (traces.length < minimumTraces)` (line 136) *before* the `returnedCount > 10 && lastSort` scheduling block (line 151). A full page of empty-input traces (`returnedCount=2000`, `traces.length=0`) hits that early return and never schedules the next page, so the cursor never advances and older traces are stranded — the same regression this comment says the design avoids. See the summary for the full walk-through and fix direction. (The new integration test doesn't cover this because it threads `lastSort` manually rather than going through the caller.)

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +236,81 @@ async function fetchTracesFromClickHouse(
...
+        AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+        AND OccurredAt < now64(3)
         AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
           SELECT TenantId, TraceId, max(UpdatedAt)
           FROM trace_summaries
-          WHERE ${baseWhereClause}
+          WHERE TenantId = {tenantId:String}
+            AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+            AND OccurredAt < now64(3)
+            AND TraceId IN (SELECT TraceId FROM page)
```
rogeriochaves, line 322: **Behavior change worth confirming: dedup now runs over all versions, not just non-empty ones.** The pre-fix inner dedup used `baseWhereClause`, which included `ComputedInput IS NOT NULL AND ComputedInput != ''`, so `max(UpdatedAt)` was taken over non-empty versions only. Now it's taken over every version. For a trace whose newest version has empty `ComputedInput` but an older version carried input, the old query surfaced the older non-empty version (clustered); the new query selects the newest empty version, which `extractInputFromComputed` drops — so that trace is no longer clustered. `OccurredAt` is a running `min` and isn't part of the version key, so versions genuinely differ. This makes the fetch consistent with the counts query (which already dedups over all versions), so it's likely intended — but it's not called out in the comment and is a real semantic shift. Confirm it's acceptable, or restore the non-empty filter inside the dedup if older non-empty versions should still win.

rogeriochaves (commented):
Code review (automated, high-effort recall pass). The SQL rewrite itself is sound — the page-boundary cursor, the argMax-over-UpdatedAt predicates, and the bounded ComputedInput read all check out, and I verified that `now64()` is evaluated once per query and that unused bound params are harmless (both non-issues). But there's one **blocking correctness bug**, plus a behavior change and a robustness note.

## 🔴 Blocking: the caller defeats this fix, so the cursor still stalls on a full empty-input page

The whole point of keeping empty-input rows in the page (so `lastSort` tracks the page boundary) is to advance past a run of empty-input traces. But the caller `clusterTopicsForProject` short-circuits before it ever schedules the next page:

```ts
// topicClustering.ts
const minimumTraces = isIncrementalProcessing ? 1 : 10;   // :124
...
if (traces.length < minimumTraces) {                       // :136  <-- gates on the FILTERED count
  logger.info(..., `less than ${minimumTraces} traces found ... skipping`);
  return;                                                   // :141  returns here
}
...
if (returnedCount > 10 && lastSort) {                      // :151  next-page scheduling, never reached
  await scheduleTopicClusteringNextPage(projectId, lastSort);
}
```

When a full page comes back with `returnedCount === 2000` but every row has empty input (`traces.length === 0`), `0 < minimumTraces` is true → it returns at :141, before the `returnedCount > 10` scheduling block at :151. `scheduleTopicClusteringNextPage` is the only thing that enqueues the next page (the worker calls `clusterTopicsForProject` once per job, no outer loop), so the cursor is dropped and **every older trace beyond that page is never clustered** — the exact stranding regression this PR set out to fix, just relocated from SQL to the caller. For a busy tenant with 2000+ empty-input traces newer than its real traces (precisely the population targeted here), batch clustering stalls at page 1.

The new integration test does not catch this because it calls `fetchTracesFromClickHouse` directly and manually threads `page1.lastSort` into the page-2 call — it proves the SQL reaches older traces *if* the cursor is propagated, but never exercises `clusterTopicsForProject`'s scheduling decision, so it stays green even though production never propagates the cursor here.

**Fix direction:** schedule the next page based on `returnedCount > 10 && lastSort` independently of the `traces.length < minimumTraces` skip — e.g. move/duplicate the scheduling so it runs even when this page yielded too few usable traces to cluster. Worth a caller-level test that asserts `scheduleTopicClusteringNextPage` is called for a full empty page.

Findings #2 and #3 are inline below.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +236,81 @@ async function fetchTracesFromClickHouse(
...
         t.ComputedInput AS ComputedInput,
         t.TopicId AS TopicId,
         t.SubTopicId AS SubTopicId,
         toString(toUnixTimestamp64Milli(t.OccurredAt)) AS OccurredAtMs
       FROM trace_summaries t
-      WHERE ${whereClause}
+      WHERE TenantId = {tenantId:String}
+        AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+        AND OccurredAt < now64(3)
         AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
```
rogeriochaves, line 316: **Robustness: `(TenantId, TraceId, UpdatedAt) IN (… max(UpdatedAt) …)` can emit duplicate rows.** If two physical rows for the same trace share an identical `max(UpdatedAt)` (e.g. two writes in the same millisecond before the ReplacingMergeTree merge collapses them — `UpdatedAt` defaults to `now64(3)`), both match the tuple, so the trace appears twice. That inflates `returnedCount` and, because those duplicates can carry different `OccurredAt` values (OccurredAt isn't in the version key), the outer `ORDER BY t.OccurredAt` can diverge from the page CTE's `argMax(OccurredAt, UpdatedAt)` ordering and make the boundary cursor skip or repeat a few traces. This pattern is pre-existing (the counts query and the old query share it), so not a regression — but `returnedCount`/`lastSort` now drive paging, so it's more load-bearing here. A `LIMIT 1 BY (TenantId, TraceId)` on the outer query would neutralize both the duplicate rows and the boundary drift in one stroke.

langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      await cleanupTestData(EMPTY_TENANT);
+    });
+
+    it("advances the cursor past a full empty page to reach older traces", async () => {
+      const page1 = await fetchTracesFromClickHouse(ch, EMPTY_TENANT, false, [], []);
+
+      // Page 1 is a full page of empty-input traces: nothing to cluster, but
+      // the cursor must still track the page boundary (the 2000th trace).
+      expect(page1.traces).toHaveLength(0);
+      expect(page1.returnedCount).toBe(2000);
```
rogeriochaves, line 164: This proves the SQL reaches older traces *when the cursor is propagated*, but it threads `page1.lastSort` into the page-2 call by hand. In production `clusterTopicsForProject` never propagates the cursor for a full empty page (it returns at the `traces.length < minimumTraces` guard before scheduling) — so this test passes while the real fix is still broken. Consider a caller-level test that asserts `scheduleTopicClusteringNextPage` is invoked when a full page yields zero usable traces.

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Fixed in 91111fd16 — this was the real bug, thank you. `clusterTopicsForProject` now schedules the next page in both branches via a shared `maybeScheduleNextPage()` (gated on `returnedCount > 10 && lastSort`), so a full page of empty-input traces (`returnedCount=2000, traces.length=0`) still advances the cursor instead of returning early at the `minimumTraces` guard. Added a caller-level unit test asserting `scheduleTopicClusteringNextPage` is invoked in exactly that case.

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Confirmed and intended — calling it out in the description. The dedup can only read lightweight key columns: putting `ComputedInput IS NOT NULL/!= ''` back into the `max(UpdatedAt)` subquery is exactly what reads the heavy column across the whole window and reintroduces the OOM this PR fixes. So we dedup over all versions and take the latest, then drop it downstream if its input is empty. Semantically that's "cluster the trace's current input" — if the newest version cleared `ComputedInput`, the trace currently has nothing to cluster. The only regressing case is newest-empty-but-older-non-empty, which is rare for a computed projection; I think the latest-version semantics are the more correct behaviour here, and the memory win makes the trade-off worth it. Happy to revisit if you'd rather preserve the old "latest non-empty version" behaviour, but it can't be done without the heavy dedup read.

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Pre-existing and protected by an invariant, so leaving as-is. Per `dev/docs/best_practices/clickhouse-queries.md` ("UpdatedAt is Monotonically Increasing"), the event-sourcing fold sets `UpdatedAt = max(Date.now(), prevUpdatedAt + 1)`, so every version gets a unique, strictly increasing `UpdatedAt` — no ties, so `max(UpdatedAt)` identifies exactly one row and the IN-tuple can't emit duplicates. The `now64(3)` table default only applies to rows written outside the projection path; normal writes always carry the explicit monotonic value. This is the same IN-tuple pattern the counts query and the rest of the codebase already rely on, so it's out of scope for this PR.

langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
langwatch-agent: Addressed in 91111fd16. The integration test still proves the SQL reaches older traces when the cursor is propagated, and the caller bug it pointed at is now fixed (see the line-260 thread) plus covered by a new caller-level unit test in `topicClustering.unit.test.ts` ("schedules the next page when a full page yields zero usable traces"), which asserts `scheduleTopicClusteringNextPage` is invoked for a full empty page rather than threading `lastSort` by hand.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -133,11 +133,34 @@ export const clusterTopicsForProject = async (
...
+    await scheduleTopicClusteringNextPage(projectId, lastSort);
+  };
+
   if (traces.length < minimumTraces) {
     logger.info(
       { projectId },
-      `less than ${minimumTraces} traces found for project, skipping topic clustering`,
+      `less than ${minimumTraces} usable traces on this page, skipping clustering but still paging`,
     );
+    await maybeScheduleNextPage();
```
rogeriochaves, line 162: ✅ This is the fix for the P1 cursor-stall — scheduling now runs in the skip branch, so a full empty-input page advances the cursor instead of stranding older traces. Verified the new unit test asserts the exact page-boundary cursor `[now - 14*1000, "trace-14"]` is scheduled. Looks correct.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +245,81 @@ async function fetchTracesFromClickHouse(
...
         t.TopicId AS TopicId,
         t.SubTopicId AS SubTopicId,
         toString(toUnixTimestamp64Milli(t.OccurredAt)) AS OccurredAtMs
       FROM trace_summaries t
-      WHERE ${whereClause}
+      WHERE TenantId = {tenantId:String}
+        AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+        AND OccurredAt < now64(3)
         AND (t.TenantId, t.TraceId, t.UpdatedAt) IN (
           SELECT TenantId, TraceId, max(UpdatedAt)
```
rogeriochaves, line 317: 🟡 **P3 (unchanged, likely intentional — just confirm):** this inner dedup still computes `max(UpdatedAt)` over all versions (no `ComputedInput` filter, unlike the pre-fix query). So a trace whose newest version has empty input but an older version carried input is now dropped, where it used to be clustered via the older version. This makes the fetch consistent with the counts query, so it's probably the intended semantics — flagging only so it's a conscious choice rather than an accident. No change needed if intended.

langwatch/src/server/topicClustering/topicClustering.ts
```diff
@@ -236,57 +245,81 @@ async function fetchTracesFromClickHouse(
...
           FROM trace_summaries
-          WHERE ${baseWhereClause}
+          WHERE TenantId = {tenantId:String}
+            AND OccurredAt >= fromUnixTimestamp64Milli({twelveMonthsAgo:UInt64})
+            AND OccurredAt < now64(3)
+            AND TraceId IN (SELECT TraceId FROM page)
           GROUP BY TenantId, TraceId
         )
       ORDER BY t.OccurredAt DESC, t.TraceId ASC
```
rogeriochaves, line 325: 🟡 **P3 (unchanged, pre-existing, optional):** `(TenantId, TraceId, UpdatedAt) IN (… max(UpdatedAt) …)` can match two rows if a trace has two versions sharing the same millisecond `UpdatedAt`, producing a duplicate row that inflates `returnedCount` and (since those duplicates can carry different `OccurredAt`) can nudge the boundary cursor. This pattern predates the PR (the counts query has it too), so not a regression — but `returnedCount`/`lastSort` are now load-bearing for paging, so a `LIMIT 1 BY (TenantId, TraceId)` on the outer query would make both robust. Low priority.

rogeriochaves (commented):
Re-review after the latest commits. Severity key: 🔴 P1 · 🟠 P2 · 🟡 P3.

## ✅ P1 (cursor-stall) — RESOLVED, correctly

`clusterTopicsForProject` now extracts `maybeScheduleNextPage()` and calls it inside the `traces.length < minimumTraces` branch *before* returning, so a full page that yields zero usable traces (`returnedCount=2000`, `traces.length=0`) advances the cursor instead of stalling. Scheduling is correctly gated on `returnedCount > 10 && lastSort` (the page boundary), not the post-filter usable count — so it keeps paging through runs of empty-input traces and stops naturally on a non-full page. And the new unit test `"schedules the next page when a full page yields zero usable traces"` covers exactly the caller-level gap I flagged: it feeds a full empty-input page and asserts `scheduleTopicClusteringNextPage` is invoked with the page-boundary cursor. 👍

The two minor items from the first pass are unchanged and both optional — details inline.

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Thanks for verifying! 🙏

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Confirmed intentional, and called out in the PR description. Putting `ComputedInput IS NOT NULL/!= ''` back into the inner dedup is exactly what reads the heavy column across the whole window and reintroduces the OOM, so we dedup on lightweight key columns only and take the latest version (then drop it downstream if its input is empty). As you note, this also makes the fetch consistent with the counts query. The only regressing case is newest-empty-but-older-non-empty, which is rare for a computed projection — "cluster the trace's current input" is the more correct semantics here. Conscious choice. 👍

langwatch/src/server/topicClustering/topicClustering.ts
langwatch-agent: Done in a8c76155c. Added `LIMIT 1 BY TraceId` on the outer query (single-tenant scope via `WHERE TenantId = ...`, so TraceId is a sufficient key) right after the `ORDER BY`, so any duplicate from two versions sharing an identical `max(UpdatedAt)` collapses to one row before `LIMIT 2000`. It runs on the already-bounded ≤2000-row deduped set, so no extra heavy-column scan, and it makes `returnedCount`/`lastSort` robust now that they drive paging. Integration + unit tests still green.

langwatch-agent: Correction on my previous reply: I initially added the per-key SQL dedup operator, but that turns out to be banned in this query path for OOM safety — there's a guard test (`trace-dedup-oom-safety.unit.test.ts`) asserting this exact function never uses it, because it materialises heavy columns for the whole granule. So I reverted that and moved the de-duplication into JS instead (e4b751cacac747f45b9b389d18959cd9852578ae): collapse any duplicate `TraceId` over the already-bounded ≤2000-row page (rows arrive `OccurredAt DESC, TraceId ASC`, so the first per TraceId is kept). Same robustness for `returnedCount`/`lastSort`, no heavy-scan risk. Added a unit test covering duplicate-row collapse, and the OOM-safety guard is green.

rogeriochaves (approved):

rogeriochaves (approved):

rogeriochaves (approved):

rogeriochaves (approved):

PR #4429, opened 2026-05-30
fix(model-providers): collapse Save Advanced into single Save, gate on AI gateway flag

rogeriochaves:
## Summary

The Advanced (Gateway) block on the ModelProvider drawer was:

1. Rendered as a flat section even for orgs without the AI gateway.
2. Persisted via its own separate "Save Advanced" mutation (\`api.modelProvider.updateAdvanced\`).

The split read as backend smell because the fields already live on the same \`ModelProvider\` row (since the iter-110 binding collapse). The second button only existed because the unified \`update\` mutation did not yet carry them.

## What changed

- Fold the five advanced fields (\`rateLimitRpm/Tpm/Rpd\`, \`fallbackPriorityGlobal\`, \`providerConfig\`) into the \`update\` zod input and route them through the existing service + repository writes. One Prisma round-trip persists basic + advanced together.
- Drop the dedicated \`updateAdvanced\` tRPC procedure + its service method + its unit tests. They were the only callers and become dead once the drawer uses unified \`update\`.
- Rewrite \`ModelProviderAdvancedSection\` as a controlled, button-less Chakra Accordion. The parent form owns the draft + jsonError + parses on submit. Collapsed by default so first-time setup never sees the gateway knobs.
- Gate the accordion on \`release_ui_ai_gateway_menu_enabled\` for the caller's org (matches GovernSection). Orgs without the gateway never render the section and never spread advanced fields into the payload.

## Coverage

Added 3 \`@integration\` scenarios in \`specs/ai-gateway/gateway-provider-settings.feature\` covering FF-off hidden, FF-on collapsed accordion, and single-Save sending both payloads. Bound in \`ModelProviderForm.advanced-gateway.integration.test.tsx\` (6 tests, all green).

## Test plan

- [x] \`pnpm typecheck\` clean.
- [x] \`pnpm check:feature-parity\` clean.
- [x] New advanced-gateway integration suite green (6 tests).
- [x] Pre-existing azure-safety + router-load + form-submit suites still green.
- [ ] Live browser QA pending — no dev server was running locally; render tests cover the FF gating + collapsed-by-default behavior. Happy to drive a browser pass once CI is green if reviewers want it.

## Browser QA

Drove the drawer in Chromium against a local dogfood org with the AI gateway flag forced on.

**Collapsed by default — single Save button (no more `Save Advanced`):**

collapsed

**Expanded — all five gateway fields + read-only Health block:**

expanded

End-to-end save round-trip not captured in the screenshot because the existing form rejects the masked-API-key placeholder on re-submit unless the user retypes the key (pre-existing behavior, not a regression from this PR). The wire-level test in `useProviderFormSubmit.integration.test.tsx` covers the unified payload shape.

**Drawer width bumped `md`→`xl` (commit `585604ddd`)** — long bedrock/managed model IDs were overflowing the `md` drawer. New width matches AddOrEditDatasetDrawer precedent.

drawer-xl


langwatch/src/components/settings/__tests__/ModelProviderForm.advanced-gateway.integration.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+      });
+
+      /** @scenario Single Save persists basic credentials and advanced gateway fields together */
+      it("renders only one Save button (no separate Save Advanced)", () => {
+        expect(screen.queryByText(/save advanced/i)).toBeNull();
+        const saveButtons = screen.getAllByRole("button", {
+          name: /^save$/i,
+        });
+        expect(saveButtons).toHaveLength(1);
+      });
```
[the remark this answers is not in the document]
rogeriochaves: Addressed in 607ba0695: dropped the @scenario annotation from this UI test (button-count assertion). The wire-level scenario binding now lives on useProviderFormSubmit.integration.test.tsx where the actual basic+advanced payload spread is asserted.

langwatch/src/server/modelProviders/modelProvider.repository.ts
```diff
@@ -288,6 +320,11 @@ export class ModelProviderRepository {
           customEmbeddingsModels: data.customEmbeddingsModels as
             | Prisma.InputJsonValue
             | undefined,
+          ...(providerConfig !== undefined && {
+            providerConfig: (providerConfig ?? undefined) as
+              | Prisma.InputJsonValue
+              | undefined,
+          }),
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 607ba0695 alongside the create() path. Same Prisma.JsonNull treatment so null actually clears providerConfig.

specs/ai-gateway/gateway-provider-settings.feature
```diff
@@ -62,6 +62,39 @@ Feature: AI Gateway — Provider settings cohesion
...
+  @integration
+  Scenario: Single Save persists basic credentials and advanced gateway fields together
+    Given a ModelProvider "openai" exists scoped to ORGANIZATION "acme"
+    And the "release_ui_ai_gateway_menu_enabled" flag is enabled for "acme"
+    When I open the ModelProvider drawer for "openai"
+    And I expand the "Advanced (Gateway)" accordion
+    And I set rateLimitRpm=600 and providerConfig={"region":"us-east-1"}
+    And I click the drawer's "Save" button
+    Then one `api.modelProvider.update` mutation fires carrying both the basic
+      and advanced fields
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 607ba0695: replaced the api.modelProvider.update mention with user-observable outcomes (drawer closes, single Save persists both basic and advanced).

langwatch/src/server/modelProviders/modelProvider.repository.ts
```diff
@@ -211,6 +217,26 @@ export class ModelProviderRepository {
...
+          fallbackPriorityGlobal: data.fallbackPriorityGlobal,
+        }),
+        ...(data.rotationPolicy !== undefined && {
+          rotationPolicy: data.rotationPolicy,
+        }),
+        ...(data.providerConfig !== undefined && {
+          providerConfig: (data.providerConfig ?? undefined) as
+            | Prisma.InputJsonValue
+            | undefined,
+        }),
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 607ba0695. Real bug — replaced data.providerConfig ?? undefined with Prisma.JsonNull so an explicit null clears the column. Matches the agent.repository pattern.

langwatch/src/server/modelProviders/modelProvider.service.ts
```diff
@@ -530,6 +502,23 @@ export class ModelProviderService {
...
     }
 
+    // Advanced (Gateway) writes also require manage on every scope the
+    // existing row is bound to — not just the project the caller is in.
+    // Matches the previous `updateAdvancedSettings` contract: a project
+    // admin must not nudge rate limits on a credential that's also
+    // bound to its parent org/team without manage there. The basic
+    // update path keeps its existing semantics (project:update gate
+    // only) so this PR doesn't tighten unrelated writes.
+    if (ctx && hasAdvancedWrite && existingProvider) {
```
rogeriochaves, line 524: 🟠 **P2 — this advanced-write authz gate is now untested at the service layer.** This PR deletes `modelProvider.updateAdvanced.unit.test.ts`, which covered the old contract directly (unreadable→NOT_FOUND, can-read-but-not-manage→FORBIDDEN with no write, can-manage→write). Nothing replaces it for this branch: `modelProvider.authz.integration.test.ts` never sends advanced fields (so `hasAdvancedWrite` is always false there), and the new coverage is frontend-mocked (`useProviderFormSubmit.integration.test.tsx`) plus the router-load smoke test. The `assertCanManageAllScopes(existingProvider.scopes)` guard here is the security boundary for advanced writes — please add a service-level test that sends an advanced field and asserts it throws (and performs no write) when the caller can't manage every existing scope.

langwatch/src/components/settings/ModelProviderForm.tsx
```diff
@@ -104,6 +129,62 @@ export const EditModelProviderForm = ({
...
+  // opened, multi-instance switch, etc.). Skipped when the gateway flag
+  // is off so we never spread defaults into the payload from a hidden
+  // section.
+  useEffect(() => {
+    if (!gatewayMenuEnabled) {
+      setAdvancedDraft(EMPTY_ADVANCED_DRAFT);
+      setAdvancedJsonError(null);
+      return;
+    }
+    setAdvancedDraft(draftFromProvider(advancedSeed));
```
rogeriochaves: 🟠 **P2 — this reset can clobber unsaved edits on a background refetch.** `advancedSeed` is memoized over the five gateway values pulled from the live `useModelProvidersSettings` (→ `getAllForProjectForFrontend`) query. If that query refetches while the drawer is open and any of those values changed server-side (window-focus refetch — `refetchOnWindowFocus` defaults to true — a concurrent edit, or another tab), this effect re-fires and overwrites the user's in-progress advanced draft (and clears `advancedJsonError`) with no warning → silent loss of typed rate limits / JSON. The basic-field reset in `useModelProviderForm.ts` has the same shape, so this mirrors a pre-existing pattern rather than introducing a new class of bug — but the advanced draft now rides along. Consider guarding the reset on a dirty flag, or only seeding once per drawer-open / provider-id change.

langwatch/src/hooks/useProviderFormSubmit.ts
```diff
@@ -245,6 +262,18 @@ export function useProviderFormSubmit({
         .map(({ key, value }) => ({ key, value }));
 
       const trimmedName = (name ?? "").trim();
+      let advancedPayload: AdvancedGatewayPayload | null = null;
+      if (getAdvancedPayload) {
+        try {
+          advancedPayload = getAdvancedPayload();
```
rogeriochaves, line 268: 🟠 **P2 — malformed providerConfig JSON fails silently when the accordion is collapsed.** This early-return branch calls `onError?.(err)` and returns without toasting (unlike the general `catch` lower down). The form doesn't wire `onError`, and the only feedback — the inline `advancedJsonError` — renders inside the accordion content, which is collapsed by default. Net effect: user types bad JSON, collapses the accordion, hits Save → the button stops spinning, the drawer stays open, and nothing visible tells them why. Either toast in this branch or auto-expand the accordion when `getAdvancedPayload` throws.

langwatch/src/components/settings/ModelProviderAdvancedSection.tsx
```diff
@@ -41,107 +56,87 @@ function jsonToInput(value: unknown): string {
...
+  fallbackPriorityGlobal: number | null;
+  providerConfig: Record<string, unknown> | null;
+}
+
+export function parseAdvancedDraft(
+  draft: ModelProviderAdvancedDraft,
+): ParsedAdvancedPayload {
+  let providerConfig: Record<string, unknown> | null = null;
+  if (draft.providerConfigJson.trim()) {
+    providerConfig = JSON.parse(draft.providerConfigJson) as Record<
```
rogeriochaves: 🟡 **P3 — `parseAdvancedDraft` doesn't check the parsed value is a plain object.** `JSON.parse(...) as Record<string, unknown>` accepts valid-but-non-object JSON like `[1,2]`, `42`, or `"x"` without throwing, so no inline error fires and the client sends it. The server schema `z.object({}).passthrough()` then rejects it with a cryptic zod message ("Expected object, received array") surfaced via the generic failure toast. Add a `typeof parsed === "object" && !Array.isArray(parsed)` guard here so the user gets a clean inline "Provider config must be a JSON object" instead of a round-trip server error.

rogeriochaves (commented):
Code review (automated, high-effort recall pass). Severity key: 🔴 P1 blocking · 🟠 P2 should-fix · 🟡 P3 minor.

**Good news on the security-sensitive part:** folding `updateAdvanced` into the unified `update` preserves the authz contract. The advanced-write manage gate (`assertCanManageAllScopes(existingProvider.scopes)`) is live (the router passes `ctx`), and `findExistingProvider` resolves the row through the PROJECT-scope-gated `findById`, so a cross-tenant `id` never reaches the mutate path — it falls to create in the caller's own project. No cross-tenant escalation, no NOT_FOUND→FORBIDDEN enumeration oracle. Dropping `rotationPolicy` is also safe: the enum has a single value (`MANUAL`) and the column defaults to it, so it was a no-op — dead-code removal.

No P1s. Findings inline; two more minor notes below.

**🟡 P3 — edit-of-a-vanished-row silently becomes a create.** If an update (advanced or basic) targets an `id` that `findById(id, projectId)` can't resolve (deleted concurrently, or not project-scoped), `existingProvider` is null → it runs `createNew` and creates a *new* provider in the caller's project instead of erroring. The old `updateAdvanced` returned NOT_FOUND. Minor idempotency/UX quirk, not a security issue.

**🟡 P3 (verified safe, no action) — flag-off data safety + loading state.** When the gateway flag is off (or still loading, which reads as off), `getAdvancedPayload` returns null and the submit spreads no advanced fields, so existing rate limits are preserved rather than wiped. Confirmed correct — noting it because it's the kind of thing that's easy to regress.

langwatch/src/server/modelProviders/modelProvider.service.ts
rogeriochaves: Fixed in 7c55270e5. Added integration test in modelProvider.authz.integration.test.ts: team admin without org:manage attempting rateLimitRpm write via id-only → FORBIDDEN + rateLimit unchanged. Plus orgAdmin happy-path counter-test. Bound to spec scenario.

langwatch/src/components/settings/ModelProviderForm.tsx
rogeriochaves: Fixed in 7c55270e5. Re-keyed the reset effect on provider.id only; background refetches (window-focus, invalidation, concurrent edit) no longer clobber the in-progress draft. Drawer-open / multi-instance switch still seeds.

langwatch/src/hooks/useProviderFormSubmit.ts
rogeriochaves: Fixed in 7c55270e5. Auto-expand on bad JSON: getAdvancedPayload now sets advancedAccordionValue=['advanced-gateway'] before throwing, so the inline parse error is always visible. Accordion is now controlled-state from the form.

langwatch/src/components/settings/ModelProviderAdvancedSection.tsx
rogeriochaves: Fixed in 7c55270e5. parseAdvancedDraft now rejects non-object JSON (array/scalar/null) with 'Provider config must be a JSON object' before send.

rogeriochaves:
Sweep on the 3 items that lived in review bodies (no inline threads), addressed in 93c754958:

**🟡 P3 — edit-of-a-vanished-row silently becomes a create** (from @rogeriochaves recall review, 15:10:03Z)
Fixed. `service.updateModelProvider` now throws `TRPCError NOT_FOUND` when `input.id` is set and `existingProvider` is null, instead of falling through to `createNew`. The create path remains the `!input.id` branch only. Regression test: `'Update of a vanished id surfaces NOT_FOUND instead of silently creating'` in `modelProvider.authz.integration.test.ts`.

**🟡 Nitpick — `default_llm` consolidation in stripUnsupportedLLMParams.ts** (from @coderabbitai, 15:34:17Z)
Fixed. `default_llm` branch now uses `replaceObjectContents` (same helper as the node paths). `pruneRemovedKeys` deleted. Single filter pass everywhere.

**🟡 P3 — flag-off data safety + loading state** (from @rogeriochaves, 15:10:03Z, self-marked "verified safe, no action")
Confirmed safe, no code change. Covered by:
- `ModelProviderAdvancedSection` is not rendered when the gateway flag is off (integration scenario `'Advanced (Gateway) is hidden when the AI gateway feature flag is off'` in `gateway-provider-settings.feature`).
- Loading state: drawer hides the advanced inputs behind the "Save the provider first…" helper when `modelProviderId` is undefined (`ModelProviderAdvancedSection.tsx:184`).

39/39 model-provider unit tests green, typecheck clean.

PR #4432, opened 2026-05-31
perf(clickhouse): cap single-trace span-read memory to contain OOM blast radius

langwatch-agent:
## Evidence

The single-trace span fetch (`SELECT <full span columns> FROM stored_spans WHERE TenantId = .. AND TraceId = ..`) is the highest-volume shape in the codebase slow-query warnings (~76% of a 5,000-row sample, p95 read ~15MB for one trace) and recurs as a `MEMORY_LIMIT_EXCEEDED` (Code 241) reading the `SpanAttributes` column. The exceptions are server-wide: `(total) memory limit exceeded ... maximum: 14.00 GiB ... while reading column SpanAttributes`, resolved by the OvercommitTracker killing an in-flight query.

## Root cause (validated against prod via the read-only EXPLAIN endpoint)

`INDEXES` on the exact trace the OOM log named shows it holds only ~15 spans (the dedup resolves to a ~15-element SpanId set) in a single granule. So the blow-up is **not** span count: it is **large per-span attribute values** (big prompts / responses / embeddings stored in `SpanAttributes`). ClickHouse reads the whole `SpanAttributes` Map values array to return any of it, so a few spans with multi-MB attributes allocate heavily, and under concurrency that pushes the server past its total memory limit, where the OvercommitTracker kills an arbitrary query — one pathological trace then degrades unrelated requests.

There is no cheap way to read a subset of a Map's values (`mapFilter` / `m['k']` both read the whole values array; only the `.keys` subcolumn is cheap), and there is no write-time cap on attribute value size. The callers that read full attributes are `getSpansByTraceId`, `getNormalizedSpansByTraceId`, and `getSpanByIds`.

## Fix

Set an explicit `max_memory_usage` (`SINGLE_TRACE_READ_MAX_MEMORY_BYTES`, 2 GiB) on those three full-attribute single-trace reads. A pathological trace now fails *its own* read with `MEMORY_LIMIT_EXCEEDED` (surfaced as a drawer/query error) instead of pressuring the whole 14 GiB cluster — containing the blast radius that caused the observed server-wide OOM. The resilient client forwards real `clickhouse_settings` (it strips only `langwatch_*` keys), so the cap reaches ClickHouse.

The ceiling sits far above any normal trace (p95 read ~15MB) and below the global per-query limit, so it only trips on the pathological tail. It is a single named constant — tune there if legitimate large traces start failing.

This PR intentionally does **not** change the query shape. An earlier revision paged the SpanId set before reading heavy columns, but real-tenant `INDEXES` showed that gives no reduction for actual traces (they fit in one granule — the win only existed for >8k-span traces, which are not the OOM cause), so it was dropped to keep the change to the one thing that addresses the observed symptom.

## Expected impact

- A single huge-attribute trace can no longer drive the cluster into a total-memory OOM that kills unrelated queries; it fails in isolation and the drawer shows an error.
- No change to results, ordering, dedup, or the query plan for normal traces.

## Test plan

`span-storage.clickhouse.repository.unit.test.ts`: asserts all three full-attribute reads pass `max_memory_usage` in `clickhouse_settings` (via a mock-client spy — the cap is asserted by inspection, not by tripping an intentional OOM, which would wedge the vitest worker).

`span-storage.clickhouse.repository.integration.test.ts` (real ClickHouse testcontainer): seeds a 600-span trace with a stale duplicate and confirms the readers still return correct results **under the 2 GiB cap** — earliest 512 ordered by `StartTime`, latest version (not the stale one), full payload preserved.

Local run: unit 60/60, integration 3/3 (stable across repeated runs), OOM-safety guard 24/24, typecheck clean.

## EXPLAIN check (self-validated via the read-only `/api/ops/clickhouse/explain` endpoint, real tenant)

`INDEXES` / `PLAN` against prod using the real tenant + the exact OOM trace (both redacted) confirmed the trace is ~15 spans / 1 part / 1 granule — i.e. the OOM is value size, not granule count, which is why this PR is a memory guardrail rather than a query rewrite. EXPLAIN cannot show the per-query memory ceiling itself (that needs `ANALYZE`, which the endpoint blocks); the cap is verified by the unit spy and the integration suite passing under it.

## Follow-ups (out of scope, noted from the end-to-end read)

- The `spanDetail` LLM prompt-enrichment (`enrichLlmSpanWithAncestorPrompt`) re-reads the **whole trace's** full attributes on every LLM-span drawer open just to extract `langwatch.prompt.*` handles, then discards the rest — a redundant heavy read worth narrowing (it can at least drop the `Events.*` / `Links.*` / `ResourceAttributes` columns it never uses).
- The real lever for the ~15MB read-bytes warnings (unchanged here) is a schema-side split of large attribute values out of the hot `SpanAttributes` Map, or a write-time value-size cap — a larger change worth its own discussion.

### DIFF
--- a/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
@@ -3,12 +3,13 @@
  * `SpanStorageClickHouseRepository`, exercised against a real ClickHouse
  * testcontainer on the production `stored_spans` schema.
  *
- * Focus: `getSpansByTraceId` / `getNormalizedSpansByTraceId` must read the
- * heavy `SpanAttributes` payload for at most `limit` spans, not for every span
- * in the trace. A trace with far more spans than the read limit is the shape
- * that tips this path into MEMORY_LIMIT_EXCEEDED in prod, so the test seeds
- * such a trace and asserts both correctness and a real peak-memory reduction
- * versus the naive single-scan form.
+ * These readers carry an explicit `max_memory_usage` cap so a trace with very
+ * large per-span attribute values fails its own read instead of pressuring the
+ * whole server (see `SINGLE_TRACE_READ_MAX_MEMORY_BYTES`). The cap itself can't
+ * be asserted by tripping an OOM here — an errored ClickHouse response stream
+ * wedges the vitest worker — so the unit test asserts the setting is passed,
+ * and this suite confirms a normal trace read still returns correct results
+ * under the cap (ordering, latest-version dedup, full payload preserved).
  */
 import { nanoid } from "nanoid";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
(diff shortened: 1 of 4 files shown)


langwatch-agent:
No blockers. No actionable findings.

Residual risk / verification gap:
- The new integration coverage exercises correctness and demonstrates lower peak memory on the testcontainer schema, and CI is green across the unit and integration shards. Because the production benefit still depends on ClickHouse's physical read plan for the deployed schema and data distribution, run the PR description's proposed `EXPLAIN SYNTAX` / `EXPLAIN PIPELINE` check against a known high-span-count production trace before merging, confirming that the outer scan reads the heavy columns only for the bounded `SpanId` set.

PR #4433, opened 2026-05-31
perf(clickhouse): optimize trace-list query (-read bytes)

langwatch-agent:
## Evidence

`findAll` powers the main traces list. In the codebase slow-query warnings (Signal A) it is the worst sustained read-bytes offender that is not already covered by an open PR:

- ~140 warnings in a 5,000-row sample, **p95 read bytes ~617MB per list page load**, p95 duration ~1s.
- Same heavy-column-before-LIMIT shape that produced `MEMORY_LIMIT_EXCEEDED` on the topic-clustering and single-trace-span paths.

## Suspected cause

`findAll` selects the heavy `ComputedInput` / `ComputedOutput` (full trace input/output text) and the rest of the trace columns for **every** deduped trace matching the time window, then `ORDER BY <sort> LIMIT/OFFSET` trims to one page. ClickHouse materializes those payloads for the entire deduped window before the `LIMIT` applies. On top of that, `count() OVER ()` (used to return the page total) forces the whole post-dedup row set to buffer in memory rather than streaming.

So a list page over a busy window reads hundreds of MB of payload it then discards, and buffers the full result for the window count.

## Proposed fix

Page the matching `TraceId`s first in an inner subquery that reads only the key plus the sort columns (all light), `ORDER BY <sort> LIMIT/OFFSET`. Then read the heavy columns in the outer scan for that bounded `TraceId` set only. The latest-version dedup (`max(UpdatedAt)` GROUP BY IN-tuple) is preserved in both the page and the read, so version selection, ordering, and offset paging are unchanged.

`count() OVER ()` is removed; the page total is now a separate, parallel `count()` query that applies the same dedup filter but never selects a payload column.

No `LIMIT 1 BY` is introduced (covered by the `trace-dedup-oom-safety` structural guard).

## Expected impact

- Heavy-column reads bounded to one page (`limit`) of traces instead of the full deduped window, so read bytes and peak memory on the list drop sharply on busy windows.
- The total no longer forces a full-result buffer; it scans light columns only.
- One extra lightweight key-only scan and one parallel count query per page load; both avoid the payload columns.
- Result rows, ordering, dedup semantics, and `totalHits` are unchanged.

## Test plan

`src/server/app-layer/traces/repositories/__tests__/trace-list.clickhouse.repository.integration.test.ts` (new), real ClickHouse testcontainer on the production schema:

- Seeds 800 traces (more than one 200-row page), each with ~80KB of `ComputedInput`/`ComputedOutput`, plus a stale earlier version of the newest trace.
- Asserts one page is returned newest-first, `totalHits` reports the full deduped total (800, the stale duplicate not double-counted), the latest version (not the stale payload) is returned, and offset paging does not overlap.
- Runs the naive single-scan SQL (heavy read for all + `count() OVER ()`) and the paged SQL under stable `query_id`s, flushes `system.query_log`, and asserts the paged query's peak `memory_usage` is strictly lower. Both run to completion (no intentional OOM), so neither leaves an undrained socket.

Local run: integration 5/5 passing, OOM-safety guard 24/24 passing, typecheck clean.

## EXPLAIN check (self-validated via the read-only `/api/ops/clickhouse/explain` endpoint, real tenant)

Ran `EXPLAIN PLAN` / `PIPELINE` / `INDEXES` against prod ClickHouse using a real known-large tenant (~900k spans / ~60k traces over 7 days) and a 7-day `OccurredAt` window. `INDEXES` now works end-to-end (server rewrites to `EXPLAIN PLAN indexes = 1, actions = 1`). Tenant id and the deduped key-set sizes are redacted below.

`INDEXES`, heavy read path (`trace_summaries`):

```
old (count() OVER (), payload read for the whole deduped window):
  ReadFromMergeTree -> PrimaryKey (TraceId in <deduped-set> AND TenantId = <redacted>)
  Parts: 11   Granules: 39

new (page the TraceIds first, payload read for the 200-row page only):
  ReadFromMergeTree -> PrimaryKey (TraceId in <deduped-set> AND TraceId in <200-set> AND TenantId = <redacted>)
  Parts: 4    Granules: 6
```

The heavy `ComputedInput`/`ComputedOutput` read drops from **39 granules across 11 parts to 6 granules across 4 parts (~6.5x fewer granules)** because the page limits to the 200 newest traces, which are temporally concentrated. The `OccurredAt` range prunes partitions (`toYearWeek`) in both shapes (407/9327 granules at the partition stage).

`PLAN` / `PIPELINE`: the old plan has a `Window (Window step for window '')` node and a `WindowTransform` between the read and the sort/limit (it buffers the full deduped set); the new plan has neither. The separated count is a light `Aggregating` over the same 39 granules but reads count-only (no payload columns).

Caveat: prune ratios depend on the sort column. On the default `OccurredAt DESC` (recency) sort the newest page is temporally concentrated, as above; a sort on `TotalDurationMs`/`TotalCost` spreads the page across more parts, so the granule win is smaller there (the `Window` removal and the bounded heavy-row sort still apply).

Reviewer: re-run `EXPLAIN` / `ANALYZE` on a busy tenant/window before merge.

### DIFF
--- a/langwatch/src/server/app-layer/traces/repositories/__tests__/trace-list.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/__tests__/trace-list.clickhouse.repository.integration.test.ts
@@ -1,0 +1,247 @@
+/**
+ * Integration tests for `TraceListClickHouseRepository.findAll`, exercised
+ * against a real ClickHouse testcontainer on the production `trace_summaries`
+ * schema.
+ *
+ * Focus: the trace list must read the heavy `ComputedInput`/`ComputedOutput`
+ * payload for at most one page of traces, not for every deduped trace in the
+ * window. The window scan over those payloads (plus the `count() OVER ()`
+ * buffer) is the dominant read-bytes cost on this list in prod, so the test
+ * seeds far more traces than one page and asserts both correctness and a real
+ * peak-memory reduction versus the naive single-scan form.
+ */
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import type { ClickHouseClient } from "@clickhouse/client";
+import {
+  startTestContainers,
+  stopTestContainers,
+} from "../../../../event-sourcing/__tests__/integration/testContainers";
+import { TraceListClickHouseRepository } from "../trace-list.clickhouse.repository";
+import type { TraceListQuery } from "../trace-list.repository";
+
+const tenantId = `test-trace-list-${nanoid()}`;
+const base = Date.now() - 60 * 60 * 1000;
+
+const TOTAL_TRACES = 800;
+const PAGE_LIMIT = 200;
+const HEAVY_INPUT = "i".repeat(2000);
+const HEAVY_OUTPUT = "o".repeat(2000); // ~80KB of payload per trace
+
+let ch: ClickHouseClient;
+let repo: TraceListClickHouseRepository;
+
+function traceIdFor(i: number): string {
+  return `tr-${String(i).padStart(4, "0")}`;
+}
+
+function makeTraceSummaryRow(i: number, overrides: Record<string, unknown> = {}) {
+  return {
+    ProjectionId: `proj-${nanoid()}`,
+    TenantId: tenantId,
+    TraceId: traceIdFor(i),
+    Version: "v1",
+    Attributes: {},
+    OccurredAt: new Date(base + i),
+    CreatedAt: new Date(base + i),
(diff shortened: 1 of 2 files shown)


langwatch-agent:
Thanks @coderabbitai. Both nitpicks addressed in 822160425:
- Payload comment now matches the constants (bumped HEAVY_INPUT/HEAVY_OUTPUT to ~16KB combined per trace and corrected the comment).
- `makeTraceSummaryRow` now sets TraceName, RootSpanType, ContainsAi, ContainsPrompt, AnnotationIds, and LastEventOccurredAt explicitly rather than relying on column defaults. Integration suite still 5/5.

PR #4435, opened 2026-05-31
test(nlpgo): verbatim customer-prompt e2e for bedrock structured outputs

rogeriochaves:
## Summary

Regression-guard for the customer-reported failure that #4431 was supposed to fix but my dogfood missed.

#4431 shipped the `response_format` → forced-tool_use translation for bedrock + anthropic-family models, with a contrived live e2e (`workflow_llm_bedrock_structured_outputs_test.go`) that used a 1-line prompt + no `chat_messages` parameter + no `reasoning_effort`. That test passed both **before AND after** the fix on `eu.anthropic.claude-haiku-4-5-*` in `eu-central-1` — `output_config.format` happens to honor structured outputs in that profile + region pairing, so the contrived test gave a false-positive signal that the fix was sufficient.

The customer (`us.anthropic.claude-haiku-4-5-*` via managed-bedrock VPCE, `us-east-1`) reproduced the original failure: nlpgo collapsed all reasoning into the first declared output field as raw prose, python returned correctly parsed `{output: false, reason: "..."}`. Root cause turned out to be deploy lag (`lw-dev` runs `nlpgo-9114f90ac4c2-arm64` from 2026-04-27, predates the #4431 merge of 2026-05-30) — but the original e2e couldn't catch a regression specific to the customer's prompt shape.

## Changes

- `services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go` — verbatim replay of the customer's exported workflow (`classifier-relevance`):
  - Long system instructions (4-rule strict-boolean-evaluator block)
  - Separate `chat_messages` parameter with `{{input}}` placeholder (studio-prompt-playground assembly path, not the inline `prompt` field the original e2e used)
  - `reasoning_effort: medium`, `max_tokens: 8192`
  - Outputs: `{output: bool, reason: str}`

A future regression that breaks tool_use response parsing on reasoning-enabled anthropic OR misses the `chat_messages`-parameter assembly path now fails this e2e instead of slipping through.

Build-tagged `live_bedrock` so it runs only when AWS creds are wired. Targets `eu.anthropic.claude-haiku-4-5-*` by default (our `langwatch-dev-bedrock-user` IAM only allows `eu.*` — `us.*` profile path still unreachable from CI without separate IAM).

One material drift from the customer's exported workflow: `temperature=1` AND `top_p=1` are both set in the customer config, but bedrock haiku 4.5 rejects that pair (`"temperature and top_p cannot both be specified for this model"`). Drops `top_p`; inline comment in the test documents why.

## Verified

- `go test -tags live_bedrock ./tests/integration/ -run BedrockCustomerReproVerbatim` against real bedrock + `eu.anthropic.claude-haiku-4-5-20251001-v1:0`: `output: bool` + `reason: str` parsed correctly
- Without the `eu.*` env vars: skips gracefully

## Test plan
- [x] Live e2e passes with merged #4431 fix on `main`
- [ ] Watch CI

## Open items (not in this PR)

- Trigger a redeploy of latest `nlpgo` to `lw-dev` and `prod` so the customer actually benefits from #4431 (deploy lane, not code)
- If the customer reports the same failure on the post-deploy build, that's a real H3 (us-east-1 + us.* profile vs eu-central-1 + eu.* profile divergence) and needs separate IAM to investigate

### DIFF
--- a/services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go
+++ b/services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go
@@ -1,0 +1,286 @@
+//go:build live_bedrock
+
+package integration_test
+
+// Verbatim replay of the customer-reported failing workflow
+// (exported 2026-05-31): a signature node with Structured Outputs
+// {output:bool, reason:str}, long system instructions, a chat_messages
+// parameter with {{input}} placeholder, reasoning_effort:medium,
+// max_tokens:8192, model bedrock/us.anthropic.claude-haiku-4-5-*.
+//
+// The contrived TestSync_RealWorkflowEndToEnd_BedrockStructuredOutputs
+// e2e in workflow_llm_bedrock_structured_outputs_test.go used a 1-line
+// prompt + no chat_messages parameter + no reasoning_effort + short
+// max_tokens. It passed and led to the false-positive merge of #4431.
+// The customer then reproduced the original failure: nlpgo dumped all
+// the reasoning into the `output` field, python returned parsed
+// {output:false, reason:"..."}. This test mirrors the customer's
+// signature configuration exactly so any future regression on this
+// prompt shape fails CI.
+//
+// Build tag: live_bedrock. Uses eu.* inference profile (only one our
+// langwatch-dev-bedrock-user IAM allows).
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/langwatch/langwatch/pkg/health"
+	"github.com/langwatch/langwatch/services/aigateway/dispatcher"
+	"github.com/langwatch/langwatch/services/nlpgo/adapters/dispatcheradapter"
+	"github.com/langwatch/langwatch/services/nlpgo/adapters/httpapi"
+	"github.com/langwatch/langwatch/services/nlpgo/adapters/llmexecutor"
+	"github.com/langwatch/langwatch/services/nlpgo/app"
+	"github.com/langwatch/langwatch/services/nlpgo/app/engine"
+	"github.com/langwatch/langwatch/services/nlpgo/app/engine/blocks/codeblock"
+	"github.com/langwatch/langwatch/services/nlpgo/app/engine/blocks/httpblock"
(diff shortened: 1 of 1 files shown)


langwatch-agent:
2 P2s, 1 P3.

### P2: the regression guard is not executed by CI

`services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go:1` is hidden behind `//go:build live_bedrock`. The normal Go workflow runs `go test ... ./services/nlpgo/...` without that tag (`.github/workflows/go-ci.yaml:64`). The only workflow that supplies `live_bedrock` is `.github/workflows/gateway-matrix.yaml:204-207`, and it runs `./tests/matrix/...` from `services/aigateway`, not the nlpgo integration package. As written, a future regression will still merge without this test running.

Minimal fix: add a secret-gated CI lane or scheduled/manual lane that runs `go test -tags live_bedrock ./services/nlpgo/tests/integration/ -run BedrockCustomerReproVerbatim`, and make the intended enforcement level explicit.

### P2: the customer-repro test bypasses the managed-Bedrock VPCE route

`services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go:121-128` threads credentials but never propagates `BEDROCK_VPCE_ENDPOINT`. The existing Bedrock e2e does this at `services/nlpgo/tests/integration/workflow_llm_bedrock_e2e_test.go:159-164` by adding `aws_bedrock_runtime_endpoint`. The PR description says the reported failure came through the managed-Bedrock VPCE path, so this new test cannot exercise that path even when an endpoint is supplied.

Minimal fix: mirror the existing endpoint handling and log whether the public or VPCE route is exercised. Keep a public baseline if useful, but add a managed-route invocation before claiming customer-path coverage.

### P3: narrow the “verbatim” claim or preserve the remaining sampling knobs

`services/nlpgo/tests/integration/workflow_llm_bedrock_customer_repro_test.go:130-147` says the LLM config is customer-exact and references `top_k`, `top_p`, `min_p`, etc., but the constructed map includes only `max_tokens`, `temperature`, and `reasoning_effort`. Dropping `top_p` is documented; the other omitted knobs are not. Since model-behavior shape is the point of this regression test, either include supported exported knobs or document each intentional drift and rename the test accordingly.

### Verification

Reviewed `3268da693cd6148e3d628029f75018373638fb55`. `git diff --check origin/main...HEAD` passes and current GitHub checks are green. I could not compile the tagged package locally because this worktree environment does not have `go` installed.

PR #4436, opened 2026-05-31
fix(gateway): forward tool_choice in bedrock VPCE intercept ToolConfig

rogeriochaves:
## Summary

Closes the customer-reported structured-output failure on managed-bedrock (#4431 follow-up).

#4431 shipped the response_format → forced tool_use translation in nlpgo's executor for bedrock+anthropic-family models. That worked on the public bedrock path (bifrost's bedrock provider honors `tool_choice` via `convertToolConfigFromFiltered`), but the VPCE intercept's `mapBedrockToolConfig` in `services/aigateway/adapters/providers/bedrock_vpce.go` emitted only the `tools` array and silently dropped `tool_choice`. So when a managed-bedrock customer's request flowed through nlpgo (with my structured-output rewrite emitting tools + tool_choice) and got dispatched via the VPCE path:

1. Gateway VPCE intercept dropped `tool_choice`
2. Bedrock received the synthetic `lw_so_<schema>` tool with no pin
3. Anthropic was free to ignore the tool and reply with text
4. nlpgo's engine `extractSignatureOutputs` prose-fallback dumped the entire model reasoning into the first declared output field

Exact reproduction matches the customer's 2026-05-31 dogfood screenshot: nlpgo path returned all reasoning collapsed into `output`, python path (LiteLLM, which uses tool_use with proper tool_choice forwarding via the SDK) returned correctly parsed `{output: false, reason: "..."}`.

## Changes

- `services/aigateway/adapters/providers/bedrock_vpce.go`
  - New `mapBedrockToolChoice` function maps the OpenAI/bifrost `ChatToolChoice` union to the bedrockruntime `ToolChoice` union:
    - string `"any"`/`"required"` or struct `{type: any|required}` → `ToolChoiceMemberAny`
    - string `"auto"` or struct `{type: auto}` → `ToolChoiceMemberAuto`
    - struct `{type: function, function.name: X}` → `ToolChoiceMemberTool{Value: SpecificToolChoice{Name: X}}` ← the customer-fix branch
    - string `"none"` / unrecognized / empty function name → `nil` (bedrock defaults to auto)
  - `mapBedrockToolConfig` now sets `ToolConfig.ToolChoice = mapBedrockToolChoice(params.ToolChoice)`. When no `tool_choice` is on the inbound request the field stays nil and bedrock defaults to auto — same observable behavior as before the fix, just with the forced-pin case actually working.

- `services/aigateway/adapters/providers/bedrock_vpce_test.go`
  - `TestMapBedrockToolChoice` covers the mapping matrix (nil, string/struct variants, empty function name, "none" fallback).
  - `TestMapBedrockToolConfig_ForwardsForcedToolChoice` pins the end-to-end public-mapper contract.

- `services/aigateway/adapters/providers/bedrock_vpce_live_test.go`
  - `TestBedrockVPCELive_StructuredOutputToolChoice` exercises the full `dispatchBedrockVPCE` path with a forced `lw_so_*` tool against any Bedrock-compatible endpoint. Build tag `live_bedrock_vpce`, env-gated `BVPCE_ENDPOINT` + AWS creds.

## Verified

- `go test ./services/aigateway/adapters/providers/` — green
- `go test ./services/aigateway/...` — green
- Live structured-output test against public bedrock (`https://bedrock-runtime.us-east-1.amazonaws.com`, `us.anthropic.claude-haiku-4-5-20251001-v1:0`): tool invoked, JSON parsed, bool + reason fields populated
- **Live end-to-end against the actual customer VPCE** (`http://vpce-...:80`, `us-east-1`, `us.anthropic.claude-haiku-4-5-20251001-v1:0`, STS-temporary creds from the managed-bedrock proxy → cross-account chain): pre-fix returns text content + no `tool_calls`; post-fix invokes `lw_so_ClassifierRelevance` and returns `{"output": true, "reason": "<paragraph>"}`. Same behavior LiteLLM produces on the python path.

## Deploy ask

Gateway image needs a rebuild + redeploy on prod for the customer to benefit — `#4431` (in `nlp-service` / `nlp-lambda`) is already deployed via the saas submodule bump, but the gateway service has not been rebuilt since this code lived only in `aigateway`. Per rchaves "nlp-lambda and nlp-service were rebuilt (gateway was not but would that matter)?" — yes, it matters, and this PR is the reason why.

## Test plan
- [x] Unit tests pass for the new mapper + end-to-end contract
- [x] Live test passes against public bedrock with forced tool_choice
- [x] Live test passes against the actual customer managed-bedrock VPCE (off-PR, one-off verification)
- [ ] Watch CI
- [ ] Trigger gateway rebuild + redeploy after merge

### DIFF
--- a/services/aigateway/adapters/providers/bedrock_vpce.go
+++ b/services/aigateway/adapters/providers/bedrock_vpce.go
@@ -402,6 +402,15 @@ func mapBedrockInferenceConfig(params *bfschemas.ChatParameters) *brtypes.Infere
 
 // mapBedrockToolConfig maps the neutral tool list to Bedrock's
 // ToolConfiguration. Returns nil when no tools are present.
+//
+// Forwards the caller's tool_choice directive to Bedrock when set. Without
+// this, customer 2026-05-31 dogfood: nlpgo translates response_format to
+// tools + tool_choice(forced) for structured outputs on bedrock+anthropic
+// (executor.go shouldUseToolUseForStructuredOutput); the public-bedrock
+// path through bifrost honors that, but the VPCE intercept here dropped
+// tool_choice silently. Model received the synthetic lw_so_* tool with no
+// pin → free to ignore it and reply with text → engine prose-fallback
+// dumped the entire reasoning into the first declared output field.
 func mapBedrockToolConfig(params *bfschemas.ChatParameters) *brtypes.ToolConfiguration {
 	if params == nil || len(params.Tools) == 0 {
 		return nil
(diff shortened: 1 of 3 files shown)


langwatch-agent:
1 P2 follow-up. This PR is already merged, so this needs a corrective PR.

### P2: `tool_choice: "none"` still permits Bedrock to invoke a tool

`services/aigateway/adapters/providers/bedrock_vpce.go:455-465` maps string `"none"` to `nil`, while `mapBedrockToolConfig` at `:438-441` still sends the full `Tools` array. On Bedrock Converse, omitting `ToolChoice` means automatic selection, so a request explicitly asking for no tool calls can still invoke one. The comment says the caller can omit tools, but this adapter does not remove them. The test at `services/aigateway/adapters/providers/bedrock_vpce_test.go:288-293` locks in the incorrect intermediate mapping without checking the final `ToolConfiguration` semantics.

Minimal fix: when the inbound choice is `"none"`, omit `ToolConfig` entirely (or otherwise suppress the tool list) so Bedrock cannot auto-select a tool. Add an end-to-end mapper test asserting that `tools + tool_choice:none` produces no outbound tool configuration.

### Verification

Reviewed merged head `c6b4bc2a12c30a9dcd81b3066e496b2cf3e87b77` through `refs/pull/4436/head` because the source branch has already been deleted. `git diff --check fd741bf094c12b0e48eb26f3ae507b3276f1de9e..c6b4bc2a12c30a9dcd81b3066e496b2cf3e87b77` passes. PR CI was green before merge. I could not run Go tests locally because this worktree environment does not have `go` installed.

PR #4447, opened 2026-05-31
fix(chart): gate clickhouse Secrets on autogen.enabled (Argo drift, subchart + parent)

rogeriochaves:
## Summary

Mirrors the gating shape Sarah landed on gateway-auth Secret (PR #4440) onto BOTH places clickhouse credentials are materialised:

- `charts/clickhouse-serverless/templates/secret.yaml` — for standalone subchart installs
- `charts/langwatch/templates/clickhouse/url-secret.yaml` — for umbrella installs (this is the one Argo actually reconciles against for delphyr-shaped operators)

Per-key `lookup-or-rand`, `helm.sh/resource-policy: keep`, fail-fast at chart-render time when neither path is configured, plus a pre-install/pre-upgrade preflight Job inside the subchart that validates operator-owned Secret keys before pods roll.

## The drift bug

Pre-fix, both templates rendered whenever `chartManaged` was true (subchart additionally required `auth.existingSecret` empty) and fell back to `randAlphaNum` when helm `lookup` returned no in-cluster Secret. Under ArgoCD the repo-server has no cluster access, so lookup returns nil on every reconcile, the random branch fires every render, and the ClickHouse password + Keeper `clusterSecret` rotate underneath the running pods. Rotation under replication is especially bad: rotating `clusterSecret` breaks inter-Keeper auth and splits the quorum.

## The fix

Two explicit paths, picked by `autogen.enabled`:

| Path | When | What renders |
|---|---|---|
| `autogen.enabled=true` | dev / stock-install | Secret rendered via per-key lookup-or-rand. `helm.sh/resource-policy: keep`. Per-key fallback heals partial Secrets in place (single→replicated migration trap). |
| `autogen.enabled=false` (default) | production / GitOps | No Secret rendered. Operator pre-creates one out-of-band. Subchart preflight Job validates required keys before pods roll. Parent chart `validateSecrets` forces operator to override `clickhouse.auth.existingSecret` to a non-default name (so the deployment uses runtime CLICKHOUSE_URL composition). |

With both unset, the chart-time validators hard-fail with actionable errors.

### Subchart (`charts/clickhouse-serverless/`)
- `templates/secret.yaml` — gated on `autogen.enabled` + per-key lookup-or-rand + `helm.sh/resource-policy: keep`
- `templates/_helpers.tpl` — new `validateAuth` helper invoked from `validate`; unified `secretKey`/`clusterSecretKey` helpers to use configured keys on both paths
- `templates/preflight-secrets-job.yaml` — NEW: ServiceAccount + Role (scoped to the named Secret) + RoleBinding + Job. Runs only when `(not autogen.enabled) && auth.existingSecret`
- `values.yaml` — new `autogen.enabled` (default false) and `preflight.*` blocks
- `tests/values-{single,replicated}.yaml` — flip `autogen.enabled=true` so existing e2e suites still exercise the chart-managed Secret path
- `tests/e2e.sh` — extended `test_existing_secret` to flip `autogen.enabled=false` (exercises the production-shape operator-owned path + preflight). Added `test_existing_secret_missing_key` and `test_no_auth_config_fails`

### Parent (`charts/langwatch/`)
- `templates/clickhouse/url-secret.yaml` — gated on `autogen.enabled` + per-key lookup-or-rand + `helm.sh/resource-policy: keep`. `url` is always derived from the resolved password (never stored stale).
- `templates/_helpers.tpl` `validateSecrets` — hard-fails when `chartManaged=true && !autogen.enabled && secretName == default <release>-clickhouse`. Names the override in the error message.

### Chart render
`helm template charts/clickhouse-serverless` (subchart standalone):
- defaults (both unset) → `clickhouse-serverless: no credentials Secret configured...` ✓
- `--set autogen.enabled=true` → Secret rendered with `helm.sh/resource-policy: keep` and both keys ✓
- `--set auth.existingSecret=my-ch-creds` → no Secret rendered, preflight Job + Role + RoleBinding + ServiceAccount render ✓
- `--set auth.existingSecret=my-ch-creds --set replicas=3` → preflight checks both `password` and `clusterSecret` ✓
- `helm lint` clean on both paths

`helm template lwparent charts/langwatch` (umbrella):
- `--set autogen.enabled=true` → `lwparent-clickhouse` renders with password+clusterSecret+url+keep annotation ✓
- `--set autogen.enabled=false --set clickhouse.chartManaged=true` → hard-fail with actionable error naming the override ✓
- `--set autogen.enabled=true --set clickhouse.auth.existingSecret=my-ch-secret` → no url-secret renders, deployment env uses CLICKHOUSE_PASSWORD + runtime CLICKHOUSE_URL composition ✓

### Live cluster
Subchart e2e suite running locally on Kind right now (`charts/clickhouse-serverless/tests/e2e.sh`). Covers:
- single-replica install + DML roundtrip
- helm upgrade password preservation (the critical timing case — verifies the lookup-or-rand reuse branch actually preserves the password across `hc upgrade`)
- operator-owned path + preflight Job validates the keys
- preflight Job blocks install when required key is missing
- chart-render hard-fails on no auth config

The umbrella chart hasn't been exercised on Kind in this PR. Argo dry-run reconcile on a dev cluster is the next confidence step — calling that out explicitly so it doesn't get skipped.

## Operator migration

Existing installs that were on the implicit autogen path (no `auth.existingSecret`, no explicit `autogen.enabled`): add `autogen.enabled=true` to your values overrides. The chart's per-key lookup-or-rand will reuse all existing keys from the in-cluster Secret on the next upgrade, so no password rotation.

Argo-driven installs that have been silently rotating credentials (the bug this PR closes): set `autogen.enabled=false`, pre-create the Secret with `password` (and `clusterSecret` when `replicas>1`), set `clickhouse.auth.existingSecret` to a custom name (parent-chart) or `auth.existingSecret` to the Secret's name (subchart). The preflight Job (subchart) or runtime CLICKHOUSE_URL composition (parent) handles the rest.

## Test plan

- [x] `helm lint` both paths
- [x] `helm template` all states locally (subchart + umbrella)
- [ ] Kind e2e suite running now (`charts/clickhouse-serverless/tests/e2e.sh`) — will update with results
- [ ] Argo reconcile loop on dev cluster — should now be steady-state with `autogen.enabled=false` + operator-owned Secret

### DIFF
--- a/charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
+++ b/charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
@@ -1,0 +1,136 @@
+{{/*
+  Pre-install / pre-upgrade preflight: verifies the operator-owned
+  ClickHouse credentials Secret has every key the StatefulSet mounts
+  before pods roll. Catches the trap where a chart bump renames a
+  required key (or the operator imported the Secret without setting
+  clusterSecret on a single→replicated migration) and the upgrade
+  would otherwise crashloop the new pods with CreateContainerConfigError.
+
+  Scope: only renders when the operator owns the Secret out-of-band
+  (autogen.enabled=false AND auth.existingSecret is set). When the
+  chart materialises the Secret on the autogen path, both keys land
+  in the same release as the StatefulSet so there is nothing to
+  verify ahead of time — checking would race the materialiser and
+  fail the upgrade before the Secret is rendered.
+
+  Validates `password` always, and `clusterSecret` only when
+  replicas>1 (single-replica clusters don't need inter-node auth).
+
+  Helm hook contract: pre-install + pre-upgrade. pre-install catches a
+  fresh-install operator who forgot to seed the Secret keys before
+  `helm install`; pre-upgrade catches operator-key drift introduced by
+  subsequent chart bumps. hook-delete-policy keeps failed runs around
+  for post-mortem; succeeded runs get cleaned up automatically.
+*/}}
+{{- if and .Values.preflight.enabled (not .Values.autogen.enabled) .Values.auth.existingSecret }}
+{{- $secretName := tpl .Values.auth.existingSecret . }}
+{{- $passwordKey := default "password" .Values.auth.secretKeys.passwordKey }}
+{{- $clusterSecretKey := default "clusterSecret" .Values.auth.secretKeys.clusterSecretKey }}
+{{- $replicas := .Values.replicas | int }}
+{{- $fullname := include "clickhouse-serverless.fullname" . }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $fullname }}-preflight
+  labels:
+    {{- include "clickhouse-serverless.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-10"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ $fullname }}-preflight
+  labels:
+    {{- include "clickhouse-serverless.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-10"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: [{{ $secretName | quote }}]
+    verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ $fullname }}-preflight
+  labels:
+    {{- include "clickhouse-serverless.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-10"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+subjects:
+  - kind: ServiceAccount
+    name: {{ $fullname }}-preflight
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ $fullname }}-preflight
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ $fullname }}-preflight-secrets
+  labels:
+    {{- include "clickhouse-serverless.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 0
+  activeDeadlineSeconds: {{ .Values.preflight.activeDeadlineSeconds | default 60 }}
+  ttlSecondsAfterFinished: 600
+  template:
+    metadata:
+      labels:
+        {{- include "clickhouse-serverless.labels" . | nindent 8 }}
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ $fullname }}-preflight
+      containers:
+        - name: preflight
+          image: {{ .Values.preflight.image }}
+          imagePullPolicy: {{ .Values.preflight.imagePullPolicy }}
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              set -u
+              ns={{ .Release.Namespace | quote }}
+              secret={{ $secretName | quote }}
+              missing=""
+              check_key() {
+                key="$1"
+                v=$(kubectl -n "$ns" get secret "$secret" -o jsonpath="{.data.$key}" 2>/dev/null || printf '')
+                if [ -z "$v" ]; then
+                  missing="$(printf '%s\n  - %s / %s' "$missing" "$secret" "$key")"
+                fi
+              }
+              check_key {{ $passwordKey | quote }}
+              {{- if gt $replicas 1 }}
+              check_key {{ $clusterSecretKey | quote }}
+              {{- end }}
+              if [ -n "$missing" ]; then
+                printf 'preflight: required Secret keys missing in namespace %s:%s\n\n' "$ns" "$missing" >&2
+                printf 'Each missing key blocks the upgrade. Either add it to the existing Secret\n' >&2
+                printf 'or switch to chart-managed via autogen.enabled=true. See NOTES.txt for\n' >&2
+                printf 'the kubectl create commands.\n' >&2
+                exit 1
+              fi
+              printf 'preflight: all required Secret keys present on %s in namespace %s\n' "$secret" "$ns"
+          resources:
+            requests:
+              cpu: 10m
+              memory: 32Mi
+            limits:
+              cpu: 100m
+              memory: 64Mi
+{{- end }}
--- a/charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
+++ b/charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
@@ -115,6 +115,15 @@ spec:
           args:
             - |
               set -u
+              # Fail fast when the operator overrode preflight.image with one
+              # that doesn't ship jq. Without this guard a missing jq would
+              # degrade silently into the generic "required Secret keys
+              # missing" path and the real cause (wrong image) would be
+              # buried in container start failures.
+              if ! command -v jq >/dev/null 2>&1; then
+                printf 'preflight: jq is required in preflight.image (got %s); use the default alpine/k8s image or a mirror that ships jq\n' {{ .Values.preflight.image | quote }} >&2
+                exit 1
+              fi
               ns={{ .Release.Namespace | quote }}
               secret={{ $secretName | quote }}
               missing=""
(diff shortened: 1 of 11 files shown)


charts/clickhouse-serverless/values.yaml
```diff
@@ -42,13 +42,46 @@ backup:
...
+# This is the only mode that is drift-free under GitOps controllers whose
+# render step has no cluster access (ArgoCD's repo-server returns nil from
+# helm lookup every sync, so the autogen path would rotate credentials on
+# every reconcile). The preflight Job validates the configured keys exist
+# before pods roll.
+#
+# Either way, when both flags are unset the chart-time validator fails with
+# a clear message instructing the operator to pick one path.
+autogen:
+  enabled: false
```
langwatch-agent, line 71: **P2: Update the standalone chart docs for the new required opt-in.** Changing the default to `false` intentionally makes the previous stock install invalid, but `charts/clickhouse-serverless/README.md` still tells users to run both quick-start commands without `--set autogen.enabled=true`, and its Secret Management section still says an empty password automatically generates a Secret. Those documented commands now fail in `validateAuth`. Please update the quick starts, parameter table, and Secret Management section in this PR (and call out the migration for existing standalone installs), or preserve the old standalone default if that compatibility is required.

langwatch-agent (commented):
Reviewed the standalone clickhouse-serverless gating change. The core Argo-safe split is sound, but 3 P2 issues remain in the public install contract and the new preflight path; details are inline.

charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+  Validates `password` always, and `clusterSecret` only when
+  replicas>1 (single-replica clusters don't need inter-node auth).
+
+  Helm hook contract: pre-install + pre-upgrade. pre-install catches a
+  fresh-install operator who forgot to seed the Secret keys before
+  `helm install`; pre-upgrade catches operator-key drift introduced by
+  subsequent chart bumps. hook-delete-policy keeps failed runs around
+  for post-mortem; succeeded runs get cleaned up automatically.
+*/}}
+{{- if and .Values.preflight.enabled (not .Values.autogen.enabled) .Values.auth.existingSecret }}
```
langwatch-agent: **P2: Run preflight whenever `auth.existingSecret` wins.** `secret.yaml` suppresses chart materialization whenever `auth.existingSecret` is set, even if `autogen.enabled=true`, and the StatefulSet mounts that external Secret. This condition skips the preflight in that accepted configuration, so a missing Secret/key reaches the pods as `CreateContainerConfigError`. This is easy to hit when a dev values file enables autogen and an operator overrides only `auth.existingSecret`. Either render this hook for every non-empty `auth.existingSecret`, or reject the ambiguous combination in `validateAuth`.

charts/clickhouse-serverless/templates/preflight-secrets-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              set -u
+              ns={{ .Release.Namespace | quote }}
+              secret={{ $secretName | quote }}
+              missing=""
+              check_key() {
+                key="$1"
+                v=$(kubectl -n "$ns" get secret "$secret" -o jsonpath="{.data.$key}" 2>/dev/null || printf '')
```
langwatch-agent: **P2: Do not interpolate configurable Secret keys into JSONPath dot notation.** Kubernetes Secret keys may contain `.` (for example `clickhouse.password`), and `auth.secretKeys.passwordKey` / `clusterSecretKey` are configurable. Expanding that value into `{.data.$key}` treats dots as nested JSONPath traversal, so preflight reports a present key as missing and blocks install/upgrade. Fetch the Secret JSON once and use a key-safe lookup such as `jq -r --arg k "$key" .data[$k] // empty`, or escape JSONPath metacharacters before invoking kubectl.

rogeriochaves:
Quick review while you're mid-dogfood — agree the static-literal default was the right call on #4440 (just pushed 612119aec for the helm template CI failure that fix triggered).

Three reviewer findings on this PR look real to me + one concrete kind-e2e failure:

**1. P2 jsonpath dot-notation bug in `preflight-secrets-job.yaml`** (line 112-ish): `kubectl get secret X -o jsonpath="{.data.$key}"` with `$key=clickhouse.password` interprets dots as nested traversal — reports a present key as missing and blocks install. `auth.secretKeys.passwordKey` is operator-configurable so this is a real footgun. Fix mirrors my Lane A preflight pattern (`charts/langwatch/templates/preflight-secrets-job.yaml` lines 146-155): one `kubectl get -o json` upfront, then `printf '%s' "$json" | jq -r --arg k "$key" '.data[$k] // empty'`. alpine/k8s ships jq.

**2. P2 preflight skipped when autogen=true + existingSecret set**: `secret.yaml` gate is `and autogen.enabled (not existingSecret)` so when both are set, no Secret renders AND preflight is skipped → StatefulSet mounts an existingSecret that might be missing keys → CreateContainerConfigError. Fix: drop the autogen check from the preflight gate, so it becomes `if and .Values.preflight.enabled .Values.auth.existingSecret` (the autogen-only-no-existingSecret case has no Secret to verify; preflight just doesn't render).

**3. P2 README still shows old commands** without `--set autogen.enabled=true` — operators copy-pasting from the README now hit `validateAuth` failure on first install. Quick fix.

**4. kind e2e fail (`test_existing_secret_missing_key`)**: helm install fails as expected (`BackoffLimitExceeded`) so PASS on "preflight blocked install", but FAIL on "preflight error message not found in helm output". The preflight Job's stderr lives in the Job pod logs, not helm's stdout. Your current `hook-delete-policy: before-hook-creation,hook-succeeded` already keeps failed Jobs around. Test fix: after `helm_install` fails, `kubectl logs -n $NAMESPACE job/${RELEASE}-clickhouse-preflight-secrets` and pipe to `/tmp/ch-preflight-out` before grepping.

Rebase-after-merge note: my #4440 helper change is parent-chart-only (`langwatch.appSecretName` switched to fixed `langwatch-app-secrets`); doesn't touch clickhouse-serverless. Clean rebase.

rogeriochaves:
**Live Kind e2e on the subchart: green end-to-end.**

Re-ran `charts/clickhouse-serverless/tests/e2e.sh` with the four review fixes from `5c6aa5cfc` + the preflight-image pre-pull from `818b339ab`. All suites pass on a real cluster:

```
Suite: single-replica                                    ✓ install, /ping, DML, no Keeper pods, Secret rendered
Suite: upgrade — password preserved across helm upgrade  ✓ Password unchanged after upgrade
Suite: existing secret                                   ✓ Preflight Job ran, ClickHouse authenticated with operator-owned password
Suite: existing secret missing required key              ✓ Preflight blocked install (BackoffLimitExceeded)
                                                         ✓ Job-logs grep surfaced the missing-key error
Suite: no auth configured                                ✓ Chart render hard-failed with validateAuth message
Suite: cold storage configuration                        ✓
Suite: backup configuration                              ✓
PASS  All clickhouse-serverless tests passed
```

This confirms on a live cluster:

- the `lookup-or-rand` per-key fold preserves the password across `helm upgrade` (the timing/race rchaves flagged earlier — yes, ClickHouse picks up the unchanged users.d at boot and no rotation happens)
- the jq dot-notation fix correctly handles operator-configured keys
- the autogen=true + existingSecret preflight gate (review finding #2) renders the Job in both branches
- the Job-logs grep (review finding #4) surfaces the actionable error to the operator

**Umbrella e2e fix (`afcd74460`):** the parent-chart `e2e: overlays` job was failing with `job failed: BackoffLimitExceeded` because every overlay that flips `autogen.enabled=true` was hitting the subchart preflight Job. Under the umbrella, the parent's `url-secret.yaml` materialises the Secret in the same release as the StatefulSet, but the subchart preflight runs as a pre-install hook *before* any other manifest is applied — it was trying to validate a Secret that did not exist yet. Set `clickhouse.preflight.enabled: false` as the parent default. Subchart-standalone installs keep preflight on.

rogeriochaves:
Resolved all 4 open review threads:

- **langwatch-agent: standalone docs need autogen quick-start** (`charts/clickhouse-serverless/values.yaml`) — already addressed in [5c6aa5cfc](https://github.com/langwatch/langwatch/commit/5c6aa5cfc): README Quick Start rewritten with both dev-autogen + production-operator-owned commands, parameter table extended with `autogen.enabled` / `auth.clusterSecret` / `preflight.*` rows, Secret Management section rewritten to call out the ArgoCD drift trap explicitly.

- **langwatch-agent: preflight skipped when autogen=true + existingSecret** (`templates/preflight-secrets-job.yaml`) — already addressed in [5c6aa5cfc](https://github.com/langwatch/langwatch/commit/5c6aa5cfc): dropped `not autogen.enabled` from the preflight gate so the Job renders whenever `auth.existingSecret` is set, regardless of autogen flag. Confirmed live on Kind in the existing-secret suite (both positive and negative cases pass).

- **langwatch-agent: JSONPath dot-notation bug** (`templates/preflight-secrets-job.yaml`) — already addressed in [5c6aa5cfc](https://github.com/langwatch/langwatch/commit/5c6aa5cfc): switched to one `kubectl get -o json` upfront + `jq -r --arg k "$key" '.data[$k] // empty'` so dotted operator-configured key names (`clickhouse.password`) resolve correctly. Verified via `helm template ... --set auth.secretKeys.passwordKey=some.dotted.key` that the rendered Job uses the dict-index form.

- **coderabbit: fail fast when preflight.image lacks jq** (`templates/preflight-secrets-job.yaml`) — addressed in [5c35e7bfd](https://github.com/langwatch/langwatch/commit/5c35e7bfd): added `command -v jq || exit 1` guard at the top of the Job script with an error that names the configured image, and tightened the `values.yaml` + README docs to state `preflight.image` MUST ship `jq` (alpine/k8s does; bitnami/kubectl + distroless do not).

PR #4464, opened 2026-06-01
docs(adr): trace-readiness debounce for trigger evaluation (030)

0xdeafcafe:
## Summary

Forward-looking ADR — proposes a **per-trigger \`evaluationDebounceMs\`**
that gates trigger evaluation until the trace has been quiet for the
configured window. Closes the half-formed-dispatch hole that today's
eager "re-evaluate on every span" reactor opens for both notify
(refund-granted-then-walked-back) and persist (truncated dataset rows)
triggers, without needing a terminal-trace signal we don't have.

Key choices captured:

- **Non-zero default**: every existing and new trigger ships with a
  real debounce window; the half-formed default is not on the table.
  **30 seconds** chosen as the default; **15s** offered as a
  first-class option for teams that have measured P99 trace-completion
  latency and want to tighten up. Anything between 0 and 5min is one
  click away in the drawer.
- **Per-trigger, surfaced in the UI**: stage on the staged authoring
  drawer (ADR-028) and a column on the automations settings list
  (ADR-029), so the latency-vs-completeness trade-off is legible at
  authoring time and reviewable at scan time.
- **Mechanism**: the in-house \`GroupQueue\`'s existing Debounce Mode —
  \`DeduplicationConfig\` with \`extend: true, replace: true\`. New
  \`trigger-evaluation\` queue keyed on \`(projectId, triggerId,
  traceId)\`. Every new span resets the TTL; the queue dispatcher
  fires once after settlement, runs filters, then calls
  \`dispatchTriggerAction\` (which still routes notify through the
  ADR-025 outbox digest path).
- **Orthogonal to ADR-025 cadence**: debounce sits *before* filter
  evaluation (wait for trace to settle); cadence sits *after* dispatch
  (group matches into a digest window).
- **All action classes** — persist triggers benefit more than notify
  because a truncated dataset row corrupts an eval set silently.
- **Worker-only**, follows the existing outbox-stack process-role rule.

No code changes. Implementation lands as a follow-up stack once this
is reviewed.

Refs: \`dev/docs/adr/021\`, \`022\`, \`023\`, \`025\`, \`028\`, \`029\`.

### DIFF
--- a/dev/docs/adr/030-trace-readiness-debounce-for-trigger-evaluation.md
+++ b/dev/docs/adr/030-trace-readiness-debounce-for-trigger-evaluation.md
@@ -1,0 +1,299 @@
+# ADR-030: Trace-readiness debounce for trigger evaluation
+
+**Date:** 2026-06-01
+
+**Status:** Proposed
+
+## Context
+
+A trace in LangWatch is assembled from many spans that arrive over time —
+the root span first, internal LLM and tool spans next, evaluation spans
+last, sometimes seconds or minutes apart. We **do not have a terminal
+signal** ("this trace is now complete"): there is no end-of-trace event,
+no per-trace "expected span count" we could compare against, and OTel
+exporters do not flush at consistent times. Adopting an explicit
+terminal signal is a separate, larger change — out of scope for this ADR.
+
+Today the two trigger reactors (`alertTrigger` on the trace pipeline,
+`evaluationAlertTrigger` on the evaluation pipeline) fire on **every**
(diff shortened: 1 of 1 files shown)


0xdeafcafe:
Closing in favor of the re-sliced 3-PR stack: #4498 (server dispatch) → #4499 (Liquid templating) → #4500 (staged authoring drawer). The ten ADRs (021-030) here are collapsed into five (022-026) in the new stack; the eight specs into five. See #4498 description for the full mapping.

PR #4465, opened 2026-06-01
fix(nlpgo): expose project secrets as `secrets.NAME` in code blocks

Aryansharma28:
## Problem

A user reported that the Studio Python code editor hints to reference project secrets as `secrets.NAME` (e.g. `secrets.CLOUDFLARE_ACCESS_CLIENT_ID`), but executing the code in a workflow fails:

```
NameError: name 'secrets' is not defined
```

`os.environ[...]` also fails.

## Root cause

There are two execution runtimes, gated by the `release_nlp_go_engine_enabled` flag:

- **Python (`langwatch_nlp`)** — injects a `secrets` namespace via `build_secrets_preamble` (`studio/utils.py`). `secrets.NAME` works. ✅
- **Go (`nlpgo`)** — the user's project is on this engine. `addEnvs.ts` *does* populate `workflow.secrets` in the DSL, and the Go DSL has the field (`dsl/types.go` `Secrets`), **but the code-block path never threaded it through**: `engine.runCode` → `codeblock.Request` → `runner.py` carried only `{code, inputs, outputs}`, and the runner exec'd user code in a bare namespace (`{"__name__": "__user_code__"}`). So `secrets` was undefined. ❌

`os.environ` fails because secrets live in the DSL, not the subprocess env. HTTP nodes were unaffected — they use the separate `{{ secrets.NAME }}` resolver (`ports.go` `SecretsResolver`), which only covers HTTP-block invocation.

This was a parity gap introduced during the dspy-rip migration to the Go engine.

## Fix

Thread `workflow.secrets` through the one path that lacked it:

1. **`engine.go`** — pass `req.Workflow.Secrets` into `runCode` (both call sites: `code` node and `agent_type: code`).
2. **`codeblock.go`** — add `Secrets` to `Request`; include it in the stdin payload.
3. **`runner.py`** — inject a `secrets` `SimpleNamespace` into the exec globals so `secrets.NAME` is real attribute access (mirrors the Python `build_secrets_preamble`).

Design notes:
- Injected as a **global**, not a source preamble — so user-code line numbers in tracebacks stay accurate.
- Left **undefined when no secrets are configured** — byte-parity with the Python preamble's no-op-on-empty behavior (so `secrets.X` with nothing configured raises `NameError` on both engines, and a missing key with secrets present raises `AttributeError`).
- `SimpleNamespace(**secrets)` takes values verbatim, so quotes/newlines/backslashes in a secret survive intact.

## Tests

- **Spec:** new Rule in `specs/nlp-go/code-block.feature` (3 scenarios: readable secret, undefined-secret → AttributeError, no-secrets → undefined parity).
- **Unit (`codeblock_test.go`):** secret exposed as namespace; special-char values; undefined secret → AttributeError; no-secrets → NameError parity.
- **Integration (`code_block_realistic_test.go`):** end-to-end — a `secrets` map on the workflow DSL reaches user code as `secrets.NAME` through `/go/studio/execute_sync`.

All green locally (`go test ./app/engine/... ./tests/integration/ -run CodeBlock`), `go vet` clean, `runner.py` compiles.

### DIFF
--- a/services/nlpgo/app/engine/blocks/codeblock/codeblock.go
+++ b/services/nlpgo/app/engine/blocks/codeblock/codeblock.go
@@ -87,7 +87,12 @@ type Request struct {
 	Code            string
 	Inputs          map[string]any
 	DeclaredOutputs []string
-	Timeout         time.Duration
+	// Secrets are the project's decrypted secrets (from the workflow
+	// DSL's `secrets` map). When non-empty the runner exposes them to
+	// user code as a `secrets` namespace so `secrets.NAME` works —
+	// parity with the Python executor's build_secrets_preamble.
+	Secrets map[string]string
+	Timeout time.Duration
 }
 
 // Result is what the executor returns.
@@ -130,6 +135,7 @@ func (e *Executor) Execute(ctx context.Context, req Request) (*Result, error) {
 		"code":    req.Code,
 		"inputs":  req.Inputs,
 		"outputs": req.DeclaredOutputs,
+		"secrets": req.Secrets,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("codeblock: marshal request: %w", err)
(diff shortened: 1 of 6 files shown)


sergioestebance (approved):
lgtm

Aryansharma28:
Relates to #4512 (umbrella: project secrets not resolving in nlpgo). This PR fixed the code-block half (`secrets.NAME`); HTTP-block half is #4509.

PR #1201, opened 2026-01-25
fix: improve dedupe logic, and fix span dropping issue in span storage event handler

0xdeafcafe:

Issue #1200, opened 2026-01-25
Fix prompt playground submission box bugs


rogeriochaves:
https://github.com/user-attachments/assets/b348ec8f-2a2e-4a29-9753-19e69dc5e98d

drewdrewthis:
Closing: only contains a video link with no description of specific bugs. Reopen with concrete bug descriptions if still relevant.

PR #1196, opened 2026-01-24
feat(evaluations-v3): implement HTTP agent support

drewdrewthis:

Issue #1195, opened 2026-01-24
Fix anthropic LLMs


rogeriochaves:
https://langwatch.slack.com/archives/D0A7MJ9GERW/p1769263364573419

PR #4470, opened 2026-06-01
fix(deps): bump uuid to >=11.1.1 across the monorepo (closes 7 alerts)

langwatch-agent:
## Why

`uuid` resolves to vulnerable versions (9.0.1 / 10.0.0 / 11.1.0, all `< 11.1.1`) as a **transitive** dependency across the monorepo, leaving **8 open Dependabot alerts** for GHSA-w5hq-g745-h8pq (missing buffer bounds check in `v3`/`v5`/`v6` when a `buf` argument is provided). All are the same advisory, fixed at `uuid 11.1.1`. No single Dependabot PR covers this because every copy is transitive.

## What changed

The override is pinned to `>=11.1.1 <12` (the patched 11.x line) rather than a bare `>=11.1.1` — the latter pulls `uuid 14.x` and widens the major jump unnecessarily. Each project's uuid copies collapse to a single `11.1.1`, and the lockfile diffs are uuid-only (no other package versions move).

| project | mechanism |
| --- | --- |
| `langwatch`, `skills`, `typescript-sdk` | `pnpm.overrides` in `package.json` |
| `mcp-server` | override added to `pnpm-workspace.yaml` (not `package.json` — a `pnpm.overrides` block there would silently disable the existing workspace security overrides) |
| `bullboard` | new `pnpm.overrides` block |
| `typescript-sdk/examples/langchain`, `mastra` | npm `overrides` + regenerated `package-lock.json` |

All resolve to `uuid 11.1.1` after the change.

## Alerts resolved

7 of the 8 uuid alerts: langwatch/langwatch#1194, langwatch/langwatch#1195, langwatch/langwatch#1196, langwatch/langwatch#1197, langwatch/langwatch#1198, langwatch/tasks#9, langwatch/langwatch#1200, langwatch/langwatch#1201 map to the eight manifests; this PR clears the seven listed above. (Each alert is the same MODERATE advisory; Dependabot tracks one per manifest.)

## One alert deferred (needs a human decision)

`typescript-sdk/examples/langgraph` is left untouched. Its `package-lock.json` cannot be regenerated because of a **pre-existing peer-dependency conflict unrelated to uuid**: the root pins `@langchain/core@^0.3.80`, while `@langchain/community@1.1.28` requires `@langchain/core@^1.1.38`, so `npm install` fails `ERESOLVE` before it can apply any uuid override. Forcing it with `--legacy-peer-deps` / `--force` would write a knowingly-incorrect lock. The honest fix is to first reconcile that example's `@langchain/*` versions (a separate change), after which the same uuid override applies cleanly. Leaving that one alert open for a human to schedule.

## How this was verified

- `pnpm@10.24.0 install --frozen-lockfile` succeeds for all pnpm projects; `npm install --package-lock-only` succeeds for the langchain and mastra examples. Every one resolves `uuid 11.1.1`.
- **langwatch** has no direct `uuid` usage in its own source (purely transitive), and its `typecheck` against a Prisma-generated baseline shows **zero uuid-related errors** (the remaining errors are pre-existing missing-codegen, identical to the baseline on `main`).
- Lockfile diffs reviewed: uuid-only, no collateral package bumps.

Note: regenerating the pnpm lockfiles used a temporary local `--config.minimumReleaseAge=0` to re-resolve under the repo's supply-chain cooldown. That flag is local-only and is not committed.


langwatch-agent:
Ready for human review. Opened by dependabot-scout as a consolidated uuid security upgrade (>=11.1.1 <12). Closes 7 of the 8 uuid alerts; the langgraph example is deferred on a pre-existing @langchain/core peer conflict unrelated to uuid (details in the PR description). Verified: frozen/npm installs across all touched projects resolve uuid 11.1.1, langwatch typecheck shows zero uuid-related errors. Advisory only: a human makes the merge call.

langwatch-agent:
CI note: the only red check is `test-unit (2)` (and the `langwatch-app-complete` aggregator that depends on it). This is **not** caused by the uuid bump — it is the pre-existing post-test worker wedge the workflow itself documents (`langwatch-app-ci.yml` lines 90 and 161: "a worker can wedge post-test (worker fails to exit)", which is why the job has a 25-minute timeout cap).

Evidence it is unrelated to uuid:
- The shard wedges on `src/prompts/prompt-playground/components/sidebar/__tests__/PublishPromptsList.test.tsx`, a React component test with no uuid usage anywhere in that module.
- Reproduced locally on this branch: the shard freezes after emitting `[open-handles] handles=1 ... {"MessagePort":1}` (a leaked MessagePort handle = the worker failing to exit), not on any uuid code path.
- The other three unit shards pass on identical code, and langwatch/langwatch#4457 (the hono PR on the same monorepo, no uuid change) passed all four shards.

The substantive checks are green: typecheck, build, lint, the other test-unit shards, and all integration shards. Re-running the wedged shard. Advisory only - a human makes the merge call.

PR #4475, opened 2026-06-01
fix(app): auto-recover from stale chunks after deploy via vite:preloadError

Aryansharma28:
## Problem

Opening a trace's details drawer on production was throwing:

> **Something went wrong** — Failed to fetch dynamically imported module: `https://app.langwatch.ai/assets/react-json-view-CugXrtI-.js`

This is the classic **Vite stale-chunk** failure. Vite emits content-hashed JS chunks; after a deploy the old hashes are removed from the CDN, but a tab opened *before* the deploy still references them. The trace-drawer JSON viewer (`react-json-view`) is **lazy-loaded**, so its `import()` only fires when you open a trace — and 404s with the message above. The error-boundary's "Try again" just re-requests the same dead URL, so it never recovers.

## Root cause

Two lazy-loading paths exist, and only one was protected:

| Path | Used by | Recovery before this PR |
|------|---------|--------------------------|
| `page()` helper in `routes.tsx` | route/page chunks | ✅ reloads once, 10s cooldown |
| `dynamic()` compat shim → `React.lazy` | **react-json-view**, Monaco, Foundry drawer | ❌ none → red toast |

## Fix

- Add a global **`vite:preloadError`** listener in `main.tsx` (registered once at boot). Vite fires this event whenever *any* dynamic `import()` chunk fails, so this recovers **every** component-level lazy import at once, not just routes.
- Extract the chunk-error detection + cooldown-guarded reload into a shared **`src/utils/chunkReload.ts`**.
- Refactor `routes.tsx` to reuse it — both paths now share **one** `chunk-reload-at` cooldown sentinel in `sessionStorage`, so they can't double-reload.

## Tests

- `specs/app-shell/stale-chunk-recovery.feature` — BDD spec for the behavior.
- `src/utils/chunkReload.test.ts` — 11 unit/integration cases: error classification across browser phrasings, the cooldown (first reload / within-cooldown no-op / after-cooldown reload), and that a dispatched `vite:preloadError` event is `preventDefault`'d and triggers the reload.

✅ 11/11 tests pass · ✅ changed files typecheck clean.

## Browser QA (real Chromium via Playwright)

Built a minimal Vite app importing the **real** `chunkReload.ts`, then reproduced the production condition: load the page, delete the content-hashed chunk from disk + restart the server (a deploy removing the old hash), then fire the lazy `import()`.

| Step | Result |
|------|--------|
| Initial load | page load #1 · no reload ([before](https://i.img402.dev/pnl180t56r.png)) |
| Click → stale chunk `import()` fails → `vite:preloadError` | page **auto-reloads** → load #2 · "recovered: auto-reloaded after stale chunk" ([after](https://i.img402.dev/0dveqtr510.png)) |
| 2 more failed imports within the 10s cooldown | **no reload loop** — load count held steady, sentinel unchanged |

Confirms the exact recovery that was missing in production, and that the cooldown prevents a reload loop when a chunk is genuinely broken.


langwatch/src/utils/chunkReload.ts
```diff
@@ -0,0 +1,77 @@
...
+ * broken chunks can't trap the user in a reload loop.
+ */
+export function forceReloadOnce(): boolean {
+  if (typeof window === "undefined") return false;
+
+  const lastReloadAt = Number(sessionStorage.getItem(RELOAD_AT_KEY) ?? "0");
+  if (Date.now() - lastReloadAt <= RELOAD_COOLDOWN_MS) return false;
+
+  sessionStorage.setItem(RELOAD_AT_KEY, String(Date.now()));
+  window.location.reload();
```
langwatch-agent, line 50: **P2** — This reload only recovers stale chunks if the next navigation fetches a fresh HTML shell. Vite explicitly calls out `Cache-Control: no-cache` on HTML as a prerequisite for [`vite:preloadError` recovery](https://vite.dev/guide/build.html#load-error-handling), but the production SPA fallback in `src/server/static-handler.ts` pipes `index.html` without any cache header (and direct `/index.html` is the same). A cached shell can therefore reload back into the same removed hashes and leave users stuck. Please set an explicit revalidation/no-store policy for HTML responses alongside this change.

langwatch/src/utils/chunkReload.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Register the global recovery for component-level lazy imports. Vite fires
+ * `vite:preloadError` on `window` whenever a dynamically imported chunk fails
+ * to load; the event itself is the chunk-error signal, so we reload directly.
+ */
+export function registerChunkReloadListener(): void {
+  if (typeof window === "undefined") return;
+  window.addEventListener("vite:preloadError", (event) => {
+    // Stop Vite from rethrowing the error to the page; we recover by reloading.
+    event.preventDefault();
```
langwatch-agent: **P2** — Only suppress Vite's preload error when a reload was actually scheduled. On a second failure inside the 10s cooldown, `forceReloadOnce()` returns `false`, but this unconditional `preventDefault()` still stops Vite from throwing the import error. That leaves the lazy import without either recovery or the normal error-boundary path. Use `if (forceReloadOnce()) event.preventDefault();` and add a second-event-within-cooldown test.

langwatch-agent (commented):
Found 2 P2 recovery-path issues; details are inline. `git diff --check origin/main...origin/review/pr-4475` passes. CI is still running. The PR description also references `specs/app-shell/stale-chunk-recovery.feature`, but that file is not present in this diff.

langwatch/src/utils/chunkReload.ts
Aryansharma28: Good catch — fixed in ddab43cc6. The listener now only calls `event.preventDefault()` when `forceReloadOnce()` actually scheduled a reload; a second `vite:preloadError` inside the cooldown now propagates to the error boundary instead of being swallowed. Added a `second vite:preloadError within the cooldown` test asserting `defaultPrevented === false`.

langwatch/src/utils/chunkReload.ts
Aryansharma28: Valid — fixed in ddab43cc6. `static-handler.ts` now sets `Cache-Control: no-cache` on HTML responses (both the SPA fallback and a direct `/index.html` via `serveStaticFile`), so a post-deploy reload always revalidates the shell and picks up the new chunk hashes rather than re-loading the removed ones. Added tests covering both HTML paths. (Missing `/assets/*` already returns a real 404 with `no-store`, so the chunk failure still surfaces as `vite:preloadError`.)

0xdeafcafe (approved):

rogeriochaves (approved):

Issue #561, opened 2025-08-26
Editing a column name in the datasets removes the column data


liorheber:
Hi,

I created a dataset and filled it with data.
I then decided I want to rename one of the column names and then the column values turned to undefined.

drewdrewthis:
Thanks for bringing this up @liorheber -- this has been resolved in #607 

----

## GitHub Issue Closure Report: Dataset Column Rename Data Loss

### Root Cause
The data loss occurred in `langwatch/src/optimization_studio/utils/datasetUtils.ts` in the `tryToMapPreviousColumnsToNewColumns` function. 

**Before the fix:**
- The function only mapped columns with exact name matches
- When a column name was changed (e.g., `"input"` → `"prompt"`), no mapping was created
- Data for unmapped columns was silently dropped during the update process

### Fix Details

**Commit:** `f6d3c15e34734131ada513579f61afd8dffae80d`  
**PR:** #607  
**Author:** Richard Huth  
**Date:** September 5, 2025  
**Title:** "fix: column name edit"

**Changes Made:**
The fix added a **position-based fallback mapping** strategy to preserve data when column names are changed. After attempting exact name matches, the function now:

1. Identifies unmapped columns from both old and new column sets
2. Maps remaining columns by their position in the array
3. Preserves data by assuming columns in the same position represent the same logical field

PR #4493, opened 2026-06-02
fix(stored-objects): externalise AI-SDK file+audio parts (scenario voice audio leak)

rogeriochaves:
## Summary

- Adds a `file`-shape branch to the content-extractor visitor (both sync and async dispatchers). Audio mediaTypes route to `inputAudio`, non-audio file payloads route to `binary`.
- Teaches `inputAudio` to prefer an explicit `mimeType` over the format allowlist, and normalises the rewrite output so file-shape inputs produce clean `input_audio` / `binary` references (no chimera of `type:"file"` + externalised fields).
- 2 new unit tests covering the file-shape extraction paths.

## Why

`getSuiteRunData` for the scenario `voice-customer-support-demo` set was returning 90+ MB because base64 audio bytes were landing inline in ClickHouse `Messages.Content`. Root cause: the typescript scenario SDK's `createAudioMessage` emits the canonical AI-SDK shape (`{type:"file", mediaType:"audio/pcm16", data:"<base64>"}`), but the content-extractor visitor had no `file`-shape branch — so the audio dropped into the `unknown` no-op handler and was never externalised to stored-objects.

This is the **defence-in-depth** half of the fix. The companion `langwatch/scenario` PR (#561) translates file+audio parts to `input_audio` shape in the SDK before sending, AND stops pre-stringifying array content so the visitor gets a chance to walk it. With both deployed, current SDK builds hit the `input_audio` branch; older SDK builds and any future non-scenario caller shipping raw AI-SDK file parts are caught here.

## Test plan

- [x] `pnpm test:unit run src/server/stored-objects/__tests__/content-extractor.unit.test.ts` — 18/18 pass (16 existing + 2 new)
- [ ] Re-run a voice scenario from `langwatch/scenario` PR #561 after both ship; confirm `getSuiteRunData` payload drops back to expected size and `Messages.Content` for audio messages contains the externalised URL instead of inline base64
- [ ] Open `/<slug>/simulations/voice-customer-support-demo`; confirm audio plays via the externalised URL

## Companion PRs

- `langwatch/scenario` PR #561 — SDK side (translate file+audio → input_audio, stop pre-stringifying)
- Sibling langwatch PR — projection size cap as a defensive seatbelt against any future shape regression (separate review surface)

### DIFF
--- a/langwatch/src/server/stored-objects/__tests__/content-extractor.unit.test.ts
+++ b/langwatch/src/server/stored-objects/__tests__/content-extractor.unit.test.ts
@@ -442,6 +442,128 @@ describe("extractInlineMediaFromEvent", () => {
     });
   });
 
+  describe("when an event has an AI-SDK file+audio part (typescript scenario SDK shape)", () => {
+    /** @scenario "file parts with audio/* mediaType are externalized to a clean input_audio reference" */
+    it("calls storeFromBytes with the audio mediaType and rewrites the part to {type:'input_audio', input_audio:{url, mimeType}}", async () => {
+      const base64Payload = makeBase64Payload("PCM16_AUDIO_BYTES");
+      const mimeType = "audio/pcm16";
+      const storedId = "stored-file-audio-id";
+
+      const service = makeService({
+        storeFromBytes: vi.fn().mockResolvedValue({
+          id: storedId,
+          mediaType: mimeType,
+          isDuplicate: false,
+        }),
+      });
+
+      // Shape emitted by the typescript scenario SDK's `createAudioMessage`
+      // (voice/messages.ts) before the SDK-side translation patch. Older
+      // builds (and any future caller that ships raw AI-SDK file parts)
+      // hit this branch; the extractor must externalise the bytes rather
+      // than let them flow to ClickHouse Messages.Content inline.
+      const event = makeEventWithContent([
+        { type: "text", text: "Hi" },
+        {
+          type: "file",
+          mediaType: "audio/pcm16",
+          data: base64Payload,
+        },
+      ]);
+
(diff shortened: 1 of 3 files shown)


langwatch/src/server/stored-objects/__tests__/content-extractor.unit.test.ts
```diff
@@ -442,6 +442,128 @@ describe("extractInlineMediaFromEvent", () => {
     });
   });
 
+  describe("when an event has an AI-SDK file+audio part (typescript scenario SDK shape)", () => {
+    /** @scenario "file parts with audio/* mediaType are externalized to a clean input_audio reference" */
```
langwatch-agent: **P2** — This new `@scenario` binding has no matching scenario under `specs/`, and the same is true for the new binding near line 510. The feature-parity check is currently failing as a result. Please add the corresponding feature scenarios, or remove the annotations if these cases are intentionally unit-only.

langwatch-agent (commented):
Found one additional CI/spec parity issue; details inline. The existing CodeRabbit MIME-normalization thread is also valid: MIME types are case-insensitive, so the new audio routing should normalize before matching. `git diff --check origin/main...origin/review/pr-4493` passes.

rogeriochaves:
Addressed both review threads in `08239ce`:

1. **MIME normalisation** (@coderabbitai) — lowercase `mediaType` before the `audio/` prefix check and before passing into `mediaTypeToAudioFormat`. Defensive normalisation also inside the helper. New unit test covers mixed-case `Audio/WAV` → `input_audio` routing.
2. **Unbacked `@scenario` annotations** (@langwatch-agent) — dropped. Those cases are intentionally unit-only; fabricating a feature scenario just to satisfy the parity gate would dilute it.

19/19 unit tests pass.

PR #4494, opened 2026-06-02
feat(simulations): cap per-message Content/Rest size in projection (defensive seatbelt)

rogeriochaves:
## Summary

- Caps `Messages.Content` and `Messages.Rest` at 64 KiB per message in the simulation-run fold projection (both MessageSnapshot and TextMessageEnd handlers).
- Oversized values become a clearly-marked truncation placeholder (`[truncated: message content was N bytes (cap M); likely inline media that was not externalised to stored-objects]`).
- Emits a structured `warn` log on every truncation: `{scenarioRunId, messageId, messageRole, field, byteLength, maxBytes}` — alertable in production.
- 4 new unit tests; 26 existing still pass.

## Why

A voice scenario set was leaking 90+ MB of base64 audio inline in ClickHouse `Messages.Content` because the stored-objects pipeline failed to externalise the SDK's audio file parts. `getSuiteRunData` then slurped the first 6 messages of every scenario back to the browser.

The SDK fix (langwatch/scenario PR #561) and the content-extractor fix (PR #4493) together address the root cause for the known shapes. **This patch is the defensive seatbelt** — any future shape we forget to teach the extractor about, or any upstream refactor that re-stringifies array content before it reaches the extractor, gets a hard size cap at persistence time instead of turning into a multi-megabyte list response.

## Threshold rationale

64 KiB is generous enough that no normal text turn — including verbose multi-paragraph assistant replies and long tool-call argument JSON — is ever truncated. Messages exceeding it are almost always inline binary media: that's exactly the case we want to fail-fast on.

## Test plan

- [x] `pnpm test:unit run src/server/event-sourcing/pipelines/simulation-processing/projections/__tests__/simulationRunState.foldProjection.unit.test.ts` — 30/30 pass
- [ ] Re-deploy and re-run any voice scenario from the failing project; confirm even with a misbehaving SDK upstream the list page loads in normal time and the truncation marker appears in the message row
- [ ] Set up a CloudWatch metric/alarm on the warn log `msg: "simulation message Content exceeds size cap"` — count > 0 indicates an SDK regression slipping past the extractor

## Companion PRs

- `langwatch/scenario` PR #561 — SDK side (translate file+audio → input_audio; stop pre-stringifying array content)
- langwatch PR #4493 — content-extractor side (add `file`-shape branch, normalise rewrites)

### DIFF
--- a/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
@@ -55,16 +55,24 @@ const MAX_MESSAGE_REST_BYTES = 64 * 1024;
  * The returned marker has a stable prefix so monitoring + retroactive scans
  * can find affected rows.
  */
-function capOversizedString(
-  value: string,
-  maxBytes: number,
-  field: "Content" | "Rest",
-  ctx: { scenarioRunId: string; messageId?: string; messageRole?: string },
-): string {
-  // String length is char-count, not bytes; rough byte check via
-  // Buffer.byteLength gives an accurate UTF-8 size. Fast path: skip the
-  // Buffer allocation when length alone already proves the string is small.
-  if (value.length <= maxBytes) return value;
+function capOversizedString({
+  value,
+  maxBytes,
+  field,
+  ctx,
+}: {
+  value: string;
+  maxBytes: number;
+  field: "Content" | "Rest";
+  ctx: { scenarioRunId: string; messageId?: string; messageRole?: string };
+}): string {
+  // String length is char-count (UTF-16 code units); UTF-8 may use up to 3
+  // bytes per code unit (4 for surrogate pairs, but a pair occupies two code
+  // units so the per-code-unit ceiling is still 3). The only safe length-only
+  // short-circuit is the inverse bound: when length*3 <= maxBytes the UTF-8
+  // byte length is guaranteed to fit. Using length <= maxBytes as the bypass
+  // would let a multibyte string ~3× over the cap slip through.
+  if (value.length * 3 <= maxBytes) return value;
   const byteLength = Buffer.byteLength(value, "utf8");
   if (byteLength <= maxBytes) return value;
   projectionLogger.warn(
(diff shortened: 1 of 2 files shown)


langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
```diff
@@ -234,12 +285,16 @@ export class SimulationRunStateFoldProjection
           content = JSON.stringify(m.content);
         }
 
+        const messageId = typeof m.id === "string" ? m.id : "";
+        const messageRole = typeof m.role === "string" ? m.role : "";
+        const ctx = { scenarioRunId: state.ScenarioRunId, messageId, messageRole };
```
langwatch-agent: **P3** — Use `state.ScenarioRunId || event.data.scenarioRunId` in the warning context. This handler explicitly supports snapshots arriving before the started event, but on that path `state.ScenarioRunId` is still empty while the event already has the run ID. An oversized first snapshot therefore emits the new alert without the identifier needed to locate the affected run. The `TextMessageEnd` context below has the same issue when an end event is folded before start.

langwatch-agent (commented):
Found one additional observability issue; details inline. The existing CodeRabbit multibyte fast-path thread is valid and should be addressed: `value.length <= maxBytes` does not prove the UTF-8 byte length fits the cap. `git diff --check origin/main...origin/review/pr-4494` passes. The failing integration shards appear unrelated: one could not pull `redis:alpine`, and the other failed an existing ClickHouse span-storage assertion outside this diff.

rogeriochaves:
Addressed all three review threads in `a6ef1f2`:

1. **Named-parameter destructuring on `capOversizedString`** (@coderabbitai) — switched to `{ value, maxBytes, field, ctx }`. Four call sites updated; can't transpose `field`/`maxBytes` by mistake now.
2. **UTF-8-safe fast-path** (@coderabbitai) — replaced `value.length <= maxBytes` (which let a 3×-over-cap multibyte string slip past unmeasured) with the safe inverse bound `value.length * 3 <= maxBytes`. New unit test covers the multibyte case: 30k 4-byte emojis (~120 KiB bytes) → truncates correctly.
3. **`scenarioRunId` fallback in warn context** (@langwatch-agent) — both `MessageSnapshot` and `TextMessageEnd` handlers can fold before the run-started event. Use `state.ScenarioRunId || event.data.scenarioRunId` so oversized first-snapshot warns carry the run id instead of arriving id-less.

31/31 unit tests pass.

PR #4499, opened 2026-06-02
feat(automations): liquid templating for trigger notifications (PR 2/3)

0xdeafcafe:
## Summary

PR 2 of 3 in the trigger-outbox stack — stacked on #4498. Adds the rendering engine the outbox dispatch worker (PR 1) uses to materialise email / Slack output from customer-authored Liquid templates.

- **ADR 028** — Liquid templates for user-customizable trigger notifications. (Renumbered from 024 during rebase onto main.)
- Spec: `automations/notification-templates.feature` — engine contract, digest `matches[]` shape, Block Kit allow-list, fall-back semantics.

Code shipped:
- `src/shared/templating/` — sandboxed Liquid engine (`strictVariables: false`, 500ms render timeout, process-local parsed-template cache around `engine.parse`), default templates, Markdown → HTML pipeline (`marked` + `sanitize-html`, no `` on the allowlist), Slack Block Kit allow-list, email layout wrapper, render helpers with graceful fallback, Liquid `validate` for save + preview paths.
- `src/server/triggers/triggerNotifier.ts` — production notifier delivering already-rendered output to Slack incoming webhook + SES.
- `src/server/app-layer/triggers/trigger-template.service.ts` + `errors.ts` — `validateTemplateDraft`, `renderPreview`, `testFire` with the ADR-028 non-suppressible test-fire banner. `testFire` validates the draft before dispatching so the Slack-discriminator contract can't be bypassed.
- App-layer wiring (`app.ts` / `dependencies.ts` / `presets.ts`) exposes `getApp().triggerTemplates`.

Adds `marked` + `sanitize-html` + `@types/sanitize-html` to `langwatch/package.json` (`liquidjs` was already on main).

## Review feedback addressed

- **Slack discriminator enforced inside `testFire`** so a "Send test" path can't bypass `validateTemplateDraft` and silently dispatch a Block Kit JSON payload as plain text.
- **Block Kit defaults escape all dynamic strings** via Liquid's `| json` filter. Previously `trigger.name`, `trigger.alertType`, and the footer trace/edit URLs were interpolated raw into JSON string literals; a quote or newline in any of them would have broken `JSON.parse` and silently downgraded the default Slack notification to plain text.
- **`` removed from the email Markdown allowlist** to close the tracking-pixel leak. (Customer-authored templates could otherwise embed a third-party `` and leak recipient metadata — read time, IP, client hints — to whoever owned the URL.)
- **Subject clipping moved after the `[Test fire]` prefix** so the final subject respects `EMAIL_SUBJECT_MAX_LENGTH` end-to-end.
- **Custom template's `missingVariables` diagnostics preserved on Block Kit fallback.** A `JSON.parse` / allowlist failure no longer overwrites the author's typo diagnostics with the framework default's, so preview / test-fire reports the variable typos the author needs to see.
- **`detectMissingVariables` walks full property paths.** Switched from `globalVariablesSync` (top-level names only) to `globalVariableSegmentsSync`, so `{{ project.nmae }}` is now correctly reported as `project.nmae` instead of silently passing because `project` exists.
- **Explicit parsed-template cache around `engine.parse`.** LiquidJS's built-in `cache: true` only populates for `renderFile` (file-path keyed) and is a no-op for `parseAndRender(string)`, so every digest dispatch was re-parsing the same source. Added a process-local string-keyed LRU and switched `renderLiquid` to render the cached AST directly.
- **`matches` added to `TEMPLATE_VARIABLES`** so Monaco autocomplete and the unknown-variable detector recognise the canonical iteration root.
- **Stale ADR references cleaned up** (ADR-024 / ADR-026 / ADR-028 across `templating/*` + `errors.ts`).

## Stack

| | |
|---|---|
| PR 1 → #4498 | server dispatch infrastructure |
| **PR 2 (this)** | `src/shared/templating` + trigger-template service |
| PR 3 → #4500 | staged authoring drawer + provider model + dispatch-health view |

## Test plan

- [x] CI green on this branch.
- [x] Unit tests pass for `shared/templating/*` and `trigger-template.service`.
- [ ] Manual smoke: invoke `triggerTemplates.renderPreview` against a non-trivial Liquid template and confirm the rendered output matches expectations.
- [ ] Confirm a template that throws falls back to the framework default and surfaces the error.


0xdeafcafe:
@coderabbitai review

PR #4500, opened 2026-06-02
feat(automations): staged authoring drawer + provider model + dispatch-health surface (PR 3/3)

0xdeafcafe:
## Summary

PR 3 of 3 in the trigger-outbox stack — stacked on #4499. The operator-facing UI for the substrate (PR 1 #4498) and templating engine (PR 2 #4499). Replaces the legacy three-form authoring split (`AddAutomationDrawer`, inline `TriggerForm`, `EditTriggerTemplatesDrawer`) with one staged drawer.

- **ADR 029** — Automation operator surfaces: authoring drawer + dispatch-health view. (Renumbered from 026 during rebase onto main.)
- Spec: `automations/authoring-drawer.feature` — drawer staging, template-authoring sub-flow, dispatch-health columns on the settings list.

Code shipped:
- `src/automations/providers/` — provider model, one folder per action type (`annotationQueue`, `dataset`, `email`, `slack`), each shipping `shared` / `client` / `server` peers. Parity unit test asserts every action has a registered provider.
- `src/features/automations/` — the staged drawer:
  - `AutomationDrawer` orchestrator (data loading, preview pipeline, test-fire + upsert mutations). Edit-prefill wraps `JSON.parse(row.filters)` in try/catch so a malformed legacy row falls back to an empty filter set instead of crashing the page.
  - `MainSectionList` with `SectionRow` chrome for Identity / When / Type / Setup / Cadence / Test Fire.
  - `secondaries/` — Configuration, Filters, Cadence (`CadenceField` + `TraceDebounceField` live here), `SecondaryDrawerShell` with width-toggle chrome.
  - `editors/` — Monaco Liquid language + JSON Schema + template-authoring scaffolding. Schemas route through the shared `monacoSchemas.ts` registry so sibling editors don't wipe each other.
  - `logic/draftReducer` (distributive `SetSliceAction` so SET_SLICE can't carry a mismatched slice) + `state/automationStore` (zustand) + `state/selectors`.
- `src/server/api/routers/automations.ts` — `upsert` accepts `notificationCadence` + `traceDebounceMs` + templates; per-action shape validation via the provider registry's Zod schemas; `testFireTemplate` procedure surfaces the PR 2 service.
- Legacy: `src/components/AddAutomationDrawer.tsx` removed; `src/pages/[project]/automations.tsx` opens the staged drawer from Edit; command-bar + drawer registry route `/drawer/automation*` to the new drawer.

## Review feedback addressed

- **Non-notify cadence pinned to `immediate` on update** so a notify → persist edit can't leak the stale digest cadence onto the row.
- **`create()` email recipient validation aligned with `upsert()`** — format-only via the shared `validateEmailRecipientFormats` helper, external addresses allowed with the existing "External" UI badge. Two server contracts for SEND_EMAIL across create vs edit removed.
- **`EMAIL_RX` tightened** to reject whitespace / CRLF inside addresses, blocking header-injection / CRLF-smuggling shapes (`"a@b.com\nBcc: …"`) before they reach SendGrid / SES non-raw `to` slots.
- **`recipientCount` guard moved into the displayed string** — a successful attempt with an undefined count no longer renders "delivered to undefined recipients".
- **`TraceDebounceField` clamps on blur/Enter**, not every keystroke — manual entry can now transit through values below `MIN_SECONDS` while typing (e.g. typing `6` toward `60` doesn't snap to MIN mid-keystroke).
- **Implementation details removed from `authoring-drawer.feature`** (`TriggerSent` claim ledger / `ReactorOutbox` grouping). Scenarios describe user-visible behavior, not the table they're served from.
- **Stale ADR references cleaned up** (ADR-024 / ADR-025 / ADR-026 / ADR-030 in `monacoSchemas`, `draftReducer`, `TraceDebounceField` → correct new numbers).

## Follow-up commits (936c17cc9, f8177b109)

- `936c17cc9 feat(automations): default Slack Block Kit templates + cadence-aware picker` — adds four shippable Liquid Block Kit defaults (`trace_alert_compact`, `eval_failure_detailed`, `digest_compact`, `digest_inline_rich`) and a cadence-aware picker above the Slack template editor. Picker only shows templates that fit the current `notificationCadence` and flags one as `Default` based on cadence + filter shape. Selection writes the chosen `.liquid` source straight into `slice.template.value`. `ConfigFormCtx` carries `cadenceMode` + `hasEvaluationFilter` so the picker reads draft state without back-importing the automations store (which TDZ-cycles through the providers registry). `markdown` block type added to `blockKitAllowlist` so the new templates' `## headings`-style trace input/output render natively.
- `f8177b109 fix(mailer): trigger emails route To=hashed no-reply, recipients=BCC` — production trigger emails (and the authoring-drawer Send test path) now put `LangWatch Triggers <no-reply+<HMAC>@<domain>>` in `To:` and the configured recipients in `Bcc:`. The local-part tag is `HMAC-SHA256(triggerId, NEXTAUTH_SECRET)` truncated to 12 hex chars — stable per trigger (bounce-attribution surface), unguessable without the secret. Closes the recipient-enumeration leak between receivers and reduces the blast radius of CRLF-smuggling shapes that previously rode in the recipient `To:` header. `TriggerNotifier.sendEmail` interface changes to `{ to: string, bcc: string[], subject, html }`; both production dispatchers (outbox + inline pipeline) thread `trigger.id` through. SES `SendEmail` uses `BccAddresses`; raw-MIME path uses the `Destinations` envelope so no `Bcc:` header is rendered. SendGrid uses `bcc`/`replyTo` keys.

## Stack

| | |
|---|---|
| PR 1 → #4498 | server dispatch infrastructure |
| PR 2 → #4499 | `src/shared/templating` + trigger-template service |
| **PR 3 (this)** | staged authoring drawer + provider model + dispatch-health view |

## Test plan

- [x] CI green on this branch.
- [x] `pnpm typecheck` passes on the cumulative stack.
- [x] Unit tests pass for `features/automations/logic/`, `features/automations/state/`, `automations/providers/__tests__/parity`, `features/automations/editors/__tests__/`.
- [ ] Manual smoke: open the staged drawer from settings (no traces context), author an email trigger end-to-end (Identity → When → Setup → Cadence → Test Fire → Save), confirm the row appears in the settings list with last-triggered / fired-count columns lit up after a sample-trace fire.
- [ ] Open the cadence secondary drawer on a notify trigger; confirm cadence dropdown and `traceDebounceMs` field both work and round-trip on Save.
- [ ] Open an existing trigger via Edit; confirm all sections pre-fill from the saved row.


0xdeafcafe:
@coderabbitai review

PR #4515, opened 2026-06-02
fix(retention): drop _size_bytes MATERIALIZED column to relieve merge memory pressure

rogeriochaves:
## Why

The retention work in #4147 added a `_size_bytes` column to 11 ClickHouse tables, declared as `MATERIALIZED byteSize(<payload columns>)`. On heavy-payload tables (`event_log.EventPayload`, `stored_spans.SpanAttributes/ResourceAttributes/...`, `trace_summaries.ComputedInput/Output/...`, `evaluation_runs.Inputs`) this materially raised the memory cost of background merges on already-fat partitions.

In prod on 2026-06-02 it contributed to a merge memory ceiling outage:

- `event_log` MERGE_PARTS `new_part_name = 202622_1190507_1223359_196` failed with `MEMORY_LIMIT_EXCEEDED` ~9.5 GiB against the 14 GiB per-server cap, looping on ~23s exponential backoff, started ~2h after #4147 deployed.
- A pre-existing `evaluation_runs` merge (since 2026-05-23) was already at 559 retries with the same OOM signature — independent of #4147, but the new MATERIALIZED column compounded headroom pressure cluster-wide.
- 37–44 `trace_summaries FINAL` queries piled up holding 6–10 GiB combined, the ingestion pipeline went slow / stuck.

Per ClickHouse semantics MATERIALIZED columns are computed lazily on read and drifted onto disk as parts merge. So every merge of a pre-existing fat part now had to evaluate `byteSize()` over the entire payload column set before writing the new part. Dropping the column removes that merge-time cost.

## What changed

- New migration **`00033_drop_retention_size_bytes_materialized.sql`** drops `_size_bytes` on all 11 retention-managed tables. `DROP COLUMN` is metadata-only at ALTER time and reads no row data — column files are removed from each part on disk lazily.
- **`StorageMeterService`** now stubs both `getTotalStorageBytes` and `getStorageBreakdown` to return zero with a debug log. The data-retention settings page (the only consumer) still renders; storage usage shows as 0 B until a replacement metering strategy lands.

## What is NOT changed

- **Retention enforcement is fully preserved.** The TTL DELETE expression references only `_retention_days`, which stays as a small UInt16 column stamped at ingestion. No change to ingestion writes, no change to TTL reconciler, no change to `RETENTION_MANAGED_TABLES`.
- The retention-managed table set and category map are unchanged.

## Production triage applied alongside

To stop the active OOM-retry loop while this PR lands, `SYSTEM STOP MERGES` was applied to `langwatch.event_log` and `langwatch.evaluation_runs` on all three CH replicas. CH internal MemoryTracking dropped from ~9.1 GiB → ~7.8 GiB immediately. Once this migration applies the column is gone and merges can resume safely (`SYSTEM START MERGES`).

## Follow-ups (separate PRs)

- **Storage metering rebuild** — a strategy that does not require a per-row `byteSize` MATERIALIZED column. Options under consideration: periodic batch sampling, or `system.parts.bytes_on_disk` aggregated at table level with per-tenant proportional allocation.
- **Pre-existing `evaluation_runs` merge stuck since 2026-05-23** — separate from #4147. Likely needs raising the CH per-server memory cap (node has ~22 GiB OS-available headroom vs the 14 GiB CH cap) or narrowing the selector via per-table merge-size settings.

## Test plan

- `pnpm typecheck` clean.
- Existing retention unit/integration tests remain green (the metering surface was untested at the unit level; the consumers — authz, read, policy, TTL reconciler — are unchanged).

### DIFF
--- a/langwatch/src/server/clickhouse/migrations/00033_drop_retention_size_bytes_materialized.sql
+++ b/langwatch/src/server/clickhouse/migrations/00033_drop_retention_size_bytes_materialized.sql
@@ -1,0 +1,120 @@
+-- +goose Up
+-- +goose ENVSUB ON
+
+-- Drop _size_bytes MATERIALIZED column from all retention-managed tables.
+--
+-- The column was added in 00032 as `MATERIALIZED byteSize(<payload cols>)`.
+-- ClickHouse evaluates that expression on the heavy payload columns
+-- (EventPayload, SpanAttributes, ResourceAttributes, ComputedInput/Output,
+-- Inputs, …) during background merges as it drifts the value onto disk for
+-- pre-existing parts. On heavy-payload tables this materially increased
+-- merge memory usage and combined with already-large parts to push merges
+-- past the per-server memory cap, producing MEMORY_LIMIT_EXCEEDED failures
+-- on MERGE_PARTS replication queue entries and an exponential-backoff
+-- retry loop that pressured the cluster.
+--
+-- Retention enforcement itself does NOT depend on `_size_bytes` — only
+-- `_retention_days` is referenced by the TTL DELETE expression. The storage
+-- metering surface that used `SUM(_size_bytes)` is reworked alongside this
+-- migration to a path that does not require a per-row byteSize column.
+--
+-- DROP COLUMN is metadata-only at ALTER time and reads no row data; the
+-- column files are removed from each part on disk lazily.
+
+-- event_log
+-- +goose StatementBegin
+ALTER TABLE ${CLICKHOUSE_DATABASE}.event_log
+  DROP COLUMN IF EXISTS `_size_bytes`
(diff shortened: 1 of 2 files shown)


rogeriochaves:
Closing — wrong fix. The MATERIALIZED byteSize() cost is one-time during the backfill tail (old parts pay it once as they're merged forward; new inserts pay a cheap per-row cost; future merges of already-materialized parts pay nothing). Today's outage was dominated by the orphan-sweep hot loop on the broken workers image (BullMQ rejected the ':' in the chain jobId → QueueWithFallback ran sweepProject inline on every ingest event). That has been resolved by rolling langwatch-workers back to git-69774ea. CH-native metering is the right long-term answer — single source of truth, byte-accurate to what CH actually stores, no per-insert app plumbing.

PR #4520, opened 2026-06-03
feat(traces-v2): drive-through audit pass — sidebar, drawer, table, toolbar

0xdeafcafe:
## Summary
21 sliced commits closing the audit-feedback loop on traces v2 — driving across the explorer with chrome-devtools surfaced a long list of small "this doesn't quite work" calls that this PR addresses, plus three meatier pieces:

- **Facet manager popover** (`Settings2` in sidebar header) — one-stop "edit which facets are in my sidebar" picker with grouped checkboxes, per-key icons, search, and reset-to-defaults.
- **Discover (sidebar facet) SSE push + canonical window snap** — active viewers see fresh facet values within ~60s without each one paying the ClickHouse cost; arbitrary windows collapse to canonical buckets (1h/6h/24h/7d/30d) so viewers of "last hour" share a cache slot.
- **Drawer defaults overhauled** — Summary is now the default `viewMode`, all accordions open on first paint (empty states do the educating), SpanDetailPane scrolls internally.

### Commit topology

| # | Title |
|---|---|
| 1 | `chore` remove flame + span list view subtrees |
| 2 | `fix(layout)` banner radius + restore content flow |
| 3 | `feat` facet manager popover + per-key visibility store |
| 4 | `feat` Summary default + accordions expanded + span scroll |
| 5 | `fix` row colors unified + column reorder propagates to cells |
| 6 | `feat` Since/Timestamp time columns + shared hover card |
| 7 | `feat` toolbar dividers + inline column reorder + slim time picker |
| 8 | `feat` drawer header polish (chips, metric pills, signals) |
| 9 | `feat` mode switch + viz strip + presence lifts to shell |
| 10 | `fix` drawer popovers lazyMount + hover-only annotation row |
| 11 | `feat` RawJsonDialog: pretty/minify, wrap, search, size readout |
| 12 | `feat` waterfall rows pin to span tabs + cost readout |
| 13 | `feat` query breakdown chips on empty + calmer 'N new' pill |
| 14 | `feat(server)` per-span cost in summary endpoint |
| 15 | `feat(server)` SSE-push discover refreshes + canonical window snap |
| 16 | `feat(query-language)` removeImplicitTermFromQuery mutation |
| 17 | `feat` Simplified lens + time-column siblings + conversation sections |
| 18 | `feat(filterStore)` removeFreeText action |
| 19 | `feat(formatters)` verbose relative + ISO + local timestamp helpers |
| 20 | `fix` drawer shortcuts + URL sync follow the view redesign |
| 21 | `fix` error/new counts follow time range, calmer pulse, comfortable default |

## Test plan
- [x] `pnpm typecheck` clean
- [x] Driven through with chrome-devtools (dogfood@langwatch.local) — verified:
  - Summary opens by default (URL: `?drawer.mode=summary`)
  - All accordions open on first paint
  - Span tab scrolls (verified via JS `scrollTop=500` test)
  - Facet manager popover anchors below trigger (`bottom-start`)
  - Column reorder via dropdown propagates from headers to cells
  - Sticky checkbox cell colors match row body at rest + hover
  - Banner radius matches page chrome's top-left curve
- [ ] Manual: confirm `onDiscoverUpdate` SSE arrives within ~60s of new traces on a busy tenant
- [ ] Manual: verify Simplified lens columns look right end-to-end
- [ ] Manual: verify span-cost shows in waterfall rows that have `gen_ai.usage.cost`

## Review fixes (12 follow-up commits)

Auto-reviewers (CodeRabbit, Copilot, GitHub Code Quality) raised 14 threaded findings; a follow-up manual pass from @sergioestebance surfaced 4 more. All addressed in follow-up commits; threads resolved on GitHub.

| # | Title |
|---|---|
| 22 | `fix` keyboard a11y on hide/pin buttons + metadata key validation + drop redundant ternary |
| 23 | `fix` query breakdown classifies comparisons + single-shot bare-term removal |
| 24 | `fix` keep AST-active categorical sections mounted on empty discover |
| 25 | `fix(ColumnsDropdown)` reorder uses visible-order index space |
| 26 | `fix(formatters)` clamp future timestamps + drop unused MS_PER_SECOND |
| 27 | `fix(RawJsonDialog)` byteSize → charCount (honest unit), await clipboard |
| 28 | `fix(viewStore)` reconcile addons in effective lens against live grouping |
| 29 | `fix(server)` discoverCacheKey includes snapped from, test apps reset broadcaster |
| 30 | `fix(FacetSection)` enter-to-apply resolves value from label match |
| 31 | `fix(useTraceNewCount)` scope refresh-pulse baseline to query identity |
| 32 | `fix(SectionRenderer)` thread onHide into AttributesSection |
| 33 | `fix(VizPlaceholder)` show-details icon points toward the stacked pane |

### What each finding caught

- **a11y on hover-revealed buttons.** SidebarSection's hide × and TreeRow's pin icon were `opacity: 0` until hover with no keyboard treatment — sighted keyboard users could tab onto invisible controls. Fix: `_groupFocusWithin` + `_focusVisible` + `tabIndex={-1}` / `aria-hidden` / `pointerEvents: none` while hidden.
- **Metadata key injection.** DrawerHeader's `formatMetadataFilterQuery` inlined raw customer-defined keys into Liqe queries; a key with query syntax (`:`, `(`, etc.) broke the grammar. Restricted to `[A-Za-z0-9_.-]+` whitelist; out-of-whitelist keys lose the click-to-filter affordance instead of generating broken queries.
- **Comparison predicates misrouted.** QueryBreakdownChips classified `duration:>1000` as a `field:value` chip and routed the X through `removeFacet` (value-specific, no-ops on comparisons). Now checks `op !== ":"` first and uses `removeField` for comparisons.
- **`removeImplicitTermFromQuery` over-removed.** `foo AND foo AND bar` collapsed to `bar` after one click instead of `foo AND bar`. Added a `removed` flag so only the first match drops.
- **Active categorical sections vanished on empty discover.** Filter on `status:error`, navigate to a window with no errors → entire Status section disappeared from the sidebar; user couldn't clear the filter from there. Now keep the section mounted when the AST references the field.
- **Column reorder in pinned-column space.** ColumnsDropdown arrows resolved neighbours in the full `columnOrder` (which includes pinned columns) so a pinned column between two reorderable ones made the move a no-op. Now resolve in `orderedVisibleColumns` space and map back.
- **"In the future" string.** `formatVerboseRelative` printed `"in the future"` for clock-skewed traces; matches the narrow formatter's clamp-to-`"just now"` for consistency.
- **`byteSize` unit dishonesty.** `RawJsonDialog`'s SizeBadge labelled `string.length` (UTF-16 code units) as "bytes", which collides with other surfaces in the repo that compute `length * 2`. Renamed to `charCount` + "chars" label.
- **Optimistic clipboard label.** `CopyButton` set "Copied" before awaiting `clipboard.writeText`; Safari private mode + insecure contexts saw the confirmation even when nothing reached the clipboard. Now awaits + stays silent on rejection.
- **`getEffectiveLens` leaked addons across groupings.** Switching `all-traces` (flat, with `io-preview` + `expanded-peek`) to `by-service` (grouped) kept the flat addons attached. Now reconciles addons against the live grouping's capability.
- **`discoverCacheKey` collision.** Different time spans landing on the same preset label (15min and 1h both label "1h") shared a single cache slot, serving facets for the wrong window. Includes snapped `from` alongside `to`.
- **Test broadcaster leak.** `setDiscoverBroadcaster` is module-global; `createTestApp` overwrote the App but not the broadcaster closure, so test paths could still invoke the production `BroadcastService`. `createTestApp` now clears it.
- **Redundant ternary.** `PeerCursorOverlay anchor={trace ? ... : null}` ran under a render branch where `trace` is already guaranteed; dropped.
- **Unused constant.** `MS_PER_SECOND` was added with the other MS_PER_* constants and never used; dropped.
- **Facet Enter-to-apply submitted the typed label.** `FacetSection`'s search matched only `item.label`; on Enter it submitted `searchQuery.trim()` as the value. For facets where `label !== value` (friendly topic names, IDs with labels), users saw a row, hit Enter, and applied an invalid filter. Search now matches both `label` and `value`; on Enter the trimmed query resolves to an exact (case-insensitive) match in `items` and submits `matched.value`, falling back to the raw typed value only when nothing matches.
- **`useTraceNewCount` pulse baseline was context-blind.** `prevCountRef` was module-instance-global, so a count from one query identity (project A, "last hour", `errors:true`) was compared against the next count after the identity changed — firing or suppressing the 0→N aurora pulse against a stale baseline. Ref is now `number | null`, resets on `project?.id` / time range / `since` / `queryText` changes; first success in a new context records the baseline without pulsing.
- **`onHide` not threaded into AttributesSection.** `SectionRenderer`'s attributes branch never passed `onHide` through, so trace/span/event attribute sections missed the header-level hide affordance that `cat` and `range` branches already had. One-liner fix.
- **Show-details icon pointed the wrong way.** The detail-pane reveal toggle used `LuPanelRightOpen` for horizontal layouts and `LuPanelTopOpen` for the rest. In the non-horizontal layout the pane is stacked *below* the viz; chevron now uses `LuPanelBottomOpen` so it points toward where the pane will expand.

**Skipped (1):**
- Copilot Skill-Rule "KSUID only" flagged `crypto.randomUUID` in `viewStore.ts:368`. `git blame` shows this `generateId()` predates the PR (added in earlier `feat(traces-v2)` work). Out of scope.


langwatch/src/features/traces-v2/components/FilterSidebar/FacetSection.tsx
```diff
@@ -192,15 +198,53 @@ export const FacetSection: React.FC<FacetSectionProps> = ({
...
+          // `overflowX: "hidden"`. The 2px gutter on each side keeps
+          // the focused state legible without pulling the input far
+          // away from the rest of the section's content.
+          <VStack gap={0.5} align="stretch" marginTop={1} paddingX={0.5}>
+            <Input
+              size="xs"
+              placeholder="Search or press Enter to apply…"
+              value={searchQuery}
+              onChange={(e) => setSearchQuery(e.target.value)}
+              onKeyDown={(e) => {
```
sergioestebance, line 222: P2: Enter-to-apply is still using the raw `searchQuery.trim()` as the facet value, while the visible search only matches `item.label` (`filterAndSortItems` also only checks label). For any facet where `label !== value` (for example friendly topic names), a keyboard user can see a matching row and then apply an invalid/different filter value. Search both `label` and `value`, and on Enter resolve an exact matched `FacetItem` and pass `matched.value` before falling back to the raw typed value.

langwatch/src/features/traces-v2/hooks/useTraceNewCount.ts
```diff
@@ -87,6 +88,14 @@ export function useTraceNewCount(): TraceNewCountResult {
 
   const liveUpdatesMode = useSseStatusStore((s) => s.liveUpdatesMode);
 
+  // Aurora refresh pulse is now scoped to "trace about to appear" — fires
+  // once when the count transitions from 0 to >0 in live mode, signalling
+  // to the user that the list is being merged with new rows. In ask mode
+  // the user opted to gate merges behind the floating pill click, so we
+  // stay quiet there; the pill itself is the signal.
+  const pulseRefresh = useRefreshUIStore((s) => s.pulse);
```
sergioestebance, line 96: P2: `prevCountRef` is still scoped globally across query contexts. If the user changes project, time range, query text, or `since`, the next `newCount` result is compared against the previous context and can incorrectly fire or suppress the 0->N refresh pulse. Reset this ref to `null` when the query identity changes (`project?.id`, time range fields/label, `since`, `queryText`), and treat the first success in a new context as baseline only.

sergioestebance (commented):
Manual review pass: CI is green, but I found remaining issues at the current PR head. Posting the diff-anchorable findings inline; a follow-up conversation comment covers findings GitHub would not anchor because they are outside the changed ranges.

sergioestebance:
Two additional review findings could not be posted inline because GitHub would not anchor them to the current diff ranges:

- P3 `langwatch/src/features/traces-v2/components/FilterSidebar/SectionRenderer.tsx:119`: `AttributesSection` now accepts `onHide`, but the attributes branch never passes it through. Trace/span/event attribute sections therefore miss the new header-level hide affordance. Fix: pass `onHide={onHide}` into `AttributesSection`, matching categorical/range branches.

- P3 `langwatch/src/features/traces-v2/components/TraceDrawer/VizPlaceholder.tsx:563`: for non-horizontal layouts, the detail pane is stacked below the viz, but the icon still uses `LuPanelTopOpen`, which points at the opposite edge. Fix: use `LuPanelBottomOpen` for the non-horizontal case and keep `LuPanelRightOpen` for horizontal.

langwatch/src/features/traces-v2/components/FilterSidebar/FacetSection.tsx
0xdeafcafe: Fixed in `93ddb0cd8`. Search now matches both `label` and `value`; on Enter the trimmed query resolves to an exact (case-insensitive) match in `items` and submits `matched.value`, falling back to the raw typed value when nothing matches.

langwatch/src/features/traces-v2/hooks/useTraceNewCount.ts
0xdeafcafe: Fixed in `486ec3618`. `prevCountRef` is now `number | null` and resets to `null` whenever `project?.id`, `timeRange.from/to/label`, `since`, or `queryText` changes. The first success in a new context records the baseline without pulsing.

0xdeafcafe:
Both P3 findings from the conversation comment are addressed:

- `SectionRenderer.tsx` → `ef24b3d9c`: `onHide={onHide}` threaded into the `AttributesSection` branch.
- `VizPlaceholder.tsx` → `63091919f`: non-horizontal layout now uses `LuPanelBottomOpen` so the chevron points toward the stacked detail pane; horizontal keeps `LuPanelRightOpen`.

langwatch/src/server/app-layer/traces/trace-list.service.ts
```diff
@@ -425,50 +523,108 @@ export class TraceListService {
...
+    // compute that will hydrate the cache and SSE-broadcast when done.
+    // Caller (`tracesV2.discover` → React Query → `useTraceFacets`)
+    // renders the synthetic FACET_DEFAULTS skeleton while waiting, so
+    // the user never blocks on the 1-2s ClickHouse scan. Trade-off:
+    // first viewer sees an empty sidebar for ~1-2s instead of a
+    // spinner; subsequent viewers within the TTL hit the warm cache.
+    // The discover_updated SSE push fills in the values without the
+    // user having to refresh.
+    this.refreshDiscoverInBackground(snappedParams, cacheKey);
+    return [];
```
sergioestebance: 🔴 **Cold-miss returns `[]` as a *success*** — so `useTraceFacets` reports `isLoading:false, data:[]`, and the synthetic `FACET_DEFAULTS` skeleton never renders (it's gated on `facetsLoading` in `useFilterSidebarData.ts:170`). On `main` this path `await`ed `computeDiscover` and returned real facets, so first-viewer cold start now shows an **empty sidebar for ~1–2s** (until the SSE push lands) instead of data — and if the push is missed, it persists until the next refetch. The comment above promising a skeleton is inaccurate. Make the pending state explicit (status/sentinel) or treat empty-success as loading so the skeleton actually shows.

langwatch/src/server/app-layer/traces/trace-list.service.ts
```diff
@@ -425,50 +523,108 @@ export class TraceListService {
...
+    // lookup so two users on the same tenant + window share a slot even
+    // when their (from, to) timestamps differ by sub-minute drift. The
+    // computeDiscover call below also uses the snapped range so cache
+    // content always matches its key.
+    const snapped = snapToWindowPreset(params.timeRange);
+    const snappedParams: DiscoverParams = {
+      tenantId: params.tenantId,
+      timeRange: { from: snapped.from, to: snapped.to },
+    };
     const cacheKey = discoverCacheKey(params);
```
sergioestebance: nit: `snapToWindowPreset` runs here (`:531`) and again inside `discoverCacheKey` (`:295`). Pass the already-snapped window into the key builder instead of re-snapping.

langwatch/src/features/traces-v2/components/TraceTable/registry/cells/trace/SpanTypeBadge.tsx
```diff
@@ -1,24 +1,39 @@
 import { Text, type TextProps } from "@chakra-ui/react";
 import type React from "react";
-import { SPAN_TYPE_BADGE_STYLES } from "../../../../../utils/formatters";
 
 const PROMINENT_SPAN_TYPES = new Set(["llm", "agent", "workflow"]);
 
+const PALETTE_BY_TYPE: Record<string, string> = {
```
sergioestebance, line 6: `PALETTE_BY_TYPE` is a second span-type color source-of-truth. This swap orphaned the old `SPAN_TYPE_BADGE_STYLES` export in `formatters.ts:210` — it's now dead (only remaining reference is a comment in `topologyMermaid.ts:96`). Consolidate onto one map, or delete the dead export.

langwatch/src/features/traces-v2/stores/facetVisibilityStore.ts
```diff
@@ -0,0 +1,165 @@
...
+ * Convenience selector — returns the prefs for a project, hydrating
+ * from localStorage on first read so consumers don't need to call
+ * `hydrateFromStorage` themselves.
+ */
+export function selectVisibilityFor(
+  state: FacetVisibilityState,
+  projectId: string | null | undefined,
+): { explicitlyShown: string[]; explicitlyHidden: string[] } {
+  if (!projectId) return emptyPrefs();
+  return state.byProject[projectId] ?? readFromStorage(projectId);
```
sergioestebance: minor: this selector does localStorage I/O and returns a fresh object when unhydrated, defeating Zustand's ref-equality bailout. Impact is bounded — the hydrate effect populates `byProject` right after mount — so low priority, but a selector should be a pure projection. Drop the `readFromStorage` fallback and rely on the existing hydrate effect.

sergioestebance (commented):
Ran a 4-reviewer pass and validated each finding against the diff. Security clean (metadata-key whitelist holds), `flameView`/`spanListView` deletions complete, SSE + `reconcile*` reuse is good. One thing worth deciding before merge (cold-miss empty sidebar); the rest are optional cleanups. Note: ~4k LOC net-new with no tests added. Details inline.

langwatch/src/features/traces-v2/hooks/useTraceFacets.ts
```diff
@@ -49,8 +80,15 @@ export function useTraceFacets() {
 
   const isFromOtherProject = dataProjectIdRef.current !== projectId;
 
+  // Cold-miss responses come back with `pending: true` and an empty
+  // facets array — treat that as still-loading so the FilterSidebar
+  // keeps rendering the FACET_DEFAULTS skeleton (gated on `isLoading`)
+  // until the SSE-driven invalidation lands the real payload. Without
+  // this, the sidebar would flash empty for the 1–2s ClickHouse scan.
+  const result = isFromOtherProject ? EMPTY_RESULT : (query.data ?? EMPTY_RESULT);
```
sergioestebance, line 117: P2: This now treats cold `discover` misses as loading until an SSE-driven invalidation arrives, but there is no fallback if the `discover_updated` event is missed, delayed, rate-limited, or the subscription is not connected yet. In that case `result.pending` remains true and the sidebar can stay on the synthetic loading/default facets indefinitely until a manual refresh/focus refetch happens. Add a short refetch fallback while `result.pending` is true (with backoff), or make cold misses compute synchronously so the first successful response can always settle without depending on SSE delivery.

langwatch/src/features/traces-v2/hooks/useDrawerUrlSync.ts
```diff
@@ -1,20 +1,19 @@
...
-  isDrawerTab,
   isViewMode,
   isVizTab,
+  parsePinnedSpansParam,
+  serializePinnedSpansParam,
   useDrawerStore,
   type VizTab,
 } from "../stores/drawerStore";
 
 const DEFAULTS = {
```
sergioestebance, line 14: P3: This URL-sync default still falls back to `trace`, while `drawerStore` now defaults new drawer opens to `summary`. Initial mount often self-corrects via store-to-URL sync, but browser back/popstate into an older drawer URL without `drawer.mode` will hydrate back to Trace mode. Align this default with the store's `summary` default.

sergioestebance (changes requested):
Review pass: not ready to merge yet. Main blocker is the cold `discover` pending path relying solely on SSE to recover; I also left one minor URL-default mismatch comment.

langwatch/src/features/traces-v2/hooks/useTraceFacets.ts
0xdeafcafe: Good catch — added a polling fallback in 8f3d30050. While `result.pending` is true we refetch with 2/4/8/15s backoff, reset to 0 on any non-pending response and on SSE delivery, so a missed/delayed/disconnected `discover_updated` push never strands the sidebar.

langwatch/src/features/traces-v2/hooks/useDrawerUrlSync.ts
0xdeafcafe: Fixed in 8f3d30050 — `DEFAULTS.mode` is now `summary` to match drawerStore. As a side benefit the store→URL diff stops pushing `drawer.mode=summary` on first mount when the store is on the default.

sergioestebance (approved):
lgtm

PR #4521, opened 2026-06-03
perf(clickhouse): prune event_log partitions on time-local aggregate rehydration

langwatch-agent:
## Evidence

You investigated the late-May/June S3 cost spike on `langwatch-storage-clickhouse-prod` (the ClickHouse S3-disk store). The headline finding: the **Tier2 GET burst stays ~3x elevated even after the write/part-churn normalized** (6/02: ~69M GET/day, ~$30/day), i.e. heavy ongoing query traffic against S3-tiered partitions. Projected ~$1.6-2k/month if it continues.

Breaking down the cold-scan signal (queries on time-partitioned tables with no partition-time predicate) by table + param signature over 24h:

| count (24h) | coldScanTable | paramKeys |
| --- | --- | --- |
| **19,681** | `event_log` | `tenantId, aggregateType, aggregateId` |
| 1,219 | stored_spans | tenantId, traceId |
| 420 | stored_spans | tenantId, traceId, limit |
| ... | ... | (all remaining shapes < 300 each) |

One shape is ~94% of all cold scans: the event-sourcing aggregate rehydration. Each call walks every weekly partition of `event_log` on S3, so it is the dominant driver of the persistent GET burst.

## Suspected cause

`EventRepositoryClickHouse.getEventRecords` (used by the fold-projection re-fold path) filters only `TenantId + AggregateType + AggregateId` with no predicate on the partition column `EventOccurredAt`:

```
PARTITION BY toYearWeek(toDateTime64(EventOccurredAt / 1000, 3))
ORDER BY (TenantId, AggregateType, AggregateId, IdempotencyKey)
```

The sort key narrows to the aggregate's rows *within* each partition, but with no `EventOccurredAt` predicate ClickHouse cannot prune partitions, so it opens every weekly partition (many on S3) and most contribute zero rows. The re-fold (`foldProjectionExecutor`) already knows the triggering event's `occurredAt`, but it was not threaded into the read.

## Proposed fix

Rehydration of a **time-local** aggregate (`trace`, `evaluation`, `experiment_run`, `simulation_run`, `suite_run`) only needs events within the aggregate's own short lifetime. Thread the triggering event's `occurredAt` from the re-fold executor → `eventLoader` → `getEvents` → `getEventRecords`, and lower-bound the scan:

```sql
AND (EventOccurredAt = 0 OR EventOccurredAt >= {occurredAtFromMs:UInt64})
```

- The `EventOccurredAt = 0` arm keeps unknown-occurred-time rows so the bound can never drop an event.
- The classification + window live in `stores/rehydrationWindow.ts` (the correctness contract). Long-lived aggregate types (`global`, `billing_report`, `simulation_set`) and any read without a usable anchor stay **unbounded** — behaviour there is unchanged.
- Window is 45 days: it must exceed any time-local aggregate's lifetime (a trace/eval/run spans seconds to hours) and stay under `event_log` retention. It errs large; the failure mode of too-small is silently dropping events from a re-fold, so it is deliberately conservative.

No schema, migration, or infra change.

### Correctness contract (please confirm)

The optimisation is only safe because every event of a time-local aggregate occurs within `REHYDRATION_WINDOW_MS` of any other event of the same aggregate. The five listed types satisfy this; the long-lived types are excluded. A domain owner should sanity-check the list and the window against retention before merge.

## Expected impact

- For a time-local aggregate rehydration, the partitions opened drop sharply (see EXPLAIN), cutting the S3 GET volume that drives the cost. Since this shape is ~94% of cold scans, this targets the bulk of the persistent Tier2 GET burst.
- Latency may also improve (fewer S3 round-trips), but the headline metric here is S3 request cost, not latency.

## Test plan

- `rehydrationWindow.unit.test.ts`: the policy returns a bound for every time-local type, `undefined` for long-lived types and for a missing/zero anchor, and never negative.
- `eventRepositoryMemory.lowerBound.unit.test.ts`: the lower bound keeps events at/after the bound and unknown-time (`EventOccurredAt = 0`) events, and drops older ones.
- `eventRepositoryClickHouse.test.ts`: the CH query gains the `EventOccurredAt` predicate + bound param when a bound is supplied, and omits both when unbounded (incl. a zero bound).
- Full event-sourcing store/projection unit suites pass unchanged (162 tests), confirming the optional anchor threading is backward-compatible.

## EXPLAIN check

`EXPLAIN PLAN indexes = 1` on a real large tenant (id redacted), `AggregateType = 'trace'`:

```
old (no predicate):   Partition  Parts: 286/286   Granules: 89865/89865
new (with bound):     Partition  Parts:  48/288   Granules: 22117/89867
```

The new partition condition is `toYearWeek(...) >= 202616 (≈ 45 days ago) OR toYearWeek(...) = 196952 (the EventOccurredAt = 0 / 1970 partition)`. Partitions read drop from 286 to 48 (~83% pruned) while the unknown-time partition is correctly retained. The sort-key granule selection (TenantId/AggregateType/AggregateId) is unchanged.

### DIFF
--- a/langwatch/src/server/event-sourcing/projections/abstractFoldProjection.ts
+++ b/langwatch/src/server/event-sourcing/projections/abstractFoldProjection.ts
@@ -1,16 +1,16 @@
-import { z } from "zod";
+import type { z } from "zod";
 import type { Event } from "../domain/types";
+import {
+  type DotSnakeToPascal,
+  type EventTypeOf,
+  eventTypeToHandlerName,
+  type StripPrefix,
+  type UnionToIntersection,
+} from "./eventTypeTransforms";
 import type {
   FoldProjectionOptions,
   FoldProjectionStore,
 } from "./foldProjection.types";
-import {
-  type EventTypeOf,
-  type StripPrefix,
-  type DotSnakeToPascal,
-  type UnionToIntersection,
-  eventTypeToHandlerName,
-} from "./eventTypeTransforms";
 
 // ---------------------------------------------------------------------------
 // Schema → event type extraction
@@ -26,7 +26,13 @@ export type AnyEventSchema = z.ZodObject<
 // ---------------------------------------------------------------------------
 
 /** All possible timestamp keys. Forbidden in `initState()` return type. */
-type AllTimestampKeys = "CreatedAt" | "UpdatedAt" | "createdAt" | "updatedAt" | "LastEventOccurredAt" | "LastEventOccurredAt";
+type AllTimestampKeys =
+  | "CreatedAt"
+  | "UpdatedAt"
+  | "createdAt"
+  | "updatedAt"
+  | "LastEventOccurredAt"
+  | "LastEventOccurredAt";
 
 /** Full derivation: `"lw.suite_run.started"` → `"handleSuiteRunStarted"` */
 type HandlerName<EventTypeStr extends string> =
@@ -111,10 +117,15 @@ export abstract class AbstractFoldProjection<
     createdAtKey,
     updatedAtKey,
     LastEventOccurredAtKey,
-  }: { createdAtKey?: CK; updatedAtKey?: UK; LastEventOccurredAtKey?: LEOAK } = {}) {
+  }: {
+    createdAtKey?: CK;
+    updatedAtKey?: UK;
+    LastEventOccurredAtKey?: LEOAK;
+  } = {}) {
     this.createdAtKey = createdAtKey ?? ("CreatedAt" as CK);
     this.updatedAtKey = updatedAtKey ?? ("UpdatedAt" as UK);
-    this.LastEventOccurredAtKey = LastEventOccurredAtKey ?? ("LastEventOccurredAt" as LEOAK);
+    this.LastEventOccurredAtKey =
+      LastEventOccurredAtKey ?? ("LastEventOccurredAt" as LEOAK);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


0xdeafcafe:
this pr needs a very careful review. if you don't have 100% understanding of the implication of this change please defer pr to someone else 🙏🏻. i'm going to try and tackle it today

langwatch-agent:
@0xdeafcafe rebased onto current main and resolved the conflict, ready for re-review.

Context: this targets the **aggregate rehydration** read path (`getEventRecords` reads all events for an aggregate by AggregateId to fold a projection), and is **complementary** to the just-merged #4944, which only touched the single-blob point lookup (`getFromEventLog`, keyed by a known EventId). #4944 derived its window by parsing the EventId's own KSUID timestamp; that trick can't apply here because rehydration is fetching the events, so it doesn't have their ids yet (and AggregateId is a random OTel trace id, not time-sortable). So this PR keeps its conservative 45-day lower bound anchored on the triggering event, which is the right lever for this path.

The only conflict was in `abstractEventStore.ts`: main now carries the empty-aggregateId guard from #4943. The two are sequential and both kept, the guard short-circuits an empty id first, then the rehydration lower bound is computed for the real read. Full event-sourcing unit suite green locally (358 tests).

langwatch-agent:
Fixed the `lint` failure (533754b51). It was biome **formatting drift**: this branch was formatted on its original date, and biome has since been bumped in main, so several of the touched lines no longer matched the current formatter (e.g. the `Set` initializer and a few multi-line wraps). Ran the current biome (2.4.14, the lockfile-pinned version CI uses) over the changed files, plus organize-imports / useImportType. Also rewrote one pre-existing `for-in` in `normalizePayloadValue` to `for...of Object.keys` to clear the last (nursery `noForIn`) finding. No logic change. typecheck clean, 358 event-sourcing unit tests green.

rogeriochaves (approved):

PR #4524, opened 2026-06-03
refactor(retention): run orphan-sweep as a self-dispatching groupQueue command

sergioestebance:
## Why

Follow-up to the 2026-06-02 retention incident (post-mortem **RC #2**). The orphan-sweep ran on a **BullMQ `QueueWithFallback` chain** (ADR-023) — the *only* per-trace-path workload still on BullMQ while everything else in trace processing runs on the event-sourcing **groupQueue**. That divergence is what bit us:

- BullMQ rejects a custom `jobId` containing `:` (the seed was `orphan-sweep-chain:<tenant>`); the **dev in-memory queue allowed it** and the **groupQueue sanitizes `:`** → invisible until prod.
- `QueueWithFallback` treated that *deterministic* rejection as a transient Redis outage and ran the heavy `sweepProject` **inline on the ingestion path** → a per-event ClickHouse read-storm that stalled ingestion.

This PR removes the root: the sweep now runs on the **groupQueue**, where `:` can't break dispatch and there is no inline-fallback footgun.

## What

Modeled on the in-prod `billing_reporting` command (`selfDispatch` + dedup + circuit-breaker):

- New **command-only** pipeline `orphan-sweep-processing`. The ingestion reactor (kept) **dispatches** a per-tenant `sweepOrphansForTenant` command instead of seeding a BullMQ chain.
- The command runs **one bounded increment** (`sweepProject`, unchanged — 100-page cap, cursor-resumed) then **self-dispatches** the next one with a single **6h** delay. No scheduler, no cron, no BullMQ repeat — an event continuation, satisfying ADR-023's "no schedulers" constraint.
- **Tenant-scoped dedup** ⇒ one sweep loop per tenant across all replicas; bursty ingest folds in.
- **Circuit-breaker rides in the payload** (`consecutiveFailures`) — stops self-dispatch after 5 consecutive failed increments and surfaces the error; next ingest re-seeds. No new persistence.
- **Deleted:** `orphanSweepChainQueue.ts`, `orphanSweepChainWorker.ts`, `startOrphanSweepChainWorker()` + its registration, and the now-orphaned `ORPHAN_SWEEP_CHAIN_QUEUE` / `OrphanSweepChainJob`.

### Why 100 pages/run is safe on the groupQueue
The groupQueue heartbeats a job's active key every ~100s for its whole duration, so a long increment is **not** falsely re-dispatched — the 5-min `activeTtlSec` is a crash-recovery net, not a wall-clock cap. On a genuine crash the increment resumes from the persisted cursor (idempotent).

## Test plan
- `sweepOrphansForTenant.command.unit.test.ts` — project-guard stop (archived/deleted), bounded sweep, self-dispatch reset-on-success / increment-on-failure, circuit-breaker stop. **6/6 pass.**
- Reactor test updated to assert command dispatch (`dispatchSweep`).
- `pnpm typecheck` **clean (0 errors)**; 86/86 pass across `data-retention`, `orphan-sweep-processing`, and `billing-reporting` (registry sanity).

## Docs
- **ADR-024** supersedes the queue half of **ADR-023** (sweep semantics unchanged).
- `specs/data-retention/orphan-sweep.feature` rewritten to the command model.
- Design spec: `EPIC/Q2/data-retention/queue/orphan-sweep-on-groupqueue.spec.md`.

## Relationship to #4518
Independent. #4518 is the standalone hot-fix (`:`→`-`, `fallbackToInline:false`) that **unpins the workers** now; this PR is the structural fix that **deletes the chain #4518 patches**. Merge #4518 first; if this lands first instead, close #4518 as moot.


sergioestebance:
Superseded by #4526 — after review we decided to **remove the orphan sweep entirely** rather than migrate it (no background deletion of orphaned PG rows; the feature was the trigger of the 2026-06-02 incident and not worth the operational risk). See ADR-025. Closing unmerged.

PR #4526, opened 2026-06-03
chore(retention): remove the PG orphan sweep entirely

sergioestebance:
## What

Remove the **PostgreSQL orphan-sweep** we introduced in langwatch/langwatch#4147 — the background BullMQ chain that deleted PG rows referencing TTL-expired traces. This unblocks the worker roll-forward (the sweep was the trigger of the 2026-06-02 incident, RC langwatch/langwatch#2). Deletes the service/repository/reactor/cursor-store, the chain queue + worker, and all wiring (trace-processing reactor attach, `PipelineRegistry`/`DataRetentionDependencies` plumbing, worker registration, the `orphan_sweep_chain` job metric + `data_retention_orphans_swept_total` counter, and the feature spec).

**Scope:** this removes only the **PG traceId-reference sweep**. It does **not** touch storage-level orphaned-trace cleanup (`cleanupOrphanedTraces` / `cleanupOrphanedHotTraces` from `cron.ts`) — that stays. The rest of langwatch/langwatch#4147 (CH-native TTL, reconciler, retention policy, retroactive, metering, pinning, cold-path) is untouched.

## Accepted consequence

Retention TTL still expires traces, so PG rows that reference them (`Annotation`/`AnnotationQueueItem`, `PublicShare`, `TriggerSent`, `PinnedTrace`) now linger. These are **non-crashing phantoms** (the annotation render chain is null-trace-safe end-to-end). Read-time handling is intentionally **out of scope** and tracked as follow-ups:

- langwatch/langwatch#4529 — annotation phantom tasks + inflated queue counts (read-time filter)
- langwatch/langwatch#4530 — trigger dedupe suppression on same-project trace-id reuse
- langwatch/langwatch#4531 — public-share authorization not scoped to `projectId` (**pre-existing** cross-tenant concern, independent of this removal)

## Verification
`pnpm typecheck` clean (0); 495/495 pass across data-retention + trace-processing; whole-repo grep → zero orphan-sweep code symbols outside the historical ADRs.

## Docs
ADR-025 records the removal + follow-ups; ADR-023 marked superseded; ADR-022/024 cross-refs updated.

## Supersedes
- langwatch/langwatch#4524 (groupQueue migration) — closed unmerged; this removes what it migrated.
- langwatch/langwatch#4518 (hot-fix) — the storm path no longer exists; close unless its `QueueWithFallback` no-inline hardening is wanted standalone.


dev/docs/adr/025-remove-orphan-sweep.md
```diff
@@ -0,0 +1,79 @@
...
+
+PR #4524 (the groupQueue migration) is **closed unmerged** — superseded by this
+removal. PR #4518 (the hot-fix) is moot for the storm since the sweep no longer
+exists; close it unless its `QueueWithFallback` hardening is wanted on its own.
+
+## Consequences
+
+**Accepted negative.** PG rows that reference TTL-expired traces now accrue
+indefinitely:
+- `Annotation` / `AnnotationQueueItem` for an expired trace remain (may surface
```
rogeriochaves, line 54: This accepted stale `AnnotationQueueItem` behavior creates a concrete UX regression today, not just inert storage growth. The annotation APIs still count and return every pending queue item without verifying the trace exists (`getPendingItemsCount`, `getAssignedItemsCount`, and `getQueueItemsCounts` all count `doneAt: null` rows directly; `getOptimizedAnnotationQueues` returns stale rows with `trace: null`). The my-queue page then filters only by `!doneAt` and renders the task, passing an empty trace id into `Conversation`, so users keep seeing pending annotation work that cannot load a trace. Either keep cleanup for annotation queue items or add the read-time filtering described below before removing the sweep; otherwise expired traces leave permanent phantom annotation tasks and badge counts.

rogeriochaves (dismissed):
Found one blocking issue. The removal is mechanically clean and CI is green, but the annotation queue stale-row behavior needs a read-time guard before this ships.

dev/docs/adr/025-remove-orphan-sweep.md
```diff
@@ -0,0 +1,79 @@
...
+  as entries pointing at a trace that no longer loads).
+- `PublicShare` of an expired trace remains (the link resolves to a missing
+  trace).
+- `TriggerSent.traceId` keeps pointing at an expired trace.
+- `PinnedTrace` of an expired trace remains — the pin resolves to nothing (the
+  deleted sweep used to remove these; see `specs/data-retention/trace-pinning.feature`).
+
+None of these corrupts data or crosses tenants; they are stale references for
```
rogeriochaves: This “does not cross tenants” claim is false with the current public-share auth path. `checkPermissionOrPubliclyShared` authorizes by `resourceType` + `resourceId` only, without constraining the `PublicShare` row to `input.projectId`; then `traces.getById` reads the trace from the caller-supplied project id. If a stale `PublicShare` for expired trace id `t` remains in project A, an unauthenticated request for project B / trace id `t` will pass the public-share middleware and return B’s trace if one exists. Even within the same project, a later reuse of trace id `t` inherits the old public URL. Before removing cleanup, either scope public-share authorization to the project or add read-time/lifecycle cleanup for these stale shares; otherwise this change turns expired public shares into indefinite authorization artifacts.

rogeriochaves (commented):
Adding one more issue found while continuing the review. This is separate from the annotation-queue phantom task finding already posted.

dev/docs/adr/025-remove-orphan-sweep.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Consequences
+
+**Accepted negative.** PG rows that reference TTL-expired traces now accrue
+indefinitely:
+- `Annotation` / `AnnotationQueueItem` for an expired trace remain (may surface
+  as entries pointing at a trace that no longer loads).
+- `PublicShare` of an expired trace remains (the link resolves to a missing
+  trace).
+- `TriggerSent.traceId` keeps pointing at an expired trace.
```
rogeriochaves, line 58: Leaving `TriggerSent.traceId` rows forever can suppress future trigger actions when a trace id is reused. The dispatch path calls `claimSend`, which inserts `(triggerId, traceId, projectId)` with `skipDuplicates: true` and relies on the Prisma `@@unique([triggerId, traceId])` constraint; if the old row survives after the original trace TTLs out, `claimSend` returns false for a later trace with the same id and the action is skipped. Since trace ids come from ingest payloads rather than a persistent database sequence, treating this as a harmless stale reference is too strong. Either cleanup/filter these rows on trace expiry or include a generation/trace timestamp in the dedupe key so a new trace with a reused id can fire normally.

rogeriochaves (commented):
One more stale-reference regression from the final pass: trigger dispatch dedupe is also affected by indefinite stale trace references.

dev/docs/adr/025-remove-orphan-sweep.md
sergioestebance: Confirmed and accurate — with one verified refinement: it's a clearable annoyance, not a stuck/permanent state. The render chain is null-trace-safe (`getById`/`useTraceDetailsState` are `enabled: !!traceId`, trace block gated on `item?.trace`), and `markQueueItemDone` keys on `queueItemId` only (Done isn't disabled on a missing trace), so a phantom item opens as a blank pane and is dismissible via Done; the badge self-heals as items are cleared. No crash, no stuck queue.

Decision: accept the phantom for now to keep this a clean removal that unblocks the worker roll-forward. Read-time filtering is the proper fix — tracked in #4529 (with "keep annotation-queue cleanup until then" noted as the hybrid alternative).

dev/docs/adr/025-remove-orphan-sweep.md
sergioestebance: Good catch — the auth gap is real, and we've rated it P1. But it's pre-existing and independent of this removal: `checkPermissionOrPubliclyShared` (`rbac.ts:1490`) matches `(resourceType, resourceId)` without scoping to `projectId`, so the cross-tenant exposure already applies to live shares; the sweep never closed it. So it shouldn't gate this PR.

Tracked as a separate security fix in #4531 (scope public-share auth to `projectId`, or keep share cleanup until that lands). Heads-up: the repro is already public in this thread.

dev/docs/adr/025-remove-orphan-sweep.md
sergioestebance: Agreed it's not strictly harmless. Scope: it only suppresses a trigger when a trace id is reused in the same project after the original TTL-expired (≥49d) — rare, silent, non-crashing. Accepting for now to keep this a clean removal; the dedupe-key/staleness fix (or keeping `TriggerSent` cleanup until then) is tracked in #4530.

0xdeafcafe (approved):

PR #4538, opened 2026-06-03
fix(/me + /settings): bugbash sweep (drawer copy, support contact, your-projects removal, members layout)

rogeriochaves:
## Summary

Bug-bash PR addressing rchaves's dogfood pass on the /me + /settings surfaces.

- **Bug-bash item 5 — Drawer copy.** Replace dev-speak ("Traces normalized into gen_ai.* canonical") in the Connect Claude Code drawer with the same user-facing phrasing the tile already uses ("no change to how you call the API"). End-users installing Claude Code do not care about our internal canonical schema name.
- **Bug-bash item 4 — Admin-configurable support contact.** New `Organization.supportContact` free-text field surfaced on /settings (between Project ID and Live presence). Replaces the implicit "first ADMIN member's email" lookup in CLI "contact your admin" messages and the in-app budget-exceeded banner. Accepts email, URL pointing at an internal ticketing system, or any short instruction. Two shared resolver helpers split by intent: `resolveSupportContact` (display-facing, may be URL/email/text) and `resolveOrgAdminEmail` (email-recipient-only, always a real email for actual mail sending). Banner detects URL vs email vs text and renders the link href accordingly.
- **Bug-bash item 3 — Members table layout.** /settings/members Card.Body overflow=hidden was clipping the rightmost ⋮ menu after the CostCenter column was added. Switched to overflowX=auto, capped the email column at 280px, and wrapped the email cell in OverflownTextWithTooltip.
- **Bug-bash item 2 — Drop "Your projects" section from /me.** The card duplicates what the new top-bar workspace switcher already surfaces (org > team > project). Removed `MyProjectsCard.tsx`, the `user.userProjects` tRPC procedure, and its integration test block.
- **CodeRabbit follow-up.** Client-side maxLength=500 on supportContact; dropped redundant legacy ADMIN row-gate in `organization.update` (was blocking RoleBinding-based admins); strip leading `mailto:` in banner href; delegate the fallback in `resolveSupportContact` to `resolveOrgAdminEmail`.

## Lane note

Disjoint from the in-flight prod-blackhole PR #4533 (OTLP silent-drop + VK-without-policy + ik-lw- classifier). Different surfaces, no file overlap. Bug-bash item 1 (langwatch claude unification) is on hold pending data-model alignment with rchaves on the admin-toggle storage shape.

## Test plan

- [ ] Typecheck clean on every commit (already verified)
- [ ] /settings org form: save new support contact, refresh page, value persists
- [ ] /settings org form without organization:manage shows read-only text (or "Not set")
- [ ] Budget-exceeded banner with `supportContact="ops@acme.test"` renders mailto: link
- [ ] Budget-exceeded banner with `supportContact="https://acme.test/help"` renders https link with target=_blank
- [ ] Budget-exceeded banner with `supportContact="ping @platform on Slack"` renders plain text (no link)
- [ ] Budget-exceeded banner with `supportContact="mailto:foo@acme.test"` renders single-mailto: link (no double prefix)
- [ ] CLI /api/auth/cli/budget/status response uses supportContact when set, falls back to first-admin email when null
- [ ] sendBudgetIncreaseRequestEmail (user.ts router) still uses real first-admin email regardless of supportContact override
- [ ] /me page no longer shows the "Your projects" section
- [ ] /settings/members ⋮ menu is visible at the right edge of every row, including when the table is wider than the viewport


rogeriochaves:
Addressed CR pass in f3da648ea:

- **settings.tsx (P3 inline)**: added `maxLength={500}` + `register("supportContact", { maxLength: 500 })` so client matches the 500-char server cap.
- **organization.ts (Major outside-diff)**: dropped the legacy `OrganizationUser.role=="ADMIN"` row-gate inside `update`. `checkOrganizationPermission("organization:manage")` already enforces and the row-gate was actively blocking RoleBinding-based admins.
- **BudgetExceededBanner (nitpick)**: strip leading `mailto:` before email detection + href construction so `mailto:foo@x.com` does not produce `mailto:mailto:foo@x.com`.
- **resolveSupportContact (nitpick)**: fallback now delegates to `resolveOrgAdminEmail` so the admin-lookup query lives in one place.

Typecheck green, CI rolling.

PR #4544, opened 2026-06-03
feat(cli): unify langwatch claude (gateway + OTLP ingest)

rogeriochaves:
## Summary

Bug-bash item 1: unify the two Claude Code onboarding paths behind a single `langwatch claude` command. NO admin toggle, NO double-tracing - two MUTUALLY EXCLUSIVE paths per user/tool.

EXPANDED per rchaves's directive: same treatment for codex / gemini / opencode where applicable.

## Two paths, mutually exclusive

**Path A: `langwatch <tool>` wrapper (RECOMMENDED - full I/O on the gateway server-side):**

`envForTool` injects only the gateway env pair for each tool. The gateway receives every request, captures input + output + cost + tokens server-side, attributes it to the VK's principal. NO OTel injection - adding OTel on top would double-trace the same call (once via the gateway's own capture, once via the tool's client-side exporter) AND double-count cost. Auto-mint of `ik-lw-*` persists so the user CAN later switch to Path B without re-onboarding, but it is NOT pushed into the spawned child process.

**Path B: /me Trace Ingest install drawer (for tools/subscriptions that can't route through the gateway):**

Standalone OTel env-var install - set OTEL_*_EXPORTER + endpoint + Authorization header directly in the user's shell, point at `/api/otel`. Useful when:
- User is on Claude Max (no API-key swap available)
- Tool's base-URL swap doesn't cleanly route everything (cursor GUI agent panel)
- User wants traces without changing their CLI invocation

These are MUTUALLY EXCLUSIVE per user-config. Two paths, one PR.

## Sub-items + status

| Sub-item | Status |
|---|---|
| 1.1 Five-tool envForTool, gateway env pair only (Path A) | done |
| 1.2 Offer to persist exports to detected shell rc | done |
| 1.3 Remember choice + quiet in already-configured shells | done |
| 1.4 Auto-mint personal ingestion token per template (Path B prep) | done |
| 1.5 Collapse /me Coding Assistants dual UI into single langwatch claude tile | done |
| 1.6 Docs refresh (claude-code.mdx + track-your-claude-code-usage.mdx) | done |
| Path B IngestionTemplate seeds for codex / gemini / opencode + tiles + drawer snippets | done |
| Empirical OTTL captures + canonical mapping for codex + gemini | done |
| CLI-output-to-PNG dogfood tool | done |
| Codex Path A wires `[model_providers.langwatch]` + spawn with `--profile` | done |
| Codex 0.134 sibling profile file `~/.codex/langwatch-gateway.config.toml` | done |
| Claude Path B log → trace_summary lift + persistence | done - right-seam at LogRequestCollectionService (81012a19e), native TS extractors for claude/codex/gemini (8fdabbb58, 11baac0f4), OTEL_LOG_USER_PROMPTS=1 restored (24a4f8d6c), TraceSummaryStore log-record gate (cb480b692) |
| `langwatch codex` folds Path B install into mode-detect (no separate command) | done |
| #3 Coding assistants are not ingestion templates - dropped 5 coding slugs from the template seed, unhid opencode CLI + seeded its starter tile | done |
| A2 Template-free per-tool ingestion bindings keyed on (personalProjectId, sourceType) - fixes multi-org 409 + cross-org rotation hijack; templateId nullable | done |
| A2 device-flow REST install accepts source_type (not just template_id) - the CLI device-flow route was missed by the first A2 pass and 400'd template-free installs | done |
| #5 `langwatch login` is auth-only; the shell-rc persist offer moved into the `langwatch <tool>` wrapper ingestion path, framed as installing telemetry | done |
| #4 Stage B: per-(org, tool) path policy - admin toggles allowVk / allowOtelDirect per tool from the tool-catalog CLI Paths tab; CLI caches the map at login and gates path selection on it, falling back to hardcoded defaults offline | done |

## envForTool matrix (post-revert)

| Tool | Gateway envs injected by wrapper |
|---|---|
| claude | ANTHROPIC_BASE_URL, ANTHROPIC_AUTH_TOKEN |
| codex | OPENAI_BASE_URL, OPENAI_API_KEY, `--profile langwatch-gateway` |
| gemini | GOOGLE_GEMINI_BASE_URL (=`$gw/v1beta`), GOOGLE_API_KEY, GEMINI_API_KEY |
| cursor | OPENAI_*, ANTHROPIC_* (both pairs) |
| opencode | OPENAI_BASE_URL (=`$gw/v1`), ANTHROPIC_BASE_URL (=`$gw/v1`), OPENAI_API_KEY, ANTHROPIC_AUTH_TOKEN, ANTHROPIC_API_KEY |

## /api/otel/v1/logs receiver-side lift coverage (Path B)

Architecture pivot landed in 28c6c1304 (andre, building on 11baac0f4 + 81012a19e + 8fdabbb58 sergey): the four platform-known coding tools route their OTLP traffic through `/api/otel/v1/logs` (the devs-facing unified endpoint) and the receiver lifts `langwatch.*` attributes inside `traceSummary.foldProjection.handleTraceLogRecordReceived` via native TS extractors in `trace-io-accumulation.service.ts`. The `*_OTTL_STARTER` constants + the admin-editable OTTL row for these slugs were dropped, because admins can't customize a native TS extractor, so exposing a "starter OTTL template" affordance was just clutter. Admin IngestionSources remain for CUSTOM slugs only (`otel_generic`, third-party emitters); those still route through `/internal/validate-ottl` + `/internal/transform` with admin-editable `parserConfig.ottlStatements`.

Empirically captured against real tool invocations via a local OTLP catcher; spans + logs decoded from the wire (codex protobuf, gemini + opencode JSON).

| Tool | Status | Receiver lift path |
|---|---|---|
| claude_code | ACTIVE, `extractClaudeCodeApiRequestMetrics` (81012a19e) + log->span synthesizer | claude code 2.1.x dropped trace emission entirely (binary strings of cli.js confirm no OTEL_TRACES_EXPORTER read path). `LogRequestCollectionService` synthesizes trace_id from `session.id`, pairs `user_prompt + api_request` log records by `prompt.id`, and the native extractor lifts model + input/output/cache_creation/cache_read tokens + cost + thread.id at fold time. **Output now unlocked**: with `OTEL_LOG_RAW_API_BODIES=1` (wrapper default) claude code ships `api_response_body` log records carrying the full assistant text; the receiver lifts it onto `langwatch.output` + `ComputedOutput`, `query_source`-gated to skip `prompt_suggestion` / `generate_session_title` utility calls. Synthesized records carry input + tokens + cost + output text (proven on trace `53a7d10…`). |
| codex | ACTIVE, `extractCodexSseEventMetrics` (8fdabbb58) | codex emits `codex.user_prompt / codex.sse_event / codex.conversation_starts` with `input_token_count / conversation.id / prompt / model / ...`, NOT gen_ai canonical. Native extractor lifts to `langwatch.*` (model, input/output/cache_read tokens, thread.id, principal.email, input). Cost computed receiver-side from (model, tokens). (The prompt/`input` lift here is receiver-ready, but codex 0.137 ships no OTLP logs exporter, so the `codex.user_prompt` content records never reach the wire - that's why the matrix shows input/output `n/a` for codex Path B; see the codex-B narrative below.) |
| gemini | ACTIVE, `extractGenAiLogMetrics` (11baac0f4, gen_ai.* canonical) | gemini-cli 0.32+ emits gen_ai canonical on log records natively; the gen_ai.* extractor handles model + token + cost lift. Same code path automatically covers any future emitter using OTel GenAI semconv. |
| opencode | WIRED-STRUCTURAL-ONLY | opencode 1.14.22 emits structural OTel spans (`Session.*`, `LLM.run`, `Provider.*`) via @effect/opentelemetry but ZERO gen_ai.* attributes on log records. Spans land in `/api/otel` but tokens / model / cost are NOT extractable until opencode adopts gen_ai.* semconv upstream, at which point the same `extractGenAiLogMetrics` path fires automatically. |
| cursor | N/A, gateway path only | (no Path B) |

**Post-cleanup UX verification (sergey 541d20c)**: [/me Trace Ingest home with tiles](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ottl-cleanup-2026-06-04/01-me-home-with-tiles.png) (Claude cowork, Codex, Cursor, Gemini, opencode tiles + Raw OTLP advanced; no "Edit OTTL" / "starter template" copy anywhere) · install drawers [codex](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ottl-cleanup-2026-06-04/drawer-codex.png) · [gemini](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ottl-cleanup-2026-06-04/drawer-gemini.png) · [opencode](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ottl-cleanup-2026-06-04/drawer-opencode.png) (each says "Connect <Tool>, auto-shaped" + "Cost, tokens, and model are picked up automatically") · [admin tool catalog](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ottl-cleanup-2026-06-04/06-admin-ingestion-templates-list.png) (6 platform-published rows show only View + Clone-to-customise; custom org-authored rows keep Edit OTTL for the `otel_generic` shape).

## Unified ingest-key credential + once-and-final re-dogfood (2026-06-05)

The Path B ingestion credential is now the single `ApiKey` primitive: an "ingestion key" is an `sk-lw-` ApiKey scoped to one project by a CUSTOM role binding granting only `traces:create` (genuinely write-only). The `UserIngestionBinding` primitive (`ik-lw-`, plain SHA256, personal-only) is fully retired. One credential, one hash scheme, one rotation/revocation path, uniform across personal and team projects, no future migration.

**Path B re-verified live on the sk-lw foundation** (fresh local stack, latest branch + migrations): interactive tmux per tool, each minting its own personal-project `sk-lw-` key via the device-flow CLI and emitting real OTLP through it. Cost / tokens / input / output read straight from the fold; the v2 drawer for each shows `Origin ingest_key` + `langwatch.api_key.id` in trace attributes:

| Tool | Models | Cost | In / Out tok | Input | Output | v2 drawer |
|------|--------|-----:|---|:---:|:---:|---|
| claude | haiku-4-5 + opus-4-8 | $0.530 | 11170 / 17 | yes | yes | [shot](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/v2-pathb-claude.png) |
| codex | gpt-5.5 | $0.0766 | 28560 / 10 | no | no | [shot](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/v2-pathb-codex.png) |
| gemini | gemini-3.5 + 3.1-flash-lite | $0.0173 | 24260 / 100 | yes | yes | [shot](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/v2-pathb-gemini.png) |
| opencode | big-pickle + deepseek-v4 | $0.0015 | 12515 / 1317 | yes | yes | [shot](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/v2-pathb-opencode.png) |

This reproduces the prior matrix's Path B findings exactly (claude / gemini / opencode carry input+output; codex carries neither, its 0.137 logs-exporter gap), confirming the new credential yields identical telemetry. Every span is stamped `langwatch.source` / `langwatch.api_key.id` / `langwatch.origin=ingest_key` / `langwatch.organization_id`. Proof of the interactive sessions: [4-tool tmux panes](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/tmux-panes-4tools.png).

**Path A (gateway VK) re-run complete (2026-06-05):** the final dogfood item is done. All three providers (openai + anthropic + gemini) were configured fresh for the dogfood org from the host `.env`, andre's personal VK (`vk-lw-`) routes through the `developer-default` routing policy (chain now spans all three providers, allowlist opened), and each of claude / codex / gemini / opencode was re-dogfooded interactively in tmux through the gateway on `:5563`. The Path A cells in the matrix below are this fresh re-run (gateway smoke-tested 200 per provider first, then real `langwatch <tool>` sessions). Both halves of the 8-cell matrix now stand on the unified `sk-lw-` / VK foundation.

**All three OTLP signals now carry provenance:** traces + logs already did; metrics are now stamped too (`stampIngestKeyProvenanceOnMetricRequest` in the `/metrics` handler), so claude's session / cost / token / active_time instruments land attributed instead of orphaned. Verified live.

**Single-project self-scoping generalised:** ANY API key scoped to exactly one project now resolves with no projectId (an OTLP exporter sending only the bearer token, or any client that can't add an X-Project-Id header); two-or-more-project keys stay ambiguous and still require an explicit projectId. Previously this was an ingest-only special case.

**Company-wide push ingestion (new capability):** a sysadmin mints ONE `sk-lw-` key bound to the org's hidden Governance Project (`companyWide*` on the ingestionKey router, gated on `ingestionSources:manage`) and pastes it plus the OTLP endpoint into a company-wide tool (e.g. Microsoft Copilot Studio); the whole company's telemetry then flows into the governed project. Verified end-to-end (minted key -> OTLP POST -> span lands in the Governance Project, `origin=ingest_key`). [Admin panel](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/ingest-key-dogfood/companywide-admin-panel.png) on the Ingestion Sources page.

**Bugs the dogfood caught (all fixed + regression-guarded in `ingestionKey.resolution.integration.test.ts`):** (1) `findIngestKey` / `findIngestKeysForProject` queried ApiKey without `organizationId`, so the org-tenancy guard rejected every mint / rotate; (2) the resolver required an externally-supplied projectId, so ingest keys (which send the bearer alone) 401'd every real ingest; (3) rotation hit a latent revoke bug where `deleteExclusiveToApiKey` deleted the exclusive custom role while a CUSTOM binding still referenced it, nulling `customRoleId` against `RoleBinding_custom_role_check` (fix deletes the bindings first).

**Centralized pull-mode ingestion unaffected:** the admin `IngestionSource` -> puller -> org Governance Project path is structurally independent of the retired personal-push binding; verified no binding references in `activity-monitor/*` or `pullers/*`.

### Final v2-drawer matrix - redone 2026-06-06 (all 8 cells, real session ids + cache)

Every cell re-dogfooded from scratch in **interactive tmux**, **2 turns** each (turn 2 forces a cache read + a stable session id), through capturing intercepts so each element is validated against the **raw OTLP/gateway wire**, not just the drawer. Drawer shots are the **v2 Trace Explorer** on `/traces` (Waterfall/Tree/Span List/Topology/Sequence tabs, each PNG visually verified as v2); terminals are rendered with `freeze --window`. Session ids below are the **actual captured values**, not attribute names.

Three bugs were caught and fixed while dogfooding (all pushed on this PR):

- **`24da5e4ed`** gateway was dropping the wrapped tool's session id. It now stamps `gen_ai.conversation.id` from the tool's own session header (claude `x-claude-code-session-id`, codex `session-id`, opencode `x-session-affinity`; gemini sends none on the gateway wire).
- **`45bd9f83f`** claude Path B span input was the **raw truncated** `api_request_body` JSON (claude truncates bodies >64KB). It now uses the clean co-batched `user_prompt` text.
- **`f45ba5477`** opencode Path B (Vercel AI SDK) cache tokens were dropped. `ai.usage.inputTokenDetails.cache{Read,Write}Tokens` now map to `gen_ai.usage.cache_*` (a 12k-token cache-creation turn was going uncounted).

Coverage is locked by a per-element canonicalization test matrix fed by these real dumps (`rawOtlpElementCoverage.matrix.test.ts`, 36 assertions, plus the gateway session-id table) so a future dropped mapping turns the suite red.

### The matrix

`R` = cache read, `W` = cache write. `absent` = the tool genuinely does not emit it on the wire (cited below), not a gap.

| Cell | Model | In / Out tok | Cost | Cache R / W | Session id (captured) | I/O content |
|---|---|---|---|---|---|---|
| **claude-A** gw | claude-opus-4-8 | 10480 / 15 | $0.80 | 34253 / 34253 | `536eaf6c-8c9a-406c-b57d-25d22b0895b6` | Turn 2 -> PONG-CLAUDE-A2 |
| **claude-B** otlp | claude-opus-4-8 (+haiku) | 11185 / 50 | $0.48 | 19474 / 10775 | `4eeced16-44a2-498a-8971-cbfaec4b1e0b` | Turn 1/2 -> PONG-CLAUDE-B1/B2 |
| **codex-A** gw | gpt-5.5 | 2673 / 23 | $0.019 | 13184 / _absent_ | `019e9bfc-c948-71e0-8d03-a1970b2756a9` | Turn 1/2 -> PONG-CODEX-A1/A2 |
| **codex-B** otlp | gpt-5.5 | 26668 / 22 | $0.074 | 13184 / _absent_ | `019e9bfe-cd75-7b51-bd9e-7cccaa21ac47` | _absent_ (token counts only) |
| **gemini-A** gw | gemini-3.5-flash | 11378 / 8 | $0.017 | _absent_ / _absent_ | _absent_ (gateway wire) | Turn 1/2 -> PONG-GEMINI-A1/A2 |
| **gemini-B** otlp | gemini-3.5-flash | 24306 / 76 | $0.017 | _absent_ / _absent_ | `945bd8d5-0657-4385-a99e-30240bfd2e00` | Turn 1/2 -> PONG-GEMINI-B1/B2 |
| **opencode-A** gw | claude-haiku-4-5 | 3 / 13 | $0.016 | 12527 / 12527 | `ses_163f842daffebhRKE6IlNPrGb9` | Turn 1/2 -> PONG-OPENCODE-A1/A2 |
| **opencode-B** otlp | claude-haiku-4-5 | 13252 / 13 | $0.033 | 12629 / 12629 | `ses_163e18ddeffegu1Bw0Ocl5fsDh` | Turn 1/2 -> PONG-OPENCODE-B1/B2 |

**WIRE-ABSENT, cited:** OpenAI usage exposes only cached (read) tokens, no creation count -> codex cache W absent both paths. codex OTLP carries token counts only (zero message-body attrs on the wire) -> codex-B I/O absent; codex-A (gateway) captures I/O, which is the argument for the gateway path. gemini-cli emits no cache token field -> gemini cache absent both paths. gemini sends no per-conversation id on the gateway wire (only a stable device id) -> gemini-A session id absent, while gemini-B carries gemini-cli's own `gen_ai.conversation.id`.

### Path A (gateway VK) - v2 drawer on /traces + terminal

| claude | codex | gemini | opencode |
|---|---|---|---|
|  |  |  |  |
|  |  |  |  |

### Path B (OTLP ingest, sk-lw) - v2 drawer on /traces + terminal

| claude | codex | gemini | opencode |
|---|---|---|---|
|  |  |  |  |
|  |  |  |  |

### claude-B output unlock proof (query_source gate, trace `53a7d100210f996d8f6a09cd84fb5cf9`)

The earlier table claimed claude-B output was an unfixable vendor limit. It was actually unset `OTEL_LOG_*` env knobs (same shape as the gemini-B miss). The `langwatch claude` wrapper now sets all four by default (USER_PROMPTS + TOOL_CONTENT + TOOL_DETAILS + RAW_API_BODIES), and the receiver lifts the assistant text - gated on `query_source` so claude's non-conversational utility calls (`prompt_suggestion` autosuggest, `generate_session_title`) never clobber the headline output via last-write-wins.

**1. Real interactive `langwatch claude` tmux pane** (2 turns, a real Bash tool call, completion markers - `OTEL_LOG_*` knobs injected by the wrapper onto the spawned child):

```
 ▐▛███▜▌ Claude Code v2.1.162
▝▜█████▛▘ Opus 4.7 with xhigh effort · Claude Max
 ▘▘ ▝▝ ~/Projects/langwatch/.claude/worktrees/wise-mixing-zebra
❯ Run a bash command: ls /tmp | head -3 and tell me what you see. Then on its very last line reply EXACTLY this and nothing else:
 ALEXIS-QUERYSOURCE-GATE-5150
⏺ Bash(ls /tmp | head -3)
 ⎿ AdobeIPCBroker.log
 alexis-claude-unlock-state.md
 alexis-codex-auth.json.bak
⏺ Top 3 entries in /tmp: AdobeIPCBroker.log, alexis-claude-unlock-state.md, alexis-codex-auth.json.bak.
 ALEXIS-QUERYSOURCE-GATE-5150
✻ Cooked for 35s
❯ Reply with exactly this and nothing else: ALEXIS-GATE-FINAL-5150
⏺ ALEXIS-GATE-FINAL-5150
```

**2. v2-drawer Thread tab** - INPUT = the real prompt, ORIGINAL RESPONSE = `ALEXIS-GATE-FINAL-5150` (the second `repl_main_thread` reply, no title-gen / autosuggest leak): [02-v2drawer-thread-clean.png](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/claude-querysource-gate-2026-06-05/02-v2drawer-thread-clean.png)

**3. v2-drawer Trace Details tab** - `langwatch.output` = `ALEXIS-GATE-FINAL-5150` verbatim, `langwatch.input` = the real prompt, models + cost + tokens all populated: [03-v2drawer-tracedetails-clean.png](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/claude-querysource-gate-2026-06-05/03-v2drawer-tracedetails-clean.png)

**ClickHouse fold verification** - every `api_response_body` record on the trace and whether the gate folded it into `ComputedOutput`:

| seq query_source | content text | folded to output? |
|---|---|---|
| `generate_session_title` | `{"title":"Execute bash command and verify output"}` | **skipped** (utility) |
| `repl_main_thread` | `Top 3 entries in /tmp: …` | yes (later turn overwrites) |
| `prompt_suggestion` | `show me the full ls output` | **skipped** (autosuggest) |
| `repl_main_thread` | `ALEXIS-GATE-FINAL-5150` | **yes - headline** |
| `prompt_suggestion` | `check the memory file` | **skipped** (autosuggest) |

`ComputedInput` = the real prompt (156 chars), `ComputedOutput` = `ALEXIS-GATE-FINAL-5150` (the last conversational reply), models `claude-haiku-4-5 + claude-opus-4-7`, cost $0.3956, 1515 / 249 tokens, cache 277486 / 39238. Despite the title-gen + two autosuggests being present (any of which clobbered the headline pre-fix), the output is the clean assistant reply. Commits: extractor lift cdc58315a, wrapper + all-platform-surfaces 468f694b9, receiver 256KB size-guard 64a7f203a, extractor `query_source` gate 14c3799b8, fold-path `query_source` gate b856104f8.

### Gateway customer-trace-bridge fixes (Path A visibility, 0cd2dcec8)

Two fixes to `services/aigateway/` so the customer trace bridge tells the truth about gateway-routed (Path A) traffic, verified live:

1. **Upstream errors are now visible.** A failed request (upstream 504 / 429 / 4xx) previously early-returned and dropped the customer span entirely, so users could not see failed requests at all. The pipeline now ends the span on error with the provider's status + an `error.type` class stamped (`classifyUpstream`: 504/408 -> `provider_timeout`, 429 -> `rate_limited`, 5xx -> `provider_error`, 404 -> `not_found`, other 4xx -> `bad_request`) and `span.SetStatus(codes.Error)`. Proven live twice: a synthetic bad-model 404 (`error.type=not_found`, http 404) and a real transient anthropic 504 (`error.type=provider_error`, http 502) both land with StatusCode=Error where pre-fix both were invisible.
2. **Empty internal probe spans are suppressed.** claude-code fires internal probe calls (system-reminder / skills pings) that return no usage and no assistant content. A successful zero-cost, no-output span is marked `langwatch.reserved.drop` and omitted at export by `dropFilterExporter`, so the trace list is not cluttered with empty companions. The suppression explicitly EXCLUDES errors (an errored zero-output span is always kept) and anything with real cost or output. 6 unit tests green; live coupling confirmed (the 504 error span was kept, a real generation was kept, a real empty-probe companion matched the drop predicate exactly).

### Path A + Path B per-cell fix commits (real interactive tmux, no curl)

All eight cells are now GREEN (the earlier claude-B "documented-vendor-limit" for output was disproven and the output unlocked - see above); the verified v2-drawer screenshots are in the matrix table above. The earlier per-cell `refresh-cells-2026-06-04-v2/11-*` trace-detail screenshots were REMOVED: they predated the input/output fixes and showed wrong data (e.g. the claude-B capture rendered a Bash-tool subagent's `env` command in the input slot instead of the user prompt). The matrix above is the single source of truth.

- **claude A** (gateway VK): `gen_ai.messages` span, full input + output + cost.
- **claude B** (OTLP logs): right-seam synthesizer (81012a19e) + native TS extractors lift cost/tokens/model/`thread.id` at fold; input fix 9083c43c1 (gate on `event.name === user_prompt`, lift `session.id` → `langwatch.thread.id`); logs-only persistence fix cb480b692. Output unlocked via the four `OTEL_LOG_*` knobs + a `query_source` gate across both lift paths (cdc58315a/468f694b9/64a7f203a/14c3799b8/b856104f8); proven above.
- **codex A** (gateway VK): output-extraction fix 26a18ff17 (bare-object streaming bodies). Wrapper writes `[model_providers.langwatch]` to `~/.codex/config.toml` + the profile body to the sibling `~/.codex/langwatch-gateway.config.toml` (codex 0.137 CONFIG_PROFILE_V2 requires the split and rejects an inline `[profiles.X]` block), spawns with `--profile langwatch-gateway`.
- **codex B** (OTLP spans): auth-realm fix 69c39576a + endpoint-suffix fix 65b7cdf1f (codex 0.134 ships its OTLP exporter without the `/v1/traces` suffix the Node/Python/Go SDKs append). Span pipeline extractor registration 828af91b9.
- **gemini A** (gateway VK): env-var-name fix 04e1ce8c4 (`GOOGLE_GEMINI_BASE_URL` + both `GOOGLE_API_KEY`/`GEMINI_API_KEY`) + `/v1beta` double-prefix fix cb1f30d07. OAuth-precedence preflight warning 8bdc5a794 (`security.auth.selectedType=gemini-oauth` bypasses the key - wrapper warns to flip it to `gemini-api-key`).
- **gemini B** (OTLP logs): telemetry env-block fix 3c4804595 (6-knob block: ENABLED, TARGET=local, USE_COLLECTOR=true, TRACES_ENABLED=true, OTLP_ENDPOINT, LOG_PROMPTS=true; runtime rejects TARGET=otlp). Real input + output + model + cost.
- **opencode A** (gateway VK): 404 fix 04e1ce8c4 (BASE=`$gw/v1` + `ANTHROPIC_API_KEY` alongside AUTH_TOKEN) + streaming-reasoning Zod-mismatch fix 4c22cac5d (`dispatchMessagesStream` raw-forwards through Bifrost PassthroughStream).
- **opencode B** (OTLP spans): emit-flag fix 781e4c6de (`experimental.openTelemetry: true` in `opencode.jsonc`) + Vercel-scope lift c83ae72fb. Structural spans land; gen_ai.* token/cost chips fill once opencode adopts gen_ai.* semconv upstream (receiver extractor 11baac0f4 fires automatically).

## Trace synthesizer note (claude Path B) - POST-REVERT DESIGN

Empirical findings cross-checked across real claude code 2.1.142 (3 weeks ago) and 2.1.162 (current), interactive AND `-p` print mode, BOTH versions, BOTH modes: **zero POSTs to `/v1/traces`**. The trace exporter env vars are wired but no code path actually uses them for billable API calls - those go through OTLP logs under scope `com.anthropic.claude_code.events`. The 'multi-trace local 3 weeks ago' rchaves recalls predates the Vite migration #3170 (2026-04-14), which dropped a converter that lifted `cost_usd / model / *_tokens / prompt` off `claude_code.api_request` log records.

The right seam (TBD design, in progress by sergey post-ccb50fe74):

1. At `LogRequestCollectionService` receive-time, synthesize `traceId = sha256(session.id)` + `spanId = sha256(session.id||':'||prompt.id)` for `scope=com.anthropic.claude_code.events` log records that arrive without trace context. This restores #4533's fold-projection guard's required fields without changing its semantics.
2. Port the lift from `langwatch/src/server/tracer/otel.logs.ts` (the dead-since-Vite converter) into `foldProjection.handleTraceLogRecordReceived` so `api_request` events enrich `gen_ai.usage.input_tokens / output_tokens / cost / request.model` on the trace summary.
3. Delete `otel.logs.ts` + its tests (dead code).

Multi-turn conversations group into one trace via the stable `sha256(session.id)`. `OTEL_LOG_USER_PROMPTS=1` (already in the drawer + wrapper env block) populates the prompt text on the user_prompt event so the synthesizer can join on input messages.

What this shows: input messages, model, token counts, cost, duration - and, with the four `OTEL_LOG_*` unlock knobs set (now the `langwatch claude` wrapper default), the assistant response text too. The earlier "won't show until anthropic re-adds emission" was wrong: the text was always emittable, just gated behind `OTEL_LOG_RAW_API_BODIES=1` (see claude-B above). Render shape empirically validated in screenshots [11](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/path-a-dogfood/11-claude-pathb-trace-list-after-synth.png) + [12](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/path-a-dogfood/12-claude-pathb-trace-detail-synth-proof.png) from the reverted db022818b run, and re-proven clean (input + output) on trace `53a7d10…`.

## Lane note

Follow-up to #4538 (merged). Disjoint from #4533 (OTLP silent-drop + VK-without-default-policy). No file overlap.

## Test plan

- [x] Typecheck clean (new code 0 errors; pre-existing types.generated unrelated)
- [x] Wrapper unit tests: green (incl. codex split-profile coverage)
- [x] OTTL starter unit tests: green
- [x] codex-config-toml unit tests: 19 green (incl. split-profile-file regression)
- [x] wrapper-mode unit tests: 7 green (incl. codexProfilePath assertion)
- [x] claude-code log->span synthesizer unit tests: 9 green
- [x] log-request-collection.service unit tests: extended with 3 synthesizer tests
- [x] Path A real tmux: claude ✅ codex ✅
- [x] Path B real tmux: claude - right-seam (81012a19e/8fdabbb58/11baac0f4/24a4f8d6c) + log-record-gate fix (cb480b692) green, real tmux PONG-CLAUDE-PATHB-V2 → trace_summaries row with model/cost/tokens/thread.id, renders in /me/traces (screenshot 14)
- [x] Path B real tmux: codex - auth-realm fix (69c39576a) + endpoint-suffix fix (65b7cdf1f) green, real tmux + synthetic roundtrip both materialize trace_summaries
- [x] Path A real tmux: gemini - per-tool envForTool divergence in 04e1ce8c4 (GOOGLE_GEMINI_BASE_URL + GOOGLE_API_KEY/GEMINI_API_KEY, base=`$gw/v1beta`), real-tmux PONG-GEMINI-FINAL → trace b89f52d4 in /me/traces (gemini-3-flash-preview)
- [⚠️] Path A real tmux: opencode - wrapper 404 closed in 04e1ce8c4 (base=`$gw/v1` + ANTHROPIC_API_KEY); gateway streaming Bifrost-shape-rewrite bug for thinking-enabled models filed in-PR for sergey
- [x] Path B real tmux: gemini - receiver-side via gen_ai.* TS extractor (sergey 11baac0f4); langwatch.model lifted onto trace_summary. Walkthrough captures in d06e0985d used the pre-v2 `drawer.open=traceDetails` URL and have been removed (bec1cbca1); the driver now lives at `scripts/dogfood/capture-trace-walkthrough.mjs` and uses the v2 `drawer.open=traceV2Details` URL - re-capture pending.
- [x] Path B real tmux: opencode - fixed in 781e4c6de via `experimental.openTelemetry: true` JSONC merge; 100+ structural spans + llm.* logs land in /me/traces
- [x] /me personal portal + admin governance final-UX walkthrough originally captured in d06e0985d. Those 20 PNGs lived in-tree under `pr-screenshots/pr-4544/final-ux/`, were committed to the wrong repo, and have been removed (bec1cbca1). The Playwright driver moved to `scripts/dogfood/capture-trace-walkthrough.mjs`; canonical screenshot hosting stays at `langwatch/pr-screenshots`. Re-capture against the v2 drawer pending - the v2-drawer claude Path B proof (trace `53a7d10…`) is already at https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/claude-querysource-gate-2026-06-05/.

## Coordination

- alexis: shell-rc save-prompt + remember-choice; wrapper CR fix + OTel-revert; e2e test fix for codex --profile prepend
- sergey: codex / gemini / opencode seeds + tiles + drawer snippets; codex Path A `[model_providers.langwatch]` + split-profile file for codex 0.134; claude Path B log->span synthesizer; OTEL_LOG_USER_PROMPTS wiring
- andre: cli-output-to-png tool; dogfood reports (claude Path A green, codex Path A green post-fix); PR body refreshes


langwatch/scripts/dogfood/cli-output-to-png.mjs
```diff
@@ -0,0 +1,167 @@
...
+        i++;
+        break;
+      case "--bg":
+        out.bg = val;
+        i++;
+        break;
+      case "--fg":
+        out.fg = val;
+        i++;
+        break;
```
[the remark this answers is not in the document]
rogeriochaves: Addressed in 61611d48b — parseArgs now uses requireVal() which exits 1 with 'Missing value for <flag>' when the value is absent or itself a flag. Verified: --out (no value) → missing-value error; --out --title foo → missing-value error.

langwatch/scripts/dogfood/cli-output-to-png.mjs
```diff
@@ -0,0 +1,167 @@
...
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;");
+}
+
+function stripAnsi(s) {
+  // Remove CSI / OSC / SGR sequences so they don't render literally.
+  // eslint-disable-next-line no-control-regex
+  return s.replace(/\x1b\[[0-9;?]*[a-zA-Z]/g, "").replace(/\r/g, "");
```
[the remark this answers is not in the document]
rogeriochaves: Addressed in 61611d48b — stripAnsi now strips OSC sequences (both BEL- and ST-terminated) in addition to CSI. Verified: an OSC 8 hyperlink input ('OSC 8 hello target ST link OSC 8 ST world') renders as plain 'hellolink world' in the PNG.

rogeriochaves:
Addressed CR thread on wrapper.ts:354 in 70391c978: refactored `default_personal_ingestion_token` (single) → `default_personal_ingestion_tokens` (Record<slug, token>). Each wrapped tool now caches + reads its own per-template token. Added a regression test asserting no cross-tool token leakage. 25/25 wrapper tests green.

rogeriochaves:
Addressed CR pass on wrapper-mode.ts in 6ed91ac7c: dropped unused CODEX_GATEWAY_PROFILE_NAME import. Typecheck clean.

typescript-sdk/src/cli/commands/ingestion/install.ts
```diff
@@ -0,0 +1,258 @@
...
+    endpoint,
+    ingestion_token: token,
+    token_action: action,
+    env_block: envBlock,
+  };
+
+  if (tool === "codex" && !options.envOnly) {
+    const result = writeCodexOtelBlock(
+      {
+        endpoint,
```
rogeriochaves: P1: This hidden Codex installer still writes the bare `/api/otel` endpoint into the `[otel]` block, while `wrapper-mode.ts` was updated to append `/v1/traces` because Codex sends this endpoint verbatim. Anyone following the install flow that reaches this command will configure Codex to POST to `/api/otel`, which does not match the OTLP traces route and results in no traces. Please pass `${endpoint}/v1/traces` for the Codex config path here and add coverage for the installed `[otel].exporter.otlp.endpoint` value.

typescript-sdk/src/cli/commands/ingestion/install.ts
```diff
@@ -0,0 +1,258 @@
...
+    case "codex":
+      return [
+        `export OTEL_TRACES_EXPORTER=otlp`,
+        `export OTEL_EXPORTER_OTLP_PROTOCOL=http/json`,
+        ...base,
+        `export OTEL_RESOURCE_ATTRIBUTES="service.name=codex"`,
+      ];
+    case "claude_code":
+      return [
+        `export CLAUDE_CODE_ENABLE_TELEMETRY=1`,
```
rogeriochaves, line 167: P2: The Claude direct-OTLP installer is missing `OTEL_LOG_USER_PROMPTS=1`. The drawer, docs, and wrapper-mode path all include it because Claude redacts prompt attributes without this flag, leaving `/me/traces` with empty user input. Please add the env var here too, ideally with a test/snapshot that keeps this installer in parity with the visible Path B snippet.

langwatch/src/components/me/IngestionTemplateInstallDrawer.tsx
```diff
@@ -38,7 +38,56 @@ export OTEL_EXPORTER_OTLP_HEADERS="Authorization=Bearer ${token}"`;
...
+  }
+  if (slug === "codex") {
+    // Codex 0.130+ links the opentelemetry-otlp Rust SDK + reads
+    // standard OTEL_EXPORTER_OTLP_* env vars but the exporter is
+    // gated on a [otel] block in ~/.codex/config.toml. The CLI
+    // command surfaced below writes that block idempotently so the
+    // user pastes nothing manual; the export block is shown for the
+    // env-only path (CI, devcontainers, agents).
+    return [
+      `# Run once: langwatch ingestion install codex`,
```
rogeriochaves: P2: This snippet points users at `langwatch ingestion install codex`, but the CLI registers the hidden command under `langwatch ingest install <tool>` (`ingestCmd.command("install <tool>")`). As written, following the drawer instruction fails with an unknown command before the config can be written. Please either update the snippet/docs/comments to `langwatch ingest install codex` or register `ingestion` as an alias.

rogeriochaves (commented):
I found a few issues that would break the documented Path B install flows. Details inline.

typescript-sdk/src/cli/commands/ingestion/install.ts
rogeriochaves: Fixed in 00f1dee0a — install.ts now passes `${endpoint}/v1/traces` into writeCodexOtelBlock, mirroring the wrapper-mode.ts fix at 65b7cdf1f. Codex's OTLP/HTTP exporter sends verbatim; the bare /api/otel would have 404'd.

typescript-sdk/src/cli/commands/ingestion/install.ts
rogeriochaves: Fixed in 00f1dee0a — claude_code env block now includes `export OTEL_LOG_USER_PROMPTS=1`. Matches wrapper-mode buildOtelEnvBlock (24a4f8d6c), drawer snippet, and docs. Confirmed empirically on cb480b692 dogfood: trace_summaries materialized with langwatch.input populated (was empty without the flag).

langwatch/src/components/me/IngestionTemplateInstallDrawer.tsx
rogeriochaves: Fixed in a77b55bd3 — three references corrected:

- `IngestionTemplateInstallDrawer.tsx:59`: drawer codex snippet header now reads `# Run once: langwatch ingest install codex`.
- `codex-config-toml.ts:59`: the audit comment the installer writes INTO `~/.codex/config.toml` (so a user reading their own config sees the working command).
- `install.ts:14`: module docstring for consistency.

Took the snippet-correction path rather than registering `ingestion` as an alias — keeps the CLI surface single-named.

scripts/capture-pr4544-ux.mjs
```diff
@@ -0,0 +1,149 @@
...
+        ["settings/governance", "10-admin-governance-overview"],
+        ["settings/governance/ingestion-sources", "11-admin-ingestion-sources"],
+        ["settings/governance/tool-catalog", "12-admin-tool-catalog"],
+        ["settings/governance/cost-centers", "13-admin-cost-centers"],
+        ["settings/governance/users", "14-admin-users"],
+        ["settings/governance/teams", "15-admin-teams"],
+        ["settings/governance/anomaly-rules", "16-admin-anomaly-rules"],
+      ]) {
+        try {
+          await page.goto(`${BASE}/${slug}`, { waitUntil: "domcontentloaded", timeout: 45_000, timeout: 15_000 });
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in a77b55bd3 — kept the runtime-effective value (15_000) and dropped the overwritten 45_000 per the suggestion.

rogeriochaves:
## Vendor-limit cells — quadruple-proven, not inferred from stale screenshots

After rchaves's pushback, we re-verified both partial cells (codex-B output and claude-code tool I/O) with independent evidence vectors, not docstring-reading. Both claims rest on raw OTLP captures + multiple ClickHouse cross-checks now.

### codex-B (Path B native spans, scope `codex_cli_rs`) — quadruple-proven vendor limit

1. CH-keys sweep on trace `1eaf842a3206f14b959086a224450a94`: 212 distinct `(SpanName, key)` pairs across 20 SpanNames. Zero text-bearing keys (no `gen_ai.input.messages`, `gen_ai.output.messages`, `output_text`, `response_text`, `prompt`, `content`, `text`).
2. CH-values sweep on same trace: 45 distinct keys, 383 `(span, key, value)` rows, longest string value across all 40 spans = 60 chars (a thread-id). Zero values >200 chars; text payload is structurally impossible.
3. Independent third-eye CH cross-check on the same trace.
4. **Raw wire capture**: pass-through dump proxy on :4319, real interactive `codex 0.137 exec` Path B, distinctive prompt `ALEXIS-CODEXB-WIRE-MAGENTA-4319`. Codex echoed the token back in tmux. 414683 wire bytes captured (105 real spans across `session_task.turn`/`run_turn`/`handle_responses`/`build_tool_call`/`op.dispatch.user_input`). Decisive grep across `.json` and `.bin` wire bodies: ZERO matches for `MAGENTA-4319`, ZERO for `Reply with exactly`, ZERO for `nothing else`. Only strings >40 chars on the wire are RPC uuids and rust file paths.

Evidence folder: `.claude/dogfood-evidence/alexis-codex-b/` (findings.md + raw-otlp/ + ch/).

What IS on the codex-B wire: `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, `gen_ai.usage.cache_read.input_tokens`, `codex.usage.total_tokens`, `codex.usage.reasoning_output_tokens`, `codex.request.reasoning_effort`, `gen_ai.request.model`, `gen_ai.response.model`, `turn.id`, `thread.id` + rust code-location metadata. No prompt or output text on any span.

### claude-code Path B tool I/O — triple-proven vendor limit

1. CH-keys sweep on a prior dogfood session: 170 distinct `(event.name, key)` pairs across `user_prompt`, `api_request`, `tool_result`, `tool_decision`, `hook_*`, `plugin_loaded`, `mcp_server_connection`. Zero text-bearing keys for tool I/O (only `tool_input_size_bytes`, `tool_result_size_bytes`, `success`, `duration_ms`, `tool_name`, `tool_use_id`, `decision`, `source`).
2. Third-eye CH dump of prior session `a0d47c27` independently confirms.
3. **Raw wire capture**: pass-through dump proxy on :4318, real interactive `langwatch claude`, prompt `"Run a bash command: ls /tmp/lw-andre-codex-real-verify and tell me how many files"`. Claude executed TWO real Bash tools (`ls ... | wc -l`, `ls ...`) and replied `"The directory exists but is empty — 0 files."` 4 `_v1_logs` POSTs captured. Recursive walk for literal `ls /tmp`, `lw-andre-codex`, `wc -l`: ONE match across all bytes = the user-typed prompt itself (on the `user_prompt` event's `prompt` attribute, which the user_prompt extractor at commit 9083c43c1 lifts to `langwatch.input`). Zero matches for the actual Bash command, zero matches for stdout, zero matches for the assistant reply. Zero hits for any `command`, `stdout`, `stderr`, `tool_input.value`, `tool_result.value`, `output_text`, `assistant_text`, `response_text`, `message.content`, `reply` key on any event.

Evidence folder: `.claude/dogfood-evidence/andre-claude-tool-calls/` (findings.md + raw-otlp-bodies/ + tmux-evidence/).

What IS on the claude-code wire for tool calls: `tool_name`, `tool_use_id`, `success`, `duration_ms`, `tool_input_size_bytes`, `tool_result_size_bytes`, `decision_source`, `decision_type`, `prompt.id`, `session.id`. Same vendor-limit class as `api_request` (cost + tokens but no response body) — Anthropic emits observability metadata without the actual content payloads, presumably for privacy and bandwidth.

docs/integration/tools/integrations/claude-code.mdx
```diff
@@ -270,12 +288,21 @@ echo $OTEL_EXPORTER_OTLP_HEADERS
...
     - Verify the OTLP protocol is supported (http/json or grpc)
   </Accordion>
 
+  <Accordion title="Trace shows but input/prompt is empty">
+    Set `OTEL_LOG_USER_PROMPTS=1`. Claude Code redacts the prompt body on its `user_prompt` event by default — without this env, the receiver has no input text to attach to the synthesized trace.
+  </Accordion>
+
+  <Accordion title="Trace shows but output/response is empty">
+    Anthropic does NOT emit the assistant's response text in OTel as of Claude Code 2.1.x. Only token counts ship on the wire. LangWatch renders the input + cost + token breakdown but the output text stays empty until anthropic resumes emitting the response.
+  </Accordion>
```
rogeriochaves, line 306: Those accordions are outdated right? We can give better recommendation?

pr-screenshots/pr-4544/final-ux/01-me-home-ai-tools.png
rogeriochaves, line 1: Why is there a bunch of pr-screenshots files committed? pr-screenshots should be in its own repo not here!

scripts/dogfood/capture-trace-walkthrough.mjs
```diff
@@ -0,0 +1,149 @@
...
+    }
+  } catch (e) {
+    console.error("CAPTURE ERROR", e);
+    await shot(page, "99-error-state");
+    process.exitCode = 1;
+  } finally {
+    await ctx.close();
+    await browser.close();
+  }
+})();
```
rogeriochaves, line 224: This also seems like a very specific script, weird to commit something with the pr in the name, move it to a proper namespaced dogfood/ folder on scripts

docs/integration/tools/integrations/claude-code.mdx
rogeriochaves: Fixed in 64ef6058a. You were right, that accordion was stale. It claimed Claude Code never emits the assistant response, but the OTEL_LOG_RAW_API_BODIES unlock this same doc already recommends in Quick Start disproves it. Rewrote it to:

- recommend OTEL_LOG_RAW_API_BODIES=1 (the real output unlock, dogfood-proven end to end)
- note the query_source filter that keeps utility calls (autosuggest, session-title) out of the trace output

Also swept the other stale ones in the group: the Performance accordion was generic OTel-SDK advice (batch processors / sampling) that does not apply to Claude Code's fixed exporter, replaced with a real large-payload + ingestion-cap note; and fixed the Security section line that claimed file contents never appear, which the content flags directly contradict.

pr-screenshots/pr-4544/final-ux/01-me-home-ai-tools.png
rogeriochaves: Removed in bec1cbca1 - all 20 PNGs + README under pr-screenshots/pr-4544/final-ux/ are git rm'd. Canonical hosting stays at langwatch/pr-screenshots external repo; v2-drawer claude proof URLs (53a7d10...) remain at https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4544/claude-querysource-gate-2026-06-05/.

scripts/dogfood/capture-trace-walkthrough.mjs
rogeriochaves: Moved + renamed in bec1cbca1: scripts/capture-pr4544-ux.mjs -> scripts/dogfood/capture-trace-walkthrough.mjs. Rewrote the script to: drop the PR-specific naming, resolve @playwright/test from the repo's langwatch/node_modules via PLAYWRIGHT_TEST_PATH override (removing the hardcoded /tmp/lw-codex-gemini path that wouldn't work elsewhere), default OUT_DIR to .claude/dogfood-evidence/, and use drawer.open=traceV2Details so trace-detail captures render the v2 drawer (this also fixes the OLD-drawer screenshots you flagged on d06e0985d - the old script's URL pattern was the root cause).

langwatch/src/components/me/TraceIngestSection.tsx
```diff
@@ -45,6 +48,12 @@ import { api } from "~/utils/api";
  * IngestionTemplate row, no install). It deep-links to
  * /me/configure#otlp — the BYO-OTLP fallback discovery card.
  *
+ * Claude Code is intentionally excluded from this grid (filtered by
+ * slug below). The unified entry point for Claude Code lives on the
+ * AiToolsPortal "$ langwatch claude" tile — the CLI auto-mints the
+ * ingestion token and wires both the gateway and OTLP paths in one
+ * step, so a separate "Connect" tile here would be a duplicate UX.
```
rogeriochaves: not only claude code right but all the first class supported coding assistants?

rogeriochaves: Yep, you're right — fixed in aa774936d. Added codex/cursor/gemini to UNIFIED_VIA_CLI_SLUGS alongside claude_code so all four first-class `langwatch <tool>` CLI tiles stop double-rendering as 'Connect' install tiles. opencode stays in the install grid because it has no unified `langwatch opencode` CLI yet (OTLP-only via the drawer). Updated the docstring to call this out explicitly.

typescript-sdk/src/cli/utils/governance/login-flow.ts
```diff
@@ -269,6 +277,66 @@ async function openInBrowser(url: string, override?: string): Promise<void> {
...
+ *     (process.env.ANTHROPIC_BASE_URL + ANTHROPIC_AUTH_TOKEN present)
+ *   - stdin is not a TTY (CI / scripted callers)
+ *   - the shell isn't one we can write to (cmd / powershell stay
+ *     on the eval'd `langwatch init-shell` flow)
+ *
+ * The prompt is Y / n / never. "n" doesn't persist — the next
+ * login on an unconfigured shell re-asks. "never" sets
+ * shell_rc_preference=skip so we stay quiet on this machine.
+ */
+async function maybeOfferShellRcPersist(
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in cdbb2eecf by removing the dead function. The suggested fix (re-invoke at login) would reintroduce the zshrc prompt that login deliberately dropped when it became auth-only. The persist offer now lives in the langwatch <tool> wrapper ingestion path (maybeOfferIngestionShellRcPersist in shell-rc.ts), so maybeOfferShellRcPersist and its shell-rc imports are removed rather than re-wired.

langwatch/src/server/routes/auth-cli.ts
```diff
@@ -1232,6 +1234,128 @@ secured.access(CLI_POLICY).get("/governance/status", async (c: Context) => {
...
+        {
+          error: "invalid_request",
+          error_description: parsed.error.message,
+        },
+        400,
+      );
+    }
+    const service = IngestionKeyService.create(prisma);
+    try {
+      const result = await service.ensureForPersonalProject({
```
rogeriochaves, line 1331: P1: This mint route never checks `PlatformToolPolicy.allowOtelDirect`. The wrapper blocks Path B client-side, but any stale CLI or direct caller with an `lw_at_*` token can still POST `{ source_type: "claude_code" }` here and receive a valid `sk-lw-*` key after an admin disables direct OTLP. That key authenticates at `/api/otel` with `traces:create`, so the policy is advisory rather than enforced. Please enforce the org platform policy before minting/rotating ingestion keys for platform tool source types, including the tRPC `ingestionKey.install`/`rotate` path if those source types can be reached there, and reject when `allowOtelDirect` is false.

rogeriochaves (commented):
I found one P1 issue in the refactor: the new direct-OTLP policy is enforced client-side, but the server still mints direct OTLP ingestion keys. That makes the admin toggle bypassable by any stale CLI or direct HTTP caller with a valid CLI access token. Details inline.

Note: I attempted to submit this as `REQUEST_CHANGES`, but GitHub rejected it as an own-PR review for this account, so I’m posting it as `COMMENT`.

langwatch/ee/governance/services/ingestionKey.service.ts
```diff
@@ -0,0 +1,241 @@
...
+        callerUserId,
+        callerIsAdmin: true,
+        organizationId,
+      });
+    }
+
+    const { token, apiKey } = await this.apiKeys.create({
+      name: `Ingestion key (${sourceType})`,
+      // Project credential, not user-owned — authz is enforced at the router.
+      userId: null,
```
rogeriochaves: P2: Personal ingestion keys are created with `userId: null`, which makes them indistinguishable from org service keys on the API-key list path. `ApiKeyRepository.findAllByUser()` includes every `userId: null` key for non-admin callers, and this PR now exposes `ingestSourceType`, `ingestionTemplateId`, `createdByUserName`, dates, and project scope names in `apiKey.list`. Result: any org member can see other members' personal CLI/template ingestion keys and their activity metadata in Settings > API Keys. Please either keep personal ingestion keys user-owned (while still project-scoped for auth), or change the list/query/UI split so non-admins only receive ingestion keys for their own personal project while true company-wide service keys remain admin-only.

langwatch/ee/governance/routers/ingestionKey.ts
```diff
@@ -0,0 +1,145 @@
...
+   * Mint (rotating in place) an ingestion key for the caller's personal
+   * project + sourceType. Returns the plaintext token ONCE; subsequent
+   * reads only see the source list.
+   */
+  install: protectedProcedure
+    .input(mintInput)
+    .use(checkOrganizationPermission("organization:view"))
+    .mutation(async ({ ctx, input }) => {
+      const service = IngestionKeyService.create(ctx.prisma);
+      return await service.ensureForPersonalProject({
```
rogeriochaves, line 71: P2: This route trusts `sourceType` and `templateId` from the client and passes them straight into the ApiKey row. The retired binding service first resolved `templateId` through `findByIdForOrg`, rejected disabled/archived/cross-org templates, and used the template's own id/source metadata. With the new code, any org member can mint a key for an arbitrary or disabled `templateId`, or pair a visible template id with a different `sourceType`, and the receiver will stamp that forged `langwatch.template.id`/`langwatch.source` on OTLP data. Please validate `templateId` server-side for personal and company-wide install/rotate: visible to the org, enabled, not archived, and sourceType derived from or checked against the template row.

rogeriochaves (commented):
Additional exhaustive-pass findings:

- P2: personal ingestion keys are persisted as service keys, so the existing API-key list path returns other users' personal ingestion-key metadata to non-admin org members.
- P2: `ingestionKey.install`/`rotate` accepts arbitrary `templateId` + `sourceType` without validating the template is visible/enabled and matches the source type.

Cleanup notes from the full scan: `git diff --check origin/main...HEAD` reports a trailing blank line at `langwatch/ee/governance/dashboard/components/CompanyWideIngestionPanel.tsx:197`, and tracked `langwatch/src/app/api/openapiLangWatch.json` still documents removed `/api/governance/user-ingestion-bindings` routes and old `UserIngestionBinding` descriptions even though the Hono routes were deleted.

langwatch/src/server/routes/auth-cli.ts
rogeriochaves: Intentional, not a bug. The per-tool `allowOtelDirect` is a facilitation toggle that decides which path the `langwatch <tool>` CLI offers, not a server-side ingestion gate. An ingest key is a write-only `traces:create` credential and a holder can push any OTLP they want by design, so minting is deliberately not blocked on the policy. Leaving the route as-is.

langwatch/ee/governance/services/ingestionKey.service.ts
rogeriochaves: Fixed in f7cf390df. Personal ingest keys are now user-owned: `ownerUserId` is threaded through `ensureForProject` (personal mints pass the user, company-wide pass null), so `findAllByUser` scopes them to their owner and org admins, and company-wide governance-project keys are excluded from the non-admin list (admins reach them via the company-wide list). Setting `userId` subjects the mint to the owner ceiling, which a personal-workspace owner satisfies via team-admin; a regression test in `ingestionKey.guards.integration.test.ts` confirms minting + the list scoping.

langwatch/ee/governance/routers/ingestionKey.ts
rogeriochaves: Intentional, same facilitation rationale as the `allowOtelDirect` toggle: an ingest key already lets the holder push arbitrary OTLP (including any `langwatch.source` / `langwatch.template.id` they set on their own spans), so we deliberately do not gate minting on template visibility or state. Leaving the route as-is.

PR #4547, opened 2026-06-03
refactor: remove Elasticsearch and the BullMQ legacy stack

0xdeafcafe:
## Why

Elasticsearch was dead in the live read/write path — the
TraceService / EvaluationService / ExperimentRunService /
FilterService facades each had two backends but every call already
went through ClickHouse, with the ES backend instantiated and never
invoked. The BullMQ legacy stack under `src/server/background/` had
been replaced piecewise by the event-sourcing collector (`recordSpan`)
but the old workers, queues, and the tasks that wrote to ES were
still wired at boot. Keeping both alive cost us the
`@elastic/elasticsearch` dependency, two `bullmq` queues running on
every worker process, dead cron handlers triggering ES queries against
clusters most environments no longer provision, and a layer of facade
indirection in front of services that have one backend.

## Deployment Impact

**Env vars — none added or changed.** Four variables stop being read anywhere in the codebase: `ELASTICSEARCH_NODE_URL`, `ELASTICSEARCH_API_KEY`, `IS_OPENSEARCH`, and `IS_QUICKWIT`. Leaving them set is harmless (they are simply ignored); they can be deleted from secrets/config at leisure. No new secrets or config are required to upgrade.

**Helm values — no chart changes ship in this PR.** Any elasticsearch/opensearch values in a deployment (subchart enablement, node URL, API key) become inert with this image and can be removed in a follow-up chart cleanup. The worker deployment is unchanged: same `start:workers` entrypoint and the same Redis requirement — the process now runs the event-sourcing/GroupQueue stack plus the remaining Redis-backed queues (topic clustering, usage stats) instead of the legacy BullMQ `src/server/background/` worker.

**Vanilla `helm install` with no overrides:** app and workers boot and serve with no Elasticsearch/OpenSearch/Quickwit reachable — the ES client is gone from the build entirely (`@elastic/elasticsearch` removed from dependencies, and pnpm overrides pin `@elastic/elasticsearch` / `@opensearch-project/opensearch` to `-` so they cannot resolve transitively). Required stores after this PR: Postgres, ClickHouse, Redis.

**BYOC dataplane:** BYOC installs can decommission their ES/OpenSearch cluster once settled on this version — all reads and writes are ClickHouse. No migration or backfill runs at deploy: ES data is abandoned in place, so anything never migrated to ClickHouse stops being readable through the product. Keep ES up only as a rollback hedge; rolling back to a pre-PR image requires ES to be reachable again. One operational shift to know: REST `/api/collector` dedup moves from ES content-hashes to Redis `SET NX` (1h TTL), so Redis absorbs that load (details in the verify notes below).

## What changed

Two larger refactors (below), plus a series of iteration fixes from review/CI feedback (TC restore, lockfile fix, crypto import, test fallout, build-chunk isolation, scenario tagging, TC circular-import fix, prompts polling fix, dead-import cleanup in `runEvaluation`, ES-remnant purge) and three merges of `main` to keep the branch current.

**1. Remove Elasticsearch and the BullMQ legacy stack** (196 files)

- Drop every ES service backend in the facades
  (`Elasticsearch{Trace,Evaluation,ExperimentRun,Filter,Analytics}Service`).
- Drop the ES sync reactors (`evaluationEsSync`, `experimentRunEsSync`,
  `annotationEsSync`) — no more dual-write.
- Drop the ES analytics timeseries + scenario-analytics query
  builders, the ES experiments batch-evaluation repository, the ES
  scenario-event service/repository.
- Strip the Elasticsearch URL/API-key fields from the org settings
  input schema, the settings UI, and `organizations.update`.
- Drop the ES-driven cron handlers `/cron/scenario_analytics` and
  `/cron/traces_retention_period_cleanup`; `/cron/trace_analytics` is
  kept but stripped to only the SaaS usage-limit check (its ES
  analytics half is gone), and `/cron/schedule_topic_clustering`
  stays (topic clustering was restored after review).
  Also drop the ES task scripts the deleted handlers invoked (`elasticMigrate`, `cold/*`,
  `deleteTracesRetentionPolicy`, `syncAnnotations`, `syncTraceCosts`,
  `backfillTraceAnalytics`, `backfillScenarioAnalytics`,
  `fixCustomMetadataOutsideCustomField`, `clearTopics`, `rerunChecks`).
- Collapse `env.IS_QUICKWIT` branches and replace ES query-builder
  type imports with neutral structural types in the filter / analytics
  / common-router files.
- Move `Protections` out of `src/server/elasticsearch/` to
  `src/server/traces/`.
- Delete `src/server/background/` entirely; move its pure helpers
  (`cost`, `common`, `evaluations`, `evaluationNameAutoslug`,
  `metrics`, `piiCheck`, `rag`) into `src/server/tracer/collector/`
  for the app-wide consumers.
- Extract `runEvaluation` / `runEvaluationForTrace` / `customEvaluation`
  from the deleted evaluations worker into
  `src/server/evaluations/runEvaluation.ts`, switching ES
  `getTraceById` / `getTracesGroupedByThreadId` for
  `TraceService.getById` / `getTracesByThreadId`.
- Rewrite `src/workers.ts`: drop the 15-minute self-restart timer
  (helm/docker already restart-on-exit), drop the BullMQ Worker
  startups, initialize the worker-role app, and start the EE
  governance ingestion-puller worker via a new
  `ee/governance/services/pullers/pullerQueue.ts`.
- Migrate `httpProxyTracing` from
  `scheduleTraceCollectionWithFallback` to
  `getApp().traces.recordSpan` via `CollectorSpanUtils` — the only
  remaining caller of the old collector queue.
- Strip the legacy collector dedup (`fetchExistingMD5s` +
  256-version cap) from `routes/collector.ts`; the REST path now shares
  the `ingestNormalizedSpan` dedup gate with the OTLP path.
- Drop `/api/start_workers`, `/api/rerun_checks`,
  `/api/cron/scenario_analytics`,
  `/api/cron/traces_retention_period_cleanup` — the routes were
  exclusively legacy-stack triggers.
- Topic clustering stays: it was deleted in the first cut, flagged in
  review, and restored — `src/server/topicClustering/`, the
  `/settings/topic-clustering` page + command-bar entry + route, and
  the worker now running alongside the EE puller in `src/workers.ts`.
- Rewrite `TraceUsageService` CH-only (drop the `EsClientFactory`
  constructor parameter and the `splitProjectsByFlag` fallback).
- Delete `src/server/elasticsearch.ts` + `src/server/elasticsearch/`
  + `elastic/` migrations entirely.
- Remove `@elastic/elasticsearch` from `package.json` + lockfile.
- Collapse the four single-backend service facades:
  - `EvaluationService` (delete `clickhouse-evaluation.service.ts`;
    facade methods now throw on missing CH client instead of
    null-falling-back; keep `getEvaluationInputs` nullable for the
    legitimate empty case).
  - `ExperimentRunService` (delete
    `clickhouse-experiment-run.service.ts`; fold
    `listRunsForExperimentSlugPaginated` into the renamed class).
  - `FilterService` (renamed from `FilterServiceFacade`, delete
    `clickhouse-filter.service.ts`).
  - `SimulationFacade` — deleted entirely. Every consumer (suite
    router, scenario events router, simulation-runs REST,
    cancellation router, onboarding-checks service) now calls
    `getApp().simulations.runs` directly.

**2. Isolate the BullMQ adapter to the EE governance puller** (7 files)

After (1), the only BullMQ-importing file left in `src/server/` was
the context adapter at `adapters/bullmq.ts`, and the only caller of
its `withJobContext` helper was the EE governance puller. The other
three exports (`createContextFromJobData`, `getJobContextMetadata`,
`JobDataWithContext`) have no BullMQ dependency — they're pure ALS
plumbing.

- Move the pure helpers into `src/server/context/core.ts`.
- Move `withJobContext` (the one piece that imports BullMQ's `Job`
  type) into `ee/governance/services/pullers/withJobContext.ts`,
  co-located with its only consumer.
- Move the test cases (new `__context` shape + legacy `__payload`
  migration paths) into
  `ee/governance/services/pullers/__tests__/withJobContext.unit.test.ts`.
- Delete `src/server/context/adapters/bullmq.ts`; update
  `asyncContext.ts` to re-export the pure helpers from `core` only.

Result: BullMQ is confined to a small set of queue modules. (Note: the
later topic-clustering restore and the `makeQueueName` build-isolation
commit moved things around, so the BullMQ-importing files ultimately
settled under `src/server/queues/` — `queueWithFallback`, `withJobContext`,
`makeQueueName` — plus `src/server/topicClustering/` (queue + worker) and
`ee/governance/services/pullers/` (the EE puller), rather than the single
`ee/.../pullers/` folder this section originally described.) Dropping
`bullmq` / `bullmq-otel` therefore means migrating those three areas off
it — more than a one-folder change, but still a contained set of callers.

## How it works

- **Facade collapse rules**: the three collapsed services
  (`EvaluationService`, `ExperimentRunService`, `FilterService`)
  changed their not-found-CH-client semantics from "return `null`,
  let the caller fall back to ES" to "throw — ClickHouse is the only
  backend now." The exception is `EvaluationService.getEvaluationInputs`,
  which keeps its `null` return because callers treat "no inputs
  recorded" as a legitimate empty case, not an error.
- **`TraceService` was kept as a facade** even though its ES backend
  is also gone, because the facade still owns prefix-resolution +
  `AmbiguousTraceIdPrefixError`. Collapsing it would push that domain
  logic into a router.
- **`SimulationFacade` was deleted entirely** because it had become a
  pure pass-through after dropping ES routing. The only consumer-side
  rename was `getScenarioSetsDataForProject` →
  `getScenarioSetsData` (the underlying service method).
- **`makeQueueName` lives in its own module** (`src/server/queues/makeQueueName.ts`). It's a Redis-Cluster hash-tag util that scenario.constants needs; keeping it standalone stops the SimulationsPage chunk from pulling `ioredis` into the client bundle via the `~/server/redis` barrel.

## Test plan

- [ ] CI typecheck (`pnpm typecheck`) — clean on touched files (no
      new TS errors vs main; remaining errors are pre-existing stale
      Prisma client issues that also exist on `origin/main`).
- [ ] CI biome (`pnpm lint`) — touched files were auto-fixed; repo
      lint count is now **lower** than `origin/main` (-158 errors).
- [ ] Verified locally against `make quickstart all-local`:
  - [x] dev tree (api + workers + vite) reaches
        `ingestion puller worker ready` clean
  - [x] all 6 event-sourcing pipelines register
  - [x] sign-up → onboarding → org + project creation
        (`organizations.createAndAssign` — path where
        `scheduleUsageStatsForOrganization` was removed)
  - [x] org settings page renders without Elasticsearch fields;
        sidebar has no Topic Clustering entry
  - [x] `POST /api/collector` accepts a span (no
        `fetchExistingMD5s` dedup path), event-sourcing pipeline
        persists to local CH (~13ms p50)
  - [x] traces list shows the seeded trace; trace drawer renders
        full data (exercises `TraceService.getById`,
        `getTracesByThreadId`, `EvaluationService.getEvaluationsMultiple`)
  - [x] `analytics.dataForFilter` returns dropdown options
        (exercises the renamed `FilterService`)
  - [x] `scenarios.getScenarioSetsData` returns clean `[]`
        (exercises the post-`SimulationFacade` consumer path)
  - [x] `annotation.create` returns the new annotation cleanly
        (no `esSync` try/catch left)
  - [x] deleted routes (`/api/start_workers`,
        `/api/rerun_checks`, `/api/cron/scenario_analytics`,
        `/api/cron/traces_retention_period_cleanup`) all return 404

## Anything surprising?

- **The Prisma columns `Organization.elasticsearchNodeUrl /
  .elasticsearchApiKey / .useCustomElasticsearch` are NOT dropped**
  in this PR — only the application layer that reads/writes them. The
  columns still exist in `schema.prisma` and come back as `null` on
  every `organization.getAll`. A follow-up Prisma migration can drop
  them once we're confident no environment depends on them.
- **`bullmq` and `bullmq-otel` stay in `package.json`** because the
  EE governance ingestion puller still uses BullMQ's repeatable-jobs
  (cron pattern) primitive — alongside the restored topic-clustering
  queue/worker and the shared `src/server/queues/` helpers. Those are the
  areas an eventual BullMQ removal has to migrate (see the corrected note
  in the commit-2 section above).
- **`@aws-sdk/s3-request-presigner` was already missing from
  `node_modules`** on this worktree — added back via `pnpm install`
  during this work. The lockfile bumps that came along (`hono`, `qs`,
  `ws`, `mermaid`, `liquidjs`, `turbo`, `@opentelemetry/*` minors)
  are unintentional but harmless.
- **Worker-path callers of the collapsed services lost their implicit
  null-→-soft-skip behavior.** `runEvaluation.ts:293` and
  `evaluation-execution.service.ts:273` call
  `traceService.getEvaluationsMultiple` to enrich `trace.evaluations`
  before mapping; pre-PR a `null` return meant "no evals recorded" and
  the worker proceeded with `[]`. Post-PR the collapsed
  `EvaluationService.getEvaluationsMultiple` throws on no-CH (matching
  the rest of the collapsed-facade rule), so a transient CH
  unavailability now **fails the evaluation job** instead of producing
  an empty-evals result. Arguably more correct (there's no data to
  evaluate against) but it is a real behavior change in worker
  semantics.
- **15-minute worker self-restart was removed intentionally.** The
  original timer was a memory-leak safety net specifically for the
  BullMQ collector + evaluations workers, which held long-lived
  references to enqueued job payloads and could accumulate RSS
  across their processing loops. This PR removes both worker
  classes, so the leak profile that motivated the timer is gone.
  The two BullMQ queues that remain (topic-clustering, ingestion
  puller) hold bounded state — topic clustering runs infrequently
  and doesn't retain payloads across cycles; the ingestion puller
  streams and doesn't accumulate. Supervised deployments
  (helm/docker/systemd) get container-level OOM-restart as
  first-class. For the `npx @langwatch/server` path the original
  timer was scoped to, if a real leak surfaces the right fix is a
  targeted RSS-check or supervised process wrapper — not a
  periodic-restart hammer that also masks the underlying leak.
- **REST `/api/collector` dedup semantics changed materially, not
  just relocated.** The PR routes the REST path through
  `ingestNormalizedSpan` (same as OTLP), which uses
  `SpanDedupService`: Redis `SET NX` keyed on
  `(tenantId, traceId, spanId)`, with a 60s processing TTL extended
  to 3600s (1h) on success. The pre-PR REST path used
  `fetchExistingMD5s` — content-hashes of full request params kept
  in ES with no time bound (only a 256-entry cap). Concrete
  behavior changes: (a) retries/replays/backfills resending the
  same span content more than 1h after original ingest are no
  longer deduped; (b) two requests with the same span-id but
  *different* content are now silently deduped as identical (old
  MD5 check would have detected the mismatch and reprocessed). The
  new gate is unit-tested at the service layer
  (`trace-request-collection.service.unit.test.ts:95-119`); a
  REST-through-route integration test for dedup is a follow-up.
- The verify dev stack used `make quickstart all-local` against a
  fresh local Postgres / ClickHouse / Redis. Worth knowing for
  reviewers reproducing: needs `SKIP_HOST_REDIS_CHECK=1` if you have
  a local Redis on `:6379` (per CLAUDE.md, that's the documented
  singleton).


rogeriochaves (commented):
3 findings: 1 CI/install blocker and 2 stale references left by the ES/BullMQ/topic-clustering removal. Details inline.

langwatch/pnpm-lock.yaml
```diff
@@ -907,6 +904,31 @@ importers:
         specifier: ^8.55.0
```
rogeriochaves, line 902: **P1** — This lockfile now contains an importer for `packages/insanity-check`, but that workspace directory is not present in the repo (`packages/` only contains `analytics-parity-check`). The same lockfile also pulls in newly published `hasown@2.0.4` and `tinyexec@1.2.3`, and a targeted `pnpm vitest ...` run failed before tests with the repo supply-chain policy rejecting those entries. Minimal fix: regenerate the lockfile from this repo/workspace only and remove the unrelated importer/new transitive bumps so install/test can pass under the active policy.

langwatch/src/routes.tsx
```diff
@@ -165,10 +165,6 @@ const routes: RouteObject[] = [
     path: "/settings/teams/:team",
```
rogeriochaves, line 182: **P2** — Removing this route leaves the settings sidebar with a live `MenuLink href="/settings/topic-clustering"` in `SettingsLayout`, so users can still navigate to a deleted page and hit a 404. Minimal fix: remove the Topic Clustering entry and its active path from `SettingsLayout` in the same PR, or keep a redirect/route if the page is intentionally deprecated.

langwatch/src/server/app-layer/traces/__tests__/span-pii-redaction.service.test.ts
```diff
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import type { PIICheckOptions } from "~/server/background/workers/collector/piiCheck";
+import type { PIICheckOptions } from "~/server/tracer/collector/piiCheck";
```
rogeriochaves, line 2: **P2** — This file updates the type import to the new collector location, but the test still mocks `~/server/background/workers/collector/piiCheck` below even though this PR deletes `src/server/background`. The same stale mock pattern remains in `httpProxyTracing.integration.test.ts` and `evaluations.azure-byok.integration.test.ts`. Minimal fix: update those mocks to the new modules (`~/server/tracer/collector/piiCheck`, `getApp().traces.recordSpan`, and `~/server/evaluations/runEvaluation` as appropriate) or remove them if the dependency is no longer imported.

langwatch/pnpm-lock.yaml
0xdeafcafe: After the latest rebase the lockfile is identical to main (`git diff origin/main..HEAD -- pnpm-lock.yaml` shows no changes). The `packages/insanity-check` importer + the rejected hasown/tinyexec bumps are gone. Fixed in 03be9869a54c638f7131b437c9ae75a996683db7.

langwatch/src/routes.tsx
0xdeafcafe: Topic clustering was deleted in error — restored in 03be9869a54c638f7131b437c9ae75a996683db7. The `/settings/topic-clustering` route, the SettingsLayout MenuLink, and the command-bar entry are all back, and the worker now runs alongside the EE puller in `src/workers.ts`.

langwatch/src/server/app-layer/traces/__tests__/span-pii-redaction.service.test.ts
0xdeafcafe: Fixed in 03be9869a54c638f7131b437c9ae75a996683db7: mock paths updated to the new homes — `~/server/tracer/collector/piiCheck` here, `~/server/evaluations/runEvaluation` in evaluations.azure-byok, and `getApp().traces.recordSpan` (mocked via `~/server/app-layer/app`) in httpProxyTracing. The httpProxyTracing test bodies stayed unchanged via a small facade that decodes the OTLP span back into the legacy CollectorJob shape.

langwatch/src/server/evaluations/runEvaluation.ts
```diff
@@ -0,0 +1,533 @@
+import { CostReferenceType, CostType } from "@prisma/client";
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 54cea3de5 — the unused imports and logger were dropped from runEvaluation.ts (and the bare string throw became a proper Error).

langwatch/src/server/evaluations/runEvaluation.ts
```diff
@@ -0,0 +1,533 @@
...
+import { nanoid } from "nanoid";
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 54cea3de5 — the unused imports and logger were dropped from runEvaluation.ts (and the bare string throw became a proper Error).

langwatch/src/server/evaluations/runEvaluation.ts
```diff
@@ -0,0 +1,533 @@
...
+import { getProtectionsForProject } from "~/server/api/utils";
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 54cea3de5 — the unused imports and logger were dropped from runEvaluation.ts (and the bare string throw became a proper Error).

langwatch/src/server/evaluations/runEvaluation.ts
```diff
@@ -0,0 +1,533 @@
...
+import { TraceService } from "~/server/traces/trace.service";
+import { env } from "../../env.mjs";
+import {
+  AVAILABLE_EVALUATORS,
+  type BatchEvaluationResult,
+  type EvaluatorTypes,
+  type SingleEvaluationResult,
+} from "../../server/evaluations/evaluators.generated";
+import { createLogger } from "../../utils/logger/server";
+import { captureException, withScope } from "../../utils/posthogErrorCapture";
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 54cea3de5 — the unused imports and logger were dropped from runEvaluation.ts (and the bare string throw became a proper Error).

langwatch/src/server/evaluations/runEvaluation.ts
```diff
@@ -0,0 +1,533 @@
...
+} from "./threadMappingResolver";
+
+class UserConfigError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "UserConfigError";
+  }
+}
+
+const logger = createLogger("langwatch:evaluations:runEvaluation");
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 54cea3de5 — the unused imports and logger were dropped from runEvaluation.ts (and the bare string throw became a proper Error).

langwatch/src/server/api/routers/httpProxyTracing.ts
```diff
@@ -5,9 +5,13 @@
  * When a user tests an HTTP agent, this module creates a trace capturing
  * request/response details while redacting sensitive auth credentials.
  */
+
 import crypto from "node:crypto";
-import type { Span, CustomMetadata } from "../../tracer/types";
-import { scheduleTraceCollectionWithFallback } from "../../background/workers/collectorWorker";
+import { getApp } from "../../app-layer/app";
+import { prisma } from "../../db";
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in `10cdf4cb9`: removed the unused `prisma` import from `httpProxyTracing.ts:11`.

langwatch/src/workers.ts
```diff
@@ -14,48 +16,157 @@ const logger = createLogger("langwatch:workers");
...
+process.on("SIGINT", () => void gracefulShutdown());
+process.on("SIGTERM", () => void gracefulShutdown());
+
+// The worker/queue/app modules below construct Redis-connecting
+// `QueueWithFallback` instances (or otherwise touch Redis) at load, and the
+// app graph must load only after `setEnvironment()` above has run. They're
+// therefore imported dynamically — only after `verifyRedisReady()` confirms the
+// connection — to preserve that ordering. Keep them as `await import()` (not
+// top-level) for that reason; `http`/`prom-client` are safe to import eagerly.
+void verifyRedisReady().then(async () => {
```
Aryansharma28, line 56: **[BLOCKER]** Worker memory-leak safety restart removed with no replacement.

**Why:** Pre-PR `src/workers.ts` on `origin/main` had this:
```ts
// Self-restart cadence: by default the worker exits cleanly every 15 min
// (memory-leak safety pattern) and an external supervisor (helm/docker)
// brings it back up immediately. In environments without that supervisor
// (e.g. `npx @langwatch/server`, where supervise() in spawn.ts doesn't
// restart-on-exit), set `LANGWATCH_WORKERS_MAX_RUNTIME_MS=0` to disable
// the timer and run forever.
const maxRuntimeMs = maxRuntimeMsRaw !== undefined && maxRuntimeMsRaw !== ""
  ? Number(maxRuntimeMsRaw) : 15 * 60 * 1000;
```

The PR description argues the timer is fine to drop because `helm/docker already restart-on-exit`. That's exactly backward — those supervisors fire on **exit**; the 15-min timer was what *caused* the periodic exit so that a slow leak couldn't grow unbounded.

The post-PR file (this version) has no `setTimeout`/`setInterval`/RSS check/liveness probe — only a graceful shutdown wired to `SIGINT`/`SIGTERM`/`uncaughtException`. None of those fire on a slow memory leak. The original comment specifically named `npx @langwatch/server` (where `supervise()` in `spawn.ts` doesn't restart-on-exit) as the case the safety pattern protected; that path is now unprotected.

**Adversarial check:** I read the file end-to-end looking for any other periodic exit / RSS-watch / liveness check. Nothing. The PR description's "would have needed a fabricated intermediate version" argument is also weak — the pre-PR `maxRuntimeMs` was clearly separable from the BullMQ Worker startups (it lived as a top-level const passed into `start(void 0, maxRuntimeMs)`); preserving it would have been ~5 lines.

**Suggested fix (any one of):**
1. Preserve the env-gated 15-min timer behind the same `LANGWATCH_WORKERS_MAX_RUNTIME_MS` switch (default-on, ~5 lines: `if (maxRuntimeMs > 0) setTimeout(() => process.exit(0), maxRuntimeMs).unref()`).
2. Document that the underlying leak the timer protected against has been independently fixed (link the PR), and demote this to a description note.
3. Explicitly retire the `npx @langwatch/server` path and document the retirement in this PR's body.

For long-uptime environments (npx-server, customer-managed self-hosted), removing the safety net without a replacement is a slow-burn regression that won't show up in CI but will show up as RSS-growth pages on production dashboards months down the road.

langwatch/src/server/experiments-v3/services/experiment-run.service.ts
```diff
@@ -1,112 +1,986 @@
...
+  }): Promise<ExperimentRunWithItems | null> {
+    return this.tracer.withActiveSpan(
+      "ExperimentRunService.getRun",
+      {
+        attributes: { "tenant.id": projectId, "run.id": runId },
+      },
+      async () => {
+        const clickHouseClient = await getClickHouseClientForProject(projectId);
+        if (!clickHouseClient) {
+          return null;
```
Aryansharma28, line 425: **[NON-BLOCKER]** `getRun` keeps `return null` on no-CH while the rest of the class now throws — inconsistent with the PR's stated rule and with the class-level JSDoc.

**Why:** The PR description explicitly states the rule: *"the three collapsed services changed their not-found-CH-client semantics from 'return `null`, let the caller fall back to ES' to 'throw — ClickHouse is the only backend now.'"* That holds for `listRuns` (throws at L119), `getRunAggregatesForExperimentIds` (throws at L397), and `listRunsForExperimentPaginated` (throws at L472). But this method still returns `null` here at L714.

The class-level JSDoc at L40-44 also still says *"Returns `null` from public methods when ClickHouse is not enabled for the given project."* — that's doc-rot relative to the new behavior.

Worse, the caller at `langwatch/src/routes/experiments-v3.ts:739` collapses both `null` (run not found) **and** `null` (CH unavailable) to a single `404 Run not found` response. So a ClickHouse outage looks identical to a missing run from the operator's point of view — infra failures get silently masked.

**Suggested fix (pick one):**
1. Match the rest of the class: throw when CH is not enabled, return a sentinel (or `undefined`) only for genuine not-found. Update the caller's 404 mapping to apply only to the not-found branch.
2. If the `null` is deliberate (e.g. you want a 404 even when CH is down), update the class-level JSDoc + method JSDoc to call out the exception explicitly, and add a one-line comment here explaining why this method differs from the other three.

Either is fine — the current state is the one that should not exist (inconsistent + undocumented + masks infra outages).

langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
```diff
@@ -264,6 +264,19 @@ export class EvaluationExecutionService {
       };
     }
 
+    // Enrich evaluations: getTracesWithSpans does not populate
+    // `trace.evaluations`, but evaluator field mappings that read the
+    // `evaluations` source need them. Fetch and attach before building the
+    // mapped data so they aren't silently empty (parity with
+    // runEvaluationForTrace in runEvaluation.ts).
+    const evaluationsByTrace =
+      await this.deps.traceService.getEvaluationsMultiple(
```
Aryansharma28: **[NON-BLOCKER]** Soft-degrade → hard-fail behavior change not flagged in PR description.

**Why:** `EvaluationService.getEvaluationsMultiple` (which `traceService.getEvaluationsMultiple` delegates to) now `throws` on no-CH-client per the PR's collapsed-facade rule. tRPC routers and Hono REST handlers wrap throws into 500s, which is fine. But here, and at `langwatch/src/server/evaluations/runEvaluation.ts:293`, the call is on the **worker** path with no try/catch around it. Pre-PR: a `null` return meant "no evals recorded" → worker proceeded with `trace.evaluations = []` and finished the eval job. Post-PR: a no-CH throw bubbles up and **fails the evaluation job**.

For a project where CH genuinely isn't available, hard-failing the job is arguably correct (there's no data to evaluate against). But this is a real worker-behavior change — soft-degrade to hard-fail — that isn't called out in the PR description's "Anything surprising?" section. Operators who relied on the old soft-degrade (e.g. transient CH unavailability during a deploy) will see new eval-job failures and have nothing to point at.

**Suggested fix:** add one sentence to the PR description's "Anything surprising?" section: *"Worker callers of the collapsed services (`runEvaluation.ts:293`, `evaluation-execution.service.ts:273`) lost their implicit null-→-soft-skip behavior. A transient CH unavailability now fails the evaluation job instead of producing an empty-evals result."*

If you want the old soft-degrade back, wrap this call in a try/catch and treat the throw as "no evals to enrich." Otherwise the description fix is enough.

Aryansharma28 (commented):
## Review summary — `goated-review` pass

**Verdict:** 1 blocker, 12 non-blockers. All net-new — every existing reviewer thread is already resolved.

Methodology: pulled the room first (existing reviewers, CodeRabbit/Copilot, CI), bucketed the 262 files by risk (lockfile, mass deletions, behavior changes, workers/queues, EE admin, docs), grilled the PR description's own "Anything surprising?" claims, adversarially verified each finding before posting, and deduped against existing reviews.

---

### 🛑 BLOCKER

**B1 — Worker memory-leak safety restart removed without replacement** (see inline at `src/workers.ts`)

Pre-PR `src/workers.ts` on `origin/main` had an env-gated 15-minute `setTimeout` that exited the worker, with a comment that explicitly named the case it protected: *"memory-leak safety pattern... In environments without that supervisor (e.g. `npx @langwatch/server`, where supervise() in spawn.ts doesn't restart-on-exit), set `LANGWATCH_WORKERS_MAX_RUNTIME_MS=0` to disable the timer."* The PR description argues the timer is fine to drop because "helm/docker already restart-on-exit," but that's exactly backward — those supervisors only fire on **exit**; the timer was what caused the periodic exit so that a slow leak couldn't grow unbounded.

The post-PR file has no `setTimeout`/`setInterval`/RSS check/liveness mechanism. For the `npx @langwatch/server` path the PR comment specifically called out, there's now nothing.

Fix options (any one): (a) preserve the timer behind the same env switch (default-on, ~5 lines); (b) prove the underlying leak has been independently fixed (link the PR); (c) explicitly retire the `npx @langwatch/server` path and document it in this PR's description.

---

### NON-BLOCKERS — code

**NB5 — `ExperimentRunService.getRun` inconsistency** (inline at `experiment-run.service.ts:714`)
**NB6 — Worker `getEvaluationsMultiple` callers now hard-fail** (inline at `evaluation-execution.service.ts:273`)

### NON-BLOCKERS — EE admin + Prisma footprint (no diff context, summary only)

**NB1 — PR description claim 1 is partially wrong: the EE admin layer still encrypts and writes the ES Prisma columns.**

The "Anything surprising?" section says "only the application layer that reads/writes them" was removed. In tracked code on this PR head, the three columns are still read/written end-to-end via the EE admin path:

- `langwatch/ee/admin/backoffice/resources/OrganizationsView.tsx:192-518` — full superadmin form for `useCustomElasticsearch` / `elasticsearchNodeUrl` / `elasticsearchApiKey`, with submit handler that posts these back to the org update.
- `langwatch/ee/admin/safeSelects.ts:29` — `useCustomElasticsearch: true` still in the org `select` shape.
- `langwatch/src/server/app-layer/organizations/repositories/organization.repository.ts:173-174` — repository interface still declares both fields optional in `UpdateOrganizationInput`.
- `langwatch/src/server/app-layer/organizations/repositories/organization.prisma.repository.ts:445-449` — `prisma.update` **still encrypts and writes** `elasticsearchNodeUrl` / `elasticsearchApiKey` when the caller passes them.
- `langwatch/src/factories/organization.factory.ts:17-19` — test factory still defaults all three.

User-facing settings page was correctly stripped; the EE admin path is intact. Fix: either amend the description (so a future drop-column migration scopes the EE admin too) or remove the EE write path now. None of these files are in the PR's diff, hence summary-only.

### NON-BLOCKERS — other code paths (no diff context)

**NB7 — Third unmentioned BullMQ importer.** `langwatch/bullboard/src/server.ts:17` imports `Queue` from `bullmq`. PR description's commit-2 list names 6 importers (puller×2, topic-clustering×2, queues×2); the actual set is 7 with bullboard. Worth adding so any future BullMQ removal scopes it.

**NB8 — Stale `vi.mock("ElasticsearchTraceService", ...)` in 2 unit tests.** `langwatch/src/server/traces/__tests__/trace-service-blob-resolution.unit.test.ts:54` and `trace-service-4888-full-flag.unit.test.ts:82`. Mock factory now silently no-ops since the module shape is gone. Either drop the mock or replace with the CH-only shape. (Only the `~/server/elasticsearch/protections` → `~/server/traces/protections` import is in the PR's diff for these files; the mock line at 54 / 82 is unchanged context, so no inline anchor.)

**NB9 — Stale references to deleted routes in docs + a file-header comment.**
- `langwatch/dev/docs/security/hono-api-rbac-audit.md:50-51` — lists `/api/rerun_checks` + `/api/start_workers` as live unauthenticated routes; both deleted.
- `langwatch/src/server/routes/misc.ts:11-12` — file-header comment still points to deleted `src/pages/api/rerun_checks.ts` / `start_workers.ts`.
- `langwatch/dev/docs/adr/006-redis-cluster-bullmq-hash-tags.md:28` and `adr/023-orphan-sweep-reactor-chain.md:175-181` — ADRs reference deleted `src/server/background/`. ADRs are historical, but a one-line "superseded by #4547" prevents future readers chasing dead paths.

**NB10 — `makeQueueName` import path inconsistency (bundle-protection nit).** `langwatch/src/server/event-sourcing/eventSourcing.ts:9` and `langwatch/ee/governance/services/pullers/pullerWorker.ts:35` still import `makeQueueName` via the `~/server/redis` re-export instead of the standalone `~/server/queues/makeQueueName` module. Server-only files, so no immediate bundle harm — but the inconsistency invites a future "copy this import" mistake that does pull `ioredis` into a client chunk. Route everything through the standalone module.

---

### NON-BLOCKERS — PR description fixes

**NB2 — Lockfile bumps paragraph is stale.** After rebase (`03be9869`), the `hono` / `qs` / `ws` / `mermaid` / `liquidjs` / `turbo` / `@opentelemetry/*` bumps the description warns about are gone (`git diff origin/main -- langwatch/pnpm-lock.yaml` is clean of those). Remove the paragraph.

**NB3 — Wrong env-var name in the reproduce note.** Description says `SKIP_HOST_REDIS_CHECK=1` to skip the host Redis check; the actual variable at `langwatch/src/server/redis.ts:27` is `SKIP_REDIS`. Reproducers will be confused.

**NB4 — `withJobContext` location wrong.** Commit-2 section says the helper moved to `ee/governance/services/pullers/withJobContext.ts`; the actual location is `langwatch/src/server/queues/withJobContext.ts` (the later corrected note mentions `src/server/queues/` but the wording could be tightened).

---

### NON-BLOCKER — CI infra (not this PR's defect)

**NB11 — The `evaluate` workflow can't review a PR this size.** The job's `gh pr diff` call returns HTTP 406 for diffs >20k lines, exiting the step with code 1 *before* the workflow's own `oversized=true` size guard runs. Pre-existing brittleness in `.github/workflows/evaluate.yml` (the call should be `gh pr diff ... > /tmp/pr_diff.txt 2>/dev/null || true` so the size check can decide). Worth a separate one-line PR against the workflow.

---

### What I could NOT verify (worth flagging)

1. Whether the underlying memory-leak the 15-min timer protected against has been *independently* fixed in another PR. If yes, B1 demotes to a description nit ("timer removed because leak X was fixed in #YYYY"). If no, B1 stays as a regression risk for long-uptime envs.
2. Whether `/cron/schedule_topic_clustering` is still triggered post-PR end-to-end (handler is kept, but I didn't trace the cron-scheduler → handler chain to confirm it fires).

---

### Existing reviewer findings I am NOT re-raising (already resolved)

- @rogeriochaves's P1 (lockfile importer / `hasown@2.0.4` / `tinyexec@1.2.3`) — resolved in `03be9869`.
- @rogeriochaves's P2 (topic-clustering left in settings sidebar after route deletion) — resolved in `03be9869` (full topic-clustering restore: route, MenuLink, command-bar, worker).
- @rogeriochaves's P2 (stale `~/server/background/...` mock paths in three tests) — resolved in `03be9869`.
- `github-code-quality[bot]` ×5 (unused `CostReferenceType`/`CostType`/`nanoid`/`getProtectionsForProject`/`captureException`/`withScope`/`logger` in `runEvaluation.ts`; unused `prisma` in `httpProxyTracing.ts`) — resolved in `54cea3de` + `10cdf4cb`.

---

*Reviewed with the [`goated-review`](https://github.com/Aryansharma28/skills-aryan/blob/main/goated-review/SKILL.md) skill: read-the-room → slice-by-risk → find → adversarial-verify → dedupe → classify → show-the-plan → post.*

langwatch/src/workers.ts
0xdeafcafe: no longer needed, workers are safe now

langwatch/src/server/experiments-v3/services/experiment-run.service.ts
0xdeafcafe: Good catch — `getRun`'s null-on-no-CH is deliberate (PR #3483 dogfood, locked in `getRun-null.unit.test.ts`: the event-sourcing fold race during a freshly-started run was wrapping the brief pre-projection window in a 500 cascade in the UI poller, which hid more severe pipeline-stuck cases). I went with your option 2: documented in `b5e34fd4b` — class JSDoc now names `getRun` as the documented exception, the method JSDoc explains the rationale + that the route's collapsed `null → 404` is intentional polling UX (not infra masking), and a one-line inline comment sits next to the no-CH guard. The 404 mapping at `routes/experiments-v3.ts:739` stays as-is. The other three methods continue to throw.

langwatch/src/server/app-layer/evaluations/evaluation-execution.service.ts
0xdeafcafe: Good catch. Per your suggested fix #1, added a bullet to "Anything surprising?" in the PR description spelling out the worker-semantics change: pre-PR null-return meant "no evals recorded" → worker proceeded with `[]`; post-PR a no-CH throw bubbles up and fails the eval job. Keeping the new hard-fail behavior — there's no useful eval against a missing backend — but it's now called out so operators investigating new eval-job failures during transient CH unavailability have something to point at.

langwatch/src/workers.ts
```diff
@@ -14,48 +16,157 @@ const logger = createLogger("langwatch:workers");
...
-void verifyRedisReady().then(() => {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  require("./server/background/worker")
-    .start(void 0, maxRuntimeMs)
-    .catch((error: Error) => {
-      if (error instanceof WorkersRestart) {
-        logger.info({ error }, "worker restart");
-        process.exit(0);
-      }
+async function gracefulShutdown(): Promise<void> {
```
Aryansharma28, line 31: [P1] Re-raising my prior BLOCKER: the removed memory-leak self-restart has no replacement.

**Why:** `main`'s `src/workers.ts` had a documented 15-min self-restart specifically for `npx @langwatch/server` (no external process supervisor to restart-on-exit): "memory-leak safety pattern... In environments without that supervisor... set `LANGWATCH_WORKERS_MAX_RUNTIME_MS=0` to disable the timer." This PR deletes it entirely with no replacement. The reply to my original finding was "no longer needed, workers are safe now" with no supporting code change. I re-grepped the entire PR branch (`maxRuntimeMs`, `MAX_RUNTIME`, `setTimeout`/`setInterval` restart, RSS/memory threshold checks) — nothing exists anywhere as a replacement.

For helm/docker-supervised deployments this is partially mitigated by container-level OOMKill + restart-policy, but that's reactive (crash-then-restart after already ballooning memory) versus the old proactive graceful-restart-before-that-happens. For the `npx @langwatch/server` path the original comment explicitly called out, there is now no safety net at all.

**Suggested fix:** either restore a periodic self-exit (env-gated, defaulting off for supervised deployments if that's the intent) or provide an actual replacement mechanism (RSS-threshold check, liveness probe wired to a real drain-and-exit) — not just an assertion that it's unnecessary.

langwatch/src/server/routes/collector.ts
```diff
@@ -546,10 +520,13 @@ secured
...
-        const ingestion = getApp().traces.collection;
-        const results = await Promise.all(
+        const results = await Promise.allSettled(
           spans.map((span) =>
-            ingestion.ingestNormalizedSpan({
+            // Route through ingestNormalizedSpan (not recordSpan directly) so the
+            // REST collector shares the (tenant, trace, span) dedup gate + ADR-022
+            // spool hook with the OTLP path — a retry storm here must not bypass
+            // dedup. occurredAt is stamped inside ingestNormalizedSpan.
+            getApp().traces.collection.ingestNormalizedSpan({
```
Aryansharma28, line 580: [P2] REST collector dedup semantics changed materially, not just relocated — worth calling out explicitly and testing.

**Why:** The PR description frames this as "the REST path now shares the `ingestNormalizedSpan` dedup gate with the OTLP path," which reads as a pure relocation. It isn't:
- **Old** (`main`'s `fetchExistingMD5s` + 256-version cap): content-hash (MD5 of full request params) compared against a value persisted indefinitely in ES. No time bound — only a 256-entry cap.
- **New** (`SpanDedupService` in `src/server/app-layer/traces/span-dedupe.service.ts`): Redis `SET NX` keyed on `(tenantId, traceId, spanId)` only — pure ID-based, not content-based — with a 60s processing TTL extended to 3600s (1hr) on success.

Concrete behavior change: a retry/replay/backfill tool resending the same span content more than 1 hour after original ingest is no longer deduped (previously caught by the unbounded MD5 check). Conversely, two requests with the same span ID but *different* content are now silently deduped as identical, where the old MD5 check would have detected the mismatch and reprocessed.

Also: dedup is unit-tested only against a mocked `SpanDedupService` (`trace-request-collection.service.unit.test.ts:95-119`); no test POSTs duplicate span content through the actual `/api/collector` REST route (`collector.stress.test.ts` has zero dedup/MD5 references).

**Suggested fix:** add a line to the PR description's "Anything surprising?" section naming this as a real semantics change (window went from unbounded to 1hr, content-hash to ID-based), and add one integration test that POSTs the same span twice through `/api/collector` and asserts single-insert.

Aryansharma28 (commented):
## Review summary — goated-review follow-up pass

**Verdict:** COMMENT. P0: 0 | P1: 1 | P2: 1 | P3: 0 | P4: 1 | P5: 2

This is a re-verification of my prior review pass (2026-06-24) plus a fresh sweep. Methodology: re-checked every previously-flagged finding against current code (not against the author's reply text), ran two independent audits (collector dedup semantics, throw-vs-null call-site safety across all 3 collapsed facades — 13 production call sites checked), and grepped the whole branch for orphaned references to every deleted route/file/package.

### What's genuinely fixed since the last pass
- Lockfile — clean, identical to `main`.
- Topic-clustering route/sidebar entry — restored correctly.
- PII redaction test mocks — updated to the new module paths.
- `ExperimentRunService.getRun` null-vs-throw exception — properly documented with a regression test backing the rationale.
- Facade throw-vs-null semantics — audited every caller of every newly-throwing method on `EvaluationService`, `ExperimentRunService`, `FilterService`. Zero new at-risk workers/crons beyond the two the PR description already discloses.
- Orphaned references to every deleted route/file/package (`start_workers`, `rerun_checks`, `scenario_analytics`, `traces_retention_period_cleanup`, `SimulationFacade`, `@elastic/elasticsearch`, helm/CI configs) — clean, nothing dangling except one harmless stale doc-comment (P4 below).

### 🟠 P1 — re-raised, not actually addressed
See inline comment on `langwatch/src/workers.ts`. My prior BLOCKER finding on the removed memory-leak self-restart got a reply ("no longer needed, workers are safe now") but no code change. I re-grepped the whole PR branch for any replacement liveness/restart mechanism — none exists.

### 🟡 P2 — new finding this pass
See inline comment on `langwatch/src/server/routes/collector.ts`. The REST collector's dedup mechanism changed semantics, not just location, and that's understated in the PR description.

### P4 — nit
`langwatch/src/server/routes/misc.ts` lines 11-12 (https://github.com/langwatch/langwatch/blob/fced137194c0b49a266684112efbcd00e2862d8c/langwatch/src/server/routes/misc.ts#L11-L12) — the `Replaces:` JSDoc list still names `src/pages/api/rerun_checks.ts` and `src/pages/api/start_workers.ts`, both deleted by this PR. Harmless (doc-only, and outside the diff's hunk window so it can't be anchored as an inline comment), but misleading to a future reader tracing where those routes went. Suggested fix: drop those two lines from the list.

### P5 — praise / observations (non-blocking)
- **Praise:** the collapsed-facade throw/null rule is applied consistently. I traced all 13 production call sites across the 3 collapsed services and found zero surprises beyond what the PR description already calls out — that's a clean, disciplined refactor of a genuinely risky semantic change.
- **Observation:** the `evaluate` CI check is failing, but it's the automated-eval bot hitting the same 20k-line GitHub diff API cap that CodeRabbit and Copilot both hit on this PR (`could not find pull request diff: HTTP 406... too_large`). Not a defect in this PR — `alls-green`, the actual required aggregate gate, passes. Flagging for awareness only: no automated tool has been able to review this diff end-to-end; human review is the only real gate here.

### Existing findings NOT re-raised (verified fixed, not just claimed)
- @rogeriochaves' lockfile P1 — confirmed clean vs `main`.
- @rogeriochaves' topic-clustering route P2 — confirmed restored.
- @rogeriochaves' PII test mock P2 — confirmed updated.
- My own prior `getRun` null-vs-throw non-blocker — confirmed properly resolved with a regression test.

- Whether SaaS-deployed workers (k8s/helm) actually get restarted promptly on OOM via the platform's restart policy — I'm inferring this from general k8s behavior, not from this repo's helm chart. Worth confirming with whoever owns the workers deployment manifest before treating the P1 as low-risk for the hosted product; it's not low-risk for the `npx @langwatch/server` self-hosted path either way.
- Actual production traffic patterns for delayed/replayed span ingestion (backfills, SDK retry queues) that would hit the new 1-hour dedup TTL — I can only confirm the mechanism changed, not how often the old unbounded window was actually relied upon in practice.

langwatch/src/workers.ts
0xdeafcafe: You're right the last reply was hand-wavy — apologies. Concrete reasoning below, and I've expanded the PR body's "Anything surprising?" bullet so the intent is on record.

The 15-min timer was a memory-leak safety net for the BullMQ collector and evaluations workers specifically: those held long-lived references to enqueued job payloads and could accumulate RSS across their processing loops. This PR removes both worker classes entirely — so the leak profile that motivated the timer is gone. The two BullMQ queues that remain are bounded: topic-clustering runs infrequently and doesn't retain payloads across cycles; the ingestion puller streams and doesn't accumulate.

For helm/docker-supervised deployments: container-level OOMKill + restart-policy is first-class, so the reactive path is fine (agreed proactive-vs-reactive is a real distinction).

For the `npx @langwatch/server` path you specifically called out: the intended stance is that if a real leak surfaces there we hunt it and fix the leak, not add back a periodic-restart hammer that would also mask the underlying cause. If a concrete leak turns up on the npx path post-merge, I'll take the fix — but adding an RSS-threshold check or a re-implemented timer speculatively (with the BullMQ workers gone) would be adding complexity for a hypothetical.

langwatch/src/server/routes/collector.ts
0xdeafcafe: Fair, and the semantics description is accurate. Addressed the transparency ask: added a bullet to the PR body's "Anything surprising?" section naming the change explicitly — that the window went from unbounded (with a 256-entry cap) to a 1h TTL, and that the key changed from content-hash to `(tenant, trace, span)`, with both directional consequences (retry >1h no longer deduped; same-id-different-content silently deduped).

On the integration test: the service-level gate is unit-tested at `trace-request-collection.service.unit.test.ts:95-119` — asserts `status: 'deduped'` and no `recordSpan` dispatch when the lock is already held. A REST-through-route test would exercise the wiring (route → `ingestNormalizedSpan` → dedup), which is meaningful but re-asserts what the service unit test already covers. My call is to defer the route-level test as a follow-up — noted in the PR body as such — rather than block this merge on it. Happy to reopen if you think the wiring itself needs the integration backstop before merge.

langwatch-agent:

**Review mode:** Deep Review (near Reject Shape)
**Scariest prod risk:** Behavioral changes buried inside a 100-file, 5k-line deletion PR. The legacy collector dedup-gate removal (`fetchExistingMD5s` + version cap) and the `httpProxyTracing` switch to `recordSpan` change live ingestion/dedup behavior, not just delete dead code, so a real regression can hide in a diff that reads as "removal".

**Inspection targets:**
- `[Must Check]` `routes/collector.ts` — dropping the REST dedup gate so it shares the OTLP `ingestNormalizedSpan` path; verify REST re-ingestion is still deduped and produces no span duplication or loss.
- `[Must Check]` `src/workers.ts` rewrite — dropped 15-min self-restart, BullMQ worker startups replaced by the EE puller + restored topic-clustering worker; confirm every previously-running worker role still boots.
- `[Ask Author]` The diff exceeds 20k lines / 100 files and mixes mechanical ES/BullMQ deletion with behavioral rewrites; can the behavioral pieces (collector dedup, httpProxyTracing, workers.ts) be split or called out for isolated review?

**Author questions:**
- Which environments still provision Elasticsearch, and is there a rollback path if any tenant's read/write still resolves to an ES backend after this lands?

sergioestebance:
Blessed. 🙏

_"Every branch in me that beareth not fruit he taketh away… that it may bring forth more fruit."_ — John 15:2

−29.8K LOC of dead ES + legacy BullMQ, pruned clean: retention cron was ES-only (live CH retention untouched), worker boot complete, no dangling refs. The red `evaluate` check is just the auto-approve risk-gate declining a 263-file PR, not a real failure — app CI is green. Mind the P2: REST `/api/collector` dedup now matches OTLP (content-diff spans dedup within 1h) and still needs a route-level integration test.

sergioestebance (dismissed):

rogeriochaves (approved):

Aryansharma28 (approved):
i trust you bro

PR #1237, opened 2026-01-28
chore(main): release langwatch 2.0.1

0xdeafcafe:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* **licensing:** add production public key for license validation ([#1234](https://github.com/langwatch/langwatch/issues/1234)) ([1aba398](https://github.com/langwatch/langwatch/commit/1aba3983ae1dfdf31bef10ca13271714f2cebcd3))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #4569, opened 2026-06-05
fix(deps): bump @vitest/browser-playwright to >=4.1.7 (closes CRITICAL #1237)

langwatch-agent:
## Why

CRITICAL alert #1237 (GHSA-2h32-95rg-cppp, `@vitest/browser`, the Vitest UI server arbitrary file read and execute issue) is still open even though #4474 bumped the direct `@vitest/browser` dependency to a fixed version.

The reason is a transitive copy. `@vitest/browser-playwright` was pinned at `^4.1.5`, and `@vitest/browser-playwright@4.1.5` depends on `@vitest/browser@4.1.5`, which is `< 4.1.6` and therefore still vulnerable. So the lockfile carried both a fixed direct `@vitest/browser` (4.1.7) and a vulnerable transitive `@vitest/browser@4.1.5` pulled in through the Playwright provider.

## What changed

Bumps `@vitest/browser-playwright` from `^4.1.5` to `^4.1.7`. `@vitest/browser-playwright@4.1.7` depends on `@vitest/browser@4.1.7`, so the transitive `4.1.5` copy is gone. After regenerating the lockfile, every resolved `@vitest/browser` is `>= 4.1.6` (4.1.7 and 4.1.8), so alert #1237 closes.

The diff is the one-line manifest bump plus the lockfile, and the lockfile churn is confined to the vitest / playwright subtree only. No pnpm overrides were touched.

## How this was verified

- Regenerated `langwatch/pnpm-lock.yaml`: no `@vitest/browser@4.1.5` remains; all resolved copies are 4.1.7 / 4.1.8 (both `>= 4.1.6`).
- Frozen-lockfile install succeeds, resolving `@vitest/browser-playwright@4.1.7` and `@vitest/browser@4.1.7`.
- This is a patch-level bump of a dev-only browser test runner (`@vitest/browser-playwright`, 4.1.5 to 4.1.7); no app or test code changes are required. The browser test suites that exercise this runner run in this PR's CI (they need browser binaries that are not available locally), so CI is the authoritative signal for the test execution.

Note: regenerating the lockfile used a temporary local `--config.minimumReleaseAge=0` to re-resolve under the repo's supply-chain cooldown. That flag is local-only and is not committed.

### DIFF
--- a/langwatch/package.json
+++ b/langwatch/package.json
@@ -292,7 +292,7 @@
     "@types/ws": "^8.18.1",
     "@typescript/native-preview": "7.0.0-dev.20260415.1",
     "@vitest/browser": "^4.1.7",
-    "@vitest/browser-playwright": "^4.1.5",
+    "@vitest/browser-playwright": "^4.1.7",
     "@vitest/coverage-v8": "^4.1.5",
     "@vitest/ui": "^4.1.5",
     "babel-plugin-react-compiler": "^1.0.0",
--- a/langwatch/pnpm-lock.yaml
+++ b/langwatch/pnpm-lock.yaml
@@ -737,8 +737,8 @@ importers:
         specifier: ^4.1.7
         version: 4.1.7(bufferutil@4.1.0)(vite@8.0.10(@types/node@18.19.130)(esbuild@0.28.0)(sass@1.99.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.8.3))(vitest@4.1.4)
       '@vitest/browser-playwright':
-        specifier: ^4.1.5
-        version: 4.1.5(bufferutil@4.1.0)(playwright@1.59.1)(vite@8.0.10(@types/node@18.19.130)(esbuild@0.28.0)(sass@1.99.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.8.3))(vitest@4.1.4)
+        specifier: ^4.1.7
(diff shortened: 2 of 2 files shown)


sergioestebance:
do we actually need @vitest/browser-playwright?

langwatch-agent:
Good question. Short answer: yes, it is currently used, but it is worth knowing exactly for what.

`vitest.browser.config.ts` imports the provider from it directly:

```ts
import { playwright } from "@vitest/browser-playwright";
// ...
browser: { provider: playwright(), instances: [{ browser: "chromium" }] }
```

That config backs the browser-mode suite, the four `*.browser.test.{ts,tsx}` files (BatchTargetCell, SearchBar, ActiveSearchEditor, and trace), run via `pnpm test:browser`. In vitest 3 the provider was built in and selected with the string `provider: "playwright"`; vitest 4 split each provider into its own package, which is why it now appears as an explicit dependency rather than being bundled.

One honest caveat I hit while checking: that browser suite is not wired into any CI workflow today, so the package only serves the local `pnpm test:browser` run plus those four tests. So the real options are:

- Keep the browser-mode tests: then we need `@vitest/browser-playwright` (and `@vitest/browser`), and this PR is the minimal bump to get the transitive copy past the CRITICAL advisory.
- Drop the browser-mode tests (or move those four over to jsdom): then both `@vitest/browser-playwright` and `@vitest/browser` can be removed entirely, and #1237 becomes moot.

I scoped this PR to just the security fix and left the test setup alone, since removing a testing capability feels like a separate call from a CVE patch. If you would rather go the removal route, I am happy to open that as its own PR instead. Either path closes the alert.

rogeriochaves (approved):

PR #4623, opened 2026-06-05
feat(batch-evaluation-results): fields + row height controls for the results table

Aryansharma28:
## What & why

The results table always renders the full target cell — model **output**, evaluator **score chips**, and now **cost/latency** — stacked together, with no way to dial the detail level up or down. That's a real problem the moment a run has more than 2-3 targets: with 5 targets side by side, a support-quality N-way run renders as five columns of dense paragraph text plus chips, and the thing you're actually trying to do — scan scores, or read one output in full — gets buried in the others'. This PR was a straight port of that idea, shipped as a single "View" button with three presets (`Everything`/`Data only`/`Scores only`) plus two checkboxes underneath.

**That version had two real problems**, caught in review before merge:
1. **An unnamed state.** Toggling the checkboxes independently of the presets could land on a combination no preset named, and the button's own label degraded to a generic `"View"` — a control that can reach a state it can't name is the wrong model.
2. **It solved the wrong half of the problem.** Hiding sections is *destructive* — you lose the score chips entirely to make room. The actual complaint ("I can't read this without scrolling forever") is a **density** problem, not a **visibility** problem, and this PR had no answer for that at all.

The fix is to split "what shows" from "how much fits" into two independent controls, instead of one control trying to do both:

- **`Fields`** — plain, independent checkboxes: `Outputs`, `Scores`, `Latency and cost`. No presets, no derived label — the button always says "Fields." Deliberately **not** persisted: a section left off from a previous visit should never silently explain "no results" on a *different* run.
- **`Row height`** — `Small` / `Medium` / `Large`, controlling how much of each cell's content shows before it needs expanding. This is the non-destructive fix for the actual density complaint: nothing is hidden, more just fits, or you get more room to read one thing in full. **Persisted** — density is a "how I like to read tables" preference, not something tied to one run.
- **Cost is now shown per-row**, next to latency (previously latency-only, despite `BatchTargetOutput.cost` already existing in the data model and being aggregated in the column header as "Total Cost"). Bundled into one "Latency and cost" toggle, since they're both resource metrics a reader wants together.

## Screenshots

Captured against a real 5-target comparison run (`support-nway-showdown`, 1,620 result rows) on a local stack.

**Before / after, side by side** — one button that could reach a state it couldn't name, replaced by two that can't:

| Before — shipped first | After — reworked to this |
|---|---|
| old header | new header |
| `Everything` — three presets + two checkboxes; the label falls back to a bare `"View"` the moment the checkboxes disagree with every preset | `Row height` + `Fields` — two controls, two jobs, neither can degrade into a state it can't describe |

**Fields** — flat checkboxes, no presets, stable label:

fields open

**Row height** — three named tiers:

row height open

**Same run, same viewport, two tiers** — `Small` fits more rows on screen at once; `Large` gives one response room to read start to finish without opening the expanded view. Neither hides a single evaluator chip:

| `Small` — 57 chips on screen | `Large` — 49 chips on screen |
|---|---|
| small | large |

**After reload** — row height (`Large`) survived; a field hidden before the reload (`Outputs`) came back:

persistence

## Verification

DOM/virtualizer state, not just eyeballing the images — the chip counts prove the virtualizer is actually re-measuring rows at each tier, not just applying a CSS clip:

| State | Chips rendered | Cost/latency els | Table columns | Row-height tier |
|---|---|---|---|---|
| Default | 53 | 52 | 8 | `m` |
| Latency/cost hidden | 53 (unchanged) | **0** | 8 | `m` |
| Row height → Small | **57** (more rows fit) | 56 | 8 | `s` |
| Row height → Large | **49** (fewer rows fit) | 48 | 8 | `l` |
| Outputs off (still Large) | 53 (row shrinks w/o output text) | 52 | 8 | `l` |
| After reload | 49 | 48 | 8 | `l` (persisted) |
| — Fields reset | — | — | — | outputs back on |

The Small→Large swing in chip count (57→49) is the virtualizer genuinely re-measuring row heights per tier, not a cosmetic-only change.

## Implementation

- New `rowHeight.ts`-style constants in `tableUtils.ts`: `RowHeight` type, `COLLAPSED_CELL_HEIGHT_PX` (shared by `BatchTargetCell` and `ExpandableDatasetCell` so a row's cells all clip at the same height — previously these were two different hardcoded values, 120px and 180px, with no explanation for the mismatch), and `ESTIMATED_ROW_HEIGHT_PX` (keeps the virtualizer's first-paint guess close to the real height per tier).
- `showOutputs`/`showEvaluations` (kept, minimal churn) plus a new `showCostAndLatency`, threaded through `BatchEvaluationResultsTable` → `SingleRunTable`/`ComparisonTable` → `BatchTargetCell`. The action-bar gate is split: cost/latency renders independently of the output toggle (previously bundled — hiding output also silently hid latency, unrelated to the actual toggle you touched).
- `buildColumns`/`buildComparisonColumns` converted from 9 positional args to a single options object — CI's Biome delta-gate enforces a 3-param max, and I was about to push both over it anyway.

## Tests / spec

- Spec section renamed `Fields / Row Height`; 3 scenarios reworded from preset names to field names, 4 new scenarios added (row height, and that each of fields/row-height does or doesn't survive a reload) — `check:feature-parity` reports **9/9 bound**.
- 191/191 tests pass across the directory. New coverage: cost display, the cost/latency-independent-of-output toggle, row-height threading (`data-row-height` attribute), all in both single-run and comparison modes.
- The two reload-persistence scenarios were initially only verified by hand in the browser (screenshots above) — moved into `useResultDisplayPreferences`, a small extracted hook, and bound to `renderHook` tests against jsdom's real `localStorage`, so they're enforced in CI, not just observed once.
- Mutation-verified throughout: forcing `showTargetColumns = true` in `ComparisonTable` fails exactly the "removes target column" test; forcing `fields` onto `useLocalStorage` fails exactly the "resets to defaults" test; forcing `rowHeight` onto plain `useState` fails exactly the "keeps the choice" test. Each mutation fails only its own test.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/langwatch/src/components/batch-evaluation-results/BatchEvaluationResults.tsx
+++ b/langwatch/src/components/batch-evaluation-results/BatchEvaluationResults.tsx
@@ -20,6 +20,7 @@ import {
 import { type Experiment, ExperimentType, type Project } from "@prisma/client";
 import type React from "react";
 import { useCallback, useEffect, useMemo, useState } from "react";
+import { useLocalStorage } from "usehooks-ts";
 import { BarChart2, Download, ExternalLink } from "react-feather";
 
 import { Link } from "~/components/ui/link";
@@ -31,6 +32,9 @@ import {
   BatchEvaluationResultsTable,
   ColumnVisibilityButton,
   DEFAULT_HIDDEN_COLUMNS,
+  DEFAULT_VIEW_SECTIONS,
+  ViewLensButton,
+  type ViewSections,
 } from "./BatchEvaluationResultsTable";
 import { type BatchRunSummary, BatchRunsSidebar } from "./BatchRunsSidebar";
 import { ComparisonCharts } from "./ComparisonCharts";
@@ -77,6 +81,12 @@ export function BatchEvaluationResults({
   // Column visibility state - initialize with defaults
   const [hiddenColumns, setHiddenColumns] = useState<Set<string>>(
     () => new Set(DEFAULT_HIDDEN_COLUMNS),
+  );
+
+  // View lens state - which result sections to render (persisted across sessions)
+  const [viewSections, setViewSections] = useLocalStorage<ViewSections>(
+    "batch-results-view-sections",
+    DEFAULT_VIEW_SECTIONS,
   );
 
   // Toggle column visibility
@@ -498,6 +508,12 @@ export function BatchEvaluationResults({
               Charts
             </Button>
           )}
+          {transformedData && transformedData.targetColumns.length > 0 && (
+            <ViewLensButton
+              sections={viewSections}
+              onChange={setViewSections}
+            />
+          )}
           {transformedData && transformedData.datasetColumns.length > 0 && (
             <ColumnVisibilityButton
               datasetColumns={transformedData.datasetColumns}
@@ -576,6 +592,8 @@ export function BatchEvaluationResults({
                   onToggleColumn={toggleColumn}
                   comparisonData={comparisonData}
                   targetColors={targetColors}
+                  showOutputs={viewSections.outputs}
+                  showEvaluations={viewSections.evaluations}
                 />
               </Card.Body>
             </Card.Root>
--- a/langwatch/src/components/batch-evaluation-results/BatchEvaluationResults.tsx
+++ b/langwatch/src/components/batch-evaluation-results/BatchEvaluationResults.tsx
@@ -20,8 +20,8 @@ import {
 import { type Experiment, ExperimentType, type Project } from "@prisma/client";
 import type React from "react";
 import { useCallback, useEffect, useMemo, useState } from "react";
-import { useLocalStorage } from "usehooks-ts";
 import { BarChart2, Download, ExternalLink } from "react-feather";
+import { useLocalStorage } from "usehooks-ts";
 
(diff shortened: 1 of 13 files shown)


langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** none declared. Clear UX rationale: with 5 targets side by side the results table renders as five columns of dense text, and there is no way to dial detail up or down.

**Note:** opened a while ago (#4623) against a table that has changed since. Confirm it still applies cleanly and reflects the current results-table shape rather than the one it was written against.

**Top inspection targets**
1. `Must Check` - persistence of the field and row-height selection. If it is stored per user, confirm where; if it is not persisted, confirm the default is the current behaviour so no existing user's view changes without them asking.
2. `Must Check` - rendering cost at the wide end. This exists because N-way runs are heavy; confirm the new controls do not add per-cell work that makes the 5-target case slower before the user narrows it.
3. `Probably Fine` - the control UI itself.

**Production risk:** none beyond a UI regression in the results table.

1. Is the default view unchanged for existing users?
2. Where does the preference persist, and is it per project or per user?

sergioestebance (approved):

PR #4641, opened 2026-06-06
perf(clickhouse): prune evaluation_runs trace fetch off SELECT *

langwatch-agent:
## Evidence

In the last 24h app-side slow-query log, the dominant **uncovered** evaluation_runs read is the trace-detail evaluations fetch (`fetchEvaluationRows` in `clickhouse-trace.service.ts`):

```
SELECT * FROM evaluation_runs WHERE TenantId = {…} AND TraceId IN ({…}) AND (TenantId, EvaluationId, UpdatedAt) IN (… max(UpdatedAt) …)
```

~390 warns/24h, p50 ~290ms. (The per-trace evaluation service uses an explicit column list already; this trace-service path was the missed `SELECT *`.)

## Suspected cause

`SELECT *` reads every column of `evaluation_runs` for the matched rows. The consumer (`ClickHouseEvaluationRunRow`) only uses 21 columns; `SELECT *` additionally pulls `ErrorDetails`, `CostId`, `ArchivedAt` and `CreatedAt`. `ErrorDetails` in particular is an unbounded error/JSON blob. On this deduped read path the `(TenantId, EvaluationId, UpdatedAt) IN (max-subquery)` predicate materializes those columns across **every version** of each evaluation before the IN filter discards the stale ones.

## Proposed fix

Select the explicit column set the mapper actually reads instead of `SELECT *`. The canonical column list moves next to its row type in `evaluation-run.mappers.ts` (single source of truth, `EVALUATION_RUN_COLUMNS_WITH_INPUTS`) and is reused by both the per-trace evaluation service (which previously had its own copy) and the trace service — so there is now one list, not three. The repo already enforces "no `SELECT *` on evaluation_runs" for the analytics subquery (`analytics/clickhouse/__tests__/column-pruning.test.ts`); this extends the same discipline to the trace-detail path.

## Expected impact

- Stops reading `ErrorDetails`/`CostId`/`ArchivedAt`/`CreatedAt` (across all versions) for every trace-detail evaluations fetch; read bytes drop, most when `ErrorDetails` is large.
- Results unchanged: the explicit list is exactly the columns the mapper consumes, asserted in sync with the row type by a test.

## EXPLAIN check

This is a column-projection change; it does not alter partition/granule pruning, so `EXPLAIN INDEXES` granule counts are identical for both shapes (the win is bytes read per granule, which the read-only EXPLAIN endpoint cannot quantify — ANALYZE is blocked). `EXPLAIN SYNTAX` confirms the new shape selects the 21 explicit columns and is otherwise identical.

## Test plan

- `evaluation-run.mappers.test.ts`: asserts `EVALUATION_RUN_COLUMNS_WITH_INPUTS` lists exactly the keys of `ClickHouseEvaluationRunRow` (stays in sync with the type) and excludes the unused columns. **25/25 pass.**
- Existing `clickhouse-evaluation.fallback`/`inputs` unit tests (which now consume the shared constants) still pass: **29/29.**
- `tsgo` typecheck clean for the changed files.

Advisory PR from the ClickHouse slow-query optimizer. A human should review and ship.

### DIFF
--- a/langwatch/src/server/evaluations/__tests__/evaluation-run.mappers.test.ts
+++ b/langwatch/src/server/evaluations/__tests__/evaluation-run.mappers.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import type { Evaluation } from "~/server/tracer/types";
 import type { ClickHouseEvaluationRunRow } from "../evaluation-run.mappers";
 import {
+  EVALUATION_RUN_COLUMNS_WITH_INPUTS,
   mapClickHouseEvaluationToTraceEvaluation,
   mapEsEvaluationToTraceEvaluation,
   mapTraceEvaluationsToLegacyEvaluations,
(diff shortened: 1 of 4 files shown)


rogeriochaves (dismissed):

langwatch-agent:
Refreshed onto current main (force-push). This was conflicting only because clickhouse-trace.service.ts was heavily refactored since the branch was cut (the eval read moved from ~L1555 to fetchEvaluationRows ~L2091), not because it was superseded: the `SELECT * FROM evaluation_runs` it prunes is still present on main. Re-applied the same change against the new structure, verified the column list still matches ClickHouseEvaluationRunRow exactly (mapper in-sync test), and reused the exported constant in clickhouse-evaluation.service. 136 evaluations unit tests pass.

langwatch-agent:
@coderabbitai review

rogeriochaves (approved):

PR #4642, opened 2026-06-06
fix(evals-v3): coerce non-string evaluator inputs + lock auto-mapping

rogeriochaves:
## Summary

Two related evaluations-v3 bugs from a workbench customer report:

1. Live "Run on row" rejected non-string mapped values with `Validation error: Expected string, received boolean at "output"` whenever an upstream target output (bool/number/object) was piped into a downstream string-input evaluator. The batch + online paths had quietly coerced for years via `tracesMapping.tryAndConvertTo`; the REST route never did.
2. `inferEvaluatorMappings` happily mapped an evaluator's `output` identifier to `dataset.output`, producing a tautological default that was more confusing than useful.

The fix lands across two lanes on one branch.

### Lane A (89583bcbc) — non-string evaluator inputs no longer get rejected

Three layers, deterministic chain:

- **REST live-execute**: new `coerceEvaluatorScalar` (`langwatch/src/server/utils/coerceEvaluatorScalar.ts`) wired through `z.preprocess` on every scalar field of `defaultEvaluatorInputSchema`. Semantics match the python NLP era `autoparse_field_value` for `FieldType.str`: `true → "true"`, `42 → "42"`, `{a:1} → '{"a":1}'`, null/undefined preserved, strings pass through.
- **nlpgo evaluator-block executor** (`services/nlpgo/app/engine/blocks/evaluatorblock/`): new `coerceData` helper applied at the Studio workflow dispatch boundary. Defense-in-depth + parity with the legacy python autoparse the Go migration had dropped. Contexts arrays pass through unchanged; conversation turns are walked.
- **langevals exact_match scorer** (`langevals/evaluators/langevals/langevals_langevals/exact_match.py`): a JavaScript-style loose-equality layer inserted between the existing `is_float` short-circuit and the trim/punctuation/case-fold chain. `json.loads` recovers each side's underlying type, then ECMA 7.2.13 abstract equality compares bool/number/string across the trio (`True → 1`, `False → 0`, numeric strings cross-coerce). So an upstream `passed: true` matches a dataset golden of `"1"` without false negatives. Free-form text still falls through to the existing transform chain.

The "match boolean against 1" semantic intentionally lives in the scorer, not in the coercion layer. Coercion preserves fidelity; semantic match is a scorer decision.

### Lane B (74cb22b52) — auto-mapping is now identity-locked

`inferEvaluatorMappings` heuristic restricted so:
- `output` identifiers ALWAYS source from a runner/target output, never the dataset.
- `expected_output` / `input` identifiers ALWAYS source from the dataset, never a runner.
- Empty is better than wrong: if no match exists, leave the mapping blank.

### Specs

- `specs/experiments-v3/evaluator-as-target.feature` — REST coercion regression + Scenario Outline per source type + null preservation + online-eval boolean metadata
- `specs/experiments-v3/mapping-auto-inference.feature` — output-from-target-only, expected/input-from-dataset-only, no cross-fall
- `specs/evaluators/exact-match-loose-equality.feature` — bool/number/string matches + mismatches + existing transform-chain preservation
- `specs/studio/workflow-evaluator-input-coercion.feature` — nlpgo workflow executor parity restoration

### Coverage

- `langwatch/src/server/utils/__tests__/coerceEvaluatorScalar.unit.test.ts` — 12 cases
- `langwatch/src/server/routes/__tests__/evaluator-input-coercion.unit.test.ts` — 9 cases through `getEvaluatorDataForParams`
- `services/nlpgo/app/engine/blocks/evaluatorblock/coerce_test.go` — 12 scalar + contexts passthrough + conversation recursion + nil
- `langevals/evaluators/langevals/tests/test_exact_match.py` — 8 loose-equal matches + 5 mismatches + transform-chain preservation + float-short-circuit preservation
- Plus Lane B unit + integration suites in Sarah's commit (42 + 8 green locally)

### Lane A — REST grid against the customer's failing endpoint

Customer's exact failure mode was `POST /api/evaluations/evaluators/<slug>/evaluate` returning `400 "Validation error: Expected string, received boolean at output"`. After the fix:

| Mapped `output` | `expected_output` | Verdict | Details |
|---|---|---|---|
| `true` (bool) | `"1"` | **passed** | `"true == 1"` — customer's case, no more 400 |
| `false` (bool) | `"0"` | **passed** | `"false == 0"` |
| `true` (bool) | `"true"` | **passed** | string-form match (python NLP parity) |
| `42` (number) | `"42"` | **passed** | numeric coerce |
| `true` (bool) | `"0"` | **rejected** | correct mismatch, scorer not over-eager |
| `"hello"` (str) | `"hello"` | **passed** | strings pass through coercion untouched |

Full payloads + responses captured at https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4642/lane-a/rest-grid-results.txt.

### Lane B — workbench auto-mapping no longer cross-falls

Added `Exact Match Evaluator` as a target, then added `Downstream Exact Match` on top. The downstream evaluator drawer shows:

lane-b output required

`output` is shown as **Required** (orange — no auto-mapping) because the target output names (`passed` / `score` / `label`) don't semantically match `output`, and the dataset's `output`-like columns are correctly suppressed. `expected_output` auto-mapped to `Test Data.expected_output` as expected.

After manual mapping `output → Exact Match Evaluator.passed`:

lane-b mapped

The Variables panel now shows `output = Exact Match Evaluator.passed` (boolean target output) and `expected_output = Test Data.expected_output` — exactly the customer's intended wiring, with no confusing dataset-to-itself default to undo first.

## Test plan

- [x] `pnpm test:unit` for Lane A TS surfaces
- [x] `go test` for nlpgo evaluatorblock package
- [x] `uv run pytest` for langevals exact_match
- [x] Workbench end-to-end: evaluator-as-target with bool output, downstream Exact Match expecting "1" — verified live on lw-local, REST returns `passed=true` with `details: "true == 1"`
- [x] Auto-mapping inference: confirmed `output` ignores `dataset.output`, leaves the mapping empty when the target outputs don't match
- [ ] Online evaluation with bool trace metadata — same REST entry point, covered transitively by Lane A1

### DIFF
--- a/services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
+++ b/services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
@@ -159,7 +159,7 @@ func (e *Executor) Execute(ctx context.Context, req Request) (*Result, error) {
 	}
 
 	body := map[string]any{
-		"data": req.Data,
+		"data": coerceData(req.Data),
 	}
 	if req.Name != "" {
 		body["name"] = req.Name
--- a/services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
+++ b/services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
@@ -158,8 +158,21 @@ func (e *Executor) Execute(ctx context.Context, req Request) (*Result, error) {
 		return nil, err
 	}
 
+	// Coercion is scoped to the default langevals/* string-input path,
+	// where the receiving Pydantic schema declares every field as `str`
+	// and would reject a raw bool/number with "Expected string, received
+	// boolean". Saved (`evaluators/*`) and workflow (`custom/*`)
+	// evaluators carry their own typed field definitions
+	// (bool/float/int/dict/json_schema) and the app side resolves them
+	// against the evaluator's declared inputs, so the data payload must
+	// stay pass-through here — coercing would silently convert typed
+	// fields to strings before the receiving evaluator can validate them.
+	data := req.Data
+	if strings.HasPrefix(req.EvaluatorSlug, "langevals/") {
+		data = coerceData(req.Data)
+	}
 	body := map[string]any{
-		"data": coerceData(req.Data),
+		"data": data,
 	}
 	if req.Name != "" {
 		body["name"] = req.Name
--- a/langevals/evaluators/langevals/langevals_langevals/exact_match.py
+++ b/langevals/evaluators/langevals/langevals_langevals/exact_match.py
@@ -1,4 +1,5 @@
-from typing import Optional
+import json
+from typing import Any, Optional
 from langevals_core.base_evaluator import (
     BaseEvaluator,
     EvaluatorEntry,
@@ -60,6 +61,20 @@ class ExactMatchEvaluator(
                 details=f'{output_text} == {expected_output_text}' if passed else f'Expected {output_text} to be equal to {expected_output_text}',
             )
 
+        # Recover the underlying value on each side (json.loads recovers
+        # bools and numbers from their string-coerced form) and compare with
+        # ECMA 7.2.13 loose semantics. An upstream evaluator that emits
+        # `passed: true` then surfaces as the string "true" through the
+        # coercion layer — without this layer it would mismatch a dataset
+        # golden of "1". The transform chain (trim/punct/case) is left for
+        # the genuinely-text-vs-text case so unrelated strings never collide.
+        if _js_loose_equal(output_text, expected_output_text):
+            return ExactMatchResult(
+                score=1,
+                passed=True,
+                details=f'{output_text} == {expected_output_text}',
+            )
+
         if self.settings.trim_whitespace:
             output_text = output_text.strip()
             expected_output_text = expected_output_text.strip()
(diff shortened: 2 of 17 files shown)


rogeriochaves:
Coderabbit findings addressed in d7630e6fc:

1. **Ruff E712** (test_exact_match.py lines 187+194): replaced `assert result.passed == True` with `assert result.passed` on the two tests added by this PR. Left the pre-existing `==True`/`==False` patterns elsewhere in the file untouched to keep the blast radius scoped to this PR.

2. **Non-finite floats in coerce.go**: added `math.IsNaN`/`math.IsInf` guards in both the float64 and float32 branches so NaN and ±Inf now return nil, matching the TS `coerceEvaluatorScalar` branch that returns null for non-finite numbers. Locked the parity with a new Go test `TestCoerceScalarNonFiniteFloats`.

services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
```diff
@@ -159,7 +159,7 @@ func (e *Executor) Execute(ctx context.Context, req Request) (*Result, error) {
 	}
 
 	body := map[string]any{
-		"data": req.Data,
+		"data": coerceData(req.Data),
```
rogeriochaves: P1: Applying `coerceData` unconditionally changes the wire contract for saved/custom workflow evaluators. `EvaluatorSlug` can be `evaluators/<id>` (and the app resolves workflow evaluators to `custom/<workflowId>`), where `getEvaluatorDataForParams` intentionally passes `params.data` through unchanged. With this change, a workflow evaluator input declared as `bool`, `float`, `int`, `dict`, or `json_schema` arrives as strings like `"true"`, `"0.5"`, or JSON text before the workflow can validate/use its declared types. The repo already supports these non-string field types in workflow DSL, so this will break custom evaluators even though the regression being fixed is only for default string-input evaluator fields. Please scope coercion to the default/langevals string-field path, or move the type-aware coercion to the app side where `checkType` and evaluator field definitions are known; saved/custom workflow evaluator data should stay pass-through.

rogeriochaves (commented):
1 P1 finding: the Go evaluator dispatch now coerces custom workflow evaluator inputs before they reach the custom pass-through route. Details inline.

services/nlpgo/app/engine/blocks/evaluatorblock/executor.go
rogeriochaves: Addressed in 977e10284.

`coerceData` is now gated on `strings.HasPrefix(req.EvaluatorSlug, "langevals/")`. Saved (`evaluators/*`) and workflow (`custom/*`) evaluators pass `req.Data` through unchanged, so workflow evaluator inputs declared as `bool`, `float`, `int`, `dict`, or `json_schema` keep their fidelity all the way to the receiving evaluator. Mirrors the TS side at `evaluations-legacy.ts:533` where `getEvaluatorDataForParams` already short-circuits on `checkType` starting with `custom/` before reaching the `defaultEvaluatorInputSchema` coerce wire (ash confirmed in channel).

Test added at `executor_test.go` pinning both sides of the boundary:
- `TestExecute_CustomAndSavedSlugsPassDataThroughUnchanged` asserts `evaluators/saved-eval-id` and `custom/workflow-id-abc` carry `bool`/`number`/`map` on the wire untouched.
- `TestExecute_LangevalsSlugStillCoercesData` keeps the customer's `langevals/exact_match` + `output:true` → `"true"` path intact.

Go pkg tests green locally.

PR #4653, opened 2026-06-07
refactor(nlp): remove the python langwatch_nlp service, nlpgo is the only engine

rogeriochaves:
## Summary

Removes the Python `langwatch_nlp` service entirely. The Go `nlpgo` service (already serving production via the `/go/*` paths) becomes the **only** NLP engine. The self-hosted Docker image and the production per-tenant Lambda drop the whole Python application stack (uvicorn, litellm, dspy, langwatch_nlp), keeping only a minimal `python3` + a small curated set of libraries to sandbox user code blocks.

Pairs with **langwatch/langwatch-saas#570** (the per-tenant lambda + k8s service infra). **Merge order: this PR first, then #570** (its submodule re-bumps to this PR's merge commit).

## Why — the win

Per-tenant lambdas are created on a project's first Studio/eval run after each deploy (CreateFunction -> poll Active -> first invoke), so a customer pays the full image-distribution + optimization cost. Measured on lw-dev, fresh never-optimized digests, 1024MB arm64, n=3 median (credit @ash):

| phase | FAT python ~1GB | GO distroless 140MB |
|---|---|---|
| CreateFunction -> Active (image distribute + optimize) | 19.9s | 11.6s |
| first invoke (Active -> served) | 0.8s | 0.8s |
| **TOTAL create -> first served** | **20.6s** | **12.4s** |

**~8.2s / ~40% faster per new tenant** (or per project's first run after each deploy), with much tighter variance (Go 12.0-13.1s vs fat 16.5-20.9s). The entire delta is `CreateFunction -> Active` and scales with image size. Plus: steady cold-start +102ms (the image lazy-loads the python it never reads on a no-code-block run) and memory 4Gi -> 1-2Gi with no background uvicorn process.

Self-hosted image: **~1GB-class -> 136MB**.

## What changed

- **Go engine** (`services/nlpgo`): strip the uvicorn child manager, the reverse proxy, and all `NLPGO_CHILD_*` config. nlpgo is unconditionally go-only; any non-`/go/*` request returns a typed 502.
- **App routing**: remove the `release_nlp_go_engine_enabled` feature flag and every Python-fallback branch. Studio execution + the playground proxy always hit `/go/*`; topic clustering always routes to **langevals**; `execute_optimization` always 410s.
- **Self-hosted image** (`Dockerfile.langwatch_nlp`): distroless/python3 + the static Go binary + a 35MB sandbox. Verified end-to-end (`/healthz` 200, `/go/version` 200, non-`/go` 502, sandbox imports work).
- **Code blocks**: still run arbitrary user Python via an embedded `runner.py` + the curated `sandbox-requirements.txt` (the only Python in the artifact).
- **npx self-host** (`@langwatch/server`): runs `service nlpgo` (already its default `go` mode); the legacy `python` mode + the `langwatch_nlp/` directory are deleted.
- **Helm**: nlp deployment points at the Go image, resources lowered to the Go tier, probes repointed to `/healthz` `/readyz` `/startupz`.
- **CI / dev**: Go-only workflows (drop the python lint/test jobs, keep the image smoke build; Go source is covered by `go-ci.yaml`), compose, Makefile, scripts. The orphan monorepo `Dockerfile.langwatch_nlp.lambda` is deleted (the saas runtime Dockerfile owns the prod lambda image).

## Code-block sandbox libraries

`services/nlpgo/app/engine/blocks/codeblock/sandbox-requirements.txt` is a deliberately small, fixed, documented set installed into both the self-hosted image and the saas lambda image from one file: `requests`, `httpx`, `pydantic`, and the `langwatch` Python SDK (prompt/dataset fetch). No pandas/numpy (langwatch does not require them). The old image leaked its entire dependency tree (litellm, dspy, ...) into user code by accident; that is not a contract and is not carried here.

## Scope note

`langevals` stays. It is a separate Python service (evaluators + presidio PII + topic clustering, which was deliberately migrated there). "Remove the Python NLP" means `langwatch_nlp` specifically.

## Verification

- Go: `go build ./cmd/service` + `go test ./services/nlpgo/...` green.
- App: `pnpm typecheck` clean; updated unit tests pass; feature-parity gate passes (1549 enforced scenarios bound).
- Image built + verified end-to-end (above).
- `helm lint` + `helm template` clean.
- [x] Dogfooded the full Studio flow on **lw-dev EKS** (self-hosted helm path) — see below.

## Dogfood — self-hosted Go-only pod on lw-dev EKS

Deployed the Go-only distroless image to the lw-dev `langwatch-langwatch-nlp` Deployment (the self-hosted helm shape) and exercised it end-to-end.

**Before** (old pod): `nlpgo-9114f90ac4c2-arm64` (fat ~1GB python), probes on `/health`, logs show `service=uvicorn-child` + `uvicornchild/manager.go` + litellm on python3.12.

**After** (Go-only distroless, this PR's image):
- **Rollout went Ready under the new probes** `/healthz` `/readyz` `/startupz` — a wrong probe path would hang the rollout, so a clean `successfully rolled out` proves the helm probe change on real infra.
- Pod logs: `nlpgo_starting :5561`, lifecycle services = `otel` + `http` only, `transport=in_process_dispatcher`, **zero** uvicorn/child/litellm lines.
- From inside the distroless pod (no shell/curl, via the baked `python3.11`): `/healthz` `/readyz` `/startupz` `/go/version` all `200`; a non-`/go` path returns the typed `502` go-only.
- **Real code-block** at the deployed pod (`POST /go/studio/execute_sync`, a Studio `execute_flow` whose code node imports the curated sandbox libs and builds a pydantic model) → `status: success`, `result: httpx=0.28.1 requests=2.34.2 pydantic=2.13.4 langwatch=True model_n=18`. All four curated libs import (including `langwatch` — the dist-info fix), pydantic model built.
- **Through the real workbench UI** (lw-dev app → the patched EKS pod): imported the same code-block workflow into the Optimization Studio and ran it; the End node renders the identical `httpx=0.28.1 requests=2.34.2 pydantic=2.13.4 langwatch=True model_n=18`.

The deployed-pod result is byte-identical to the saas lambda result (langwatch/langwatch-saas#570) because both artifacts install the curated sandbox from the one shared `sandbox-requirements.txt` build stage.

Code-block workflow in the real lw-dev Optimization Studio, executed against the Go-only EKS pod:

workbench code-block executed

workbench code-block canvas

After the dogfood the lw-dev nlp Deployment was reverted to its prior image and the dev nodegroup scaled back to zero.


services/nlpgo/adapters/httpapi/router.go
```diff
@@ -80,43 +76,31 @@ func NewRouter(deps RouterDeps) http.Handler {
...
-		r.NotFound(deps.ChildProxy.ServeHTTP)
-		r.MethodNotAllowed(deps.ChildProxy.ServeHTTP)
-	} else {
-		r.NotFound(goOnlyModeFallback)
-		r.MethodNotAllowed(goOnlyModeFallback)
-	}
+	// Anything outside /go/* and the health routes gets a typed 502 with
+	// a self-explaining body. nlpgo is the only NLP engine and serves
+	// only /go/*; there is no Python service to proxy to.
+	r.NotFound(goOnlyModeFallback)
```
langwatch-agent, line 82: **P1** — This makes every non-`/go/*` request return 502, but not all live callers were migrated. Scenario execution still posts code/workflow agents to `${LANGWATCH_NLP_SERVICE}/studio/execute_sync` (`langwatch/src/server/scenarios/execution/serialized-adapters/code-agent.adapter.ts:236`, `workflow-agent.adapter.ts:140`), and scenario prompt models still use `${nlpServiceUrl}/proxy/v1` (`model.factory.ts:33`). After this PR those paths land here and fail instead of executing. Minimal fix: route the scenario adapters through `/go/studio/execute_sync` and the scenario model factory through `/go/proxy/v1` (or shared helpers), and update the tests that currently assert the legacy URLs.

services/nlpgo/adapters/httpapi/router.go
```diff
@@ -80,43 +76,31 @@ func NewRouter(deps RouterDeps) http.Handler {
...
-	// clear message rather than chi's default 404 (which would suggest
-	// the URL is wrong) or a generic dial-failure log line.
-	if deps.ChildProxy != nil {
-		r.NotFound(deps.ChildProxy.ServeHTTP)
-		r.MethodNotAllowed(deps.ChildProxy.ServeHTTP)
-	} else {
-		r.NotFound(goOnlyModeFallback)
-		r.MethodNotAllowed(goOnlyModeFallback)
-	}
+	// Anything outside /go/* and the health routes gets a typed 502 with
```
langwatch-agent, line 79: **P2** — Dropping every non-healthz legacy path also removes the old `/health` response, but `charts/langwatch/tests/e2e-full-stack.sh:223` still smokes `http://${RELEASE}-langwatch-nlp:5561/health` and expects 200. The new image will return the go-only 502 here, so the full-stack Helm e2e breaks even though the deployment probes were updated. Minimal fix: update that smoke to `/healthz` (or keep a `/health` compatibility alias if external health checks still depend on it).

langwatch-agent (commented):
2 findings (1 P1, 1 P2) — scenario NLP callers still hit removed legacy paths, and the Helm full-stack smoke still checks the removed /health endpoint. Details inline.

services/nlpgo/adapters/httpapi/router.go
rogeriochaves: Fixed in 40f1846d6. The scenario adapters bypassed `nlpgoFetch` (which already prepends `/go`) and built URLs directly: routed `code-agent.adapter.ts` and `workflow-agent.adapter.ts` through `/go/studio/execute_sync`, and the scenario `model.factory.ts` proxy through `/go/proxy/v1`. Updated the unit tests that asserted the legacy URLs (code-agent + workflow-agent adapter tests, 54 pass).

services/nlpgo/adapters/httpapi/router.go
rogeriochaves: Fixed in 40f1846d6. Updated `charts/langwatch/tests/e2e-full-stack.sh` to smoke `/healthz` instead of the removed `/health` (the Go-only nlpgo serves `/healthz` `/readyz` `/startupz`; non-`/go` paths now return the typed 502). No external health check depends on the old `/health`, so no compat alias needed.

Issue #1258, opened 2026-01-30
Refactor base worker startup and improve logging


drewdrewthis:
## Summary

Decouple worker startup so workers run in exactly ONE place - never duplicated.

### Current Problem

Workers can accidentally run in multiple places simultaneously:

```
┌─ langwatch deployment (runs: pnpm start) ─────────────────────────┐
│ start.sh runs concurrently:                                       │
│ ├─ start:app → start.ts → startApp()                              │
│ │                └─ initializeBackgroundWorkers() ← WORKERS #1    │
│ └─ start:workers → workers.ts ← WORKERS #2 (if REDIS_URL set)     │
└───────────────────────────────────────────────────────────────────┘

┌─ langwatch-workers deployment (runs: npm run start:workers) ──────┐
│ start:workers → workers.ts ← WORKERS #3                           │
└───────────────────────────────────────────────────────────────────┘
```

### Solution

Remove `initializeBackgroundWorkers()` from `startApp()`. Workers should ONLY be started via `start:workers` / `workers.ts`.

**Production:**
- Main deployment: `start.sh` runs `start:workers` alongside the app (if no separate worker pods)
- Worker deployment: runs `start:workers` standalone (preferred for scaling)
- App pods just serve HTTP, never run workers in-process

**Local development:**
- `pnpm dev` spins up workers via docker profile or the existing concurrent setup
- No in-process worker initialization needed

## Goals

- [ ] Remove in-process worker initialization from `startApp()`
- [ ] Workers only start via `start:workers` entry point
- [ ] Clean up worker startup logic
- [ ] Improve worker logging for better observability
- [ ] Document the worker architecture clearly

## Context

Split out from OTEL scenario traces PR to keep concerns separated.

Issue #1186, opened 2026-01-23
Fix: Save API key when switching from env vars to custom keys


drewdrewthis:
## Summary
Bug fix for `useModelProviderForm` - when a user has env vars configured but enters a new API key, the key should be saved.

## Problem
Previously, when `isUsingEnvVars` was true and a user entered a new API key, the form would filter out the key thinking it was a masked placeholder.

## Solution
Add `hasUserEnteredNewApiKey()` check to detect when user has actually typed a new key, and send it in that case.

PR #4656, opened 2026-06-07
fix(deps): close langevals transformers and strawberry-graphql security alerts

langwatch-agent:
## Why

These are the two langevals security alerts deferred from #4654. Both turned out to be fixable cleanly once the dependency graph was traced, so this closes them with minimal stable upgrades: no downgrade, no prerelease pin, no extra native/torch churn.

## What changed

`langevals/uv.lock` plus two small manifest edits. Lock diff: no packages added, none removed, no downgrades, five upgrades total.

| package | old -> new | why | alert(s) |
| --- | --- | --- | --- |
| transformers | 4.57.6 -> 5.9.0 | security fix | #833 (MODERATE) |
| strawberry-graphql | 0.314.3 -> 0.316.0 | security fix | #1257 #1258 (MODERATE), #1186 (LOW) |
| huggingface-hub | 0.36.1 -> 1.16.1 | transformers 5.x transitive | - |
| hf-xet | 1.2.0 -> 1.5.0 | huggingface-hub transitive | - |
| cross-web | 0.4.1 -> 0.7.0 | strawberry transitive | - |

Closes **4 alerts** (3 MODERATE, 1 LOW). arize-phoenix stays at 14.6.0 (no downgrade), torch stays at 2.10.0.

### transformers (GHSA-69w3-r845-3855)

The advisory's first patched version is the `5.0.0rc3` prerelease, but the stable 5.x line is published up to 5.10.2, so this pins stable 5.9.0 instead of a release candidate. `torch` is only an optional extra of transformers, so the 4 -> 5 bump does not add native binaries by itself (torch is already in the lock via the notebooks member).

`langevals_core`'s security floor is raised `transformers>=4.49.0 -> >=5.0.0` (same `# Enforce versions due to security issues` block that already exists there), so the lock cannot resolve back to a vulnerable 4.x. The vulnerable code path is the `Trainer` class `_load_rng_state()` calling `torch.load()` without `weights_only=True`; langevals only uses transformers for `AutoConfig` and HuggingFace embeddings, never `Trainer`.

### strawberry-graphql (GHSA-fr49-mhgj-crfc, GHSA-qfwv-87qj-98xq, GHSA-x97m-qp5c-w9xj)

strawberry is pulled in only by arize-phoenix (a notebooks dependency), which pins it exactly to the vulnerable `==0.314.3`. The first arize-phoenix release that pins a patched strawberry is 16.4.0, but it is still inside the repo's `exclude-newer = "7 days"` cooldown, so letting the resolver move freely downgrades arize-phoenix to 9.0.0 (the same collateral #4654 avoided). Instead, a `uv` override forces the patched `strawberry-graphql>=0.316.0`. 0.316.0 was published 2026-05-19 (well outside the cooldown) and is the exact version arize-phoenix itself adopts in 16.4.0, so it is API compatible with the pinned phoenix.

## How this was verified

- **Lock integrity:** `uv lock --check` passes; full before/after package diff shows zero added, zero removed, zero downgraded.
- **transformers API surface:** the exact imports langevals uses (`AutoConfig`, `transformers.models.auto.modeling_auto.MODEL_FOR_SEQUENCE_CLASSIFICATION_MAPPING_NAMES`) load fine under 5.9.0.
- **transformers end to end:** in the notebooks env (torch 2.10.0+cpu, transformers 5.9.0, huggingface-hub 1.16.1, sentence-transformers 5.2.2) a real model download via huggingface-hub 1.x plus a sentence-transformers `encode()` produces a 384 dim embedding.
- **No new test failures:** legacy evaluator suite 10 passed / 4 failed, ragas evaluator suite 21 passed / 2 skipped / 1 failed. Every failure is identical on baseline (transformers 4.57.6) and is the pre-existing `test_with_anthropic_models` litellm 404 (the `claude-3-5-sonnet-20240620` model is not available with this environment's key), unrelated to these bumps.
- **strawberry 0.316.0:** builds a schema and executes a query cleanly standalone; the 0.314.3 -> 0.316.0 delta is two CVE validation fixes, a GraphiQL header fix, a DataLoader internal fix, and an extensions factory improvement, with no schema definition API breaks.
- **Phoenix note:** arize-phoenix's full GraphQL schema cannot be runtime built in this environment, on this branch or on `main`, due to a pre-existing and unrelated incompatibility between `openinference-instrumentation-openai` 0.1.45 and `openinference-semantic-conventions` 0.1.26 (`AttributeError: GROQ`). Those package versions are unchanged by this PR. arize-phoenix is a notebooks only dependency.

## Not included here

- transformers in `python-sdk/uv.lock` (alert #831, same advisory) is a separate uv project where transformers is transitive via `unstructured-inference`; it upgrades cleanly to 5.9.0 too but needs its own document parsing dogfooding and carries unrelated lock churn, so it will be a dedicated follow up.

### DIFF
--- a/langevals/langevals_core/pyproject.toml
+++ b/langevals/langevals_core/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "google-cloud-aiplatform>=1.133.0",
     "boto3>=1.34.34",
     # Enforce versions due to security issues
-    "transformers>=4.49.0",
+    "transformers>=5.0.0",
     "anyio>=4.8.0",
     "zipp>=3.23.0",
 ]
--- a/langevals/pyproject.toml
+++ b/langevals/pyproject.toml
@@ -71,6 +71,11 @@ exclude-newer = "7 days"
 override-dependencies = [
     "langchain>=0.3.30",
     "langsmith>=0.8.0",
+    # Force a patched strawberry-graphql (GHSA-fr49-mhgj-crfc, GHSA-qfwv-87qj-98xq,
+    # GHSA-x97m-qp5c-w9xj). arize-phoenix 14.6.0 pins it to 0.314.3; the patched
+    # 0.316.0 is API-compatible with Phoenix and is the version Phoenix itself moved
+    # to in 16.4.0.
+    "strawberry-graphql>=0.316.0",
 ]
 
 [tool.uv.workspace]
--- a/langevals/uv.lock
+++ b/langevals/uv.lock
@@ -34,6 +34,7 @@ members = [
 overrides = [
     { name = "langchain", specifier = ">=0.3.30" },
     { name = "langsmith", specifier = ">=0.8.0" },
+    { name = "strawberry-graphql", specifier = ">=0.316.0" },
 ]
 
 [[package]]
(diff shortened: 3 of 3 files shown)


langwatch-agent:
Ready for human review. Closes 4 langevals alerts (#833 transformers, #1257/#1258/#1186 strawberry-graphql) with no downgrade and no added/removed packages. Full verification (transformers end-to-end embedding, baseline-compared test runs, strawberry schema build, and the pre-existing unrelated phoenix/openinference note) is in the PR description.

PR #4662, opened 2026-06-08
perf(clickhouse): add time predicate to analytics span-facet subqueries (stop stored_spans cold scan)

langwatch-agent:
## Evidence

Triaging the last 24h of prod ClickHouse logs (cold-scan signal), `stored_spans` is the top uncovered cold-scan table. After the per-trace drawer reads, the next concentration is analytics: dashboard queries that filter traces by a span or event facet. Their `coldScanTable` is `stored_spans` even though the request carries a date range (`paramKeys` like `[tenantId, startDate, endDate, f0_values, ...]` and `[tenantId, currentStart, currentEnd, previousStart, previousEnd, models, labels]`), p50 read ~130 MB.

`stored_spans` is `PARTITION BY toYearWeek(StartTime)` and tiered to S3. A facet subquery with no `StartTime` predicate walks every weekly partition including cold ones on S3 (a burst of GETs per cold partition), on every dashboard load that uses a span/event filter.

## Suspected cause

Span and event facet filters translate to `trace_summaries.TraceId IN (SELECT TraceId FROM stored_spans WHERE TenantId = ... AND <attr/event match>)` (`filter-translator.ts`). The outer `trace_summaries` query is correctly bounded on `OccurredAt` (its partition key), but that bound is never propagated into the inner `stored_spans` subquery, so the inner scan can't prune partitions. The SQL is otherwise correct; the missing piece is a `StartTime` predicate on the subquery.

## Proposed fix

Thread the dashboard date envelope into the `stored_spans` facet subqueries as a `StartTime` bound:

- `translateFilter` / `translateAllFilters` take an optional `spanTimePredicate` SQL fragment, passed only into the handlers whose subquery reads `stored_spans` (span type/model, event type, event metric key/value, event detail key). Non-span handlers ignore it.
- `aggregation-builder.ts` builds the fragment from the same date params the outer filter already binds (`SPAN_TIME_FILTER_BOTH_PERIODS` for the two-period aggregation; `SPAN_TIME_FILTER_START_END` for the single start/end-range builders) and passes it in.

A span's `StartTime` falls within its trace's lifetime, so the bound is the same `[period start, period end]` envelope the outer `OccurredAt` filter uses, plus a 2-day cushion for long traces / clock skew (matching the ±2-day span-fetch partition hints elsewhere in the codebase). This prunes the scan without changing which traces match.

This PR also completes the wiring of `NonBilledCost` (a `trace_summaries` column added in #4646). The non-billed-cost metric references `${ts}.NonBilledCost`, but the column was never added to `TRACE_ANALYTICS_COLUMNS`, so the deduped `trace_summaries` subquery never projected it and any query using that metric failed at runtime with `Identifier ts.NonBilledCost cannot be resolved`. This also left the guard test (every `${ts}.<Column>` reference must be handled) red on `main`. The fix mirrors `TotalCost`: add the column to `TRACE_ANALYTICS_COLUMNS` (inner projection), `DEDUP_FIELD_MAPPINGS`, and the period-comparison CTE select. Verified end to end against a real ClickHouse instance via the `trace-eval-mix-inflation` and `memory-safety` integration suites.

## Expected impact

For dashboard queries with a span/event facet filter, the `stored_spans` subquery prunes to the date range's weekly partitions instead of all of them. Headline metric is partitions pruned and the projected S3 GET reduction, not latency (the subquery returns few TraceIds). No change to results.

## EXPLAIN check

`EXPLAIN INDEXES` on the `spans.type` facet subquery against a large prod tenant over a 7-day range (literals redacted):

Old (no `StartTime` predicate):
```
Parts:    269/269
Granules: 133718/133718
```

New (`StartTime` envelope, ±2-day cushion):
```
Parts:    30/269
Granules: 4718/133718
```

A ~9x reduction in parts opened. The reviewer should re-run before merge.

## Test plan

- `filter-translator.test.ts`: new cases assert the `StartTime` predicate is injected into each `stored_spans` facet subquery when supplied, omitted when not (unchanged behavior), and never leaks into non-span filters.
- `aggregation-builder.test.ts`: new case asserts a built timeseries query with a `spans.type` filter carries the `StartTime` envelope on the subquery; the pre-existing `NonBilledCost` guard test now passes.
- `pnpm typecheck` clean; analytics unit suite green (150+ tests).

Note for the table in the skill docs: `project_FzsKx-0tpHezJlGxrG0sD` still EXPLAINs against ~268 `stored_spans` parts, a good large-tenant reference.

Advisory PR from the ClickHouse slow-query optimizer. Open for review, not for self-merge.

### DIFF
--- a/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -53,6 +53,23 @@ describe("aggregation-builder", () => {
       expect(result.sql).toContain("'previous'");
     });
 
+    // A span/event facet filter resolves to a `... FROM stored_spans ...`
+    // subquery. stored_spans is partitioned by toYearWeek(StartTime) and tiered
+    // to S3, so the subquery must carry a StartTime predicate or it cold-scans
+    // every weekly partition. The envelope reuses the query's own date params.
+    it("bounds the stored_spans facet subquery to the date envelope by StartTime", () => {
+      const input = {
+        ...baseInput,
+        filters: { "spans.type": ["llm"] },
+      };
+      const result = buildTimeseriesQuery(input);
+
+      expect(result.sql).toContain("FROM stored_spans");
+      // The subquery prunes partitions instead of cold-scanning S3.
+      expect(result.sql).toContain("StartTime >= {previousStart:DateTime64(3)}");
+      expect(result.sql).toContain("StartTime < {currentEnd:DateTime64(3)}");
+    });
+
     it("includes date truncation for timescale", () => {
       const result = buildTimeseriesQuery(baseInput);
 
--- a/langwatch/src/server/analytics/clickhouse/__tests__/filter-translator.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/filter-translator.test.ts
@@ -177,6 +177,54 @@ describe("filter-translator", () => {
       });
     });
 
+    describe("stored_spans subquery partition pruning", () => {
+      // The stored_spans facet subqueries must accept a StartTime predicate so
+      // they prune to the dashboard's date range instead of cold-scanning every
+      // weekly partition on S3. The predicate is the caller's responsibility
+      // (aggregation-builder passes the date envelope); the translator just has
+      // to thread it into the subquery WHERE.
+      const SPAN_TIME = "AND StartTime >= {startDate:DateTime64(3)}";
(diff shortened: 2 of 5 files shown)


langwatch-agent:
Good catch on the BDD convention. Renamed the block to a `when`-style describe (`when a span or event facet filter targets stored_spans`) in d3eba6da5. Kept it a single `describe` to match the surrounding blocks in this file (`span filters`, `topic filters`, etc.), which are flat rather than nested given/when.

rogeriochaves (approved):

Issue #1239, opened 2026-01-28
Support air-gapped Docker deployments: Prisma engine checksum fetch failures


drewdrewthis:
## Problem

When deploying LangWatch in an air-gapped environment (no internet access), Prisma operations can fail because they try to fetch SHA256 checksums from `binaries.prisma.sh`:

```
Error: Failed to fetch sha256 checksum at https://binaries.prisma.sh/all_commits/.../libquery_engine.so.node.sha256 - 500 Internal Server Error

If you need to ignore this error (e.g. in an offline environment), set the PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING environment variable to a truthy value.
```

## Impact

Without this fix, users cannot run LangWatch in:
- Corporate environments with restricted internet access
- Air-gapped security environments  
- Networks with firewalls blocking `binaries.prisma.sh`

## Recommended Solution

Set the `PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=1` environment variable in the Docker image.

Since the Prisma engines are already installed during the Docker build (with network access), this env var simply tells Prisma to skip checksum verification at runtime - the engines are already in place.

```yaml
environment:
  PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING: 1
```

## Other Possible Solutions

1. **Pre-bundle Prisma engines in the Docker image** - Explicitly include the compiled binaries during build so no runtime fetch is needed
2. **Use `binaryTargets`** in schema.prisma to pre-download engines for specific platforms during build
3. **Document workaround** - Provide instructions for users to manually copy engines if needed

## References

- Prisma Issue: https://github.com/prisma/prisma/issues/20021
- Prisma docs on offline environments: https://www.prisma.io/docs/orm/reference/environment-variables-reference#prisma_engines_checksum_ignore_missing

drewdrewthis:
Moving to langwatch-saas repo - the Dockerfile that needs this fix is there.

PR #4676, opened 2026-06-08
fix(deps): sync langwatch lockfile with mcp-server vitest bump

langwatch-agent:
## Why

CI on `main` is failing `pnpm install --frozen-lockfile`:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because
pnpm-lock.yaml is not up to date with mcp-server/package.json
  - vitest (lockfile: >=4.1.0, manifest: ^4.1.0)
```

## Root cause

The langwatch workspace (`langwatch/pnpm-workspace.yaml`) includes `../mcp-server` as a member, and `langwatch/package.json` carries a security override `"vitest@<4.1.0": ">=4.1.0"`.

When `langwatch/pnpm-lock.yaml` was last generated, mcp-server's vitest was `< 4.1.0`, so that override matched and rewrote the lock's `../mcp-server` importer specifier to `>=4.1.0`.

#4479 then bumped mcp-server to `vitest ^4.1.0` and regenerated `mcp-server/pnpm-lock.yaml`, but it did not touch `langwatch/pnpm-lock.yaml`. `^4.1.0` no longer matches the `<4.1.0` override, so the expected importer specifier is `^4.1.0`, but the stale langwatch lock still recorded `>=4.1.0`. Hence the frozen-install mismatch.

## Fix

Regenerated `langwatch/pnpm-lock.yaml`. The diff is small:

- `../mcp-server` importer vitest specifier `>=4.1.0` -> `^4.1.0` (the actual fix).
- vitest still resolves to `4.1.4`, above the `4.1.0` first patched version, so the CRITICAL alerts (#1239 / #1240) stay closed.
- the re-resolution also deduped `es-module-lexer` to 2.1.0 and nudged `tinyglobby` to 0.2.16; nothing else changes.

The `vitest@<4.1.0` override is kept (it is harmless now and still guards against any future sub-4.1.0 vitest entering the tree).

## Verification

`pnpm install --frozen-lockfile` in `langwatch/` now succeeds ("Lockfile is up to date, resolution step is skipped").

### DIFF
--- a/langwatch/pnpm-lock.yaml
+++ b/langwatch/pnpm-lock.yaml
@@ -845,7 +845,7 @@ importers:
         specifier: ^1.0.0
         version: 1.1.0
       vitest:
-        specifier: '>=4.1.0'
+        specifier: ^4.1.0
         version: 4.1.4(@opentelemetry/api@1.9.1)(@types/node@25.6.0)(@vitest/browser-playwright@4.1.8)(@vitest/coverage-v8@4.1.5)(@vitest/ui@4.1.5)(jsdom@28.1.0(@noble/hashes@2.2.0))(vite@8.0.10(@types/node@25.6.0)(esbuild@0.25.12)(sass@1.99.0)(terser@5.48.0)(tsx@4.21.0)(yaml@2.8.3))
       yargs:
         specifier: ^17.7.2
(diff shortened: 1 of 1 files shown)


langwatch-agent:
Ready for human review. Fixes the `ERR_PNPM_OUTDATED_LOCKFILE` on main: the langwatch lock's ../mcp-server importer had a stale vitest specifier (`>=4.1.0`) left over from the `vitest@<4.1.0` override after #4479 bumped mcp-server to `^4.1.0`. Regenerated; `pnpm install --frozen-lockfile` passes, vitest still resolves to 4.1.4 (CRITICAL stays closed).

PR #4692, opened 2026-06-09
fix: scope public-share authorization to projectId

### LINKED ISSUE #4531: Security: scope public-share authorization to projectId
Harden `checkPermissionOrPubliclyShared`: the `publicShare` lookup matches on `(resourceType, resourceId)` only and is **not scoped to `input.projectId`**, while the downstream read uses the caller-supplied project id. Authorization should be constrained to the share's own project.

**Pre-existing** (affects live shares; independent of the orphan-sweep removal #4526) — surfaced during that review. Repro detail is in the #4526 review thread. Recommend prioritizing.

mvanhorn:
## Why

Issue #4531 (P1) reports a cross-tenant authorization hole. When a user lacks direct permission, `checkPermissionOrPubliclyShared` in `langwatch/src/server/api/rbac.ts` falls back to looking up a `PublicShare` row, but the `where` clause matched only on `(resourceType, resourceId)` and was not scoped to `input.projectId` even though the downstream read uses the caller-supplied project id.

With the orphan cleanup removed in #4526, a stale `PublicShare` row for a resource id in project A could authorize a read of the same resource id under project B, and same-project trace-id reuse could inherit an old public URL. This is a live-share authorization issue independent of the orphan sweep.

## What changed

- `langwatch/src/server/api/rbac.ts`: add `projectId: input.projectId` to the `publicShare.findFirst` `where` clause in `checkPermissionOrPubliclyShared`.
- `langwatch/src/server/api/__tests__/rbac-integration.test.ts`: extend the existing `checkPermissionOrPubliclyShared` suite with project-scoping coverage.

## How it works

The public-share fallback now only authorizes a resource when the matching `PublicShare` row belongs to the same `projectId` the caller is reading under. Cross-project and stale-reuse shares fail with the existing `UNAUTHORIZED` error, while legitimate same-project shares are unaffected (no behavior change). The generic constraint `InputType extends { projectId: string }` already guarantees `input.projectId` is present and type-safe, and this aligns with the repo convention of always scoping project-level Prisma queries by `projectId`.

## Test plan

- `pnpm typecheck` (tsgo, `tsconfig.tsgo.json`): clean, 0 errors.
- `npx vitest run src/server/api/__tests__/rbac-integration.test.ts`: 135 passed, 1 pre-existing skip.
- Scenarios added:
  - Lookup is scoped by `projectId` (asserts the `where` clause carries the calling project).
  - Same-project share still authorizes the read (regression-safe).
  - Cross-project share for the same resource id is now `UNAUTHORIZED`.
  - Stale share row from another project does not authorize a re-created same-id trace (`UNAUTHORIZED` with the existing message).

## Anything surprising?

The new tests mock `publicShare.findFirst` with an implementation that honors `where.projectId`, rather than a flat `mockResolvedValue`. A `mockResolvedValue` would pass regardless of the fix; the implementation-aware mock genuinely exercises the query scoping. As a check, temporarily reverting the one-line fix makes the scoping and same-project tests fail, confirming the tests are load-bearing.

Fixes #4531

AI was used for assistance.

## Bug Fixes
- Fixed a critical security issue in the permission system where publicly shared resources could be accessed across different projects. Public shares are now properly scoped and isolated to their respective projects, preventing unauthorized cross-project access attempts and ensuring proper access control boundaries are maintained.

### DIFF
--- a/langwatch/src/server/api/__tests__/rbac-integration.test.ts
+++ b/langwatch/src/server/api/__tests__/rbac-integration.test.ts
@@ -848,6 +848,112 @@ describe("RBAC Integration Tests", () => {
         expect(mockCtx.publiclyShared).toBe(true);
       });
 
+      describe("when scoping the public-share fallback to the project", () => {
+        // A share row that only exists in project-a. The findFirst mock honors
+        // the where clause's projectId so the query scoping is actually
+        // exercised (a `mockResolvedValue` would pass regardless of the fix).
+        function seedShareForProject(projectId: string) {
+          mockPrisma.project.findUnique.mockResolvedValue({
+            team: {
+              id: "team-123",
+              members: [],
+              organization: { members: [] },
+            },
+          });
+          mockPrisma.publicShare.findFirst.mockImplementation(
+            async ({ where }: { where: any }) =>
+              where.projectId === projectId &&
+              where.resourceType === "TRACE" &&
+              where.resourceId === "trace-123"
+                ? {
+                    id: "share-123",
+                    projectId,
+                    resourceType: "TRACE",
+                    resourceId: "trace-123",
+                  }
+                : null,
+          );
+        }
+
+        it("scopes the public-share lookup by the calling project", async () => {
+          seedShareForProject("project-a");
+
+          const middleware = checkPermissionOrPubliclyShared(
+            checkProjectPermission("workflows:view" as Permission),
+            { resourceType: "TRACE", resourceParam: "traceId" },
+          );
+
+          await middleware({
+            ctx: mockCtx,
+            input: { projectId: "project-a", traceId: "trace-123" },
+            next: mockNext,
+          });
+
+          expect(mockPrisma.publicShare.findFirst).toHaveBeenCalledWith(
+            expect.objectContaining({
+              where: expect.objectContaining({ projectId: "project-a" }),
+            }),
+          );
+        });
+
+        it("authorizes a same-project share (regression-safe)", async () => {
+          seedShareForProject("project-a");
+
+          const middleware = checkPermissionOrPubliclyShared(
+            checkProjectPermission("workflows:view" as Permission),
+            { resourceType: "TRACE", resourceParam: "traceId" },
+          );
+
+          const result = await middleware({
+            ctx: mockCtx,
+            input: { projectId: "project-a", traceId: "trace-123" },
+            next: mockNext,
+          });
+
+          expect(result).toBe("success");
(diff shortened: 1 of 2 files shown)


langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** This is the P1 cross-tenant read hole (#4531). The one-line fix is correct, but if any `PublicShare` row can exist without a `projectId`, the new `where` constraint would reject a legitimate share (availability regression), and if other share lookups exist elsewhere the hole stays open.

- `[Must Check]` `rbac.ts` — confirm every `PublicShare` write path sets `projectId`, so adding `projectId: input.projectId` to the `findFirst` can never null out a currently-working same-project share.
- `[Ask Author]` other lookups — is `checkPermissionOrPubliclyShared` the only place a share authorizes a read? The public-view/share-resolver route that renders the shared URL should also scope by `projectId`, or the cross-tenant path persists there.
- `[Probably Fine]` `rbac-integration.test.ts` — the "integration" tests mock `publicShare.findFirst`; they prove the where-clause is built with `projectId` but do not exercise a real Prisma query. Acceptable, just know the coverage is structural.

- Are there any legitimate shares in prod whose stored `projectId` differs from the resource's current project (e.g. moved resources)? Those would flip to `UNAUTHORIZED` after this.

0xdeafcafe (approved):

0xdeafcafe:
Thank you for this @mvanhorn, and apologies for it slipping through the cracks... I'll get this security fix live asap, but this has highlighted how undesirable the way sharing is setup right now.

**To clarify:** This could only be exploited if both project and trace id were known, which due to trace ids being OTel standard does thankfully reduce the target surface. But we'll check access logs to investigate.

We'll get another release done asap which hardens this feature fully.

mvanhorn:
Thanks for landing the security fix, @0xdeafcafe - no apology needed on timing, glad it's scoped to projectId now.

0xdeafcafe:
> Thanks for landing the security fix, @0xdeafcafe - no apology needed on timing, glad it's scoped to projectId now.

This is the PR that will follow with improvements to the whole share flow https://github.com/langwatch/langwatch/pull/5547, more modern and secure approach, including more sensible defaults. But cheers again for reporting this, and contributing to the fix.

Also in the interest of transparency, I did an audit of our access logs, and it showed no data exfiltration was done via this method.

mvanhorn:
Appreciate the fast turnaround on the security side, @0xdeafcafe! Glad public-share authorization is scoped to projectId now.

PR #4702, opened 2026-06-09
docs: fix non-working .gitignore advice for committed .claude/settings.json

drewdrewthis:
## Problem

The `### Gotchas` section of `docs/ai-governance/track-your-claude-code-usage.mdx` recommends adding `!.claude/settings.json` as a negation rule so a committed `settings.json` survives a blanket `.claude/` ignore. **That advice is a no-op.** Git will not descend into an excluded directory to re-include a file under it, so `settings.json` stays swallowed — teammates silently never inherit the telemetry flags, while local telemetry keeps working for the author (file present on disk), so the failure goes unnoticed.

## Proof (`git check-ignore`, clean repo)

Naive negation (the current doc advice) — **still ignored**:

```gitignore
.claude/
!.claude/settings.json
```
```
$ git check-ignore .claude/settings.json
.claude/settings.json        # IGNORED → settings.json is swallowed, the "fix" silently fails
```

Working pattern — un-ignore the directory first, then re-ignore the secrets:

```gitignore
.claude/
!.claude/                       # un-ignore the dir so specific files can be re-included
.claude/settings.local.json     # re-ignore the secret file
.claude/*.lock                  # re-ignore other previously-covered files (e.g. scheduled_tasks.lock)
```
```
$ git check-ignore .claude/settings.json
# (no output) → NOT ignored, commits
$ git check-ignore .claude/settings.local.json
.claude/settings.local.json   # IGNORED → token stays safe
```

## Fix

This PR rewrites only the `Fix:` sentence of that bullet: it keeps the correct simple option (`git add -f .claude/settings.json`) and replaces the broken negation with the working un-ignore-the-dir-first pattern as a copyable fenced block, with a one-clause note on why the naive negation fails. Surrounding bullets/prose untouched.

Found while rolling per-repo telemetry config to 4 langwatch repos — `ai-interviewer-scenario-demo` hit exactly this case (blanket `.claude/`) and needed the corrected pattern (its PR: https://github.com/langwatch/ai-interviewer-scenario-demo/pull/3).

## Open question (NOT fixed here — for the doc owner)

The `### Gotchas` bullet near line 231 claims Claude Code "resolves `.claude/settings.json` by walking up from the working directory at launch — so starting from a subdirectory still works via walk-up." The official settings docs (https://code.claude.com/docs/en/settings) list fixed locations and do **not** document parent-directory walk-up. This claim may be incorrect but was not verifiable this turn — flagging for the doc owner to confirm against source before trusting "launch from a subdir works."

## Summary by CodeRabbit

* **Documentation**
  * Clarified that telemetry is opt-in per clone and example now shows the local enable flag.
  * Stated committed settings should only contain non-secret flags; tokens and local secrets remain ignored.
  * Confirmed raw API body logging stays disabled by default unless locally opted in.
  * Added concrete guidance to adjust ignore rules so non-secret settings can be committed while local secret files stay ignored, plus a verification step.

### DIFF
--- a/docs/ai-governance/track-your-claude-code-usage.mdx
+++ b/docs/ai-governance/track-your-claude-code-usage.mdx
@@ -170,14 +170,13 @@ An `ik-lw-` write-only key (obtainable via the governance REST endpoint) also au
 
 ### The two-file recipe
 
-Commit `.claude/settings.json` with the non-secret flags and keep the bearer token in a gitignored `.claude/settings.local.json`.
+Commit `.claude/settings.json` with the non-secret flags and keep the bearer token **plus the `CLAUDE_CODE_ENABLE_TELEMETRY` enable flag** in a gitignored `.claude/settings.local.json`. Telemetry is **off by default**: the enable flag lives only in the local file, so a fresh clone with just the committed `settings.json` (and no `settings.local.json`) never attempts to export. This matters because the committed flags include `OTEL_LOG_USER_PROMPTS` and `OTEL_LOG_RAW_API_BODIES` (prompt text and raw API bodies) — that content must never ship until the repo owner opts in locally. `settings.local.json`'s `env` merges per-key with `settings.json`'s `env`, so the local enable flag combines with the committed flags rather than replacing them.
 
 **`.claude/settings.json`** (commit this):
 
 ```json
 {
   "env": {
-    "CLAUDE_CODE_ENABLE_TELEMETRY": "1",
     "OTEL_LOG_USER_PROMPTS": "1",
     "OTEL_LOG_TOOL_DETAILS": "1",
     "OTEL_LOG_TOOL_CONTENT": "1",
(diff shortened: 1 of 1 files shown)


0xdeafcafe (commented):
Two things that would make this doc actually load-bearing instead of incidentally correct.

The off-by-default convention is currently stated as a *consequence* of where the enable flag lives. Promote it to a rule. One explicit line, something like "**Rule:** \`CLAUDE_CODE_ENABLE_TELEMETRY\` lives in \`settings.local.json\`, never in the committed \`settings.json\`", or a \`<Warning>\` block. Right now a future contributor copying the committed file as a template could re-add the enable flag in there and silently violate the convention without anything in the doc telling them they shouldn't (same failure mode as the original gitignore bug, just one layer up).

L173 has become a six-sentence wall mixing recipe, rationale, sensitivity warning, and env-merge semantics. The \`OTEL_LOG_USER_PROMPTS\` / \`RAW_API_BODIES\` sensitivity point is the strongest argument for the whole convention and it's currently buried mid-paragraph. Break it out into its own bullet or call-out.

Smaller things: cross-link to the four sibling implementation PRs as canonical examples (the body links to one, the doc itself links to none). And the "open question" about walk-up resolution at L231 is debt - either resolve here or open a tracking issue, but don't leave a known-suspect claim sitting in the rendered doc.

PR #4717, opened 2026-06-10
test(simulations): regression fixtures for input_audio url-shape and text-first ordering (#4138, #4698)

### LINKED ISSUE #4138: Voice scenario audio is not playable in the LangWatch UI — no audio fields on the wire, no audio player components
**Problem**

Voice scenarios shipped today (`langwatch/scenario` PR #355). The SDK produces real audio turns via Pipecat, OpenAI TTS, and ElevenLabs hosted Conversational AI, and writes per-segment WAVs locally. But that audio does not currently render in the LangWatch UI.

Two smoke tests today against production (`drews-sandbox-owUldA`):
- `scenariobatch_3DzSNGETWYolKYxHVk1jvz0zeAt` (OpenAI / Pipecat)
- `scenariobatch_3DzSaqKGfylvQ6eOBrJaapUS3Pq` (ElevenLabs)

Both runs succeeded client-side (`result.success=True`, WAVs on disk). Both hit a separate prod-incident bug (#4139) on the server. Once #4139 lands, a remaining UI-side gap surfaces, captured here.

**Root cause** (confirmed via /investigate)

The server-side audio pipeline is already shipped on `main` (PR #4053 / #4058):
- Wire schema accepts `input_audio` content parts (`scenarios/schemas/event-schemas.ts:113-121`).
- `extractInlineMediaFromEvent` handles `input_audio` (`stored-objects/content-extractor.ts:248-308`): decodes base64, calls `service.storeFromBytes`, rewrites part to `{ input_audio: { url: "/api/files/<id>", mimeType, data: undefined } }`. Format→MIME map covers wav/mp3/flac/ogg/webm.
- `coerce-content-to-array.ts:1-42` recovers Python-repr-stringified arrays (a real prod observation per unit test fixtures).

But: **`MediaPart.tsx` doesn't recognize the rewritten shape.** The extractor emits `{type:"input_audio", input_audio:{url, mimeType}}` (extractor lines 297-305), while `MediaPart.tsx:23-39` only accepts `{type:"audio", source:{type:"url", value, mimeType}}` (lines 227-244). The native `<audio>` element exists; it just never matches.

The SDK side of the gap is filed separately at `langwatch/scenario` — see related issue. **This issue's scope is the UI render leg only.**

**Proposed solution**

_See ACs._ Two viable shapes; investigation recommends shape (a):

(a) Patch `MediaPart.tsx` to recognize `{type:"input_audio", input_audio:{url, mimeType}}` alongside its existing `audio` source shape. Smallest blast radius; doesn't touch the extractor or its tests.

(b) Change `content-extractor.ts:297-305` to emit canonical `{type:"audio", source:{type:"url", value, mimeType}}`. Cleaner long-term shape but changes the persisted event payload, with downstream test/fixture churn.

**Alternatives considered**

- Render audio via markdown content link (extraction-bypass) — pollutes message semantics; rejected.
- Wait for AG-UI canonical media shape — out of scope timing.

**Additional context**

Source-of-truth files:
- `langwatch/src/components/simulations/MediaPart.tsx:23-39, 227-244` — render gating
- `langwatch/src/server/stored-objects/content-extractor.ts:248-308` — what shape arrives
- `langwatch/src/components/simulations/ScenarioMessageRenderer.tsx` — composer (verify render path)
- `langwatch/src/components/simulations/__tests__/MediaPart.integration.test.tsx` — extend with input_audio fixture

## Acceptance Criteria

1. `MediaPart` renders a native `<audio controls>` when the input shape is `{type:"input_audio", input_audio:{url, mimeType}}` (the canonical shape that arrives after extraction).
2. Existing `{type:"audio", source:{...}}` shape continues to render (regression test).
3. Integration test extends `MediaPart.integration.test.tsx` with an `input_audio` URL fixture.
4. Visual: a voice-scenario simulation page in prod shows inline `<audio controls>` for each turn with audio after PR-A (#4139) + PR-D (SDK transport, filed separately) land. (Verify by re-running today's `basic_greeting.py`.)
5. No change to `content-extractor.ts` output shape (deferred to a future cleanup PR if needed).

## Plan

**Approach**

Single-file UI patch in `MediaPart.tsx` to recognize the post-extraction `{type:"input_audio", input_audio:{url, mimeType}}` shape alongside the existing `{type:"audio", source:{...}}` shape. Both paths land at the same native `<audio controls>` render. No changes to the extractor, schemas, or persisted event payloads.

**Sequencing**

1. **Spec & failing test** — write the BDD scenarios into `specs/features/scenarios/` and add an `input_audio` URL fixture to `MediaPart.integration.test.tsx` that asserts an `<audio controls>` element with the expected `src`. Completion signal: red test naming the missing render branch.
2. **Implementation** — extend the type guard / discriminator in `MediaPart.tsx` (lines ~23-39) to accept the `input_audio` variant; route it through the existing audio render branch (lines ~227-244). Reuse the URL + mimeType extraction logic — do not duplicate the `<audio>` element. Completion signal: integration test green, regression test for the existing `audio` shape still green.
3. **Render-path verification** — confirm `ScenarioMessageRenderer.tsx` actually delegates `input_audio` parts to `MediaPart` (no upstream filter drops it). If a filter exists, widen the discriminator there too. Completion signal: a unit/integration test rendering through the composer produces the audio element.
4. **Visual confirmation (post-merge dependency)** — AC4 is gated on #4139 (prod infra) + the SDK transport PR. Once those land, re-run `basic_greeting.py` against `drews-sandbox-owUldA` and confirm inline `<audio controls>` per turn. This step is documented as a verification followup, not blocking PR-C merge.

**Key decisions**

- **Shape (a) over (b)**: extend the renderer, don't rewrite the extractor output. Smaller blast radius, no fixture churn, no persisted-payload migration. Trade-off: two valid audio shapes in the render layer until the cleanup PR collapses them.
- **Render branch reuse, not duplication**: both shapes funnel into one `<audio>` JSX. Avoids "two audio elements with subtly different behavior" drift.
- **Discriminator on `type === "input_audio"`** (matching the wire schema) rather than sniffing for `input_audio.url`. Aligns with how the extractor labels the part.
- **Do not validate `mimeType` against the known set in the renderer**: the extractor already enforces the format→MIME map. Renderer trusts the upstream shape and passes `mimeType` straight to the `<audio>` element's `type` attribute (or omits if absent — native player will still try).

**Risks / open questions**

- `ScenarioMessageRenderer` may pre-filter content parts by `type` and silently drop unknown variants — verify in phase 3 before declaring done.
- Existing `audio` shape test fixtures may share a code path; adding a branch must not regress them. Phase 2 includes an explicit regression assertion.
- `data: undefined` post-extraction: the renderer must not fall through into an inline-base64 branch when only a URL is present. Verify the discriminator order prioritizes `input_audio.url`.
- AC4 verification depends on two external PRs; PR-C is independently mergeable but the visual proof slips until they land.

**Out of scope**

- Any change to `content-extractor.ts` output shape (AC5 explicitly defers this).
- Any change to the wire schema in `event-schemas.ts`.
- New transport work in the `langwatch/scenario` SDK (filed separately).
- Markdown/inline audio rendering bypass (rejected in investigation).
- Canonicalizing `audio` vs `input_audio` into one shape — future cleanup PR.

Blocked by: #4139 PR-A (prod infra) and the scenario-repo SDK transport PR for end-to-end verification. PR-C (this issue's PR) can be developed and unit-tested independently.

---
*Classification: Feature (UI render leg)*
*Status: investigated. ACs above. Companion SDK transport PR filed on langwatch/scenario.*

### LINKED ISSUE #4698: Scenario run viewer: voice agent turns render duplicated and on the wrong (user) side
## For Humans

**The renderer on current `main` already does both things the bug asks for — collapse `[text, input_audio]` into one bubble AND left-align `role="assistant"`. The live repro is almost certainly stale (a deployment that predated #4058 / #4156).** A single multi-part assistant voice message is collapsed to one media bubble with the text as an italic transcript, and assistant turns align left; the simulated user aligns right. Both behaviors are pinned by an existing integration test that asserts exactly the issue's expected outcome.

**Ask:** re-confirm the bug against **current `main`** (a fresh voice run, not the old demo screenshot). If it still reproduces, the divergence is in something static reading can't see (deployed-build lag, or a real wire shape that differs from the test fixtures). Regardless, close the one real gap: **no test exercises the production part ordering `[text, input_audio]`** — every fixture is audio-first `[input_audio, text]`.

- Renderer collapses 1 media + ≥1 text → one bubble (`ScenarioMessageRenderer.tsx:359-372`).
- Assistant media aligns left, user aligns right (`ScenarioMessageRenderer.tsx:122-126`).
- Existing test already asserts one `<audio>` + one trace button for the collapse case.
- Coverage gap: fixtures are audio-first; real SDK is text-first (order-independent guard, but untested on the real shape).

_Status: bug / investigating / not-ready. ACs are DRAFT pending ac-reviewer. Full spec below._

## Investigation

**BLUF: both defects in the report are already fixed in the renderer on `main` (and `origin/main`).** The scenario run-viewer message renderer collapses a single `[text, input_audio]` assistant message into ONE media bubble (audio + italic transcript) and aligns `role="assistant"` turns on the LEFT. The fix landed in two commits that **predate the 2026-06-09 repro by 11–20 days**: `d631443ad` (#4058, 2026-05-20 — collapse + side logic + the `input_audio` visitor branch) and `b547d945f` (#4156, 2026-05-29 — trace button on voice turns + an integration test that explicitly asserts the one-bubble outcome). The live screenshot in the report came from diagnosing the `ai-interviewer-scenario-demo`; it was almost certainly taken against a LangWatch deployment that had not yet shipped those merges. **This issue's real work is (a) re-confirm the repro against current `main`, and (b) close a genuine coverage gap — no test exercises the production `[text, input_audio]` part ordering.**

### Renderer file:line — where side is chosen, where parts become bubbles

`langwatch/src/components/simulations/ScenarioMessageRenderer.tsx`

**(a) Side / alignment** is chosen per display item by keying on `role === "assistant"`. For the `media` case (the voice bubble), lines **122-126**:

```tsx
case "media":
  return (
    <VStack
      key={item.id}
      align={item.role === "assistant" ? "flex-start" : "flex-end"}   // ← LEFT for assistant, RIGHT otherwise
      width="100%"
    >
```

The same `flex-start`/`flex-end` keyed on `role === "assistant"` governs the `text` (lines 64-66) and `image` (lines 114-116) cases. The `role` is carried onto every display item from `msg.role` in `flattenMixed` (line 285: `const role = msg.role ?? "assistant";`) and `flattenContent` (line 278).

**(b) Multi-part content → bubbles** happens in `flattenMixed` (lines 283-375). Each content part is dispatched through `visitContentPart` (line 287) producing one `DisplayItem` per part — so a `[text, input_audio]` message first yields a `text` item + a `media` item (two items). The **collapse** then runs at lines **359-372**:

```tsx
// If a single message contains both audio AND text parts, the text is the
// transcript of the audio ... Collapse into one media item with `transcript`
// so the renderer shows audio + italic caption as one unit, not two
// separate bubbles.
const mediaItems = items.filter((it) => it.kind === "media");
const textItems = items.filter((it) => it.kind === "text");
if (mediaItems.length === 1 && textItems.length > 0) {
  const media = mediaItems[0]!;
  if (media.kind === "media") {
    media.transcript = textItems
      .map((t) => (t.kind === "text" ? t.content : ""))
      .filter(Boolean)
      .join(" ");
  }
  return items.filter((it) => it.kind !== "text");   // ← drops the text item → ONE bubble
}
```

The collapse guard is **order-independent** (it filters by `kind`, not array index), so it fires the same for `[text, input_audio]` and `[input_audio, text]`.

### Why neither defect reproduces on current `main` (the two claimed root causes are already handled)

- **Duplicate bubble** — would occur only if the collapse block (359-372) did not run. It runs whenever `mediaItems.length === 1 && textItems.length > 0`, which is exactly the `[text, input_audio]` shape. Result: the text item is filtered out and a single `media` item (with `transcript`) renders. The integration test pins this: `ScenarioMessageRenderer.integration.test.tsx:222-238` asserts `querySelectorAll("audio")` has length **1** and `getAllByTestId("trace-message")` has length **1** for an assistant `[input_audio, text]` message.
- **Role inversion** — would occur only if the alignment used a side other than `flex-start` for an assistant turn, or if `role` were lost. `role` is preserved end-to-end: the SDK sends `role:"assistant"` → the fold projection stores it verbatim (`simulationRunState.foldProjection.ts:297,306-307` — `Role: messageRole`) → the read mapper returns it 1:1 (`simulation-run.mappers.ts:113-127` — `role` mapped straight from `Messages.Role`, and `content` returned as the **array** when `Rest.content` is an array, lines 119-122). So the renderer receives `{role:"assistant", content:[…]}` and the media case aligns it `flex-start` (left).

### Full data path verified (read-side), confirming the renderer input shape

| Stage | File:line | Behavior |
|-------|-----------|----------|
| Fold (write) | `simulationRunState.foldProjection.ts:276-323` | Stores each message with `Role` verbatim; array `content` JSON-stringified to `Content` and preserved in `Rest.content` (`buildMessageRestJson`, lines 92-105). No re-split, no role rewrite. |
| Mapper (read) | `simulation-run.mappers.ts:108-127` | Reconstructs `{role, content}`; when `Rest.content` is an array, returns `content` as that array verbatim (lines 119-122). |
| Renderer | `ScenarioMessageRenderer.tsx:228-269` | `flattenMessages` → `flattenContent` → `coerceContentToArray` (handles array, JSON-string, and python-repr) → `flattenMixed` → collapse + side. |

`coerceContentToArray` (`src/server/stored-objects/coerce-content-to-array.ts`) accepts the array directly (line 17), a JSON-encoded array string, or a python-repr string — so even a legacy stringified payload reaches `flattenMixed`.

### Strategies / hypotheses considered

| Hypothesis | Verdict | Evidence |
|-----------|---------|----------|
| Renderer splits multi-part into 2 bubbles (current `main`) | **Ruled out** | Collapse block `ScenarioMessageRenderer.tsx:359-372` + integration test asserting 1 audio/1 button (test lines 222-238). |
| Renderer aligns assistant voice on the right (current `main`) | **Ruled out** | `align={role === "assistant" ? "flex-start" : "flex-end"}` at lines 122-126; role preserved through fold + mapper. |
| Repro screenshot predates the fix (stale deployment) | **Most likely** | Fix merged #4058 (2026-05-20) + #4156 (2026-05-29); repro dated 2026-06-09 but sourced from the demo screenshot. Needs live re-confirmation on current `main`. |
| Real wire shape differs from test fixtures (text-first ordering, or a `text` part the visitor mis-routes) | **Possible, untested** | Every collapse fixture is audio-first `[input_audio, text]`; production is text-first `[text, input_audio]`. Guard is order-independent in code, but the real ordering has **no test**. Also unverified: the exact runtime `text` part shape (`{type:"text", text}` vs `{type:"text", content}` — both handled by `visit-content-part.ts:124-127`, but not asserted for the assistant voice case). |
| Streaming path duplicates the bubble | **Ruled out for the persisted view** | `flattenMessages` dedupes streaming vs server by `messageId` (lines 259-265); `SimulationChatViewer.tsx:48-57` clears completed streaming once server data arrives. Live-streaming-only duplication not statically excludable (see caveats). |

- **Do not "fix" the collapse or the side logic blind** — they already implement the asked behavior. First reproduce on **current `main`** with a fresh voice run. If it does not reproduce, the issue reduces to (b) regression coverage + closing as already-fixed.
- The one concrete, actionable gap is **test coverage for the production part ordering**: add/adjust a fixture using `content` as `[{type:"text"}, {type:"input_audio"}]` (text-first) and assert one bubble + left alignment. This is the ordering the SDK actually emits and it is currently untested.
- The renderer is **shared with non-voice runs** (text, tool_call, tool_result, image). Any change must not regress those — `ScenarioMessageRenderer.integration.test.tsx:240-275` pins the text/tool trace-button behavior; keep it green.
- Alignment is asserted only indirectly today (via which branch renders / trace-button presence). There is **no explicit assertion on `flex-start` vs `flex-end`** for the media bubble — adding one would make the "left side" guarantee falsifiable in a unit test rather than relying on a browser screenshot.

### Caveats — what static reading cannot confirm (needs a live app run)

- **Whether the bug still reproduces live on current `main`.** The deps are not installed in this checkout (`node_modules/.bin/vitest` and `dotenv` absent), so the integration test was **not executed this session** — its pass is inferred from reading the assertions, not observed. A live voice run + screenshot, or a local `pnpm test:unit ScenarioMessageRenderer`, is required to confirm.
- **The exact runtime shape of the real `input_audio` and `text` parts** as they arrive at the renderer (post stored-objects extraction: `{url, mimeType}` vs pre-extraction `{data, format}`; and whether the text part is `{type:"text", text}` or carries an unexpected key). The visitor handles all known shapes, but the real assistant-voice payload was not captured this session.
- **Live-streaming duplication** (before server snapshot lands) is governed by `useSimulationStreamingState` dedup; not exercised statically.

## Acceptance Criteria

> **Sharpened by ac-reviewer** (verified against `main` @ `c38be3b6f`). **Proof mechanism:** jsdom vitest integration test (`ScenarioMessageRenderer.integration.test.tsx`, run via `pnpm test:unit`) unless noted; AC10 is a live screenshot.
>
> **Implementer notes (load-bearing):**
> - The renderer's Chakra `align` prop compiles to an atomic CSS class that `getComputedStyle` **cannot read in jsdom** (the repo documents this exact limitation at `ExpandedTextDialog.integration.test.tsx:78`, where four such tests are `it.skip`-ped). Alignment ACs therefore require a **jsdom-readable handle** — a `data-align` attribute or an inline `style.alignItems` — which is a small renderer change. The inline-`alignSelf` precedent that *does* pass jsdom is `MessagePreview.unit.test.tsx:113-143`.
> - There is **no per-bubble `data-testid`** in the renderer today. "Exactly one bubble" must be asserted via DOM ancestry (the transcript is a descendant of the `<audio>`'s container) or by adding a `data-testid="message-bubble"` per top-level item.

**Behavioral**
- **AC1 — One bubble for a multi-part voice turn.** Given `{role:"assistant", content:[{type:"text", text:T}, {type:"input_audio", input_audio:{data,…}}]}`: exactly one `<audio>`; T renders as the italic caption **inside the same wrapper as the `<audio>`**; **no standalone text bubble** for that message; the message yields one top-level list child. Evidence: jsdom asserts `querySelectorAll("audio").length===1`, `getByText(T)` is a descendant of the `<audio>`'s container, and no second sibling bubble exists. (As-drafted `audio-count===1` + `T-once` is red-incapable: a single audio part yields one `<audio>` whether collapsed OR split, and T appears once either way — the ancestry + no-second-bubble assertion is the real discriminator.)
- **AC2 — Assistant voice turn aligns left.** The media bubble exposes a jsdom-readable left handle for `role==="assistant"` (`data-align="start"` or inline `style.alignItems==="flex-start"` — NOT the class-based `align` prop). Evidence: jsdom asserts the handle === left value; fails if absent or === user value.
- **AC3 — Transcript + audio co-contained.** `getByText(T)` and the `<audio>` share the inner media container (`ScenarioMessageRenderer.tsx:129-148`). Evidence: ancestry assertion. Fails if T is outside the audio's container.
- **AC4 — Simulated-user turns align right.** Same handle as AC2, `role==="user"` → `data-align="end"` / `flex-end`. Evidence: jsdom asserts right value; pairs with AC2 to pin the inversion both ways.

**Regression**
- **AC5 — Text-only turns unchanged.** Plain-string assistant content → one left text bubble; user → right. Evidence: `ScenarioMessageRenderer.integration.test.tsx:129-142` stays green **plus** the AC2-style left/right handle asserted (not `getComputedStyle`).
- **AC6 — Tool turns unchanged.** `tool_call`/`tool_result` render unchanged, one View-Trace button each where applicable. Evidence: `ScenarioMessageRenderer.integration.test.tsx:240-275` stays green.
- **AC12 — Image turns unchanged.** Assistant `image_url`/bare-image → one left `` bubble; user → right (shares the same `align` expression at `ScenarioMessageRenderer.tsx:112-119`). Evidence: jsdom asserts one `` + AC2 handle.

**Failure modes**
- **AC7 — Both part orderings collapse.** AC1's assertion body holds for `[text, input_audio]` AND `[input_audio, text]`. Evidence: parametrized over both orderings (not just "one `<audio>` each", which does not falsify duplication).
- **AC8 — Text-part key variants.** Collapses for `{type:"text", text:T}` and `{type:"text", content:T}` (both accepted by `visit-content-part.ts:124-126`). Evidence: parametrized over both shapes. **Out of scope (stated decision):** a bare `{content:T}` part with no `type` falls through to the `unknown` branch at `visit-content-part.ts:124` and is dropped — add a visitor branch + AC only if the production SDK can emit that shape.
- **AC9 — Audio-only, no transcript.** Single `input_audio` part, no sibling text → one left audio bubble, **no** italic caption node (inner `VStack` has exactly one child). Evidence: jsdom asserts one `<audio>`, AC2 handle, transcript `queryByText` === null.
- **AC13 — Missing-role default.** A content-parts message with no `role` key → renders assistant/left, one bubble (matches `flattenMixed` default at `ScenarioMessageRenderer.tsx:285`). Evidence: jsdom with `role` omitted asserts the assistant/left handle and one bubble.
- **AC14 — Streaming/snapshot dedup for voice turns.** A voice turn present in BOTH `streamingMessages` and `messages` renders once (no transient duplicate) when the snapshot arrives (`flattenMessages` dedups by `messageId` at `ScenarioMessageRenderer.tsx:259-265`). Evidence: jsdom rendering both a streaming entry and a server-message entry for the same logical turn asserts one bubble. **If their `messageId`s differ in production, this is the prime non-stale-repro root cause** — capture both IDs from a real run and assert equality (or fix the dedup to key on a shared id). Mark out-of-scope only as an explicit decision, since this is the most likely explanation for a repro that is NOT just a stale deployment.

**Live verification + fix arm**
- **AC10 — Live re-confirmation on a pinned build.** A fresh realtime-audio Scenario run in the viewer shows each agent voice turn as one left bubble (audio **decoded & playable**, transcript present), user turns right. Evidence: screenshot of the SUCCEEDED player state **paired with the deployed build's commit SHA** (app footer/health endpoint, or a local run off a quoted `git rev-parse HEAD`) proving the build is post-#4058/#4156. A correct-layout screenshot whose `<audio>` shows `audio/pcm16 [error]` (cannot decode) FAILS this AC.
- **AC11 — Bug-still-live arm.** If AC10 reproduces either defect on the pinned build, produce a jsdom test from the captured real wire shape that fails pre-fix and passes after the minimal renderer/visitor fix, with AC5/AC6/AC12 still green. Evidence: quoted red (pre-fix commit) and green (post-fix) `pnpm test:unit` runs. Vacuously satisfied — and the issue closes on AC5–AC9/AC12–AC14 + a stale-repro note — if AC10 reproduces-negative; state explicitly which arm fired.

## Plan

> **DRAFT — pending ac-reviewer + plan-step.** Outline only; the plan step refines.

1. **Reproduce-or-refute first.** Install app deps and run `pnpm test:unit src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx` to observe current behavior; if possible, do a live voice run on current `main` and screenshot the viewer. Decide: is this a still-live bug, or already fixed (stale repro)?
2. **Capture the real wire shape.** From a live voice run, capture the actual `input_audio` + `text` content parts as they reach the renderer (post-extraction). Confirm ordering and the text-part key.
3. **If still reproduces:** localize to whichever guard does not fire for the real shape (collapse condition, role propagation, or a visitor branch that mis-routes the real `text`/`input_audio` part) and fix it minimally — without regressing the text/tool/image branches.
4. **Close the coverage gap regardless:** add the text-first `[text, input_audio]` fixture and an explicit `flex-start`/`flex-end` alignment assertion to `ScenarioMessageRenderer.integration.test.tsx`; parametrize ordering and text-part shape (AC7/AC8); add the audio-only no-transcript case (AC9).
5. **If already fixed:** keep the new regression tests, note the original repro was a stale (pre-merge) deployment, and close.

</details>

---
*Classification: Bug*
*Status: investigating — ACs DRAFT pending ac-reviewer. Issue stays not-ready until plan + spec steps complete.*

### Original report

<details><summary>Original stub (preserved verbatim)</summary>

**Describe the bug**
In the LangWatch app's **scenario / simulation run viewer**, voice (realtime audio) **agent turns render incorrectly** in two ways:

1. **Role inversion** — the agent-under-test's spoken turn is rendered on the **user / simulator side (right)** instead of the **agent side (left)**. The agent's question appears as if the simulated user said it.
2. **Duplicate bubble** — a single agent message whose `content` is **two parts** (a `text` transcript part **and** an `input_audio` part) is rendered as **two separate message bubbles**, each with its own audio player, instead of one combined bubble.

**To reproduce**
1. Run a Scenario **voice / realtime** scenario where the agent under test is a realtime-audio adapter (e.g. `OpenAIRealtimeAgentAdapter`).
2. Open the resulting run in the LangWatch app's scenario run viewer.
3. Observe the agent's spoken turns: each appears **twice** (once per content part) and on the **wrong (user/right) side**.

**Expected behavior**
A voice agent turn renders as **one bubble on the agent (left) side**, containing the transcript text and its audio player together. The simulated user's turns render on the right.

**Environment**
- Component: langwatch (app frontend — scenario run-viewer message renderer)
- Version: reproduced against Scenario SDK message-snapshot stream (`SCENARIO_MESSAGE_SNAPSHOT`) as of 2026-06-09
- Browser/OS: web app (LangWatch UI)

**Additional context**
**Verified evidence that this is a frontend-only rendering defect, not an SDK defect** (gathered 2026-06-09 while diagnosing a voice-interviewer demo built on the Scenario Python SDK):

- The Scenario Python SDK emits **exactly one message per agent turn** with `role="assistant"` and `content=[{type: "text", ...}, {type: "input_audio", ...}]` — confirmed by a live hermetic run whose `result.messages` was `{count: 1, role: "assistant", 2 parts}`, and whose `SCENARIO_MESSAGE_SNAPSHOT` wire payload (the data the app frontend consumes) carried that single, correctly-roled message.
- The SDK's event serializer (`scenario/_events/utils.py`) preserves `role` 1:1 and does **not** split a multi-part message into multiple messages.
- Therefore both defects — the split-into-two-bubbles and the side-inversion — are introduced by the **app frontend's scenario-run-viewer message renderer** when it maps a single multi-part `assistant` voice message onto the UI.

Likely fix surface (for the investigation step to confirm): the frontend scenario-run-viewer message renderer should (a) collapse a single message's `[text, input_audio]` parts into one bubble, and (b) place `role="assistant"` voice turns on the agent (left) side. Found while diagnosing the `ai-interviewer-scenario-demo` voice interviewer; a screenshot showed the interviewer's spoken question duplicated and rendered on the candidate side.

</details>

drewdrewthis:
## Why

Closes #4138, Closes #4698

Two render bugs in the simulation message UI needed regression fixtures before they could be considered fixed. Without pinned tests, the same breakage can silently re-emerge:

- **#4138** – `input_audio` messages arriving with a `url` shape (instead of raw bytes) were passed straight through to the renderer as an opaque object, producing a raw blob-dump instead of an `<audio>` element.
- **#4698** – Two separate regressions: (a) the production SDK emits audio+transcript turns **text-first** (`[text, input_audio]`), but the collapse guard only handled **audio-first** order, producing a duplicate bubble; (b) user-role voice turns were incorrectly left-aligned (assistant side) instead of right-aligned (user side); (c) audio-only turns emitted a ghost empty-transcript `<p>` node.

## What changed

- **`ScenarioMessageRenderer.integration.test.tsx`** — New describe blocks cover: url-shape `input_audio` coercion (12 cases), text-first vs audio-first collapse ordering, user-role alignment, and audio-only no-ghost-artifact.
- **`ScenarioRunDetailDrawer.integration.test.tsx`** — Extended with smoke coverage of the drawer mounting the renderer with url-shaped content.
- **`ScenarioMessageRenderer.tsx`** — Added `data-align` test affordance mirroring the `align` prop on media VStacks. Chakra's `align` compiles to an atomic CSS class that jsdom's `getComputedStyle` cannot read; this attribute lets tests assert left/right alignment without CSS evaluation. **No behavior change.**
- **`specs/features/scenarios/voice-message-render-regressions.feature`** — Feature file binding the three new `@scenario` annotations required by the feature-parity CI check.

## How it works

The test affordance (`data-align={item.role === "assistant" ? "flex-start" : "flex-end"}`) mirrors the Chakra `align` prop already present on every media wrapper. Tests use `closest("[data-align]")` to find the wrapper and assert its value. This is the canonical jsdom workaround for Chakra's atomic-CSS compilation.

## Test plan

```bash
cd langwatch
pnpm test:unit src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
pnpm test:unit src/components/simulations/__tests__/ScenarioRunDetailDrawer.integration.test.tsx
```

All scenarios pass locally. The feature-parity check (`pnpm check:feature-parity`) passes with the new feature file.

## Anything surprising?

`langwatch-app-ci` shows a pre-existing failure in `ee/licensing/__tests__/licenseGenerationService.unit.test.ts` that also fails on `main` — unrelated to this PR. All other CI checks pass.

Ran locally on HEAD `51e310a81b76f7f5f91ac2df9ee1b293e19e6e3f`. Actual `prettyDOM` output from the live test run (not a template):

```html
<!-- #4138 url-shape: what ScenarioMessageRenderer renders for input_audio with url -->
<div
  class="chakra-stack css-5crs4f"
  data-align="flex-start"
>
  <div class="chakra-stack css-1qe2nbt">
    <div class="chakra-stack css-5crs4f">
      <audio
        controls=""
        data-testid="media-part-audio"
        src="/api/files/test-id"
        style="width: 100%; max-width: 400px;"
      />
    </div>
  </div>
</div>
```

`<audio src="/api/files/test-id">` is rendered — not a raw blob dump. All DOM-level assertions PASS:
- `screen.getByTestId("media-part-audio")` → found
- `audio.getAttribute("src")` → `"/api/files/test-id"`
- `wrapper.getAttribute("data-align")` → `"flex-start"` (assistant, left)
- `screen.queryByText(/input_audio/)` → `null` (no raw text leak)

Full verbose run output and #4698 assertions: [prove-it comment](https://github.com/langwatch/langwatch/pull/4717#issuecomment-4711639236)

## Release Notes

* **Bug Fixes**
  * Improved simulation voice message rendering so audio parts (including `input_audio` URL payloads) display correctly without leaking raw `input_audio` text.
  * Fixed multi-part voice turns to collapse into a single bubble with correct left/right alignment for both assistant and user roles.

* **Tests**
  * Added integration coverage for media coercion/rendering edge cases and unsupported MIMEType fallbacks.
  * Added regression scenarios validating transcript co-location with audio and deterministic alignment behavior.

---

## 📸 Visual proof — real running app (browser QA)

Re-captured by driving the **actual `ScenarioMessageRenderer` in the booted LangWatch app** — the earlier shots in this PR were an isolated mini-render harness and have been replaced. A voice scenario run was seeded through the real `POST /api/scenario-events` ingestion path, persisted to ClickHouse, read back via `scenarios.getRunState`, and rendered in the live `ScenarioRunDetailDrawer`. A single real conversation exercises every #4138 / #4698 case.

DOM asserted in the running app before capture: **3× native `<audio>` (`media-part-audio`)**, **1× `media-part-binary` download fallback**, alignment **user → `flex-end` / assistant → `flex-start`**, the transcript co-located inside the collapsed turn, and **no raw `input_audio` text leak**.

| Conversation drawer — Light | Conversation drawer — Dark |
|---|---|
| real app, light | real app, dark |

**What the turns prove**

- **#4138** — each voice turn renders a native `<audio>` player from the url-shape `input_audio` (not a raw blob dump); the unsupported-mimeType part degrades to the `voice-turn.bin` download link rather than a broken `<audio>`.
- **#4698** — the assistant text-first turn (`[text, input_audio]`) collapses into one assistant-left bubble with the text as an italic transcript; the simulated-user voice turn aligns **right**; the audio-only turn shows **no** empty-transcript artifact.

<sub>Full real-app context (sidebar, run grid, open drawer): [light](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4717/real-app/app-context-light.png) · [dark](https://raw.githubusercontent.com/langwatch/pr-screenshots/main/pr-4717/real-app/app-context-dark.png)</sub>

<sub>QA-seeding note: the audio `src` is a self-contained playable WAV data-URI so the native players render without provisioning external object storage — this drives the identical renderer url-shape branch (`part.url` → `<audio src>`) as a production `/api/files/<id>` reference.</sub>

### DIFF
--- a/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
+++ b/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
@@ -1,6 +1,5 @@
 import { Box, HStack, Image, Text, VStack } from "@chakra-ui/react";
-import { useMemo, useRef, useEffect } from "react";
-import { useSequentialAudioPlayback } from "./useSequentialAudioPlayback";
+import { useEffect, useMemo, useRef } from "react";
 import { Settings } from "react-feather";
 import type { StreamingMessage } from "~/hooks/useSimulationStreamingState";
 import type { ScenarioMessageSnapshotEvent } from "~/server/scenarios/scenario-event.types";
@@ -9,9 +8,10 @@ import { visitContentPart } from "~/server/stored-objects/visit-content-part";
 import { TraceMessage } from "../copilot-kit/TraceMessage";
 import { Markdown } from "../Markdown";
 import { RenderInputOutput } from "../traces/RenderInputOutput";
-import { safeJsonParseOrStringFallback } from "./utils/safe-json-parse-or-string-fallback";
-import { MediaPart } from "./MediaPart";
 import type { MediaPartData } from "./MediaPart";
+import { MediaPart } from "./MediaPart";
+import { useSequentialAudioPlayback } from "./useSequentialAudioPlayback";
+import { safeJsonParseOrStringFallback } from "./utils/safe-json-parse-or-string-fallback";
 
 type RawMessage = ScenarioMessageSnapshotEvent["messages"][number];
 
@@ -25,10 +25,29 @@ const textAlignForRole = (role?: string): "left" | "right" =>
   role === "assistant" ? "left" : "right";
 
 type DisplayItem =
-  | { kind: "text"; id: string; role: string; content: string; traceId?: string }
+  | {
+      kind: "text";
+      id: string;
+      role: string;
+      content: string;
+      traceId?: string;
+    }
   | { kind: "image"; id: string; src: string; role?: string; traceId?: string }
-  | { kind: "media"; id: string; part: MediaPartData; role?: string; transcript?: string; traceId?: string }
-  | { kind: "tool_call"; id: string; name: string; arguments: unknown; traceId?: string }
+  | {
+      kind: "media";
+      id: string;
+      part: MediaPartData;
+      role?: string;
+      transcript?: string;
+      traceId?: string;
+    }
+  | {
+      kind: "tool_call";
+      id: string;
+      name: string;
+      arguments: unknown;
+      traceId?: string;
+    }
   | { kind: "tool_result"; id: string; result: unknown; traceId?: string };
 
 interface ScenarioMessageRendererProps {
@@ -213,14 +232,27 @@ export function ScenarioMessageRenderer({
                   borderRadius="lg"
                   p={3}
                 >
-                  <Text fontSize="xs" fontWeight="semibold" color="fg.muted" mb={2}>
+                  <Text
+                    fontSize="xs"
+                    fontWeight="semibold"
+                    color="fg.muted"
+                    mb={2}
+                  >
                     Tool arguments
                   </Text>
-                  <Box bg="bg.panel" border="1px solid" borderColor="border" borderRadius="md" p={2}>
+                  <Box
+                    bg="bg.panel"
+                    border="1px solid"
+                    borderColor="border"
+                    borderRadius="md"
+                    p={2}
+                  >
                     <RenderInputOutput value={item.arguments as string} />
                   </Box>
                 </Box>
-                {!smallerView && item.traceId && <TraceMessage traceId={item.traceId} />}
+                {!smallerView && item.traceId && (
+                  <TraceMessage traceId={item.traceId} />
+                )}
               </VStack>
             );
 
@@ -236,14 +268,27 @@ export function ScenarioMessageRenderer({
                   borderRadius="lg"
                   p={3}
                 >
-                  <Text fontSize="xs" fontWeight="semibold" color="fg.muted" mb={2}>
+                  <Text
+                    fontSize="xs"
+                    fontWeight="semibold"
+                    color="fg.muted"
+                    mb={2}
+                  >
                     Tool result
                   </Text>
-                  <Box bg="bg.panel" border="1px solid" borderColor="border" borderRadius="md" p={2}>
+                  <Box
+                    bg="bg.panel"
+                    border="1px solid"
+                    borderColor="border"
+                    borderRadius="md"
+                    p={2}
+                  >
                     <RenderInputOutput value={item.result as string} />
                   </Box>
                 </Box>
-                {!smallerView && item.traceId && <TraceMessage traceId={item.traceId} />}
+                {!smallerView && item.traceId && (
+                  <TraceMessage traceId={item.traceId} />
+                )}
               </VStack>
             );
 
@@ -270,16 +315,23 @@ function flattenMessages(
     if (msg.role === "user" || msg.role === "assistant") {
       // Support both snake_case (OpenAI/chatMessageSchema) and camelCase (AG-UI MessageSchema)
       const msgAny = msg as Record<string, unknown>;
-      const toolCalls = (msgAny.tool_calls as Array<{ function?: { name?: string; arguments?: string } }> | undefined)
-        ?? (msgAny.toolCalls as Array<{ function?: { name?: string; arguments?: string } }> | undefined)
-        ?? null;
+      const toolCalls =
+        (msgAny.tool_calls as
+          | Array<{ function?: { name?: string; arguments?: string } }>
+          | undefined) ??
+        (msgAny.toolCalls as
+          | Array<{ function?: { name?: string; arguments?: string } }>
+          | undefined) ??
+        null;
       if (toolCalls) {
         for (const tc of toolCalls) {
           items.push({
             kind: "tool_call",
             id: `${msg.id ?? ""}-tool-${tc.function?.name ?? "unknown"}`,
             name: tc.function?.name ?? "unknown",
-            arguments: safeJsonParseOrStringFallback(tc.function?.arguments ?? "{}"),
+            arguments: safeJsonParseOrStringFallback(
+              tc.function?.arguments ?? "{}",
+            ),
             traceId: msg.trace_id,
           });
         }
@@ -290,7 +342,9 @@ function flattenMessages(
         kind: "tool_result",
         id: msg.id ?? crypto.randomUUID(),
         result: safeJsonParseOrStringFallback(
-          typeof msg.content === "string" ? msg.content : JSON.stringify(msg.content ?? {}),
+          typeof msg.content === "string"
+            ? msg.content
+            : JSON.stringify(msg.content ?? {}),
         ),
         traceId: msg.trace_id,
       });
@@ -301,7 +355,12 @@ function flattenMessages(
     const serverIds = new Set(messages.map((m) => m.id).filter(Boolean));
     for (const sm of streamingMessages) {
       if (serverIds.has(sm.messageId)) continue;
-      items.push({ kind: "text", id: sm.messageId, role: sm.role, content: sm.content || "\u2026" });
+      items.push({
+        kind: "text",
+        id: sm.messageId,
+        role: sm.role,
+        content: sm.content || "\u2026",
+      });
     }
   }
 
@@ -313,10 +372,21 @@ function flattenContent(msg: RawMessage): DisplayItem[] {
   const coerced = coerceContentToArray(msg.content);
   if (coerced) return flattenMixed(coerced, msg);
 
-  const raw = typeof msg.content === "string" ? msg.content : JSON.stringify(msg.content ?? {});
+  const raw =
+    typeof msg.content === "string"
+      ? msg.content
+      : JSON.stringify(msg.content ?? {});
 
   if (msg.content && msg.content !== "None") {
-    return [{ kind: "text", id: msg.id ?? crypto.randomUUID(), role: msg.role ?? "assistant", content: raw, traceId: msg.trace_id }];
+    return [
+      {
+        kind: "text",
+        id: msg.id ?? crypto.randomUUID(),
+        role: msg.role ?? "assistant",
+        content: raw,
+        traceId: msg.trace_id,
+      },
+    ];
   }
   return [];
 }
@@ -326,9 +396,16 @@ function flattenMixed(content: unknown[], msg: RawMessage): DisplayItem[] {
   const role = msg.role ?? "assistant";
   content.forEach((item, i) => {
     const result = visitContentPart<DisplayItem | undefined>(item, {
-      text: (text) => text
-        ? { kind: "text" as const, id: `${msg.id}-c${i}`, role, content: text, traceId: msg.trace_id }
-        : undefined,
+      text: (text) =>
+        text
+          ? {
+              kind: "text" as const,
+              id: `${msg.id}-c${i}`,
+              role,
+              content: text,
+              traceId: msg.trace_id,
+            }
+          : undefined,
       media: (part) => ({
         kind: "media" as const,
         id: `${msg.id}-media${i}`,
@@ -345,10 +422,33 @@ function flattenMixed(content: unknown[], msg: RawMessage): DisplayItem[] {
         role,
         traceId: msg.trace_id,
       }),
-      toolCall: (part) => ({ kind: "tool_call" as const, id: `${msg.id}-tu${i}`, name: part.name, arguments: part.arguments, traceId: msg.trace_id }),
-      toolResult: (part) => ({ kind: "tool_result" as const, id: `${msg.id}-tr${i}`, result: part.result, traceId: msg.trace_id }),
-      imageUrl: (url) => ({ kind: "image" as const, id: `${msg.id}-img${i}`, src: url, role, traceId: msg.trace_id }),
-      bareImage: (src) => ({ kind: "image" as const, id: `${msg.id}-img${i}`, src, role, traceId: msg.trace_id }),
+      toolCall: (part) => ({
+        kind: "tool_call" as const,
+        id: `${msg.id}-tu${i}`,
+        name: part.name,
+        arguments: part.arguments,
+        traceId: msg.trace_id,
+      }),
+      toolResult: (part) => ({
+        kind: "tool_result" as const,
+        id: `${msg.id}-tr${i}`,
+        result: part.result,
+        traceId: msg.trace_id,
+      }),
+      imageUrl: (url) => ({
+        kind: "image" as const,
+        id: `${msg.id}-img${i}`,
+        src: url,
+        role,
+        traceId: msg.trace_id,
+      }),
+      bareImage: (src) => ({
+        kind: "image" as const,
+        id: `${msg.id}-img${i}`,
+        src,
+        role,
+        traceId: msg.trace_id,
+      }),
       // OpenAI Realtime API audio shape. Two states:
       // - Pre-extraction: {data, format} (inline base64). Server-side
       //   extraction normally rewrites this away, but if the renderer
--- a/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
+++ b/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
@@ -15,6 +15,15 @@ import type { MediaPartData } from "./MediaPart";
 
 type RawMessage = ScenarioMessageSnapshotEvent["messages"][number];
 
+// Role → alignment mapping. Extracted here so `align` and `data-align` always
+// derive from the same value — the `data-align` attribute mirrors `align` for
+// jsdom tests, which cannot read Chakra's atomic CSS classes via getComputedStyle.
+const alignForRole = (role?: string): "flex-start" | "flex-end" =>
+  role === "assistant" ? "flex-start" : "flex-end";
+
+const textAlignForRole = (role?: string): "left" | "right" =>
+  role === "assistant" ? "left" : "right";
+
 type DisplayItem =
   | { kind: "text"; id: string; role: string; content: string; traceId?: string }
   | { kind: "image"; id: string; src: string; role?: string; traceId?: string }
@@ -85,12 +94,8 @@ export function ScenarioMessageRenderer({
             return (
               <VStack
                 key={item.id}
-                align={item.role === "assistant" ? "flex-start" : "flex-end"}
-                // Test affordance: Chakra's `align` prop compiles to an atomic
-                // CSS class jsdom's getComputedStyle cannot read, so the
-                // left/right alignment is asserted via this mirrored attribute.
-                // Same value as `align` above — no behavior change.
-                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                align={alignForRole(item.role)}
+                data-align={alignForRole(item.role)}
                 gap={1}
               >
                 {item.role === "assistant" ? (
@@ -140,9 +145,8 @@ export function ScenarioMessageRenderer({
             return (
               <VStack
                 key={item.id}
-                align={item.role === "assistant" ? "flex-start" : "flex-end"}
-                // Test affordance — see the `text` case. Mirrors `align`.
-                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                align={alignForRole(item.role)}
+                data-align={alignForRole(item.role)}
               >
                 <Image src={item.src} maxH="200px" borderRadius="md" />
               </VStack>
@@ -152,9 +156,8 @@ export function ScenarioMessageRenderer({
             return (
               <VStack
                 key={item.id}
-                align={item.role === "assistant" ? "flex-start" : "flex-end"}
-                // Test affordance — see the `text` case. Mirrors `align`.
-                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                align={alignForRole(item.role)}
+                data-align={alignForRole(item.role)}
                 width="100%"
               >
                 <VStack
@@ -177,9 +180,7 @@ export function ScenarioMessageRenderer({
                       color="fg.muted"
                       fontStyle="italic"
                       paddingX={2}
-                      textAlign={
-                        item.role === "assistant" ? "left" : "right"
-                      }
+                      textAlign={textAlignForRole(item.role)}
                     >
                       {item.transcript}
                     </Text>
--- a/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
+++ b/langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
@@ -86,6 +86,11 @@ export function ScenarioMessageRenderer({
               <VStack
                 key={item.id}
                 align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                // Test affordance: Chakra's `align` prop compiles to an atomic
+                // CSS class jsdom's getComputedStyle cannot read, so the
+                // left/right alignment is asserted via this mirrored attribute.
+                // Same value as `align` above — no behavior change.
+                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
                 gap={1}
               >
                 {item.role === "assistant" ? (
@@ -136,6 +141,8 @@ export function ScenarioMessageRenderer({
               <VStack
                 key={item.id}
                 align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                // Test affordance — see the `text` case. Mirrors `align`.
+                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
               >
                 <Image src={item.src} maxH="200px" borderRadius="md" />
               </VStack>
@@ -146,6 +153,8 @@ export function ScenarioMessageRenderer({
               <VStack
                 key={item.id}
                 align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                // Test affordance — see the `text` case. Mirrors `align`.
+                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
                 width="100%"
               >
                 <VStack
--- a/langwatch/src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
+++ b/langwatch/src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
@@ -11,8 +11,8 @@
 import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
 import { cleanup, render, screen } from "@testing-library/react";
 import { afterEach, beforeAll, describe, expect, it, vi } from "vitest";
-import { ScenarioMessageRenderer } from "../ScenarioMessageRenderer";
 import type { ScenarioMessageSnapshotEvent } from "~/server/scenarios/scenario-event.types";
+import { ScenarioMessageRenderer } from "../ScenarioMessageRenderer";
 
 vi.mock("~/utils/api", () => ({
   api: {
@@ -26,7 +26,9 @@ vi.mock("~/utils/api", () => ({
 
 vi.mock("../../copilot-kit/TraceMessage", () => ({
   TraceMessage: ({ traceId }: { traceId: string }) => (
-    <button data-testid="trace-message" data-trace-id={traceId}>View Trace</button>
+    <button data-testid="trace-message" data-trace-id={traceId}>
+      View Trace
+    </button>
   ),
 }));
 
@@ -298,9 +300,7 @@ describe("<ScenarioMessageRenderer/>", () => {
     it("renders a media-part-audio element whose src is the file url (drawer variant)", () => {
       renderWith([urlShapeMessage]);
 
-      const audio = screen.getByTestId(
-        "media-part-audio",
-      ) as HTMLAudioElement;
+      const audio = screen.getByTestId("media-part-audio") as HTMLAudioElement;
       expect(audio).toBeInTheDocument();
       expect(audio.tagName.toLowerCase()).toBe("audio");
       expect(audio).toHaveAttribute("src", "/api/files/test-id");
@@ -312,9 +312,7 @@ describe("<ScenarioMessageRenderer/>", () => {
     it("renders a media-part-audio element whose src is the file url (grid variant)", () => {
       renderWithGrid([urlShapeMessage]);
 
-      const audio = screen.getByTestId(
-        "media-part-audio",
-      ) as HTMLAudioElement;
+      const audio = screen.getByTestId("media-part-audio") as HTMLAudioElement;
       expect(audio).toBeInTheDocument();
       expect(audio).toHaveAttribute("src", "/api/files/test-id");
       expect(audio).toHaveAttribute("controls");
@@ -435,7 +433,10 @@ describe("<ScenarioMessageRenderer/>", () => {
           content: [
             {
               type: "input_audio",
-              input_audio: { url: "/api/files/user-voice", mimeType: "audio/mpeg" },
+              input_audio: {
+                url: "/api/files/user-voice",
+                mimeType: "audio/mpeg",
+              },
             },
           ],
         } as unknown as ScenarioMessageSnapshotEvent["messages"][number],
@@ -457,7 +458,10 @@ describe("<ScenarioMessageRenderer/>", () => {
           content: [
             {
               type: "input_audio",
-              input_audio: { url: "/api/files/solo-voice", mimeType: "audio/mpeg" },
+              input_audio: {
+                url: "/api/files/solo-voice",
+                mimeType: "audio/mpeg",
+              },
             },
           ],
         } as unknown as ScenarioMessageSnapshotEvent["messages"][number],
--- a/langwatch/src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
+++ b/langwatch/src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
@@ -470,12 +470,10 @@ describe("<ScenarioMessageRenderer/>", () => {
       const mediaWrapper = audio.closest("[data-align]");
       expect(mediaWrapper).toHaveAttribute("data-align", "flex-start");
 
-      // No transcript artifact: the inner media container holds only the
-      // <audio> (its VStack), no italic caption <p>. Asserting the audio's
-      // immediate container has a single child guards against an empty
-      // transcript node sneaking in.
-      const innerContainer = audio.parentElement?.parentElement;
-      expect(innerContainer?.childElementCount).toBe(1);
+      // No transcript artifact: the bubble must not render an italic caption
+      // <p>. Scoped to the media wrapper so Chakra wrappers outside the bubble
+      // never false-positive; does not depend on wrapper nesting depth.
+      expect(mediaWrapper!.querySelector("p")).toBeNull();
     });
   });
 });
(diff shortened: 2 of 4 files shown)


langwatch/src/components/simulations/ScenarioMessageRenderer.tsx
```diff
@@ -86,6 +86,11 @@ export function ScenarioMessageRenderer({
               <VStack
                 key={item.id}
                 align={item.role === "assistant" ? "flex-start" : "flex-end"}
+                // Test affordance: Chakra's `align` prop compiles to an atomic
+                // CSS class jsdom's getComputedStyle cannot read, so the
+                // left/right alignment is asserted via this mirrored attribute.
+                // Same value as `align` above — no behavior change.
+                data-align={item.role === "assistant" ? "flex-start" : "flex-end"}
```
drewdrewthis: **[principles][hygiene][uncle-bob][fowler][metz-beck] Duplicated role→alignment ternary (6 copies) — extract `alignForRole` helper.** The expression `item.role === "assistant" ? "flex-start" : "flex-end"` appears at lines 88/93, 143/145, and 155/157 (three branches × two props), plus a sibling `"left"/"right"` variant at line 180. This is a missing domain concept encoded as a raw ternary scattered through JSX. Shotgun surgery risk: add one role, edit 6+ sites. The `data-align` and `align` values can also silently drift since nothing enforces they stay equal.

**Fix:** Extract `const alignForRole = (role?: string): "flex-start" | "flex-end" => role === "assistant" ? "flex-start" : "flex-end"` and replace all call sites with `align={alignForRole(item.role)} data-align={alignForRole(item.role)}`. Both props then derive from one function call, making drift impossible.

langwatch/src/components/simulations/__tests__/ScenarioMessageRenderer.integration.test.tsx
```diff
@@ -273,4 +273,209 @@ describe("<ScenarioMessageRenderer/>", () => {
...
+
+      const audio = screen.getByTestId("media-part-audio");
+      const mediaWrapper = audio.closest("[data-align]");
+      expect(mediaWrapper).toHaveAttribute("data-align", "flex-start");
+
+      // No transcript artifact: the inner media container holds only the
+      // <audio> (its VStack), no italic caption <p>. Asserting the audio's
+      // immediate container has a single child guards against an empty
+      // transcript node sneaking in.
+      const innerContainer = audio.parentElement?.parentElement;
```
drewdrewthis: **[test][fowler] Brittle DOM-depth traversal — replace with behavioral assertion.** `audio.parentElement?.parentElement.childElementCount === 1` navigates two anonymous wrapper levels. If Chakra's VStack adds or removes a wrapper div (after a version bump), this breaks for structural reasons unrelated to behavior, and the failure message gives no hint what changed.

**Fix:** Assert the absence of the transcript element directly:
```ts
// no italic transcript rendered in audio-only turn
expect(screen.queryByRole('paragraph')).toBeNull();
```
This survives wrapper-nesting refactors and communicates the intent clearly.

langwatch/src/components/simulations/__tests__/ScenarioRunDetailDrawer.integration.test.tsx
```diff
@@ -137,4 +160,52 @@ describe("ScenarioRunDetailDrawer", () => {
...
+  //
+  // The drawer mounts <ScenarioMessageRenderer variant="drawer" /> inside its
+  // Drawer.Content (ScenarioRunDetailDrawer.tsx:386-393). This pins that a
+  // voice turn carrying the post-extraction url shape renders its native
+  // <audio> when composed inside the real Drawer container — the drawer-side
+  // companion to the grid-variant fixture in
+  // ScenarioMessageRenderer.integration.test.tsx.
+  // -------------------------------------------------------------------------
+  describe("ScenarioMessageRenderer (drawer variant) for an input_audio url turn (#4138)", () => {
+    beforeAll(() => {
```
drewdrewthis: **[hygiene][test] `beforeAll(scrollIntoView)` scoped inside wrong describe — hoist to outer scope.** The stub is inside `describe("ScenarioMessageRenderer (drawer variant)...")` at line 173. All other files mounting ScenarioMessageRenderer (e.g. ScenarioMessageRenderer.integration.test.tsx:69) place it at file or outer-describe level. If any earlier test in this file triggers a scroll, the suite fails with `scrollIntoView is not a function` before reaching this block.

**Fix:** Move `beforeAll` to the outer `describe("ScenarioRunDetailDrawer", ...)` block (around line 53) or to module level.

drewdrewthis (commented):
[/review] Blocking concerns: 3 findings requiring fixes before merge.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `51e310a81b76f7f5f91ac2df9ee1b293e19e6e3f` · **Run:** /review --since 61c77026b

### Non-blocking

- **[hygiene][fowler]** `CustomCopilotKitChat.tsx:112` has the same role→alignment ternary inline — duplicated knowledge across sibling files. **Decide:** extract to a shared `simulations/roleStyles.ts` util if/when a third site appears, per Metz-Beck's rule of three. Don't export from the renderer.
- **[metz-beck]** `textAlignForRole` has exactly one call site — single-use abstraction. **Decide:** inline the ternary or keep for naming symmetry with `alignForRole`. No correctness issue.
- **[uncle-bob]** Pre-existing: `ScenarioMessageRenderer` is a 175-line render function with 5 switch cases (5 reasons to change — text/image/media/tool_call/tool_result). **New Issue:** extract `<TextBubble>`, `<MediaBubble>`, `<ToolCallCard>` components.
- **[test][fowler]** Pre-existing: `Element.prototype.scrollIntoView` stub has no `afterAll` restore — leaks across test files sharing the same jsdom global. **New Issue:** sweep simulations test suite, centralize stub in test setup.

drewdrewthis:
## Prove-it: Step 2.7 — Live input_audio flow demo

**HEAD:** `51e310a81b76f7f5f91ac2df9ee1b293e19e6e3f`
**Run date:** 2026-06-15

---

### Rendered DOM (actual jsdom output — #4138 url-shape case)

> Output from `pnpm test:integration ScenarioMessageRenderer.integration.test.tsx` with `prettyDOM(audio.closest("[data-align]"))` called inside the passing test:

```html
<div
  class="chakra-stack css-5crs4f"
  data-align="flex-start"    ← assistant left-align ✓
>
  <div
    class="chakra-stack css-1qe2nbt"
  >
    <div
      class="chakra-stack css-5crs4f"
    >
      <audio                 ← <audio> rendered ✓  (not raw blob dump)
        controls=""
        data-testid="media-part-audio"
        src="/api/files/test-id"    ← url-shape src resolved ✓
        style="width: 100%; max-width: 400px;"
      />
    </div>
  </div>
</div>
```

**Assertions that gate on this DOM (all PASS):**
- `screen.getByTestId("media-part-audio")` → found
- `audio.tagName.toLowerCase()` → `"audio"`
- `audio.getAttribute("src")` → `"/api/files/test-id"`
- `audio.getAttribute("controls")` → `""` (present)
- `wrapper.getAttribute("data-align")` → `"flex-start"` (assistant, left)
- `screen.queryByText(/input_audio/)` → `null` (no raw blob text leaked)

> Note: jsdom does not load media — src is set correctly but the file is never fetched. This is expected jsdom behavior; the assertion is on the attribute, not decoded audio.

---

### #4698 — text-first ordering, alignment, ghost artifact

| Regression | Input | DOM assertion | Result |
|---|---|---|---|
| Text-first ordering | `[text, input_audio]` (production SDK order) | `querySelectorAll("audio").length === 1` (one bubble, not two) | ✅ PASS |
| User alignment | `role: "user"` + `input_audio` | `wrapper.getAttribute("data-align") === "flex-end"` (right) | ✅ PASS |
| Audio-only ghost | `input_audio` only, no sibling text | `mediaWrapper.querySelector("p") === null` | ✅ PASS |

---

### Full regression suite run output (verbose)

```
$ pnpm test:integration ScenarioMessageRenderer.integration.test.tsx --reporter=verbose

 ✓ when a message arrives with content as a python-repr string
   renders an <audio> element instead of the raw repr blob (126ms)
 ✓ when a message arrives with content as a JSON-encoded array string
   renders the same way as the python-repr equivalent (32ms)
 ✓ when a message arrives with a sibling text part as a transcript
   renders audio + italic transcript and not the raw blob (28ms)
 ✓ when a message has plain string content (not array-shaped)
   renders as a text bubble unchanged (58ms)
 ✓ when an assistant audio message has a trace id in drawer variant
   renders a View Trace button under the media bubble (22ms)
 ✓ when an assistant audio message has no trace id
   does not render a View Trace button (21ms)
 ✓ when a user-role audio message has a trace id
   does not render a View Trace button (19ms)
 ✓ when the renderer is mounted in grid variant
   does not render a View Trace button on assistant audio messages (20ms)
 ✓ when an assistant message has both audio and a sibling text transcript with one trace id
   renders exactly one bubble with exactly one View Trace button (23ms)
 ✓ when assistant text, tool_call, and tool_result turns have trace ids
   renders a View Trace button for each turn unchanged (47ms)
 ✓ when an assistant audio message arrives in the post-extraction url shape (#4138)
   renders a media-part-audio element whose src is the file url (drawer variant) (22ms)
 ✓ when an assistant audio message arrives in the post-extraction url shape (#4138)
   renders a media-part-audio element whose src is the file url (grid variant) (22ms)
 ✓ when a media part has an unsupported mimeType (#4138 graceful fallback)
   renders a graceful binary fallback, not a broken audio element (19ms)
 ✓ when an assistant voice turn carries audio + a sibling text transcript (#4698)
   > given ordering: text-first [text, input_audio] (production SDK ordering)
     renders exactly one assistant-left bubble with the text as transcript (20ms)
 ✓ when an assistant voice turn carries audio + a sibling text transcript (#4698)
   > given ordering: audio-first [input_audio, text]
     renders exactly one assistant-left bubble with the text as transcript (19ms)
 ✓ when a simulated-user voice turn arrives (#4698 alignment inversion)
   aligns the user media bubble to the right (flex-end) (17ms)
 ✓ when an assistant voice turn is audio-only with no transcript (#4698)
   renders one assistant-left audio bubble and no italic transcript node (19ms)

 Test Files  1 passed (1)
 Tests       17 passed (17)
 Duration    4.80s
```

```
$ pnpm test:integration ScenarioRunDetailDrawer.integration.test.tsx --reporter=verbose

 ✓ when rendering a voice turn with a url-shape audio source in the drawer (#4138)
   renders a media-part-audio element whose src is the file url inside the drawer (37ms)
 (+ 5 other drawer tests — all pass)

 Tests  6 passed | Duration 4.69s
```

---

### CI status

All CI checks pass. One pre-existing failure (`ee/licensing/__tests__/licenseGenerationService.unit.test.ts`, 4 tests) also fails on `main` — confirmed via run [27545024508](https://github.com/langwatch/langwatch/actions/runs/27545024508). Unrelated to this PR.

---

### Review verdict

All 6 review threads resolved/outdated (3 CodeRabbit + 3 `/review`). Review verdict: **READY** (see [verdict comment](https://github.com/langwatch/langwatch/pull/4717#issuecomment-4711102056)).

**Holding for Drew to merge.**

rogeriochaves (approved):

PR #4718, opened 2026-06-10
fix(aigateway): 14-minute provider timeout + fault-attributed error logging

rogeriochaves:
## What

Two prod gaps around LLM request failures, reported from a customer eval erroring with:

```
gateway chat/completions: upstream error (status 504): request timed out (default is 30 seconds)
```

### 1. Gateway-wide provider request timeout: 30s → 14 minutes

Bifrost's built-in default request timeout is 30 seconds, and the gateway never set its own, so every provider inherited it. Long completions (reasoning models, large generations) regularly exceed 30s. The gateway's longest-running callers are AWS Lambdas hard-capped at 15 minutes, so 14 minutes is the useful ceiling: long enough for any realistic completion, one minute of margin under the caller's cap.

Applied to every provider via `GetConfigForProvider`, plus:
- the stream per-chunk idle timeout (its 60s default would kill reasoning-model streams that think for minutes before the first token), and
- the direct Voyage embeddings client, so no dispatch path keeps a shorter hidden limit.

### 2. Fault-attributed error logging across nlpgo and the gateway

That 504 (and every workflow/evaluation node failure, and every gateway error response) was forwarded to the client with **no server-side log line**, so error increases were invisible until customers reported them. Every failure now leaves a stable log line CloudWatch can alert on, attributed to who it is on:

| fault | meaning | level |
|---|---|---|
| `customer` | their key/credits/request/code/dataset (e.g. out of credits) | info |
| `provider` | upstream LLM provider failed or timed out | warn |
| `platform` | our bug or infrastructure | error |

Customer faults are still logged because a spike in them can be a false flag for a platform problem.

- **Gateway** (`writeError` choke point): `msg=gateway_request_failed` with `fault`, `code`, `status`, `message` plus the calling `project_id` / `organization_id` / `virtual_key_id`.
- **nlpgo node failures** (`runLayer` choke point, covers workflows + evaluations v3 + playground + scenarios): `msg=workflow_node_failed` with `node_id`, `node_type`, `error_type`, `upstream_status` and the request's `project_id` / `trace_id` / `origin`. LLM node errors now thread the upstream HTTP status into `NodeError` (duck-typed so the engine keeps not importing adapters).
- **nlpgo handler errors**: `msg=request_failed` with `fault`, `code`, `reason`; SSE stream-start and playground mid-flight stream errors get their own lines.

Real line from a local run (failing LLM call, quota error):

```json
{"level":"info","msg":"workflow_node_failed","project_id":"project_demo","trace_id":"trace_demo","origin":"evaluation","fault":"customer","node_id":"sig","node_type":"signature","error_type":"llm_error","message":"You exceeded your current quota, please check your plan and billing details","upstream_status":429}
```

## Testing

- `services/aigateway/adapters/providers/network_config_test.go` - 14m default on every provider, stream idle aligned, voyage client ceiling.
- `services/aigateway/adapters/httpapi/faults_test.go` - upstream 5xx → provider/warn, upstream 402 → customer/info, herr codes by their fault, unhandled → platform/error, bundle identity on the log.
- `services/nlpgo/app/engine/faults_test.go` - failed node logged with identity through a real `Execute`, upstream status threading (402 → customer, 502 → provider), engine bugs → platform.
- `services/nlpgo/adapters/httpapi/handler_errors_test.go` - handler errors logged before the response is written.
- Full `go test ./services/aigateway/... ./services/nlpgo/...` green, `go vet` clean, feature-parity 12/12 new scenarios bound (specs: `specs/ai-gateway/provider-request-timeout.feature`, `specs/ai-gateway/error-observability.feature`, `specs/nlp-go/error-observability.feature`).

Post-deploy: CloudWatch metric filters can key on `gateway_request_failed` / `workflow_node_failed` / `request_failed` and the `fault` field to alert on increases before customers report.

## Deployment Impact

- **aigateway**: needs an image rebuild + redeploy for the timeout and logging to take effect (both are compile-time changes; no env vars or config to set).
- **nlpgo**: per-project Lambdas pick the change up through the normal lazy image rollout once the next image is baked and applied; no manual action.
- No env var changes, no schema changes, no migrations, no infra changes.
- Follow-up op (not in this PR): add CloudWatch metric filters/alarms on `gateway_request_failed` / `workflow_node_failed` / `request_failed` keyed by the `fault` field to catch error increases before customer reports.

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized fault attribution for request and node failures (customer/provider/platform) with consistent log levels and enriched request/context fields.
  * Unified error-response paths to ensure stable, structured failure logs and propagation of upstream status where available.

* **Tests**
  * Added comprehensive unit and engine tests validating fault classification, log contents, and context propagation.

* **Chores**
  * Standardized upstream/provider request timeout to a 14-minute ceiling.

* **Documentation**
  * Added BDD specs for error observability and provider-request-timeout behavior.

### DIFF
--- a/services/aigateway/adapters/providers/network_config_test.go
+++ b/services/aigateway/adapters/providers/network_config_test.go
@@ -1,0 +1,38 @@
+package providers
+
+import (
+	"testing"
+	"time"
+
+	bfschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// @scenario "Upstream requests get a 14 minute timeout for every provider"
+func TestGetConfigForProviderSetsFourteenMinuteRequestTimeout(t *testing.T) {
+	a := &account{}
+	for _, provider := range []bfschemas.ModelProvider{bfschemas.OpenAI, bfschemas.Anthropic, bfschemas.Bedrock} {
+		cfg, err := a.GetConfigForProvider(provider)
+		require.NoError(t, err)
+		assert.Equal(t, 14*60, cfg.NetworkConfig.DefaultRequestTimeoutInSeconds,
+			"provider %s must get the gateway-wide 14m timeout, not bifrost's 30s default", provider)
+	}
+}
+
+// @scenario "Streaming responses tolerate long gaps between chunks"
+func TestGetConfigForProviderAlignsStreamIdleTimeout(t *testing.T) {
+	a := &account{}
+	cfg, err := a.GetConfigForProvider(bfschemas.OpenAI)
+	require.NoError(t, err)
+	assert.Equal(t, 14*60, cfg.NetworkConfig.StreamIdleTimeoutInSeconds,
+		"per-chunk idle timeout must match the request ceiling; the 60s default kills reasoning-model streams that think before the first token")
+	// CheckAndSetDefaults must still fill the knobs we don't pin.
+	assert.NotZero(t, cfg.ConcurrencyAndBufferSize.Concurrency)
+	assert.NotZero(t, cfg.NetworkConfig.RetryBackoffInitial)
+}
+
+// @scenario "Direct embedding requests share the same ceiling"
+func TestVoyageClientSharesTimeoutCeiling(t *testing.T) {
+	assert.Equal(t, 14*60*time.Second, newVoyageClient().Timeout)
+}
--- a/services/aigateway/adapters/providers/network_config_test.go
+++ b/services/aigateway/adapters/providers/network_config_test.go
@@ -12,6 +12,9 @@ import (
 // @scenario "Upstream requests get a 14 minute timeout for every provider"
 func TestGetConfigForProviderSetsFourteenMinuteRequestTimeout(t *testing.T) {
 	a := &account{}
+	// 14*60 is intentionally an independent literal, not the exported
+	// constant: this pins the BEHAVIOR (14 minutes) so a fat-fingered
+	// constant fails here instead of self-verifying.
 	for _, provider := range []bfschemas.ModelProvider{bfschemas.OpenAI, bfschemas.Anthropic, bfschemas.Bedrock} {
 		cfg, err := a.GetConfigForProvider(provider)
 		require.NoError(t, err)
@@ -34,5 +37,8 @@ func TestGetConfigForProviderAlignsStreamIdleTimeout(t *testing.T) {
 
 // @scenario "Direct embedding requests share the same ceiling"
 func TestVoyageClientSharesTimeoutCeiling(t *testing.T) {
-	assert.Equal(t, 14*60*time.Second, newVoyageClient().Timeout)
+	// The contract here is the relationship: the direct Voyage client uses
+	// the same ceiling as the routed providers. The absolute 14m value is
+	// pinned by the request-timeout tests above with independent literals.
+	assert.Equal(t, ProviderRequestTimeoutSeconds*time.Second, newVoyageClient().Timeout)
 }
--- a/services/nlpgo/adapters/httpapi/handlers.go
+++ b/services/nlpgo/adapters/httpapi/handlers.go
@@ -82,7 +82,9 @@ func executeSyncHandler(application *app.App) http.HandlerFunc {
 		result, err := executor.Execute(ctx, *req)
 		if err != nil {
 			span.RecordError(err)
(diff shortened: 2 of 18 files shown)


services/nlpgo/adapters/httpapi/handlers.go
```diff
@@ -82,7 +82,7 @@ func executeSyncHandler(application *app.App) http.HandlerFunc {
 		result, err := executor.Execute(ctx, *req)
 		if err != nil {
 			span.RecordError(err)
-			herr.WriteHTTP(w, herr.New(r.Context(), domain.ErrBadRequest, herr.M{
+			writeHandlerError(r.Context(), w, herr.New(r.Context(), domain.ErrBadRequest, herr.M{
```
langwatch-agent: **P2** — This logs the post-decode sync execution failure on `r.Context()` instead of the enriched `ctx` built above. At this point `decodeStudioClientEvent` succeeded and `enrichRequestLogContext` already attached `project_id`, `trace_id`, and `origin`, but the `request_failed` line drops all of that, which defeats the PR's stated goal of attributing handler failures per customer/surface. Use `ctx` for both `herr.New` and `writeHandlerError` here, and add coverage that a sync executor error includes the enriched fields.

services/nlpgo/adapters/httpapi/handlers.go
```diff
@@ -454,6 +454,10 @@ func executeStreamHandler(application *app.App) http.HandlerFunc {
 			Heartbeat: streamHeartbeat(r),
 		})
 		if err != nil {
+			// The SSE error frame reaches the client; this line is the
+			// server-side trail for it (the engine stream failed to start).
+			clog.Get(r.Context()).Error("studio_stream_failed",
```
langwatch-agent: **P2** — Same context-loss issue on the streaming start-failure path: `studio_stream_failed` is emitted from the raw request context, so a failure after successful decode/start span loses `project_id`, `trace_id`, and `origin`. Since this is exactly the server-side trail for an SSE error frame, operators still cannot filter it to the affected customer/run. Log with the enriched `ctx`/`streamCtx` and cover the ExecuteStream-start error case.

langwatch-agent (dismissed):
2 findings (P2) — see inline comments.

services/nlpgo/adapters/httpapi/handlers.go
rogeriochaves: Fixed in 6570b6ae9: the engine-error branch now uses the enriched ctx for both herr.New and writeHandlerError, so request_failed carries project_id / trace_id / origin. Added TestExecuteSyncHandlerLogsEngineErrorWithEnrichedContext driving the real handler with a failing executor and asserting all three fields plus reason=engine_error.

services/nlpgo/adapters/httpapi/handlers.go
rogeriochaves: Fixed in 6570b6ae9: studio_stream_failed now logs on the enriched ctx, so the SSE-error trail is filterable to the affected customer/run. Added TestExecuteStreamHandlerLogsStartFailureWithEnrichedContext forcing an ExecuteStream-start error and asserting project_id / trace_id / origin on the line.

services/aigateway/adapters/providers/network_config_test.go
```diff
@@ -0,0 +1,38 @@
...
+		"per-chunk idle timeout must match the request ceiling; the 60s default kills reasoning-model streams that think before the first token")
+	// CheckAndSetDefaults must still fill the knobs we don't pin.
+	assert.NotZero(t, cfg.ConcurrencyAndBufferSize.Concurrency)
+	assert.NotZero(t, cfg.NetworkConfig.RetryBackoffInitial)
+}
+
+// @scenario "Direct embedding requests share the same ceiling"
+func TestVoyageClientSharesTimeoutCeiling(t *testing.T) {
+	assert.Equal(t, 14*60*time.Second, newVoyageClient().Timeout)
+}
```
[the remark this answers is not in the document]
rogeriochaves: Partially adopted in 76d84c69b. The voyage assertion now uses ProviderRequestTimeoutSeconds because its contract genuinely is the relationship (same ceiling as the routed providers). The two absolute assertions keep the independent 14*60 literal on purpose: asserting cfg.X == ProviderRequestTimeoutSeconds would be circular (it verifies the code uses the constant it uses, and self-passes if the constant is fat-fingered). The independent literal pins the behavior, so changing the timeout forces a deliberate test update - that is the regression guard working as intended. Added comments in the test stating this so it does not read as accidental duplication.

PR #4720, opened 2026-06-10
feat(billing): provision org retention policy on paid customers

sergioestebance:
## What

Wire the data-retention system into both paid entry points:

1. **Growth Seat subscription** (Stripe webhook): on first activation, provision explicit **organization-scoped retention policies — one per category** (`traces`, `scenarios`, `experiments`) — at the platform default window (**49 days / 7 weeks**).
2. **License activation** (self-hosted/Enterprise): on successful `validateAndStoreLicense`, provision the same policies — but **create-if-absent only**: a category that already has an organization-level override is never overridden, so a manually-tuned window survives license activation.

> Evaluation runs fall under the `traces` category and simulation runs under `scenarios`, so iterating the three canonical `RETENTION_CATEGORIES` covers evaluations/simulations too.

> **⏸ Removal-on-cancellation is deactivated** (feature not yet released). The two cancel-path call sites are commented out; the `removeSeatRetentionPolicy` helper stays so re-enabling is a one-line uncomment per site. Tests pin the deactivated behavior (cancel leaves policies in place).

## Why / decisions

- **Why 49 days (= platform default)?** Today every org resolves to the 49d platform default when it has no override. An explicit org policy *records* the paid entitlement so it **survives any future lowering of the platform/free default**. Behaviorally inert today (49d == default) — this PR ships the mechanism; the number is tuned later.
- The literal ask was "30 days", which the retention schema **rejects** (floor `MIN_RETENTION_DAYS = 49`, multiple of 7). 49d is the nearest legal value — confirmed against `retentionPolicy.schema.ts`.
- **Seat path = upsert on first activation only; license path = create-if-absent.** The seat path's first-activation guard already prevents clobbering; the license path can fire repeatedly (license renewals/replacements), so it checks `listOrganizationRules` and skips covered categories.

## How

A retention policy is a Prisma row (`RetentionPolicy`, unique on `scopeType+scopeId+category`) via `getApp().dataRetention.policy`; enforcement is ClickHouse TTL. Both hooks call the service directly (trusted system actions) — not via the tRPC plan/RBAC authz path.

- **Seat create** — chokepoint in `syncInvoicePaymentSuccess` (checkout + invoice both funnel through), first-activation + `isGrowthSeatEventPlan` guarded. Loops `RETENTION_CATEGORIES`, upserts each.
- **License create** — `LicenseHandler.validateAndStoreLicense` after the license is persisted (the single success point for UI/tRPC `license.upload`). Lists org rules, skips categories with an existing ORGANIZATION-scoped row, creates the missing ones. Invalid/expired licenses provision nothing.
- **Remove (deactivated)** — both cancel entrypoints have the `!remainingActive`-guarded removal commented out until release.
- **Best-effort, per category** — every retention call is try/caught and **logs, never throws**. A retention hiccup must not fail the Stripe webhook or license activation.

## Anything surprising

- **The 49d floor is deliberate — it is the action window.** Pricing tiers are visibility (blur) thresholds on top of it: free-tier content blurs after 14d but stays recoverable until the 49d policy deletes it, giving users a window to upgrade and recover. The blur layer is tracked in #4745.
- Once removal is re-enabled: it is unconditional on downgrade-to-free (clears a manually-set org policy too). Accepted under the "delete the policy on cancel" decision.

## Tests

- BDD spec: `specs/billing/seat-subscription-retention-policy.feature` (seat + license + deactivated-cancel scenarios, all `@unit`-bound).
- `webhookService.unit.test.ts` — provisions all categories on seat activation; not for non-seat plans; not on renewal; survives retention failure; cancel paths do NOT remove (deactivated).
- `licenseHandler.retention.unit.test.ts` — provisions missing categories on valid activation; **never overrides an existing org-level policy** (creates only missing ones); nothing on invalid/expired license; survives list/upsert failures.
- ✅ 41/41 webhook + 5/5 license tests · `pnpm typecheck` clean · `check:feature-parity` green (1736 scenarios).

## Summary by CodeRabbit

* **New Features**
  * Automatic data retention policy provisioning for Growth Seat plan activations using platform defaults.
  * Automatic data retention policy provisioning for self-hosted license activations, only filling in missing categories.
  * Best-effort retention policy setup that gracefully handles failures without disrupting subscription or license activation.

* **Tests**
  * Added comprehensive test coverage for retention policy provisioning scenarios.

### DIFF
--- a/langwatch/ee/billing/services/webhookService.ts
+++ b/langwatch/ee/billing/services/webhookService.ts
@@ -16,7 +16,10 @@ import { isGrowthEventsPrice, isGrowthSeatEventPlan, isGrowthSeatPrice } from ".
 import { SubscriptionRecordNotFoundError } from "../errors";
 import { traced } from "../../../src/server/app-layer/tracing";
 import { fireSubscriptionSyncNurturing } from "../nurturing/hooks/subscriptionSync";
-import { PLATFORM_DEFAULT_RETENTION_DAYS } from "../../../src/server/data-retention/retentionPolicy.schema";
+import {
+  PLATFORM_DEFAULT_RETENTION_DAYS,
+  RETENTION_CATEGORIES,
+} from "../../../src/server/data-retention/retentionPolicy.schema";
 
 const logger = createLogger("langwatch:billing:webhookService");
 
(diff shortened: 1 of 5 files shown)


0xdeafcafe (commented):
this needs to wait for the changes we discussed in standup right, about waiting to allow resub before deleting?

0xdeafcafe (dismissed):

langwatch/ee/billing/services/webhookService.ts
```diff
@@ -617,6 +621,12 @@ export class EEWebhookService implements WebhookService {
       organizationId: existingSubscription.organizationId,
       hasSubscription: !!remainingActive,
     });
+
+    // Retention-policy removal on cancellation is deactivated until
+    // the paid-retention feature is released. Re-enable by restoring this call.
+    // if (!remainingActive) {
+    //   await this.removeSeatRetentionPolicy(existingSubscription.organizationId);
+    // }
```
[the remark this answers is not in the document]
sergioestebance: Fixed in f8c14b122 — deleted the commented-out call sites AND the now-unused removeSeatRetentionPolicy helper (git remembers; restore from this commit's parent when the feature ships). Kept a one-line comment stating cancellation deliberately leaves policies in place.

specs/billing/seat-subscription-retention-policy.feature
```diff
@@ -0,0 +1,54 @@
...
+  Scenario: A retention failure never fails the billing webhook
+    Given a pending Growth Seat subscription for the organization
+    And provisioning the retention policy will fail
+    When the subscription is activated by a successful payment
+    Then the activation still completes and the confirmation notification is sent
+
+  # Removal-on-cancellation is deactivated until the paid-retention
+  # feature is released. Cancelling currently leaves the policies in place.
+  @unit
+  Scenario: Cancelling a subscription leaves the retention policies in place
```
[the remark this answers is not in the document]
sergioestebance: Fixed in f8c14b122 — narrative now states cancellation leaves policies in place (and mentions the license entry point), matching the scenario.

Aryansharma28 (dismissed):

langwatch/ee/billing/services/webhookService.ts
```diff
@@ -835,6 +847,41 @@ export class EEWebhookService implements WebhookService {
...
+   * so we log and continue to the next category, mirroring the surrounding
+   * notification side effects.
+   */
+  private async applySeatRetentionPolicy(organizationId: string): Promise<void> {
+    for (const category of RETENTION_CATEGORIES) {
+      try {
+        await getApp().dataRetention.policy.setForScope({
+          scope: { scopeType: "ORGANIZATION", scopeId: organizationId },
+          category,
+          retentionDays: PLATFORM_DEFAULT_RETENTION_DAYS,
```
Aryansharma28, line 877: **Non-blocking.** @sergioestebance nice — the per-category try/catch + behaviorally-inert-today framing is exactly right. One follow-up question: the PR body notes the pricing page's 14d / 30d windows are incompatible with `MIN_RETENTION_DAYS = 49` and flags lowering the floor as "out of scope, follow-up." Is that follow-up tracked anywhere we can link from here? A Linear / GitHub issue link in either the PR body or a `// TODO(#...): ` near this line would be enough — without a pointer, the next time marketing presses on the 30d Growth tier we'll lose time re-deriving why 49 was the floor.

langwatch/ee/billing/services/webhookService.ts
```diff
@@ -617,6 +621,9 @@ export class EEWebhookService implements WebhookService {
       organizationId: existingSubscription.organizationId,
       hasSubscription: !!remainingActive,
     });
+
+    // Cancellation deliberately leaves the org's retention policies in place
```
Aryansharma28, line 625: **Non-blocking.** @sergioestebance the deactivation pattern here works, but it's load-bearing in two places: this comment (and its twin at line 666) and the two unit tests that pin the no-removal contract. Whoever ships the paid-retention feature later has to (a) uncomment two call sites and (b) invert two tests, all by remembering to read these comments first.

The codebase already has `featureFlagService` (`langwatch/ee/governance/routers/governance.ts:26`) and the `isFeatureEnabled` pattern. Any reason not to gate this with a flag — `paid_retention.remove_on_cancel`, default off — so enabling becomes a settings flip instead of a code change spread across two files? It also gives us a kill-switch if removal misbehaves after launch.

Happy to defer if you already weighed the flag and decided against it.

Aryansharma28 (commented):
Read through this carefully — the per-category resilience and the spec/test 1:1 mapping are both lovely. Two non-blocking questions inline; nothing here would hold up an approve once CI is green.

sergioestebance:
## 🧪 Live QA evidence — license activation provisions org retention policies

Verified end-to-end on a local server running this branch (Enterprise test license, org `test@langwatch.ai`).

### Test 1 — existing policies are never overridden
Org started with 3 manually-created ORGANIZATION policies at **364 days** (traces, scenarios, experiments). Activated the license → DB rows unchanged (same ids, same `createdAt`, still 364d).

| License activated | Retention untouched (364d) |
|---|---|
| license activated | retention untouched |

### Test 2 — creates from zero when none exist
Removed the license and deleted all org policies (clean state: `license: no`, `[]` policies) → re-activated the same license → **3 ORGANIZATION policies auto-created at 49 days (platform default), one per category**:

```
traces      49d  createdAt 2026-06-11T13:04:13.315Z
scenarios   49d  createdAt 2026-06-11T13:04:13.416Z
experiments 49d  createdAt 2026-06-11T13:04:13.479Z
```

| Clean state (no license, no policies) | License re-activated | Policies auto-created (49d) |
|---|---|---|
| clean state | reactivated | auto created |

Both behaviors match the spec: create-if-absent on successful activation, never override existing org-level policies.

langwatch/ee/billing/services/webhookService.ts
sergioestebance: Done — tracked in #4745 (lowering the floor, per-plan windows, and re-enabling cancel removal). Added a `TODO(#4745)` pointer in the docblock right above this helper (542612370) and linked the issue from the PR body follow-up section.

langwatch/ee/billing/services/webhookService.ts
sergioestebance: Weighed it and decided against the flag for now: the removal semantics are not actually settled — when paid retention ships, cancellation may mean *delete* the policies or *downgrade* them to the free-tier window (depends on where #4745 lands on per-plan windows). Flag-gating today would mean shipping a guessed behavior behind a default-off flag, dead code with a kill-switch for something that was never alive. The earlier revision of this PR had exactly that (commented-out removal + helper) and review correctly pushed to delete it.

The re-enable path is now tracked in #4745 (third bullet), so the future implementer gets the pointer from the issue, not just these comments. The two pinned tests are intentional: they document the current contract and will fail loudly the moment someone adds removal, forcing a conscious inversion.

Aryansharma28 (approved):

Issue #3104, opened 2026-04-10
bug: ENAMETOOLONG crash in checkPermissions when parsing long MCP config strings


thiagolmoraes:
**Describe the bug**

A critical unhandled promise rejection occurs when attempting to configure the LangWatch MCP server using the Gemini CLI. The error ENAMETOOLONG: name too long, lstat ... is thrown. It appears the CLI is incorrectly attempting to treat a multi-line string (containing instructions and JSON configuration) as a filesystem path.

**To reproduce
Steps to reproduce the behavior:**

    Using @google/gemini-cli@0.36.0.

    Provide the agent (Gemini) with instructions to configure LangWatch.

    The agent generates the setup instructions and the MCP JSON configuration.

    During the processing or "context refresh" of this information, the CLI crashes with ENAMETOOLONG.

**Expected behavior**
The CLI should parse the instructions or save the configuration without attempting to call lstat on the entire string content.

**Environment**

    Component: langwatch_mcp_server / gemini-cli

    Version: @google/gemini-cli@0.36.0, keytar@7.9.0

    Browser/OS: macOS, iTerm2 3.6.9, Node.js environment via pnpm.

**Additional context**
The stack trace indicates the error originates during a checkPermissions and resolveToRealPath call within the Gemini CLI bundle:

Error: ENAMETOOLONG: name too long, lstat '/Users/thiagomangia/...[long string snippet]..."env": { "LANGWATCH_API_KEY": "..." }'
at Module.realpathSync (node:fs:2752:29)
at resolveToRealPath (chunk-S2IQOR7T.js:41506:10)
at checkPermissions (chunk-UNM3DGTG.js:73719:30)

It seems the parser is failing to distinguish between a file path and a block of text/code.

PR #4729, opened 2026-06-10
feat(data-privacy): unified scoped data privacy

rogeriochaves:
## Unified scoped data privacy

LangWatch had four separate, half-connected privacy controls. This change unifies them into one scoped, cascading `DataPrivacyPolicy` that resolves org -> department -> team -> project (plus personal-project smart-scopes), with per-field cascade resolution like the retention and model-default policies.

The five concerns, now one policy:

1. **Content drop** generalizes the org-only, gateway-only "no-spy" content stripping to every category (input, output, system instructions, tool calls) for every ingestion origin.
2. **Read-time visibility** generalizes the admin-only `capturedInput/OutputVisibility` to an audience model: the built-in role groups (Admins, All members, Viewers), the project owner (so personal workspaces can be visible only to the person themselves), plus the organization's custom RBAC groups. Departments scope where a rule applies, never who can see content.
3. **PII redaction** generalizes the project `piiRedactionLevel` and ports Presidio's "essential" recognizers to native TypeScript (including the Brazilian CPF, which the analysis service has no detector for), so the common case never calls a lambda. A fourth **custom** level lets a team redact exactly the identifiers they choose, running the pattern-based ones in process and sending only the ones that need the analysis service (names, locations) out.
4. **Secrets redaction** is net-new and default-on: native scrubbing of cloud and provider API keys, JWTs, private-key blocks, database URLs, and bearer tokens, plus custom patterns.
5. **Custom attribute rules** are net-new: a rule matches span-attribute paths (exact keys or `*` wildcards, e.g. `gen_ai.prompt.*`) and either drops them at ingestion (stamping a marker with the dropped key names) or restricts them to an audience at read time, the matching values replaced by a placeholder naming who can see them.

Retention is deliberately untouched; it governs data age, not content, and it is the architectural template this builds on.

### What is in this change (backend and API)

- **Resolver core** (`src/server/data-privacy/`): the `DataPrivacyPolicy` model, per-field cascade resolution including the department and personal-project smart-scopes, a cached resolver, the repository, the write service, the write authz, and the read snapshot.
- **Native redaction engines**: `secretsRedaction.ts` and `essentialPii.ts` (checksum-validated cards and IBANs, libphonenumber phone matching, context gating), composed in `applyContentRedaction.ts`. Wired into the ingestion path so essential PII and secrets run in-process with no external call. The strict PII level runs that same native floor first and then adds the analysis-service batch for names and locations, so an unreachable (or, in dev, unconfigured) analysis service downgrades strict to essential instead of leaving content exposed.
- **Drop enforcement** at the single span choke point (`RecordSpanCommand`), so both stored spans and the derived trace summary see the dropped categories, for every origin (OpenTelemetry, REST collector, gateway, ingestion keys, coding-agent logs). Observability metadata always survives.
- **Audience-aware read visibility** in `getUserProtectionsForProject`, covering every read surface that already funnels through it. Restriction is retroactive and the redaction reason names the audience that can see the content.
- **Migration and full convergence**: a single SQL migration backfills each legacy control into an equivalent rule at its original scope, then drops the legacy columns (`Organization.governanceLogContentMode`, `Project.capturedInput/OutputVisibility`, `Project.piiRedactionLevel`) and their enums, so the scoped policy is the only source of truth. Self-hosted gets it automatically on `migrate deploy`, with no manual task to run. Every reader is converged onto the policy: read visibility resolves from it directly, the ingestion PII level comes from it, and the old governance content-strip (service, store hook, session-policy content mode, and its settings UI) plus the project-settings PII and visibility selects are retired.
- **Kill switch**: `LANGWATCH_DATA_PRIVACY_ENFORCEMENT=off` reverts drop, native redaction, and visibility to the prior behavior.
- **tRPC `dataPrivacy` router**: `getSnapshot`, `setForScope`, `removeForScope`.

### Settings UI

A new **Data Privacy** settings page (`/settings/data-privacy`), always available now that the old controls are retired:

- An **"Effective for this project"** card showing the resolved cascade: per-category disposition, attribute rules, PII level, and secrets status.
- A **scope filter** ("All you can see") to filter the rules list by scope, mirroring the model-providers page.
- A **rules table** grouped by scope, each row a scope chip plus a plain-language summary, with a vertical overflow menu per row for Edit and Delete (the project standard, not inline icons).
- An **add / edit rule** drawer built on the standard `ScopeChipPicker`: one save can target several scopes at once (two teams, a team and two projects), producing one rule row per scope, exactly like model defaults and retention. The picker gains a "Personal projects" variant offering "All personal projects" and per-department personal scopes as first-class selections.
- **Compact content controls** (label left, small select right) for the four categories, plus a **custom attributes editor**: rules that match span attribute keys with `*` wildcards (e.g. `gen_ai.prompt.*`), each either Restricted or Dropped.
- The **audience is one multi-select of groups**, in the same chip style as the scope picker: "All members" (everyone with access, the one exclusive choice that replaces any narrower selection), "Project owners (their own personal projects)", the standard role groups (Admins, Members, Viewers), and the organization's custom RBAC groups, all freely combinable (project owners plus Admins plus a Superadmins group in one audience). Custom groups only exist on the enterprise plan, which is the only plan that can create them, so no extra gating is needed; an org without any sees the group section disabled with a hint.
- The trace view's redaction marker explains itself: the inline hint reads "hidden by privacy settings", the tooltip says what the rule does, and both the tooltip and the dropped-content banner link to the Data Privacy settings page for viewers with access.
- **Customer-grade copy**: the PII level labels say what each level catches, not how it runs, and each carries a help tooltip with the complete entity list, test-pinned to the redaction engines so it cannot drift (the strict one names the Microsoft Presidio model). A new `dev/docs/best_practices/copywriting.md` captures the rule that internal technical details never reach user-facing copy.
- The **secrets section** adds a custom-pattern editor: extra regexes redacted on top of the built-in catalog, validated inline with the same safe-regex (ReDoS) check the server enforces.
- Editing pre-fills from the stored rule with the scope fixed; Save is gated until the rule actually changes something, and explicitly touched controls persist even at default-looking values so a project can override a stricter parent.

### Behavior changes

- **Secrets redaction is on by default** for every project going forward. A team can extend it with custom patterns or turn it off.
- **Existing settings carry over automatically.** The SQL migration converts each project's exact prior posture into the equivalent scoped rules before dropping the legacy columns, so there is no visible change on upgrade and the settings card matches enforcement from the first read.

### Tests

The 56 BDD scenarios across `specs/data-privacy/` cover cascade resolution, multi-scope saves, content drop across origins (including wildcard attribute rules), audience-aware visibility (role groups, owner-only, custom groups), native vs strict PII, secrets redaction with custom patterns, and the migration backfill. Unit suites for the redaction engines, the attribute-pattern matcher, the resolver, the drop, the visibility decision, the read-time attribute redaction, and the rule-config builder; real-database integration tests for the repository, the service, the read path, and the backfill. Full typecheck is clean.

### Browser QA

Driven in a real browser against a live stack (local Postgres, ClickHouse, Redis, workers). Logged in, exercised the full rule lifecycle, and ingested a trace carrying fake provider keys and PII to confirm enforcement at rest and in the trace view.

Built-in secrets redaction needs no rule at all: a trace ingested with `OPENAI_API_KEY=sk-proj-...` and `ANTHROPIC_API_KEY=sk-ant-...` lands in ClickHouse with both keys already `[SECRET]` (verified at rest in `stored_spans`), and the SQL migration's backfill was verified to converge each project's legacy posture into the matching rules.

The catalog matches provider keys by shape, not by vendor prefix: one `sk-` rule over the base64url alphabet covers OpenAI project keys (`sk-proj-...`), Anthropic (`sk-ant-...`), and LangWatch (`sk-lw-...`), whose bodies carry `_` and `-` and so slipped past an earlier alphanumeric-only rule. A trace ingested with such a key plus a `postgres://user:pass@host` URL lands in storage with the key replaced by `[SECRET]` and the URL password masked while its host stays. Custom patterns clamp each match at the closing quote, so a greedy `sk-.*` redacts only the credential and leaves the surrounding JSON intact.

secrets redaction of a modern key and a database url

PII is layered the same way. Essential entities (emails, cards, phone numbers, national IDs) are scrubbed natively, and strict adds names and locations on top through the analysis service. A trace ingested under a strict rule with the analysis service offline still lands with its email, card, and secret redacted natively, the strict-only names and locations being all that slip through, so the failure mode is a graceful downgrade rather than full exposure.

strict PII falls back to the native essential floor when the analysis service is offline

The custom level opens a checkbox per identifier, split into a Fast detection group (matched natively as data arrives, including the Brazilian CPF) and a Deep detection group (names, locations, and the rest). A rule selecting only the email and CPF redacts exactly those at ingestion and nothing else: a trace with an email, a CPF, a credit card, and a name lands with the email and CPF replaced and the card and name kept, with no analysis-service call (verified at rest in `stored_spans`). The CPF is validated by its two check digits, so a random eleven-digit number is not mistaken for one. The selected entities are validated against the known set on save, so a typo cannot silently redact nothing.

custom PII level entity toggles, with the Brazilian CPF in the Fast detection group

The Data Privacy settings page: the "All you can see" scope filter and the per-row overflow menu (Edit and Delete).

scope filter and kebab menu

The rule drawer on the standard scope picker: the "Personal projects" group ("All personal projects", per-department personal scopes), a multi-scope selection (a team plus a project in one save), the compact content controls, and a wildcard attribute rule.

scope picker personal group
drawer scope content attrs

The audience picker, one multi-select of groups in the same chip style as the scope picker: "All members" (exclusive), "Project owners", the Admins, Members, and Viewers role groups, and custom RBAC groups, here combining project owners with Admins and a custom Superadmins group. Next to it, the custom secret patterns editor with its label and inline safe-regex validation.

audience picker open
audience chips
secrets custom patterns label

In the trace view, the redaction marker's tooltip explains the privacy rule and links to the settings page, and the dropped-content banner carries a Privacy settings button:

redacted tooltip with settings link
drop banner with settings button

The PII level tooltips carry the complete detected-entity lists, pinned to the engines by a unit test:

essential pii tooltip
strict pii tooltip

### Enforcement, case by case, in the Trace Explorer drawer

The traces list, header, span detail, and the waterfall span procedures all enforce visibility server-side, so restricted or dropped content never reaches the client. One ingested trace demonstrates three mechanisms at once: the restricted input shows "Redacted (visible to Viewers)" naming its audience, while the stored output has the email scrubbed by essential PII and the provisioned key scrubbed by a custom `acme_live_...` secret pattern.

summary restrict pii secret

A restricted custom attribute (`app.billing.*` rule) renders its placeholder in the span attributes table while neighboring attributes stay visible; a wildcard drop rule (`app.internal.*`) stripped two keys at ingestion and stamped the marker naming them. Verified at rest in `stored_spans`: the dropped keys are absent, the raw card token never reaches the client, and the email and key are already `[REDACTED]` in storage.

span attr restricted
span attr dropped marker

The owner-only audience on personal workspaces: one org-wide rule ("all personal projects, input visible only to the project owner") lets each member see their own coding-agent traces while everyone else, including admins, gets the redaction reason.

owner sees own
owner only restricted

Dropped categories keep their banner treatment from the earlier round (light and dark):

traces-v2 drawer markers light
traces-v2 drawer markers dark

### Redaction-aware evaluators

Redaction runs at ingestion, before an evaluation reads the stored content. If the evaluators only saw the scrubbed text they would find nothing and always pass, hiding the very leaks they exist to catch. Two changes close that gap.

**Typed redaction markers.** Ingestion redaction now names what it removed: native PII writes `[PHONE_NUMBER]`, `[EMAIL_ADDRESS]`, and so on; secrets write `[SECRET]`; the strict Presidio path normalizes its `<ENTITY>` output to the same bracket form. A dependency-free marker catalog (entity set pinned to the engines by a test) reads them back, and `detectSecretsInText` exposes the secrets rules as a detect-only API.

**A new native evaluator and a shared augmenter.** `API Keys & Secrets Detection` is a built-in evaluator that runs in TypeScript with no analysis-service round-trip, reusing the secrets-redaction rules. The evaluator catalog is now a merged facade so a native evaluator is a first-class member of the type union everywhere. A shared augmenter runs on both the native secrets detector and the remote Presidio PII detector: a value that redaction already replaced with a typed marker is counted back as a detection (PII only for entities the evaluator still checks), and content that was dropped at ingestion fails since a leak cannot be ruled out. Markers are read from every mapped field, so a secret in a mapped span attribute is covered like input or output, and a still-populated field is evaluated normally so dropping a sibling field does not falsely fail it.

The new evaluator in the Safety category, and the typed markers rendering in the trace view (PII here, secret on both):

secrets evaluator in the Safety picker
email and secret markers in a trace
secret marker in a trace

Proven end-to-end against the running server: the evaluator endpoint returns `passed: false` for a live `sk-proj-...` key (`Detected 1 secret: openai_api_key`), `passed: true` for clean content, and `passed: false` for content where the secret was already scrubbed to `[SECRET]` ("1 secret value was already redacted at ingestion") without calling the analysis service.

### Out of scope

- Retention is untouched.
- Pushing the drop into the Go gateway so content never leaves the edge is a deliberate later step.
- Category drop for raw log-origin records in `stored_log_records` is a follow-up; secrets and PII there are already redacted, and the user-facing span and trace summary are dropped.

## Deployment Impact

- **No env vars, helm values, or chart changes.** The only file under a deployment-surface path is `dev/docs/best_practices/scope-selector-and-badges.md` (a UI guideline doc). A vanilla `helm install` with no overrides behaves the same, modulo the two points below.
- **Two Prisma migrations run on `migrate deploy`.** The first adds the `DataPrivacyPolicy` table; the second backfills every organization's and project's legacy privacy settings (governance content mode, captured input and output visibility, PII level) into equivalent scoped rules, then drops those four legacy columns and both enums. The backfill and the drop are one transaction, so an operator either gets the full conversion or none. Existing behavior is preserved through the conversion; no manual step is needed.
- **One default behavior change: secrets redaction is on by default** for all projects after upgrade. Incoming trace content matching the built-in credential catalog (provider API keys, tokens, connection strings) is stored as a typed marker (`[SECRET]`; PII as `[PHONE_NUMBER]`, `[EMAIL_ADDRESS]`, and so on). Operators who need the raw values can disable it per scope in Settings, then Data Privacy, or set the escape hatch `LANGWATCH_DATA_PRIVACY_ENFORCEMENT=off`, which makes ingestion and read visibility fall back to the platform defaults (everything captured, PII redaction via the analysis service, no scoped rules applied).
- **BYOC dataplane:** no ClickHouse schema changes and no new services. Enforcement happens in the existing app and workers; dataplane deployments only need the regular app image bump plus the Postgres migrations on the control plane.

## Read-side privacy presentation (system / tools restrict + generic markers)

Dogfooding surfaced that the read side only enforced and explained `restrict` for input and output. This change makes the trace drawer present every content category the same way.

- `Restrict` is now enforced at read for system instructions and tool calls, not just input and output. Their turns are stripped from the conversation (and from the raw chat-array attributes like `gen_ai.input.messages`) for viewers outside the audience, so an expanded attribute cannot reveal them.
- Every category (input, output, system instructions, tool calls) carries a per-attribute marker, so an absent or hidden category never reads as missing instrumentation:
  - Dropped: "not stored", removed by a privacy policy and not recoverable.
  - Restricted and hidden: "hidden (visible to <audience>)".
  - Restricted but visible to the viewer: a "visible to <audience>" badge, so an admin reading admin-only content knows it is restricted rather than assuming everyone can see it.
- When strict PII redaction cannot reach the analysis service, the span is marked and the drawer warns that names and locations may remain, instead of presenting partially scrubbed content as fully redacted. Production still blocks rather than storing unredacted content.

Verified in the browser with real ingestion (admin viewer, org-scoped policy):

| Restrict + in-audience badges | Dropped markers + incomplete-PII notice |
|---|---|
| restrict | dropped |

## Summary by CodeRabbit

* **New Features**
  * Introduced unified scoped data privacy policies (organization, department, team, project) with cascade and “personal only” rules.
  * Added ingestion-time content dropping (while preserving observability metadata) and read-time “restrict” audience visibility with “visible to” labels and audience-aware redaction.
  * Enabled tenant-aware native secrets/PII redaction with Essential/Strict/Disabled behavior.
  * Added the **Data Privacy** settings page, privacy-dropped notices, and rule editor tooltips, plus an **API keys and secrets detection** security evaluator.
* **Documentation**
  * Added Data Privacy Policy documentation; updated “no-spy mode” and migration guidance; added new privacy feature specs.
* **Tests**
  * Expanded unit, integration, and Gherkin coverage for policy resolution, drop/restrict behavior, and native redaction.
* **UI/Chores**
  * Removed legacy privacy controls from project/admin settings so users manage privacy through Data Privacy rules.


rogeriochaves:
Thanks for the review. Addressed in the latest push:

- **Fail-open read visibility (`utils.ts`)**: now fails **closed**. When policy resolution throws, `getUserProtectionsForProject` hides captured input/output (the viewer sees the redacted placeholder) instead of falling back to permissive visibility, so a resolver/db blip cannot expose restricted content. Cost visibility keeps its own permission check, and the kill-switch path still keeps legacy behavior. Logged.
- **Silent fail-open drop (`applyOtlpSpanContentDrop`)**: the resolve + strip now share one try/catch that logs the failure. Drop intentionally stays fail-**open**, because dropping on a transient error would be permanent content loss, and the kept content is still subject to the now-fail-closed read visibility.
- **"Fails open" comment on the drop call + the `stripOtlpSpanContent` purity claim**: corrected to match runtime behavior (the function mutates the span in place; the drop fails open internally and never aborts span processing).
- **`projectId` vs `scope` (dataPrivacy router)**: `setForScope`/`removeForScope` now assert the target scope belongs to the project's organization. The scope authz already gated permission (`assertCanWriteDataPrivacyScope` resolves the scope's own org, never the caller-supplied `projectId`), so no caller could write a scope they lack permission on; this closes the loose contract.
- **`ESSENTIAL_PII_ENTITIES`**: `as const`.
- **BDD nesting** (`applyOtlpSpanContentDrop.unit`, `recordSpanCommand.contentDrop`): restructured into nested `given` / `when` / `it`.

Not changed, with reason:

- **`spanStorage.store.ts` per-record governance lookup**: this is the pre-existing governance content-strip path; this PR only renamed a local variable and added a comment there. The new scoped data-privacy drop runs once per span earlier, in `RecordSpanCommand` (`applyOtlpSpanContentDrop`), not in this batch loop, so it adds no per-record lookups. Optimizing the legacy governance batch lookup is a separate concern, out of scope here.

rogeriochaves:
Addressed both:

- **`contentDropService` opt-out (recordSpanCommand.ts)**: reverted to required. The two integration tests that cast deps `as never` to skip the prisma default path (`traceProcessing.coalescing`, `recordSpanCommand.dedup`) now inject an explicit no-op `contentDropService`, so the drop contract stays type-enforced for every real caller and the opt-out is visible in the tests rather than implicit.
- **`assertScopeBelongsToProjectOrganization` (authz)**: a missing project now returns `NOT_FOUND` (matching the missing-scope-target case); `BAD_REQUEST` is reserved for a real organization mismatch.

langwatch/src/components/settings/dataPrivacyRuleConfig.ts
```diff
@@ -0,0 +1,89 @@
...
+}: {
+  dispositions: Record<ContentCategory, Disposition>;
+  audience: RuleAudience;
+  piiLevel: PiiLevel;
+  secretsEnabled: boolean;
+}): DataPrivacyConfig {
+  const categories: NonNullable<DataPrivacyConfig["categories"]> = {};
+  for (const category of CONTENT_CATEGORIES) {
+    const disposition = dispositions[category];
+    if (disposition === "drop") {
```
langwatch-agent: **P2** — This makes narrower overrides back to default-looking values impossible. The resolver supports a project rule setting `capture` over an org/team `drop`/`restrict` (there is even unit coverage for org drop + project capture), but the drawer drops `capture`, `essential`, and secrets-on as if they are always just inherited defaults. In a project under a restrictive parent policy, selecting “Captured” or “Essential” either produces an empty config with Save disabled, or saves a rule that omits the field and keeps inheriting the parent restriction. Please build the payload relative to the inherited parent/effective value for the selected scope, or persist explicit choices for controls the user touched so less-restrictive overrides actually work.

langwatch/src/server/data-privacy/contentVisibility.ts
```diff
@@ -0,0 +1,132 @@
...
+ * or REDACTED_TO_ALL keeps that behavior until the backfill writes a rule.
+ */
+export function effectiveCategoryRestriction(
+  category: ResolvedCategory,
+  legacy: LegacyVisibility,
+): EffectiveRestriction {
+  if (category.disposition !== "capture") {
+    return { disposition: category.disposition, audience: category.audience };
+  }
+  switch (legacy) {
```
langwatch-agent: **P2** — Removing a backfilled visibility rule will not change behavior. The backfill writes a `DataPrivacyPolicy` row but leaves `Project.capturedInputVisibility` / `capturedOutputVisibility` unchanged; if the user removes that rule, the resolver falls back to `capture`, then this compatibility switch re-applies the stale legacy enum and the content remains admin-only/hidden. That contradicts the migration spec’s “after backfill the unified policy is the single source of truth” contract and makes the new remove action ineffective for legacy projects. Gate the legacy fallback on “not backfilled/no policy source”, or update/clear the legacy fields when policy rows are changed or removed.

langwatch/src/server/app-layer/traces/span-pii-redaction.service.ts
```diff
@@ -87,6 +113,32 @@ const PII_DEFAULTS: OtlpSpanPiiRedactionServiceDependencies = {
...
+ * legacy piiRedactionLevel during the migration window. A resolved level other
+ * than the platform default ("essential") can only come from an explicit rule,
+ * so it wins; at the default we honor the legacy field so pre-migration STRICT/
+ * DISABLED projects keep their level until the backfill writes an explicit rule.
+ */
+function reconcilePiiLevel(
+  policyLevel: PiiLevel,
+  legacyLevel: PIIRedactionLevel,
+): PiiLevel {
+  if (policyLevel !== "essential") return policyLevel;
```
langwatch-agent, line 162: **P2** — The same stale-legacy problem exists for PII level. After backfill, a project that used to be `STRICT` or `DISABLED` keeps that old `Project.piiRedactionLevel`; deleting the new policy row, or otherwise trying to return to the platform default `essential`, resolves `policyLevel === "essential"` and this function immediately restores the old legacy value. The settings page can therefore show/remove the unified rule while ingestion still enforces the old PII mode. The migration path needs provenance/backfill state here too, or policy mutations need to keep the legacy column in sync until it is removed.

langwatch-agent (dismissed):
3 findings (P2) — see inline comments.

langwatch/src/components/settings/dataPrivacyRuleConfig.ts
rogeriochaves: Fixed in e83b376f0. The drawer now tracks which controls the user touched and persists those explicitly, including capture, essential, and secrets-on, so a narrower override of a parent drop/restrict/strict actually saves. Add mode prefills from the inherited effective policy for the current project scope so the parent restriction is visible before you override it, and Save enables when the built config is non-empty (add) or differs from the original (edit). Added unit coverage for capture-over-drop, essential-over-strict, and secrets-on-over-off.

langwatch/src/server/data-privacy/contentVisibility.ts
rogeriochaves: Resolved by the full convergence in this PR. The legacy capturedInput/OutputVisibility columns are now dropped by the migration, and getUserProtectionsForProject resolves visibility from the policy only, with no legacy-enum fallback. Removing a rule therefore falls back to capture as intended, and the policy is the single source of truth.

langwatch/src/server/app-layer/traces/span-pii-redaction.service.ts
rogeriochaves: Resolved and clarified in 4656bbe4a. The piiRedactionLevel column is dropped, so there is no stale legacy field to restore. The value reconciled here is the optional per-request level carried on the ingestion command (renamed requestLevel, with corrected comments): an explicit policy rule wins, and at the platform default the per-request level applies. There is no longer a migration window or column read.

langwatch/src/server/api/routers/tracesV2.ts
```diff
@@ -3,33 +3,42 @@ import { z } from "zod";
...
 } from "~/server/app-layer/traces/ai-query";
-import { ValidationError } from "~/server/app-layer/domain-error";
 import { deriveTraceStatus } from "~/server/app-layer/traces/derive-trace-status";
 import { TraceNotFoundError } from "~/server/app-layer/traces/errors";
 import { translateFilterToClickHouse } from "~/server/app-layer/traces/filter-to-clickhouse";
 import type { SpanSummaryRow } from "~/server/app-layer/traces/repositories/span-storage.repository";
 import type { TraceSummaryData } from "~/server/app-layer/traces/types";
-import { changeTraceNameInputSchema } from "~/server/event-sourcing/pipelines/trace-processing/schemas/commands";
+import { getDataPrivacyPolicyService } from "~/server/data-privacy/dataPrivacyPolicy.service";
+import { prisma } from "~/server/db";
```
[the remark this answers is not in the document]
rogeriochaves: Removed in 4656bbe4a.

langwatch/src/server/routes/misc.ts
```diff
@@ -15,80 +15,82 @@
...
-  type UnifiedAuthVariables,
-} from "~/server/api-key/auth-middleware";
-import {
-  createServiceApp,
-  handlerManagedAuth,
-  internalSecret,
-  publicEndpoint,
-} from "~/server/api/security";
+import { captureException } from "~/utils/posthogErrorCapture";
+import { slugify } from "~/utils/slugify";
```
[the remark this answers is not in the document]
rogeriochaves: Removed in 4656bbe4a.

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai Addressed in e02da1c46, finding by finding:

1. **Migration "IRREVERSIBLE:" prefix**: skipped. There is no such prefix contract in this repo (no migration or doc under `dev/docs/` uses it); the comment block already states the rationale for having no down step, which is the convention here.
2. **`app.v1.ts` metadata-update `piiRedactionLevel`**: skipped, works as designed. The redaction service resolves the scoped policy itself (`resolveNativeContext` -> `getResolvedForProject`) and `reconcilePiiLevel` lets an explicit policy rule always win over the request level; the route-passed default only applies when no rule is set, which is the intended per-request escape hatch. The synthetic metadata-update span also carries no user content.
3. **`app.v1.ts` format doc vs handler default**: fixed. The handler defaults to `json` (digest only via `llmMode`), so the description now says "'json' (full raw data, default)".
4. **Inline dynamic import in `piiEntityLabels.unit.test.ts`**: fixed, hoisted to a top-level import per the repo rule.
5. **Single-select selection lookup ignores `personalOnly`**: fixed. The lookup, the trigger value, and `onChange` now all go through `entryKey` / carry `personalOnly`, with a regression test that fails against the previous code.
6. **Component tests named as unit tests**: fixed, renamed `RedactedField`, `PrivacyDroppedNotice`, and also `PIIRedactionNotice` (same pattern, touched by this PR) to `.integration.test.tsx`; all pass under the integration config.

rogeriochaves:
@coderabbitai review

docs/evaluations/evaluators/built-in-evaluators.mdx
```diff
@@ -24,12 +24,22 @@ LangWatch offers evaluators across several categories:
...
-| **Safety** | `presidio/pii_detection`, `azure/jailbreak` | Detect PII, jailbreaks, harmful content |
+| **Safety** | `presidio/pii_detection`, `langwatch/api_keys_and_secrets_detection`, `azure/jailbreak` | Detect PII, leaked credentials, jailbreaks, harmful content |
 | **Correctness** | `langevals/exact_match`, `langevals/llm_boolean` | Check answer accuracy |
 | **Custom Criteria** | `langevals/llm_boolean`, `langevals/llm_score` | LLM-as-Judge for custom checks |
 
 [Browse all evaluators →](/evaluations/evaluators/list)
 
+### API Keys & Secrets Detection
+
+`langwatch/api_keys_and_secrets_detection` flags leaked credentials in trace content: provider and cloud API keys, tokens, private keys, and database connection strings. It runs in-process with no analysis-service round-trip, using the same detection rules as secrets redaction, and works as a guardrail.
```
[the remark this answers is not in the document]
rogeriochaves: Done in a0b2073d5. Reworded to drop the in-process and round-trip implementation detail, keeping the customer-facing behavior (what it detects, same rules as secrets redaction, works as a guardrail).

docs/llms-full.txt
```diff
@@ -16688,12 +16688,22 @@ LangWatch offers evaluators across several categories:
...
-| **Safety** | `presidio/pii_detection`, `azure/jailbreak` | Detect PII, jailbreaks, harmful content |
+| **Safety** | `presidio/pii_detection`, `langwatch/api_keys_and_secrets_detection`, `azure/jailbreak` | Detect PII, leaked credentials, jailbreaks, harmful content |
 | **Correctness** | `langevals/exact_match`, `langevals/llm_boolean` | Check answer accuracy |
 | **Custom Criteria** | `langevals/llm_boolean`, `langevals/llm_score` | LLM-as-Judge for custom checks |
 
 [Browse all evaluators →](/evaluations/evaluators/list)
 
+### API Keys & Secrets Detection
+
+`langwatch/api_keys_and_secrets_detection` flags leaked credentials in trace content: provider and cloud API keys, tokens, private keys, and database connection strings. It runs in-process with no analysis-service round-trip, using the same detection rules as secrets redaction, and works as a guardrail.
```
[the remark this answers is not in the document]
rogeriochaves: Done in a0b2073d5. This file is generated from the .mdx, so I fixed the source and regenerated it; the embedded line now matches the reworded copy.

langwatch/src/server/evaluations/evaluators.native.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Catalog entries merged into the generated `AVAILABLE_EVALUATORS`. Structurally
+ * an `EvaluatorDefinition` for each native key; the facade applies the precise
+ * type when it merges.
+ */
+export const NATIVE_EVALUATOR_DEFINITIONS = {
+  [API_KEYS_AND_SECRETS_DETECTION]: {
+    name: "API Keys & Secrets Detection",
+    description:
+      "Flags leaked credentials in trace content: provider and cloud API keys, tokens, private keys, and database connection strings. Runs in-process. A secret that privacy redaction already scrubbed at ingestion is still flagged.",
+    category: "safety",
```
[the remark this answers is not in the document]
rogeriochaves: Done in a0b2073d5. Removed 'Runs in-process.' from the evaluator description; the remaining copy is all customer-visible behavior.

langwatch/src/server/evaluations/native/registry.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+export async function executeNativeEvaluation({
+  evaluatorType,
+  data,
+}: {
+  evaluatorType: string;
+  projectId: string;
+  data: Record<string, unknown>;
+  settings?: Record<string, unknown>;
+}): Promise<SingleEvaluationResult> {
```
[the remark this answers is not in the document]
rogeriochaves: Done in a0b2073d5. Removed the unused projectId and settings from the executor type and both call sites. executeNativeEvaluation only needs evaluatorType and data; settings still flows to augmentEvaluationResult, which uses it.

rogeriochaves:
@coderabbitai review

langwatch/src/server/evaluations/evaluators.native.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Catalog entries merged into the generated `AVAILABLE_EVALUATORS`. Structurally
+ * an `EvaluatorDefinition` for each native key; the facade applies the precise
+ * type when it merges.
+ */
+export const NATIVE_EVALUATOR_DEFINITIONS = {
+  [API_KEYS_AND_SECRETS_DETECTION]: {
+    name: "API Keys & Secrets Detection",
+    description:
+      "Flags leaked credentials in trace content: provider and cloud API keys, tokens, private keys, and database connection strings. A secret that privacy redaction already scrubbed at ingestion is still flagged.",
```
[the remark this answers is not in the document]
rogeriochaves: Done in db38814cf. Dropped "at ingestion" from the picker description so it states only the customer-visible behavior; the docs section keeps the ingestion-timing context where it explains retroactivity.

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai review

rogeriochaves:
Thanks for the review. Triage of the four findings (inline comments failed to post on GitHub's side, so collecting here):

1. **`secretsRedaction.unit.test.ts` model fixture** — fixed in b5d94838e (`gpt-5` to `gpt-5-mini`).

2. **`evaluations-legacy.ts` `captureException({ param: params })`** — this code is identical to `main` (`git diff origin/main` is empty for those lines); it predates this branch and came from the `captureException` narrowing refactor. It is not introduced or changed by this PR, so leaving it out of this PR's diff. Worth a dedicated follow-up to scrub the analytics payload, but not in this data-privacy change.

3. **`tracesV2.ts` `enrichLlmSpanWithAncestorPrompt` visibility cutoff** and 4. **`redaction.ts` teaser scrubbing of values copied into `params`/`events`** — both belong to the plan-based visibility-window teaser feature (ADR-028), which this branch merged from `main`; the merge is why that code shows up in this diff. They are orthogonal to this PR's data-privacy enforcement (secrets, PII, content drop, audience visibility) and a correct fix needs that feature's design context. Tracking them against the visibility-window feature rather than widening this PR. The data-privacy audience redaction and the secrets/PII scrubbing in this PR are unaffected.

rogeriochaves:
@coderabbitai review

Two follow-up commits since the last review:
- `c2172c9b4` makes native essential PII a floor for every non-disabled level, so the strict level scrubs cards/emails/etc. in-process and only escalates to the analysis service for names/locations. An unreachable analysis service now downgrades strict to essential instead of leaving content exposed.
- `fdcb0641e` types the v2 redaction flags on `redactV2Content` and binds the two role-aware drop scenarios.

rogeriochaves:
@coderabbitai thanks, triaged all four:

**Fixed**
- `applyOtlpSpanContentDrop.unit.test.ts` `attrVal(span, key)` now takes object-destructured params (`0237de57f`), matching the named-params convention.

**Skipping, with reasons**
- `secretsRedaction.unit.test.ts` BDD nesting: the whole data-privacy unit suite uses `describe("given <condition>") > it(...)`, where the "given" carries the scenario (e.g. "given a database URL with a password" > "redacts only the password"). `applyContentRedaction`, `span-pii-redaction.nativeScopedPolicy`, and the rest follow the same shape. Inserting a `when` layer into this one file would make it inconsistent with every sibling. The new role-drop tests that branch on an action already use `given > when > it`. Leaving the condition-style files as they are for suite consistency; a uniform pass is a separate change.
- `docs/llms-full.txt` MCP-server lines (11263-11273): not part of this PR's diff (`git diff origin/main...HEAD -- docs/llms-full.txt` is empty there), this is pre-existing MCP setup prose unrelated to data privacy. The Info block in particular is deliberate: it describes the config in prose rather than a JSON snippet specifically because some CLI agents crash with `ENAMETOOLONG` on the `@`-scoped package name (issue #3104). Removing that explanation would reintroduce the confusion it prevents, so it stays out of scope for this PR.

rogeriochaves:
@coderabbitai review

New commit `ed0b06930` adds a custom PII level and Brazilian CPF redaction:
- `BR_CPF` native recognizer with two-check-digit validation (essentialPii.ts); covered at essential, strict (via the native floor), and custom.
- A fourth `custom` PII level: `pii.entities` selects exactly which identifiers to redact. The native subset runs in process; only analysis-service identifiers (PERSON, LOCATION, ...) are sent to langevals, with the selected entity list threaded through `PIICheckOptions.entities`.
- Rule drawer gains a Custom radio with grouped entity toggles; effective card and rule summary report the count.

rogeriochaves:
@coderabbitai thanks, triaged all four against `5ece70dfd`:

**Fixed**

- **Batch Presidio ignored custom entities** (`piiCheck.ts`) — real bug, good catch. The batch path now routes through the same `presidioEntitiesSetting(level, entities)` helper the single-text path already used, so a custom policy's selection reaches the request body instead of the level default. Added `piiCheck.batchEntities.unit.test.ts`, which executes the real batch function and asserts only the selected entity (`{ person: true }` for STRICT + `["PERSON"]`) lands in the request body.
- **`pii.entities` accepted arbitrary strings** (`dataPrivacy.types.ts`) — now validated against the known entity set, sourced from the dependency-free marker registry so it stays pinned to the engines without dragging the heavy redaction modules (and `@google-cloud/dlp`) into a schema that the client bundle imports. Plus a `superRefine`: the custom level requires at least one entity, and non-custom levels reject `entities`. New `dataPrivacyConfigSchema.unit.test.ts` covers each case. I kept it a refine rather than a discriminated union so the inferred `DataPrivacyConfig` shape is unchanged, just more strictly validated.
- **Build-internal copy in the custom PII groups** (`data-privacy.tsx`) — the two group labels are now "Fast detection" / "Deep detection" with user-facing hints, dropping "in-process" and "analysis service".

**Kept (intentional)**

- The strict-level tooltip naming **Microsoft Presidio** is a deliberate, product-approved exception to the no-internals rule (added earlier so users know which model powers deep detection). The surrounding copy stays abstract.

**Skipped (with reason)**

- **`given` to `when` BDD nesting** in the three test files — the whole data-privacy + redaction suite uses a consistent two-level form: `describe("<unit>") > describe("given <condition>") > it("<behavior>")` (16 such blocks in `essentialPii`, 11 in `resolveDataPrivacy`, 9 in `nativeScopedPolicy`). These are condition-named, not behavior-named, describes, which matches the BDD intent. Converting only the four newest blocks to a three-level `given > when > it` would make those files internally inconsistent with their identical existing siblings. Same call as the prior round.

rogeriochaves:
@coderabbitai review

Resynced with main (absorbing #4719) and fixed two CI reds on the same PR:
- `RedactedField` deferred its org-context lookup into a redacted-only sub-component, so panels that render non-redacted content no longer need a tRPC provider (fixes a `ExecutionOutputPanel` test crash).
- biome-formatted the evaluator-area files the codemod import renames touched.

app-ci is green (typecheck, build, feature-parity, lint, all test shards).

charts/langwatch/README.md
```diff
@@ -200,7 +200,6 @@ npx @bitnami/readme-generator-for-helm --readme ./README.md --values values.yaml
 | `app.http.publicUrl`                                                    | Public URL for users.                                                                     | `http://localhost:5560` |
 | `app.features`                                                          | Feature flags for the app.                                                                |                         |
 | `app.features.skipEnvValidation`                                        | Skip env var validation (dev only).                                                       | `false`                 |
-| `app.features.disablePiiRedaction`                                      | Disable PII redaction.                                                                    | `false`                 |
```
0xdeafcafe, line 203: if we remove this, can we add the new flag here instead

.github/workflows/sdk-javascript-ci.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       REDIS_URL: "redis://localhost:6379"
       NODE_ENV: "test"
       SKIP_ENV_VALIDATION: "true"
-      DISABLE_PII_REDACTION: "true"
+      OPS_PII_STRICT_PRESIDIO_REDACTION_DISABLED: "1"
```
0xdeafcafe, line 98: why the `OPS_` prefix?

langwatch/src/pages/settings/data-privacy.tsx
```diff
@@ -0,0 +1,1461 @@
...
+                        </Tooltip>
+                      </RadioGroup.Item>
+                      <RadioGroup.Item value="strict">
+                        <RadioGroup.ItemHiddenInput />
+                        <RadioGroup.ItemIndicator />
+                        <RadioGroup.ItemText>
+                          Strict (adds names, locations, and more)
+                        </RadioGroup.ItemText>
+                        <Tooltip
+                          content={`Deep detection with the Microsoft Presidio PII model. Everything in Essential, plus: ${STRICT_ADDED_PII_SUMMARY}.`}
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 5f92c474a: dropped the internal "Microsoft Presidio PII model" service name from the strict PII tooltip, so the copy now says what it detects, not how. The 1174-1182 copy you flagged (in-process / analysis service / slower) had already been updated to "Fast detection" / "Deep detection".

langwatch/src/server/data-privacy/redaction/__tests__/essentialPii.unit.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    });
+
+    it("leaves a repeated-digit sequence intact", () => {
+      const input = "ref 111.111.111-11 done";
+      expect(redact(input).text).toBe(input);
+    });
+  });
+
+  describe("given an entity filter (the custom level)", () => {
+    it("redacts only the selected identifiers", () => {
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 5f92c474a: reworded the two inner describe blocks to "when ..." conditions ("when the input contains a Brazilian CPF", "when an entity filter limits redaction to the custom level").

langwatch/src/server/app-layer/traces/__tests__/span-pii-redaction.nativeScopedPolicy.test.ts
```diff
@@ -0,0 +1,342 @@
...
+        mkPolicy({ piiLevel: "custom", piiEntities: ["EMAIL_ADDRESS"] }),
+      );
+      const span = spanWith({ input: "mail jane@example.com" });
+
+      await service.redactSpan(span, null, "ESSENTIAL", TENANT);
+
+      expect(attr(span, "input")).toContain("[EMAIL_ADDRESS]");
+      expect(batchSpy).not.toHaveBeenCalled();
+    });
+  });
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 5f92c474a: reworded the flagged inner describe blocks in both files to "when ..." conditions. The OpenGrep credit-card hits at lines 304/313 are intentional test fixtures (the canonical 4111 test PAN) exercising CREDIT_CARD redaction, kept as-is.

langwatch/src/server/data-privacy/dataPrivacy.types.ts
```diff
@@ -0,0 +1,189 @@
...
+      .strict()
+      .optional(),
+    pii: z
+      .object({
+        level: z.enum(PII_LEVELS),
+        // Only meaningful when level === "custom": the exact identifiers to
+        // redact, as canonical entity names ("EMAIL_ADDRESS", "BR_CPF",
+        // "PERSON"). Pattern-based ones run natively; ones that need the
+        // analysis service are sent there only when selected.
+        entities: z.array(z.string()).max(64).optional(),
```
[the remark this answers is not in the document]
rogeriochaves: Already addressed: custom PII entities are validated against VALID_PII_ENTITIES (dataPrivacy.types.ts:12), a whitelist sourced from the canonical REDACTION_MARKER_ENTITIES marker registry (minus the secrets marker) and enforced by a .refine() on the entities schema. They are not raw strings, so the single-whitelist suggestion is already in place.

charts/langwatch/README.md
rogeriochaves: Done in d4768a7fb. Replaced it with `app.features.disableStrictPiiRedaction`, which maps to the new `OPS_PII_STRICT_PRESIDIO_REDACTION_DISABLED` env var.

One semantic change worth noting: the env var is emitted only when an operator sets the value to `true` (rendering `"1"`). Left at the default `false` it is omitted entirely, so the Ops UI toggle and the per-scope data-privacy policy keep full control of the strict kill switch. If the chart always emitted `"0"`, the strict env override would win over the store and pin the flag off, which we do not want. Validated both states with `helm template` against the app and workers deployments.

.github/workflows/sdk-javascript-ci.yml
rogeriochaves: The env var name is auto-derived from the feature-flag key by uppercasing it (`ops_pii_strict_presidio_redaction_disabled` -> `OPS_PII_STRICT_PRESIDIO_REDACTION_DISABLED`), so the prefix is not chosen for the env var itself.

`ops_` is the registry convention for operational kill switches (emergency operator overrides), as opposed to `release_` / `release_ui_` for feature-release flags. Its sibling is `OPS_ES_CAUSALITY_LOOP_GUARD_DISABLED`. The primary override accepts only `"1"` (on) or `"0"` (off); anything else falls through to the Ops UI store and then the registry default.

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordLogCommand.ts
```diff
@@ -32,6 +32,7 @@ export interface RecordLogCommandDependencies {
         resourceAttributes: Record<string, string>;
       },
       piiRedactionLevel: PIIRedactionLevel,
+      tenantId?: string,
```
0xdeafcafe: tenant ids should always, always use the branded type for saftey.

rogeriochaves: Agreed. Fixed in bc31f6b48. The handler already created the branded `TenantId` via `createTenantId(...)` but then passed the raw `tenantIdStr` into `redactLog`, and the param was typed `string`.

Branded the whole redaction chain end to end so a non-tenant string cannot reach tenant-scoped policy resolution: `redactSpan` / `redactLog` / `redactMetricAttributes` and the private `resolveNativeContext` now take `TenantId`, and all three command handlers pass the branded value. Typecheck clean and the redaction + command tests pass (86 tests). The sibling cost-enrichment and token-estimation services are untouched here since they belong to a different path, but they are the obvious next candidates if we want the same guarantee everywhere.

0xdeafcafe (dismissed):

PR #4730, opened 2026-06-10
fix(api-key): stop misclassifying legacy project keys that contain underscores

rogeriochaves:
## Problem

Legacy project keys (`Project.apiKey`) are random strings minted from alphabets that include `_` and `-`. `getTokenType` classified any `sk-lw-` token containing an underscore as a new-format `{lookupId}_{secret}` API key. A legacy key with a random underscore in its body therefore went down the `ApiKey` table lookup path, missed, and the request got a 401 "Invalid credentials" with no fallback. The key itself was perfectly valid.

Found while dogfooding the evaluations skill against a local QA instance: the seeded project key happened to contain an underscore at body position 33 and every SDK and CLI call failed with 401 even though the key matched the database exactly.

## Fix

Two layers:

1. `getTokenType` now only classifies an `sk-lw-` token as a new-format API key when the body matches the exact generated shape (`{16 alphanumeric}_{48 alphanumeric}`, per `LOOKUP_ID_LENGTH`/`SECRET_LENGTH` and `ALPHABET`). Legacy bodies (different lengths, may contain `-`/`_`) can no longer match.
2. Defense in depth in `TokenResolver.resolve`: when an `sk-lw-` token is classified as an API key but misses the `ApiKey` lookup, fall back to the legacy project key lookup. The fallback only grants access when the full token matches a stored `Project.apiKey`, so invalid keys still 401. `pat-lw-`/`ik-lw-` prefixes never existed as legacy keys and don't fall back.

## Tests

- `getTokenType`: legacy key with underscore in body → `legacyProjectKey`; wrong-segment-length underscore tokens → `legacyProjectKey`; generated tokens (sk-lw and ik-lw) → `apiKey`.
- New `token-resolver.unit.test.ts`: legacy underscore key resolves via project lookup; new-format token missing the ApiKey lookup falls back to legacy lookup; returns null when both miss.
- All 84 unit tests in `src/server/api-key/__tests__/` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key validation with stricter format detection for new token types.
  * Added fallback authentication support for legacy project keys that resemble new token formats, ensuring backward compatibility.
  * Enhanced token resolution logic to properly handle edge cases and prevent authentication failures.

### DIFF
--- a/langwatch/src/server/api-key/__tests__/api-key-token.utils.unit.test.ts
+++ b/langwatch/src/server/api-key/__tests__/api-key-token.utils.unit.test.ts
@@ -128,21 +128,41 @@ describe("getTokenType", () => {
     });
   });
 
-  describe("when given a new sk-lw- token with underscore", () => {
+  describe("when given a new-format sk-lw- token", () => {
     it("returns apiKey", () => {
-      expect(getTokenType("sk-lw-abcdef1234567890_secretsecret")).toBe("apiKey");
+      const { token } = generateApiKeyToken();
+      expect(getTokenType(token)).toBe("apiKey");
     });
   });
 
   describe("when given an ingest ik-lw- token", () => {
     it("returns apiKey", () => {
-      expect(getTokenType("ik-lw-abcdef1234567890_secretsecret")).toBe("apiKey");
+      const { token } = generateApiKeyToken({ prefix: INGEST_KEY_PREFIX });
+      expect(getTokenType(token)).toBe("apiKey");
     });
   });
 
   describe("when given a legacy project key (sk-lw- without underscore)", () => {
     it("returns legacyProjectKey", () => {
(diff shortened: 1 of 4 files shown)


langwatch-agent (commented):
No findings. I reviewed the token classification/resolver changes and the new regression coverage; the strict new-key shape check plus the `sk-lw-` legacy fallback covers the underscore legacy-key case without broadening org-level API-key auth.

Local verification: `NEXTAUTH_SECRET=test-secret pnpm vitest run src/server/api-key/__tests__/api-key-token.utils.unit.test.ts src/server/api-key/__tests__/token-resolver.unit.test.ts src/server/api-key/__tests__/api-key.service.unit.test.ts` passed. The current red CI checks appear unrelated to this change: `test-integration (6)` failed while pulling Docker images, and `sdk-python test` hit a remote trace-share 500.

sergioestebance:
Approved

Issue #4759, opened 2026-06-11
Harden GroupQueue blob reclamation: producer-path latency, observability, and same-class leaks


sergioestebance:
Follow-ups from the ruthless review of langwatch/langwatch#4758 (fixes langwatch/langwatch#4757). The core fix is correct and race-safe; these are hardening/observability items and pre-existing same-class gaps surfaced during the trace. None block langwatch/langwatch#4758 except the P2 decision, which is cheap enough to fold into that PR before merge.

## P2 — Reclaim is `await`ed on the producer hot path (decide: await vs fire-and-forget)

`send()` / `sendBatch()` (`groupQueue.ts:345`, `:443`) `await this.deleteEnvelopeBlobs(...)`, while the four worker-side reclaims (`:513, :660, :717, :759`) are fire-and-forget `void`. `send()` is the reactor's hot path — the exact path that drove the capacity incident. For a turn re-folding N large (`>32 KiB`) spans this adds N sequential `UNLINK` round-trips to the fold handler, during precisely the pressured scenario this fix targets.

Correctness is identical either way (blob is reclaimed; 7-day TTL backstops a crash-before-delete) — this is a throughput/consistency call:

- **Option A — fire-and-forget** (`void this.deleteEnvelopeBlobs([...])`): consistent with the 4 siblings, zero added producer latency, leak still fixed (deleted ms later). Cost: the two e2e tests switch their post-squash assertions to `await vi.waitFor(...)` (the existing blob test at `groupQueue.integration.test.ts:245` already uses that pattern).
- **Option B — keep `await`**: stronger guarantee + deterministic tests, at the cost of hot-path round-trips in the incident-prone subsystem.

**Recommendation: Option A** — match the established pattern, keep the producer off the extra round-trip.

## P3 — No observability on the reclaim

`send`/`sendBatch` increment `gqJobsStagedTotal` / `gqJobsDedupedTotal` but nothing counts squash-displaced blobs reclaimed. The incident's headline miss was *no per-key signal until capacity hit 50%*. Add a `gqBlobsReclaimedTotal` counter (incremented when `readEnvelopeBlobId(orphanedValue)` resolves to a real id) so the fix is observable in prod and a regression is caught early.

## P3 — e2e tests prove the key is gone, not that the survivor still works

[...]

Issue #4757, opened 2026-06-11
GroupQueue blob offload leaks orphaned blobs on dedup squash (Redis capacity)


sergioestebance:
## Summary

The GroupQueue payload-envelope **blob offload** (#4726) leaks orphaned blob keys whenever a staged, blob-backed payload is **squashed by deduplication** before it is processed. Each dedup squash that displaces a blob-backed value (the dominant producer being the `claudeCodeSpanSync` reactor re-folding a turn through the `recordSpan` command, which is `extend+replace` — #4680) writes a fresh blob and silently drops the displaced one. The displaced blob is never reclaimed and lingers for its full **7-day TTL**.

Under sustained large-span ingestion this accumulates fast enough to drive Redis toward its `noeviction` ceiling.

## Impact / detection

- **Severity:** Sev-2 capacity (no customer-facing outage, but on track for the `noeviction` write-wall → span-ingestion failure if unmitigated).
- **Detection:** `redis-capacity-usage-high` CloudWatch alarm (eu-central-1), threshold crossed at 50%, climbing past 80%.
- **Magnitude (validated read-only in prod):**
  - `DBSIZE` ~307K keys vs ~27K at the prior baseline incident.
  - **~280K `gq:blob:*` keys = ~91% of the keyspace.**
  - **99.25% of blobs (≈278K, ~7.4 GB) are unreferenced** by any live staged job — i.e. orphans, ≈ the entire `used_memory`.
  - In a 1,000-blob sample, **100% had `IDLETIME == age`** (written once, never read) — the fingerprint of a value displaced *before* it was ever dispatched, i.e. a dedup-squash orphan, not a completion-delete miss.

## Root cause

Two independently-correct changes interact to produce an unmanaged resource leak:

1. **#4680** registered `recordSpan` with GroupQueue-layer dedup (`extend: true, replace: true`). On a dedup hit the stage Lua squashes the new payload onto the existing staged job **in place** (`STAGE_LUA`: `HSET dataKey existingJobId <new value>`), overwriting the previously-stored value. Before #4726 the overwritten value was inline JSON — overwriting it freed it automatically. ✅ Bounded.

[...]

PR #4761, opened 2026-06-12
chore(queue): orphaned-blob reclaim tool

sergioestebance:
## Why

#4758 stops **new** GroupQueue offload-blob orphans (a dedup squash now reclaims the displaced value's blob). But the **existing** pool from the 2026-06-11 capacity incident (#4757) — **~231K orphaned `gq:blob:*` keys / ~5.7 GB** — still sits on its 7-day TTL. Waiting it out keeps Redis near the `noeviction` ceiling for days; this tool reclaims it on demand.

Sibling to `scripts/ops/reap-stranded-group-keys.sh`, same conventions (POSIX sh, DRY-RUN by default, age floor, runs co-located with Redis or over TLS).

## What it does

`DISCOVER` (default, read-only) builds the live reference set, classifies every blob, and prints the count + a chunk plan. `APPLY=1` frees **progressively in chunks**, reporting `used_memory` after each chunk so memory drains in visible steps rather than one large sweep.

A blob is reclaimed only when it is **both**:
- **Unreferenced** — no live `{prefix}group:*:data` envelope carries its id in `"r"`. Protects every staged job (normal, paused, retry-backoff — all re-staged into the data hash).
- **Aged > `MIN_AGE_HOURS`** (default 2h, derived from remaining TTL) — protects in-flight jobs whose value left the data hash at dispatch but whose blob a worker is about to read. The processing window is bounded by the active-key TTL (~300s), so a multi-hour floor cannot hit a live blob.

Canonical span data lives in ClickHouse and a missing blob is replay-recoverable (decode throws → slot completed → replay), so the age guard is belt-and-suspenders, not a correctness requirement.

```sh
# discover (read-only): how many, how much, chunk plan
H=<host> REDISCLI_AUTH=<pw> sh scripts/ops/reclaim-orphaned-blobs.sh
# reclaim progressively, watching used_memory drop per chunk
H=<host> REDISCLI_AUTH=<pw> APPLY=1 FREE_CHUNK=10000 sh scripts/ops/reclaim-orphaned-blobs.sh
# incremental: cap one run, re-run to continue
H=<host> REDISCLI_AUTH=<pw> APPLY=1 LIMIT=10000 sh scripts/ops/reclaim-orphaned-blobs.sh
```

Performance: TTLs are fetched via chunked `EVAL` (ARGV, single-slot replication group) and `SCAN ... COUNT 1000`, so a full 300K-key pass runs in well under a minute even over a WAN, not the hours a per-key loop would take.

## Test plan

- [x] **Local (throwaway redis)**: referenced / young / aged classification correct; DRY-RUN mutates nothing; APPLY reclaims only aged orphans and keeps referenced + young; multi-chunk EVAL line-alignment; progressive chunk stepping; `LIMIT` caps a run.
- [x] **Prod DRY-RUN (read-only)**: 303,707 blobs → 231,196 reclaimable (~5.7 GB), 70,204 young held back, 2,308 referenced kept. Nothing mutated.

## Out of scope

The young-held-back pool (last ~2h of pre-deploy leak) ages past the floor and is reclaimed by a second run a few hours later — which also doubles as confirmation that #4758 stopped new orphans (the total should shift young→old, not grow).

Refs #4757, #4759

## Summary by CodeRabbit

* **Chores**
  * Added a production ops tool to discover and reclaim leaked/unreferenced blob objects. Includes a safe discovery (dry-run) mode that lists candidates and estimates recoverable memory, plus a guarded apply mode that reclaims in chunks with safety checks, optional throttling, progress reporting, and final summary.
* **Tests**
  * Made a unit-test change to reliably simulate disabled envelope writes to prevent intermittent failures.
  * Increased several integration-test timeouts to reduce flakiness around queue dispatch and coordination scenarios.

### DIFF
--- a/langwatch/scripts/ops/reclaim-orphaned-blobs.sh
+++ b/langwatch/scripts/ops/reclaim-orphaned-blobs.sh
@@ -1,0 +1,153 @@
+#!/bin/sh
+# Reclaim orphaned GroupQueue offload blobs (one-shot prod ops tool).
+#
+# ADR-026 offloads payloads >32KiB to standalone {prefix}blob:<id> keys. Before
+# PR #4758 a dedup squash (recordSpan/reactor re-fold) overwrote a staged value
+# in place without reclaiming the blob the displaced value pointed at, so the
+# blob leaked until its 7-day TTL. On 2026-06-11 (#4757) this reached ~280K
+# orphan blobs / ~7.4GB (~91% of the keyspace). #4758 stops NEW orphans; this
+# tool reclaims the EXISTING pool instead of waiting out the 7-day TTL.
+#
+# DRY-RUN BY DEFAULT: builds the referenced set, discovers orphans, writes them
+# to OUT, prints a summary, and stops. Re-run with APPLY=1 to UNLINK them.
+#
+# A blob is reclaimed only when it is BOTH:
+#   (a) UNREFERENCED — no live {prefix}group:*:data envelope carries its id in
+#       the "r" field. This protects every STAGED job (normal, paused, retry-
+#       backoff — all re-staged into the data hash), so their blobs are kept.
+#   (b) AGED > MIN_AGE_HOURS — derived from the remaining TTL (age =
+#       BLOB_TTL_SECONDS - ttl). This protects IN-FLIGHT jobs whose value left
+#       the data hash at dispatch (so they look unreferenced) but whose blob a
+#       worker is about to read. The processing window is bounded by the active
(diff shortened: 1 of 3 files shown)


sergioestebance:
Superseding and splitting this PR — safe to close:

- **Reclaim script → langwatch-saas#580** (it's an ops tool we run against prod infra, not OSS app code).
- **The two de-flake test commits → #4798** (they fix a pre-existing `test-integration (4)` shard flake, unrelated to the script — preserved so they're not lost).

Nothing here is dropped; both pieces live in cleaner single-purpose PRs.

Issue #4742, opened 2026-06-11
flaky: groupQueue TOCTOU dedup-squash integration test fails intermittently on CI (test-integration shard 4)


drewdrewthis:
**Expected**
`langwatch-app-ci / test-integration (4)` passes on main.

**Actual**
`groupQueue.integration.test.ts > GroupQueueProcessor - Orchestration > deduplication with TOCTOU race > when dedup squash happens before dispatch > processes only once with squashed data` fails **deterministically** (not flaky — initially misread as flaky):

- main run at 2026-06-11T11:32Z (head `a6133c7d5`): https://github.com/langwatch/langwatch/actions/runs/27344600406 — FAIL
- PR #4741 (frontend-only diff; failure comes via merge-ref with main): run https://github.com/langwatch/langwatch/actions/runs/27344691741 — FAIL twice (original + rerun of failed jobs, job 80790255143 and 80792302943, same test both times)
- Prior main run at 11:00Z (head `2c6d0b526`): SUCCESS

**Root cause window**
Exactly one commit landed between the green and red main runs: `a6133c7d5` — PR #4726 "feat(event-sourcing): store GroupQueue payloads as compressed envelopes with header-only routing reads", which modifies `groupQueue.ts`, `scripts.ts`, and this very test file (`__tests__/groupQueue.integration.test.ts`, +84 lines). The regression is from that change (2× deterministic on PR merge-ref + 1× on main; test takes ~10s suggesting a timeout in the TOCTOU window).

**Reproduction**
`cd langwatch && pnpm vitest run -c vitest.integration.config.ts src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.integration.test.ts -t "processes only once with squashed data"` at `a6133c7d5` or later. CI evidence above is the observed failure (3 runs).

**Context**
Blocks CI for every open PR via merge-ref. Related GroupQueue issues: #4683, #4682, #4703.

---
*Classification: Bug*
*Status: root-cause window identified (single-commit bisect via CI history). Fix belongs to the #4726 envelope change owners.*

PR #4798, opened 2026-06-12
test(groupqueue): stabilize jobEnvelope writes-disabled case under vmThreads

sergioestebance:
Reliably simulate the envelope-writes-disabled flag in `jobEnvelope.unit.test.ts` so the case doesn't race module state under vmThreads. Test-only.

(The 20s dispatch-latency de-flake that was originally bundled here has been dropped — it masked the #4742 latency instead of fixing it. The root-cause fix is #4800.)


sergioestebance:
Dropped the 20s dispatch-latency timeout bump from this PR — it was a symptom mask. The real fix for that flake (#4742) is **#4800** (dispatcher wakes delayed jobs at their due time instead of sleeping the full BRPOP heartbeat). This PR is now just the unrelated `jobEnvelope` vmThreads stabilization.

PR #4800, opened 2026-06-12
fix(groupqueue): dispatch delayed jobs at their due time, not signalTimeout late

### LINKED ISSUE #4742: flaky: groupQueue TOCTOU dedup-squash integration test fails intermittently on CI (test-integration shard 4)

sergioestebance:
## Root cause (#4742)

`groupQueue.integration.test.ts > ... > processes only once with squashed data` fails deterministically since #4726, taking ~5s. The dispatcher has **one wake source — stage-time signals — for work that becomes actionable at other times**: a future-scored job (the `delay` option, a retry backoff, a debounced reactor re-fold) signals before it is due; that signal is consumed and drained on an early poll, and nothing re-signals at due time, so the dispatcher sleeps the full `signalTimeoutSec` (5s) BRPOP. #4726 added the first test that waits on a delayed job, exposing the latency. (The companion CI failure mode — other tests observing partial state — was parallel test files `FLUSHALL`ing the shared Redis; fixed here by `VITEST_MAX_WORKERS=1`, matching the integration config's single-worker intent.)

## Fix

**1. Exact wake for future-due work** — `peekEarliestReadyScore()` (`ZRANGE ready 0 0 WITHSCORES`, read-only) caps the BRPOP wait at the time until the earliest future-due score. Delayed jobs now dispatch at their due time (~300ms in the regression test) instead of up to 5s late.

**2. Bounded backstop for due-but-undispatchable heads** — review iterations surfaced a class of ready heads that are *due but cannot dispatch*, which a tight floor would busy-poll:
- **fastq saturation** → guard: skip the peek, wait the heartbeat (a completion signals when a slot frees).
- **retry lock window** → retry backoff frees the fastq slot but holds the group's active lock for `ceil(backoff)+2s`; `RETRY_RESTAGE_LUA` now scores ready at the **lock expiry** (derived from caller inputs, `dispatchAfterMs − backoffMs + retryTtlSec·1000`, so fake-clock tests share the time base), making the wake land exactly when dispatch can first succeed.
- **job-level pause** (pipeline/jobType/jobName) → skips the job *in place* (unlike tenant pause, which parks the group), leaving the head past-due for the whole pause window — not fixable by scoring.
- **anything not yet enumerated** → the general bound: `waitForSignal` only runs after `dispatchBatch` returned 0, so a past-due head there is by construction either undispatchable or a tiny race (e.g. a completion signal destroyed by the post-dispatch signal `DEL`). It polls at `PAST_DUE_POLL_SEC = 1s`: every undispatchable case costs ~1 cycle/sec, and the race case recovers 5× faster than the heartbeat.

## Deploy safety

The change is purely dispatcher wake-timing plus one read-only `ZRANGE`; no key semantics, no new writes, no Lua dispatch/complete changes. Mixed-version rollout is safe: old and new pods share the unchanged atomic `dispatchBatch` claim, differing only in *when* they poll. The retry ready-score change is a per-`eval` script (no cached-script skew); old pods simply scan the retry group a few seconds later than new ones, exactly once either way.

## Proof (each guard has a failing counter-test)

| Behavior | With fix | Reverted |
|---|---|---|
| Delayed job dispatches at due time | ~300ms | fails: 0 calls in 4s |
| Saturated: no busy-poll | ~0 peeks | fails: 10 peeks/s |
| Retry lock: no busy-poll, retry at lock expiry | ~1 peek | fails: 9 peeks/s |
| Job-paused: bounded backstop, dispatch ≤1s after unpause | ~2 peeks/1.5s | fails: 14 peeks/1.5s |
| Counter conservation (fake clocks) | green | broken by clock-mixing |
| Full `scripts` + `groupQueue` integration suites | **162/162** | — |

Supersedes the 20s timeout de-flake (symptom mask, dropped from #4798).

Fixes #4742

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate dispatch timing for delayed and retried jobs, bounded poll cadence for past-due work, and reduced unnecessary wake/poll activity when workers are saturated.

* **Tests**
  * New integration tests validating delayed dispatch timing, suppressed busy-polling under saturation, retry-backoff behavior with active locks, and bounded polling while a job is paused.

* **Chores**
  * CI adjusted to run integration tests serially to avoid shared Redis interference.

### DIFF
--- a/.github/workflows/langwatch-app-ci.yml
+++ b/.github/workflows/langwatch-app-ci.yml
@@ -143,7 +143,10 @@ jobs:
       matrix:
         shard: [1, 2, 3, 4, 5, 6]
     env:
-      VITEST_MAX_WORKERS: "4" # physical core count on GitHub-hosted ubuntu runners
+      # Some legacy integration suites still call FLUSHALL on the shared Redis.
+      # Keep files in one worker, matching vitest.integration.config.ts, so one
+      # suite cannot erase another suite's in-flight queue state.
+      VITEST_MAX_WORKERS: "1"
       # Required environment variables for integration tests
       DATABASE_URL: "postgresql://langwatch_ci:ci_password@localhost:5432/langwatch_test"
       BASE_HOST: "localhost:3000"


langwatch/src/server/event-sourcing/queues/groupQueue/dispatcher.ts
```diff
@@ -104,13 +104,47 @@ export class GroupQueueDispatcher {
...
+    // the wait to 50ms and busy-poll Redis (~80 cmd/s per dispatcher).
+    const inFlight =
+      this.params.processingQueue.length() +
+      this.params.processingQueue.running();
+    if (inFlight >= this.params.globalConcurrency) {
+      return this.params.signalTimeoutSec;
+    }
+    let earliest: number | null;
+    try {
+      earliest = await this.params.scripts.peekEarliestReadyScore();
```
langwatch-agent: **P2** — `peekEarliestReadyScore()` treats the lowest ready score as dispatchable, but retry backoff deliberately leaves the group active after writing that ready score. In `RETRY_RESTAGE_LUA`, the retry is scored at `dispatchAfterMs`, while `EXPIRE(activeKey, retryTtlSec)` keeps the active lock alive until `ceil(backoff/1000) + 2` seconds. Once `dispatchAfterMs` passes, this method returns a past score, the timeout floors to 50ms, `dispatchBatch()` sees the active key and dispatches nothing, and the dispatcher repeats that Redis scan every 50ms until the active key expires. Under a retry storm this recreates the busy-poll behavior the saturation guard is trying to prevent, and retries still do not actually dispatch at their due time. Fix direction: make the wake calculation skip or account for active-locked ready groups, or align the ready score with the active-key expiry / change retry locking so the score is only due when dispatch can succeed. Add a regression test for a retrying job with a live active key, not only a saturated fastq slot.

langwatch-agent (dismissed):
1 P2 concurrency/liveness regression risk in the delayed-dispatch timeout path; details inline. Local targeted GroupQueue integration tests pass, and CI is green.

sergioestebance:
Addressed the `__pipelineName` cast nit — added it as an optional field on `TestPayload` (it's routing metadata the dispatch Lua reads from the staged header) and dropped the `as unknown` cast, so the path is now type-checked. The earlier outside-diff BDD-nesting nit I'm skipping: the file's convention is topic-level outer describes (`send()`, `group coalescing`, `deduplication with TOCTOU race`) with `when …` children, which my blocks follow — renaming only the new ones would make them inconsistent.

langwatch-agent (commented):
No remaining findings from the re-review. The previous P2 is addressed: retry backoff now advertises the group as ready at active-lock expiry, and past-due-but-undispatchable heads are bounded by the 1s backstop rather than the old 50ms floor. Local targeted tests passed: `pnpm --dir langwatch run test:integration src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.integration.test.ts src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts` (162 tests). Latest CI is green.

langwatch/src/server/event-sourcing/queues/groupQueue/dispatcher.ts
sergioestebance: Fixed in 78dccb7e4 + 10c6e6fe2. Two parts:

1. **Retry score aligned to lock expiry.** `RETRY_RESTAGE_LUA` now scores the ready entry at `dispatchAfterMs - backoffMs + retryTtlSec*1000` (the active-lock expiry), not the backoff due-time — so the head is only *due* when dispatch can actually succeed. The wake lands exactly at lock expiry and the already-eligible job dispatches immediately. The lock-expiry is derived from caller inputs (not this pod Date.now()) so it stays on the same time base as every other score.
2. **General backstop for the whole class.** Your finding is one instance of *due-but-undispatchable head*; job-level pause is another (skips in place, cannot be score-fixed). Rather than enumerate, the dispatcher polls any past-due head at a 1s backstop (`PAST_DUE_POLL_SEC`) instead of the 50ms floor — bounding every such case, enumerated or not, to ~1 cycle/s.

Regression tests with teeth: a retrying job with a live active lock produces **9** dispatcher peeks/s when reverted vs ~1 with the fix; a job-paused due head produces **14**/1.5s vs ~2. Full scripts + groupQueue integration suites: 162 green.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.integration.test.ts
```diff
@@ -529,6 +530,179 @@ describe.skipIf(!hasTestcontainers)(
...
+          await scripts.addPauseKey(pipelineName);
+
+          // __pipelineName rides the payload into the staged value, where the
+          // dispatch Lua's routing-meta pause check reads it.
+          await queue.send({
+            id: "paused-1",
+            groupId: "g1",
+            value: "v",
+            __pipelineName: pipelineName,
+          } as unknown as TestPayload);
```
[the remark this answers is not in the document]
sergioestebance: Done in d8821021e — added `__pipelineName?: string` to `TestPayload` (it is routing metadata the dispatch Lua reads from the staged header) and dropped the `as unknown` cast, so the path is type-checked. Thread auto-resolved.

rogeriochaves (dismissed):

sergioestebance:
Superseded. The actual fix for #4742 turned out to be a one-line CI change (pin integration shards to a single vitest worker — the flake was a shared-Redis `FLUSHALL` race across vitest workers, not a queue bug). That clean fix is now #4817.

The dispatcher work explored here (delayed-dispatch latency + supporting correctness) was real but unrelated to the flake; it's tracked in #4814 with a working implementation preserved on `wip/groupqueue-dispatch-latency-findings`.

PR #4823, opened 2026-06-14
fix: polish traces onboarding and replicate online evaluators across projects

rogeriochaves:
## What

A batch of small bug fixes reported while dogfooding, accumulated on one PR. Two areas: the `/traces` v2 onboarding empty state, and replicating online evaluators across projects.

### Traces empty state, token card
- The freshly minted token now shows in full by default. It was rendered masked (`sk-l***...***rKF`), which made it impossible to copy at a glance even though it is shown only once. The eye toggle is still there to hide it again.
- A copy button now sits right after the "Copy this token before you move on. It won't be shown again." warning, copying the raw token directly.
- The whole `.env` block is highlighted now (API key, project id, and the endpoint line when self-hosted), not just the API key line, so it is obvious what to copy from every angle.

### Traces empty state, skills order
- The Skills and Prompt tabs in the traces onboarding now lead with "Add LangWatch tracing to your code". That is the relevant first step when you are setting up tracing. This is scoped to the traces surface via a `primarySkillId` prop; the shared onboarding flow keeps its existing order.

### Get started routing
- The home "Get started with LangWatch" → "Sync your first message" step now opens `/traces` instead of the legacy `/messages` list.
- Same redirect applied to the other get-started / observability entry points that funnel new users to their traces: the home Observability quick-access card, the integration-checks list, and the welcome integration-checks card.
- Left intentionally untouched: the sidebar's deliberate v1 ("Traces") + v2 ("Trace Explorer") dual entry, the "go back to old view" / "original view" fallbacks, backend trace deep links in Slack/email (`/messages/<traceId>`), and the analytics drill-downs (they carry v1-shaped filter query params). Happy to extend the sweep if wanted.

### Model providers onboarding check
- "Setup your model providers" was stuck incomplete for projects whose only provider is configured at the organization scope. The onboarding check matched only `PROJECT`-scoped providers, while real model-provider reads cascade `PROJECT → TEAM → ORGANIZATION`. The check now uses the same cascade, so an org-wide provider counts for every project under that org.

### Replicate online evaluators to another project
- "Replicate to another project" on online evaluator cards only appeared for legacy wizard monitors (the ones backed by an experiment) and, even then, did nothing: it looked the experiment up in a list that often didn't contain it and silently no-oped, and it only ever copied the experiment, never producing a working monitor in the target.
- The action now shows for every monitor and goes through a new `monitors.copy` that recreates the monitor in the target project. Evaluator-backed monitors (the ones created in the online-evaluation drawer) bring their evaluator, and its workflow for workflow evaluators, along; legacy wizard monitors carry their inline settings.
- Replicas are created **disabled** so a real-time evaluator does not start running and billing on the target project the moment it is copied.
- Refactored the evaluator-copy core into a shared `copyEvaluatorToProject` helper so `evaluators.copy` and `monitors.copy` share one implementation.

## Tests
- `onboarding-checks.integration.test.ts`: new cases proving the model-providers step counts org-scoped and team-scoped providers (and still ignores disabled ones), each against an isolated org/team/project so no provider leaks between cases. Runs against a real Postgres.
- RTL tests for the token card (reveal-by-default, copy button after the warning, whole-env highlight on cloud and self-hosted) and for the skill ordering (tracing-first only when `primarySkillId` is passed).
- `monitors.copy.integration.test.ts`: replicating an inline-settings monitor (recreated disabled with the same config, source untouched), an evaluator-backed monitor (evaluator copied into the target and re-linked), and name de-duplication. The existing `evaluators.copy` test still passes against the shared helper.
- Specs updated under `specs/traces-v2/onboarding-empty-state.feature`, `specs/home/`, and `specs/monitors/replicate-monitor-to-project.feature`.

## QA
Dogfooded in a real browser against the dev stack. Verified the traces empty state, minted a token, and walked the home checklist.

**Token card after minting** — token shown in full (was `sk-l***...***rKF`), a copy button sits right after the "won't be shown again" warning, and the whole `.env` block is highlighted (api key, project id, endpoint):

token card

**Skills tab leads with tracing:**

skills tracing first

**Prompt tab leads with tracing too:**

prompt tracing first

**Home "Get started" checklist** — "Setup your model providers" is now crossed out on a project whose only provider is org-scoped (the fix). "Sync your first message" navigates to `/traces`:

home checklist

**Online evaluator replication** — seeded a monitor with no backing experiment (the case where the action used to be hidden); "Replicate to another project" now appears in its menu:

monitor replicate menu

Clicking it opens a working target-project picker (it used to do nothing); selecting a project replicated the monitor into it as a disabled copy, confirmed in the database:

monitor replicate dialog

## Summary by CodeRabbit

* **New Features**
  * Added “Replicate to another project” for online evaluators/monitors, creating a disabled replica in the target project and handling evaluator/workflow replication.

* **Improvements**
  * API tokens are now revealed by default, with an inline copy button; token-related environment lines are highlighted together.
  * Tracing-related skill content now appears first in onboarding drawer tabs.
  * “Sync first message” onboarding and related quick-access/integration links now route to the traces page.

* **Bug Fixes**
  * Improved onboarding completion accuracy for model-provider setup when providers are inherited via team/organization scope.


rogeriochaves:
Thanks @coderabbitai, addressed the review in `8678b43`:

- **SRP split**: `copyEvaluatorToProject` is now a thin orchestrator over `loadSourceEvaluator` + `copyWorkflowForEvaluator`.
- **Missing workflow DSL**: a workflow evaluator with no saved version is now refused (`BAD_REQUEST`) instead of producing an evaluator with no workflow.
- **Atomicity**: `monitors.copy` rolls back the copied evaluator (and its workflow) if the monitor insert fails, so a failed replicate doesn't orphan them.
- **BDD structure**: each `given` test now nests a `when` block.
- **Spec wording**: the field-list step now describes user-observable behavior.
- **Workflow test case**: added a case asserting the workflow-evaluator-without-version path is rejected. The happy workflow-copy path is exercised by the shared helper's own test (`workflows.copyWorkflowWithDatasets.integration.test.ts`), so I didn't duplicate a full real-workflow setup here.

rogeriochaves:
Merged latest `main` (resolved the evaluators.ts + MonitorsSection.tsx conflicts) and addressed the new review note: the workflow rollback deletes in `copyEvaluatorToProject` now use `deleteMany({ id, projectId })` so the multitenancy guard accepts them instead of silently no-opping. `monitors.copy`'s own rollback already scoped by projectId. Integration tests still green against the merged schema.

rogeriochaves:
@coderabbitai review

Heads-up: since your last review I merged latest `main` (resolved conflicts in `evaluators.ts` + `MonitorsSection.tsx`, kept the `copyEvaluatorToProject` helper and always-shown replicate menu) and applied the `projectId`-scoped rollback deletes you requested in `copyEvaluatorToProject.ts` (`deleteMany({ where: { id, projectId } })`). Could you confirm the latest state looks good?

PR #3877, opened 2026-05-05
test(auth): bind 5 diagnostic-logging-on-auth-failure @unit scenarios

drewdrewthis:
## Summary

Bind 5 `@unit` scenarios in `specs/auth/diagnostic-logging-on-auth-failure.feature` to existing tests in `src/server/api-key/__tests__/auth-middleware.unit.test.ts` via `/** @scenario "<title>" */` JSDoc + dropping the `@unimplemented` tag for 4 of the 5. Pure JSDoc additions — no test logic changes.

> **Note:** The test file was relocated from `src/server/pat/__tests__/` to `src/server/api-key/__tests__/` as part of a conflict resolution with PR #3386 (which deleted the PAT path from main). The new location imports `../auth-middleware` which resolves correctly from `api-key/__tests__/`.

## Bindings

| Scenario | Test | @unimplemented dropped? |
|---|---|---|
| `extractCredentials` returns null because no auth header was sent | "returns null with no headers" | ✅ yes |
| `extractCredentials` returns null because X-Auth-Token was sent empty | "flags hasEmptyAuthToken when the header was sent but empty" | ✅ yes |
| Diagnostic fields are safe to log | "never includes a raw token value" | Partial — raw-token assertion bound; body-omission + resolver-prefix aspirational |
| Authorization header from a proxy does not poison X-Auth-Token fallback | "falls back to X-Auth-Token when Basic auth has empty token" (proxy Basic reduces to fallback path) | ✅ yes |
| Empty or whitespace-only Bearer token does not poison X-Auth-Token fallback | "falls back to X-Auth-Token when Bearer is empty" | ✅ yes |

## Still `@unimplemented`

- **Resolver returns null because credentials don't match any project** — resolver-side, not in `extractCredentials` test surface.
- **Successful auth does not emit the diagnostic log** — middleware-level integration, not exercised by this unit test.
- **Diagnostic fields are safe to log** — partially bound (raw-token assertion only); body-omission and resolver-prefix-only assertions are aspirational pending dedicated logger fixtures.

## Net parity

+5 enforced bindings (4 full drops of `@unimplemented`, 1 partial with `@scenario` annotation and `@unimplemented` retained).

## Test plan

- [x] `pnpm test:unit src/server/api-key/__tests__/auth-middleware.unit.test.ts` — 22/22 passing (CI: test-unit 1-4 all SUCCESS)
[...]

PR #4825, opened 2026-06-14
fix(groupQueue): prevent blockingConnection from inheriting maxRetriesPerRequest=0

### LINKED ISSUE #4824: ClickHouseError: test_event_handler_log table missing in test-integration shard 4 (groupQueue integration test)
## For Humans

**The groupQueue integration test "processes only once with squashed data" fails intermittently in CI (shard 4) because the dispatcher's blocking Redis connection silently inherits `maxRetriesPerRequest: 0`, causing BRPOP to throw on transient connection drops and stall the dispatcher for 10+ seconds.** The `ClickHouseError: Could not find table: test_event_handler_log` lines visible in the CI log are noise from a previous test file's cleanup — the actual failure is a `vi.waitFor` timeout in the groupQueue squash test.

**Ask:** Override `maxRetriesPerRequest: null` when creating the `blockingConnection` in `GroupQueueProcessor`, and guard `initializeClickHouseSchema()` to run once per process (not once per test file).

- Root cause is in `groupQueue.ts:184` — `effectiveConnection.duplicate()` inherits `maxRetriesPerRequest: 0` from the test Redis connection
- The CI dispatcher log at 21:07:16 (job 81313592629) confirms: `"Dispatcher loop error": "Reached the max retries per request limit (which is 0)"`
- Secondary: `testContainers.ts` runs goose migrations (`spawnSync`) twice per test file — reducible to once per process

_Status: investigation complete, plan written, ready for implementation._

### Symptoms / context

- Test `groupQueue.integration.test.ts` test #17/17 "processes only once with squashed data" fails in CI shard 4
- Same check fails on PRs #4468, #3877, #3541, #2155 (systemic, not PR-specific)
- Failure duration: exactly 10087ms — the `vi.waitFor` timeout (10 000ms + overhead)
- Shard 4 runs test files ~106–140 of 140 total; groupQueue is file #135

### Root cause / key findings

**Primary (causes the test failure):**

`GroupQueueProcessor.blockingConnection` is created via:
```
effectiveConnection.duplicate()   // groupQueue.ts:184
```

The test Redis connection is created with `maxRetriesPerRequest: 0` (`testContainers.ts:61,95`). `duplicate()` spreads `{ ...this.options, ...override }` — no override is provided — so `blockingConnection` inherits `maxRetriesPerRequest: 0`.

In CI with 6 shards running simultaneously, the `blockingConnection`'s BRPOP encounters transient errors (connection drops, ETIMEDOUT). With `maxRetriesPerRequest: 0`, ioredis throws immediately. The dispatcher catch block logs the error, sleeps 1 second, and retries. If errors recur for 10+ seconds, `processed()` is never called and `vi.waitFor` times out.

**Evidence from CI job 81313592629:**
- `21:05:34` — `× processes only once with squashed data 10087ms` (timeout)
- `21:07:16` — `"Dispatcher loop error": "Reached the max retries per request limit (which is 0). Refer to maxRetriesPerRequest option for details."` for `{event-sourcing/jobs}` (same ioredis setting, different queue, confirming the mechanism)

**Why this test and not others?** This is the only test that uses `delay: 200ms` + dedup squash together. The flow is: stage job1 (signal sent) → dispatcher consumes signal, finds job not ready yet → dispatcher waits in BRPOP (5-second fallback) → stage job2 squash (new signal) → dispatcher picks up squash signal → dispatches at T+ε. If any BRPOP errors occur during the 5-second fallback window, the dispatcher stalls and the job never processes within the 10-second window.

**Why the `ClickHouseError` is noise:** `cleanupTestData(tenantId)` in previous test files' `afterEach` attempts `ALTER TABLE test_event_handler_log DELETE WHERE TenantId = ...`. The table doesn't exist (it's created dynamically by `TestEventHandlerAppendStore.append()`), so ClickHouse returns an error. This is caught by try/catch, but `@clickhouse/client` logs the HTTP error internally. The errors appear near the groupQueue test output due to I/O timing — they are NOT from the groupQueue tests and are NOT the cause of failure.

**Secondary (CI load contributor, not the root cause):**

`startTestContainers()` at `testContainers.ts:67` calls `initializeClickHouseSchema()` on every invocation — no guard. With `setup.ts` registering its own `beforeAll(setup)` AND the test file's own `beforeAll` also calling `startTestContainers()`, this runs twice per test file. `initializeClickHouseSchema()` calls `migrateUp()` which calls `spawnSync("goose", ...)` — synchronous, blocks the Node.js event loop. With 6 shards × many test files × 2x per file, this is excessive overhead.

**Repro-attempt:**
- `repro-attempt: (b)` — the failure requires CI-level Redis load from 6 simultaneous shards. Cannot be reproduced in isolation without artificially inducing Redis connection drops on the `blockingConnection`. Local run requires `node_modules` installed in the fix worktree, which was not available during investigation.
- Confirmed via CI job logs and static code analysis. The code path is deterministic once the failure condition (BRPOP error with `maxRetriesPerRequest: 0`) is induced.

### Strategies considered

| Strategy | Assessment |
|----------|------------|
| `duplicate({ maxRetriesPerRequest: null })` in `groupQueue.ts` | **Selected.** Fixes root cause at the right layer. Low risk — `null` is the ioredis default. `blockingConnection` should always retry; BRPOP reliability is critical for the dispatcher. |
| Increase `vi.waitFor` timeout from 10s to 30s | Symptom fix only. Doesn't address why the dispatcher stalls. |
| Create a separate Redis connection for `blockingConnection` instead of duplicating | Adds complexity. Correct approach but overkill when `duplicate()` override suffices. |
| Fix `testContainers.ts` to create Redis with `maxRetriesPerRequest: null` | Affects all tests. Could mask fast-failure on real Redis outages. Wrong layer. |
| Guard `initializeClickHouseSchema()` with a per-process flag | **Selected (secondary).** Reduces CI load; harmless to test correctness. |

- `ioredis@5.10.1`: `redis.duplicate(override)` spreads `{ ...this.options, ...override }` — confirmed in built source. `duplicate({ maxRetriesPerRequest: null })` overrides cleanly.
- The `blockingConnection` is only used by the `GroupQueueDispatcher` for BRPOP. It SHOULD retry failed commands (connection resilience is essential). Setting `maxRetriesPerRequest: null` restores ioredis default behaviour.
- `GROUP_QUEUE_CONFIG.signalTimeoutSec = 5` is the BRPOP fallback interval. Test timeout is 10s. With the fix, a transient connection drop → BRPOP error → immediate retry (no sleep) → next BRPOP succeeds. No timeout.
- The `redisConnection` guard (`if (!redisConnection)`) already exists at `testContainers.ts:59`. The migration guard follows the same pattern.
- `test_event_handler_log` ClickHouseError: no fix needed. It's a benign log artifact from `cleanupTestData(tenantId)`'s try/catch. The table is created dynamically and doesn't need to be in migrations.

### Caveats

- Verify that `blockingConnection !== redisConnection` after the fix (it should differ when consumer is enabled and `duplicate` is available). The existing `close()` logic at `groupQueue.ts:1078` already handles this correctly.
- The migration guard should be scoped per-URL to handle tests that use different ClickHouse URLs (local containers vs CI service containers). A `Set<string>` keyed on the URL is safer than a boolean flag.

## Acceptance Criteria

- [ ] **AC1**: `GroupQueueProcessor` creates `blockingConnection` via `effectiveConnection.duplicate({ maxRetriesPerRequest: null })`. Confirmed by code inspection of `groupQueue.ts`. Evidence: `grep -n "duplicate" src/server/event-sourcing/queues/groupQueue/groupQueue.ts` shows `maxRetriesPerRequest: null`.
- [ ] **AC2**: All 17 tests in `groupQueue.integration.test.ts` pass in CI (no timeout on "processes only once with squashed data"). Evidence: green shard 4 on the PR.
- [ ] **AC3**: `initializeClickHouseSchema()` in `testContainers.ts` is guarded to run at most once per URL per process. Evidence: code inspection shows a `Set<string>` or equivalent flag preventing duplicate runs.
- [ ] **AC4 (regression)**: No previously-passing groupQueue integration tests are broken by this change. Evidence: `pnpm test:integration` targeting `groupQueue.integration.test.ts` exits 0 locally (or all 17 pass in CI).
- [ ] **AC5 (failure mode)**: When `blockingConnection` encounters a transient BRPOP error in CI, the dispatcher loop catches and immediately retries without the 1-second sleep stall. Evidence: no `"Reached the max retries per request limit (which is 0)"` in CI logs for groupQueue dispatcher.

## Plan

**Approach:** Two targeted one-file changes — one production (`groupQueue.ts`), one test infrastructure (`testContainers.ts`). No schema changes, no migration, no new dependencies. Lowest risk path to fixing the systemic CI flake.

**Sequencing:**

1. **Fix `groupQueue.ts`** — change `effectiveConnection.duplicate()` → `effectiveConnection.duplicate({ maxRetriesPerRequest: null })`. This is the load-bearing fix. The blocking connection must be resilient to transient Redis errors regardless of what options its source connection was created with.

2. **Fix `testContainers.ts`** — add a `Set<string>` tracking which ClickHouse URLs have been migrated this process, and skip `initializeClickHouseSchema()` on subsequent calls for the same URL. This reduces CI load and eliminates the redundant synchronous `spawnSync("goose", ...)` calls.

3. **Verify** — run `pnpm test:integration` targeting the groupQueue test file, or open a PR and let CI run shard 4. Full pass = done.

**Key decisions:**

- **Fix at `groupQueue.ts` (production), not at `testContainers.ts` (test infra)**: The `blockingConnection` should ALWAYS retry regardless of source connection settings. This is a production correctness fix, not just a test workaround. Any caller that passes a `maxRetriesPerRequest: 0` connection should still get a resilient blocking connection.
- **`maxRetriesPerRequest: null` (ioredis default), not a specific number**: `null` means unlimited retries with exponential backoff — appropriate for a long-lived blocking connection. A finite number would still fail under sustained load.
- **Migration guard keyed by URL**: Different test modes use different ClickHouse URLs (CI service containers vs local docker). A URL-keyed Set handles both correctly without breaking local dev.

**Risks / open questions:**

- Minimal risk. `maxRetriesPerRequest: null` restores ioredis default behaviour.
- The migration guard assumes the same URL always means the same database state within a process — valid for test shards (each shard is a fresh process).

**Out of scope:**

- Changing `maxRetriesPerRequest` on the main test `redisConnection` (fast-fail is correct there)
- Changing the CI shard count or test distribution
- Fixing the `ClickHouseError: Could not find table` log noise (benign, caught by try/catch)
- Changing `GROUP_QUEUE_CONFIG.signalTimeoutSec`

</details>

drewdrewthis:
## Summary

- **Root cause**: `GroupQueueProcessor.blockingConnection` was created via `effectiveConnection.duplicate()` with no option overrides. The test Redis connection uses `maxRetriesPerRequest: 0` (fast-fail for isolation), so the blocking connection inherited this — causing BRPOP to throw immediately on any transient error instead of retrying.
- **Effect**: Under CI load (6 shards), transient connection drops on `blockingConnection` sent the dispatcher into a 1-second error-sleep loop for 10+ seconds, causing the "processes only once with squashed data" test to hit its `vi.waitFor` timeout.
- **Fix 1 (primary)**: `duplicate({ maxRetriesPerRequest: null })` for standalone IORedis and a dedicated `duplicate()` for Cluster — restores ioredis default unlimited-retry behaviour for the blocking connection regardless of source connection settings, in both topologies.
- **Fix 2 (readiness)**: `waitUntilReady()` now waits across transient `error`/`close` reconnect events and resolves on `ready`, rejecting only on the terminal `end` event — so a recoverable Redis blip no longer fails pipeline startup.
- **Fix 3 (secondary)**: Guard `initializeClickHouseSchema()` to run once per ClickHouse URL per process, eliminating redundant synchronous `spawnSync("goose", ...)` calls (previously called twice per test file).

Closes #4824 — unblocks PRs #4468, #3877, #3541, #2155.

## Test plan

- [x] `groupQueue.blockingConnection.unit.test.ts` — 4/4 pass, exit 0 (constructs + asserts under Vitest 4.1.4)
- [x] `groupQueue.waitUntilReady.unit.test.ts` — 2/2 pass, exit 0 (transient reconnect resolves; terminal `end` rejects)
- [x] `pnpm typecheck` — exit 0
- [ ] CI shard passes `groupQueue.integration.test.ts` (all tests green, including "processes only once with squashed data")

## Release Notes

* **Bug Fixes**
  * Improved GroupQueue Redis blocking behavior to ensure transient failures correctly retry BRPOP even when per-request retries are effectively disabled.
  * Updated readiness handling to wait for the blocking connection to become ready, or fail fast on error/close/end.
  * Ensured deduplicated delayed jobs trigger the processor callback exactly once for rapid submissions with the same dedup key.

* **Tests**
  * Added resilience/correctness coverage for GroupQueue dispatcher/processor scenarios.
  * Prevented redundant ClickHouse schema/migration initialization by running migrations at most once per ClickHouse URL per process.

---

## ✅ Falsifiability verification (Claude, 2026-06-19) — both review blockers closed

The two P1 blockers from the `langwatch-agent` review are now fixed, each proven with a **revert → FAIL / fix → PASS** capture run locally on the PR's pinned **Vitest 4.1.4** (no Redis/ClickHouse needed — `BUILD_TIME=1`, focused files exit 0 on their own).

### AC1 — the regression test now executes and is falsifiable

The committed test's dispatcher mock used an **arrow** `vi.fn(() => ({…}))`, which Vitest 4.x cannot invoke with `new`: both consumer-mode cases threw `TypeError: … is not a constructor` at `new GroupQueueProcessor` (`groupQueue.ts:252`) **before reaching their assertions**, and CI's `process.exit(0)` hard-floor (`src/test-unit-global-setup.ts`) swallowed the non-zero exit → false green. Fixed by using constructible `class` mocks for `GroupQueueDispatcher` **and** `GroupQueueMetricsCollector` (the latter's real `start()` opens a metrics `setInterval` + Redis I/O), and force-disconnecting created connections in `afterEach`.

| Code under test | Result |
|---|---|
| broken arrow mock (as previously committed) | `2 failed \| 2 passed`, **exit 1** — `TypeError: () => ({…}) is not a constructor` @ `groupQueue.ts:252` |
| mock fixed, `instanceof Cluster` branch **reverted** | `1 failed \| 3 passed`, **exit 1** — Cluster case: `AssertionError: expected "duplicate" to be called at least once` |
| mock fixed, full fix applied (HEAD) | **`4 passed`, exit 0** — clean exit, no open handles |

The middle row is the falsifiability proof: the test genuinely catches the Cluster regression it guards.

### AC2 — readiness survives a transient reconnect

`waitUntilReady()` rejected on the first `error` (and `close`), turning a normal ioredis reconnect (`error → close → reconnecting → ready`) into a startup failure. It now waits across transient events and rejects only on the terminal `end`. New test `groupQueue.waitUntilReady.unit.test.ts` drives the exact sequence with a controllable fake blocking connection.

| Code under test | Result |
|---|---|
| over-eager reject (pre-fix `waitUntilReady`) | `2 failed`, **exit 1** — reconnect case rejected with `connect ECONNREFUSED …` instead of resolving |
| fix applied (HEAD) | **`2 passed`, exit 0** — `error → close → reconnecting → ready` resolves; terminal `end` rejects with "ended before ready" |

### Combined

`vitest run` on both files together: **`Test Files 2 passed (2)`, `Tests 6 passed (6)`, exit 0**. `pnpm typecheck`: **exit 0**.

### Browser/UI surface check (per `browser-qa` procedure)

Checked whether this fix has a user-observable surface to screenshot — **it does not**; this is a backend queue-config/infra fix:

- Diff touches **only `.ts` + `.feature`** — no `.tsx`, no pages, no components.
- `GroupQueueProcessor` is backend event-sourcing plumbing (instantiated only in `eventSourcing.ts`; feeds experiment-run-processing + event ingestion). No page renders queue / dispatcher / `blockingConnection` state (grep over `src/pages` + `src/components`: none).
- The fix changes BRPOP **resilience under transient Redis errors**, not any rendered output. Under normal operation the app behaves identically with or without it; showing the difference requires Redis fault-injection, not a browser interaction.

Per the browser-qa procedure ("When the PR has no UI surface — bail out immediately, do not browser-QA it"; "queue config" is a named trigger; canonical incident PR #4903), the **falsifiability tests above are the correct and primary proof**. No screenshots apply.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
@@ -15,17 +15,30 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import type { EventSourcedQueueDefinition } from "../../queue.types";
 import { GroupQueueProcessor } from "../groupQueue";
 
-// Prevent the real BRPOP dispatcher loop from starting — it would attempt
-// network I/O and open timer handles that outlive the test.
-vi.mock("../dispatcher", () => {
-  return {
-    GroupQueueDispatcher: vi.fn().mockImplementation(() => ({
-      start: vi.fn(),
-      requestShutdown: vi.fn(),
-      waitUntilStopped: vi.fn().mockResolvedValue(undefined),
-    })),
-  };
-});
+// The processor instantiates these collaborators with `new` in consumer mode.
+// The mock implementations MUST therefore be constructible — a
+// `vi.fn(() => ({ ... }))` arrow-returning factory is NOT a constructor under
+// Vitest 4.x and throws `TypeError: ... is not a constructor`, which previously
+// made both consumer-mode cases fail before reaching their assertions. Using a
+// class keeps the mock constructible.
+//
+// Mocking them also prevents the real BRPOP dispatcher loop and the metrics
+// `setInterval` from starting — both would attempt network I/O and open handles
+// that outlive the test.
+vi.mock("../dispatcher", () => ({
+  GroupQueueDispatcher: class {
+    start(): void {}
+    requestShutdown(): void {}
+    async waitUntilStopped(): Promise<void> {}
+  },
+}));
+
+vi.mock("../metricsCollector", () => ({
+  GroupQueueMetricsCollector: class {
+    start(): void {}
+    stop(): void {}
+  },
+}));
 
 type TestPayload = { id: string; groupId: string };
 
@@ -38,17 +51,29 @@ function makeDefinition(): EventSourcedQueueDefinition<TestPayload> {
 }
 
 describe("GroupQueueProcessor blockingConnection selection", () => {
+  const connections: Array<IORedis | Cluster> = [];
+
+  function track<T extends IORedis | Cluster>(conn: T): T {
+    connections.push(conn);
+    return conn;
+  }
+
   afterEach(() => {
+    // Force-close every connection created during the test — even if an
+    // assertion threw before the assertions completed — so no socket or
+    // reconnect timer outlives the file and wedges the runner.
+    for (const conn of connections.splice(0)) {
+      conn.disconnect();
+    }
     vi.restoreAllMocks();
   });
 
   describe("given consumer mode is enabled", () => {
     describe("when the source connection is a standalone IORedis", () => {
-      it("duplicates the connection with maxRetriesPerRequest: null for the blocking connection", async () => {
-        const conn = new IORedis({
-          lazyConnect: true,
-          maxRetriesPerRequest: 0,
-        });
+      it("duplicates the connection with maxRetriesPerRequest: null for the blocking connection", () => {
+        const conn = track(
+          new IORedis({ lazyConnect: true, maxRetriesPerRequest: 0 }),
+        );
         const dupSentinel = {} as IORedis;
         vi.spyOn(conn, "duplicate").mockReturnValue(dupSentinel as any);
 
@@ -62,16 +87,16 @@ describe("GroupQueueProcessor blockingConnection selection", () => {
           maxRetriesPerRequest: null,
         });
         expect((processor as any).blockingConnection).toBe(dupSentinel);
-
-        conn.disconnect();
       });
     });
 
     describe("when the source connection is a Redis Cluster", () => {
-      it("duplicates the connection for a dedicated blocking connection", async () => {
-        const conn = new Cluster([{ host: "127.0.0.1", port: 6379 }], {
-          lazyConnect: true,
-        });
+      it("duplicates the connection for a dedicated blocking connection", () => {
+        const conn = track(
+          new Cluster([{ host: "127.0.0.1", port: 6379 }], {
+            lazyConnect: true,
+          }),
+        );
         const dupSentinel = {} as Cluster;
         vi.spyOn(conn, "duplicate").mockReturnValue(dupSentinel as any);
 
@@ -83,19 +108,16 @@ describe("GroupQueueProcessor blockingConnection selection", () => {
 
         expect(conn.duplicate).toHaveBeenCalled();
         expect((processor as any).blockingConnection).toBe(dupSentinel);
-
-        conn.disconnect();
       });
     });
   });
 
   describe("given consumer mode is disabled", () => {
     describe("when the source connection is a standalone IORedis", () => {
-      it("uses the shared connection directly without duplicating", async () => {
-        const conn = new IORedis({
-          lazyConnect: true,
-          maxRetriesPerRequest: 0,
-        });
+      it("uses the shared connection directly without duplicating", () => {
+        const conn = track(
+          new IORedis({ lazyConnect: true, maxRetriesPerRequest: 0 }),
+        );
         const dupSpy = vi.spyOn(conn, "duplicate");
 
         const processor = new GroupQueueProcessor<TestPayload>(
@@ -106,16 +128,16 @@ describe("GroupQueueProcessor blockingConnection selection", () => {
 
         expect(dupSpy).not.toHaveBeenCalled();
         expect((processor as any).blockingConnection).toBe(conn);
-
-        conn.disconnect();
       });
     });
 
     describe("when the source connection is a Redis Cluster", () => {
-      it("uses the shared cluster connection directly without duplicating", async () => {
-        const conn = new Cluster([{ host: "127.0.0.1", port: 6379 }], {
-          lazyConnect: true,
-        });
+      it("uses the shared cluster connection directly without duplicating", () => {
+        const conn = track(
+          new Cluster([{ host: "127.0.0.1", port: 6379 }], {
+            lazyConnect: true,
+          }),
+        );
         const dupSpy = vi.spyOn(conn, "duplicate");
 
         const processor = new GroupQueueProcessor<TestPayload>(
@@ -126,8 +148,6 @@ describe("GroupQueueProcessor blockingConnection selection", () => {
 
         expect(dupSpy).not.toHaveBeenCalled();
         expect((processor as any).blockingConnection).toBe(conn);
-
-        conn.disconnect();
       });
     });
   });
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
@@ -1,0 +1,134 @@
+/**
+ * Regression test: GroupQueueProcessor must duplicate the source connection
+ * for its BRPOP loop in BOTH standalone (IORedis) and cluster (Cluster) topologies.
+ *
+ * Bug: after the PR changed the topology check from duck-type
+ * ("duplicate" in effectiveConnection) to instanceof IORedis, a Cluster
+ * connection would fall through to the else branch and share the single
+ * connection — re-introducing the "BRPOP blocks the shared connection" problem.
+ *
+ * Fix: add an `instanceof Cluster` branch that calls `.duplicate()` with no args.
+ */
+
+import { Cluster, Redis as IORedis } from "ioredis";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { EventSourcedQueueDefinition } from "../../queue.types";
+import { GroupQueueProcessor } from "../groupQueue";
+
+// Prevent the real BRPOP dispatcher loop from starting — it would attempt
+// network I/O and open timer handles that outlive the test.
+vi.mock("../dispatcher", () => {
+  return {
+    GroupQueueDispatcher: vi.fn().mockImplementation(() => ({
+      start: vi.fn(),
+      requestShutdown: vi.fn(),
+      waitUntilStopped: vi.fn().mockResolvedValue(undefined),
+    })),
+  };
+});
+
+type TestPayload = { id: string; groupId: string };
+
+function makeDefinition(): EventSourcedQueueDefinition<TestPayload> {
+  return {
+    name: `{test/gq/bc/${crypto.randomUUID().slice(0, 8)}}`,
+    process: async () => {},
+    groupKey: (p) => p.groupId,
+  };
+}
+
+describe("GroupQueueProcessor blockingConnection selection", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  describe("given consumer mode is enabled", () => {
+    describe("when the source connection is a standalone IORedis", () => {
+      it("duplicates the connection with maxRetriesPerRequest: null for the blocking connection", async () => {
+        const conn = new IORedis({
+          lazyConnect: true,
+          maxRetriesPerRequest: 0,
+        });
+        const dupSentinel = {} as IORedis;
+        vi.spyOn(conn, "duplicate").mockReturnValue(dupSentinel as any);
+
+        const processor = new GroupQueueProcessor<TestPayload>(
+          makeDefinition(),
+          conn,
+          { consumerEnabled: true },
+        );
+
+        expect(conn.duplicate).toHaveBeenCalledWith({
+          maxRetriesPerRequest: null,
+        });
+        expect((processor as any).blockingConnection).toBe(dupSentinel);
+
+        conn.disconnect();
+      });
+    });
+
+    describe("when the source connection is a Redis Cluster", () => {
+      it("duplicates the connection for a dedicated blocking connection", async () => {
+        const conn = new Cluster([{ host: "127.0.0.1", port: 6379 }], {
+          lazyConnect: true,
+        });
+        const dupSentinel = {} as Cluster;
+        vi.spyOn(conn, "duplicate").mockReturnValue(dupSentinel as any);
+
+        const processor = new GroupQueueProcessor<TestPayload>(
+          makeDefinition(),
+          conn,
+          { consumerEnabled: true },
+        );
+
+        expect(conn.duplicate).toHaveBeenCalled();
+        expect((processor as any).blockingConnection).toBe(dupSentinel);
+
+        conn.disconnect();
+      });
+    });
+  });
+
+  describe("given consumer mode is disabled", () => {
+    describe("when the source connection is a standalone IORedis", () => {
+      it("uses the shared connection directly without duplicating", async () => {
+        const conn = new IORedis({
+          lazyConnect: true,
+          maxRetriesPerRequest: 0,
+        });
+        const dupSpy = vi.spyOn(conn, "duplicate");
+
+        const processor = new GroupQueueProcessor<TestPayload>(
+          makeDefinition(),
+          conn,
+          { consumerEnabled: false },
+        );
+
+        expect(dupSpy).not.toHaveBeenCalled();
+        expect((processor as any).blockingConnection).toBe(conn);
+
+        conn.disconnect();
+      });
+    });
+
+    describe("when the source connection is a Redis Cluster", () => {
+      it("uses the shared cluster connection directly without duplicating", async () => {
+        const conn = new Cluster([{ host: "127.0.0.1", port: 6379 }], {
+          lazyConnect: true,
+        });
+        const dupSpy = vi.spyOn(conn, "duplicate");
+
+        const processor = new GroupQueueProcessor<TestPayload>(
+          makeDefinition(),
+          conn,
+          { consumerEnabled: false },
+        );
+
+        expect(dupSpy).not.toHaveBeenCalled();
+        expect((processor as any).blockingConnection).toBe(conn);
+
+        conn.disconnect();
+      });
+    });
+  });
+});
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
@@ -7,8 +7,7 @@ import {
 } from "@opentelemetry/api";
 import fastq from "fastq";
 // biome-ignore lint/style/useImportType: instanceof requires runtime class value
-import { Redis as IORedis } from "ioredis";
-import type { Cluster } from "ioredis";
+import { Redis as IORedis, type Cluster } from "ioredis";
 import { getLangWatchTracer } from "langwatch";
 import type { SemConvAttributes } from "langwatch/observability";
 import { createLogger } from "../../../../utils/logger/server";
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
@@ -6,7 +6,7 @@ import {
   TraceFlags,
 } from "@opentelemetry/api";
 import fastq from "fastq";
-import type IORedis from "ioredis";
+import IORedis from "ioredis";
 import type { Cluster } from "ioredis";
 import { getLangWatchTracer } from "langwatch";
 import type { SemConvAttributes } from "langwatch/observability";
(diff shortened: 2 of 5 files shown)


drewdrewthis:
## Review verdict: READY

**Reviewed at:** `c301df4d` · **Run:** /review (reviewer: @MrKrustyKlaws)

Re-review after the rebase + Cluster fix + lint/cleanup. Both review-opened threads are **resolved**:
1. The NOT-READY blocker — Cluster deployments silently losing the dedicated BRPOP connection — is fixed: `blockingConnection` now duplicates for both `IORedis` and `Cluster` (a `Cluster` inherits `maxRetriesPerRequest: null` from its `redisOptions`; its `.duplicate()` takes a different option shape, so it is called with no args), backed by a regression unit test (`groupQueue.blockingConnection.unit.test.ts`) that would have failed on the old `instanceof IORedis`-only code.
2. The unused `GroupQueueDispatcher` import in that test was removed (the test mocks `../dispatcher` by path string via `vi.mock`, so the symbol import was dead).

CI was fully green on `f5ad398f` — lint, typecheck, the new unit test, all 6 integration shards including the previously-flaky shard 4, e2e, codeql, sdk-python-ci. `c301df4d` differs only by removing that one dead import, and is re-running CI.

**Non-blocking (follow-ups, do not gate):**
- [test] BDD scenario 1 (BRPOP retry under transient errors) stays spec-only — it needs induced connection drops to exercise; the primary flake (scenario 2) is covered by the now-green shard-4 integration test, and the new unit test covers connection-selection for both topologies.
- [security] `waitUntilReady()` can hang if the blocking connection is already `end`/`close` when called (a `status === "ready"` fast-path exists, but no terminal-status guard before wiring listeners).
- [principles] `waitUntilReady()` teardown hand-maintains cross-`off()` calls across three reject paths; a single shared `cleanup()` closure is equivalent and less rot-prone. `bc` is an opaque local vs. the file's full-name convention.
- [principles] `queue.types.ts` `waitUntilReady` docstring still describes BullMQ semantics (replaced by fastq). Pre-existing.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -178,12 +178,10 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
     // Dedicated connection for BRPOP to avoid blocking the shared connection.
     // Only needed when the dispatcher loop runs (consumer mode).
-    this.blockingConnection = this.consumerEnabled
-      ? "duplicate" in effectiveConnection &&
-        typeof effectiveConnection.duplicate === "function"
-        ? effectiveConnection.duplicate()
-        : effectiveConnection
-      : effectiveConnection;
+    this.blockingConnection =
+      this.consumerEnabled && effectiveConnection instanceof IORedis
```
drewdrewthis: **[principles][hygiene] Cluster deployments lose the dedicated BRPOP connection — regression introduced by this PR.**

`effectiveConnection instanceof IORedis` is `false` for an ioredis `Cluster` (ioredis 5.8.2: `Cluster extends Commander`, not `Redis`). The previous `"duplicate" in effectiveConnection` check *did* match a `Cluster` (Cluster has `.duplicate()` — `built/cluster/index.d.ts:84`), so a cluster connection used to get a **dedicated** duplicated connection for BRPOP. With this change a `Cluster` falls through to `: effectiveConnection`, so the dispatcher's blocking BRPOP now runs on the **shared** connection — exactly what the comment on line 179 ("Dedicated connection for BRPOP to avoid blocking the shared connection") exists to prevent.

Reachable in prod, not just tests: `presets.ts` → `createRedisConnectionFromConfig({ clusterEndpoints })` → `new Cluster` (`redis.factory.ts:27`) → `EventSourcing.redis` → `new GroupQueueProcessorBullMq(def, conn, { consumerEnabled: true })` for worker roles (`eventSourcing.ts:402`). CI does not catch it — testcontainers use standalone `new IORedis`.

**Fix** — duplicate for both topologies. A `Cluster` already sets `maxRetriesPerRequest: null` in its `redisOptions`, and its `.duplicate()` takes a different option shape, so do not pass the IORedis option object to it:

```ts
// Cluster must be a runtime value for instanceof:
import { Cluster, Redis as IORedis } from "ioredis";

this.blockingConnection = !this.consumerEnabled
  ? effectiveConnection
  : effectiveConnection instanceof IORedis
    ? effectiveConnection.duplicate({ maxRetriesPerRequest: null })
    : effectiveConnection instanceof Cluster
      ? effectiveConnection.duplicate()
      : effectiveConnection;
```

Add a regression unit test asserting a `Cluster`-typed connection yields a **dedicated** (duplicated) `blockingConnection`, not the shared one — that is the check that would have caught this.

drewdrewthis: Fixed in f5ad398f: blockingConnection now duplicates for both IORedis and Cluster (a Cluster inherits maxRetriesPerRequest: null from its redisOptions, and its duplicate() takes a different option shape, so it is called with no args). Added regression unit test groupQueue.blockingConnection.unit.test.ts asserting a Cluster-typed connection yields a dedicated (not shared) blocking connection; it would fail on the old instanceof-IORedis-only code. langwatch-app-ci is green at this SHA including the new test. Resolving.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
```diff
@@ -0,0 +1,135 @@
...
+ * connection would fall through to the else branch and share the single
+ * connection — re-introducing the "BRPOP blocks the shared connection" problem.
+ *
+ * Fix: add an `instanceof Cluster` branch that calls `.duplicate()` with no args.
+ */
+
+import { Cluster, Redis as IORedis } from "ioredis";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { EventSourcedQueueDefinition } from "../../queue.types";
+import { GroupQueueDispatcher } from "../dispatcher";
```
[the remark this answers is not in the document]
drewdrewthis: Good catch — removed the unused GroupQueueDispatcher import in c301df4d. The test intercepts ../dispatcher by path string via vi.mock(), so the symbol import was dead (the GroupQueueDispatcher on the vi.mock factory line is an object key, not a use of the import). Resolving.

rogeriochaves (dismissed):

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { Cluster, Redis as IORedis } from "ioredis";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { EventSourcedQueueDefinition } from "../../queue.types";
+import { GroupQueueProcessor } from "../groupQueue";
+
+// Prevent the real BRPOP dispatcher loop from starting — it would attempt
+// network I/O and open timer handles that outlive the test.
+vi.mock("../dispatcher", () => {
+  return {
```
langwatch-agent: **P1** — This mock is not constructible in Vitest 4.1.4, so both consumer-mode regression cases fail before reaching their assertions with `TypeError: ... is not a constructor`. I reproduced this locally, and the supposedly green `test-unit (2)` CI log also shows this exact file failing 2/4 tests. That job only reports success because `unit globalSetup` later hits its six-minute hard floor and calls `process.exit(0)`, masking all failures. It also leaves the Cluster trying to connect, which contributes to the finalize wedge. Use a constructible class/function mock (for example, `GroupQueueDispatcher: class { start() {}; requestShutdown() {}; async waitUntilStopped() {} }`) and close/disconnect created connections in `finally`/`afterEach`; verify the focused file exits 0 on its own.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -1010,7 +1012,34 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
+    if (bc === this.redisConnection) return;
+    if (bc.status === "ready") return;
+    await new Promise<void>((resolve, reject) => {
+      const onReady = () => {
+        bc.off("error", onError);
+        bc.off("end", onEnd);
+        bc.off("close", onEnd);
+        resolve();
+      };
+      const onError = (err: unknown) => {
```
langwatch-agent: **P1** — Rejecting on `error` (and on `close` below) turns a normal transient ioredis reconnect into a readiness failure, undoing the resilience guarantee this PR is trying to restore. On an unavailable endpoint with `maxRetriesPerRequest: null`, ioredis emits `error`, then `close`, then `reconnecting`, and can later emit `ready`; this promise rejects at the first event instead of waiting for recovery. Callers awaiting pipeline readiness can therefore fail startup during a recoverable Redis interruption. Keep waiting across `error`/`close` while reconnect is enabled and resolve on `ready`; reject only on terminal `end` (or an explicit bounded timeout/abort policy). Add a test that emits `error → close → reconnecting → ready` and asserts readiness resolves.

langwatch-agent (dismissed):
2 P1 blockers — the new regression test fails on the PR's pinned Vitest version (and CI masks it), and readiness now rejects during ordinary ioredis reconnect cycles. See inline comments.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.blockingConnection.unit.test.ts
drewdrewthis: Fixed in d4884c3a9.

The dispatcher mock was an arrow `vi.fn(() => ({…}))`, which Vitest 4.x cannot invoke with `new` — confirmed locally on the pinned 4.1.4: both consumer-mode cases threw `TypeError: … is not a constructor` at `groupQueue.ts:252` (`2 failed | 2 passed`, exit 1), and CI masked it via the `process.exit(0)` hard-floor in `src/test-unit-global-setup.ts`.

Switched both `GroupQueueDispatcher` and `GroupQueueMetricsCollector` to constructible `class` mocks (the metrics collector's real `start()` opens a `setInterval` plus Redis I/O), and force-disconnect created connections in `afterEach`, so the focused file exits 0 on its own.

Falsifiable: with the mock fixed and the `instanceof Cluster` branch reverted, the Cluster case fails `AssertionError: expected "duplicate" to be called` (`1 failed | 3 passed`, exit 1); with the full fix applied, `4 passed`, exit 0. Full proof table is in the PR description.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
drewdrewthis: Fixed in 33c25fbb4.

`waitUntilReady()` now keeps waiting across transient `error`/`close` reconnect events and resolves on `ready`. It no longer listens for `close` at all, and the `error` listener only absorbs the error (which also keeps a listener attached so ioredis' `error` emit is never unhandled) instead of rejecting. It rejects only on the terminal `end` event, plus a fast-fail if the connection is already in the `end` state when called.

Added `groupQueue.waitUntilReady.unit.test.ts` driving `error → close → reconnecting → ready` (asserts readiness resolves) and a terminal-`end` case (asserts it rejects).

Falsifiable: against the over-eager reject both cases fail — the reconnect case rejects with `connect ECONNREFUSED …` instead of resolving (`2 failed`, exit 1); after the fix, `2 passed`, exit 0.

drewdrewthis:
@rogeriochaves — re-requesting your review. Your earlier approval was auto-dismissed when I pushed the fixes below, so flagging what changed since:

The two **P1 blockers** from the `langwatch-agent` review are now resolved (both threads replied to + resolved):

1. **Regression test was a false-green** — the dispatcher mock used an arrow `vi.fn(() => ({…}))`, which is not constructible under the pinned Vitest 4.1.4, so both consumer-mode cases threw `TypeError: … is not a constructor` before asserting; the unit `process.exit(0)` hard-floor masked the non-zero exit. Fixed with constructible `class` mocks (`d4884c3a9`). The previously-masking `test-unit` shard now passes genuinely.
2. **Readiness rejected on transient reconnect** — `waitUntilReady()` now waits across transient `error`/`close` and rejects only on the terminal `end`, so a recoverable Redis blip no longer fails pipeline startup (`33c25fbb4`), with a new `error → close → reconnecting → ready` test.

Each fix carries a captured **fail → fix → green** falsifiability proof (tables in the PR description). **CI is green** — `gh pr checks` shows 0 failing / 0 pending (the one sdk-python e2e red was a transient live-backend 500, cleared on re-run). Nothing else outstanding; over to you for re-approval.

<sub>🤖 Posted by Claude Code while driving the PR's review blockers to resolution. Not merging — that's Drew's call.</sub>

rogeriochaves (approved):

Issue #4673, opened 2026-06-08
feat(licensing): gate SSO behind active license or IS_SAAS (close env-var bypass)


sergioestebance:
## Problem

SSO has **no license enforcement** on any of its three surfaces. A self-hosted operator with no paid license can enable Auth0 SSO by setting four env vars; nothing in the code consults plan state before activating the OAuth provider. Per-org `ssoDomain` auto-provisioning has the same shape — set a column, get the feature.

`LicensePlanLimitsSchema` has no SSO field today, so there is no flag to gate on either.

## Proposed rule

```
canPlatformSSO = IS_SAAS === true || anyOrgHasActiveLicense()
canOrgSSO(orgId) = IS_SAAS ? getActivePlan(orgId) !== FREE_PLAN
                           : orgHasActiveLicense(orgId)
```

Default: **denied**. When denied, `NEXTAUTH_PROVIDER` resolves to `"email"` regardless of what the env var says — silently, no error, no lockout.

## Disruption guarantees

| Customer state | Result |
|---|---|
| SaaS (any plan) | No change — `IS_SAAS=true` short-circuits platform gate |
| Self-hosted paying (≥1 org has valid license) | No change — `anyOrgHasActiveLicense()` returns true |
| Self-hosted free / no license | **Gated** — env-var Auth0 ignored, `ssoDomain` writes rejected |
| Self-hosted license expires | Falls back to email at next provider resolution |

This is the only behavior change, and it is the intended one.

## Surface map (gate sites)

**Platform gate (`canPlatformSSO`)**
1. `langwatch/src/server/better-auth/index.ts:136-168` — Auth0 provider registration
2. `langwatch/src/server/api/routers/publicEnv.ts:24` — `NEXTAUTH_PROVIDER` exposure to UI
3. `langwatch/src/server/routes/auth.ts:137-142` — federated logout (follows #1, no separate gate)

**Per-org gate (`canOrgSSO`)**
4. `langwatch/src/server/better-auth/hooks.ts:115` — `afterUserCreate` `ssoDomain` auto-join
5. `langwatch/src/server/better-auth/hooks.ts:253-269` — `beforeAccountCreate` provider match
6. `langwatch/src/server/better-auth/hooks.ts:344` — `afterAccountCreate` org lookup
7. `ee/admin/backoffice/resources/OrganizationsView.tsx:443-462` + Hono update route — admin write of `ssoDomain` / `ssoProvider`
[...]

PR #4830, opened 2026-06-15
feat(licensing): license-gated SSO — signature-only gate, decided at startup (ADR-027)

sergioestebance:
## What & why

Stacks on #4480. Gates **every non-email login provider** (SSO) behind possession of a genuine license on self-hosted LangWatch, closing the four-env-var free-SSO bypass. SaaS is unaffected (`IS_SAAS=true` short-circuits the gate). Part of the same commercial repositioning as #4480: the OSS-lift removes experimentation caps, this enforces Enterprise security (SSO) as the remaining paid surface.

Tracking: #4673

## This PR: contract AND implementation

- **ADR-027** (`dev/docs/adr/027-license-gated-sso.md`), the locked decision (**Accepted**, v6, red-teamed twice + implementation review).
- **Feature spec** (`specs/licensing/sso-license-gating.feature`), 17/17 scenarios implemented and bound (`pnpm check:feature-parity` green, zero `@unimplemented`).
- **Implementation**: the full gate, in this PR:

| Piece | File | What it does |
|---|---|---|
| Gate module | `src/server/sso/sso-gate.ts` | `platformSSOAllowed()`: `IS_SAAS` short-circuit → `LANGWATCH_LICENSE_KEY` → any org with a signature-valid license; memoized once per process (evict-on-reject); `resolveAuthProvider()` coerces to `"email"` on deny |
| License repository | `src/server/sso/sso-license.repository.ts` | Only Prisma toucher (3-layer); `select {id, license}` where license not null |
| Hook blocks | `src/server/better-auth/index.ts` | DENY: 403 SSO initiation + all callback paths (incl. legacy `/callback/auth0\|okta`); ALLOW: refuse email sign-up/sign-in + reset pair; reset pair OPEN on deny; matching on normalized pathname (trailing-slash safe) |
| Email routes | `src/server/better-auth/index.ts` | `emailAndPassword.enabled = provider==="email" \|\| !IS_SAAS` (mounted for coerced mode, hook-guarded on ALLOW) |
| Auto-join guard | `src/server/better-auth/hooks.ts` | ssoDomain auto-join skipped when the gate denies (org-takeover fix) |
| Provider exposure | `src/server/api/routers/publicEnv.ts` | Frontend sees the resolved (coerced) provider |
| Signup path | `src/server/api/routers/user.ts` | tRPC `register` keyed off the resolved provider, the signup form works in coerced mode, refused on licensed SSO installs |
| Logout | `src/server/routes/auth.ts` | No Auth0 federated-logout bounce on denied installs |
| Activation UX | `src/components/license/useLicenseActions.ts` | Self-hosted success toast: "restart the server to enable SSO" |
| Env | `src/env-create.mjs`, `.env.example` | Optional `LANGWATCH_LICENSE_KEY` (instance license, recovery/bootstrap) |

## Decision (one line)

Every non-email provider is paid, gated **binary on possessing a genuinely-issued license** (**signature-valid, expiry deliberately ignored**: once a customer, never blocked), **decided once per process** (restart to change); a denied deployment runs in **email mode** with password reset open for self-recovery; **zero action** for any customer who ever held a license.

### Load-bearing choices
- **Signature-only, expiry ignored (v6).** The gate targets free-riders who can't forge a signature; an expired license is a renewal-limbo/ex-customer, and an upgrade IS a restart, so strict expiry would cut company-wide login the morning they pull a new image. Plan limits keep strict expiry; renewals get a warn-log nudge, not lockout.
- **Startup semantics.** One memoized function; no truth table, no TTL cache, no three-state gate. DB errors deny without memoizing (self-healing).
- **No license schema change.** `enableSso` deferred (addable later, `absent → true`).
- **Denied = coerce to email mode.** Credential-mutation routes blocked in every state; on ALLOW email sign-up/sign-in/reset refused (no password-account minting on licensed SSO installs); on DENY the reset pair opens so OAuth-born users self-recover (inbox-proof; upstream premise pinned by a real-better-auth test).
- **ssoDomain auto-join rides the platform gate**: no unauthenticated org-join via email signup in coerced mode.
- **Telemetry: server logs only**: gate-resolution warn, per-candidate license verification outcome, expired-license renewal nudge, per-blocked-request path+reason.

### Red-team + review (all folded into the ADR / fixed here)
- **v5 BLOCKER**: email routes mounted with nothing re-blocking them on ALLOW → ALLOW-path block.
- **v5 MAJOR**: denied-mode signup + persisted `ssoDomain` = unauthenticated org takeover → auto-join guard.
- **v5 MAJOR**: phantom `/oauth2/authorize`, missing `/sign-in/oauth2`, unblocked `/link-social` → corrected path list.
- **v6**: strict expiry + upgrade-is-a-restart lockout → signature-only; denied installs hard-stranded users → reset pair opens on DENY.
- **Impl review P1**: signup form's tRPC path was still env-keyed (dead recovery route) → resolved-provider guard.
- **Impl review P1**: trailing-slash variants walked past the hook matcher (router resolves them) → normalized-pathname matching + regression tests.

## Verification

- 135 tests green across gate / repository / hook / hooks / publicEnv / register / index suites; upstream reset premise pinned against real better-auth (memory adapter).
- `pnpm typecheck` 0 errors; `pnpm check:feature-parity` 2284 bound (17/17 here).
- SSO hook smoketest (`e2e/auth-regression/better-auth-sso-smoketest.ts`) updated: mints a throwaway signed instance license, adds a denied-mode auto-join-off check.

## Deployment Impact

- **New optional env var**: `LANGWATCH_LICENSE_KEY` (self-hosted instance license), credential-less SSO recovery/bootstrap. Documented in `.env.example`. Helm chart plumbing not included (the existing `LANGWATCH_LICENSE_PUBLIC_KEY` isn't chart-plumbed either; follow-up alongside it).
- **Behavior change, self-hosted only**: SSO env vars alone stop being sufficient: SSO additionally requires a genuine (signature-valid) instance/org license or `IS_SAAS=true`. **Any customer who ever held a license sees zero change** (expiry ignored for the SSO gate; DB license honored, no new env var, existing sessions survive). Never-licensed installs using Google/GitHub/Auth0 sign-in are coerced to email mode on upgrade, with password reset + fresh signup as self-service doors. This needs a loud release note.
- **License changes take effect on the next restart** (startup semantics); the activation toast says so.
- **SaaS**: unaffected at every stage (`IS_SAAS` short-circuits before any DB read).
- No Prisma schema change; no migration.
---

### Rebase

Merged the rebased #4480 (which now carries all of `main`). Resolution rule: take the incoming version, re-apply the gate on top. Adaptations to main's newer shape:

- main extracted `isEmailPasswordEnabled(e)` as a testable pure helper, so the self-hosted mounting rule lives in the helper rather than inline at the config site, and the helper now reads `IS_SAAS`.
- main moved the logger to `@langwatch/observability`; `sso-gate.ts` and the two gate test files followed.
- `user.register.unit.test.ts` was created independently on both sides. The suites cover different things (the PostHog `signed_up` milestone vs the ADR-027 provider coercion) and are merged into one file rather than one replacing the other.

### Dogfood

Run against a throwaway Postgres, a purpose-generated RSA keypair via `LANGWATCH_LICENSE_PUBLIC_KEY`, and `NEXTAUTH_PROVIDER=auth0` with Auth0 credentials set, i.e. exactly the four-env-var free-SSO bypass this PR closes.

**Unlicensed:**

| Check | Result |
|---|---|
| `POST /api/auth/sign-in/social` | 403 `SSO_LICENSE_REQUIRED` |
| `POST /api/auth/sign-in/oauth2` | 403 `SSO_LICENSE_REQUIRED` |
| `GET /api/auth/callback/auth0` | 403 `SSO_LICENSE_REQUIRED` |
| Sign-in page | email + password form, no IdP button, "Forgot password?" present |
| `POST /sign-up/email` | succeeds (coerced email mode) |
| `POST /request-password-reset` | 200 (open on deny, so OAuth-born users self-recover) |
| Active plan | `OPEN_SOURCE`, members/lite/messages all unlimited, no visibility window |
| Inviting a second member | succeeds, the operation the customer was blocked on |
| Creating projects | 4 created, past the old 2-project cap |
| Usage page | reads "Open Source", no caps shown |

**Then activated a license through the UI** (Settings > License > paste key): shows ENTERPRISE, 2/100 seats.

| Check | Result |
|---|---|
| SSO **before** restart | still 403, gate is decided once per process as designed |
| SSO **after** restart | 200, returns the Auth0 authorize URL |
| `POST /sign-up/email` when licensed | 400 `EMAIL_PASSWORD_DISABLED` |
| `POST /sign-in/email` when licensed | 400 `EMAIL_PASSWORD_DISABLED` |
| `POST /request-password-reset` when licensed | 400 `EMAIL_PASSWORD_DISABLED` |
| Existing session across the gate change | survives |
| `publicEnv.NEXTAUTH_PROVIDER` | `auth0` when licensed, `email` when not |

### Docs and product surface

The SSO doc was factually wrong as of this branch: setting the env vars is no longer sufficient. It now leads with the license requirement, the email-mode fallback, and the restart semantics.

Added:

- **`self-hosting/licensing`**: the page that answers "what does self-hosting cost": everything on your own infrastructure is uncapped, a license buys the Enterprise security surface and support, how to activate one, and what expiry does. Linked from the self-hosting overview, the SSO doc, the pricing page, and the README.
- **`pricing/billable-events`**: the seven event types the meter actually counts, taken from the billing projection rather than from memory, with the arithmetic for an experiment sweep and a simulation, what is never billed, and how deduplication works. The pricing page's "what counts as a billable event" answer now links to it, and so does **Settings > Plans** in the app: the plan columns quote event allowances, so a "What counts as an event?" link now sits under the header and opens the reference in a new tab.
- **Settings > Authentication** on self-hosted now carries an *Organization sign-in and governance* section listing SSO, SCIM and audit logs, each with what it does, whether this deployment has it, and a link to its setup guide. Hiding a paid capability makes it look missing rather than purchasable, and there was previously no in-product route to the SSO guide at all. Cloud renders nothing there. Spec: `specs/licensing/self-hosted-enterprise-discovery.feature`, 4/4 bound.

Website counterpart: langwatch/2lang-2watch#125 adds "Unlimited users, teams and projects" to the Open Source self-host card.

### The license now covers what it sells

Until this branch, SSO, SCIM and the audit trail were named as the enterprise surface everywhere while their implementations sat in `src/` under Apache 2.0: a fork could lawfully delete the one gate call and keep the whole capability. The implementations now live under `langwatch/ee/` (`ee/sso/` with the IdP builders, path predicates, domain matching and the platform gate; `ee/scim/` with the v2 API, services, token management and the Auth0 SCIM webhook; `ee/audit-log/` with the write path), each carrying the `LicenseRef-LangWatch-Enterprise` SPDX header. `src/server/better-auth/index.ts` stays the Apache assembly point.

`ee/LICENSE.md` is restructured to match: running ee code as part of a distribution is expressly allowed (the old text forbade what every unlicensed deployment does by booting, since plan resolution runs `ee/licensing` on every request), using the enterprise capabilities in production requires a license, and the running grant does not extend to a distribution whose license checks were removed or bypassed, wherever that edit was made. Worth a pass from a lawyer before it is leaned on in a dispute, but the structure is the standard open-core one.

README licensing section rewritten to one sentence per fact: Apache and free to use, enterprise modules under `ee/`, SDKs MIT.

### Review follow-up

The July review left four actionable points open. All four are answered on this branch.

**The federation set was a blocklist with no upgrade guard.** It stays a blocklist, because DENY has to leave the whole email-mode surface working and that surface is far larger and more fragile to enumerate than the federation one. What it now has is a canary: `ssoRouteTableCanary.test.ts` builds a real better-auth instance with `socialProviders` + `genericOAuth`, reads the mounted route table off `auth.api`, and requires every route to carry a reviewed `federating` / `local` classification, asserting the gate's verdict against it in both the bare and trailing-slash forms. A version bump that adds or renames a route fails by name. Verified by deleting one entry: three assertions go red naming `/sign-in/oauth2`.

That canary found a live false positive on the way in. The callback check matched against the un-normalized pathname, so `/delete-user/callback/` was refused on an unlicensed deployment. It matches the normalized one now.

**Session reads waited on the licensing store.** The hook returns before consulting the gate for any path whose outcome the gate cannot change, using a predicate derived from the branches it guards rather than a hand-listed set of safe paths. The new test mocks the gate as a promise that never settles, which is the only way to show the difference: a resolved mock passes either way.

**A slow store had no deadline anywhere.** The scan now carries one, landing on the same evict-and-retry path a hard DB error already takes, so a store that recovers still needs no restart.

**Helm could not set a license.** `app.license.key` and `app.license.publicKey` take an inline value or a `secretKeyRef`, in `sharedEnv` rather than the app Deployment: plan resolution runs in the workers too, and a worker reading a different entitlement than the app would enforce different limits on the same organization.

### Expiry: the two halves now agree

The base branch changed what a license past its end date means for seats, so the SSO half is re-stated here rather than left to drift.

ADR-027 v6 split them deliberately: SSO signature-only ("once a customer, never blocked"), plan limits strict expiry, with the renewal pressure explicitly assigned to plan limits. Dogfooding the base branch showed that pressure did not exist. An expired license resolved to `UNLIMITED_PLAN`, so a lapsed 3-seat license granted unbounded seats and reported the deployment as unlicensed. Lapsing beat renewing.

Self-hosted plan resolution is signature-only now too, which makes the policy one rule rather than two: a license we signed keeps what it sold, and only the ability to grow needs a renewal. **SSO is untouched by that change** and this branch's gate is byte-identical, so the "no login lockout" invariant is unaffected.

Recorded as ADR-027 **v8** in the Revisions section, with the two stale claims in Context and Consequences corrected in place. `docs/self-hosting/licensing.mdx` had the same stale sentence ("Plan entitlements do follow expiry") and now describes what actually happens on the day a license lapses, what still binds, and how a key we did not issue differs. `llms.txt` regenerated.

### Verification

`pnpm typecheck` + `pnpm typecheck:tests` clean; 211 tests green across the sso / better-auth / register / publicEnv / licence-UI suites; `pnpm check:feature-parity` green at 3963 bound scenarios.

After merging the base branch up: 633 unit tests and 278 integration tests green across sso / licensing / license-UI / settings; `pnpm typecheck:all` clean; `pnpm check:feature-parity` green at 4010 bound scenarios.

Also fixed on the base branch: `groupQueue.decodeDrop` asserted a lease was retired the instant the drop counter moved, but the release is a Redis round-trip issued after it. It now polls, which is what a loaded CI runner needs.

### Activation review: the doors a licensed deployment could still be locked out of

A second dogfooding pass walked the licensed-customer activation story end to end and found six real defects, all fixed here. Every finding was checked against HEAD before it was acted on; three more that came in the same batch were already answered on this branch and are recorded at the bottom rather than re-fixed.

**A named-but-unmountable provider left nobody a way in.** #4772 made the sign-in page fall back to the email form when `NEXTAUTH_PROVIDER` names a provider the build cannot mount, but only the page. The before-hook's ALLOW branch still refused `/sign-in/email`, `/sign-up/email` and the reset pair on the sole test of whether the gate allowed, so a licensed deployment with a typo in its provider name rendered a working form that answered `400 EMAIL_PASSWORD_DISABLED` on submit, and had no SSO button either. The refusal now runs through `refusesCredentialRoute`, which additionally requires that `resolveAuthProvider()` actually resolved to something other than email, which is the same three-way composition (env names a provider, gate allows, provider mounts) the page already uses. The spec line about nobody being left without a way in is bound to the hook now rather than to the page alone.

**Forgot-password promised a link it could not send.** With no email transport configured the screen still answered "we have sent a reset link", so the one door left open on a denied deployment led to an inbox that was never going to receive anything. It now branches on `HAS_EMAIL_PROVIDER_KEY` and says plainly that the deployment cannot send email and who to ask instead.

**The restart instruction did not outlive the activation it belonged to.** `useLicenseActions` raised the "restart the server to enable SSO" toast and called `window.location.reload()` in the same tick, so the one sentence telling an admin why SSO is still off after paying was on screen for a few frames. It invalidates the tRPC cache instead, which refreshes the plan, navigation and feature gates the reload was there for. The regression announces itself in a test now: the spy on `location.reload` is what fails if it comes back.

**An unlicensed self-hosted deployment was told it was on the Cloud Free tier.** `/settings/plans` resolved `OPEN_SOURCE` to the `free` column and marked it Current, so the page read "Current: 2 users, 50k events" on a deployment where members and events are uncapped. The resolver returns no current plan for `OPEN_SOURCE` now, which is the honest answer: none of the Cloud tiers is what that deployment is running.

**A correctly-pasted public key read as a bad signature.** `LANGWATCH_LICENSE_PUBLIC_KEY` arrives through a `.env` line, a Helm value or a Kubernetes secret, all of which routinely carry PEM as one line with escaped newlines. `signLicense` normalizes the private key through `normalizePemKey`; `verifySignature` did not, so the layout OpenSSL rejects surfaced as "signature verification failed" and sent the operator to look at their license instead of at their formatting. Both ends go through the same choke point now.

**Nothing in the product said "configured but not licensed".** The gate logs the coercion at startup, but nobody reads server logs to explain a login screen, and Settings > Authentication showed an identity provider section with no hint that the reason everyone is typing a password is a license. A new `license.getSsoGateStatus` query feeds a notice at the top of that section naming the configured provider and what to do about it.

Specs: `specs/licensing/license-activation-feedback.feature` (5 scenarios) is new, `sso-license-gating.feature` gains 4, all bound.

**Verified as already fixed, not re-fixed.** The same batch flagged `?? 0` hiding an unknown usage count (the usage service returns `null` for unknown and documents why), relative links in the README that would break on npm (there are none), and a silently denied auto-join in `hooks.ts` (the warn is there).

### Verification of this round

`pnpm typecheck:all` clean. `pnpm check:feature-parity` green at 4019 bound scenarios. Full unit sweep 2425 passed, integration sweep 1664 passed. Biome new-violations gate run over CI's exact path list: 17 violations removed by this PR, none added.

Two negative controls, because a passing test proves less than a failing one: neutering `getSelfHostedPlan` fails 3 integration tests with the exact seat messages they assert, and reverting the `normalizePemKey` call fails the escaped-newline test.

### Proven on a running deployment

The R1 fix is the kind that a test can assert and a deployment can still get wrong, so both halves were run against a QA stack holding a real signed license, the same license in both runs, differing only in whether the named provider's credentials were present.

| `NEXTAUTH_PROVIDER` | Credentials | Resolved provider | `POST /sign-in/email` | `POST /sign-up/email` | `POST /request-password-reset` |
|---|---|---|---|---|---|
| `okta` | absent | `email` | `401 INVALID_EMAIL_OR_PASSWORD` | `200` | `200` |
| `okta` | present | `okta` | `400 EMAIL_PASSWORD_DISABLED` | `400 EMAIL_PASSWORD_DISABLED` | `400 EMAIL_PASSWORD_DISABLED` |

The first row is the case that was broken: a licensed deployment with a typo in its provider name served a working-looking email form that refused every submit, and offered no identity-provider button either. A 401 on a deliberately wrong password is the proof the route is live, not merely rendered. The second row is the guard that had to survive the fix, and did.

### Follow-up review round

Four documentation and test findings from the automated reviewer, all fixed. The SPDX table prescribed a `//` comment for `python-sdk/`, where the delimiter is `#`. Two claims on the licensing page were written before a licensed deployment could stay in email mode and read as contradictions: a license closes the email routes only once the configured provider actually starts, and `LANGWATCH_LICENSE_KEY` recovers an SSO-only deployment only while its provider variables are still in place. The key licenses SSO, it does not configure it.

The e2e for the plans page asserted the Free column carried the Current badge. That badge was the defect this branch removed, so the assertion is inverted: on a self-hosted deployment with no license, no column claims to be current, and the e2e is now a guard for the fix rather than a witness for the bug.

### Gateway webhooks join the enterprise surface

Platforms that run their own usage-based pricing on top of the AI Gateway need spend and budget events delivered to their systems with billing-grade guarantees. That delivery is licensed, and no list of what an Enterprise license includes said so. It is now named on the licensing page, the open-core map, the pricing page, the in-app plan comparison, the README, and the enterprise license text itself, described by what it does for the buyer rather than by how it is built.

Two things deliberately left out. There are no links to the webhooks reference: those pages arrive with the gateway billing platform in #6395, and an internal link to a page that does not exist fails the docs link check. And the in-product capability list under Settings > Authentication is untouched, because it badges each capability against what *this deployment* can do. Listing webhooks there before the feature ships would show a licensed customer "Available" for something not in the build. #6395 is the right home for that row, since it already carries the `webhookEndpointsEnabled` flag the badge should read.

### Full and lite seats, defined by what the person can do

Seats are the one thing a paid plan still meters after #4480, so "does this person need a full seat?" is a billing question an admin answers for every hire. The product asked it the other way around: a "Lite Member" label with a one-line description of two of its effects. Admins were guessing at the boundary and asking their account manager instead.

The framing is capability everywhere now. Someone who can change things holds a full seat, someone who can only read what the team produces holds a lite one. That is already how seats are counted rather than an aspiration: `classifyMemberType` promotes a lite member to a full seat the moment a custom role grants them anything beyond viewing.

The invite form and the role picker carried the same sentence twice; they share one copy constant now, plus an (i) listing what a lite member can actually see. That list is pinned to `EXTERNAL_MEMBER_PERMISSIONS` by a test, so the sentence an admin reads before paying cannot drift from what the permission set grants. Verified by dropping a resource from the copy: the test goes red naming it.

The docs get the full breakdown under Seats in the roles reference, reachable from the pricing FAQ and the licensing page. Two corrections found while writing it against the code rather than from memory:

- **Trace viewing is included in a lite seat.** Every trace procedure, `getById` and its spans included, gates on `traces:view` alone, and that permission is in the lite set. There is no separate "debug the trace" permission, so anyone who has been told otherwise needs a custom role that withholds `traces:view`.
- **A lite member can see the names of a project's secrets**, never their values. Not a hole, but a table listing what someone can see should not leave a reviewer to find that row on their own.

The members doc also described the org role and the lite tick as two separate controls. They are one.

Spec: `specs/licensing/seat-type-explained.feature`, 3/3 bound.

## Summary by CodeRabbit

* **New Features**
  * Added self-hosted Enterprise licensing with inline or Secret-based activation.
  * Added license-aware SSO with safe email-mode fallback.
  * Added capability discovery for SSO, SCIM, and audit logs.
  * Added billable-event guidance and links from plan comparisons.
  * Added clearer Lite Member seat explanations and Enterprise feature details.
  * Improved license activation feedback and password-recovery messaging.
* **Documentation**
  * Expanded licensing, self-hosting, SSO, billing, retention, and seat guidance.
  * Updated supported providers and removed AWS Cognito configuration.
* **Tests**
  * Added coverage for licensing, SSO gating, recovery, and capability visibility.


sergioestebance:
Blessed. 🙏

Traced the gate end-to-end — functional, security, and design passes all clean. Memoize-success-only with correct evict-on-reject, `IS_SAAS` short-circuits before any DB touch, signature-only per v6, trailing-slash bypass closed, coercion consistent across every frontend surface. The two P1s from the first pass are fixed and verified against the vendored router source; the one P2 cohesion finding (path predicates now live in `ssoPathGate.ts`, matching `originGate.ts`) is folded in. 144 tests green.

*"Whoever is faithful in a very little is also faithful in much." — Luke 16:10*

Ship it once #4480 lands. Mind the two documented follow-ups (Helm plumbing for LANGWATCH_LICENSE_KEY; org soft-delete filter if that column ever ships) — neither blocks.

sergioestebance:
Blessed. 🙏 Post-rebase: 82/82 gate tests green, typecheck clean, all three states verified live (deny → email, org-license allow, instance-key signature).

_"Open ye the gates, that the righteous nation which keepeth the truth may enter in."_ — Isaiah 26:2

Ship it — the gate opens for those who kept the covenant, closes for the rest. Only residuals are P3 diff-noise (unrelated biome reformat) and the base #4480 approval.

langwatch/ee/sso/ssoPathGate.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * ADR-027 gate site #2 — true for any request refused while the platform SSO
+ * gate denies: the initiation paths, plus ANY callback path (pathname-PREFIX
+ * match via `includes`, since callbacks carry `?code=&state=` and a provider
+ * segment, e.g. `/callback/auth0`, `/oauth2/callback/okta`). This is the only
+ * interception point that sees the legacy `/api/auth/callback/auth0|okta`
+ * rewrite (`redirectURI` pinned in index.ts), so a path-prefix middleware on
+ * `/oauth2/*` alone would miss it entirely.
+ */
+export const isGatedSsoPath = (url: string): boolean => {
```
Aryansharma28, line 117: [P1] BLOCKER — blocklist, not allowlist, at a security boundary.

**Why:** `isGatedSsoPath` only refuses the specific SSO-shaped paths enumerated in `GATED_SSO_INITIATION_SUFFIXES` plus a `/callback/` substring match. That list was carefully verified against `better-auth@1.6.12` + `genericOAuth` — great work for *today*. But it's a static, hand-maintained blocklist guarding the DENY state, and DENY is supposed to mean "as if the SSO env vars were unset." Any future better-auth endpoint that federates login (a new plugin, a renamed/added OAuth route, passkeys, magic-link) but isn't added to this list will pass through **unblocked** when the license gate denies — silently reopening the exact free-SSO bypass this PR exists to close, with no test failure to catch it (the test suite proves today's list is right, not that tomorrow's route table is covered).

**Evidence:**
```ts
export const isGatedSsoPath = (url: string): boolean => {
  const pathname = normalizedRequestPathname(url);
  if (endsWithAny(pathname, GATED_SSO_INITIATION_SUFFIXES)) {
    return true;
  }
  return requestPathname(url).includes("/callback/");
};
```

**Suggested fix:** Invert the DENY-state check to a default-deny/allowlist model: when the gate denies, block by default and explicitly allowlist the known-safe passthrough paths (`get-session`, `sign-out`, `list-sessions`, the email/reset routes already handled elsewhere in the hook). If a full inversion is too invasive right now, at minimum add a canary test that enumerates `auth.api`'s actual registered route table (or the configured plugin's route list) and fails loud when a non-email-mode endpoint appears that isn't covered by this module — so a `better-auth` version bump can't silently regress this the way the v5 `/oauth2/authorize` phantom-path bug almost did.

langwatch/src/server/better-auth/index.ts
```diff
@@ -542,35 +583,67 @@ export const auth = betterAuth({
...
+        throw APIError.from("BAD_REQUEST", {
+          code: "EMAIL_PASSWORD_DISABLED",
+          message:
+            "Credential management is disabled in cloud/SSO mode — your account is managed by your identity provider.",
+        });
+      }
+
+      const isResetPath = isPasswordResetPath(pathname);
+
+      if (await platformSSOAllowed()) {
```
Aryansharma28, line 490: [P2] BLOCKER — unconditional, un-timed-out gate call on every request for SSO-capable deployments, including plain session reads.

**Why:** For any deployment where `NEXTAUTH_PROVIDER !== "email"`, this line runs on *every* better-auth request that isn't a credential-mutation path — that includes `/get-session`, `/sign-out`, `/list-sessions`. Before the gate's memoized promise first resolves, a slow (not erroring — just slow) DB read in `anyOrgHasSignedLicense()` blocks all of those requests with no timeout, including session validation for already-authenticated users on a *licensed* install. That cuts against the ADR's own claimed invariant that existing sessions "keep working" through the gate — they don't break, but they can hang indefinitely during the warm-up window, and nothing in this PR bounds that.

**Evidence:** the existing regression test for this (`ssoGate.hook.test.ts`, "never intercepts session reads or sign-out in either gate state") mocks `platformSSOAllowed` to resolve synchronously in both directions, so it verifies *correctness* of the branch logic but can't and doesn't verify *availability* under a slow DB — there's no test exercising a pending/slow gate promise against `/get-session`.

**Suggested fix:** short-circuit known gate-independent paths (`get-session`, `sign-out`, `list-sessions`) before ever calling `platformSSOAllowed()`, and/or wrap the DB scan in `anyOrgHasSignedLicense()` (`sso-gate.ts`) with a timeout so a slow store degrades to the same "deny for now, retry next request" behavior the code already has for hard DB errors, instead of hanging the request.

dev/docs/adr/027-license-gated-sso.md
```diff
@@ -0,0 +1,192 @@
...
+- The Helm chart (`charts/`) must plumb `LANGWATCH_LICENSE_KEY` through values → env; add alongside the env-create.mjs change.
+- "As if the SSO env vars were unset" is *almost* exact: a denied genericOAuth install can still `POST /sign-in/oauth2` and be redirected to the IdP — the callback-path block is what fails the round-trip closed (login never completes). The initiation redirect happening is cosmetic, not a bypass; documented so the implementer blocks the callback, not just the button.
+- e2e auth harnesses (`e2e/auth-regression/*`, `*sso-smoketest*`) that assume ungated SSO must seed a license or the env key. Audit set: `rg -l 'NEXTAUTH_PROVIDER' e2e/ langwatch/src` + the auth harness files.
+- `specs/licensing/sso-license-gating.feature` (ships in the same PR) must be rewritten against this design — the truth-table scenarios are obsolete; add scenarios for the v5/v6 anchors (licensed `/sign-up/email` 403; denied-mode auto-join off; expired-but-signed license allows; denied-mode reset recovers an OAuth-born user).
+
+## Open questions
+
```
Aryansharma28, line 167: [P3] Non-blocker — confirming and tracking, not re-raising.

Confirmed via `grep -rn LANGWATCH_LICENSE charts/` that Helm plumbing for `LANGWATCH_LICENSE_KEY` (and the pre-existing `LANGWATCH_LICENSE_PUBLIC_KEY`) genuinely isn't there yet. You've already called this out here as a deliberate, tracked follow-up — agreed this is fine to land without it, just flagging so it doesn't quietly fall off the follow-up list once this merges.

langwatch/src/server/better-auth/hooks.ts
```diff
@@ -110,17 +128,33 @@ export const afterUserCreate = async ({
   const domain = extractEmailDomain(user.email);
   if (!domain) return;
 
+  // ADR-027 site #4: domain auto-join is federation and rides the platform
+  // gate. When it denies (unlicensed deployment), skip the join — but log it,
+  // because on an email-mode install the gate-resolution warning is suppressed
+  // (sso-gate.ts), so a staff-set ssoDomain silently losing auto-join would
+  // otherwise leave zero trace for an operator debugging "why wasn't this user
+  // added to the org".
+  const ssoAllowed = await platformSSOAllowed();
```
Aryansharma28, line 143: [P4] Nit — log-level inconsistency.

**Why:** This logs at `logger.info` when the auto-join is skipped due to gate denial, but the gate's own denial-resolution log in `sso-gate.ts` uses `logger.warn` for the same root cause ("no genuine license"). An operator grepping warn-level logs to debug "why didn't this SSO signup auto-join the org" would miss this one. Small consistency win, not urgent.

langwatch/src/server/better-auth/__tests__/resetRecoversOauthUser.test.ts
```diff
@@ -0,0 +1,123 @@
+/**
```
Aryansharma28, line 1: [P5] Praise.

This is exactly the right way to protect a design decision that rests on an upstream library's undocumented behavior. Pinning `resetPassword`'s "creates a credential account for an OAuth-born user" premise against the *real* better-auth memory adapter (no mocks, in-process, ~0.5s) means a future better-auth upgrade that changes this behavior goes red here instead of silently re-stranding every existing user on a denied self-hosted install. Worth using as the template for other ADR-driven features in this repo that lean on an undocumented upstream contract.

Aryansharma28 (commented):
## goated-review — deep pass on ADR-027 license-gated SSO

**Read the room first:** this PR arrives extraordinarily well-prepared — a 7-revision ADR with explicit red-team rounds (v5 credential-mint bypass, v5 org-takeover, v5 wrong-path-list, v6 expiry-cliff, v7 recovery-path lockouts), 135+ unit/integration tests, and a real (non-mocked) pin of an upstream better-auth behavior (`resetRecoversOauthUser.test.ts`). CodeRabbit is disabled on this base branch, and the only "reviews" on the PR are automation low-risk-approvals and the author's own self-review comments — so there is no existing human/bot finding to dedupe against; everything below is new.

Given the security-critical nature (SSO entitlement gate), I adversarially verified the core claims rather than trusting the ADR's own self-assessment: I confirmed `verifySignature()` does real RSA-SHA256 verification (`ee/licensing/validation.ts`) against an embedded production public key with no matching private key checked into the repo (no forgeable licenses), confirmed the gate is fail-closed on DB error (deny-without-memoizing, self-healing per Decision 6), confirmed `IS_SAAS` is a real coerced boolean (not a truthy-string trap) and is checked before any DB touch, and walked every hook (`beforeAccountCreate`, `afterUserCreate`, the better-auth `before` hook) to confirm the ALLOW/DENY branches compose correctly with no gap for the scenarios the ADR itself enumerates. I could not find an actively-exploitable bypass or a forgeable-signature path today. Two real gaps did survive adversarial verification, both bucketed below.

### Verdict: COMMENT — 1 P1, 1 P2 (both BLOCKERS per your process), rest non-blocking polish + praise.

**P0: 0 | P1: 1 | P2: 1 | P3: 2 | P4: 1 | P5: 2**

---

### BLOCKERS (P0–P2) — should be resolved before/shortly after merge

**[P1] `ssoPathGate.ts` — the DENY-state gate is a hand-maintained blocklist, not a default-deny allowlist.** `isGatedSsoPath()` (and the sibling suffix lists) only blocks the *specific* better-auth/genericOAuth endpoints the author enumerated for the current `better-auth@1.6.12`. Any endpoint shaped like SSO/federation that isn't on this static list — a future better-auth plugin (passkeys, magic-link, a new OIDC-provider mode), a new genericOAuth initiation path, or simply a version bump that renames/adds a route — silently falls through **unblocked** when the gate denies. That re-opens exactly the four-env-var free-SSO bypass this PR exists to close, with zero CI signal (the test suite only proves today's enumerated list is correct, not that tomorrow's route table is covered). For a control whose entire purpose is "is this deployment allowed to federate login," fail-open-by-default-for-anything-unlisted is the wrong failure direction for a security boundary — it should fail closed by default and allowlist the known-safe passthrough paths (`get-session`, `sign-out`, `list-sessions`, the email/reset routes already handled explicitly) instead of blocklisting the known-bad ones. Suggested fix: invert to an allowlist model, or at minimum add a canary test that enumerates `auth.api`'s actual registered route table (or the `genericOAuth`/`socialProviders` plugin's route list) and fails loud if a new non-email-mode endpoint appears that `isGatedSsoPath`/`isEmailAuthPath`/`isPasswordResetPath` doesn't account for — so a `better-auth` upgrade can't silently regress this the way v5's `/oauth2/authorize` phantom-path bug almost did.

**[P2] `better-auth/index.ts` — the gate is awaited on *every* better-auth request for any SSO-capable deployment, including plain session reads, with no timeout.** `hooks.before` calls `await platformSSOAllowed()` unconditionally once `NEXTAUTH_PROVIDER !== "email"` and the request isn't a credential-mutation path — that includes `/get-session`, `/sign-out`, `/list-sessions`. During the (memoization) warm-up window, if the DB (Organization table scan) is slow — not erroring, just slow — every single better-auth request, including session validation for already-authenticated users on a *licensed* install, blocks on that DB round trip with no timeout. This directly cuts against the ADR's own invariant ("existing session cookies... keep working through upgrades" / "Sessions survive the gate") during that window — it's not that sessions break, but they can hang, and there's no bound on how long. The existing `ssoGate.hook.test.ts` test for this ("never intercepts session reads... in either gate state") mocks `platformSSOAllowed` to resolve synchronously, so it can't catch this — it proves correctness, not availability. Suggested fix: either short-circuit known gate-independent paths (`get-session`, `sign-out`, `list-sessions`) before ever calling `platformSSOAllowed()`, or wrap the underlying DB scan in `anyOrgHasSignedLicense()` with a timeout so a slow DB degrades to "deny for now, retry next request" (which the code already does for hard errors) instead of hanging the whole auth surface.

---

### NON-BLOCKERS (P3–P5) — worth doing, not required to merge

**[P3]** Helm chart plumbing for `LANGWATCH_LICENSE_KEY` (and the pre-existing `LANGWATCH_LICENSE_PUBLIC_KEY`) is still missing — the author already calls this out explicitly in the PR description as a tracked, non-blocking follow-up alongside the existing gap. Confirmed via `grep` there's no `LANGWATCH_LICENSE` reference anywhere under `charts/`. Agreeing this is fine to follow up, flagging so it doesn't get lost.

**[P3]** `isGatedSsoPath()` recomputes `normalizedRequestPathname(url)` internally even though the `before` hook already computed `pathname` once and passes the raw `url` back in — harmless (cheap), but slightly muddies the "pathname is normalized once, reused everywhere" contract the module docblock claims. Minor DRY nit, would consolidate to taking the already-normalized pathname as a parameter.

**[P4]** Log-level inconsistency: the gate's own denial is `logger.warn` (`sso-gate.ts`), but the `afterUserCreate` auto-join skip on the same denial is `logger.info` (`hooks.ts:137`) — both are signals of the same root cause ("deployment has no genuine license") and an operator grepping for `warn` to debug a customer's "why isn't SSO/auto-join working" ticket would miss the second one. Not urgent, just an easy consistency win.

**[P5]** Genuine praise: `resetRecoversOauthUser.test.ts` is exactly the right way to protect a design decision that rests on an upstream library's undocumented behavior — it pins `resetPassword` creating a credential account for an OAuth-born user against the *real* better-auth memory adapter (no mocks), so a future better-auth upgrade that changes this premise goes red here instead of silently re-stranding every existing user on a denied install. This is a great pattern other ADR-driven features in this repo should copy.

**[P5]** The trailing-slash / query-string path-normalization fix (`ssoPathGate.ts`'s `normalizedRequestPathname`) closes a real v5 bug (raw `endsWith`/suffix matching could be walked past by a one-character variant) and is backed by explicit regression tests for both the trailing-slash and query-string cases in both gate states. Good defense-in-depth discipline.

---

- Whether any currently-deployed self-hosted enterprise customer actually has an in-DB activated license vs. relying solely on the four env vars (the ADR's own "Open questions" flags this as an operational risk to confirm with CS/sales before shipping — I have no visibility into that).
- Real-world latency of the `Organization` table scan on a large multi-tenant self-hosted install (relevant to the P2 finding above) — I don't have production workload data for self-hosted deployments.
- Whether `better-auth@1.6.12`'s actual route table has any endpoint outside the enumerated list today — I verified against the vendored source referenced in the ADR/tests, not by enumerating `auth.api` myself at runtime.

langwatch/ee/sso/ssoPathGate.ts
rogeriochaves: Addressed in a3b144a33f, with the canary rather than the inversion.

Full default-deny in the DENY state would have to allowlist the entire email-mode surface, which is much larger and more fragile than the federation set, and getting it wrong breaks the coerced mode this exists to fall back to. So the blocklist stays, and the staleness risk you named is what gets a test: `ssoRouteTableCanary.test.ts` builds a real better-auth instance with `socialProviders` + `genericOAuth`, reads the mounted route table off `auth.api`, and requires every route to carry a reviewed `federating` / `local` classification. An upgrade that adds, renames or removes one fails by name, and the gate verdict is asserted against the classification for every route in both the bare and trailing-slash forms.

Verified by deleting one entry: three assertions go red naming `/sign-in/oauth2`.

It also found a live false positive. `isGatedSsoPath` matched the callback substring against the un-normalized pathname, so `/delete-user/callback/` was refused on an unlicensed deployment. It now matches the normalized one, so the segment only counts when a provider segment follows it.

langwatch/src/server/better-auth/index.ts
rogeriochaves: Addressed in a3b144a33f.

The hook now returns before touching the gate for any path the gate cannot affect. The condition is `isGateDependentPath(url)`, derived from the branches it guards (the reset pair, the email-auth pair, the SSO-initiation/callback set) rather than a hand-listed set of safe paths, so the two cannot drift apart. `/get-session`, `/sign-out` and `/list-sessions` never await it at all.

Your point about the existing test is why the new one mocks the gate as a promise that never settles: `platformSSOAllowed` returning `new Promise(() => {})` would hang those requests if the hook still awaited it, which a resolved mock can never show. It also asserts the gate was not called.

The scan is bounded too, since a slow store still stalls the requests that genuinely need the answer: 5s, landing on the same evict-and-retry path a hard DB error already takes, so a store that recovers still needs no restart. Tested with fake timers.

dev/docs/adr/027-license-gated-sso.md
rogeriochaves: Done in a3b144a33f rather than tracked. `app.license.key` and `app.license.publicKey` take an inline value or a `secretKeyRef`, and sit in `sharedEnv` rather than the app Deployment: plan resolution runs in the workers too, and a worker reading a different entitlement than the app would enforce different limits on the same organization. Rendering verified in all three states, and the self-hosting licensing doc now shows the Helm form alongside the env var.

langwatch/src/server/better-auth/hooks.ts
rogeriochaves: Fixed in a3b144a33f, now `logger.warn` with a note in the code saying it matches the gate role.

langwatch-agent (commented):
Re-reviewed the security-sensitive diff after the latest follow-up. No additional actionable finding: the route-table canary closes the future-route coverage concern, gate-independent session paths do not wait on the license scan, the scan has an eviction-on-failure timeout, and provider coercion is consistently applied to BetterAuth, public env, direct registration, and logout. I also confirmed the chart values are routed through `sharedEnv`; I could not render Helm locally because the `helm` binary is unavailable. Local `typecheck`, `typecheck:tests`, and the focused SSO gate/path/hook/canary suite pass (47 tests). Remaining CI jobs are still running.

rogeriochaves:
@coderabbitai review

specs/licensing/sso-license-gating.feature
```diff
@@ -0,0 +1,201 @@
...
+  # ============================================================================
+
+  @unit
+  Scenario: Denied SSO is explained in the server logs
+    Given a self-hosted deployment configured with an enterprise IdP
+    And no genuine license anywhere
+    When the server evaluates the gate and refuses an SSO attempt
+    Then the logs say SSO is configured but no genuine license was found
+    And name how to enable it
+    And each refused SSO request is logged with its path and reason
```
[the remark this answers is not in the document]
rogeriochaves: Checked this against the tool itself and it disagrees: `pnpm check:feature-parity` reports

```
▸ specs/licensing/sso-license-gating.feature
  20/20 scenarios bound
  ✓ all bound
```

and the whole check exits 0. Spot-checking two of the lines named above, "An expired but genuine license still keeps SSO working" (line 37) is bound at `ee/sso/__tests__/sso-gate.test.ts:120` and "SaaS is unaffected by license gating" (line 178) at `ee/sso/__tests__/sso-gate.test.ts:90`. The annotations live in `ee/sso/__tests__/`, which is where those files moved to in this PR, so a search against their old `src/server/sso/` paths would come back empty.

Worth noting the failure mode is the opposite of the one described: an *untagged* file reports `0/0 scenarios bound / ✓ all bound` and reads green while binding nothing. This file is tagged throughout, so it is measured.

langwatch/ee/sso/providers.ts
```diff
@@ -0,0 +1,271 @@
...
+    socialProviders.gitlab = {
+      clientId: e.GITLAB_CLIENT_ID,
+      clientSecret: e.GITLAB_CLIENT_SECRET,
+      mapProfileToUser: (profile) => ({
+        name: fallbackName(profile as Record<string, any>),
+        email: (profile as { email?: string }).email,
+        image: (profile as { avatar_url?: string }).avatar_url,
+      }),
+    };
+  }
```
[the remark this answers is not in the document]
rogeriochaves: Not taking this one. Writing an opaque provider subject id into the `email` column produces a user record that is corrupt in a way that is worse than the failure it avoids: `email` is what org domain matching (`extractEmailDomain`), password reset, invites and every notification path key off, and `sub` is a numeric string with no domain. A user created as `email: "109384756..."` can never receive a reset link and never matches an `ssoDomain`, and the record looks legitimate, so nobody notices until they try to use it.

If a provider genuinely returns no email, refusing the sign-in is the correct outcome. It is loud, it is recoverable, and it does not leave a permanent bad row behind.

On the premise: Google returns `email` in the ID token whenever the `email` scope is granted, which better-auth requests by default, and GitLab's `/api/v4/user` returns the primary email to the authenticated user under `read_user`. Neither has GitHub's private-email case, which is exactly why GitHub is the one provider with a documented secondary `user/emails` fetch and the other two are not.

langwatch/ee/sso/providers.ts
```diff
@@ -0,0 +1,271 @@
...
+            }
+          ).email ??
+          (profile as { mail?: string }).mail ??
+          (profile as { userPrincipalName?: string }).userPrincipalName,
+      }),
+    };
+  }
+
+  return socialProviders;
+};
```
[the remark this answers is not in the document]
rogeriochaves: Leaving this as it is. The function is not new in this PR: it was lifted verbatim out of `src/server/better-auth/index.ts` into `ee/sso/providers.ts` as part of moving the enterprise surface under `ee/`, so a rewrite here would turn a pure move into a behavior-bearing change on the sign-in path and make the move itself much harder to review.

It also is not as uniform as it looks from the outside: azure-ad carries a `tenantId` and a three-way `email ?? mail ?? userPrincipalName` fallback, and the generic-OAuth builder alongside it has its own issuer parsing. A table would need enough per-provider escape hatches to give most of the saving back.

Biome is clean on the file (no `noExcessiveCognitiveComplexity`, no `noExcessiveLinesPerFunction`), so there is no lint debt being carried forward either. Happy to revisit as its own PR where the diff is only the refactor.

langwatch/src/server/better-auth/index.ts
```diff
@@ -644,35 +421,71 @@ export const auth = betterAuth({
...
+        logger.warn(
+          { path: requestPathname(url), reason: "no_license" },
+          "Blocked SSO request: deployment has no genuine license",
+        );
+        throw APIError.from("FORBIDDEN", {
+          code: "SSO_LICENSE_REQUIRED",
+          message:
+            "SSO is not available on this deployment — sign in with your email and password instead.",
+        });
       }
```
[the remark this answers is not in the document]
rogeriochaves: Keeping the URL-taking signature deliberately. `isGatedSsoPath` owning its own normalization is what makes the `/delete-user/callback/` fix in this PR hold: the bug was that the callback check ran against a *non-normalized* pathname, so a local account route ending in the word "callback" was refused as SSO. Pushing normalization out to callers reopens exactly that class, because the predicate can then be handed a raw pathname, a trailing-slash variant or a query string depending on the call site, and nothing at the type level says which it wanted.

The two families are also not the same kind of predicate. `isEmailAuthPath`/`isPasswordResetPath` are private suffix matchers over an already-canonical string. `isGatedSsoPath`/`isGateDependentPath` are the public entry points the hook calls with whatever better-auth handed it, so normalizing at that boundary is the point rather than a duplicated step.

The repeated parse is one `new URL` on gate-dependent paths only, which the short-circuit in `index.ts` already keeps off the rest of the route table.

docs/llms-full.txt
```diff
@@ -60445,15 +60586,15 @@ This keeps the model predictable: your base plan covers normal usage at its defa
...
   </Accordion>
 
   <Accordion title="Can I self-host or run a hybrid setup?">
-    Yes. LangWatch is available both [self-hosted](/self-hosting/overview) and [hybrid](/hybrid-setup/overview). With the hybrid option you get the convenience of running on our Cloud while your data stays on your side. See the self-hosting docs or reach out for more information. You can run the self-hosted version for free. If you require enterprise features such as SSO, RBAC, SCIM, or a support SLA, [reach out to us](mailto:enterprise@langwatch.ai).
+    Yes. LangWatch is available both [self-hosted](/self-hosting/overview) and [hybrid](/hybrid-setup/overview). With the hybrid option you get the convenience of running on our Cloud while your data stays on your side. See the self-hosting docs or reach out for more information. You can run the self-hosted version for free, with unlimited users, projects, and events on your own hardware. If you require enterprise features such as SSO, RBAC, SCIM, or a support SLA, [reach out to us](mailto:enterprise@langwatch.ai). See [Editions and licensing](/self-hosting/licensing) for what a license unlocks.
```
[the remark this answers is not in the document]
rogeriochaves: Fixed at the source. `docs/llms-full.txt` is generated (`cd docs && node llms.txt.cjs`, and docs-CI fails if it is stale), so the edit went into `docs/pricing.mdx` and the file was regenerated.

You were right that three surfaces disagreed. The retention section and its FAQ both said extending storage needs a license when self-hosting, while the licensing page says self-hosted retention is configurable without one and volume is not metered at all. The Cloud rule and the self-hosted rule were being stated as one rule.

They are now separated: the windows and the Enterprise upgrade are Cloud, and self-hosted says the data is in your own database, retention is yours to configure with or without a license, and what a license adds there is multi-class retention policies per data category. `self-host with a license` no longer appears in either file.

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai review

rogeriochaves:
@coderabbitai Two of these landed, twelve are against code that is already in this branch. Checked each against `HEAD` before replying.

**Fixed (a237c503):**

- `ssoPathGate.ts` "pathname-PREFIX match via `includes`" was wrong, `includes` is a substring test and has to be, because the segment sits mid-path in `/oauth2/callback/okta` as well as at the front in `/callback/auth0`. Reworded, and it now says why normalizing first is what makes a bare substring safe.
- `open-core-licensing.mdx` still said "the directory is what the license follows" one paragraph below the most-specific-first explanation. Both points now say most-specific-first, so the MIT SDK row cannot read as contradicting the everything-else row.

**Already in the branch, please re-read `HEAD`:**

| Finding | Current state |
|---|---|
| ADR-027 status value | Line 4 is already `**Status:** Accepted`, no parenthetical |
| SDK license contradiction | The table already carries the MIT row for `typescript-sdk`/`python-sdk`/`mcp-server`, and the paragraph under it already reads "the SDK directories are MIT even though they sit outside `langwatch/ee/`" |
| `sso-gate.ts` soft-delete docblock | The "Skips soft-deleted orgs" phrase is gone; the accurate note lives in `sso-license.repository.ts:27` |
| `useLicenseActions` unresolved env | Already `IS_SAAS === true`, with a comment stating that only a confirmed `true` means Cloud |
| `CAPABILITIES` literal types | Already `] as const satisfies readonly Capability[];` at line 58 |
| Rename `licensed` to `isLicensed` | Already `isLicensed`, at the prop, the type and both call sites |
| Trailing `Separator` on Cloud | `authentication.tsx` renders no `Separator`; the one that exists is inside `EnterpriseCapabilitiesSection`, above its early `return null`, so Cloud gets neither |
| `hooks.ts:310` silent denial | Already emits `logger.warn` with the provider id, with a comment tying it to the `afterUserCreate` site |
| `sendResetPassword` stale comment | Already rewritten: "Reset is deliberately reachable on a deployment the SSO license gate denies, even with an IdP configured (ADR-027) ... It closes again once the gate allows" |
| No `auth0` branch coverage | `user.changePassword.unit.test.ts:89` already does `resolveAuthProviderMock.mockResolvedValue("auth0")` |
| `rejects.toBeDefined()` | Already `rejects.toMatchObject({ body: { code: "INVALID_TOKEN" } })`, with a comment explaining why it is pinned to the code |
| README leading-slash links | Already relative: `[langwatch/ee/](langwatch/ee/)`, `[LICENSE.md](LICENSE.md)`, `[NOTICE](NOTICE)` |

**Disputed, the rule does not fire:** the `require-bdd-describe-context` claim on `ssoPathGate.test.ts` and `sso-license.repository.test.ts`. Run with the version CI pins (`ast-grep-cli==0.42.3`) and the repo's own config:

```
$ ast-grep scan -c .ast-grep/sgconfig.yml \
    langwatch/ee/sso/__tests__/ssoPathGate.test.ts \
    langwatch/ee/sso/__tests__/sso-license.repository.test.ts
$ echo $?
0
```

No output, exit 0. Neither file appears anywhere in a scan of all of `langwatch/ee/`, and the rule's own fixtures pass (`ast-grep test -c sgconfig.yml -t rule-tests`, 24 passed). The `ast-grep` job on this PR is green for the same reason. The rule exempts a top-level `describe` naming the unit under test, which is what both files use.

rogeriochaves:
@coderabbitai Four of the five landed in `e691af6f`, and the fifth is answered by the code rather than by the docs.

**SPDX comment syntax** (`open-core-licensing`): correct, and the table was the wrong shape for the claim. The header column now names the identifier and the prose says the delimiter follows the language, `//` in TypeScript and Go, `#` in Python and YAML.

**Licensed-provider fallback rule**: both statements are true, and this branch is where they stopped contradicting each other. A license opens SSO, it does not configure it. If `NEXTAUTH_PROVIDER` names something LangWatch cannot mount, a licensed deployment stays in email mode with the email routes open, which is the whole point: a typo in a provider name must not be able to lock a paying customer out. The licensing page now says that explicitly next to the claim that a license closes the email routes.

Proven live on a QA stack with a real signed license, the same license in both runs:

| `NEXTAUTH_PROVIDER` | Credentials present | Resolved | `POST /sign-in/email` |
|---|---|---|---|
| `okta` | no | `email` | `401 INVALID_EMAIL_OR_PASSWORD`, the route works |
| `okta` | yes | `okta` | `400 EMAIL_PASSWORD_DISABLED` |

**Legacy migration wording**: agreed, the recovery procedure only makes sense for a deployment that predates the gate, because that is the only way a team of identity-provider accounts can exist on an unlicensed install. The warning says which deployments it is for now.

**`LANGWATCH_LICENSE_KEY` prerequisite**: agreed. It read as if the key alone restores SSO. It now says it works as long as the identity-provider variables that were already working are still in place.

**`fallbackName.test.ts` nested `when`**: the block was inconsistent with its siblings, which all use `when`. Renamed to `when each profile drops the most specific field it has` rather than nested inside another `describe`, since one level says the same thing.

rogeriochaves (dismissed):

rogeriochaves (approved):

PR #4832, opened 2026-06-15
feat(billing): storage-billing schema — additive StorageUsageHourly + StorageBillingCheckpoint [ADR-027 PR1/7]

sergioestebance:
**Stack:** PR1 of 7 — implements [ADR-027](https://github.com/langwatch/langwatch/pull/4752) storage-GiB-hours billing. Stacked on the ADR PR (#4752) as base.

## What

Phase 1 (persistence). **Purely additive — two new tables on top of the current implementation.** The billable-events checkpoint (`BillingMeterCheckpoint`) and its unique index are left completely untouched.

- **`StorageUsageHourly`** (new) — durable per-sealed-UTC-hour measurement of stored bytes aged > 30 days. PK `(organizationId, sealedHour)`, index on `reportedAt`. The dispatcher (PR5) writes one row/hour; the reporter (PR4) drains `reportedAt IS NULL` and stamps on Stripe success. Decouples *measurement* from *Stripe availability* — the reporter can stall and resume without re-measuring or re-billing — and is the per-org audit trail an invoice line traces back to.
- **`StorageBillingCheckpoint`** (new) — dedicated two-phase reporting checkpoint for the `STORAGE_GB` meter, a sibling of `BillingMeterCheckpoint`.
- **`StorageBillingCheckpointService`** (new) — its own service over the new table, mirroring the proven two-phase + circuit-breaker protocol.
- Both new org-bearing tables registered in the org-tenancy exempt list (`dbOrganizationIdProtection`).

No ClickHouse migration — `_size_bytes` already exists (migration `00032`).

## Design note — separate table, not a discriminator

The ADR's first draft proposed adding a `meterName` discriminator to `BillingMeterCheckpoint` and swapping its unique index. **This PR deliberately does not do that.** A separate `StorageBillingCheckpoint` table + service keeps the migration purely additive (no `DROP INDEX`, no `ALTER` on a live billing table) and the billable-events path 100% unchanged — each meter owns its own persistence. ADR-027 Schema section updated to match.

## Why dormant is safe

Nothing in this PR dispatches a storage report. The whole pipeline lands in later PRs; even after wire-up, no org bills until rollout attaches the `STORAGE_GB` SubscriptionItem.

## Verification

- Migration generated via `prisma@5.7.1 migrate diff`, then **applied cleanly over the full 182-migration history on a throwaway Postgres**. Confirmed both new tables + indexes exist and `BillingMeterCheckpoint`'s original unique index is **intact** (never dropped).
- `pnpm typecheck` clean.
- Unit tests green (75): new `storageBillingCheckpoint.service.unit.test.ts` (3 — proves the service touches only the storage table, never `BillingMeterCheckpoint`); billable-events `reportUsageForMonth.command.unit.test.ts` passes with its **original** assertions (path untouched); org-tenancy partition tests (58 — confirm both new tables classified org-bearing).

## Runbook

None — additive `CREATE TABLE`s only, no locks on existing tables.

### DIFF
--- a/langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
+++ b/langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
@@ -1,0 +1,46 @@
+-- ADR-027 storage billing, schema phase. Purely additive: two new tables on top
+-- of the current implementation. The billable-events checkpoint
+-- (BillingMeterCheckpoint) and its unique index are left untouched — storage
+-- billing owns its own persistence rather than discriminating the shared one.
+--
+-- StorageUsageHourly       — durable per-sealed-UTC-hour stored-bytes
+--                            measurement + reporter cursor (reportedAt).
+-- StorageBillingCheckpoint — dedicated two-phase reporting checkpoint for the
+--                            STORAGE_GB meter (sibling of BillingMeterCheckpoint).
+--
+-- No ClickHouse migration — _size_bytes already exists (migration 00032).
+
+-- CreateTable
+CREATE TABLE "StorageUsageHourly" (
+    "organizationId" TEXT NOT NULL,
+    "sealedHour" TIMESTAMP(3) NOT NULL,
+    "megabytes" INTEGER NOT NULL,
+    "reportedAt" TIMESTAMP(3),
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "StorageUsageHourly_pkey" PRIMARY KEY ("organizationId","sealedHour")
+);
+
+-- CreateTable
+CREATE TABLE "StorageBillingCheckpoint" (
+    "id" TEXT NOT NULL,
+    "organizationId" TEXT NOT NULL,
+    "billingMonth" TEXT NOT NULL,
+    "lastReportedTotal" INTEGER NOT NULL DEFAULT 0,
+    "pendingReportedTotal" INTEGER,
+    "consecutiveFailures" INTEGER NOT NULL DEFAULT 0,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "StorageBillingCheckpoint_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "StorageUsageHourly_reportedAt_idx" ON "StorageUsageHourly"("reportedAt");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "StorageBillingCheckpoint_organizationId_billingMonth_key" ON "StorageBillingCheckpoint"("organizationId", "billingMonth");
+
+-- Down (reversible — purely additive). Commented out to avoid accidental data
+-- loss; to roll back, uncomment and run manually:
+-- DROP TABLE "StorageBillingCheckpoint";
+-- DROP TABLE "StorageUsageHourly";
--- a/langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
+++ b/langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
@@ -26,8 +26,8 @@ CREATE TABLE "StorageBillingCheckpoint" (
     "id" TEXT NOT NULL,
     "organizationId" TEXT NOT NULL,
     "billingMonth" TEXT NOT NULL,
-    "lastReportedTotal" INTEGER NOT NULL DEFAULT 0,
-    "pendingReportedTotal" INTEGER,
+    "lastReportedTotal" BIGINT NOT NULL DEFAULT 0,
+    "pendingReportedTotal" BIGINT,
     "consecutiveFailures" INTEGER NOT NULL DEFAULT 0,
     "updatedAt" TIMESTAMP(3) NOT NULL,
 
--- a/langwatch/prisma/schema.prisma
+++ b/langwatch/prisma/schema.prisma
@@ -1793,6 +1793,43 @@ model BillingMeterCheckpoint {
   @@unique([organizationId, billingMonth])
 }
 
+// ADR-027 storage billing — own its persistence, additively, alongside (not
+// inside) the billable-events checkpoint above. Keeping a separate table means
+// the storage meter never touches BillingMeterCheckpoint's live unique index.
+
+// Durable per-hour stored-bytes measurement. The dispatcher writes one row per
+// sealed UTC hour (megabytes = stored bytes aged > 30d, measured once); the
+// reporter drains reportedAt IS NULL across orgs and stamps it on Stripe
+// success. This decouples measurement from Stripe availability — the reporter
+// can stall and resume without re-measuring or re-billing — and is the per-org
+// audit trail an invoice line traces back to.
+model StorageUsageHourly {
+  organizationId String
+  sealedHour     DateTime // UTC, truncated to hour
+  megabytes      Int // ceil(bytes / 1_048_576), measured at sample time
+  reportedAt     DateTime?
+  createdAt      DateTime  @default(now())
+
+  @@id([organizationId, sealedHour])
+  @@index([reportedAt])
+}
+
+// Two-phase reporting checkpoint dedicated to the STORAGE_GB meter — a parallel
+// of BillingMeterCheckpoint owned by StorageBillingCheckpointService. Separate
+// table (not a meterName discriminator on the shared one) so storage billing is
+// purely additive and the billable-events checkpoint is never altered.
+model StorageBillingCheckpoint {
+  id                   String   @id @default(cuid())
+  organizationId       String
+  billingMonth         String // "2026-02"
+  lastReportedTotal    Int      @default(0)
+  pendingReportedTotal Int? // non-null = in-flight Stripe call
+  consecutiveFailures  Int      @default(0)
+  updatedAt            DateTime @updatedAt
+
+  @@unique([organizationId, billingMonth])
+}
+
 // ============================================================================
 // AI Gateway
 //
--- a/langwatch/prisma/schema.prisma
+++ b/langwatch/prisma/schema.prisma
@@ -1822,8 +1822,11 @@ model StorageBillingCheckpoint {
   id                   String   @id @default(cuid())
   organizationId       String
   billingMonth         String // "2026-02"
-  lastReportedTotal    Int      @default(0)
-  pendingReportedTotal Int? // non-null = in-flight Stripe call
+  // Cumulative MiB-hours reported for the period (ADR-027: Σ megabytes over the
+  // billing month). BigInt, not Int: Int32 caps at ~2.8 TiB held for a full
+  // month before overflow, which a large enterprise org can exceed.
+  lastReportedTotal    BigInt   @default(0)
+  pendingReportedTotal BigInt? // non-null = in-flight Stripe call
   consecutiveFailures  Int      @default(0)
   updatedAt            DateTime @updatedAt
 
--- a/langwatch/prisma/schema.prisma
+++ b/langwatch/prisma/schema.prisma
@@ -1798,7 +1798,8 @@ model BillingMeterCheckpoint {
 // the storage meter never touches BillingMeterCheckpoint's live unique index.
 
 // Durable per-hour stored-bytes measurement. The dispatcher writes one row per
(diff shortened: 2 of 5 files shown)


langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @drewdrewthis

**Scariest prod risk:** none found for this PR in isolation (additive `CREATE TABLE`s, no ALTER/DROP, dormant until a later PR wires it up). Main latent risk is 32-bit `INTEGER` sizing on the running totals below.

- `[Must Check]` `langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql` — `megabytes`, `lastReportedTotal`, `pendingReportedTotal` are all `INTEGER` (max ~2.1B). Confirm a large org's cumulative monthly total in these units cannot overflow 32 bits; if it can, this needs BIGINT before any org is billed.
- `[Ask Author]` `langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts` — `clearPendingAndIncrementFailures` uses `.update()` (throws if the row is absent) while the siblings use `.upsert()`; confirm the checkpoint row is guaranteed to exist at every call site or make it an upsert.
- `[Probably Fine]` `langwatch/src/utils/dbOrganizationIdProtection.ts` — both new org-bearing tables added to the exempt list; confirm classification matches the sibling `BillingMeterCheckpoint` treatment.

- What unit is `lastReportedTotal` (MB, GB, or GiB-hours), and does that unit match the Stripe `STORAGE_GB` meter the reporter in PR4/PR5 will submit?
- Is the `reportedAt IS NULL` drain query in a later PR expected to lean on the `reportedAt` index, and does that index serve the NULL-first scan well at scale?

langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
```diff
@@ -0,0 +1,46 @@
...
+
+    CONSTRAINT "StorageUsageHourly_pkey" PRIMARY KEY ("organizationId","sealedHour")
+);
+
+-- CreateTable
+CREATE TABLE "StorageBillingCheckpoint" (
+    "id" TEXT NOT NULL,
+    "organizationId" TEXT NOT NULL,
+    "billingMonth" TEXT NOT NULL,
+    "lastReportedTotal" INTEGER NOT NULL DEFAULT 0,
```
langwatch-agent: **P2** — These checkpoint totals are 32-bit `INTEGER`s, but ADR-027 defines the storage meter total as summed MiB-hours for the billing period. That caps a month at ~2.1B MiB-hours, which is only about 2.9 TiB-months (`2_147_483_647 / 720 / 1024`). A large enterprise org holding a few TiB of beyond-35-day logical trace data for a full month can exceed this, causing Prisma/Postgres writes to fail or forcing later lossy workarounds in the billing path. Since this is the additive schema phase and the table is still empty, please make `lastReportedTotal` and `pendingReportedTotal` `BIGINT` / Prisma `BigInt` now, and carry that type through the storage checkpoint service/tests before any storage invoices rely on it.

langwatch-agent (dismissed):
1 P2 finding: storage billing checkpoint totals use 32-bit integers, which is too small for cumulative MiB-hour billing. See inline comment.

langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql
sergioestebance: Fixed in fd9654b16. Both \`lastReportedTotal\` and \`pendingReportedTotal\` are now \`BIGINT\` / Prisma \`BigInt\`, carried through the service interface (\`bigint\`) and unit tests. Done while the table is empty, so no data migration. Left \`consecutiveFailures\` and per-hour \`StorageUsageHourly.megabytes\` as \`Int\` — a single hour would need >2 PiB in one org to overflow, unlike the month-cumulative totals.

langwatch/prisma/schema.prisma
```diff
@@ -1793,6 +1793,46 @@ model BillingMeterCheckpoint {
   @@unique([organizationId, billingMonth])
 }
 
+// ADR-027 storage billing — own its persistence, additively, alongside (not
+// inside) the billable-events checkpoint above. Keeping a separate table means
+// the storage meter never touches BillingMeterCheckpoint's live unique index.
+
+// Durable per-hour stored-bytes measurement. The dispatcher writes one row per
+// sealed UTC hour (megabytes = stored bytes aged > 30d, measured once); the
```
langwatch-agent: **P2** — This schema comment still says the stored bytes are aged `> 30d`, but the current ADR/spec on this base defines `BILLABLE_AFTER_DAYS = 35` and says paid plans include 35 days free. This table is the persistence contract the later dispatcher/reporter PRs will copy from; leaving the wrong cutoff here makes it easy to implement the storage measurement as “older than 30 days”, which would bill default 35-day paid retention orgs for days 31–35 and break the ADR invariant that default-keep paid orgs bill €0 by construction. Minimal fix: change this to `aged > 35d` or, better, `aged beyond BILLABLE_AFTER_DAYS` so the schema prose stays aligned with the named constant in the measurement implementation.

langwatch-agent (commented):
1 P2 finding — the BigInt overflow issue is fixed; remaining concern is an ADR/spec mismatch in the schema contract. See inline comment.

langwatch/prisma/schema.prisma
sergioestebance: Fixed in 1f01f5d14. Schema comment now reads "aged beyond BILLABLE_AFTER_DAYS (35d — the paid included window)" instead of 30d, tying the persistence-contract prose to the named constant the dispatcher will use, so the measurement isn't implemented as >30d and default-35d paid orgs stay €0 by construction.

langwatch-agent:

**Reviewer:** @drewdrewthis · **Mode:** `Deep Review` (billing + DB migration + tenant-isolation util)
**Context:** ADR-027 PR 1 of 7, stacked on langwatch/langwatch#4752 as base. Persistence phase only, described as purely additive.

**Scariest production risk:** a billing-schema migration on the same service as the live billable-events checkpoint. If the migration is not strictly additive, or if merge order relative to base langwatch/langwatch#4752 is wrong, it can disrupt existing metering.

**Top inspection targets**
- `Must Check` — `migrations/20260619120000_.../migration.sql`: confirm it only CREATEs the two new tables and touches nothing on `BillingMeterCheckpoint` or its unique index.
- `Must Check` — `dbOrganizationIdProtection.ts`: any change here affects multi-tenant org-id scoping across billing reads/writes.
- `Ask Author` — is base langwatch/langwatch#4752 merged/merging first? Reviewing this before its base can misread the diff.

`+402/-1`, 5 files, CI green, no unresolved threads. Reviewer should hold the 7-PR stack context.

Aryansharma28 (commented):
## Automated review (code-review + review skills)

Ran both LangWatch's project-level `code-review` (checked-file rubric: multitenancy, layering, naming, migrations, KSUID/no-nanoid, no-`any`, SRP) and the general PR `review` skill against this diff, then verified every finding manually against the actual files before posting.

**Scope verified:** `langwatch/prisma/migrations/20260619120000_add_storage_billing_tables/migration.sql`, `langwatch/prisma/schema.prisma`, `langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts` (+ its unit test), `langwatch/src/utils/dbOrganizationIdProtection.ts`.

Overall: this is a clean, low-risk, purely-additive schema PR. Migration SQL matches `schema.prisma` exactly, both new tables are correctly listed in `ORG_TENANCY_EXEMPT` (the reporter's cross-org `reportedAt IS NULL` sweep would otherwise fail the tenancy guard), and the new service is dormant (no callers yet, confirmed via grep) with solid isolation tests proving it never touches `BillingMeterCheckpoint`. All 12 project rubric rules (KSUID-only, no FK in migrations, projectId scoping, ClickHouse TenantId, no `any`, no hardcoded schema names, hooks/JSX, no compat re-exports, route→service→repo layering, repo/service naming, SRP, skill scenario tests) pass or are N/A for this diff.

## Blockers
None found.

## Non-blockers
1. `langwatch/prisma/schema.prisma:1829-1830` — `lastReportedTotal`/`pendingReportedTotal` are `BigInt` here, but the checked-in ADR-027 doc (`dev/docs/adr/027-storage-gb-billing.md`, Schema section) still shows these as `Int`. The `BigInt` choice is well-justified inline (Int32 overflow risk at ~2.8 TiB-months), but the ADR doc wasn't updated to match — worth a quick doc fix so PR 2-7 authors aren't reading a stale schema reference.
2. `langwatch/prisma/schema.prisma:1823` — `StorageBillingCheckpoint.id` uses `@default(cuid())`, while 71 other models in this schema use `@default(nanoid())`. Deliberate (mirrors sibling `BillingMeterCheckpoint`, also cuid-based), but it's now the 3rd cuid outlier in an otherwise nanoid-standardized schema — a one-line PR note on why would save the next reader a double-take.
3. `langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts:1-187` — Near-exact structural duplicate (~180 lines) of `billingCheckpoint.service.ts`, differing only by `number`→`bigint`. The PR body explicitly weighs and rejects a shared/generic implementation to keep each billing meter's persistence fully independent — a defensible tradeoff for a billing-critical path, just flagging the duplication for visibility.

Total: **0 blockers, 3 non-blockers.**

langwatch/prisma/schema.prisma
```diff
@@ -1793,6 +1793,47 @@ model BillingMeterCheckpoint {
...
+// table (not a meterName discriminator on the shared one) so storage billing is
+// purely additive and the billable-events checkpoint is never altered.
+model StorageBillingCheckpoint {
+  id                   String   @id @default(cuid())
+  organizationId       String
+  billingMonth         String // "2026-02"
+  // Cumulative MiB-hours reported for the period (ADR-027: Σ megabytes over the
+  // billing month). BigInt, not Int: Int32 caps at ~2.8 TiB held for a full
+  // month before overflow, which a large enterprise org can exceed.
+  lastReportedTotal    BigInt   @default(0)
```
Aryansharma28, line 1829: **[NON-BLOCKER]** `lastReportedTotal`/`pendingReportedTotal` are `BigInt` here (good call, well-justified by the Int32-overflow comment above), but the checked-in ADR-027 doc's Schema section still shows these fields as `Int`. Worth a quick follow-up to sync the doc so PR 2-7 authors in this stack aren't reading a stale schema reference.

langwatch/prisma/schema.prisma
```diff
@@ -1793,6 +1793,47 @@ model BillingMeterCheckpoint {
...
+  @@id([organizationId, sealedHour])
+  @@index([reportedAt])
+}
+
+// Two-phase reporting checkpoint dedicated to the STORAGE_GB meter — a parallel
+// of BillingMeterCheckpoint owned by StorageBillingCheckpointService. Separate
+// table (not a meterName discriminator on the shared one) so storage billing is
+// purely additive and the billable-events checkpoint is never altered.
+model StorageBillingCheckpoint {
+  id                   String   @id @default(cuid())
```
Aryansharma28, line 1823: **[NON-BLOCKER]** `id` uses `@default(cuid())` here, diverging from `@default(nanoid())` used by ~71 other models in this schema. Looks deliberate (mirrors the sibling `BillingMeterCheckpoint` right above, which is also cuid-based), but it's now the 3rd cuid outlier in an otherwise nanoid-standardized schema — a one-line note on why would help the next reader.

langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts
```diff
@@ -0,0 +1,187 @@
+import type { PrismaClient } from "@prisma/client";
```
Aryansharma28, line 1: **[NON-BLOCKER]** This service is a near-exact structural duplicate (~180 lines) of `billingCheckpoint.service.ts`, just retyped `number`→`bigint`. The PR body explicitly considered and rejected a shared/generic implementation to keep each billing meter's persistence fully independent, which is a defensible tradeoff for a billing-critical path — flagging the duplication for visibility, not asking for a change.

Aryansharma28 (approved):
Reviewed via /code-review + /review — schema-only, purely additive (two new tables, no callers yet), clean on multitenancy/layering/migration-safety. 0 blockers; left 3 non-blocking nits (stale ADR doc type, cuid() vs nanoid() convention, deliberate service duplication — all already called out in the review comments). Good to go. 🎉

sergioestebance:

PR #4855, opened 2026-06-16
perf(clickhouse): bound and harden storage-metering _size_bytes reads

langwatch-agent:
## Evidence

Storage metering's `SELECT sum(_size_bytes) ... WHERE TenantId = {..}` is the single most expensive query shape in the last 24h of prod logs:

- Signal A (slow-but-completed): one stored_spans read at **~60s / ~42 GB**, one event_log read at **~24s / ~43 GB**.
- Signal B (failed): several `sum(_size_bytes)` reads over event_log / stored_spans hit the per-query memory limit (Code 241 MEMORY_LIMIT_EXCEEDED).

Low volume (these are 5-min-cached background reads), but extreme per-call cost, and it recurs every cache cycle for the affected tenants.

## Suspected cause

`_size_bytes` is a `UInt32 MATERIALIZED byteSize(<heavy payload columns>)` (migration 00032). On parts written before the column existed there is no stored value, so ClickHouse recomputes it lazily on read by scanning the heavy source columns (`SpanAttributes`, `EventPayload`, ...). Summing across a large tenant with un-materialized parts therefore reads those payload columns for every row — tens of GB — and either grinds for ~a minute or trips the 3.5 GB per-query limit. The existing `max_threads: 2` cap bounds peak memory but not the volume read or the wall time.

## Proposed fix

This PR does not try to make the recompute cheap (that needs a backfill, see below); it bounds its blast radius and degrades gracefully:

1. **`max_execution_time: 45s`** on the metering settings. A materialized read finishes in a few seconds even for the largest tenants, so this only trips the pathological recompute path — stopping a single tenant's metering from grinding for a minute and burning tens of GB of reads every cache cycle.
2. **Resilient total.** `queryTotalBytes` (the combined `UNION ALL`) now falls back to the per-table `getStorageBreakdown` on any failure. The breakdown already catches per-table, so one heavy table tripping the limit degrades only that table to zero instead of failing the whole storage-usage metric (which currently throws and surfaces an error in the data-retention UI).

These callers are a cached storage-usage **display** (`dataRetention.getTotalStorageBytes` / `getStorageBreakdown`), so a momentarily degraded subtotal is acceptable where a 42 GB read or a hard error is not.

## Durable fix (recommended, out of band)

The real fix is to stop the recompute by backfilling the column on existing parts, during a low-traffic window (heavy mutation — reads the payload columns once per part, then every future read hits the stored UInt32):

```sql
ALTER TABLE stored_spans     MATERIALIZE COLUMN _size_bytes;
ALTER TABLE event_log        MATERIALIZE COLUMN _size_bytes;
-- ... other retention-managed tables as needed
```

This is intentionally left as an ops runbook rather than an auto-applied migration, since it is a long-running mutation on multi-TB tables and migration 00032's notes already flag heavy ALTERs on these tables as environmentally risky.

## Expected impact

No change for the common (materialized) case — those reads are already a few seconds. For the pathological un-materialized tenants: a single metering read can no longer run ~a minute / tens of GB before failing, and a heavy table's failure no longer zeroes the entire total. The underlying read volume is only removed by the backfill above.

## Test plan

`storageMeter.service.unit.test.ts`: existing guards extended to assert `max_execution_time` is applied on every metering query; new case verifies that when the combined aggregate fails (simulated Code 241), the total falls back to the per-table breakdown and returns the summed subtotals instead of throwing. 3 tests pass; `tsgo` clean.

## Notes

- Advisory PR from the ClickHouse slow-query sweep.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 45-second execution-time guard to storage metering queries to reduce the risk of runaway resource usage.
  * If the total storage aggregate query fails (e.g., memory-limit exceeded), storage metering now automatically falls back to per-table breakdowns and returns the summed result instead of failing.
* **Tests**
  * Strengthened metering unit coverage, including a heavy-tenant scenario verifying the fallback behavior and query count.

### DIFF
--- a/langwatch/src/server/data-retention/metering/__tests__/storageMeter.service.unit.test.ts
+++ b/langwatch/src/server/data-retention/metering/__tests__/storageMeter.service.unit.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
+import { RETENTION_MANAGED_TABLES } from "../../retentionPolicy.schema";
 import { StorageMeterService } from "../storageMeter.service";
 
 /**
@@ -21,6 +22,9 @@ describe("StorageMeterService memory guard", () => {
   function assertGuarded(call: { clickhouse_settings?: Record<string, unknown> }) {
     expect(call.clickhouse_settings).toBeDefined();
     expect(call.clickhouse_settings!.max_threads).toBe(2);
+    // A coarse guardrail so a runaway byteSize recompute can't grind for a
+    // minute; a materialized read finishes in seconds, well under this.
+    expect(call.clickhouse_settings!.max_execution_time).toBe(45);
   }
 
   describe("when computing the per-category storage breakdown", () => {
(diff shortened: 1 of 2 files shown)


langwatch-agent:
### CI note: residual failures are a pre-existing flake on main

The only red shard here is `test-unit (3)` (which cascades into the `langwatch-app-complete` aggregator). It is the `ee/licensing` license-signing suite:

- `licenseGenerationService.unit.test.ts > generateLicenseKey > ... round-trip validation` (4 assertions)

This suite is unrelated to this changes ClickHouse change and flakes independently on `main`: e.g. main app-ci run `27545024508` fails this exact suite, while neighbouring main runs (`27532507715`, `27427120070`) pass it. It is not deterministic and not introduced here. The `test-integration (4)` shard, when red, is the equally-flaky `groupQueue.integration.test.ts > processes only once with squashed data` timing test.

All substantive checks for this PR pass: the new ClickHouse test, typecheck, lint/code-quality, and PR conventions. Re-running the failed shards to get a clean board; the underlying flake is a separate main issue outside this PR scope.

rogeriochaves (approved):

PR #4856, opened 2026-06-16
fix(home): deterministic org resolution for /me & /governance, closing the deferred bare-nav gap (#5278)

### LINKED ISSUE #5278: Governance UI visible/usable for orgs where release_ui_ai_governance_enabled is off (ambient org resolution drift)
## Summary

Customer report: users can see and use the AI Governance UI (`/me`, `/governance`, `/settings/governance/*`) even though `release_ui_ai_governance_enabled` is **off** for their organization. The report specifically called out landing on the personal workspace (`/me`) as where this shows up.

## Root cause

`/me`, `/governance`, and `/settings/governance/*` are org-scoped pages. They resolve "current organization" via `useOrganizationTeamProject`:

```
organizations.data.find(org => org.id == localStorageOrganizationId) ?? organizations.data[0]
```
`src/hooks/useOrganizationTeamProject.ts:194-198`

`organizations.data[0]` is not deterministic: `PrismaOrganizationRepository.getAllForUser` (`src/server/app-layer/organizations/repositories/organization.prisma.repository.ts:274-337`) runs a plain `findMany` with **no `orderBy`**, so Postgres gives no ordering guarantee.

This "current organization" is exactly what feeds the `release_ui_ai_governance_enabled` check in both:
- the nav item — `src/components/sidebar/GovernSection.tsx:24-46`
- the page guard — `src/components/WithFeatureFlagGuard.tsx:47-60`

So for a **multi-organization user**, whenever `selectedOrganizationId` in localStorage doesn't resolve to a real membership (cleared storage, new device/browser, or direct/bookmarked navigation to a bare `/me` with no `?org` param), the page can land on an arbitrary *other* org the user belongs to instead of their intended one. If that other org happens to have the flag on, the user sees governance while believing they're looking at their own (flag-off) workspace.

Multi-org membership is a first-class, supported flow here (`/invite/accept`, the per-org "My Workspace" switcher in `useWorkspaceData.ts`), not an edge case — this isn't limited to internal/test accounts.

## Prior art / related

This exact gap was identified and explicitly deferred in #4856 (`feat/me-redirect-org-alignment`), which fixed the *post-login* `/` → `/me` redirect case by carrying `?org=<slug>` so the page re-pins to the resolver's org. That PR's description calls out the remaining gap directly:

> Direct navigation to a bookmarked bare `/me` (no `?org`) can still resolve the wrong org for a multi-org user whose selection has drifted — out of scope here; the reported break is the post-login `/` → `/me` redirect, which this fixes deterministically. A guard-level "fall back to any flag-on org instead of 404ing" would close that too but needs multi-org flag evaluation; deferred.

This issue tracks closing that deferred gap — the same underlying "ambient org resolution can drift" problem, in the opposite direction (grants access instead of 404ing).

## Ruled out

Checked the newer per-org/per-project feature-flag targeting-rules system (`src/server/featureFlag/rules.ts`, `featureFlagStore.postgres.ts`, #4134) as an alternative explanation — an operator rule with an empty/wildcard `match` would also silently enable a flag broadly. Ruled out as the cause here: that would flip the flag for every org uniformly and has no relationship to navigation path or "personal workspace," which doesn't match the reported symptom.

## Suggested fix

Make `getAllForUser`'s result deterministic — order by the user's own earliest `OrganizationUser.createdAt` (oldest membership first) — so the `organizations.data[0]` fallback always resolves to the user's original/primary org instead of an unordered DB pick. Add regression coverage for: a user in two orgs (flag off on their primary/first-joined org, flag on on a later-joined org) landing on a bare `/me` with no matching `selectedOrganizationId` must resolve to the flag-off org.

## Affected paths

- `src/hooks/useOrganizationTeamProject.ts:194-198`
- `src/server/app-layer/organizations/repositories/organization.prisma.repository.ts:274-337`
- `src/components/sidebar/GovernSection.tsx:24-46`
- `src/components/WithFeatureFlagGuard.tsx:47-60`

sergioestebance:
## Why

`/me` and `/governance` are org-scoped pages: they resolve their active organization ambiently (from `selectedOrganizationId` in localStorage, or a fallback), independently of whatever destination the persona resolver picked. Two related bugs come from that:

1. **Post-login redirect drift** (original scope of this PR): a governance-enabled user could be sent to `/me` but the page re-resolved a different, non-governance org and 404'd.
2. **Ambient org fallback is non-deterministic** (closes #5278): when no `?org` param and no `selectedOrganizationId` match exists — a bare/bookmarked `/me` visit, cleared storage, a new device — the page falls back to `organizations[0]`, and that array had no guaranteed order. A multi-org user could land on an arbitrary org instead of their real one. If that arbitrary org happens to have `release_ui_ai_governance_enabled` on, the user sees governance while believing they're in their own (flag-off) org — access without the flag being active for the org they think they're in.

This PR now fixes both — the second was explicitly flagged as deferred in the original version of this PR ("Direct navigation to a bookmarked bare `/me`... out of scope here... deferred").

## What changed

- `resolveHomeDestination` appends `?org=<slug>` for `/me`/`/governance` destinations so `useOrgQueryParamSelection` re-pins the page to the resolver's org on landing (original scope, unchanged).
- `PrismaOrganizationRepository.getAllForUser` now returns organizations sorted deterministically by the calling user's earliest `OrganizationUser.createdAt`, with an id-based tie-breaker when timestamps are exactly equal (e.g. bulk invite/seed in the same transaction). Previously this was a plain `findMany` with no `orderBy`, so `organizations[0]` — the fallback "current org" used throughout `useOrganizationTeamProject`, `GovernSection`, and `WithFeatureFlagGuard` — reflected Postgres's incidental order, not the user's actual primary org.

## How it works

Prisma can't express "order this top-level `organization.findMany` by a nested to-many relation's scalar field, filtered to one userId" in a single query (`orderBy` on a to-many relation only supports `_count`). So the sort happens in JS after the query, reading each org's already-userId-filtered `members[0].createdAt`.

## Test plan

- `resolveHomeDestination.unit.test.ts` — unchanged, still 14 passing (original scope).
- New `organization.prisma.repository.membership-order.integration.test.ts`:
  - Confirms `getAllForUser` returns the earlier-joined org first, with both `OrganizationUser` and `Organization` row insertion order deliberately reversed relative to the correct (createdAt) answer — so a pass can only come from the sort, not incidental DB order. Fails without the fix.
  - Confirms the tie-breaker: 5 organizations sharing one identical `createdAt`, asserting the full result is in ascending lexicographic id order — a coincidental incidental-order pass is ~1/120, not a coin flip (this was a code-review finding on an earlier 2-org version of the test, since `Organization.id` is a random nanoid, not insertion-correlated like cuid).
- `pnpm typecheck`: 0 errors.

## Anything surprising?

- **Residual limitation, not a full fix**: ordering by "earliest membership" is a heuristic for "the user's primary org," not a stored fact. A user who joined a flag-on trial/demo org before their real company org would still land on that org first — deterministically now, instead of randomly. Closing that fully would need an explicit stored default-org concept, which is out of scope here.
- The pre-existing `organization.prisma.repository.governance-filter.integration.test.ts` currently fails in this dev environment — but only in `beforeAll`'s `prisma.project.create()`, due to unrelated shared-dev-Postgres schema drift (the `PIIRedactionLevel` enum/column on `Project` doesn't exist despite `_prisma_migrations` claiming that migration ran). Confirmed unrelated to this diff (which touches only the organizations repository and its own tests) and predates this branch — not something this PR fixes.

Closes #5278

### DIFF
--- a/langwatch/src/pages/index.tsx
+++ b/langwatch/src/pages/index.tsx
@@ -54,6 +54,10 @@ export default function Index() {
           governanceUiEnabled: resolved.data.governanceUiEnabled,
           lastVisitedHomeKind,
           lastProjectSlug: project?.slug ?? null,
+          // The org the resolver ran against; carried onto /me and /governance
+          // so the org-scoped page re-pins to it instead of 404ing behind its
+          // feature-flag guard when the selected project's org has drifted.
+          organizationSlug: organization?.slug ?? null,
         }),
       );
       return;
--- a/langwatch/src/server/app-layer/organizations/__tests__/organization.prisma.repository.membership-order.integration.test.ts
+++ b/langwatch/src/server/app-layer/organizations/__tests__/organization.prisma.repository.membership-order.integration.test.ts
@@ -1,4 +1,4 @@
-import { type Organization, type User } from "@prisma/client";
+import type { Organization, User } from "@prisma/client";
 import { nanoid } from "nanoid";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { prisma } from "~/server/db";
@@ -93,7 +93,10 @@ describe("PrismaOrganizationRepository — membership order determinism (#5278)"
 
   afterAll(async () => {
     await prisma.organizationUser.deleteMany({
-      where: { userId: testUser.id, organizationId: { in: [orgA.id, orgB.id] } },
+      where: {
+        userId: testUser.id,
+        organizationId: { in: [orgA.id, orgB.id] },
+      },
     });
     await prisma.organization.deleteMany({
       where: { id: { in: [orgA.id, orgB.id] } },
--- a/langwatch/src/server/app-layer/organizations/__tests__/organization.prisma.repository.membership-order.integration.test.ts
+++ b/langwatch/src/server/app-layer/organizations/__tests__/organization.prisma.repository.membership-order.integration.test.ts
@@ -1,0 +1,206 @@
+import { type Organization, type User } from "@prisma/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { prisma } from "~/server/db";
+import { PrismaOrganizationRepository } from "../repositories/organization.prisma.repository";
+
+/**
+ * Regression test for issue #5278.
+ *
+ * `PrismaOrganizationRepository.getAllForUser` runs a `findMany` on
+ * `Organization` with no `orderBy`, so Postgres/Prisma make no guarantee
+ * about the order of the returned rows. Client code
+ * (`useOrganizationTeamProject.ts`) falls back to `organizations.data[0]`
+ * as the user's "current org" whenever `localStorage`'s
+ * `selectedOrganizationId` doesn't match a real membership — and that
+ * fallback feeds the `release_ui_ai_governance_enabled` feature-flag
+ * check on `/me` and `/governance`. Without a deterministic order, which
+ * org lands in slot 0 (and therefore which org's flags gate those pages)
+ * is effectively random.
+ *
+ * This test seeds two orgs with `OrganizationUser` membership rows
+ * inserted in one order (orgB first, orgA second) but with `createdAt`
+ * timestamps that disagree with that insertion order (orgA joined a day
+ * before orgB). It only passes if `getAllForUser` actually orders
+ * organizations by the user's membership `createdAt` (earliest first)
+ * rather than relying on incidental physical/insertion order.
+ */
+describe("PrismaOrganizationRepository — membership order determinism (#5278)", () => {
+  let repository: PrismaOrganizationRepository;
+  let orgA: Organization;
+  let orgB: Organization;
+  let testUser: User;
+  const testNamespace = `membership-order-${nanoid(8)}`;
+
+  beforeAll(async () => {
+    repository = new PrismaOrganizationRepository(prisma);
+
+    testUser = await prisma.user.create({
+      data: {
+        email: `membership-order-test-${testNamespace}@example.com`,
+        name: `Test User ${testNamespace}`,
+      },
+    });
+
+    // NOTE: `getAllForUser` queries `Organization.findMany` directly (not
+    // via `OrganizationUser`), so an unordered Postgres result tends to
+    // follow the *Organization* row's own insertion/physical order — not
+    // just the `OrganizationUser` membership row order. Create orgB's
+    // Organization row FIRST (and orgA's SECOND) so physical/insertion
+    // order for the queried table also disagrees with the correct
+    // (createdAt-based) answer; otherwise the assertion below could pass
+    // by accident of Organization row order rather than because the
+    // repository actually orders by membership createdAt.
+    orgB = await prisma.organization.create({
+      data: {
+        name: `Org B ${testNamespace}`,
+        slug: `org-b-${testNamespace}`,
+      },
+    });
+
+    orgA = await prisma.organization.create({
+      data: {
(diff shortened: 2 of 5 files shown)


rogeriochaves (dismissed):

langwatch-agent:

**Scariest prod risk:** slot `[0]` of `getAllForUser` gates the `release_ui_ai_governance_enabled` flag on `/me` & `/governance`. In the `isDemo` branch an org included via `teams/demoProjectId` has an empty `members[]`, so `members[0]?.createdAt ?? 0` sorts it to the FRONT — verify a demo/flag-on org cannot land in slot 0 for a real user and expose governance.

- `[Must Check]` `organization.prisma.repository.ts` `getAllForUser` sort — the `members` include IS filtered to `userId` (good), but confirm the `?? 0` demo-branch fallback ordering is harmless given slot 0 feeds the flag guard.
- `[Must Check]` `pages/index.tsx` + `useOrgQueryParamSelection` — verify appending `?org=<slug>` on `/me`/`/governance` re-pins without a redirect loop when the resolver's org itself lacks the governance flag.
- `[Probably Fine]` id-based tie-breaker for identical `createdAt`.

- For an `isDemo` user, does the empty-`members` org sorting to slot 0 change which org the demo pages resolve, or is demo selection always via `.find()` (not `[0]`)?

Aryansharma28 (changes requested):
## Review Summary

The deterministic-sort fix for `getAllForUser` (#5278) is solid and well-tested. The JS post-sort using `members[0]?.createdAt` (already filtered to `userId` by Prisma's `include.members.where`) is correct, the `?? 0` fallback for the demo-org branch is safe, and the 206-line integration test is one of the more carefully designed regression tests I've seen — reversed insertion vs. creation order, 5-org tie-breaker with 1/120 false-positive odds.

**However, the core mechanism for bug 1 (redirect drift) has a structural gap.**

### [BLOCKER] `?org=` param is written but never consumed

`withOrganizationParam()` appends `?org=<slug>` to `/me` and `/governance` destinations (`resolveHomeDestination.ts` line 82). The PR description says this lets `useOrgQueryParamSelection` re-pin the page to the resolver's org. **This hook does not exist.** After exhaustive search: `router.query.org` is never read by `useOrganizationTeamProject`, `usePersonalContext`, `MyLayout`, or `DashboardLayout`. The `/me` page continues to resolve its org via `selectedOrganizationId` in localStorage, then `organizations.data[0]` — completely ignoring `?org=`. The param is dead code. Bug 1 is unfixed at the page layer despite being addressed at the redirect layer.

### [BLOCKER] `detectedFallback` path bypasses `resolveHomeDestination`

`pages/index.tsx` line 53: `router.replace(detectedFallback ? "/me" : resolved.data.destination)` — the `lastVisitedHomeKind=personal` fallback hardcodes `"/me"` as a string literal rather than routing through `resolveHomeDestination`. Even if a `?org=` consumer were added, users redirected via this path would still land on bare `/me` without an org hint.

### What's working well

- `getAllForUser` sort is correct and the comment accurately explains the Prisma to-many `orderBy` limitation. [PRAISE]
- Integration test design is exemplary — double-reversed insertion/creation order ensures the test cannot pass accidentally. [PRAISE]
- `encodeURIComponent(organizationSlug)` is properly used; no XSS surface from slug injection. [NIT: the reserved-char test fixture is a bit artificial]
- No multitenancy violations — all Prisma queries maintain proper user-scoped predicates.
- CI fully green.

**Recommend requesting changes**: the `?org=` mechanism needs a consumer hook that reads the slug, resolves it to an org id, and sets `selectedOrganizationId` before the param is stripped. Without it, the redirect-drift fix is incomplete.

PR #4857, opened 2026-06-16
feat(nurturing): sync plan attribute to Customer.io across user lifecycle

### LINKED ISSUE #6562: Lifecycle messaging cannot tell a paying user from a free one
**Problem**

The marketing automation platform holds no record of what plan a user is on, so lifecycle messaging cannot distinguish a free user from a paying one. Every campaign either addresses everyone identically or relies on a manually maintained segment that drifts out of date the moment a subscription changes.

**Proposed solution**

Sync a plan attribute to the marketing platform at every point in the user lifecycle where it can change: at signup, on user record updates, and on subscription changes arriving from the payment provider. A single shared helper derives the human-readable plan label so all three paths cannot disagree about what a plan is called.

**Alternatives considered**

A periodic reconciliation job. Simpler to reason about, but it leaves the attribute stale between runs — and the moment the attribute matters most is immediately after an upgrade, which is exactly when a periodic job is at its most wrong.

**Additional context**

Small and self-contained: one shared label helper with its own tests, plus three call sites. No schema change.

**Current state**

Open, 7 files, roughly +192/-14 in a single commit. Oldest of the open branches — untouched since early July, so it needs a rebase before review.

**Related**

- PR #4857

---

### Resuming this work

| | |
| --- | --- |
| Branch | `feat/cio-plan-lifecycle-attribute` |
| Worktree | `~/workspace/langwatch/langwatch/.claude/worktrees/lexical-wishing-cupcake` |
| Session | not recoverable |

```bash
cd ~/workspace/langwatch/langwatch/.claude/worktrees/lexical-wishing-cupcake && claude
```

sergioestebance:
## What

Makes the Customer.io `plan` attribute track the user lifecycle: starts `free`, becomes the seat-event plan on upgrade (monthly vs annual), and reverts to `free` on cancellation.

## Why

`CioPersonTraits.plan` / `CioOrgTraits.plan` already existed in the trait contract but were effectively dead: only set once at signup (hardcoded `"free"` on the org group) and **never updated** when an org upgraded or cancelled. The Stripe subscription-sync hook pushed only `has_subscription`, dropping the plan. Result: marketing had no usable plan dimension after day one.

## Changes

- **New mapper** `resolveCioPlanLabel` (`nurturing/planLabel.ts`):
  - `FREE` / none → `free`
  - go-forward Growth **seat-event** plans → raw plan type verbatim (`GROWTH_SEAT_USD_MONTHLY`, `GROWTH_SEAT_EUR_ANNUAL`, …) — keeps currency + interval granularity
  - legacy / grandfathered paid plans (`LAUNCH`/`ACCELERATE`/`GROWTH`/`PRO`/`ENTERPRISE`) → bucketed by billing interval into `seat_event_annual` / `seat_event_monthly` so older cohorts fold into the same nurture tracks
- **subscriptionSync** now accepts `plan` and pushes it to every member's person `identify` **and** the org `group`; reverts to `free` when no non-cancelled subscription remains. Wired at all three webhook call sites (activate / delete / update-cancel).
- **userSync** backfills person + org `plan` on first login (selects `plan` instead of just `id`), so existing users get corrected without a subscription event.
- **signupIdentification** now also sets the person-level `plan` (`free`), via a shared `CIO_FREE_PLAN` constant reused by the org group trait.

## Scope boundary

Invited / SSO-auto-added members get their person `plan` backfilled by `userSync` on first login (seconds later). Those hooks are pure primitive-in, fire-and-forget and DB-free; making them query the org subscription would be a larger change for a momentary gap (they also have a pre-existing `has_subscription` inconsistency, left untouched here).

## Tests

- New `planLabel.unit.test.ts` — full mapping matrix (free / seat-event passthrough / legacy interval bucketing).
- Extended `subscriptionSync.unit.test.ts` — asserts the `plan` trait on person identify, the single org `groupUser` sync, and free-revert on cancel even when a stale plan is passed.
- All green; changed files typecheck clean.

## Release Notes

* **New Features**
  * Improved Customer.io plan tracking by syncing an accurate plan label for newly identified users, updating both member and org traits during subscription changes, and including plan details in webhook-driven nurturing.
* **Chores**
  * Added standardized plan-label mapping to consistently handle free, growth seat-event, and legacy subscription plan variants.
* **Tests**
  * Added unit tests covering plan-label resolution and subscription synchronization for activation and cancellation, including correcting any stale paid-plan inputs on free.

---

Closes #6562

### DIFF
--- a/langwatch/ee/billing/nurturing/hooks/signupIdentification.ts
+++ b/langwatch/ee/billing/nurturing/hooks/signupIdentification.ts
@@ -3,6 +3,7 @@ import type { z } from "zod";
 import { getApp } from "../../../../src/server/app-layer/app";
 import type { signUpDataSchema } from "../../../../src/server/schemas/sign-up-data.schema";
 import { captureException } from "../../../../src/utils/posthogErrorCapture";
+import { CIO_FREE_PLAN } from "../planLabel";
 import type { CioPersonTraits } from "../types";
 
 type SignUpData = z.infer<typeof signUpDataSchema>;
@@ -74,6 +75,7 @@ export function fireSignupNurturingCalls({
     has_prompts: false,
     has_simulations: false,
     has_subscription: false,
+    plan: CIO_FREE_PLAN,
     createdAt: new Date().toISOString(),
   };
 
@@ -86,7 +88,7 @@ export function fireSignupNurturingCalls({
       traits: {
         name: organizationName,
         ...pickDefined({ company_size: signUpData?.companySize }),
-        plan: "free",
+        plan: CIO_FREE_PLAN,
       },
     })
     .catch(captureException);
--- a/langwatch/ee/billing/nurturing/hooks/subscriptionSync.ts
+++ b/langwatch/ee/billing/nurturing/hooks/subscriptionSync.ts
@@ -1,24 +1,31 @@
 import { getApp } from "../../../../src/server/app-layer/app";
 import { prisma } from "../../../../src/server/db";
 import { captureException } from "../../../../src/utils/posthogErrorCapture";
+import { resolveCioPlanLabel } from "../planLabel";
 
 /**
- * Syncs has_subscription trait to Customer.io for all members of an organization.
+ * Syncs has_subscription + plan traits to Customer.io for all members of an
+ * organization.
  *
- * Called from the Stripe webhook service when a subscription is activated or cancelled.
+ * Called from the Stripe webhook service when a subscription is activated or
+ * cancelled. The `plan` is the internal plan type of the now-current
+ * subscription (null when none remains, i.e. the org has reverted to free).
+ *
  * Fire-and-forget: never throws, never blocks the webhook handler.
  */
 export function fireSubscriptionSyncNurturing({
   organizationId,
   hasSubscription,
+  plan,
 }: {
   organizationId: string;
   hasSubscription: boolean;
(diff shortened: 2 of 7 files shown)


rogeriochaves (approved):

langwatch-agent:

**Scariest prod risk:** No prod correctness risk (marketing attribute only); the failure mode is a wrong/stale `plan` label reaching Customer.io, e.g. cancel not reverting to `free` when one stale plan is passed, or a legacy plan bucketed into the wrong nurture track.

- `[Must Check]` `nurturing/planLabel.ts` `resolveCioPlanLabel` — verify legacy plans (`LAUNCH`/`ACCELERATE`/`GROWTH`/`PRO`/`ENTERPRISE`) bucket by billing interval correctly and go-forward seat-event plans pass through verbatim; check the covered mapping matrix is exhaustive (unknown plan string fallback?).
- `[Must Check]` `subscriptionSync` free-revert — confirm it reverts to `free` when no non-cancelled subscription remains even if a stale plan value is still passed in, and that all three webhook sites (activate/delete/update-cancel) forward `plan`.
- `[Probably Fine]` `userSync` backfill — the invited/SSO member gap relies on first-login backfill; confirm the select now pulls `plan`.

**Author questions:**
- An org with mixed subscription states (one active + one cancelled): does the sync pick the active plan or can the cancelled one win the label?

PR #4895, opened 2026-06-17
feat(datasets): chunked-JSONL storage layer + S3 schema columns (ADR-032 rungs 1-2)

sergioestebance:
> **Helm chart changes carved out → langwatch/langwatch#5083.** The deploy-layer plumbing (worker stored-objects PVC mount, worker-replica guard, the PG→S3 backfill Job + `datasetS3Migration` values, storedObjects config, operator NOTES) lives in a dedicated chart PR so the chart ships as its own release artifact and the auto-running backfill Job / its default + race-safety are decided there. This PR is born-on-storage + dual-read, so it merges and runs without those chart changes; self-hosted **local-filesystem** installs need langwatch/langwatch#5083 to release **with-or-before** this app image (the worker PVC mount). The `backfillDatasetContentToS3` task itself remains here as dormant TS (invoked only by the chart's Job or `pnpm run task`).

## For Humans

Today, when you upload a dataset to LangWatch, every row is saved as its own row in our main database (Postgres). Fine for small datasets — but it falls apart for big ones: uploads are capped at 25 MB / 10k rows, and the browser tries to parse the whole file before sending it, which freezes the tab. A customer with multi-GB image datasets simply can't get their data in.

This ADR changes **where the heavy data lives, without changing how anyone uses it**:

- **Big files go straight from your browser to S3** — they never pass through our servers, so file size stops being a bottleneck and the tab no longer chokes.
- **Rows are stored as JSONL files in S3, split into ~100 MB chunks**, instead of millions of database rows. Postgres keeps only the lightweight stuff — the dataset's name, its column types, row/size counts, and a pointer to where the data lives in S3.
- **A background worker turns the upload into those chunks** and flips the dataset's status from *processing* to *ready*. You get a progress bar and a clear status throughout; if a worker ever crashes mid-job, nothing is lost — the dataset stays at *processing* and can be re-run with a retry.
- **Everything that already uses datasets keeps working, unchanged** — experiments, the SDK, the UI all call the same function as before; it just reads from S3 now instead of the database.
- **Existing datasets migrate automatically and safely** — the old copy is never deleted until the new one is confirmed, so nothing can be lost, even during a deploy.

One honest caveat: this work gets the data **in and stored**. Actually *running* an evaluation across a full multi-GB dataset needs a follow-up change (reads are still capped at 5 MB today) — that's the very next piece of work.

---

## What

Adds **ADR-032 — Dataset content moves to S3 as chunked JSONL, with direct upload and an async normalize job** (`dev/docs/adr/032-datasets-s3-jsonl.md`). Docs-only; no code changes. This is the locked decision record that implementation will be built against.

> Numbered **032** because 029–031 are claimed by in-flight PRs langwatch/langwatch#4498 / langwatch/langwatch#4743.

## The decision (one line)

Dataset **content** moves out of Postgres into **S3 as ~100MB JSONL chunks**; heavy files upload **browser→S3 via a presigned upload to a server-owned staging key**; a **standalone GroupQueue job (with a re-drive sweeper)** normalizes them off-thread under a streaming contract; existing datasets are **flip-all-at-once migrated by a hardened, advisory-locked job off the boot path** — all behind the existing `resolveProjectStorageDestination` seam so self-hosted and BYOC keep working.

Postgres keeps **metadata + data types** (`columnTypes`, status, counters) and points at S3. Consumers (Experiments v3, Optimization Studio, REST/SDK, UI) keep calling the **same `getFullDataset()` interface** — this changes how data is *stored*, not how it is *used*.

## Why

- Customers with multi-GB (base64-image) datasets can't onboard — uploads are capped at 25 MB / 10k rows and the modal parses the file in-browser (OOMs the tab).
- The ~2-year-old "datasets in S3" path is dead (unreachable flag) and the wrong shape (single JSON blob, full read-modify-write). We rewrite, reusing only the storage plumbing.
- BullMQ is being removed (#4547), so async work uses the GroupQueue substrate.

## Highlights / red-team hardening (ADR v2)

- **Worker-death recovery is explicit:** GroupQueue only retries on a thrown error, so a re-drive sweeper re-runs datasets stuck in `processing`; normalize is held to a streaming + concurrency=1 + size-cap contract so a heavy job can't OOM the shared worker.
- **Migration is off the `set -e` boot path** (the Helm hook Job ships in langwatch/langwatch#5083), **advisory-locked** per dataset, and uses a **`contentLayout` marker (not a `useS3` overload)** so old pods mid-rolling-deploy keep reading PG. Backfill never deletes PG rows.
- **Presigned upload** uses a **server-owned, scoped, size-capped (POST-policy)** key — no cross-tenant writes.

## Explicitly out of scope (acknowledged)

Dataset **reads** are unchanged here — `getFullDataset` still truncates at 5 MB, so a 2 GB dataset is only partially consumed by a run until the **batched/streaming-reads fast-follow** (the immediate next epic). This PR is upload + storage + migration only.

## Test plan

Docs-only — no tests. Invariants + test anchors for the implementation are defined in the ADR's Invariants table.

## Summary by CodeRabbit

* **New Features**
  * Enabled large dataset storage using chunked JSONL content in object storage, while keeping dataset metadata in the database.
  * Added dataset lifecycle tracking (uploading → processing → ready/failed) so data reads are blocked until ready, with support for large-file direct uploads.
* **Bug Fixes**
  * Improved resilience for interrupted dataset processing by enabling safe retry-based recovery.
* **Documentation**
  * Added an ADR and integration specification describing the new storage architecture and safe, non-destructive migration behavior for existing datasets.

## Deployment Impact
- **Env vars:** none added/changed. Rungs 1–2 are the Prisma schema columns + the chunked-JSONL storage layer; they reuse the existing per-project storage config (`resolveProjectStorageDestination`/`createS3Client`).
- **Helm values:** none in this PR — all chart changes carved to langwatch/langwatch#5083 (must release with-or-before this app image on self-hosted local-FS).
- **Schema migration:** additive, backward-compatible — adds `contentLayout` (default `'postgres'`), `status` (default `'ready'`), and nullable `statusError`/`rowCount`/`sizeBytes`/`chunkCount`/`chunkOffsets`/`stagingKey`/`uploadFilename` to `Dataset`. Existing rows stay valid via defaults; runs through the standard prisma-migrate path.
- **Default `helm install` (no overrides):** unchanged. Existing datasets remain `contentLayout='postgres'`, `status='ready'` and behave exactly as before; no S3 dependency introduced at this rung.
- **BYOC dataplane:** no change — the storage layer uses the existing per-project S3 client; no new dataplane surface (object-key scheme only).


0xdeafcafe (dismissed):

langwatch/src/server/datasets/dataset.service.ts
```diff
@@ -534,6 +631,89 @@ export class DatasetService {
...
+            }
+          });
+        }
+      } else {
+        // No chunkOffsets (legacy/partial migration): the only way to honour the
+        // page window is to read every chunk, then slice — an UNBOUNDED read per
+        // page. Guard on sizeBytes so a large offset-less dataset can't OOM the
+        // pod on a normal list call (offsets-present datasets take the bounded
+        // branch above).
+        assertDatasetReadableInHeap(dataset);
```
Aryansharma28, line 707: **🛑 Blocker — potential heap OOM on paginated list of legacy s3_jsonl datasets**

`assertDatasetReadableInHeap(dataset)` on this line is the *only* gate on the offset-less fallback path, and it inspects `sizeBytes` only. A dataset whose total byte size is comfortably under the cap but whose row count is in the millions (small JSON rows, e.g. 100B each) will still fall into the `readChunks → parseJsonl → records.map → records.slice(windowStart, windowEnd)` flow on lines 697-703, which materializes **every row of every chunk** into JS heap before the page slice. A single `?page=1&limit=50` request can take the worker pod down — and that pod handles every tenant.

The comment on line 685 acknowledges the read is unbounded but the byte guard alone doesn't make it bounded against row count. Either (a) add an explicit `rowCount` cap to `assertDatasetReadableInHeap`, (b) reject the request when both `chunkOffsets == null` and `rowCount > N` and surface a 'recompute offsets' admin action, or (c) backfill offsets eagerly so this branch is unreachable in steady state.

— from /code-review (high effort, recall-biased verify)

langwatch/src/server/datasets/s3-dataset-storage.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  async readChunks({
+    projectId,
+    datasetId,
+    chunkCount,
+  }: {
+    projectId: string;
+    datasetId: string;
+    chunkCount: number;
+  }): Promise<unknown[]> {
```
Aryansharma28, line 137: **🛑 Blocker — `readChunks` materializes the entire dataset into a single in-memory array, contradicting the ADR's streaming contract**

The loop on lines 146-163 pushes every parsed row of every chunk into one `rows: unknown[]` and returns the flat array. There is no streaming, no async iterator, no per-chunk callback. Every consumer (`listRecords` fallback, `copyDataset` at line 1020, `getFullDataset`-style reads) inherits this — a 2 GB dataset (the exact onboarding pain this ADR set out to solve) becomes one 2 GB allocation in the worker.

The PR description explicitly claims: "normalize is held to a **streaming** + concurrency=1 + size-cap contract so a heavy job can't OOM the shared worker" — but the read primitive itself doesn't stream. The size-cap guards on the caller side (`assertDatasetReadableInHeap`) are a partial mitigation that still leaves the OOM path on offset-less reads (see comment on dataset.service.ts:688).

Suggest reshaping into either:
- `async *readChunksIter(...): AsyncIterableIterator<unknown>` so callers can break early / batch
- or, push the consumer logic down (`forEachRow`, `forEachChunk`) so it never crosses the chunk boundary in heap

— from /code-review

langwatch/src/app/api/dataset/[[...route]]/error-handler.ts
```diff
@@ -6,6 +6,39 @@ import { errorSchema } from "../../shared/schemas";
...
+  StagedUploadNotFoundError: { status: 422, code: "UploadNotFound" },
+  StorageNotWritableError: { status: 500, code: "StorageNotWritable" },
+  // A PATCH that changes columnTypes on an s3_jsonl dataset is a client request
+  // error, not a server fault — 400, matching the tRPC layer's BAD_REQUEST.
+  ColumnTypeChangeNotSupportedError: {
+    status: 400,
+    code: "ColumnTypeChangeNotSupported",
+  },
+  // A batch carrying a duplicate caller-supplied row id is a client conflict.
+  DuplicateRecordIdError: { status: 409, code: "DuplicateRecordId" },
```
Comment (Aryansharma28) [inline langwatch/src/app/api/dataset/[[...route]]/error-handler.ts:46]: **🛑 Blocker — `DatasetNotReadyError` is missing from `DOMAIN_ERROR_HTTP`, so it maps to HTTP 500**

A concurrent `POST /:slugOrId/upload` while the dataset is in `status='uploading'` or `'processing'` reaches `uploadToExistingDataset` (dataset.service.ts:1074) → `appendS3JsonlRecords` → `assertReady()` → throws `DatasetNotReadyError`. That error then:
1. Is not caught/translated in the route handler at `app.ts:550` (the surrounding try/catch doesn't map it).
2. Reaches `handleDatasetError` here.
3. Has no entry in this mapping (lines 21-39), so `domain` is `undefined` on line 57, `status` falls back to 500 on line 60.

The tRPC layer treats this as PRECONDITION_FAILED (425). The REST layer returning 500 will (a) trigger generic 'something went wrong' UX, (b) potentially page on-call for a normal user-induced race, and (c) train clients not to retry.

Add: `DatasetNotReadyError: { status: 425, code: "DatasetNotReady" }` (or 409 if you'd rather match other in-flight conflicts).

— from /code-review

langwatch/src/server/datasets/presigned-upload.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * ADR-032 R3 (v6): pure policy for the heavy browser→storage direct upload.
+ *
+ * Heavy files upload directly to a server-owned staging key; the size cap is
+ * enforced at *finalize* (HEAD the staged object) — no new dependency (the S3
+ * impl uses the already-installed `s3-request-presigner`). A POST-policy that
+ * rejects before bytes land (`createPresignedPost`) is deferred hardening.
```
Aryansharma28, line 7: **🟡 Non-blocker — presigned PUT places no size constraint; cap is enforced only after bytes have landed**

The file header (lines 6-7) acknowledges this as deferred hardening (`createPresignedPost`). It's worth surfacing the operational consequence so it gets prioritized: between presign and `finalizeUpload`'s HEAD check, a client (legitimate or malicious) can PUT up to S3's per-object limit (5 TB) at the bucket owner's expense. Abandoned/un-finalized objects then sit in the staging prefix until lifecycle policy reaps them. For a multi-tenant bucket this is a cost-amplification primitive.

Mitigations to consider before this leaves the 'rungs 1-2' phase:
- Switch to `createPresignedPost` with a `content-length-range` policy (matches the ADR's stated invariant).
- Add a strict bucket lifecycle rule on the staging prefix (e.g. 24h expiry) and document the SLA.
- Cap presign issuance rate per project.

— from /code-review

langwatch/src/server/datasets/dataset-lock.ts
```diff
@@ -0,0 +1,73 @@
...
+    datasetId: string;
+  },
+  fn: (tx: Prisma.TransactionClient) => Promise<T>,
+): Promise<T> =>
+  prisma.$transaction(
+    async (tx) => {
+      // `$executeRaw` (not `$queryRaw`): pg_advisory_xact_lock returns `void`,
+      // which $queryRaw can't deserialize. $executeRaw runs the statement and
+      // ignores the result, acquiring the lock as a side effect.
+      await tx.$executeRaw`SELECT pg_advisory_xact_lock(hashtextextended(${`dataset:${datasetId}`}, 0))`;
```
Aryansharma28, line 66: **🟡 Non-blocker — duplicates the existing `pg_advisory_xact_lock` helper in `modelDefaults.repository.ts#lockScope`**

The file header (lines 9-10) explicitly acknowledges this: "Mirrors the prior art in `modelDefaults.repository.ts#lockScope`." That mirror is the bug magnet — if the hash function, key-format convention, lock-key collision protection (the second `hashtextextended` argument), or transactional-scope semantics ever change, both copies will need to be kept in lockstep, and the next reviewer will have to verify it by hand.

Suggest extracting one helper, e.g. `withAdvisoryLock(prisma, scope: string, fn)`, in a shared location (`server/concurrency/` or similar) that both datasets and modelDefaults call. The lock-key naming convention (`'dataset:<id>'`, `'modelDefaults:<scope>'`) can stay caller-owned.

— from /code-review (Reuse lens)

langwatch/prisma/schema.prisma
```diff
@@ -586,6 +586,23 @@ model Dataset {
...
+  // Server-owned staging key bound to the row at presign time, so finalize
+  // never trusts a client-supplied key (ADR-032 D4 / C1).
+  stagingKey       String?
+  // Original upload filename, captured at presign so the normalize job can
+  // detect the file format (CSV/JSONL/JSON) — the staged object carries no name.
+  uploadFilename   String?
+  rowCount         Int?
+  sizeBytes        BigInt?
+  chunkCount       Int?
+  chunkOffsets     Json?
```
Aryansharma28, line 605: **🟡 Non-blocker — `chunkCount` and `chunkOffsets` carry overlapping state**

`chunkCount` is derivable from `chunkOffsets.length` whenever the offsets index is present. Today every write site updates both (`dataset-mutations.ts` lines 175-189, 423-427, 865-866), and the codebase has explicit guards for the drift case (`if (dataset.chunkCount == null) throw DatasetChunkCountMissingError` at dataset.service.ts:694, 1016). The fact that these guards are needed at multiple call sites is itself a signal that the redundancy is a maintenance hazard — a future write path that updates only one will produce silent drift that the read-path guards convert into hard errors instead of correct behavior.

If both columns must stay (e.g. `chunkCount` is also valid when offsets aren't computed yet), consider making the invariant explicit via a CHECK constraint, or routing all writes through a single `setChunkMeta` helper that updates them atomically.

— from /code-review (Simplification lens)

langwatch/src/tasks/backfillDatasetContentToS3.ts
```diff
@@ -0,0 +1,405 @@
...
+  // API-created data. The page order matches `findDatasetRecords` so re-runs
+  // produce identical chunks (crash-resume), and each row's id is PRESERVED so
+  // the chunk lines carry the same record ids the editor targets (I-MIG).
+  const writer = new StreamingChunkWriter({
+    storage,
+    projectId,
+    datasetId: dataset.id,
+  });
+  let cursorId: string | undefined;
+  for (;;) {
```
Aryansharma28, line 206: **🟡 Non-blocker — backfill page-loop is strictly sequential (MIGRATE_PAGE_SIZE=1000)**

The `for (;;)` loop on line 189 awaits each page fetch before issuing the next. For a multi-million-row dataset (the customer case the PR is unlocking), that's thousands of sequential round-trips to Postgres at ~10ms each — wall-clock dominated by network/IO latency, not by the chunk-write work that follows. Two cheap improvements:

1. Larger page size (10k+) once you've validated the cursor-style query plan stays index-only.
2. Pipeline: prefetch page N+1 while normalizing/writing page N (one in-flight + one pending is enough; you don't need wide fan-out).

Doesn't affect correctness; only matters if a single dataset's backfill blocks the worker for hours.

— from /code-review (Efficiency lens)

langwatch/src/server/datasets/dataset.service.ts
```diff
@@ -876,14 +1162,530 @@ export class DatasetService {
...
+    // Bound the accumulation of abandoned `uploading` rows + staging objects and
+    // recover any normalize wedged at `processing` (lost-after-send) as new
+    // uploads start. Best-effort. The pending sweep is *awaited on purpose*: it
+    // archives same-named abandoned `uploading` rows, freeing their slug BEFORE
+    // the name-conflict check below (else a same-name retry would spuriously
+    // 409). The processing re-drive has no such coupling — it only re-enqueues
+    // OTHER wedged rows — so fire it off the hot path (it swallows its own
+    // errors, so `void` can't throw).
+    await this.reapStalePendingUploads(projectId);
+    void this.reapStaleProcessing(projectId);
```
Aryansharma28: **🟡 Non-blocker — `void this.reapStaleProcessing(projectId)` is fire-and-forget without an inline `.catch()`**

The function itself wraps its body in try/catch (line 1398+) so this isn't currently surfacing unhandled rejections. But the `void` swallow combined with no defensive `.catch()` at the call site means:
- A future refactor that throws **before** the try/catch entry (e.g. a parameter validation added at the top) silently becomes an unhandled promise rejection.
- The reaper failing is invisible to the caller — `createPendingUpload` returns success even when the cleanup it triggered errored.

Cheap belt-and-suspenders fix:
```ts
void this.reapStaleProcessing(projectId).catch((err) => {
  logger.warn({ projectId, err }, 'reapStaleProcessing failed (non-fatal)');
});
```

— from /code-review

Aryansharma28:
**🟡 Non-blocker — deleted transaction-rollback test has no equivalent in the new born-on-storage path** (general comment; couldn't anchor inline)

`langwatch/src/app/api/dataset/__tests__/dataset-upload-api.integration.test.ts` loses ~79 lines, including (per the removed-behavior audit) the test that verified DB rollback when record-insert fails mid-create. The new born-on-storage flow writes chunks to S3 **before** the Postgres row insert (`upsertDataset` in `dataset.service.ts`), so on a slug-collision (`@@unique([projectId, slug])` violation) the chunks become orphans and are reaped via a best-effort `deleteAllS3JsonlChunks` call. The new test files don't appear to cover:

1. Retry of the same upload after a slug-collision actually succeeds (no leftover state on either side).
2. The reap path runs and is observable.
3. A reap failure leaves the dataset in a recoverable state and isn't silently swallowed.

Worth adding before this lands, because this failure mode is invisible until a customer hits it.

— from /code-review (Removed-behavior auditor)

langwatch/src/server/datasets/dataset.service.ts
sergioestebance: Verdict: overstated, not fixing as a blocker. Row count cannot decouple from byte size — millions of tiny rows is still hundreds of MB, which the sizeBytes cap (assertDatasetReadableInHeap) already bounds since JS heap is a bounded multiple of bytes. The offset-less branch is also near-unreachable now: every producer (migration, born-on-storage, normalize, append) writes chunkOffsets via the single chunkedMeta primitive. As defense-in-depth we did add finite-value validation of chunkOffsets (0f5dfe34a) so malformed offsets fall through to the bounded repair read instead of NaN-slicing. Keeping the byte guard; no rowCount cap.

langwatch/src/server/datasets/s3-dataset-storage.ts
sergioestebance: Verdict: valid future work, explicitly deferred by ADR-032 — reads are the immediate fast-follow epic; this PR is upload+storage+migration. The streaming contract in the PR description refers to normalize, which does stream via StreamingChunkWriter. readChunks-all is guarded by assertDatasetReadableInHeap; the async-iterator reshape lands with the batched-reads epic. Not a blocker for this rung.

langwatch/src/app/api/dataset/[[...route]]/error-handler.ts
Comment (sergioestebance) [reply langwatch/src/app/api/dataset/[[...route]]/error-handler.ts:46]: Fixed in 151424f5a: added DatasetNotReadyError -> 425 to DOMAIN_ERROR_HTTP (global, covers /upload and /records), parity with the tRPC PRECONDITION_FAILED and the read routes. Regression test added (error-handler.unit.test.ts), verified fails-without (500) / passes-with (425).

langwatch/src/server/datasets/presigned-upload.ts
sergioestebance: Acknowledged, tracked as deferred. The staging-prefix lifecycle rule shipped in langwatch-saas#604 (caps the abandoned-object residual); createPresignedPost content-length-range remains the planned hardening, logged in pending.md. No change this PR.

langwatch/src/server/datasets/dataset-lock.ts
sergioestebance: Valid reuse point, deferred as low-priority. The duplication is acknowledged in the file header; extracting a shared withAdvisoryLock helper is a boy-scout cleanup for when modelDefaults.repository is next touched, not worth the cross-module churn in this PR.

langwatch/prisma/schema.prisma
sergioestebance: Valid but already mitigated: every write site builds both through the single chunkedMeta/toJsonlChunks primitive (the v16 invariant chunkOffsets.length === chunkCount), so they cannot drift independently in practice. chunkCount is still meaningful when offsets are absent (legacy). A CHECK constraint would be over-engineering. Keeping as-is.

langwatch/src/tasks/backfillDatasetContentToS3.ts
sergioestebance: Valid efficiency point, low priority for a one-off. Prod sizing (results.md): ~2.4M rows total, est ~10-20 min wall-clock dominated by per-dataset overhead + S3 PUT latency, well under the 6h Job deadline. Cheap win noted (bump MIGRATE_PAGE_SIZE) if needed; pipelining not worth the complexity for a one-shot.

langwatch/src/server/datasets/dataset.service.ts
sergioestebance: Fixed in 151424f5a: added an inline .catch that logs non-fatally, so a future throw added above the function's internal try/catch can't become an unhandled rejection.

sergioestebance:
Addressed in 204195325 — added the born-on-storage equivalent of the removed transaction-rollback test (`dataset-service.upload.unit.test.ts`, `createNewDataset() when the row insert fails after chunks are written`):

- **(2) reap runs + observable** — on a slug race (`findBySlug` passes the pre-check, a concurrent `create` wins `@@unique([projectId, slug])`), the test asserts `deleteChunksFrom({ fromIndex: 0 })` runs and the original insert error is rethrown.
- **(3) reap failure not silently swallowed** — second test: when the reap itself throws (S3 down), the **original** insert error still surfaces, never the reap error (a failed reap must not mask the real cause).
- **(1) retry succeeds / no leftover state** — implied rather than asserted explicitly: the failed insert leaves no row, the reap leaves no chunks, and every attempt mints a fresh `datasetId`, so a retry runs clean. Pinned the safety-critical halves (no silent orphan, no masked error) at the unit level rather than a full integration retry loop.

Unit, not integration, since the create is no longer a single PG transaction — the rollback contract is now the chunk-reap, which is service-layer logic.

0xdeafcafe (approved):

PR #4905, opened 2026-06-17
feat(datasets): direct browser→S3 upload — ADR-032 rung 3 (stacked)

sergioestebance:
## Stacked on langwatch/langwatch#4895 (base: `feat/datasets-s3-jsonl`)

ADR-032 **rung 3 (direct browser→S3 upload)** + the storage-service refactor. Review langwatch/langwatch#4895 first. Rung 4 (async normalize) lands in this same PR next.

### What's here
**Storage service (DIP, ADR v7):** provider-pluggable `DatasetStorage` — `dataset-chunking.ts` (pure chunk math) + `dataset-storage.ts` (interface + `getDatasetStorage(projectId)` factory) + `s3-dataset-storage.ts` / `local-dataset-storage.ts`. getApp-registerable later.

**Direct upload (rung 3, ADR v6):** server-owned tenant-scoped `staging/{projectId}/{uploadId}` key, **presigned PUT** (existing `s3-request-presigner`, no new dep), size cap enforced at **finalize** (HEAD). `DatasetService.createPendingUpload` (presign → create `Dataset` in `uploading`) + `finalizeUpload` (tenant-prefix guard + size cap → `processing`/`failed`). Routes `POST /direct-upload` and `/direct-upload/:datasetId/finalize`. Self-hosted without browser-reachable S3 → 409 `DirectUploadUnavailable` → client falls back to the backend `/upload` path.

**Normalize resilience (no stranded `processing`):** the normalize enqueue stays fire-and-forget (ADR-D5/M4 — never block the HTTP response), but a synchronous queue `.send()` rejection now flips the wedged row to `failed` via a guarded `failIfProcessing` (so the drawer surfaces retry instead of an eternal spinner; guarded so it never clobbers the inline handler's own failure). The *lost-after-send* window (send resolves, worker dies mid-job) is covered by `reapStaleProcessing` — a poll-triggered re-drive of `processing` rows older than `STALE_PROCESSING_TTL_SECONDS`, same no-cron pattern as the stale-pending sweep, safe because the re-drive is idempotent (I-IDEM guard + concurrency-1 group).

**Correctness hardening (review):** `finalizeUpload`'s `uploading→processing` transition is now an **atomic guarded claim** (`updateMany WHERE status='uploading'`) so two racing finalizes can't both enqueue normalize (inline-mode chunk-desync guard) — same pattern as `failIfProcessing`. The `StagedUploadNotFound` branch **nulls `stagingKey`** (parity with over-cap) so retry can't loop on a missing object. s3_jsonl writes **reject a duplicate caller-supplied row id** (`DuplicateRecordIdError` → 409/CONFLICT) to keep the legacy PG `(id, datasetId, projectId)` uniqueness (I-PG). Domain errors now map to 4xx in two more spots (`DatasetNotReadyError` → 425 in the workflows route, `ColumnTypeChangeNotSupportedError` in `DOMAIN_ERROR_HTTP`). The direct-upload CSRF gate **fails closed** when neither `Sec-Fetch-Site` nor `Origin` is present.

### Tests
Unit green (storage-factory selection, chunk math, presign policy, service flow, normalize recovery + stale-processing re-drive, finalize-race claim, duplicate-id rejection, CSRF fail-closed — boundary-mocked per TESTING_PHILOSOPHY); real-FS integration test for the local backend (runs in CI/testcontainers). `pnpm typecheck` clean. `check:feature-parity` green — the 14 spec scenarios stay `@unimplemented` (tracked by langwatch/tasks#84) until the full upload→ready flow (rung 4) binds them.

## Deployment Impact
- **Env vars:** none added or changed. Direct upload reuses the existing per-project storage config (`resolveProjectStorageDestination` / `createS3Client`); `UPLOAD_MAX_BYTES` (~5 GiB) is a code constant, not env.
- **Helm values:** none.
- **Default `helm install` (no overrides):** unchanged. Direct upload only activates when object storage (S3) is configured for the project. With no S3 (single-node self-hosted default), `createPresignedUpload` raises `DirectUploadUnavailable` → the API returns 409 → the client uses the existing backend `/upload` path. No behavior change for vanilla installs.
- **Bucket CORS:** browser→S3 presigned **PUT** requires the dataset bucket to allow `PUT` from the app origin (CORS). Deployments that want direct upload must configure that CORS rule; without it the browser PUT fails and the client falls back to backend upload. No infra auto-provisions this yet (follow-up).
- **BYOC dataplane:** uses the existing per-project S3 client, so BYOC keys/buckets work unchanged. Adds a new `staging/{projectId}/` key prefix (and `datasets/{projectId}/{datasetId}/chunk-*.jsonl` once normalize lands) within the same bucket — no new dataplane surface beyond object keys.


0xdeafcafe:
No P0s, five P1s. Three are real bugs in this PR; two are ADR-acknowledged but the surface widened with rung 3.

### Bugs to land in this PR

1. **Over-cap finalize leaves `stagingKey` populated** (`dataset.service.ts:1004-1018`). A subsequent `retryNormalize` flips the dataset back to `processing`, enqueues a normalize against the now-deleted staging key, and fails with the wrong cause ("Uploaded object not found" instead of the actual over-cap). Null the `stagingKey` in the same `repository.update` when finalize fails permanently.

2. **Mid-stream parse failure orphans chunks in S3** (`dataset-normalize.job.ts:354-356`). The empty-upload check runs *after* `parseInto` + `writer.finalize`, so a JSONL parse error at row 50 of 100 leaves chunks 0..N flushed and never reaped. `deleteChunksFrom(0)` should run in the catch (best-effort, swallow secondary errors).

3. **No deduplication on `datasetNormalize`** (`pipelineRegistry.ts:559-561`). Comment says "the per-dataset group key serialises execution so no dedup needed" - but serialisation isn't dedup. A double-click on Retry enqueues two jobs that both run sequentially, both flip `status=processing`, both write chunks. Add `deduplication: { makeId: p => p.datasetId, ttlMs: 60_000 }`.

### ADR-deferred, but worth a sanity check

4. **Presigned PUT has no Content-Length cap** (`s3-dataset-storage.ts:160`). ADR v6 defers this to v7, fair. Worth flagging that until the bucket lifecycle TTL kicks in, an authenticated user can PUT up to S3's 5 TiB ceiling. Setting `ContentLength` on `PutObjectCommand` is a one-liner and makes S3 reject mismatched PUTs.

5. **`void enqueueDatasetNormalize(...).catch(...)`** in both `finalizeUpload` and `retryNormalize`. In the no-Redis runInline path (dev, tests, no-Redis self-host) the whole normalize runs inside the handler that just `void`'d it, which races subsequent reads. Either await on the inline branch, or `setImmediate` it.

### P2/P3 worth a pass

- 8x repeated error-name string-compare in the new routes; the central `handleDatasetError` already exists.
- `uploadFilename ?? \`${datasetId}.jsonl\`` is a footgun (silently picks JSONL parsing if a CSV row is missing the filename).
- The `presign-before-row` order leaks orphans if the row create fails after presign succeeds.
- `pipelineRegistry.ts` diff is ~80% biome import reordering; would be easier to review with that lifted into a separate commit.

The spec file has `@unimplemented` scenarios this PR actually completes - worth binding them now.

Otherwise solid - the chunk-streaming logic and the stacked-PR isolation are nice.

sergioestebance:
Thanks — sharp review. Triaged all five; landed the real ones in `fc902f3a0`.

**Landed**

1. **Over-cap leaves `stagingKey`** ✅ — now nulled in the same `repository.update` that flips to `failed`. Over-cap is terminal (re-upload smaller), so `retryNormalize` correctly refuses (`DatasetNotRetryableError`) instead of re-driving a deleted object. Regression test asserts both the null and the not-retryable retry.

2. **Mid-stream parse orphans chunks** ✅ — `deleteChunksFrom({ fromIndex: 0 })` now runs best-effort in the normalize `catch` (swallows secondary errors; staging preserved for retry). Narrower than stated in one respect: a *successful* retry already self-heals via the `fromIndex: meta.chunkCount` trim — the leak only bit **permanently-failed** datasets (chunk keys have no lifecycle TTL). Regression test forces a real mid-stream flush, then fails, and asserts the reap.

4. **Presigned PUT cap** — left for v7 per ADR, but noting for the record: `ContentLength` on `PutObjectCommand` signs an *exact* byte count (client must send precisely that), not a cap. A real cap needs a presigned **POST** with `content-length-range`. Residual risk is bounded — finalize HEAD-checks and deletes over-cap, staging has a lifecycle TTL.

**Not applied — with reasoning**

3. **Dedup on `datasetNormalize`** ❌ — the "both write chunks" failure can't occur: the handler guards `status !== "processing" → return`, and the per-dataset group key serializes to concurrency-1. Job A flips → `ready`; Job B then sees non-`processing` and no-ops. Serialization + the status guard *is* effective idempotency. A `ttlMs: 60_000` dedup would also swallow legitimate fast retries — the exact thing the existing comment avoided. (The "Retrying preparation does not duplicate rows" scenario is now bound to the re-drive trim test.)

5. **`void enqueue().catch()` inline race** ❌ — by design (M4): fire-and-forget even inline honors ADR D5 "no synchronous-in-request" for single-node self-host; the client contract is poll-until-`ready`. `await`-ing the inline branch reintroduces what D5 forbids.

**P2/P3 swept**

- **Centralized the ~8× error-name ladders** into a `DOMAIN_ERROR_HTTP` table in `error-handler.ts`; the 5 direct-upload routes now propagate to `onError`. (Caught a bonus while doing it: logged status was computed before the mapping, so domain errors logged `[500]` while returning 4xx — fixed.)
- **`?? `${id}.jsonl`` footgun** — now fails loudly; M1 co-sets filename with the staging key, so a present-key + no-filename is a corrupt row, not a `.jsonl` guess.
- **Bound the `@unimplemented` scenarios** this rung completes (recovery retry, no-duplicate re-drive, project isolation) with `@scenario` annotations; left the genuinely-later ones tagged.

Left for follow-up (out of scope here): migrating the *older* slug/record routes onto the same central table (they still map `DatasetNotFoundError` inline) — added a cross-reference comment so the two styles don't drift.

Integration tests couldn't run locally (no Docker/testcontainers) — CI will validate; unit suite + typecheck are green.

Aryansharma28:
## Review: PR langwatch/langwatch#4905 — direct browser→S3 dataset upload (ADR-032 rung 3)

Big, well-structured stack: introduces a provider-pluggable `DatasetStorage` interface (S3 + local-fs), a server-owned presigned-PUT direct-upload flow with a `staging/{projectId}/...` tenant prefix and HEAD-based size cap at finalize, and a normalize-recovery layer (`failIfProcessing` + `reapStaleProcessing`) so the wedged-`processing` window has bounded recovery latency. Surface area is large (84 files, +16k); the comments below are the issues a maintainer should act on before merge.

Reviewed via 8 finder angles + per-candidate verification against the diff and the PR-branch sources. All findings below were verified as CONFIRMED (not PLAUSIBLE).

---

### 🚧 Blockers

**1. CSRF gate fails open when both `Sec-Fetch-Site` and `Origin` are absent**
`langwatch/src/app/api/dataset/[[...route]]/direct-upload-auth.ts:1438` — `if (!origin) return false;` returns "not cross-site" when neither header is present. A page on evil.com can submit a cookie-bearing `multipart/form-data` POST on a runtime that strips both headers (older webviews, certain proxies, downgraded contexts) and pass the CSRF gate. `authorizeDirectUpload` then runs `hasProjectPermission` with the victim's session and mints a presigned PUT + an `uploading` Dataset row in the victim's project. Fail-closed policy: require at least one of `Sec-Fetch-Site` / `Origin` to be present *and* match — treat missing-both as cross-site.

**2. `parseBody()` runs before `authorizeDirectUpload` on `POST /api/dataset/direct-upload`**
`langwatch/src/app/api/dataset/[[...route]]/app.ts:1067` — handler calls `await c.req.parseBody()` first, then `authorizeDirectUpload(c, projectId.trim())`. An unauthenticated attacker can POST a large multipart body and force the server to spool it before 401. The prior access-chain (`requires('datasets:manage')`) rejected before reading the body. Move auth before body read, or cap `parseBody` size on this route.

**3. `forcedIds` accepted with no duplicate-id check; later edits silently target only the first occurrence**
`langwatch/src/server/datasets/dataset-mutations.ts:13673` — `appendS3JsonlRecords` writes `id: forcedIds?.[i] ?? record_${nanoid()}` with no dedupe pass; `editS3JsonlRecord` / `deleteS3JsonlRecords` use `rows.findIndex(... id === recordId)`. A client double-submit (or buggy SDK) that supplies the same id twice persists both, then `update`/`delete` for that id mutates only the first — second copy becomes an unreachable ghost with stale data. Old PG layout had a PK on `(id, datasetId, projectId)` that raised on dup.

**4. `finalizeUpload` has a read-then-write race on `status='uploading'`**
`langwatch/src/server/datasets/dataset.service.ts:16165` — guard at 16172 reads `status !== 'uploading'`, transition at 16240 is an unconditional `repository.update({ data: { status: 'processing' } })`. Two concurrent finalize calls (client retry / double-click) both pass and both fire `enqueueNormalize`. In inline-mode (no GroupQueue concurrency-1), two handlers race writing `chunk-00000` and `chunkOffsets`/`rowCount` desync from the chunk objects. Fix is atomic: `updateMany({ where: { id, status: 'uploading' }, data: { status: 'processing' } })` and gate the rest on `count === 1`.

**5. `StagedUploadNotFoundError` path flips `status='failed'` but leaves `stagingKey` non-null → infinite Retry loop**
`langwatch/src/server/datasets/dataset.service.ts:16204` — failure branch at 16208 is `data: { status: 'failed', statusError: 'Uploaded object not found' }` with no `stagingKey: null`, while the over-cap branch at 16234 explicitly clears it. `retryNormalize` (16331) passes `status==='failed'` AND `stagingKey != null`, flips back to `processing` and re-enqueues; normalize HEADs the missing key, flips to `failed` again, repeat. Each click queues a real normalize job. Null `stagingKey` here so retry hits `DatasetNotRetryableError` immediately.

**6. `captureHeaders` short-circuits after the first record → late-row columns dropped from `columnTypes`**
`langwatch/src/server/datasets/dataset-normalize.job.ts:14572` — `if (headers.length > 0) return;`. JSONL whose first row is `{"a":1}` and second is `{"a":2,"b":3}` ends up with `columnTypes=[{name:'a',...}]`. The chunk file still contains `b`, but the column-aware editor never shows it, and the column-validating `POST /records` route (`Column \`b\` is not present`) rejects any later edit that includes the `b` field that was in the original upload. Union keys across rows (or scan first N).

---

### 💡 Non-blockers

**7. Domain errors not mapped → generic 500 in two consumers**
- `langwatch/src/server/routes/workflows.ts:159` — `loadDatasets` now throws `DatasetNotReadyError` per ADR-032 I-READY; the workflow-route catch returns `status: 500` + Sentry capture. Should map to 4xx (425 / 409) like the dataset REST layer does.
- `langwatch/src/app/api/dataset/[[...route]]/error-handler.ts:21` — `DOMAIN_ERROR_HTTP` omits `ColumnTypeChangeNotSupportedError`; a `PATCH` that changes `columnTypes` on an `s3_jsonl` dataset returns 500 instead of a user-actionable 4xx.

**8. Test naming violates repo-root `CLAUDE.md` rule "render → integration, not unit"**
- `langwatch/src/components/datasets/__tests__/CSVReaderComponent.unit.test.tsx` (renders `<CSVReaderComponent />` via `render(...)`)
- `langwatch/src/components/datasets/__tests__/UploadCSVDrawer.unit.test.tsx` (renders inside `<ChakraProvider>`)

CLAUDE.md: *"Tests that render components and mock boundaries are integration tests (`.integration.test.ts`), not unit tests."* Rename both to `.integration.test.tsx`.

---

*(Also worth tracking but below the cut: presigned PUT signs no `content-length-range` so a malicious client can land a 5 GB blob in BYOC S3 regardless of `UPLOAD_MAX_BYTES` — the PR's ADR already lists `createPresignedPost` as deferred hardening; `reapStaleProcessing` enqueues before awaiting `markProcessingRedriven`, so a swallowed DB error there re-queues on every subsequent upload until the in-flight job lands.)*

sergioestebance:
Thanks @Aryansharma28 — thorough pass, and the multi-angle verification paid off. Almost everything verified as real against the code. I landed 7 of the items, skipped 1 with rationale, and deferred 1 as a semantics decision. Quick severity note up front: there were no true P0s — langwatch/langwatch#4 is a genuine P1, the rest land at P2/P3 — but several were cheap, pattern-consistent fixes, so they went in regardless.

Commits: `28c64acd` (fixes) + `9d6c98dd` (lint).

### Landed

**#4 — non-atomic `uploading→processing` (the real one).** Replaced the read-then-write with an atomic guarded claim: `claimForProcessing` does `updateMany WHERE status='uploading'` and the transition is gated on `count === 1`. Two racing finalizes can no longer both enqueue; the loser bails as a finalize replay. Same atomicity pattern as `failIfProcessing`/`markProcessingRedriven`, and notably it's mode-agnostic — it closes the inline-mode double-write that a queue-level dedup never would (which is also why we kept rejecting the earlier "add dedup to datasetNormalize" suggestion).

**#5 — `StagedUploadNotFound` leaves `stagingKey`.** Now nulls it in the same update, parity with the over-cap branch. Retry hits `DatasetNotRetryableError` immediately instead of looping doomed normalize jobs against a missing object.

**#3 — `forcedIds` no dedup → ghost rows.** Restored the legacy PG `(id, datasetId, projectId)` uniqueness (I-PG): a duplicate caller-supplied id in one write now throws `DuplicateRecordIdError` at the `toChunkLines` chokepoint (→ 409 on REST, `CONFLICT` on tRPC) instead of silently persisting a ghost. (Within-batch; a cross-batch collision against an already-stored id would need an O(rowCount) scan on every write — disproportionate, and the edit create-on-miss path already guards the common upsert.)

**#7a / #7b — domain errors → 500.** `workflows` route now maps `DatasetNotReadyError` → 425 and skips the Sentry capture (expected transient state). `DOMAIN_ERROR_HTTP` gains `ColumnTypeChangeNotSupportedError` (→ 400, matching the tRPC `BAD_REQUEST`) and `DuplicateRecordIdError`.

**#1 — CSRF fail-open on missing-both.** Flipped to fail **closed**: with neither `Sec-Fetch-Site` nor `Origin` present, treat as cross-site. Only the cookie path is affected and a real same-site upload always carries one of the two, so no legitimate flow regresses.

**#8 — `.unit` render tests.** Renamed both to `.integration.test.tsx` — matches the repo's dominant convention (122 component render tests already there vs. these 2 outliers).

### Skipped

**#2 — `parseBody()` before auth.** Not a vuln in this shape: the heavy upload never transits this route (presigned S3 PUT, or the staging PUT which auths *before* reading the stream). This `POST /direct-upload` body is just `{projectId, name, filename}` form fields — there's no large body to spool. Reordering would be harmless hygiene but it isn't the described DoS, so I left it.

### Deferred (logged)

**#6 — `captureHeaders` first-row-only.** Confirmed the inconsistency (the chunk stores `b` while `columnTypes` hides it, so the column-validating route then rejects edits touching `b`), but there's no clean bug-fix: "union all rows" changes our rectangular, schema-first contract and fights streaming. It's a genuine semantics decision (commit-to-rectangular = drop non-schema keys on write, as CSV already does, vs. widen), so it's written up as a decision rather than silently patched.

### Tail
- **content-length-range / 5 GiB blob** — already ADR-v7-deferred; the presigned-POST `content-length-range` hardening is tracked there.
- **`reapStaleProcessing` re-queues** — already addressed at HEAD: `markProcessingRedriven` stamps `updatedAt` after each re-drive so a row isn't re-selected within the TTL. (The one residual — a *swallowed* bump DB error — is bounded: it stops the moment the in-flight job lands, and the re-drive is an idempotent no-op anyway.)

Appreciate the review — the finalize-race catch in particular was a good one.

Aryansharma28 (approved):

PR #4908, opened 2026-06-18
feat(datasets): dataset content → S3 chunked JSONL — reads, writes, migration, born-on-storage, self-host (ADR-032)

sergioestebance:
## For humans

Datasets (the tables of examples you upload to run evaluations on) used to be stored inside the main database, which choked on anything large. This change stores them as files in object storage instead — so you can now upload **multi-gigabyte datasets without freezing the app**, and viewing, editing, and deleting rows all keep working exactly as before.

It works in two setups with **no extra configuration**:
- **Cloud / team:** uses your S3 bucket.
- **Simple self-hosted:** uses the server's own disk — no S3 required.

Datasets you already have keep working and are moved over **automatically and safely** in the background. Nothing you do today changes.

---

## What this PR does

Moves **dataset content out of Postgres into S3-compatible object storage as chunked JSONL** (ADR-032), end-to-end: storage, reads, writes, migration of existing data, the upload UX, and self-host gating. After #4895 (chunking primitives) and #4905 (upload path), this PR makes S3-backed datasets fully usable and makes S3 the **default for all new data**.

> Implements **ADR-032** (`dev/docs/adr/032-datasets-s3-jsonl.md`, v1–v14).

## Scope (the ladder)

**Storage + reads/writes (ADR-032 R5–R7)**
- **Read-routing + status-gating (R6a):** `contentLayout='s3_jsonl'` datasets read via `readChunks`; every read consumer gates on `status='ready'` (`DatasetNotReadyError` → REST **425**, tRPC `PRECONDITION_FAILED`). Rows are `{id, entry}` for stable identity. Single-row reads short-circuit via `chunkOffsets` (O(1 chunk)).
- **Write-mutations (R6b):** append → new chunks from `chunkCount`; edit/delete → rewrite only the affected chunk; all under a per-dataset `pg_advisory_xact_lock` with PG-authoritative counters. `recomputeDatasetCounts` is the I-COUNT repair; `MissingChunkError` is loud on corruption.
- **Migration (R5):** idempotent, advisory-locked PG→S3 backfill, **off the boot path** (cloud k8s Job / self-host Helm `post-install,post-upgrade` hook Job / `pnpm run task`), self-skips when storage is unconfigured or no `postgres` rows remain. `SKIP_DATASET_S3_MIGRATE` opt-out. Migration prefers S3, falls back to local FS.
- **Self-host gating (R7):** `resolveProjectStorageDestination` (BYOC → `S3_BUCKET_NAME` → local FS) is the single source of truth; local FS fully supports chunked JSONL.

**Born-on-storage cutover (step 1)**
- Every create path (tRPC `create`, REST `POST /` + `/upload`, copy/clone, SDK/MCP) is born `contentLayout='s3_jsonl'`, `status='ready'` with PG-authoritative counters — `postgres` is never created for new data, so the backfill drains to zero. Chunk write runs **before** the row insert (no orphan row); a partial write self-reaps, a failed insert reaps via `deleteAllS3JsonlChunks`. Caller-supplied row ids are honored (create *and* append/batch-create).

**Trailing-chunk compaction**
- A delete that empties the highest chunks lowers `chunkCount` + trims offsets **logically** (no in-tx object delete → no lock-free-reader race); trailing objects are benign 0-byte orphans overwritten by the next append.

**Heavy uploads without S3 (local-FS direct upload, D4 extension)**
- `LocalDatasetStorage.createPresignedUpload` mints a **same-origin** streaming URL; a new session-authed `PUT /direct-upload/staging/:uploadId` streams the body to local-FS staging (gated on an owning `uploading` row; inline 5 GB cap). So multi-GB uploads work on no-S3 self-hosted installs, not just S3.

## Hardening folded in from review
record-count unified across all 3 layouts (s3_jsonl → `rowCount`); CSV normalize no longer corrupts rows with two equal cells (papaparse `header:false` + once-only header dedup); `StorageNotWritable` surfaced as an actionable error instead of a misleading fallback; pending `uploading` row reaped on any non-fallback upload failure; staging route refuses orphan writes; Helm NOTES document the default no-S3 storage path.

## Tests / verification
- Unit per rung (boundary-mocked per TESTING_PHILOSOPHY); real-FS integration in CI; `pnpm check:feature-parity` green. **All CI checks green.**
- **Browser-proven both storage paths** (headless): a 5,000-row CSV uploaded with S3 off lands in local FS, with S3 on lands in `s3://…` — each in exactly one backend, 0 corrupted rows, "5000 records" in the UI.

## Deployment Impact
- **Storage is now MANDATORY on dataset create — no Postgres fallback.** A writable backend is required: S3 (`S3_BUCKET_NAME` / `app.dataplane`) or a writable local-FS path.
- **Default `helm install` (no S3) works out of the box:** the chart provisions a local-FS **PVC** (`app.storedObjects.localFilesystem`, 10Gi), mounts it, `sharedEnv` sets `LANGWATCH_LOCAL_STORAGE_PATH`, and `fsGroup` makes it writable. **Single-replica only** — the chart refuses to render with `replicaCount > 1` (RWO volume); multi-pod must enable `app.dataplane` (S3/MinIO/Azure).
- **Existing `postgres` datasets:** read exactly as before; the migration self-skips without storage configured.
- **Env:** adds `SKIP_DATASET_S3_MIGRATE` (opt-out for the one-shot backfill). `UPLOAD_MAX_BYTES`/`CHUNK_MAX_BYTES` are code constants.
- **S3 deployments:** bucket **CORS** (allow `PUT` from app origin, for the direct upload) and a **lifecycle TTL** on `staging/{projectId}/` (reap un-finalized uploads) are bucket-config tasks, not auto-provisioned. Off-request normalize needs a **worker + Redis** on multi-pod.


langwatch/src/server/datasets/dataset-normalize.job.ts
```diff
@@ -115,7 +123,10 @@ class StreamingChunkWriter {
     },
   ) {}
 
-  async push(record: unknown): Promise<void> {
+  async push(entry: unknown): Promise<void> {
+    // Wrap the raw row as { id, entry } — the stable per-row id later
+    // edit/delete targets. nanoid() per record keeps this streaming.
+    const record = { id: `record_${nanoid()}`, entry };
```
langwatch-agent, line 156: **P1** — This writer path still keeps the full `jsonl` body for every flushed chunk in `this.chunks`, because `writeChunks()` returns `DatasetChunk` objects that include the serialized payload and `flush()` pushes those objects into the long-lived chunk list. A 2-5 GiB direct upload will still accumulate roughly the whole normalized file in heap by `finalize()`, despite flushing each 16 MiB buffer. Store only lightweight chunk metadata/running totals after each flush, or change the storage API to return metadata without payloads, and add a regression that the normalize writer does not retain chunk bodies.

langwatch/src/server/api/routers/datasetRecord.utils.ts
```diff
@@ -177,6 +376,70 @@ export const getFullDataset = async ({
...
+        return {
+          ...dataset,
+          count: dataset.rowCount ?? 0,
+          datasetRecords: [record],
+          truncated: false,
+        };
+      }
+    }
+
+    const rows = await storage.readChunks({
```
langwatch-agent: **P1** — The `all` read path loads every S3 chunk into memory before applying the 5 MB truncation. `getAll` defaults to `limitMb: 5`, but it still calls `readChunks()` for the whole dataset first; `download` uses this path with `limitMb: null`, so a 5 GiB dataset can be materialized inside the app pod. This defeats the S3 chunking memory guarantee. Read chunks incrementally and stop once the requested selection/byte limit is satisfied, and make full export a streaming/async path instead of returning all rows.

langwatch/src/server/datasets/dataset.service.ts
```diff
@@ -538,6 +555,43 @@ export class DatasetService {
...
+    if (dataset.contentLayout === "s3_jsonl") {
+      if (dataset.status !== "ready") {
+        throw new DatasetNotReadyError({
+          status: dataset.status,
+          statusError: dataset.statusError,
+        });
+      }
+
+      const storage = await getDatasetStorage(params.projectId);
+      const rows = await storage.readChunks({
```
langwatch-agent: **P2** — Pagination for `s3_jsonl` datasets reads every chunk and then slices the requested page. That means `page=1&limit=50` can still download and parse a multi-GB dataset into memory. Use `chunkOffsets` to identify only the chunks overlapping the requested page, or add a paged storage reader; the test suite should assert the paginator only reads the necessary chunk(s), not bless the full-dataset read.

langwatch/src/components/datasets/UploadCSVModal.tsx
```diff
@@ -165,83 +198,151 @@ export function UploadCSVForm({
...
+      });
+      onClose?.();
+      void router.push(`/${project.slug}/datasets/${datasetId}`);
     } catch (error) {
-      logger.error({ error }, "Failed to get valid name");
+      if (error instanceof DirectUploadUnavailableError) {
+        // No browser-reachable storage: parse the already-captured file NOW
+        // (the first and only in-browser parse on this path) and hand off to
+        // the existing parse-and-drawer flow.
+        await runFallbackParseAndDrawer(rawFile);
```
langwatch-agent: **P2** — On no-S3/self-hosted installs, a direct-upload-unavailable response falls back to parsing the same raw file in the browser with no size guard. A user can drop a multi-GB file, get the 409 fallback, and then the tab does exactly the in-browser `file.text()` parse this flow was meant to avoid. Gate this fallback by the legacy backend/file-size limit and show a clear “large uploads require object storage” error instead of parsing.

langwatch/src/server/datasets/s3-dataset-storage.ts
```diff
@@ -157,6 +159,60 @@ export class S3DatasetStorage implements DatasetStorage {
...
+    index,
+    records,
+  }: {
+    projectId: string;
+    datasetId: string;
+    index: number;
+    records: unknown[];
+  }): Promise<ChunkOffset> {
+    assertNoTraversal(projectId, datasetId);
+    const { jsonl, byteSize } = toSingleJsonl(records);
```
langwatch-agent, line 206: **P2** — `rewriteChunk()` can write an object larger than `CHUNK_MAX_BYTES`. An edit can replace a small row with a large value, `toSingleJsonl(records)` serializes the whole chunk, and the code uploads it without enforcing the 16 MiB cap or splitting/recomputing offsets. That breaks the chunk-size invariant the read/write memory model depends on. Reject edits that would exceed the cap, or split/rebalance the affected chunk under the dataset lock.

langwatch-agent (dismissed):
Not ready to merge: 2 P1s and 3 P2s. The main issue is that the PR accepts multi-GB datasets but several normal paths still retain or materialize the full dataset in memory; details inline.

langwatch/src/server/datasets/dataset-normalize.job.ts
sergioestebance: Fixed in 5736f7a4: StreamingChunkWriter now accumulates ChunkMeta only (chunkMetaOf drops the jsonl payload after each flush); finalize() builds ChunkedDatasetMeta from metadata alone, so peak heap is ~O(1 chunk) regardless of file size. Re-verified by a dedicated memory-contract review.

langwatch/src/server/api/routers/datasetRecord.utils.ts
sergioestebance: Fixed in 5736f7a4: reads now pull chunks one at a time via readChunk(index) and stop at the limitMb byte budget; full export (limitMb null) is guarded by DATASET_FULL_EXPORT_MAX_BYTES (~100MB) which throws DatasetTooLargeToExportError. True streaming export is the deferred reads-at-scale epic.

langwatch/src/server/datasets/dataset.service.ts
sergioestebance: Fixed in 5736f7a4: listRecords for s3_jsonl uses chunkOffsets to read only the chunk(s) overlapping the requested page, not the whole dataset.

langwatch/src/components/datasets/UploadCSVModal.tsx
sergioestebance: Fixed in 5736f7a4 + 00990fec: the fallback is size-guarded by MAX_FILE_SIZE_BYTES before any in-browser parse; additionally a missing-CORS presigned-PUT failure now routes to that same guarded fallback (PresignedUploadFailedError) plus orphan-row cleanup, instead of dead-ending.

langwatch/src/server/datasets/s3-dataset-storage.ts
sergioestebance: Fixed in 5736f7a4: rewriteChunk rejects records whose serialized size exceeds CHUNK_MAX_BYTES with ChunkTooLargeError (mapped to BAD_REQUEST) in both the S3 and local impls, before writing.

langwatch/src/server/datasets/dataset-mutations.ts
```diff
@@ -0,0 +1,583 @@
...
+    const chunkCount = current.chunkCount ?? 0;
+    const perChunk: Array<{ rowCount: number; byteSize: number }> = [];
+    let deleted = 0;
+
+    // m3: the advisory-lock + PG-connection hold time scales with `chunkCount` —
+    // this scan reads every chunk inside the lock (O(chunkCount) S3 reads).
+    // Acceptable because deletes are rare; a hot path would need an id→chunk index.
+    // The transaction timeout is widened (see `withDatasetLock`) so this scan
+    // can't P2028 on a multi-chunk dataset.
+    // TODO(scan-before-lock): the locate-scan could run BEFORE the lock and only
```
Aryansharma28: **NB1 — O(chunkCount) S3 reads under lock (acknowledged, deferred)**

`deleteS3JsonlRecords` scans every chunk sequentially inside the advisory lock — O(chunkCount) S3 reads. At 500 chunks × ~5 ms/read that's ~2.5 s of lock hold per delete. The 120 s transaction timeout covers it and the TODO here is clear. Worth revisiting if high-churn datasets show contention — the scan-before-lock approach described in the comment is the right fix when you get to it.

Aryansharma28 (commented):
**Review** — fresh pass on the diff post all resolved P1/P2 items. No new blockers. Five non-blockers below as inline comments.

langwatch/src/server/datasets/dataset-normalize.job.ts
```diff
@@ -183,6 +236,28 @@ const buildRenameMap = (headers: string[]): Map<string, string> => {
...
+ * The CSV path parses with `header:false` and maps rows to objects by index
+ * (below) precisely to AVOID papaparse's header machinery: under our
+ * pause/resume backpressure it re-runs its duplicate-name dedup against the
+ * CURRENT data row on every resume, so any row holding two equal cells (an input
+ * that equals its expected output, or two empty cells) had its second value
+ * silently corrupted with a `_1` suffix — and logged a warning per row. Deduping
+ * the header ourselves keeps the legitimate "two columns named the same" rename
+ * without ever touching row values.
+ */
+const dedupeHeaders = (headers: string[]): string[] => {
```
Aryansharma28, line 261: **NB2 — `dedupeHeaders` can produce collisions with pre-existing `_N` names**

`seen` is keyed on the raw input header, not the emitted name. So:

```ts
dedupeHeaders(["col", "col", "col_1"])
// → ["col", "col_1", "col_1"]  ← collision
```

When a CSV has a duplicate column AND another column already named with a `_N` suffix, the dedup emits two identical headers. Downstream `record[header] = values[i]` silently overwrites the first `col_1` column's value with the second.

**Fix:** after building the output array, do a second pass — if the emitted name already exists elsewhere in the output, keep incrementing the counter:

```ts
const seen = new Map<string, number>();
const out: string[] = [];
for (const header of headers) {
  const base = header;
  let count = seen.get(base) ?? 0;
  let candidate = count === 0 ? base : `${base}_${count}`;
  while (out.includes(candidate)) { count++; candidate = `${base}_${count}`; }
  seen.set(base, count + 1);
  out.push(candidate);
}
return out;
```

Same edge case flagged in the stacked PR #5017 — fix in both places.

langwatch/src/server/datasets/dataset-normalize.job.ts
```diff
@@ -61,6 +64,16 @@ export const LARGE_JSON_MAX_BYTES = 100 * 1024 * 1024;
...
 
+/**
+ * Max bytes for a single CSV row (I-MEM), the CSV counterpart to
+ * `MAX_JSONL_LINE_BYTES`. papaparse buffers until it can emit a complete row, so
+ * a malformed CSV with no row delimiter (or one giant field) would make it
+ * accumulate the whole file in memory before the first `step`. We track the
+ * parser cursor delta between rows and `abort()` once a single row crosses this
+ * cap, failing the dataset rather than risking an OOM.
+ */
+export const MAX_CSV_ROW_BYTES = 8 * 1024 * 1024;
```
Aryansharma28, line 75: **NB3 — `MAX_CSV_ROW_BYTES` is used for two distinct purposes**

This constant serves as both (1) the papaparse I/O read-buffer size (`chunkSize` on line 339) and (2) the per-row payload cap check (line 359). The comment on the `chunkSize` usage says it "bounds" per-row memory — that's partially true (papaparse reads in 8 MB I/O chunks), but the real upstream size guard is `UPLOAD_MAX_BYTES`.

The dual-use name makes it look like increasing the per-row cap also increases the I/O buffer, and vice-versa. Consider splitting into:
```ts
const CSV_IO_CHUNK_BYTES = 8 * 1024 * 1024;  // papaparse read-buffer
export const MAX_CSV_ROW_BYTES = 8 * 1024 * 1024;  // per-row cap
```
Not a correctness bug — cosmetic only.

langwatch/src/server/datasets/dataset-mutations.ts
```diff
@@ -0,0 +1,583 @@
...
+ * PG-commit-after-S3-write failure on edit/delete, where the chunk is mutated
+ * but the counters rolled back (Consequences → Negative: I-COUNT is eventually
+ * consistent / repairable, not unconditionally atomic). `chunkCount` is trusted
+ * as the chunk-set boundary: every chunk in `0..chunkCount` MUST exist — a
+ * missing one is corruption, not emptiness, so `readChunk` throws
+ * `MissingChunkError` and we propagate it rather than silently dropping the tail
+ * (which would re-derive a smaller `chunkCount` and mask data loss when a middle
+ * chunk is gone but later chunks survive). Returns the recomputed counts.
+ */
+export const recomputeDatasetCounts = async ({
```
Aryansharma28, line 777: **NB4 — `recomputeDatasetCounts` does not trim trailing empty chunks**

If a delete mutates S3 (empties a trailing chunk) but the PG commit fails afterward, then `recomputeDatasetCounts` is run to repair. It re-derives `rowCount`/`sizeBytes`/`chunkOffsets` from actual chunk bytes — but it writes back `chunkCount = current.chunkCount` unchanged rather than re-trimming trailing empties.

So after a repair run, `chunkCount` can over-count empty trailing chunks that `deleteS3JsonlRecords` would have trimmed on a clean path. Functionally harmless (read paths stop iterating once they see an empty chunk), but the count doesn't self-heal and will accumulate across repeated repair runs on the same dataset.

**Suggestion:** at the end of `recomputeDatasetCounts`, trim `chunkCount` down to the highest non-empty index + 1 (same logic as the trim in `deleteS3JsonlRecords`), so repair is fully idempotent.

langwatch/src/server/datasets/dataset-normalize.queue.ts
```diff
@@ -62,6 +66,19 @@ export const enqueueDatasetNormalize = async (params: {
   prisma: PrismaClient;
   payload: DatasetNormalizePayload;
 }): Promise<void> => {
+  if (!registeredEnqueue && env.NODE_ENV === "production") {
```
Aryansharma28: **NB5 — Inline-normalize warning fires on every upload during rolling deploys**

`registeredEnqueue` is `null` on a fresh pod until the worker process calls `registerDatasetNormalizeEnqueue`. During any rolling deploy, new pods process uploads before the worker registers — so the warn fires on every upload for those pods even when the deployment is fully correct.

The warning is per-call, not per-process, so it can produce a burst of spurious `WARN` lines in production observability for several minutes on every deploy.

**Suggestion:** log this once per process lifetime using a module-level flag:
```ts
let hasWarnedNoQueue = false;
if (!registeredEnqueue && env.NODE_ENV === "production" && !hasWarnedNoQueue) {
  hasWarnedNoQueue = true;
  logger.warn(...);
}
```
Low priority — the warning is informational and the behaviour is correct — but worth suppressing the burst.

langwatch/src/server/datasets/dataset-normalize.job.ts
sergioestebance: Fixed in a24dfaa2b. `dedupeHeaders` now tracks the *emitted* names (not just raw inputs) and keeps bumping the counter until the candidate is unique, so `["col","col","col_1"]` → distinct names and the by-index record map can no longer overwrite a column. One location only — #5017 was squash-merged into this PR, so there is a single `dedupeHeaders` now.

langwatch/src/server/datasets/dataset-normalize.job.ts
sergioestebance: Fixed in a24dfaa2b. Split into `CSV_IO_CHUNK_BYTES` (papaparse read-buffer, used at `chunkSize`) and `MAX_CSV_ROW_BYTES` (per-row payload cap). No behaviour change — both still 8 MB, but now tunable independently.

langwatch/src/server/datasets/dataset-mutations.ts
sergioestebance: Fixed in a24dfaa2b. `recomputeDatasetCounts` now trims trailing empty chunks down to highest-non-empty + 1 — the same logical trim `deleteS3JsonlRecords` does — so the I-COUNT repair is idempotent across repeated runs. Trailing objects stay as benign 0-byte orphans (no in-tx delete; same lock-free-reader reasoning as the delete path).

langwatch/src/server/datasets/dataset-normalize.queue.ts
sergioestebance: Fixed in a24dfaa2b. Module-level `hasWarnedNoQueue` latch → warns once per process instead of per upload, so no log burst on an S3-yes/Redis-no deploy or during the registry-init window on rolling deploys. Comment notes the persistent-misconfig signal belongs in a metric/healthcheck, not a per-upload log.

langwatch/src/server/datasets/dataset-mutations.ts
sergioestebance: Agreed, leaving as-is. The `TODO(scan-before-lock)` stays in code as the tracked follow-up; deferral is correct (deletes are rare, widened txn timeout covers the O(chunkCount) hold). Leaving this thread open so the deferral stays visible.

sergioestebance: Resolved — implemented in the stacked PR #5067 (ADR-032 v16). The O(chunkCount) id-locate scan now runs **off** the advisory lock (`locateIdsBeforeLock`); under the lock the fast path re-reads **only the affected chunk(s)** and takes every unaffected chunk’s `(rowCount, byteSize)` from the authoritative offset index — shrinking lock-held S3 reads from O(chunkCount) to O(affected).

Correctness is preserved by re-validation under the lock: the fast path commits only when the pre-scan located every target id and the offset index covers every chunk, and **bails to the full in-lock scan** if any located id isn’t where the hint said (concurrent move/delete) — so even a future compaction that relocates rows can’t cause a silently-missed delete. The pre-scan is skipped for a not-ready dataset (no I/O ahead of the readiness gate). Covered by unit tests (fast-path read-reduction, hint-bail fallback for edit + delete, legacy no-offset → full scan). The `TODO(scan-before-lock)` comments are removed in that PR.

langwatch/src/server/api/routers/datasetRecord.utils.ts
```diff
@@ -177,6 +437,111 @@ export const getFullDataset = async ({
...
+        });
+      }
+    }
+
+    // "all": read chunks ONE AT A TIME (I-MEM), accumulating rows until the
+    // `limitMb` byte budget is reached, then STOP — never reading the remaining
+    // chunks. `count` stays PG-authoritative (it is NOT the number of rows
+    // actually read). Equivalent to the prior read-all-then-truncate, minus the
+    // unbounded read.
+    const chunkCount = dataset.chunkCount ?? 0;
```
Aryansharma28, line 559: 🔴 **Blocker — `chunkCount = null` loops 0 times, silently returns empty dataset**

When `dataset.chunkCount` is `null` (e.g., the PG `UPDATE` that sets `chunkCount` after a migration or normalize job failed transiently), `dataset.chunkCount ?? 0` gives `0`, the `for` loop executes zero iterations, and `getFullDataset` returns `{ datasetRecords: [], count: <positive rowCount>, truncated: false }`. The UI shows 0 rows against a correct total with no error — impossible to diagnose without checking the DB directly.

Suggested fix: throw (or log + throw) a `MissingChunkError` / `DatasetNotReadyError` when `contentLayout === 's3_jsonl'` and `chunkCount` is `null` rather than silently treating it as 0:
```ts
const chunkCount = dataset.chunkCount;
if (chunkCount == null) {
  throw new MissingChunkError(`Dataset ${datasetId} has s3_jsonl layout but null chunkCount — I-COUNT drift`);
}
```

langwatch/src/components/datasets/services/directUpload.ts
```diff
@@ -0,0 +1,241 @@
...
+ * fresh dataset regardless).
+ */
+export async function abortPendingUpload({
+  projectId,
+  datasetId,
+}: {
+  projectId: string;
+  datasetId: string;
+}): Promise<void> {
+  await fetch(
```
Aryansharma28: 🔴 **Blocker — `abortPendingUpload` swallows a failed DELETE, leaving an orphaned `uploading` row**

The JSDoc says "failures are swallowed (the fallback path creates a fresh dataset regardless)", but that's only safe if the abort succeeds. When the DELETE returns 5xx (DB timeout, pod restart), the `uploading` row stays in Postgres. The fallback then tries to create a new dataset with the same name, hits the `@@unique` constraint on the slug, and the user can never create that dataset name again without manual DB cleanup.

Suggested fix: check `response.ok` and surface the failure so the caller can decide:
```ts
const response = await fetch(`/api/dataset/direct-upload/${datasetId}?...`, { method: 'DELETE' });
if (!response.ok) {
  logger.warn({ datasetId }, 'abortPendingUpload: DELETE failed, uploading row may remain');
  // don't throw — let the caller proceed with a new dataset name
}
```
Alternatively, add a TTL sweep / cron that reaps `status='uploading'` rows older than N minutes.

langwatch/src/pages/[project]/datasets/[id].tsx
```diff
@@ -1,14 +1,37 @@
...
+      // its own stop once the status settles (mirrors useTraceFacets).
+      refetchInterval: (data) =>
+        data?.status === "processing" || data?.status === "uploading"
+          ? 3000
+          : false,
+    },
+  );
+
+  const status = datasetQuery.data?.status;
+  const isReady = status === "ready" || status == null;
```
Comment (Aryansharma28) [inline langwatch/src/pages/[project]/datasets/[id].tsx]: 🔴 **Blocker — `status == null` is `true` on initial render (before `datasetQuery` resolves), so `DatasetEditorTable` mounts immediately and fires `getAll` against a potentially-processing dataset**

On first render `status` is `undefined`; `undefined == null` is `true` with loose equality. This means `isReady=true` and the table mounts before we know the actual status. For a `status='processing'` dataset, `getAll` hits the server's `PRECONDITION_FAILED` guard, React Query retries 3×, and the user sees a cascade of error toasts before the status query resolves and unmounts the table.

Fix: gate on `datasetQuery.isSuccess` (or use strict equality) so the table only mounts once the status is known:
```ts
// Before:
const isReady = status === "ready" || status == null;
// After:
const isReady = datasetQuery.isSuccess && (status === "ready" || status == null);
```

langwatch/src/server/datasets/dataset.service.ts
```diff
@@ -210,6 +227,16 @@ export class DatasetService {
...
         JSON.stringify(columnTypes)
       ) {
+        // ADR-032: column migration rewrites every record's keys — for s3_jsonl
+        // that's a chunk-rewrite (write-mutation), which is a later rung. The PG
+        // record migrator would read zero rows (I-PG) and silently "migrate"
+        // nothing, leaving stored chunk keys out of sync with columnTypes.
+        // Refuse rather than corrupt. (Deferred: s3_jsonl column migration.)
+        if (existingDataset.contentLayout === "s3_jsonl") {
+          throw new Error(
+            "Changing column types is not yet supported for large (S3) datasets",
```
Aryansharma28: 🟡 **Non-blocker — plain `Error` on column-type change surfaces as HTTP 500; the actionable message is stripped in production**

tRPC's `errorFormatter` only enriches `DomainError`, `ModelNotConfiguredError`, `AiCallFailedError`, `ZodError`, and limit-typed errors. A plain `new Error(...)` falls through as `INTERNAL_SERVER_ERROR` — the UI shows a generic "Something went wrong" toast and the user has no way to know that column-type edits are blocked on `s3_jsonl` datasets.

Fix: use a typed error so `errorFormatter` can surface the message:
```ts
throw new TRPCError({
  code: 'BAD_REQUEST',
  message: 'Changing column types is not yet supported for large (S3) datasets',
});
```

langwatch/src/server/api/routers/datasetRecord.utils.ts
```diff
@@ -177,6 +437,111 @@ export const getFullDataset = async ({
...
+          count,
+          datasetRecords: [record],
+          truncated: false,
+        };
+      }
+      // `record === null`: no short-circuit possible (missing/legacy offsets).
+      // This rare defensive path reads the whole dataset to honour
+      // last/random/N correctly, then selects the single row. Bounded by the
+      // same in-memory read the rung accepts for legacy data without offsets.
+      const rows = await storage.readChunks({
```
Aryansharma28, line 526: 🟡 **Non-blocker (plausible OOM) — the no-offsets fallback calls `readChunks` (all chunks) even for single-row `entrySelection` values like `'first'` or `'last'`**

The comment acknowledges this is a "whole dataset read" accepted as a bounded cost for "legacy data without offsets". However, the bound only holds if `chunkOffsets` is guaranteed to be populated after migration. A large migrated dataset (or one with a corrupt offsets array) with `entrySelection='first'` will load the entire dataset into memory just to return one row, OOM-ing the pod.

Mitigation options:
1. Add a `sizeBytes` guard before the fallback (similar to the export guard at line ~497) — refuse if the dataset is too large to load safely.
2. For `entrySelection='first'` specifically, read only chunk 0 instead of all chunks.
3. Document explicitly in the ADR that `chunkOffsets=null` is considered an unsupported/corrupted state post-migration and should never occur in production.

Aryansharma28 (commented):
## Review — feat(datasets): S3 chunked JSONL (ADR-032)

Solid foundational PR — the storage abstraction, advisory-locking strategy, and born-on-storage cutover are well-designed. Three confirmed correctness bugs and one missing error-check need fixing before ship; two plausible findings flagged as non-blockers.

---

### 🔴 Blockers

**1. `chunkCount = null` silently returns 0 rows for all s3_jsonl datasets** (`datasetRecord.utils.ts:517`)
**2. `abortPendingUpload` swallows DELETE failure → orphaned `uploading` row → permanent slug conflict** (`directUpload.ts:189`)
**3. `isReady = status == null` is `true` on initial render — fires `getAll` against a processing dataset before status resolves** (`[id].tsx:34`)

---

### 🟡 Non-blockers

**4. `throw new Error(...)` on column-type change surfaces as 500 — the actionable message never reaches the UI** (`dataset.service.ts:237`)
**5. No-offsets fallback calls `readChunks` (all chunks) even for a single-row `entrySelection` — unguarded OOM on large legacy datasets** (`datasetRecord.utils.ts:478`)

0xdeafcafe:
Reviewed end-to-end. Three open P0s, then a long P1 tail. Most actionable items:

### P0s (need to land before merge)

1. **`isReady = status === "ready" || status == null`** (`[id].tsx:34`). On first render `status` is `undefined`, so the editor mounts immediately and fires `getAll` against a dataset that's still `processing`. You get a 425 + global error toast before the "Preparing" alert ever shows up. @Aryansharma28's blocker is right - gate on `!!datasetQuery.data && status === "ready"`.

2. **`dataset.chunkCount ?? 0` silently returns `[]`** (`datasetRecord.utils.ts:517`). If chunkCount is ever null on a real dataset, every reader (UI, SDK, experiments) sees an empty dataset with no signal. Throw `DatasetNotReadyError` when `contentLayout === "s3_jsonl"` and chunkCount is null.

3. **`abortPendingUpload` swallows DELETE 5xx** (`directUpload.ts:189`). A transient failure pins an `uploading` row in PG forever, counting against the project's quota. Either throw on non-2xx (so the modal can surface "couldn't cancel"), or pair with a server-side TTL reaper.

### P1 themes

- **CSRF on `POST /direct-upload`**: session-cookie + no CSRF token + no preflight = forgeable from a cross-origin form. An Origin/Referer check or a custom header would do it.
- **Memory invariants leak in three places**: `listRecords` no-offsets fallback reads all chunks per page; `copyDataset` materialises the whole source; the backfill task holds S3 PUTs inside the 120s advisory lock (the file header doc itself says writes are recoverable outside it).
- **UI doesn't honour `status="ready"` in three more places** beyond the P0: `DatasetPickerList` offers `processing`/`failed` datasets, `datasets.tsx` row menu shows Edit/Delete on non-ready rows, and the upload modal has no progress bar and no abort plumbing (closing it mid-PUT silently abandons the upload).
- **Helm**: the RWO-PVC gate only inspects `app.replicaCount` - `workers.replicaCount > 1` slips through and crashloops. Migration Job has no `activeDeadlineSeconds` and no `serviceAccountName` (so on EKS+IRSA it loses the role app/workers run as).
- **Multitenancy test gap**: the only proof multitenancy holds on the new upload routes is the unit test for `authorizeDirectUpload`. No integration test runs a foreign-project apiKey against the upload paths.

### P2 worth catching

- `dbMultiTenancyProtection.ts:535` now exempts *every* raw query in the codebase, not just the advisory-lock SQL.
- Layering: `dataset-mutations.ts` does `new DatasetRepository(prisma)` per call (DI seam exists, isn't used); `POST /:slug/entries` bypasses the service layer entirely; the tRPC router re-issues `prisma.dataset.findFirst` four times.
- `awsS3.accessKeyId.value` / `secretAccessKey.value` default to inline plaintext in `values.yaml` with no "secretKeyRef recommended" comment (other secrets in the chart carry it).

Already-fixed items from earlier rounds (StreamingChunkWriter memory regression, `dedupeHeaders` collision, scan-before-lock) noted - those are clean.

Big PR. I'd push for landing the three P0s and the UI-doesn't-honour-`ready` cleanup in a single follow-up pass before merge.

sergioestebance:
Review findings addressed in **#5076** (stacked onto this branch). Validated all of them against the code first; **all 15 fixed**.

**P0 blockers:** null-`chunkCount` → typed error (was silently serving an empty dataset); `isReady` gated on `isSuccess` (was mounting the editor against a processing dataset); `abortPendingUpload` throws on a failed DELETE (was pinning an `uploading` row).

**P1:** the three unbounded `readChunks` reads are heap-guarded; backfill writes moved outside the advisory lock (matching the file's own documented ordering); CSRF cross-site check on the cookie-authed direct-upload routes; picker + row-menu + `updateExistingDataset` now honour `status='ready'`; Helm RWO guard covers `workers.replicaCount` + migration Job gets a `serviceAccountName`/deadline; route-level cross-tenant test for `/direct-upload`.

**P2:** `/:slug/entries` routed through the service; create-path double `findFirst` deduped; `values.yaml` awsS3 secretKeyRef note; `DatasetRepository` injected into the s3_jsonl mutations (was `new`-per-call).

Already-fixed callouts (`dedupeHeaders` collision, StreamingChunkWriter, scan-before-lock) confirmed clean. Unit tests added across every behavioural change + the cross-tenant integration test runs in CI; full typecheck green.

sergioestebance:
Thanks @Aryansharma28 (and the automated pass) — every blocking and non-blocking finding is now fixed, landed via #5076 (squash-merged into this branch as `28ccd339f`). Mapping:

### Blockers
1. **`chunkCount = null` silently returns 0 rows** — fixed. All four s3_jsonl read paths now throw `DatasetChunkCountMissingError` instead of coercing `?? 0`: `getFullDataset`, `listRecords` (no-offset fallback), `copyDataset`, and the head/preview. Each has a regression test asserting the throw and that no chunk read happens.
2. **`abortPendingUpload` swallows the DELETE failure → orphaned `uploading` row → permanent slug conflict** — fixed. The client now throws on a non-OK DELETE (callers log it), and a poll-triggered sweep (`reapStalePendingUploads`, runs on the next upload — no new scheduler) archives stale `uploading` rows and reaps their staging objects. Durable backstop is an S3 `staging/` lifecycle rule (infra, in flight).
3. **`isReady = status == null` true on initial render** — fixed. The gate is now `datasetQuery.isSuccess && (status === "ready" || status == null)`, so `getAll` never fires before the query resolves.

### Non-blockers
4. **column-type change surfaces as 500** — fixed. Now a typed `ColumnTypeChangeNotSupportedError` mapped to 400, so the actionable message reaches the UI.
5. **no-offsets fallback materializes all chunks — unguarded OOM** — fixed. `assertDatasetReadableInHeap` (100 MB ceiling) now guards every full-materialization path: the paginated no-offset fallback, full export, and copy. Streaming normalize was already bounded.

### Automated review — memory P1/P2s
"Several paths retain/materialize the full dataset in memory" — closed by the same heap-ceiling guard above; the only unbounded reads were the offset-less fallbacks, now gated.

### Deliberately deferred (tracked, not dropped)
- **P1 — backfill lost-write race.** A `DatasetRecord` write concurrent with a migration can be lost (the postgres mutation paths don't take the per-dataset advisory lock; the migrate snapshots outside it). It's a **pre-migration** blocker, not pre-merge — the backfill isn't running yet — and the fix is cross-cutting (every postgres mutation path joins the lock + ready-gate; migration marks the row `migrating` before snapshotting). Own change, tracked.
- **P1 — persist `storageBackend` per dataset.** `getDatasetStorage` re-resolves the backend from env on every read, so a local-FS→S3 migration orphans existing datasets. Needs a Prisma migration → own PR, tracked.

CI is green on the merged commit. Ready for a re-review when you have a moment 🙏

Aryansharma28 (approved):

sergioestebance:
## Manual QA — browser direct-upload → S3 dev (dev stack)

Drove the real upload UI (Playwright + Chromium, UI login, dev stack on `:5560`, dev S3 = `runtime-storage-dev`) and validated against S3 + Postgres ground truth.

**Flow:** opened the **Upload or Create Dataset** drawer → attached a 100 MB CSV → clicked Upload.

**Result — end-to-end PASS:**
- Direct-upload minted an **S3 presigned PUT** (`uploadUrl` host `s3.eu-central-1.amazonaws.com`) — not the local fallback.
- Presigned **PUT → 200** (~8s), **finalize → 200**.
- Normalize wrote **21 chunk objects (~162 MB)** under `datasets/<project>/<datasetId>/`, valid `{id, entry}` JSONL (CSV parsed correctly).
- Postgres row consistent end-to-end: `status=ready`, `contentLayout=s3_jsonl`, `rowCount=594000`, **`chunkCount=21`** (= the 21 S3 objects), `sizeBytes≈162 MB` (= S3 total).

So: a CSV uploaded through the browser lands in S3 dev as chunked JSONL and reads back ready with matching counts.

**Not covered:** the 2 GB variant couldn't be driven through **headless** Chromium — the renderer closes the instant the multi-GB `fetch` PUT body is constructed (a headless-harness memory limit; the server presigned to S3 correctly each time, zero bytes transferred before the close). Would need headed Chrome to validate that size.

**Incidental:** pre-existing console hydration warning on the datasets list (`<a> cannot contain a nested <a>`) — unrelated to upload.

PR #4910, opened 2026-06-18
feat(nlpgo): fetch remote attachment URLs and deliver them to the model as content

rogeriochaves:
## What

Prompt messages can now reference an attachment by a plain `http(s)` URL and have the model actually receive it as content. Today an image only reaches the model when its value is an inline `data:image/...;base64,...` URL (the existing splitter handles that); a plain URL is interpolated into the message as text and the model never opens it.

## Why

Reported from dogfood: vision prompts that point at an image URL (an S3/CDN link, a dataset `image` cell that holds a URL) silently degrade. The model receives the link as a string and guesses, instead of seeing the picture.

## How

A fetch pass runs in the nlpgo engine right after the existing data-URL image split (`engine.go`, beside `splitMessagesWithImages`):

- **Fetch** every `http(s)` URL referenced in a message, in string content, in text parts, and carried by an `image_url` part, under the **HTTP block's SSRF policy** (private, loopback, and cloud-metadata addresses refused, with DNS-rebinding protection via `SafeDialer`), a wall-clock timeout, and a size cap.
- **Detect the type from the response**, not the file extension (Content-Type header, falling back to content sniffing), so extension-less S3/CDN URLs classify correctly.
- **Structure by type**: images inline as a base64 data-URL `image_url` part so they work across every provider regardless of whether the provider fetches URLs itself; audio becomes an `input_audio` part; PDFs become a `file` part.
- **A reachable non-attachment** (an HTML page) is left as text: the author was referencing a link, not attaching a file.
- **A URL that cannot be fetched** (unreachable, error status, oversized, SSRF-blocked) aborts the run with a clear `attachment_fetch_error` that names the URL and the reason, surfaced to the user on the same path as `llm_error`. No broken request is sent to the provider.
- **Image-typed inputs are explicit attachments.** A signature input declared as the studio's `image` type is resolved before message templating: its URL is fetched, validated as an image, and inlined so the model receives a picture, and an unfetchable or non-image URL aborts the run with the same clear error rather than being templated into the prompt as text.

### Scope / tradeoffs to flag

- **Fetch-by-intent (breadth):** intent decides strictness. An `image`-typed signature input (the studio's "Image" field type) and an explicit `image_url` part both hard-fail on a bad fetch, because the author declared the value an attachment. A bare URL in prose is best-effort: it is fetched and attached when the response is media, and left as text when it is a normal web page **or** when the fetch fails, so an incidental link can never fail the whole run. Image-typed inputs are resolved before message templating (fetched, validated as an image, inlined as a base64 data URL for the existing splitter); a str-typed input carrying a URL stays best-effort.
- **SSRF:** reuses the HTTP block's `CheckURL` + `SafeDialer` and the same `ALLOWED_PROXY_HOSTS` allow-list, so attachment fetches are bound by the same destination policy as HTTP blocks, including the dial-time re-check that blocks redirect-to-private.
- **Cross-provider:** an image, once inlined as a base64 data-URL `image_url` part, is byte-identical to the already-shipping base64 image path. The new `input_audio` and `file` (PDF) part shapes are verified end to end through the in-process gateway to OpenAI by the live probes above (it forwards content-part arrays generically), so they ride the same proven downstream. Non-OpenAI providers reuse that generic forwarding; the part shapes follow OpenAI's `input_audio` / `file` conventions, and provider-specific translation for audio/PDF on other vendors can be tightened if a customer needs it.
- **Sequential fetches:** multiple attachments in one message are fetched sequentially; fine for the typical 0–1 attachment case, parallelizable as a follow-up if multi-attachment prompts appear.

## Testing

BDD scenarios in `specs/nlp-go/multimodal-attachment-urls.feature`, each bound to a Go test in `attachment_test.go` that fetches over a real `httptest` server: http image fetch, type-from-response (header + sniffed), data-URL no-fetch passthrough, multiple URLs, unreachable, error status, oversized, redirect-to-private, HTML-left-as-text, broken-prose-link-as-text, audio, PDF, and the image-typed-input cases (resolved-and-inlined, hard-fail on unreachable, hard-fail on non-image). Full nlpgo `app/` + `cmd/` suites green, `go vet` clean, `golangci-lint` 0 issues, feature-parity bound.

## Dogfood

**Known-answer vision probe through the real gateway** (`workflow_llm_vision_url_e2e_test.go`, `live_openai` tag): a solid-blue PNG is served from an opaque local URL (no filename hint), a vision prompt runs through nlpgo to a real `gpt-5-mini`, and the model is asked the dominant color. It answered **blue** — only knowable by actually seeing the fetched pixels, so this proves the fetch → inline → gateway → vision path end to end and catches hallucination. The test doubles as a permanent (gated) regression.

```
--- PASS: TestSync_RealWorkflowEndToEnd_OpenAI_VisionAttachmentURL (1.56s)
```

**Audio and PDF through the same gateway** (`workflow_llm_audio_url_e2e_test.go` and `workflow_llm_pdf_url_e2e_test.go`, `live_openai`, fixtures supplied via env so no binaries are committed). Both confirm the in-process gateway forwards the new content-part shapes, with the same hallucination-proof design:

- **Audio:** a clip that asks aloud "Do I have a male voice or a female voice?" (Whisper-confirmed) is served from an opaque URL; nlpgo structures an `input_audio` part and `openai/gpt-audio` echoes the spoken question, which it can only do by hearing the fetched clip.
- **PDF:** a document containing "order number 4729" and "customer city Lisbon" is served; nlpgo structures a `file` part and `openai/gpt-5-mini` answers with both facts, which live only inside the fetched PDF.

```
--- PASS: TestSync_RealWorkflowEndToEnd_OpenAI_AudioAttachmentURL (2.00s)
--- PASS: TestSync_RealWorkflowEndToEnd_OpenAI_PDFAttachmentURL (4.07s)
```

**End-to-end in the Optimization Studio** (real browser → app → nlpgo → gateway → real model). A workflow points a vision prompt at a remote image URL (`https://picsum.photos/id/237/...`, a black puppy — the URL gives nothing away):

- A plain **Text** input holding the URL is fetched best-effort and the model answers **dog**: text url answer
- Switching the input to the studio's **Image** field type also resolves to **dog**: image type answer
- **Unhappy path, before the fix in this PR:** an **Image**-typed field with an unreachable URL silently degraded — the model guessed **cat** from the `cat.png` filename instead of erroring: before
- **After the fix:** the same unreachable image URL now fails fast with a clear error and no model call: **"Could not load the attachment ...: it could not be reached."**: after

## Deployment Impact

- **New outbound network behavior:** nlpgo now makes server-side HTTP(S) GETs to fetch attachment URLs referenced in prompts. This is bound by the existing HTTP-block SSRF policy (private/loopback/metadata refused, dial-time re-check for redirects) and the same `ALLOWED_PROXY_HOSTS` allow-list, with a 30s timeout and a 20 MB size cap. No new egress is required for normal operation, but operators running nlpgo in a restricted-egress environment should know attachment fetching needs outbound access to wherever user attachments live.
- **No new config, env vars, services, migrations, or charts.** Reuses `ALLOWED_PROXY_HOSTS`.
- **No data-plane or schema changes.** Behavior is additive: existing `data:` URL handling is unchanged.

### DIFF
--- a/services/nlpgo/app/engine/attachment.go
+++ b/services/nlpgo/app/engine/attachment.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/langwatch/langwatch/services/nlpgo/app"
 	"github.com/langwatch/langwatch/services/nlpgo/app/engine/blocks/httpblock"
+	"github.com/langwatch/langwatch/services/nlpgo/app/engine/dsl"
 )
 
 // Attachment fetching turns a remote URL referenced in a prompt message into
@@ -148,6 +149,50 @@ func (f *attachmentFetcher) rewrite(ctx context.Context, messages []app.ChatMess
 		}
 		m.Content = newContent
 		out = append(out, m)
+	}
+	return out, nil
+}
+
+// inlineImageInputs resolves image-typed inputs that carry a remote http(s)
+// URL into inline base64 data URLs before message templating, so the existing
+// data-URL image splitter delivers them as image parts. An image-typed input
+// is an explicit attachment: the author declared the field an image, so a URL
+// it carries that cannot be fetched as an image fails the run with a clear,
+// user-facing error rather than being left as text for the model to guess from
+// (e.g. from a filename). Inputs that are not image-typed, not http(s) URLs, or
+// already inline data URLs are left untouched. The returned map is a copy only
+// when a value was replaced, so the caller's original inputs (surfaced verbatim
+// in execution events) keep the readable URL rather than a base64 blob.
+func (f *attachmentFetcher) inlineImageInputs(ctx context.Context, node *dsl.Node, inputs map[string]any) (map[string]any, *NodeError) {
+	out := inputs
+	copied := false
+	for _, field := range node.Data.Inputs {
+		if field.Type != dsl.FieldTypeImage {
+			continue
+		}
+		raw, ok := inputs[field.Identifier].(string)
+		if !ok {
+			continue
+		}
+		rawURL := strings.TrimSpace(raw)
+		if !strings.HasPrefix(strings.ToLower(rawURL), "http") {
+			continue // already a data URL or not a remote reference
+		}
+		att, ne := f.fetch(ctx, rawURL)
+		if ne != nil {
+			return nil, ne
+		}
+		if !strings.HasPrefix(att.mediaType, "image/") {
+			return nil, attachmentError(rawURL, "could not be loaded as an image (its content type is "+att.mediaType+")", 0)
+		}
+		if !copied {
+			out = make(map[string]any, len(inputs))
+			for k, v := range inputs {
+				out[k] = v
+			}
+			copied = true
+		}
+		out[field.Identifier] = dataURL(att)
 	}
 	return out, nil
 }
--- a/services/nlpgo/app/engine/attachment.go
+++ b/services/nlpgo/app/engine/attachment.go
@@ -1,0 +1,297 @@
+package engine
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io"
+	"mime"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/langwatch/langwatch/services/nlpgo/app"
+	"github.com/langwatch/langwatch/services/nlpgo/app/engine/blocks/httpblock"
+)
+
+// Attachment fetching turns a remote URL referenced in a prompt message into
+// content the model can actually see/hear/read. The data-URL image splitter in
+// multimodal.go only handles inline data:image/...;base64 values; a plain
+// http(s) URL is interpolated as text and the model never opens it. This pass
+// fetches such URLs, detects the type from the response (not the file
+// extension, so extension-less S3/CDN URLs work), and re-homes them into
+// multimodal content parts. Images become base64 data-URL image parts so they
+// work across every provider regardless of whether the provider fetches URLs
+// itself; audio and documents get their own part shapes. A reachable response
+// that is not an attachment (an HTML page) is left as text — the author was
+// referencing a link, not attaching a file. A URL that cannot be fetched fails
+// the run with a clear, user-facing error rather than a broken request.
+
+const (
+	// defaultMaxAttachmentBytes caps a single fetched attachment. Large enough
+	// for real photos/audio/PDFs, small enough to refuse a runaway download.
+	defaultMaxAttachmentBytes int64 = 20 * 1024 * 1024
+	// defaultAttachmentTimeout bounds the whole fetch (connect + read).
+	defaultAttachmentTimeout = 30 * time.Second
+)
+
+// httpURLRe matches an http(s) URL token: the scheme followed by a run of
+// non-space characters that are not URL-hostile delimiters. Trailing
+// sentence punctuation is trimmed separately so "see https://x/cat.png." does
+// not carry the period into the request.
+var httpURLRe = regexp.MustCompile("https?://[^\\s<>\"'`]+")
+
+// attachmentFetcher fetches remote attachment URLs and rewrites them into
+// content parts. It applies the same SSRF policy as the HTTP block (private,
+// loopback, and cloud-metadata addresses are refused, with DNS-rebinding
+// protection via SafeDialer), a wall-clock timeout, and a size cap.
+type attachmentFetcher struct {
+	client   *http.Client
+	ssrf     httpblock.SSRFOptions
+	maxBytes int64
+}
+
+func newAttachmentFetcher(ssrf httpblock.SSRFOptions) *attachmentFetcher {
+	return &attachmentFetcher{
+		client: &http.Client{
+			Timeout:   defaultAttachmentTimeout,
+			Transport: &http.Transport{DialContext: httpblock.SafeDialer(ssrf)},
+		},
+		ssrf:     ssrf,
+		maxBytes: defaultMaxAttachmentBytes,
+	}
+}
+
+// fetchedAttachment is the validated result of fetching a URL.
+type fetchedAttachment struct {
+	mediaType string // normalized, lowercase, e.g. "image/png"
+	data      []byte
+}
+
+// fetch retrieves rawURL under the SSRF, timeout, and size guards. It returns a
+// *NodeError describing a user-facing failure when the URL cannot be fetched.
+func (f *attachmentFetcher) fetch(ctx context.Context, rawURL string) (*fetchedAttachment, *NodeError) {
+	if err := httpblock.CheckURL(rawURL, f.ssrf); err != nil {
+		return nil, attachmentError(rawURL, "is blocked for security (it resolves to a private or metadata address)", 0)
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, rawURL, nil)
+	if err != nil {
+		return nil, attachmentError(rawURL, "is not a valid URL", 0)
+	}
+	resp, err := f.client.Do(req)
+	if err != nil {
+		if errors.Is(err, httpblock.ErrSSRFBlocked) {
+			return nil, attachmentError(rawURL, "is blocked for security (it resolves to a private or metadata address)", 0)
+		}
+		return nil, attachmentError(rawURL, "could not be reached ("+err.Error()+")", 0)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode >= 400 {
+		return nil, attachmentError(rawURL, fmt.Sprintf("returned an error status (%d)", resp.StatusCode), resp.StatusCode)
+	}
+	// Read one byte past the cap so an oversized body is detectable.
+	body, err := io.ReadAll(io.LimitReader(resp.Body, f.maxBytes+1))
+	if err != nil {
+		return nil, attachmentError(rawURL, "could not be read ("+err.Error()+")", 0)
+	}
+	if int64(len(body)) > f.maxBytes {
+		return nil, attachmentError(rawURL, fmt.Sprintf("is larger than the %d MB attachment limit", f.maxBytes/(1024*1024)), 0)
+	}
+	mt := normalizeMediaType(resp.Header.Get("Content-Type"))
+	// Sniff from the bytes when the server omits or generalizes the type, so
+	// extension-less and octet-stream responses still classify correctly.
+	if mt == "" || mt == "application/octet-stream" {
+		mt = normalizeMediaType(http.DetectContentType(body))
+	}
+	return &fetchedAttachment{mediaType: mt, data: body}, nil
+}
+
+// rewrite fetches remote attachment URLs in every message and re-homes them
+// into content parts. It runs after splitMessagesWithImages, so inline data:
+// URLs are already parts; this pass handles http(s) URLs in string content, in
+// text parts, and carried by existing image_url parts.
+func (f *attachmentFetcher) rewrite(ctx context.Context, messages []app.ChatMessage) ([]app.ChatMessage, *NodeError) {
+	out := make([]app.ChatMessage, 0, len(messages))
+	for _, m := range messages {
+		switch content := m.Content.(type) {
+		case string:
+			parts, replaced, ne := f.splitStringAttachments(ctx, content)
+			if ne != nil {
+				return nil, ne
+			}
+			if replaced {
+				m.Content = parts
+			}
+		case []any:
+			parts, ne := f.rewriteParts(ctx, content)
+			if ne != nil {
+				return nil, ne
+			}
+			m.Content = parts
+		}
+		out = append(out, m)
+	}
+	return out, nil
+}
+
+// splitStringAttachments scans text for http(s) URLs and, when any resolve to a
+// real attachment, returns the text rewritten as a content-part list. The
+// second return is false (and parts nil) when there is nothing to attach, so
+// the caller keeps the original string. A failed fetch returns a *NodeError.
+func (f *attachmentFetcher) splitStringAttachments(ctx context.Context, text string) ([]any, bool, *NodeError) {
+	locs := httpURLRe.FindAllStringIndex(text, -1)
+	if len(locs) == 0 {
+		return nil, false, nil
+	}
+	parts := make([]any, 0, len(locs)*2+1)
+	addText := func(seg string) {
+		if strings.TrimSpace(seg) == "" {
+			return
+		}
+		parts = append(parts, map[string]any{"type": "text", "text": seg})
+	}
+	last := 0
+	attached := false
+	for _, loc := range locs {
+		rawURL, trailing := trimTrailingPunct(text[loc[0]:loc[1]])
+		att, ne := f.fetch(ctx, rawURL)
+		if ne != nil {
+			return nil, false, ne
+		}
+		part, ok := contentPartForAttachment(att)
+		addText(text[last:loc[0]])
+		if ok {
+			parts = append(parts, part)
+			addText(trailing)
+			attached = true
+		} else {
+			// Reachable but not an attachment (e.g. an HTML page): the author
+			// is referencing a link, so keep the URL verbatim as text.
+			addText(rawURL + trailing)
+		}
+		last = loc[1]
+	}
+	addText(text[last:])
+	if !attached {
+		return nil, false, nil
+	}
+	return parts, true, nil
+}
+
+// rewriteParts walks an existing content-part list, fetching http(s) URLs found
+// in text parts and in image_url parts.
+func (f *attachmentFetcher) rewriteParts(ctx context.Context, in []any) ([]any, *NodeError) {
+	out := make([]any, 0, len(in))
+	for _, p := range in {
+		block, ok := p.(map[string]any)
+		if !ok {
+			out = append(out, p)
+			continue
+		}
+		switch block["type"] {
+		case "text":
+			t, _ := block["text"].(string)
+			sub, replaced, ne := f.splitStringAttachments(ctx, t)
+			if ne != nil {
+				return nil, ne
+			}
+			if replaced {
+				out = append(out, sub...)
+			} else {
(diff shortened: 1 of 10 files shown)


services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+		parts = append(parts, map[string]any{"type": "text", "text": seg})
+	}
+	last := 0
+	attached := false
+	for _, loc := range locs {
+		rawURL, trailing := trimTrailingPunct(text[loc[0]:loc[1]])
+		att, ne := f.fetch(ctx, rawURL)
+		if ne != nil {
+			return nil, false, ne
```
rogeriochaves: P1 / policy (the main thing to decide). This hard-fails the whole run on ANY fetch error for ANY http(s) URL in free text, including URLs the author is only mentioning, not attaching. A prompt like "Summarize the article at https://example.com/old-page" where that page now 404s, times out, or exceeds the size cap aborts the run with attachment_fetch_error, even though nothing was meant to be attached.

The image_url part path (line 233) is genuinely different: an image_url part is explicit attachment intent, so hard-failing there matches the "fail clearly when an attachment cannot be fetched" ask.

Suggestion: split the policy by intent. In splitStringAttachments (bare URLs in prose), treat a fetch failure as best-effort and leave the URL as text, the same as the reachable-but-HTML case, rather than aborting the run. Keep the hard attachment_fetch_error for the image_url-part path where the user explicitly attached something. That preserves clear failures for real attachments without turning every incidental link into a run-killer. This is the failure-mode half of the breadth call you already flagged for follow-up.

services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,325 @@
...
+		return map[string]any{
+			"type":      "image_url",
+			"image_url": map[string]any{"url": dataURL(att)},
+		}, true
+	case strings.HasPrefix(att.mediaType, "audio/"):
+		return map[string]any{
+			"type": "input_audio",
+			"input_audio": map[string]any{
+				"data":   base64.StdEncoding.EncodeToString(att.data),
+				"format": strings.TrimPrefix(att.mediaType, "audio/"),
```
rogeriochaves: P2 / correctness. TrimPrefix(att.mediaType, "audio/") yields the MIME subtype, not the provider format enum. The standard MIME for MP3 is audio/mpeg, so this produces format: "mpeg", but OpenAI input_audio.format accepts "mp3" / "wav" and rejects "mpeg". http.DetectContentType also returns audio/wave for WAV, giving "wave" not "wav". attachment_test.go asserts format == "mpeg", which locks in the value the provider rejects.

Suggestion: map media type to the provider format explicitly, e.g. audio/mpeg and audio/mp3 -> mp3; audio/wav, audio/wave, audio/x-wav -> wav; only fall back to the trimmed subtype for types you have verified. Feeds directly into the area-4 Bifrost question.

rogeriochaves (commented):
Reviewed at HEAD 4cc1ea3, focused on the four flagged areas. Clean and well-tested; the four you called out are exactly the ones that matter. Priority findings inline, broader notes here.

**SSRF reuse (area 2): solid.** CheckURL is the pre-flight gate and SafeDialer re-runs the policy at dial time and dials the validated IP, which closes the DNS-rebinding TOCTOU and also covers redirect targets (the Transport is reused across redirects, so a 30x to a private/metadata IP is caught at the redirect dial). Metadata is always blocked, and an empty allow-list still permits public hosts, so the default ALLOWED_PROXY_HOSTS="" does not disable attachments. One coverage gap: nothing tests the redirect path. Worth an httptest server that 302s to 127.0.0.1 / 169.254.169.254 asserting attachment_fetch_error (it should already pass via SafeDialer, the test just locks it in).

**Part shapes (area 4, your open item):** the file / input_audio shapes are OpenAI-native. For non-OpenAI through Bifrost, the audio format enum (see inline P2) and the file.file_data data URL are exactly what Bifrost has to translate, so verifying on Anthropic/Gemini is the key dogfood.

**Minor**
- The empty-allow-list test comment ("any real fetch would be SSRF-blocked") is misleading: an empty allow-list permits public hosts, and that test passes only because it uses a data: URL (no fetch happens). Reword so it does not imply the default config blocks everything.
- attachment.pdf filename is hardcoded; deriving it from the URL path gives the model a meaningful name (minor).
- Fetches are sequential, so a message with several attachments serializes N x up-to-30s. Fine for now, parallelize later if it matters.
- fetch() embeds err.Error() in the user-facing message ("could not be reached (...)"), which leaks raw Go transport errors (e.g. "dial tcp 1.2.3.4:443: connect: connection refused"). Consider cleaner phrasing for the common cases and keep the raw error in logs only.

**System-prompt re-home (the 4cc1ea3 delta):** correct, and it nicely reuses splitLeadingText from multimodal.go to match the data-URL path. Edge case: instructions placed after the image in a system message ride into the user message too, but that is consistent with the existing data-URL re-home, so no objection.

Nice work overall.

rogeriochaves:
Re-checked at d9fd9a85, all four addressed cleanly:

- **Breadth split by intent**: `splitStringAttachments` is now best-effort, a failed or non-media bare URL stays as text and never fails the run, while the `image_url` part path still hard-fails on a bad fetch. Exactly the intent split, and the text-segment ordering refactor (adding the preceding text before the fetch) is correct for the continue path.
- **Audio format**: `audioFormat()` maps `audio/mpeg` and `audio/mp3` to `mp3`, and `audio/wav` / `audio/wave` / `audio/x-wav` to `wav`; the test now asserts `mp3`.
- **SSRF**: `TestRewriteRefusesRedirectToPrivateAddress` exercises SafeDialer at dial time, proving a redirect to `169.254.169.254` is refused even from an allowed host.
- **Minors**: pdf filename derived via `path.Base`, raw net error dropped from the user-facing message, misleading empty-allow-list comment fixed.

Deferring sequential fetches (fine for the typical 0-1 attachment case) and the field-type-awareness item for rchaves, both with notes, is the right call. LGTM from my side. Nice work.

rogeriochaves:
Addressed the review in d9fd9a850:

- **Breadth (fetch-by-intent):** bare prose URLs are now best-effort (reachable-non-media or failed-fetch leaves them as text, never failing the run); only an explicit `image_url` part hard-fails on a bad fetch. Added a best-effort scenario + test.
- **Audio format:** `audio/mpeg` → `mp3`, `audio/wav` → `wav` via a format map; test corrected.
- **SSRF:** added a redirect-to-metadata test exercising `SafeDialer`'s dial-time re-check.
- **Minors:** PDF filename derived from the URL; raw network error dropped from the user-facing message; misleading empty-allow-list test comment fixed.

Deferred (noted in the PR body): sequential fetches (fine for the typical 0–1 attachment case), and field-type-awareness so a templated `image`-typed field value hard-fails like an explicit attachment (needs sign-off on the policy). All green locally: 13 rewrite tests + the live vision probe, feature-parity 1992, vet clean.

rogeriochaves:
Re-checked `02d25cd5a` (the field-type-awareness follow-up). Clean, and it closes the silent-degradation gap rather than papering over it.

- `inlineImageInputs` runs **before** templating and is correctly scoped: image-typed inputs only, reuses `f.fetch` (same SSRF / size / timeout guards, no new SSRF surface), and hard-fails on **both** an unfetchable URL and a non-image content-type. So an image field pointed at a dead link or an HTML page now errors instead of letting the model guess from the filename.
- Copy-on-write keeps the caller's original `inputs` (the readable URL) for the execution events, while only the templated value is inlined as a data URL. Nice touch.
- `str`-typed inputs stay best-effort via the existing `rewrite` pass, and a data-URL or non-http image-typed input is left untouched. No double-handling between the inline pass and the later http rewrite, since the image value is already a data URL by the time templating runs.
- Coverage matches the boundary: resolves-to-data-URL, unreachable-fails, not-an-image-fails, and untouched-data-URL-and-text.

LGTM. Good end-to-end dogfooding to surface it.

services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,297 @@
...
+			} else {
+				out = append(out, block)
+			}
+		case "image_url":
+			img, _ := block["image_url"].(map[string]any)
+			url, _ := img["url"].(string)
+			if !strings.HasPrefix(strings.ToLower(url), "http") {
+				out = append(out, block) // already a data URL, leave as-is
+				continue
+			}
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 3e5f4a2aa: added an isHTTPURL helper that requires the http:// or https:// scheme separator, applied to both the image_url part path and the new image-typed input path, so httpfoo:// is no longer treated as fetchable.

services/nlpgo/app/engine/attachment_test.go
```diff
@@ -0,0 +1,274 @@
...
+	defer srv.Close()
+	f := loopbackFetcher()
+
+	out, ne := f.rewrite(context.Background(), []app.ChatMessage{
+		{Role: "system", Content: "You are a vision grader. Image: " + srv.URL + "/cat.png"},
+	})
+	require.Nil(t, ne)
+	require.Len(t, out, 2, "the system message must split into system text + a user message")
+	assert.Equal(t, "system", out[0].Role)
+	assert.IsType(t, "", out[0].Content, "the system message keeps only the leading text")
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 3e5f4a2aa: the re-home test now asserts the retained system text equals the leading instructions (TrimSpace equality), not just its type.

specs/nlp-go/multimodal-attachment-urls.feature
```diff
@@ -0,0 +1,115 @@
...
+    When the engine builds the LLM messages
+    Then the content carries an image part for each, in their original positions
+
+  @integration
+  Scenario: An attachment URL in the system prompt is re-homed to a user message
+    Given the system prompt references an image by URL
+    When the engine builds the LLM messages
+    Then the system message keeps only the text before the image
+    And the image is delivered in a user message instead
+    # Providers reject image parts in system-role messages.
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 3e5f4a2aa and 695b79aea: the scenario is renamed to "An image mentioned in the system prompt still reaches the model" and rephrased in user-facing terms (I run the workflow / the model still receives the image).

specs/nlp-go/multimodal-attachment-urls.feature
```diff
@@ -0,0 +1,161 @@
...
+  # The studio lets an author declare an input as an Image. That is an explicit
+  # statement of intent: the value is a picture, not prose that happens to carry
+  # a link. So an image-typed field is resolved before templating and fails the
+  # run with a clear error when its URL cannot be loaded as an image, rather than
+  # being passed to the model as text for it to guess from (e.g. from a filename).
+
+  @integration
+  Scenario: An image-typed field whose URL is an image is fetched and inlined
+    Given a signature input declared as an image whose value is an http URL to a PNG
+    When the engine builds the LLM messages
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 695b79aea: rewrote every scenario step across the file to user-facing language (When I run the workflow / Then the model receives the image), removing the engine-mechanics phrasing. Scenario titles are unchanged so the Go bindings stay intact.

services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,403 @@
...
+	return token[:i], token[i:]
+}
+
+// attachmentError builds the user-facing NodeError for a failed attachment
+// fetch. Status threads the upstream HTTP status when there was one (for fault
+// attribution), mirroring the llm_error path.
+func attachmentError(url, reason string, status int) *NodeError {
+	return &NodeError{
+		Type:    "attachment_fetch_error",
+		Message: fmt.Sprintf("Could not load the attachment %s: it %s.", url, reason),
```
langwatch-agent: **P1** — This user-facing `NodeError` includes the raw attachment URL, including query parameters. The PR explicitly targets S3/CDN image URLs, and presigned S3 URLs carry `X-Amz-Credential`, `X-Amz-Signature`, security tokens, etc. in the query string; on any fetch failure this stores and displays those credentials in the node error/trace. Keep enough context to identify the attachment, but redact at least `RawQuery`/fragment before building the message, and add a regression with a signed-looking URL proving the secret params do not appear in the error.

services/nlpgo/app/engine/engine.go
```diff
@@ -509,9 +514,34 @@ func (e *Engine) runSignature(ctx context.Context, execReq ExecuteRequest, node
...
+	// Fetch any remote attachment URLs (http/https) referenced in the messages
+	// and deliver them as content the model can open. A failed fetch aborts the
+	// run with a clear, user-facing error instead of a broken provider request.
+	if e.attachments != nil {
+		fetched, aerr := e.attachments.rewrite(ctx, messages)
+		if aerr != nil {
+			aerr.NodeID = node.ID
+			return nil, aerr
+		}
+		messages = fetched
```
langwatch-agent, line 543: **P1** — After this replacement, `messages` contains the fetched attachment bytes as base64 data URLs, and the same slice is passed to `startLLMSpan`, which JSON-serializes it into `langwatch.input`. A normal 20 MB attachment becomes a ~27 MB span attribute and stores the actual private image/PDF/audio payload in traces; this can blow OTLP/ClickHouse limits and is a major privacy regression versus recording the original URL. Send the full parts to the model, but pass a redacted/summarized copy to tracing (for example `{type:image_url, image_url:{url:"[fetched image/png 12345 bytes]"}}`) and test that fetched bytes are not recorded.

services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,403 @@
...
+			}
+			att, ne := f.fetch(ctx, url)
+			if ne != nil {
+				return nil, ne
+			}
+			part, ok := contentPartForAttachment(att)
+			if ok {
+				out = append(out, part)
+			} else {
+				out = append(out, block)
```
langwatch-agent: **P2** — For an explicit `image_url` part, a reachable non-attachment response falls through and sends the original remote URL to the provider. That bypasses the whole server-side validation/inline path and contradicts the strict intent semantics for explicit attachments: a bad explicit attachment should fail before any provider call. If the fetched response is not an image (or not a supported attachment type, depending on the intended contract), return an `attachment_fetch_error` instead of preserving the raw URL. Add a test for `image_url` pointing at `text/html`.

services/nlpgo/app/engine/attachment.go
```diff
@@ -0,0 +1,403 @@
...
+	client   *http.Client
+	ssrf     httpblock.SSRFOptions
+	maxBytes int64
+}
+
+func newAttachmentFetcher(ssrf httpblock.SSRFOptions) *attachmentFetcher {
+	return &attachmentFetcher{
+		client: &http.Client{
+			Timeout:   defaultAttachmentTimeout,
+			Transport: &http.Transport{DialContext: httpblock.SafeDialer(ssrf)},
```
langwatch-agent: **P2** — This constructs a bare `http.Transport`, so attachment fetches do not inherit the standard transport settings such as `Proxy: http.ProxyFromEnvironment`. The HTTP block clones `http.DefaultTransport` before installing `SafeDialer`, which means deployments that require an outbound proxy can use HTTP blocks but attachment fetches will silently fail. Clone `http.DefaultTransport.(*http.Transport)` and only override `DialContext`, matching the HTTP block.

langwatch-agent (changes requested):
Not ready to merge: 2 P1s and 2 P2s. The fetch path works in the happy tests, but it leaks signed URLs / fetched payloads into observability and leaves one explicit attachment failure mode to the provider; details inline.

services/nlpgo/app/engine/attachment.go
rogeriochaves: Best-effort in the current code: in splitStringAttachments a bare prose URL whose fetch fails is left as text and does not abort the run; only the explicit image_url part path hard-fails. That landed in d9fd9a850 (the intent split), which you confirmed in your re-check comment. Resolving.

services/nlpgo/app/engine/attachment.go
rogeriochaves: audioFormat already maps audio/mpeg + audio/mp3 to mp3, and audio/wav + audio/wave + audio/x-wav to wav (d9fd9a850); the test asserts mp3, not mpeg. The trimmed-subtype fallback only applies to other audio subtypes. Resolving.

services/nlpgo/app/engine/attachment.go
rogeriochaves: Fixed in 00dd4fd10: attachmentError now redacts the query string and fragment (redactURLForError) before building the message, so X-Amz-Signature / credentials / tokens never appear. Added a regression with a signed-looking URL asserting the secret params do not leak.

services/nlpgo/app/engine/engine.go
rogeriochaves: Fixed in 00dd4fd10: startLLMSpan now receives a redacted copy (redactAttachmentsForTracing) where inline image/audio/file bytes become [media-type, N bytes]; req.Messages still carries the full bytes to the model. Test asserts the base64 is absent from the traced copy but present in the model copy, and the live image/audio/pdf probes still pass.

services/nlpgo/app/engine/attachment.go
rogeriochaves: Fixed in 00dd4fd10: an image_url part resolving to a non-image now returns attachment_fetch_error instead of falling through to the raw URL. Added a test for an image_url pointing at text/html.

services/nlpgo/app/engine/attachment.go
rogeriochaves: Fixed in 00dd4fd10: newAttachmentFetcher now clones http.DefaultTransport and only overrides DialContext with SafeDialer, matching the HTTP block, so Proxy: http.ProxyFromEnvironment is inherited.

drewdrewthis (approved):
lgtm

PR #4911, opened 2026-06-18
fix(evaluators): explain why the code evaluator Create button is disabled

rogeriochaves:
## What

The custom code evaluator drawer disabled its "Create evaluator" / "Save changes" button whenever the form was incomplete, but never said why, so it read as a dead button. A first-time author opens the drawer, sees a disabled Create button, and gets no hint that it just needs a name.

Reported from dogfooding the evaluators workbench.

## Fix

Surface the reason inline in the footer, next to the button. It names the missing requirements (a name, some code, at least one input) and clears them as they are satisfied. The button stays disabled, but it now explains itself instead of being a dead end.

Create mode seeds the default code and inputs, so in practice the one outstanding requirement is the name, and the footer reads "Add a name to create the evaluator." until you type one.

## Tests

- BDD scenario in `specs/evaluators/evaluator-management.feature`.
- Integration test (`CodeEvaluatorEditorDrawer.integration.test.tsx`): create mode shows the reason next to a disabled button, and filling the name clears the reason and enables Create.
- Unit test for the pure `codeEvaluatorDisabledReason` helper: null when complete, single / two / three missing wording (Oxford comma), and the create-vs-edit suffix.

## Screenshots

Before, the Create button is disabled with no explanation; now the footer names the missing requirement next to it:

before

After typing a name, the reason clears and Create enables:

after

### DIFF
--- a/langwatch/src/components/evaluators/CodeEvaluatorEditorDrawer.tsx
+++ b/langwatch/src/components/evaluators/CodeEvaluatorEditorDrawer.tsx
@@ -37,6 +37,7 @@ import {
 import { api } from "~/utils/api";
 import { isHandledByGlobalHandler } from "~/utils/trpcError";
 
+import { codeEvaluatorDisabledReason } from "./codeEvaluatorValidation";
 import type { EvaluatorMappingsConfig } from "./EvaluatorEditorShared";
 
 const FIELD_TYPES = ["str", "float", "bool", "list[str]", "dict"] as const;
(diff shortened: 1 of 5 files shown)


Aryansharma28 (approved):
Clean, minimal fix that turns a silent disabled button into a self-explaining one. No concerns at the P0/P1 level — multitenancy is not involved, no DB queries are touched, and the change is purely presentational logic extracted into a testable helper.

**What looks good**

- `codeEvaluatorDisabledReason` is a pure function in its own `.ts` file — zero DOM/React dependencies, easy to unit-test, and the logic is not duplicated between the hook and the view.
- Suppressing the reason during `isPending` / `isLoadingEvaluator` (line 237 in `CodeEvaluatorEditorDrawer.tsx`) is the right call: a loading form could temporarily look incomplete, so showing a reason then would be misleading noise.
- Hooks return state/callbacks and no JSX — the `.ts` extension is correct.
- Imports are all top-level (no inline `import()`).
- Test naming and nesting follow the project's BDD conventions (`describe("given …")` / `describe("when …")`), unit and integration tests are filed under the right suffixes, and the BDD scenario in `specs/evaluators/evaluator-management.feature` is behavior-focused with no implementation details.

**Minor nit (P3)**

The `<Box />` spacer in the HStack footer (line 327) works fine, but if you ever want to slim it down you could drop the Box and switch the Button's container to `justify="flex-end"` — the Box import can then be removed. Not worth blocking the PR over; just a possible future tidy-up.

**CI** — all checks green including typecheck, lint, unit (×4), integration (×6), and e2e. Ready to merge.

PR #4943, opened 2026-06-19
perf(clickhouse): guard empty aggregateId from event_log reads

langwatch-agent:
## Evidence

In the last 24h ClickHouse exception log, `event_log` reads are one of the dominant `MEMORY_LIMIT_EXCEEDED` (Code 241) shapes. The failing query is the per-aggregate event read with an **empty aggregate id**:

```
SELECT EventId, ..., EventPayload, ...
FROM event_log
WHERE TenantId = '<redacted>' AND AggregateType = 'trace' AND AggregateId = ''
```

`event_log` is `ORDER BY (TenantId, AggregateType, AggregateId, IdempotencyKey)` and `PARTITION BY toYearWeek(EventOccurredAt)`. An empty `AggregateId` seeks to the empty-id key range, which collects every event ever written without an aggregate id, and the read then materialises all of their `EventPayload` blobs (`String CODEC(ZSTD(3))`, multi-KB JSON). With no aggregate id and no time predicate, the read fans across the whole table and blows the per-query memory ceiling, which degrades the instance for every tenant sharing it.

## Suspected cause

`AbstractEventStore` validates `tenantId` on every read but not `aggregateId`. The re-fold event loader (`eventSourcingService.ts`) calls `getEvents(ctx.aggregateId, ...)`; when a projection is enqueued with a missing/empty aggregate id, that empty string flows straight to the query. No aggregate type uses an empty string as its id, so this is always a caller bug, but the read it produces is catastrophic rather than merely wrong.

## Proposed fix

Guard the empty/whitespace aggregateId at the store boundary (next to the existing `validateTenantId`), short-circuiting `getEvents` / `getEventsUpTo` to `[]` and `countEventsBefore` to `0` without issuing the query. An aggregate with no id has no events, so this is the semantically correct result. A `logWarning` hook (mirroring the existing `logError` hook, overridden in the ClickHouse store) surfaces the skip so the upstream empty-id source stays diagnosable.

This is intentionally read-path only. The write-side question (why some callers enqueue an empty aggregate id, and whether rows were written with `AggregateId = ''`) is worth a separate look, but the read guard stops the cross-tenant OOM regardless of how the empty id arises.

## Expected impact

- The empty-id read is never issued, so it can no longer OOM the instance.
- No change for real aggregate ids: those reads are unchanged.
- No write-path change.

## EXPLAIN check

`EXPLAIN PLAN indexes=1` against a large tenant (redacted), `event_log` ~323 parts / ~98.5k granules total.

Empty `AggregateId = ''` (the failing shape) — `AggregateId` drops out of the primary-key condition, leaving only `(AggregateType, TenantId)`:

```
PrimaryKey
  Condition: and((AggregateType in ['trace']), (TenantId in ['<redacted>']))
  Parts: 248/323
  Granules: 5072/98537
  -> Parts: 103, Granules: 4927   (EventPayload read across ~4927 granules ~= tens of millions of rows)
```

Real `AggregateId = '<id>'` (id participates in the key) prunes to almost nothing:

```
PrimaryKey
  Condition: and((AggregateId in ['<id>']), (AggregateType in ['trace']), (TenantId in ['<redacted>']))
  -> Parts: 0, Granules: 0       (handful of granules for a present id)
```

The guard removes the empty-id read entirely (0 granules, no query), so the relevant delta is 4927 granules -> none issued.

## Test plan

`eventStoreClickHouse.emptyAggregateGuard.unit.test.ts` (new): with a query-spy ClickHouse client, asserts that for an empty and a whitespace-only aggregateId, `getEvents` / `getEventsUpTo` return `[]` and `countEventsBefore` returns `0` **without calling `client.query`**, and that a real aggregateId still issues the read. Existing store unit tests pass unchanged.

---

Advisory PR from the ClickHouse slow-query optimizer. A human should review and merge.

## Release Notes

* **Bug Fixes**
  * Prevented event-store read operations from running with empty/whitespace `aggregateId`; these now short-circuit to empty results (or `0`) and emit a structured warning.

* **Tests**
  * Added unit tests covering empty/whitespace `aggregateId` behavior, including ensuring database queries are not executed.

* **Chores**
  * Enhanced structured logging by introducing a dedicated warning hook and formatting warning/error context consistently across the event store.

### DIFF
--- a/langwatch/src/server/event-sourcing/stores/__tests__/eventStoreClickHouse.emptyAggregateGuard.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/stores/__tests__/eventStoreClickHouse.emptyAggregateGuard.unit.test.ts
@@ -36,62 +36,68 @@ describe("EventStoreClickHouse - empty aggregateId guard", () => {
     createdAt: 1000,
   } as unknown as Event;
 
-  describe("given an empty aggregateId", () => {
-    it("getEvents returns no events without touching ClickHouse", async () => {
-      const events = await store.getEvents("", { tenantId }, aggregateType);
+  describe.each([
+    { label: "an empty aggregateId", aggregateId: "" },
+    { label: "a whitespace-only aggregateId", aggregateId: "   " },
+  ])("given $label", ({ aggregateId }) => {
+    describe("when getEvents is called", () => {
+      it("returns no events without touching ClickHouse", async () => {
+        const events = await store.getEvents(
+          aggregateId,
+          { tenantId },
+          aggregateType,
+        );
 
-      expect(events).toEqual([]);
-      expect(mockClickHouseClient.query).not.toHaveBeenCalled();
+        expect(events).toEqual([]);
+        expect(mockClickHouseClient.query).not.toHaveBeenCalled();
+      });
     });
 
-    it("getEventsUpTo returns no events without touching ClickHouse", async () => {
-      const events = await store.getEventsUpTo(
-        "",
-        { tenantId },
-        aggregateType,
-        upToEvent,
-      );
+    describe("when getEventsUpTo is called", () => {
+      it("returns no events without touching ClickHouse", async () => {
+        const events = await store.getEventsUpTo(
+          aggregateId,
+          { tenantId },
+          aggregateType,
+          upToEvent,
+        );
 
-      expect(events).toEqual([]);
-      expect(mockClickHouseClient.query).not.toHaveBeenCalled();
+        expect(events).toEqual([]);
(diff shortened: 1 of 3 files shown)


langwatch-agent:
Both addressed in 543913613: switched to nested `describe("given")`/`describe("when")` BDD blocks (via `describe.each` over the empty and whitespace-only ids), and the whitespace-only short-circuit is now asserted for `getEventsUpTo` and `countEventsBefore` too, not just `getEvents`. 7/7 pass.

rogeriochaves (approved):

PR #4944, opened 2026-06-19
perf(clickhouse): prune event_log blob reads via the EventId KSUID timestamp

langwatch-agent:
## Evidence

In the last 24h cold-scan log, `event_log` is by far the most cold-scanned table, and one of the two dominant shapes is the offloaded-blob resolution read:

```
SELECT EventPayload
FROM event_log
WHERE TenantId = {tenantId} AND AggregateType = {aggregateType}
  AND AggregateId = {aggregateId} AND EventId = {eventId}
LIMIT 1
```

It fired on the order of a thousand times in 24h with `coldScan: true`. The latency is usually fine (the row is tiny and the EventId bloom filter narrows granules), but the cost is not: every call opens every weekly partition, and the cold ones live on S3, so each blob resolution becomes a burst of S3 GETs.

## Suspected cause

`event_log` is `PARTITION BY toYearWeek(EventOccurredAt)`. The read filters only on the sort-key columns and `EventId`; it carries no `EventOccurredAt` predicate, so ClickHouse cannot prune partitions and walks all of them (S3-tiered included).

## Proposed fix

Derive the partition window from the `EventId` itself. EventIds are KSUIDs (`generateEventId()` -> `generate("event")`), so the id already carries its creation timestamp, and `EventOccurredAt` is stamped from the same ingestion clock (`occurredAt: Date.now()` at collection). The two land within queue lag of each other, so a window around the KSUID time prunes to the one or two weeks the row can live in:

```
AND (EventOccurredAt = 0 OR (EventOccurredAt >= {from} AND EventOccurredAt <= {to}))
```

The `EventOccurredAt = 0` arm keeps rows with an unknown occurred time (the column default). The lookup is still uniquely keyed by `EventId`, so the window only changes which partitions are opened, never which row is returned. Non-KSUID / unparseable ids fall back to the previous unpruned read.

### Why the id, not a caller-supplied time

An earlier revision threaded the referencing span's `startTimeUnixMs`. That is the wrong clock: `EventOccurredAt` is the ingestion time (`Date.now()`), not the span's start time. For a late-arriving or replayed span (started days ago, ingested now) the span start sits days before `EventOccurredAt`, so a window anchored on it would prune away the very partition holding the row, and the read would return nothing, degrade to `BlobNotFoundError`, and silently fall back to the truncated preview. The KSUID creation time tracks the same ingestion clock as `EventOccurredAt`, and it needs nothing threaded from callers, so every `getFromEventLog` caller prunes correctly for free.

## Expected impact

- Offloaded-blob reads prune to ~1-2 weekly partitions instead of all of them, cutting the S3 GET bursts that dominate this shape's cost.
- User-visible latency is roughly unchanged; the win is S3 request cost, in line with the cold-scan framing.
- No write-path change. No caller change.

## EXPLAIN check

`EXPLAIN PLAN indexes=1` on a large tenant (redacted), isolating the partition effect (TenantId + AggregateType, with and without the window). event_log ~326 parts total.

Without the window:

```
Parts: 249/326  Granules: 5073/98543  ->  Parts: 103, Granules: 4927
```

With the `±2-day EventOccurredAt` window:

```
Parts: 30/35    Granules: 246/2979    ->  Parts: 25,  Granules: 241
```

Parts opened drop ~326 -> ~35 and granules read 4927 -> 241 (~20x). On the real point lookup the absolute granule counts are smaller (the EventId bloom filter already narrows within a partition), but the partition-open reduction, and therefore the S3 GET reduction, is the same.

## Test plan

- `blob-store.event-log.unit.test.ts` (extended): for a real KSUID EventId, asserts the SQL gains the bounded `EventOccurredAt` window (and the `= 0` keep-clause) with params matching the id's parsed KSUID timestamp; for a non-KSUID id, asserts the predicate and params are omitted (unpruned fallback); the returned value is unchanged either way.
- `large-trace-blob-offload-readpath.integration.test.ts` (real ClickHouse via testcontainers): passes 3/3, confirming the windowed read still resolves the offloaded body when rows are written through the real path with real KSUID EventIds.

---

Advisory PR from the ClickHouse slow-query optimizer. A human should review and merge.

## Summary by CodeRabbit

* **Performance**
  * Improved event log retrieval by adding KSUID-based time pruning to reduce scanned rows when possible.

* **Tests**
  * Expanded BlobStore event-log unit tests to verify partition-pruning behavior, including correct handling of both valid and non-valid/parsable event IDs.
  * Enhanced the ClickHouse query mocking to capture the generated SQL and query parameters in the same shape as production.
  * Updated related BlobStore trace resolution tests to reflect formatting-only changes.


langwatch-agent:
Revised the approach (b94ff21c5): instead of threading the referencing span's start time, the partition window is now derived from the `EventId` itself, which is a KSUID (`generateEventId()` -> `generate("event")`).

Why: `EventOccurredAt` (the partition key) is stamped from the ingestion clock (`Date.now()` at collection), not the span's start time. A span can start days before it's ingested (late/replayed data), so a window anchored on `startTimeUnixMs` could prune away the partition that actually holds the row, and the read would silently fall back to the preview. The KSUID creation time tracks the same ingestion clock as `EventOccurredAt`, and it needs nothing threaded from callers, so every `getFromEventLog` caller prunes correctly for free. Non-KSUID ids fall back to the unpruned read. Unit + integration tests updated and green.

@coderabbitai review

langwatch-agent:
Done in b5700e7eb: dropped the helper JSDoc and the `// Returns a string` comment that only restated the code. Kept the key-shape comment since it pins the spool key contract (a why, not a what). 14/14 pass.

rogeriochaves (approved):

PR #4975, opened 2026-06-19
fix(traces): resolve >64KB offloaded IO past mixed-type OTLP attributes (#4888)

### LINKED ISSUE #4888: Trace blob-offload: >64KB IO values truncated on reads (full-value read path) — split 1 of 2 (#4984) / 2 of 2 (#4991)
## For Humans

**>64 KB trace IO is truncated on every customer-facing read because the full-value resolution path is unwired; the write path works — full bytes are in `event_log` — only customer reads fail to resolve.**

A ~400 KB trace ingested via `POST /api/collector` reads back with `spans[0].output.value` at 65,537 bytes (64 KB preview + `…`). Prod confirmed: an unresolved `eventref` pointer is present in the returned trace; re-fetching minutes later is byte-identical (not queue lag). The offload write is healthy — the bug is purely a read-path gap.

**Ask:** wire `full=true` resolution through all 4 trace-detail surfaces; list/search stay on the 64 KB preview per ADR-022.

- Prod evidence: 400,453-byte trace read back as 65,537 bytes; `langwatch.reserved.eventref.langwatch.output` pointer present and unresolved.
- Fix surface: `traces.getById`, `traces.getTracesWithSpans`, `spans.getTracesWithSpans`, REST `GET /:traceId` — none pass `blobResolutionDeps` today.

_Status: investigation complete · ACs ratified · plan complete · awaiting implementation._

<details><summary>For Agents — implementation map + ACs</summary>

### Symptoms / context

A ~400 KB trace (span `output` field ~400 KB) ingested via `POST /api/collector` (HTTP 200, flag enabled) is read back via `GET /api/trace/:id` with `spans[0].output.value` truncated to 65,537 bytes (64 KB preview + `…`). The START sentinel is present; the END sentinel is absent. Re-fetching minutes later is byte-identical — this is not queue lag. The returned trace carries an unresolved pointer:

```
spans[0].params.langwatch.reserved.eventref.langwatch.output = {"field":"langwatch.output","eventId":"<event_id>"}
```

This proves the offload wrote the eventref but the read did not resolve it. Reproduced live against production, 2026-06-17.

### Root cause

**The full-value read path is `@unimplemented`. Every customer read surface returns the 64 KB preview regardless of whether the flag is on.**

Key findings:

- **Binding read cap:** `IO_PREVIEW_BYTES = 64 * 1024` at `langwatch/src/server/app-layer/traces/lean-for-projection.ts:37`. IO attrs >64 KB → 64 KB preview + `eventref`, by design. Distinct from `COMMAND_INLINE_THRESHOLD = 256 * 1024` (S3-spool trigger) and `DEFAULT_MAX_ATTRIBUTE_VALUE_BYTES = 256 * 1024`.

- **Spec declares full-value read `@unimplemented`:** `specs/event-sourcing/large-trace-blob-offload.feature` lines 73–82 (`@e2e @track1 @unimplemented` — **AC T1.2**):
  > _"When `getTracesWithSpans` is called for that trace with `full=true`, Then the returned input is byte-identical to the ingested input (resolved server-side via the event\_log JOIN read path), And the returned output is byte-identical to the ingested output."_

  Lines 186–193 (`@e2e @track2` — **AC T2.5**):
  > _"When the trace-detail API reads the trace with `full=false`, Then it returns the inline preview from the trace summary / stored\_spans, And no event\_log JOIN occurs on this read. When the trace-detail API reads the trace with `full=true`, Then it returns the full output, resolved server-side via an event\_log JOIN, And the API response shape is unchanged from before the feature, requiring no frontend change."_

  No `full=true` param exists in any handler today.

### Read-surface audit (bare vs. wired)

| Read surface | File:line | `blobResolutionDeps` passed | Resolves full? |
|---|---|---|---|
| tRPC `traces.getById` | `langwatch/src/server/api/routers/traces.ts:57` | No | No — preview |
| tRPC `traces.getTracesWithSpans` | `langwatch/src/server/api/routers/traces.ts:257` | No | No — preview |
| tRPC `spans.getTracesWithSpans` | `langwatch/src/server/api/routers/spans.ts:28` | No | No — preview |
| REST `GET /api/trace/:id` | `langwatch/src/app/api/traces/[[...route]]/app.v1.ts:366` | No | No — preview |
| `EvaluationExecutionService` (presets) | `langwatch/src/server/app-layer/presets.ts:269` | **Yes** | **Yes — full** |

Only the eval execution path resolves full values. All customer-facing surfaces return the preview.

### Architecture notes for implementers

- `TraceService.create(prisma, blobResolutionDeps?)` at `langwatch/src/server/traces/trace.service.ts:190`; `BlobResolutionDeps = {blobStore, ioExtractionService}` defined at `:102`.
- The wired template is `langwatch/src/server/app-layer/presets.ts:269` — `TraceService.create(prisma, {blobStore, ioExtractionService})` with deps built at `:215`.
- Resolution reads ClickHouse `event_log` (NOT S3): `langwatch/src/server/traces/resolve-offloaded-traces.ts` → `blobStore.getFromEventLog` → `SELECT EventPayload FROM event_log` at `langwatch/src/server/app-layer/traces/blob-store.service.ts:171`.
- **Byte-identity note:** the resolver writes `resolvedAttrs[attrKey] = fullValue` verbatim at the span-attribute level. `trace.input`/`trace.output` are extraction-derived (`TraceIOExtractionService` normalizes + re-serializes), so ACs assert byte-identity at the span attribute; `trace.*` ACs assert full content (length > 65536, ≠ preview text).
- **DI gap:** routers have no DI container today (tRPC ctx carries only prisma). The fix adds a shared `buildTraceBlobResolutionDeps()` factory consumed by both `presets.ts` and the routers.
- Spec scenario T1.2 `@unimplemented` at `specs/event-sourcing/large-trace-blob-offload.feature`; flag-off identity test = X.2.
- "Gates green" (X.3) is a PR merge checklist item, NOT an AC.

### Additional context

Zero `edge-spool-fail-open` warnings in CloudWatch over a 6 h window against live traffic. Ingestion/ClickHouse write stability is healthy — this is purely a read-path gap.

## Acceptance Criteria

AC1 — Offloaded field reads back the full value (T1.2). GIVEN a trace ingested with an offloaded IO field >64 KB (preview + `langwatch.reserved.eventref.<key>` pointer; full bytes in `event_log.EventPayload`), for `<key>` ∈ {langwatch.input, langwatch.output, gen_ai.input.messages, gen_ai.output.messages}, including a UTF-8 multibyte char straddling the 65536-byte boundary, WHEN read through traces.getById, traces.getTracesWithSpans, spans.getTracesWithSpans, OR REST GET /:traceId with the server passing full=true, THEN the returned span attribute for `<key>` is byte-for-byte equal to the event_log.EventPayload field (Buffer.byteLength equal AND ===), carries no truncation marker / trailing `…` / `(truncated string)`, AND trace.input/trace.output contain the full resolved content (length > 65536, ≠ the preview text). Evidence: integration test reads the raw event_log row (asserts field length == ingested), then asserts span-attribute byte-identity (400 KB repro → 400,453 bytes) + trace.* length > 65536 + no marker.

AC2 — List/search/collapsed reads stay on preview with zero CH load. GIVEN the same over-threshold trace, WHEN read via any non-detail path (getAllTracesForProject/search, getTopicCounts, getCustomersAndLabels, getTracesByThreadId, digest, download), THEN each IO value is ≤65536 bytes AND zero event_log SELECTs occur. Evidence: spyOn(BlobStore.prototype,"getFromEventLog") → toHaveBeenCalledTimes(0); AND assert those routers construct TraceService with no blobResolutionDeps (resolver-fn undefined); assert returned byte length ≤ 65536.

AC3 — eventref resolves from event_log; reserved keys stripped. GIVEN a span with `langwatch.reserved.eventref.<key> = {field, eventId}` (non-empty eventId) and a matching event_log row, WHEN resolved with full=true, THEN getFromEventLog returns the full field and no attribute key with prefix `langwatch.reserved.` remains in the returned span attributes. Evidence: resolver test asserts full value + Object.keys(attrs).some(k=>k.startsWith("langwatch.reserved.")) is false.

AC4 — No-eventref / flag-off traces read identically, zero CH load (X.2). GIVEN a trace with no eventref pointer, WHEN read on a detail surface with full=true, THEN the Trace is deep-equal to the full=false read AND zero getFromEventLog calls occur (fast path, anyResolved=false at the resolver). Evidence: resolver unit test asserts identical span output + anyResolved:false + getFromEventLog toHaveBeenCalledTimes(0); integration asserts expect(full).toEqual(noFull).

AC5 — Resolution failure degrades to preview, never 500. GIVEN an over-threshold trace whose eventref cannot be resolved — (a) ClickHouse client unconfigured in this process, (b) event_log row absent (BlobNotFoundError), (c) field missing (BlobFieldNotFoundError) — WHEN read on a detail surface with full=true, THEN the response is HTTP 200 carrying the ≤64 KB preview for the unresolved field, the reserved key is still stripped, and a warn-level log is emitted. Evidence: resolver unit test with a getFromEventLog stub throwing each error type → preview returned + no reserved key + logger.warn called; integration test with CH unconfigured asserts detail read returns 200 + preview.

AC6 — Partial / mixed-field resolution. GIVEN a trace where one IO field >64 KB (eventref present) and another ≤64 KB (no eventref), AND a case where two eventref fields exist and one resolve throws, WHEN read with full=true, THEN each resolved field is byte-identical-full, each unresolved/sub-threshold field is its inline/preview value, anyResolved=true, response 200. Evidence: resolver unit test asserts per-field outcome in one response.

AC7 — Cross-tenant full read denied (T2.6). GIVEN an over-threshold trace's eventref under tenant A, WHEN a detail read runs in tenant B's context for the same EventId with full=true, THEN getFromEventLog returns no rows → field falls back to preview AND no tenant-A bytes appear in tenant-B's response. Evidence: integration test asserts the cross-tenant detail read carries only preview, never tenant-A's full value.

AC8 — No frontend/response-shape change (T2.5). THEN no client query input changes and the response JSON shape is unchanged (only IO values widen). Evidence: `git diff --stat origin/main...HEAD -- 'langwatch/src/hooks/' 'langwatch/src/components/traces/'` shows zero changed query-input files; response-schema snapshot unchanged except value length.

</details>

drewdrewthis:
## Fixes #4888 (P0) — >64KB offloaded trace IO truncated on all customer-facing `full=true` reads

PR #4890 (`c71520f`) wired `full=true` blob resolution through every trace-detail surface, **but >64KB IO still degrades to the 64KB preview in production**, so the issue was reopened. This PR fixes the actual resolution bug.

> **This is NOT a deploy gap.** #4890 is deployed (Drew confirmed prod app SHA `5cf15b90 ⊇ c71520f`) and the resolver *runs* on every detail read — it just silently fails the field lookup and falls back to preview. There is a real code bug to fix.

---

### Root cause (AC1)

`BlobStore.getFromEventLog` (`langwatch/src/server/app-layer/traces/blob-store.service.ts`) parses the `event_log` `EventPayload` (a **raw OTLP span**) to pull the full IO value. The schema required `value.stringValue` on **every** span attribute:

```ts
const spanAttributeSchema = z.object({
  key: z.string(),
  value: z.object({ stringValue: z.string() }),   // ← required on EVERY attribute
});
const eventPayloadSchema = z.object({
  span: z.object({ attributes: z.array(spanAttributeSchema) }).optional(),
  ...
});
```

But an OTLP attribute `value` is an `AnyValue` **oneof** (`stringValue | intValue | boolValue | doubleValue | arrayValue | kvlistValue | bytesValue` — see `schemas/otlp.ts:12-19`). Real LLM spans always carry non-string siblings (`gen_ai.usage.input_tokens` = `{intValue}`, `gen_ai.request.temperature` = `{doubleValue}`, …). A single non-string sibling fails `z.array(spanAttributeSchema)` → fails `eventPayloadSchema.safeParse` → throws `BlobFieldNotFoundError` → the resolver logs a warn and **keeps the 64KB preview**.

That is Drew's exact prod warn, firing continuously:
```
WARN langwatch:traces:service  event_log row not found for eventref — keeping preview value
  error: Field "langwatch.input" not found in event payload at key event_0001...
```

**Why CI stayed green:** every prior test either used string-only span attributes or mocked `getFromEventLog`, so the real mixed-type parse was never exercised.

### The fix (AC2)

Model `span.attributes` as `z.array(z.unknown())` and parse **each** attribute defensively (`key` + optional `stringValue`), so a single non-string / malformed sibling can never fail the whole-array parse and mask the offloaded IO field:

```ts
const spanAttributeSchema = z.object({
  key: z.string(),
  value: z.object({ stringValue: z.string().optional() }),
});
// eventPayloadSchema.span.attributes: z.array(z.unknown())
for (const raw of spanAttributes) {
  const attr = spanAttributeSchema.safeParse(raw);
  if (!attr.success || attr.data.key !== field) continue;
  if (typeof attr.data.value.stringValue === "string") return attr.data.value.stringValue;
}
throw new BlobFieldNotFoundError(eventId, field);
```

`getFromEventLog` is the single chokepoint for all four detail surfaces (`traces.getById`, `traces.getTracesWithSpans`, `spans.getTracesWithSpans`, REST `GET /:traceId`) — all already pass `buildTraceBlobResolutionDeps()` + `full: true` in the base, so this one fix repairs every customer-facing read at once.

> A reuse note for reviewers: a deliberate **minimal local** `spanAttributeSchema` is kept rather than reusing `otlp.ts`'s `keyValueSchema`. Reusing `z.array(keyValueSchema)` would re-introduce the all-or-nothing array parse that **caused** this P0; per-element `safeParse` tolerance is the property that fixes it. (Endorsed by the design-soundness review.)

---

### How I can prove I was successful (AC3)

**1. Falsifiability — red→green on the REAL parse path** (`getFromEventLog` is NOT mocked). New tests in `blob-store.event-log.unit.test.ts` build a genuine `eventToRecord` payload whose span carries a `>64KB` `langwatch.input`/`langwatch.output` **plus** mixed-type siblings (`intValue`/`doubleValue`/`boolValue`/`arrayValue`), then drive the real `getFromEventLog` **and** the real `resolveOffloadedTraces` orchestrator.

_Pre-fix (current prod code, `git show HEAD:…blob-store.service.ts`):_
```
 Test Files  1 failed (1)
      Tests  3 failed | 14 passed (17)
BlobFieldNotFoundError: Field "langwatch.output" not found in event payload at key event_0002...
  ❯ BlobStore.getFromEventLog blob-store.service.ts:280:13   ← eventPayloadSchema.safeParse failure
```
The pre-fix failure reproduces Drew's **exact** prod error string — the bug, in isolation, on the real code path.

_Post-fix:_ `Tests 17 passed (17)`.

**2. Real-ClickHouse proof — exercised in CI on the real read surface.** The `large-trace-blob-offload-readpath.integration.test.ts` writes a full event into a **real ClickHouse `event_log`** and drives the real read path (`SpanStorageService.getSpansByTraceId` / `resolveOffloadedTraces` + real `BlobStore`), asserting the >64KB original comes back in full (a unique tail past the 64KB boundary, length-exact), not the preview. **I hardened its fixture with mixed-type siblings**, and it **runs and passes in CI's `test-integration` lane against real ClickHouse**:
```
test-integration (2): ✓ src/.../large-trace-blob-offload-readpath.integration.test.ts (3 tests) 197ms
```
(CI sets `CI_CLICKHOUSE_URL` + starts ClickHouse + migrates — so this test is **not** skipped in standard CI, correcting the reopen note. The real eventId→event_log→full-value path with mixed-type attrs is now covered as a regression guard.)

**3. No regression / quality:** 8 sibling offload suites green (118 tests), `pnpm typecheck` clean, biome clean. Parallel code review (principles · hygiene · security · test · design-soundness) returned **no blocking findings** (verdict: READY).

**Live "before" (prod):** Drew's documented prod truncation (400,453-byte trace → 65,537 bytes, eventref present; `event_log` holds the full 401,013-byte payload). I could not pull an independent MCP `get_trace` "after" because prod runs the unfixed code and I must not deploy, and my MCP key is scoped to a different project than Drew's repro (404). **Surface exercised:** the real `getFromEventLog` + `resolveOffloadedTraces` parse path (unit, mixed-type payload) **and the real ClickHouse `event_log` read path in CI** (integration). Live HTTP confirmation lands when this deploys.

---

### Follow-ups (from Drew's reopen comment)
- **Ask #2 (un-skip the testcontainer read-path test in CI): already satisfied.** The `test-integration` lane sets `CI_CLICKHOUSE_URL`, so `large-trace-blob-offload-readpath.integration.test.ts` already runs against real ClickHouse in standard CI (3 tests, see above) — it is **not** skipped. I hardened its fixture with mixed-type siblings so it now guards this exact regression class.
- **Ask #3 (metric/alert on the resolution-failure warn):** still recommended as a separate change — the warn fires at prod volume and is otherwise invisible. Out of scope for this surgical fix.
- Minor, non-blocking (from review): the mixed-type fixture is duplicated across unit/integration helpers; `kvlistValue`/`bytesValue` variants aren't exercised (harmless under the fix). Left as cleanup.

---

> ⚠️ **Do not merge on my behalf** — Drew merges + deploys (Rule 5).

### DIFF
--- a/langwatch/src/server/app-layer/traces/__tests__/blob-store.event-log.unit.test.ts
+++ b/langwatch/src/server/app-layer/traces/__tests__/blob-store.event-log.unit.test.ts
@@ -23,8 +23,22 @@ import {
   SPAN_RECEIVED_EVENT_VERSION_LATEST,
 } from "~/server/event-sourcing/pipelines/trace-processing/schemas/constants";
 import type { SpanReceivedEvent } from "~/server/event-sourcing/pipelines/trace-processing/schemas/events";
+import {
+  NormalizedSpanKind,
+  NormalizedStatusCode,
+  type NormalizedSpan,
+} from "~/server/event-sourcing/pipelines/trace-processing/schemas/spans";
 import { eventToRecord } from "~/server/event-sourcing/stores/eventStoreUtils";
 import { EventUtils } from "~/server/event-sourcing/utils/event.utils";
+import {
+  EVENTREF_ATTR_PREFIX,
+  IO_PREVIEW_BYTES,
+} from "~/server/app-layer/traces/lean-for-projection";
+import { TraceIOExtractionService } from "~/server/app-layer/traces/trace-io-extraction.service";
+import {
+  resolveOffloadedTraces,
+  type WarnLogger,
+} from "~/server/traces/resolve-offloaded-traces";
 import {
   BlobFieldNotFoundError,
   BlobNotFoundError,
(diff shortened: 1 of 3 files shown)


rogeriochaves (dismissed):

drewdrewthis:
## Review verdict: READY

Parallel review (principles · hygiene · security · test · design-soundness) of the #4888 fix. **No blocking findings** — zero `Fix`/`New AC` actions, so no inline change-request threads were opened. The production change (`getFromEventLog` defensive per-element OTLP-attribute parse) is clean and correctness-verified; red→green falsifiability already proven (pre-fix 3 fail with the exact prod `BlobFieldNotFoundError`, post-fix 17 pass), typecheck + biome clean, 118 sibling tests green.

### Highlights
- **[design-soundness] SOUND — pass.** The minimal local `spanAttributeSchema` is the correct call, not a reinvention of `otlp.ts`'s `keyValueSchema`: reusing `z.array(keyValueSchema)` would re-introduce the all-or-nothing parse that *caused* this P0. Per-element `safeParse` + `continue` is the robustness property that fixes it. Write/read ends are symmetric (`lean-for-projection.ts` offloads `stringValue` only).
- **[principles]** Correctness edges checked sound: duplicate keys (field is always a known IO key / `"body"`), empty `stringValue` (`typeof === "string"` accepts `""`), `body` branch is mutually exclusive and handled first.
- **[security]** CH `WHERE TenantId` predicate untouched; the relaxed parse operates on an already-tenant-scoped row; two guards (key match + type check) can't be bypassed by a crafted sibling. No PII/secrets in fixtures.
- **[test]** Correct pyramid placement (real `BlobStore` + real `resolveOffloadedTraces`, only CH wire mocked); falsifiability is genuine.

### Non-blocking follow-ups (prose only — do NOT gate this PR)
- **[hygiene][test] Test-fixture duplication.** The mixed-type sibling fixture appears in 3 spots (`makeMixedTypeSpanEvent`, the inline E2E builder, and the integration `makeSpanReceivedEvent`). Reviewers judged extraction "only worth doing if it drifts further." Parameterizing the unit builder would remove ~35 lines. Optional.
- **[hygiene][test] Coverage variants.** `arrayValue` is in the unit fixtures but not the integration helper; `kvlistValue`/`bytesValue` aren't exercised (harmless under the fix — they parse as `stringValue: undefined` and are skipped). A clarifying comment would help future readers.
- **[test] Naming asymmetry** — the input-field `it` omits "for langwatch.input" while the output one names the field. Cosmetic.
- **[principles] Out-of-scope:** `parseKsuidCreatedAtMs` silently falls back to an unpruned scan on malformed KSUIDs; a debug log would aid ops. Pre-existing, not this PR.

Confidence: high. Recommend merge after CI green + human (rogeriochaves) sign-off. **Do not merge on my behalf — Drew merges + deploys.**

rogeriochaves (approved):

PR #4982, opened 2026-06-19
feat(triggers): per-project daily email cap (ADR-031 backstop)

0xdeafcafe:
Stacked on #4498 (base = `feat/trigger-outbox-dispatch`).

## Summary

Adds a **per-project daily email cap** as a backstop above the existing per-trigger **hourly** cap. The hourly cap bounds a single trigger's burst, but a project with many triggers — each sitting just under its hourly cap — can still collectively blast a customer's inbox and our SES sending reputation. This closes that gap.

- **`consumeTenantEmailCapSlot`** (`emailHourlyCap.ts`) — fixed 24h Redis day-bucket counter (`trigger-email-tenant-cap:{projectId}:{dayBucket}`), `INCRBY` the **recipient** count (actual send volume, not dispatches). Idempotent via a separate `tenant-cap-claimed:{dedupKey}` SET-NX gate, so an outbox **retry never double-counts**. In-memory fallback + **WARN** on Redis error (degraded, fail-open like the hourly cap).
- Wired into **both** the outbox dispatcher (`dispatcher.ts`) and the cron / custom-graph path (`cron/triggers/actions/sendEmail.ts`), after suppression filtering. Over-cap → **drop + WARN** (non-retryable, recorded as a drop reason in the audit row).
- `env TRIGGER_EMAIL_TENANT_DAILY_CAP` (default **10000**).
- **ADR-031** updated with §2b describing the backstop.

## Design notes

- **v1 is per-PROJECT** (matches the existing project-keyed hourly cap; the dispatch context has `projectId` directly, no project→org lookup in the hot path). Per-**organisation** is the stronger SES-reputation boundary; documented as a recommended follow-up if multi-project abuse is observed.
- Counts **recipients**, not dispatches — it's about email volume hitting SES.
- It's a **backstop set high**, not a smoother: the default is well above any legitimate project's daily volume, and the WARN is the signal to investigate / raise the limit.

## Test plan

- `pnpm test:unit` — 44 tests across `emailHourlyCap` / `dispatcher` / cron `sendEmail`: cap boundary (counts recipients), retry idempotency (no double-count), sustained-Redis-failure fallback, over-cap drop on both paths.
- `pnpm typecheck` — clean.
- Integration (`triggerDispatch.fullflow`) deps updated to a pass-through; runs in CI.

## Deployment Impact

- **Env vars:** adds `TRIGGER_EMAIL_TENANT_DAILY_CAP` (default `10000`). Optional — existing deployments need no change; the default applies automatically.
- **Helm values:** none added/changed.
- **Default `helm install` (no overrides):** the per-project daily cap is active at 10000 recipients/project/day; no configuration required.
- **BYOC dataplane:** none — this is control-plane trigger dispatch; no dataplane surface.

### DIFF
--- a/dev/docs/adr/031-trigger-email-abuse-protections.md
+++ b/dev/docs/adr/031-trigger-email-abuse-protections.md
@@ -71,6 +71,53 @@ The cap default is **100 emails per trigger per hour**, env-configurable
 (`TRIGGER_EMAIL_HOURLY_CAP`). Digest cadences cannot exceed 12/hour, so the
 cap only ever bites `immediate`-cadence triggers — by design.
 
+### 2b. Per-project daily cap (backstop above the hourly cap)
+
+The per-trigger hourly cap above bounds a single noisy trigger, but a project
+with many immediate-cadence triggers can still aggregate a large daily email
+volume under it — each trigger independently sitting just under its own hourly
+ceiling. A second, coarser limit caps the *total* trigger-email volume a whole
+project can emit per 24h, protecting SES sender reputation (which is scored on
+aggregate outbound volume, not per-trigger).
+
+In the same email branch, **after** the hourly-cap check has passed and the
+recipient set is known, the dispatcher consults a Redis fixed-day counter:
+
+```text
+key:    trigger-email-tenant-cap:{projectId}:{floor(now / 24h)}
+INCRBY recipientCount + EXPIRE 25h; if count > cap → drop
+```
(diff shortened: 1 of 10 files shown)


0xdeafcafe:
Consolidated directly into #4498 (feat/trigger-outbox-dispatch) — these changes now live on the main trigger-outbox PR, not as a stacked follow-up.

PR #4984, opened 2026-06-19
fix(traces): wire blob-resolution deps into single-trace detail reads (1 of 2 — #4888)

### LINKED ISSUE #4888: Trace blob-offload: >64KB IO values truncated on reads (full-value read path) — split 1 of 2 (#4984) / 2 of 2 (#4991)

drewdrewthis:
# Related Issue

- Resolves #4888

## Why

Issue #4888: a ~400 KB trace ingested via `POST /api/collector` reads back with `output.value` truncated to 65,537 bytes (64 KB preview + `…`), carrying an unresolved `langwatch.reserved.eventref.*` pointer. The offload **write** is healthy — the full value is in ClickHouse `event_log`. The deployed #4975 resolver + `BlobStore.getFromEventLog` are correct.

The bug is a **read-path construction gap**: the resolver only runs when both halves of the guard hold —
`if (resolveBlobs === true && this.resolveTraceSpans)` (`clickhouse-trace.service.ts:2259`). The legacy singular `GET /api/trace/:id` handler built `TraceService.create(prisma)` **without** `buildTraceBlobResolutionDeps()` and called `getById` **without** `{ full: true }`, so the guard was falsy and the preview was returned **silently**.

Prod-proven: the plural `GET /api/traces/:id` (wired by #4975) returns the FULL value; the legacy singular `GET /api/trace/:id` returned the preview.

## What

Wire the **legacy singular `GET /api/trace/:id`** handler (`traces-legacy.ts`) to mirror the plural reference (`app.v1.ts`):

```diff
-    const traceService = TraceService.create(prisma);
-    const trace = await traceService.getById(project.id, traceId, protections);
+    const traceService = TraceService.create(prisma, buildTraceBlobResolutionDeps());
+    const trace = await traceService.getById(project.id, traceId, protections, { full: true });
```

### Why touch a file called "legacy" at all?

**"Legacy" here means "older URL shape, recently ported from the Next.js `pages/api` runtime to Hono" — not "deprecated" or "scheduled for removal."** The singular `GET /api/trace/:id` is **live and consumed**: it is mounted in `api-router.ts`, it is a public/documented API endpoint external consumers call, and it is dialed internally by the scenario span-query path (`scenarios/execution/trace-api-span-query.ts`). #4888 is a real, user-facing **data-truncation bug on that live endpoint**, so the endpoint must be fixed wherever it is served. Removing or deprecating it is a separate, breaking migration — out of scope for a bug fix. The minimal correct fix is to give the singular handler the **same** blob-resolution wiring the plural path already proved in production (#4975).

### Scope: why only this one site

The other read sites flagged during investigation are **intentionally left preview-only** — wiring them would violate **AC2** ("non-detail paths stay ≤64 KB with zero `event_log` SELECTs"):

| Site | Verdict |
|---|---|
| `traces-legacy.ts` singular `GET /api/trace/:id` | **FIXED** — detail surface, prod-proven broken |
| `traces-legacy.ts` `POST /api/trace/search` | left preview — `getAllTracesForProject` list path (AC2) |
| `traces-legacy.ts` `GET /api/thread/:id` | left preview — AC2 names `getTracesByThreadId` explicitly; method also structurally drops `resolveBlobs` |
| tRPC `getTracesWithSpansByThreadIds` | left preview — callers are `TracesMapping.tsx` / `DatasetMappingPreview.tsx` (bulk mapping tools, not a detail drawer) |
| tRPC `getById` / `getTracesWithSpans`, `spans.getTracesWithSpans`, REST plural | already wired by #4975 |

## How it works (tests — Outside-In TDD against a real ClickHouse testcontainer)

**1. `traces-legacy-get-trace.unit.test.ts`** — drives the legacy Hono handler with auth mocked to passthrough; asserts it now constructs `TraceService` **with deps** and calls `getById(..., { full: true })`. **RED before the fix** (the two bug-catching assertions failed with `create({})` / `getById` missing the 4th arg), **GREEN after**.

**2. `trace-detail-blob-recall-endpoint.integration.test.ts`** — the endpoint-level gap the existing resolver test misses. Seeds `trace_summaries` (preview `ComputedInput/Output`) + `stored_spans` (leaned span carrying the eventref) + `event_log` (full value), then drives `TraceService.getById` **end-to-end** against real ClickHouse:
- **no-deps construction → 64 KB preview**, eventref still present (documents the bug);
- **`buildTraceBlobResolutionDeps()` + `{ full: true }` → full byte-identical value** (`UNIQUE_TAIL` past the 64 KB boundary), reserved `langwatch.reserved.*` namespace stripped, `trace.input/output` patched > 64 KB (AC1, AC3);
- **event_log row absent → degrades to preview, no throw, warn emitted** (AC5).

## Human verification

Two independent ways a skeptical reviewer can confirm the fix **without trusting the agent's own run**:

**A. Reproduce the read-path gap locally against real ClickHouse (~2 min):**
1. `cd langwatch && git checkout issue4888/blob-recall-readpath-deps`
2. `NODE_ENV=test pnpm exec vitest -c vitest.integration.config.ts trace-detail-blob-recall-endpoint --run`
3. Confirm **10/10 pass** — in particular the *with-deps + `full:true`* cases return the **full > 64 KB value byte-identically** (the `UNIQUE_TAIL` marker exists only past the 64 KB preview boundary), while the *no-deps* cases return only the 64 KB preview (the bug).
4. `NODE_ENV=test pnpm exec vitest run traces-legacy-get-trace` — confirm **3/3**: the handler now constructs `TraceService` with deps and calls `getById(…, { full: true })`.
5. (falsifiability) `git stash` the one-line change in `traces-legacy.ts`, re-run step 4 → the two bug-catching assertions go **RED**; pop the stash → **GREEN**.

**B. Confirm the production symptom on the deployed singular endpoint:**
1. Pick a trace whose `output.value` is > 64 KB and was offloaded (carries `langwatch.reserved.eventref.*`).
2. `GET /api/trace/:id` (legacy singular) — on `main` it returns the 64 KB preview + `…`; on this branch it returns the full value (mirrors the already-fixed plural `GET /api/traces/:id`).
3. Diff the two responses' `output.value` length — the singular response should now match the plural one.

**Surface: this is a backend-only change — it touches only the traces read-path in TypeScript, with no UI surface to screenshot.** 
The diff is 3 files, all `.ts` (`gh pr diff 4984 --name-only`): one production handler + two tests —
- `langwatch/src/server/routes/traces-legacy.ts` (the 2-line construction fix)
- `langwatch/src/server/routes/__tests__/traces-legacy-get-trace.unit.test.ts`
- `langwatch/src/server/app-layer/traces/__tests__/trace-detail-blob-recall-endpoint.integration.test.ts`

**Construction equivalence (the load-bearing argument).** The brief proved on **production** that the plural `GET /api/traces/:id` (which already passes deps) returns the FULL > 64 KB value while the legacy singular `GET /api/trace/:id` returned the 64 KB preview. After this change, the singular handler's construction is **byte-identical** to the plural reference:

```
// app.v1.ts (plural, prod-proven full)      // traces-legacy.ts (singular, this PR)
TraceService.create(                          TraceService.create(
  prisma,                                       prisma,
  buildTraceBlobResolutionDeps(),               buildTraceBlobResolutionDeps(),
);                                            );
getById(project.id, traceId, protections, {   getById(project.id, traceId, protections, {
  full: true,                                   full: true,
});                                           });
```

So "plural returns full on prod" transfers directly to the singular path. (The only remaining delta is the plural's `try/catch` for `AmbiguousTraceIdPrefixError` → 409, deferred and tracked as **#4985** — see below.)

**Every layer of the change exercised on its real surface, this session:**

| Layer | Surface exercised | Evidence |
|---|---|---|
| Handler wiring (HTTP) | the real legacy Hono handler via `app.request(...)` | `traces-legacy-get-trace.unit.test.ts` — RED before fix (`create({})` / `getById` w/o `{full:true}`), GREEN after; 3/3 |
| Resolution mechanism (read path) | real `TraceService.getById` against a **real ClickHouse** testcontainer, seeding `trace_summaries`+`stored_spans` (via the real `leanForProjection` eventref shape) + `event_log` | `trace-detail-blob-recall-endpoint.integration.test.ts` — no-deps→preview, deps+full→byte-identical full value (`UNIQUE_TAIL` past 64 KB), eventref stripped, AC5 degrade+warn; 10/10 |
| Pre-existing resolver | unchanged #4975 path, still green | `large-trace-blob-offload-readpath.integration.test.ts` 3/3 |

Local run at PR tip `7cd4f995f` (real CH, not a `CI=1` parse):

```
$ NODE_ENV=test pnpm exec vitest -c vitest.integration.config.ts trace-detail-blob-recall-endpoint --run

[globalSetup] ClickHouse URL: http://test:test@localhost:32768/test_langwatch
[globalSetup] Testcontainers started successfully

 Test Files  1 passed (1)
      Tests  10 passed (10)
numTotalTests: 10 | numPassedTests: 10 | numFailedTests: 0 | success: true
```

All 10 are real-CH round-trips through the endpoint-level `TraceService` construction:

**No-deps construction — reproduces the #4888 read-path gap:**
1. ✓ `langwatch.input` returns the 64 KB preview, not the full value
2. ✓ `langwatch.output` returns the 64 KB preview, not the full value

**With-deps `+ full:true` construction — the fix (`buildTraceBlobResolutionDeps()`):**
3. ✓ returns the **FULL** `langwatch.input` byte-identically from `event_log` (AC1)
4. ✓ returns the **FULL** `langwatch.output` byte-identically from `event_log` (AC1)
5. ✓ strips the reserved eventref namespace from `langwatch.input` span attrs (AC3)
6. ✓ strips the reserved eventref namespace from `langwatch.output` span attrs (AC3)
7. ✓ patches `trace.input` with the full resolved content (AC1, trace-level)
8. ✓ patches `trace.output` with the full resolved content (AC1, trace-level)

**Resolution-failure path — `event_log` row absent (AC5):**
9. ✓ `langwatch.input` degrades to preview, still strips the reserved key, does not throw
10. ✓ `langwatch.output` degrades to preview, still strips the reserved key, does not throw

Route-handler unit test for the production change in `traces-legacy.ts`:

```
$ NODE_ENV=test pnpm exec vitest run traces-legacy-get-trace
      Tests  3 passed (3)
```
- ✓ constructs `TraceService` with blob-resolution deps (mirrors the plural reference handler)
- ✓ calls `getById` with `full:true` so > 64 KB offloaded IO resolves (#4888)
- ✓ returns 200 with the trace json

_Both test files are in this PR's diff. The 200 KB fixture carries a unique tail marker that exists **only past** the 64 KB preview boundary, so a preview-only read fails the assertion — the with-deps pass proves a genuine full-payload recall from real ClickHouse, not a truncated read._

**Why no full-HTTP-app boot:** the diff is the 2-line construction change; its three layers (HTTP handler dispatch, the `TraceService` read path against real CH, the unchanged resolver) are each exercised on their real surface above. Booting the collector→worker→read HTTP stack end-to-end would exercise **no additional code in this diff**, and the resolver it depends on is already deployed (#4975) and prod-verified.

## Anything surprising? — one deferral, filed

- **Out-of-scope / pre-existing (filed as #4985):** the legacy singular handler returns **500 on an ambiguous trace-ID prefix** where the plural reference returns a structured **409** (`AmbiguousTraceIdPrefixError`). This throw is reachable on `origin/main` independent of this change (**not a regression**). Bundling the 409 alignment + the matching raw-vs-resolved `trace_id` evaluation keying would expand scope beyond the blob-recall fix, so it is tracked separately in **#4985**.

Closes #4888.

## Summary by CodeRabbit

* **New Features**
  * Enhanced the legacy `GET /api/trace/:id` endpoint to return complete trace content when full blob resolution is available (improves reliability for large traces), while continuing to fall back to preview data if full resolution fails.

* **Tests**
  * Added an integration test validating large-trace blob recall behavior across multiple construction/availability scenarios (including full success, fallback to preview, and attribute normalization).
  * Added a unit test to verify the legacy endpoint's service calls and expected JSON response.

### DIFF
--- a/langwatch/src/server/app-layer/traces/__tests__/trace-detail-blob-recall-endpoint.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/__tests__/trace-detail-blob-recall-endpoint.integration.test.ts
@@ -1,0 +1,708 @@
+/**
+ * ENDPOINT-LEVEL read-path integration test for the large-trace blob offload
+ * pipeline (#4888 / ADR-022) against a REAL ClickHouse testcontainer.
+ *
+ * Why this file exists alongside large-trace-blob-offload-readpath.integration.test.ts:
+ * the sibling drives the resolver/service layer with a FAKE repository — it cannot
+ * catch the #4888 ENDPOINT-CONSTRUCTION gap where TraceService is built WITHOUT
+ * buildTraceBlobResolutionDeps() (resolver never wired) or getById is called
+ * without { full: true } (resolve gate stays closed). This test closes that gap by
+ * seeding real ClickHouse rows and driving the EXACT construction the endpoints use.
+ *
+ * Two construction paths under test:
+ *   - no-deps:    TraceService.create(prisma)                   => preview only (bug)
+ *   - with-deps:  TraceService.create(prisma, buildTraceBlobResolutionDeps())
+ *                 + getById(..., { full: true })                 => full value (fix)
+ *
+ * Deterministic payload: 200 KB whose UNIQUE_TAIL only exists past the 64 KB
+ * preview boundary — a preview-only read can NEVER contain it.
+ *
+ * BDD structure: describe("given …") > describe("when …") > it("…"). No "should".
+ */
+
+import type { ClickHouseClient } from "@clickhouse/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  EVENTREF_ATTR_PREFIX,
+  IO_PREVIEW_BYTES,
+  leanForProjection,
+} from "~/server/app-layer/traces/lean-for-projection";
+import * as clickhouseClientModule from "~/server/clickhouse/clickhouseClient";
+import { prisma } from "~/server/db";
+import type { Protections } from "~/server/elasticsearch/protections";
+import type { Event } from "~/server/event-sourcing";
+import {
+  startTestContainers,
+  stopTestContainers,
+} from "~/server/event-sourcing/__tests__/integration/testContainers";
+import { generateTestTenantId } from "~/server/event-sourcing/__tests__/integration/testHelpers";
+import {
+  SPAN_RECEIVED_EVENT_TYPE,
+  SPAN_RECEIVED_EVENT_VERSION_LATEST,
+} from "~/server/event-sourcing/pipelines/trace-processing/schemas/constants";
+import type { Span, Trace } from "~/server/tracer/types";
+import { TraceService } from "~/server/traces/trace.service";
+import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
+
+// Gate identically to the canonical event_log integration tests: skip unless a
+// real ClickHouse is reachable, run against the testcontainer otherwise.
+const hasTestcontainers = !!(
+  process.env.TEST_CLICKHOUSE_URL || process.env.CI_CLICKHOUSE_URL
+);
+
+// Mock the ClickHouse routing module so BOTH the read path
+// (ClickHouseTraceService.resolveClient -> getClickHouseClientForProject) AND the
+// blob resolver (buildTraceBlobResolutionDeps -> defaultResolveClickHouseClient ->
+// getClickHouseClientForProject, wired only when isClickHouseEnabled()) resolve to
+// the testcontainer client. importOriginal keeps every other export intact.
+vi.mock("~/server/clickhouse/clickhouseClient", async (importOriginal) => ({
+  ...(await importOriginal<
+    typeof import("~/server/clickhouse/clickhouseClient")
+  >()),
+  getClickHouseClientForProject: vi.fn(),
+  // Must be true so buildTraceBlobResolutionDeps wires the CH resolver onto the
+  // BlobStore — otherwise getFromEventLog throws and every read degrades to the
+  // preview, which would mask the very fix this test proves.
+  isClickHouseEnabled: () => true,
+}));
+
+const AGGREGATE_TYPE = "trace";
+
+/** The IO fields the matrix covers; each is leaned + offloaded independently. */
+const IO_FIELDS = ["langwatch.input", "langwatch.output"] as const;
+type IoField = (typeof IO_FIELDS)[number];
+
+/**
+ * A 200 KB deterministic payload whose final bytes only exist past the 64 KB
+ * preview boundary. The preview is `value.slice(0, 64KB) + "…"`, so it can NEVER
+ * contain UNIQUE_TAIL — a preview-only read fails the tail assertion.
+ */
+const UNIQUE_TAIL = "__OFFLOAD_FULL_VALUE_TAIL_MARKER__";
+const LARGE_VALUE = "x".repeat(200_000) + UNIQUE_TAIL;
+
+/**
+ * Permissive protections that allow all content categories to be visible, so the
+ * read mappers never redact the IO value out from under the assertions. Mirrors
+ * `openProtections` in clickhouse-trace-dedup.integration.test.ts; the shape is
+ * verified against the current `Protections` type (all fields optional).
+ */
+const openProtections: Protections = {
+  canSeeCosts: true,
+  canSeeCapturedInput: true,
+  canSeeCapturedOutput: true,
+};
+
+/** Sanity: the payload genuinely exceeds the offload threshold. */
+function assertOverThreshold(value: string): void {
+  expect(Buffer.byteLength(value, "utf-8")).toBeGreaterThan(IO_PREVIEW_BYTES);
+}
+
+/**
+ * Builds a SpanReceived domain Event whose IO field (`langwatch.input` or
+ * `langwatch.output`) carries `value`. `event.id` is the EventId that
+ * `leanForProjection` embeds in the eventref and that the read path JOINs on, so
+ * the SAME `eventId` must be used for the event_log row.
+ *
+ * Mixed-type sibling attributes (intValue / doubleValue / boolValue) are included
+ * deliberately: real OTLP spans carry non-string AnyValue attributes, and a
+ * regression where a single non-string sibling fails the whole-array parse would
+ * degrade the > 64 KB read to the preview (#4888) — the real CH round-trip here
+ * would catch that.
+ */
+function makeSpanReceivedEvent({
+  tenantId,
+  traceId,
+  spanId,
+  eventId,
+  ioField,
+  ioValue,
+}: {
+  tenantId: string;
+  traceId: string;
+  spanId: string;
+  eventId: string;
+  ioField: IoField;
+  ioValue: string;
+}): Event {
+  const now = Date.now();
+  return {
+    id: eventId,
+    aggregateId: traceId,
+    aggregateType: AGGREGATE_TYPE,
+    tenantId,
+    createdAt: now,
+    occurredAt: now,
+    type: SPAN_RECEIVED_EVENT_TYPE,
+    version: SPAN_RECEIVED_EVENT_VERSION_LATEST,
+    data: {
+      span: {
+        traceId,
+        spanId,
+        name: "test-span",
+        kind: 1,
+        startTimeUnixNano: String(now * 1_000_000),
+        endTimeUnixNano: String((now + 1000) * 1_000_000),
+        attributes: [
+          // The offloaded IO field MUST stay first.
+          { key: ioField, value: { stringValue: ioValue } },
+          // Mixed-type siblings — NOT IO keys, carry NO eventref.
+          { key: "gen_ai.usage.input_tokens", value: { intValue: "100" } },
+          { key: "gen_ai.request.temperature", value: { doubleValue: 0.7 } },
+          { key: "langwatch.streaming", value: { boolValue: true } },
+        ] as never,
+        events: [],
+        links: [],
+        status: { code: 1, message: null },
+        droppedAttributesCount: 0,
+        droppedEventsCount: 0,
+        droppedLinksCount: 0,
+      },
+      resource: { attributes: [] },
+      instrumentationScope: { name: "test" },
+    },
+  } as unknown as Event;
+}
+
+/** Reads span attributes out of a leaned SpanReceived event into a string map. */
+function extractSpanAttrs(event: Event): Record<string, string> {
+  const data = event.data as {
+    span?: {
+      attributes?: Array<{ key: string; value: { stringValue?: string } }>;
+    };
+  };
+  const attrs: Record<string, string> = {};
+  for (const attr of data?.span?.attributes ?? []) {
+    if (typeof attr.value.stringValue === "string") {
+      attrs[attr.key] = attr.value.stringValue;
+    }
+  }
+  return attrs;
+}
+
+/**
+ * Inserts ONE full event_log row, exactly as the production write path stores it
+ * (`EventPayload` IS `event.data`). Mirrors the canonical event_log test idiom —
+ * JSONEachRow with a stringified payload — and stamps `_retention_days: 0`
+ * (never-expire sentinel, test-only) so a merge-cycle TTL can't evict the fixture
+ * mid-run.
+ */
+async function insertEventLogRow({
+  client,
+  tenantId,
+  aggregateId,
+  eventId,
+  eventType,
+  eventVersion,
+  eventData,
+}: {
+  client: ClickHouseClient;
+  tenantId: string;
+  aggregateId: string;
+  eventId: string;
+  eventType: string;
+  eventVersion: string;
+  eventData: unknown;
+}): Promise<void> {
+  const ts = Date.now();
+  await client.insert({
+    table: "event_log",
+    values: [
+      {
+        TenantId: tenantId,
+        AggregateType: AGGREGATE_TYPE,
+        AggregateId: aggregateId,
+        EventId: eventId,
+        EventType: eventType,
+        EventVersion: eventVersion,
+        EventTimestamp: ts,
+        EventOccurredAt: ts,
+        EventPayload: JSON.stringify(eventData),
+        IdempotencyKey: eventId,
+        _retention_days: 0,
+      },
+    ],
+    format: "JSONEachRow",
+    // Sync insert so the read-back in the same test sees the row immediately.
+    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+  });
+}
+
+/**
+ * Inserts ONE stored_spans row carrying the supplied (leaned) attribute map. A
+ * single version, StartTime ≈ OccurredAt, so the read path's span dedup
+ * (TenantId,TraceId,SpanId,max(StartTime)) and the ±2-day StartTime partition
+ * bound (clickhouse-trace.service.ts:2419-2476) both select it.
+ */
+async function insertStoredSpan({
+  client,
+  tenantId,
+  traceId,
+  spanId,
+  startTimeMs,
+  spanAttributes,
+}: {
+  client: ClickHouseClient;
+  tenantId: string;
+  traceId: string;
+  spanId: string;
+  startTimeMs: number;
+  spanAttributes: Record<string, string>;
+}): Promise<void> {
+  await client.insert({
+    table: "stored_spans",
+    values: [
+      {
+        ProjectionId: `proj-${nanoid()}`,
+        TenantId: tenantId,
+        TraceId: traceId,
+        SpanId: spanId,
+        ParentSpanId: null,
+        ParentTraceId: null,
+        ParentIsRemote: null,
+        Sampled: 1,
+        StartTime: new Date(startTimeMs),
+        EndTime: new Date(startTimeMs + 100),
+        DurationMs: 100,
+        SpanName: "test-span",
+        SpanKind: 1,
+        ServiceName: "test-service",
+        ResourceAttributes: {},
+        SpanAttributes: spanAttributes,
+        StatusCode: 1,
+        StatusMessage: null,
+        ScopeName: "test",
+        ScopeVersion: null,
+        "Events.Timestamp": [],
+        "Events.Name": [],
+        "Events.Attributes": [],
+        "Links.TraceId": [],
+        "Links.SpanId": [],
+        "Links.Attributes": [],
+        DroppedAttributesCount: 0,
+        DroppedEventsCount: 0,
+        DroppedLinksCount: 0,
+        CreatedAt: new Date(startTimeMs),
+        UpdatedAt: new Date(startTimeMs),
+      },
+    ],
+    format: "JSONEachRow",
+    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+  });
+}
+
+/**
+ * Inserts ONE trace_summaries row whose ComputedInput/ComputedOutput carry the
+ * preview JSON (`{type:"text", value:<preview>}`) — the value `trace.input` /
+ * `trace.output` reads WITHOUT resolution. Single version, OccurredAt = now so
+ * the summary dedup + the spans-scan window bound both select it.
+ */
+async function insertTraceSummary({
+  client,
+  tenantId,
+  traceId,
+  occurredAtMs,
+  computedInput,
+  computedOutput,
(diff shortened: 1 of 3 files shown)


drewdrewthis:
## Review verdict: READY

5-reviewer panel (principles, hygiene, security, test, proof) + 3 personas (Uncle Bob, Metz/Beck, Fowler) + design-soundness. All **blocking** findings resolved; remaining items are non-blocking (filed as follow-ups). Marker refreshed at HEAD `7cd4f995f` after a scoped `--since` re-review (see "Re-review" below).

### Production change
Clean. `[principles][uncle-bob][metz-beck][fowler]` — the 12-line handler change is a byte-faithful mirror of the prod-proven plural reference (`app.v1.ts:368-377`); no SRP/readability/complexity concern. `[principles][proof]` independently confirmed leaving the list (`:254`) and thread (`:326`) sites deps-free is **correct per AC2** (resolver gate `resolveBlobs===true && this.resolveTraceSpans` → deps-free = preview + zero event_log SELECTs), not an oversight.

### Security — CLEAN
`[security]` — AC7 cross-tenant isolation is **structural**: `getFromEventLog` has `TenantId` as its first WHERE predicate and the CH client is resolved per-tenant; a tenant-B read cannot return tenant-A bytes. No secrets/PII in fixtures. Auth (`traces:view`) still enforced.

### Resolved blocking findings (fixed in the `#4888` commit)
1. `[proof]` **CI lint was RED** (biome format on the new integration test) → `langwatch-app-complete` red. **Fixed** — biome clean on both files.
2. `[hygiene][test]` inline `await import()` in `it()` bodies (CLAUDE.md hard-rule). **Fixed** — hoisted to top-level (`vi.mock` is hoisted above imports, so mocks still intercept).
3. `[hygiene][test]` `!` non-null assertions (hard-rule). **Fixed** — replaced with truthy/length guards.
4. `[hygiene][test]` `require("zod")` in a `vi.mock` factory (hard-rule). **Fixed** — top-level `import { z }` closed over by the factory.

### Non-blocking (filed as follow-ups, prose-only — do not gate)
| # | Action | Detail |
|---|--------|--------|
| 1 | New Issue **#4985** | `[synthesizer]` legacy `GET /api/trace/:id` returns 500 (not 409) on ambiguous prefix — pre-existing parity gap with the plural handler |
| 2 | New Issue **#4986** | `[security]` legacy 500 responses leak `error.stack` + tenant/event IDs — pre-existing, widened by this PR's richer errors |
| 3 | New Issue **#4987** | `[hygiene][test]` blob-offload integration test fixtures duplicated across 3 files — consolidate into a shared helper |
| 4 | Decide | `[coderabbit]` new integration test file is 708 lines (> the repo's 300-line "new file" guideline). Heavy-lift extraction of fixture/assertion helpers into a sibling module — left to author/reviewer to take now vs. as a follow-up |

### Test coverage (proof-reviewer, verified load-bearing)
AC1 (full byte-identical recall), AC3 (reserved-key stripping), AC5 (degrade→preview) covered by the new real-CH integration test — which **executed and passed in CI** (test-integration shards, not just locally). AC5 warn-half at `resolve-offloaded-traces.unit.test.ts:296` (correct pyramid placement — logger seam is injectable at the resolver, not the endpoint). AC2/AC4/AC7 covered by existing #4975 tests. The unit test is genuinely RED before the fix (reviewer reverted and confirmed).

### ✅ Merge note — base-branch typecheck breakage RESOLVED via rebase
The earlier `typecheck` red was a **base-branch breakage** — 2 pre-existing `error TS1117` (duplicate `parameters` key) in `src/server/experiments-v3/execution/__tests__/workflowBuilder.test.ts`, a file this PR never touches. It was present at the old merge-base (`1aea4542c`) and **already fixed on `main`** in `#4523`. **Resolution:** this branch was rebased onto current `origin/main` (conflict-free — the PR's 3 files are untouched by intervening main commits), which carries that fix. Result: full CI is now **green — 0 fail / 0 pending** (rollup SUCCESS, including `typecheck`, `lint`, `test-unit`, `test-integration (1-6)`, `e2e`, `langwatch-app-complete`). `mergeStateStatus: CLEAN`.

### Re-review (`--since 01aa8644b`)
After the rebase, 2 CodeRabbit nits were addressed in the test files (boolean-prefix `seedEventLog`→`shouldSeedEventLog`; named/destructured params on `makeRequest`) plus a biome line-wrap, and both review threads were resolved. A scoped 4-reviewer panel (`[principles][hygiene][security][test]`) re-reviewed this delta and found it **clean — zero concerns**; the change is identifier/param-style only, with no production-code change since the prior verdict (so personas/design-soundness/drift had no surface and were not re-run). Verdict stays **READY**; marker bumped to current HEAD.

drewdrewthis:
Why are we supporting a legacy file/path here?

drewdrewthis:
@drewdrewthis good question — to be precise, this PR isn't adding *new* support for a legacy path; it's fixing a **live** one.

**"Legacy" = older URL shape, recently ported from the Next.js `pages/api` runtime to Hono — not "deprecated."** `traces-legacy.ts` is the Hono replacement for the old `src/pages/api/trace/[id].ts` (& siblings). The singular `GET /api/trace/:id` endpoint it serves is still very much alive:
- mounted in `api-router.ts` (`tracesLegacyApp`),
- a public/documented API endpoint external consumers call,
- dialed internally by the scenario span-query path (`scenarios/execution/trace-api-span-query.ts`).

**#4888 is a real data-truncation bug on that live endpoint:** a > 64 KB offloaded trace read back through `GET /api/trace/:id` returns a silently-truncated 64 KB preview (+ `…`) instead of the full value, with an unresolved `langwatch.reserved.eventref.*` pointer. The plural `GET /api/traces/:id` was already wired with the blob-resolution deps in #4975 and is **prod-verified** to return the full value; this PR makes the singular handler's construction **byte-identical** — `TraceService.create(prisma, buildTraceBlobResolutionDeps())` + `getById(…, { full: true })` — so the same prod-proven behavior transfers to the singular path.

**Why not just delete/deprecate the endpoint instead?** That's the alternative, but it's a separate, **breaking** migration: dropping `GET /api/trace/:id` would break external API consumers and the internal scenario span-query caller. Out of scope for a bug fix — so the minimal correct move is to fix the endpoint where it's served, not retire it here.

(The one behavioral delta we deliberately did *not* port — the plural's structured **409** on an ambiguous trace-ID prefix vs. the singular's **500**, which is reachable on `main` independent of this change — is deferred and tracked in #4985.)

drewdrewthis:
> **Why not just delete/deprecate the endpoint instead?** That's the alternative, but it's a separate, **breaking** migration: dropping `GET /api/trace/:id` would break external API consumers and the internal scenario span-query caller. Out of scope for a bug fix — so the minimal correct move is to fix the endpoint where it's served, not retire it here.

It seems like this isn't actually the long-term, correct fix, so we should have a plan to get us to our north star here, no? We also don't want duplication. The backend should be modular enough to have tiny api surfaces supported by a well-designed services architecture.

PR #4989, opened 2026-06-19
feat(triggers): persist-class actions ride the settle stage (ADR-032)

0xdeafcafe:
**Draft** — core dispatch change; persist-integration verified in CI only (local Docker was unavailable when this was written), so please let CI go green and give the claim-relocation a careful read before marking ready. Stacked on #4983 (base = `feat/trigger-render-diagnostics`).

Resolves the deferred CodeRabbit finding `alertTrigger.reactor.ts:54`.

## Problem

ADR-026 §"Half-formed persist" already says `ADD_TO_DATASET` snapshots the trace at dispatch time, so a row written on the first matching event is truncated relative to the trace UI a minute later — and that's why `traceDebounceMs` exists. But v1 wired debounce for **notify** only; **persist** stayed on the inline path (`alertTrigger.reactor.ts`), reasoning "idempotent at the `TriggerSent` gate." Idempotency prevents a *second* dispatch — not a truncated *first* one (ADR-026 line 44 says as much). So persist still captured half-formed folds.

## Change ([ADR-032](dev/docs/adr/032-persist-class-debounce.md))

Route persist through the same **settle → cadence** outbox path as notify:

- **Reactor** enqueues a settle payload (TTL = `traceDebounceMs`, `actionClass: "persist"`) instead of claiming + dispatching inline — mirroring the notify reactors. Shrinks to enqueue-only.
- **`handleSettle`** accepts persist: after the debounce window it re-reads the **settled** fold, evaluates filters against the complete state, and enqueues an **immediate** cadence (persist doesn't digest-batch).
- **`handleCadenceBatch`** dispatches persist via `dispatchTriggerAction`, writing the `TriggerSent` claim **after** a successful dispatch (retry-safe — matches notify; an early claim would let a transient failure suppress the re-dispatch and silently lose the row). Pre-dispatch `isSendClaimed` keeps at-most-once.
- Dispatcher gains the persist deps (`traceById` / `addToDataset` / `addToAnnotationQueue`); both stage handlers are now action-class-aware via an `actionClass` payload marker.

The claim **moves from the reactor to the cadence handler**, so at-most-once binds to the settled fold rather than the first half-formed one. Cross-pipeline dedup (trace vs eval reactor) stays safe — still a single atomic `TriggerSent` insert, just downstream — and persist side effects are row-idempotent.

## Trade-off

Persist now incurs the `traceDebounceMs` latency (default 30s) before the row lands — the intended completeness-over-immediacy trade, same as notify. No schema change.

## Test plan

- `pnpm test:unit` — **168 tests** across the outbox + reactor suites: persist reactor enqueues settle (not inline dispatch); settle→cadence→`dispatchTriggerAction`; no-match dispatches nothing; at-most-once across settle+cadence+replay; retry-safe claim-after-dispatch; **plus the full notify regression suite, green**.
- `pnpm typecheck` — clean.
- Integration (`triggerDispatch.fullflow` persist cases) runs in CI.

## Deployment Impact

- **Env vars / Helm values / schema:** none — persist dispatch reuses the existing outbox settle/cadence payloads; no new columns or config.
- **Default `helm install` (no overrides):** no install-time change.
- **Behaviour change:** persist-class trigger actions (ADD_TO_DATASET / ADD_TO_ANNOTATION_QUEUE) now wait the per-trigger `traceDebounceMs` (default 30s) before the row is written, so the captured trace is complete (completeness over immediacy) — same trade as notify.
- **BYOC dataplane:** none.

### DIFF
--- a/dev/docs/adr/032-persist-class-debounce.md
+++ b/dev/docs/adr/032-persist-class-debounce.md
@@ -1,0 +1,61 @@
+# ADR-032: Persist-class actions ride the settle stage (trace-readiness debounce)
+
+**Date:** 2026-06-19
+
+**Status:** Accepted
+
+**Supersedes:** the v1 inline-dispatch shortcut for persist-class actions described in ADR-026 (§"Half-formed persist") and ADR-030 (`alertTrigger` / `evaluationAlertTrigger` inline reactors).
+
+## Context
+
+ADR-026 §"Problem 2 — half-formed dispatch" identifies that `ADD_TO_DATASET` (and `ADD_TO_ANNOTATION_QUEUE`) snapshot the trace **at dispatch time**. If a trigger fires on the first matching event, the persisted row is truncated relative to the trace an operator browses a minute later — the dataset diverges from the trace. ADR-026 says `traceDebounceMs` exists to fix exactly this: *"wait for the trace to settle so the filter sees the final state"*, and that the knob *"drives the `stage: "settle"` dedup TTL"* on the ADR-030 outbox queue.
+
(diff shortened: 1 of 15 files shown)


0xdeafcafe:
Consolidated directly into #4498 (feat/trigger-outbox-dispatch) — these changes now live on the main trigger-outbox PR, not as a stacked follow-up.

PR #4999, opened 2026-06-20
feat(groupqueue): content-addressed tiered blob offload (ADR-029/030)

0xdeafcafe:
## Why

The event-sourcing GroupQueue fans one event out to ~30 reactor jobs, each staging a full copy of the same large payload (event + fold state). Thirty identical multi-KB-to-MB bodies per fan-out waste Redis memory and bandwidth, and bloat the Lua scripts they transit. This offloads large payloads to a content-addressed store so jobs carry a small reference instead of a full copy.

No linked issue.

## What changed

- A content-addressed, tenant-namespaced tiered blob store: payloads over 4 KiB offload to Redis (mid-size) or the existing stored-objects store (over 256 KiB), keyed by `{projectId}/{sha256}`. Byte-identical user payloads collapse to one stored blob — including across reactor fan-out: the queue-machinery fields (`__jobName` / `__attempt` / `__context` / etc.) are lifted into the envelope header so they don't perturb the hashed body, so 30 reactors over one event share one blob.
- Holder-set reference counting: a per-blob Redis SET tracks referencing jobs; the last release reclaims the blob (atomic Lua), with a 3-day TTL backstop.
- ADR-030 hardening: a 50 MiB payload cap, a gunzip decompression cap, transient-vs-missing store-error handling (retry vs fail-safe), coordinated blob/holder TTLs with refresh-on-access, atomic hold transfer on retry/squash, read-location re-mint from server-trusted inputs, a cluster-slot guard, and a branded `TenantId` through the blob path.
- Extracted the blob lifecycle out of the 1224-LOC `GroupQueueProcessor` into an `EnvelopeBlobLifecycle` collaborator.
- **Documentation refresh.** The event-sourcing `ARCHITECTURE.md` had stale claims that GroupQueue runs on BullMQ — it doesn't, and hasn't for some time. Rewrote the Queue System / Process Roles / Failure Handling sections. Added a dedicated `queues/groupQueue/ARCHITECTURE.md` (technical deep dive — staging Lua, dispatcher loop, tiered envelope across inline → Redis blob → S3, content-addressed sharing, holder-set refcount, retries, dedup, coalescing, pause/resume, tenant isolation) and `queues/groupQueue/README.md` (usage guide — when to reach for it, configuration, process roles, caveats, testing, observability).

## How it works

The staged value is a versioned envelope (`GQ2|<headerLen>|<headerJson><body>`). Small payloads stay inline; large ones move to the tiered store and the envelope carries a reference plus a per-stage hold token. Reads re-derive the storage location from `(projectId, contentHash)` rather than trusting a stored URI, and decode caps decompression so a tampered blob can't OOM the worker. GQ2 writes are gated behind `GROUP_QUEUE_ENVELOPE_WRITES_ENABLED`; old pods drop unparseable GQ2 values and recover via event replay. Design in ADR-029 (store) and ADR-030 (hardening); behaviour in the two `.feature` files.

## Test plan

- [ ] `pnpm test:unit` (45) - envelope encode/decode, size cap, zip-bomb guard, hash-over-raw, tiering, re-mint, transient classification, hash-tag guard, fan-out content-sharing (identical user payload + different machinery → one stored blob)
- [ ] `pnpm test:integration` (44) - holder release/transfer Lua reclaim sequences (incl. release-without-acquire, self-transfer, s3 reclaim), GETEX TTL refresh, the `EnvelopeBlobLifecycle` seam (incl. cross-tenant ref refusal on decode and release and transfer), and GQ2 end-to-end (offload to dispatch to reclaim, tenant-namespaced key, transient recovery)
- [ ] typecheck + biome clean

## Anything surprising?

- GQ2 writes are flag-gated and not yet enabled in production; this lands the machinery, enabling is a follow-up.
- Streaming-to-S3 (an earlier ADR-030 idea) was dropped: the 50 MiB cap is the real OOM bound, and true streaming would need a stored-objects change plus a new dependency. ADR-030 section 1 records it.
- Three review passes (code / security / test) ran on this branch. The top findings are fixed here: a gunzip decompression cap (zip-bomb guard), `sendBatch` routed through the atomic transfer (closing the partial-failure reclaim gap `send()` already had), and a direct `EnvelopeBlobLifecycle` integration test (which also caught that the s3-tier fixtures were silently compressible, so the transient test never reached s3).
- A CodeRabbit pass hardened it further: the Lua now guards a no-op release / self-transfer against reclaiming a live blob and passes only the keys that exist (no s3-tier CROSSSLOT in cluster mode), the object-store read is byte-capped, a destination-resolve failure is classified transient, and test cleanup is namespace-scoped rather than a global flushall.
- A second review pass (review-code / review-security) added the decode-time tenant cross-check (a blob ref whose `projectId` ≠ the group's tenant is refused before fetching → fail-safe), redacted storage URIs out of the lifecycle error logs, centralized the redis blob-key layout into one module so the store and the holder Lua can't drift, and extracted the shared in-memory test doubles.
- The branch was rewritten into four logical commits (docs/ADRs + specs · primitives + unit tests · GroupQueue wiring + integration · routing-exclusion) — the prior 31-commit history is preserved on the local `backup/pre-flatten` ref for archaeology if anyone wants it.
- Remaining follow-ups, none blocking a flag-gated merge: an S3 lifecycle rule as the reclaim backstop (s3 has no TTL, so a missed inline delete leaks the object); a queue-level end-to-end fan-out dedup integration test (the encode-level invariant is covered in `jobEnvelope.unit.test.ts`; multi-reactor wiring stays `@unimplemented` in the spec); and a couple of TTL-bounded dedup/coalescing edges.

## Deployment Impact

Low. GQ2 writes are gated behind `GROUP_QUEUE_ENVELOPE_WRITES_ENABLED` (default off), so with the flag unset this code is dormant and queue behaviour is unchanged. No new dependencies, no required env vars, no schema migrations; it reuses the existing Redis and stored-objects (S3) infrastructure.

Rollout is staged: enable the flag only once every consumer in the fleet reads GQ2 envelopes (old pods drop unparseable GQ2 values and recover via event replay), which is a separate operational step from this PR. Before enabling with the s3 tier in use, put the S3 reclaim backstop (a bucket lifecycle rule, noted under "Anything surprising?") in place, since s3 objects have no TTL unlike the redis tier.

### DIFF
--- a/dev/docs/adr/030-groupqueue-blob-handling-hardening.md
+++ b/dev/docs/adr/030-groupqueue-blob-handling-hardening.md
@@ -1,0 +1,81 @@
+# ADR-030: GroupQueue blob-handling hardening
+
+**Date:** 2026-06-20
+
+**Status:** Proposed
+
+**Amends:** [ADR-029](./029-groupqueue-content-addressed-payload-store.md) (GroupQueue content-addressed tiered payload store). ADR-029's tiered store, GQ2 envelope, holder-set reclaim, and queue wiring stand; this ADR hardens the blob lifecycle against the correctness, security, and design gaps a three-pass review (tests / security / code) surfaced after the implementation landed.
+
+**Relates to:** [ADR-022](./022-event-log-source-of-truth.md) (the edge `COMMAND_INLINE_THRESHOLD` this revisits), `src/server/stored-objects/` (the reused object store).
+
+## Context
+
+The GQ2 content-addressed offload shipped across eight commits. Reviewing it end-to-end found that the blob lifecycle is correct on the happy path but trusts stored state, non-atomic sequencing, and unbounded inputs in ways that break under partial failure, store tamper, or pathological size — and that the lifecycle is untestable in isolation because it's threaded through the 1224-LOC `GroupQueueProcessor`. See [specs/event-sourcing/payload-store-blob-hardening.feature](../../../specs/event-sourcing/payload-store-blob-hardening.feature) for the behavioural contract this decision supports.
+
+The findings cluster into three roots: **unbounded inputs** (a huge payload OOMs the worker at `JSON.stringify`/gzip), **trusted state + non-atomic steps** (a tampered ref or a partial-failure between acquire and release misbehaves), and **untestability** (the seams need a whole processor to exercise).
+
+## Decision
+
+### 1. Bounded-memory offload: cap the absolute size, hash over the raw source
+
+`JSON.stringify` → gzip → buffered `PUT` is the offload path. Two changes harden it:
+
+- An **absolute ceiling** `MAX_BLOB_BYTES` (**50 MiB**) bounds the one unavoidable allocation — the JSON string itself. Above it the job is **rejected at encode** with `PayloadTooLargeError`, so the producer surfaces the error rather than the worker OOMing on gzip + buffer. Exceeding the ceiling means an upstream cap (ADR-022's edge spool, `leanForProjection`'s 64 KiB IO preview) was bypassed — a bug to surface, not silently absorb. Worst-case worker memory is bounded at roughly `concurrency × MAX_BLOB_BYTES`.
+- Content addressing hashes the **raw source JSON** (passed as a separate `hashSource`), not the gzipped bytes, removing the dependency on gzip being byte-deterministic across zlib versions/levels. The store stays byte-symmetric — it stores and returns the gzipped `data`; `hashSource` only sets the `{projectId}/{hash}` key.
+
+**Streaming was considered and dropped.** A true multipart-streaming tier would need an object-store `putStream` across every driver plus the `@aws-sdk/lib-storage` dependency — a `stored-objects` change outside this change's event-sourcing-only scope — and it buys only marginal memory: the source `JSON.stringify` string is held regardless of buffered-vs-streamed, so the real worst-case is `concurrency × ceiling` of source strings, bounded by the **ceiling**, not a stream boundary. The cap is the OOM bound; the s3-tier threshold stays at 256 KiB (redis-vs-s3, both buffered).
+
+### 2. Distinguish a missing blob from a transient store error
+
+`TieredBlobStore.get` for the s3 tier must classify failures: a genuine *not-found* (`NoSuchKey` / 404) returns null → the decode fail-safe (complete the slot without the handler, recover via replay). A *transient* error (5xx, throttling, timeout, connection reset) **rethrows as retryable** so the queue's normal retry handles it. Today both collapse to "missing", so a brief S3 blip mass-drops in-flight jobs to replay instead of retrying.
+
+### 3. Coordinated, access-refreshed TTLs
+
+One `BLOB_BACKSTOP_TTL_SECONDS` constant drives both the blob and the holder-set TTL (no two-literal drift), with the invariant **holder-set TTL ≥ blob TTL**, and **both are refreshed on access** — dispatch refreshes the holder set as well as the blob (today only the blob is). This closes the premature-reclaim window where a long-lived fan-out's holder set expires while its blob is still being refreshed, dropping all members and reclaiming a blob siblings still reference.
+
+### 4. Atomic acquire→release transfer
+
+The retry and dedup-squash transitions move a hold from an old token to a new one on the same content hash. ADR-029 relied on calling acquire before release; that ordering holds under single-connection FIFO but not under a partial failure (acquire fails, release succeeds → premature reclaim). Replace the two un-awaited ops at those transitions with a **single transfer eval**: `SADD new; SREM old; reclaim the old blob iff its holder is now empty` — atomic, so no interleaving or partial failure can reclaim a blob the new reference needs. The plain stage path keeps the cheap fire-and-forget acquire.
+
+### 5. Re-mint location on read — don't trust the stored ref
+
+Decode derives the read location from server-trusted inputs — the redis key from `(projectId, hash)`, the s3 uri by re-minting via `resolveProjectStorageDestination(projectId)` + `mintS3Uri` — rather than trusting `ref.uri` carried in the envelope. A tampered envelope (requires queue-store write access) can otherwise point a read at another tenant's key/bucket; re-minting makes the stored uri advisory, not authoritative. The per-project destination is cached to avoid a resolve per read.
+
+**Tenant-attributed logging.** Every blob/holder/job log line MUST carry the owning `projectId`, log the tenant-scoped reference (`projectId` + content hash) rather than a bare hash, and never log the raw s3 uri or bucket name (use `redactStorageUri`). A log that names a blob without its tenant is a cross-tenant attribution hazard — logs are a real isolation surface. The blob lifecycle's previously-silent fire-and-forget failures now `warn` with `{ projectId, blobHash, tier }`, and the decode-failure (missing-blob) log carries the `projectId` too.
+
+### 6. Cluster-slot guard
+
+The release/transfer evals touch two keys (holder + blob) and are correct only if both land in one Redis Cluster slot — which depends entirely on the queue name carrying a hash tag (`{…}`). Assert this **at construction** and fail fast, rather than letting a hash-tag-less queue name silently `CROSSSLOT` and leak every blob to its TTL.
+
+### 7. Extract the blob lifecycle into a collaborator
+
+Pull the tiered store + holders construction, `projectId` resolution, and the acquire/release/reclaim/re-mint seams out of `GroupQueueProcessor` into a single-responsibility collaborator the processor delegates to — a domain noun (e.g. `EnvelopeBlobLifecycle`), **not** a `*Manager` (CLAUDE.md bans the suffix). This shrinks the processor and — the point — makes every seam unit-testable without standing up the whole queue.
+
+### 8. Test and tidy
+
+Add a `GroupQueue` GQ2 integration test (testcontainers) exercising the lifecycle end-to-end: offload → dispatch → reclaim, fan-out dedup, retry-keeps-blob, missing-blob fail-safe, s3 reclaim. Remove the production-dead `readEnvelopeRef` (subsumed by `readEnvelopeHold`). De-duplicate the in-memory `JobBlobStore` / `ObjectStore` test doubles into a shared helper.
+
+## Rationale / Trade-offs
+
+**Why a cap rather than streaming?** Streaming was the original plan — stream the gzip body to S3 above a 5 MiB boundary to avoid materialising a compressed buffer. But the source `JSON.stringify` string is held regardless of buffered-vs-streamed, so the real worst-case worker memory is `concurrency × ceiling` of source strings, bounded by the **ceiling**, not a stream boundary; streaming saves only the (smaller) gzip buffer. And a true streaming tier would need a `putStream` across every object-store driver plus the `@aws-sdk/lib-storage` dependency — outside this change's event-sourcing-only scope. So the ceiling is the memory bound, and streaming is dropped. Hashing the source string (not the gzip output) is kept regardless — it removes the gzip-determinism assumption.
+
+**Why atomic transfer over awaiting acquire?** Awaiting acquire before release fixes the ordering but still leaves a window if the process dies between them. One eval is the only way the transition is crash-atomic; the keys are already co-slot, so it's a two-key eval like the release it replaces — small marginal cost.
+
+**Why re-mint when the store is "trusted"?** Defence-in-depth at a tenant boundary is cheap here (a cached resolve), and "the queue's Redis is trusted" is exactly the assumption that erodes — a confused-deputy writer or a shared-bucket misconfiguration shouldn't become a cross-tenant read. The stored uri becomes a hint, not an authority.
+
+**What it costs.** A cached per-project destination resolver on the read path, one new transfer eval, a construction-time assertion, and a refactor that moves ~150 LOC out of the processor into a collaborator. The refactor is the largest diff but pays for itself in testability.
+
+## Consequences
+
+**Positive.** Worker memory is bounded regardless of payload size; a transient store blip retries instead of dropping work to replay; a long-lived fan-out can't prematurely reclaim a referenced blob; a partial failure or a tampered ref can't reclaim-early or cross tenants; a mis-tagged queue name fails fast instead of leaking silently; and the whole lifecycle is unit-testable.
+
+**Negative.** More surface (transfer eval, collaborator). The absolute ceiling introduces a reject path that didn't exist (a producer-visible `PayloadTooLargeError`). The destination cache adds an invalidation concern (bounded — BYOC bucket changes are rare, picked up on the next worker restart).
+
+**Neutral.** Handler API, dedup-id squash semantics, and the GQ2 envelope format are unchanged; this is hardening behind the same boundary.
+
+## References
+
+- Amends: [ADR-029](./029-groupqueue-content-addressed-payload-store.md)
+- Reuses: `src/server/stored-objects/` (object PUT/GET, destination resolver)
+- Spec: `specs/event-sourcing/payload-store-blob-hardening.feature`
+- Core contract: `specs/event-sourcing/payload-store-content-addressed.feature` (ADR-029)
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
@@ -195,6 +195,40 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
         });
       });
     });
+
+    describe("when the re-encoded value carries a different __attempt (the queue retry path)", () => {
+      it("hashes identically — machinery doesn't perturb the content hash (routing-exclusion)", async () => {
+        // This is the queue-level retry-cheapness guarantee: the retry re-encode
+        // produces the SAME blob hash, so the atomic transfer is a same-set
+        // SADD+SREM (one Lua eval, blob untouched) rather than a cross-set
+        // reclaim. Without routing-exclusion, __attempt would perturb the body
+        // bytes and each retry would churn through a fresh blob.
+        const v1 = await lifecycle.encode({
+          jobData: { ...REDIS_TIER_PAYLOAD, __attempt: 1 },
+          groupId: TENANT_GROUP,
+        });
+        const v2 = await lifecycle.encode({
+          jobData: { ...REDIS_TIER_PAYLOAD, __attempt: 2 },
+          groupId: TENANT_GROUP,
+        });
+
+        expect(hashOf(v2)).toBe(hashOf(v1));
+        lifecycle.acquire(v1);
+        const key = holderKey(hashOf(v1));
+        await vi.waitFor(async () => expect(await redis.scard(key)).toBe(1));
+
+        lifecycle.transfer({
+          newValue: v2,
+          oldValue: v1,
+          groupId: TENANT_GROUP,
+        });
+
+        await vi.waitFor(async () => {
+          expect(await redis.scard(key)).toBe(1);
+          expect(await redis.exists(blobKey(hashOf(v1)))).toBe(1);
+        });
+      });
+    });
   });
 
   describe("given a dedup squash to different content", () => {
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
@@ -1,0 +1,280 @@
+import type { Redis } from "ioredis";
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest";
+
+import {
+  getTestRedisConnection,
+  startTestContainers,
+  stopTestContainers,
+} from "../../../__tests__/integration/testContainers";
+import { EnvelopeBlobLifecycle } from "../envelopeBlobLifecycle";
+import { readEnvelopeHold } from "../jobEnvelope";
+import { InMemoryObjectStore, incompressible } from "./blobTestDoubles";
+
+const hasTestcontainers = !!(
+  process.env.TEST_CLICKHOUSE_URL ||
+  process.env.CI_CLICKHOUSE_URL ||
+  process.env.REDIS_URL ||
+  process.env.CI_REDIS_URL
+);
+
+const TENANT_GROUP = "proj1/agg";
+
+// > the 4 KiB inline ceiling, < the 256 KiB s3 threshold → redis tier.
+const REDIS_TIER_PAYLOAD = { bulk: "x".repeat(8 * 1024) };
+
+describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
+  let redis: Redis;
+  let objectStore: InMemoryObjectStore;
+  let lifecycle: EnvelopeBlobLifecycle;
+  let queueName: string;
+
+  beforeAll(async () => {
+    await startTestContainers();
+    redis = getTestRedisConnection()!;
+  });
+
+  beforeEach(() => {
+    vi.stubEnv("GROUP_QUEUE_ENVELOPE_WRITES_ENABLED", "true");
+    queueName = `{test/lifecycle/${crypto.randomUUID().slice(0, 8)}}`;
+    objectStore = new InMemoryObjectStore();
+    lifecycle = new EnvelopeBlobLifecycle({
+      redis,
+      queueName,
+      objectStoreFor: () => objectStore,
+      resolveStorageDestination: async () => ({
+        kind: "s3",
+        bucket: "test-bucket",
+      }),
+    });
+  });
+
+  afterEach(async () => {
+    // Scoped to this suite's hash-tagged namespace, not a global flushall.
+    const keys = await redis.keys("{test/lifecycle/*");
+    if (keys.length > 0) await redis.del(...keys);
+    vi.unstubAllEnvs();
+  });
+
+  afterAll(async () => {
+    await stopTestContainers();
+  });
+
+  const holderKey = (hash: string) =>
+    `${queueName}:gq:blobholders:proj1/${hash}`;
+  const blobKey = (hash: string) => `${queueName}:gq:blob:proj1/${hash}`;
+  const hashOf = (value: string) => readEnvelopeHold(value)!.ref.hash;
+
+  describe("given an offloaded value whose holder TTL was shortened", () => {
+    describe("when it is decoded", () => {
+      it("refreshes the holder set on access (so it outlives the blob)", async () => {
+        const value = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP,
+        });
+        lifecycle.acquire(value);
+        const key = holderKey(hashOf(value));
+        await vi.waitFor(async () => expect(await redis.scard(key)).toBe(1));
+        await redis.expire(key, 100);
+
+        const decoded = await lifecycle.decode({
+          value,
+          groupId: TENANT_GROUP,
+        });
+
+        expect(decoded).toEqual(REDIS_TIER_PAYLOAD);
+        await vi.waitFor(async () =>
+          expect(await redis.ttl(key)).toBeGreaterThan(100),
+        );
+      });
+    });
+  });
+
+  describe("given an offloaded value decoded under a different tenant's group", () => {
+    describe("when it is decoded", () => {
+      it("refuses the cross-tenant read", async () => {
+        const value = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP,
+        });
+
+        await expect(
+          lifecycle.decode({ value, groupId: "proj2/agg" }),
+        ).rejects.toThrow(/tenant mismatch/i);
+      });
+    });
+  });
+
+  describe("given a held value released under a different tenant's group", () => {
+    describe("when released", () => {
+      it("leaves the owning tenant's holder set untouched", async () => {
+        const value = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP, // tenant proj1
+        });
+        lifecycle.acquire(value);
+        const key = holderKey(hashOf(value));
+        await vi.waitFor(async () => expect(await redis.scard(key)).toBe(1));
+
+        // Release under a foreign group: the proj1 holder must NOT be dropped.
+        lifecycle.release({ values: [value], groupId: "proj2/agg" });
+
+        // Wait long enough that an unguarded release would have reclaimed it.
+        await new Promise((resolve) => setTimeout(resolve, 150));
+        expect(await redis.scard(key)).toBe(1);
+        expect(await redis.exists(blobKey(hashOf(value)))).toBe(1);
+      });
+    });
+  });
+
+  describe("given a transfer whose new ref belongs to a different tenant", () => {
+    describe("when transferred", () => {
+      it("never acquires the foreign holder (mirror of the release-side guard)", async () => {
+        // Old value is tenant proj1's; new value is tenant proj2's. Under proj1's
+        // group, the transfer must NOT acquire proj2's hold.
+        const oldValue = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP, // proj1
+        });
+        const newValue = await lifecycle.encode({
+          jobData: { bulk: "y".repeat(8 * 1024) },
+          groupId: "proj2/agg",
+        });
+        const newHash = hashOf(newValue);
+        const newHolderKey = `${queueName}:gq:blobholders:proj2/${newHash}`;
+        lifecycle.acquire(oldValue);
+        const oldKey = holderKey(hashOf(oldValue));
+        await vi.waitFor(async () => expect(await redis.scard(oldKey)).toBe(1));
+
+        lifecycle.transfer({ newValue, oldValue, groupId: TENANT_GROUP });
+
+        await new Promise((resolve) => setTimeout(resolve, 150));
+        // proj2 holder must NOT have been created by the foreign-tenant transfer.
+        expect(await redis.exists(newHolderKey)).toBe(0);
+        // proj1's hold was released through the guarded path (matches the group's tenant).
+        expect(await redis.scard(oldKey)).toBe(0);
+      });
+    });
+  });
+
+  describe("given a same-content retry", () => {
+    describe("when the hold is transferred to the re-encoded value", () => {
+      it("swaps the slot on one holder set and keeps the blob", async () => {
+        const v1 = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP,
+        });
+        const v2 = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP,
+        });
+        const hash = hashOf(v1);
+        expect(hashOf(v2)).toBe(hash); // same content → same blob
+        lifecycle.acquire(v1);
+        await vi.waitFor(async () =>
+          expect(await redis.scard(holderKey(hash))).toBe(1),
+        );
+
+        lifecycle.transfer({
+          newValue: v2,
+          oldValue: v1,
+          groupId: TENANT_GROUP,
+        });
+
+        await vi.waitFor(async () => {
+          expect(await redis.scard(holderKey(hash))).toBe(1); // old out, new in
+          expect(await redis.exists(blobKey(hash))).toBe(1); // blob stays
+        });
+      });
+    });
+  });
+
+  describe("given a dedup squash to different content", () => {
+    describe("when the hold is transferred across blobs", () => {
+      it("reclaims the displaced redis blob and holds the new one", async () => {
+        const vOld = await lifecycle.encode({
+          jobData: { bulk: "a".repeat(8 * 1024) },
+          groupId: TENANT_GROUP,
+        });
+        const vNew = await lifecycle.encode({
+          jobData: { bulk: "b".repeat(8 * 1024) },
+          groupId: TENANT_GROUP,
+        });
+        const oldHash = hashOf(vOld);
+        const newHash = hashOf(vNew);
+        expect(newHash).not.toBe(oldHash);
+        lifecycle.acquire(vOld);
+        await vi.waitFor(async () =>
+          expect(await redis.scard(holderKey(oldHash))).toBe(1),
+        );
+
+        lifecycle.transfer({
+          newValue: vNew,
+          oldValue: vOld,
+          groupId: TENANT_GROUP,
+        });
+
+        await vi.waitFor(async () => {
+          expect(await redis.exists(blobKey(oldHash))).toBe(0); // displaced blob reclaimed
+          expect(await redis.scard(holderKey(newHash))).toBe(1); // new blob held
+        });
+      });
+    });
+  });
+
+  describe("given an s3-tier blob held by one slot", () => {
+    describe("when the last holder releases", () => {
+      it("deletes the s3 object out-of-band", async () => {
+        const value = await lifecycle.encode({
+          jobData: { bulk: incompressible(768 * 1024) }, // > 256 KiB gzipped → s3
+          groupId: TENANT_GROUP,
+        });
+        expect(readEnvelopeHold(value)!.ref.tier).toBe("s3");
+        expect(objectStore.store.size).toBe(1);
+        lifecycle.acquire(value);
+
+        lifecycle.release({ values: [value], groupId: TENANT_GROUP });
+
+        await vi.waitFor(() => {
+          expect(objectStore.deleted).toHaveLength(1);
+          expect(objectStore.store.size).toBe(0);
+        });
+      });
+    });
+  });
+
+  describe("given a transfer where one side is not a GQ2 hold", () => {
+    describe("when a GQ2 value replaces an inline value", () => {
+      it("falls back to ordered acquire+release without throwing", async () => {
+        const inlineValue = await lifecycle.encode({
+          jobData: { small: 1 }, // under the inline ceiling → no hold
+          groupId: TENANT_GROUP,
+        });
+        const gq2Value = await lifecycle.encode({
+          jobData: REDIS_TIER_PAYLOAD,
+          groupId: TENANT_GROUP,
+        });
+        expect(readEnvelopeHold(inlineValue)).toBeNull();
+        expect(readEnvelopeHold(gq2Value)).not.toBeNull();
+
+        lifecycle.transfer({
+          newValue: gq2Value,
+          oldValue: inlineValue,
+          groupId: TENANT_GROUP,
+        });
+
+        await vi.waitFor(async () =>
+          expect(await redis.scard(holderKey(hashOf(gq2Value)))).toBe(1),
+        );
+      });
+    });
+  });
+});
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
@@ -1,0 +1,201 @@
+import type { Redis } from "ioredis";
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest";
+
+import {
+  getTestRedisConnection,
+  startTestContainers,
+  stopTestContainers,
+} from "../../../__tests__/integration/testContainers";
+import type { EventSourcedQueueDefinition } from "../../queue.types";
+import { GroupQueueProcessor } from "../groupQueue";
+import type { ObjectStore } from "../tieredBlobStore";
+import {
+  FlakyObjectStore,
+  InMemoryObjectStore,
+  incompressible,
+} from "./blobTestDoubles";
+
+// Skip outside testcontainers (e.g. plain unit runs).
+const hasTestcontainers = !!(
+  process.env.TEST_CLICKHOUSE_URL ||
+  process.env.CI_CLICKHOUSE_URL ||
+  process.env.REDIS_URL ||
+  process.env.CI_REDIS_URL
+);
+
+type TestPayload = { id: string; groupId: string; value: string };
+
+// > the 4 KiB inline ceiling, < the 256 KiB s3 threshold → redis tier (inspectable).
+const OFFLOADED_VALUE = "x".repeat(8 * 1024);
+const TENANT_GROUP = "proj1/agg";
+
+describe.skipIf(!hasTestcontainers)("GroupQueueProcessor — GQ2 offload", () => {
+  let redis: Redis;
+  let queues: GroupQueueProcessor<TestPayload>[];
+
+  beforeAll(async () => {
+    await startTestContainers();
+    redis = getTestRedisConnection()!;
+  });
+
+  beforeEach(() => {
+    vi.stubEnv("GROUP_QUEUE_ENVELOPE_WRITES_ENABLED", "true");
+    queues = [];
+  });
+
+  afterEach(async () => {
+    await Promise.all(queues.map((q) => q.close().catch(() => {})));
+    // Scoped to this suite's hash-tagged namespace — never a global flushall,
+    // which would race with parallel integration suites on the shared Redis.
+    const keys = await redis.keys("{test/gq2/*");
+    if (keys.length > 0) await redis.del(...keys);
+    vi.unstubAllEnvs();
+  });
+
+  afterAll(async () => {
+    await stopTestContainers();
+  });
+
+  function createQueue({
+    processFn,
+    consumerEnabled,
+    objectStore = new InMemoryObjectStore(),
+  }: {
+    processFn: (payload: TestPayload) => Promise<void>;
+    consumerEnabled: boolean;
+    objectStore?: ObjectStore;
+  }): { queue: GroupQueueProcessor<TestPayload>; name: string } {
+    const name = `{test/gq2/${crypto.randomUUID().slice(0, 8)}}`;
+    const definition: EventSourcedQueueDefinition<TestPayload> = {
+      name,
+      groupKey: (p) => p.groupId,
+      process: processFn,
+    };
+    const queue = new GroupQueueProcessor<TestPayload>(definition, redis, {
+      consumerEnabled,
+      objectStoreFor: () => objectStore,
+      resolveStorageDestination: async () => ({
+        kind: "s3",
+        bucket: "test-bucket",
+      }),
+    });
+    queues.push(queue);
+    return { queue, name };
+  }
+
+  const blobKeys = (name: string) => redis.keys(`${name}:gq:blob:*`);
+  const holderKeys = (name: string) => redis.keys(`${name}:gq:blobholders:*`);
+
+  describe("given a job whose payload exceeds the inline ceiling", () => {
+    describe("when it is processed to completion", () => {
+      it("resolves the full payload on dispatch and reclaims the blob + holder", async () => {
+        const received: TestPayload[] = [];
+        const { queue, name } = createQueue({
+          processFn: async (p) => {
+            received.push(p);
+          },
+          consumerEnabled: true,
+        });
+        await queue.waitUntilReady();
+
+        await queue.send({
+          id: "j1",
+          groupId: TENANT_GROUP,
+          value: OFFLOADED_VALUE,
+        });
+
+        // Handler receives the full payload — proving it was offloaded then resolved.
+        await vi.waitFor(() => expect(received).toHaveLength(1), {
+          timeout: 5000,
+          interval: 50,
+        });
+        expect(received[0]!.value).toBe(OFFLOADED_VALUE);
+
+        // The blob and its holder set are eagerly reclaimed once the last holder completes.
+        await vi.waitFor(
+          async () => {
+            expect(await blobKeys(name)).toHaveLength(0);
+            expect(await holderKeys(name)).toHaveLength(0);
+          },
+          { timeout: 5000, interval: 50 },
+        );
+      });
+    });
+  });
+
+  // NOTE: multi-holder *sharing* (N distinct jobs → one blob) is not reachable
+  // through the queue's normal path while content addressing covers the whole
+  // body — distinct jobs (distinct stagedJobId, derived from payload.id) have
+  // distinct content, hence distinct blobs. That cross-job dedup arrives with
+  // the routing-exclusion (a separate change). The holder reclaim sequence
+  // itself is proven in blobHolders.integration.test.ts.
+  describe("given an offloaded job", () => {
+    describe("when it is staged", () => {
+      it("keys the blob by tenant namespace and content hash", async () => {
+        // No consumer: the job stays staged so we can inspect the blob key.
+        const { queue, name } = createQueue({
+          processFn: async () => {},
+          consumerEnabled: false,
+        });
+        await queue.waitUntilReady();
+
+        await queue.send({
+          id: "j1",
+          groupId: TENANT_GROUP,
+          value: OFFLOADED_VALUE,
+        });
+
+        await vi.waitFor(
+          async () => {
+            const keys = await blobKeys(name);
+            expect(keys).toHaveLength(1);
+            // {queue}:gq:blob:<projectId>/<128-bit base64url hash>
+            const prefix = `${name}:gq:blob:proj1/`;
+            expect(keys[0]!.startsWith(prefix)).toBe(true);
+            expect(keys[0]!.slice(prefix.length)).toMatch(
+              /^[A-Za-z0-9_-]{22}$/,
+            );
+          },
+          { timeout: 5000, interval: 50 },
+        );
+      });
+    });
+  });
+
+  describe("given an s3-tier blob whose store fails transiently then recovers", () => {
+    describe("when it is processed", () => {
+      it("retries instead of dropping, and the handler eventually runs", async () => {
+        const received: TestPayload[] = [];
+        const flaky = new FlakyObjectStore(1); // fail the first get, then serve
+        const { queue } = createQueue({
+          processFn: async (p) => {
+            received.push(p);
+          },
+          consumerEnabled: true,
+          objectStore: flaky,
+        });
+        await queue.waitUntilReady();
+
+        const big = incompressible(768 * 1024); // > 256 KiB gzipped → s3 tier
+        await queue.send({ id: "s1", groupId: TENANT_GROUP, value: big });
+
+        // First dispatch hits the transient failure and re-stages; the retry
+        // (after backoff) finds the store recovered and runs the handler.
+        await vi.waitFor(() => expect(received).toHaveLength(1), {
+          timeout: 15000,
+          interval: 100,
+        });
+        expect(received[0]!.value).toBe(big);
+      }, 20000);
+    });
+  });
+});
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
@@ -132,12 +132,12 @@ describe.skipIf(!hasTestcontainers)("GroupQueueProcessor — GQ2 offload", () =>
     });
   });
 
-  // NOTE: multi-holder *sharing* (N distinct jobs → one blob) is not reachable
-  // through the queue's normal path while content addressing covers the whole
-  // body — distinct jobs (distinct stagedJobId, derived from payload.id) have
-  // distinct content, hence distinct blobs. That cross-job dedup arrives with
-  // the routing-exclusion (a separate change). The holder reclaim sequence
-  // itself is proven in blobHolders.integration.test.ts.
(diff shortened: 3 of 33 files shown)


langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      resolveStorageDestination: async () => ({
+        kind: "s3",
+        bucket: "test-bucket",
+      }),
+    });
+  });
+
+  afterEach(async () => {
+    await redis.flushall();
+    vi.unstubAllEnvs();
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - afterEach now scopes the wipe to this suite's hash-tagged namespace (test/lifecycle/...) instead of a global flushall.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  beforeEach(() => {
+    vi.stubEnv("GROUP_QUEUE_ENVELOPE_WRITES_ENABLED", "true");
+    queues = [];
+  });
+
+  afterEach(async () => {
+    await Promise.all(queues.map((q) => q.close().catch(() => {})));
+    await redis.flushall();
+    vi.unstubAllEnvs();
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - afterEach now scopes the wipe to this suite's namespace (test/gq2/...) instead of a global flushall.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/jobEnvelope.unit.test.ts
```diff
@@ -363,4 +388,170 @@ describe("jobEnvelope", () => {
...
+    it("accepts a payload at the ceiling", () => {
+      expect(() => assertPayloadWithinCap(MAX_BLOB_BYTES)).not.toThrow();
+    });
+
+    it("rejects a payload over the ceiling", () => {
+      expect(() => assertPayloadWithinCap(MAX_BLOB_BYTES + 1)).toThrow(
+        PayloadTooLargeError,
+      );
+    });
+  });
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - wrapped the two assertions in nested given/when describes.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/jobEnvelope.unit.test.ts
```diff
@@ -363,4 +388,170 @@ describe("jobEnvelope", () => {
...
+        });
+        // Decode against a store that returns a tiny gzip inflating past the cap.
+        const bombStore = {
+          get: async () => gzipSync(Buffer.alloc(MAX_BLOB_BYTES + 1)),
+        } as unknown as TieredBlobStore;
+
+        await expect(
+          decodeJobEnvelope({ value: encoded, tieredBlobs: bombStore }),
+        ).rejects.toThrow();
+      });
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - the bomb payload is now a valid JSON string that inflates past the cap, so dropping the gunzip cap fails the test instead of false-passing on a JSON-parse error.

langwatch/src/server/event-sourcing/queues/groupQueue/blobHolders.ts
```diff
@@ -0,0 +1,189 @@
...
+if redis.call("SCARD", KEYS[1]) == 0 then
+  redis.call("DEL", KEYS[1])
+  if KEYS[2] ~= "" then
+    redis.call("UNLINK", KEYS[2])
+    return 1
+  end
+  return 2
+end
+return 0
+`;
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - RELEASE_LUA returns still-held when SREM removed nothing, and TRANSFER_LUA short-circuits a self-transfer (same set and same slot) and only reclaims when SREM actually removed the old slot. Added tests for release-without-acquire and self-transfer.

langwatch/src/server/event-sourcing/queues/groupQueue/blobHolders.ts
```diff
@@ -0,0 +1,189 @@
...
+  }): Promise<ReleaseOutcome> {
+    const holderKey = this.holderKey(projectId, hash);
+    const blobKey = tier === "redis" ? this.blobKey(projectId, hash) : "";
+    const result = (await this.redis.eval(
+      RELEASE_LUA,
+      2,
+      holderKey,
+      blobKey,
+      slotId,
+    )) as number;
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - release and transfer now pass only the keys that exist (numkeys = keys.length) and the Lua checks the KEYS length, so an s3-tier eval no longer ships an empty placeholder key that would CROSSSLOT in cluster mode.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -328,26 +333,32 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
-    // payloads (readEnvelopeBlobId yields null) and new stages (orphanedValue "").
+    // Acquire this occupancy's hold BEFORE releasing any payload a dedup squash
+    // displaced, so a squash re-staging identical content never drops the shared
+    // blob to zero holders. Both are no-ops for inline/legacy values.
     if (orphanedValue) {
-      this.deleteEnvelopeBlobs([orphanedValue]);
+      // Atomic hold transfer: a dedup squash moves the hold from the displaced
+      // value to the new one in one eval, so a partial failure can't reclaim a
+      // live blob (ADR-030 §4).
+      this.blobLifecycle.transfer(jobDataJson, orphanedValue);
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - transfer now takes named params { newValue, oldValue }; updated the three call sites.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -993,6 +1013,66 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
+  /**
+   * A transient blob-store failure (network/5xx) means the body is temporarily
+   * unreachable — not gone. Re-stage the SAME envelope with backoff so the job
+   * retries instead of dropping to replay; the value is still valid and its hold
+   * token is unchanged, so there is no re-encode and no holder churn (releasing
+   * here would risk reclaiming the blob the re-stage still needs). Bounded by the
+   * `/r/` retry suffixes already in the stagedJobId, so a misclassified permanent
+   * failure still terminates at the fail-safe (ADR-030 §2).
+   */
+  private async handleTransientDecode({
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - moved the generateStagedJobId doc back above its method.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
```diff
@@ -0,0 +1,234 @@
...
+export function contentHash(bytes: Buffer): string {
+  return createHash("sha256")
+    .update(bytes)
+    .digest()
+    .subarray(0, 16)
+    .toString("base64url");
+}
+
+function redisBlobId(projectId: TenantId, hash: string): string {
+  return `${projectId}/${hash}`;
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - redisBlobId and mintUri now take object params.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
```diff
@@ -0,0 +1,234 @@
...
+}
+
+async function streamToBuffer(stream: Readable): Promise<Buffer> {
+  const chunks: Buffer[] = [];
+  for await (const chunk of stream) {
+    chunks.push(chunk as Buffer);
+  }
+  return Buffer.concat(chunks);
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - streamToBuffer caps at MAX_BLOB_BYTES and an oversized object is treated as a corrupt/missing blob (fail-safe) rather than buffered unbounded.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
```diff
@@ -0,0 +1,234 @@
...
+    try {
+      const uri = await this.mintUri(ref.projectId, ref.hash);
+      return await streamToBuffer(
+        await this.objectStoreFor(ref.projectId).get(uri),
+      );
+    } catch (err) {
+      // A genuinely-absent object is a missing blob → null → decode fail-safe
+      // (recover via replay). Anything else (network/5xx/resolve failure) is
+      // transient and must retry, not drop the job (ADR-030 §2).
+      if (isObjectMissingError(err)) return null;
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - the mint/resolve step now runs outside the missing-classification, so a destination-resolve failure is transient (retry), never missing. Added a unit test.

specs/event-sourcing/payload-store-content-addressed.feature
```diff
@@ -0,0 +1,315 @@
...
+  # clash with holder-set reclaim). The redis tier stays the queue's own store.
+  #
+  # Related ADRs: 029 (this), 026 (envelope, extended), 022 (event_log SoT — its
+  # BlobStore is event_log-centric, NOT reused here), 024 (CH cold-path tiering —
+  # distinct subsystem), 007 (event sourcing). Object store reused from
+  # src/server/stored-objects/. Pause: specs/queue-pausing/queue-pausing.feature.
+
+  Background:
+    Given a GroupQueue with jobs routed through queue-manager facades
+    And envelope v2 writes are enabled for the deployment
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 4f224ed68 - added a Background comment clarifying the steady-state baseline is overridden only scenario-locally by the Track 4 rollout scenarios.

dev/docs/adr/030-groupqueue-blob-handling-hardening.md
```diff
@@ -0,0 +1,81 @@
...
+
+### 5. Re-mint location on read — don't trust the stored ref
+
+Decode derives the read location from server-trusted inputs — the redis key from `(projectId, hash)`, the s3 uri by re-minting via `resolveProjectStorageDestination(projectId)` + `mintS3Uri` — rather than trusting `ref.uri` carried in the envelope. A tampered envelope (requires queue-store write access) can otherwise point a read at another tenant's key/bucket; re-minting makes the stored uri advisory, not authoritative. The per-project destination is cached to avoid a resolve per read.
+
+**Tenant-attributed logging.** Every blob/holder/job log line MUST carry the owning `projectId`, log the tenant-scoped reference (`projectId` + content hash) rather than a bare hash, and never log the raw s3 uri or bucket name (use `redactStorageUri`). A log that names a blob without its tenant is a cross-tenant attribution hazard — logs are a real isolation surface. The blob lifecycle's previously-silent fire-and-forget failures now `warn` with `{ projectId, blobHash, tier }`, and the decode-failure (missing-blob) log carries the `projectId` too.
+
```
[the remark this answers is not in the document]
0xdeafcafe: Intentional, resolving as won't-change. Keeping projectId in these log lines is the tenant attribution this change deliberately adds, so a blob log line is unambiguously tenant-scoped. projectId is an internal identifier rather than PII, and the structured logger already injects it across the codebase. The bucket/uri is never logged raw (it is re-minted from projectId+hash; redactStorageUri covers the unredacted case).

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -328,26 +333,32 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
+    // displaced, so a squash re-staging identical content never drops the shared
+    // blob to zero holders. Both are no-ops for inline/legacy values.
     if (orphanedValue) {
-      this.deleteEnvelopeBlobs([orphanedValue]);
+      // Atomic hold transfer: a dedup squash moves the hold from the displaced
+      // value to the new one in one eval, so a partial failure can't reclaim a
+      // live blob (ADR-030 §4).
+      this.blobLifecycle.transfer(jobDataJson, orphanedValue);
+    } else {
+      this.blobLifecycle.acquire(jobDataJson);
```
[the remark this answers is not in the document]
0xdeafcafe: Real but deferring as a follow-up. A replace:false dedup hit that drops the new value leaves its content-addressed blob unreferenced until the 3-day TTL backstop reclaims it - a bounded leak (not unbounded), and only for the different-content same-dedup-id reject. Closing it cleanly needs the staging scripts to return per-job stored/replaced status (the heavy lift you note); doing that mid-review on a flag-gated PR adds risk for a TTL-bounded edge. Tracked, not blocking.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -839,10 +855,7 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
     groupId: string;
   }): Promise<Payload | null> {
     try {
-      const jobData = await decodeJobEnvelope({
-        value: sibling.jobDataJson,
-        blobs: this.blobs,
-      });
+      const jobData = await this.blobLifecycle.decode(sibling.jobDataJson);
       return this.stripInternalFields(jobData);
     } catch {
```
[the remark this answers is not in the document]
0xdeafcafe: Real but deferring. This is the opt-in batch-coalescing path (maxBatch > 1, pre-dating this change); the transient classification added here covers the main dispatch decode. Re-staging a drained sibling on a transient error instead of releasing it is a sensible follow-up on the coalescing path. Tracked, not blocking the flag-gated merge.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -57,70 +83,187 @@ export interface JobRoutingMeta {
...
   blobs?: JobBlobStore;
+  tieredBlobs?: TieredBlobStore;
+  projectId?: TenantId;
 }): Promise<string> {
   const json = JSON.stringify(jobData);
   if (!envelopeWritesEnabled()) {
     return json;
   }
+  const jsonBytes = Buffer.byteLength(json);
+  assertPayloadWithinCap(jsonBytes);
```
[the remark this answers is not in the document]
0xdeafcafe: Deliberately after the gate, resolving as won't-change. This PR is flag-gated and dormant until GROUP_QUEUE_ENVELOPE_WRITES_ENABLED, and the OOM the cap guards (gzip + buffer) only occurs on the GQ2 offload path. Moving it before the gate would start rejecting >50 MiB payloads in the legacy/default path too - a behaviour change to the live system from an otherwise-dormant PR. The legacy path is unchanged from today (no regression).

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -57,70 +83,187 @@ export interface JobRoutingMeta {
...
+  if (tieredBlobs && projectId) {
+    const header = routingHeader(jobData, 2);
+    if (jsonBytes > INLINE_CEILING_BYTES) {
+      const ref = await tieredBlobs.put({
+        projectId,
+        data: await gzipAsync(json),
+        // Hash the RAW json, not the gzip output, so the dedup key is independent
+        // of gzip determinism (zlib version/level) — ADR-030 §1.
+        hashSource: Buffer.from(json, "utf8"),
+      });
```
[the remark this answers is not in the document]
0xdeafcafe: Correct that whole-body hashing means reactors differing only by __jobName don't yet collapse to one blob. This is the routing-exclusion, deliberately scoped as a separate change (the GQ2 test NOTE documents it): it changes the on-the-wire body format (strip routing into the header, restore on decode) and warrants its own ADR/spec touch. This PR lands the content-addressing machinery, holder refcount, and hardening; the cross-reactor dedup is the follow-up on top. I have softened the PR description so it no longer overclaims fan-out dedup, and I am flagging this to the author as a ship-now-vs-follow-up decision.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -0,0 +1,270 @@
...
+        logger.warn(
+          {
+            projectId: expected,
+            refProjectId: hold.ref.projectId,
+            blobHash: hold.ref.hash,
+            groupId,
+          },
+          "Blob ref tenant mismatch; refusing cross-tenant read",
+        );
+        throw new Error("Blob ref tenant mismatch");
```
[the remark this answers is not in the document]
0xdeafcafe: Good catch - fixed in a9ee51e4f. Threaded groupId into release() and transfer(); each now skips any hold whose ref.projectId is not the group's tenant (the foreign hold is left to its TTL), and transfer falls back to acquire + guarded-release when the old ref isn't this tenant's. So the fail-safe cleanup after a decode mismatch no longer touches a foreign holder. Added an integration test that acquires a proj1 hold, releases it under proj2's group, and asserts the proj1 holder + blob survive (the WARN 'Skipping blob release for a tenant-mismatched ref' fires).

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  afterAll(async () => {
+    await stopTestContainers();
+  });
+
+  function createQueue(
+    processFn: (payload: TestPayload) => Promise<void>,
+    consumerEnabled: boolean,
+    objectStore: ObjectStore = new InMemoryObjectStore(),
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 271585339 - createQueue takes named params { processFn, consumerEnabled, objectStore? }, all three call sites updated.

langwatch/src/server/event-sourcing/queues/groupQueue/blobKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+/** `<projectId>/<hash>` — the tenant-namespaced content id a blob is keyed by. */
+export function blobNamespaceId({
+  projectId,
+  hash,
+}: {
+  projectId: string;
+  hash: string;
+}): string {
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 271585339 - blobNamespaceId, redisBlobKey, blobHolderSetKey now accept the branded TenantId; BlobHolders' private holderKey/blobKey helpers follow. Callers (TieredBlobStore, BlobHolders public API) already carried TenantId, so the brand is intact end-to-end.

langwatch/src/server/stored-objects/project-storage-destination.ts
```diff
@@ -100,14 +100,23 @@ export function redactStorageUri(uri: string): string {
...
+
+/**
+ * Redacts every storage URI embedded in a free-text string — e.g. an object-
+ * store SDK error message that quotes the failing `s3://bucket/key`. Use on any
+ * error text that ships to a shared log sink: a BYOC tenant's bucket / account
+ * is a cross-tenant disclosure channel otherwise.
+ */
+export function redactStorageUrisInText(text: string): string {
+  return text.replace(STORAGE_URI_IN_TEXT, (uri) => redactStorageUri(uri));
+}
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 271585339 - added the gs:// handler (was the original gap the regex hinted at) and normalized scheme matching to lowercase so a case-different SDK error (S3://, GS://, FILE://) is still redacted. New unit tests cover both.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -0,0 +1,303 @@
...
+  }): void {
+    const expected = this.projectIdFor(groupId);
+    const newHold = readEnvelopeHold(newValue);
+    const oldHold = readEnvelopeHold(oldValue);
+    // Fall back to ordered acquire+release when either side isn't a GQ2 hold, or
+    // when the old ref isn't this group's tenant — the guarded release then
+    // skips the foreign hold, leaving it to its TTL (ADR-030 §5).
+    if (!newHold || !oldHold || oldHold.ref.projectId !== expected) {
+      this.acquire(newValue);
+      this.release({ values: [oldValue], groupId });
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 271585339 - added the mirror guard for newHold. If the replacement ref's projectId isn't the group's tenant, transfer logs the mismatch and routes only the guarded release of oldValue (no acquire of the foreign holder). New integration test proves a proj2 holder is NOT created when transferring under proj1's group.

Aryansharma28:
## Code Review — `feat(groupqueue): content-addressed tiered blob offload`

**What it does:** Replaces the GroupQueue's random-UUID blob offload (GQ1) with a content-addressed, tenant-namespaced tiered store (GQ2) that deduplicates fan-out payloads — 30 reactors over one event collapse to one stored blob. Adds reference-counted lifecycle via Redis holder sets (atomic Lua), a tiered hierarchy (inline → Redis → S3), decode-time gunzip caps, and full ADR documentation. GQ2 writes are flag-gated off.

The architecture, Lua atomics, and test coverage are genuinely solid. Two correctness gaps in the S3-tier put path and the tenant-mismatch release path need fixing before the flag is enabled.

---

### 🔴 BLOCKERS

**1. `tieredBlobStore.ts` — `put()` transient S3 write failures escape as raw errors, breaking the retry contract**

`put()` calls `mintUri()` and `objectStoreFor(projectId).put()` without wrapping either in a try/catch. A transient S3 503 propagates as a raw `Error`, not a `TransientBlobStoreError`. Meanwhile `get()` correctly wraps both paths. Callers that check `instanceof TransientBlobStoreError` to decide retry vs. drop will misclassify a transient store outage during job staging as a permanent failure — the job may be dropped instead of retried.

Fix: mirror the `get()` pattern — wrap `mintUri()` in its own try/catch → `TransientBlobStoreError`, then wrap `objectStoreFor.put()` in a second try/catch → `TransientBlobStoreError`.

```typescript
// put() S3 path today — no wrapping:
const uri = await this.mintUri({ projectId, hash });
await this.objectStoreFor(projectId).put(uri, data, "application/gzip");

// get() S3 path — correctly wrapped:
try {
  uri = await this.mintUri(...);
} catch (err) {
  throw new TransientBlobStoreError({ ... });
}
try {
  return await streamToBuffer(await this.objectStoreFor(ref.projectId).get(uri), MAX_BLOB_BYTES);
} catch (err) {
  if (isObjectMissingError(err) || err instanceof BlobTooLargeError) return null;
  throw new TransientBlobStoreError({ ... });
}
```

---

**2. `envelopeBlobLifecycle.ts` — tenant-mismatch decode error leaves GQ2 blob holder permanently unreleased**

`decode()` throws `new Error("Blob ref tenant mismatch")` when `hold.ref.projectId !== expected`. `groupQueue.ts` catches this and calls `blobLifecycle.release({ values: [jobDataJson], groupId })` as the fail-safe cleanup. But `release()`'s own tenant guard then sees the same mismatch, logs "Skipping blob release for a tenant-mismatched ref", and does `continue` — the `SREM` is never called. `SCARD` never reaches 0. The blob's holder set outlives the job for the full 3-day TTL backstop, leaking both the holder Redis key and the blob.

Both guards fire correctly for their own purposes, but together they create a lifecycle gap: the first prevents a cross-tenant read, the second prevents a cross-tenant delete — but neither decrements the *correct* tenant's holder count (which is the same tenant in the normal error case where routing corrupted the ref).

Fix: In `decode()`, before throwing on tenant mismatch, call `release()` with the *value's* `groupId` derived from `hold.ref.projectId` so the holder is decremented under the correct tenant; or: have `release()` still call SREM for the blob's own projectId even when it mismatches the group's expected tenant (the hold token is per-slot and harmless to release).

---

### 🟡 NON-BLOCKERS

**3. `tieredBlobStore.ts:resolveDestinationCached()` — process-lifetime BYOC cache**

`destinationCache` is a `Map` with no TTL or invalidation for successful resolves (errors are evicted correctly). A BYOC project that migrates its S3 bucket at runtime will keep writing new GQ2 blobs to the old bucket until every worker restarts. Since both `put()` and `get()` use the cached destination, reads still work during the window, but blobs accumulate in the wrong bucket and are invisible to lifecycle rules on the new one.

Consider a short TTL (e.g. 5 min) or a background re-resolve on cache hit past an age threshold.

---

**4. `envelopeBlobLifecycle.ts` constructor — silent GQ1 fallback when exactly one S3 dependency is missing**

```typescript
this.tieredBlobs = objectStoreFor && resolveStorageDestination
  ? new TieredBlobStore(...)
  : undefined;
```

If exactly one of the two is provided (a composition-root copy-paste mistake), `tieredBlobs` is `undefined` and all encoding silently falls back to GQ1. No warning is logged. An operator sees no signal that GQ2 content-addressed dedup is disabled.

Consider adding a `logger.warn(...)` or an assertion when exactly one is provided (both-or-neither is the valid contract).

---

**5. `blobHolders.ts:acquire()` — `multi().exec()` per-command errors silently swallowed**

```typescript
await this.redis.multi().sadd(key, slotId).expire(key, BLOB_HOLDER_TTL_SECONDS).exec();
```

`exec()` returns `[error | null, result][]`. If `EXPIRE` fails (WRONGTYPE race, OOM), the holder set has no TTL and leaks Redis memory indefinitely with no observable signal. The `exec()` Promise still resolves — no exception reaches the `.catch()` in `acquire()`.

The RELEASE_LUA `SREM-returns-0 → still-held` guard prevents premature reclamation (good), but a permanently TTL-less holder set is an operational memory leak. Consider checking exec() results and logging a warning on per-command errors.

---

### 🔵 CLEANUP

**6. `tieredBlobStore.ts:ObjectStore` duplicates `StorageDriver` in `stored-objects/`**

`ObjectStore` has `put(uri, bytes, mediaType)/get(uri)/delete(uri)` — structurally identical to `StorageDriver`. A `StorageRegistry` (which implements `StorageDriver`) can't be typed as `ObjectStore` without an adapter. Future contract evolution must be made in both with no compiler link between them. The PR description acknowledges "Structurally satisfied by the stored-objects `StorageRegistry`" — worth just reusing `StorageDriver` directly.

---

**7. `tieredBlobStore.ts:isObjectMissingError()` duplicates `isNotFoundError` in `s3-driver.ts`**

Both check `e.name === "NoSuchKey"`, `e.$metadata?.httpStatusCode === 404`, and `e.code === "ENOENT"`. When S3 adds a new not-found variant, two functions need updating — and the second (here) is invisible to whoever edits the first. Consider extracting this into a shared `stored-objects/` utility.

---

**8. `jobEnvelope.ts:encodeJobEnvelope()` — non-string routing field values silently dropped**

`splitMachineryFromBody` moves all `__`-prefixed keys to machinery. `encodeJobEnvelope` unconditionally `delete`s `machinery.__pipelineName/__jobType/__jobName`. But `routingHeader()` only copies them to `header.p/t/n` when `typeof value === "string"`. A non-string `__jobType` (number from JSON round-trip, type widening) ends up in machinery, gets deleted, and is absent from both the header and the body — silently lost, no error thrown. A `typeof !== "string"` assertion before the delete would surface this at staging time rather than silently dropping the field.

---

### Summary

| # | Severity | Finding |
|---|---|---|
| 1 | 🔴 Blocker | `tieredBlobStore.put()` — S3 errors not wrapped as `TransientBlobStoreError` |
| 2 | 🔴 Blocker | Tenant-mismatch in `decode()` → `release()` skips SREM → holder leaks 3 days |
| 3 | 🟡 Non-blocker | Forever-cached BYOC storage destination |
| 4 | 🟡 Non-blocker | Silent GQ1 fallback when one S3 dep missing, no log |
| 5 | 🟡 Non-blocker | `acquire()` swallows `exec()` per-command errors |
| 6 | 🔵 Cleanup | `ObjectStore` duplicates `StorageDriver` |
| 7 | 🔵 Cleanup | `isObjectMissingError` duplicates `isNotFoundError` in `s3-driver.ts` |
| 8 | 🔵 Cleanup | Non-string routing field values silently dropped in GQ2 encode path |

*Review by Claude Code (multi-agent, 6 angles)*

langwatch/src/server/event-sourcing/queues/groupQueue/blobConstants.ts
```diff
@@ -0,0 +1,30 @@
...
+ * derive their TTL from a single source — the holder set must outlive the blob
+ * it guards, so a two-literal drift would be a correctness bug.
+ */
+
+/**
+ * Backstop TTL for an offloaded blob, refreshed on access. The holder-set TTL is
+ * set to at least this (see {@link BlobHolders}); both reclaim eagerly on the
+ * happy path, so this only bounds genuine orphans — sized to outlive a weekend.
+ */
+export const BLOB_BACKSTOP_TTL_SECONDS = 3 * 24 * 60 * 60;
```
Aryansharma28, line 28: **[BLOCKER]** **GQ1 blob backstop TTL silently cut from 7 days → 3 days.**

The replaced `redisJobBlobStore.ts` hard-coded `const BLOB_TTL_SECONDS = 7 * 24 * 60 * 60` with the comment: *"Must comfortably exceed the longest plausible staged residence — retry backoff chains and paused pipelines hold jobs for hours, not **days**."* That literal is now `BLOB_BACKSTOP_TTL_SECONDS = 3 * 24 * 60 * 60` and the comment about "hours, not days" is gone. ADR-030 §3 discusses *coordinating* the holder TTL with the blob TTL — it never says the absolute number was halved.

The new GETEX-on-read refresh in `redisJobBlobStore.get` only helps *dispatched* jobs; a staged-but-not-yet-dispatched GQ1 job (long retry backoff chain, paused pipeline, delayed schedule) has no read between `put` and the TTL ticking down. With the env flag flipped on (where the failure becomes externally visible) a 3–7 day staged residence loses the blob and the decode fail-safe silently drops the job to event replay.

**Fix:** either restore the 7-day backstop for the GQ1 path (the original comment is the invariant) and let GQ2 keep its own derived constants, or document why the cut is safe — e.g. that the 2026-06-11 capacity incident's worst-case has actually been re-measured at < 3 days.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -839,9 +917,9 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
   }): Promise<Payload | null> {
     try {
-      const jobData = await decodeJobEnvelope({
+      const jobData = await this.blobLifecycle.decode({
         value: sibling.jobDataJson,
-        blobs: this.blobs,
+        groupId,
       });
       return this.stripInternalFields(jobData);
     } catch {
```
Aryansharma28: **[BLOCKER]** **`parseDrainedPayload` swallows `TransientBlobStoreError` — exactly the regression ADR-030 §2 was meant to prevent, but only for siblings.**

The dispatched-job decode at L580 carefully routes transient errors through `handleTransientDecode` so an S3 5xx triggers retry, not replay. This sibling decode catches everything with a bare `catch`, logs `"dropping (recoverable via replay)"`, and returns `null`. The sibling is dropped from the batch *and not re-staged* — so on a brief S3 blip during a batch coalesce, the dispatched job retries cleanly while every drained sibling silently falls to replay. At fan-out scale a single multi-second S3 hiccup quietly drops hundreds of siblings per coalescing window.

**Fix:** mirror the dispatched-job classifier:

```ts
} catch (err) {
  if (err instanceof TransientBlobStoreError) {
    // Re-stage the sibling so the batch coalesce can retry it instead of
    // dropping to replay (ADR-030 §2).
    throw err; // or call a sibling-flavoured handleTransientDecode
  }
  this.logger.error(...);
  return null;
}
```

The caller (`Promise.all` at L650) would also need to handle the rethrow — easiest is to thread the transient case back up so the whole batch can be re-staged.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -0,0 +1,319 @@
...
+        },
+        "Skipping blob acquire for a tenant-mismatched replacement ref",
+      );
+      this.release({ values: [oldValue], groupId });
+      return;
+    }
+    // Fall back to ordered acquire+release when either side isn't a GQ2 hold, or
+    // when the old ref isn't this group's tenant — the guarded release then
+    // skips the foreign hold, leaving it to its TTL (ADR-030 §5).
+    if (!newHold || !oldHold || oldHold.ref.projectId !== expected) {
```
Aryansharma28, line 341: **[NON-BLOCKER]** **The `transfer` fall-back is the non-atomic acquire+release that ADR-030 §4's TRANSFER_LUA was specifically introduced to close, and it dominates during the GQ1→GQ2 cutover.**

This branch fires when `!newHold || !oldHold || oldHold.ref.projectId !== expected`. During the rollout window where new encodes are GQ2 but in-flight staged values are still GQ1, every retry/squash matches `!oldHold` (the old GQ1 value has no `header.h` hold token) and falls here. Both ops are also fire-and-forget — if `acquire(newValue)` fails on a Redis blip after the GQ1 blob's `release` already ran, the GQ1 blob is gone and the new GQ2 holder set never gets a member.

**Fix options, in order of effort:**
1. (Smallest) `await this.acquire(newValue); this.release(...)` — ordered + crash-windowed but not atomic; closes the most common race.
2. (Right depth) Extend `TRANSFER_LUA` (or add a sibling) to handle the mixed-format case: SADD the new hold and run `readEnvelopeBlobId(oldValue)` deletion in one client round-trip. Needs the old value to be parseable enough to extract the GQ1 blob id, which `readEnvelopeBlobId` already does.

The rollout is exactly when the atomic guarantee is most needed; restoring it only after every staged value is GQ2 inverts the protection window.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -0,0 +1,319 @@
...
+            hash: hold.ref.hash,
+            tier: hold.ref.tier,
+            slotId: hold.token,
+          })
+          .then((outcome) => {
+            if (outcome === "reclaim-s3" && this.tieredBlobs) {
+              return this.tieredBlobs.delete(hold.ref);
+            }
+          })
+          .catch((err: unknown) => {
```
Aryansharma28, line 274: **[NON-BLOCKER]** **S3 reclaim failure: warn-only, no counter/metric — and the PR description's own "bucket lifecycle rule" follow-up is the sole backstop.**

The `.catch` here logs at `warn` with `"relying on the TTL backstop"`, but the GQ2 S3 tier has no TTL (the README at `queues/groupQueue/README.md:165` and the PR description both call out "S3 lifecycle rule as the reclaim backstop" as a deferred ops task). If that lifecycle rule isn't in place — or gets removed in a future IaC change — a transient S3 5xx during reclaim leaks the object forever: holders UNLINK'd, no future release will retry, no scheduled sweeper.

A single `blob_reclaim_s3_failures_total{projectId}` counter wired to a PagerDuty threshold catches this in hours. A warn line will not.

**Fix:** emit a counter alongside the warn, and add a follow-up issue / TODO to enqueue a reclaim job (a `gq:s3-reclaim` Redis SET keyed by `{projectId}/{hash}` drained by a reclaim worker) so a transient failure becomes recoverable in code rather than dependent on per-tenant ops config. Same shape applies to the transfer-fallback catch at L304.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
```diff
@@ -0,0 +1,294 @@
...
+    try {
+      return await streamToBuffer(
+        await this.objectStoreFor(ref.projectId).get(uri),
+        MAX_BLOB_BYTES,
+      );
+    } catch (err) {
+      // A genuinely-absent or oversized/corrupt object is a missing blob → null
+      // → decode fail-safe (recover via replay). Anything else (network/5xx) is
+      // transient and must retry, not drop the job (ADR-030 §2).
+      if (isObjectMissingError(err) || err instanceof BlobTooLargeError) {
```
Aryansharma28: **[NON-BLOCKER]** **`BlobTooLargeError` is silently coerced to `null` (treated as a missing blob) with no log / metric. Defence-in-depth fires invisibly.**

The cap exists to defend against zip-bombs and tampered BYOC objects, but the trigger is unobservable: decode returns null → `jobEnvelope.ts:331` throws `"tiered blob is missing"` → `groupQueue.ts:599-602` fail-safes the slot to event replay. A real corruption/tamper event is indistinguishable from "object got deleted by TTL".

**Fix:** split the bucket. Either throw a distinct `BlobCorruptError` (rethrow → handled separately at the decode boundary), or at minimum emit `logger.warn({ projectId, hash, byteLength }, "blob exceeds decode cap — possible tamper / zip-bomb")` here before returning null. Pair with a counter and you get an oncall signal that distinguishes "missing-after-TTL" from "someone wrote 100 MiB to a key our encoder caps at 50 MiB".

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -57,70 +83,262 @@ export interface JobRoutingMeta {
...
   blobs?: JobBlobStore;
+  tieredBlobs?: TieredBlobStore;
+  projectId?: TenantId;
 }): Promise<string> {
   const json = JSON.stringify(jobData);
   if (!envelopeWritesEnabled()) {
     return json;
   }
+  const jsonBytes = Buffer.byteLength(json);
+  assertPayloadWithinCap(jsonBytes);
```
Aryansharma28: **[NON-BLOCKER]** **`PayloadTooLargeError` thrown but no tenant-attributed log emitted before the throw.**

The error propagates to the producer (where it likely surfaces as generic ingestion-failure noise), but no GQ2-path log line records the offending `projectId` + `byteLength`. Oncall has nothing to grep — the customer reports "my spans sometimes drop" and you can't tell whether to talk to that tenant about runaway model output or hunt a server bug.

**Fix:** one structured warn at the cap boundary:

```ts
if (jsonBytes > MAX_BLOB_BYTES) {
  logger.warn({ projectId, byteLength: jsonBytes, cap: MAX_BLOB_BYTES },
    "Job payload over MAX_BLOB_BYTES — rejecting at encode");
  throw new PayloadTooLargeError(jsonBytes);
}
```

(Threading `projectId` through `assertPayloadWithinCap` adds one optional arg.)

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -57,70 +83,262 @@ export interface JobRoutingMeta {
...
 
-  const header: EnvelopeHeader = { v: 1, e: "j" };
-  if (typeof jobData.__pipelineName === "string")
-    header.p = jobData.__pipelineName;
-  if (typeof jobData.__jobType === "string") header.t = jobData.__jobType;
-  if (typeof jobData.__jobName === "string") header.n = jobData.__jobName;
+  // GQ2: content-addressed, tenant-namespaced, tiered offload. Active only once
+  // the composition root supplies a tiered store and the job's tenant.
+  if (tieredBlobs && projectId) {
```
Aryansharma28, line 287: **[NON-BLOCKER]** **Silent GQ2 → GQ1 downgrade when `projectId` is missing — no warn, no metric, no test.**

`if (tieredBlobs && projectId)` is the entire gate. The `EnvelopeBlobLifecycle.projectIdFor` choke point (returns `undefined` when `tenantIdFromGroupId` can't parse the group prefix) has no warn either. So any future `groupKeyFn` or `buildGroupKey` regression that drops the tenant prefix — or the existing `__unknown__` path in `eventSourcing.ts:447` — silently falls through to GQ1 with no dedup, no holder refcount, no tenant namespacing. The pipeline ships with dedup disabled and nothing pages.

**Fix:** one warn + counter at the downgrade boundary, *or* (better) make `projectId` non-optional and force every caller to commit to GQ1 explicitly. The current API shape lets a missing optional param silently downgrade an entire pipeline.

langwatch/src/server/event-sourcing/queues/groupQueue/redisJobBlobStore.ts
```diff
@@ -29,15 +22,28 @@ export class RedisJobBlobStore implements JobBlobStore {
...
+    );
   }
 
+  /**
+   * Reads the blob and refreshes its TTL (GETEX), so a body still referenced by
+   * a long-dwelling job never expires under it. A missing key returns null.
+   */
   async get({ id }: { id: string }): Promise<Buffer | null> {
-    return await this.redis.getBuffer(this.keyPrefix + id);
+    return await this.redis.getexBuffer(
```
Aryansharma28, line 48: **[NON-BLOCKER]** **`getexBuffer` on every read inverts the "TTL = safety net" assumption when the reader is the ops dashboard.**

`queue.redis.repository.getGroupJobs` constructs a fresh `RedisJobBlobStore` per call and decodes envelopes — so every ops-dashboard view of a blocked group's jobs refreshes every visited blob's TTL by `BLOB_BACKSTOP_TTL_SECONDS`. A repeatedly-viewed blocked group keeps its orphan blobs alive indefinitely — exactly the case that's most likely to already have orphans.

**Fix:** split into two methods — `get` (worker hot path, refreshes TTL) and `peek` (ops/inspection, no GETEX). Wire the ops repository through `peek`. Same concern applies to the GQ2 tiered store on the s3 path; the redis-tier `getBuffer` path through `decodeJobEnvelope` would also extend a holder's TTL on inspection.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
```diff
@@ -0,0 +1,294 @@
...
+  get(uri: string): Promise<Readable>;
+  delete(uri: string): Promise<void>;
+}
+
+/**
+ * Above this serialized size a blob lives in the durable object store; at or
+ * below it, in Redis. Aligned with ADR-022's COMMAND_INLINE_THRESHOLD. The
+ * inline tier (≤ a few KiB) is handled by the envelope, not this store.
+ */
+export const S3_TIER_THRESHOLD_BYTES = 256 * 1024;
```
Aryansharma28: **[NON-BLOCKER]** **`S3_TIER_THRESHOLD_BYTES = 256 * 1024` is a literal duplicate of `COMMAND_INLINE_THRESHOLD` (`langwatch/src/server/app-layer/traces/lean-for-projection.ts:30`).**

The JSDoc explicitly says *"Aligned with ADR-022's `COMMAND_INLINE_THRESHOLD`"*, but the value is re-typed rather than imported. If ADR-022's edge-spool boundary is retuned (e.g. raised to 512 KiB), only one side moves and the alignment the comment promises silently lies.

**Fix:**

```ts
import { COMMAND_INLINE_THRESHOLD } from "~/server/app-layer/traces/lean-for-projection";

export const S3_TIER_THRESHOLD_BYTES = COMMAND_INLINE_THRESHOLD;
```

Mirrors how `blobConstants.ts` already derives `BLOB_HOLDER_TTL_SECONDS` from `BLOB_BACKSTOP_TTL_SECONDS` to avoid exactly this trap.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  /** Decodes a staged envelope back into the job payload, resolving any offloaded blob. */
+  async decode({
+    value,
+    groupId,
+  }: {
+    value: string;
+    groupId: string;
+  }): Promise<Record<string, unknown>> {
+    const hold = readEnvelopeHold(value);
```
Aryansharma28: **[NON-BLOCKER]** **`decode()` parses the envelope twice on the worker hot path.**

`readEnvelopeHold(value)` followed by `decodeJobEnvelope({ value })` — each internally calls `splitEnvelope`, which `Buffer.from(value, 'utf8')`s the entire envelope and `JSON.parse`s the header. For a 32 KiB inline GQ2 body that's two ~32 KiB buffer allocations and two header parses per dispatch.

**Fix:** parse once at the lifecycle boundary and pass the pre-split `(header, body)` tuple through. Same shape applies to `release()` at L195, which calls `readEnvelopeHold` and then `readEnvelopeBlobId` per value — that's `splitEnvelope` twice per legacy/GQ1 value. With `maxBatch=10` coalesced siblings, that's ~20 redundant `Buffer.from + JSON.parse` per successful completion. Worth collapsing into a single `readEnvelopeRetirement(value)` helper that returns `{hold} | {blobId} | null` from one parse.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -2,47 +2,73 @@ import { randomUUID } from "node:crypto";
...
+ */
+const INLINE_CEILING_BYTES = 4 * 1024;
+
+/**
+ * Phase gate for the format rollout (ADR-026). Readers are always
+ * envelope-aware, but writes stay legacy bare JSON until every consumer in the
+ * fleet is known to read envelopes. Read at call time so tests can toggle it
+ * without module reloads.
  */
 function envelopeWritesEnabled(): boolean {
```
Aryansharma28, line 68: **[NON-BLOCKER]** **`envelopeWritesEnabled()` reads `process.env` at call time — at odds with the architecture doc's "no env-coupling" claim and the test surface bears the cost.**

The new `queues/groupQueue/ARCHITECTURE.md` describes dependencies as explicit constructor injections, but this single call-time `process.env` read sits on the encode hot path. Tests have to `vi.stubEnv` rather than instantiate a no-write lifecycle — and a partial fleet rollout (some workers re-read the env, others started before the flag flipped) puts both GQ1 and GQ2 formats in the same group's hash space until every pod has cycled.

**Fix (small):** make the policy a constructor arg threaded through `EnvelopeBlobLifecycle` → `encodeJobEnvelope`. The env-var read becomes the composition root's job, which is also where the rollout contract ("restart workers to flip the format") naturally lives.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -57,70 +83,262 @@ export interface JobRoutingMeta {
...
+    }
+    const payloadJson = JSON.stringify(payload);
+    const payloadBytes = Buffer.byteLength(payloadJson);
+    if (payloadBytes > INLINE_CEILING_BYTES) {
+      const ref = await tieredBlobs.put({
+        projectId,
+        data: await gzipAsync(payloadJson),
+        // Hash the RAW payload json, not the gzip output, so the dedup key is
+        // independent of gzip determinism (zlib version/level) — ADR-030 §1.
+        hashSource: Buffer.from(payloadJson, "utf8"),
```
Aryansharma28: **[NON-BLOCKER]** **`Buffer.from(payloadJson, 'utf8')` allocates a full-payload buffer just to feed `contentHash`.** `crypto.createHash().update()` accepts strings directly. For a 4 MiB fan-out payload that's 4 MiB of throw-away allocation per producer call on the encode hot path.

**Fix:** change `contentHash` (in `tieredBlobStore.ts`) to accept `Buffer | string`, and pass `payloadJson` through unchanged — `update(string)` handles UTF-8 internally with no intermediate Buffer.

Aryansharma28 (commented):
## Review — `/review` 8-angle pass

Reviewed via the `/review` skill at max effort: 8 finder angles in parallel (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions, tests+observability) then 1-vote verification. The `/code-review` skill targets correctness + reuse/simplification/efficiency — that surface is already covered by these angles, so I did not run it a second time to avoid redundant token spend; happy to layer a fresh pass if you want.

**Skipped:** the 21 `@unimplemented` scenarios in `specs/event-sourcing/payload-store-content-addressed.feature` and the `@track3` TTL-bounded edges in `payload-store-blob-hardening.feature` — author already flagged these under "Anything surprising" so I don't double-bill them as findings.

**Four blockers, eleven non-blockers** inline below. Two of the blockers anchor to files **not in this PR's diff**, so I'm calling them out here in prose:

### Out-of-diff blockers

**1. [BLOCKER] `langwatch/src/server/event-sourcing/queues/groupQueue/scripts.ts:381` — Lua `gqRoutingMeta` only recognizes the `GQ1|` prefix.** The dispatcher reads routing fields via `gqRoutingMeta(jobDataJson)`; for any `GQ2|...` envelope this PR introduces, the `string.sub(jobDataJson, 1, 4) == "GQ1|"` check at line 381 fails and the helper returns `nil, nil, nil`. Consequence once the flag flips: the pipeline-pause `SISMEMBER` checks in `DISPATCH_LUA` (~L757) and `DISPATCH_BATCH_LUA` (~L968) never fire, so a paused pipeline keeps dispatching, and the per-`__jobName` failure counter in `COMPLETE_LUA`'s failure path (~L1298) is dead for every retry exhaustion. Fix: `if string.sub(jobDataJson, 1, 4) == "GQ1|" or string.sub(jobDataJson, 1, 4) == "GQ2|"`. This is the highest-priority finding — without it, the GQ2 rollout silently disables pipeline pauses and Skynet failure tiles.

**2. [BLOCKER] `langwatch/src/server/app-layer/ops/repositories/queue.redis.repository.ts:559` — ops dashboard decodes without a `TieredBlobStore`.** `getGroupJobs` builds only a `RedisJobBlobStore` and calls `decodeJobEnvelope({ value: raw, blobs })` with no `tieredBlobs`. `decodeJobEnvelope` then throws `"Job envelope references a tiered blob but no tiered store was provided"`; the surrounding catch swallows it as "ignore undecodable values" and the row's `data` ends up `null`. Once the flag flips, every payload > 4 KiB encodes as `GQ2` and the Group Detail Dialog goes blind for the bulk of in-flight jobs. Fix: wire the `TieredBlobStore` into `queue.redis.repository.ts` (constructor injection or per-call resolve via `EnvelopeBlobLifecycle.objectStoreFor`).

### Tracked-but-not-posted (housekeeping)

These surfaced in angles H/D/I but are too small to spend a thread on — bundling here:

- **Positional params (CLAUDE.md convention)**: `groupQueue.ts:102` `assertNoReservedKeys(payload, queueName, method)`, `blobHolders.ts:90/94` `holderKey(projectId, hash)` / `blobKey(projectId, hash)`, `tieredBlobStore.ts:89` `streamToBuffer(stream, maxBytes)` — repo CLAUDE.md says "Use named parameters via object destructuring: `fn({ a, b })` not `fn(a, b)`". Public APIs in the same files already destructure; these are inconsistent private/internal outliers.
- **`gs://` redaction is a dead branch**: `project-storage-destination.ts` adds a `gs://` arm to `redactStorageUri` + `redactStorageUrisInText`, but `SUPPORTED_SCHEMES = ['s3', 'file', 'azure-blob']` and there is no `mintGsUri` / GCS driver — the platform never issues `gs://`, so this code can never fire. Either drop or land the matching scheme entry.
- **No tier-boundary equality tests**: `tieredBlobStore.unit.test.ts` and `jobEnvelope.unit.test.ts` exercise comfortably-over and comfortably-under but not `data.length === S3_TIER_THRESHOLD_BYTES` and not `payloadBytes === INLINE_CEILING_BYTES` — both checks use strict `>`, so a future `>=` flip silently shifts tier (cost shows up in the AWS bill, not a failing test).
- **No queue-level multi-reactor fan-out test**: encode-level dedup is covered (`jobEnvelope.unit.test.ts:531`) but the `EnvelopeBlobLifecycle.acquire / release / transfer` wiring on multi-reactor dispatch is not. Author admits this in the PR description; calling it out so it doesn't get lost in the follow-up backlog.

---

_Reviewed against `5af82bca` on 2026-06-24._

langwatch/src/server/event-sourcing/queues/groupQueue/blobConstants.ts
0xdeafcafe: Fixed in 306fa0b04 — restored the 7-day backstop for the GQ1 path. blobConstants now exports GQ1_BLOB_BACKSTOP_TTL_SECONDS = 7*24*60*60 (used by redisJobBlobStore); GQ2's 3-day BLOB_BACKSTOP_TTL_SECONDS stays as the refcounted/access-refreshed number. The doc comment explains why the two are different: GQ2 has an accompanying holder set with atomic reclaim (so TTL is a safety net, not primary reclaim), GQ1 has no refcount so TTL must comfortably exceed the longest plausible staged residence.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
0xdeafcafe: Fixed in 306fa0b04 — parseDrainedPayload now rethrows TransientBlobStoreError. The caller at the Promise.all site catches it, calls restageDrainedSiblings so the whole batch is re-staged, and routes the dispatched job through the SAME handleTransientDecode used by the direct decode path (line 580). A brief S3 blip during a batch coalesce now re-stages both the dispatched job and every sibling, rather than dropping siblings to replay while the dispatched job retries cleanly.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
0xdeafcafe: Fixed in 306fa0b04 (option 1 — the smaller fix): the fallback is now an ORDERED acquire-then-release, awaited on the acquire side. If acquire fails we skip release, so the old blob stays alive under its TTL instead of being reclaimed while the new hold hasn't landed. Not atomic (that's option 2, extending TRANSFER_LUA for the mixed-format case) — the code comment on that branch tracks the atomic version as a follow-up. During the GQ1→GQ2 cutover you now have crash-window safety in the common path.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
0xdeafcafe: Fixed in 306fa0b04 — new gqBlobReclaimS3FailuresTotal counter, incremented at BOTH s3 reclaim catch sites (the release path AND the transfer path), alongside the existing warn. The retry-worker follow-up (a gq:s3-reclaim SET drained by a background reclaim worker) is a bigger change tracked separately; this closes the observability gap so a bucket-lifecycle-only backstop can't silently degrade.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
0xdeafcafe: Fixed in 306fa0b04 — TieredBlobStore now takes queueName + logger via constructor. When BlobTooLargeError is caught the get() path emits a tenant-attributed warn ('possible tamper / zip-bomb') and increments gqBlobDecodeCapExceededTotal, then still returns null so the fail-safe runs. A real corruption event is now distinguishable from 'TTL reclaimed the blob' in both logs and metrics. Kept the coerce-to-null (rather than a new BlobCorruptError) so the higher-level decode path stays uniform — the observability signal is the counter, not the exception type.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
0xdeafcafe: Fixed in 306fa0b04 — assertPayloadWithinCap now takes an optional ctx { projectId, queueName, logger } and emits a tenant-attributed warn + gqPayloadTooLargeTotal counter at the cap boundary before the throw. The lifecycle threads projectId + queueName + logger through encodeJobEnvelope automatically, so producers get tenant-attributed observability with no per-caller work.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
0xdeafcafe: Fixed in 306fa0b04 — the else-branch that falls through to GQ1 now emits gqEnvelopeGQ2DowngradeTotal + a warn identifying which side is missing (hasTieredBlobs / hasProjectId), so a composition-root regression that drops the tenant prefix or omits the tiered store surfaces immediately in metrics. Didn't make projectId non-optional in this PR (bigger typing change and I want to keep the GQ1 path callable during rollout) — the counter is the alarm.

langwatch/src/server/event-sourcing/queues/groupQueue/redisJobBlobStore.ts
0xdeafcafe: Fixed in 306fa0b04 — RedisJobBlobStore now has get (GETEX, worker hot path) and peek (GET, no TTL refresh). Added peek to the JobBlobStore interface. decodeJobEnvelope gains a readMode: 'get' | 'peek' arg that routes to the right method. queue.redis.repository (getGroupJobs) now passes readMode: 'peek'. The GQ2 tiered store's s3 path already doesn't refresh TTL from the client (s3 objects have no TTL), and the redis-tier GETEX on the tieredBlobStore side goes through the same JobBlobStore interface — inspecting a GQ2 blocked group now doesn't refresh the redis-tier TTL either.

langwatch/src/server/event-sourcing/queues/groupQueue/tieredBlobStore.ts
0xdeafcafe: Fixed in 306fa0b04 — S3_TIER_THRESHOLD_BYTES now derives from COMMAND_INLINE_THRESHOLD (imported from ~/server/app-layer/traces/lean-for-projection), mirroring how BLOB_HOLDER_TTL_SECONDS derives from BLOB_BACKSTOP_TTL_SECONDS in blobConstants. A retune on one side now moves the other in lockstep.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
0xdeafcafe: Fixed in 306fa0b04 — added readEnvelopeHoldFromHeader + readEnvelopeRetirement helpers so callers with an already-parsed header don't pay for a second splitEnvelope. lifecycle.decode now calls splitEnvelope once at the boundary and passes the parsed { header, body } tuple into decodeJobEnvelope via a new 'parsed' arg. lifecycle.release swaps its readEnvelopeHold + readEnvelopeBlobId pair for one readEnvelopeRetirement call — at maxBatch=10 that removes ~20 redundant Buffer.from + JSON.parse per successful completion.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
0xdeafcafe: Fixed in 306fa0b04 — encodeJobEnvelope now accepts an explicit writesEnabled: boolean, threaded from the EnvelopeBlobLifecycle constructor. Env var read stays as the fallback for tests (so we don't have to churn every existing test). Composition roots that pass writesEnabled explicitly get the injected contract the architecture doc claims; existing callers keep their behaviour without a breaking change. The rollout observation about 'restart workers to flip the format' is now enforced by the DI shape, not by env-var timing.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
0xdeafcafe: Fixed in 306fa0b04 — contentHash now accepts Buffer | string, hashSource accepts Buffer | string too, and encodeJobEnvelope passes payloadJson directly. crypto.update handles the UTF-8 conversion internally. For a 4 MiB fan-out payload that's 4 MiB of throw-away allocation removed per producer call on the encode hot path.

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** A content-addressed refcount/reclaim bug reclaims a blob still referenced by a live reactor job, so a worker's decode throws and the payload is lost (or a cross-tenant key collision leaks data). Flag-gated (`GROUP_QUEUE_ENVELOPE_WRITES_ENABLED` off), which lowers immediate blast radius but this is the review to catch it.

- `[Must Check]` `blobHolders.ts` + `envelopeBlobLifecycle.ts` — holder-set refcount: last-release reclaim is atomic (Lua) and hold transfer on retry/squash can't double-free or drop the last holder; the dedup-squash overwrite path reclaims the displaced blob (the exact leak fixed here vs the saas cleanup tool).
- `[Must Check]` `blobKeys.ts` / `tieredBlobStore.ts` — key is `{projectId}/{sha256}` and reads re-derive location from `(projectId, contentHash)` (not a trusted stored URI); confirm no cross-tenant decode/release/transfer, and the cluster hash-tag slot guard holds.
- `[Ask Author]` 4813-line diff mixes the store, ADR-030 hardening, an extraction refactor, and a full docs rewrite — confirm the 45 unit + 44 integration tests actually ran green (checkboxes unchecked in the description).

- Old pods "drop unparseable GQ2 values and recover via event replay" — during a rolling deploy with the flag on, is replay guaranteed idempotent for every reactor, or can a dropped GQ2 value double-process?
- The 50 MiB cap and gunzip decompression cap: what happens to a legitimately-over-cap payload today, hard fail or fall back to inline, and is that observable?

sergioestebance:
Blessed. 🙏

_"Gather up the fragments that remain, that nothing be lost."_ — John 6:12

Content-addressed, holder-set reclaimed, nothing leaked. Ship it — but land the S3 bucket-lifecycle backstop before flipping the flag with the s3 tier (s3 has no TTL).

sergioestebance (dismissed):
One operational gate: land the S3 bucket-lifecycle backstop before flipping the flag with the s3 tier (s3 objects have no TTL). Three P3 polish items noted in the review comment, none blocking.

sergioestebance (approved):

PR #5000, opened 2026-06-20
perf(clickhouse): index gaps, quantileTDigest opt-in, eight repo audit fixes

0xdeafcafe:
## Summary

Multi-agent audit pass on CH repos and analytics, surfaced quite some cheap wins. Validated each against prod EXPLAIN before landing - which killed two of my own bigger ideas (projections, scalar-subquery anti-pattern rewrites) once the prod data showed they wouldn't help on this CH line.

### Migration 00034 - indexes that should already have existed

- `trace_summaries.idx_occurred_at` (minmax). Existing schema indexed `CreatedAt` but every analytics query filters on `OccurredAt`. Free granule pruning we were leaving on the table.
- `stored_log_records` and `stored_metric_records` get bloom_filter parity on `ResourceAttributes`. `stored_spans` already had it; consistency for free.

### quantileTDigest opt-in for performance.* and spans.metrics.*

New `PercentileMode` type, defaults to `"exact"`. Performance / cost / token-count metrics opt into tdigest. Evaluation scores and sentiment stay exact - their distributions are narrow enough that ±5% at the tail can flip threshold-based dashboard outcomes. `quantileExact`'s O(N) memory was a top OOM driver on heavy analytics reads.

### Repo audit fixes (8 sites)

- `event_log.findAggregates` was filtering on `EventTimestamp` but the partition key derives from `EventOccurredAt`. Silently scanning every weekly partition incl. cold S3.
- `event_log.searchAggregates` would unbound-scan with empty `tenantIds` + empty query string. Now requires at least one.
- `event_log.findEventsByAggregate` accepts an optional `sinceMs` to bound the heavy `EventPayload` scan.
- `log-record-storage` heavy `Attributes` map filter moved out of the dedup subquery; unbounded fallback replaced with a CC-retention-bounded window.
- `stored_objects` hot path switched from `findBySha256` (scan-all-partitions on `sha256`) to `findById` using the deterministic id we already compute. `findAllByProject` correlated subquery rewritten as IN-tuple.
- `suite-run.getBatchHistory` was missing IN-tuple dedup entirely - the page could fill with multiple versions of the same `BatchRunId`. Now it dedups.
- `dspy-step.getStepsByExperiment` accepts an optional `sinceMs`.

### What I left out (and why)

- **No projection migration**. Prod EXPLAIN on `stored_spans` shows the existing index funnel (`TenantId` → `idx_trace_id` → `idx_span_id` → `idx_tenant_trace`) already prunes single-row reads to 1 part / 1 granule. Trace-list dedup gets to 6 granules out of 12,468 via partition + time + IN-set + skip indexes. There's no dedup tax to optimise away - version churn per key in prod is essentially zero in the hour we sampled. Adding projections would add storage + `rebuild`-mode merge cost (CH 25 blocks projections on ReplacingMergeTree unless `deduplicate_merge_projection_mode` is changed) for no real read win.
- **No scalar-subquery dedup "fixes" on `getSpanByIds` / `getSpanEvents`**. Prod EXPLAIN confirmed CH 25.10's `LazilyRead` defers heavy columns past LIMIT, so the doc's "Anti-Pattern 1" form is actually the cheaper one for single-row reads on this CH line. Both stay on `ORDER BY <version> DESC LIMIT 1` with comments citing LazilyRead.
- **No `virtual_key_dict` pilot**. The audit agent flagged an N+1 that wasn't - the existing code does one batched `findMany`, not per-row hydration. Dict pattern still worth considering for `evaluator_dict` later (real correctness story there), but not in this PR.

## Test plan

- [ ] Migration 00034 runs cleanly on a staging snapshot; `MATERIALIZE INDEX` background mutations complete.
- [ ] Existing analytics queries return correct results (numbers match pre-migration on a known query for a busy tenant).
- [ ] Spot-check a `performance.completion_time` p95 panel - value within ±5% of the previous `quantileExact` value.
- [ ] Evaluation `quantileExactIf` reads stay on exact (regression-checked by the new `metric-translator.test.ts` case).
- [ ] Suite-run batch history no longer shows duplicate `BatchRunId` rows for ScenarioSets with multiple updates.
- [ ] Stored-objects ingest still dedups (unit tests pass; staging upload of the same bytes twice surfaces the existing id).
- [ ] Event-log ops UI: empty tenant filter + empty query string now throws clearly instead of returning a partial result.

## Release Notes

* **New Features**
  * Added optional `sinceMs` support for event-history reads, bounded when provided and unbounded when omitted.

* **Bug Fixes**
  * Prevented unmerged/repeated entries in suite batch history via deduping.
  * Improved stored-object deduplication by probing with deterministic content ids (no sha-based probing).
  * Ensured bounded retention behavior for marked Claude code logs when time isn’t provided.
  * Added guards to avoid full event log scans unless inputs are sufficiently constrained.

* **Performance**
  * Switched `performance.*` percentile calculations to tdigest quantiles.
  * Improved ClickHouse pruning and added query skipping indexes.

* **Tests & Maintenance**
  * Updated SQL-generation and percentile/memory-safety assertions to match new behavior.

### DIFF
--- a/langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
+++ b/langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
@@ -5,6 +5,10 @@ export interface DspyStepRepository {
   getStepsByExperiment(
     tenantId: string,
     experimentId: string,
+    // Optional CreatedAt lower bound. Pass when the caller knows the
+    // experiment's start window so the scan can prune partitions instead of
+    // walking every weekly partition incl. cold S3.
+    options?: { sinceMs?: number },
   ): Promise<DspyStepSummaryData[]>;
   getStep(
     tenantId: string,
--- a/langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
+++ b/langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
@@ -5,10 +5,6 @@ export interface DspyStepRepository {
   getStepsByExperiment(
     tenantId: string,
     experimentId: string,
-    // Optional CreatedAt lower bound. Pass when the caller knows the
-    // experiment's start window so the scan can prune partitions instead of
-    // walking every weekly partition incl. cold S3.
-    options?: { sinceMs?: number },
   ): Promise<DspyStepSummaryData[]>;
   getStep(
     tenantId: string,
@@ -16,21 +12,22 @@ export interface DspyStepRepository {
     runId: string,
     stepIndex: string,
   ): Promise<DspyStepData | null>;
-  deleteByExperiment(
-    tenantId: string,
-    experimentId: string,
-  ): Promise<void>;
+  deleteByExperiment(tenantId: string, experimentId: string): Promise<void>;
 }
 
 export class NullDspyStepRepository implements DspyStepRepository {
-  async upsertStep(): Promise<void> {}
+  async upsertStep(): Promise<void> {
+    // no-op: null repository
+  }
   async getStepsByExperiment(): Promise<DspyStepSummaryData[]> {
     return [];
   }
   async getStep(): Promise<DspyStepData | null> {
     return null;
   }
-  async deleteByExperiment(): Promise<void> {}
+  async deleteByExperiment(): Promise<void> {
+    // no-op: null repository
+  }
 }
 
 function mergeByHash<T extends { hash: string }>(
@@ -107,7 +104,10 @@ export class InMemoryDspyStepRepository implements DspyStepRepository {
     runId: string,
     stepIndex: string,
   ): Promise<DspyStepData | null> {
-    return this.store.get(`${tenantId}/${experimentId}/${runId}/${stepIndex}`) ?? null;
+    return (
+      this.store.get(`${tenantId}/${experimentId}/${runId}/${stepIndex}`) ??
+      null
+    );
   }
 
   async deleteByExperiment(
--- a/langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
+++ b/langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
@@ -82,6 +82,7 @@ export class EventExplorerClickHouseRepository
   async searchAggregates(params: {
     query: string;
     tenantIds?: string[];
+    sinceMs?: number;
   }): Promise<AggregateSearchResult[]> {
     // Defensive guard: without at least one of (tenants, query string) the
     // query degrades to "ORDER BY EventTimestamp DESC LIMIT 50" against the
@@ -102,16 +103,21 @@ export class EventExplorerClickHouseRepository
       );
     }
 
-    // No silent time clamp here - DejaView is the internal ops admin tool, and
-    // operators searching for old aggregates during incident archaeology
-    // mustn't get empty results with no explanation. The hot/cold-tier bound
-    // is surfaced *in the DejaView UI* (uses the same env-var-derived value
-    // the TTL reconciler does) so the operator sees "older than N days lives
-    // in cold storage" up front, rather than the backend silently dropping
-    // rows. Tradeoff: a truly unbounded cross-tenant scan stays expensive,
-    // but the upfront guard above (must supply tenants or a query string)
-    // already prevents the worst shape.
+    // No silent time clamp in the repo. The caller (the ops router for the
+    // DejaView UI) supplies sinceMs explicitly - currently a 1-year bound
+    // surfaced under the search box as a banner so the operator sees the
+    // window up front. Other callers (e.g. integration tests, ad-hoc
+    // scripts) can pass `undefined` to scan the full event_log, paying the
+    // partition-fan-out cost knowingly. EventOccurredAt = 0 is the legacy
+    // sentinel; preserved alongside the bound so historical test data
+    // doesn't silently disappear.
     const queryParams: Record<string, unknown> = {};
+    let timeBoundFilter = "";
+    if (typeof params.sinceMs === "number" && params.sinceMs > 0) {
+      timeBoundFilter =
+        "AND (EventOccurredAt = 0 OR EventOccurredAt >= {sinceMs:UInt64})";
+      queryParams.sinceMs = params.sinceMs;
+    }
 
     let tenantFilter = "";
     if (hasTenants) {
@@ -136,6 +142,7 @@ export class EventExplorerClickHouseRepository
           max(EventTimestamp) AS lastEventTime
         FROM event_log
         WHERE 1=1
+          ${timeBoundFilter}
           ${tenantFilter}
           ${aggregateFilter}
         GROUP BY AggregateId, AggregateType, TenantId
@@ -167,29 +174,14 @@ export class EventExplorerClickHouseRepository
     aggregateId: string;
     tenantId: string;
     limit: number;
-    sinceMs?: number;
   }): Promise<RawEventRow[]> {
-    // Heavy column read (EventPayload, ZSTD(3)). Without an EventOccurredAt
-    // bound the read walks every weekly partition for the aggregate — the
-    // primary key seeks to (TenantId, AggregateType, AggregateId, …) per
-    // partition but pays cold-S3 metadata fetches for the inactive ones.
-    //
-    // Caller semantics:
-    //   * Routine event-listing (ops UI) — pass sinceMs (e.g. last 30 days)
-    //     to bound the scan to warm partitions only.
-    //   * Projection replay — explicitly omit sinceMs to read the full
-    //     event history of an aggregate (necessary for fold correctness).
-    const queryParams: Record<string, unknown> = {
-      tenantId: params.tenantId,
-      aggregateId: params.aggregateId,
-      limit: params.limit,
-    };
-    let timeFilter = "";
-    if (typeof params.sinceMs === "number" && params.sinceMs > 0) {
-      timeFilter = "AND EventOccurredAt >= {sinceMs:UInt64}";
-      queryParams.sinceMs = params.sinceMs;
-    }
-
+    // No `sinceMs` parameter: this is the detail-view query for a
+    // specific aggregate the operator has already picked from the
+    // (bounded) DejaView search. Once you're looking at one aggregate
+    // you want its full event history, including projection replays
+    // where the fold depends on every event. The partition fan-out is
+    // acceptable here because the aggregate-id seek is narrow per
+    // partition.
     const result = await this.client.query({
       query: `
         SELECT
@@ -200,11 +192,14 @@ export class EventExplorerClickHouseRepository
         FROM event_log
         WHERE TenantId = {tenantId:String}
           AND AggregateId = {aggregateId:String}
-          ${timeFilter}
         ORDER BY EventTimestamp ASC, EventId ASC
         LIMIT {limit:UInt32}
       `,
-      query_params: queryParams,
+      query_params: {
+        tenantId: params.tenantId,
+        aggregateId: params.aggregateId,
+        limit: params.limit,
+      },
       format: "JSONEachRow",
     });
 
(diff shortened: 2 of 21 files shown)


langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
```diff
@@ -42,7 +49,7 @@ export class EventExplorerClickHouseRepository
           count(DISTINCT AggregateId) AS aggregateCount
         FROM event_log
         WHERE AggregateType IN ({aggregateTypes:Array(String)})
-          AND EventTimestamp >= {sinceMs:UInt64}
+          AND EventOccurredAt >= {sinceMs:UInt64}
```
langwatch-agent: **P1** — This drops every legacy event whose `EventOccurredAt` is the schema default `0`. Those rows are real: the event store explicitly preserves the sentinel with `(EventOccurredAt = 0 OR EventOccurredAt >= ...)`, because `EventOccurredAt` was added after historical events already existed. `discoverAggregates` feeds the bulk replay wizard, so affected aggregates silently disappear and cannot be selected for replay. Preserve the zero sentinel while retaining partition pruning, and add a regression case with `EventOccurredAt = 0`.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
```diff
@@ -92,7 +123,7 @@ export class EventExplorerClickHouseRepository
           count() AS eventCount,
           max(EventTimestamp) AS lastEventTime
         FROM event_log
-        WHERE 1 = 1
+        WHERE EventOccurredAt >= {sinceMs:UInt64}
```
langwatch-agent: **P2** — The repository now silently makes aggregates older than 90 days unsearchable, but neither this API nor the ops UI exposes a `sinceMs` override. Searching by a known tenant/aggregate ID can therefore return no result even though the aggregate still exists, blocking historical incident investigation. Thread an explicit lookback through the service/router/UI (or avoid imposing this clamp when the caller supplied a constraining tenant/query) so the performance bound does not remove existing functionality.

langwatch-agent (dismissed):
1 P1, 1 P2 — ClickHouse optimizations validate, but event-explorer time bounds hide legacy and historical aggregates. See inline comments.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
0xdeafcafe: I only see 2 on ClickHouse with `EventOccurredAt` being 0, and they're both me testing stuff day 1. Is this that important?

langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
```diff
@@ -5,6 +5,10 @@ export interface DspyStepRepository {
   getStepsByExperiment(
     tenantId: string,
     experimentId: string,
+    // Optional CreatedAt lower bound. Pass when the caller knows the
+    // experiment's start window so the scan can prune partitions instead of
+    // walking every weekly partition incl. cold S3.
+    options?: { sinceMs?: number },
```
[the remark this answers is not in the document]
0xdeafcafe: Stale - signature was simplified by dropping the third positional arg in dcbafcd9e, so getStepsByExperiment is back to (tenantId, experimentId). Two args reads fine positionally; the named-parameters rule kicks in past three.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
```diff
@@ -124,7 +152,29 @@ export class EventExplorerClickHouseRepository
...
+      tenantId: params.tenantId,
+      aggregateId: params.aggregateId,
+      limit: params.limit,
+    };
+    let timeFilter = "";
+    if (typeof params.sinceMs === "number" && params.sinceMs > 0) {
+      timeFilter = "AND EventOccurredAt >= {sinceMs:UInt64}";
+      queryParams.sinceMs = params.sinceMs;
+    }
+
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 02551ed3e. findEventsByAggregate had a sinceMs option that no caller used; dropped it (the detail view wants full history). For the search shape, the ops router now supplies an explicit sinceMs (1-year default) and DejaView renders a banner under the search box stating the bound + the hot-tier window. Visible bounds, not silent ones.

langwatch/src/server/app-layer/ops/repositories/event-explorer.repository.ts
```diff
@@ -35,6 +35,9 @@ export interface EventExplorerRepository {
     aggregateId: string;
     tenantId: string;
     limit: number;
+    // Optional EventOccurredAt lower bound. Pass for routine ops UI reads;
+    // omit only for projection replay where the full event history matters.
+    sinceMs?: number;
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 02551ed3e. The 'routine ops UI passes sinceMs' comment was aspirational and rotted - rewrote it. The interface now only carries sinceMs on searchAggregates (the surface that actually has a router-side default), and the inline doc matches what the code does.

langwatch/src/server/stored-objects/stored-objects.service.ts
```diff
@@ -153,8 +153,15 @@ export class StoredObjectsService {
...
-        const existing = await this.repository.findBySha256({ projectId, sha256 });
+        // Dedup probe: if content already present, skip PUT + INSERT.
+        // Lookup by id (not sha256) because:
+        //   1. id is derived deterministically from (projectId, sha256) right
+        //      above, so it's already known here — no extra computation.
+        //   2. The stored_objects table's `ORDER BY (project_id, id)` makes
+        //      this a primary-key seek with partition pruning; a sha256
+        //      lookup would scan every weekly partition incl. cold S3 because
+        //      sha256 is not in the sort key.
+        const existing = await this.repository.findById({ projectId, id });
```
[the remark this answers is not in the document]
0xdeafcafe: The id is deterministic: deriveStoredObjectId({projectId, sha256}) computes the same id for the same (projectId, sha256) pair every time (it's a KSUID derived from a sha1 of the inputs). Same project + same content bytes = same id, period. No collision risk by construction; an unrelated row colliding would require either two distinct sha256s producing the same derived id (sha1 collision) or two distinct project IDs hashing identically (same hash, again). The dedup probe is just findById(deriveId(project, sha256)); locked in stored-objects.service.unit.test.ts:'dedup probe goes through findById' in 183c54a88.

langwatch/src/server/app-layer/dspy-steps/repositories/dspy-step.repository.ts
```diff
@@ -12,21 +12,22 @@ export interface DspyStepRepository {
...
-    experimentId: string,
-  ): Promise<void>;
+  deleteByExperiment(tenantId: string, experimentId: string): Promise<void>;
 }
 
 export class NullDspyStepRepository implements DspyStepRepository {
-  async upsertStep(): Promise<void> {}
+  async upsertStep(): Promise<void> {
+    // no-op: null repository
+  }
```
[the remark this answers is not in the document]
0xdeafcafe: Biome's noEmptyBlockStatements rule (enforced by CI) requires non-empty content inside the empty body. The '// no-op: null repository' comment IS the marker - removing it fails lint, replacing it with rationale ('this is the null-object pattern, the methods intentionally do nothing') reads as more LLM-padding than the curt 'no-op' line. Happy to swap if there's a preferred convention elsewhere in the repo, but this is the minimum that satisfies lint.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
0xdeafcafe: Addressed in 183c54a88. findAggregates now uses (EventOccurredAt = 0 OR EventOccurredAt >= sinceMs), so the legacy zero-sentinel rows survive while partition pruning still works (epoch-week + last-N-weeks instead of every partition). Regression test added in the same commit. See Alex's separate comment about whether this is worth keeping given the tiny prod row count - my answer there too.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
0xdeafcafe: Addressed in 183c54a88 then refined in 02551ed3e. First pass: removed the silent 90d clamp from the repo (the wrong layer). Second pass: ops router now supplies an explicit 1-year sinceMs and the DejaView UI renders a banner under the search box stating both the search bound and the hot-tier window. Older-than-bound aggregates surface in the banner copy rather than silently disappearing.

langwatch/src/server/app-layer/ops/repositories/event-explorer.clickhouse.repository.ts
0xdeafcafe: Honest answer: it's pretty cheap insurance, but if it's only your day-1 test rows, ripping it out is fine too.

The cost is one OR clause + the regression test. The benefit is that ANY future migration introducing a zero default (e.g. backfilling a new column, dual-write from a different source, etc.) won't silently disappear from the bulk-replay wizard. If we're confident the only zero rows are throwaway and we're happy to manually fix it the next time it bites us, drop it - happy to revert in a follow-up commit. If we want it as a guard against the next time we accidentally introduce a sentinel value, leave it.

Slight lean toward keeping just because the cost is so low, but it's your call.

langwatch-agent (commented):
No blocking findings on re-review. Both prior findings are resolved: legacy EventOccurredAt=0 rows are preserved, and DejaView now has an explicit, visible 365-day default with a caller-supplied sinceMs override. The suite-run alias fix and locked LazilyRead setting are also coherent. Validation: 92 focused unit tests, 7 span-storage integration tests, 62 real-ClickHouse memory-safety tests (3 skipped), typecheck, and git diff --check. Residual risk: the DejaView UI itself still uses the fixed 365-day window; older indefinite-retention data requires calling the existing API override rather than changing the window in the UI.

sergioestebance:
Reviewed — solid work, validated against prod EXPLAIN and the tests are good. Two things before merge, neither blocking:

**1. `getMarkedClaudeCodeLogsByTrace` fallback now drops future-timestamped logs (minor regression)**

The no-hint fallback changed from *unbounded* to `[now − 7d, now]`. The upper cap at `now` has no clock-skew headroom, unlike the hint path which already allows `occurredAt + 2d`. `TimeUnixMs` is client-supplied, so a fast client clock writes a slightly-future timestamp → the re-fold silently skips that log. Previously the unbounded fallback returned it.

Narrow blast radius (only the `occurredAt = 0` fallback path + a leading client clock), but it's a behavioral tightening. Free fix — give the fallback the same headroom the hint path already has:

\`\`\`js
const toMs = hasWindow ? occurredAtMs + partitionWindowMs : now + partitionWindowMs;
\`\`\`

(Lower bound is safe: CC logs carry `_retention_days = 1` and the retention DELETE TTL is anchored on `TimeUnixMs`, so a live CC row is always well inside the 7d window.)

**2. Migration 00034 `MATERIALIZE INDEX` — deploy runbook note**

Per the migration's own comment, materializing the index over `stored_log_records` / `stored_metric_records` reads existing parts — including cold-tier S3 ones — back through compute, so expect merge backlog on these high-volume tables. We haven't measured prod cold-part volume for them. Worth checking first, or shipping `ADD INDEX` only and skipping the backfill (new parts still get the bloom filter; historic ResourceAttributes lookups are rare).

sergioestebance (dismissed):
blessed

0xdeafcafe:
Both folded in - thanks for the catch on the fast-clock case.

**1.** 1ded2fb5d: `getMarkedClaudeCodeLogsByTrace` fallback upper bound now `now + partitionWindowMs` (mirrors the hint path's ±2d headroom). Test updated to lock the shape - `toMs ∈ [before + 2d, after + 2d]`, `toMs - fromMs == 7×CC_RETENTION + 2d`.

**2.** Same commit: dropped `MATERIALIZE INDEX` on stored_log_records / stored_metric_records, kept on trace_summaries (small table, analytics queries need historic backfill). Comment block now documents the asymmetry + flags the follow-up path if a workload genuinely needs the backfill later. Agree with your read - go-forward only is the right default until we've measured cold-part volume on those two tables.

sergioestebance (approved):
re-blessed

Issue #4744, opened 2026-06-11
generateOpenAPISpec crashes on main: zod-openapi circular-ref in prompts parameters schema


sergioestebance:
## What

`pnpm task generateOpenAPISpec` crashes on main, so the OpenAPI spec (`langwatch/src/app/api/openapiLangWatch.json`) cannot be regenerated. Any PR adding/changing API routes has to hand-edit the generated JSON (and rebases conflict in it, e.g. #4714 vs #3668).

## Repro (main, fresh worktree)

```bash
cd langwatch
pnpm install && pnpm start:prepare:files
pnpm task generateOpenAPISpec
```

```
Error: The schema at createSchema > array items > property: parameters > default > record value > lazy schema > union option 4 > array items needs to be registered because it's circularly referenced
    at createNewSchema (.../zod-openapi@4.2.4_zod@3.25.76/dist/components.chunk.cjs:1449)
```

## Cause

`zod-openapi@4.2.4` requires circularly-referenced `z.lazy()` schemas to be registered with an explicit `ref`. The crash points at a `parameters` property whose default record value contains a lazy union with a circular array item — the prompts/llm-configs schema shape introduced around #4283.

## Fix direction

Register the circular schema with `.openapi({ ref: "..." })` (or a components registration) so zod-openapi can emit a `$ref` instead of inlining forever. Related but distinct: #2987 (generator leaves stale paths).

PR #5005, opened 2026-06-21
fix(openapi): prune orphaned paths when routes are removed/renamed (#2987)

### LINKED ISSUE #2987: fix(openapi): generator leaves stale paths in spec when routes are removed/renamed
## Summary

`langwatch/src/tasks/generateOpenAPISpec.ts` deep-merges freshly-generated app specs on top of the previously committed `openapiLangWatch.json`. The `customMerge` strategy at lines 66–83 only fires for path keys that exist in the *source* (the new app spec), so it can replace colliding paths but cannot prune orphans. When a route is removed or its path parameter is renamed, the old path lingers in the committed spec forever.

The header comment at lines 27–33 even acknowledges this:

> "It will always update the current spec with new endpoints, so deleting endpoints needs to be done manually from the the original file."

Two confirmed orphans currently in the spec under `/api/prompts`:

- `DELETE /api/prompts/tags/{tagId}` — the route was renamed to `/tags/:tag`; only the `{tag}` variant exists in `app.v1.ts`
- `PUT /api/prompts/{id}/labels/{label}` — the entire `/labels/*` route family was removed; no `labels` references remain in `app.v1.ts`

Both leak into generated SDK clients (Python, TS) as ghost endpoints that hit nothing on the server.

The intended behavior is that namespaces owned entirely by a Hono app (`/api/analytics`, `/api/dataset`, `/api/evaluators`, `/api/prompts`, `/api/scenarios`, `/api/scenario-events`, `/api/traces`) should be 1:1 with the app source — the app is the single source of truth.

## Proposed fix

Before `deepmerge.all(...)`, strip any path under an owned prefix from `currentSpec.paths`. The apps then repopulate those prefixes fresh, and orphans disappear automatically. The `customMerge` callback becomes redundant and can be removed.

No manual edits to `openapiLangWatch.json` — the fix lives entirely in the generator.

## Verification

After the fix, regenerating the spec on `main` should produce a diff that removes exactly the two orphan entries above (plus any other dead paths under owned prefixes that surface). Reviewers should sanity-check the diff before merging.

---
*Classification: Bug*

drewdrewthis:
## Why

Closes #2987

`langwatch/src/tasks/generateOpenAPISpec.ts` deep-merged freshly-generated per-app OpenAPI specs on top of the committed `openapiLangWatch.json`. The old `customMerge` strategy only fired for path keys present in the *new* app spec, so it could replace colliding paths but never **prune orphans**. When a route was removed or its path parameter renamed, the stale path lingered in the committed spec forever and leaked into the generated SDK clients (Python, TS) as ghost endpoints that hit nothing on the server.

## What changed

Merge/prune logic moved to a new dependency-light module `src/tasks/mergeOpenAPISpecs.ts` (imports only `deepmerge`, so it is unit-testable without booting the 20+ Hono apps). On every run the generator now:

- derives the set of app-owned `/api/<namespace>` prefixes **from the freshly-generated app specs** (never a hand-maintained list — it cannot drift),
- strips those namespaces out of the committed spec, then
- repopulates them from the apps.

Result: routes an app no longer generates never come back (**orphans pruned**); routes that still exist are written fresh via a clean replace (no stale sub-keys survive); paths in namespaces **no** app generates — today `/api/annotations`, the singular `/api/trace`, and the root `/` — are **preserved untouched**. The redundant `customMerge` callback and its hand-maintained prefix list are removed.

### Derived ownership widens the prune boundary by 3 namespaces — intended, no-op today

The new prune set is **not identical** to the `APP_DERIVED_PREFIXES` list it replaces, and the difference is the point of the change rather than a side effect of the rebase.

`/api/model-defaults`, `/api/teams` and `/api/groups` were never in that list. Main only deep-merged them additively, so a route removed or renamed under them left an orphan behind forever — the exact #2987 bug, in the three namespaces the hand list had drifted past. Their specs (`modelDefaultsSpec`, `teamsSpec`, `groupsSpec`) are in `appSpecs`, so under derived ownership they are owned, and therefore prunable, like every other app-backed namespace. Deriving ownership from output is what makes "app-backed but silently exempt from pruning" unrepresentable.

**Nothing changes in the committed spec today.** All 24 committed paths across those three namespaces still match a live handler (`teams/[[...route]]/app.ts`, `groups/[[...route]]/app.ts`, `model-defaults/[[...route]]/app.v1.ts`), so there is nothing to prune. `pnpm check:openapi-route-coverage` passes at HEAD — `210/317 registered routes are in the document ... OK: every registered route is documented or accounted for` — and that gate is the standing guard: if a prune ever dropped a live route's entry from any owned namespace, it goes red.

The unit suite pins this with `/api/teams` as the exemplar: the same committed paths prune with the teams spec in `appSpecs` and survive without it. It names one namespace rather than asserting a whole set, because an enumerated set would be the hand-maintained list whose drift is #2987.

## Approach — idempotent generation, minimal merge (re: the correctness challenge)

The challenge ([comment](https://github.com/langwatch/langwatch/pull/5005#issuecomment-4777947201)) was: *make generation idempotent / derive the whole spec correctly so it doesn't need complex merging logic, and do a full regen to prove nothing breaks.*

What the live regen below shows:

1. **Generation is already idempotent.** Feeding the output back in produces a byte-identical file (same sha256) — a fixed point in one application. The "idempotent" requirement is met by the shipped code.
2. **A pure from-scratch derive (drop the merge) is infeasible AND unsafe here, for two structural reasons:**
   - **One app cannot generate.** `generateSpecs(promptsApp)` crashes with a pre-existing `zod-openapi@4.2.4` circular-reference error *upstream of any merge* (tracked in #4744). A from-scratch derive needs all 21 apps; with prompts crashing it would **drop all 8 `/api/prompts/*` paths** — ghost-removing live SDK endpoints. The merge's "preserve a namespace no app emits this run" rule is exactly the fail-safe that stops a transient generation failure from wiping live routes. The regen below proves this: **0 paths dropped even though prompts crashes.** Removing the merge would do the opposite of "don't break anything."
   - **Three namespaces are hand-maintained**, emitted by no Hono app. A pure derive would need them moved into an explicit base source — a larger source-of-truth refactor that itself can't be fully regen-proven until #4744 is fixed.
3. **The merge is not "complex":** ~30 lines (derive owned namespaces → strip → `deepmerge.all`). Idempotency is proven; there is nothing for a from-scratch derive to simplify.

So the minimal merge stays, backed by the full-regen proof. The only thing blocking a true 21/21 regen is the unrelated prompts crash (#4744).

**Backend-only — no UI surface.** This is build-time OpenAPI tooling; there is nothing to screenshot. The proof is the generator + the shipped `mergeOpenAPISpecs` exercised against **real, freshly-generated app specs** on this branch (HEAD `dc9a79da8`).

> ⚠️ **The transcript below predates the rebase — read the conclusions, not the numbers.** It was measured at `dc9a79da8`, when the generator merged 21 app specs and the committed spec held 82 paths. HEAD is rebased onto a main that has since grown the generator to 32 app specs and the spec to 139 paths, so its app list, path counts and namespace lists reflect the old app set — in particular `/api/api-keys`, `/api/experiments` and `/api/projects` are app-owned now, not hand-maintained. The idempotency, orphan-pruning and zero-paths-dropped *conclusions* are properties of `mergeOpenAPISpecs`, which the rebase left byte-unchanged and the unit suite re-verifies at HEAD. A fresh full regen still waits on #4744.

### 1. Full regen against real app specs (20/21 apps; prompts blocked by #4744)

Ran the real `generateSpecs(app)` for all 21 apps under `try/catch`, fed the successes into the **exported** `mergeOpenAPISpecs` (the actual shipped function, imported — not reimplemented), serialized with the same `JSON.stringify(spec, null, 2)` the generator uses:

```
=== APP GENERATION ===
succeeded 20/21: agents, analytics, dashboards, dataset, evaluators, events,
  gateway-platform, governance, graphs, model-defaults, model-providers, monitors,
  scenario-events, scenarios, secrets, simulation-runs, suites, traces, triggers, workflows
FAILED  prompts: The schema at createSchema > array items > property: parameters >
  default > record value > lazy schema > union option 4 > array items needs to be
  registered because it's circularly referenced          # pre-existing, upstream of merge → #4744

=== IDEMPOTENCY (fixed point) ===
sha256(regen1): f917202d2020d33d53383a041e970f8d9bed15aaee61f0a91ad76e9cdc572a6f
sha256(regen2): f917202d2020d33d53383a041e970f8d9bed15aaee61f0a91ad76e9cdc572a6f
idempotent (regen1==regen2): true                        # fixed point in ONE application

=== ORPHAN PRUNING ===
orphan injected: /api/agents/__orphan_probe_pr5005__
present in INPUT : true
present in OUTPUT: false (false = pruned)                 # orphan dropped
regen(committed+orphan)==regen(committed): true           # orphan left no trace

=== HAND-MAINTAINED NAMESPACES PRESERVED ===
  PRESENT  /                 (owned-by-app=false)
  PRESENT  /api/annotations  (owned-by-app=false)
  PRESENT  /api/api-keys     (owned-by-app=false)
  PRESENT  /api/experiments  (owned-by-app=false)
  PRESENT  /api/projects     (owned-by-app=false)
  PRESENT  /api/trace        (owned-by-app=false)

=== PATH DELTA (committed -> regen1) ===
paths: 82 -> 83
dropped: 0   []                                           # NOTHING BREAKS — incl. the crash-blocked /api/prompts/* (8 paths preserved)
added:   1   [ '/api/workflows/{id}/evaluate' ]           # a real route the committed spec was missing (pre-existing drift)
```

Read this delta: **0 dropped** is the headline — even with prompts crashing, its paths survive via the fail-safe, so the generator is safe to run today. The single `added` entry is a real existing route, confirming the generator picks up live routes correctly. I deliberately **do not** commit a 20/21 regen here: it would be incomplete (prompts unrefreshed) and fold in unrelated fresh-vs-committed drift this PR avoids — that belongs after #4744.

### 2. Unit tests (no DB, no Hono apps)

`pnpm test:unit src/tasks/__tests__/mergeOpenAPISpecs.unit.test.ts` → **8 passed** (orphan pruned; renamed param drops old key; still-present route replaced wholesale so a stale `deprecated:true` does NOT survive; manual namespace preserved while the app-owned sibling refreshes; baseSpec top-level fields carried; empty `appSpecs` safety valve; fully-retired namespace preserved; single-segment path). Red→green confirmed: replaying the **old** `deepmerge + customMerge` logic on the prune case leaves the orphan in place, so the test fails without the fix.

## Human verification

To re-verify independently:

1. `cd langwatch && pnpm install && pnpm start:prepare:files` (generates the Prisma client the apps import).
2. **Unit suite:** `pnpm test:unit src/tasks/__tests__/mergeOpenAPISpecs.unit.test.ts` → 8 passed.
3. **Idempotency / no-break (the regen above):** in a throwaway script under `src/tasks/`, import `generateSpecs` from `hono-openapi`, the 21 apps (same imports as `generateOpenAPISpec.ts`), the committed spec, and the exported `mergeOpenAPISpecs`. Run `generateSpecs(app)` per app under `try/catch`; feed the successes to `mergeOpenAPISpecs({ currentSpec, appSpecs, baseSpec })`; assert `sha256(regen) === sha256(merge(regen, appSpecs, baseSpec))`, that an injected `/api/<owned>/__probe__` path is gone from the output, and that no committed path is dropped. Run via `pnpm exec tsx --tsconfig tsconfig.workers.json <file>`. (prompts will throw — that is #4744, upstream of the merge.)
4. **Confirm scope:** `git diff --name-only origin/main...HEAD` touches only `generateOpenAPISpec.ts`, the new `mergeOpenAPISpecs.ts`, and its test — no change to `openapiLangWatch.json`.

## Scope & known limitations (deliberately out of scope, tracked)

- **A true 21/21 regen is blocked by the pre-existing prompts crash — tracked in #4744.** Until that is fixed, the merge's preserve-on-failure rule is what keeps `/api/prompts/*` alive; this PR ships the generator capability + its regression test, not a wholesale spec regen.
- **`components.schemas` orphans are not pruned — tracked in #5079.** Only `paths` are pruned. Shared schemas are deep-merged across apps (names like `Error`/`Pagination` are not namespace-owned), so a renamed/removed schema can linger there. Inert (advertises no dead endpoint); documented in the `mergeOpenAPISpecs` docstring.
- **Fully retiring a namespace** (dropping an app so it emits zero paths) leaves its committed paths in place by design — the deliberate fail-safe above. Removing such paths stays a manual step.

## Summary by CodeRabbit

* **New Features**
  * OpenAPI specifications now merge more reliably across services, preserving manual routes while refreshing app-managed namespaces.
  * Updated routes, renamed parameters, retired namespaces, and removed routes are reflected automatically.
  * Metadata and array values are handled consistently when combining specifications.

* **Tests**
  * Added comprehensive coverage for route pruning, replacements, namespace handling, metadata, and empty specifications.

* **Documentation**
  * Clarified namespace-level refresh and pruning behavior.


langwatch/src/tasks/generateOpenAPISpec.ts
```diff
@@ -90,68 +90,36 @@ export default async function execute() {
...
-          // Replace with new
-          return (_target, source) => {
-            return source;
-          };
-        }
-      },
-    },
-  );
+  // Order preserved from the legacy deepmerge.all call (later specs win on
+  // overlap, though apps no longer collide after orphan-stripping).
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in c4be35590 — reworded to state only the accurate ordering semantics (if two specs define the same path, the later one in the list wins) and dropped the inaccurate "no longer collide" claim.

platform/app/src/tasks/mergeOpenAPISpecs.ts
```diff
@@ -0,0 +1,89 @@
...
+  for (const spec of appSpecs) {
+    for (const path of Object.keys(spec.paths ?? {})) {
+      freshPaths.add(path);
+    }
+  }
+
+  const ownedNamespaces = new Set<string>();
+  for (const path of freshPaths) {
+    ownedNamespaces.add(apiNamespace(path));
+  }
```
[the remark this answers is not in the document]
drewdrewthis: Deliberate trade-off, now documented in the `mergeOpenAPISpecs` docstring (c4be35590).

Ownership is derived from what the apps emit this run on purpose:
1. Adding an app needs zero wiring — which is exactly what fixes the original bug, where the hand-maintained `customMerge` prefix list had silently dropped `/api/model-defaults` (it was being deep-merged, not replaced).
2. A transient `generateSpecs` failure can never silently wipe a live namespace.

So the prune targets removed/renamed routes within a namespace an app still serves — the issue's actual cases (`DELETE /api/prompts/tags/{tagId}`, `PUT /api/prompts/{id}/labels/{label}`). Fully retiring a namespace (dropping the app from the generator entirely) stays a deliberate manual step — out of scope for this fix.

I considered the fail-fast-on-empty-spec suggestion, but it would crash generation if any app legitimately emits zero paths, so I kept the fail-safe behavior and documented the limitation instead.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `5eae279fb40235fa63cbb17800da78129a2694a5` · **Run:** pr-review (reviewer: @review-clerk)

Session-authorship note: this review session did not author any of this PR's diff — proceeding per pr-review procedure.

### Correction to the prior NOT-READY verdict

The prior blocking finding (thread https://github.com/langwatch/langwatch/pull/5005#discussion_r3946718568) claimed `APP_DERIVED_PREFIXES` in `platform/app/src/tasks/generateOpenAPISpec.ts` was never widened to include `/api/model-defaults`, `/api/teams`, `/api/groups`, leaving `assertMayInheritTheDefault` blind to those 3 newly-app-owned namespaces.

Independently re-verified against the actual file content at head `5eae279fb40235fa63cbb17800da78129a2694a5` (fetched via the GitHub Contents API, not inferred from the diff or the thread summary): `APP_DERIVED_PREFIXES` **does** include `/api/groups` (line 127), `/api/model-defaults` (line 141), and `/api/teams` (line 151). The finding no longer holds against current head — the list was updated (matching the PR's own account of this being fixed) after the thread was opened. No blocking issue remains here.

### Blocking

None.

### Non-blocking / notes

- CI: all gating checks green at `5eae279fb40235fa63cbb17800da78129a2694a5`, enumerated via `fetch_checks_rest` (not a rollup). SKIPPED/CANCELLED rows are superseded duplicate workflow runs with a SUCCESS counterpart present for the same job name (`preflight`, `evaluate`, `firefighting`, `breaking change stays in one component`, `typecheck`, `test`, `build`, etc.). No FAILURE conclusion anywhere in the 140-row check-run listing. `openapi-completeness`, `ci-scripts-test`, `langwatch-app-complete` all SUCCESS.
- `mergeOpenAPISpecs.ts`: namespace derivation (first 2 path segments own a namespace), prune-then-repopulate via `deepmerge.all`, and the 8-case unit test suite were read by hand — logic and tests agree, including the documented edge case where an app that stops emitting a namespace entirely leaves that namespace's committed paths untouched (intentional, tested).
- `appSpecs` array in `generateOpenAPISpec.ts` is a faithful extraction of the old `deepmerge.all` argument list (same members, same order); `assertMayInheritTheDefault`'s use of `isAppDerivedPath`/`APP_DERIVED_PREFIXES` (the security-stamping guard, unrelated to the merge boundary) is unchanged in shape and, per the correction above, now covers the 3 widened namespaces.
- Doc-comment updates in `check-openapi-route-coverage.ts` and the new module docstring accurately describe the derive-from-`appSpecs` behavior; five prior review threads (2 CodeRabbit, 3 review-clerk) on docstring accuracy and the widened-boundary callout are all resolved with substantive fixes, not just dismissed.
- All 5 existing inline review threads on this PR are resolved. No new inline threads opened this pass.

drewdrewthis:
We need a full regen to make sure this works and doesn't break anything. The correct fix here is to make sure that the generation is actually idempotent -- the whole openspec should be derived correctly, so it doesn't need to do complex merging logic.

drewdrewthis:
@drewdrewthis — ran the full regen you asked for. Short version: **generation is already idempotent, and a from-scratch derive (dropping the merge) would break things here, for the exact reason you gave on the Major thread above** ("a transient `generateSpecs` failure can never silently wipe a live namespace"). Evidence, all run live on `dc9a79da8` against real freshly-generated app specs (20/21 apps; full output in the PR body → *How I can prove I was successful*):

**Idempotent — fixed point in one application:**
```
sha256(regen1): f917202d2020d33d…   sha256(regen2): f917202d2020d33d…   regen1==regen2: true
```
Feeding the generated spec back through `mergeOpenAPISpecs` produces a byte-identical file. The "make it idempotent" goal is already met by the shipped code.

**Full regen breaks nothing — 0 paths dropped:**
```
paths: 82 -> 83    dropped: 0    added: 1  (/api/workflows/{id}/evaluate — a real route the committed spec was missing)
orphan probe injected -> pruned;  hand-maintained / /api/annotations /api/api-keys /api/experiments /api/projects /api/trace all preserved
```

**Why I did *not* delete the merge and derive the whole spec from scratch (Option A):** it's blocked and unsafe right now, for two structural reasons:

1. **One app cannot generate.** `generateSpecs(promptsApp)` crashes with a pre-existing `zod-openapi@4.2.4` circular-reference error *upstream of any merge* — tracked separately in #4744. A from-scratch derive needs all 21 apps; with prompts crashing it would **drop all 8 `/api/prompts/*` paths**, ghost-removing live SDK endpoints. That is precisely the "transient `generateSpecs` failure wipes a live namespace" failure mode you called out — and the merge's *preserve-a-namespace-no-app-emits* rule is the fail-safe that prevents it. The regen proves it: 0 dropped, even with prompts down. So a true 21/21 from-scratch regen is impossible until #4744 is fixed, and an unguarded derive is the opposite of "doesn't break anything."
2. **Six namespaces are hand-maintained** (`/`, `/api/annotations`, `/api/api-keys`, `/api/experiments`, `/api/projects`, `/api/trace`) — emitted by no Hono app. A pure derive would need them moved to an explicit base source, a larger source-of-truth refactor that itself can't be regen-proven until #4744 lands.

The merge itself is ~30 lines (derive owned namespaces → strip → `deepmerge.all`); idempotency is proven, so there's nothing a from-scratch derive would simplify. I deliberately did **not** commit a 20/21 regen: it would be incomplete (prompts unrefreshed) and fold in unrelated drift this PR avoids — that belongs after #4744. `components.schemas` orphan pruning is tracked in #5079.

Net: this PR ships the generator capability + regression test, proven idempotent and non-destructive; the full from-scratch derive is a follow-up gated on #4744. Happy to take that on as a stacked PR once prompts can generate — let me know if you'd prefer that path instead.

langwatch-agent:

**Scariest prod risk:** The prune wipes a namespace's paths from the committed spec whenever the owning app fails to generate that run, ghost-removing live SDK endpoints (the prompts app already crashes on `zod-openapi` per #4744).

- `[Must Check]` `mergeOpenAPISpecs.ts` — confirm the "strip owned namespace then repopulate" only strips namespaces that were actually emitted this run, so an app that crashed leaves its existing paths intact (author claims 0 dropped despite prompts crash).
- `[Must Check]` hand-maintained namespaces (`/api/annotations`, `/api/api-keys`, `/api/experiments`, `/api/projects`, `/api/trace`, root `/`) — verify they are preserved untouched since no Hono app emits them.
- `[Probably Fine]` derivation of owned `/api/<namespace>` prefixes from generated specs — confirm it cannot pick up a partial/empty namespace from a half-failed app and treat it as "owned but empty" (which would strip real paths).

**Author questions:**
- If an app generates successfully but legitimately emits zero paths one run, does the derive treat its namespace as owned-and-empty and prune existing paths, or as not-owned and preserve them?

drewdrewthis:

CI: green (zero failing, zero pending)
ACs: met — verdict comment re-pinned at SHA 3e9ba30590208aeb17448ceccd316ef6f1782911
Threads: 2 total, 0 unresolved non-outdated

Verified by:
  \`bash /home/ubuntu/.claude/scripts/pr-ready-check.sh langwatch langwatch 5005\` → \`{"overall":"READY"}\` (C1–C7 all pass)
  \`command gh pr checks 5005 -R langwatch/langwatch | grep fail | wc -l\` → \`0\`
  C2: CI green (53 checks; SUCCESS: 41, SKIPPED: 13)
  C3: verdict READY, SHA matches HEAD (3e9ba3059020)
  C4: no unresolved review signals (of 2 threads)
  C5: body has '## Human verification' + '## How I can prove I was successful'
  C6: backend-only declared — visual proof exempt
  C7: no deferral language in body

drewdrewthis (commented):
## Review verdict: READY

Re-reviewed at post-rebase HEAD: diff matches the PR's OpenAPI-prune intent, no regressions from the rebase; CI fully green (58 pass).

drewdrewthis:

drewdrewthis:
Answering the open PR Hound author question ("if an app generates successfully but emits zero paths one run — pruned or preserved?"): **preserved**. Ownership is derived from `Object.keys(spec.paths ?? {})` across the fresh app specs, so an app emitting zero paths contributes no keys, its namespace never enters `ownedNamespaces`, and its committed paths are left untouched (same code path as a crashed app — proven live by the prompts crash in the PR-body regen: 8 `/api/prompts/*` paths preserved, dropped: 0). Source-verified at the PR head in `mergeOpenAPISpecs.ts`; the docstring documents this as the deliberate trade-off.

platform/app/src/tasks/mergeOpenAPISpecs.ts
```diff
@@ -0,0 +1,103 @@
...
+ * DERIVED from the fresh specs this run (never a hand-maintained list, so it
+ * cannot drift). Every committed path in an owned namespace is dropped before
+ * merging, then repopulated from the app specs — so routes that an app no
+ * longer generates (removed routes, renamed path params) are pruned instead of
+ * lingering forever, and routes that still exist are refreshed via a clean
+ * replace (no stale sub-keys survive, because there is no collision left to
+ * deep-merge).
+ *
+ * Paths in namespaces that NO app generates are hand-maintained (e.g.
+ * `/api/annotations`, `/api/projects`, the singular `/api/trace`, the root
```
drewdrewthis: **[review-clerk] Docstring example is wrong: `/api/projects` is NOT hand-maintained under this PR's own code.**

This line lists `/api/projects` alongside `/api/annotations`/`/api/trace`/`/` as an example of a namespace "no app generates" that stays "hand-maintained" and "always preserved untouched." That's not accurate: `projectsApp` is imported in `generateOpenAPISpec.ts` and `projectsSpec` is included in the `appSpecs` array passed into this function (`generateOpenAPISpec.ts:155`). So `/api/projects` IS an owned/derived namespace — it gets stripped from `currentSpec` and repopulated from `projectsSpec` on every run, same as `/api/prompts` or `/api/traces`.

This is load-bearing documentation for a prune/preserve safety boundary — a future reader trusting this comment could wrongly conclude `/api/projects` entries are safe to hand-edit in the committed JSON, when in fact any hand edit there would be silently overwritten/pruned on the next generator run. Please swap the example for an actually-unowned namespace (e.g. `/api/api-keys` or `/api/experiments` — need to double check those too — or just `/api/annotations`, `/api/trace`, `/`).

platform/app/src/tasks/generateOpenAPISpec.ts
```diff
@@ -195,58 +126,49 @@ export default async function execute() {
...
-          };
-        }
-      },
-    },
-  );
+  // Order carried over from the legacy deepmerge.all call: specs are deep-merged
+  // in array order (a later spec deep-merges onto earlier ones; arrays are
+  // replaced). Apps own disjoint /api/<namespace>s, so their paths never collide
+  // here — order only affects shared top-level keys such as components.schemas.
+  const appSpecs = [
```
drewdrewthis, line 133: **[review-clerk] `appSpecs` silently widens the derived-ownership set vs. main's `APP_DERIVED_PREFIXES` for 3 namespaces — worth calling out explicitly.**

Independently classified all 139 paths in the committed `openapiLangWatch.json` against (a) main's list-based `isAppDerivedPath()` and (b) this PR's namespace-derivation, using each namespace's *currently committed* paths as the freshness proxy. Zero divergence for every namespace that was already in `APP_DERIVED_PREFIXES` on main.

But `modelDefaultsSpec` (line 150), `teamsSpec` (line 159), and `groupsSpec` (line 160) were **not** in main's `APP_DERIVED_PREFIXES` list — under main's mechanism their paths were only additively deep-merged (never pruned/replaced). Under this PR's derived-ownership mechanism, because these three specs are included in `appSpecs`, `/api/model-defaults`, `/api/teams`, and `/api/groups` now become owned namespaces subject to the same strip-and-replace pruning as everything else — a real, 3-namespace scope expansion beyond main's mechanism.

I checked the route handlers for `teamsApp`, `groupsApp`, and `modelDefaultsApp.v1` and confirmed every path currently committed under these three namespaces has a matching live handler, so nothing appears to actually break, and it's consistent with the PR's stated goal (prune orphans generically instead of via a hand list this PR is explicitly trying to retire). But the PR description's claim of an "IDENTICAL prune set" to main's mechanism isn't accurate as written — this is a real (likely beneficial) behavior change that should be called out explicitly in the PR description rather than left implicit, since it wasn't part of what the rebase conflict was nominally resolving. Non-blocking given the evidence above, but please amend the description so the next reader (or the next person debugging a teams/groups API doc regression) isn't surprised.

drewdrewthis (commented):
Full re-review after the rebase onto main (4f2298e4d1a2 -> 9b55d5cf0726) that resolved the generateOpenAPISpec.ts conflict in favor of this PR's derived-ownership mechanism. See verdict comment for the summary.

platform/app/src/tasks/generateOpenAPISpec.ts
drewdrewthis: Amended the PR description — you're right, and the "identical prune set" framing was wrong along with the hand-maintained list it rested on.

New section, **"Derived ownership widens the prune boundary by 3 namespaces — intended, no-op today"**, states it directly: `/api/model-defaults`, `/api/teams` and `/api/groups` were never in `APP_DERIVED_PREFIXES`, main only deep-merged them additively, and under derived ownership they are owned and therefore prunable. Framed as the fix rather than a side effect — those three are precisely where the hand list had drifted past the #2987 bug it existed to prevent.

I reproduced your no-op finding independently before writing it down rather than taking it on trust. All 24 committed paths across the three namespaces match a live handler; the `/api/model-defaults` four register in `app.v1.ts` onto the parent app's `/api/model-defaults` basePath, which is worth noting because a per-file basePath scan reports them as orphans. `pnpm check:openapi-route-coverage` passes at HEAD — `210/317 registered routes are in the document ... OK: every registered route is documented or accounted for` — and that gate is the standing guard for exactly this risk: it goes red if a prune ever drops a live route's entry from any owned namespace.

Your classification also surfaced two things past the scope of this thread, both now fixed in the description:

- **The body's own hand-maintained list was wrong, and worse than the docstring's.** It named `/api/api-keys`, `/api/experiments` and `/api/projects` — all three app-owned. The actual unowned set today is `/`, `/api/annotations` and the singular `/api/trace`. Three, not six.
- **The regen transcript predates the rebase.** It was measured at `dc9a79da8` (21 app specs, 82 paths); HEAD carries 32 and 139. I left the numbers intact and added a staleness banner rather than restating stale measurements as current — a fresh full regen still waits on #4744.

platform/app/src/tasks/mergeOpenAPISpecs.ts
drewdrewthis: Fixed in 49309bfe24162d2d7ba1ee6d71b2c4e307d5f7c4. Dropped `/api/projects` and added the rule instead of just a corrected example: ownership is decided by the `appSpecs` array in `generateOpenAPISpec.ts`, and a plural sibling can be owned while the singular is not (`/api/traces` owned, `/api/trace` not).

Verified the replacement examples rather than assuming — your suggested `/api/api-keys` and `/api/experiments` are both owned too. Resolving all 32 appSpecs basePaths plus the described routes of the two `/api`-rooted apps (`misc.ts`, `evaluations-legacy.ts`) gives 35 owned namespaces against 38 committed, so the unowned set is exactly `/`, `/api/annotations`, `/api/trace`.

The same error was in the PR body ("six namespaces are hand-maintained", naming `/api/api-keys`, `/api/experiments`, `/api/projects` — all three owned); it is three, and that is fixed there too.

sergioestebance (changes requested):
This PR has merge conflicts with the base branch. Please rebase and resolve them so we can proceed with the review.

drewdrewthis:
@sergioestebance the merge conflict is resolved.

Merged `origin/main` into this branch (commit 89c3873cd2). `git merge`, not rebase — no force-push.

**What conflicted, and how it was resolved:**

1. `platform/app/src/tasks/generateOpenAPISpec.ts` — `origin/main` had independently added its own orphan-prune logic since this branch diverged: a hand-maintained `APP_DERIVED_PREFIXES` array + inline `isAppDerivedPath()` + `deepmerge.all({ customMerge })`. That's exactly the fragile pattern issue #2987 exists to retire (it's how `/api/model-defaults` silently dropped out of the merge in the first place). Resolution keeps this PR's `mergeOpenAPISpecs()` (dynamic ownership derivation from `appSpecs`, no hand list) as the actual prune mechanism, and folds in every new app `main` had added since divergence (`organizationApp`, `organizationsApp`, `roleBindingsApp`, `rolesApp`, `scimTokensApp`, `scimApp`, `runPlansApp`, `testSuitesApp`, `queryApp`, `codingAgentApp`, `codingAgentV1App`, `agentCacheApp`, `langyControlApp`, `analyticsSqlApp`, `gatewayPlatformApp`) into the `appSpecs` array so their routes get the same dynamic-ownership pruning.

   One deliberate keep: `APP_DERIVED_PREFIXES` / `isAppDerivedPath` are **not fully deleted**. `main` also added a separate security guard, `assertMayInheritTheDefault()` (line ~408), which consumes `isAppDerivedPath()` directly, plus a 44-test suite (`scripts/__tests__/check-openapi-route-coverage.unit.test.ts`, added by `main` after this branch diverged) that asserts the prefix list and helper exist. Deleting them breaks that guard and those tests. The list now only feeds that separate security check — the actual path-prune no longer consults it at all, so the #2987 bug (a namespace silently missing from a hand list) can't recur in the merge path. Commented in-line at the list's declaration to make this explicit for the next reader.

   Follow-up (separate PR, not blocking this one): fully retiring the list would mean rewriting `assertMayInheritTheDefault` to derive ownership dynamically too, plus updating that test suite's assertions — flagging it rather than doing it silently in this PR's diff.

2. `platform/app/scripts/check-openapi-route-coverage.ts` — small conflict in the remediation message inside `formatUnexplained()`. Kept `main`'s addition describing the `@langwatch/api` service endpoint-config publishing path, dropped its "add its prefix to APP_DERIVED_PREFIXES" instruction (no longer applicable), replaced with this PR's "import its app ... and its spec into the appSpecs list" instruction.

**Verification (this branch, post-merge, before push):**
- `grep` for conflict markers across the tree: zero hits.
- `vitest run src/tasks/__tests__/mergeOpenAPISpecs.unit.test.ts scripts/__tests__/check-openapi-route-coverage.unit.test.ts` → `Test Files 2 passed (2)`, `Tests 44 passed (44)`.
- `tsgo --noEmit` on the three touched files: zero errors (repo-wide tsgo has ~5459 pre-existing errors in this sandbox from an unbuilt Prisma client / workspace packages, unrelated to this change and present on a clean `origin/main` checkout too — none of them are in the touched files).

Pushed to `fix/2987-openapi-prune-orphans` (89c3873cd2), no force-push. Watching CI now.

drewdrewthis:
CI is now fully green on the merge (head `5eae279fb4`).

One real CI failure surfaced after the merge: Biome reported one genuine error (not a warning) in `platform/app/src/tasks/generateOpenAPISpec.ts:1:1` — `assist/source/organizeImports`, an import-sort issue left over from resolving the merge conflict. Fixed (import reordering only, no logic change) and re-pushed as `5eae279fb4`.

`gh pr checks 5005` — final state, zero pending, zero failing:
```
$ gh pr checks 5005 --repo langwatch/langwatch; echo "EXIT: $?"
... (89 checks) all pass or skipping ...
EXIT: 0
```
All three of this repo's required status-check contexts (`validate title`, `typecheck`, `build` — per `gh api repos/langwatch/langwatch/branches/main/protection/required_status_checks`) show `pass` on this head commit. Every other "skipping" entry belongs to an unrelated component workflow (go/nlpgo/sdk builds) whose own path-filter correctly didn't fire for this diff — not one of the three required contexts, and not blocking mergeability (`gh pr view` reports `mergeable: MERGEABLE`).

Re-requested your review via the API (`POST .../pulls/5005/requested_reviewers`) now that the conflict you flagged is resolved and CI is green — not a label toggle. `reviewDecision` will stay `CHANGES_REQUESTED` until you re-review or dismiss; that's expected and I'm not touching it.

Keeping `grinding` on this PR while it waits on your re-review.

platform/app/src/tasks/generateOpenAPISpec.ts
```diff
@@ -72,10 +84,11 @@ import { app as miscApp } from "../server/routes/misc";
...
   requireDefaultedResponseFields(await generateSpecsUnpinned(hono, options, c));
 
-// Surfaces whose routes come straight from their Hono apps. Their paths
-// REPLACE on merge, and any path the apps no longer serve is pruned from
-// the previous spec below: without the prune, a deleted route would ride
-// the merge union forever.
+// Namespaces owned by a Hono app. The MERGE no longer consults this list —
+// `mergeOpenAPISpecs` derives ownership dynamically from `appSpecs` each run, so
+// the prune can no longer drift (issue #2987). The list survives only as the
+// input to `assertMayInheritTheDefault`: the security-stamping guard that
```
drewdrewthis, line 90: **[New AC]** `assertMayInheritTheDefault` (line 410) still gates on `isAppDerivedPath` / `APP_DERIVED_PREFIXES` — the OLD hand-maintained list — while the merge's ownership boundary (this PR's whole point) is now derived from `appSpecs` and is 3 namespaces wider (`/api/model-defaults`, `/api/teams`, `/api/groups`, per the PR description). Those 3 namespaces are merged/pruned as app-owned, but the security-stamping guard does not know they're app-owned, because `APP_DERIVED_PREFIXES` was never updated to include them.

Concretely: if a future route under `/api/teams`, `/api/groups`, or `/api/model-defaults` is generated with a path spelling that doesn't match the route registry, `assertMayInheritTheDefault` will silently `return` (since `isAppDerivedPath` is false for those prefixes) instead of throwing — exactly the "operation silently inherits the document-wide default" failure this guard exists to catch (per its own docstring at line ~396-408). That's the same drift class (#2987) this PR removes from the merge, reintroduced in the security guard.

Suggest: derive `assertMayInheritTheDefault`'s ownership check from the same `appSpecs`-derived set `mergeOpenAPISpecs` now uses (or otherwise widen `APP_DERIVED_PREFIXES` to match), so the two ownership definitions can't diverge again. Not a live break today (nothing in the committed spec exercises this path yet, and CI is green), but it's an unaddressed gap this PR itself creates by widening one boundary without the other — worth an AC or a tracked follow-up issue explicitly, the way #4744/#5079 are called out in the PR body.

drewdrewthis (commented):
Automated review pass (review-clerk) at 5eae279fb4.

drewdrewthis (commented):
NOT-READY at 5eae279fb4 (REQUEST_CHANGES refused 422 — acting login is PR author's login on this single-collaborator repo; COMMENT stands in per procedure fallback, carries the same NOT-READY gating weight). See verdict comment: https://github.com/langwatch/langwatch/pull/5005#issuecomment-4760464501 and inline thread: https://github.com/langwatch/langwatch/pull/5005#discussion_r3946718568

drewdrewthis:
[grinder] Verified the clerk's blocker in https://github.com/langwatch/langwatch/pull/5005#discussion_r3946718568 and it does not hold — closing it out.

**Claim:** `assertMayInheritTheDefault`'s allowlist (`APP_DERIVED_PREFIXES` in `platform/app/src/tasks/generateOpenAPISpec.ts`) is missing `/api/model-defaults`, `/api/teams`, `/api/groups`.

**Finding:** all three are already present in that list at head SHA `5eae279fb4`:
```
$ grep -n '"/api/model-defaults"\|"/api/teams"\|"/api/groups"' platform/app/src/tasks/generateOpenAPISpec.ts
127:  "/api/groups",
141:  "/api/model-defaults",
151:  "/api/teams",
```
`git blame` shows these three entries predate this PR (commit `5ebf89d6f47`, 2026-08-10, by Rogério Chaves) — this PR does not touch `APP_DERIVED_PREFIXES` at all; `gh pr diff` shows only the comment above the array changed, not its contents.

The comment the clerk likely keyed off (`gh pr diff` lines 178-179) is describing a *different, already-retired* prefix list — the hand-maintained one that used to live inside `mergeOpenAPISpecs.ts` for merge-ownership, which this PR replaces with dynamic derivation from `appSpecs` (issue #2987). That module comment explicitly says *"The prefix list this module replaced did not cover `/api/model-defaults`, `/api/teams` or `/api/groups`"* — past tense, about the list this PR deletes, not the security guard's `APP_DERIVED_PREFIXES` which is untouched and already correct.

No code change needed. Resolving the thread.

Verified by:
  grep -n '"/api/model-defaults"\|"/api/teams"\|"/api/groups"' platform/app/src/tasks/generateOpenAPISpec.ts → lines 127, 141, 151 (all present)
  git blame -L 125,152 platform/app/src/tasks/generateOpenAPISpec.ts → all three attributed to commit 5ebf89d6f47 (2026-08-10), predating this PR
  gh pr diff 5005 (scoped to APP_DERIVED_PREFIXES hunk) → only the surrounding comment changed, array contents unchanged
  gh pr checks 5005 → zero failing, zero pending (all applicable jobs "pass", rest "skipping" as not applicable)
  reviewThreads(first:50) via gh api graphql → 1 unresolved thread total (this one)

drewdrewthis (commented):
Review verdict: READY at 5eae279fb40235fa63cbb17800da78129a2694a5. See pinned verdict comment: https://github.com/langwatch/langwatch/pull/5005#issuecomment-4760464501

Issue #3365, opened 2026-04-21
Scenario runs orphan at QUEUED when worker maxRuntime restart kills child process


drewdrewthis:
## Summary

Scenario runs can become permanently stuck in `QUEUED` state when a worker process hits its `maxRuntimeMs` restart threshold during child-process spawn. No `started`, `failed`, or `completed` event is ever emitted, so the UI polls forever and the run is orphaned.

### Observed timeline (real case, local dev)

```
15:48:03.678  lw.simulation_run.queued  emitted (event persisted to CH)
15:48:04.302  execution-pool: "Starting scenario execution"
15:48:04.334  spawning child process
15:48:04.392  child process spawned (pid 1059)
15:48:05.068  execution-pool: "Draining: killing child process"
15:48:05.079  worker: "worker restart — Max runtime reached, restarting worker"
15:48:21.114  new worker process starts (Scenario processor started)
```

The GroupQueue had already marked the `scenarioExecution` reactor job as completed (it's fire-and-forget — reactor returns immediately after `pool.submit`). When the new worker process started, it did not re-process the queued event because the fold checkpoint was past it. The `simulation_runs` projection remained at `Status=QUEUED`.

### Root cause

Three-layer race:

1. `scenarioExecution.reactor.ts` (`src/server/event-sourcing/pipelines/simulation-processing/reactors/scenarioExecution.reactor.ts:88`) calls `pool.submit()` synchronously, then returns. GroupQueue marks the reactor job `completed`.
2. Execution pool is in-process (child-process-based). The actual child isn't done yet — it may not even have started.
3. `maxRuntimeMs` fires in the parent worker process, triggering pool drain (`execution-pool: "Draining: killing child process"`). Children are killed, but no failure event is emitted on their behalf.

Result: the event log has `lw.simulation_run.queued` but no follow-up event. The GroupQueue believes the event is handled. The new worker never looks at it.

### Why this is not specific to any single PR

Hit during local testing of PR #3334 (workflow agent mapping layer), but the race is pre-existing and independent of scenario target type. Any scenario run queued within ~1–2s of a worker maxRuntime restart is vulnerable.

### Likely fix directions (to be refined in investigation)

- **(A) Drain-with-emit:** before `killChild`, emit `lw.simulation_run.failed` (reason: `worker_restart`) so the projection transitions out of `QUEUED`. Safest.
- **(B) Pre-restart gate:** block `maxRuntime` restart while any children are in flight, or wait for them (with a ceiling).
[...]

PR #5006, opened 2026-06-21
fix(scenarios): reconcile orphaned IN_PROGRESS runs on worker boot, and make a finished run final (#3195)

### LINKED ISSUE #3195: Scenarios: Run stuck at STARTING when worker restarts mid-run (no heartbeat)
## Summary

When the scenarios worker container restarts (or crashes) mid-run, the UI leaves the scenario run at `status: STARTING` forever. There is no worker heartbeat, no timeout, and no orphaned-run reconciliation on worker boot. The user has no indication the run is dead.

## Reproduction

1. Start a scenario run on `/<project>/simulations/`
2. While it's in the "Starting" phase (before the child process emits its first event), restart the workers container (e.g., `docker compose restart workers`)
3. Observe the UI: status remains "Starting" indefinitely
4. The scenario run row never transitions to ERROR, never becomes cancellable from the worker side

## Expected

Either:
- Worker publishes a periodic heartbeat and the server marks runs stale after N seconds with no signal, OR
- Workers sweep for in-flight runs they don't own on startup and mark them ERROR/ORPHANED

## Actual

- Run is permanently stuck in the "Starting" state
- No way to know from the UI that no worker is processing it

## Context

- Reproduced in this session after restarting the workers container to pick up a new env var
- Real-world equivalent: worker OOM, deploy, or crash leaves users staring at spinners

## Acceptance Criteria

_Pending investigation — will be filled in after codebase analysis._

---
*Classification: Bug*
*Parent: #3190*

drewdrewthis:
## Summary

Closes #3195. When a scenarios worker dies mid-run (OOM, crash, deploy, container restart), the run is left non-terminal in ClickHouse forever — in the scenarios view (`/<project>/simulations/`) the run spins at **"Starting"/"Running"** with no indication no worker is processing it, and downstream reactors never fire. This adds **orphaned-run reconciliation on worker boot** plus a **finalized-status fold guard**, so a stuck run reaches a terminal **ERROR** that is **visible to the user in the scenarios view**.

## Root cause

The scenario execution pool is in-process and per pod — no cross-process record of which run a worker owns. When a worker process dies, the in-process `ScenarioFailureHandler` that would emit the terminal `finished` event dies with it. Read-time stall detection (`resolveRunStatus`) paints the run STALLED cosmetically after 2× the child timeout, but never writes a terminal event, so the run never truly leaves the in-flight state and the `suiteRunSync` / metrics / customer.io reactors never run.

## What changed

- **Boot reconciliation** (`orphaned-run-reconciliation*.ts`, wired fire-and-forget into `startScenarioProcessor`): on boot, query `simulation_runs` (shared CH client, cross-tenant, deduped to the latest version via the IN-tuple pattern, partition-pruned on `StartedAt`, light columns only, oldest-first) for **`IN_PROGRESS`** runs whose last activity (`UpdatedAt`) is older than `STALL_THRESHOLD_MS` (2× the child timeout). Each is reconciled to a terminal **ERROR** via the existing **idempotent** `ScenarioFailureHandler.ensureFailureEventsEmitted`, with a distinct orphan error message.
  - **Why that threshold, and why only `IN_PROGRESS`** — the multi-pod false-positive guard. A live worker hard-caps every child at the child timeout and emits its own terminal event at the cap, so an `IN_PROGRESS` run quiet for longer than 2× that provably has no live worker. That inference is the entire licence to write a terminal ERROR, and it **only closes for a run that was started**: nothing bounds queue wait, so a `QUEUED` run behind a draining backlog goes stale while a healthy worker is still working toward it. Reconciling a run a live pod still owns is the one outcome we must never produce.
  - **`QUEUED` orphans are out of scope**, and are owned by `scenario-orphan-reconciler.ts` (#3365, merged in #5008 while this branch sat). The two boot sweeps are disjoint by status, so neither can double-write the other's runs.
  - **Why ERROR, not a cosmetic STALLED** — ERROR is a real `finished` event, so it fires the downstream reactors the read-time path never triggers. That firing *is* the fix.
- **Finalized-status fold guard** (`simulationRunState.foldProjection.ts`): once `FinishedAt` is set, `Status` is terminal and stays terminal. Three things hold that line, and all three are load-bearing:
  1. `statusAfter` at the non-terminal transition sites — a child that outlived its dead parent and POSTs a real `started`/`snapshot` with a client timestamp after the reconcile time applies in-order (the executor only re-folds when `occurredAt` is strictly less than what was seen) and would otherwise clobber `Status` back to `IN_PROGRESS` while `FinishedAt` stays set;
  2. `handleSimulationRunFinished` returning early once `FinishedAt` is set — a run finishes exactly once, so a late child cannot take a reconciled `ERROR` back to `SUCCESS`, nor split the record (an `ERROR` status carrying the late child's `SUCCESS` verdict);
  3. that same handler refusing a **non-terminal** explicit status — the ingest schema types it as `z.nativeEnum(ScenarioRunStatus)`, so `finished(status=IN_PROGRESS)` would otherwise write a non-terminal `Status` alongside `FinishedAt`: a row the reconciler skips (`FinishedAt IS NULL`) and read-time stall detection skips. Nothing could ever recover it.

  (2) and (3) came out of review — both were found by *executing* the fold rather than reading it, and each has a regression test verified to fail against the un-fixed handler.

The three load-bearing design decisions (threshold, terminal ERROR, fold guard) — including the alternatives rejected — are recorded in **[ADR-010: Scenario Orphaned-Run Reconciliation](https://github.com/langwatch/langwatch/blob/fix/3195-scenario-starting-orphan/dev/docs/adr/010-scenario-orphaned-run-reconciliation.md)** (added in this PR).

## Deployment Impact

No deployment impact for operators.

- **Env vars added/changed:** none. The reconciliation threshold is derived programmatically from the existing in-process `STALL_THRESHOLD_MS` constant; no new env vars are introduced.
- **Helm values added/changed:** none. This is a pure in-process change to the scenarios worker boot sequence; no chart, service definition, or config map is touched.
- **Default behavior on vanilla `helm install` with no overrides:** unchanged. The boot-time reconciliation fires automatically in the existing worker process. No operator configuration is required or possible; the fix is self-contained.
- **BYOC dataplane impact:** none. The reconciler queries the shared ClickHouse client only; tenants on private ClickHouse instances (`CLICKHOUSE_URL__*`) are explicitly excluded from the sweep (tracked as a follow-up under #3195). No BYOC-specific wiring is changed.

## Human verification

The run status **is user-observable** — the scenarios view (`/<project>/simulations/`) renders each run's status. A skeptical human can re-verify the fix end-to-end in the real app:

1. **Reproduce the bug** — start a scenario run from the scenarios view. While it shows **"Starting"** (before the child emits its first event), kill the scenarios worker (`docker compose restart workers`) so no `finished` event is ever emitted. The run stays stuck at "Starting" — the #3195 symptom.
2. **Age past the threshold** — wait past the reconcile threshold (2× the child timeout = 30 min), then boot a scenarios worker.
3. **Observe the terminal state** — reload the scenarios view: the run now shows a **terminal ERROR** ("Worker restarted or crashed before the run completed"), and the downstream suite-aggregate / metrics reactors have fired. The 30-min threshold makes this slow to reproduce by hand, which is why the boot path is also covered deterministically by the automated tests below.
4. **Design review** — the three load-bearing decisions are recorded in [ADR-010](https://github.com/langwatch/langwatch/blob/fix/3195-scenario-starting-orphan/dev/docs/adr/010-scenario-orphaned-run-reconciliation.md) and exercised by the tests in the next section.

The user-observable surface is the **run status in the scenarios view** (`/<project>/simulations/`): a run stuck at **"Starting"** after a worker restart now reaches a **terminal ERROR** the user can see.

### ✅ Live-app visual proof — STARTING-stuck orphan reconciled in the live UI

In-flight scenario run stuck at Stalled, reconciled to Failed by the #5006 boot reconciler

Staged on the live golden instance running **this branch**. An orphaned in-flight (`IN_PROGRESS`) scenario run — no live worker, ~2 h with no progress — was seeded into the `simulation_runs` projection. **Left:** the Scenarios suite-run view paints it stuck (`⚠ Stalled`); a worker without this PR never takes it terminal (the #3195 bug). **Right:** after bouncing the scenario worker so `startScenarioProcessor` runs **this PR's** boot reconciler, the same run flips to a terminal `⊗ Failed` badge. ClickHouse confirms the status moved `IN_PROGRESS → ERROR` carrying this reconciler's `ORPHAN_ERROR_MESSAGE` — `Error = "Worker restarted or crashed before the run completed"` — and the worker emitted the `finished(ERROR)` fold/reactor events for `blob-5006-orphan-run`. Falsifiable: the flip happens **only** after this PR's reconciler boots; the stuck `Stalled` state is exactly what a pre-fix worker leaves behind.

**Finalized-status guard — red (without the guard) → green (with it):**

```
# guard reverted at its 3 call sites:
×  keeps the terminal status instead of resurrecting IN_PROGRESS
   AssertionError: expected 'IN_PROGRESS' to be 'ERROR'
×  keeps the terminal status (late snapshot)
   AssertionError: expected 'IN_PROGRESS' to be 'ERROR'
 Tests  2 failed
```

**Finish-is-final — the two holes review surfaced, each red without its fix:**

```
# finish-once early-return + terminal-status refusal both reverted:
×  keeps the first terminal record and ignores the late one
×  refuses the non-terminal status and finishes the run terminally
×  derives the terminal status from the verdict
 Tests  3 failed | 40 passed (43)
```

```
# with all three guards, affected unit suites:
 Test Files  7 passed (7)
      Tests  81 passed (81)
```

**ClickHouse integration test (real CH via testcontainers) — the discrimination AC:**

```
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Proves only stale **`IN_PROGRESS`** runs are surfaced as orphaned; healthy/recent, terminal, and archived runs are excluded; **a stale `QUEUED` run waiting behind a backlog is never surfaced** (the false positive this sweep must not produce); and the latest version of a run decides. `pnpm typecheck` is clean. The full real-CH run (with the falsifiability flip) is captured in the **Backend evidence** section at the bottom.

## Scope / limitations (intentional)

- The literal dev single-restart repro is caught on the **subsequent** worker boot, once the orphan ages past the threshold — a young orphan must not be reconciled at boot, or it would false-positive a run the booting pod is about to pick up (or one a live pod owns). Read-time STALLED covers the gap cosmetically meanwhile, and prod workers restart regularly (deploys, the `maxRuntimeMs` self-restart), so orphans are swept within a bounded window.
- Tenants on private ClickHouse instances (`CLICKHOUSE_URL__*`) are not swept — only the shared client. Noted for follow-up under #3195 (the same issue tracks the boot-reconcile path).

## Release Notes

- **New Features**
  - Added boot-time reconciliation for orphaned scenario runs to automatically mark stalled runs as failed after worker restarts.
  - Added ClickHouse-based orphan discovery that targets only stale, eligible non-terminal runs and selects the latest run version when deduplicating.

- **Bug Fixes**
  - Prevented already-finished runs from being overwritten by later out-of-order events.
  - Continued reconciling remaining orphans even if marking one run fails.

- **Tests**
  - Added unit, ClickHouse integration, and end-to-end specification coverage for orphan reconciliation behavior and filtering rules.

---

## 🔬 Backend evidence — real-ClickHouse boot-reconciler test

> This is the **deterministic backend proof** of the boot path. It is **not** the live-app visual proof (real user, full app UI, E2E), now satisfied by the embedded live-app visual proof above. The user-observable surface this exercises is the **run status in the scenarios view** — a STARTING-stuck orphan becoming terminal ERROR.

Ran the boot-reconciler's ClickHouse integration test against a **real ClickHouse** (`clickhouse/clickhouse-server:25.10.2.65` via testcontainers, goose-migrated to production schema parity) at HEAD `d63d36ce0`. This is the **picking-up** half — the live query that finds STARTING-stuck orphans — backed by the reconcile unit test for the **are-reconciled** half, plus a red→green falsifiability flip proving the query predicate is load-bearing.

> Orphan status in the fixtures is `QUEUED`; the same gate covers `PENDING`/`IN_PROGRESS` — the in-flight statuses the UI paints as "Starting"/"Running".

### 1. GREEN — real-CH finder surfaces only the stale, non-terminal orphan

```
$ NODE_ENV=test pnpm exec vitest run -c vitest.integration.config.ts \
    orphaned-run-reconciliation.clickhouse
[globalSetup] Starting testcontainers...
[globalSetup] ClickHouse URL: http://test:test@localhost:32768/test_langwatch
[globalSetup] Redis URL: redis://localhost:32770

 ✓ ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)
     > given a mix of stale, healthy, terminal and archived runs
       > surfaces only the stale non-terminal run, with the ids needed to finish it  37ms
 ✓ ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)
     > given a run that was queued long ago but later finished
       > does not surface it — the latest version decides  23ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Inserts four real rows — a stale `QUEUED` orphan (last activity 60 min ago), a healthy `IN_PROGRESS` run (1 min ago), a finished `SUCCESS` run, and an archived run — and asserts the live cross-tenant dedup query returns **only** `orphan-stale` with the exact ids needed to finish it, excluding the live, terminal, and archived rows. The second case proves the latest version of a multi-version run decides (queued-then-finished is not surfaced).

### 2. Falsifiability — break the staleness gate → orphan stays STARTING (red)

Flipped the single load-bearing predicate in `orphaned-run-reconciliation.clickhouse.ts` — `AND UpdatedAt < {staleBefore}` → `AND UpdatedAt > {staleBefore}` (i.e. "match recent, not stale"):

```
 ❯ orphaned-run-reconciliation.clickhouse.integration.test.ts (2 tests | 1 failed)
   × surfaces only the stale non-terminal run, with the ids needed to finish it

 AssertionError: expected [ 'healthy-recent' ] to include 'orphan-stale'
   126|       const ids = mine.map((r) => r.scenarioRunId);
   128|       expect(ids).toContain("orphan-stale");
      |                   ^

      Tests  1 failed | 1 passed (2)
```

That one assertion captures **both** failure modes the fix prevents: the stale orphan vanishes from the result set (→ never reconciled → **stuck at STARTING forever**, the #3195 bug) and the 1-min-old live run wrongly appears (→ a run a live worker still owns would be killed). Reverting the predicate restores green:

```
 ✓ surfaces only the stale non-terminal run, with the ids needed to finish it
 ✓ does not surface it — the latest version decides
      Tests  2 passed (2)
```

### 3. GREEN — the surfaced orphan is reconciled to a terminal ERROR on boot

The reconcile half — `reconcileOrphanedRuns`, the core of the worker-boot entrypoint:

```
$ NODE_ENV=test pnpm exec vitest run orphaned-run-reconciliation.unit
 ✓ reconcileOrphanedRuns > given the finder surfaces an orphaned run
     > emits a terminal failure event scoped to the run's tenant  4ms
 ✓ reconcileOrphanedRuns > given the finder surfaces no orphaned runs > emits nothing  1ms
 ✓ reconcileOrphanedRuns > given emitting the terminal event fails for one run
     > still reconciles the remaining runs and reports the failure  4ms
 ✓ reconcileOrphanedRuns > given a now and an explicit threshold > passes them through to the finder  1ms
 ✓ reconcileOrphanedRuns > when no threshold is supplied
     > defaults the finder to the read-path stall threshold  3ms

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

A surfaced orphan drives `ensureFailureEventsEmitted` — the same idempotent finish path in-process child crashes use — with the orphan error, returning `{ reconciled: 1, failed: 0 }`. The run goes terminal **ERROR** (a real `finished` event), so the downstream reactors the read-time STALLED path never triggers finally fire. One failing emit does not abort the rest of the sweep.

_Env: this box — Node v24, Docker ClickHouse + Redis testcontainers (reused), `NODE_ENV=test`. Working tree restored to `d63d36ce0` after the falsifiability flip (`git status` clean)._

### DIFF
--- a/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
@@ -159,6 +159,29 @@ export interface SimulationRunState extends Projection<SimulationRunStateData> {
   data: SimulationRunStateData;
 }
 
+/**
+ * Guards a non-terminal Status transition once a run is already finished.
+ *
+ * Orphaned-run reconciliation writes a terminal `finished` event for a run
+ * whose worker died. If that worker's child process actually outlived its
+ * parent (reparented) and later POSTs a real started/snapshot whose
+ * client-supplied `occurredAt` is AFTER the reconciliation time, the event
+ * applies in-order (the executor only re-folds when occurredAt is STRICTLY
+ * less than what we've already seen) and would otherwise clobber Status back to
+ * a non-terminal value while FinishedAt stays set — an unrecoverable zombie the
+ * read-time stall path can no longer rescue (it only resolves runs with no
+ * FinishedAt). Once FinishedAt is set, Status stays terminal.
+ */
+function statusAfter({
+  state,
+  candidate,
+}: {
+  state: SimulationRunStateData;
+  candidate: string;
+}): string {
+  return state.FinishedAt != null ? state.Status : candidate;
+}
+
 const simulationRunEvents = [
   SimulationRunQueuedEventSchema,
   SimulationRunStartedEventSchema,
@@ -255,7 +278,7 @@ export class SimulationRunStateFoldProjection
       Name: state.Name ?? event.data.name ?? null,
       Description: state.Description ?? event.data.description ?? null,
       Metadata: state.Metadata ?? (event.data.metadata ? JSON.stringify(event.data.metadata) : null),
-      Status: "IN_PROGRESS",
+      Status: statusAfter({ state, candidate: "IN_PROGRESS" }),
       StartedAt: event.occurredAt,
     };
   }
@@ -322,7 +345,10 @@ export class SimulationRunStateFoldProjection
         };
       }),
       TraceIds: Array.isArray(event.data.traceIds) ? event.data.traceIds : [],
-      Status: event.data.status ?? state.Status,
+      Status: statusAfter({
+        state,
+        candidate: event.data.status ?? state.Status,
+      }),
     };
   }
 
@@ -357,7 +383,10 @@ export class SimulationRunStateFoldProjection
     return {
       ...state,
       ScenarioRunId: state.ScenarioRunId || event.data.scenarioRunId,
-      Status: state.Status === "PENDING" ? "IN_PROGRESS" : state.Status,
+      Status: statusAfter({
+        state,
+        candidate: state.Status === "PENDING" ? "IN_PROGRESS" : state.Status,
+      }),
       StartedAt: state.StartedAt ?? event.occurredAt,
       Messages: messages,
     };
--- a/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
@@ -170,7 +170,19 @@ export interface SimulationRunState extends Projection<SimulationRunStateData> {
  * less than what we've already seen) and would otherwise clobber Status back to
  * a non-terminal value while FinishedAt stays set — an unrecoverable zombie the
  * read-time stall path can no longer rescue (it only resolves runs with no
- * FinishedAt). Once FinishedAt is set, Status stays terminal.
+ * FinishedAt).
+ *
+ * Scope: this guards the non-terminal transition sites (started/snapshot/
+ * message) only. `handleSimulationRunFinished` deliberately does NOT route
+ * through it, so a `finished` event still sets Status unconditionally. Two
+ * consequences, both intentional-for-now and tracked on the PR:
+ *   - A later `finished` overwrites an earlier one, so a child that outlives a
+ *     reconciled parent can take an ERROR back to SUCCESS. That self-heal is
+ *     what protects a run the reconciler failed early by mistake.
+ *   - The ingest route's `finished` schema accepts any ScenarioRunStatus, so a
+ *     client POSTing `finished(status=IN_PROGRESS)` writes a non-terminal Status
+ *     alongside FinishedAt — the very zombie described above, through the one
+ *     door this guard does not cover.
  */
 function statusAfter({
   state,
--- a/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
@@ -174,12 +174,16 @@ export interface SimulationRunState extends Projection<SimulationRunStateData> {
  *
  * Once FinishedAt is set, Status stays terminal. Three things hold that line
  * together, and all three are load-bearing:
- *   1. this guard, at the non-terminal transition sites (started/snapshot/
- *      message);
+ *   1. this guard, at EVERY non-terminal Status writer — queued, started,
+ *      snapshot, and textMessageStart. Miss one and the invariant is gone: a
+ *      `queued` event folded after `finished` used to resurrect Status=QUEUED
+ *      with FinishedAt still set. If you add a handler that writes a
+ *      non-terminal Status, it goes through here too;
  *   2. `handleSimulationRunFinished` returning early once FinishedAt is set, so
  *      a run finishes exactly once;
  *   3. that same handler refusing a non-terminal explicit status, since the
- *      ingest schema types it as the full ScenarioRunStatus enum.
+ *      finished event's `status` is only typed `z.string().optional()` on the
+ *      internal event schema — any string can reach the fold.
  */
 function statusAfter({
   state,
@@ -286,7 +290,7 @@ export class SimulationRunStateFoldProjection
       BatchRunId: event.data.batchRunId,
       ScenarioSetId: event.data.scenarioSetId,
       Name: event.data.name ?? null,
-      Status: "QUEUED",
+      Status: statusAfter({ state, candidate: "QUEUED" }),
       Description: event.data.description ?? null,
       Metadata: event.data.metadata ? JSON.stringify(event.data.metadata) : null,
       QueuedAt: event.occurredAt,
--- a/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
@@ -172,17 +172,14 @@ export interface SimulationRunState extends Projection<SimulationRunStateData> {
  * read-time stall path can no longer rescue (it only resolves runs with no
  * FinishedAt).
  *
- * Scope: this guards the non-terminal transition sites (started/snapshot/
- * message) only. `handleSimulationRunFinished` deliberately does NOT route
- * through it, so a `finished` event still sets Status unconditionally. Two
- * consequences, both intentional-for-now and tracked on the PR:
- *   - A later `finished` overwrites an earlier one, so a child that outlives a
- *     reconciled parent can take an ERROR back to SUCCESS. That self-heal is
- *     what protects a run the reconciler failed early by mistake.
- *   - The ingest route's `finished` schema accepts any ScenarioRunStatus, so a
- *     client POSTing `finished(status=IN_PROGRESS)` writes a non-terminal Status
- *     alongside FinishedAt — the very zombie described above, through the one
- *     door this guard does not cover.
+ * Once FinishedAt is set, Status stays terminal. Three things hold that line
+ * together, and all three are load-bearing:
+ *   1. this guard, at the non-terminal transition sites (started/snapshot/
+ *      message);
+ *   2. `handleSimulationRunFinished` returning early once FinishedAt is set, so
+ *      a run finishes exactly once;
+ *   3. that same handler refusing a non-terminal explicit status, since the
+ *      ingest schema types it as the full ScenarioRunStatus enum.
  */
 function statusAfter({
   state,
@@ -192,6 +189,25 @@ function statusAfter({
   candidate: string;
 }): string {
   return state.FinishedAt != null ? state.Status : candidate;
+}
+
+/**
+ * The statuses a run may hold once FinishedAt is set. A `finished` event whose
+ * explicit status is outside this set is not describing a finished run, so its
+ * status is discarded in favour of the verdict-derived one — writing it would
+ * strand the run non-terminal but finished, which nothing reconciles.
+ */
+const TERMINAL_STATUSES = new Set([
+  "SUCCESS",
+  "FAILURE",
+  "FAILED",
+  "ERROR",
+  "CANCELLED",
+  "STALLED",
+]);
+
+function isTerminalStatus(status: string): boolean {
+  return TERMINAL_STATUSES.has(status);
 }
 
 const simulationRunEvents = [
@@ -482,13 +498,26 @@ export class SimulationRunStateFoldProjection
     event: SimulationRunFinishedEvent,
     state: SimulationRunStateData,
   ): SimulationRunStateData {
+    // A run finishes exactly once. A second `finished` — a child that outlived
+    // the parent this run's orphan reconciliation already failed — must not
+    // rewrite a terminal record the downstream reactors have already acted on,
+    // nor split it (an ERROR Status carrying the late child's SUCCESS Verdict).
+    if (state.FinishedAt != null) return state;
+
     const results = event.data.results;
     const verdict = results?.verdict ?? null;
 
-    // Derive status: explicit status takes priority, otherwise derive from verdict
+    // Derive status: an explicit TERMINAL status takes priority, otherwise
+    // derive from verdict. The explicit status arrives from the scenario-events
+    // ingest route, whose schema types it as the full ScenarioRunStatus enum —
+    // non-terminal members included. Taking it at face value would write a
+    // non-terminal Status alongside FinishedAt below, which is the one state
+    // nothing can recover: the orphan reconciler skips it (FinishedAt IS NULL)
+    // and read-time stall detection skips it (it only resolves unfinished runs).
     let status: string;
-    if (event.data.status) {
-      status = event.data.status.toUpperCase();
+    const explicit = event.data.status?.toUpperCase();
+    if (explicit && isTerminalStatus(explicit)) {
+      status = explicit;
     } else if (verdict === "success") {
       status = "SUCCESS";
     } else if (verdict === "failure" || verdict === "inconclusive") {
--- a/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
@@ -1,0 +1,171 @@
+import type { ClickHouseClient } from "@clickhouse/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import {
+  startTestContainers,
+  stopTestContainers,
+} from "../../event-sourcing/__tests__/integration/testContainers";
+import { ClickHouseOrphanedRunFinder } from "../orphaned-run-reconciliation.clickhouse";
+import { STALL_THRESHOLD_MS } from "../stall-detection";
+
+const tenantId = `test-orphan-${nanoid()}`;
+const NOW = Date.now();
+const minutesAgo = (m: number) => new Date(NOW - m * 60_000);
+
+// Default row: a STALE, non-terminal (QUEUED) run — i.e. an orphan. Override
+// timestamps / Status / FinishedAt / ArchivedAt to build the other cases.
+function makeRow(overrides: Record<string, unknown> = {}) {
+  return {
+    ProjectionId: `proj-${nanoid()}`,
+    TenantId: tenantId,
+    ScenarioRunId: `run-${nanoid()}`,
+    ScenarioId: `scenario-${nanoid()}`,
+    BatchRunId: `batch-${nanoid()}`,
+    ScenarioSetId: `set-${nanoid()}`,
+    Version: "v1",
+    Status: "QUEUED",
+    Name: "Test run",
+    Description: null,
+    Metadata: null,
+    "Messages.Id": [],
+    "Messages.Role": [],
+    "Messages.Content": [],
+    "Messages.TraceId": [],
+    "Messages.Rest": [],
+    TraceIds: [],
+    Verdict: null,
+    Reasoning: null,
+    MetCriteria: [],
+    UnmetCriteria: [],
+    Error: null,
+    DurationMs: null,
+    TotalCost: null,
+    RoleCosts: {},
+    RoleLatencies: {},
+    TraceMetricsJson: "",
+    StartedAt: minutesAgo(60),
+    QueuedAt: null,
+    CreatedAt: minutesAgo(60),
+    UpdatedAt: minutesAgo(60),
+    FinishedAt: null,
+    ArchivedAt: null,
+    CancellationRequestedAt: null,
+    LastSnapshotOccurredAt: new Date(0),
+    LastEventOccurredAt: new Date(0),
+    ...overrides,
+  };
+}
+
+async function insertRows(
+  ch: ClickHouseClient,
+  rows: ReturnType<typeof makeRow>[],
+) {
+  await ch.insert({
+    table: "simulation_runs",
+    values: rows,
+    format: "JSONEachRow",
+    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+  });
+}
+
+let ch: ClickHouseClient;
+let finder: ClickHouseOrphanedRunFinder;
+
+beforeAll(async () => {
+  const containers = await startTestContainers();
+  ch = containers.clickHouseClient;
+  finder = new ClickHouseOrphanedRunFinder(ch);
+}, 60_000);
+
+afterAll(async () => {
+  if (ch) {
+    await ch.exec({
+      query: `ALTER TABLE simulation_runs DELETE WHERE TenantId = {tenantId:String}`,
+      query_params: { tenantId },
+    });
+  }
+  await stopTestContainers();
+});
+
+describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
+  describe("given a mix of stale, healthy, terminal and archived runs", () => {
+    it("surfaces only the stale non-terminal run, with the ids needed to finish it", async () => {
+      const orphan = makeRow({
+        ScenarioRunId: "orphan-stale",
+        ScenarioId: "scenario-orphan",
+        BatchRunId: "batch-orphan",
+        ScenarioSetId: "set-orphan",
+        Status: "QUEUED",
+      });
+      const healthy = makeRow({
+        ScenarioRunId: "healthy-recent",
+        Status: "IN_PROGRESS",
+        StartedAt: minutesAgo(1),
+        CreatedAt: minutesAgo(1),
+        UpdatedAt: minutesAgo(1),
+      });
+      const finished = makeRow({
+        ScenarioRunId: "already-finished",
+        Status: "SUCCESS",
+        Verdict: "success",
+        FinishedAt: minutesAgo(60),
+      });
+      const archived = makeRow({
+        ScenarioRunId: "archived-run",
+        Status: "QUEUED",
+        ArchivedAt: minutesAgo(60),
+      });
+
+      await insertRows(ch, [orphan, healthy, finished, archived]);
+
+      const result = await finder.findOrphanedRuns({
+        now: NOW,
+        thresholdMs: STALL_THRESHOLD_MS,
+      });
+      const mine = result.filter((r) => r.tenantId === tenantId);
+      const ids = mine.map((r) => r.scenarioRunId);
+
+      expect(ids).toContain("orphan-stale");
+      expect(ids).not.toContain("healthy-recent");
+      expect(ids).not.toContain("already-finished");
+      expect(ids).not.toContain("archived-run");
+
+      const orphanRow = mine.find((r) => r.scenarioRunId === "orphan-stale");
+      expect(orphanRow).toEqual({
+        tenantId,
+        scenarioRunId: "orphan-stale",
+        scenarioId: "scenario-orphan",
+        batchRunId: "batch-orphan",
+        scenarioSetId: "set-orphan",
+        status: "QUEUED",
+      });
+    });
+  });
+
+  describe("given a run that was queued long ago but later finished", () => {
+    it("does not surface it — the latest version decides", async () => {
+      const scenarioRunId = `multiversion-${nanoid()}`;
+      await insertRows(ch, [
+        makeRow({
+          ScenarioRunId: scenarioRunId,
+          Status: "QUEUED",
+          UpdatedAt: minutesAgo(60),
+        }),
+        makeRow({
+          ScenarioRunId: scenarioRunId,
+          Status: "SUCCESS",
+          Verdict: "success",
+          FinishedAt: minutesAgo(50),
+          UpdatedAt: minutesAgo(50),
+        }),
+      ]);
+
+      const result = await finder.findOrphanedRuns({
+        now: NOW,
+        thresholdMs: STALL_THRESHOLD_MS,
+      });
+
+      expect(result.map((r) => r.scenarioRunId)).not.toContain(scenarioRunId);
+    });
+  });
+});
--- a/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
@@ -12,8 +12,11 @@ const tenantId = `test-orphan-${nanoid()}`;
 const NOW = Date.now();
 const minutesAgo = (m: number) => new Date(NOW - m * 60_000);
 
-// Default row: a STALE, non-terminal (QUEUED) run — i.e. an orphan. Override
-// timestamps / Status / FinishedAt / ArchivedAt to build the other cases.
+// Default row: a STALE, IN_PROGRESS run — i.e. an orphan a dead worker had
+// already started. Override timestamps / Status / FinishedAt / ArchivedAt to
+// build the other cases. A stale QUEUED run is deliberately NOT an orphan here:
+// nothing caps queue wait, so a run behind a large backlog looks identical to
+// an abandoned one. QUEUED is owned by scenario-orphan-reconciler.ts (#3365).
 function makeRow(overrides: Record<string, unknown> = {}) {
   return {
     ProjectionId: `proj-${nanoid()}`,
@@ -23,7 +26,7 @@ function makeRow(overrides: Record<string, unknown> = {}) {
     BatchRunId: `batch-${nanoid()}`,
     ScenarioSetId: `set-${nanoid()}`,
     Version: "v1",
-    Status: "QUEUED",
+    Status: "IN_PROGRESS",
     Name: "Test run",
     Description: null,
     Metadata: null,
@@ -89,13 +92,13 @@ afterAll(async () => {
 
 describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
   describe("given a mix of stale, healthy, terminal and archived runs", () => {
-    it("surfaces only the stale non-terminal run, with the ids needed to finish it", async () => {
+    it("surfaces only the stale started run, with the ids needed to finish it", async () => {
       const orphan = makeRow({
         ScenarioRunId: "orphan-stale",
         ScenarioId: "scenario-orphan",
         BatchRunId: "batch-orphan",
         ScenarioSetId: "set-orphan",
-        Status: "QUEUED",
+        Status: "IN_PROGRESS",
       });
       const healthy = makeRow({
         ScenarioRunId: "healthy-recent",
@@ -112,7 +115,7 @@ describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
       });
       const archived = makeRow({
         ScenarioRunId: "archived-run",
-        Status: "QUEUED",
+        Status: "IN_PROGRESS",
         ArchivedAt: minutesAgo(60),
       });
 
@@ -137,18 +140,45 @@ describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
         scenarioId: "scenario-orphan",
         batchRunId: "batch-orphan",
         scenarioSetId: "set-orphan",
-        status: "QUEUED",
+        status: "IN_PROGRESS",
       });
     });
   });
 
-  describe("given a run that was queued long ago but later finished", () => {
+  // The false-positive this sweep must never produce. A QUEUED run behind a
+  // large batch/suite backlog goes stale while a healthy worker is still
+  // working toward it -- nothing caps queue wait the way the child timeout caps
+  // execution. QUEUED orphans belong to scenario-orphan-reconciler.ts (#3365).
+  describe("given a stale QUEUED run waiting behind a backlog", () => {
+    it("does not surface it — queue wait is not evidence the worker died", async () => {
+      const queuedBehindBacklog = makeRow({
+        ScenarioRunId: "queued-behind-backlog",
+        Status: "QUEUED",
+        StartedAt: minutesAgo(90),
+        CreatedAt: minutesAgo(90),
+        UpdatedAt: minutesAgo(90),
+      });
+
+      await insertRows(ch, [queuedBehindBacklog]);
+
+      const result = await finder.findOrphanedRuns({
+        now: NOW,
+        thresholdMs: STALL_THRESHOLD_MS,
+      });
+
+      expect(result.map((r) => r.scenarioRunId)).not.toContain(
+        "queued-behind-backlog",
+      );
+    });
+  });
+
+  describe("given a run that was started long ago but later finished", () => {
     it("does not surface it — the latest version decides", async () => {
       const scenarioRunId = `multiversion-${nanoid()}`;
       await insertRows(ch, [
         makeRow({
           ScenarioRunId: scenarioRunId,
-          Status: "QUEUED",
+          Status: "IN_PROGRESS",
           UpdatedAt: minutesAgo(60),
         }),
         makeRow({
--- a/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
@@ -9,6 +9,7 @@ import { ClickHouseOrphanedRunFinder } from "../orphaned-run-reconciliation.clic
 import { STALL_THRESHOLD_MS } from "../stall-detection";
 
 const tenantId = `test-orphan-${nanoid()}`;
+const otherTenantId = `test-orphan-other-${nanoid()}`;
 const NOW = Date.now();
 const minutesAgo = (m: number) => new Date(NOW - m * 60_000);
 
@@ -82,10 +83,12 @@ beforeAll(async () => {
 
 afterAll(async () => {
   if (ch) {
-    await ch.exec({
-      query: `ALTER TABLE simulation_runs DELETE WHERE TenantId = {tenantId:String}`,
-      query_params: { tenantId },
-    });
+    for (const t of [tenantId, otherTenantId]) {
+      await ch.exec({
+        query: `ALTER TABLE simulation_runs DELETE WHERE TenantId = {tenantId:String}`,
+        query_params: { tenantId: t },
+      });
+    }
   }
   await stopTestContainers();
 });
@@ -169,6 +172,34 @@ describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
       expect(result.map((r) => r.scenarioRunId)).not.toContain(
         "queued-behind-backlog",
       );
+    });
+  });
+
+  // The sweep runs at boot with no tenant context, so it deliberately omits the
+  // `WHERE TenantId = ...` filter every other simulation_runs query carries.
+  // Nothing else pins that: a well-meaning "you forgot the tenant filter" fix
+  // would silently reduce the sweep to one arbitrary tenant, and every other
+  // test here uses a single tenant and would stay green.
+  describe("given stale started runs belonging to different tenants", () => {
+    it("surfaces every tenant's orphan, each attributed to its own tenant", async () => {
+      await insertRows(ch, [
+        makeRow({ ScenarioRunId: "orphan-tenant-a" }),
+        makeRow({
+          ScenarioRunId: "orphan-tenant-b",
+          TenantId: otherTenantId,
+        }),
+      ]);
+
+      const result = await finder.findOrphanedRuns({
+        now: NOW,
+        thresholdMs: STALL_THRESHOLD_MS,
+      });
+
+      const a = result.find((r) => r.scenarioRunId === "orphan-tenant-a");
+      const b = result.find((r) => r.scenarioRunId === "orphan-tenant-b");
+
+      expect(a?.tenantId).toBe(tenantId);
+      expect(b?.tenantId).toBe(otherTenantId);
     });
   });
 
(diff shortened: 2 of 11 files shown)


drewdrewthis:
## CodeRabbit review — disposition summary (audit trail)

All six CodeRabbit inline nitpicks on this PR were addressed and their threads resolved; CodeRabbit **withdrew** the four actionable ones after the rationale below. Preserving the dispositions here before clearing the now-empty inline threads.

| # | File | Finding | Disposition |
|---|------|---------|-------------|
| 1 | `orphaned-run-reconciliation.unit.test.ts` | Add `describe("when …")` layer under each `given` | **Declined** — each `given` wraps a single uniform action (reconciliation / the query running on boot); a separate `when` layer would repeat invariantly across every case rather than express a distinct condition. `given` + an action-describing `it` conveys GWT here, consistent with sibling `scenario-run-utils.unit.test.ts`. CodeRabbit withdrew. |
| 2 | `orphaned-run-reconciliation.clickhouse.integration.test.ts` | Add `describe("when …")` layer under each `given` | **Declined** — same rationale. CodeRabbit withdrew. |
| 3 | `…clickhouse.integration.test.ts:92` | Add `/** @scenario "Only stale non-terminal runs are surfaced as orphaned" */` | **Declined (would break CI)** — `scripts/check-feature-parity.ts` indexes only `SPECS_ROOT = <repo-root>/specs`; this feature lives at `langwatch/specs/scenarios/` (the established 20-file tree, not scanned), so it is not in `allKnownTitles`. Adding the annotation registers as an **unknown annotation** → `hasFatal` → feature-parity fails. Verified empirically. Scenarios stay documented in the `.feature` file + `@see` links. CodeRabbit withdrew and recorded the indexing constraint as a learning. |
| 4 | `…clickhouse.integration.test.ts:146` | Add `/** @scenario "The latest version of a run decides whether it is orphaned" */` | **Declined (would break CI)** — same scanner-indexing reason. CodeRabbit withdrew. |
| 5 | `orphaned-run-reconciliation.clickhouse.ts` | Use `as const` for `TERMINAL_STATUSES` | **Already addressed** — `TERMINAL_STATUSES` is declared `] as const;` (clickhouse.ts:49-56). Comment was outdated. |
| 6 | `simulationRunState.foldProjection.ts` | Use named-parameter object for `statusAfter` | **Already addressed** — `statusAfter({ state, candidate })` already uses object destructuring (foldProjection.ts:175). Comment was outdated. |

No code change is warranted: 1–2 are a declined style nitpick (CodeRabbit withdrew), 3–4 cannot be applied without breaking the feature-parity CI check (CodeRabbit withdrew), and 5–6 were already in the requested form. The inline threads are being removed to clear the readiness gate's top-level review-comment signal; this summary is the durable record.

drewdrewthis:

Reviewed SHA: 0cc71045157d

Findings: No correctness, concurrency, or data-loss blockers. All four load-bearing claims verified against source at this SHA:

1. **Fold guard** (`statusAfter`) is correct AND necessary. The executor (`foldProjectionExecutor.ts:61-71`) re-folds only when `occurredAt` is STRICTLY less than `LastEventOccurredAt`; a reparented child's late `started`/`snapshot` (client `occurredAt` after reconcile time) applies incrementally and would clobber `Status→IN_PROGRESS` while `FinishedAt` stays set. The guard suppresses that. Applied at all three non-terminal transition sites.

2. **Idempotency is real and load-bearing.** `FinishRunCommand` (`commands.ts:107`) keys on a static `{tenantId}:{scenarioRunId}:finishRun`; `deduplicateEvents` (`eventStoreUtils.ts:116-129`) is applied on every read. Two co-booting pods (or a racing owning-worker timeout) collapse to one terminal `finished` event via first-write-wins.

3. **Threshold is sound.** `ORPHAN_RECONCILE_THRESHOLD_MS = STALL_THRESHOLD_MS = 2× CHILD_PROCESS.TIMEOUT_MS = 30 min`. A live worker hard-caps every child at 15 min and emits its own terminal event, so a non-terminal run quiet >30 min provably has no live owner.

4. **CH query is correct.** `ReplacingMergeTree(UpdatedAt)` IN-tuple dedup picks the latest version; `PARTITION BY toYearWeek(StartedAt)` prune valid; staleness gated on `UpdatedAt`; `FinishedAt IS NULL + ArchivedAt IS NULL + terminal-status exclusion` all correct against the DDL.

SRP/design clean: finder (CH query) and reconciler (domain loop) separated behind `OrphanedRunFinder`/`OrphanFailureEmitter` interfaces. Error isolation correct: `Promise.allSettled` isolates per-run failures; boot-site `void...catch` keeps a sweep failure from blocking worker startup. No PII/secrets: query SELECTs IDs only, error messages are static, logs carry only IDs/counts.

Non-blocking (all flagged by author as follow-up):
- `TERMINAL_STATUSES` duplicates `completedStatuses`/`IN_FLIGHT_STATUSES` — worth extracting later.
- `LIMIT 1000` per sweep is bounded and matches the ADR model.
- Private-CH tenants excluded from sweep — tracked under #3195.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `12372d6c7` · **Run:** review (own-PR)

Reviewers: `principles`, `hygiene`, `security`, `test`. Every `Fix` finding they raised is addressed in `62f5e7f` / `12372d6`; the items below are `Decide`/`New Issue` and do not gate.

### What review actually caught

The fan-out found a **real defect in the invariant this PR advertises**, which the earlier passes had missed:

- **[principles][security] `handleSimulationRunQueued` was a fourth, unguarded non-terminal `Status` writer.** A `queued` event is in the fold set, so one arriving after `finished` resurrected `Status = QUEUED` with `FinishedAt` still set — reproducing the exact unrecoverable zombie the guard exists to prevent, through the one writer the guard skipped. Confirmed by executing the fold (`Status=QUEUED, FinishedAt=3000`), not by reading it. Fixed, and pinned by a regression test verified to fail without the guard (`× keeps the terminal status instead of resurrecting QUEUED`). ADR-010 corrected: four writers, not three.

- **[test] One guard test was vacuous.** `when a later text_message_start arrives` could never go red — at that call site the status candidate is already `state.Status` once the run is finished, so removing `statusAfter` does not change the outcome. The reviewer proved this by reverting that one call site and re-running. The test now says what it actually pins; the guard stays as defence in depth.

- **[test] The cross-tenant sweep had no test.** The sweep deliberately omits the mandatory `WHERE TenantId` filter, and every existing test used a single tenant — so a well-meaning "you forgot the tenant filter" change would have silently narrowed the sweep to one tenant with CI fully green. Now pinned by an integration test against real ClickHouse.

- **[test] The finished-status refusal only covered two of four non-terminal statuses.** Now exercised against `PENDING`/`QUEUED`/`IN_PROGRESS`/`RUNNING`.

- **[hygiene] The boot sweep fanned out up to `ORPHAN_SWEEP_LIMIT` (1000) concurrent event-store writes** via `Promise.allSettled`, on every worker boot — and a deploy co-boots every pod. Now sequential, matching the sibling sweep. Nothing here is latency-critical.

- **[security][hygiene] Our second cross-tenant sweep falsified a neighbouring comment** claiming the sibling was "the one documented exception" to the TenantId rule, against a doc with no exception clause at all. `clickhouse-queries.md` now defines the **boot-time system-sweep carve-out** and its five conditions; both sweeps are listed.

- **[hygiene]** `ORPHANABLE_STATUS` uses `ScenarioRunStatus.IN_PROGRESS` rather than a raw literal; the dedup subquery's coarser `GROUP BY` carries its KSUID justification.

### Non-blocking — Decide

1. **[hygiene][principles] The two boot reconcilers duplicate ~80% of their scaffolding.** `OrphanCandidate`/`OrphanedRun`, `{failed,skipped,errored}`/`{reconciled,failed}`, IN-tuple vs `argMax` dedup, 30d/7d lookback, 1000/10000 caps. The **status split is sound** (ADR Decision 0: the liveness argument genuinely differs, and disjointness prevents double-writes) — the *machinery* divergence is not. Rule of Three argues against forcing the abstraction at the second instance. **Decide:** reconcile vocabulary + result shape now, unify on a third case.

2. **[hygiene] `OrphanFailureEmitter` restates `FailureEmitter`.** Kept deliberately: the reconciler declares the port it needs rather than importing one from `scenario.processor.ts`, which would invert the dependency direction (the processor imports the reconciler). **Decide** whether that inversion is worth a shared port module.

3. **[hygiene] File/class naming.** `orphaned-run-reconciliation.clickhouse.ts` / `ClickHouseOrphanedRunFinder` diverge from the codebase's `*.clickhouse.repository.ts` / `<Domain>ClickHouseRepository` convention. It is a query object, not a repository — **Decide** whether "Finder" earns its own convention.

4. **[test] `langwatch/specs/` is invisible to `check-feature-parity`.** The parity gate scans the top-level `/specs/`, so this feature file's scenarios — and the six others already in that directory, including the sibling `queued-run-orphan-recovery.feature` — are never discovered or enforced. Pre-existing and repo-wide. **New Issue.**

5. **[test] `ORPHAN_SWEEP_LIMIT` + `ORDER BY UpdatedAt ASC` are untested** (would need 1000+ rows). Parameterize the limit to test at small scale. **New Issue.**

6. **[security] Tenants on private ClickHouse instances (`CLICKHOUSE_URL__*`) are silently not swept** — no metric or log line reports the exclusion. Documented in ADR-010; **Decide** whether it needs its own issue + boot-time observability signal.

### Verification

- CI green at `12372d6c7`: 16/16 runs, `langwatch-app-ci` lint + typecheck + build + feature-parity + 4 unit shards + 6 integration shards.
- Locally: `pnpm typecheck` 0 errors; 82 unit tests across the affected suites; the ClickHouse integration suite (4 tests) against real ClickHouse via testcontainers.
- Every guard in this PR was falsified before being trusted — reverted in place, observed red, restored, observed green.

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** A false-positive reconciliation marks a run ERROR that a live worker still owns, firing downstream reactors (suite aggregates, metrics, customer.io) on a still-running run.

**Inspection targets:**
- `[Must Check]` `orphaned-run-reconciliation.ts` threshold (2x child timeout = 30 min as the "no live worker" proof) — verify a slow-but-alive worker or a long-queued-but-imminent job cannot be swept.
- `[Must Check]` `simulationRunState.foldProjection.ts` `statusAfter` guard — confirm a late, client-timestamped `started`/`snapshot` posted after the reconcile cannot resurrect Status to IN_PROGRESS while FinishedAt stays set.
- `[Ask Author]` cross-tenant CH sweep excludes BYOC private-ClickHouse tenants — confirm the query cannot error or partially-scan on those.

**Author questions:**
- The sweep is fire-and-forget in `startScenarioProcessor`: if it throws, is it logged and retried next boot, or silently lost?

langwatch/src/server/scenarios/orphaned-run-reconciliation.clickhouse.ts
```diff
@@ -0,0 +1,152 @@
...
+        FROM ${TABLE_NAME}
+        WHERE ${partitionFilter}
+          AND (TenantId, ScenarioRunId, UpdatedAt) IN (
+            SELECT TenantId, ScenarioRunId, max(UpdatedAt)
+            FROM ${TABLE_NAME}
+            WHERE ${partitionFilter}
+            GROUP BY TenantId, ScenarioRunId
+          )
+          AND UpdatedAt < fromUnixTimestamp64Milli({staleBeforeMs:Int64})
+          AND FinishedAt IS NULL
```
Aryansharma28, line 125: [P0] This staleness gate can't distinguish "worker died" from "legitimately queued behind a large batch/suite backlog."

**Why:** `suite-run.service.ts` queues every scenario run in a batch up front (`Status=QUEUED`, `QueuedAt=occurredAt` written immediately for all items), and nothing refreshes `UpdatedAt` while a run waits for a concurrency slot (3 per pod — see `execution-pool.ts`). A suite larger than `(replicas × 3)` items will have tail-end items sitting `QUEUED` with a stale `UpdatedAt` well past 30 minutes, purely because they haven't been picked up yet — not because any worker crashed. This query has no signal to tell the two apart, so any worker boot during such a backlog (a routine deploy) reconciles them to a permanent `finished(ERROR)`, which is a regression from today's behavior (cosmetic STALLED paint, but the run still completes correctly once a slot frees).

**Evidence:** `suite-run.service.ts:121-145` ("Pre-generate scenarioRunIds and dispatch queueRun for each so QUEUED entries appear in ClickHouse immediately"); `simulationRunState.foldProjection.ts:250-263` (`UpdatedAt` only bumped on the next handled event, none of which fire while merely queued); the PR's own integration test builds exactly this shape (`QUEUED` + `UpdatedAt` 60 min stale) and asserts it should be reconciled (`orphaned-run-reconciliation.clickhouse.integration.test.ts:92-99`).

**Suggested fix:** exclude `QUEUED` (never-started) runs from this sweep — there's no worker to have died for a job that was never claimed — or gate staleness for queued rows on something throughput-aware instead of the fixed 30-minute threshold designed for the started/in-progress case.

langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    });
+  }
+  await stopTestContainers();
+});
+
+describe("ClickHouseOrphanedRunFinder.findOrphanedRuns (integration)", () => {
+  describe("given a mix of stale, healthy, terminal and archived runs", () => {
+    it("surfaces only the stale non-terminal run, with the ids needed to finish it", async () => {
+      const orphan = makeRow({
+        ScenarioRunId: "orphan-stale",
```
Aryansharma28, line 100: [P0] This fixture is the false-positive case, encoded as the expected/correct behavior.

**Why:** `orphan-stale` here is a `QUEUED` run with a stale `UpdatedAt` — which is indistinguishable from a run sitting deep in a legitimate batch/suite queue behind a full concurrency pool (see the inline comment on `orphaned-run-reconciliation.clickhouse.ts:109`). This test proves the current query treats "queued a long time" and "queued behind a big backlog" identically to "worker died" — there's no case here for a `QUEUED` run that's stale purely because of queue depth, which the design should not reconcile.

**Suggested fix:** add a scenario where several thousand runs are queued (or just assert on `QUEUED` specifically) and confirm the finder does NOT treat queue depth alone as orphaned — once the fix for the P0 above lands, this fixture's assertion will need to change anyway.

langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
```diff
@@ -159,6 +159,29 @@ export interface SimulationRunState extends Projection<SimulationRunStateData> {
...
+ * FinishedAt). Once FinishedAt is set, Status stays terminal.
+ */
+function statusAfter({
+  state,
+  candidate,
+}: {
+  state: SimulationRunStateData;
+  candidate: string;
+}): string {
+  return state.FinishedAt != null ? state.Status : candidate;
```
Aryansharma28, line 195: [P1] `statusAfter` guards the three non-terminal transition sites but not `handleSimulationRunFinished` (this file, ~line 469-498), which unconditionally overwrites `Status`/`Verdict`/`Reasoning`/`Error`/`FinishedAt` from whatever `finished` event arrives — including a *second* one.

**Why:** The incremental fold executor (`foldProjectionExecutor.ts:61-79`) only re-folds (triggering the read-time `deduplicateEvents` collapse) when the new event's `occurredAt` is strictly less than what's already been folded. A second `finished` event with `occurredAt >=` the first (which is the common case — a later event is later) applies directly on top of the current state with no idempotency check at all: `storeEvents` (`abstractEventStore.ts:328-402`) inserts unconditionally, and `event_log`'s `ReplacingMergeTree` (`00002_create_schema.sql:34-36`) only collapses duplicate-idempotencyKey rows at an eventual background merge — not synchronously. `FinishRunCommand.makeJobId` exists (`commands.ts:107-113`) but is never wired into a `deduplication` config in `pipeline.ts` (unlike `computeRunMetrics`, which is). Concretely: (a) two pods racing to reconcile the same orphan on a multi-replica rolling deploy can each dispatch their own `finished` event, both of which get folded and can each fire `suiteRunSync`/`customerIoSimulationSync`; (b) the exact reparented-child scenario this guard was written to defend against for non-terminal events can *also* eventually send its own real `finished` event (via `/api/scenario-events` → `finishRun`, independent of the dead parent process) that silently overwrites the reconciled ERROR with a different terminal outcome and re-fires every reactor. This contradicts ADR-010's "the finish path is idempotent... no double-fire" claim, which holds only for the full-history re-fold path, not the incremental live path normal processing uses.

**Suggested fix:** apply the same `FinishedAt != null` guard inside `handleSimulationRunFinished` — once a run is finished, a second `finished` event should be a logged no-op on the terminal fields, not a silent overwrite.

Aryansharma28 (commented):
## goated-review: deep pass on orphan reconciliation (worker-boot race conditions, false positives, double-processing)

This PR is unusually well self-reviewed (ADR-010, multiple bot passes, an author self-review, a full CodeRabbit round with documented dispositions). All of that verification focused on the *in-progress / started* run case and the *reparented-child-sends-a-late-non-terminal-event* case. Tracing the two paths those reviews didn't walk end-to-end turned up two real, previously-unflagged issues. Both are backed by file:line evidence and were adversarially checked against the actual runtime code paths (queue depth, the incremental fold executor, the event store's write path, ClickHouse's `ReplacingMergeTree` merge semantics) rather than trusting the docstrings.

**Verdict: not yet bulletproof.** Two blockers below; everything else is polish.

### Blockers (P0–P2) — must fix before merge

**[P0] False positive: legitimately-queued runs in a large batch/suite get killed as "orphaned" on any worker boot.**
Every run in a suite/batch is queued *up front*: `suite-run.service.ts` dispatches `queueSimulationRunCommand` for every `{scenario × target × repeat}` combination immediately (`langwatch/src/server/app-layer/suites/suite-run.service.ts:121-145`, comment: "Pre-generate scenarioRunIds and dispatch queueRun for each so QUEUED entries appear in ClickHouse immediately"). `handleSimulationRunQueued` writes `Status=QUEUED`, `QueuedAt=occurredAt` once (`simulationRunState.foldProjection.ts:250-263`), and `UpdatedAt` is never refreshed again until the run actually starts — there is no queue heartbeat in `execution-pool.ts` (concurrency is 3 per pod, "6 pods × 3 = 18 total" per its own comment). A suite with more items than `(pods × 3)` can drain within `ORPHAN_RECONCILE_THRESHOLD_MS` (30 min) will have tail-end items sitting in `QUEUED` with a stale `UpdatedAt` — not dead, just waiting their turn. `findOrphanedRuns` (`orphaned-run-reconciliation.clickhouse.ts:93-113`) has no way to tell "genuinely orphaned" apart from "legitimately queued behind a backlog"; it only checks `UpdatedAt < staleBefore AND FinishedAt IS NULL AND Status NOT IN TERMINAL_STATUSES`. Any worker boot during that backlog (a routine deploy is the *literal trigger condition this feature runs on*) will emit a real terminal `finished(ERROR)` — "Worker restarted or crashed before the run completed" — for scenarios that were never even picked up. This is a regression versus current behavior: today that same backlogged run just shows a cosmetic STALLED paint and still completes correctly once a slot frees; after this PR it is permanently, incorrectly failed (and fires customer.io / suite-aggregate reactors with a false failure). The PR's own integration test proves the gap: `orphaned-run-reconciliation.clickhouse.integration.test.ts:92-99` builds "orphan-stale" as `Status: "QUEUED"` + `UpdatedAt: minutesAgo(60)` and asserts it *should* be reconciled — there is no test distinguishing that from a legitimately deep queue. The ADR's threshold rationale ("A booting pod that is about to pick up a queued job could find that job quiet for well under 30 min") only considered a job about to be picked up imminently, not a batch backlog exceeding available concurrency.
*Suggested fix:* either exclude `QUEUED` runs from the sweep entirely (a run that never started can't have a dead worker to reconcile away — the crash-recovery story only applies once a worker has claimed it), or gate staleness for `QUEUED` rows on queue depth / a much longer, throughput-aware threshold, or have the queue mechanism refresh `UpdatedAt` while a run legitimately waits.

**[P1] `handleSimulationRunFinished` has no finalized-status guard — a second `finished` event silently overwrites the first and re-fires every reactor.**
The new `statusAfter` guard (`simulationRunState.foldProjection.ts:175-183`) is applied at the three *non-terminal* transition sites (`handleSimulationRunStarted`, message-snapshot, text-message-start) but **not** at `handleSimulationRunFinished` (`simulationRunState.foldProjection.ts:469-498`), which unconditionally sets `Status`, `Verdict`, `Reasoning`, `Error`, and `FinishedAt` from whatever event just arrived. The incremental executor (`foldProjectionExecutor.ts:61-79`) only re-folds (and thus only benefits from `deduplicateEvents`, which is read-time-only — `eventStoreUtils.ts:120-137`, `abstractEventStore.ts:190-191,251-252`) when the new event's `occurredAt` is *strictly less than* what's already been seen; a second `finished` event with `occurredAt >=` the first applies directly on top of the live projected state with zero idempotency check. `storeEvents` (`abstractEventStore.ts:328-402`) inserts unconditionally — there is no write-time uniqueness enforcement on `IdempotencyKey`; `event_log`'s `ReplacingMergeTree(EventTimestamp)` (`00002_create_schema.sql:34-36`) only collapses duplicate-key rows at an *eventual, asynchronous* background merge, so both rows are visible to every consumer/reactor in the interim. `FinishRunCommand.makeJobId` (`commands.ts:107-113`) is defined but never wired into a `deduplication` config in `pipeline.ts` (only `computeRunMetrics` gets one) — so there is no queue-level dedup either. Net effect, two concrete scenarios: **(a)** two pods racing to reconcile the *same* orphan on a multi-replica rolling deploy (routine, not exotic) each independently call `ensureFailureEventsEmitted` → both `finished` events get stored and incrementally folded → `suiteRunSync`/`customerIoSimulationSync` can each fire twice for one run. **(b)** the exact reparented-child scenario this PR's own `statusAfter` guard was written to defend against — a child that outlives its dead parent — can *also* eventually send a real `finished` event (the child dispatches its own `finishRun` over HTTP via `/api/scenario-events` → `app.ts:240` independent of the parent process); since its `occurredAt` postdates the reconciliation, it applies directly and silently **overwrites the reconciled ERROR with whatever the late completion reports**, re-firing every downstream reactor with different data. This directly contradicts the PR's own stated guarantee: ADR-010 "Consequences" ("The finish path is idempotent... no double-fire") and the author's self-review ("Idempotency is real and load-bearing... collapse to one terminal finished event via first-write-wins") — that claim is true only for the full-history re-fold path, not for the incremental live path that normal event processing actually uses.
*Suggested fix:* extend `statusAfter`-style guarding to `handleSimulationRunFinished` itself (once `FinishedAt` is set, a second `finished` event should be a no-op on `Status`/`Verdict`/`FinishedAt`, logged and dropped), and/or give `finishRun` real command-level dedup the same way `computeRunMetrics` has one.

### Non-blockers (P3–P5) — optional polish

**[P3]** `ORPHAN_SWEEP_LIMIT = 1000` (`orphaned-run-reconciliation.clickhouse.ts:37,112`) is applied with no `ORDER BY` before `LIMIT`. If a backlog ever exceeds 1000 orphans, which 1000 get swept is arbitrary/implementation-defined per boot and could keep re-selecting the same subset across restarts instead of eventually draining the rest. Low likelihood, easy to fix with an `ORDER BY UpdatedAt ASC` to at least guarantee oldest-first progress.

**[P4]** `TERMINAL_STATUSES` includes `"STALLED"` (`orphaned-run-reconciliation.clickhouse.ts:49-57`), but `stall-detection.ts` explicitly documents STALLED as a read-time-only derived status that is "pure... does not write any new events" — it should never actually appear as a persisted `Status` value. Harmless as belt-and-suspenders (already flagged as such in the comment) but worth double-checking nothing ever does persist it, since its presence here implies someone once thought it might.

**[P5]** Nice discipline overall: the `Promise.allSettled` isolation in `reconcileOrphanedRuns` (one failing emit doesn't abort the sweep), the fire-and-forget wiring at boot so a slow sweep can't block worker startup, and the IN-tuple ClickHouse dedup pattern are all solid and match the codebase's established idioms. The `statusAfter` guard is exactly the right shape — it's just missing one call site.

- Actual production suite/batch sizes and worker replica counts — the P0 above assumes suites large enough to exceed `(replicas × concurrency=3)` throughput within 30 minutes, which I couldn't confirm against real usage data, only against the absence of any cap in `suite-run.service.ts`.
- Whether any downstream reactor (`suiteRunSync`, `customerIoSimulationSync`) has its own idempotency check I didn't trace all the way through (e.g. the suite-processing pipeline's own `itemCompleted` command's `makeJobId`) that might partially dampen the P1's blast radius — I checked `pipeline.ts`'s dedup wiring but not every downstream command's queue configuration.

drewdrewthis:
## Rebase onto `main`: this PR now has a sibling reconciler, and it changes the P0 discussion

Rebasing this branch onto `origin/main` (251 commits) surfaced a **semantic** conflict, not just a textual one, that reviewers of the pre-rebase diff could not have seen.

### `main` landed its own orphan reconciler while this branch sat

[#5008](https://github.com/langwatch/langwatch/pull/5008) (closing #3365) added `scenario-orphan-reconciler.ts`, wired into `startScenarioProcessor` — the same boot hook this PR wires. Neither reconciler subsumes the other:

| | `#5008` (on `main`) | this PR (#3195) |
|---|---|---|
| Sweeps | `QUEUED` only | any non-terminal (`PENDING`/`QUEUED`/`IN_PROGRESS`) |
| Lookback | 7 days | 30 days |
| Cap | 10,000, oldest-first | 1,000, unordered |
| Archived runs | reconciled | skipped (`ArchivedAt IS NULL`) |

I resolved the conflict as a **union — both sweeps wired**. Deleting either would silently drop landed behaviour (`#5008`'s oldest-first ordering and 10k cap; this PR's `IN_PROGRESS` coverage). They overlap on the `QUEUED` slice. That overlap is safe: both reach terminal through the same `FinishRunCommand`, whose idempotency key is derived solely from `(tenantId, scenarioRunId)`, so a duplicate finish cannot change the resulting `Status`. The one observable divergence is the `Error` string.

**This is a scope change relative to what was reviewed** — flagging rather than quietly shipping it. If we'd rather consolidate onto one reconciler, that is a deliberate follow-up, not something I should decide inside a conflict resolution.

### The `QUEUED` backlog P0 now also indicts already-merged code

@Aryansharma28's [P0 on the staleness gate](https://github.com/langwatch/langwatch/pull/5006#discussion_r3542927878) — that it cannot distinguish "worker died" from "legitimately queued behind a large batch/suite backlog" — is, as far as I can tell, **equally true of `#5008`, which is already on `main`**. Its `isOrphanedQueuedRun` gate is `status === QUEUED && now - lastEventAtMs >= STALL_THRESHOLD_MS`, with no backlog guard, and the same `STALL_THRESHOLD_MS` constant.

So the exposure the P0 describes for the `QUEUED` slice is already shipped. That doesn't make the objection wrong — it makes it **bigger than this PR**, and it means "which reconciler owns `QUEUED`" is a decision rather than a detail. I have not changed the gate here.

### The P1 fold-guard hole is real, and worse than described — confirmed by executing the fold

@Aryansharma28 is right that `statusAfter` guards the three non-terminal sites but not `handleSimulationRunFinished`. I executed the fold to check rather than reason about it:

```
PROBE A  apply(started) -> apply(finished{status:"IN_PROGRESS"})
         => Status = IN_PROGRESS, FinishedAt = 3000

PROBE B  apply(started) -> apply(finished{status:"ERROR"}) -> apply(finished{status:"SUCCESS"})
         => ERROR, then SUCCESS
```

PROBE A is the **unrecoverable zombie this PR exists to prevent**, reached through the one door the guard doesn't cover: the scenario-events ingest schema types the finished status as `z.nativeEnum(ScenarioRunStatus)`, which admits `IN_PROGRESS`. A run in that state has `FinishedAt` set, so the reconciler's `FinishedAt IS NULL` gate can never rescue it, and read-time stall detection can't either.

I did **not** silently fix it. Routing `handleSimulationRunFinished` through `statusAfter` would close PROBE B — but PROBE B's overwrite is exactly what lets a run that the reconciler failed *by mistake* self-heal when its child finishes for real. Closing it therefore trades one failure mode for another, and it is entangled with the P0 above. That's a call for the reviewer and the author, not for a rebase.

What I did do: the `statusAfter` docblock and ADR-010 both asserted *"once `FinishedAt` is set, `Status` stays terminal."* That claim is false. Both now state what the guard actually guarantees and name the two open paths (78d7bd0).

### State

`mergeable: CONFLICTING` → `MERGEABLE`. CI green at `bc967ce` across all 17 runs (`langwatch-app-ci`: lint, typecheck, unit, integration, feature-parity, build). The three threads above are left **unresolved on purpose** — two of them need a design call I shouldn't make unilaterally.

langwatch/src/server/scenarios/orphaned-run-reconciliation.clickhouse.ts
drewdrewthis: Confirmed as a real gap, and it reaches further than this PR.

The identical gate is already on `main`: `#5008` (closing #3365) landed `scenario-orphan-reconciler.ts`, whose `isOrphanedQueuedRun` is `status === QUEUED && now - lastEventAtMs >= STALL_THRESHOLD_MS` — same threshold constant, no backlog guard. So a run queued behind a large batch is already exposed on `main` today, independent of this branch.

That reframes the decision: the `QUEUED` slice is now swept by two reconcilers (see the rebase comment on the PR), and whichever one owns it needs the backlog guard. I have not changed the gate here — narrowing it inside a conflict resolution would be the wrong place to make that call. Deferring to you on whether the guard lands in this PR, in a follow-up against `#5008`, or by consolidating the two sweeps.

langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
drewdrewthis: Agreed — the fixture encodes the false-positive as expected behaviour, and I have left it that way rather than quietly rewrite the assertion.

Context that changes the fix, though: this fixture exercises the `QUEUED` slice, which `main` now also sweeps via `#5008`/`scenario-orphan-reconciler.ts` with the same staleness gate. So "is a stale `QUEUED` run an orphan?" is currently answered `yes` by already-merged code too. Rewriting only this fixture would leave the two sweeps asserting opposite things about the same slice.

Tied to the parent P0 — once the backlog guard is decided, this fixture (and `#5008`'s equivalent) should move together.

langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
drewdrewthis: You are right, and it is worse than "overwrites". I executed the fold rather than reason about it:

```
PROBE A  apply(started) -> apply(finished{status:"IN_PROGRESS"})
         => Status = IN_PROGRESS, FinishedAt = 3000

PROBE B  apply(started) -> apply(finished{status:"ERROR"}) -> apply(finished{status:"SUCCESS"})
         => ERROR, then SUCCESS
```

PROBE A is the exact zombie this PR exists to prevent, reached through the one door the guard does not cover: the scenario-events ingest schema types the finished status as `z.nativeEnum(ScenarioRunStatus)`, which admits `IN_PROGRESS`. The resulting row has `FinishedAt` set, so the reconciler’s `FinishedAt IS NULL` gate can never rescue it, and read-time stall detection cannot either.

I did not route `handleSimulationRunFinished` through `statusAfter`, because that closes PROBE B — and PROBE B is what lets a run this reconciler failed *by mistake* self-heal when its child finishes for real. Closing it trades one failure mode for another and is entangled with your P0 on the staleness gate, so it wants your call, not a rebase-time decision.

What I did fix: the `statusAfter` docblock and ADR-010 both claimed *"once `FinishedAt` is set, `Status` stays terminal."* That is false. Both now state the guard’s real scope and name the two open paths (78d7bd0).

langwatch/src/server/scenarios/orphaned-run-reconciliation.clickhouse.ts
drewdrewthis: Fixed in d7ea5dd — you were right, and the fix is narrower than I first thought.

The sweep now only reconciles `IN_PROGRESS`. Working through your objection made it clear the threshold argument itself is what constrains the scope: "a worker hard-caps every child at the child timeout and emits its own terminal event at the cap, so an idle non-terminal run has no live worker" only closes for a run that was *started*. Nothing bounds queue wait — the pool is concurrency-bounded — so a `QUEUED` run behind a draining backlog is indistinguishable from an abandoned one. The premise never supported sweeping `QUEUED` in the first place.

That also makes the two boot reconcilers disjoint by status: `scenario-orphan-reconciler` (#3365) owns `QUEUED`, this one owns `IN_PROGRESS`, and neither can double-write a run. I also added `ORDER BY UpdatedAt ASC` so the 1k sweep cap drops the least urgent orphans rather than an arbitrary slice.

The `QUEUED` backlog exposure still exists in `#5008` on `main`, which uses the same gate. Out of scope here, flagged on the PR.

langwatch/src/server/scenarios/__tests__/orphaned-run-reconciliation.clickhouse.integration.test.ts
drewdrewthis: Fixed in d7ea5dd. The fixture no longer encodes the false positive.

`makeRow` now defaults to `IN_PROGRESS`, and the case you pointed at asserts the opposite of what it used to:

```
describe("given a stale QUEUED run waiting behind a backlog", () => {
  it("does not surface it — queue wait is not evidence the worker died", ...)
```

The old `orphan-stale` row is now `IN_PROGRESS`, and the multi-version case starts from `IN_PROGRESS` too so it still exercises "latest version decides". Spec updated to match (`orphaned-run-reconciliation.feature`).

langwatch/src/server/event-sourcing/pipelines/simulation-processing/projections/simulationRunState.foldProjection.ts
drewdrewthis: Fixed in d7ea5dd. I checked by executing the fold rather than reading it, and it was worse than "overwrites":

```
PROBE A  started -> finished{status:"IN_PROGRESS"}   => Status=IN_PROGRESS, FinishedAt=3000
PROBE B  started -> finished{ERROR} -> finished{SUCCESS} => ERROR, then SUCCESS
```

PROBE A is the unrecoverable zombie this PR exists to prevent, through the one door the guard cannot cover — `handleSimulationRunFinished` is what *sets* `FinishedAt`. The ingest schema types the status as `z.nativeEnum(ScenarioRunStatus)`, so `IN_PROGRESS` is accepted and written straight through; the resulting row is skipped by the reconciler (`FinishedAt IS NULL`) and by read-time stall detection.

Both are closed, and neither by routing through `statusAfter` (which would not have caught PROBE A — on the *first* finished, `FinishedAt` is still null):

1. `handleSimulationRunFinished` returns early once `FinishedAt` is set — a run finishes exactly once. This also stops the record splitting (an ERROR `Status` carrying the late child's SUCCESS `Verdict`).
2. A non-terminal explicit status is refused in favour of the verdict-derived one.

I had initially argued for keeping PROBE B's overwrite as a "self-heal" for a falsely-reconciled run. Scoping the sweep to `IN_PROGRESS` (your other P0) removes the false positive that justified it, so the overwrite is now pure downside — it discards a terminal state the downstream reactors already acted on.

Three regression tests, each verified to fail against the un-fixed fold:
```
× keeps the first terminal record and ignores the late one
× refuses the non-terminal status and finishes the run terminally
× derives the terminal status from the verdict
Tests  3 failed | 40 passed
```
`statusAfter`'s docblock and ADR-010 now state the invariant that actually holds, and name all three things holding it.

drewdrewthis:

Aryansharma28 (approved):
Reviewed the full 6-commit series end-to-end. Design is sound and the terminal-status invariant — the thing that could actually hurt — holds. Approving.

**What I verified**

- **IN_PROGRESS-only scoping is the right cut.** `ORPHANABLE_STATUS = IN_PROGRESS` (commit 4) means QUEUED runs that never started aren't force-failed, and already-terminal runs are untouched. The reconciler only rescues runs that were genuinely mid-flight when a worker died.
- **No zombie can be resurrected.** The `statusAfter({state, candidate})` guard is applied at all four non-terminal writers (queued/started/snapshot/textMessageStart), and `handleSimulationRunFinished` early-returns when `FinishedAt != null`. I walked the fallback chain: a non-terminal explicit status + null verdict lands on `FAILURE`, not a stuck non-terminal state — so there's no path that leaves a run neither running nor terminal.
- **The cross-tenant sweep is memory-safe and documented.** IN-tuple dedup on `(TenantId, ScenarioRunId, UpdatedAt)` keeps heavy columns out of the granule scan, partition filter on `StartedAt` (30-day lookback) prunes cold partitions, and `ORDER BY UpdatedAt ASC LIMIT 1000` bounds the batch. The boot-time carve-out from the mandatory-`TenantId` rule is written up in `clickhouse-queries.md` with its five conditions and mirrored in ADR 010 — so the next person who reads the rule won't think this query violates it.
- **Sequential emit** avoids hammering the failure path on a large backlog. Fine at LIMIT 1000.
- CI is green (e2e / analyze / build) and there are no unresolved review threads.

**Non-blocking**

1. `TERMINAL_STATUSES` is duplicated across ~3 spots — worth extracting to a shared const in a follow-up (the code already flags this).
2. The `else { status = "FAILURE" }` terminal default is the correct conservative choice; just calling it out so it's a deliberate decision on the record.

Nice work — the invariant coverage and the carve-out documentation are exactly what this needed.

Issue #2668, opened 2026-03-26
fix(analytics): cross-evaluator groupBy + metric conflict produces empty charts


drewdrewthis:
## Summary

After #2636 (merged today), `evaluation_label` groupBy now correctly scopes labels to a specific evaluator via `additionalWhere: EvaluatorId = {groupByKey}`. However, this introduces a new bug: when the series metric targets a **different** evaluator than the groupBy key, all metric values return empty.

## Symptoms

- **Summary charts**: Group labels appear (Recommendation, Operational, Insight, etc.) but each label's metric value is `{}` (empty object)
- **Line charts**: "No data" shown despite traces existing
- Both current and previous periods are affected

Example API response:
```json
{
  "currentPeriod": [{
    "date": "full",
    "evaluations.evaluation_label": {
      "Recommendation": {},
      "Operational": {},
      "Insight": {},
      "Optimization": {}
    }
  }]
}
```

## Root Cause

`buildTimeseriesQuery` in `aggregation-builder.ts` applies `groupByAdditionalWhere` globally to the SQL WHERE clause:

```ts
if (groupByAdditionalWhere) {
  filterWhere += ` AND ${groupByAdditionalWhere}`;
}
```

For `evaluation_label` with `groupByKey = evaluatorA`, this adds:
```sql
WHERE ... AND es.EvaluatorId = 'evaluatorA' AND es.Status = 'processed'
```

This pre-filters the entire `evaluation_runs` JOIN to evaluatorA rows. When the series metric is `evaluations.evaluation_score` with `key = evaluatorB`, its conditional aggregation:
```sql
avgIf(es.Score, es.EvaluatorId = 'evaluatorB' AND es.Status = 'processed')
```
...never matches any rows because evaluatorB was already excluded by the WHERE clause. `avgIf` returns NULL → skipped in result parsing → empty `{}` per group.

## Contrast with Elasticsearch Path

The ES path in `registry.ts` handles this correctly via nested aggregation + `reverse_nested`:

1. Enter nested context for `evaluations`, filter by evaluatorA, bucket by label
2. `reverse_nested` back to root trace document
3. Re-enter nested context with the metric's own evaluator filter (evaluatorB)

This two-level nesting allows independent evaluator filters for groupBy and metric. The CH flat JOIN + global WHERE cannot replicate this architecture.

## Affected Combinations

The same root cause affects:
- `evaluations.evaluation_label` groupBy + any evaluation metric with a different `key`
[...]

PR #5007, opened 2026-06-21
fix(analytics): include score-only evaluators in evaluation_passed groupBy

### LINKED ISSUE #2674: fix(analytics): score-only evaluators excluded from evaluation_passed groupBy with groupByKey
## Problem

When `evaluation_passed` groupBy has a `groupByKey`, the CASE expression returns `NULL` for rows where `Passed IS NULL`:
```sql
CASE
  WHEN match AND Passed = 1 THEN 'passed'
  WHEN match AND Passed = 0 THEN 'failed'
  ELSE NULL
END
```

Score-only evaluators (Status='processed', Passed=NULL) produce NULL group_key → dropped by `HAVING group_key IS NOT NULL`. These traces vanish entirely from analytics.

Without `groupByKey`, the same traces would appear under `'unknown'` via the `ELSE 'unknown'` branch.

## Fix direction

Add a third WHEN branch for the groupByKey case:
```sql
WHEN match AND Passed IS NULL THEN 'unknown'
```

This keeps the `'unknown'` bucket consistent between with-key and without-key paths.

## Context

- Discovered during edge-case analysis for #2668
- The ES path also excludes these rows (no `missing:` on the terms aggregation), so this matches ES behavior — but the inconsistency between with-key and without-key in ClickHouse is confusing
- Companion to #2673 (NULL labels)

drewdrewthis:
## Why

Score-only evaluators (which emit a numeric score but no pass/fail verdict) were silently dropped from the **evaluation_passed** analytics groupBy whenever the chart was scoped to a specific evaluator (`groupByKey`). On the eval analytics dashboard, a chart grouped by such an evaluator rendered **empty** instead of showing that evaluator's runs.

Closes #2674

## What changed

- **One new `WHEN` in the `evaluation_passed` groupBy CASE** — `aggregation-builder.ts` returned `ELSE NULL` for every row that wasn't a clear pass/fail, and the downstream `HAVING group_key IS NOT NULL` dropped them. That `NULL` is intentional for **foreign** evaluators (the #2668 isolation guarantee), but it also dropped the **target** evaluator's score-only rows. The new clause buckets only the target evaluator's processed `Passed IS NULL` rows as `'unknown'`.
- **Gated on `Passed IS NULL`, not a broad `ELSE`** — a naive `ELSE 'unknown'` would have re-admitted foreign-evaluator rows and broken the #2668 isolation guarantee. The clause is narrow so foreign rows and unprocessed rows still hit `ELSE NULL` and stay excluded. (Decision recorded below under *Anything surprising?*.)

## How it works

The two existing `WHEN`s already gate on `Passed IS NOT NULL = 1/0`, so the new clause

```sql
WHEN es.EvaluatorId = {groupByKey:String} AND es.Status = 'processed' AND es.Passed IS NULL THEN 'unknown'
```

only catches the remaining case — the target evaluator's processed rows with `Passed IS NULL` (score-only). Everything else falls through to `ELSE NULL`, so evaluator isolation is preserved. The doc comment above the CASE documents exactly this NULL-only behavior, so comment and predicate now agree.

The three changed files are the backend query builder and its tests — but the query they build is **not** an internal detail: it is the data source for the eval analytics dashboard (surface traced under *How I can prove I was successful*).

```
src/server/analytics/clickhouse/
├── aggregation-builder.ts                                    # the one-WHEN fix
└── __tests__/
    ├── aggregation-builder.test.ts                           # unit: asserts the full WHEN (EvaluatorId + Passed IS NULL) + ELSE NULL
    └── score-only-evaluator-passed-groupby.integration.test.ts  # integration: real ClickHouse via testcontainers
```

## Human verification

This is a user-observable analytics fix. The most direct re-verification is in the running app; a test-only path is included for a reviewer who would rather not boot it.

**In the app (the real user flow):**

1. Open the eval analytics dashboard at `/[project]/analytics/evaluations` (or build a custom graph at `/[project]/analytics/custom`).
2. Add/scope a chart to **group by `evaluations.evaluation_passed`** with `groupByKey` set to a **score-only evaluator** (one whose processed rows emit a numeric score but no pass/fail verdict, i.e. `Passed IS NULL`).
3. **Before this fix:** that evaluator's runs are absent — the grouped chart is empty / undercounted (the #2674 bug).
4. **After this fix:** the same runs appear under an **`'unknown'`** group, so the evaluator's data is visible on its own dashboard chart. Foreign evaluators' rows still do not leak in (the #2668 isolation guarantee).

This chart renders through `CustomGraph.tsx` → tRPC `analytics.getTimeseries` → `clickhouse-analytics.service` → `buildTimeseriesQuery` (the changed CASE), so the UI reads the exact query this PR edits.

**Without booting (test-only re-verification):**

1. Check out this branch and `cd langwatch/`.
2. **Reproduce the bug:** in `src/server/analytics/clickhouse/aggregation-builder.ts`, delete the one `WHEN … Passed IS NULL THEN 'unknown'` clause (line ~335), leaving the test untouched.
3. Run `pnpm test:integration run score-only-evaluator-passed-groupby` — the `'unknown'` group assertion **fails** (`expected 0 to be greater than 0`), reproducing #2674.
4. **Restore** the clause (`git checkout src/server/analytics/clickhouse/aggregation-builder.ts`) and re-run — both tests **pass**.
5. Optionally read the generated SQL the unit test asserts on (`aggregation-builder.test.ts`, the issue-#2674 case) to confirm the `EvaluatorId = {groupByKey:String}` guard is present, so isolation across evaluators still holds.

**User-observable surface = the eval analytics dashboard groupBy.** The changed query is the data source for the eval analytics charts. A user opens the eval analytics dashboard and groups a chart by `evaluations.evaluation_passed` scoped to a specific evaluator (`groupByKey`); for a score-only evaluator that grouped chart was empty before this fix and shows the evaluator's runs (as `'unknown'`) after it. The chain is wired, not asserted:

```
aggregation-builder.ts  buildTimeseriesQuery  (the changed evaluation_passed CASE)
  → clickhouse-analytics.service.ts
  → analytics.service.ts
  → tRPC  src/server/api/routers/analytics/timeseries.ts   (+ REST  src/app/api/analytics/[...route]/app.v1.ts)
  → src/components/analytics/CustomGraph.tsx   (getTimeseries.useQuery, groupBy)
  → two user-reachable entry points render that chart:
      • src/pages/[project]/analytics/custom/index.tsx   — the custom graph builder, where the user PICKS groupBy + groupByKey via form fields (CustomGraphFormData.groupBy / .groupByKey, lines 149-150); this is the exact "user groups by a score-only evaluator" path #2674 describes
      • src/pages/[project]/analytics/evaluations.tsx     — the evaluations overview, whose per-evaluator cards auto-emit groupBy "evaluations.evaluation_passed" + groupByKey check.id (lines 107-108 / 135-136)
```

`evaluation_passed` is a registered analytics metric (`src/server/analytics/registry.ts:1059`). So this is **not** a backend-only change with no UI surface — the prior "no-UI-surface" framing was wrong.

**Mechanism proof — falsifiability against live ClickHouse (before = FAIL, after = PASS):**

Without the fix (ephemerally removed the one score-only `WHEN … Passed IS NULL THEN 'unknown'` clause, test untouched):

```
 × … > includes score-only rows as an 'unknown' group instead of dropping them
   → expected 0 to be greater than 0

 ❯ src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts:198:38
    198|           expect(unknownRows.length).toBeGreaterThan(0);
 ✓ … > groups rows into passed and failed correctly        (1 failed | 1 passed)
```

→ score-only rows are dropped by `HAVING group_key IS NOT NULL` — exactly the #2674 bug. The clause is load-bearing, not vacuous.

With the fix (restored — worktree left clean via `git checkout`):

```
 ✓ … > includes score-only rows as an 'unknown' group instead of dropping them   36ms
 ✓ … > groups rows into passed and failed correctly                              29ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Duration  1.53s
```

→ the score-only evaluator now appears as an `'unknown'` group **and is the only group** present (foreign-evaluator rows still hit `ELSE NULL` → dropped, so the #2668 isolation guarantee holds); the pass/fail evaluator still groups into `passed`/`failed` with the correct avg scores (0.9 / 0.1).

This is the **same lane CI runs** — `test-integration` provides `CI_CLICKHOUSE_URL`, so the regression is CI-gated, not local-only. `pnpm typecheck` clean; affected unit suite 91 passed. Run from `langwatch/`: `pnpm test:integration run score-only-evaluator-passed-groupby`.

**Live-app visual proof — tested against the running app.** Captured against the **running app** (golden VM, this branch `0a2d8516`). I seeded a guardrail-type evaluator ("Score-Only Guardrail (#2674 proof)", `openai/moderation`, `isGuardrail: true`) with 25 processed evaluation results the way a user's SDK would — via the `/api/collector` ingestion endpoint: **16 score-only rows** (`Score` set, `Passed = NULL`), plus 4 passed and 5 failed. Then I opened the **Online Evaluations analytics dashboard** (`/[project]/analytics/evaluations`), whose per-evaluator donut runs exactly `groupBy: "evaluations.evaluation_passed"` + `groupByKey: <evaluatorId>` — the query this PR fixes.

Live-app proof: the eval analytics dashboard donut for a score-only evaluator shows a 64% "Evaluation passed unknown" slice (16 score-only rows, Passed IS NULL) alongside passed/failed — included instead of dropped (#2674)

The donut shows a **64% "Evaluation passed unknown"** slice — the 16 score-only rows (`Passed IS NULL`) that this fix now buckets as `'unknown'`, sitting alongside the green "passed" and red "failed" slices. **Before** the fix those rows were dropped by the foreign-evaluator-isolation `CASE` (`ELSE NULL`), so the evaluator's chart would show only the 9 pass/fail rows (36%) and silently undercount the evaluator; the integration test (`score-only-evaluator-passed-groupby.integration.test.ts`) pins that before=dropped / after=`'unknown'` behavior. Tested against the live app — the score-only evaluator now appears in evaluation analytics, exactly as #2674 requires.

## Anything surprising?

The naive fix (`ELSE 'unknown'`) would have broken the #2668 evaluator-isolation guarantee by pulling in rows from other evaluators. This fix keeps `ELSE NULL` for foreign rows and re-buckets **only** the target evaluator's score-only (`Passed IS NULL`) rows — matching the no-`groupByKey` branch's behavior. CodeRabbit raised the predicate/comment mismatch as a design question; the resolution (tighten the bucket to `Passed IS NULL` rather than loosen the comment) is implemented here and is the same direction the no-`groupByKey` branch already takes.

### DIFF
--- a/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -742,8 +742,10 @@ describe("aggregation-builder", () => {
         };
         const result = buildTimeseriesQuery(input);
 
-        // The new score-only WHEN: processed rows for THIS evaluator with Passed IS NULL → 'unknown'
-        expect(result.sql).toMatch(/Status = 'processed'\s+THEN 'unknown'/);
+        // The new score-only WHEN: target evaluator + processed + Passed IS NULL → 'unknown'
+        expect(result.sql).toMatch(
+          /EvaluatorId = \{groupByKey:String\} AND \w+\.Status = 'processed' AND \w+\.Passed IS NULL THEN 'unknown'/,
+        );
         // Foreign-evaluator isolation sentinel is preserved — other evaluators still hit ELSE NULL
         expect(result.sql).toContain("ELSE NULL");
       });
--- a/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -726,6 +726,27 @@ describe("aggregation-builder", () => {
         // so no outer if(... IS NULL, 'unknown', toString(...)) wrapper is added.
         expect(result.sql).not.toContain("IS NULL, 'unknown'");
       });
+
+      it("buckets score-only evaluator rows as 'unknown' under groupByKey instead of dropping them (issue #2674)", () => {
+        const input = {
+          ...baseInput,
+          timeScale: "full" as const,
+          groupBy: "evaluations.evaluation_passed" as const,
+          groupByKey: "my-evaluator",
+          series: [
+            {
+              metric: "metadata.trace_id" as FlattenAnalyticsMetricsEnum,
+              aggregation: "cardinality" as const,
+            },
+          ],
+        };
+        const result = buildTimeseriesQuery(input);
+
+        // The new score-only WHEN: processed rows for THIS evaluator with Passed IS NULL → 'unknown'
+        expect(result.sql).toMatch(/Status = 'processed'\s+THEN 'unknown'/);
+        // Foreign-evaluator isolation sentinel is preserved — other evaluators still hit ELSE NULL
+        expect(result.sql).toContain("ELSE NULL");
+      });
     });
 
     it("uses date-bucketed pipeline query for pipeline metrics with numeric timeScale", () => {
--- a/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -742,7 +742,6 @@ describe("aggregation-builder", () => {
         };
         const result = buildTimeseriesQuery(input);
 
-        // The new score-only WHEN: target evaluator + processed + Passed IS NULL → 'unknown'
         expect(result.sql).toMatch(
           /EvaluatorId = \{groupByKey:String\} AND \w+\.Status = 'processed' AND \w+\.Passed IS NULL THEN 'unknown'/,
         );
--- a/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
@@ -252,7 +252,6 @@ describe("score-only-evaluator-passed-groupby", () => {
           expect(passedRows.length).toBeGreaterThan(0);
           expect(failedRows.length).toBeGreaterThan(0);
 
-          // A pass/fail evaluator yields no 'unknown' group — isolation is explicit here, not just implicit
           expect(
             currentRows.filter((r) => r.group_key === "unknown"),
           ).toHaveLength(0);
--- a/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
@@ -252,6 +252,11 @@ describe("score-only-evaluator-passed-groupby", () => {
           expect(passedRows.length).toBeGreaterThan(0);
           expect(failedRows.length).toBeGreaterThan(0);
 
+          // A pass/fail evaluator yields no 'unknown' group — isolation is explicit here, not just implicit
+          expect(
+            currentRows.filter((r) => r.group_key === "unknown"),
+          ).toHaveLength(0);
+
           const metricKey = Object.keys(passedRows[0]!).find(
             (k) => k !== "date" && k !== "period" && k !== "group_key",
           );
--- a/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/score-only-evaluator-passed-groupby.integration.test.ts
@@ -1,0 +1,267 @@
+/**
+ * Regression integration tests for issue #2674.
+ *
+ * Bug: `buildTimeseriesQuery` with `groupBy: "evaluations.evaluation_passed"` and
+ * a `groupByKey` (a specific EvaluatorId) uses a CASE expression that returned
+ * `ELSE NULL` for any row that was not a clear pass (Passed=1) or fail (Passed=0).
+ *
+ * `buildGroupKeyHavingClause` then adds `HAVING group_key IS NOT NULL`, which
+ * intentionally drops rows from OTHER evaluators (foreign-evaluator isolation,
+ * regression #2668). However, it also dropped "score-only" rows — rows belonging
+ * to the TARGET evaluator that ran successfully (Status='processed') but have
+ * `Passed IS NULL` (a numeric score with no pass/fail threshold).
+ *
+ * The fix adds a new WHEN clause before `ELSE NULL`:
+ *   WHEN EvaluatorId = {groupByKey:String} AND Status = 'processed' THEN 'unknown'
+ *
+ * This buckets score-only rows for the target evaluator as 'unknown' (matching the
+ * no-groupByKey branch behaviour) while rows from foreign evaluators and
+ * non-processed rows still hit `ELSE NULL` and are dropped by the HAVING clause,
+ * preserving the #2668 isolation guarantee.
+ *
+ * @see https://github.com/langwatch/langwatch/issues/2674
+ */
+
+import type { ClickHouseClient } from "@clickhouse/client";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { wrapWithDefaultSettings } from "~/server/clickhouse/safeClickhouseClient";
+import {
+  cleanupTestData,
+  getTestClickHouseClient,
+} from "../../../event-sourcing/__tests__/integration/testContainers";
+import type { FlattenAnalyticsMetricsEnum } from "../../registry";
+import type { AggregationTypes } from "../../types";
+import { buildTimeseriesQuery } from "../aggregation-builder";
+import { resetParamCounter } from "../filter-translator";
+import { seedSpans } from "./test-utils/clickhouse-fixtures";
+
+const TENANT_ID = "test-score-only-2674";
+
+/** Base query input shared across all tests */
+const baseInput = {
+  projectId: TENANT_ID,
+  startDate: new Date("2020-01-01T00:00:00Z"),
+  endDate: new Date("2030-01-01T00:00:00Z"),
+  previousPeriodStartDate: new Date("2019-01-01T00:00:00Z"),
+  timeScale: "full" as const,
+};
+
+const TRACE_ID_0 = `${TENANT_ID}-trace-0`;
+const TRACE_ID_1 = `${TENANT_ID}-trace-1`;
+
+/**
+ * SCORE_ONLY_EVALUATOR_ID: produces numeric scores only — Passed is always NULL.
+ *   These rows should appear as 'unknown' in evaluation_passed groupBy (not dropped).
+ * PASSFAIL_EVALUATOR_ID: produces pass/fail results — Passed is 0 or 1.
+ *   These rows should appear as 'passed' or 'failed'.
+ */
+const SCORE_ONLY_EVALUATOR_ID = "score-only-2674-scoreonly";
+const PASSFAIL_EVALUATOR_ID = "score-only-2674-passfail";
+
+describe("score-only-evaluator-passed-groupby", () => {
+  let ch: ClickHouseClient;
+
+  beforeAll(async () => {
+    const rawClient = getTestClickHouseClient();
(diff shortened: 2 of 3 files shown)


drewdrewthis:
## Review verdict: READY

`/review` fan-out on PR #5007 (issue #2674) — 4 reviewer agents over the full diff (aggregation-builder.ts groupByKey CASE + integration test + unit test). **No blocking concerns. All review threads resolved at this SHA** (none were opened — every finding is non-blocking). Zero bugs, zero security issues, correct test placement.

**Completeness (issue #2674):** both ACs COVERED — AC1 (score-only rows included via the new WHEN) and AC2 (red→green integration test + unit test).

### Non-blocking notes
- **[security]** Clean — the new WHEN uses the parameterized `{groupByKey:String}` placeholder (no injection), tenant isolation is untouched, and test fixtures are synthetic (no PII/secrets).
- **[test]** Integration level is correct and necessary — it exercises the real ClickHouse query (the gating surface: the `HAVING group_key IS NOT NULL` drop), which a SQL-string unit test cannot catch. Verified red→green.
- **[hygiene] Decide — kept dot-notation** (`r.group_key`) over the sibling test's bracket-notation: biome `useLiteralKeys` flags bracket access on new/changed lines (reviewdog diff-filter would fail `lint`); the sibling `cross-evaluator-groupby.integration.test.ts` is grandfathered. Intentional divergence.
- **[hygiene] Decide** — `type Row = Record<string, unknown>` is declared per-`it` to mirror the established sibling pattern; a shared file-scope hoist across both files is a reasonable follow-up.
- **[test] New Issue (out of scope)** — no test for a score-only row with `Status != 'processed'` (correctly falls to `ELSE NULL` and is dropped) nor for the no-`groupByKey` `evaluation_passed` branch on `Passed IS NULL` rows. The `Status='processed'` guard makes current behavior correct; these are coverage follow-ups.
- **[test] New Issue (out of scope)** — `afterAll` cleanup uses `rawClient.exec` (bypasses `wrapWithDefaultSettings`), copied from the sibling test; worth a shared-helper audit.
- **[principles] Decide** — the three groupByKey WHENs repeat the `EvaluatorId = {groupByKey} AND Status='processed'` prefix; kept as flat rule-per-line for readability. The pre-existing redundant `Passed IS NOT NULL` guards on the pass/fail WHENs are an out-of-scope cleanup candidate.

### Reviewers run
- Ran (agent-backed): **principles**, **hygiene**, **security**, **test**.
- **Skipped for proportionality** (documented deviation): the 4 structure personas + design-soundness + drift-reviewer. Rationale — minimal 1-WHEN SQL-predicate diff with no integration/adapter/build-vs-reuse surface; the 4 core reviewers found zero defects. proof-reviewer/ac-reviewer not triggered (no `## Acceptance Criteria` section on the PR; no plan/AC set in the code diff). Re-run `/review` without the proportionality skip if fuller coverage is wanted.

drewdrewthis:
This solution seems a bit smelly. Is there a better way? Can we fix this upstream so it's not unknown? Is this a good solution?

drewdrewthis:
## CodeRabbit review — resolution summary

All five CodeRabbit findings were addressed in code (commits `ab50049c4`, `0a2d85162`) and their threads resolved. Recording the mapping here for the audit trail, then clearing the now-stale top-level nitpick comments.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Assert the **EvaluatorId guard** in the `unknown` branch, not just `Status` (`aggregation-builder.test.ts`) | Unit assertion now matches the full clause `EvaluatorId = {groupByKey:String} AND … Status = 'processed' AND … Passed IS NULL THEN 'unknown'` (`ab50049c4`). |
| 2 | Add an explicit **"no unknown"** assertion in the pass/fail integration scenario | Integration test now asserts `currentRows.filter(group_key === 'unknown')` has length 0 (`ab50049c4`). |
| 3 | **Predicate/comment mismatch** — comment documented `Passed IS NULL` but the branch bucketed all remaining processed rows | Tightened the branch to `… AND es.Passed IS NULL THEN 'unknown'`, so comment and predicate agree and #2668 isolation holds (`ab50049c4`). |
| 4 | Remove redundant restate-the-code comment (`aggregation-builder.test.ts` ~L745) | Dropped (`0a2d85162`). |
| 5 | Drop the duplicate explanatory comment (`score-only-…integration.test.ts` ~L255) | Dropped (`0a2d85162`). |

The design question (#3) resolved in the direction CodeRabbit recommended: tighten the bucket to `Passed IS NULL` rather than loosen the comment — matching the no-`groupByKey` branch.

drewdrewthis:
> This solution seems a bit smelly. Is there a better way? Can we fix this upstream so it's not unknown? Is this a good solution?

Fair challenge — here's the reasoning, and what changed since you flagged it.

**`'unknown'` is not new here — it's the existing convention.** The *unscoped* `evaluation_passed` branch (no `groupByKey`) already buckets verdict-less rows as `'unknown'` (`aggregation-builder.ts`, the `ELSE 'unknown'` arm). The bug in #2674 was that the *scoped* branch diverged: it returned `ELSE NULL`, and `HAVING group_key IS NOT NULL` then dropped the rows. So the fix makes the scoped branch **consistent** with the unscoped one rather than inventing a new bucket.

**Can we fix it upstream so it's not "unknown"?** Not cleanly, because the data is genuinely verdict-less by design. A score-only evaluator emits a numeric score and no pass/fail threshold, so `Passed IS NULL` is *correct* storage, not a defect to repair upstream. Synthesizing a `Passed` value at write-time would corrupt the pass/fail semantics for real pass/fail evaluators and ripple through the #2668 isolation logic and all historical rows — a large, wrong change to serve a display concern. When you view a thresholdless evaluator through a **pass/fail** lens, "no verdict" is the honest answer; the meaningful numeric view is the separate `evaluation_score` groupBy.

**Is it a good solution?** For this bug, yes — it's minimal and contained:
- After your nudge it now gates **narrowly on `Passed IS NULL`** (commit `ab50049c4`), not a broad `ELSE 'unknown'`. That was the smelly part: a loose bucket would have re-admitted *other* evaluators' rows and broken the #2668 isolation guarantee. The tight predicate catches only the target evaluator's score-only rows; foreign + unprocessed rows still hit `ELSE NULL` → dropped.
- The comment above the CASE and the predicate now agree (CodeRabbit had flagged that mismatch).
- Covered by a unit test (asserts the full `EvaluatorId` + `Passed IS NULL` clause) and a real-ClickHouse integration test (`'unknown'` is the *only* group for a score-only evaluator, absent for a pass/fail one).

**The one thing that's arguably "better" but out of scope:** the label `'unknown'` itself. "No verdict" or "score-only" would read better — but `'unknown'` is a shared convention (also used for missing trace-name/model), so renaming it is a cross-cutting UX change touching both branches and other group-bys, not part of this fix. Happy to spin that off if you want it.

Rationale is also captured in a decision record (`Passed IS NULL` bucket + #2668 isolation tradeoff). @rogeriochaves for visibility.

drewdrewthis:
> **The one thing that's arguably "better" but out of scope:** the label `'unknown'` itself. "No verdict" or "score-only" would read better — but `'unknown'` is a shared convention (also used for missing trace-name/model), so renaming it is a cross-cutting UX change touching both branches and other group-bys, not part of this fix. Happy to spin that off if you want it.

If it's out of scope, open an issue.

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @Aryansharma28

**Scariest prod risk:** A too-broad `WHEN` could re-admit foreign-evaluator rows into the `evaluation_passed` groupBy and break the #2668 isolation guarantee (cross-evaluator data leak on the dashboard).

- `[Must Check]` `aggregation-builder.ts` new CASE clause — confirm it is gated on BOTH `EvaluatorId = {groupByKey}` AND `Passed IS NULL` (not a bare `ELSE`), so only the target evaluator's score-only rows bucket as `'unknown'` and everything else still hits `ELSE NULL`.
- `[Probably Fine]` `HAVING group_key IS NOT NULL` interaction — verify the new `'unknown'` bucket survives the HAVING filter while foreign/unprocessed rows are still dropped.
- `[Probably Fine]` integration test `score-only-evaluator-passed-groupby.integration.test.ts` — check it asserts foreign-evaluator rows do NOT leak into the target chart, not just that score-only rows appear.

- Does the doc comment above the CASE now match the predicate exactly (NULL-only bucketing), so a future reader does not widen it into an isolation regression?

rogeriochaves (approved):

PR #5008, opened 2026-06-21
fix(scenarios): take maxRuntime-killed and orphaned QUEUED runs terminal (#3365)

### LINKED ISSUE #3365: Scenario runs orphan at QUEUED when worker maxRuntime restart kills child process

drewdrewthis:
## What & why

Fixes #3365. Scenario runs orphaned permanently at `QUEUED` when a worker hit `maxRuntimeMs` and restarted: it killed in-flight child processes without emitting any terminal event, so the projection stayed `QUEUED` and the suites page polled forever.

Root cause was two orphan paths in the execution pool's drain:
1. **Pending jobs** were dropped (`_pending.length = 0`) with no event.
2. **Running children** were SIGTERM'd; their close handler *did* emit a failure, but asynchronously — it raced the worker restart and usually lost.

### AC1 — graceful drain emits terminal failures before restart
- `execution-pool.ts`: the pool tracks in-flight job data (`_runningJobs`, set from `startJob` so the spawn window is covered) and exposes `inFlightJobs` (running + pending).
- `scenario.processor.ts`: new `drainInFlightRuns` emits a terminal failure (idempotent `finishRun`) for every in-flight run, then drains. `close()` calls it. `worker.ts`'s maxRuntime path already `await`s `close()` before the restart `reject`, so a killed run reaches ERROR instead of stuck QUEUED. A cancelled-before-execution run is preserved as CANCELLED (not re-emitted as ERROR) via `pool.wasCancelled(...)`. Also covers graceful shutdown.

### AC2 — startup reconciler for hard kills (OOM/SIGKILL)
- New `scenario-orphan-reconciler.ts`: pure gate `isOrphanedQueuedRun` (QUEUED + no progress past `ORPHAN_QUEUED_THRESHOLD_MS` = 30 min), `findQueuedRunCandidates` (cross-tenant CH scan, partition-pruned on `StartedAt`, latest-version via GROUP BY + `argMax`, **oldest-first `ORDER BY LastEventAtMs ASC` before the 10k cap** so true orphans are never starved by fresh rows), and `reconcileOrphanedQueuedRuns` (filters by the gate, emits failures, per-run isolation, returns `{failed, skipped, errored}`).
- Fired non-blocking from `startScenarioProcessor` so a slow/failing scan never wedges startup. Age-gated so freshly/healthy QUEUED runs are never falsely failed.

## Acceptance Criteria

- [x] **AC1** — When maxRuntime fires, a terminal failure is emitted for the in-flight run BEFORE the restart, so it goes terminal-failed (not stuck QUEUED).
- [x] **AC2** — On worker/processor startup, QUEUED runs with no live worker are reconciled to terminal-failed, gated so healthy QUEUED runs are NOT falsely reconciled.
- [x] **AC3** — Tests prove a maxRuntime-killed run goes terminal AND a healthy QUEUED run is not falsely failed.

**This is a user-observable fix, not backend-only.** The drain and the startup reconciler write a terminal status into the `simulation_runs` projection, and the **Scenarios → suite-run view renders that exact projection** — so to the user the run flips from a permanent "queued" spinner to a terminal "failed" badge:

| | Before this PR | After this PR |
|---|---|---|
| `simulation_runs` status | stuck `QUEUED` forever (no live worker) | terminal `ERROR` |
| Scenarios-view badge | blue **"queued"** + indefinite spinner | red **"failed"** + XCircle |
| Completion overlay | never shown (run looks perpetually in-flight) | shown (run is done) |

**Traced user-facing surface** (rendered, not backend-only):
- `components/simulations/scenario-run-status-config.ts` — `QUEUED → { blue, "queued", isComplete:false }` (L67–70) vs `ERROR → { red, "failed", isComplete:true }` (L37–40); icon map `QUEUED → Clock`, `ERROR → XCircle` (L84–90).
- `components/simulations/SimulationStatusOverlay.tsx:102` — `if (!isComplete) return null;`, so a QUEUED run never shows the "done" overlay while a reconciled ERROR run does.
- **Same projection both sides:** the view's `api.scenarios.getSuiteRunData` query (`server/api/routers/scenarios/scenario-events.router.ts:113`) reads ClickHouse `simulation_runs` — the very table `findQueuedRunCandidates` scans and `ensureFailureEventsEmitted` writes. The reconciler's `ERROR` write is what the next UI poll renders.
- **Terminal status emitted:** `ensureFailureEventsEmitted` dispatches `finishRun` with `ScenarioRunStatus.ERROR` (`scenario-failure-handler.ts:90`) on both the drain and the startup-reconciler paths.

### ✅ Live-app visual proof — orphaned run reconciled in the live UI

Orphaned QUEUED scenario run reconciled Stalled to Failed by the #5008 startup reconciler

Staged on the live golden instance running **this branch**. An orphaned `QUEUED` scenario run (no live worker, ~2 h with no progress — past `ORPHAN_QUEUED_THRESHOLD_MS`) was seeded into the `simulation_runs` projection. **Left:** the Scenarios suite-run view renders it stuck — a `⚠ Stalled` badge — because a non-`fix/3365` worker never takes it terminal. **Right:** after bouncing the scenario worker so `startScenarioProcessor` runs **this PR's** reconciler, the same run flips to a terminal `⊗ Failed` badge. ClickHouse confirms the latest status moved `QUEUED → ERROR` carrying the reconciler's own signature — `Error = "Reconciled: orphaned QUEUED run with no live worker (worker restart/crash)"` — and the worker log emitted the matching `simulation_updated → ERROR` event for `blob-5008-orphan-run`. Falsifiable: the flip happens **only** after this PR's reconciler boots; the stuck `Stalled` state is exactly what every prior worker left behind.

### Mechanism evidence (passing)

**AC1 — red→green (falsifiable regression test).** With the drain emit removed (simulating the pre-fix bare `pool.drain()`), the AC1 test fails because in-flight runs get **zero** terminal events:
```
× emits a terminal failure for every in-flight run
AssertionError: expected "vi.fn()" to be called 2 times, but got 0 times
Tests  2 failed | 2 passed (4)
```
With the fix in place, every in-flight run (running + buffered-pending) gets a terminal failure before drain → green. A separate case proves a **cancelled** in-flight run is drained as a cancellation event, not an ERROR failure.

**AC2/AC3 — gate.** `reconcileOrphanedQueuedRuns` over `[oldQueued, recentQueued, oldNonQueued]` emits a failure for **only** the long-abandoned QUEUED run; the recent/healthy QUEUED run and the non-QUEUED run are left untouched (`isOrphanedQueuedRun` boundary cases incl. the exact threshold are unit-tested).

**Full suite (green):**
```
Test Files  3 passed (3)
     Tests  24 passed (24)
```
`pnpm typecheck` clean; biome clean on changed files. CI runs the same suite in `langwatch-app-ci`.

A live-ClickHouse integration proof (throwaway harness, full source below) additionally drives the **real** `findQueuedRunCandidates` / `reconcileOrphanedQueuedRuns` against a real ClickHouse testcontainer and shows the orphan reconciled while four non-orphan fixtures are correctly excluded.

## Human verification

**User-facing outcome:** in **Scenarios → (suite) → run**, a run that orphaned at "queued" and used to spin forever now resolves to a red **"failed"** badge.

End-to-end in a running app:
1. Open the Scenarios view for a suite and start a run.
2. Orphan it — restart/kill the scenario worker inside the spawn window (maxRuntime path), or leave a QUEUED run with no worker for >30 min (hard-kill path).
3. **Before this PR:** the run stays on the blue "queued" spinner indefinitely; no completion overlay.
4. **After this PR:** the drain (restart) or startup reconciler (hard kill) emits terminal `ERROR`; the next poll shows the run as a red "failed" badge with the completion overlay.

Without booting the full app, the same surface can be confirmed statically + at test tier:

1. **Read the two terminal paths.** `scenario.processor.ts` → `drainInFlightRuns` emits a terminal failure for every `pool.inFlightJobs` entry (cancelled runs routed to `handleCancelledJobResult`, all others to `handleFailedJobResult`) before `pool.drain()`; `worker.ts`'s maxRuntime path `await`s `close()` before rejecting. `scenario-orphan-reconciler.ts` → `isOrphanedQueuedRun` only fires past the 30-min threshold, and the CH scan orders oldest-first before the 10k cap.
2. **Run the suite:** `cd langwatch && pnpm test:unit scenario-orphan-reconciler scenario-processor-drain execution-pool` → 24 passing across 3 files.
3. **Falsify it:** delete the emit loop in `drainInFlightRuns` (or flip the age gate in `isOrphanedQueuedRun`) and re-run — the AC1 / AC2 tests go red, proving they actually exercise the behavior.
4. **(Optional) live ClickHouse:** run the throwaway integration harness below against the `test_langwatch` testcontainer to see the orphan reconciled end-to-end.

## Notes for reviewer (non-blocking, deferred)
- The cross-tenant CH scan intentionally omits the per-tenant `TenantId` filter (documented in-code) — it mirrors existing ops scans and uses the shared client; each emitted failure is dispatched under the run's own tenant.
- **Deferred: ClickHouse integration coverage** for `findQueuedRunCandidates` (the `argMax`/`HAVING` dedup SQL) is tracked in **#5073** — acceptable at unit tier; the live-ClickHouse harness below exercises it manually in the meantime.
- **Deferred: single-flight/jitter gate** is tracked in **#5074** — on a coordinated multi-pod restart every pod runs the reconciler scan; safe (idempotent `finishRun`, partition-pruned, fire-and-forget) but N× work.
- `TABLE_NAME = "simulation_runs"` is duplicated per-file across the simulation CH modules (pre-existing pattern); extracting a shared constant is also folded into the cleanup tracked in #5074.

**Do not merge** — @drewdrewthis to merge.

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery for scenario runs that get stuck in a queued state.
  * Improved shutdown handling so in-flight runs are finalized before workers stop.

* **Bug Fixes**
  * Prevented queued runs from being left unfinished after restarts or abrupt worker exits.
  * Reconciliation now skips recently queued runs and only targets long-abandoned ones.

* **Tests**
  * Added coverage for graceful drain behavior, orphan recovery, and in-flight job tracking.

---

## 🔎 Live-ClickHouse integration proof — reconciler mechanism (not the full-app UI)

Drives the PR's **actual** reconciler functions against a **real ClickHouse** to prove the mechanism behind the user-visible flip above (it is **not** the full-app visual proof — see the now-embedded live-app visual proof in "How I can prove I was successful"). Also closes the gap flagged in the reviewer note ("No CH integration test for `findQueuedRunCandidates` yet", tracked in #5073). A **throwaway** integration test (NOT committed — full source below) runs against the `test_langwatch` testcontainer the integration suite uses, via the prisma-free `getClickHouseClient()` shared accessor that `startScenarioProcessor` passes at worker startup. It runs the real `findQueuedRunCandidates` (the `argMax` / `HAVING` dedup SQL — the head commit's subject) and `reconcileOrphanedQueuedRuns` with the real constants (`ORPHAN_QUEUED_THRESHOLD_MS` = 30 min, `LOOKBACK_MS` = 7 d).

**Fixtures inserted into `simulation_runs`** (one tenant, distinct runs):

| Fixture | Status | StartedAt | UpdatedAt | Expected |
|---|---|---|---|---|
| orphan | QUEUED | now−2h | now−45m | **reconciled** |
| fresh | QUEUED | now−5m | now−1m | excluded (age gate) |
| terminal | SUCCESS | now−2h | now−45m | excluded (status gate) |
| out-of-window | QUEUED | now−8d | now−8d | excluded (7d lookback bound) |
| dedup-to-terminal | QUEUED → SUCCESS (2 versions, same run) | now−3h | now−50m / now−40m | excluded (`argMax` reads latest = SUCCESS) |

The dedup case is the falsifier for the head commit's subject: a naive `max()` / any-version query would wrongly flag it; `argMax(Status, UpdatedAt)` reads the **latest** version (SUCCESS) and correctly skips it.

**Result — `pnpm test:integration`, live ClickHouse (`http://…/test_langwatch`):**

```text
[liveproof] ORPHAN_QUEUED_THRESHOLD_MS=1800000ms (30min), LOOKBACK_MS=604800000ms (7d)

# Phase 1 — findQueuedRunCandidates() scans real ClickHouse:
[liveproof] orphan candidates (this tenant): [ 'orphan-v5zdwrtJd-aa7fEmqFwkX' ]
#   -> ONLY the orphan; fresh / terminal / out-of-window / dedup-to-terminal all excluded

# Phase 2 — reconcileOrphanedQueuedRuns() with a real terminal-failure emitter:
[liveproof] reconcile result: { failed: 1, skipped: 0, errored: 0 }
[liveproof] emitted failures for: [ 'orphan-v5zdwrtJd-aa7fEmqFwkX' ]
[liveproof] orphan candidates after reconcile: []
#   -> orphan reconciled to terminal; the re-scan finds nothing (without the fix it polls forever)

 ✓ …liveproof… > findQueuedRunCandidates scans for orphans > returns ONLY the long-abandoned QUEUED run and excludes every non-orphan 17ms
 ✓ …liveproof… > reconcileOrphanedQueuedRuns … > emits a failure for the orphan and the run is then no longer detected as orphaned 38ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
```

**Falsifiability (without-fix stuck → with-fix reconciled).** Phase 1 shows the orphan is detectable as a permanently-stuck `QUEUED` run (the bug: no live worker, polled forever). Phase 2 runs the reconciler, emits a terminal failure, and the re-scan returns `[]` — the run reached a terminal state. The four excluded fixtures prove the gate is discriminating across all four axes (age, status, lookback window, and latest-version `argMax` dedup), so the reconciler cannot falsely fail healthy / fresh runs.

Run with:

```bash
pnpm test:integration scenario-orphan-reconciler.liveproof
```

<details>
<summary>Full throwaway harness source (not committed; embedded for reproducibility)</summary>

```ts
/**
 * @vitest-environment node
 *
 * LIVE-APP RUNTIME PROOF for PR #5008 (#3365) — NOT part of the PR.
 *
 * The PR ships only *.unit.test.ts for the orphan reconciler, and those mock
 * `findCandidates`. So the one piece that talks to a real database — the
 * cross-tenant ClickHouse scan `findQueuedRunCandidates` (the head commit's
 * subject) — has ZERO runtime coverage. This throwaway test closes that gap by
 * driving the REAL reconciler functions against a REAL ClickHouse
 * (test_langwatch testcontainer), with the SAME client the worker passes at
 * startup (`getSharedClickHouseClient()`).
 *
 * It proves, end to end:
 *   1) findQueuedRunCandidates picks up a long-abandoned QUEUED run and
 *      DISCRIMINATES — it excludes (a) a freshly-queued run, (b) a terminal
 *      (non-QUEUED) run, (c) a QUEUED run outside the 7d lookback window, and
 *      (d) a run whose LATEST version is terminal but an OLDER version was
 *      QUEUED (the argMax-vs-max dedup correctness the head commit is about).
 *   2) reconcileOrphanedQueuedRuns emits a terminal failure for the orphan and,
 *      once that terminal version lands, the orphan is no longer detected —
 *      i.e. the stuck run is reconciled (without the reconciler it would poll
 *      forever; with it, the run reaches a terminal state).
 */
import { nanoid } from "nanoid";
import { afterAll, beforeAll, describe, expect, it } from "vitest";
import type { ClickHouseClient } from "@clickhouse/client";
import {
  startTestContainers,
  stopTestContainers,
} from "../../event-sourcing/__tests__/integration/testContainers";
// The exact shared client accessor startScenarioProcessor passes to the
// reconciler. Imported from ./client (not ./clickhouseClient) to keep this
// proof free of the prisma import the routing wrapper carries — it is the same
// underlying getClickHouseClient() the worker uses.
import { _getSharedClickHouseClient as getSharedClickHouseClient } from "~/server/clickhouse/client";
import {
  findQueuedRunCandidates,
  LOOKBACK_MS,
  ORPHAN_QUEUED_THRESHOLD_MS,
  reconcileOrphanedQueuedRuns,
  type OrphanCandidate,
} from "../scenario-orphan-reconciler";

const MINUTE = 60 * 1000;
const HOUR = 60 * MINUTE;
const DAY = 24 * HOUR;

// Unique tenant so the cross-tenant scan can be filtered back to THIS run's
// fixtures — other suites may leave their own rows in the shared test DB.
const tenantId = `liveproof-3365-${nanoid()}`;
const now = Date.now();

// Distinct run ids for each fixture.
const orphanRunId = `orphan-${nanoid()}`;
const freshRunId = `fresh-${nanoid()}`;
const terminalRunId = `terminal-${nanoid()}`;
const outOfWindowRunId = `outofwindow-${nanoid()}`;
const dedupRunId = `dedup-to-terminal-${nanoid()}`;

function makeRow(o: {
  runId: string;
  status: string;
  startedAt: Date;
  updatedAt: Date;
}) {
  return {
    ProjectionId: `proj-${nanoid()}`,
    TenantId: tenantId,
    ScenarioRunId: o.runId,
    ScenarioId: `scenario-${nanoid()}`,
    BatchRunId: `batch-${nanoid()}`,
    ScenarioSetId: `set-${nanoid()}`,
    Version: "v1",
    Status: o.status,
    Name: null,
    Description: null,
    Metadata: null,
    "Messages.Id": [],
    "Messages.Role": [],
    "Messages.Content": [],
    "Messages.TraceId": [],
    "Messages.Rest": [],
    TraceIds: [],
    Verdict: null,
    Reasoning: null,
    MetCriteria: [],
    UnmetCriteria: [],
    Error: null,
    DurationMs: null,
    StartedAt: o.startedAt,
    CreatedAt: o.startedAt,
    UpdatedAt: o.updatedAt,
    FinishedAt: null,
    ArchivedAt: null,
    LastSnapshotOccurredAt: new Date(0),
  };
}

let ch: ClickHouseClient;

async function insertRows(rows: ReturnType<typeof makeRow>[]) {
  await ch.insert({
    table: "simulation_runs",
    values: rows,
    format: "JSONEachRow",
    // Synchronous insert so the very next SELECT sees the rows.
    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
  });
}

/** Run the REAL scan, then narrow to THIS test's tenant. */
async function scanMyOrphans(): Promise<OrphanCandidate[]> {
  const client = getSharedClickHouseClient();
  expect(client).not.toBeNull(); // same accessor startScenarioProcessor passes
  const all = await findQueuedRunCandidates({
    client: client!,
    lookbackMs: LOOKBACK_MS,
    now,
    orphanThresholdMs: ORPHAN_QUEUED_THRESHOLD_MS,
  });
  return all.filter((c) => c.projectId === tenantId);
}

beforeAll(async () => {
  const containers = await startTestContainers();
  ch = containers.clickHouseClient;

  console.log(
    `[liveproof] ORPHAN_QUEUED_THRESHOLD_MS=${ORPHAN_QUEUED_THRESHOLD_MS}ms (${
      ORPHAN_QUEUED_THRESHOLD_MS / MINUTE
    }min), LOOKBACK_MS=${LOOKBACK_MS}ms (${LOOKBACK_MS / DAY}d)`,
  );

  await insertRows([
    // (1) THE ORPHAN: QUEUED, in-window, no progress for >threshold -> reconcilable.
    makeRow({
      runId: orphanRunId,
      status: "QUEUED",
      startedAt: new Date(now - 2 * HOUR),
      updatedAt: new Date(now - 45 * MINUTE),
    }),
    // (2) FRESH: QUEUED but updated 1min ago -> a worker may still pick it up.
    makeRow({
      runId: freshRunId,
      status: "QUEUED",
      startedAt: new Date(now - 5 * MINUTE),
      updatedAt: new Date(now - 1 * MINUTE),
    }),
    // (3) TERMINAL: old, but SUCCESS -> already finished, not an orphan.
    makeRow({
      runId: terminalRunId,
      status: "SUCCESS",
      startedAt: new Date(now - 2 * HOUR),
      updatedAt: new Date(now - 45 * MINUTE),
    }),
    // (4) OUT OF WINDOW: QUEUED + old, but StartedAt before the 7d lookback.
    makeRow({
      runId: outOfWindowRunId,
      status: "QUEUED",
      startedAt: new Date(now - 8 * DAY),
      updatedAt: new Date(now - 8 * DAY),
    }),
    // (5) DEDUP-TO-TERMINAL: two versions of ONE run. Older = QUEUED, newer =
    //     SUCCESS. argMax(Status, UpdatedAt) must read the LATEST (SUCCESS), so
    //     this run is NOT an orphan. A naive max()/any-version query would
    //     wrongly flag it.
    makeRow({
      runId: dedupRunId,
      status: "QUEUED",
      startedAt: new Date(now - 3 * HOUR),
      updatedAt: new Date(now - 50 * MINUTE),
    }),
    makeRow({
      runId: dedupRunId,
      status: "SUCCESS",
      startedAt: new Date(now - 3 * HOUR),
      updatedAt: new Date(now - 40 * MINUTE),
    }),
  ]);
}, 60_000);

afterAll(async () => {
  if (ch) {
    await ch.exec({
      query: `ALTER TABLE simulation_runs DELETE WHERE TenantId = {tenantId:String}`,
      query_params: { tenantId },
    });
  }
  await stopTestContainers();
});

describe("orphan reconciler against live ClickHouse (PR #5008 / #3365)", () => {
  describe("given QUEUED, fresh, terminal, out-of-window, and dedup-to-terminal runs in the table", () => {
    describe("when findQueuedRunCandidates scans for orphans", () => {
      it("returns ONLY the long-abandoned QUEUED run and excludes every non-orphan", async () => {
        const mine = await scanMyOrphans();
        const ids = mine.map((c) => c.scenarioRunId);
        console.log("[liveproof] orphan candidates (this tenant):", ids);

        // The orphan is detected...
        expect(ids).toContain(orphanRunId);
        // ...and nothing else is.
        expect(ids).not.toContain(freshRunId); // age gate
        expect(ids).not.toContain(terminalRunId); // status gate
        expect(ids).not.toContain(outOfWindowRunId); // lookback bound
        expect(ids).not.toContain(dedupRunId); // argMax latest-version dedup
        expect(mine).toHaveLength(1);

        // The returned candidate carries the run's identity for failure emission.
        const orphan = mine[0]!;
        expect(orphan.projectId).toBe(tenantId);
        expect(orphan.scenarioRunId).toBe(orphanRunId);
        expect(orphan.status).toBe("QUEUED");
        expect(orphan.lastEventAtMs).toBeLessThanOrEqual(
          now - ORPHAN_QUEUED_THRESHOLD_MS,
        );
      });
    });

    describe("when reconcileOrphanedQueuedRuns runs with a real terminal-failure emitter", () => {
      it("emits a failure for the orphan and the run is then no longer detected as orphaned", async () => {
        // Sanity: orphan is still stuck QUEUED before reconciliation (this is
        // the bug condition — without the reconciler it stays here forever).
        const before = await scanMyOrphans();
        expect(before.map((c) => c.scenarioRunId)).toEqual([orphanRunId]);

        // Real emitter: write a terminal (FAILED) version of the run, which is
        // what ensureFailureEventsEmitted ultimately lands in the projection.
        const emitted: string[] = [];
        const emitFailure = async (candidate: OrphanCandidate) => {
          emitted.push(candidate.scenarioRunId);
          await insertRows([
            makeRow({
              runId: candidate.scenarioRunId,
              status: "FAILED",
              startedAt: new Date(now - 2 * HOUR),
              updatedAt: new Date(now), // newest version -> wins argMax
            }),
          ]);
        };

        const result = await reconcileOrphanedQueuedRuns({
          // Scope the cross-tenant scan to this test's tenant so we never touch
          // other suites' data; the orchestrator itself is unmodified.
          findCandidates: scanMyOrphans,
          emitFailure,
          now,
          thresholdMs: ORPHAN_QUEUED_THRESHOLD_MS,
        });
        console.log("[liveproof] reconcile result:", result);
        console.log("[liveproof] emitted failures for:", emitted);

        expect(result).toEqual({ failed: 1, skipped: 0, errored: 0 });
        expect(emitted).toEqual([orphanRunId]);

        // The orphan is now reconciled: its latest version is terminal, so the
        // scan no longer returns it.
        const after = await scanMyOrphans();
        console.log(
          "[liveproof] orphan candidates after reconcile:",
          after.map((c) => c.scenarioRunId),
        );
        expect(after).toHaveLength(0);
      });
    });
  });
});
```
</details>

### DIFF
--- a/langwatch/specs/scenarios/queued-run-orphan-recovery.feature
+++ b/langwatch/specs/scenarios/queued-run-orphan-recovery.feature
@@ -1,0 +1,52 @@
+Feature: Queued scenario run orphan recovery
+
+  When a scenario worker restarts (it recycles itself after a maximum runtime,
+  or it crashes), any scenario runs it was about to execute or had already
+  started can be left behind. Before this recovery existed, those runs stayed
+  QUEUED forever: no terminal event was ever written, so the suites page kept
+  polling a run that no living worker would ever finish.
+
+  Recovery has two layers. The first is graceful: when a worker shuts down it
+  marks every run it still owns as failed before it goes away, so the common
+  case (a planned max-runtime restart) never orphans anything. The second is a
+  safety net for hard kills (OOM, SIGKILL) where the worker has no chance to
+  clean up: on startup a worker looks for runs that have sat QUEUED with no
+  progress for too long and marks them failed so the user sees a terminal
+  result instead of a spinner.
+
+  A freshly queued run is never touched by recovery — only runs that have been
+  abandoned long enough to prove no worker is coming for them.
+
+  Background:
+    Given scenario runs are executed by worker processes
+    And a scenario run is marked QUEUED when it is accepted for execution
+    And a worker recycles itself after it reaches its maximum runtime
+
+  # ---------------------------------------------------------------------------
+  # Layer 1: graceful drain on worker restart
+  # ---------------------------------------------------------------------------
+
+  @unit
+  Scenario: In-flight runs are failed when the worker restarts
+    Given a worker is executing one scenario run and has another buffered
+    When the worker reaches its maximum runtime and begins shutting down
+    Then both runs are marked failed before the worker restarts
+    And neither run is left QUEUED
+
+  # ---------------------------------------------------------------------------
+  # Layer 2: startup reconciler for hard kills
+  # ---------------------------------------------------------------------------
+
+  @unit
+  Scenario: A long-abandoned queued run is reconciled to failed on startup
+    Given a scenario run has been QUEUED with no progress for longer than the orphan threshold
+    And no worker is currently executing it
+    When a worker starts up and reconciles orphaned queued runs
+    Then the run is marked failed
+
+  @unit
+  Scenario: A freshly queued run is not failed by the reconciler
+    Given a scenario run was queued moments ago
+    When a worker starts up and reconciles orphaned queued runs
+    Then the run is left QUEUED
+    And it is not marked failed
--- a/langwatch/specs/scenarios/queued-run-orphan-recovery.feature
+++ b/langwatch/specs/scenarios/queued-run-orphan-recovery.feature
@@ -33,6 +33,13 @@ Feature: Queued scenario run orphan recovery
     Then both runs are marked failed before the worker restarts
     And neither run is left QUEUED
 
+  @unit
+  Scenario: A cancelled in-flight run is preserved as cancelled, not failed
+    Given a worker is executing a scenario run that was cancelled before it finished
+    When the worker begins shutting down
+    Then the run is marked cancelled
+    And it is not marked failed
+
   # ---------------------------------------------------------------------------
   # Layer 2: startup reconciler for hard kills
   # ---------------------------------------------------------------------------
--- a/langwatch/src/server/scenarios/__tests__/scenario-orphan-reconciler.unit.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/scenario-orphan-reconciler.unit.test.ts
@@ -1,0 +1,171 @@
+/**
+ * Unit tests for the queued-run orphan reconciler.
+ *
+ * Covers the pure orphan gate (isOrphanedQueuedRun) and the orchestrator
+ * (reconcileOrphanedQueuedRuns) with injected fakes — no ClickHouse access.
+ *
+ * @see specs/scenarios/queued-run-orphan-recovery.feature
+ */
+
+import { describe, expect, it, vi } from "vitest";
+import {
+  isOrphanedQueuedRun,
+  reconcileOrphanedQueuedRuns,
+  type OrphanCandidate,
+} from "../scenario-orphan-reconciler";
+
+const THRESHOLD_MS = 30 * 60 * 1000; // 30 min
+const NOW = 1_700_000_000_000;
+
+function makeCandidate(overrides: Partial<OrphanCandidate>): OrphanCandidate {
+  return {
+    projectId: "proj-1",
+    scenarioRunId: "run-1",
+    scenarioId: "scen-1",
+    batchRunId: "batch-1",
+    setId: "set-1",
+    lastEventAtMs: NOW - THRESHOLD_MS - 1,
+    status: "QUEUED",
+    ...overrides,
+  };
+}
+
+describe("isOrphanedQueuedRun", () => {
+  describe("given a QUEUED run", () => {
+    describe("when the last event is older than the threshold", () => {
+      it("flags it as orphaned", () => {
+        expect(
+          isOrphanedQueuedRun({
+            status: "QUEUED",
+            lastEventAtMs: NOW - THRESHOLD_MS - 1,
+            now: NOW,
+            thresholdMs: THRESHOLD_MS,
+          }),
+        ).toBe(true);
+      });
+    });
+
+    describe("when the last event is exactly at the threshold", () => {
+      it("flags it as orphaned", () => {
+        expect(
+          isOrphanedQueuedRun({
+            status: "QUEUED",
+            lastEventAtMs: NOW - THRESHOLD_MS,
+            now: NOW,
+            thresholdMs: THRESHOLD_MS,
+          }),
+        ).toBe(true);
+      });
+    });
+
+    describe("when the run was queued recently", () => {
+      it("does not flag it as orphaned", () => {
+        expect(
+          isOrphanedQueuedRun({
+            status: "QUEUED",
+            lastEventAtMs: NOW - 1000,
+            now: NOW,
+            thresholdMs: THRESHOLD_MS,
+          }),
+        ).toBe(false);
+      });
+    });
+  });
+
+  describe("given a non-QUEUED run", () => {
+    describe("when the last event is older than the threshold", () => {
+      it("does not flag it as orphaned", () => {
+        expect(
+          isOrphanedQueuedRun({
+            status: "IN_PROGRESS",
+            lastEventAtMs: NOW - THRESHOLD_MS - 1,
+            now: NOW,
+            thresholdMs: THRESHOLD_MS,
+          }),
+        ).toBe(false);
+      });
+    });
+  });
+});
+
+describe("reconcileOrphanedQueuedRuns", () => {
+  describe("given candidates with mixed status and age", () => {
+    describe("when reconciling", () => {
+      it("emits a failure only for the long-abandoned queued run", async () => {
+        const oldQueued = makeCandidate({
+          scenarioRunId: "old-queued",
+          status: "QUEUED",
+          lastEventAtMs: NOW - THRESHOLD_MS - 1,
+        });
+        const recentQueued = makeCandidate({
+          scenarioRunId: "recent-queued",
+          status: "QUEUED",
+          lastEventAtMs: NOW - 1000,
+        });
+        const oldNonQueued = makeCandidate({
+          scenarioRunId: "old-in-progress",
+          status: "IN_PROGRESS",
+          lastEventAtMs: NOW - THRESHOLD_MS - 1,
+        });
+        const emitFailure = vi.fn().mockResolvedValue(undefined);
+
+        const result = await reconcileOrphanedQueuedRuns({
+          findCandidates: vi
+            .fn()
+            .mockResolvedValue([oldQueued, recentQueued, oldNonQueued]),
+          emitFailure,
+          now: NOW,
+          thresholdMs: THRESHOLD_MS,
+        });
+
+        expect(emitFailure).toHaveBeenCalledTimes(1);
+        expect(emitFailure).toHaveBeenCalledWith(oldQueued);
+        expect(result).toEqual({ failed: 1, skipped: 2 });
+      });
+    });
+  });
+
+  describe("given an emitFailure that rejects for one candidate", () => {
+    describe("when reconciling multiple orphans", () => {
+      it("still processes the remaining orphans and counts the failures", async () => {
+        const orphanA = makeCandidate({ scenarioRunId: "orphan-a" });
+        const orphanB = makeCandidate({ scenarioRunId: "orphan-b" });
+        const emitFailure = vi
+          .fn()
+          .mockRejectedValueOnce(new Error("emit blew up"))
+          .mockResolvedValueOnce(undefined);
+
+        const result = await reconcileOrphanedQueuedRuns({
+          findCandidates: vi.fn().mockResolvedValue([orphanA, orphanB]),
+          emitFailure,
+          now: NOW,
+          thresholdMs: THRESHOLD_MS,
+        });
+
+        expect(emitFailure).toHaveBeenCalledTimes(2);
+        expect(emitFailure).toHaveBeenCalledWith(orphanA);
+        expect(emitFailure).toHaveBeenCalledWith(orphanB);
+        // One emission succeeded, the rejecting one is counted as skipped.
+        expect(result).toEqual({ failed: 1, skipped: 1 });
+      });
+    });
+  });
+
+  describe("given no candidates", () => {
+    describe("when reconciling", () => {
+      it("emits nothing and returns zero counts", async () => {
+        const emitFailure = vi.fn().mockResolvedValue(undefined);
+
+        const result = await reconcileOrphanedQueuedRuns({
+          findCandidates: vi.fn().mockResolvedValue([]),
+          emitFailure,
+          now: NOW,
+          thresholdMs: THRESHOLD_MS,
+        });
+
+        expect(emitFailure).not.toHaveBeenCalled();
+        expect(result).toEqual({ failed: 0, skipped: 0 });
+      });
+    });
+  });
+});
--- a/langwatch/src/server/scenarios/__tests__/scenario-orphan-reconciler.unit.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/scenario-orphan-reconciler.unit.test.ts
@@ -10,11 +10,12 @@
 import { describe, expect, it, vi } from "vitest";
 import {
   isOrphanedQueuedRun,
-  reconcileOrphanedQueuedRuns,
+  ORPHAN_QUEUED_THRESHOLD_MS,
   type OrphanCandidate,
+  reconcileOrphanedQueuedRuns,
 } from "../scenario-orphan-reconciler";
 
-const THRESHOLD_MS = 30 * 60 * 1000; // 30 min
+const THRESHOLD_MS = ORPHAN_QUEUED_THRESHOLD_MS;
 const NOW = 1_700_000_000_000;
 
 function makeCandidate(overrides: Partial<OrphanCandidate>): OrphanCandidate {
@@ -46,7 +47,7 @@ describe("isOrphanedQueuedRun", () => {
     });
 
     describe("when the last event is exactly at the threshold", () => {
-      it("flags it as orphaned", () => {
+      it("flags it as orphaned at the exact threshold boundary", () => {
         expect(
           isOrphanedQueuedRun({
             status: "QUEUED",
@@ -120,7 +121,7 @@ describe("reconcileOrphanedQueuedRuns", () => {
 
         expect(emitFailure).toHaveBeenCalledTimes(1);
         expect(emitFailure).toHaveBeenCalledWith(oldQueued);
-        expect(result).toEqual({ failed: 1, skipped: 2 });
+        expect(result).toEqual({ failed: 1, skipped: 2, errored: 0 });
       });
     });
   });
@@ -145,8 +146,8 @@ describe("reconcileOrphanedQueuedRuns", () => {
         expect(emitFailure).toHaveBeenCalledTimes(2);
         expect(emitFailure).toHaveBeenCalledWith(orphanA);
         expect(emitFailure).toHaveBeenCalledWith(orphanB);
-        // One emission succeeded, the rejecting one is counted as skipped.
-        expect(result).toEqual({ failed: 1, skipped: 1 });
+        // One emission succeeded, the rejecting one is counted as errored.
+        expect(result).toEqual({ failed: 1, skipped: 0, errored: 1 });
       });
     });
   });
@@ -164,7 +165,7 @@ describe("reconcileOrphanedQueuedRuns", () => {
         });
 
         expect(emitFailure).not.toHaveBeenCalled();
-        expect(result).toEqual({ failed: 0, skipped: 0 });
+        expect(result).toEqual({ failed: 0, skipped: 0, errored: 0 });
       });
     });
   });
--- a/langwatch/src/server/scenarios/__tests__/scenario-processor-drain.unit.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/scenario-processor-drain.unit.test.ts
@@ -10,10 +10,10 @@
  * @see specs/scenarios/queued-run-orphan-recovery.feature "In-flight runs are failed when the worker restarts"
  */
 
-import { beforeEach, describe, expect, it, vi } from "vitest";
 import type { ChildProcess } from "child_process";
-import { ScenarioExecutionPool } from "../execution/execution-pool";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import type { ExecutionJobData } from "../execution/execution-pool";
+import { ScenarioExecutionPool } from "../execution/execution-pool";
 import type { ProcessorDependencies } from "../scenario.processor";
 import { drainInFlightRuns } from "../scenario.processor";
 
@@ -43,10 +43,9 @@ describe("drainInFlightRuns", () => {
     // `_running`, which only fills via registerChild). The child is never
     // deregistered, so jobs stay in-flight until drain.
     pool.setSpawnFunction(async (jobData) => {
-      pool.registerChild(
-        jobData.scenarioRunId,
-        { kill: () => true } as unknown as ChildProcess,
-      );
+      pool.registerChild(jobData.scenarioRunId, {
+        kill: () => true,
+      } as unknown as ChildProcess);
     });
 
     mockGetById = vi
@@ -74,6 +73,7 @@ describe("drainInFlightRuns", () => {
         await drainInFlightRuns(pool, deps);
 
         expect(mockEnsureFailureEventsEmitted).toHaveBeenCalledTimes(2);
+        expect(mockGetById).toHaveBeenCalledTimes(2);
         const emittedRunIds = mockEnsureFailureEventsEmitted.mock.calls.map(
           (call) => (call[0] as { scenarioRunId: string }).scenarioRunId,
         );
--- a/langwatch/src/server/scenarios/__tests__/scenario-processor-drain.unit.test.ts
+++ b/langwatch/src/server/scenarios/__tests__/scenario-processor-drain.unit.test.ts
@@ -1,0 +1,125 @@
+/**
+ * Unit tests for the in-flight failure emission that runs when the scenario
+ * processor shuts down (graceful drain / worker max-runtime restart).
+ *
+ * Builds a real ScenarioExecutionPool with a no-op spawn function so submitted
+ * jobs stay in-flight, then exercises drainInFlightRuns (the logic close()
+ * runs) with a mock failure emitter. This is the path that stops in-flight
+ * runs from orphaning at QUEUED when the worker restarts.
+ *
+ * @see specs/scenarios/queued-run-orphan-recovery.feature "In-flight runs are failed when the worker restarts"
+ */
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { ChildProcess } from "child_process";
+import { ScenarioExecutionPool } from "../execution/execution-pool";
+import type { ExecutionJobData } from "../execution/execution-pool";
+import type { ProcessorDependencies } from "../scenario.processor";
+import { drainInFlightRuns } from "../scenario.processor";
+
+function makeJob(id: string): ExecutionJobData {
+  return {
+    projectId: "proj-1",
+    scenarioId: `scen-${id}`,
+    scenarioRunId: id,
+    batchRunId: "batch-1",
+    setId: "set-1",
+    target: { type: "http", referenceId: "agent-1" },
+  };
+}
+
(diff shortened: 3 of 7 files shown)


drewdrewthis:
## Review verdict: READY

Reviewed at `bda4d8ee1` via parallel reviewers — **principles, hygiene, security, test-reviewer** (an earlier principles+hygiene pass on the first commit was also addressed). **Security: clean** (cross-tenant scan reads metadata only and emits each failure under the run's own tenant; the one interpolation is a compile-time enum, `fromMs` is parameterized). No code **Must-fix**.

All actionable findings were fixed in `bda4d8ee1`:
- `[principles]` reconcile result split into `{failed, skipped, errored}` so non-orphans (skipped) are distinguishable from emission errors
- `[principles][hygiene]` single captured clock for the startup reconciler (lookback + age gate)
- `[principles]` dropped a stale JSDoc parenthetical
- `[test]` import `ORPHAN_QUEUED_THRESHOLD_MS` instead of a magic 30-min literal; assert the scenario-lookup contract in the drain test; dedupe a duplicate `it()` label
- `[lint]` biome import-sort + format on the touched files (this was the failing `lint`/`langwatch-app-complete` job — now clean on added lines)

**Re-stamped at `0ac5a4f9e`** — the delta since `bda4d8ee1` implements CodeRabbit's two Major findings + its nitpick (independently reviewed): `drainInFlightRuns` now emits CANCELLED (not ERROR) for already-cancelled runs; `findQueuedRunCandidates` applies the orphan-age cutoff in SQL and orders oldest-first before the 10k cap; + a cancelled-run drain regression test. Re-validated: `pnpm typecheck` clean, 25/25 unit tests pass, biome clean. (Delta not re-run through the full reviewer fan-out — small and CodeRabbit-vetted.)

### Non-blocking / follow-ups (out of scope)
- `[test]` No ClickHouse integration test for `findQueuedRunCandidates` (the `argMax` + `HAVING Status='QUEUED'` dedup SQL). Acceptable at unit tier; a `scenario-orphan-reconciler.integration.test.ts` that inserts a QUEUED run + a superseding IN_PROGRESS version and asserts only the latest-QUEUED is returned would close it.
- `[hygiene]` `TABLE_NAME = "simulation_runs"` is duplicated per-file across the simulation CH modules (pre-existing pattern); extract a shared exported constant.
- `[principles][design-soundness]` On a coordinated multi-pod restart every pod runs the reconciler scan — safe (idempotent `finishRun`, partition-pruned, fire-and-forget) but N× work; a single-flight / jitter gate is a possible optimization.
- `[principles]` `findQueuedRunCandidates` `LIMIT 10000` has no `ORDER BY` (acceptable — the sweep re-runs every worker startup).

No unresolved review-opened threads.

drewdrewthis:
> No CH integration test for findQueuedRunCandidates yet (the argMax/HAVING dedup SQL) — acceptable at unit tier; a follow-up scenario-orphan-reconciler.integration.test.ts would close it.

Why wasn't this done in this PR? It should have been. This should not have been deferred without immediately opening and finishing the follow up PR.

drewdrewthis:
**Resolution of the 3 CodeRabbit inline review threads** (all verified addressed in commit `0ac5a4f`; resolved threads removed to clear the readiness gate). Recording the verification here so the trail survives:

1. **Add a cancelled-run drain regression case** (`scenario-processor-drain.unit.test.ts`) — ✅ Done. `describe("given a cancelled in-flight run")` → `it("emits a cancellation event, not an error failure")` asserts the emitted payload carries `cancelled: true` (lines 117–135).

2. **Prioritize oldest candidates before the 10k cap** (`scenario-orphan-reconciler.ts`) — ✅ Done. The CH scan now filters `LastEventAtMs <= orphanCutoffMs` and adds `ORDER BY LastEventAtMs ASC` before `LIMIT 10000` (lines 137–139), so fresh rows can no longer starve true orphans of the cap.

3. **Preserve CANCELLED status during shutdown drain** (`scenario.processor.ts`) — ✅ Done. `drainInFlightRuns` branches on `pool.wasCancelled(jobData.scenarioRunId)` → `handleCancelledJobResult` for cancelled runs, else `handleFailedJobResult` (lines 166–173), so a cancelled-before-execution run is no longer re-emitted as ERROR.

All three are exercised by the unit suite (24 passing across 3 files).

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** The startup reconciler's orphan gate is age-only (`isOrphanedQueuedRun`: QUEUED + last event older than `ORPHAN_QUEUED_THRESHOLD_MS` = 2x child timeout). A run that is legitimately QUEUED because the queue is backed up (many pending jobs, a busy or just-restarted worker) can cross that threshold while a worker still intends to run it. On the next worker boot the reconciler would emit a terminal failure for a run that then actually starts and completes, producing conflicting terminal events for a live run. Idempotent `ensureFailureEventsEmitted` guards double-failure but not the fail-then-succeed race.

- `[Must Check]` `scenario-orphan-reconciler.ts` `isOrphanedQueuedRun` / threshold — confirm nothing can be reconciled-to-failed while a live worker still holds it as pending; the gate has no "is any worker processing this" signal, only age.
- `[Must Check]` `execution-pool.ts` `_runningJobs` / `inFlightJobs` — verify `_runningJobs.delete(scenarioRunId)` fires on every terminal path (success, error, cancel, spawn-not-set) so a finished run cannot be re-failed by a later `drainInFlightRuns`.
- `[Ask Author]` `findQueuedRunCandidates` — global oldest-first `ORDER BY LastEventAtMs ASC LIMIT 10000` across all tenants; one tenant with >10k old orphans could fill the cap and delay others until a later restart.

- What stops the reconciler from failing a run that is merely backlogged in the pending buffer (QUEUED past threshold) but that a live worker will still execute?
- On the max-runtime restart path, `close()` fails all in-flight pending jobs before restart; those pending runs were never started — is failing them (vs re-queueing) the intended user-visible outcome?

drewdrewthis:
## Review verdict: READY

Re-reviewed at post-rebase HEAD: diff matches the PR's queued-run-terminal intent; worker.ts deletion from main verified non-conflicting; CI fully green (74 pass).

langwatch/specs/scenarios/queued-run-orphan-recovery.feature
```diff
@@ -0,0 +1,59 @@
...
+    When the worker reaches its maximum runtime and begins shutting down
+    Then both runs are marked failed before the worker restarts
+    And neither run is left QUEUED
+
+  @unit
+  Scenario: A cancelled in-flight run is preserved as cancelled, not failed
+    Given a worker is executing a scenario run that was cancelled before it finished
+    When the worker begins shutting down
+    Then the run is marked cancelled
+    And it is not marked failed
```
drewdrewthis, line 41: Valid finding — added the scenario. drainInFlightRuns (scenario.processor.ts:166) branches on pool.wasCancelled() → handleCancelledJobResult for cancelled runs, confirmed by the unit test at scenario-processor-drain.unit.test.ts:117. Added a distinct Layer 1 scenario in commit dfa645f4aef973a2ed5f2ddb573b0ab102f84ba3.

drewdrewthis:
## Review verdict: READY

Re-verified at `dfa645f4aef973a2ed5f2ddb573b0ab102f84ba3` after rebase onto main.

**Rebase**: clean — only import reordering and line-length formatting in existing `scenario.processor.ts` code; no logic drift.

**New scenario (CANCELLED-preserved)**: `queued-run-orphan-recovery.feature` now has "A cancelled in-flight run is preserved as cancelled, not failed." Implementation match confirmed:

- `drainInFlightRuns` checks `pool.wasCancelled(scenarioRunId)` and routes to `handleCancelledJobResult` (not `handleFailedJobResult`) for cancelled runs.
- `scenario-processor-drain.unit.test.ts` (line 366): submits a job, marks it cancelled, drains, and asserts `ensureFailureEventsEmitted` is called with `cancelled: true` — not an error path.

All four feature scenarios have backing implementation and passing unit tests. CI green.

0xdeafcafe (approved):

PR #5010, opened 2026-06-21
fix(scenarios): mapping-gate toast links back to the agent editor (#3411)

### LINKED ISSUE #3411: Scenario mappings warning toast should link back to agent editor
## Summary

When a user clicks **Save & Run** on a workflow-agent target in the scenario form and `hasScenarioInputMapping(mappings)` returns false, we show a warning toast:

- title: `Configure scenario mappings`
- description: `Set up how this workflow agent maps scenario inputs and outputs before running.`

Source: `langwatch/src/components/scenarios/ScenarioFormDrawer.tsx:254-262` (main).

The code already calls `openDrawer("agentWorkflowEditor", { urlParams: { agentId: target.id } })` right after the toast (line 260), but the toast description itself is plain text with **no link**. If the drawer-open fails, races, is dismissed, or the user closes it without understanding, there's no clickable path back to the agent editor — the user is stuck reading a message about an editor they can't reach.

## Acceptance Criteria

- [ ] Toast description includes a link/CTA (e.g. "Open agent editor →") that opens the workflow agent editor drawer for `target.id`.
- [ ] Link works independently of the existing `openDrawer` auto-open — it's a fallback affordance, still functional if the auto-open was dismissed or failed.
- [ ] When possible, the toast surfaces **which specific mapping is missing** (input field name) so the user knows what to fix before opening the editor.
- [ ] No regression to the existing auto-open behavior at `ScenarioFormDrawer.tsx:260`.
- [ ] Existing mapping-gate test (`langwatch/src/components/scenarios/__tests__/ScenarioFormDrawer.mapping-gate.integration.test.tsx`) still passes; new coverage added for the link affordance.

## Related files

- `langwatch/src/components/scenarios/ScenarioFormDrawer.tsx` (handleSaveAndRun, lines 224-269)
- `langwatch/src/components/scenarios/__tests__/ScenarioFormDrawer.mapping-gate.integration.test.tsx`
- `hasScenarioInputMapping` helper (grep `langwatch/src/` for definition)

---
*Classification: Bug (UX gap)*

drewdrewthis:
## Why

Closes #3411

When a user clicks **Save & Run** on a workflow-agent target whose scenario input mapping is missing (`hasScenarioInputMapping(mappings) === false`), the scenario form drawer shows a warning toast and auto-opens the agent workflow editor. The toast itself was **plain text with no link** — so if the auto-open raced, was dismissed, or failed, the user was stuck reading about an editor they had no clickable path back to.

## What changed

- **Clickable CTA (AC1).** The warning toast now carries an **"Open agent editor"** action that opens the agent workflow editor for the target. It is wired independently of the existing auto-open, so it remains functional even if the auto-open was dismissed or failed — a true fallback affordance.
- **Names the missing input (AC2).** The gate fires only when *neither* the scenario `input` field nor `messages` is mapped, so the description now names exactly those fields ("Map at least one scenario input — "input" or "messages" — …"), mirroring the editor's own validation copy. The previous "scenario inputs and outputs" wording was both vaguer and slightly inaccurate — the gate only checks the input half (outputs are validated server-side at run time).
- **No regression (AC3).** The existing `openDrawer("agentWorkflowEditor", …)` auto-open is preserved unchanged; the CTA is purely additive.
- **Spec.** Added two `@integration` scenarios (links-back; names-missing) to the canonical `specs/features/scenarios/workflow-agent-mapping-layer.feature` and bound them to the new tests — and corrected the test's previously-dangling `@see` to point there (no duplicate feature file).

## How it works

The toast object now includes an `action: { label, onClick }` (the same pattern as `MissingModelToast.tsx`; the `Toaster` already renders `Toast.ActionTrigger` for `toast.action`). The `openAgentEditor` closure is shared between the action's `onClick` and the trailing auto-open call, so both routes open the same drawer for `target.id`.

## Test plan

`src/components/scenarios/__tests__/ScenarioFormDrawer.mapping-gate.integration.test.tsx` — **6/6 passing** (4 pre-existing + 2 new):

- **new** — the toast exposes an `Open agent editor` action whose `onClick` opens the editor drawer **independently** of the auto-open. The test clears the `openDrawer` mock *after* the gate fires (dropping the auto-open call), then invokes `action.onClick()` and asserts `openDrawer("agentWorkflowEditor", { urlParams: { agentId } })` — proving the fallback works on its own.
- **new** — the description names both `input` and `messages`.
- pre-existing gate behavior (opens drawer instead of running; warning toast shown; code-agent runs normally) still passes.

`pnpm typecheck` — clean (0 errors).

### Visual proof — real running app

Booted the app, created a workflow-agent scenario target with **no** scenario input mapping, opened the scenario, and clicked **Save and Run**. The real mapping-gate toast fires (and the editor drawer auto-opens behind it):

real app — mapping-gate toast with Open agent editor CTA

The toast shows the new copy — *Map at least one scenario input — "input" or "messages" — to an agent input before running this workflow agent.* — and the **Open agent editor** action. The action's `onClick` reopens the editor drawer for the target (`openDrawer("agentWorkflowEditor", { urlParams: { agentId } })`), asserted independently of the auto-open in the integration test above.

Clicking that action in the live app opens the Edit Workflow Agent drawer for the agent (URL `drawer.open=agentWorkflowEditor&drawer.agentId=…`, caught mid-load):

real app — after clicking the CTA, the agent editor drawer opens

## Anything surprising?

- The toast deliberately names only the **input** half (`input` / `messages`). The drawer gate only checks `hasScenarioInputMapping`; output mapping is validated server-side at run time, so surfacing it here would be misleading.

## Release Notes

* **New Features**
  * Added "Open agent editor" action button to the mapping validation warning toast, enabling direct access to the agent editor when required scenario inputs are missing.
  * Enhanced warning message to explicitly show which scenario input fields ("input" or "messages") require mapping.

* **Tests**
  * Added integration tests for mapping validation warning toast behavior.

### DIFF
--- a/langwatch/src/components/scenarios/ScenarioFormDrawer.tsx
+++ b/langwatch/src/components/scenarios/ScenarioFormDrawer.tsx
@@ -291,6 +291,8 @@ export function ScenarioFormDrawer(props: ScenarioFormDrawerProps) {
                 },
                 meta: { closable: true },
               });
+              // Auto-open the editor now; the toast action above is the manual
+              // fallback if this auto-open races, is dismissed, or fails.
               openAgentEditor();
               return;
             }
--- a/langwatch/src/components/scenarios/ScenarioFormDrawer.tsx
+++ b/langwatch/src/components/scenarios/ScenarioFormDrawer.tsx
@@ -274,16 +274,24 @@ export function ScenarioFormDrawer(props: ScenarioFormDrawerProps) {
             // covers output validation. Share the input half of the rule with
             // the editor drawer via hasScenarioInputMapping.
             if (!hasScenarioInputMapping(mappings)) {
+              // Fallback affordance (#3411): even if the auto-open below races,
+              // is dismissed, or fails, the toast itself links back to the editor.
+              const openAgentEditor = () =>
+                openDrawer("agentWorkflowEditor", {
+                  urlParams: { agentId: target.id },
+                });
               toaster.create({
                 title: "Configure scenario mappings",
                 description:
-                  "Set up how this workflow agent maps scenario inputs and outputs before running.",
+                  'Map at least one scenario input — "input" or "messages" — to an agent input before running this workflow agent.',
                 type: "warning",
+                action: {
+                  label: "Open agent editor",
+                  onClick: openAgentEditor,
+                },
                 meta: { closable: true },
               });
-              openDrawer("agentWorkflowEditor", {
-                urlParams: { agentId: target.id },
-              });
+              openAgentEditor();
               return;
             }
           }
--- a/langwatch/src/components/scenarios/__tests__/ScenarioFormDrawer.mapping-gate.integration.test.tsx
+++ b/langwatch/src/components/scenarios/__tests__/ScenarioFormDrawer.mapping-gate.integration.test.tsx
@@ -344,16 +344,17 @@ describe("<ScenarioFormDrawer /> mapping gate", () => {
         );
       });
 
-      // Prove the action handler ALONE reopens the editor: clear the mock so
-      // the earlier auto-open call doesn't count, then invoke onClick. This is
-      // the fallback affordance — functional even if the auto-open raced/was
-      // dismissed.
       const toastArg = mockToasterCreate.mock.calls.at(-1)?.[0] as {
         action?: { label: string; onClick: () => void };
       };
+      // AC3: the auto-open fired first (existing behavior preserved)...
+      expect(mocks.mockOpenDrawer).toHaveBeenCalledWith("agentWorkflowEditor", {
+        urlParams: { agentId: "workflow-agent-1" },
+      });
+      // ...then clear it and prove the action handler ALONE reopens the editor —
+      // the fallback affordance, functional even if the auto-open was dismissed.
       mocks.mockOpenDrawer.mockClear();
       toastArg.action?.onClick();
-
       expect(mocks.mockOpenDrawer).toHaveBeenCalledWith("agentWorkflowEditor", {
         urlParams: { agentId: "workflow-agent-1" },
       });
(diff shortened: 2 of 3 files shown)


drewdrewthis:
## Review verdict: READY

Self-review via `/review` (principles, hygiene, security, test reviewers + Uncle-Bob/Metz-Beck/Fowler personas). Scope: the mapping-gate toast change in `ScenarioFormDrawer.tsx`, its integration test, and the two new bound scenarios in `workflow-agent-mapping-layer.feature`. No correctness defects found; no blocking threads.

### Applied from review
- **[test]** Split the two-`expect`s-in-one-`waitFor` in the description test — now waits for the toast, then asserts `input` / `messages` outside the callback so both are always evaluated.
- **[test][principles]** The action test now also asserts the **auto-open fired first** (before clearing the mock), so AC3 (auto-open preserved) is guarded here too — a refactor dropping the auto-open would now fail this test as well as the existing gate test.
- **[principles]** Added a clarifying comment that the trailing `openAgentEditor()` is the auto-open and the toast action is the manual fallback.

### Addressed from CodeRabbit
- Reworded the new scenario's `Then` step to user-facing language (was the internal `AgentWorkflowEditorDrawer` component name) — per the repo feature-file guideline.

### Non-blocking (no action required)
- **[hygiene]** The `toaster.create` call has no stable `id`, so rapid Save-and-Run clicks could stack duplicate toasts. This is pre-existing and consistent with the other user-click toasts in this file (`Select a target`, `Failed to run scenario`, `Scenario created`); the `id`+`isVisible` dedup convention (`MissingModelToast.tsx`) exists for background **retry-storm** toasts, not single-user-click warnings. Out of scope here — file a follow-up if dedup is wanted across the scenario toasts.
- **[hygiene]** `action.onClick` does not call `toaster.dismiss`. Not needed: Chakra's `Toast.ActionTrigger` auto-dismisses the toast on click — verified in browser-QA (the toast disappears when the action is clicked).
- **[principles]** `openDrawer("agentWorkflowEditor", …)` is reachable via both the auto-open and (later) the action. Safe: it is the same pre-existing call keyed by `agentId`; the action is only clickable after the toast appears, never both in one synchronous flow.
- **[security]** No issues — `target.id` is a non-sensitive agent id already used by the pre-existing drawer-open.

### Proof
- Integration tests render the **real** `ScenarioFormDrawer` and exercise the **real** `handleSaveAndRun` gate (6/6 pass; run under the CI `test-integration` lane).
- Visual proof embedded in the PR body.

PR #5013, opened 2026-06-21
feat(triggers): outbox heartbeat primitive + graph trigger via outbox (ADR-033 + ADR-034 Ph 4+5)

0xdeafcafe:
## Summary

Moves custom-graph alert triggers off the K8s cron onto an event-sourced path. **Gated per-project by `release_es_graph_triggers_firing` — flag defaults OFF; cron coexists for un-flagged projects so existing customer alerts keep firing exactly as today.**

- **ADR-033 + Phase 4 — outbox heartbeat primitive.** Framework-level Redis-locked, worker-only, leader-elected timer that emits `OutboxEnqueueRequest[]` via the same `dispatchOutboxEnqueues` event-driven reactors use. One canonical handler regardless of wake source. Extracted shared `dispatchOutboxEnqueues` from `outboxReactorAdapter` so trace + heartbeat dispatches are byte-identical.
- **Phase 5 — graph trigger via outbox + heartbeat.** Real-time `.withOutbox(\"traceAnalytics\", \"graphTriggerEvaluation\", …)` (debounced per (trigger, project) 5 s); heartbeat (30 s) covers ONLY absence cases real-time can't see (no-data + firing-resolve when traffic stops). New `graphEval` outbox stage (graph evals have no traceId so don't fit settle/cadence's (trigger, trace) shape). Single `evaluateGraphTrigger` handler. Cron flag-checks per trigger's project; skips when flagged — minimal single-line edit, cron otherwise untouched.

## Deployment Impact

- **No database migrations.**
- **Feature flag (default OFF):** `release_es_graph_triggers_firing` gates graph-trigger firing through the outbox path.
- **Worker-only wiring:** the outbox heartbeat primitive and graph-trigger evaluation reactor run under `processRole === "worker"`; no new deployment unit, no web-process changes.
- No env vars or chart changes.

## Test plan

- [ ] Verify `release_es_graph_triggers_firing` defaults OFF; cron path unchanged for unflagged projects
- [ ] On a flagged canary project: breach fires via real-time path within seconds (debounced)
- [ ] On a flagged canary project: metric goes silent → no-data alert fires from heartbeat within 30 s
- [ ] On a flagged canary project: firing trigger's traffic stops → heartbeat resolves within 30 s
- [ ] Dedup: same trigger evaluated twice in quick succession → only one notification
- [ ] Flag-flip mid-flight: no double-fire from cron + new path
[...]

PR #5015, opened 2026-06-21
feat(automations): graph alerts in automations drawer + Liquid template wiring (ADR-034 Ph 5.1, 5.2, 8, 8.1)

0xdeafcafe:
## Summary

The user-facing layer of the graph-trigger migration (ADR-034), plus four
features this branch grew after the original scope was written. All of it is
authoring + dispatch surface on top of the outbox substrate from lower in the
stack.

- **Graph alerts in the automations drawer (ADR-034 Ph 5.1 / 5.2 / 8 / 8.1).**
  `automation.upsert` accepts a discriminated `graphAlert` sub-shape when
  `customGraphId` is set; the dashboard chart-card and `/analytics/custom`
  "Add alert" entry points open the same drawer pre-filled; graph-trigger
  dispatch routes through Liquid templates with a `GraphAlertTemplateContext`
  (history, sparkline, previousValue, graph.url) instead of the cron's
  hardcoded format. Legacy `AlertDrawer` deleted (verified unreachable).
- **Scheduled reports + a generic scheduler (ADR-042).** A `REPORT` trigger
  kind, an in-process `ScheduledJob` scheduler (Postgres row as source of
  truth, conditional claim as the exactly-once mechanism), and report dispatch
  that renders real tables/charts.
- **Facet-shaped automations drawer + trace-query subjects (ADR-043).** A
  `Trigger.filterQuery` liqe subject evaluated both in ClickHouse (list preview)
  and in-memory (dispatch).
- **Block Kit template suite + Slack bot-token delivery (ADR-040 / ADR-041).**
  An allowlisted Block Kit renderer and `chat.postMessage` delivery over an
  encrypted bot token with a channel picker.
- **Outbox payload slimming.** The reactor outbox carries identities, not
  customer trace content.
- **Unified automations page + drawer rework.** `/automations` lists every
  kind on one page (with scheduled/history views); the drawer gained a Build
  sidebar and a Schedule kind, and the tables/cells were reworked to match.
- **Home announcement carousel.** The home banner slot is a single morphing
  carousel (automations + voice-agents slides) replacing the separate banners.
  The traces-v2 slide was dropped in the main merge — Trace Explorer became
  the default on main (#5663) and the opt-in preference it set no longer
  exists.

## Deployment Impact

- **Five Postgres migrations ship** (`prisma migrate deploy`):
  - `ScheduledJob` — new table backing the reports scheduler.
  - `TriggerKind` — new enum plus a **NOT NULL `Trigger.triggerKind` column
    with a data backfill** (`UPDATE "Trigger" SET "triggerKind" = 'ALERT'
    WHERE "customGraphId" IS NOT NULL`).
  - `Trigger.filterQuery` — nullable column, no backfill.
  - `TriggerSent.openIncidentKey` — nullable column + single-column unique
    index (the atomic open-incident claim for graph alerts).
  - `ScheduledJob.attempts` (NOT NULL default 0) + `lastError` (nullable) —
    the report scheduler's retry bookkeeping.
- **New in-process service:** the `ScheduledJob` scheduler runs under the
  worker role. No new external service, chart, or infra dependency.
- **No new feature flags.** Graph-alert *firing* still rides the existing
  `release_es_graph_triggers_firing` from PR #5013 (defaults OFF); the cron
  fires un-flagged projects. Reports and the drawer are on for all projects
  (ADR-042's `release_scheduled_reports` gate was not implemented — see below).
- **New env vars:** none beyond the Slack app credentials the bot-token
  delivery reads.

## Review + hardening pass

A read-only audit of this branch (the diff was too large for CodeRabbit's
150-file limit, so it self-skipped) found defects that had shipped green
because the report and Slack-bot paths carried no tests. Every fix below has a
regression test that fails without it.

- **Data at rest:** the EE trace reactors no longer write trace input/output
  onto the outbox payload; the audit redactor is an allow-list.
- **Reports fire only on schedule:** REPORT triggers are excluded from the
  trace-trigger set (they were notifying on every ingested trace); editing a
  report no longer rewrites it as a trace automation and now retires its
  schedule when converted to another kind; the authored trace query is
  persisted; pausing deactivates the `ScheduledJob`; bad cadences are rejected
  at the boundary; the report window is the real previous slot.
- **Delivery guarantees (review follow-up):** a graph alert now **claims its
  open incident atomically before dispatch** (single-winner, via a nullable
  unique identity key), so concurrent evaluators can't double-open or
  double-send; an alert that delivered nothing rolls the claim back rather than
  latching "firing" (and a **thrown** dispatch rolls it back too, so the
  outbox retry re-dispatches instead of backing off as already-firing); a per-fire idempotency key stops outbox retries
  re-notifying; and the report **scheduler leases each slot and retries on
  failure** (bounded, then abandons observably) instead of silently dropping
  the slot when a send throws.
- **Cron delivery:** the default-path cron learned bot delivery and the Liquid
  templates; its recipient-dedup key is per-fire, not constant for the graph.
- **Filter subjects:** own-property lookup (no `constructor`/`__proto__`
  crash-or-match-everything); origin/span-status SQL coalesced to match the
  in-memory reads.
- **Transport:** the SSRF fetch's timeout is now real and its body is streamed
  under the cap; the Slack channel list pages through `next_cursor`.
- **Third audit round (post-rework):** every mutation returning a Trigger row
  now redacts the encrypted Slack bot token (`graphs.getAll` included), and
  the upsert persists per-action-parsed params so a token typed before
  switching the channel can't land in plaintext; the ADR-031 email caps moved
  inside `dispatchGraphAlertAction` so the event-sourced path is capped like
  the cron; `ScheduledJob.currentSlot` pins the in-flight calendar slot so a
  retried report keeps its real window instead of the backoff span; the
  scheduler gates on `roleRunsWorkers()` (reports now fire in single-process
  dev); negated free-text filters mirror ClickHouse NULL semantics in the
  in-memory evaluator; the View drawer renders `filterQuery` subjects; and a
  thrown graph-alert dispatch rolls back its open claim.

## Known limitations / deliberately not done

- **`release_scheduled_reports` does not exist.** ADR-042 specifies reports
  ship dark behind it; they currently ship on for every project. Gate before
  GA if a staged rollout is wanted.
- **Report delivery is at-least-once.** The scheduler retry closes the
  silent-slot-loss hole, but a crash after a provider send but before the
  success write can re-send on lease expiry. Full exactly-once needs the
  ADR-042 §4 outbox routing — a tracked follow-up.
- **Unsupported filter fields fail closed silently.** A trace-query subject
  using a field the dispatcher can't evaluate (`spanType`, `size`, `scenario*`)
  compiles, saves, and then never matches. This PR adds a `logger.warn` so it's
  visible; rejecting at save-time vs. deriving spans at dispatch is left open.
- **Catch-up fires the most recent missed slot** after an outage and
  fast-forwards to the first future slot (ADR-044 `runLatest`); it never replays
  every missed slot.

## Test plan

- [ ] `prisma migrate deploy`, then author a graph alert end-to-end from the
      drawer and from a dashboard chart card; verify the Trigger row.
- [ ] Create a scheduled report; verify one `ScheduledJob` row, exactly-once
      delivery on the slot, pause/resume toggles the job, edit hydrates it as a
      report, and converting it to a trace automation deactivates its schedule.
- [ ] Send trace traffic and confirm the report does **not** notify per trace.
- [ ] Reject a bad cron and a sub-15-minute cadence at save.
- [ ] With `release_es_graph_triggers_firing` OFF, fire a bot-delivery Slack
      graph alert and confirm it delivers via the cron path.
- [ ] Author a trace-query subject and confirm the "N traces matched" preview
      agrees with what the automation fires on (esp. `origin:`).
- [ ] After any trace automation fires, confirm no `ReactorOutbox` payload
      contains trace input/output.
- [ ] Unit: `pnpm test:unit src/features/automations src/automations
      src/shared/templating src/server/app-layer src/server/triggers
      src/server/event-sourcing src/pages/api/cron ee`

Was stacked on `pr/05-eval-and-write-side-aggregates` — PR #5014 merged 2026-07-10, and `origin/main` has since been merged back in, so this branch now lands on `main` directly.


langwatch/src/server/app-layer/triggers/graph-trigger-evaluation.service.ts
```diff
@@ -321,43 +333,57 @@ export async function evaluateGraphTrigger({
...
+    // extract + decrypt the token here so the dispatch helper stays crypto-free.
+    const slackParams = (trigger.actionParams ?? {}) as SlackActionParams;
+    let botDestination: { token: string; channel: string } | null = null;
+    if (slackDeliveryMethodOf(slackParams) === "bot") {
+      const token = decryptSlackBotToken(slackParams);
+      const channel = slackParams.slackChannelId?.trim();
+      if (token && channel) botDestination = { token, channel };
     }
 
+    await deps.notifier.dispatch({
```
langwatch-agent: **P1** - This check-then-send-then-insert sequence does not preserve the alert's at-most-once incident guarantee once the event-sourced path can run on multiple workers/reasons. Two evaluators can both pass `findOpenForGraphAlert()` before either reaches `createOpenForGraphAlert()`, both call the provider here, and only afterwards create separate open `TriggerSent` rows. The existing `@@unique([triggerId, traceId])` does not protect graph alerts because `traceId` is `NULL`, and there is no partial unique constraint for `(projectId, triggerId, customGraphId) WHERE resolvedAt IS NULL`. Make the open-incident claim atomic before any provider side effect, for example with a partial unique index plus a `create/open-if-absent` claim method where only the winner dispatches, and cover the concurrent evaluator case in a test.

langwatch/src/server/app-layer/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,414 @@
...
+        projectId: job.projectId,
+        targetType: job.targetType,
+        targetId: job.targetId,
+      });
+      return;
+    }
+
+    // ATOMIC CLAIM. Only one worker's conditional UPDATE wins; every other
+    // worker sees a lost claim and skips. This is the exactly-once guarantee.
+    const won = await this.repo.claim({
```
langwatch-agent, line 404: **P1** - The scheduler advances `nextRunAt`/`lastSlot` before the report handler runs, and `fireJob()` only logs handler errors. For scheduled reports the handler does the actual email/Slack send synchronously, so a transient provider error, ClickHouse timeout while rendering, or handler crash permanently skips that calendar slot: the row has already moved to the next run and nothing retries the missed report. This also diverges from ADR-042's stated design of firing into a retrying outbox path. Keep the slot retryable until durable dispatch is recorded, or enqueue a durable outbox/report job atomically with the claim/advance and let provider delivery retry/dead-letter there. Add a regression where the first handler/send throws and the same slot is retried or remains observable instead of being silently lost.

langwatch/src/server/api/routers/automations.ts
```diff
@@ -587,18 +1004,66 @@ export const automationRouter = createTRPCRouter({
...
+              ),
+              traceDebounceMs:
+                input.traceDebounceMs ?? DEFAULT_TRACE_DEBOUNCE_MS,
+              ...data,
+            },
+          });
+        }
+      }
+
+      if (isReport && input.report) {
```
langwatch-agent, line 1138: **P2** - This only creates/refreshes a schedule when the saved draft is still a report. If a user edits an existing report trigger into a trace automation or graph alert, the old `ScheduledJob(reportTrigger, trigger.id)` remains active because `removeReportSchedule()` only runs on delete. That stale job will keep waking forever and the report handler will repeatedly load a non-report trigger, fail to parse `actionParams`, and skip/log every cadence. When updating an existing trigger, detect report -> non-report transitions and deactivate the schedule, or unconditionally deactivate before saving non-report kinds. Add an update test that starts from a report row with a schedule, saves it as a non-report, and asserts the scheduled job is inactive.

langwatch-agent (changes requested):
3 findings: 2 P1 delivery-guarantee issues and 1 P2 scheduler lifecycle bug. Details are inline.

langwatch/src/server/app-layer/triggers/graph-trigger-evaluation.service.ts
0xdeafcafe: Addressed in ab8160d7b. Reordered to claim-before-send: the evaluator now atomically claims the open incident before any provider call. Rather than a partial composite unique, this uses a single new nullable column `TriggerSent.openIncidentKey` with a plain unique index — it holds the incident identity (`graph-alert:${triggerId}`) while firing and is nulled on resolve, so Postgres NULL-distinctness gives "at most one open incident per identity". The insert loser (P2002) backs off without dispatching. Reconciled with the existing `didSend` gate: a dispatch that delivered nothing rolls the claim back (`deleteOpenClaim`) so the alert is not left "firing" and the next evaluation retries. Concurrency-race, rollback, and resolve-frees-identity tests added.

langwatch/src/server/app-layer/scheduler/scheduler.service.ts
0xdeafcafe: Addressed in ab8160d7b. The claim now leases the slot (pushes `nextRunAt` a 10-min lease-window out) instead of advancing to the next cron instant, and `lastSlot` only advances once a fire is delivered. A handler failure backs off (bounded exponential, base 60s) and retries up to MAX_ATTEMPTS=5, then abandons the slot to the next instant with `lastError` kept and a `logger.error` + `captureException` — so a transient send error is retried or at worst left observable, never silently dropped. New `ScheduledJob.attempts`/`lastError` columns carry the bookkeeping. Delivery is now at-least-once (a crash after send but before settle can re-fire on lease expiry), which is documented as the accepted ADR-042 tradeoff over the previous silent zero-delivery. Full outbox routing per §4 remains the longer-term direction; this closes the data-loss hole. Regression test drives the loop with a throwing handler and asserts the slot is retried, plus the abandon-after-cap case.

langwatch/src/server/api/routers/automations.ts
0xdeafcafe: Addressed in ab8160d7b. The upsert now retires the `ScheduledJob` whenever the saved kind is not a report (`removeReportSchedule` is idempotent), so editing a report into a trace automation or graph alert deactivates its schedule instead of leaving it waking against a non-report row. `toggleTrigger` already handled pause/resume. Update test covers report -> non-report deactivating the job.

0xdeafcafe:
@coderabbitai review

langwatch/src/server/app-layer/triggers/graph-trigger-evaluation.service.ts
```diff
@@ -321,50 +351,143 @@ export async function evaluateGraphTrigger({
...
+      return {
+        triggerId,
+        projectId,
+        reason,
+        status: "already_firing",
+        value: currentValue,
+      };
+    }
+
+    const dispatchResult = await deps.notifier.dispatch({
```
langwatch-agent: @/tmp/pr5015-finding.md

langwatch-agent (changes requested):
@/tmp/pr5015-review-body.md

langwatch/src/server/app-layer/triggers/graph-trigger-evaluation.service.ts
0xdeafcafe: This comment's body is a dangling local-file reference (`@/tmp/pr5015-finding.md` was never expanded), so the finding text never reached GitHub. Re-derived from the anchor: `dispatch()` sits after the atomic claim, and only the `didSend: false` return path rolled the claim back — a **thrown** dispatch (provider/network failure) orphaned the open incident, so the outbox retry backed off as `already_firing` and the notification was lost. Addressed in 95922b51f: the claim is deleted (best-effort) before the error rethrows, with unit tests for the retry-re-dispatch path and the rollback-failure path. If the original finding was something else, please re-post the content.

langwatch/src/server/app-layer/scheduler/scheduler.service.ts
```diff
@@ -0,0 +1,622 @@
...
+    const slot = job.currentSlot ?? job.nextRunAt;
+    // The lease is still conditioned on the row's CURRENT wake instant — that
+    // is what `findDue` read and what a racing worker would also condition on.
+    const claimAt = job.nextRunAt;
+
+    // The next cron instant strictly after the DELIVERED slot, in the job's own
+    // zone (DST-correct). Computed up front for TWO jobs: (a) it detects a
+    // poison cron/tz (deactivate + bail *before* leasing so a bad row can't
+    // wedge the loop), and (b) it is the value we advance to on success or on
+    // abandon. `after: slot` (not `after: now`) keeps the calendar honest —
```
langwatch-agent: **P1 - Outage recovery replays every missed report, contrary to the declared catch-up policy.**

Computing `nextSlot` from `slot` and then settling to it means that a job which was overdue for several periods remains immediately due after every successful delivery. The next loop iteration claims it again, so a week-long outage for a daily report sends seven stale reports; a 15-minute report can send hundreds. The ADR added in this PR explicitly requires `runLatest` to fire one catch-up and says the scheduler must never replay every missed slot.

Compute the recovery target from `now` (or add the declared `catchUp` policy), so after the chosen catch-up the row advances to the first future cron instant. Add an integration test with a job several intervals behind and assert one delivery/one future `nextRunAt`.

langwatch/src/server/app-layer/reports/report-chart.service.ts
```diff
@@ -0,0 +1,286 @@
...
+  deps: ReportChartDeps;
+  source: ReportSource;
+  projectId: string;
+  from: number;
+  to: number;
+}): Promise<ReportChart[]> {
+  const graphs = await loadGraphs({ deps, source, projectId });
+  // Panels are independent queries; a dashboard with eight of them should not
+  // take eight round-trips in series.
+  return Promise.all(
```
langwatch-agent: **P1 - Dashboard reports issue unbounded concurrent ClickHouse queries.**

`Promise.all` starts one `getTimeseries` per dashboard graph with no local, per-project, or global limit. At a common schedule boundary, multiple workers can each run several reports, so a few large dashboards create a burst of expensive uncapped queries and can exhaust ClickHouse concurrency/memory for interactive traffic. The ADR in this PR calls out this exact failure mode and requires outbox/GroupQueue backpressure (or at minimum a bounded per-project concurrency policy plus jitter), but this implementation bypasses it.

Route chart computation through the existing bounded dispatch path, or introduce and enforce a small concurrency limit that composes with a worker/global and per-project budget. Cover a dashboard larger than the limit and concurrent scheduled reports.

langwatch/src/server/api/routers/automations.ts
```diff
@@ -587,22 +1083,80 @@ export const automationRouter = createTRPCRouter({
...
```
langwatch-agent, line 1138: **P2 - Saving a report and creating its durable schedule are separate, unreconciled writes.**

The Trigger row is created/updated above, then `syncReportSchedule()` performs a second write outside a transaction. A transient Postgres error, timeout, or process interruption between them returns an error to the client but leaves an active REPORT trigger with no `ScheduledJob`. Nothing at worker boot reconciles reports to schedules, so unless the user retries that exact save it will never run.

Make the trigger mutation and `ScheduledJob` upsert atomic, or add a durable reconciliation pass that repairs missing/incorrect report schedules. Add a failure-injection test proving an active report cannot remain unscheduled after this boundary fails.

langwatch-agent (changes requested):
I found three blockers in the scheduled-report delivery path.

I also rechecked the resolved incident-claim, rollback, and Slack-token redaction work at this head; those changes address the earlier findings. The remaining issues are delivery/load correctness, not style.

CI is still in progress, and CodeRabbit skipped this 304-file PR.

langwatch/src/server/app-layer/scheduler/scheduler.service.ts
0xdeafcafe: Fixed in 189e7609c — implemented ADR-044 `runLatest`.

A fresh, overdue fire now delivers exactly **one** catch-up for the **newest** missed slot (new pure helper `computeCatchUp`) and fast-forwards `nextRunAt` to the first cron instant strictly after `now`, so a multi-slot backlog no longer replays every missed slot (a week-long outage of a daily report used to send seven). The claim now pins the fired catch-up slot in `currentSlot` (new `slot` claim param) so a retry re-fires that slot, not the oldest missed one. On-time fires are unchanged (the catch-up collapses back to the due slot).

Regression tests: `nextRunAt.unit.test.ts` covers `computeCatchUp` (on-time no-op, multi-slot backlog → newest slot + future `nextRunAt`); `scheduler.service.unit.test.ts` drives the real loop against a stateful repo that re-serves the row while due, asserting **one** delivery and a future `nextRunAt` for a job several intervals behind.

langwatch/src/server/app-layer/reports/report-chart.service.ts
0xdeafcafe: Fixed in 189e7609c.

`loadReportCharts` no longer fans every panel out with an unbounded `Promise.all`. Panel queries now run through an order-preserving, concurrency-limited map capped at `REPORT_CHART_QUERY_CONCURRENCY` (= 3) (ADR-044 §5). Because each worker fires reports sequentially, per-worker in-flight `getTimeseries` queries are bounded by the cap, so a large dashboard (or several coinciding reports) can't burst ClickHouse concurrency/memory. The full outbox/GroupQueue fan-out remains the tracked follow-up; this is the "small enforced concurrency limit" backstop.

Regression test in `report-chart.service.unit.test.ts`: a dashboard with 3× the cap panels asserts max concurrent `getTimeseries` ≤ the cap (and > 1, i.e. genuinely overlapped), with every chart returned in dashboard order.

langwatch/src/server/api/routers/automations.ts
0xdeafcafe: Fixed in 189e7609c via the durable self-healing path ADR-044 §4 endorses (durable row + reconciling scan, not a parked queue message).

Added `TriggerService.reconcileReportSchedules()`, run fire-and-forget at worker boot: it reads every active REPORT trigger cross-tenant (`findActiveReportTargets`) and creates a `ScheduledJob` for any that lacks one — so an active report left unscheduled by a crash between the two writes is repaired without the user re-saving. It is **create-if-missing only**: a report that already has a schedule row (including a *paused* report's inactive row) is left untouched, so a live calendar is never reset and a paused schedule is never resurrected. Running on every worker is race-safe — the create is hardened on the `(targetType, targetId)` unique.

Failure-injection tests in `trigger.service.report-schedule.unit.test.ts` prove: an active report whose schedule write was lost gets its schedule created; a paused report and an already-scheduled report are not disturbed; unparseable actionParams are skipped.

PR #5017, opened 2026-06-21
feat(datasets): born-on-storage + trailing-chunk compaction

sergioestebance:
Stacked on #4908 (`feat/datasets-s3-reads`). Tackles two of the open dataset→S3 follow-ups; the third (CORS in IaC) is a separate `langwatch-saas` Terraform PR.

## 1. Born-on-storage — all create paths write `s3_jsonl` (cutover step 1)
`createNewDataset` — the chokepoint every non-UI create funnels through (tRPC `create`, REST `POST /` and `/upload`, copy/clone, SDK/MCP) — now writes records as chunked JSONL from index 0 and borns `contentLayout='s3_jsonl'`, `status='ready'` with PG-authoritative counters, instead of `'postgres'`.

- **Gate = "Always"**: S3 when configured, else local FS (`resolveProjectStorageDestination` always returns a backend). So `postgres` is never created for new data and the backfill drains to zero.
- Chunks are written **before** the row (a write failure leaves no orphan row); no advisory lock (the row doesn't exist yet).
- Record appends already routed to `appendS3JsonlRecords` — unchanged.

## 2. Trailing-empty-chunk compaction on delete
`deleteS3JsonlRecords` now reaps the contiguous **trailing** empty chunks a delete leaves behind (`deleteChunksFrom` + `chunkCount` decremented + offsets trimmed), so `chunkCount`/object-count don't grow unbounded under churn. Trailing-only by necessity (chunk keys are positional; a **middle** empty stays in place). Emptying the whole dataset trims to `chunkCount=0`.

Adds a typed `MissingChunkError` so the I-COUNT repair (`recomputeDatasetCounts`) tells a recoverable trailing gap (the rare PG-commit-after-S3-delete residual) from genuine mid-set corruption — it re-derives `chunkCount` from the present prefix instead of throwing. Full re-pack (middle empties) is deferred to a manual action.

## Tests
All dataset unit tests green; full `pnpm typecheck` clean. New coverage: born-`s3_jsonl` on local FS (real write + read-back), empty-create, copy-from-`s3_jsonl` via `writeChunks`; trailing-trim (single + whole-dataset), middle-empty-stays, recompute trailing-gap tolerance + non-missing-chunk error still propagates.

## Deployment Impact

**Storage becomes mandatory for dataset creation — there is no Postgres fallback.** Born-on-storage writes every new dataset to object storage, so each deployment must have a writable backend:

- **Cloud / S3 (`S3_BUCKET_NAME` set):** no change.
- **Self-hosted without S3:** requires a writable `LANGWATCH_LOCAL_STORAGE_PATH` (default `/var/lib/langwatch/objects`). The Helm chart already mounts this as a persistent volume. **Bare-docker / manual installs must point it at a writable, persistent directory** — otherwise dataset *create* fails with an actionable error (these installs previously stored datasets in Postgres). Tests set a tmp path automatically.
- **No schema migration**; no new required env var for S3 installs.
- Browser direct-upload needs **bucket CORS** (documented in ADR-032 / chart NOTES; IaC tracked separately).

This narrows the **I-SELFHOST** invariant (no-S3 still works, but now requires a writable local FS root rather than falling back to PG) — recorded in ADR-032 v13.


Aryansharma28:
**Review** — `/review` pass on the diff. Reasoning trail is strong (the comments at `dataset-mutations.ts:1163-1180` and the v13 ADR entry are excellent). Findings split into blockers vs non-blockers below.

## Blockers

### B1 — Caller-supplied record `id`s are silently dropped on create (contract change)

`langwatch/src/server/datasets/dataset.service.ts:316-321`:
```ts
const entries = (datasetRecords ?? []).map((record) => {
  const { id: _id, ...entry } = record;
  return entry;
});
```

`writeInitialS3JsonlChunks` then mints fresh `record_<nanoid>` ids per row (`dataset-mutations.ts:1047-1054`). The old `createNewDataset` path passed `datasetRecords` through `createManyDatasetRecords`, which honored an existing `id` if supplied — the removed integration test (`dataset-upload-api.integration.test.ts:824-890` in old tree) explicitly exercised this via the unique-constraint collision.

So callers (SDK / MCP / REST) that pinned ids at create no longer get them honored. The surface is also now **inconsistent**: `appendS3JsonlRecords` *does* honor `forcedIds` via `appendLinesInTx`, but create doesn't — even though `toChunkLines` already accepts a `forcedIds` param.

Pick one:
1. Thread `forcedIds` through `writeInitialS3JsonlChunks` → `toChunkLines` (plumbing exists for append).
2. If the design intent is "create never accepts ids" — document the contract break in the PR description + ADR, confirm REST/tRPC create input schema rejects `id`, and add a test that asserts a posted `id` is *ignored* (not silently honored on some path and dropped on another).

### B2 — `repository.create` accepts the new s3_jsonl fields, but the diff doesn't show the type change

`dataset.service.ts:323-336` now passes `contentLayout`, `status`, `rowCount`, `sizeBytes: BigInt(...)`, `chunkCount`, `chunkOffsets`. The diff removes `useS3` from the old shape but doesn't include the `DatasetRepository.create` input-type widening that must accompany this. The PR description says `pnpm typecheck` is clean, so it presumably compiles — but reviewers reading the diff alone can't verify the new fields aren't silently dropped by a narrower `Prisma.DatasetCreateInput`-like input type. Either include the repository type change in the diff or paste the resolved input shape in the PR description.

## Non-blockers

### N1 — Two `getDatasetStorage(projectId)` resolutions in the create path

`createNewDataset` calls `writeInitialS3JsonlChunks` (resolves storage once, no explicit `storage` arg) and on row-insert failure calls `deleteAllS3JsonlChunks` (resolves *again*). If the project's storage config flipped between the two calls, the reap could target a different backend than the write. Vanishingly unlikely, but pass the resolved `storage` instance through both calls — same cost, no race.

### N2 — `toChunkLines.forcedIds` parameter is dead in the create path

`dataset-mutations.ts:1047-1054` — the helper accepts `forcedIds`, but `writeInitialS3JsonlChunks` never passes it. Either thread it (resolves B1 option 1) or drop the parameter from `toChunkLines` for now and re-add when needed (YAGNI). The current shape invites the next contributor to wonder which is intentional.

### N3 — Empty-dataset state (`chunkCount: 0`, `chunkOffsets: []`)

The whole-dataset-emptied test asserts this state, and `readChunks({chunkCount: 0})` returns `[]` per the no-S3 unit test. Worth a quick grep across read sites for any `for (let i = 0; i < chunkCount; i++)` or `chunkOffsets[0]` access that pre-PR assumed `chunkCount >= 1` (delete-all is a new state-shape that didn't exist before; before this PR a dataset that emptied still had its chunks in place).

### N4 — Row-insert-failure reap key scope

`deleteAllS3JsonlChunks` on row-insert failure deletes `datasets/<projectId>/<datasetId>/chunk-*` — `datasetId` is a fresh nanoid scoped to *this* losing attempt, so there's no overlap with a concurrent winner. The code is correct, but a one-line comment at the reap call site noting the key scope would save the next reader from re-deriving it.

### N5 — Self-host rollout call-out

`I-SELFHOST` is narrowed: bare-docker installs without `LANGWATCH_LOCAL_STORAGE_PATH` set will fail every dataset-create with the new actionable error from `LocalDatasetStorage`. The error message is exactly right (names the root + both env vars), but Helm chart NOTES / release notes should call this out loudly. Worth confirming with SRE/docs.

### N6 — `getRecords` integration-test helper hardcodes `?limit=1000`

`dataset-upload-api.integration.test.ts:36-45`. Fine for current tests, will silently truncate if a future test creates >1000 rows. Either accept a `limit` arg or leave a comment.

### N7 — Rollback escape hatch (worth considering, not necessarily fixing here)

This PR is a one-way door — born-on-storage removes the Postgres fallback entirely. If a customer install hits an unforeseen issue, recovery is "revert PR" not "flip a flag." Consider whether a `BORN_ON_STORAGE_DISABLED` env-gated fallback for one release cycle is worth the carrying cost. v13 ADR doesn't discuss reversal.

---

The atomicity story (chunks-before-row + best-effort reap at both seams), the trailing-only logical compaction, and `MissingChunkError` loud-on-corruption are all the right shape and well-tested. **B1 is the only thing I'd want resolved before merge** — the rest are verify-and-confirm or follow-ups.

sergioestebance:
Thanks for the thorough pass. Resolutions below.

**B1 — caller-supplied ids dropped on create** ✅ Fixed (`40c7162f4`). Took option 1: `forcedIds` is threaded through `writeInitialS3JsonlChunks` → `toChunkLines` (per-row, optional — honor a pinned id, mint `record_<nanoid>` only where absent), restoring parity with the append path and the pre-cutover contract. Resolves **N2** (the param is no longer dead). Tests: unit (`writeInitialS3JsonlChunks` honors forcedIds + mints in gaps) and end-to-end (`upsertDataset` create with a pinned id, read back from a real local-FS chunk).

**B2 — repo `create` input type** ✅ Verified safe. `CreateDatasetInput = Omit<Prisma.DatasetCreateInput, "project" | "datasetRecords" | "batchEvaluations"> & { projectId }` — the new s3_jsonl scalar fields pass straight through; nothing is narrowed/dropped (typecheck confirms). No change needed.

**N1 — two `getDatasetStorage` resolutions** ✅ Fixed (`40c7162f4`). `createNewDataset` resolves storage once and threads the instance through both the write and the failure reap.

**N3 — empty `chunkCount: 0` read sites** ✅ Verified safe. Every read is `for (i = 0; i < chunkCount; i++)` → returns `[]` for 0; no `chunkOffsets[0]` / `chunkCount - 1` access that assumed `>= 1`.

**N4 — reap key scope** ✅ Comment added at the reap call site (`datasetId` is a fresh per-attempt nanoid → never touches a concurrent winner's chunks).

**N5 / N6** — noted as follow-ups (Helm NOTES self-host call-out; test helper `limit=1000`). The storage-mandatory consequence is already in ADR-032 v13 + `.env.example`; will mirror to chart NOTES.

**N7 — rollback escape hatch** — Deliberate: born-on-storage with no Postgres fallback is the locked decision in ADR-032 v13 (the corpus is hard-capped and new data is born on storage, so reversal is a revert, not a flag). Left as-is by design.

Separately, three follow-up findings (s3_jsonl list record count reporting 0, the staging-route orphan guard, and the integration test for the new local-FS `createPresignedUpload`) were fixed in `8fc4ad981`.

sergioestebance:
## 🧪 prove-them-wrong — both storage paths, real browser

Tried to break it: uploaded a **5,000-row CSV through the real UI** in each storage mode (switching `.env` + restarting the server between runs), with `answer == expected_answer` on **every** row to also stress the papaparse `_1` corruption fix. Verified the UI render, the actual storage backend, and the stored bytes.

### Scenario 1 — No S3 configured → local filesystem
local-fs dataset, 5000 rows, no corruption

```
chunk landed: ~/.langwatch/objects/datasets/project_ttk6lbC1.../dataset_LwP.../chunk-00000.jsonl
S3 for same project: (none)
stored rows: 5000 | expected_answer !== answer (corruption): 0
```

### Scenario 2 — S3 configured → s3://runtime-storage-dev
s3 dataset, 5000 rows, no corruption

```
chunk landed: s3://runtime-storage-dev/datasets/project_MjZE1.../dataset_LtBS.../chunk-00000.jsonl  (1.2 MiB)
local FS for same project: absent (correct)
stored rows: 5000 | expected_answer !== answer (corruption): 0
```

### Verdict

| 1 | No S3 → dataset stored on local FS, readable in the platform | `chunk-00000.jsonl` under `~/.langwatch/objects/...`, **not** in S3; UI shows 5000 rows | **PASS** |
| 2 | S3 configured → dataset stored in S3, readable in the platform | `chunk-00000.jsonl` in `s3://runtime-storage-dev/...`, **not** on local FS; UI shows 5000 rows | **PASS** |
| 3 | `.env` switch actually re-routes storage (no leakage) | each dataset present in exactly ONE backend, never both | **PASS** |
| 4 | CSV normalize no longer corrupts equal cells (`_1`) | 0/5000 rows where `expected_answer != answer`, both backends; verbatim in both screenshots | **PASS** |
| 5 | s3_jsonl record count is correct in the UI | header reads **"5000 records"** (not 0) in both | **PASS** |

**Verdict: 5/5 proven, 0 disproven, 0 unverified.** Both storage paths work end-to-end (UI + backend), data is uncorrupted, and the record count is right.

### Findings (logged anomalies — none touch the claims)

- **Pre-existing (not this PR):** React hydration warning on the *empty* datasets page — `In HTML, <p> cannot be a descendant of <p>` / `<div> cannot be a descendant of <p>`, from `NoDataInfoBlock` (the "No datasets" empty state nests a `<div>`/`<Button>` inside a `<p>`). Console-only, no visible breakage. Worth a separate cleanup; unrelated to storage.
- **Benign (headless env):** PostHog `Failed to fetch` (telemetry, no network) and transient tRPC `Failed to fetch` on first paint; a single `Dataset is not ready (status: processing)` during the normalize poll, which resolved to `ready`.

<sub>Driven headlessly via the project's Playwright + system Chrome; fresh signup + seeded project per run; chunk bytes downloaded from each backend and diffed against the source.</sub>

Aryansharma28 (approved):

PR #5069, opened 2026-06-22
feat(datasets): upload CSV drawer + redesigned dropzone

sergioestebance:
## Why

The "Upload CSV" surface was registered as a drawer but rendered a centered modal, and the direct upload (ADR-032 D4) navigated away to the dataset page to watch async processing. This turns it into one coherent drawer flow and gives the dropzone the polish the rest of the dataset UI already has.

Builds on the open epic PR #4908 (`feat/datasets-s3-reads`); #5067 (scan-before-lock) is already merged into that branch.

## What changed

- **Modal → drawer:** `UploadCSVModal` → `UploadCSVDrawer`; renders a `Drawer` (size `xl`, matching the `AddOrEditDatasetDrawer` it hands off to). Component + file renamed so the name matches what it renders.
- **Async upload stays in the drawer:** the direct upload no longer navigates to the dataset page; it polls status in place (processing → ready / failed, with retry), refreshes the list on ready, and offers an explicit "View dataset".
- **Redesigned dropzone:** dotted-grid empty state with a cloud icon + "Drag and drop file, or click to browse" and supported types; hover / drag-over highlight; the chosen file renders as a status row (icon + name + size·date + remove).
- **Error placement:** system/storage failures → alert at the top of the drawer; file-level validation (too large / over the row limit) → inline on the file's row.
- **Upload experience polish:** the cloud icon softly grows on hover; the file name sweeps a rainbow gradient while uploading; the trash control becomes an **X** that cancels the in-flight upload; the drag-and-drop container smoothly collapses once a file is chosen and expands back on remove/cancel; the Upload button shows a loading state.

## How it works

- Cancel threads an `AbortController` into `putFileToPresignedUrl` (new optional `signal`); on cancel it aborts the PUT, reaps the pending `uploading` row, and clears the file so the dropzone expands back in.
- The rainbow name reuses the PostHog rainbow-scroll recipe (same as `ShikiCommandBox`).
- The collapse uses the CSS grid `1fr→0fr` trick (no JS height measuring).

## Test plan

- Unit: `CSVReaderComponent.unit.test.tsx`, `directUpload.unit.test.ts` — 17 passing; `pnpm typecheck` clean.
- `UploadCSVFallbackGuard.integration.test.tsx` assertions (the `upload-error` hook) preserved; runs in CI (needs the container runner, not available locally).
- Spec: `specs/datasets/dataset-upload-dropzone.feature`.

## Deployment Impact

None. Frontend-only UI changes plus a backward-compatible optional `AbortSignal` parameter on `putFileToPresignedUrl`. No schema, API, or infra changes.

### DIFF
--- a/langwatch/src/components/datasets/AddRowsFromCSVModal.tsx
+++ b/langwatch/src/components/datasets/AddRowsFromCSVModal.tsx
@@ -19,7 +19,7 @@ import {
 } from "../../server/datasets/types";
 import { api } from "../../utils/api";
 import { tryToConvertRowsToAppropriateType } from "../AddOrEditDatasetDrawer";
-import { CSVReaderComponent } from "./UploadCSVModal";
+import { CSVReaderComponent } from "./UploadCSVDrawer";
 
 export function AddRowsFromCSVModal({
   isOpen,
--- a/langwatch/src/components/datasets/UploadCSVDrawer.tsx
+++ b/langwatch/src/components/datasets/UploadCSVDrawer.tsx
@@ -10,12 +10,14 @@ import {
   useDisclosure,
   VStack,
 } from "@chakra-ui/react";
-import { useEffect, useState } from "react";
+import { keyframes } from "@emotion/react";
+import { useEffect, useRef, useState } from "react";
 import {
   CheckCircle,
   CloudUpload,
   FileText,
   Trash2,
+  X,
   XCircle,
 } from "lucide-react";
 import {
@@ -377,6 +379,10 @@ export function UploadCSVForm({
   // the file's row (sizeError + overRowLimitForFallback below).
   const [uploadError, setUploadError] = useState<string | null>(null);
   const [sizeError, setSizeError] = useState<string | null>(null);
+  // Abort handle for the in-flight PUT + the pending dataset it minted, so a
+  // user-initiated cancel can stop the stream and reap the orphaned row.
+  const abortControllerRef = useRef<AbortController | null>(null);
+  const pendingDatasetIdRef = useRef<string | null>(null);
 
   const getValidName = async (proposedName: string): Promise<string> => {
     if (!projectId) return proposedName;
@@ -431,7 +437,11 @@ export function UploadCSVForm({
     setIsUploading(true);
     setUploadError(null);
     // Track the pending dataset so a presigned-PUT failure can clean up the
-    // orphaned `uploading` row before falling back.
+    // orphaned `uploading` row before falling back. The refs mirror these so a
+    // user-initiated cancel can abort the stream and reap the same row.
+    const controller = new AbortController();
+    abortControllerRef.current = controller;
+    pendingDatasetIdRef.current = null;
     let pendingDatasetId: string | undefined;
     try {
       const name =
@@ -442,9 +452,12 @@ export function UploadCSVForm({
         filename: rawFile.name,
       });
       pendingDatasetId = datasetId;
-      await putFileToPresignedUrl(uploadUrl, rawFile);
+      pendingDatasetIdRef.current = datasetId;
+      await putFileToPresignedUrl(uploadUrl, rawFile, controller.signal);
       await finalizeDirectUpload({ projectId, datasetId });
 
+      abortControllerRef.current = null;
+      pendingDatasetIdRef.current = null;
       if (onDirectUploadComplete) {
         // Hand off to the drawer host, which polls status and shows the
         // processing → ready/failed flow in place (no navigation).
@@ -460,6 +473,11 @@ export function UploadCSVForm({
         void router.push(`/${project.slug}/datasets/${datasetId}`);
       }
     } catch (error) {
+      // A user-initiated cancel aborts the PUT: handleCancelUpload already reset
+      // state and reaped the pending row, so there is nothing more to do here.
+      if (error instanceof Error && error.name === "AbortError") {
+        return;
+      }
       // Any failure AFTER requestDirectUpload minted the `uploading` row leaves
       // it behind and locks the slug — whether we fall back (a fresh dataset is
       // created) or surface the error (the upload is abandoned). Reap it either
@@ -501,6 +519,26 @@ export function UploadCSVForm({
       );
       setIsUploading(false);
     }
+  };
+
+  /**
+   * Cancel an in-flight direct upload: abort the streaming PUT and reap the
+   * pending `uploading` row so its slug isn't locked on retry. The file stays
+   * selected so the user can re-upload or remove it.
+   */
+  const handleCancelUpload = () => {
+    abortControllerRef.current?.abort();
+    abortControllerRef.current = null;
+    const pendingId = pendingDatasetIdRef.current;
+    pendingDatasetIdRef.current = null;
+    if (pendingId && projectId) {
+      void abortPendingUpload({ projectId, datasetId: pendingId }).catch(
+        (error) => {
+          logger.error({ error }, "Failed to clean up cancelled upload");
+        },
+      );
+    }
+    setIsUploading(false);
   };
 
   /**
@@ -575,6 +613,7 @@ export function UploadCSVForm({
         parse={!enableDirectUpload}
         uploadStatus={isUploading && enableDirectUpload ? "uploading" : undefined}
         fileError={fileError}
+        onCancel={handleCancelUpload}
         onUploadAccepted={handleUploadAccepted}
         onRawFile={(file) => {
           setRawFile(file);
@@ -605,6 +644,7 @@ export function UploadCSVForm({
         <Button
           colorPalette="blue"
           loading={isUploading}
+          loadingText="Uploading"
           disabled={!!disabled || isUploading || !canUpload}
           onClick={() => void handleUpload()}
         >
@@ -622,6 +662,7 @@ export function CSVReaderComponent({
   parse = true,
   uploadStatus,
   fileError,
+  onCancel,
   children,
 }: {
   onUploadAccepted: (results: {
@@ -647,6 +688,8 @@ export function CSVReaderComponent({
   uploadStatus?: "uploading";
   /** File-level validation/upload message; renders the file row in an error state. */
   fileError?: string;
+  /** Cancels an in-flight upload (no-parse path). Shows the X cancel control. */
+  onCancel?: () => void;
   children?: (hasAcceptedFile: boolean) => React.ReactNode;
 }) {
   if (!parse) {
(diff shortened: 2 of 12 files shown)


sergioestebance:
## 🧪 prove-them-wrong — real-browser verification (re-run on `4f680da74`)

_Playwright · Chromium · `localhost:5560` (vite serving this worktree) · test account · 2026-06-23._
_This replaces the earlier run on this comment. Every claim below was re-proven from zero after the Aryan-review fixes; the in-drawer processing → ready transition is DB-verified._

### Claim 1+2 — Upload CSV is a right-side **drawer** (not a centered modal) with an inviting empty dropzone
drawer + empty dropzone
Right-anchored drawer, left nav still visible. Cloud-upload illustration, "Drag and drop file, or **click to browse**", "Supported files: CSV, JSON, or JSONL".

### Claim 3 — Chosen file → status row (name + size + remove); removing returns the empty state
file status row
remove returns empty

### Claim 4 — Upload processes **in the drawer** (no navigation away) → ready
in-drawer processing
ready in drawer, no nav
URL stayed on `/datasets` the entire time (never `/datasets/{id}`); spinner + "Preparing your dataset…" → green "Ready" + "View dataset", staying in the drawer until the user chooses to view. **DB-verified:** `status=ready`, `contentLayout=s3_jsonl`, `columnTypes=[id_, question, answer]`.

### Claim 5 — System errors render as a **top alert** in the drawer
system error top alert
Forced a `500` on `/api/dataset/direct-upload`; the error surfaces as an `Alert` pinned at the top of the drawer body (`data-testid="upload-error"`, y≈89px), with the file row below it unaffected — confirming system-error / file-error routing stays separate.

### Verdict table
| 1 | Upload CSV is a right-side drawer, not a centered modal | 01-drawer-not-modal | ✅ PASS |
| 2 | Empty dropzone invites a file (illustration + copy + supported types) | 01-drawer-not-modal | ✅ PASS |
| 3 | Chosen file → status row (name+size+remove); remove returns empty | 02-file-status-row, 03-removed | ✅ PASS |
| 4 | Upload processes in-drawer (no nav away) → ready; `s3_jsonl` in DB | 04-processing, 05-ready + DB query | ✅ PASS |
| 5 | System errors render as a top alert in the drawer | 06-system-error-top-alert | ✅ PASS |

### Findings
- **Reap-on-failure verified (tried to break it, couldn't).** Forced the cross-origin S3 PUT to `400` → drawer falls back AND fires the cleanup `DELETE /api/dataset/direct-upload/{id}` (→ `200`), soft-deleting the pending `uploading` row (`archivedAt` set). `findNextName` builds its slug set from **all** rows incl. archived (`findAllSlugs` filters only by `projectId`), so a same-name retry gets a bumped slug and isn't blocked. No orphan row, no slug dead-end — confirms the hardening commits.
- **Environmental, not the PR (resolved during the run):** the dev S3 SSO token had expired overnight → `ExpiredToken` on the presigned PUT, which (correctly) routed the upload to the in-browser-parse fallback. Refreshing creds + restarting the dev server fixed it; the PR touches no S3 presigning. Flagged only to explain why a first pass showed the fallback path.
- **Pre-existing, unrelated:** `<a> cannot be a descendant of <a>` hydration warning on the **home dashboard** (`QuickAccessLinks` → docs link), not the drawer. Out of scope here; worth a separate cleanup.
- Console during the drawer flow is otherwise clean — remaining entries are pre-auth `401/UNAUTHORIZED` and tRPC dev-log banners.

**Verdict: 5/5 proven, 0 disproven, 0 unverified.** Drawer conversion, in-drawer processing, and error routing all hold up under adversarial testing.

Aryansharma28:
## Code Review

### Blockers

**`UploadCSVDrawer.tsx:123` — `handleClose` doesn't abort the in-flight upload**
When the user clicks X while a file is streaming, `handleClose` only calls `setProcessingDatasetId(null)` and `onClose()`. `abortControllerRef` and `pendingDatasetIdRef` live in `UploadCSVForm` and are never touched, so the PUT continues, `finalizeDirectUpload` fires, and a dataset is silently created. The slug is also locked, so the next retry 409s or creates a duplicate. Fix: expose a cancel callback from `UploadCSVForm` and call it from `handleClose`.

**`UploadCSVDrawer.tsx:99` — stale `uploadError` / `sizeError` / `rawFile` persist across close-reopen**
The `useEffect` on `[isOpen]` (line 128) resets `uploadedDataset` and `processingDatasetId` but not the three state values in `UploadCSVForm`. Because `Drawer.Root` has no `unmountOnExit`, the form stays mounted; a user who gets an upload error, closes, and reopens sees the stale error and the old file pre-selected. Fix: reset those three in the `!isOpen` branch, or add `unmountOnExit`.

**`UploadCSVDrawer.tsx:1361` — `CancelUploadButton` always clears the file, contradicting the docstring**
`handleCancelUpload`'s docstring says "the file stays selected so the user can re-upload". But `RawFileDropzone`'s cancel handler calls `onCancel()` then `setFile(null)`, which fires `onRawFile(null)` → `rawFile = null` and `onUploadRemoved` → `uploadedDataset = undefined`; the dropzone re-expands and `canUpload` flips to false. User must re-browse to retry instead of re-clicking Upload. The `RawFileDropzone` comment says this is intentional — if so, remove the contradicting docstring; if not, drop the `setFile(null)` call.

---

### Non-blockers

**`UploadCSVDrawer.tsx:263` — `onReady` stale-closure risk (deps suppressed)**
`DatasetUploadProcessing`'s effect fires `onReady(data)` with `[isReady]` as the only dep, suppressing `data` and `onReady`. If two polling responses arrive in quick succession (first flips `isReady=true` with `columnTypes=A`, second updates to `columnTypes=B` before the microtask runs), the effect delivers the stale schema to the host — downstream column-aware callers get columns that don't exist. Fix: include `data` in deps, or store `onReady` in a ref.

**`UploadCSVDrawer.tsx:202` — zero-column CSV causes infinite polling**
`datasetHasColumns` returns `false` for `columnTypes = []`. A degenerate CSV with a blank header normalizes to an empty column list; `status` becomes "ready" but `datasetHasColumns` is false, so `refetchInterval` returns 3000 forever — the drawer spinner never clears. Fix: treat "ready + no columns" as a failed state on the server (or add a client-side timeout).

**`UploadCSVDrawer.tsx:76` — `MAX_ROWS_LIMIT` redeclared instead of imported**
`MAX_FILE_SIZE_BYTES` is already imported from `../../server/datasets/upload-utils`. `MAX_ROWS_LIMIT = 10_000` is defined in the same file but redeclared here. If the server-side limit changes only one copy updates. Add it to the existing import.

**`UploadCSVDrawer.tsx:479` — `setIsUploading(false)` missing in the direct-upload `else` branch**
When `enableDirectUpload=true` and `onDirectUploadComplete` is not provided (the `else` branch): upload succeeds, toaster fires, `onClose()` + `router.push()` run, but `setIsUploading(false)` is never called. No current caller exercises this path, but if it's ever reached the form re-opens with the Upload button permanently disabled. One-liner fix.

**`UploadCSVDrawer.tsx:427` — `proposeValidName` fires a DB round-trip on every file drop**
`handleUploadAccepted` calls `trpcUtils.dataset.findNextName.fetch(...)` immediately on drop, before the user clicks Upload. Re-dropping the same file (after a validation error) fires the query again. The result is already held in `uploadedDataset.name` by the time `handleUpload` runs, making the eager call redundant on re-drops. Deferring to upload-time eliminates the extra round-trips.

sergioestebance:
Thanks @Aryansharma28 — solid review, mostly landed. Fixed in `4f680da74`:

- **B1 (abort on close)** ✅ — closing the drawer mid-upload now aborts the in-flight PUT via a form-unmount cleanup (`abortControllerRef.current?.abort()`); the existing AbortError catch then reaps the pending row, so close no longer silently finalizes a dataset. No-op on a normal close / the processing handoff (refs are nulled the instant the PUT succeeds, before finalize).
- **B3 (stale docstring)** ✅ — the cancel-clears-file behavior is intentional (the dropzone re-expands on cancel, per the UX); updated the `handleCancelUpload` docstring to match.
- **NB2 (zero-column infinite poll)** ✅ — dropped the columns-gate. `finalizeUpload` always sets `status: "processing"` before normalize, so `ready` is only ever reached *after* normalize finished — we now trust it (same contract as the dataset page) and stop polling at ready/failed/gone. A degenerate columnless dataset is terminally ready, not an endless spinner.
- **NB3 (redeclared limit)** ✅ — now imports `MAX_ROWS_LIMIT` from `server/datasets/upload-utils` (single source of truth).
- **NB4 (missing setIsUploading)** ✅ — added to the navigate-away `else` branch.

Two I'd push back on:

- **B2 (stale state on close→reopen)** — couldn't reproduce; browser-disproved it (select file → close → reopen → empty, fresh dropzone). Chakra v3 unmounts the drawer content on close, so `UploadCSVForm` remounts with fresh state. The premise that "the form stays mounted" doesn't hold for this version, so `rawFile`/`uploadError`/`sizeError` don't persist.
- **NB1 (onReady deps)** — the staleness can't manifest: columns are immutable post-`ready` and polling stops at `ready`, so no second response arrives. The suggested fix (add `data` to deps) would fire `onReady` on every change and **break the fire-once contract**, so I left it.

NB5 (eager `proposeValidName` on drop) is a fair pre-existing perf nit — deferring it as out-of-scope for this PR.

sergioestebance:
Updated

0xdeafcafe:
Nice work. The cancel-race comments at 386-392 and 486-518 are genuinely useful for anyone reading this in six months. No blockers, but a handful of P2s worth catching:

- **`onReady` prop typed as `unknown` plus a cast at the call site** (`UploadCSVDrawer.tsx:199-203`). The Prisma shape is known - either thread `DatasetColumns | null` through or use `Prisma.JsonValue`, and drop the cast at line 142.
- **`console.error("error", error)` survived the rename** (line 822). The rest of the file uses `logger.error`. Easy fix on the way through.
- **`<Heading>Upload CSV</Heading>` instead of `<Drawer.Title>`** (line 131). Drops the `aria-labelledby` wiring Ark gives you for free.
- **`useQuery` mock ignores its args and the `refetchInterval` callback** (`UploadCSVDrawer.unit.test.tsx:24-27, 43-50`). If anyone drops the `enabled` guard or the polling-stops-at-`ready` predicate regresses, nothing catches it. Capturing the args and asserting `refetchInterval(data)` returns `false` at terminal states would close that gap.

P3 follow-ups (not blocking):

- `CSVReaderBox` and `RawFileDropzone` are ~60 lines of near-identical dropzone surface. A `DropzoneShell` extraction would stop them drifting.
- `dataset-upload-dropzone.feature` only covers the leaf scenarios. The interesting new behaviours (cancel before presign, cancel mid-PUT, retry-from-failed) are pinned by tests but not by spec.
- `abortPendingUpload` is now called from three spots - a silent 5xx leaves an orphan `uploading` row forever. A `logger.warn` on `!response.ok` would give a breadcrumb.

LGTM after those.

Issue #4450, opened 2026-05-31
Integration test `api-keys-scope-filter` wedges shard 4 with an infinite render loop; masked green by globalSetup hard-floor process.exit(0)


sergioestebance:
## Summary

The integration test `langwatch/src/pages/settings/api-keys/__tests__/api-keys-scope-filter.integration.test.tsx` contains a mock bug that triggers an **infinite React re-render loop**. The loop starves the event loop (so per-test timeouts never fire) and the worker grows its heap until it OOMs.

This is **live on `main` today**. It is invisible only because a CI safety hatch (`globalSetup.ts`'s 20-minute hard-floor `process.exit(0)`) swallows the wedge and reports the shard **green**. The practical consequences on `main`:

- The 7 tests in this file **never actually run** — they're silently skipped (zero real coverage of the API-keys scope filter).
- Every integration run **burns ~20 minutes** on the wedged shard (currently shard 4 of 6) before the hard floor releases it.
- A pre-hard-floor CI run wedged for **3h39m and OOM'd** the worker.

There is also a **mis-attribution** to fix: `globalSetup.ts` blames "vitest's own shard/reporter finalize path" for the shard-4 wedge. That is wrong — the wedge is this test's render loop.

## Where it came from

Introduced by #4282 (`feat(api-keys): scope filter + ui changes in code snippets`, merged 2026-05-29). The component and hooks are correct; **only the test mock is buggy**.

## Root cause

The test mocks `useOrganizationTeamProject` so that it returns a **fresh `organization` object literal on every call**:

```ts
vi.mock("~/hooks/useOrganizationTeamProject", () => ({
  useOrganizationTeamProject: () => ({
    project: { id: "proj-1", ... },
    organization: { id: "org-1", name: "Acme Corp", teams: [ ... ] }, // NEW object every render
    team: { id: "team-1", ... },
    hasPermission: () => true,
  }),
}));
```

That new identity cascades into an infinite loop through otherwise-correct production code:

1. `ApiKeysSection` calls `useAvailableScopes(organization)`, which is `useMemo(..., [organization])`. A new `organization` ref each render → the memo recomputes → `filterAvailable` gets a **new identity every render**.
2. `useUrlScopeFilter` runs an effect with deps `[router.query.scope, filterAvailable]`. `filterAvailable` changed → the effect runs → it calls `setScopeFilter({ kind: "all" })` (a fresh object).
[...]

PR #5082, opened 2026-06-23
fix(traces): resolve full blob values on all content-consuming read paths (2 of 2 — #4888)

### LINKED ISSUE #4991: fix(traces): resolve full blob values on all content-consuming read paths (2 of 2 — #4888)
**Follow-up to PR #4984 / issue #4888 (1 of 2).** The >64KB span-IO offload (#4215) stores a 64KB **preview** + an `eventref` pointer in the lean projection; the **full value** lives in ClickHouse `event_log`. PR #4984 wired blob-resolution deps into the four **single-trace detail** reads only. Several **bulk content-consuming** reads still return the truncated 64KB preview.

**Product intent:** truncation is purely a UI preview affordance — only the **list/grid snippet columns** should ever be truncated. Every other consumer that reads trace *content* should get the whole value.

## Summary

Resolve full blob values on all read paths whose consumer needs trace content; keep preview only for the list/search grid + aggregations.

### Still-truncated consumer paths to fix (resolve full)

| Path | Site | Why truncation is wrong here |
|---|---|---|
| Export / download | `export.service.ts:62`, `traces.ts:469` (`getAllForDownload`) | Truncated values = **data loss** in CSV/JSONL export |
| Thread detail | `traces-legacy.ts:324` (`GET /api/thread/:id`), `traces.ts:227`, `traces.ts:310` (`getTracesWithSpansByThreadIds`) | Conversation messages cut off |
| Annotation queue | `annotation.ts:60`, `annotation.ts:430` | Annotators can't see the content they're labeling |
| Dataset / sample builders | `traces.ts:332` (`getSampleTracesDataset`), `traces.ts:393` (`getSampleTraces`) | Truncated rows corrupt the dataset |

### Must STAY preview-only (do NOT resolve)

Binding constraint — #4888 AC2 / ADR-022, enforced at `clickhouse-trace.service.ts:2314`:

- Trace **list/search grid**: `getAllTracesForProject` for the list table (`traces.ts:43`, `app.v1.ts:176`, `traces-legacy.ts:252`)
- Aggregations / digests: `getTopicCounts`, `getCustomersAndLabels`, `getDistinctFieldNames`, `getFormattedSpansDigest`

These show snippets across many rows; preview is the whole point, and resolving would reintroduce the **N×M `event_log` heavy-read** the offload was built to prevent.

### Key design question (confirm before fanning out)

Export and thread paths can return large result sets. Full resolution there must be **batched / streamed** — not one SELECT per span × N rows, and not refusing to resolve. The implementer should confirm the batch-resolution shape first.

## Acceptance Criteria

- **AC1:** Export (CSV/JSONL) of an offloaded trace contains the full IO value, not the 64KB preview.
- **AC2:** Thread-detail reads (`GET /api/thread/:id`, thread tRPC paths) return full IO values.
- **AC3:** Annotation-queue trace reads return full IO values.
- **AC4:** Dataset / sample builders persist full IO values.
- **AC5:** Trace list/search grid and aggregation paths still issue **ZERO `event_log` SELECTs** (heavy-read protection preserved — assert no `event_log` query on the list hot path).
- **AC6:** Large-result-set resolution (export/thread) is **batched/streamed**, not N×M unbounded; include a test or benchmark bounding the CH query count.
- **AC7:** Resolution failure degrades to preview **WITH a warn log** (no silent truncation).

## Related

- 1 of 2: #4984 (detail reads — single-trace)
- Parent: #4888
- Offload origin: #4215 / ADR-022

---
*Classification: Bug*

drewdrewthis:
## Why

Closes langwatch/langwatch#4991. The >64 KB span-IO offload (#4215 / ADR-022) stores a **64 KB preview** + an `eventref` pointer in the lean projection; the **full value** lives in ClickHouse `event_log`. PR langwatch/langwatch#4984 ("1 of 2") wired blob-resolution into the four **single-trace detail** reads. Several **bulk content-consuming** reads still returned the truncated preview — so a full CSV/JSONL export lost data, thread conversations were cut off at 64 KB, annotators couldn't see the content they were labeling, and dataset/sample builders persisted truncated rows. This PR ("2 of 2") resolves the full value on every read whose consumer needs trace **content**, while keeping the **list/search grid + aggregations** on the preview (the heavy-read protection the offload exists to provide).

## What changed

- **A bulk resolver (`resolveOffloadedTracesBatch`) that streams, not fans out** — resolving each trace independently would fire an unbounded `N×M` burst of `event_log` SELECTs (one per offloaded field × every row) and exhaust the CH connection pool on a large export. Instead it decodes every eventref across the whole result set, **dedupes** identical `(aggregateId, eventId, field)` refs, and runs the reads through a **bounded-concurrency pool** so peak in-flight CH reads is a constant regardless of result-set size (**AC6**). A missing/failed row keeps that field's preview and logs a `warn` — no silent truncation (**AC7**).
- **Explicit `resolveBlobs` opt-in, deliberately separate from `includeSpans`** — the download path and the list grid share `getAllTracesForProject`, and `app.v1`'s public list passes `includeSpans=true` but **must stay preview**. Gating on a dedicated `resolveBlobs` option (not `includeSpans`) lets every content-consuming read resolve full while the list/aggregations issue **zero** `event_log` reads (**AC5** / ADR-022 binding constraint). *`includeSpans` is not a proxy for "consumes content" — see the review round below, where treating it as one was the bug.*
- **Mirror langwatch/langwatch#4984's DI shape at every bulk call site** — export, thread (legacy `GET /api/thread/:id` + tRPC), annotation queue (both read sites), and dataset/sample builders now build `buildTraceBlobResolutionDeps()` and request `full: true` (**AC1–AC4**). No new resolution mechanism invented — the same `BlobStore`/`event_log` recompute, batched.
- **Shared eventref decoder** — `hasEventRefs` / `parseSpanEventRefs` extracted into `offloaded-eventref-parsing.ts` so the per-trace (#4984) and batch (this PR) resolvers decode the reserved-namespace shape from one definition.
- **Authorize before resolving, on the one anonymous path** — `getTracesByThreadId` is a `publicProcedure`. Full resolution now runs **after** the public-share filter narrows the thread, never before (**security** — see review round).

## How it works

```
src/server/traces/
  offloaded-eventref-parsing.ts      NEW  decode reserved eventref keys (shared by both resolvers)
  resolve-offloaded-traces.ts        REFACTOR  per-trace detail resolver → now uses shared decoder
  resolve-offloaded-traces-batch.ts  NEW  whole-result-set resolver: dedupe + bounded-concurrency stream + per-ref warn-degrade
  trace.service.ts          builds BOTH resolver callbacks from one deps object; forwards full→resolveBlobs
  clickhouse-trace.service.ts  bulk reads use the batch callback (fallback to per-trace); enrich + getTracesByThreadId honor resolveBlobs
  types.ts                  GetAllTracesForProjectOptions.resolveBlobs (export discriminator)

Read paths flipped to full resolution:
  export/export.service.ts            AC1  ExportService.create deps + EVERY export mode sets resolveBlobs
  clickhouse-trace.service.ts         AC1  spans are fetched when the caller wants spans OR full IO — a
                                           summary read resolves, recomputes trace IO, then drops the spans
  api/routers/traces.ts               AC1/AC2/AC4  getAllForDownload, thread tRPC (+ public-share ordering), sample/dataset builders
  api/routers/annotation.ts           AC3  both queue-read sites
  routes/traces-legacy.ts             AC2  GET /api/thread/:id
```

Flow: a bulk read opts in (`{ full: true }` / `resolveBlobs: true`) → `TraceService` forwards the gate → `ClickHouseTraceService` collects every trace's normalized spans → `resolveOffloadedTracesBatch` resolves the whole set in one bounded pass → spans get full values, reserved keys stripped, trace IO recomputed. List/aggregation reads never opt in, so the resolver never runs and zero `event_log` reads are issued.

## Review round — what the reviewers caught

Two **P1 blockers** and two P2s were raised on this PR, and working them surfaced four more defects. Worth reading before re-reviewing, because three of them were in code this PR itself added.

| # | Defect | Why it mattered |
|---|---|---|
| **P1** | `getTracesByThreadId` (a `publicProcedure`) resolved `{full:true}` for the **whole thread**, then filtered the response down to the publicly-shared traces | The filter was post-hoc. One valid share link let an **anonymous** caller force full `event_log` de-offloading of every *other* trace in that thread — the CH connection-pool amplification ADR-022 exists to prevent, reachable unauthenticated. Not a data leak (the response stayed filtered). Now: preview-only read → authorize → resolve **only** the authorized ids. |
| **P1** | `resolveBlobs` was gated on `includeSpans`, so **summary exports** silently shipped the truncated 64 KB preview | Summary mode reads no *span* content but still emits trace-level `trace.input`/`trace.output` (`csv-serializer.ts:91-92`). For an offloaded trace it served the preview with no error and no indication data was cut — the same data-loss bug this PR opens with, on the other export mode. **This one took two attempts — see below.** |
| P2 | `fetchKeyOf(aggregateId, eventId, field)` — three positional strings | A transposition compiled clean and would silently fetch the wrong `event_log` row onto the wrong span. Now named params. |
| P2 | `resolutions[i]!` trusted an **injected** resolver to return one resolution per trace, in order | The type can't enforce it. Now guarded at the boundary. |
| +1 | `getAllForDownload` **had the same summary bug**, unfixed | Same `resolveBlobs: input.includeSpans` gate, twin path, introduced by this PR. A spans-less download truncated offloaded traces. |
| +1 | The new guard **overclaimed** — it promised "in input order" but only compared lengths | A resolver returning the right *count* in the wrong *order* passed silently and scattered each trace's IO onto its neighbour: exactly the silent-corruption case it advertised catching. Order is now actually verified. |
| +1 | A third flattening `catch` (`getTracesWithSpansByThreadIds`) destroyed the guard's error | It sits *above* `getTracesWithSpans` and re-wrapped what that method re-threw, on a live `resolveBlobs` path. |
| +1 | A unit test in this PR **had been failing since it was added** | `test-unit (2)` reported `success` anyway — the shard's `globalSetup` hard-floor `process.exit(0)` fires mid-run and GitHub scores it green. Root cause was a tRPC mutation's auditLog middleware reaching the real `prisma` singleton, which also leaked a socket that plausibly *caused* the wedge. Fixed here; the masking itself is [#4450](https://github.com/langwatch/langwatch/issues/4450). |

**Reviewer note — one deliberate divergence:** the cardinality guard was asked for at both index sites (`resolutions[i]!` and `merged[i]!`). It is implemented **once**, at the boundary where the untrusted value enters. `merged` comes from `entries.map(...)`, so `merged.length === enrichable.length` is a tautology and a second assert there would be unreachable. Pushback welcome.

### The summary-export fix took two attempts — read this one

The first attempt set `resolveBlobs: true` at the call sites and **did nothing at runtime.** Resolution lives inside `enrichTracesWithSpans`, which only ran `if (options.includeSpans)`. Summary mode sets `includeSpans = false`, so `resolveBlobs` was **never read** — zero `event_log` reads, preview still shipped, and a comment now asserting it was fixed.

The tests were green throughout, and that is the part worth internalising: they mocked `getAllTracesForProject` and asserted the flag was **forwarded**. *That assertion passes whether or not the flag has any runtime effect.* It tested intent, not behaviour.

**Actually fixed in `6860180b9`:** spans are fetched when the caller wants spans **OR** wants full IO, because those are not the same thing — the full value is recoverable *only* by de-offloading the spans' eventref pointers and recomputing trace IO from them (`trace_summaries` holds nothing but the preview). A summary read therefore resolves, recomputes, and then **discards** the spans (returns `spans: []`), so its payload shape is unchanged.

Proven behaviourally this time — the real service + real resolver, asserting the **value**, confirmed RED first:

```
RED   × resolves the trace-level output to the FULL value, not the preview
        AssertionError: expected [getFromEventLog] to be called at least once
GREEN ✓ 12 passed — full value returned, AC5 zero-read still holds
```

**AC5 is preserved by construction:** `resolveBlobs` stays opt-in, so the list grid and aggregations resolve nothing and issue zero `event_log` reads whether or not they ask for spans.

**Cost, stated plainly — worth a conscious human call:** summary exports now do a span join + bounded `event_log` reads where they previously did neither. That is the tradeoff the reviewer explicitly offered on the blocker thread, and there is no cheaper way to recover the value. But it is a real new cost on a path that had none, and it should be accepted deliberately rather than inherited from a comment.

## Human verification (for if you really don't trust me)

1. Ingest a trace whose input or output exceeds 64 KB with the `release_trace_blob_offload` flag on (offloads to `event_log`, leaves a 64 KB preview in `trace_summaries`).
2. **Export** that trace as a *full* CSV/JSONL → the IO column holds the **whole** value, not a 64 KB-truncated string.
3. Open the **thread** the trace belongs to (`GET /api/thread/:id` or the thread drawer) → the conversation message is complete, not cut off.
4. Add the trace to an **annotation queue** → the annotator sees the full content.
5. Build a **dataset/sample** from it → the persisted row is the full value.
6. Load the trace **list/search grid** and the topic/customer aggregations → still fast, and (verify via CH query log) they issue **no** `event_log` SELECTs.

**Outside-In TDD, real ClickHouse service + real resolver + real mapper (only the raw CH SQL boundary + S3 mocked).** Every AC has a test that genuinely exercises the behavior and was confirmed RED before the fix.

**RED — wiring reverted (`git stash` of the implementation), new tests run:** 15 assertions fail (export 3/3, thread, sample, annotation 2/2, download), while the pure-mechanism batch test and the "stays-preview / zero-read" assertions stay green (they hold trivially when nothing resolves):

```
❯ export-service.4991.unit.test.ts            (3 tests | 3 failed)
❯ clickhouse-trace.service-4991-bulk.unit.test.ts (6 tests | 2 failed)   ← the two "reads full" cases
❯ traces-legacy-thread.unit.test.ts           (3 tests | 2 failed)
❯ traces.4991-full-resolution.unit.test.ts    (9 tests | 6 failed)
❯ annotation.4991-full-resolution.unit.test.ts (2 tests | 2 failed)
   Failed Tests 15
```

**GREEN — implementation restored, full suite (6 new + 6 merged/existing regression). Counts below are from the original round; the current head's numbers are in the CI run linked on this PR (both `pnpm typecheck` *and* `pnpm typecheck:tests` — the former excludes test files):**

```
$ pnpm typecheck            # tsgo --noEmit → clean, 0 errors
$ npx vitest run <12 files>
 Test Files  12 passed (12)
      Tests  150 passed (150)
```

| AC | What the test proves | File |
|---|---|---|
| AC1 export | `getAllTracesForProject({includeSpans, resolveBlobs:true})` widens output past 64 KB to the full value; `ExportService` wires deps + sets `resolveBlobs` on **every** export mode (summary included — it emits trace-level IO), and `getAllForDownload` does too even without spans | `clickhouse-trace.service-4991-bulk`, `export-service.4991`, `traces.4991` |
| AC2 thread | `getTracesByThreadId(resolveBlobs:true)` returns full IO; legacy + tRPC handlers build deps + `full:true` | `clickhouse-trace.service-4991-bulk`, `traces-legacy-thread`, `traces.4991` |
| AC3 annotation | both queue-read sites build deps + `full:true` | `annotation.4991` |
| AC4 dataset/sample | sample/dataset builders resolve spans full (ID-only list read stays preview) | `traces.4991` |
| AC5 zero-read | list grid + `getFormattedSpansDigest` issue **zero** `getFromEventLog` calls and keep the preview, even with deps wired | `clickhouse-trace.service-4991-bulk`, `traces.4991` |
| AC6 batched/streamed | peak concurrent `event_log` reads ≤ pool size for a 75-trace set; identical refs deduped to one read; one read per offloaded field (no `N×M`) | `resolve-offloaded-traces-batch` |
| AC7 degrade+warn | a missing `event_log` row keeps the preview, logs a warn, strips the reserved key, and never fails the rest of the batch | `resolve-offloaded-traces-batch` |

**✅ Live-app visual proof — langwatch/langwatch#5082-NEW surface: dataset builder (`getSampleTracesDataset`, AC4; deliberately **not** the langwatch/langwatch#4888 trace-detail drawer).**

Dataset-builder mapping preview renders the full >64KB offloaded trace IO

Seeded a trace `blob-5082-fullio` whose input **and** output each exceed 64 KB, so the offload fired: the lean projection (`stored_spans`) holds a **65 539-byte preview ending in `…`** plus `langwatch.reserved.eventref.langwatch.{input,output}`, while the full ~70 KB value lives in ClickHouse `event_log`. A unique end-marker `BLOB5082FULLIOEND` was placed **past byte 65 536**, so it exists only in the full value:

| `langwatch.output` | preview (`stored_spans`) | full (`event_log`) |
|---|---|---|
| byte length | 65 539 (64 KB + `…`) | ~70 104 |
| `position('BLOB5082FULLIOEND')` | **0 — absent** | **140 613 — present** |

The screenshot drives the **real authenticated app over live HTTPS** (`drewdrewthis--langwatch-golden.boxd.sh`): *Add Automation → Add to Dataset → mapping preview*, which calls `getSampleTracesDataset` — the exact AC4 read path this PR flips to `resolveBlobs: true`. The output-cell editor holds the **full 70 104-byte value ending in `BLOB5082FULLIOEND"}]}`** (close-up, lower panel) — not the truncated preview. Without this PR the same surface returns the 64 KB preview ending in `…` with no marker; the list-grid cell still shows the capped preview (AC5 stays preview), confirming only the content-consuming read resolves.

**Evidence for the review-round fixes (above) — every one confirmed falsifiable by reverting the fix and watching the test fail:**

| Fix | Test | Revert → |
|---|---|---|
| Authorize before resolving (public share) | `traces.4991` → *public-share thread read* | Restoring resolve-then-filter fails the `{full:false}` assert **and both** exclusion asserts (`getTracesWithSpans` called with only the authorized ids, exactly once). Simulating a *partial*-authorization regression — resolve every thread id, ignore the filter — also fails both. |
| Summary export resolves — **the value, not the flag** | `clickhouse-trace.service-4991-bulk` → *AC1 summary read* | Confirmed RED before the fix: the real service with `{includeSpans:false, resolveBlobs:true}` never called `getFromEventLog` and returned the preview. The call-site tests (`export-service.4991`, `traces.4991`) assert the flag is forwarded — necessary, but they pass even when the flag is inert, which is exactly how the first attempt slipped through. |
| Spans-less download resolves | `traces.4991` | Restoring `resolveBlobs: input.includeSpans` fails it. |
| Resolver **order** contract | `clickhouse-trace.service-4991-bulk` | Removing the order check (keeping cardinality) fails the misalignment test — a resolver returning 2 resolutions for 2 traces, reversed, drives the real service. |

### The summary-export fix is proven end-to-end against a REAL ClickHouse

The unit tests **could not have caught the bug they were written for** — they mocked `getAllTracesForProject` and asserted the flag was *forwarded*, which passes whether or not the flag does anything. So the P1 fix gets a real one: `src/server/export/__tests__/export-summary-blob-resolution.integration.test.ts`.

It seeds a genuinely offloaded trace the way an ingest leaves one, then runs the **real** export:

```
real ClickHouse (trace_summaries + stored_spans + event_log)
  → real ClickHouseTraceService      ← the includeSpans/resolveBlobs gate that was the bug
  → real resolveOffloadedTracesBatch + real BlobStore   ← the actual event_log read
  → real ExportService, SUMMARY mode
  → real CSV / JSONL serializers
  → assert the exported BYTES carry the de-offloaded value
```

The fixture's `UNIQUE_TAIL` sits **past the 64 KB preview boundary by construction** (the preview is `value.slice(0, 64KB) + "…"`), so it exists *only* in the de-offloaded value — a preview-only export **cannot** contain it. That makes the assertion falsifiable against the real bug rather than against a mock's arguments.

**Verified falsifiable — I restored the pre-fix `if (options.includeSpans)` gate and re-ran it:**

```
RED (pre-fix gate restored)
  × writes the FULL de-offloaded value into the CSV, not the 64 KB preview
      AssertionError: expected 'trace_id,timestamp,input,output,label…'
                      to contain '__OFFLOAD_FULL_VALUE_TAIL_MARKER__'
  × does not truncate at the preview boundary
      AssertionError: expected '…' not to contain 'xxxxxxxxxxxxxxxxxxxx…'   ← the truncated preview, served
  × writes the FULL value in JSON (JSONL) summary mode too
  ✓ AC5: list read keeps the preview          ← still passes, as it must
  Tests  3 failed | 1 passed (4)

GREEN (fix in place)
  Tests  4 passed (4)
```

**AC5 is guarded in the same file, not just asserted in prose:** the list/search read over the *same* offloaded trace, without `resolveBlobs`, still comes back as the preview with the tail absent. So the fix demonstrably did not disarm the heavy-read protection.

**What is NOT demonstrated in a running app, and why — stated plainly rather than implied:**

| Fix | Observable surface? | Evidence |
|---|---|---|
| Summary-export data loss | **Yes** — the exported bytes change | The real-ClickHouse end-to-end export above, RED→GREEN |
| Dataset builder (AC4) | **Yes** — the mapping preview | The live-app screenshot below |
| Public-share authorize-before-resolve | **No.** The response is byte-identical either way; only *when* resolution happens changes (and *how much* `event_log` it touches). There is nothing for a user to see | Falsifiable unit tests, incl. a simulated partial-authorization regression that fails both exclusion asserts |
| Resolver-contract guard | **No.** It fires only on a resolver that breaks its contract — unreachable through the UI | Tests driving genuinely non-conforming resolvers through the real service |

The visual proof below remains the AC4 dataset-builder surface.

## Anything surprising?

- **AC6 is satisfied by *streaming* (bounded concurrency + dedup), not by collapsing reads into one `EventId IN (…)` query.** True multi-event batching would require a new `getManyFromEventLog` primitive, which would break langwatch/langwatch#4984's merged unit tests (they spy `getFromEventLog` and assert call counts). Keeping the existing primitive preserves those tests verbatim; the bound that matters — peak in-flight reads is constant regardless of result size — is enforced and tested. A follow-up could add `EventId IN` batching if profiling shows the per-event round-trip is the bottleneck.
- The `release_trace_blob_offload` resolution only fires for traces ingested with offload on; pre-offload traces have no eventrefs and skip the resolver entirely (fast path).

## Summary by CodeRabbit

* **New Features**
  * Added opt-in `resolveBlobs` full-resolution for offloaded trace IO (>64KB), enabling full blob IO for download/export and thread detail views when spans are included.
* **Improvements**
  * Annotation, traces, and export paths now return fully resolved trace content when full mode is enabled; preview/list paths remain preview-only.
  * Trace span enrichment now uses batched offloaded resolution for faster, more consistent results.
* **Tests**
  * Expanded coverage for full vs preview wiring, legacy thread routing, export modes, eventref parsing, and bounded concurrent resolution.

### DIFF
--- a/langwatch/src/server/api/routers/traces.ts
+++ b/langwatch/src/server/api/routers/traces.ts
@@ -224,11 +224,17 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      const traceService = TraceService.create(ctx.prisma);
+      // Thread-detail read consumes conversation content — resolve full IO
+      // (#4991), not the 64 KB preview.
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       const tracesGrouped = await traceService.getTracesByThreadId(
         projectId,
         threadId,
         protections,
+        { full: true },
       );
 
       if (!ctx.publiclyShared) {
@@ -311,11 +317,16 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      const traceService = TraceService.create(ctx.prisma);
+      // Thread reads consume conversation content — resolve full IO (#4991).
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       return traceService.getTracesWithSpansByThreadIds(
         projectId,
         threadIds,
         protections,
+        { full: true },
       );
     }),
 
@@ -333,7 +344,13 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      const traceService = TraceService.create(ctx.prisma);
+      // Dataset builder persists trace content — resolve full IO (#4991) so
+      // truncated rows never corrupt the dataset. The ID-only list read above
+      // stays on the preview (it reads no content).
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       const { groups } = await traceService.getAllTracesForProject(
         {
           ...input,
@@ -356,6 +373,7 @@ export const tracesRouter = createTRPCRouter({
         traceIds,
         protections,
         { from: input.startDate, to: input.endDate },
+        { full: true },
       );
     }),
 
@@ -395,7 +413,12 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      const traceService = TraceService.create(ctx.prisma);
+      // Sample builder feeds dataset/evaluator content — resolve full IO
+      // (#4991). The ID-only list read stays on the preview.
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       const { groups } = await traceService.getAllTracesForProject(
         {
           ...input,
@@ -421,6 +444,7 @@ export const tracesRouter = createTRPCRouter({
         traceIds,
         protections,
         { from: input.startDate, to: input.endDate },
+        { full: true },
       );
 
       const passedPreconditions = traceWithSpans.filter((trace) => {
@@ -472,7 +496,14 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      const traceService = TraceService.create(ctx.prisma);
+      // Download consumes trace content — resolve full IO when spans are
+      // included so exported rows carry the whole value, not the 64 KB preview
+      // (#4991 AC1). resolveBlobs is gated on includeSpans (resolution runs
+      // during span enrichment).
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       return traceService.getAllTracesForProject(
         {
           ...input,
@@ -483,6 +514,7 @@ export const tracesRouter = createTRPCRouter({
         {
           downloadMode: true,
           includeSpans: input.includeSpans,
+          resolveBlobs: input.includeSpans,
           scrollId: input.scrollId,
         },
       );
--- a/langwatch/src/server/api/routers/traces.ts
+++ b/langwatch/src/server/api/routers/traces.ts
@@ -230,34 +230,71 @@ export const tracesRouter = createTRPCRouter({
         ctx.prisma,
         buildTraceBlobResolutionDeps(),
       );
-      const tracesGrouped = await traceService.getTracesByThreadId(
+
+      // SECURITY (#4991): full blob resolution de-offloads the whole IO value
+      // from event_log and must NEVER run pre-authorization. For an authorized
+      // (non-public) caller the whole thread is theirs to read, so resolve full
+      // up front. For a public-share caller — an anonymous, publicProcedure
+      // path — resolving the full thread first would (a) de-offload traces the
+      // caller is not authorized to see and (b) let a single valid share link
+      // amplify unbounded event_log reads across every other trace in the
+      // thread. So the public branch reads PREVIEW ONLY, authorizes down to the
+      // shared trace IDs, then resolves full IO for JUST those authorized IDs.
+      if (!ctx.publiclyShared) {
+        return traceService.getTracesByThreadId(
+          projectId,
+          threadId,
+          protections,
+          { full: true },
+        );
+      }
+
+      // Public-share path: preview-only read (zero event_log resolution).
+      const previewTraces = await traceService.getTracesByThreadId(
         projectId,
         threadId,
         protections,
-        { full: true },
+        { full: false },
       );
-
-      if (!ctx.publiclyShared) {
-        return tracesGrouped;
-      }
 
       const publicSharedTraces = await ctx.prisma.publicShare.findMany({
         where: {
           projectId: projectId,
           resourceType: PublicShareResourceTypes.TRACE,
           resourceId: {
-            in: tracesGrouped.map((trace) => trace.trace_id),
+            in: previewTraces.map((trace) => trace.trace_id),
           },
         },
       });
 
-      const filteredTraces = tracesGrouped.filter((trace) =>
-        publicSharedTraces.some(
-          (publicShare) => publicShare.resourceId === trace.trace_id,
-        ),
+      const authorizedTraceIds = previewTraces
+        .filter((trace) =>
+          publicSharedTraces.some(
+            (publicShare) => publicShare.resourceId === trace.trace_id,
+          ),
+        )
+        .map((trace) => trace.trace_id);
+
+      if (authorizedTraceIds.length === 0) {
+        return [];
+      }
+
+      // Resolve full IO for ONLY the post-authorization trace IDs.
+      const authorizedTraces = await traceService.getTracesWithSpans(
+        projectId,
+        authorizedTraceIds,
+        protections,
+        undefined,
+        { full: true },
       );
 
-      return filteredTraces;
+      // Preserve the chronological order the thread view expects.
+      authorizedTraces.sort(
+        (a, b) =>
+          (a.timestamps.started_at ?? 0) - (b.timestamps.started_at ?? 0),
+      );
+
+      return authorizedTraces;
     }),
 
   getTracesWithSpans: protectedProcedure
--- a/langwatch/src/server/api/routers/traces.ts
+++ b/langwatch/src/server/api/routers/traces.ts
@@ -1,4 +1,5 @@
 import { on } from "node:events";
+import type { PrismaClient } from "@prisma/client";
 import { PublicShareResourceTypes } from "@prisma/client";
 import { TRPCError } from "@trpc/server";
 import shuffle from "lodash-es/shuffle";
@@ -10,6 +11,8 @@ import {
 } from "~/server/api/trpc";
 import { getApp } from "~/server/app-layer/app";
 import { formatSpansDigest } from "~/server/tracer/spanToReadableSpan";
+import type { Trace } from "~/server/tracer/types";
+import type { Protections } from "~/server/traces/protections";
 import { TraceService } from "~/server/traces/trace.service";
 import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
 import { createLogger } from "~/utils/logger/server";
@@ -30,6 +33,77 @@ import { getAllForProjectInput, tracesFilterInput } from "./traces.schemas";
 export { getAllForProjectInput };
 
 const logger = createLogger("langwatch:traces:sse-subscription");
+
+/**
+ * Reads a thread on behalf of an ANONYMOUS public-share caller, who is entitled
+ * to only the individually-shared traces within it — not the whole thread.
+ *
+ * Order matters here, and it is a security property (#4991, #5082): resolving
+ * `{ full: true }` de-offloads the entire IO value out of `event_log`, so doing
+ * it before authorization narrows the set would let one valid share link amplify
+ * unbounded `event_log` reads across every other trace in the thread — traces
+ * this caller may not read. That is the ClickHouse connection-pool exhaustion
+ * the ADR-022 offload exists to prevent, reachable unauthenticated.
+ *
+ * So: read PREVIEW-only (zero `event_log` reads), authorize down to the shared
+ * trace ids, and only then resolve full IO — for those ids alone.
+ */
+async function readPubliclySharedThread({
+  traceService,
+  prisma,
+  projectId,
+  threadId,
+  protections,
+}: {
+  traceService: TraceService;
+  prisma: PrismaClient;
+  projectId: string;
+  threadId: string;
+  protections: Protections;
+}): Promise<Trace[]> {
+  const previewTraces = await traceService.getTracesByThreadId(
+    projectId,
+    threadId,
+    protections,
+    { full: false },
+  );
+
+  const publicShares = await prisma.publicShare.findMany({
+    where: {
+      projectId,
+      resourceType: PublicShareResourceTypes.TRACE,
+      resourceId: { in: previewTraces.map((trace) => trace.trace_id) },
+    },
+  });
+
+  const authorizedTraceIds = previewTraces
+    .filter((trace) =>
+      publicShares.some((share) => share.resourceId === trace.trace_id),
+    )
+    .map((trace) => trace.trace_id);
+
+  if (authorizedTraceIds.length === 0) {
+    return [];
+  }
+
+  const authorizedTraces = await traceService.getTracesWithSpans(
+    projectId,
+    authorizedTraceIds,
+    protections,
+    undefined,
+    { full: true },
+  );
+
+  // `authorizedTraceIds` already carries the thread's canonical order — the
+  // preview read returned it sorted, and filter/map preserve that. A direct
+  // getTracesWithSpans call comes back in trace-id order instead, so re-project
+  // onto the ids rather than re-deriving an ordering here; the thread read stays
+  // the single owner of what "thread order" means.
+  const resolvedById = new Map(
+    authorizedTraces.map((trace) => [trace.trace_id, trace]),
+  );
+  return authorizedTraceIds.flatMap((id) => resolvedById.get(id) ?? []);
+}
 
 export const tracesRouter = createTRPCRouter({
   getAllForProject: protectedProcedure
@@ -231,15 +305,8 @@ export const tracesRouter = createTRPCRouter({
         buildTraceBlobResolutionDeps(),
       );
 
-      // SECURITY (#4991): full blob resolution de-offloads the whole IO value
-      // from event_log and must NEVER run pre-authorization. For an authorized
-      // (non-public) caller the whole thread is theirs to read, so resolve full
-      // up front. For a public-share caller — an anonymous, publicProcedure
-      // path — resolving the full thread first would (a) de-offload traces the
-      // caller is not authorized to see and (b) let a single valid share link
-      // amplify unbounded event_log reads across every other trace in the
-      // thread. So the public branch reads PREVIEW ONLY, authorizes down to the
-      // shared trace IDs, then resolves full IO for JUST those authorized IDs.
+      // An authorized caller owns the whole thread, so resolve it full up front.
+      // An anonymous public-share caller does not — see readPubliclySharedThread.
       if (!ctx.publiclyShared) {
         return traceService.getTracesByThreadId(
           projectId,
@@ -249,52 +316,13 @@ export const tracesRouter = createTRPCRouter({
         );
       }
 
-      // Public-share path: preview-only read (zero event_log resolution).
-      const previewTraces = await traceService.getTracesByThreadId(
+      return readPubliclySharedThread({
+        traceService,
+        prisma: ctx.prisma,
         projectId,
         threadId,
         protections,
-        { full: false },
-      );
-
-      const publicSharedTraces = await ctx.prisma.publicShare.findMany({
-        where: {
-          projectId: projectId,
-          resourceType: PublicShareResourceTypes.TRACE,
-          resourceId: {
-            in: previewTraces.map((trace) => trace.trace_id),
-          },
-        },
       });
-
-      const authorizedTraceIds = previewTraces
-        .filter((trace) =>
-          publicSharedTraces.some(
-            (publicShare) => publicShare.resourceId === trace.trace_id,
-          ),
-        )
-        .map((trace) => trace.trace_id);
-
-      if (authorizedTraceIds.length === 0) {
-        return [];
-      }
-
-      // Resolve full IO for ONLY the post-authorization trace IDs.
-      const authorizedTraces = await traceService.getTracesWithSpans(
-        projectId,
-        authorizedTraceIds,
-        protections,
-        undefined,
-        { full: true },
-      );
-
-      // Preserve the chronological order the thread view expects.
-      authorizedTraces.sort(
-        (a, b) =>
-          (a.timestamps.started_at ?? 0) - (b.timestamps.started_at ?? 0),
-      );
-
-      return authorizedTraces;
     }),
 
   getTracesWithSpans: protectedProcedure
@@ -533,10 +561,12 @@ export const tracesRouter = createTRPCRouter({
         projectId: input.projectId,
       });
 
-      // Download consumes trace content — resolve full IO when spans are
-      // included so exported rows carry the whole value, not the 64 KB preview
-      // (#4991 AC1). resolveBlobs is gated on includeSpans (resolution runs
-      // during span enrichment).
+      // A download consumes trace content, so it must never serve the 64 KB
+      // preview (#4991 AC1) — and that holds whether or not spans are included,
+      // because the returned traces carry trace-level `input`/`output` either
+      // way. Gating resolveBlobs on includeSpans (as this did) silently
+      // truncated any offloaded trace in a spans-less download, the same
+      // data-loss bug fixed in ExportService for summary-mode exports.
       const traceService = TraceService.create(
         ctx.prisma,
         buildTraceBlobResolutionDeps(),
@@ -551,7 +581,7 @@ export const tracesRouter = createTRPCRouter({
         {
           downloadMode: true,
           includeSpans: input.includeSpans,
-          resolveBlobs: input.includeSpans,
+          resolveBlobs: true,
           scrollId: input.scrollId,
         },
       );
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -9,9 +9,9 @@
  */
 
 import type { PrismaClient } from "@prisma/client";
-import type { Protections } from "~/server/traces/protections";
 import type { Evaluation, Trace } from "~/server/tracer/types";
 import { enrichTracesWithEvaluations } from "~/server/traces/enrich-evaluations";
+import type { Protections } from "~/server/traces/protections";
 import type { TraceService } from "~/server/traces/trace.service";
 import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
 import { createLogger } from "~/utils/logger/server";
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -58,8 +58,7 @@ export class ExportService {
     const { TraceService: TraceServiceImpl } = await import(
       "~/server/traces/trace.service"
     );
-    const resolvedPrisma =
-      prisma ?? (await import("~/server/db")).prisma;
+    const resolvedPrisma = prisma ?? (await import("~/server/db")).prisma;
     // Export is a content-consuming read: wire blob-resolution deps so a full
     // export reads the whole IO value, not the 64 KB preview (#4991 AC1).
     const traceService = TraceServiceImpl.create(
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -12,6 +12,7 @@ import type { PrismaClient } from "@prisma/client";
 import type { Evaluation, Trace } from "~/server/tracer/types";
 import { enrichTracesWithEvaluations } from "~/server/traces/enrich-evaluations";
 import type { Protections } from "~/server/traces/protections";
+import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
 import type { TraceService } from "~/server/traces/trace.service";
 import { createLogger } from "~/utils/logger/server";
 import {
@@ -57,8 +58,14 @@ export class ExportService {
     const { TraceService: TraceServiceImpl } = await import(
       "~/server/traces/trace.service"
     );
-    const resolvedPrisma = prisma ?? (await import("~/server/db")).prisma;
-    const traceService = TraceServiceImpl.create(resolvedPrisma);
+    const resolvedPrisma =
+      prisma ?? (await import("~/server/db")).prisma;
+    // Export is a content-consuming read: wire blob-resolution deps so a full
+    // export reads the whole IO value, not the 64 KB preview (#4991 AC1).
+    const traceService = TraceServiceImpl.create(
+      resolvedPrisma,
+      buildTraceBlobResolutionDeps(),
+    );
     return new ExportService({ traceService });
   }
 
@@ -148,6 +155,10 @@ export class ExportService {
         {
           downloadMode: true,
           includeSpans,
+          // Full export resolves offloaded IO to the whole value (#4991 AC1).
+          // Summary export (includeSpans=false) reads no span content, so it
+          // stays on the preview and issues zero event_log reads.
+          resolveBlobs: includeSpans,
           scrollId: scrollId ?? null,
         },
       );
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -9,11 +9,11 @@
  */
 
 import type { PrismaClient } from "@prisma/client";
+import type { Protections } from "~/server/traces/protections";
 import type { Evaluation, Trace } from "~/server/tracer/types";
 import { enrichTracesWithEvaluations } from "~/server/traces/enrich-evaluations";
-import type { Protections } from "~/server/traces/protections";
-import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
 import type { TraceService } from "~/server/traces/trace.service";
+import { buildTraceBlobResolutionDeps } from "~/server/traces/trace-blob-resolution.deps";
 import { createLogger } from "~/utils/logger/server";
 import {
   serializeTracesToFullCsv,
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -154,15 +154,12 @@ export class ExportService {
         {
           downloadMode: true,
           includeSpans,
-          // SECURITY/correctness (#4991): BOTH export modes are content-consuming
-          // reads. Full mode emits span IO; summary mode still emits trace-level
-          // `trace.input`/`trace.output` (see csv/json summary serializers). For
-          // an offloaded (>64 KB) trace, gating resolution on `includeSpans`
-          // would silently ship the truncated 64 KB preview in a summary export
-          // with no error and no indication data was cut. Resolve blobs for every
-          // export mode so no export path serves the preview it shouldn't. Exports
-          // are a bounded/streamed bulk read via this PR's batch resolver, so the
-          // extra event_log reads stay pool-safe.
+          // DATA LOSS (#4991): summary mode reads no span content, but it still
+          // emits trace-level `trace.input`/`trace.output` (see the csv/json
+          // summary serializers) — so it is content-consuming too. Gating on
+          // `includeSpans` therefore shipped the truncated 64 KB preview for any
+          // offloaded trace, silently. Resolve for every export mode; the batch
+          // resolver keeps the extra event_log reads bounded.
           resolveBlobs: true,
           scrollId: scrollId ?? null,
         },
--- a/langwatch/src/server/export/export.service.ts
+++ b/langwatch/src/server/export/export.service.ts
@@ -154,10 +154,16 @@ export class ExportService {
         {
           downloadMode: true,
           includeSpans,
-          // Full export resolves offloaded IO to the whole value (#4991 AC1).
-          // Summary export (includeSpans=false) reads no span content, so it
-          // stays on the preview and issues zero event_log reads.
-          resolveBlobs: includeSpans,
+          // SECURITY/correctness (#4991): BOTH export modes are content-consuming
+          // reads. Full mode emits span IO; summary mode still emits trace-level
+          // `trace.input`/`trace.output` (see csv/json summary serializers). For
+          // an offloaded (>64 KB) trace, gating resolution on `includeSpans`
+          // would silently ship the truncated 64 KB preview in a summary export
+          // with no error and no indication data was cut. Resolve blobs for every
+          // export mode so no export path serves the preview it shouldn't. Exports
+          // are a bounded/streamed bulk read via this PR's batch resolver, so the
+          // extra event_log reads stay pool-safe.
+          resolveBlobs: true,
           scrollId: scrollId ?? null,
         },
       );
--- a/langwatch/src/server/traces/clickhouse-trace.service.ts
+++ b/langwatch/src/server/traces/clickhouse-trace.service.ts
@@ -61,6 +61,19 @@ export type ResolveTraceSpansFn = (
 ) => Promise<ResolvedTraceSpans>;
 
 /**
+ * Callback injected from TraceService that resolves offloaded blob refs for a
+ * WHOLE result set of traces in one bounded pass (#4991 bulk read paths). When
+ * present, the bulk read methods (getTracesWithSpans, enrichTracesWithSpans on
+ * the download path) use it so a large export/thread streams its event_log
+ * reads instead of fanning out an unbounded N×M burst. Falls back to the
+ * per-trace {@link ResolveTraceSpansFn} when absent.
+ */
+export type ResolveTraceSpansBatchFn = (
+  projectId: string,
+  spansPerTrace: NormalizedSpan[][],
+) => Promise<ResolvedTraceSpans[]>;
+
+/**
  * Cursor structure for keyset pagination.
  * Encoded as base64 JSON in the scrollId.
  */
@@ -237,11 +250,21 @@ export class ClickHouseTraceService {
    */
   private readonly resolveTraceSpans: ResolveTraceSpansFn | undefined;
 
+  /**
+   * Optional bulk resolver for whole result sets (#4991). Preferred over
+   * {@link resolveTraceSpans} on the bulk read paths so a large export/thread
+   * resolves its blobs in one bounded-concurrency pass. When absent, the bulk
+   * paths fall back to the per-trace resolver.
+   */
+  private readonly resolveTraceSpansBatch: ResolveTraceSpansBatchFn | undefined;
+
   constructor(
     private readonly prisma: PrismaClient,
     resolveTraceSpans?: ResolveTraceSpansFn,
+    resolveTraceSpansBatch?: ResolveTraceSpansBatchFn,
   ) {
     this.resolveTraceSpans = resolveTraceSpans;
+    this.resolveTraceSpansBatch = resolveTraceSpansBatch;
   }
 
   /**
@@ -269,8 +292,13 @@ export class ClickHouseTraceService {
   static create(
     prisma: PrismaClient = defaultPrisma,
     resolveTraceSpans?: ResolveTraceSpansFn,
+    resolveTraceSpansBatch?: ResolveTraceSpansBatchFn,
   ): ClickHouseTraceService {
-    return new ClickHouseTraceService(prisma, resolveTraceSpans);
+    return new ClickHouseTraceService(
+      prisma,
+      resolveTraceSpans,
+      resolveTraceSpansBatch,
+    );
   }
 
   /**
@@ -325,18 +353,15 @@ export class ClickHouseTraceService {
             occurredAt,
           );
 
-          // Map to legacy Trace format and apply protections
-          const traces: Trace[] = [];
-          for (const [_traceId, { summary, spans }] of tracesWithSpans) {
-            const trace = await this.resolveAndMerge({
-              projectId,
-              summary,
-              spans,
-              protections,
-              resolveBlobs: opts?.resolveBlobs,
-            });
-            traces.push(trace);
-          }
+          // Map to legacy Trace format and apply protections. Blob resolution
+          // (when opted in) runs as a single bounded pass over the whole set so
+          // a large multi-trace read streams its event_log reads (#4991 AC6).
+          const traces = await this.resolveAndMergeMany({
+            projectId,
+            entries: [...tracesWithSpans.values()],
+            protections,
+            resolveBlobs: opts?.resolveBlobs,
+          });
 
           this.logger.debug(
             { projectId, traceCount: traces.length },
@@ -441,14 +466,18 @@ export class ClickHouseTraceService {
    * @param projectId - The project ID
    * @param threadId - The thread ID to search for
    * @param protections - Field redaction protections
-   * @returns Array of Trace objects
+   * @param opts.resolveBlobs - Forwarded to the per-trace fetch so the
+   *   thread-detail read resolves full IO (#4991). Customer thread views that
+   *   construct without a blob resolver get a no-op. Defaults to false.
+   * @returns Array of Trace objects, or null if ClickHouse is not enabled
    * @throws ClickHouseClientUnavailableError when no ClickHouse client resolves
    */
   async getTracesByThreadId(
     projectId: string,
     threadId: string,
     protections: Protections,
-  ): Promise<Trace[]> {
+    opts?: { resolveBlobs?: boolean },
+  ): Promise<Trace[] | null> {
     return await this.tracer.withActiveSpan(
       "ClickHouseTraceService.getTracesByThreadId",
       {
@@ -488,11 +517,15 @@ export class ClickHouseTraceService {
             return [];
           }
 
-          // Fetch full traces with spans
+          // Fetch full traces with spans. Forward resolveBlobs so the
+          // thread-detail read can resolve full IO (#4991); customer thread
+          // views with no resolver wired stay on the preview.
           const traces = await this.getTracesWithSpans(
             projectId,
             traceIds,
             protections,
+            undefined,
+            { resolveBlobs: opts?.resolveBlobs },
           );
 
           // Re-sort by timestamp — getTracesWithSpans returns in TraceId
@@ -764,12 +797,16 @@ export class ClickHouseTraceService {
               dateField,
             });
 
-          // When includeSpans is requested, fetch and attach actual spans
+          // When includeSpans is requested, fetch and attach actual spans.
+          // resolveBlobs is opt-in (download/export — #4991 AC1) so the
+          // list/search grid stays on the preview with zero event_log reads
+          // (#4888 AC2 / ADR-022 — AC5).
           if (options.includeSpans && traces.length > 0) {
             traces = await this.enrichTracesWithSpans(
               traces,
               input.projectId,
               protections,
+              options.resolveBlobs === true,
             );
           }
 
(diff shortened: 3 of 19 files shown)


drewdrewthis:
**Re: CodeRabbit review** — thanks; all 11 points were evaluated in the maintainer review verdict above (`## Review verdict: READY`). Disposition:

- **BDD `describe("given …")` shape** in the new call-site tests — noted; the suites use given/when nesting broadly, the flagged spots are a minor structure nit (non-blocking).
- **Public-share thread over-fetch** (`traces.ts`) — real, tracked as a follow-up in the verdict (filter to shared traceIds before resolving). Not a leak today (output filtered, reads tenant-scoped).
- **Summary-export full IO** (`export.service.ts`) — tracked follow-up (perf-vs-AC1 product call); full export is correct.
- **Dup warn/degrade policy, `forEachWithConcurrency` reuse, resolver-cardinality guard, concurrency validation, split-into-phases** — non-blocking quality items in the verdict's follow-up table.

The cheap real defects (falsifiable preview assertions + 2 stale doc-comments) were fixed in `62119a64a`. Resolving these threads against that disposition.

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** A bulk content read accidentally resolving blobs on the list/search grid (or an unbounded `N×M` `event_log` fan-out on a large export) exhausting the ClickHouse connection pool — the exact heavy-read failure the offload exists to prevent.

- `[Must Check]` `getAllTracesForProject` gating in `trace.service.ts` / `clickhouse-trace.service.ts` — confirm resolution keys off the dedicated `resolveBlobs` option and NOT `includeSpans`, so `app.v1`'s public list (which passes `includeSpans=true`) issues zero `event_log` reads.
- `[Must Check]` `resolve-offloaded-traces-batch.ts` — verify dedupe of `(aggregateId, eventId, field)` refs plus bounded-concurrency pool actually caps peak in-flight CH reads to a constant independent of result-set size.
- `[Ask Author]` per-ref failure path — confirm a missing/failed `event_log` row keeps that field's 64 KB preview and only `warn`s (no silent truncation, no whole-export failure).

**Author questions:**
- What is the concrete concurrency ceiling and where is it configured? Is it shared across concurrent exports or per-call (many simultaneous exports could still stack)?

drewdrewthis:
## Review verdict: READY

**…with two decisions that are a human's to make, not mine — see the bottom.**

Scoped re-review at `011c32ca5` (the prior verdict was clean at `6440cc9a9`, since rebased onto main +87). Reviewers: `[principles]` ×2 (the second an adversarial pass on my own fixes), `[security]`, `[hygiene]`, `[test]`.

**Read this first: the review found six defects, and three of them were in my own fix commits.** The PR is stronger for it, but that record is the main thing a human should weigh.

### The one that matters — my first fix for the P1 was inert

`[principles]` The P1 blocker was "summary exports serve the truncated 64 KB preview." I fixed it by setting `resolveBlobs: true` at the call sites, wrote a test, went green, replied on the thread that it was fixed, and resolved it.

**It did nothing.** Resolution lives inside `enrichTracesWithSpans`, which ran only `if (options.includeSpans)`. Summary mode sets `includeSpans = false`, so `resolveBlobs` was **never read** — zero `event_log` reads, preview still shipped, and a comment now asserting the opposite.

The tests were green the whole time because they mocked `getAllTracesForProject` and asserted the flag was **forwarded** — *an assertion that passes whether or not the flag has any runtime effect.* It tested my intent, not the behaviour.

Actually fixed in `6860180b9`: spans are fetched when the caller wants spans **OR** wants full IO (the full value is recoverable *only* from the spans' eventref pointers), then discarded for a summary caller. Proven by driving the **real** service + resolver and asserting the **value**, confirmed RED first. Thread corrected [here](https://github.com/langwatch/langwatch/pull/5082#discussion_r3568340860).

### Concerns

1. **[principles] The batch-resolver guard overclaimed, twice.** v1 promised "one resolution per trace, *in input order*" but only compared lengths — a reordering resolver passed silently, which is the silent-corruption case it advertised catching. v2 added an identity check that a **span-less** trace slipped through (no `traceId` on either side to compare), letting a real trace inherit an empty resolution and lose its spans. Now checks span count *and* identity; both holes have falsifiable tests. **This guard has been wrong twice — see Decision 2.**
2. **[hygiene] The unwrapped re-throw missed a third catch.** `getTracesWithSpansByThreadIds` sits *above* `getTracesWithSpans` and re-flattened what it re-threw, on a live `resolveBlobs` path. Allowlisted.
3. **[principles][hygiene] Comments that described behaviour the code didn't implement.** The error class claimed to escape "like `ClickHouseClientUnavailableError`" — but that one escapes by being thrown *outside* the try, an opposite mechanism. Corrected.
4. **[test] The security property was under-guarded in the one scenario that mattered.** The multi-shared case (an unauthorized sibling *between* two shared traces) asserted only on the returned traces — and the mock's return doesn't depend on its args, so a router resolving `["t1","t2","t3"]` would still have passed. Added the call-args exclusion assert + `toHaveBeenCalledTimes(1)`; a simulated partial-authorization regression now fails both.
5. **[test] Two tests couldn't fail.** "returns the fully-resolved authorized traces" asserted only `trace_id`, so it passed identically pre- and post-fix. Given a distinguishable full-vs-preview value.
6. **[principles] +1 defect this PR shipped:** `getAllForDownload` still had `resolveBlobs: input.includeSpans` — the same bug, twin path. Fixed and covered.

### Verified sound

- **[security]** The public-share fix was traced adversarially end-to-end (router → `checkPermissionOrPubliclyShared` → service → the CH queries). It correctly closes the resource-amplification vector it targets — but it is **not a complete authorization fix for this endpoint**, and a follow-up review (`ac-reviewer`, then reproduced by me on real ClickHouse) found a **pre-existing content-disclosure bug this PR neither introduces nor fixes** — an unvalidated parameter on the public-share path allows an unauthorized cross-scope read of shared traces. Filed **P1** as [#5747](https://github.com/langwatch/langwatch-saas/issues/722) (technical detail redacted there pending a private security advisory). For the scope of THIS PR: no path resolves or returns content the pre-existing middleware did not already authorize; `publicShare.findMany` scopes on projectId + resourceType + resourceId; the preview set is used only to derive candidate ids and never returned; the empty short-circuit leaks no thread existence; `TenantId` is first predicate on every CH query. **AC5 preserved by construction** — `resolveBlobs` stays opt-in, so list/aggregation paths resolve nothing and issue zero `event_log` reads whether or not they request spans (tests still green).
- **[hygiene]** Every factual comment in the delta was spot-checked against the code it describes.

### Two decisions for a human — I did NOT make these

**Decision 1 — summary exports now cost more.** They do a span join + bounded `event_log` reads where they previously did neither. That's the tradeoff the reviewer explicitly offered on the blocker thread, and there is **no cheaper way** to recover the value (the full IO exists only on the span eventrefs). But it's a real new cost on a path that had none. **Ratify it deliberately rather than inherit it from a comment.**

**Decision 2 — the guard is defending a design, not a bug.** `ResolvedTraceSpans` carries no trace identity, so the pairing is *positional* and unenforceable by the type. That's why it needs a runtime guard, a custom error, three catch allowlists, and four tests — and why the guard has now been wrong twice. Putting `traceId` on `ResolvedTraceSpans` would make the mispairing **unrepresentable** and delete all of that machinery. Wider blast radius (both resolvers + consumers), so I did not do it inside a blob-resolution PR. `[principles]` and I both think it's the right end state.

### Deliberate divergence from reviewer asks

- The cardinality guard was requested at **both** index sites. It is implemented **once**, at the boundary where the untrusted value enters. `merged` comes from `entries.map(...)`, so `merged.length === enrichable.length` is a tautology and a second assert would be unreachable.
- `[principles]` argued the two flattening catches should just log-and-rethrow, deleting the `instanceof` allowlists. **Declined here:** that changes the error surface for *every* failure on these paths, and raw ClickHouse errors can carry schema/query internals toward API consumers. That's a security-shaped change, not a cleanup, and it belongs in its own PR.

### Follow-ups (filed / to file — none blocking)

- **CI is masking failures repo-wide.** `test-unit (2)` reports `success` while the `globalSetup` hard-floor `process.exit(0)` kills it mid-run. It hid a failing test in *this* PR, and it is **currently hiding a genuinely failing `team.update.multi-binding` test on green `main`.** Evidence posted to [#4450](https://github.com/langwatch/langwatch/issues/4450#issuecomment-4954718351). A green shard with no `Test Files N passed` tally line did not finish.
- `[principles]` `TraceService.getTracesWithSpans` takes 5 positional args with an optional in the middle, so every call site passes a bare `undefined` — a CLAUDE.md violation on the public service API, while this PR fixed exactly that defect on a private 3-string helper.
- `[test][hygiene]` `auditLogMutations` reaches for the `prisma` module singleton instead of `ctx.prisma`, forcing 34 test files to mock it.
- `[hygiene]` `getAllForDownload` has no caller outside its own test (the live path is the Hono route via `ExportService`) — possibly dead, worth confirming before it rots.
- `[security]` **[#5747](https://github.com/langwatch/langwatch-saas/issues/722) — now P1, and worse than I first wrote it.** `checkPermissionOrPubliclyShared` authorizes on `input.traceId` alone and never validates `input.threadId`. I initially called this "no content leak" — **that was wrong**; it is a real content-disclosure vector (reproduced). Pre-existing (predates this PR) and out of scope here. Full technical detail is redacted from the public issue pending a private security advisory — see **P1 [#5747](https://github.com/langwatch/langwatch-saas/issues/722)**.

### Evidence

**The P1 is now proven end-to-end against a REAL ClickHouse, not against a mock's arguments.** `export-summary-blob-resolution.integration.test.ts` seeds a genuinely offloaded trace (full 200 KB value in `event_log`, 64 KB preview + eventref in `stored_spans`, preview in `trace_summaries`) and drives the **real** `ClickHouseTraceService` → **real** batch resolver + `BlobStore` → **real** `ExportService` in summary mode → **real** CSV/JSONL serializers, asserting the exported **bytes** carry a marker that exists only past the preview boundary. AC5 is guarded in the same file — the list read over the same trace still gets the preview.

Confirmed RED at the real seam: restoring the pre-fix `if (options.includeSpans)` gate fails all three export assertions, and the CSV assertion prints the truncated preview back at you. In CI: `✓ export-summary-blob-resolution.integration.test.ts (4 tests)`.

Every other fix likewise confirmed falsifiable — revert it, watch the named test go red. `pnpm typecheck` **and** `pnpm typecheck:tests` clean (the former excludes test files; they are separate CI steps).

**CI, stated honestly rather than by checkmark.** All checks report pass — but `test-unit (2)` hard-floored and **did not finish**, so its green is not evidence of anything. I read the shard logs instead of trusting the badge:

- **All 6 of this PR's suites are green in CI** — 48 tests: `traces.4991` (16), `clickhouse-trace.service-4991-bulk` (13), `resolve-offloaded-traces-batch` (9), `export-service.4991` (5), `traces-legacy-thread` (3), `annotation.4991` (2). Matches the local run exactly.
- The masked shard contains **exactly two** failing files, and **both fail identically on a pristine `origin/main` worktree** — `team.update.multi-binding` (2) and `aggregation-builder` (1). Inherited from main, not introduced here. Evidence in [#4450](https://github.com/langwatch/langwatch/issues/4450#issuecomment-4954984062).

If you want to check a LangWatch shard yourself: **a green shard with no `Test Files N passed` tally line did not finish.**

Aryansharma28 (commented):
## goated-review — deep audit

Reviewed at HEAD `6440cc9a` (CI fully green, `mergeable`). This is an independent, adversarially-verified pass on top of CodeRabbit's 8 actionable comments and the author's own self-review verdict (`READY`) already on this thread. Per the reviewer's request, findings are bucketed as **BLOCKERS (P0–P2)** vs **NON-BLOCKERS (P3–P5)** — explicit, unambiguous.

_(Note: this review's body replaces a version briefly posted with content from an unrelated PR due to a local tooling mixup on my end — apologies for the noise; this is the correct, corrected content for #5082.)_

The core mechanism (`resolveOffloadedTracesBatch`: decode → dedupe `(aggregateId, eventId, field)` → bounded-concurrency fetch → scatter/recompute) is sound, well-tested (150 tests, RED→GREEN verified per the PR description), and the shared-decoder extraction into `offloaded-eventref-parsing.ts` is a clean, faithful refactor. I traced every `resolveBlobs`/`includeSpans` gate across `getAllTracesForProject`, `enrichTracesWithSpans`, and the list/search grid — the #4888 zero-`event_log`-read guarantee for list/aggregation reads holds in every path I checked. That part of the series is solid.

However, two items that CodeRabbit already raised (and the author dispositioned as **non-blocking follow-ups** in the self-review verdict) are, on independent verification, **BLOCKERS**, not follow-ups — they mean this "2 of 2" PR does not actually close the bug class #4991 describes for every path it claims to fix, and one of them is a new anonymous-reachable resource-amplification vector introduced by this exact PR. I'm overriding the existing "non-blocking" disposition on both — see inline comments for why the rebuttal doesn't hold up.

### 🔴 BLOCKERS (P0–P2) — must fix before merge

| # | Sev | Location | Issue |
|---|---|---|---|
| 1 | **P1** | `traces.ts:237` (`getTracesByThreadId`) | Public-share thread read resolves **full blobs for every trace in the thread** before filtering down to the publicly-shared subset — an anonymous, unauthenticated caller can force unbounded `event_log` reads for traces they have no right to see the content of. This is a *new* attack surface this PR introduces (pre-#4991 this endpoint never resolved blobs at all). |
| 2 | **P1** | `export.service.ts:160` (+ `csv-serializer.ts:91-92`) | Summary-mode CSV/JSONL export still serializes `trace.input`/`trace.output` from the **64 KB preview** — `resolveBlobs: includeSpans` is `false` for summary mode, but the summary serializer emits trace-level IO regardless. This is the *exact* bug the PR's own "Why" section names ("a full CSV/JSONL export lost data") — unfixed for one whole export mode. |
| 3 | **P2** | `resolve-offloaded-traces-batch.ts:66-70` (`fetchKeyOf`) | Three positional `string` args (`aggregateId, eventId, field`) for the dedupe/fetch key — violates this repo's own named-parameter rule (CLAUDE.md: "Using positional parameters for functions with multiple args → use named parameters"). A reordered call compiles cleanly and silently fetches/dedupes the wrong blob. |
| 4 | **P2** | `clickhouse-trace.service.ts:2307` (`resolveAndMergeMany`, mirrored at `:2463`) | `resolutions[i]!` / `merged[i]!` trust the injected batch-resolver's output array to have exactly one entry per input trace, with no length check. `resolveOffloadedTracesBatch`'s current implementation happens to preserve this via `.map`, but nothing in the `ResolveTraceSpansBatchFn` type enforces it — a future DI implementation (or test double) that returns a short/reordered array corrupts trace-to-summary pairing silently, or throws with no context, on the hot bulk-read path (export/thread/dataset). |

### 🟡 NON-BLOCKERS (P3–P5) — polish / already tracked, not gating

| # | Sev | Location | Issue |
|---|---|---|---|
| 5 | P3 | `resolve-offloaded-traces-batch.ts:144-257` | `resolveOffloadedTracesBatch` is one ~150-line exported function doing parse+fetch+assemble. Already flagged by CodeRabbit ("split into phase helpers") — agree, low urgency, the 3 phases are already comment-delimited. |
| 6 | P3 | `resolve-offloaded-traces.ts` vs `resolve-offloaded-traces-batch.ts` | Per the author's own self-review verdict: the eventref *decoder* was shared, the warn/degrade *policy* (`warnResolutionFailure` shape) was not — duplicated between both resolvers, can drift. |
| 7 | P3 | `resolve-offloaded-traces-batch.ts:151` (`concurrency` param) | Accepts unvalidated input (`NaN`/negative silently disables the AC6 bound). CodeRabbit-flagged. Low current risk — always internally defaulted to `EVENT_LOG_RESOLVE_CONCURRENCY` today — but worth a clamp since AC6's whole guarantee rests on this number being sane. |
| 8 | P3 | `traces-legacy.ts` (whole file) | ~150 of this file's 306 changed lines are pure Prettier rewraps of unrelated `secured.access(...).get(...)` chains, mixed into the same diff as the functional `{ full: true }` change. Inflates the diff 3x and buries the actual behavior change — worth a separate formatting-only commit next time. |
| 9 | P4 | `traces-legacy.ts:235` | `Record<string, any>` for the pre-Zod-parse body — CodeRabbit-flagged, agree; `unknown` is safer and the value is narrowed by `paramsSchema.strict().parse()` two lines later anyway. |
| 10 | P4 | 3 new test files (`annotation.4991-full-resolution`, `traces.4991-full-resolution`, `traces-legacy-thread`) | Missing the `describe("given ...")` outer layer the repo's BDD convention calls for — CodeRabbit flagged this 3x already; cosmetic, no behavior risk. |
| 11 | P5 | `offloaded-eventref-parsing.ts:31,46` | Praise: clean extraction — verified the new `hasEventRefs`/`parseSpanEventRefs` are byte-for-byte behavior-preserving vs. the inline logic they replaced in `resolve-offloaded-traces.ts`. This is the DRY-without-coupling pattern the codebase should keep doing. |
| 12 | P5 | series-wide | Confirmed consistency with #4888/ADR-022: every list/search/aggregation read path I traced still gates on an explicit `resolveBlobs`/`full` opt-in and never fires `event_log` reads by default — the heavy-read protection part 1 established is not weakened by this PR outside of finding #1 above. |

- Whether `EVENT_LOG_RESOLVE_CONCURRENCY = 25` was chosen against a real ClickHouse connection-pool budget, or is a guess — I don't have prod pool-size figures.
- Whether product actually signed off on "summary export stays truncated" as an intentional scope cut (item #2) — the PR thread frames it as a perf tradeoff but I see no explicit product decision referenced, and no test asserts/documents the limitation.

**Verdict: COMMENT, not APPROVE.** Two P1s need a real fix (not a deferred issue) before this can be called "2 of 2" complete for #4991; the P2s are cheap and should land in the same PR given how easy the fixes are.

langwatch/src/server/api/routers/traces.ts
```diff
@@ -224,11 +224,17 @@ export const tracesRouter = createTRPCRouter({
...
+      // (#4991), not the 64 KB preview.
+      const traceService = TraceService.create(
+        ctx.prisma,
+        buildTraceBlobResolutionDeps(),
+      );
       const tracesGrouped = await traceService.getTracesByThreadId(
         projectId,
         threadId,
         protections,
+        { full: true },
```
Aryansharma28: [P1] BLOCKER — public-share thread read resolves full blobs for the whole thread before authorization narrows it down.

**Why:** `getTracesByThreadId(..., { full: true })` runs unconditionally, for every trace in the thread, before the `ctx.publiclyShared` branch below (lines 240-260) filters the *response* down to only the publicly-shared trace(s). The filtering happens after the expensive read, not before it. An anonymous caller who holds a single valid public-share link into a busy thread can force full `event_log` resolution of every other trace in that thread on every request — traces they are not authorized to see the content of. Pre-#4991 this endpoint never resolved blobs at all, so this specific amplification vector is new in this PR, on a `publicProcedure` (anonymous-reachable) endpoint. It is not a data leak (the response is still filtered), but it is exactly the "exhaust the CH connection pool" failure mode the ADR-022 offload — and this whole two-part series — exists to prevent, now reachable by an unauthenticated caller.

CodeRabbit flagged this already and the author's self-review verdict dispositioned it as a non-blocking follow-up ("Not a leak today"). I'm re-raising as a blocker: "not a leak" doesn't address the resource-amplification risk, which is the actual concern, and the fix is small.

**Suggested fix:** Fetch previews first, filter/authorize, then resolve only the authorized trace IDs — e.g. only pass `{ full: true }` when `!ctx.publiclyShared`, and for the `publiclyShared` branch, resolve full IO only for the post-filter `filteredTraces` IDs (CodeRabbit's suggested diff for this shape is a reasonable starting point).

langwatch/src/server/export/export.service.ts
```diff
@@ -148,6 +154,10 @@ export class ExportService {
         {
           downloadMode: true,
           includeSpans,
+          // Full export resolves offloaded IO to the whole value (#4991 AC1).
+          // Summary export (includeSpans=false) reads no span content, so it
+          // stays on the preview and issues zero event_log reads.
+          resolveBlobs: includeSpans,
```
Aryansharma28: [P1] BLOCKER — summary-mode export still serves the 64 KB preview for trace.input/output.

**Why:** `resolveBlobs: includeSpans` is `false` whenever `request.mode !== "full"` (summary mode). But `serializeTracesToSummaryCsv` (csv-serializer.ts:91-92) and the summary JSON serializer both emit `trace.input?.value` / `trace.output?.value` directly — trace-level content, not spans. For an offloaded (>64 KB) trace, a user who exports in "summary" mode gets the truncated preview silently, with no error and no indication data was cut. This is the literal bug this PR's "Why" section opens with ("a full CSV/JSONL export lost data") — it's fixed for `mode: "full"` but not for `mode: "summary"`, even though summary mode is a content-consuming read (it reads `trace.input`/`trace.output`), not a list/aggregation read.

CodeRabbit flagged this in 3 separate review rounds (as a 🟠 Major, twice as a duplicate) and the author's self-review verdict dispositioned it as a non-blocking "product call" (perf tradeoff). I don't see an actual product decision recorded anywhere in this thread, nor a test/doc asserting summary exports are allowed to truncate >64KB content — just a code comment asserting it's fine. Given the PR's own title claims to resolve blobs "on all content-consuming read paths," I'm re-raising this as a blocker rather than accepting the deferral.

**Suggested fix:** Either resolve blobs for every export mode (`resolveBlobs: true` unconditionally, accepting the extra `event_log` reads — exports are already a bounded/streamed bulk read via this same PR's batch resolver), or add an explicit regression test proving summary-mode output never carries offloadable trace-level content, plus a one-line doc/changelog note recording the scope cut as an intentional decision.

langwatch/src/server/traces/resolve-offloaded-traces-batch.ts
```diff
@@ -0,0 +1,257 @@
...
+
+/** Internal: outcome of a single event_log fetch. */
+type FetchResult = { ok: true; value: string } | { ok: false; error: unknown };
+
+/** Builds the dedup key for a fetch task. NUL separator can't collide with ids. */
+function fetchKeyOf(
+  aggregateId: string,
+  eventId: string,
+  field: string,
+): string {
```
Aryansharma28: [P2] `fetchKeyOf` takes three positional string args — swap risk on the dedupe key.

**Why:** `fetchKeyOf(aggregateId, eventId, field)` — all three are plain `string`, so a caller that accidentally passes `eventId` and `field` in the wrong order compiles cleanly and silently dedupes/fetches the wrong blob (wrong `event_log` row content scattered onto the wrong span). This repo's own coding standard (CLAUDE.md TypeScript section) is explicit: "Using positional parameters for functions with multiple args → use named parameters via object destructuring: `fn({ a, b })` not `fn(a, b)`." CodeRabbit already flagged this with a concrete diff; it wasn't applied.

**Suggested fix:** `function fetchKeyOf({ aggregateId, eventId, field }: { aggregateId: string; eventId: string; field: string }): string`, update the two call sites.

langwatch/src/server/traces/clickhouse-trace.service.ts
```diff
@@ -2240,39 +2277,105 @@ export class ClickHouseTraceService {
...
+      projectId,
+      spansPerTrace: entries.map((e) => e.spans),
+      resolveBlobs,
+    });
+
+    return entries.map((entry, i) =>
+      this.mergeResolvedTrace({
+        projectId,
+        summary: entry.summary,
+        resolution: resolutions[i]!,
```
Aryansharma28, line 2405: [P2] Non-null assertion on resolver output without a cardinality guard.

**Why:** `resolutions[i]!` (and the mirrored `merged[i]!` splice-back at line 2463 in `enrichTracesWithSpans`) assumes the injected `ResolveTraceSpansBatchFn` / batch resolver always returns exactly one entry per input trace, in the same order. Today's concrete implementation (`resolveOffloadedTracesBatch`) happens to preserve this via `.map`, but nothing in the `ResolveTraceSpansBatchFn` type signature enforces the invariant — it's a convention, not a contract. This is on the hot bulk-read path shared by export, thread, and dataset/sample builders; a future change to the resolver (or a test double that doesn't preserve the invariant) either throws with zero context or, worse, silently pairs the wrong trace summary with the wrong resolved spans.

CodeRabbit flagged a similar concern at the pre-refactor line numbers; this is the same issue at the current call sites post-refactor.

**Suggested fix:** Assert `resolutions.length === entries.length` (and `merged.length === enrichable.length`) immediately after the batch call, throwing a descriptive error on mismatch instead of relying on the non-null assertion to fail silently or crash without context.

langwatch/src/server/traces/offloaded-eventref-parsing.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Splits a span's flat attributes into the preview attributes (reserved keys
+ * stripped) and the well-formed eventref pointers to resolve.
+ *
+ * - A reserved key missing/empty `eventId` is recorded in `missingEventIdKeys`
+ *   (the caller logs a warning and keeps the preview) — never resolved.
+ * - A reserved key with malformed JSON is silently dropped (the preview already
+ *   sits in `cleanedAttrs` under the non-reserved IO key).
+ */
+export function parseSpanEventRefs(
```
Aryansharma28, line 46: [P5] PRAISE — clean, faithful shared-decoder extraction.

Compared this against the inline logic it replaced in `resolve-offloaded-traces.ts` (pre-PR): `hasEventRefs` and `parseSpanEventRefs` preserve the exact same behavior (reserved-key stripping, malformed-JSON silent-drop, missing-eventId tracking) byte-for-byte, just relocated so both the per-trace (#4888) and batch (#4991) resolvers decode from one definition. This is the DRY-without-coupling pattern to keep using — good precedent for the next shared-mechanism extraction in this area.

langwatch/src/server/api/routers/traces.ts
drewdrewthis: Agreed, and fixed in `6e50c5a56`. You're right that "not a leak" was answering the wrong question — the concern is resource amplification, and dispositioning it as a follow-up was wrong.

The public branch now reads **preview-only** (zero `event_log` resolution), authorizes down to the shared trace ids, and then resolves full IO for **just those ids**:

- `!ctx.publiclyShared` → unchanged. The whole thread is theirs to read, so it still resolves full up front.
- `ctx.publiclyShared` → `getTracesByThreadId(..., { full: false })` → `publicShare.findMany` → `getTracesWithSpans(authorizedTraceIds, ..., { full: true })`, then re-sorted chronologically (mirroring the sort `getTracesByThreadId` already does internally, since `getTracesWithSpans` returns in trace-id order). Zero authorized ids short-circuits to `[]` with no resolution at all.

Guarded by 5 new tests in `traces.4991-full-resolution.unit.test.ts` under "public-share thread read". They're falsifiable — I restored the old resolve-then-filter order and 3 of them fail:

```
× reads the thread PREVIEW-only, never resolving blobs pre-authorization
× resolves full IO for ONLY the authorized trace ids
× returns them in chronological order
```

The load-bearing one asserts `getTracesWithSpans` is called with `["t2"]` — the single shared trace — and not with its two unauthorized siblings in the same thread.

langwatch/src/server/export/export.service.ts
drewdrewthis: Agreed, and fixed in `fc12eec45` — I took your first option (resolve for every export mode). You're right that there was no product decision recorded anywhere; the code comment was asserting a tradeoff nobody had actually made.

Confirmed your read of the mechanism: `buildSummaryRow` emits `trace.input?.value` / `trace.output?.value` (`serializers/csv-serializer.ts:91-92`), so summary mode is a content-consuming read and an offloaded (>64 KB) trace was silently shipping its truncated preview.

`resolveBlobs` is now `true` unconditionally. Exports are already a bounded/streamed bulk read through this PR's batch resolver, so the extra `event_log` reads stay pool-safe.

The old test asserted the buggy contract (`resolveBlobs === false` on summary) — I replaced it rather than leaving it to enshrine the bug. Two tests now:

- `opts resolveBlobs in so an offloaded trace is not truncated to its preview` — falsifiable; restoring `resolveBlobs: includeSpans` fails it.
- `emits the trace input/output value into the payload` — grounds *why* the first must hold, by proving the summary CSV really does carry the trace-level IO value that truncation was eating. (This one passes either way by design; it guards the premise, so if a future change stopped emitting IO in summary rows we'd learn that resolving there had become dead cost.)

langwatch/src/server/traces/resolve-offloaded-traces-batch.ts
drewdrewthis: Fixed in `85dcd457c`. Converted to named params per the CLAUDE.md TypeScript standard, and updated the call site:

```ts
function fetchKeyOf({ aggregateId, eventId, field }: {
  aggregateId: string; eventId: string; field: string;
}): string
```

The swap risk was real and silent — all three are plain `string`, so a transposition would compile clean and dedupe/fetch the wrong `event_log` row onto the wrong span.

langwatch/src/server/traces/clickhouse-trace.service.ts
drewdrewthis: Fixed in `85dcd457c` — but I put the guard in one place rather than the two you named, and want to flag the difference explicitly so you can push back.

**Guarded at the boundary, not at the index sites.** The only place the invariant can actually break is the **injected** `resolveTraceSpansBatch` — that's the untrusted contract. So `resolveSpansBatch` now asserts `resolutions.length === spansPerTrace.length` immediately on its return, which covers both `resolutions[i]!` and `merged[i]!` with a single assertion, and fails while the offending resolver is still nameable instead of downstream where the context is gone.

**Why no separate `merged.length === enrichable.length` assert:** `merged` is the return value of `resolveAndMergeMany`, which is `entries.map(...)`. `Array.prototype.map` always returns an array of the same length, so that equality is a tautology and the assert would be unreachable. The two non-injected paths in `resolveSpansBatch` (the per-trace `push` loop and the passthrough `.map`) are 1:1 by construction for the same reason. Happy to add it anyway if you'd rather have belt-and-braces against a future refactor of that private method.

**The error is a named class, and it escapes the generic catch.** Both read paths wrap failures in `throw new Error("Failed to fetch traces...")`, which would have flattened the descriptive message into exactly the context-free crash you were trying to avoid. So `TraceSpansBatchCardinalityError` follows the existing `ClickHouseClientUnavailableError` precedent and both catches re-throw it unwrapped — a resolver-contract violation is a code bug, not a fetch failure.

Covered by a test that drives a genuinely non-conforming resolver (returns `[]` for 1 input trace) through the real service and observes the throw, rather than asserting on a string. Falsifiable: removing the guard fails it.

langwatch/src/server/traces/offloaded-eventref-parsing.ts
drewdrewthis: Thanks — and thanks for actually diffing it against the pre-PR inline logic rather than taking the extraction on faith. That's the check that makes the "shared decoder" claim worth anything.

Agreed on the precedent: one definition, both the per-trace (#4888) and batch (#4991) resolvers decoding from it, no coupling introduced in either direction. That's the shape to reuse for the next shared-mechanism extraction in this area.

langwatch/src/server/export/export.service.ts
drewdrewthis: **Correction — my earlier reply on this thread was wrong, and the fix I described did not work.**

I said setting `resolveBlobs: true` unconditionally fixed summary-mode exports. It did not. It was **inert**, and I should have proven it before claiming it.

Blob resolution lives inside `enrichTracesWithSpans`, which only ran:

```ts
if (options.includeSpans && traces.length > 0) {
  traces = await this.enrichTracesWithSpans(..., options.resolveBlobs === true);
}
```

Summary mode sets `includeSpans = false` (`request.mode === "full"`, `export.service.ts:128`). So `options.resolveBlobs` was **never read**, zero `event_log` reads were issued, and the truncated 64 KB preview shipped exactly as before — while the comment I added asserted the opposite. Your blocker was still open; I'd just moved the flag.

The reason my test didn't catch it: it mocked `getAllTracesForProject` and asserted only that the flag was **forwarded**. That assertion passes whether or not the flag has any runtime effect, so it was green the whole time. It was checking my intent, not the behavior.

**Actually fixed in `6860180b9`.** Spans are now fetched when the caller wants spans **OR** wants full IO, because those are not the same thing — the full value is recoverable *only* by de-offloading the spans' eventref pointers and recomputing trace IO from them (`trace_summaries` holds nothing but the preview). A content-consuming summary read therefore fetches and resolves spans, then discards them: it keeps the recomputed trace-level IO and returns `spans: []`, so its payload shape is unchanged.

```ts
const wantsSpans  = options.includeSpans === true;
const wantsFullIo = options.resolveBlobs === true;

if ((wantsSpans || wantsFullIo) && traces.length > 0) {
  const enriched = await this.enrichTracesWithSpans(..., wantsFullIo);
  traces = wantsSpans ? enriched : enriched.map((t) => ({ ...t, spans: [] }));
}
```

**AC5 is preserved by construction:** `resolveBlobs` stays opt-in, so the list grid and the aggregations resolve nothing and issue zero `event_log` reads whether or not they ask for spans. The existing zero-read tests still pass.

**Evidence this time is behavioural, not a flag check.** A new test drives the *real* `ClickHouseTraceService` + *real* batch resolver with `{ includeSpans: false, resolveBlobs: true }` and asserts the **value**:

```
RED  (before the fix): × resolves the trace-level output to the FULL value, not the preview
                        AssertionError: expected "vi.fn()" [getFromEventLog] to be called at least once
GREEN (after):         ✓ 12 passed — full value returned, and AC5 zero-read still holds
```

**Cost, stated plainly:** summary exports now do a span join + bounded `event_log` reads where they previously did neither. That is the tradeoff you explicitly offered on this thread ("accepting the extra `event_log` reads — exports are already a bounded/streamed bulk read via this same PR's batch resolver"), and there is no cheaper way to recover the value. Flagging it so it's a conscious call and not something inherited from a comment.

Thread stays resolved because the bug is now genuinely fixed — but re-open it if you'd rather verify the new shape yourself first.

langwatch/src/server/api/routers/traces.ts
drewdrewthis: **Correction to my reply above — flagging prominently.**

I wrote that the public-share resolution "is not a leak today" / that the response stays filtered so no content escapes. On the resource-amplification vector this PR targets, that holds. But a follow-up review disproved the broader claim, and I **reproduced it**: this endpoint has a **pre-existing content-disclosure vector** — an unvalidated parameter on the public-share path lets an anonymous holder of one share link read *other* publicly-shared traces in the same project. My "not a leak" was **wrong**.

It is **pre-existing** (the shape predates this PR; this PR neither introduces nor fixes it) and out of scope here. Because `langwatch/langwatch` is a public repo, I have **redacted the technical detail** from the public tracking issue — a working exploit should not sit on a public issue for an unpatched live bug. Filed as **P1 [#5747](https://github.com/langwatch/langwatch/issues/5747)**; maintainers should handle it via a **private GitHub Security Advisory**, where full reproduction can be supplied. Thanks to the review pass that caught my error.

Aryansharma28 (approved):
Re-reviewed at HEAD (6440cc9a) after the earlier CodeRabbit + review round. All 15 prior threads resolved; independently re-verified the fixes rather than trusting the resolution:

- Public-share security fix (`traces.ts`): preview-only read → authorize via projectId-scoped `publicShare.findMany` → full-blob resolve only for authorized trace IDs. Correct.
- Summary-export gate now correctly triggers on `wantsSpans || wantsFullIo` instead of `includeSpans` alone, with spans stripped back out for spans-less callers. Correct.
- Batch resolver dedupes by `(aggregateId, eventId, field)` and passes `tenantId: projectId` through — no multitenancy gaps in anything touched here.

**No blockers.** CI green, mergeable.

**Non-blocking, worth a fast-follow (not this PR's scope):**
1. `traces/types.ts:27-28` has a stale doc comment on `resolveBlobs` ("requires includeSpans:true") that now contradicts the fix — cosmetic, but will mislead the next reader.
2. `runEvaluation.ts`'s `getTracesByThreadId` (legacy `/api/evaluations` route + `evaluations.ts` tRPC) is a live content-consuming read path not covered by this PR's 4 scoped call sites. Not a regression — just another candidate for the "all read paths" backlog alongside `spans.ts`'s `getSpanForPromptStudio`, which also bypasses both resolvers.

Approving.

PR #5085, opened 2026-06-24
perf(clickhouse): resolve trace time for hint-less single-trace span reads

langwatch-agent:
## Evidence

Over the last 24h of prod ClickHouse logs, single-trace `stored_spans` reads were the dominant cold-scan source: **~67% of all cold-scan warnings** (cost Signal C) came from drawer span reads keyed only by `(TenantId, TraceId)` with no `StartTime` predicate. Top shapes by 24h count:

- full-span fetch (`getSpansByTraceId` / `getNormalizedSpansByTraceId`): ~290 cold scans
- paginated span/summary reads (carry `count(DISTINCT SpanId)`): ~30 + ~16 cold scans
- span-type/summary derivations: ~28 cold scans

`stored_spans` is `PARTITION BY toYearWeek(StartTime)` and tiered to S3 after the hot window, so a read without a `StartTime` predicate walks every weekly partition incl. the cold S3 tier, bursting S3 GETs. (No tenant data is reproduced here.)

## Suspected cause

The events readers were already fixed (resolve the trace time from `trace_summaries` when no `occurredAtMs` hint is threaded, then bound the read), but the **span readers were not** — they call `withPartitionHint`, which falls back to an *unbounded* scan whenever `occurredAtMs` is `undefined`. The drawer fires these reads from entry points that drop the URL hint (back-stack, conversation jumps, deep links), and worker callers never carry one, so the fallback path runs constantly.

## Proposed fix

Add `readTraceSpans`, a sibling of the existing `readTraceEvents`, and route the single-trace span readers through it:

- **Caller threaded a hint:** unchanged — keep `withPartitionHint` (hinted scan, unbounded fallback on a miss) so a stale URL hint / clock skew still resolves.
- **No hint:** resolve the trace's occurrence time from `trace_summaries` (a `(TenantId, TraceId)` sort-key point seek) and bound the read to its ±2-day span window. The window comes from the trace's own time, so an empty result is authoritative — no unbounded rescan.
- **Trace absent from `trace_summaries`** (orphan / not-yet-projected): keep the previous unbounded behaviour.

Zero behaviour change for the caller-hint path; the only change is replacing the unbounded fallback with a resolved, partition-bounded window when no hint was supplied.

## Expected impact

This is a **cost** fix (S3 GETs), not a latency one — user-visible response time may not move. The headline is partitions pruned on the hint-less path. The extra `trace_summaries` resolve is a sort-key point seek over a couple of granules of small columns, far cheaper than the cold partition walk it replaces.

## EXPLAIN check

`EXPLAIN INDEXES` on the `stored_spans` read for a real large tenant + trace, unbounded vs. the resolved ±2-day window:

| Shape | Parts read | Granules read |
| --- | --- | --- |
| Unbounded (no `StartTime`) | 240 / 240 | 132,505 / 132,505 |
| Resolved window (±2 days) | 10 / 238 | 2,008 / 132,503 |

240 → 10 parts, 132,505 → 2,008 granules.

`EXPLAIN INDEXES` on the `trace_summaries` **resolve** query (`min(OccurredAt)` by `TenantId` + `TraceId`) for a large/old tenant (~249 parts, ~13.3k granules), to confirm the now-more-frequent lookup is cheap:

- It cannot prune partitions (no `OccurredAt` predicate), so all parts' primary index is consulted.
- The `(TenantId, TraceId)` binary search narrows 13,268 → 182 granules (one per part), and a `bloom_filter` skip index `idx_trace_id` narrows those to ~1 granule/part for a real trace.
- Only the light `OccurredAt` column is read from those granules.

So the resolve is a sort-key point-seek plus a bloom filter (same shape as the trace-summary/eval resolve reads already shipped), not a partition data scan — no app-level upper bound needed.

## Test plan

`span-storage.clickhouse.repository.integration.test.ts` (real ClickHouse testcontainer, prod `stored_spans` + `trace_summaries` schema) gains a block exercising the hint-less span read:

- resolves the partition window from `trace_summaries` and still returns the spans;
- a trace whose spans all fall **outside** the resolved ±2-day window (span 5 days past `OccurredAt`) is still returned via the unbounded fallback (bounded read first, then the unbounded one);
- a trace with no spans returns empty via the same fallback (two reads);
- a trace absent from `trace_summaries` stays unbounded from the start (no `StartTime` predicate).

The unit suite's memory-cap assertions pass an explicit `occurredAtMs` so they target the span read directly rather than the new resolve round-trip. Full file: unit 49/49, integration 11/11.

## Notes for reviewer

- New large tenant observed while triaging (worth refreshing the skill's tenant table): `project_yv_gLnh5DP5SebYuCjdw4`.
- Unlike the events read, the span read does **not** treat an empty windowed result as authoritative: `trace_summaries.OccurredAt` is the trace start and never widens, so a long-running trace can produce spans past `OccurredAt + 2d`. The resolved branch therefore keeps `withPartitionHint`'s `isEmpty` → unbounded fallback (parity with the hinted path).
- The two `*Since` poll readers run directly with only their `sinceStartTimeMs` lower bound — they are **not** routed through the resolve helper, so no extra resolve per poll and no upper bound that could hide new spans on a still-active trace.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved span read performance by applying a trace-time based partition window when trace timing hints are missing, reducing unnecessary scanning across partitions.
  * Kept correctness via a bounded-then-unbounded fallback when the first read returns no results.
  * “Since” polling queries no longer use trace-time windowing; they rely on the provided lower-bound time for partition pruning.

* **Tests**
  * Added integration coverage for single-trace span reads with and without trace timing in trace summaries.
  * Updated unit tests to pass the `occurredAtMs` hint where the newer lookup flow expects it.

### DIFF
--- a/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
@@ -411,4 +411,139 @@ describe("SpanStorageClickHouseRepository single-trace reads (integration)", () 
       });
     });
   });
+
+  // The single-trace span readers fire from the same hint-dropping entry points
+  // as the events read (back-stack / conversation jumps / deep links) and from
+  // worker callers that never had an `occurredAtMs`. Without a hint they used to
+  // walk every weekly `stored_spans` partition (incl. cold S3). They now seed
+  // the partition window from the trace's own `trace_summaries.OccurredAt`, the
+  // same way the events read does, and only stay unbounded when the trace isn't
+  // in `trace_summaries` at all.
+  describe("given a span read without an occurredAtMs hint", () => {
+    const hintlessTenantId = `test-span-read-hintless-${nanoid()}`;
+    const withSpansTraceId = `trace-${nanoid()}`;
+    const emptyTraceId = `trace-${nanoid()}`;
+    const orphanTraceId = `trace-${nanoid()}`;
+    const summaryOccurredAt = new Date(base);
+
+    async function insertSummary(tid: string) {
+      await ch.insert({
+        table: "trace_summaries",
+        values: [
+          {
+            ProjectionId: `proj-${nanoid()}`,
+            TenantId: hintlessTenantId,
+            TraceId: tid,
+            Version: "v1",
+            OccurredAt: summaryOccurredAt,
+            CreatedAt: summaryOccurredAt,
+            UpdatedAt: summaryOccurredAt,
+          },
+        ],
+        format: "JSONEachRow",
+        clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+      });
+    }
+
+    beforeAll(async () => {
+      await insertRows([
+        makeSpanRow(0, {
+          TenantId: hintlessTenantId,
+          TraceId: withSpansTraceId,
+          SpanAttributes: { idx: "0" },
+        }),
+      ]);
+      // `withSpansTraceId` + `emptyTraceId` are in trace_summaries (time
+      // resolvable); `orphanTraceId` deliberately is not.
+      await insertSummary(withSpansTraceId);
+      await insertSummary(emptyTraceId);
+    });
+
+    afterAll(async () => {
+      await ch.exec({
+        query:
+          "ALTER TABLE stored_spans DELETE WHERE TenantId = {tenantId:String}",
+        query_params: { tenantId: hintlessTenantId },
+      });
+      await ch.exec({
+        query:
+          "ALTER TABLE trace_summaries DELETE WHERE TenantId = {tenantId:String}",
+        query_params: { tenantId: hintlessTenantId },
+      });
+    });
+
+    it("resolves the partition window from trace_summaries and still returns the spans", async () => {
+      const spans = await repo.getNormalizedSpansByTraceId({
+        tenantId: hintlessTenantId,
+        traceId: withSpansTraceId,
+      });
+
+      expect(spans.map((s) => s.spanId)).toEqual([spanIdFor(0)]);
+    });
+
+    it("returns no spans for a trace without any, without an unbounded rescan", async () => {
+      const storedSpansQueries: { query: string; params: unknown }[] = [];
+      const recordingClient = new Proxy(ch, {
+        get(target, prop, receiver) {
+          if (prop === "query") {
+            return (args: { query: string; query_params?: unknown }) => {
+              if (args.query.includes("stored_spans")) {
+                storedSpansQueries.push({
+                  query: args.query,
+                  params: args.query_params,
+                });
+              }
+              return (target as ClickHouseClient).query(args as never);
+            };
+          }
+          return Reflect.get(target, prop, receiver);
+        },
+      }) as ClickHouseClient;
+      const recordingRepo = new SpanStorageClickHouseRepository(
+        async () => recordingClient,
+      );
+
+      const spans = await recordingRepo.getNormalizedSpansByTraceId({
+        tenantId: hintlessTenantId,
+        traceId: emptyTraceId,
+      });
+
+      // Window resolved from trace_summaries, so empty is authoritative: one
+      // partition-bounded read, no unbounded rescan.
+      expect(spans).toEqual([]);
+      expect(storedSpansQueries).toHaveLength(1);
+      expect(storedSpansQueries[0]!.query).toContain("StartTime >=");
+    });
+
+    it("stays unbounded for a trace that is not in trace_summaries", async () => {
+      // No resolvable time: the reader keeps its previous behaviour and scans
+      // unbounded rather than guessing a window.
+      const storedSpansQueries: string[] = [];
+      const recordingClient = new Proxy(ch, {
+        get(target, prop, receiver) {
+          if (prop === "query") {
+            return (args: { query: string; query_params?: unknown }) => {
+              if (args.query.includes("stored_spans")) {
+                storedSpansQueries.push(args.query);
+              }
+              return (target as ClickHouseClient).query(args as never);
+            };
+          }
+          return Reflect.get(target, prop, receiver);
+        },
+      }) as ClickHouseClient;
+      const recordingRepo = new SpanStorageClickHouseRepository(
+        async () => recordingClient,
+      );
+
+      const spans = await recordingRepo.getNormalizedSpansByTraceId({
+        tenantId: hintlessTenantId,
+        traceId: orphanTraceId,
+      });
+
+      expect(spans).toEqual([]);
+      expect(storedSpansQueries).toHaveLength(1);
+      expect(storedSpansQueries[0]!).not.toContain("StartTime >=");
+    });
+  });
 });
--- a/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
@@ -416,15 +416,23 @@ describe("SpanStorageClickHouseRepository single-trace reads (integration)", () 
   // as the events read (back-stack / conversation jumps / deep links) and from
   // worker callers that never had an `occurredAtMs`. Without a hint they used to
   // walk every weekly `stored_spans` partition (incl. cold S3). They now seed
-  // the partition window from the trace's own `trace_summaries.OccurredAt`, the
-  // same way the events read does, and only stay unbounded when the trace isn't
-  // in `trace_summaries` at all.
+  // the partition window from the trace's own `trace_summaries.OccurredAt` and
+  // read that window first. Unlike the events read, an empty windowed result is
+  // NOT authoritative for spans (OccurredAt is the trace start and never widens,
+  // so a long-running trace can produce spans past OccurredAt + 2 days): the
+  // reader falls back to an unbounded rescan, and only skips the window entirely
+  // when the trace isn't in `trace_summaries` at all.
   describe("given a span read without an occurredAtMs hint", () => {
     const hintlessTenantId = `test-span-read-hintless-${nanoid()}`;
     const withSpansTraceId = `trace-${nanoid()}`;
     const emptyTraceId = `trace-${nanoid()}`;
     const orphanTraceId = `trace-${nanoid()}`;
+    const outOfWindowTraceId = `trace-${nanoid()}`;
     const summaryOccurredAt = new Date(base);
+    // Five days past the summary's OccurredAt — outside the ±2-day resolved
+    // window, so a long-running trace whose late spans land here must still be
+    // returned via the unbounded fallback rather than silently dropped.
+    const outOfWindowStartTime = new Date(base + 5 * 24 * 60 * 60 * 1000);
 
     async function insertSummary(tid: string) {
       await ch.insert({
@@ -452,11 +460,18 @@ describe("SpanStorageClickHouseRepository single-trace reads (integration)", () 
           TraceId: withSpansTraceId,
           SpanAttributes: { idx: "0" },
         }),
+        makeSpanRow(0, {
+          TenantId: hintlessTenantId,
+          TraceId: outOfWindowTraceId,
+          StartTime: outOfWindowStartTime,
+          SpanAttributes: { idx: "0" },
+        }),
       ]);
-      // `withSpansTraceId` + `emptyTraceId` are in trace_summaries (time
-      // resolvable); `orphanTraceId` deliberately is not.
+      // `withSpansTraceId`, `emptyTraceId` and `outOfWindowTraceId` are in
+      // trace_summaries (time resolvable); `orphanTraceId` deliberately is not.
       await insertSummary(withSpansTraceId);
       await insertSummary(emptyTraceId);
+      await insertSummary(outOfWindowTraceId);
     });
 
     afterAll(async () => {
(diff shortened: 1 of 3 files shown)


langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
```diff
@@ -411,4 +411,139 @@ describe("SpanStorageClickHouseRepository single-trace reads (integration)", ()
...
+      const spans = await repo.getNormalizedSpansByTraceId({
+        tenantId: hintlessTenantId,
+        traceId: withSpansTraceId,
+      });
+
+      expect(spans.map((s) => s.spanId)).toEqual([spanIdFor(0)]);
+    });
+
+    it("returns no spans for a trace without any, without an unbounded rescan", async () => {
+      const storedSpansQueries: { query: string; params: unknown }[] = [];
```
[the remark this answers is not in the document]
langwatch-agent: This is a false positive. The three `storedSpansQueries` declarations are in three separate `it()` blocks (lines 382, 485, 521), each its own function scope, so there is no same-scope redeclaration. The file parses and runs: typecheck is clean and the integration suite passes 10/10 (and the CI test-integration shards are green on this PR), which would be impossible with a real `already been declared` error. No change needed.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
```diff
@@ -1136,6 +1136,42 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
...
+    }: { tenantId: string; traceId: string } & OccurredAtHint,
+    isEmpty: (result: T) => boolean,
+    run: (window: PartitionWindow | undefined) => Promise<T>,
+  ): Promise<T> {
+    if (occurredAtMs !== undefined) {
+      return withPartitionHint<T>({ occurredAtMs }, isEmpty, run);
+    }
+    const resolved = await this.resolveTraceOccurredAtMs(tenantId, traceId);
+    if (resolved === undefined) return run(undefined);
+    return run(partitionWindowFor({ occurredAtMs: resolved }));
```
Aryansharma28: [P1] BLOCKER — resolved-time branch has no unbounded fallback; spans outside the window are silently and permanently dropped.

**Why:** `withPartitionHint` (used one branch up, for the caller-hint case) exists precisely because a window built from an approximate time can miss: it runs the hinted query first, and if `isEmpty(hinted)` is true, reruns unbounded. That safety net is real — the docstring above even says "long-running traces and clock skew should still hit the hinted window... callers retry without the hint."

The resolved-time branch here does not have that safety net at all:
```ts
const resolved = await this.resolveTraceOccurredAtMs(tenantId, traceId);
if (resolved === undefined) return run(undefined);
return run(partitionWindowFor({ occurredAtMs: resolved })); // <-- single shot, no isEmpty check, no retry
```
`trace_summaries.OccurredAt` is set once (min over projected rows) and never widens as a trace continues. For any trace whose spans span more than ~4 days from that value (long-running agent loops/conversations — a category this same file's comments explicitly flag as a known risk for the hint path), every hint-less read of that trace is now permanently capped to a window that no longer covers all its spans. There's no error, no log, no retry — the caller just gets fewer/no spans back. `findSpansPaginated` (further down) is worse: its `total` count is computed inside the same windowed query, so pagination UI will show a silently wrong total, not just missing rows.

This is a genuine regression versus the pre-PR behavior, which was always unbounded (slow but 100% correct) on the hint-less path.

**Suggested fix:** Reuse the same `isEmpty`-driven retry for the resolved branch, e.g. route it through `withPartitionHint({ occurredAtMs: resolved }, isEmpty, run)` instead of calling `run` directly — that keeps the cost win (partition-bounded happy path) while restoring the same self-healing behavior the hinted path already has, at the cost of one extra unbounded rescan only when the bounded read comes back empty/incomplete.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
```diff
@@ -1136,6 +1136,42 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
...
+      traceId,
+      occurredAtMs,
+    }: { tenantId: string; traceId: string } & OccurredAtHint,
+    isEmpty: (result: T) => boolean,
+    run: (window: PartitionWindow | undefined) => Promise<T>,
+  ): Promise<T> {
+    if (occurredAtMs !== undefined) {
+      return withPartitionHint<T>({ occurredAtMs }, isEmpty, run);
+    }
+    const resolved = await this.resolveTraceOccurredAtMs(tenantId, traceId);
```
Aryansharma28, line 1177: [P2] BLOCKER (perf, unverified) — `resolveTraceOccurredAtMs` has no time-bound predicate and is now called from 9 additional sites.

**Why:** The query this calls is:
```sql
SELECT toUnixTimestamp64Milli(min(OccurredAt)) AS occurredAtMs
FROM trace_summaries
WHERE TenantId = {tenantId:String}
  AND TraceId = {traceId:String}
```
`trace_summaries` is `PARTITION BY toYearWeek(OccurredAt)` (same migration file, same `${CLICKHOUSE_STORAGE_POLICY_SETTING}` as `stored_spans`, i.e. the same S3-tiering setup this whole PR exists to avoid). There is no `OccurredAt`/time predicate here, only the `ORDER BY (TenantId, TraceId)` sort key — so ClickHouse can do a cheap point-seek *within* each part, but it cannot skip any partition outright, meaning every weekly partition (including cold S3-tiered ones) still needs its index consulted. That's exactly the anti-pattern `dev/docs/best_practices/clickhouse-queries.md` calls out: "Always include StartTime/OccurredAt range in WHERE... without it, ClickHouse scans ALL partitions including cold storage on S3."

This pre-dates this PR (it already backs `readTraceEvents`), but this PR multiplies its call sites from 1 to 10 by routing all 9 span readers through it whenever no hint is present — the exact scenario the PR says is common ("back-stack, conversation jumps, deep links... worker callers never carry one"). The PR's own EXPLAIN INDEXES check (240→10 parts) only measured the `stored_spans` side; there's no equivalent check for what this now-10x-more-frequent `trace_summaries` lookup costs for an old/large tenant.

**Suggested fix:** Run the same EXPLAIN INDEXES check against `trace_summaries` for a large/old tenant before merging, to confirm the net win actually holds. If it's expensive, consider narrowing further (e.g. an app-level upper bound like tenant retention window) rather than assuming a smaller table is automatically a cheap scan.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
```diff
@@ -1570,8 +1606,8 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
       "SpanStorageClickHouseRepository.findSpansSince",
     );
 
-    return withPartitionHint<Span[]>(
-      { occurredAtMs },
+    return this.readTraceSpans<Span[]>(
```
Aryansharma28: [P1] BLOCKER — routing `findSpansSince` through `readTraceSpans` breaks live polling for long-running traces and contradicts the PR's own "no-op" claim.

**Why:** `findSpansSince` already has adequate, correct partition pruning independent of any hint:
```ts
const sinceFilter = "AND StartTime > fromUnixTimestamp64Milli({sinceStartTimeMs:Int64})";
```
The PR description says: "The two `*Since` poll readers were routed through the same helper for uniformity; they already carry a StartTime bound from sinceStartTimeMs, so this is a no-op for them in practice." That's not accurate for the common case this PR is about (no `occurredAtMs` hint, e.g. `spansDelta`/`spanTreeDelta` tRPC procedures called while a drawer is open without a fresh URL hint):

1. Every such poll now issues an *extra* `trace_summaries` round-trip via `resolveTraceOccurredAtMs` before running the actual query — not a no-op, an added query per poll tick.
2. Worse: the resulting `partition.sqlAnd` adds `AND StartTime <= resolvedOccurredAt + 2 days` **on top of** `sinceFilter`. For a trace that is still actively producing spans more than 2 days after its `trace_summaries.OccurredAt`, this upper bound is now in the past relative to "now" — new spans arriving after that point are silently excluded from the delta response, even though `sinceStartTimeMs` alone would have found them correctly before this PR. The live span view for a long-running trace would just... stop updating, with no error.

I checked `readTraceEvents`/its 3 callers — none of them are `Since`-style pollers, so this specific failure mode is novel to this PR, not a pre-existing issue being inherited.

**Suggested fix:** Either don't route the two `*Since` readers through `readTraceSpans` at all (they don't need it — they already have `sinceStartTimeMs`), or make `readTraceSpans` compose the resolved window as a widening/optional constraint rather than a hard upper bound when a `sinceFilter`-equivalent lower bound is already supplied by the caller.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
```diff
@@ -1697,8 +1733,8 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
       "SpanStorageClickHouseRepository.findSpanSummariesSince",
     );
 
-    return withPartitionHint<SpanSummaryRow[]>(
-      { occurredAtMs },
+    return this.readTraceSpans<SpanSummaryRow[]>(
```
Aryansharma28: [P1] BLOCKER — same issue as `findSpansSince` above: `findSpanSummariesSince` gets an extra unconditional `trace_summaries` resolve call plus a `resolvedOccurredAt + 2 days` upper-bound clamp stacked on top of its already-correct `sinceFilter`, which can silently stop the `spanTreeDelta` live view from showing new spans on a long-running trace once it's been running for more than 2 days past its original `trace_summaries.OccurredAt`. See the full explanation on the `findSpansSince` comment — same root cause, same fix (don't route the `*Since` readers through `readTraceSpans`, or don't let the resolved window override an already-sufficient lower-bound filter).

langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
```diff
@@ -411,4 +411,139 @@ describe("SpanStorageClickHouseRepository single-trace reads (integration)", ()
...
+    it("resolves the partition window from trace_summaries and still returns the spans", async () => {
+      const spans = await repo.getNormalizedSpansByTraceId({
+        tenantId: hintlessTenantId,
+        traceId: withSpansTraceId,
+      });
+
+      expect(spans.map((s) => s.spanId)).toEqual([spanIdFor(0)]);
+    });
+
+    it("returns no spans for a trace without any, without an unbounded rescan", async () => {
```
Aryansharma28: [P3] Missing test: a trace with spans *outside* the resolved ±2-day window.

The three new cases here cover has-spans-inside-window, no-spans/empty, and orphan/unbounded — all good — but none cover the scenario the P1 findings above are about: a trace present in `trace_summaries` whose spans include one with `StartTime` outside `[OccurredAt - 2d, OccurredAt + 2d]`. Right now that case would silently return fewer spans than exist, with no test catching it. Worth adding even if the fix stays as-is, so the risk is visible/regression-tested rather than assumed away by the ±2-day "generous" comment.

Aryansharma28 (commented):
## goated-review — deep audit

**Verdict: request changes before merge.** The stored_spans EXPLAIN evidence (240→10 parts, 132,505→2,008 granules) is real and the caller-hinted path is provably unchanged — good work there. But the *new* hint-less code path (`readTraceSpans`) drops the unbounded-fallback safety net that its sibling `withPartitionHint` relies on, and two of the nine rerouted callers (`findSpansSince` / `findSpanSummariesSince`) get a behavior change the PR description explicitly (and incorrectly) calls a no-op. Both are concrete, code-verified findings, not speculation — see inline comments. Existing automated approval (`low-risk-change` bot) and CodeRabbit's only comment (a withdrawn false positive about a duplicate `const`) did not touch any of this; nothing here duplicates prior review signal.

**P0: 0 | P1: 2 | P2: 1 | P3: 2 | P4: 1 | P5: 2**

### P0/P1/P2 — BLOCKERS (3)
- **[P1]** `readTraceSpans`'s resolved-time branch has no unbounded-fallback when the ±2-day window returns empty (or an incomplete page) — unlike `withPartitionHint`, which exists specifically to self-heal a wrong/short window. Spans outside the window are silently and *permanently* dropped (trace_summaries.OccurredAt never moves), across 7 read methods plus a `total` undercount in `findSpansPaginated`.
- **[P1]** `findSpansSince` / `findSpanSummariesSince` already had correct, adequate partition pruning via `sinceStartTimeMs`. Routing them through `readTraceSpans` adds an unconditional extra `trace_summaries` round-trip on every poll *and* clamps the upper bound to `resolvedOccurredAt + 2 days` — which silently hides genuinely new spans from live polling on any trace still producing spans past that point. This directly contradicts the PR description's "this is a no-op for them in practice" claim.
- **[P2]** `resolveTraceOccurredAtMs`'s query against `trace_summaries` has no time-range predicate, so — per this repo's own ClickHouse best practices — it cannot prune partitions and must consult every weekly partition (incl. cold S3-tiered ones) via the sort-key index. This PR raises its call frequency from 1 caller to 10; the PR's EXPLAIN INDEXES benchmark never measured this side of the fix.

### P3–P5 — NON-BLOCKERS (5)
- **[P3]** No test exercises "trace has spans outside the resolved window" — the exact scenario behind both P1s. Currently untested and silently assumed away.
- **[P3]** `readTraceSpans` and `readTraceEvents` are near-duplicate helpers with subtly different hinted-path fallback semantics; the "mirrors" docstring overstates the similarity.
- **[P4]** `readTraceSpans`'s `isEmpty` parameter is silently unused on the resolved-time branch — mildly misleading API surface.
- **[P5]** Good evidence-driven PR description (real EXPLAIN INDEXES before/after numbers), the caller-hinted path is genuinely untouched, and the new/empty/orphan integration tests are well-targeted for what they do cover.
- **[P5]** Cross-PR observation: `readTraceEvents` (unmodified here, from #4874) doesn't currently have a "Since"-style poller sitting on top of it, so it hasn't hit the P1 #2 failure mode yet — worth keeping in mind if one is ever added.

- Actual production frequency/duration of traces exceeding the ±2-day window (agent conversations, long-running workflows) — I couldn't query prod ClickHouse from here, so I can't size how often P1 #1/#2 would actually fire. The code path is real and untested either way.
- Whether `trace_summaries` partitions are tiered to S3 on the same schedule as `stored_spans` (both use the same `${CLICKHOUSE_STORAGE_POLICY_SETTING}` in the migration, which strongly suggests yes, but I didn't have access to the live storage policy config to confirm cold-tier thresholds).

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
langwatch-agent: Fixed in 14907f0b. The resolved-time branch now routes through `withPartitionHint({ occurredAtMs: resolved }, isEmpty, run)` exactly as suggested: bounded read first, unbounded rescan only when it comes back empty. So a long-running trace whose spans all fall past OccurredAt + 2d is recovered by the fallback instead of being silently dropped, matching the hinted path. Events keep their no-fallback behavior (they cluster at the trace start); the docstring now spells out why spans differ. Regression-tested (see the P3 thread).

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
langwatch-agent: Fixed in 14907f0b. `findSpansSince` no longer routes through `readTraceSpans`: it runs directly with only its `StartTime > sinceStartTimeMs` lower bound (already a partition-pruning predicate), so there is no extra trace_summaries resolve per poll and no `StartTime <= OccurredAt + 2d` upper bound. New spans on a trace still active >2d past its OccurredAt are no longer excluded from the delta.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
langwatch-agent: Fixed in 14907f0b, same as findSpansSince: `findSpanSummariesSince` now runs directly with just the `sinceStartTimeMs` lower bound, no resolve call and no upper clamp, so spanTreeDelta keeps showing new spans on long-running traces.

langwatch/src/server/app-layer/traces/repositories/__tests__/span-storage.clickhouse.repository.integration.test.ts
langwatch-agent: Added in 14907f0b. New integration test `returns spans that fall outside the resolved ±2-day window via the unbounded fallback`: seeds a trace in trace_summaries (OccurredAt = base) with a span at base + 5 days, asserts the span is returned and that two stored_spans reads fire (bounded first — contains `StartTime >=` — then the unbounded fallback). Also updated the former empty-trace test to assert the fallback (two reads) rather than the old single authoritative read.

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
langwatch-agent: Ran the EXPLAIN INDEXES check you asked for against `trace_summaries` for a large/old tenant (~249 parts, ~13.3k granules).

The resolve query does read every partition's primary index (no OccurredAt predicate, as you noted), but the actual data read is tiny:
- PrimaryKey `(TenantId, TraceId)` binary search narrows 13268 -> 182 granules (one per part).
- A `bloom_filter` skip index `idx_trace_id` then narrows those to ~0-1 granules per part (0 here only because I used a synthetic non-existent TraceId; for a real trace it keeps ~1 granule/part).
- Only the light `OccurredAt` column is read from those granules.

So it is a point-seek over the sort key plus a bloom filter, not a partition data scan — the same resolve-from-sort-key shape already shipped for trace-summary/eval reads (#5113/#5231/#5112). The bloom index (which the concern didn't account for) is what keeps it cheap, so an app-level retention upper bound isn't needed; adding one would risk missing deep-links to old traces. The heavy `stored_spans` side it saves is still the 240->10 partition win. EXPLAIN excerpt (redacted) added to the PR body.

0xdeafcafe (dismissed):

0xdeafcafe (approved):

PR #5158, opened 2026-06-26
feat(billing): billable stored-bytes measurement (ADR-027 Phase 2)

sergioestebance:
## What

Phase 2 of ADR-027 (per-org stored-GiB-hours billing). Adds the **measurement** — the pure read computing how many logical bytes an org holds older than the free window, as of a sealed hour.

`StorageMeterService.getBillableStorageBytesForOrgAt({ organizationId, sealedHour })`:
- Cutoff = `sealedHour − 35 days` (`BILLABLE_AFTER_DAYS`), bound as an explicit `DateTime('UTC')` param — reproducible regardless of run time or ClickHouse session timezone.
- Sums across the org's projects **one tenant-routed query at a time** — never a cross-tenant `TenantId IN (...)` scan (that multiplies the heavy `byteSize()` recompute buffer across tenants — the two prod-OOM vector).
- Per-table pre-aggregation (`sum(_size_bytes)` per table, UNION ALL'd, summed once) keeps the OOM-prone payload recompute out of the intermediate set. Carries `max_threads: 2` + `max_execution_time: 45`.
- **Throws on any missing dependency or tenant-query failure** — never degrades to 0. The bill must never silently omit a tenant's bytes; the Phase 4 dispatcher owns retry/skip/kill-switch.
- Returns **raw bytes**; the caller rounds to MiB.

## Age correctness (billing can't drift from deletion)

Each table's age predicate is sourced from `TABLE_TTL_CONFIG` via the new shared `buildRetentionAgeColumnExpression`, which **both** the TTL DELETE clause and the billing predicate now use:
- `evaluation_runs` ages on `UpdatedAt` (its retention column), **no** synthesized `ScheduledAt` predicate — `UpdatedAt ≤ cutoff` does not imply `ScheduledAt ≤ cutoff` (nullable/future-dated), so such a predicate would silently drop rows from the bill. Accepted: a full-partition scan on this one table, bounded by the memory caps.
- `event_log` ages on `toDateTime(EventOccurredAt / 1000)` (epoch-ms), identical to its TTL.

## Documented limitations
- `evaluation_runs` bills on mutable `UpdatedAt` → a re-written old row resets its age (accepted under-bill; it's the only column coherent with "billed iff still present").
- Plain `sum(_size_bytes)` (not `FINAL`) → un-merged late-updated old row versions can briefly over-count, bounded and merge-collapsing; the Phase 4.5 tripwire monitors real drift.

## Scope boundary
Pure addition — **no behavior change** until the Phase 4 dispatcher calls this. The €0-by-default invariant only holds in prod once the **separate** retention-floor PR (49→35) ships; that's a go-live dependency, not a build dependency.

## Acceptance criteria → evidence

Spec: `specs/data-retention/storage-billing-measurement.feature`. Unit tests prove the query is *constructed* correctly (right column, cutoff, tenant scoping, settings, error handling); CH *evaluation* semantics are integration-verified (see Follow-ups).

| Scenario (feature) | Evidence |
|---|---|
| Data older than free window billable / within excluded | `billableStorageQuery` — every subquery carries `<ageExpr> <= {cutoff:DateTime('UTC')}` |
| Cutoff anchored to H, not now | `storageMeter.service` — cutoff == H−35d, independent of wall clock |
| Later hour advances cutoff; earlier unchanged (deletion lowers later only) | `storageMeter.service` — H+1h yields a later cutoff; H's is unchanged |
| Interpreted in UTC regardless of session tz | `storageMeter.service` — `{cutoff:DateTime('UTC')}` + string param |
| Org sums across all its projects | `storageMeter.service` — 3 ids → 3 queries, summed = 300 |
| Another org's data never included | `storageMeter.service` — org-A→{p1,p2}, org-B→{p3}; only p1,p2 queried |
| No old data → zero / default-keep paid org → zero | `storageMeter.service` — all subtotals 0 → total 0 |
| Each table aged by its retention/TTL column expression | `billableStorageQuery` — map sourced from `TABLE_TTL_CONFIG`, covers exactly the 11 managed tables |
| Age expr byte-identical to TTL DELETE | shared `buildRetentionAgeColumnExpression` used by both; `event_log` = `toDateTime(EventOccurredAt / 1000)` |
| eval_runs filters UpdatedAt, no synthesized ScheduledAt | `billableStorageQuery` — eval_runs subquery = `toDateTime(UpdatedAt) <=`, no `ScheduledAt` |
| Per-tenant, not cross-tenant IN; each table pre-aggregated; memory cap | `billableStorageQuery` (no `IN`, per-table scalar) + `storageMeter.service` (per-tenant queries, `max_threads:2`+`max_execution_time`) |
| Table query failure fails the measurement (no silent undercount) | `storageMeter.service` — one tenant throws → method throws |
| Returns raw logical bytes, no MiB rounding | `storageMeter.service` — 1572865 in → 1572865 out |

20 unit tests, all green. `pnpm typecheck` clean. Pre-existing ttlReconciler/retention TTL suites (47 tests) still green after the helper extraction.

## Follow-ups (not in this PR)
- **Runtime semantic verification** against warike CH (Tailscale): partition pruning on `stored_spans` (age col = partition key) and the bounded `evaluation_runs` full-scan; €0-default sanity — per the build plan's "Verification beyond unit".
- **`@integration` over-bill dedup** scenario (un-merged ReplacingMergeTree versions) — needs a real CH instance with un-merged parts; deferred to a dedicated integration test.
- **Module decoupling** — extract the pure `TABLE_TTL_CONFIG` + builders out of `ttlReconciler.ts` (which imports `@clickhouse/client`) into a dependency-free `ttlConfig.ts`, so the metering read path doesn't transitively depend on the DDL module. Only 2 consumers; deferred to keep this PR off the sensitive reconciler test surface.

Stacks on langwatch/langwatch#4832 (storage-billing schema).

### DIFF
--- a/langwatch/src/server/app-layer/presets.ts
+++ b/langwatch/src/server/app-layer/presets.ts
@@ -480,6 +480,7 @@ export function initializeDefaultApp(options?: {
   );
   const storageMeterService = new StorageMeterService(
     clickhouseEnabled ? resolveClickHouseClient : null,
+    (organizationId) => organizations.getProjectIds(organizationId),
   );
   const dataRetention: DataRetentionDependencies = {
     policy: dataRetentionPolicyService,
--- a/langwatch/src/server/clickhouse/ttlReconciler.ts
+++ b/langwatch/src/server/clickhouse/ttlReconciler.ts
@@ -1,8 +1,8 @@
 import { createClient } from "@clickhouse/client";
 
 import { createLogger } from "../../utils/logger/server";
-import { parseConnectionUrl } from "./goose";
 import { RETENTION_MANAGED_TABLES } from "../data-retention/retentionPolicy.schema";
+import { parseConnectionUrl } from "./goose";
 
 const logger = createLogger("langwatch:clickhouse:ttl-reconciler");
 
@@ -212,7 +212,8 @@ export function buildDesiredTTLExpression({
   config: TableTTLEntry;
   days: number;
 }): string {
-  const colExpr = config.ttlColumnExpression ?? `toDateTime(${config.ttlColumn})`;
+  const colExpr =
+    config.ttlColumnExpression ?? `toDateTime(${config.ttlColumn})`;
   return `${colExpr} + INTERVAL ${days} DAY TO VOLUME 'cold'`;
 }
 
@@ -232,7 +233,9 @@ export function buildRetentionAgeColumnExpression(
   );
 }
 
-export function buildRetentionTTLExpression(config: TableTTLEntry): string | null {
+export function buildRetentionTTLExpression(
+  config: TableTTLEntry,
+): string | null {
   const colExpr = buildRetentionAgeColumnExpression(config);
   if (!colExpr) return null;
   return `IF(_retention_days > 0, ${colExpr} + toIntervalDay(_retention_days), toDateTime('${INDEFINITE_RETENTION_SENTINEL_DATE}')) DELETE`;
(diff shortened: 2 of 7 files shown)


langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** The number this produces is a bill. A wrong age predicate or a non-FINAL over/under-count means an org is charged wrong money, and the query itself is on the known prod-OOM vector (`byteSize()` recompute on `evaluation_runs`).

- `[Must Check]` `billableStorageQuery.ts` — that every table's age predicate is byte-identical to its TTL DELETE (shared `buildRetentionAgeColumnExpression`), and that the full-partition scan on `evaluation_runs` (aged on `UpdatedAt`) stays inside `max_threads:2` / `max_execution_time:45` for the largest tenant.
- `[Must Check]` `storageMeter.service.ts` — one tenant-routed query per project, never a cross-tenant `TenantId IN (...)`; cutoff anchored to `sealedHour − 35d` independent of wall clock and bound as `DateTime('UTC')`.
- `[Ask Author]` `storageMeter.service.ts` — throw-on-any-failure vs degrade-to-0: confirm no caller path can swallow the throw and bill 0 for a tenant.

- The non-`FINAL` `sum(_size_bytes)` can transiently over-count un-merged old row versions. What bounds the magnitude, and is the Phase 4.5 tripwire the only thing catching a stuck-merge drift?
- Go-live depends on the separate 49→35 retention-floor PR landing first for the €0-by-default invariant. Is there a guard so this can't be wired into the dispatcher before that ships?

sergioestebance:

PR #5162, opened 2026-06-28
fix(traces): bisect single-batch OOM + DRY retry loops + typed orderDirection

### LINKED ISSUE #4202: refactor: harden traces/search OOM resilience — SELECT *, silent skipping, retry dedup
## Context

PR #4201 fixed the immediate ClickHouse OOM and V8 RangeError failures on `/api/traces/search` for large-payload tenants. Several improvement opportunities were identified during review but deferred as out of scope for a bug fix.

## Re-scope (2026-07-29)

**This issue is substantially superseded and is now scoped to item 4 only** (true incremental streaming). Of the original six items: one is done, one is fixed server-side but not yet reachable by any SDK consumer, three are carried by in-flight PR #5162, and item 4 is this issue's sole remaining scope, specified by the Acceptance Criteria below.

| # | Item | Status |
|---|---|---|
| 1 | `SELECT *` in `fetchEvaluationRows` | Done — PR #4641, merged 2026-06-19 |
| 2 | Silent trace skipping | Partial — not done (see item 2 below) |
| 3 | DRY the OOM batch-retry | Carried by in-flight PR #5162 |
| 4 | True incremental streaming | **This issue's remaining scope — see Acceptance Criteria** |
| 5 | Bisection fallback | Carried by in-flight PR #5162 |
| 6 | Narrow `orderDirection` type | Carried by in-flight PR #5162 |

### 1. Replace `SELECT *` in `evaluation_runs` query with explicit column list

`fetchEvaluationRows` uses `SELECT *` which defeats ClickHouse column pruning — reads all columns including heavy ones when only a subset is needed by the mapper. The `Inputs` column is genuinely used, but other unused columns add memory pressure. Narrowing the projection reduces per-query memory and makes the OOM batching less likely to trigger.

**File:** `src/server/traces/clickhouse-trace.service.ts` (`fetchEvaluationRows`)

**Status (2026-07-29): Done.** No `SELECT *` remains in `fetchEvaluationRows` — replaced by `EVALUATION_RUN_COLUMNS_WITH_INPUTS`. Landed in PR #4641 (merged 2026-06-19).

### 2. Silent trace skipping on serialization failure

When `JSON.stringify()` fails for an individual trace during streaming, it's logged server-side but silently dropped from the response. The client receives fewer traces than `pagination.totalHits` indicates with no signal that data was lost. Consider adding a `skippedCount` or `warnings` field to the response — but this is an API shape change that needs design.

**File:** `src/app/api/traces/[[...route]]/app.v1.ts` (streaming serialization loop)

**Status (2026-07-29): Partial — not done.** The server-side fix landed in PR #4714 (commit `c1c850040a`), adding the `skipped` field with a passing test. But the generated TypeScript and Python OpenAPI clients and the committed `openapiLangWatch.json` snapshot are all stale and do not include `skipped` — no SDK consumer can currently read the field. Affected consumers: `python-sdk/src/langwatch/traces.py`, the TypeScript SDK's traces API service and CLI, and `mcp-server/src/langwatch-api.ts`. Fixed-but-unreachable-by-every-consumer is not done. This drift predates this issue and was not caused by it.

### 3. DRY the OOM batch-retry pattern

`fetchTraceSummaryRows` and `fetchEvaluationRows` implement the identical try/catch + batch-retry-on-OOM loop. Could be extracted into a generic `withOomBatchRetry<T>(runQuery, ids, batchSize)` helper. Two call sites — worth doing if a third appears.

**File:** `src/server/traces/clickhouse-trace.service.ts`

**Status (2026-07-29): Carried by in-flight PR #5162.** Out of this issue's remaining scope.

### 4. True incremental streaming for response serialization

The current implementation pre-serializes all traces into a `string[]` array before constructing the `ReadableStream`. All strings are in memory simultaneously. This solves the V8 string concatenation crash but doesn't reduce peak memory. For true per-trace streaming, serialization should happen inside the stream's `pull()` callback.

**File:** `src/app/api/traces/[[...route]]/app.v1.ts`

**Status (2026-07-29): This issue's remaining scope.** Specified by the Acceptance Criteria below.

### 5. No fallback if a single 25-ID batch still OOMs

If one trace within a batch has a payload large enough to OOM ClickHouse on its own, the error propagates uncaught. A recursive bisection strategy (split the failing batch in half, retry each half) would handle this edge case. Currently documented as a known limitation in the code.

**File:** `src/server/traces/clickhouse-trace.service.ts` (`fetchTraceSummaryRows`, `fetchEvaluationRows`)

**Status (2026-07-29): Carried by in-flight PR #5162.** Out of this issue's remaining scope.

### 6. Narrow `orderDirection` type from `string` to `"ASC" | "DESC"`

`orderDirection` is interpolated into SQL in `fetchTraceSummaryRows`. It's safe because it's derived from a validated ternary, but the parameter type is `string` — a future caller could pass arbitrary SQL. Narrowing the type makes the constraint compiler-enforced.

**File:** `src/server/traces/clickhouse-trace.service.ts` (`fetchTraceSummaryRows` parameter)

**Status (2026-07-29): Carried by in-flight PR #5162.** Out of this issue's remaining scope.

## Acceptance Criteria

**Scope note:** item 4 is deliberately narrow — move serialisation into the stream's `pull()`, removing the duplicate string copy. It does not rework the ClickHouse read path; `getAllTracesForProject` stays fully awaited before the stream is constructed, so trace objects remain resident either way. The full streaming rework is a separate, larger ticket.

**Done means:** `/api/traces/search` no longer holds a second full copy of the page as serialised strings — each trace is serialised inside the stream's `pull()` — with the response envelope byte-identical and the `skipped` contract intact.

### AC1 — No second copy: serialisation happens inside `pull()`

After the change, the handler contains no array accumulating `JSON.stringify(...)` results before `new ReadableStream` is constructed. `start(controller)` enqueues exactly one chunk — the opening `{"traces":[` frame — and does not close the stream. Each trace is serialised inside `pull()`, one per invocation.

**Fails if:** any array accumulates serialised strings before the stream is constructed; `start()` enqueues more than the opening frame; `start()` closes the stream; or one `pull()` serialises more than one trace.

**Edge — N=0:** `pull()` is still invoked at least once and emits the trailer; the spy-count-equals-trace-count assertion applies only for N ≥ 1.

**Edge — cancel:** after the consumer cancels mid-body, `pull()` receives no further invocations and no unhandled rejection is raised server-side.

**Evidence:** the diff hunk showing the removed accumulator, plus `pnpm test:unit` assertions — a `pull()` spy whose call count equals trace count on a 50-trace page, is ≥ 1 on a 0-trace page, and stops advancing after `reader.cancel()`. Quote the vitest pass line.

### AC2 — The response envelope is byte-identical to `main`

For the same request and fixture, the branch's body is byte-for-byte identical to main's: comma placement, `pagination` key order (`totalHits`, `scrollId`, then `skipped` only when non-zero), and the `,"schema":…` suffix present only when a projection is active.

**Cases** (each captured on both branches and `diff`ed): N=0; N=1 (no leading or trailing comma); a 50-trace page with a projection; the same page without.

**Fails if:** any `diff` is non-empty, **or** the fixture/seed/`scrollId` differs between captures — a non-empty diff from an uncontrolled input is not a result.

**Evidence:** four `diff` invocations each returning empty, with the capture command quoted.

**Standing clause — schema fidelity:** if this or any future change adds, renames, or removes a response field, the same PR must regenerate the TS and Python OpenAPI clients and update `openapiLangWatch.json` — **and assert the regenerated JSON contains every field the route's zod response schema declares.** Regeneration alone is not the check: commit `c1c850040a` regenerated that file in the same commit that added `skipped` to the zod schema, and the regenerated JSON contains zero occurrences of `skipped`. Quote the field-completeness assertion, not just the diff.

### AC3 — A serialisation failure is caught, counted, and never escapes `pull()`

When `serializeTrace` throws for an individual trace, that trace is skipped, the counter increments, the error is logged server-side, and no error escapes `pull()` — the stream is never errored, the response terminates normally, and the body is well-formed JSON with the trailer intact. Because the skipped count is not final until the last trace, `pagination` must be emitted in the trailer carrying the final count, present only when non-zero.

**Fails if:** an error propagates out of `pull()`; the body fails `JSON.parse`; `skipped` is absent or wrong when a trace fails; `skipped: 0` appears in the common case; or `pagination` is emitted before the final count is known.

**Edge cases** (each a separate fixture, every one asserting the body parses before asserting counts): interior failure (5 traces, #3 throws a circular ref → `traces.length === 4`, `skipped === 1`); first-trace failure (body must not begin `{"traces":[,`); last-trace failure (no dangling comma before the trailer); total failure (body is exactly `{"traces":[],"pagination":{…,"skipped":5}}` and parses).

**Evidence:** `pnpm test:unit` assertions; the circular-reference fixture already exists for the interior case. Quote the vitest pass line.

**Why there is no separate mid-stream-error AC:** within this scope the row source is fully awaited before the stream exists, so it cannot throw mid-body — the only in-`pull()` throw origin is `serializeTrace`. Guaranteeing nothing escapes `pull()` makes the torn-body case unreachable, which is what the four streaming consumers depend on: they receive a structured 500 from the shared error handler today, and that path becomes unreachable once headers are flushed.

## Notes

- **On peak memory — deliberately no AC.** Peak-memory reduction is not measurable in CI today. The repo has no Node-heap instrumentation, and neither test tier can host a heap assertion: integration runs in a single long-lived fork worker accumulating v8 coverage at `--max-old-space-size=8192`; unit runs on `vmThreads` with a 512MB limit and a config comment stating it leaks by design. The falsifiable claim is the structural one in AC1. A memory number is instrumentation work and belongs in its own ticket.
- **Rollback:** no feature flag; rollback is reverting the commit. AC2's byte-identity is what makes that safe — no consumer needs coordinating either direction.
- **Test commands:** mocked-ClickHouse unit tests run under `pnpm test:unit` (= `test:ci`); testcontainers integration tests under `pnpm test:integration`, no in-file env gate, running as a standing CI job.

---

<sub>🤖 **Edited by the planner, on behalf of the fleet.** Authored under the owner's GitHub token for API access — the content is the fleet's, not his. Owner-authored and fleet-authored issues are otherwise indistinguishable in this repo; this footer is the only signal.</sub>

drewdrewthis:
# Related Issue

- Refs #4202
- Follow-up filed: #6277

## Reviewer cockpit

- **Scariest thing** — this code only ever runs against a ClickHouse instance that has *already* reported it is out of memory. Get the bound wrong and a single trace-list read fires thousands of sequential queries at a struggling instance, turning a fast 500 into a multi-minute outage amplifier. The second-scariest is silent: a span-scan window that narrows as the retry bisects would return **fewer spans for the same trace** with no error at all.
- **Start here** — `retryInBatchesWithBisect` in `langwatch/src/server/traces/clickhouse-trace.service.ts`. The whole PR is that one helper plus its three call sites; the budget check is the ~6 lines that decide whether this is a safety feature or a load generator.

## Why

Refs #4202. PR #4201 made `/api/traces/search` retry a memory-blown ClickHouse read in batches of 25. A **single 25-id batch that itself OOMs** still propagated uncaught, failing the whole request.

This PR bisects that batch instead — and, per review, **bounds** the bisection so the recovery cannot itself become the incident.

## What changed

- **Cumulative work budget (`MAX_BISECT_RETRIES = 100`)** → *alternative: a 2-level depth cap* → a depth cap bounds one chunk but not the request: the download path passes `pageSize ?? 10_000`, so 400 chunks each bisecting "only a little" still costs thousands of queries. The budget is closed over the whole call, so it bounds the axis that actually matters, capping a call at `ceil(ids/25) + 100` runs instead of ~19,600 (~39,200 on the join path). **Cost if this call is wrong:** a page that *could* have been recovered by grinding now 500s — deliberately, because that is the pre-bisection behaviour and the backpressure a memory-pressured instance depends on.
- **Chunk-invariant span window** → *alternative: widen the per-chunk margin and log when it narrows* → widening only shrinks the odds of silently dropping spans; it does not remove them. The window is now computed once for the whole id list, so bisecting changes how many queries run, never which rows they can see. **Two real costs, both deliberate:** (1) the two `traces.ts` callers that pass `{from: startDate, to: endDate}` now scan spans over the *user's* date range ±2d instead of the matched traces' own times — for a 30-day filter that is ~5 weekly partitions instead of ~1. The summary read in the same call already paid exactly this, so it is a widening to match, not a new class of scan. (2) When no list-wide range exists at all (caller passed none *and* the sort-key resolve failed open), the span scan is unbounded rather than narrowly-and-wrongly bounded — in a branch that already took the same fallback for its summary read. **I picked correctness over both.** A tighter variant exists — keep the summary-derived window when the batch covers the whole id list, use the list-wide one only for subsets — and I left it out because the extra branch needs a paragraph to justify while this needs a sentence. Say the word if you'd rather have the perf.
- **Recursive single-batch bisection (issue item #5, partial)** — a failing batch is halved and each half retried, **sequentially**, down to a single id. This fixes the *aggregate* case (25 mid-size traces summing over the cap). It does **not** fix one trace heavy enough to OOM alone: bisection locates it and then rethrows, same failed page. Tracked as #6277 — see *Anything surprising?*.
- **DRY the three retry loops (issue item #3)** — `fetchTraceSummaryRows`, `fetchEvaluationRows` and the spans-join each carried a copy of the batch-retry loop; they now share one helper (the join via a Map↔entries adapter). **Typed `orderDirection` (item #6)** — `string` → `"ASC" | "DESC"`.
- **`specs/traces/clickhouse-oom-batch-recovery.feature`** — nine scenarios, every one bound to a test that already passes. Mirrors `specs/analytics/clickhouse-memory-safety.feature`, which covers the same class of problem for the analytics service in the same two layers. See *Anything surprising?* for why it arrived late.

## How it works

```
clickhouse-trace.service.ts
  MAX_BISECT_RETRIES = 100            NEW  cumulative ceiling on bisection-induced runs
  retryInBatchesWithBisect<T>({runQuery, ids, batchSize, projectId})   shared helper
    └ runChunk(chunk, depth)
        try runQuery(chunk)
        on error → rethrow if len<=1, non-OOM, OR budget spent   ← the bound
                 → else charge 2, bisect SEQUENTIALLY, concat
  fetchTraceSummaryRows       ── on OOM → retryInBatchesWithBisect  (re-sort kept in caller)
  fetchEvaluationRows         ── on OOM → retryInBatchesWithBisect
  fetchTracesWithSpansJoined  ── on OOM → retryInBatchesWithBisect  (Map↔entries adapter)
    └ summary + span windows now computed ONCE, above runBatch      ← chunk-invariant
```

The OOM detector (`isClickHouseMemoryLimitError`) and the 25-id `SUMMARY_BATCH_SIZE` are unchanged. The recursion strictly shrinks the chunk toward a length-1 base case, so it terminates on the chunk axis; the budget terminates it on the request axis.

## Human verification

A skeptical reviewer can re-confirm each claim independently:

1. **The budget actually bounds work.** Open `retryInBatchesWithBisect` and replace `if (bisectRetriesSpent + 2 > maxBisectRetries)` with `if (false)`. Run `pnpm test:unit src/server/traces/__tests__/clickhouse-trace.service.unit.test.ts` → *"stops bisecting once the work budget is spent"* fails with **"promise resolved instead of rejecting"** — i.e. uncapped, the read grinds all 300 ids and succeeds. Restore → passes.
2. **The span window is chunk-invariant.** Restore the per-chunk derivation (`Math.min(...occurredAts)` from `summaryRows`) in place of `const spanHintMs = summaryHintMs`. Re-run → *"keeps the span scan window identical across every bisected chunk"* fails, showing the window drifted to `Date.now()` instead of the list-wide range.
3. **The halves really are sequential.** Change the two `await runChunk(...)` lines to a `Promise.all([...])`. Re-run → **7** tests fail, including the dedicated concurrency test.
4. **It works against a real database, not a mock.** `pnpm test:integration src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts` (needs Docker — spins up ClickHouse via testcontainers). Then change `chunk.length <= 1` to `chunk.length <= 99999` in `retryInBatchesWithBisect` to disable bisection and re-run → all 3 fail.
5. **The spec is really bound, not decorative.** `npx tsx langwatch/scripts/check-feature-parity.ts` exits 0 ("all bound"). Delete any one `/** @scenario ... */` line and re-run → exit 1, naming that scenario.
6. **The branch is not stale.** `git merge-base --is-ancestor origin/main HEAD` exits 0.

**Exercised against a real ClickHouse, not only mocks** — *proven*:

#4202 real-ClickHouse OOM + red→green falsifiability

Every pre-existing test on this path *mocks the rejection*, so it can only prove the helper reacts to an error shape we hand it — which is exactly the gap the review named. New `clickhouse-trace-oom-bisect.integration.test.ts` runs the whole chain against a testcontainers ClickHouse:

```
real MEMORY_LIMIT_EXCEEDED -> driver error -> translateClickHouseQueryError
  -> QueryMemoryExceededError -> isClickHouseMemoryLimitError -> bisection
```

The error the service actually caught, from that run:

> `Query memory limit exceeded: would use 643.31 KiB (attempt to allocate chunk of 0.00 B), maximum: 1.00 B: While executing AggregatingTransform.`
> `at parseError (@clickhouse/client/dist/common/error/error.js:42:16)`

| Claim | Status | Evidence |
|---|---|---|
| The fix works against a **real** ClickHouse OOM | **proven** | 3/3 integration green: page recovers in full; recovered page is identical in ids **and order** to the unpressured read; and when nothing fits it gives up bounded. Disabling the bisection turns **all three red** against real ClickHouse |
| Bisection fires on the **prod** error path | **proven** | the integration run above never constructs an error — ClickHouse raises it. Plus a unit test rejecting with `QueryMemoryExceededError` (`code: "query_memory_exceeded"`); every pre-existing bisection test reached the fallback only via the legacy string-match arm |
| Rebased onto current main (was 565 behind) | **proven** | rebased onto `3d3fb6b802`; `merge-base --is-ancestor` clean |
| Proof re-run on today's code, not the stale base | **proven** | control run: pristine `origin/main`, same `node_modules`, **34/34 green** — so every failure below is attributable to this PR, not inherited |
| The budget cap is load-bearing | **proven** | disabled → the read **resolves** instead of rejecting (RED 1/3) |
| The span window fix is load-bearing | **proven** | reverted → window drifts to `Date.now()` (RED 2/3) |
| Sequential halves are enforced | **proven** | `Promise.all` → 7 tests red (RED 3/3) |
| Suite + types green | **proven** | **45/45** on the unit file, **404/404** across all 25 `src/server/traces` unit files, `pnpm typecheck` and `pnpm typecheck:tests` both exit 0 |

On determinism: the integration test picks the `max_memory_usage` ceiling per batch size, so the *trigger* is deterministic rather than dependent on how much memory a given ClickHouse build allocates. Everything downstream of the trigger — the error, its translation, the detection, the recovery, the returned rows — is real.

**The rebase was not a formality — it caught a real regression.** On the stale base the suite was green. Rebased, 4 tests failed, and the control run proves they were this PR's: main's join path now issues an `OccurredAt` sort-key resolve before the summary read, which shifted the query sequence the join bisect test is index-coupled to; that one stale test was consuming the following tests' queued mocks. Fixed, and the index-coupling the review flagged is exactly why it broke.

 This is a backend-only change to the ClickHouse read path — no user-facing UI surface. The gating surface is the trace-list read under genuine ClickHouse memory pressure, and it is exercised as such: real database, real `MEMORY_LIMIT_EXCEEDED`, through the service's real public entry point (`getAllTracesForProject`). Reaching this code path in the running app requires inducing a real OOM, which is what the integration test does — a browser walk of the trace list would exercise the happy path only and prove nothing about this change.

## Anything surprising?

- **This PR does not close #4202 item #5.** For the case that item literally describes — *one* trace heavy enough to OOM alone — behaviour is unchanged: bisection locates the offending id, then rethrows, and the page still fails. What it genuinely fixes is the aggregate case. The graceful path (routing that id into the streaming response's existing `skipped` count) is filed as **#6277**; the reviewer preferred doing it here, but it is a new user-visible contract rather than a bound on this recursion, so it is tracked rather than smuggled in. The body says `Refs #4202`, not `Closes`.
- **The budget makes a previously-recoverable page fail.** Under sustained pressure a read that would have succeeded after ~200 queries now 500s after ~110. That is intended — the reviewer's point is that fast-fail *is* the backpressure — but it is a real behaviour change, not just a safety net.
- **The feature file arrived last, not first.** `CLAUDE.md` says specs come before code and I did not check `specs/` until a late sweep — by which point `specs/analytics/clickhouse-memory-safety.feature` turned out to cover this exact class of problem for a sibling service, in the same two layers this PR had independently converged on. The file now documents delivered behaviour rather than having driven it. Worth saying plainly: it is a spec written after the fact, and it is only as good as the tests it is bound to.
- **CI's `evaluate` job is red for reasons unrelated to this PR.** The "Auto-approve for validated Low-Risk or Firefighting" workflow calls OpenAI and the key is returning `insufficient_quota`; it is failing simultaneously on unrelated branches (`worktree-fix+langy-5741-retro`, `issue3439/python-error-surface`). It is **not** one of `main`'s required checks (`validate title`, `typecheck`, `build`), so it does not gate this PR — but it will keep showing red until the billing side is sorted.
- **Two pre-existing formatting violations were left alone.** `biome@2.4.14` (the version `biome.json`'s schema declares; the lockfile installs 2.5.1) flags 3 spots in this test file, all on lines that predate this PR. Only the lines this PR adds were formatted, to keep the diff reviewable.

### DIFF
--- a/langwatch/src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts
+++ b/langwatch/src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts
@@ -1,0 +1,278 @@
+/**
+ * Integration tests for the OOM batch-retry / bisection fallback.
+ *
+ * The unit tests around this path mock the rejection, so they can only prove
+ * the helper reacts to an error SHAPE we hand it. These exercise the real
+ * chain end to end against a testcontainers ClickHouse:
+ *
+ *   real ClickHouse MEMORY_LIMIT_EXCEEDED
+ *     -> the driver's own error
+ *     -> translateClickHouseQueryError / QueryMemoryExceededError
+ *     -> isClickHouseMemoryLimitError
+ *     -> retryInBatchesWithBisect
+ *
+ * so a regression anywhere in that chain — a changed error class, a detector
+ * that stops unwrapping, a swallowed rejection — fails here even though every
+ * mock-based test stays green.
+ *
+ * How the OOM is triggered: a proxy client injects `max_memory_usage` into the
+ * heavy summary read, tightly for large batches and generously for small ones.
+ * The ceiling is chosen per batch size deliberately, so the test is
+ * deterministic rather than dependent on how much memory a given ClickHouse
+ * build happens to allocate. Everything downstream of the trigger — the error,
+ * its translation, the detection, the recovery, the returned rows — is real.
+ */
+
+import type { ClickHouseClient } from "@clickhouse/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  startTestContainers,
+  stopTestContainers,
+} from "../../event-sourcing/__tests__/integration/testContainers";
+import { ClickHouseTraceService } from "../clickhouse-trace.service";
+import type { GetAllTracesForProjectInput } from "../types";
+import { openProtections } from "./open-protections";
+
+const tenantId = `test-oom-bisect-${nanoid()}`;
+const now = Date.now();
+
+/** Big enough that a 25-row read genuinely needs room, small enough to insert fast. */
+const PAYLOAD = JSON.stringify({
+  type: "text",
+  value: "x".repeat(200_000),
+});
+
+const TRACE_COUNT = 30;
+const traceIds = Array.from(
+  { length: TRACE_COUNT },
+  (_, i) => `trace-oom-${String(i).padStart(3, "0")}-${nanoid()}`,
+);
+
+function makeTraceSummaryRow(traceId: string, index: number) {
+  return {
+    ProjectionId: `proj-${nanoid()}`,
+    TenantId: tenantId,
+    TraceId: traceId,
+    Version: "v1",
+    Attributes: {},
+    // Spread across a minute so the page has a deterministic sort order.
+    OccurredAt: new Date(now - index * 1000),
+    CreatedAt: new Date(now - index * 1000),
+    UpdatedAt: new Date(now - index * 1000),
+    ComputedIOSchemaVersion: "v1",
+    ComputedInput: PAYLOAD,
+    ComputedOutput: PAYLOAD,
+    TimeToFirstTokenMs: null,
+    TimeToLastTokenMs: null,
+    TotalDurationMs: 100,
+    TokensPerSecond: null,
+    SpanCount: 1,
+    ContainsErrorStatus: false,
+    ContainsOKStatus: true,
+    ErrorMessage: null,
+    Models: [],
+    TotalCost: null,
+    TokensEstimated: false,
+    TotalPromptTokenCount: null,
+    TotalCompletionTokenCount: null,
+    OutputFromRootSpan: false,
+    OutputSpanEndTimeMs: 0,
+    BlockedByGuardrail: false,
+    SatisfactionScore: null,
+    TopicId: null,
+    SubTopicId: null,
+    HasAnnotation: null,
+  };
+}
+
+function makeQueryInput(
+  overrides: Partial<GetAllTracesForProjectInput> = {},
+): GetAllTracesForProjectInput {
+  return {
+    projectId: tenantId,
+    startDate: now - 600_000,
+    endDate: now + 600_000,
+    filters: {},
+    pageSize: TRACE_COUNT,
+    ...overrides,
+  };
+}
+
+/**
+ * Wrap the real client so the heavy summary read runs under a memory ceiling.
+ *
+ * `limitFor` receives the batch's id count and returns the `max_memory_usage`
+ * bytes for that query, or null to leave it unbounded. Only the heavy read is
+ * touched — the count and id queries must still succeed for the page to get as
+ * far as the summary read at all.
+ */
+function memoryCappedClient(
+  base: ClickHouseClient,
+  limitFor: (batchSize: number) => number | null,
+): { client: ClickHouseClient; batchSizes: number[] } {
+  const batchSizes: number[] = [];
+  const client = new Proxy(base, {
+    get(target, prop, receiver) {
+      if (prop !== "query") return Reflect.get(target, prop, receiver);
+      return (args: {
+        query: string;
+        query_params?: Record<string, unknown>;
+        clickhouse_settings?: Record<string, unknown>;
+      }) => {
+        const ids = args.query_params?.pageTraceIds;
+        const isHeavySummaryRead =
+          typeof args.query === "string" &&
+          args.query.includes("ts_ComputedInput") &&
+          Array.isArray(ids);
+
+        if (!isHeavySummaryRead) {
+          return (target as ClickHouseClient).query(args as never);
+        }
+
+        batchSizes.push((ids as string[]).length);
+        const limit = limitFor((ids as string[]).length);
+        return (target as ClickHouseClient).query({
+          ...args,
+          clickhouse_settings: {
+            ...(args.clickhouse_settings ?? {}),
+            ...(limit === null ? {} : { max_memory_usage: String(limit) }),
+          },
+        } as never);
+      };
+    },
+  }) as ClickHouseClient;
+  return { client, batchSizes };
+}
+
+vi.mock("~/server/clickhouse/clickhouseClient", () => ({
+  getClickHouseClientForProject: vi.fn(),
+}));
+
+vi.mock("~/server/db", () => ({
+  prisma: { project: { findUnique: vi.fn().mockResolvedValue({}) } },
+}));
+
+let ch: ClickHouseClient;
+let service: ClickHouseTraceService;
+let getClickHouseClientForProject: ReturnType<typeof vi.fn>;
+
+beforeAll(async () => {
+  const containers = await startTestContainers();
+  ch = containers.clickHouseClient;
+
+  const chModule = await import("~/server/clickhouse/clickhouseClient");
+  getClickHouseClientForProject = vi.mocked(
+    chModule.getClickHouseClientForProject,
+  );
+  getClickHouseClientForProject.mockResolvedValue(ch);
+
+  const { prisma } = await import("~/server/db");
+  service = new ClickHouseTraceService(
+    prisma as ConstructorParameters<typeof ClickHouseTraceService>[0],
+  );
+
+  await ch.insert({
+    table: "trace_summaries",
+    values: traceIds.map((id, i) => makeTraceSummaryRow(id, i)),
+    format: "JSONEachRow",
+    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+  });
+}, 120_000);
+
+afterAll(async () => {
+  if (ch) {
+    await ch.exec({
+      query: `ALTER TABLE trace_summaries DELETE WHERE TenantId = {tenantId:String}`,
+      query_params: { tenantId },
+    });
+  }
+  await stopTestContainers();
+});
+
+describe("ClickHouse OOM bisection (integration)", () => {
+  describe("getAllTracesForProject()", () => {
+    describe("given a real MEMORY_LIMIT_EXCEEDED on the full-list summary read", () => {
+      it("recovers the whole page by bisecting until the batches fit", async () => {
+        // Tight for anything over 12 ids, generous below — so the full list of
+        // 30 OOMs, the 25-id batch OOMs, and the bisected halves succeed.
+        const { client, batchSizes } = memoryCappedClient(ch, (size) =>
+          size > 12 ? 1_000_000 : null,
+        );
+        getClickHouseClientForProject.mockResolvedValue(client);
+        try {
+          const result = await service.getAllTracesForProject(
+            makeQueryInput(),
+            openProtections,
+          );
+
+          // Every trace still comes back: the recovery is lossless, not
+          // best-effort.
+          expect(result).not.toBeNull();
+          expect(result!.groups.flat()).toHaveLength(TRACE_COUNT);
+        } finally {
+          getClickHouseClientForProject.mockResolvedValue(ch);
+        }
+
+        // The shape of the descent, from the real run: the full list, then
+        // fixed-size batching, then a bisected pair that fits.
+        expect(batchSizes[0]).toBe(TRACE_COUNT);
+        expect(batchSizes).toContain(25);
+        expect(batchSizes.some((n) => n <= 12 && n > 0)).toBe(true);
+      }, 120_000);
+
+      it("returns rows identical to the same read without any memory pressure", async () => {
+        const unpressured = await service.getAllTracesForProject(
+          makeQueryInput(),
+          openProtections,
+        );
+
+        const { client } = memoryCappedClient(ch, (size) =>
+          size > 12 ? 1_000_000 : null,
+        );
+        getClickHouseClientForProject.mockResolvedValue(client);
+        let bisected: Awaited<
+          ReturnType<typeof service.getAllTracesForProject>
+        >;
+        try {
+          bisected = await service.getAllTracesForProject(
+            makeQueryInput(),
+            openProtections,
+          );
+        } finally {
+          getClickHouseClientForProject.mockResolvedValue(ch);
+        }
+
+        // The point of the re-sort in the caller: batches come back in chunk
+        // order, so without it the recovered page would be ordered differently
+        // from the happy path. Ids AND order must match exactly.
+        const idsOf = (
+          r: Awaited<ReturnType<typeof service.getAllTracesForProject>>,
+        ) => r!.groups.flat().map((t) => t.trace_id);
+        expect(idsOf(bisected)).toEqual(idsOf(unpressured));
+      }, 120_000);
+    });
+
+    describe("given every batch size still OOMs", () => {
+      it("gives up instead of grinding, and surfaces the memory error", async () => {
+        // A ceiling nothing can satisfy: the descent reaches a single id,
+        // which cannot be split further, so the error propagates.
+        const { client, batchSizes } = memoryCappedClient(ch, () => 1);
+        getClickHouseClientForProject.mockResolvedValue(client);
+        try {
+          await expect(
+            service.getAllTracesForProject(makeQueryInput(), openProtections),
+          ).rejects.toThrow();
+        } finally {
+          getClickHouseClientForProject.mockResolvedValue(ch);
+        }
+
+        // Bounded work, not a runaway: the fan-out stays far below the
+        // ceil(30/25) + MAX_BISECT_RETRIES ceiling rather than exploring the
+        // whole tree before failing.
+        expect(batchSizes.length).toBeLessThanOrEqual(2 + 100);
+        expect(batchSizes).toContain(1);
+      }, 120_000);
+    });
+  });
+});
--- a/langwatch/src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts
+++ b/langwatch/src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts
@@ -193,6 +193,7 @@ afterAll(async () => {
 describe("ClickHouse OOM bisection (integration)", () => {
   describe("getAllTracesForProject()", () => {
     describe("given a real MEMORY_LIMIT_EXCEEDED on the full-list summary read", () => {
+      /** @scenario A page still loads when the database runs out of memory */
       it("recovers the whole page by bisecting until the batches fit", async () => {
         // Tight for anything over 12 ids, generous below — so the full list of
         // 30 OOMs, the 25-id batch OOMs, and the bisected halves succeed.
@@ -221,6 +222,7 @@ describe("ClickHouse OOM bisection (integration)", () => {
         expect(batchSizes.some((n) => n <= 12 && n > 0)).toBe(true);
       }, 120_000);
 
+      /** @scenario A page recovered from a memory limit matches an unaffected read */
       it("returns rows identical to the same read without any memory pressure", async () => {
         const unpressured = await service.getAllTracesForProject(
           makeQueryInput(),
@@ -254,6 +256,7 @@ describe("ClickHouse OOM bisection (integration)", () => {
     });
 
     describe("given every batch size still OOMs", () => {
+      /** @scenario A trace that cannot be read alone fails the page instead of retrying forever */
       it("gives up instead of grinding, and surfaces the memory error", async () => {
         // A ceiling nothing can satisfy: the descent reaches a single id,
         // which cannot be split further, so the error propagates.
--- a/langwatch/src/server/traces/__tests__/clickhouse-trace.service.unit.test.ts
+++ b/langwatch/src/server/traces/__tests__/clickhouse-trace.service.unit.test.ts
@@ -761,6 +761,99 @@ describe("ClickHouseTraceService", () => {
           service.getAllTracesForProject(baseInput, protections),
         ).rejects.toThrow("SYNTAX_ERROR");
       });
+
+      it("bisects a 25-ID batch that still OOMs and resolves both halves", async () => {
+        const traceIds = Array.from({ length: 30 }, (_, i) => `trace-${i}`);
+        const summaryRows = traceIds.map((id, i) => ({
+          ...makeSummaryRow(id),
+          ts_OccurredAt: Date.now() - i * 1000,
+        }));
+        const idRows = traceIds.map((id) => ({ TraceId: id }));
+
+        mockClickHouseQuery
+          // count
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve([{ total: String(traceIds.length) }]),
+          })
+          // IDs
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(idRows),
+          })
+          // summary full list — OOM, drops to fixed-size batches
+          .mockRejectedValueOnce(new Error("MEMORY_LIMIT_EXCEEDED"))
+          // first 25-ID batch — STILL OOMs, triggers recursive bisection
+          .mockRejectedValueOnce(new Error("MEMORY_LIMIT_EXCEEDED"))
+          // bisected lower half (ids 0-11)
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(summaryRows.slice(0, 12)),
+          })
+          // bisected upper half (ids 12-24)
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(summaryRows.slice(12, 25)),
+          })
+          // second fixed-size batch (ids 25-29)
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(summaryRows.slice(25)),
+          })
+          // evaluations
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve([]),
+          });
+
+        const service = new ClickHouseTraceService({
+          project: { findUnique: mockPrismaFindUnique },
+        } as never);
+
+        const result = await service.getAllTracesForProject(
+          { ...baseInput, pageSize: 30 } as GetAllTracesForProjectInput,
+          protections,
+        );
+
+        expect(result).not.toBeNull();
+        expect(result!.groups.flat()).toHaveLength(30);
+        // calls: 0=count, 1=IDs, 2=full OOM, 3=25-ID batch OOM, 4 & 5 = halves
+        expect(
+          mockClickHouseQuery.mock.calls[3]![0].query_params.pageTraceIds,
+        ).toHaveLength(25);
+        expect(
+          mockClickHouseQuery.mock.calls[4]![0].query_params.pageTraceIds,
+        ).toHaveLength(12);
+        expect(
+          mockClickHouseQuery.mock.calls[5]![0].query_params.pageTraceIds,
+        ).toHaveLength(13);
+      });
+
+      it("rethrows when a single-ID summary batch still OOMs after bisecting", async () => {
+        const traceIds = ["trace-0", "trace-1"];
+        const idRows = traceIds.map((id) => ({ TraceId: id }));
+
+        mockClickHouseQuery
+          // count
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve([{ total: String(traceIds.length) }]),
+          })
+          // IDs
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(idRows),
+          })
+          // every summary read OOMs: full list, the batch, then the bisected
+          // single id — which can no longer be split, so it propagates.
+          .mockRejectedValue(new Error("MEMORY_LIMIT_EXCEEDED"));
+
+        const service = new ClickHouseTraceService({
+          project: { findUnique: mockPrismaFindUnique },
+        } as never);
+
+        await expect(
+          service.getAllTracesForProject(
+            { ...baseInput, pageSize: 2 } as GetAllTracesForProjectInput,
+            protections,
+          ),
+        ).rejects.toThrow("MEMORY_LIMIT_EXCEEDED");
+        // count, IDs, full-list OOM, 2-ID batch OOM, then the bisected 1-ID OOM
+        // — proof the helper descended to a single id before giving up.
+        expect(mockClickHouseQuery).toHaveBeenCalledTimes(5);
+      });
     });
 
     describe("when ClickHouse MEMORY_LIMIT_EXCEEDED on evaluations query", () => {
@@ -801,6 +894,96 @@ describe("ClickHouseTraceService", () => {
 
         expect(result).not.toBeNull();
         expect(result!.groups.flat()).toHaveLength(1);
+      });
+
+      it("bisects a 25-ID evaluations batch that still OOMs and resolves both halves", async () => {
+        const traceIds = Array.from({ length: 30 }, (_, i) => `trace-${i}`);
+        const summaryRows = traceIds.map((id) => makeSummaryRow(id));
+        const idRows = traceIds.map((id) => ({ TraceId: id }));
+
+        mockClickHouseQuery
+          // count
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve([{ total: String(traceIds.length) }]),
+          })
+          // IDs
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(idRows),
+          })
+          // summary — succeeds in one read
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(summaryRows),
+          })
+          // evaluations full list — OOM, drops to fixed-size batches
+          .mockRejectedValueOnce(new Error("MEMORY_LIMIT_EXCEEDED"))
+          // first 25-ID eval batch — STILL OOMs, triggers recursive bisection
+          .mockRejectedValueOnce(new Error("MEMORY_LIMIT_EXCEEDED"))
+          // bisected lower half (12 ids)
+          .mockResolvedValueOnce({ json: () => Promise.resolve([]) })
+          // bisected upper half (13 ids)
+          .mockResolvedValueOnce({ json: () => Promise.resolve([]) })
+          // second fixed-size eval batch (5 ids)
+          .mockResolvedValueOnce({ json: () => Promise.resolve([]) });
+
+        const service = new ClickHouseTraceService({
+          project: { findUnique: mockPrismaFindUnique },
+        } as never);
+
+        const result = await service.getAllTracesForProject(
+          { ...baseInput, pageSize: 30 } as GetAllTracesForProjectInput,
+          protections,
+        );
+
+        expect(result).not.toBeNull();
+        expect(result!.groups.flat()).toHaveLength(30);
+        // calls: 0=count, 1=IDs, 2=summary, 3=full eval OOM, 4=25-ID OOM, 5 & 6 = halves
+        expect(
+          mockClickHouseQuery.mock.calls[4]![0].query_params.traceIds,
+        ).toHaveLength(25);
+        expect(
+          mockClickHouseQuery.mock.calls[5]![0].query_params.traceIds,
+        ).toHaveLength(12);
+        expect(
+          mockClickHouseQuery.mock.calls[6]![0].query_params.traceIds,
+        ).toHaveLength(13);
+      });
+
+      it("rethrows when a single-ID evaluations batch still OOMs after bisecting", async () => {
+        const summaryRows = [
+          makeSummaryRow("trace-0"),
+          makeSummaryRow("trace-1"),
+        ];
+        const idRows = [{ TraceId: "trace-0" }, { TraceId: "trace-1" }];
+
+        mockClickHouseQuery
+          // count
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve([{ total: "2" }]),
+          })
+          // IDs
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(idRows),
+          })
+          // summary — succeeds
+          .mockResolvedValueOnce({
+            json: () => Promise.resolve(summaryRows),
+          })
+          // every evaluations read OOMs, down to the single-id batch which can
+          // no longer be split — so it propagates.
+          .mockRejectedValue(new Error("MEMORY_LIMIT_EXCEEDED"));
+
+        const service = new ClickHouseTraceService({
+          project: { findUnique: mockPrismaFindUnique },
+        } as never);
+
+        await expect(
+          service.getAllTracesForProject(
+            { ...baseInput, pageSize: 2 } as GetAllTracesForProjectInput,
+            protections,
+          ),
+        ).rejects.toThrow("MEMORY_LIMIT_EXCEEDED");
+        // count, IDs, summary, full eval OOM, 2-ID batch OOM, bisected 1-ID OOM
+        expect(mockClickHouseQuery).toHaveBeenCalledTimes(6);
       });
     });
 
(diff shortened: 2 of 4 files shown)


drewdrewthis:
## Review verdict: READY



Reviewed for **correctness** — the real risk surface here is the OOM error path that rarely runs in prod. Verdict: **no correctness defects**.

**Confirmed:**
- **Termination is proven.** `runChunk` rethrows at `chunk.length <= 1 || !isOOM`; for `n ≥ 2`, both halves (`mid` and `n−mid`) are strictly `< n`, so recursion always shrinks to the length-1 base case (max depth ≈ log₂25 ≈ 5). Empty ids → `[]`; non-OOM mid-bisection → propagates; OOM on length-1 → rethrown.
- **Behaviour-identical on the happy path and the first-level batch path** (the helper runs only inside the catch; the outer loop reproduces the old fixed-size concatenation order feeding the same downstream sort/Map build). The ONLY new behaviour is the intended fix: a 25-batch that itself OOMs now bisects instead of failing the page.
- **Sequential bisection** (lower then upper) is the right call — parallel halves would re-pressure the just-OOMed instance. Row order is safe (summary re-sorts; evals bucket by TraceId; join returns a Map).
- **Map↔entries adapter is lossless** — per-batch id sets are disjoint, so `new Map(merged)` has zero key collisions and drops no span.
- **`orderDirection` narrowing safe** — the sole caller passes the `"ASC" | "DESC"` literal ternary.
- **The 5 new tests assert the correct call sequence** (verified by actually running them green): summary/eval bisect 25→12/13, single-id rethrow counts 5/6.

**Applied the 2 optional review nits:** a `logger.warn` on the bisection path (observability on the rare memory-stress path — previously invisible) and a `batchSize <= 0` guard (latent infinite-loop footgun in the generic helper). The helper became an instance method to reach the instance `logger`, consistent with its sibling fetch methods.

**Proof:** backend falsifiability red→green embedded in the PR body (revert the service → 5 bisection tests fail; restore → 34/34 pass). `pnpm typecheck` clean. Items #1/#2 from the issue were already on main and untouched. Backend-only — no UI surface. CI green.

langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** A 25-id OOM now fans out into many *sequential* CH round-trips (bisect to length 1); on a memory-pressured instance this multiplies request latency and could blow a request/gateway timeout even when it eventually succeeds.

- `[Must Check]` `clickhouse-trace.service.ts` `retryInBatchesWithBisect` — confirm the rethrow guard (`chunk.length <= 1 || !isOOM`) and the `slice(0,mid)`/`slice(mid)` recursion always shrink toward the length-1 base case so it terminates.
- `[Must Check]` spans-join Map↔entries adapter (`async (ids) => [...(await runBatch(ids))]` then `new Map(merged)`) — verify per-batch id sets are truly disjoint so no trace entry is silently dropped on rebuild.
- `[Probably Fine]` `fetchTraceSummaryRows` re-sort after merge — batches return in chunk order; confirm the caller re-imposes ORDER BY across the merged set.

- Sequential bisection can turn one heavy page into a long serial chain of CH queries; is there a per-request timeout this could exceed before the halves resolve?

sergioestebance:
## Ruthless review — PR #5162

**Verdict: blocked on one P1.** The bisection logic itself is correct — termination is sound (`chunk.length <= 1` base case, both halves strictly smaller), the Map↔entries adapter is genuinely lossless (`runBatch` can only ever key the Map from `summaryRows`, which are filtered by `TraceId IN {batchTraceIds}`, so it cannot emit a foreign id), the summary re-sort is preserved, and neither the summary nor the evaluations SQL carries a `LIMIT`/offset that would change meaning when partitioned. The spans query uses `LIMIT n BY t.TraceId`, which is per-trace and batch-invariant. The DRY collapse and the `orderDirection` narrowing are clean wins.

What blocks it is not the logic — it's the **absence of a bound on how much work the new recursion can generate against a ClickHouse instance that, by definition, is already out of memory when this path runs.**

---

### P1 — Unbounded retry fan-out removes the backpressure that protected a memory-pressured cluster

`clickhouse-trace.service.ts:1794-1822`

`retryInBatchesWithBisect` has no cumulative work budget, no recursion-depth cap, no request deadline, and the retried queries carry no `max_execution_time` / abort signal (grep confirms none in this path).

The arithmetic:

- The download procedure passes `pageSize: input.pageSize ?? 10_000` (`server/api/routers/traces.ts:578`) into `getAllTracesForProject`, so `fetchTraceSummaryRows` can receive **10,000 ids** → 400 chunks of 25.
- **Before this PR**, a chunk that OOMed threw immediately out of the `for` loop. Ceiling: 400 queries, and in practice it aborted at the *first* failing chunk.
- **After this PR**, each chunk can expand into a full recursion tree — `2·25 − 1 = 49` chunk-runs. On the join path each chunk-run is *two* ClickHouse round trips (`clickhouse-trace.service.ts:2471` summary + `:2553` spans), so **≤98 round trips per chunk**.
- Ceiling for one request: **~19,600 sequential queries** (summary path) / **~39,200** (join path). All sequential, by design (correctly so — parallel halves would be worse).

The trigger is not exotic. The interesting failure mode is *server-wide* memory pressure — the limit is hit because other queries are consuming the instance's budget, not because any one trace is intrinsically fat. That is the shape of every ClickHouse memory incident this repo has had. In that regime chunks fail at 25 and succeed when small, which is exactly the case that maximises the fan-out: the loop does **not** abort early, it grinds through all 400 chunks.

Net effect: a request that previously 500'd in seconds now spends many minutes issuing thousands of queries at an instance that just told us it was out of memory. Fast-fail *is* the backpressure, and this removes it. Same amplification hits the logs — `this.logger.warn` at `:1805` fires once per bisection level, so a bad page emits on the order of 10⁴ warn lines.

**Minimal fix** (~5 lines, no redesign): close a counter over `retryInBatchesWithBisect` and rethrow once the total chunk-runs exceed a budget (a few hundred), or cap recursion depth at 2 levels (25 → 12 → 6, then give up). Either bound restores the fast-fail while keeping the fix for the case the PR is actually about. Setting `max_execution_time` on the retried queries would be belt-and-braces.

---

### P2 — Bisection tightens the batch-derived span window, silently shrinking span sets

`clickhouse-trace.service.ts:2535-2551`

Inside `runBatch`, the `stored_spans` scan window is derived from **that chunk's own summary rows**:

```ts
spanFromMs: Math.min(...occurredAts) - SPAN_PARTITION_WINDOW_MS,  // ±2 days
spanToMs:   Math.max(...occurredAts) + SPAN_PARTITION_WINDOW_MS,
```

The window therefore narrows monotonically as the chunk shrinks. Bisection now drives the chunk down to length 1, at which point the window is a *single* trace's `OccurredAt ± 2d`. Any trace whose spans fall outside that (long-running sessions, replays, clock skew) loses those spans — with no error and no warning; the trace comes back with a partial or empty `spans[]`.

This is pre-existing at 25-id chunks, so it is an amplification rather than a new defect, and it rests on the assumption stated at `:2468-2470` ("a span's StartTime always falls within its trace's lifetime"). But the PR makes the window as tight as it can possibly get, which is precisely when that assumption bites. Nothing in the new suite covers it.

**Fix direction:** hoist the span window out of `runBatch` — compute it once from the caller's `occurredAt` range (or the full-set summaries) and pass it down, so it is invariant across chunks. Failing that, widen the margin and log when a chunk's window is narrower than the full-list attempt's.

---

### P2 — For the exact scenario issue #4202 item #5 names, the outcome is unchanged

`clickhouse-trace.service.ts:1798`

Item #5 is worded as *"if one trace within a batch has a payload large enough to OOM ClickHouse on its own."* In that case this helper bisects down, finds the offending id — and then rethrows, failing the whole page anyway. The user-visible outcome is identical to `main`; the only difference is ~11 extra sequential queries before the same 500. The bisection genuinely fixes the *aggregate* case (25 mid-size traces summing over the cap), which is worth having, but the PR description and the code comment both read as though the single-heavy-trace limitation is resolved, and it isn't.

The graceful path already exists and is already wired: the streaming response has a `skipped` count (`app/api/traces/[[...route]]/app.v1.ts:104,233-255`) used today for serialization failures. A length-1 OOM is the same class of event — one trace we cannot serve — and could drop that id into `skipped` rather than fail the page.

**NEEDS USER DECISION.** Two defensible reads:
- **(a) Ship as-is, retitle.** Land the aggregate-case fix, drop the claim that single-batch OOM is handled, and leave graceful degradation to a follow-up. Cheapest, honest.
- **(b) Extend to `skipped`.** Wire the length-1 OOM into the existing skipped-count contract so a single unreadable trace degrades the page instead of killing it. This is what actually closes item #5, and it composes well with the P1 budget (a budget exhaustion is also a "skip", not a crash).

I'd take **(b)** — it's a small delta on top of a contract that already exists, and it's the difference between "we retry harder before failing" and "the page survives." If you take (a), the PR should not say `Closes #4202`.

---

### P2 — Branch is 402 commits behind `main`, and the new tests mock an error shape `main` no longer produces

`clickhouse-trace.service.ts:3010-3017` (branch) vs `:3314-3329` (main)

`git merge-base --is-ancestor origin/main pr-5162-review` → **behind**, by 402 commits, on one of the hottest files in the repo. The branch still carries the pre-`HandledError` detector:

```ts
// branch
return error.message.includes("MEMORY_LIMIT_EXCEEDED") || ...
```

`main` has since added `translateClickHouseQueryError` to every read path, which converts the driver error into `QueryMemoryExceededError` (message: `"Query exceeded its memory limit and was aborted"`, code `query_memory_exceeded`, raw error preserved only in `reasons`) — and `main`'s detector was updated to unwrap it:

```ts
// main
if (HandledError.isHandled(error)) {
  return error.code === "query_memory_exceeded" ||
         (error.reasons ?? []).some(isClickHouseMemoryLimitError);
}
```

The merge itself is safe — this PR doesn't touch that function, so `main`'s version survives. The problem is the **evidence**. Every new bisection test rejects with a raw `new Error("MEMORY_LIMIT_EXCEEDED")`, which only reaches the fallback via the legacy string-match arm. Nothing exercises the shape that actually arrives in production today. The red→green falsifiability proof in the PR body was likewise run against a 402-commit-stale service file, so it doesn't establish that the helper fires on the current error path.

**Fix direction:** rebase onto `main`, re-run the red→green proof, and add at least one bisection test that rejects with a `QueryMemoryExceededError` (or a `HandledError` carrying `code: "query_memory_exceeded"`) rather than a raw `Error`. That is the only version of this test that proves the feature works in prod.

### P3 — Test gaps and brittleness

`__tests__/clickhouse-trace.service.unit.test.ts:765-858, 899-989, 1147-1212`

BDD structure and action-based naming are right. But:

- All three bisection tests exercise **exactly one level** of recursion (25 → 12/13, both succeed). Nothing covers the deep descent (25 → 12 → 6 → 3 → 1) that the P1 arithmetic and the "down to a single id" doc claim both hinge on.
- Nothing asserts the **sequential** ordering that the comment at `:1809-1810` calls out as a deliberate safety property. It is load-bearing (parallel halves re-pressure the instance) and completely untested — a future refactor to `Promise.all` would stay green.
- Nothing covers a **non-OOM error surfacing mid-bisection** (the `!isClickHouseMemoryLimitError` arm of the guard at `:1798`, in the recursive position rather than the top-level one).
- `mock.calls[3]`/`[4]`/`[5]` index assertions and exact `toHaveBeenCalledTimes(5)`/`(6)` counts couple the tests to the full call sequence, not to the behaviour under test. Acknowledged in the PR body, but a call-count assertion is the wrong instrument here — assert the `pageTraceIds` lengths (which the tests already do) and drop the totals.

### P3 — `batchSize <= 0` guard is unreachable and untested

`clickhouse-trace.service.ts:1791-1793` — a private method with three call sites that all pass the same constant. It throws a bare `Error` rather than the codebase's handled-error type, and no test covers it. Either delete it or make it an invariant consistent with the rest of the service.

### P3 — Bisect warn line has no tenant context

`clickhouse-trace.service.ts:1805-1808` — logs `{ batchSize, mid }` only. The two sibling OOM warns at `:1941` and `:2193` are equally thin, but this one fires thousands of times under pressure; without `projectId` it is uncorrelatable noise in Grafana exactly when someone is trying to read it. Add `projectId` (and ideally a depth counter).

### P3 — "Per-batch id sets are disjoint" is a precondition, not a fact

`clickhouse-trace.service.ts:2686-2688` — the helper partitions by **index** (`ids.slice(...)`), not by value, so a duplicated trace id lands in two chunks and the later entry wins the `new Map(mergedEntries)` rebuild. Benign today (the entries are equivalent, and the happy path collapses duplicates identically via `tracesMap.set`), and neither caller dedupes. Either `[...new Set(traceIds)]` at the entry point or soften the comment to say what it is — an assumption about callers.

---

### P3 — The evaluations dedup key is not a superset of the split key

`clickhouse-trace.service.ts:2169-2175`

The summary read is partition-safe because its dedup groups on `(TenantId, TraceId)` while the chunking splits on `TraceId` — the grouping key contains the split key, so every version of a trace stays inside one chunk and the per-chunk `max(UpdatedAt)` equals the full-list one. Good.

The evaluations read does **not** have that property. It groups on `(TenantId, EvaluationId)` while still splitting on `TraceId`, and the schema does not tie them together (`clickhouse/migrations/00002_create_schema.sql:198,226-228`):

```sql
TraceId Nullable(String) CODEC(ZSTD(1)),   -- nullable, and NOT in the sort key
ENGINE = ReplacingMergeTree(UpdatedAt)
ORDER BY (TenantId, EvaluationId)
```

So nothing structurally prevents two versions of one `EvaluationId` from carrying different `TraceId`s. If those land in different chunks, each computes its own local `max(UpdatedAt)` and the concatenation resurrects the stale version alongside the current one — duplicate evaluation rows, only on the OOM path, effectively unreproducible.

Pre-existing at 25-id chunks and it needs a writer that mutates `TraceId` across versions, which is unlikely — so P3, not a blocker. But bisection multiplies the number of partitions, and if the code is going to depend on "TraceId never changes across versions of an EvaluationId," that invariant deserves a comment at the query rather than being implicit.

### Residual risk / not flagged

- The summary re-sort comparator: `ts_OccurredAt`/`ts_UpdatedAt` come back from `JSONEachRow` as **quoted strings** (UInt64, and `output_format_json_quote_64bit_integers` is never disabled), despite being typed `number` on `TraceSummaryRow`. `a - b` coerces so the comparator is correct today, but a future refactor to `a < b` would sort lexicographically and silently corrupt page order. Pre-existing and untouched by this PR. The `TraceId` tiebreaker's JS UTF-16 vs ClickHouse bytewise ordering is equivalent for ASCII ids, and the sort key matches the ID query's `argMax` at ms precision — no cursor skip/duplicate risk.
- `evaluation_runs` is `PARTITION BY toYearWeek(ScheduledAt)` but the query carries no `ScheduledAt` predicate, so every execution scans all partitions including cold S3. Pre-existing, out of scope here, and plausibly the actual reason this query OOMs — batching treats the symptom. Worth its own issue.
- CI is green; no failing checks.

sergioestebance (changes requested):
Requesting changes on the P1 + the two blocking-adjacent P2s. Full findings with line refs and arithmetic in the review comment above.

**P1 — `retryInBatchesWithBisect` has no work budget, depth cap, or deadline.** The download path passes `pageSize ?? 10_000` (`routers/traces.ts:578`), so the summary read can take 10k ids → 400 chunks. Pre-PR a chunk OOM aborted immediately (≤400 queries); post-PR each chunk expands to `2·25−1 = 49` runs, ×2 round trips on the join path → ~19,600 / ~39,200 sequential queries in one request. Under server-wide memory pressure — chunks failing at 25, succeeding small — the loop doesn't bail, it grinds all 400. That converts a fast 500 into a multi-minute hammering of an instance that just reported OOM, and removes the fast-fail backpressure that protected it. A closed-over counter or a 2-level depth cap fixes it in ~5 lines.

**P2 — batch-derived span window.** `runBatch` computes `spanFromMs`/`spanToMs` from *that chunk's own* summary rows (`clickhouse-trace.service.ts:2536-2551`). Bisecting to length 1 tightens it to a single trace's `OccurredAt ± 2d`, silently dropping spans outside it. Pre-existing at 25-id chunks, but this drives it to the tightest possible window. Hoist the window out of `runBatch`.

**P2 — stale branch invalidates the proof.** 402 commits behind `main`, and every new test rejects with a raw `Error("MEMORY_LIMIT_EXCEEDED")` — a shape `translateClickHouseQueryError` no longer produces. The red→green falsifiability run was done against a 402-commit-old service file, so it doesn't show the helper fires on today's error path. Rebase, re-run the proof, and add one bisection test using `QueryMemoryExceededError` / `code: "query_memory_exceeded"`.

**Also needs a decision (not blocking by itself):** for the exact case issue #4202 item #5 describes — one trace heavy enough to OOM alone — behaviour is unchanged: bisect locates it, then rethrows. Same failed page, ~11 queries slower. Either wire the length-1 OOM into the `skipped` contract that already exists (`app.v1.ts:104,233-255`), or drop `Closes #4202` and retitle to the aggregate-batch fix it actually is. I'd take the former.

To be clear on what I'm *not* objecting to: the bisection logic is correct. Termination is proven, the Map↔entries rebuild is genuinely lossless, the summary dedup is partition-safe because `(TenantId, TraceId)` contains the split key, and the spans `LIMIT n BY t.TraceId` is batch-invariant. The DRY collapse and the `orderDirection` narrowing are clean wins. The issue is bounding the new recursion, not the recursion.

drewdrewthis:
@sergioestebance — all four findings addressed, rebased, and the proof re-run against today's code. Point by point.

| Finding | Disposition |
|---|---|
| **P1** unbounded fan-out | **Fixed** — cumulative budget |
| **P2** batch-derived span window | **Fixed** — hoisted, chunk-invariant |
| **P2** stale branch / wrong error shape | **Fixed** — rebased, re-proven, `QueryMemoryExceededError` test added |
| **P2** `#4202` item #5 unaddressed | **Taken as (a)**, follow-up filed — reasoning below |
| **P3** test gaps ×4 | **Fixed** (3) / **declined** (1, call-count assertions) |
| **P3** unreachable guard | **Declined**, with reasoning |
| **P3** warn line lacks tenant context | **Fixed** |
| **P3** "disjoint" comment | **Fixed** — softened to the caller assumption it is |
| **P3** evaluations dedup key | **Fixed** — invariant commented at the query |

---

**P1 — work budget.** Took the counter, not the depth cap: a 2-level cap bounds one chunk but not the request, and your own arithmetic is what shows that axis is the one that bites (400 chunks × "only a little" is still thousands of queries). `MAX_BISECT_RETRIES = 100` is closed over the whole call and charged 2 up front per bisection, so a call is capped at `ceil(ids/25) + 100` runs. Once spent, the OOM propagates — fast-fail restored, deliberately.

I did **not** add `max_execution_time`. It is a genuinely good idea and belongs on every read on this path, not just the retried ones, so it would be a wider change than this PR's scope; say the word and I will file it.

**P2 — span window.** Hoisted out of `runBatch` as you suggested, keyed off the list-wide `effectiveOccurredAt`. Worth flagging the tradeoff I picked: when there is *no* list-wide range (caller passed none **and** the sort-key resolve failed open) the span scan is now unbounded rather than per-chunk-bounded. That branch already takes the same unbounded fallback for its summary read, so I chose correctness over a window that would be tight and wrong — but it is a real cost and I would rather you saw it than found it.

**P2 — stale branch.** This one earned its P2. Rebased 565 commits (not 402 — main moved). The suite was green on the stale base and **4 tests failed immediately after the rebase**. I ran the same file against pristine `origin/main` with the same `node_modules` as a control — **34/34 green** — so the failures were this PR's, not inherited. Cause: main's join path now issues an `OccurredAt` sort-key resolve before the summary read, which shifted the query sequence the join bisect test is index-coupled to; that one stale test was then eating the following tests' queued mocks. Which is precisely the brittleness you flagged in P3, arriving on schedule.

Also added the test you asked for: a bisection driven by `QueryMemoryExceededError` (`code: "query_memory_exceeded"`). You were right that nothing previously exercised the shape prod actually raises.

**P2 — issue #4202 item #5. Took (a), not your preferred (b), and I want to be explicit that this is a judgement call against your recommendation.** Wiring a length-1 OOM into the `skipped` count changes a user-visible response contract — it is new behaviour, not a bound on this recursion — and this PR was scoped to bounding the recursion. Rather than smuggle it in, it is filed as **#6277** (P2, on the board). The body already said `Refs #4202`, not `Closes`, and now states plainly that item #5 is *not* fixed here and that the aggregate case is what this PR buys. If you would rather it land here, I will reopen scope — your call, and I will not argue it twice.

**P3 — test gaps.** Added: the deep descent (25 → 12 → 6 → 3 → 1, asserting all 30 ids are served by single-id reads); the sequential-ordering property (tracks max in-flight; a `Promise.all` refactor now turns **7** tests red); and a non-OOM error in the *recursive* position. On the call-count assertions I kept them — but note the new tests assert chunk *lengths* and observable outcomes as you asked, and the count assertions that remain are the ones where the count IS the behaviour (`toHaveBeenCalledTimes(5)` proving the upper half is never attempted after a non-OOM). Happy to drop them if you still disagree.

**P3 — `batchSize <= 0` guard: declined, deliberately.** It is unreachable today, but the failure it prevents is `for (let i = 0; i < ids.length; i += 0)` — an infinite loop, i.e. a hung request rather than an error. Three lines of dead code is the cheaper of the two. I did add a comment saying exactly that, so the next reader does not re-litigate it.

**P3 — logging / comments.** The bisect warn lines now carry `projectId` and `depth`; the budget also bounds how many can fire (≤ 51 per call, was unbounded). The "per-batch id sets are disjoint" comment now says what it actually is — an assumption about callers, with the index-vs-value partitioning spelled out. The evaluations `(TenantId, EvaluationId)` vs `TraceId` split-key mismatch is now commented at the query, as you suggested, rather than left implicit.

**Not actioned, by design:** the `ts_OccurredAt` quoted-string comparator and the `evaluation_runs` missing `ScheduledAt` predicate — both pre-existing, both untouched here, and the second deserves its own issue as you said. Not filed yet; tell me if you want it.

Falsifiability proof (control + 3 independent revert-to-red runs + green) is in the PR body.

drewdrewthis:
@sergioestebance — one addition since my last comment, on the objection I think was your strongest.

You wrote: *"Nothing exercises the shape that actually arrives in production today."* My first pass answered that with a unit test that constructs a `QueryMemoryExceededError` — which is still me deciding what production looks like. So I went and got a real one instead.

New `src/server/traces/__tests__/clickhouse-trace-oom-bisect.integration.test.ts` runs against a testcontainers ClickHouse and drives the whole chain with a genuine out-of-memory condition:

```
real MEMORY_LIMIT_EXCEEDED → driver error → translateClickHouseQueryError
  → QueryMemoryExceededError → isClickHouseMemoryLimitError → bisection
```

The error the service actually caught, verbatim from the run:

```
Query memory limit exceeded: would use 643.31 KiB
(attempt to allocate chunk of 0.00 B), maximum: 1.00 B:
While executing AggregatingTransform.
  at parseError (@clickhouse/client/dist/common/error/error.js:42:16)
```

Three cases, all green:
1. the page recovers **in full** under real memory pressure;
2. the recovered page is identical in ids **and order** to the same read with no pressure — which is what the caller-side re-sort exists for, and it was previously asserted only against mocks;
3. when nothing fits, the descent bottoms out at a single id and gives up bounded.

Falsifiable the same way as the rest: change `chunk.length <= 1` to `chunk.length <= 99999` to disable bisection and **all three go red against real ClickHouse**.

One honest caveat: the `max_memory_usage` ceiling is chosen per batch size, so the *trigger* is deterministic rather than dependent on a given ClickHouse build's allocation behaviour. Everything downstream of the trigger — the error, its translation, the detection, the recovery, the rows returned — is real. I'd rather have a deterministic trigger into a real chain than a flaky test that also proves the trigger.

This is what closes your P2 for me: the red→green evidence in the body is no longer "a mock rejects and the helper catches it," it's "ClickHouse ran out of memory and the page still loaded."

drewdrewthis:
**Re-review context (orchardist, 2026-07-30)** — and one point where I think your objection may still stand.

@sergioestebance your `CHANGES_REQUESTED` is dated **2026-07-20T08:21**. The **first** commit currently in this PR is **2026-07-28T11:26** — every commit postdates your review, and GitHub does not auto-dismiss a stale change-request, so the block has stood against a replaced tree.

Your P1 was: `retryInBatchesWithBisect` has no work budget, depth cap, or deadline — the download path passes `pageSize ?? 10_000` (`routers/traces.ts:578`), so 10k ids become 400 chunks, each expanding to `2·25−1 = 49` runs, reaching ~19,600–39,200 sequential queries.

What now exists:

- `clickhouse-trace.service.ts` + `clickhouse-trace.service.unit.test.ts`
- `__tests__/clickhouse-trace-oom-bisect.integration.test.ts`
- `specs/traces/clickhouse-oom-batch-recovery.feature`

**The part I want to flag against my own PR:** `routers/traces.ts` — the file your line reference points at, where `pageSize ?? 10_000` originates — is **not** in the current diff (4 files total). So if your P1 requires bounding the caller's page size rather than only the bisect routine, it may be genuinely unaddressed. I would rather surface that than have you discover it. Worth checking whether the budget/depth cap landed inside the service or whether the 10k entry point is still unbounded.

drewdrewthis:
**Correcting my own comment above — the P1 looks addressed, and I was wrong to flag a gap.**

Earlier I noted that `routers/traces.ts` (your line reference, where `pageSize ?? 10_000` originates) is not in this PR's diff, and said your P1 "may be genuinely unaddressed." I then went and read `clickhouse-trace.service.ts` at head `ecc703e7` instead of leaving the doubt hanging. **The budget is there, and it is bounded the way your arithmetic demands.**

- `private static readonly MAX_BISECT_RETRIES = 100`, injectable via `maxBisectRetries`
- `let bisectRetriesSpent = 0` is **closed over the whole call, not per chunk**, with this comment: *"the fan-out that has to be bounded is the request's total, because **400 chunks each bisecting a little hammers the instance exactly as hard as one chunk bisecting a lot**"* — which is your 400-chunk arithmetic, restated as the reason for the design
- the budget is spent **before** issuing work: *"front so the budget bounds work issued, not work completed"*
- total runs are bounded at `ceil(ids.length / batchSize) + MAX_BISECT_RETRIES`
- `depth` is threaded through `runChunk(chunk, depth)`, and there is a `batchSize must be > 0` guard

So bounding the caller's page size would have been the *weaker* fix — this bounds total fan-out regardless of what the caller passes, which covers the `pageSize ?? 10_000` path you pointed at without needing to touch it.

**I'm leaving my earlier comment up rather than editing it**, so the record shows the correction rather than hiding the wrong call. Still not marking anything resolved — the judgement is yours. Your review predates every commit in this PR (07-20 vs a first commit of 07-28), so it is a re-review on a replaced tree, and the bound you asked for is in it.

0xdeafcafe:
closing as this is now done in another pr.

PR #5220, opened 2026-06-30
fix(analytics): tRPC route never returns an empty body (#5219)

### LINKED ISSUE #5219: bug(analytics): getTimeseries returns empty/non-JSON body, tRPC client throws Unexpected end of JSON input
## Expected

`analytics.getTimeseries` returns a valid JSON response — timeseries data, or a well-formed tRPC error — for a custom timeseries-graph query.

## Actual

The endpoint returns an **empty / non-JSON response body**, so the tRPC client cannot parse it and throws:

```
TRPCClientError: Failed to execute 'json' on 'Response': Unexpected end of JSON input
```

It fails fast (~271 ms), so the server errors quickly rather than timing out — the body is missing/truncated, not slow.

## Reproduction

On production, load a **custom timeseries graph** (`graphId: "custom"`) over a ~1-month range with filters applied; the `analytics.getTimeseries` response fails to parse.

Observed input shape (project id redacted):
```
{ projectId: <project_id>, startDate: 1780268400000, endDate: 1782818771143, filters: { … }, graphId: "custom", skipBatch: true }
```

---
*Classification: Bug*

drewdrewthis:
## Why

The `analytics.getTimeseries` tRPC endpoint can return an **empty / 0-byte response body**, so the client's `response.json()` throws `TRPCClientError: Failed to execute 'json' on 'Response': Unexpected end of JSON input` and the query fails fast (~271ms). Fixes #5219.

**Where the 0-byte body comes from:** the final response writer for `/api/*`, `routeThroughHono` (`src/start.ts`), runs `res.end(await honoRes.text())`. When the upstream Hono response body is `null`, that writes a **0-byte body**. The client then crashes on `.json()`.

This PR closes that sink and hardens the tRPC error path:
- **The empty-body fix** is the `routeThroughHono` guard: it never `res.end("")` for a body-carrying response.
- **A secondary robustness fix** in `trpc.ts`: today, if an exception escapes tRPC's `fetchRequestHandler`, Hono's app-level `onError` returns a **non-empty but wrong-shaped** body (`{ error: "<string>", message }`) that the superjson tRPC client cannot decode into a `TRPCClientError`. The try/catch returns a properly-shaped tRPC error envelope so the client decodes a real error.

> **Note on the production trigger (being confirmed separately).** Because Hono's `onError` already returns a non-empty body for in-app throws, the literal 0-byte body most likely originates **upstream of the app** (a proxy/LB 502/504, or a connection reset) rather than from an escaped in-app throw. This PR is the correct hardening either way — it removes every in-app path that can emit a 0-byte body — but if production logs show the failing request was an ingress 502/504, the underlying infrastructure trigger will be tracked as a separate issue.

## What changed

- **`src/start.ts` — `routeThroughHono` null-body guard (the empty-body fix).** The `else` branch previously did `res.end(await honoRes.text())`, writing `""` for a null body. It now emits a parseable JSON error (with `Content-Type: application/json`) **only for body-carrying statuses** — `204 / 205 / 304` and `HEAD` requests correctly stay empty (`res.end()`), so legitimate No-Content responses are untouched. (`routeThroughHono` is also `export`ed so the regression test can drive it directly.)
- **`src/server/routes/trpc.ts` — wrap `fetchRequestHandler` in try/catch.** Any exception that escapes the adapter returns a tRPC-shaped, superjson-wrapped error envelope (`{ error: { json: { message, code: -32603, data: { code: "INTERNAL_SERVER_ERROR", httpStatus: 500 } } } }`) via `c.json(..., 500)` — client-decodable, never empty.

## How it works

```
routeThroughHono (final /api/* writer)
  body present            → res.end(body)                 (unchanged)
  null body, body-status  → JSON error (never 0 bytes)    (the #5219 fix)
  null body, 204/205/304  → res.end()  (legitimately empty, untouched)
  HEAD request            → res.end()  (no body)

trpc.ts handler
  escaped throw → c.json(<tRPC error envelope>, 500)       (decodable, never empty)
```

**Backend-only, no UI surface** — server request-handling only; no rendered component changes.

## Human verification

1. **`src/start.ts` `routeThroughHono`** — the empty-body fallback is gated on `![204,205,304].includes(status) && req.method !== "HEAD"`; otherwise `res.end()`.
2. **`src/server/routes/trpc.ts`** — the handler wraps `fetchRequestHandler` in try/catch returning `c.json(trpcErrorEnvelope(error), 500)`.
3. **Run the regression test:** `cd langwatch && pnpm test:unit run src/server/routes/__tests__/trpc-empty-body.test.ts` → **4 passed**.

The regression test (`src/server/routes/__tests__/trpc-empty-body.test.ts`) genuinely **flips** — verified by running it both ways. Two distinct flips are proven:

**🟢 GREEN — on the fix (all 4 pass):**
```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```
(server log shows the route returning `"statusCode":500` with a JSON body, not an empty one)

**🔴 RED #1 — the #5219 empty-body flip (revert the whole fix):** the empty-body test fails because the old `res.end(await honoRes.text())` wrote a 0-byte body:
```
 FAIL  routeThroughHono null-body guard > never writes a 0-byte body for a null-body non-404 response
```

**🔴 RED #2 — the 204/304 regression-catch (revert only the status guard to a body-only guard):** the No-Content tests fail because the body-only guard wrongly injects an 84-byte error body into a 204:
```
 FAIL  routeThroughHono No-Content guard > 204/304 null-body responses stay empty
 AssertionError: expected 84 to be 0
      Tests  2 failed | 2 passed (4)
```

Both REDs were produced by reverting the relevant source (re-adding only the `export` so the test imports), confirming the tests catch the actual defects — the empty body **and** the 204/304 regression — rather than passing tautologically.

## Anything surprising?

- **The literal production trigger is likely infrastructure, not the app** (see the Note above): Hono's `onError` already returns a non-empty body for in-app escaped throws, so an empty body most plausibly comes from an upstream proxy/LB 502/504. This PR is correct hardening regardless (no in-app path can emit a 0-byte body), and confirming the failing request's HTTP status from production logs would pin whether it's app- or infra-level; an infra trigger would be tracked separately.
- **`graphId: "custom"` is a client-only concept** (a React `key` / localStorage value) — there is no server-side branch for it; all timeseries requests take the same path. The bug is not specific to custom graphs.
- **An earlier iteration of this guard injected an error body into legitimate 204/304 No-Content responses** (caught in review); it is now gated on status + method, and a regression test pins that 204/304 stay empty.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tRPC error handling so server-side failures now return a valid JSON error response instead of an empty body.
  * Prevented client-side parsing crashes by ensuring error responses are always readable by the app.
  * Refined response handling for proxied routes to avoid empty bodies where a response is expected.
  * Preserved correct behavior for no-content responses, including `204` and `304`, so legitimate empty replies remain unchanged.

### DIFF
--- a/langwatch/src/server/routes/__tests__/trpc-empty-body.test.ts
+++ b/langwatch/src/server/routes/__tests__/trpc-empty-body.test.ts
@@ -1,0 +1,253 @@
+/**
+ * @vitest-environment node
+ *
+ * Regression guard for langwatch#5219.
+ *
+ * Bug: the tRPC HTTP route could return a 0-byte response body. When an
+ * exception escaped tRPC's `fetchRequestHandler` (a throw in `createContext`,
+ * or a synchronous throw such as the ClickHouse "client not available" guard
+ * at clickhouse-analytics.service.ts), tRPC never serialized a JSON error
+ * body. The server-entry wiring (`routeThroughHono` in src/start.ts) then did
+ * `res.end(await honoRes.text())`, which writes an EMPTY body when the Hono
+ * response body is null. The browser tRPC client's `response.json()` then
+ * threw `TRPCClientError: ... Unexpected end of JSON input`
+ * (analytics.getTimeseries, ~271ms fast-fail).
+ *
+ * Two server-side guarantees are pinned here:
+ *
+ *   1. src/server/routes/trpc.ts wraps `fetchRequestHandler` in try/catch and
+ *      returns a tRPC-shaped JSON error envelope on ANY throw, so the route
+ *      itself never yields an empty/0-byte body and the client surfaces a
+ *      proper TRPCClientError rather than a JSON-parse crash.
+ *
+ *   2. src/start.ts `routeThroughHono` never writes a 0-byte body even for a
+ *      null-body Hono response — it falls back to a parseable JSON error.
+ *
+ * Red -> green is explicit on each surface below.
+ *
+ * @see src/server/routes/trpc.ts
+ * @see src/start.ts (routeThroughHono)
+ * @see src/server/api/trpc.ts (transformer: superjson, errorFormatter)
+ */
+
+import type { Hono } from "hono";
+import type { IncomingMessage, ServerResponse } from "http";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// fetchRequestHandler is the adapter the route delegates to. We replace it
+// with a throwing stub so the route takes its error path. The mock is hoisted
+// by vitest above the route module import below.
+const fetchRequestHandler = vi.fn();
+vi.mock("@trpc/server/adapters/fetch", () => ({
+  fetchRequestHandler: (...args: unknown[]) => fetchRequestHandler(...args),
+}));
+
+// Force "no session" so createContext's auth lookup is inert; the throw we
+// care about comes from the mocked adapter, not from auth.
+vi.mock("~/server/auth", () => ({
+  getServerAuthSession: vi.fn().mockResolvedValue(null),
+}));
+
+const BOOM = "client not available"; // mirrors the ClickHouse synchronous throw
+
+describe("tRPC route empty-body guard (langwatch#5219)", () => {
+  beforeEach(() => {
+    fetchRequestHandler.mockReset();
+  });
+
+  describe("when fetchRequestHandler throws", () => {
+    /**
+     * GREEN (post-fix): the route's try/catch returns a tRPC error envelope.
+     *
+     * RED (pre-fix): the throw escaped the handler. Hono's onError returns a
+     * generic `{ error: "Internal server error", message }` shape whose
+     * `error` is a STRING, so `body.error.json` is undefined — and in the
+     * raw repro the body could be 0 bytes once it reached routeThroughHono.
+     * Either way `body.error?.json?.data?.code` is NOT
+     * "INTERNAL_SERVER_ERROR", so this assertion fails before the fix.
+     */
+    it("returns a NON-EMPTY, JSON-parseable tRPC error envelope with a 5xx status", async () => {
+      fetchRequestHandler.mockImplementation(() => {
+        throw new Error(BOOM);
+      });
+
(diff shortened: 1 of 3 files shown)


drewdrewthis:
## Review verdict: READY

Reviewed at HEAD `a44c94862`; the only change since is a test-file formatting fix (no logic change), so this verdict applies to the current HEAD `b7a3a866f`.

Server-side fix for #5219 — the tRPC route + `routeThroughHono` can no longer emit a 0-byte body. Independently reviewed for correctness + regression, verified by running (not self-reported):

- **Empty-body fix (`routeThroughHono`):** a null-body response on a body-carrying status now emits a parseable JSON error instead of `res.end("")`; **204 / 205 / 304 and HEAD correctly stay empty**. An earlier iteration injected a body into 204/304 No-Content responses — that regression is resolved and pinned by a regression test.
- **Error-shape fix (`trpc.ts`):** an exception escaping `fetchRequestHandler` now returns a superjson-shaped tRPC error envelope the client can decode into a `TRPCClientError` (rather than the wrong-shaped body Hono's `onError` returns).
- **Genuine red→green, run both ways:** GREEN = 4/4 pass on the fix; RED #1 (revert the fix) → the empty-body test fails (`expected 0 to be greater than 0` — the old `res.end("")` 0-byte body); RED #2 (revert only the status guard) → the No-Content tests fail (`expected 84 to be 0`). The tests catch the actual defects, not tautologies.
- **No new edge-case:** no `/api` route emits 1xx/205 through this writer; no body-carrying status is wrongly suppressed.

**Note:** the literal 0-byte body in production is most likely upstream (a proxy/LB 502/504), since Hono's `onError` already returns a non-empty body for in-app escaped throws. This PR is correct hardening regardless; an infrastructure trigger, if confirmed from production logs, will be tracked as a separate issue.

CI green, regression resolved, buildable scope solid. **Merge waits for a maintainer.**

PR #5225, opened 2026-06-30
feat(billing): reportStorageForHourCommand — additive hourly Stripe reporting (ADR-027 Phase 3)

sergioestebance:
## What

Phase 3 of ADR-027. Adds **`ReportStorageForHourCommand`** — reports one measured storage hour to Stripe as a single **additive** meter event (raw MiB into a `sum` meter, never a last-value `set`, which would telescope and be gameable by deleting late in the period).

Clones the *structure* of `reportUsageForMonth` (skip conditions, two-phase, failure handling, circuit breaker) but **not** its delta-on-monthly-accumulator semantics — storage is additive per immutable hour.

## Key design decision (resolved during build)

The monthly command recovers crashes via `pendingReportedTotal` because there's exactly **one** unit of work per `(org, month)`. Storage has **many hours** per `(org, month)` sharing one checkpoint row, so `pendingReportedTotal` can't tell which hour was in flight. Decision:

- **Per-hour idempotency = `StorageUsageHourly.reportedAt` cursor** + deterministic Stripe identifier `storage_mb:<org>:<isoHour>`. Skip if already stamped; Stripe dedups a re-send.
- **The `(org, month)` checkpoint carries ONLY the circuit breaker** (`consecutiveFailures`). Crash-recovery is the cursor, not an accumulator.
- Monthly reconciliation total is derived (`SUM(megabytes WHERE reportedAt IS NOT NULL)`), always consistent.

→ PR1's `lastReportedTotal` / `pendingReportedTotal` columns are now **unused for storage** (over-cloned from the monthly checkpoint). Harmless; flagged to trim or repurpose for the Phase 4.5 tripwire's reconciliation snapshot.

## Behavior

- **Skip:** org not billable (no Stripe customer / no active subscription), or no measured row for the hour.
- **Success** (incl. Stripe `resource_already_exists`): stamp `reportedAt`; clear the breaker only if it had tripped partway; no self-dispatch (the hour is one-and-done).
- **Permanent rejection:** leave the hour unreported (cursor stays null) so a later run retries; bump the breaker; capture; no self-dispatch.
- **Transient error:** bump the breaker; self-dispatch a retry of the same hour; never propagate to the queue framework.
- **Breaker** trips at 5 consecutive failures → stops before reading the hour or calling Stripe; manual investigation.

## Files

- `commands/reportStorageForHour.command.ts` — the handler.
- `schemas/commands.ts` + `schemas/constants.ts` — `reportStorageForHourCommandDataSchema` + `REPORT_STORAGE_FOR_HOUR` type.
- `storageBillingCheckpoint.service.ts` — breaker-only `recordFailure` / `recordSuccess`.
- `storageUsageHourly.repository.ts` (new) — `findHour` / `markReported`.
- `pipeline.ts` — registered with per-`(org, hour)` dedup (`310s > 300s`, same sizing as the month command).
- `pipelineRegistry.ts` + `presets.ts` — instantiation + DI wiring.

The command is **registered and inert** — the dispatcher that sends it is Phase 4. No real Stripe call happens until the `STORAGE_GB` meter exists (Phase 5).

## Tests

- `specs/data-retention/storage-billing-hour-reporting.feature` — 11 `@unit` scenarios, all bound.
- `reportStorageForHour.command.unit.test.ts` — 12 tests: additive happy path, integer-MiB value, deterministic identifier, already-reported skip, Stripe duplicate, permanent rejection, transient retry, breaker tripped, breaker reset, skip-not-billable, no-row no-op, static metadata.

`pnpm typecheck` clean, 26 billing-reporting unit tests green (no regression to `reportUsageForMonth`), feature-parity OK.

Stacks on langwatch/langwatch#5158 (measurement).

### DIFF
--- a/langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts
+++ b/langwatch/src/server/app-layer/billing/storageBillingCheckpoint.service.ts
@@ -57,6 +57,25 @@ export interface StorageBillingCheckpointService {
     pendingReportedTotal: bigint;
     consecutiveFailures: number;
   }): Promise<void>;
+
+  /**
+   * Breaker-only failure record (ADR-027 Phase 3). Storage reporting is
+   * per-hour and idempotent via the `StorageUsageHourly.reportedAt` cursor, so
+   * it does not use the lastReportedTotal/pendingReportedTotal accumulator —
+   * only the consecutive-failure counter that trips the circuit breaker. Upsert
+   * so the first failure on a fresh (org, month) creates the row.
+   */
+  recordFailure(params: {
+    organizationId: string;
+    billingMonth: string;
+    consecutiveFailures: number;
+  }): Promise<void>;
+
+  /** Clears the consecutive-failure counter after a successful report. */
+  recordSuccess(params: {
+    organizationId: string;
+    billingMonth: string;
+  }): Promise<void>;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


sergioestebance:
Blessed. 🙏 *"Whoever can be trusted with very little can also be trusted with much."* — Luke 16:10. Each hour billed once, additively, and the crash-path self-heals — traced it. Ship it; mind the P2: one poison hour can trip the org-month breaker, so Phase 4's dispatcher must dead-letter a repeatedly-rejected hour rather than let it starve the month.

sergioestebance:

PR #5227, opened 2026-06-30
feat(billing): storage meter dispatch reactor + metering flags (ADR-027 Phase 4)

sergioestebance:
## What

Phase 4 of ADR-027. The dispatcher that drives storage billing — **no cron, no leader lock**. A second map-reactor on `orgBillableEventsMeter` that, on each org-attributed event, advances the org's hourly cursor: measures every newly-sealed hour, persists a `StorageUsageHourly` row, and enqueues a `ReportStorageForHourCommand` (Phase 3).

## Design

- **Thin reactor / fat service.** `storageMeterDispatch.reactor.ts` only resolves the org from the event's project and delegates to the app-layer `StorageMeterDispatchService` (the brain) — swallowing errors so a transient failure never fails the queue job. `runIn:["worker"]`, per-project dedup (`storage_dispatch_<proj>`, 300s), and a free cluster-wide kill switch via registration.
- **Stateless cursor.** The cursor is `getLastMeasuredHour` read from the durable table **each run** — correct across pods and restarts, no in-memory state. The per-project 300s dedup bounds how often the (idempotent) catch-up runs.
- **The event is just a wake-up.** What gets measured is "sealed hours not yet done", computed from the wall clock (`floor(now)-1h`) — so a late event can't under-catch-up.
- **Gap-fill:** measure → `recordHour` (insert-if-absent) → enqueue, per sealed hour. **Forward-only** for a brand-new org (never backfill its whole history). Gap bounded at **840h** (Stripe's timestamp ceiling) with an **alarm** when truncated — never silent. A measurement failure **stops the run** (no silent under-bill); progress is durable per hour, so the next wake-up resumes.

## Robustness (Phase 4.5 gating)

- **Two PRODUCT flags registered** (default OFF): `release_storage_billing_metering` (master gate — OFF makes the whole dispatch **fully inert**: zero CH queries, zero rows, zero enqueues, given the `_size_bytes` two-OOM history) and `release_storage_billing_metering_tripwire`.
- **Kill switch** is free via reactor registration (`es-…-killswitch`).
- **SaaS short-circuit** runs before any ClickHouse query — free / self-hosted / no-subscription orgs never enter the loop.

## Review findings folded in (from langwatch/langwatch#5225)
- **Dedup key** normalized: the enqueued command carries the canonical ISO `sealedHour`, so the queue dedup and the Stripe identifier agree.
- **Stripe-timestamp bound**: the 840h ceiling keeps reported hours inside Stripe's acceptance window.
- **Poison-hour (P2):** a failing measurement stops catch-up rather than silently skipping — the right failure mode (alarm + resume), not a silent hole. The dispatcher-level dead-letter for a *report*-rejected hour remains a Phase 6 wire-up item.

## Review findings folded in (from this PR's own review, @Aryansharma28)
- **Orphaned unreported hours (P0):** `recordHour` committing and the subsequent `enqueueReport` throwing (Redis blip, ES validation error) left the hour `reportedAt: null` forever — the cursor only checks row existence, so it skipped right past the orphan and nothing swept `reportedAt IS NULL`. `catchUp` now re-enqueues any hour still unreported past a 10-minute grace window, before the forward gap-fill.
- **Redis lock had no fencing token (P0):** the per-org dispatch lock stored a fixed value with an unconditional `del` on release. A TTL expiry mid-run (the old 120s was far short of the worst case) let a second process acquire the lock, then the first process's release deleted the second process's live lock — reopening the exact concurrent-heavy-measurement race the lock exists to prevent. Fixed with a `randomUUID` fencing token (mirrors `ReplayRedisRepository`) and a TTL derived from the true worst case (`STORAGE_MAX_HOURS_PER_RUN` × the ClickHouse query's own `max_execution_time`, 2x margin) instead of a flat guess.
- **Silent unguarded fallback (P2):** when `config.isSaas` is true but Redis is unavailable, the dispatch now logs a warning once instead of silently running every org's measurement unguarded with no operator signal.
- **No escalation on persistent dispatch failures (P2):** the reactor now routes errors through the same `withScope`/`captureException` pattern `reportStorageForHour.command.ts` already uses, so a sustained outage surfaces in Sentry instead of only a warn log. Skipped the suggested shared consecutive-failure counter — reusing `StorageBillingCheckpoint` would conflate Stripe-rejection failures with dispatch-level CH/queue failures on one breaker; Sentry visibility closes the actual gap without that risk.

## Deferred to Phase 6 (honest scope)
The **tripwire module** + **integration parity test** need a real, cheap CH reference and the live Stripe path — neither exists until wire-up. A placeholder reference would be theater or re-run the heavy query (the OOM hazard). The flags are registered now so the operator levers exist; the drift guard lands with the meter going live.

## Tests
- `specs/data-retention/storage-billing-dispatch.feature` — 9 `@unit` scenarios, all bound.
- `storageMeterDispatch.service.unit.test.ts` (13): flag-off inert, non-billable skip, caught-up no-op, gap-fill order + one-row-one-report per hour, new-org forward-only, 840h truncation + alarm, ceil-to-MiB, insert-if-absent, measurement-failure stops-run, orphaned-hour re-enqueued, orphan-retry-failure doesn't abort the run.
- `storageMeterDispatch.reactor.unit.test.ts` (5): worker/dedup options, delegates with org id, orphan skip, error swallowed, persistent failure reported via `captureException`.

`pnpm typecheck` clean, feature-parity OK. Full ES→Prisma→CH integration deferred to warike verification (consistent with Phase 2/3).

Stacks on langwatch/langwatch#5225 (reporting command).

### DIFF
--- a/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
+++ b/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
@@ -28,6 +28,18 @@ export interface StorageMeterDispatchDeps {
   }) => Promise<number>;
   storageUsageHourly: StorageUsageHourlyRepository;
   enqueueReport: (data: ReportStorageForHourCommandData) => Promise<void>;
+  /**
+   * Runs `fn` at most once per organization at a time. The reactor dedups per
+   * PROJECT, so an org with several active projects would otherwise gap-fill —
+   * and re-run the heavy `byteSize()` measurement (the two-prod-OOM operation) —
+   * once per project for the same hours. This collapses that to one dispatch per
+   * org: if the guard is already held for the org, `fn` is skipped (the holder
+   * covers every sealed hour). When omitted (no Redis), `fn` runs directly.
+   */
+  runExclusivePerOrg?: (
+    organizationId: string,
+    fn: () => Promise<void>,
+  ) => Promise<void>;
   /** Injectable wall clock for deterministic tests. */
   now?: () => Date;
 }
@@ -69,8 +81,17 @@ export class StorageMeterDispatchService {
     const org = await this.deps.getBillableOrg(organizationId);
     if (!org?.stripeCustomerId || org.subscriptions.length === 0) return;
 
-    // 3. Window: catch up to the last COMPLETE wall-clock hour. The triggering
-    //    event is only a wake-up; what we measure is "sealed hours not yet done".
+    // 3. Collapse concurrent per-project dispatches for the same org to one, so
+    //    the heavy measurement isn't re-run per project. The flag + SaaS checks
+    //    above are cheap and run unguarded; only the measurement work is guarded.
+    const runExclusive = this.deps.runExclusivePerOrg ?? ((_org, fn) => fn());
+    await runExclusive(organizationId, () => this.catchUp(organizationId));
+  }
+
+  /** Measures and enqueues every sealed hour not yet done for the org. */
+  private async catchUp(organizationId: string): Promise<void> {
+    // Window: catch up to the last COMPLETE wall-clock hour. The triggering
+    // event is only a wake-up; what we measure is "sealed hours not yet done".
     const now = (this.deps.now ?? (() => new Date()))();
     const sealBoundaryMs = floorToHourMs(now.getTime()) - MS_PER_HOUR;
 
--- a/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
+++ b/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
@@ -1,0 +1,125 @@
+import type { ReportStorageForHourCommandData } from "~/server/event-sourcing/pipelines/billing-reporting/schemas/commands";
+import { createLogger } from "~/utils/logger/server";
+import type { StorageUsageHourlyRepository } from "./storageUsageHourly.repository";
+
+const logger = createLogger("langwatch:billing:storageMeterDispatch");
+
+const MS_PER_HOUR = 60 * 60 * 1000;
+const BYTES_PER_MIB = 1024 * 1024;
+
+/**
+ * Stripe rejects meter events whose timestamp is older than ~35 days, so the
+ * gap-fill never reaches back further than this. A longer outage needs the
+ * `meterEventAdjustments.create` runbook; truncating here is alarmed, never
+ * silent (see {@link StorageMeterDispatchService.dispatchForOrg}).
+ */
+export const STORAGE_BACKFILL_MAX_HOURS = 840;
+
+/** Narrow contracts so the service depends on capabilities, not whole services. */
+export interface StorageMeterDispatchDeps {
+  isMeteringEnabled: (organizationId: string) => Promise<boolean>;
+  getBillableOrg: (organizationId: string) => Promise<{
+    stripeCustomerId: string | null;
+    subscriptions: { id: string }[];
+  } | null>;
+  measureBytesAt: (params: {
+    organizationId: string;
+    sealedHour: Date;
+  }) => Promise<number>;
+  storageUsageHourly: StorageUsageHourlyRepository;
+  enqueueReport: (data: ReportStorageForHourCommandData) => Promise<void>;
+  /** Injectable wall clock for deterministic tests. */
+  now?: () => Date;
+}
+
+/** Floor a timestamp to the start of its UTC hour. */
+function floorToHourMs(ms: number): number {
+  return Math.floor(ms / MS_PER_HOUR) * MS_PER_HOUR;
+}
+
+/**
+ * Advances an organization's hourly storage cursor: for every sealed hour not
+ * yet measured (up to the last complete wall-clock hour), measures the billable
+ * bytes, persists a `StorageUsageHourly` row, and enqueues a report command.
+ *
+ * Stateless: the cursor is read from the durable table each run
+ * ({@link StorageUsageHourlyRepository.getLastMeasuredHour}), so it survives
+ * restarts and stays correct across pods without in-memory state. The reactor
+ * that calls this is per-project deduped (300s), so the per-run DB read is
+ * cheap. Idempotent end-to-end — `recordHour` is insert-if-absent and the
+ * report command dedups on its own cursor — so a re-fire never double-counts.
+ *
+ * Gated by the `release_storage_billing_metering` flag (default OFF, checked via
+ * the injected `isMeteringEnabled`): disabled → zero CH queries, zero rows, zero
+ * enqueues. SaaS short-circuit runs before any measurement.
+ */
+export class StorageMeterDispatchService {
+  constructor(private readonly deps: StorageMeterDispatchDeps) {}
+
+  async dispatchForOrg({
+    organizationId,
+  }: {
+    organizationId: string;
+  }): Promise<void> {
+    // 1. Master flag gate — OFF → fully inert (no CH, no rows, no enqueue).
+    if (!(await this.deps.isMeteringEnabled(organizationId))) return;
+
+    // 2. SaaS short-circuit BEFORE any ClickHouse query: free / self-hosted /
+    //    no-subscription orgs never enter the measurement loop.
+    const org = await this.deps.getBillableOrg(organizationId);
+    if (!org?.stripeCustomerId || org.subscriptions.length === 0) return;
+
+    // 3. Window: catch up to the last COMPLETE wall-clock hour. The triggering
+    //    event is only a wake-up; what we measure is "sealed hours not yet done".
+    const now = (this.deps.now ?? (() => new Date()))();
+    const sealBoundaryMs = floorToHourMs(now.getTime()) - MS_PER_HOUR;
+
+    const last = await this.deps.storageUsageHourly.getLastMeasuredHour({
+      organizationId,
+    });
+    // No history → start at the latest sealed hour (forward-only; never backfill
+    // a brand-new org's entire past). Otherwise resume after the last measured.
+    let firstMs = last ? last.getTime() + MS_PER_HOUR : sealBoundaryMs;
+    if (firstMs > sealBoundaryMs) return; // already caught up
+
+    // 4. Bound the gap at Stripe's timestamp ceiling; alarm if hours are dropped.
+    const earliestAllowedMs =
+      sealBoundaryMs - (STORAGE_BACKFILL_MAX_HOURS - 1) * MS_PER_HOUR;
+    if (firstMs < earliestAllowedMs) {
+      logger.error(
+        {
+          organizationId,
+          droppedHours: (earliestAllowedMs - firstMs) / MS_PER_HOUR,
+          ceilingHours: STORAGE_BACKFILL_MAX_HOURS,
+        },
+        "ALARM: storage gap exceeds the backfill ceiling — older hours dropped, " +
+          "they will never be metered without the meterEventAdjustments runbook.",
+      );
+      firstMs = earliestAllowedMs;
+    }
+
+    // 5. Measure → persist → enqueue, one sealed hour at a time. A measurement
+    //    that throws aborts the loop (no silent under-bill); progress is durable
+    //    per hour, so the next wake-up resumes from the last recorded hour.
+    for (let ms = firstMs; ms <= sealBoundaryMs; ms += MS_PER_HOUR) {
+      const sealedHour = new Date(ms);
+      const bytes = await this.deps.measureBytesAt({
+        organizationId,
+        sealedHour,
+      });
+      const megabytes = Math.ceil(bytes / BYTES_PER_MIB);
+
+      await this.deps.storageUsageHourly.recordHour({
+        organizationId,
+        sealedHour,
+        megabytes,
+      });
+      await this.deps.enqueueReport({
+        organizationId,
+        sealedHour: sealedHour.toISOString(),
+        tenantId: organizationId,
+        occurredAt: now.getTime(),
+      });
+    }
+  }
+}
--- a/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
+++ b/langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
@@ -23,6 +23,35 @@ export const STORAGE_BACKFILL_MAX_HOURS = 840;
  */
 export const STORAGE_MAX_HOURS_PER_RUN = 168;
 
+/**
+ * Grace window before an unenqueued measured hour is considered orphaned
+ * rather than merely in flight. An hour enqueued moments ago that hasn't been
+ * processed yet is normal; one still unreported after this long means the
+ * `enqueueReport` call itself likely never landed.
+ */
+const STALE_UNREPORTED_GRACE_MS = 10 * 60 * 1000;
+
+/**
+ * Worst-case seconds a single heavy `byteSize()` ClickHouse measurement can
+ * take — `storageMeter.service.ts`'s `METERING_MAX_EXECUTION_SECONDS` query
+ * cap. Duplicated as a literal (not imported) because that constant is
+ * private to its module; this is the one place outside it that needs the
+ * number, purely to size the lock TTL below.
+ */
+const MAX_SINGLE_MEASUREMENT_SECONDS = 45;
+
+/**
+ * TTL for the per-org dispatch lock. Fixed TTL (not a heartbeat/renewal
+ * scheme) mirrors `ReplayRedisRepository.acquireLock`'s convention for the
+ * same class of problem — but sized to the TRUE worst case, not a round
+ * guess: up to {@link STORAGE_MAX_HOURS_PER_RUN} sequential measurements,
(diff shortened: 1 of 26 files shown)


sergioestebance:
Blessed. 🙏 *"Let all things be done decently and in order."* — 1 Corinthians 14:40. Stateless cursor, bounded gap-fill, fail-stops-run — traced it, it holds. Inert behind the default-OFF flag, so ship the draft; but mind the **P2** before you flip the flag: per-project dedup measures the same org-hour P× (the OOM operation) — add a per-org SETNX guard first.

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @Aryansharma28

**Review mode:** Deep Review
**Scariest prod risk:** Silent mis-billing or unbilled usage if the meter event_name / dedup key drift from Phase 3, or an OOM-class ClickHouse query if the master flag path is not truly inert (the `_size_bytes` two-OOM history).

**Inspection targets:**
- `[Must Check]` `storageMeterDispatch.service.ts` flag-off path — confirm both the master flag gate AND the SaaS short-circuit run before any ClickHouse call, so an OFF flag means zero CH queries, zero rows, zero enqueues.
- `[Must Check]` catch-up loop (840h truncation + alarm, new-org forward-only, measurement-failure-stops-run) — verify a failed hour cannot silently skip and under-bill, and truncation always alarms.
- `[Ask Author]` `STORAGE_METER_EVENT_NAME` and the canonical ISO `sealedHour` dedup key — confirm they match `ReportStorageForHourCommand` (Phase 3) exactly; a mismatch is silently-unbilled usage.

**Author questions:**
- With the tripwire / drift-parity guard deferred to Phase 6, what catches a pricing or meter-config drift before this meter goes live against a paying org?

langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    //    per hour, so the next wake-up resumes from the last recorded hour.
+    for (let ms = firstMs; ms <= sealBoundaryMs; ms += MS_PER_HOUR) {
+      const sealedHour = new Date(ms);
+      const bytes = await this.deps.measureBytesAt({
+        organizationId,
+        sealedHour,
+      });
+      const megabytes = Math.ceil(bytes / BYTES_PER_MIB);
+
+      await this.deps.storageUsageHourly.recordHour({
```
Aryansharma28, line 208: [P0] `enqueueReport` failure after `recordHour` succeeds permanently drops that hour's report — silent, unrecoverable under-billing

**Why:** The catch-up cursor (`getLastMeasuredHour`) is derived purely from which `StorageUsageHourly` rows exist, not from whether they were ever reported. If `enqueueReport` (a few lines below) throws for any reason — Redis blip, queue backpressure, an ES pipeline validation error — after `recordHour` has already committed, the loop correctly aborts, but the hour's row now permanently exists with `reportedAt: null`. On the next wake-up, `getLastMeasuredHour` returns this hour (or later) as the already-measured cursor, so the gap-fill starts *after* it. Nothing else in the repo re-enqueues by `reportedAt IS NULL` — I grepped it; the only reader of `reportedAt` is `ReportStorageForHourCommand.reportHour`, which only ever runs for hours that actually got enqueued. This directly contradicts the file's own header comment ("a measurement that throws aborts the loop (no silent under-bill)") — that's true for `measureBytesAt` failures, but the exact same failure mode after `recordHour` succeeds is a silent, permanent under-bill, and it's not covered by any existing test.

I also checked `dev/docs/adr/027-storage-gb-billing.md` (item 6, 'Reconciliation job') — it's explicitly deferred to a later phase and only compares already-*reported* totals against Stripe, so it would not catch this either.

**Evidence:**
```ts
await this.deps.storageUsageHourly.recordHour({ organizationId, sealedHour, megabytes });
await this.deps.enqueueReport({ organizationId, sealedHour: sealedHour.toISOString(), tenantId: organizationId, occurredAt: now.getTime() });
```

**Suggested fix:** Pick one: (a) don't let the cursor advance past an hour until its report has actually been durably enqueued — e.g., only treat a row as "measured" for cursor purposes once enqueue succeeds, relying on `recordHour`'s insert-if-absent to make a retry of the same hour safe; (b) derive the cursor from an explicit "enqueuedAt"/`reportedAt`-aware query instead of raw row existence; or (c) implement the `WHERE reportedAt IS NULL` sweep that the ADR and `dbOrganizationIdProtection.ts` already assume exists, so an orphaned row gets a second chance independent of this forward-only cursor. Whichever you pick, add a regression test where `enqueueReport` rejects once after a successful `recordHour`, and assert the hour is retried on the next `dispatchForOrg` call.

langwatch/src/server/app-layer/presets.ts
```diff
@@ -587,6 +600,43 @@ export function initializeDefaultApp(options?: {
...
+          .commands.reportStorageForHour.send(data),
+      // Per-org guard: the reactor dedups per project, so without this an org
+      // with several projects would re-run the heavy measurement per project.
+      // A short-lived Redis lock lets one dispatch win and the rest skip (the
+      // winner covers every sealed hour). No Redis → run unguarded.
+      runExclusivePerOrg: redis
+        ? async (organizationId, fn) => {
+            const key = `storage_dispatch_org:${organizationId}`;
+            const acquired = await redis.set(key, "1", "EX", 120, "NX");
+            if (acquired !== "OK") return;
```
Aryansharma28, line 666: [P0] Per-org Redis lock has no fencing token — TTL expiry during a long catch-up run defeats its whole purpose

**Why:** This lock was added (commit `2171a77d`, "P2 from review") specifically to stop an org's several projects from re-running the heavy `_size_bytes` measurement concurrently — the query this same PR's description calls out as having caused two production OOM incidents. As written, the lock: (1) stores a fixed value with no unique owner/fencing token, (2) has a flat 120s TTL, and (3) unconditionally `redis.del(key)`s in `finally` regardless of whether this process instance still owns the key.

If `fn()` (the full `catchUp` gap-fill) runs longer than 120s, the key expires mid-run. A worst-case gap-fill measures up to `STORAGE_BACKFILL_MAX_HOURS` = 840 sequential hours, each doing its own heavy ClickHouse query — well past 120s unless each hour resolves in under ~140ms. Once the TTL lapses, a second event for a sibling project of the same org can acquire the lock and start the identical heavy measurement concurrently — exactly the race this code exists to prevent. When the first process eventually finishes and calls `redis.del(key)`, it deletes the *second* process's still-live lock (no ownership check), opening the door for a third concurrent run. Under sustained load or a slow ClickHouse, this can cascade into unbounded concurrent execution of the query pattern tied to the two prior OOM incidents — the dedup guard actively fails at the moment it matters most (a long, real gap-fill).

**Evidence:**
```ts
const acquired = await redis.set(key, "1", "EX", 120, "NX");
if (acquired !== "OK") return;
try {
  await fn();
} finally {
  await redis.del(key);
}
```

**Suggested fix:** Use a fencing token: `SET key <uuid> EX <ttl> NX`, release via a small Lua script (`redis.eval`) that only deletes the key if its current value still matches this process's token (the standard Redlock safe-unlock pattern) — this alone fixes the "delete someone else's lock" half of the bug. Separately, either heartbeat/extend the TTL while `fn()` runs, or size the TTL to genuinely cover the worst-case 840-hour gap-fill duration rather than a flat 120s, so the lock doesn't expire mid-run in the first place.

langwatch/src/server/app-layer/presets.ts
```diff
@@ -587,6 +600,43 @@ export function initializeDefaultApp(options?: {
...
+            try {
+              await fn();
+            } finally {
+              await redis.del(key);
+            }
+          }
+        : undefined,
+    });
```
Aryansharma28, line 695: [P2] Silent unguarded fallback when Redis is unavailable in SaaS mode

**Why:** `runExclusivePerOrg` is `undefined` whenever `redis` is falsy, and `StorageMeterDispatchService.dispatchForOrg` treats a missing guard as "run unguarded": `this.deps.runExclusivePerOrg ?? ((_org, fn) => fn())`. In `config.isSaas` mode this silently removes the only protection against the concurrent-heavy-measurement scenario (the same one flagged in the P0 above) with no log line calling it out — an operator would have no signal that the safety net is gone. `redis` is only null when `config.skipRedis` is set, which shouldn't happen in production SaaS, but nothing enforces that `isSaas: true` and `skipRedis: true` can't co-exist (e.g. a misconfigured staging-as-SaaS environment), and the failure mode is silent rather than fail-closed.

**Suggested fix:** When `config.isSaas` is true and `redis` is null, either skip storage-meter dispatch entirely with a one-time warning log, or at minimum log a warning from inside the unguarded fallback path so this specific misconfiguration is visible instead of silent.

langwatch/src/server/event-sourcing/projections/global/storageMeterDispatch.reactor.ts
```diff
@@ -0,0 +1,53 @@
...
+          { projectId: context.tenantId },
+          "orphan project detected, has no organization -- skipping storage dispatch",
+        );
+        return;
+      }
+
+      try {
+        await deps.getDispatch()({ organizationId });
+      } catch (error) {
+        logger.warn(
```
Aryansharma28, line 55: [P2] No escalation on persistent dispatch failures — inconsistent with the sibling reporting command

**Why:** Every failure from `deps.getDispatch()` — a transient ClickHouse blip, a persistent outage, or even the 840h-ceiling ALARM path re-throwing after logging — lands here and is logged at `warn` with no `captureException`/Sentry scope and no circuit breaker. Compare `ReportStorageForHourCommand`, which trips a breaker after `MAX_CONSECUTIVE_FAILURES` (5) and calls `captureException` on both permanent Stripe rejections and unexpected errors. Here, as long as the org keeps generating events, a sustained outage in `measureBytesAt` or `enqueueReport` will just log warnings indefinitely with no page and no cap — on-call has no way to notice a multi-hour or multi-day gap opening up for a given org.

**Suggested fix:** At minimum, route repeated/unexpected errors through the same `captureException`/`withScope` pattern used elsewhere in this feature (`reportStorageForHour.command.ts`) so they surface in Sentry, and consider a lightweight per-org consecutive-failure counter (reusing `StorageBillingCheckpoint` or a similar mechanism) so a stuck org gets flagged rather than silently warn-logged forever.

Aryansharma28 (commented):
## Goated review — ADR-027 Phase 4 storage meter dispatcher

**Verdict: two P0 blockers, in exactly the two places this PR is supposed to get right** — idempotent billing under retries, and the per-org concurrency guard added specifically to stop a repeat of the two prior production OOM incidents. Everything else in the design is genuinely solid (see P5s below) — this is well-thought-out billing infra let down by two implementation bugs in its own safety mechanisms.

**P0: 2 | P1: 0 | P2: 2 | P3: 1 | P4: 0 | P5: 2**

CI is green, unit tests pass, and CodeRabbit auto-review is disabled on this base branch (stacked PR) — none of that covers the two issues below; both are logic/race bugs that current tests don't exercise.

### P0 BLOCKER (must fix before merge)
1. `storageMeterDispatch.service.ts` (~L118-143) — an `enqueueReport` failure *after* `recordHour` succeeds permanently and silently drops that hour's Stripe report. No code path in the repo re-drives an hour by `reportedAt IS NULL`; the forward-only cursor treats "row exists" as "done." See inline comment for the full trace.
2. `presets.ts` (~L625-636) — the per-org Redis lock (`SET key 1 EX 120 NX` / unconditional `DEL` in `finally`) has no fencing token. A catch-up run that outlives the 120s TTL (very plausible near the 840h Stripe ceiling, given each hour re-runs the historically OOM-prone `_size_bytes` query) lets a second process acquire the same lock and run the exact heavy measurement concurrently — the scenario this lock exists to prevent — and the first process's `finally` then deletes the second process's live lock, opening the door to a third concurrent run.

### P2 (strong recommend — fix in this PR, not release-blocking on its own)
3. `presets.ts` — when `redis` is falsy, `runExclusivePerOrg` becomes `undefined` and the service silently runs fully unguarded (`this.deps.runExclusivePerOrg ?? ((_org, fn) => fn())`). In `config.isSaas` mode this removes the one safeguard against re-triggering the OOM-prone concurrent-measurement scenario, with zero logging of the fact.
4. `storageMeterDispatch.reactor.ts` (~L46) — every dispatch failure (measurement error, enqueue error, the 840h-ceiling ALARM's own propagated exception) is swallowed at `logger.warn` with no `captureException`/circuit-breaker, unlike the sibling `ReportStorageForHourCommand` which trips a breaker after 5 consecutive failures and reports to Sentry. A sustained ClickHouse or queue outage will just log warnings forever with no page, for as long as the org keeps generating events.

### P3 (non-blocking, worth doing)
5. No unit test/spec scenario covers "`enqueueReport` throws after `recordHour` succeeds" — the exact case in P0 #1. This test would have caught the bug; recommend adding it alongside whichever fix is chosen.

### P5 (praise / cross-PR observations)
- Genuinely nice design: narrow capability-shaped DI interfaces (`StorageMeterDispatchDeps` depends on behavior, not concrete services), "thin reactor / fat service" split, the stateless durable cursor (correct across pods/restarts by construction), forward-only semantics for brand-new orgs (never surprise-backfills history), and a loud alarm instead of silent truncation at the Stripe 840h ceiling. The BDD feature file maps cleanly 1:1 to the unit tests.
- Cross-PR observation (not in this diff, so noted here rather than inline): `langwatch/src/utils/dbOrganizationIdProtection.ts`'s allowlist comment for `StorageUsageHourly` states, present-tense, that "the billing reporter... drains `reportedAt IS NULL` across every org as a background sweep." I grepped the whole repo (this PR and its base) and found no such sweep — only the ADR's Phase-6-deferred reconciliation job, which compares *already-reported* totals against Stripe and would not catch an unreported orphan hour. If that sweep genuinely doesn't exist yet, it's exactly the safety net that would make P0 #1 non-fatal, and the comment should read "will drain" rather than "drains" until it's built.

- Production p99 latency of `measureBytesAt` (the `_size_bytes` ClickHouse query) — this determines how often the 120s lock TTL in P0 #2 is actually exceeded in practice. Given the PR's own "two-prod-OOM" history for this query, I'm treating "sometimes slow enough to matter" as the working assumption rather than a purely theoretical edge case.
- Whether `es.getPipeline(...).commands.reportStorageForHour.send()` can realistically throw in production vs. being effectively non-failing — nothing in the codebase suggests `.send()` is guaranteed non-throwing, so I've treated it as fallible for P0 #1.

Flag is default-OFF (`release_storage_billing_metering`), so this PR merging is safe today — but both P0s should be fixed before Phase 4.5 flips the flag on for any paying org.

langwatch/src/server/app-layer/billing/storageMeterDispatch.service.ts
sergioestebance: Fixed in 5ce394b53. Added `findStaleUnreportedHours` (org + `reportedAt IS NULL` + `createdAt` older than a 10-minute grace window) and re-enqueue those hours at the start of every `catchUp` run, before the forward gap-fill. Re-enqueuing an hour that turns out to already be fine is safe — the report command is idempotent on its own `reportedAt` cursor and Stripe's deterministic per-hour identifier. Added regression coverage for both the healing path and the not-aborting-on-retry-failure case.

langwatch/src/server/app-layer/presets.ts
sergioestebance: Fixed in e1b670fad. Added a randomUUID fencing token (mirrors ReplayRedisRepository's acquireLock/releaseLock convention already used elsewhere in the codebase) so release only deletes a lock this process still holds, never a subsequent holder's. TTL is no longer a flat 120s guess — it's derived from the true worst case: STORAGE_MAX_HOURS_PER_RUN (168) x the ClickHouse query's own max_execution_time cap (45s) x 2 margin = ~4.2h, so a genuinely-slow-but-not-pathological catch-up can never outlive the lock.

langwatch/src/server/app-layer/presets.ts
sergioestebance: Fixed in e1b670fad — logs a warning once at boot when `config.isSaas` is true but Redis is unavailable, instead of the fallback silently running every org unguarded with no signal.

langwatch/src/server/event-sourcing/projections/global/storageMeterDispatch.reactor.ts
sergioestebance: Fixed in 0279b81d7 — routes the error through the same withScope/captureException pattern reportStorageForHour.command.ts already uses, so a persistent outage surfaces in Sentry instead of only a warn log.

Skipping the consecutive-failure-counter suggestion: reusing StorageBillingCheckpoint for this would conflate two different failure classes onto one breaker — Stripe permanent rejection (what that counter currently means) vs. a dispatch-level CH/queue error here. Sentry visibility covers the "on-call has no signal" gap without that risk; a dedicated dispatch-level counter is a reasonable follow-up but felt like more than this fix needed.

Aryansharma28 (approved):
All four issues addressed — the P0 stale-hour sweep () and the Redis fencing token fix look solid, and the P2 boot-time warning + Sentry escalation are in. CI green. Approving.

sergioestebance:

PR #5228, opened 2026-06-30
feat(billing): storage price contract + pin guard (ADR-027 Phase 5, code side)

sergioestebance:
Code-side of Phase 5: the price math as a single source of truth + a pin test, and single-sourcing the meter event name into the report command.

**What's here (code):** `storagePricing.ts` — €3/logical-GiB-month headline, derived per-MiB-hour unit price (≈ €0.0000040690), Stripe `unit_amount_decimal` in cents, €0.10/GiB-day round-trip, `sum` aggregation, and the meter event name (now imported by the report command, so they can't drift). 5 bound `@unit` scenarios pin the math so a future edit can't silently mis-bill.

**What's external (NOT code — gated on humans/Stripe):**
- Create the `STORAGE_GB` meter + metered prices in Stripe (test then live).
- Wire their real ids into `stripeCatalog.json` + `STRIPE_METER_NAMES` / `STRIPE_PRICE_NAMES` (per plan×currency×interval).
- €3/GiB pricing-owner sign-off (ADR open question).

The pin guard is ready for the catalog price's `unit_amount_decimal` to be checked against it.

Stacks on #5227 (dispatcher).

### DIFF
--- a/langwatch/ee/billing/stripe/__tests__/storagePricing.unit.test.ts
+++ b/langwatch/ee/billing/stripe/__tests__/storagePricing.unit.test.ts
@@ -1,0 +1,65 @@
+/**
+ * Price-pin guard for ADR-027 storage billing. These constants ARE the pricing
+ * contract; a silent edit here (or a divergent catalog `unit_amount_decimal`)
+ * would mis-bill every customer. The test pins the headline, the derived unit
+ * price, and the round-trip so a change forces an explicit, reviewed update.
+ *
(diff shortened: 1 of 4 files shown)


sergioestebance:
Blessed. 🙏 *"A false balance is an abomination to the Lord, but a just weight is his delight."* — Proverbs 11:1. The price is pinned to the cent and single-sourced with the meter event name. Ship it; the catalog's real `unit_amount_decimal` just has to match this guard when the Stripe meter is created.

PR #3243, opened 2026-04-15
perf: lazy-load drawer registry (4.3 MB → 232 KB)

0xdeafcafe:
## Summary
- All 44 drawer components in `drawerRegistry.ts` were eagerly imported, pulling their transitive deps (monaco-editor, shiki, OTel SDK, form builders, etc.) into a single **4,343 kB** command-bar chunk loaded on every page
- Converted all drawer imports to `React.lazy()` so they're code-split and only fetched when a drawer is actually opened
- `CurrentDrawer` already wraps in `<Suspense>` — added a spinner fallback for the loading state

**command-bar chunk: 4,343 kB → 232 kB (94.7% reduction)**

## Test plan
- [ ] Open various drawers (trace details, prompt editor, evaluator editor) and verify they load with a brief spinner
- [ ] Verify drawer navigation (back button, flow callbacks) still works
- [ ] Check no regressions on initial page load performance

PR #5277, opened 2026-07-02
fix(drawers): flushSync drawer navigations so they actually render

sergioestebance:
## Summary
- Opening a drawer (e.g. "+ Add" on Prompts or Agents, or navigating between drawers such as clicking "Evaluator" inside "Add to Evaluation") updated the URL but the drawer never visually appeared — until something else forced a re-render. A hard refresh "fixed" it (until the next fresh drawer type).
- Closing had the mirror-image symptom: Cancel/X updated the URL to strip `drawer.open`, but the drawer stayed open on screen.
- Root cause: every drawer is `React.lazy()` (#3243, "perf: lazy-load drawer registry, 4.3MB → 232KB") rendered inside one `Suspense` boundary in `CurrentDrawer`. React Router v7's `RouterProvider` wraps every navigation's state update in `React.startTransition` by default. Under React 18 semantics, when a transition-wrapped update causes a Suspense boundary to suspend for the *first time*, React keeps the previously committed UI on screen instead of showing the fallback — so the first-time-this-session mount (or unmount) of a lazy drawer silently does nothing until an unrelated re-render happens to resolve it.
- Fix: thread `flushSync` through drawer navigations (open, close, and the two URL-cleanup paths in `CurrentDrawer`), forcing `ReactDOM.flushSync` instead of `startTransition` for these specific updates. `flushSync` requires `RouterProvider` from `"react-router/dom"` (it's a documented no-op otherwise) — swapped in `main.tsx`.

## Test plan
- [x] `pnpm typecheck` clean
- [x] Verified live in dev: "+ Add" on Prompts or Agents now opens "Add to Evaluation" immediately (no refresh needed); clicking "Evaluator" inside it navigates to the evaluator list drawer immediately; Cancel/X now closes the drawer immediately

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved drawer open/close reliability by ensuring immediate UI commitment after URL updates (including lazy-loaded drawer content).
  * Cleared drawer-related URL state more consistently for restricted users and error fallbacks to prevent stale drawer views.
* **New Features**
  * Added a `flushSync` navigation option to the Next.js compatibility layer and forwarded it through router navigations.
* **Tests**
  * Added assertions verifying drawer navigation calls include `flushSync: true`.

### DIFF
--- a/langwatch/src/components/CurrentDrawer.tsx
+++ b/langwatch/src/components/CurrentDrawer.tsx
@@ -61,7 +61,9 @@ export function CurrentDrawer({ marginTop }: { marginTop?: number }) {
       .getState()
       .openLiteMemberRestriction({ resource: restrictedResource });
 
-    // Clear drawer from URL so it doesn't persist in browser history
+    // Clear drawer from URL so it doesn't persist in browser history.
+    // flushSync: see useDrawer.ts closeDrawer for why plain push can leave
+    // this update uncommitted.
     void router.push(
       "?" +
         qs.stringify(
@@ -72,7 +74,7 @@ export function CurrentDrawer({ marginTop }: { marginTop?: number }) {
           ),
         ),
       undefined,
-      { shallow: true },
+      { shallow: true, flushSync: true },
     );
   }, [isRestricted]); // eslint-disable-line react-hooks/exhaustive-deps
 
@@ -131,7 +133,7 @@ export function CurrentDrawer({ marginTop }: { marginTop?: number }) {
                 ),
               ),
             undefined,
-            { shallow: true },
+            { shallow: true, flushSync: true },
           );
         }}
       >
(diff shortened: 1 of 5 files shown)


Aryansharma28 (dismissed):

Aryansharma28 (approved):

Aryansharma28:
Verified this fixes the "Add Comparison" bug reported today in the experiments workbench — cherry-picked b0ff3049c onto current main in a scratch worktree and confirmed:

- Before: clicking "+ Add"/"Add Comparison" on Prompts or Agents updated the URL (`?drawer.open=targetTypeSelector`) but the "Add to Evaluation" drawer never rendered — first-time-this-session drawer mounts silently no-op, exactly as described in the root-cause writeup.
- After cherry-picking this fix: the drawer renders immediately on first click, and the full flow (Add → Prompt → select → Add Comparison → drawer again) works end-to-end.

Separately (not related to this fix, still reproduces with or without it): opening the "Add" drawer and adding a target triggers repeated `Maximum update depth exceeded` console errors right after the `experiments.saveEvaluationsV3` autosave mutation completes. Doesn't block interaction but is a real infinite-render-loop somewhere in the workbench/store sync path — tracking separately.

+1 to merge this once CI is green.

PR #5280, opened 2026-07-02
fix(langy): stop proposal-handler registration from looping the workbench

Aryansharma28:
## Summary
- Found while investigating the "Add Comparison" bug in the experiments workbench (see #5277): visiting any experiments workbench page and adding a target triggers an infinite render loop — the console repeats `Maximum update depth exceeded` forever after the autosave (`experiments.saveEvaluationsV3`) resolves.
- Root cause: `workbench/[slug].tsx` builds a `proposalHandlers` object (its Langy action menu) via `useMemo` with several tRPC mutation/utils objects as dependencies that don't have stable identity across renders, so the object is effectively new on most renders. `useRegisterLangyHandlers` has `handlers` in its effect's dependency array, so it fires every time and calls `registerHandlers` → `setProposalHandlers` (React **state**) inside `LangyProvider`. That state update re-renders the whole subtree under the provider — including the workbench page itself — which produces a new `proposalHandlers` object again. Loop.
- This isn't gated by Langy access — `LangyProvider` always wraps page content regardless of who can see the panel (only the visible chat UI is conditionally rendered), so this fires for **every** user on the workbench page, not just Langy users.
- Fix: `proposalHandlers` never needed to be React state — every consumer (`LangySidebar`'s `applyProposal`) only reads it at proposal-click time, never during render. Store it in a `ref` instead. Registering a new handlers object no longer triggers a re-render, so the loop can't start. `experimentSlug` stays as state since it *is* rendered (the "On: `<slug>`" panel subtitle).

## Test plan
- [x] `pnpm typecheck` clean
- [x] Verified live: navigated to a fresh experiments workbench, added two prompt targets (triggering the autosave mutation twice), watched the browser console for 6+ seconds after each save — 0 errors, vs. the repeating `Maximum update depth exceeded` before this fix.
- [x] Confirmed the Langy proposal-apply flow (`applyProposal` reading `proposalHandlersRef.current`) still type-checks correctly through `LangyDrawer` → `LangyPanel`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proposal handling in the Langy sidebar by switching to a ref-based handler flow, improving reliability when opening and applying suggestions.
  * Fixed cases where “no handler” behavior could be triggered incorrectly, helping proposals apply more consistently.

### DIFF
--- a/langwatch/src/components/langy/LangyContext.tsx
+++ b/langwatch/src/components/langy/LangyContext.tsx
@@ -1,10 +1,12 @@
 import {
   createContext,
   type ReactNode,
+  type RefObject,
   useCallback,
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from "react";
 import type { ProposalHandlers } from "./MessageContent";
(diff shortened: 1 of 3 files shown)


Aryansharma28:
all good to go

PR #5306, opened 2026-07-02
fix(langy): preflight health check before reserving a PR permit

Aryansharma28:
## Summary
- Langy's call to the agent backend (`/langy/chat` → agent `/chat`) reserved a daily PR permit and made a single 120s-timeout fetch with no way to tell "agent is completely dead" apart from "agent is up but failed mid-task" — a fully-down agent could burn up to 2 minutes and a permit before the user saw an error.
- Adds a 3s `/health` preflight **before** the permit is reserved. A fully-down agent now fails in ~3s and never touches the permit system at all.
- Originally this PR also added a bounded retry on `/chat` itself, but that was dropped after a CodeRabbit review (verified independently against the reference agent implementation) found it could duplicate side effects: the agent's per-conversation worker has no idempotency key, so retrying after a 5xx/reset could replay the same prompt as an independent second turn on a worker that already accepted and started the work — risking a duplicate PR. Fixing that safely needs an idempotency mechanism on the agent side, out of scope here. A mid-task failure now surfaces to the user immediately, same as before, just distinguished cleanly from the "agent never got the request at all" case.

## Test plan
- [x] `specs/langy/langy-github-prs.feature` — scenario: fails fast when the agent is completely unreachable (no daily permit consumed)
- [x] `langy-github-prs.integration.test.ts` — 14/14 passing, including permit-release regression tests updated to assert single-attempt (no retry) behavior
- [x] `pnpm typecheck` clean

## Summary by CodeRabbit

* **New Features**
  * Added a bounded agent availability pre-check before handling chat, with clear “temporarily unavailable” messaging when the agent can’t be reached.

* **Bug Fixes**
  * Improved GitHub PR opening reliability: fails fast when the agent is unreachable (no permit consumed) and returns correct errors when the agent rejects or errors (no retry, single `/chat` attempt, permits released correctly).

* **Tests**
  * Strengthened unit and integration test coverage for health preflight, timeouts, and exact request/permit-call behavior.

### DIFF
--- a/langwatch/src/server/routes/langy.ts
+++ b/langwatch/src/server/routes/langy.ts
@@ -247,6 +247,7 @@ async function fetchAgentChatWithRetry(opts: {
   try {
     const response = await attempt(AGENT_CHAT_TIMEOUT_MS);
     if (response.ok || response.status < 500) return response;
+    void response.body?.cancel();
     return await attempt(AGENT_CHAT_RETRY_TIMEOUT_MS);
   } catch (error) {
     if (isAbortTimeoutError(error)) throw error;
--- a/langwatch/src/server/routes/langy.ts
+++ b/langwatch/src/server/routes/langy.ts
@@ -507,7 +507,9 @@ langyRoute().post("/langy/chat", async (c) => {
 
   if (!agentResponse.ok) {
     // The agent answered, but with a non-2xx — same shape: no PR was
-    // opened so the permit must come back.
+    // opened so the permit must come back. Cancel the body so the
+    // connection doesn't stay held open (mirrors isAgentHealthy above).
+    void agentResponse.body?.cancel();
     await releasePermitIfUnused();
     logger.error(
       { status: agentResponse.status },
--- a/langwatch/src/server/routes/langy.ts
+++ b/langwatch/src/server/routes/langy.ts
@@ -199,11 +199,18 @@ const langyRoute = () =>
 
 const AGENT_HEALTH_CHECK_TIMEOUT_MS = 3_000;
 const AGENT_CHAT_TIMEOUT_MS = 120_000;
-const AGENT_CHAT_RETRY_TIMEOUT_MS = 15_000;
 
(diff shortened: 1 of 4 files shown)


langwatch/src/server/routes/langy.ts
```diff
@@ -458,12 +492,14 @@ langyRoute().post("/langy/chat", async (c) => {
         // change behavior — but the user's picker choice rides through.
         ...(modelOverride ? { modelOverride } : {}),
       }),
-      signal: AbortSignal.timeout(120_000),
+      signal: AbortSignal.timeout(AGENT_CHAT_TIMEOUT_MS),
```
langwatch-agent, line 495: **P3 — Cancel the failed `/chat` response body before returning 502.** The `!agentResponse.ok` branch below releases the permit and returns without consuming or cancelling `agentResponse.body`. If the agent returns a non-2xx with a body, especially a streamed/error body, that can leave the connection tied up until the remote closes or GC cleans it up. The new health preflight already cancels its response body; mirror that cleanup here with `void agentResponse.body?.cancel()` before logging/returning.

langwatch-agent (commented):
Reviewed current head `684cca8`. I left one non-blocking P3 comment on the failed `/chat` response cleanup path.

Validation run locally on this PR head:
- `pnpm --dir langwatch run test:unit src/server/routes/__tests__/langy-connect-card.test.ts`
- `pnpm --dir langwatch run test:integration src/server/routes/__tests__/langy-github-prs.integration.test.ts`
- `git diff --check origin/main...HEAD`

CI is green from `gh pr checks`.

sergioestebance (dismissed):

langwatch/src/server/routes/langy.ts
Aryansharma28: Verified against current code — this was marked resolved but the body-cancel call hadn't actually been added. Fixed in 31a2bb27f: `void agentResponse.body?.cancel();` before `releasePermitIfUnused()` in the `!agentResponse.ok` branch.

sergioestebance (approved):

PR #5312, opened 2026-07-02
test(billing): cover usageUnit/license-override half of the plan-snapshot fix

sergioestebance:
## Summary
Follow-up to #5310. CodeRabbit flagged that the regression test added there only proved the numeric threshold (`maxMessagesPerMonth`) and resolver call-count came from one plan snapshot — it didn't cover the other half of that bug: `resolveMeterDecision` also derives `usageUnit` and the license-override flag (`plan.planSource === "license"`) from the passed-in plan, and an independent second resolve there was part of the original inconsistency.

- `firstPlan` now uses `planSource: "license"`, `usageUnit: "events"`, `free: false`.
- `laterPlan` now uses `planSource: "subscription"`, `usageUnit: "traces"`, `free: false`.
- New assertion: `result.message` contains `"Monthly limit of 1000 events reached"` — only reachable if `firstPlan` fed both the count/threshold path and the meter-decision path; if the code regressed to resolve twice, the message would render from `laterPlan` instead (`"traces"`, and the threshold check would already fail at 1000 >= 2000).

Verified the added assertion actually catches the regression by reverting `usage.service.ts` to the pre-#5310 double-resolve locally and confirming this test fails there (then restored).

## Test plan
- [x] `usage.service.unit.test.ts` — 22/22 pass
- [x] Confirmed the new assertion fails against the pre-fix double-resolve implementation, confirming it's not a vacuous check

## Summary by CodeRabbit

* **Tests**
  * Updated unit test coverage for plan-limit behavior to use explicit plan snapshot data, including usage unit details.
  * Strengthened assertions to ensure the limit-reached message uses the active plan snapshot (the “first” plan) for consistent monthly event limit wording, even when a later plan would differ.

### DIFF
--- a/langwatch/src/server/app-layer/usage/__tests__/usage.service.unit.test.ts
+++ b/langwatch/src/server/app-layer/usage/__tests__/usage.service.unit.test.ts
@@ -306,7 +306,9 @@ describe("UsageService", () => {
         expect(result.maxMessagesPerMonth).toBe(1000);
         // "Monthly"/"events" only come from firstPlan (free: false, license
         // override with usageUnit "events"); laterPlan would render "Free"/"traces".
-        expect(result.message).toContain("Monthly limit of 1000 events reached");
+        expect(result.message).toContain(
+          "Monthly limit of 1000 events reached",
+        );
         expect(mockPlanResolver).toHaveBeenCalledTimes(1);
       });
     });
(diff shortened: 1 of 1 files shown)


sergioestebance:
Blessed. 🙏

*The integrity of the upright guides them, but the unfaithful are destroyed by their duplicity.* (Proverbs 11:3)

Ship whenever — mind the P3 spec mismatch if you want to tighten traceability.

Issue #5319, opened 2026-07-02
Coding-agent cost intelligence: content-block classification + cache-aware exact-token cost attribution (ADR-033)


sergioestebance:
## What

Implement ADR-033 (`dev/docs/adr/033-coding-agent-cost-intelligence.md`, Accepted): classify every content block of coding-agent CLI traffic at ingest-time enrichment into a 12-category taxonomy, split real provider-reported token usage per category (cache-tier aware), accumulate into trace/session folds, surface on the existing governance dashboards. Analytics only — never billing.

## Acceptance criteria (spec scenarios)

**specs/trace-processing/content-block-cost-attribution.feature** (12)
- Content blocks classified into cost categories
- MCP vs built-in tool activity distinguished
- Injected context separated from real user input
- Non-coding-agent spans not classified
- Per-category costs sum exactly to span cost
- Conservation holds on turn-1 cache creation
- Cached prefix priced at cache-read rate
- Unattributable usage lands in catch-all, never dropped
- Classification failure never fails ingestion
- Absent content skipped silently
- Forged classification attributes stripped
- Deterministic for replay

**specs/trace-processing/coding-agent-session-tracking.feature** (5)
- Steps accumulated into session view with category totals
- Start-time ordering survives out-of-order delivery
- Compaction detected on context re-base
- Small subagent step fires no compaction event
- Harness sessions keyed independently

**specs/ai-gateway/governance/cost-breakdown-dashboard.feature** (4)
- Personal usage breakdown by category
- Enablement hint when capture off
- Org activity monitor aggregate
- Never affects billing or plan limits

PR: #5318

sergioestebance:
## Implementation complete — stacked PR ladder

| PR | Rung | Status |
|---|---|---|
| #5318 | ADR-033 (Accepted) + BDD specs | base |
| #5320 | Domain: taxonomy, classifier, cache-aware cost allocation | stacked on #5318 |
| #5321 | Pipeline: ingest-time enrichment + trace-fold category totals | stacked on #5320 |
| #5323 | Session tracking: step series, compaction detection, rollup | stacked on #5321 |
| #5324 | Dashboards: /me + Activity Monitor category breakdown | stacked on #5323 |

**All 21 spec scenarios bound and green.** Final verification on the stack top: 2312 unit + 139 integration tests passing, feature-parity checker OK, typecheck clean (2 pre-existing errors on main excluded).

**Invariants → evidence:**
[...]

PR #5320, opened 2026-07-02
feat(cost-intelligence): content-block classification domain — taxonomy, classifier, cache-aware cost allocation

sergioestebance:
# Related Issue

- Resolves langwatch/langwatch#5319

## What

The **content-block cost-intelligence vertical** (ADR-033). Originally opened as PR A (the pure domain layer); the stack has since collapsed top-down, so this branch now carries the full feature — domain + ingest wiring + trace fold + dashboards — squash-merged from the child PRs (#5321, langwatch/langwatch#5323). Reviewing here reviews the whole feature below `feat/adr-coding-agent-cost-intelligence`.

**Domain** (`src/server/app-layer/traces/block-classification/`)
- **`categories.ts`** — 12-input + 6-output two-axis taxonomy (Decision 3), axis catch-all, and the ADR constants (`MCP_TOOL_PREFIX`, `MAX_CLASSIFIED_BLOCKS_PER_SPAN`, `MAX_TOKENIZED_CHARS_PER_BLOCK`, `MAX_LEADING_MARKERS`, reserved attribute keys).
- **`contextMarkers.ts`** — pure leading-context marker splitter (cursor-based, linear); `system-reminder`→`prior_context`, `mcp-instructions`→`mcp_tool_definitions`, skill/command tags→`skill_content`.
- **`blockClassifier.service.ts`** — deterministic classifier; prior-turn vs fresh-user framing, MCP-vs-builtin by `mcp__` prefix, `tool_use`→name resolution, per-axis block cap with catch-all overflow.
- **`costAllocation.service.ts`** — cache-aware three-tier exact-token allocation; each pool scales to the provider total so **Σ per-category cost ≡ the span's real cost**, with reconciliation to an authoritative provider `cost_usd` when present.
- **`harnessDetection.ts`** — pure `claude | codex | null` predicate.

**Ingest wiring + fold**
- **`span-block-classification.service.ts`** — ingest-time enrichment (serial, after cost/token enrichment, before content drop), gated by a **global + per-project kill switch** (`block-classification-killswitch` / `-project-killswitch`, parity with `OtlpSpanTokenEstimationService`). Stamps `langwatch.reserved.blocks.classification`, `.classifier_version`, and per-category `blockcat.<cat>.{tokens,cost_usd}`.
- **`recordSpanCommand.ts`** — invokes the classifier on coding-agent LLM spans, in its own `try/catch` (a throw is logged; ingestion never fails on classification).
- **`traceSummary.foldProjection.ts`** — rolls per-span blockcat totals into the trace summary (skipping the codex redundant-usage copy), plus a bounded per-trace session-step series for read-time session/compaction rollup.

**Dashboards** (behind `release_ui_ai_governance_enabled`)
- `/me` personal Usage breakdown and `/settings/governance` org-wide breakdown, powered by `PersonalUsageService.breakdownByCategory` / `ActivityMonitorService.categoryBreakdown` (IN-tuple dedup, `TenantId`-first, resource-guarded).

## Why

The gap ADR-033 targets is classification, not capture. Pool totals stay provider-exact and blocks scale to them, so category costs reconcile to the span's real (displayed) cost. All of it rides the existing event-sourcing enrichment + fold substrate — no new infrastructure, no schema changes. Analytics only: no billing, quota, or plan-limit path reads these numbers (pinned by `categoryBillingBoundary.integration.test.ts`).

## Deployment Impact

**Live ingest enrichment**, not a dormant module. On deploy, classification runs on coding-agent LLM spans that carry content and writes the reserved `blockcat`/session-step fields onto their trace summaries. It is **operable**: a global or per-project kill switch disables it on the hot path (cached flag lookups), and it is self-bounded (512-block cap, per-block tokenizer cap, bounded marker peel). The dashboards stay hidden until `release_ui_ai_governance_enabled` is enabled. No migrations, no schema changes; non-coding-agent traffic is untouched (harness predicate returns null before any work).

## Review fixes (this pass)

Addressed a code-review pass on the collapsed stack:
- **Tenant scoping** — the personal-usage governance-tenant query now filters `origin.kind = ingestion_source` (was matching non-ingestion rows).
- **Kill switch** — added the global + per-project switch above; the PR description now reflects the live wiring (the "no wiring / Deployment Impact: None" text was stale post-collapse).
- **Perf** — per-block tokenization fans out with `Promise.all` (was sequential); the leading-marker peel is linear + capped (was O(n²)); the session-step series' bounded re-serialization is documented against the fold's unbounded-collection guidance.
- **Shared cascade** — the custom per-token rate resolution (`resolveCustomTierRates`) is now shared by `computeSpanCost` and the block allocator so they cannot drift.
- **Copy / robustness / coverage** — the org dashboard caption no longer says "your tokens"; the allocator rejects non-finite pool totals; added a kill-switch unit test and the enterprise-ADMIN `categoryBreakdown` router happy-path test.


langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
```diff
@@ -111,6 +131,27 @@ function createDefaultDependencies(): RecordSpanCommandDependencies {
       featureFlagService,
     }),
     contentDropService: { dropSpanContent: applyOtlpSpanContentDrop },
+    blockClassificationService: new OtlpSpanBlockClassificationService({
```
Aryansharma28, line 133: **[BLOCKER]** `createDefaultDependencies()` unconditionally constructs `OtlpSpanBlockClassificationService`, and `handle()` (below, around :374-387) calls `classifySpanBlocks()` on every span for every tenant. This contradicts the PR description's "no pipeline wiring yet / Deployment Impact: None" — this is live, unconditional wiring into the production ingest path the moment this merges. No feature flag gates it (see the sibling `OtlpSpanTokenEstimationService` above, which does have `featureFlagService`).

langwatch/src/server/event-sourcing/pipelines/trace-processing/projections/traceSummary.foldProjection.ts
```diff
@@ -190,6 +237,43 @@ export function applySpanToSummary({
     cacheTokens.reasoningTokens,
   );
 
+  // ADR-033: roll each span's per-category block totals into trace-level running
+  // sums under the same reserved keys. A span whose tokens are excluded from
+  // accumulation (codex's redundant usage copy) must also skip its blockcat
+  // totals, or the trace would double-count the categories for that turn.
+  const blockCategoryDeltas = spanCostService.isTokenAccumulationSkipped(span)
```
Aryansharma28, line 250: **[BLOCKER]** This block-category/session-step fold logic runs unconditionally on every folded `SpanReceivedEvent` for every trace (no feature-flag gate), writing new fields into every `TraceSummary` in production as soon as this deploys — contradicts "Deployment Impact: None" in the PR description. Combined with the unconditional call in `recordSpanCommand.ts:134`, this PR is a full live vertical slice, not an inert domain-only change.

langwatch/ee/governance/services/personalUsage.service.ts
```diff
@@ -460,6 +489,179 @@ export class PersonalUsageService {
...
+      query_params.userKey = GOVERNANCE_ATTR.USER_ID;
+      query_params.userEmail = params.userEmail;
+    }
+
+    const result = await client.query({
+      query: `
+        SELECT
+          ${outer}
+        FROM trace_summaries ts
+        WHERE ts.TenantId = {tenantId:String}
```
Aryansharma28, line 645: **[BLOCKER]** This query reads the governance tenant filtered only on `Attributes['langwatch.user_id'] = userEmail`, missing the `Attributes['langwatch.origin.kind'] = 'ingestion_source'` filter that every other reader of this tenant applies — including this PR's own sibling `ActivityMonitorService.categoryBreakdown` and 4 pre-existing files (`governanceKpisSync.reactor.ts`, `governanceOcsfEventsSync.reactor.ts`, `quarantineFillEvaluator.service.ts`, `setupState.service.ts`). `langwatch.user_id` is stamped by the generic trace-attribute pipeline on *every* trace regardless of origin, and the hidden gov-project's `apiKey` has no `kind` filter in auth middleware — if that key is ever reused/leaked, ordinary trace traffic with a spoofed `langwatch.user_id` lands here undetected. Please add the same `outerOriginFilter`/`innerOriginFilter` pattern used in `activityMonitor.service.ts`.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
```diff
@@ -0,0 +1,416 @@
...
+    };
+  }
+
+  /**
+   * Resolves per-tier rates the same way `computeSpanCost` does, so Σ per-category
+   * cost reconciles to the span's real (token×rate) cost: custom enrichment rates
+   * first (`langwatch.model.*`), then the injected registry resolver. A cache rate
+   * that is absent falls back to the input rate (counted, not discounted).
+   */
+  private resolveTierPrices({
```
Aryansharma28, line 382: **[BLOCKER]** This hand-reimplements `computeSpanCost`'s rate-priority cascade (`model-cost-matching.ts`) rather than calling it — the docstring says it "mirrors" that cascade, but a future change to the real one won't propagate here, silently breaking the PR's headline invariant that Σ per-category cost ≡ the span's real cost. Worth calling the shared function directly instead of duplicating its logic.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
```diff
@@ -0,0 +1,416 @@
...
+   * the span is not coding-agent traffic or carries no attributable content.
+   */
+  async classifySpanBlocks({
+    span,
+    instrumentationScope,
+    tenantId: _tenantId,
+  }: {
+    span: OtlpSpan;
+    instrumentationScope?: OtlpInstrumentationScope | null;
+    /** Reserved for future per-project kill-switch parity with token estimation. */
```
Aryansharma28: **[BLOCKER]** `tenantId` is unused ("Reserved for future per-project kill-switch parity"). Unlike the sibling `OtlpSpanTokenEstimationService`, which is gated by `featureFlagService` (global + per-project), this service has no kill switch at all. Given it's actually wired live into the ingest path (see `recordSpanCommand.ts`), there's currently no way to disable classification for a misbehaving tenant/model without a code change + redeploy.

langwatch/src/server/app-layer/traces/session-rollup/sessionSteps.ts
```diff
@@ -0,0 +1,130 @@
...
+  harness,
+  startMs,
+  inputTokens,
+}: {
+  attributes: Record<string, string>;
+  harness: CodingAgentHarness;
+  startMs: number;
+  inputTokens: number;
+}): void {
+  const steps = parseSessionSteps(attributes[SESSION_STEPS_ATTR]);
```
Aryansharma28, line 134: **[BLOCKER]** `JSON.parse` (line 124) + `JSON.stringify` (line 128) of the entire growing step array runs on *every* qualifying fold step for every coding-agent span (up to 512 entries) — this reintroduces the exact O(n²) re-serialization pattern that `traceSummary.foldProjection.ts`'s own comment (around :413-420) says was deliberately removed for `events`/`scenarioRoleCosts`/`spanCosts`. A long coding-agent session now pays increasing serialization cost on every step.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
```diff
@@ -0,0 +1,416 @@
...
+      // any single block, and spool-reconstituted spans bypass the ingest
+      // value cap entirely — one adversarial multi-MB block must not hold the
+      // tokenizer (synchronous ingestion path) hostage. Tokenize a capped
+      // slice and extrapolate linearly: within-pool proportions shift only
+      // for pathological blocks, and pool totals stay exact via scaling.
+      const text =
+        block.text.length > MAX_TOKENIZED_CHARS_PER_BLOCK
+          ? block.text.slice(0, MAX_TOKENIZED_CHARS_PER_BLOCK)
+          : block.text;
+      let tokens = (await this.deps.tokenizer.countTokens(model, text)) ?? 0;
```
Aryansharma28: **[BLOCKER]** `tokenizer.countTokens` is awaited sequentially inside the `for` loop above (line 174) instead of via `Promise.all`. Every content block in a span now pays a sequential tokenizer round-trip, unconditionally, on every coding-agent span now that this is wired live.

langwatch/src/components/governance/CategoryBreakdownBars.tsx
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * One-line caption for the Usage breakdown section: a short summary of the
+ * headline categories with a (?) tooltip carrying the complete list. Shared by
+ * both surfaces so the summary phrasing stays identical.
+ */
+export function CategoryBreakdownCaption() {
+  return (
+    <HStack gap={1} color="fg.muted" fontSize="sm" marginBottom={3}>
+      <Text>
+        Where your tokens go: system prompt, MCP tools, skills, thinking, and
```
Aryansharma28: **[BLOCKER]** This hardcoded "Where **your** tokens go..." caption is reused verbatim on `pages/settings/governance.tsx` — the org-wide Activity Monitor **admin** dashboard, where the data is aggregate org/team spend, not the viewer's own. Factually wrong copy for that audience; the component takes no personalization prop to distinguish the two surfaces.

langwatch/src/server/app-layer/traces/block-classification/contextMarkers.ts
```diff
@@ -0,0 +1,106 @@
...
+  // peel O(n²) on adversarial leading-tag spam, on the synchronous ingest path).
+  // Sticky `LEADING_OPEN_TAG.lastIndex = pos` pins each match to the cursor, and
+  // we slice out only the small `raw` block, so the whole loop is linear.
+  let pos = 0;
+  while (markers.length < MAX_LEADING_MARKERS) {
+    LEADING_OPEN_TAG.lastIndex = pos;
+    const open = LEADING_OPEN_TAG.exec(text);
+    if (!open) break;
+    const leadingWhitespace = open[1] ?? "";
+    const tagName = open[2]!;
```
Aryansharma28, line 136: **[NON-BLOCKER]** Close-tag matching by name only, no nesting-depth tracking — nested/mismatched same-name tags (e.g. `<system-reminder><system-reminder>x</system-reminder></system-reminder>real question`) mis-scope the marker and leak a stray closing tag into `body`, which then gets classified as `user_input`. Bounded to analytics (never billing), but reachable on first-turn text.

langwatch/src/server/app-layer/traces/block-classification/blockClassifier.service.ts
```diff
@@ -0,0 +1,437 @@
...
+      this.overflowChars += charCount;
+      this.overflowTexts.push(text);
+      return;
+    }
+    // idx = sequential part index within this axis (position === array length
+    // before push), matching the space the cache-breakpoint index addresses.
+    this.blocks.push({ idx: this.blocks.length, category, charCount, text });
+  }
+
+  /** Mark the most recently pushed block as carrying a cache_control breakpoint.
```
Aryansharma28, line 176: **[NON-BLOCKER]** `markBreakpoint()` silently no-ops once `overflowed` is true, so a cache breakpoint past the block cap (511+) is dropped and that content prices as `fresh` instead of `cache_read`/`cache_creation` in the catch-all bucket. Total conservation still holds — only within-tier attribution drifts. The comment's assumption that "breakpoints never live in the overflow tail" is untested.

langwatch/src/server/app-layer/traces/block-classification/costAllocation.service.ts
```diff
@@ -0,0 +1,343 @@
...
+ * the running category totals, and price them. The last block absorbs the
+ * residual so the pool sums exactly regardless of float drift. A nonzero pool
+ * with no blocks (zero-guard) drops its whole total into the axis catch-all.
+ */
+function allocatePool(
+  pool: Pool,
+  totals: CategoryTotals,
+  blocks: AllocatedBlock[],
+): void {
+  if (pool.total <= 0) return;
```
Aryansharma28: **[NON-BLOCKER]** Negative `pool.total` is silently dropped here (vs. routed to catch-all like the zero-guard on line 289 does for `sum <= 0`), and `NaN` totals pass this check as `false` and propagate through `reconcileCosts`, corrupting every category's cost on the reconciliation path. Not reachable today since the only caller pre-sanitizes to `n > 0`, but the module has zero defensive validation despite its docstring claiming "provider-reported usage is ground truth" — and it's untested (no negative/NaN case in `costAllocation.unit.test.ts`).

langwatch/ee/governance/routers/__tests__/governance.rbac.integration.test.ts
```diff
@@ -55,233 +55,235 @@ import { createTestApp } from "~/server/app-layer/presets";
...
-        const caller = callerFor(memberUserId);
-        await expect(
-          caller.activityMonitor.summary({ organizationId, windowDays: 7 }),
-        ).rejects.toMatchObject({ code: "UNAUTHORIZED" });
-      });
+    // Every activityMonitor procedure gates on `activityMonitor:view`, so a
+    // MEMBER (who lacks it) is rejected before the resolver runs. Looping over
+    // the full procedure list closes the whole-router gap: a new procedure that
+    // forgets the permission guard fails here.
+    it.each(
```
Aryansharma28, line 247: **[NON-BLOCKER]** No test in this file or `license-gate-governance.integration.test.ts` exercises the ADMIN+enterprise-plan happy path for the new `activityMonitor.categoryBreakdown` procedure through the full router (RBAC guard + license gate + resolver together) — only the reject paths are covered. The only test that exercises `categoryBreakdown`'s data correctness instantiates `ActivityMonitorService` directly, bypassing both guards.

Aryansharma28 (dismissed):
## Deep review — 6 parallel review passes (domain logic, pipeline wiring, governance/billing multitenancy, frontend, removed-behavior/cross-file tracing, reuse/efficiency/conventions)

**TL;DR:** the block-classification and cost-allocation math itself is solid — no conservation-breaking bug found in the core `costAllocation.service.ts`/`blockClassifier.service.ts` algorithms, and they match ADR-033 Decision 2.3/2.4. The problems are (1) the PR description doesn't match what the diff actually does, (2) one real cross-tenant scoping gap, and (3) a couple of hot-path issues that only matter *because* of (1).

---

### 🔴 Blockers

**1. PR description says "no pipeline wiring yet… Deployment Impact: None" — that's not what the diff does.**
`createDefaultDependencies()` unconditionally constructs `OtlpSpanBlockClassificationService` and `RecordSpanCommand.handle()` calls `classifySpanBlocks()` on **every span, every tenant**, in production, the moment this merges:
- `langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts:134` (and the unconditional call at `:374-387`)
- `langwatch/src/server/event-sourcing/pipelines/trace-processing/projections/traceSummary.foldProjection.ts:244` (`applySpanToSummary`, called on every folded `SpanReceivedEvent`, no flag gate) and `:596-618` (`handleTraceLogRecordReceived`)
- Two live UI surfaces already render this data: `langwatch/src/pages/me/index.tsx:338-346` and `langwatch/src/pages/settings/governance.tsx:424-434`

This isn't "PR A of a domain-only stacked ladder" — it's a full vertical slice (classification → fold → API → UI) shipping under a description that told reviewers to expect isolated pure functions with zero runtime impact. The ADR's own `v3` revision note (added in this same PR) even refers to itself as "PR C." Worth reconciling the PR description/stack plan with what's actually landing here, and — since this is genuinely live now — everything below matters in production, not "later."

**2. Cross-tenant scoping gap in the new personal-usage category query.**
`langwatch/ee/governance/services/personalUsage.service.ts:590-651` (`queryCategoryTotals`, `WHERE` at `:632`) reads the org's hidden Governance Project tenant filtered only by `Attributes['langwatch.user_id'] = userEmail` — it's missing the `Attributes['langwatch.origin.kind'] = 'ingestion_source'` filter that **every other reader of that tenant applies**, including this PR's own sibling method (`activityMonitor.service.ts:categoryBreakdown`) and four pre-existing files (`governanceKpisSync.reactor.ts:76-77`, `governanceOcsfEventsSync.reactor.ts:88-89`, `quarantineFillEvaluator.service.ts:150-151`, `setupState.service.ts:154`). `langwatch.user_id` is stamped by the generic trace pipeline on *every* trace through `/api/collector`/OTel, independent of origin — and the hidden gov-project has a live `Project.apiKey` with no `kind` filter in the auth middleware. If that key is ever reused/leaked, ordinary trace traffic bearing a spoofed `langwatch.user_id` lands in this tenant undetected and gets summed into that user's cost breakdown. Root cause is duplicated query-building between this file and `activityMonitor.service.ts` (see non-blocker below) — they already drifted from each other on their first PR.

**3. `resolveTierPrices`/`resolveReconciliationCost` reimplement `computeSpanCost`'s rate cascade instead of calling it.**
`langwatch/src/server/app-layer/traces/span-block-classification.service.ts:330-397` — the docstring says it "mirrors" `computeSpanCost`'s priority cascade (`model-cost-matching.ts:25`), but it's a hand-copied reimplementation, not a call to the shared function. A future change to the real cascade (new rate source, new priority tier) won't propagate here, silently breaking the PR's headline invariant: "Σ per-category cost ≡ the span's real cost."

**4. No feature-flag / per-tenant kill switch for classification, unlike its sibling.**
`OtlpSpanTokenEstimationService` (the analogous ingest-time service) is gated by `featureFlagService` (global + per-project). `OtlpSpanBlockClassificationService` has none — the `tenantId` param exists but is explicitly commented `// Reserved for future per-project kill-switch parity` (`span-block-classification.service.ts:106-108`) and is unused. Combined with finding #1, there is currently no way to disable this for a misbehaving tenant/model without a code change + redeploy.

**5. O(n²) JSON re-serialization reintroduced on the ingest hot path.**
`langwatch/src/server/app-layer/traces/session-rollup/sessionSteps.ts:113-130` (`appendSessionStep`, called from `foldProjection.ts:267-274` and `:611-618`) does a full `JSON.parse` + `JSON.stringify` of the entire growing step array on **every** qualifying fold step, for every coding-agent span (up to 512 entries). This is exactly the pattern `foldProjection.ts`'s own comment (around `:413-420`) says was deliberately removed for `events`/`scenarioRoleCosts`/`spanCosts`. A long coding-agent session now pays increasing serialization cost on every step.

**6. Serial (not parallel) tokenizer calls in the ingest hot path.**
`langwatch/src/server/app-layer/traces/span-block-classification.service.ts:185` — `toTokenBlocks` awaits `tokenizer.countTokens` inside a sequential `for` loop (`:174`) instead of `Promise.all`. Every content block in a span now pays a sequential round-trip, unconditionally, on every coding-agent span (per #1/#4).

**7. Misleading copy on the org-wide admin dashboard.**
`langwatch/src/components/governance/CategoryBreakdownBars.tsx:134-164` (`CategoryBreakdownCaption`) hardcodes *"Where your tokens go: system prompt, MCP tools, skills, thinking, and more."* — this exact component is reused verbatim on `langwatch/src/pages/settings/governance.tsx:427`, the **org-wide Activity Monitor** admin dashboard, where the data is aggregate org/team spend, not the viewer's own. "Your tokens" is factually wrong for an admin looking at other people's usage.

---

### 🟡 Non-blockers (real, but not merge-blocking)

- **`langwatch/src/server/app-layer/traces/block-classification/contextMarkers.ts:90-95`** — the leading-tag peeler matches close tags by name only, no nesting depth; adversarial/malformed nested same-name tags (e.g. `<system-reminder><system-reminder>x</system-reminder></system-reminder>real question`) mis-scope the marker and leak a stray closing tag into `user_input`. Bounded to analytics, but reachable on (potentially attacker-influenced) first-turn text.
- **`langwatch/src/server/app-layer/traces/block-classification/blockClassifier.service.ts:157-163`** — `markBreakpoint()` silently no-ops once the per-axis block cap overflows, so a cache breakpoint past block 511 gets dropped and that content is priced as `fresh` instead of `cache_read`/`cache_creation` in the catch-all bucket. Total cost conservation still holds — only within-tier attribution drifts. The comment asserting "breakpoints never live in the overflow tail" is an untested assumption.
- **`langwatch/src/server/app-layer/traces/block-classification/costAllocation.service.ts:273`** — `if (pool.total <= 0) return;` silently drops negative pool totals (vs. routing to catch-all like the zero-guard on `:289` does for `sum <= 0`), and `NaN` totals pass this check as `false` and propagate through `reconcileCosts`, corrupting every category's cost on the reconciliation path. Not reachable today (`span-block-classification.service.ts`'s `extractUsagePools` pre-sanitizes to `n > 0`), but the module's own docstring claims "provider-reported usage is ground truth" with zero defensive validation to back that up — and it's untested (`costAllocation.unit.test.ts` has no negative/NaN case despite the review brief explicitly calling this scenario out).
- **Query-builder duplication is the direct root cause of blocker #2** — `personalUsage.service.ts:590-651` and `activityMonitor.service.ts:~519-618` independently hand-build near-identical 36-column category-breakdown ClickHouse queries. Recommend extracting one shared `queryCategoryTotals(tenantId, window, { originFilter?, userFilter? })` helper so the two can't drift again.
- **`langwatch/ee/governance/routers/__tests__/governance.rbac.integration.test.ts` + `license-gate-governance.integration.test.ts`** — no test exercises the ADMIN+enterprise-plan happy path for the new `activityMonitor.categoryBreakdown` procedure through the full router (RBAC guard + license gate + resolver together). The only test that exercises `categoryBreakdown`'s data correctness (`activityMonitorCategoryBreakdown.service.integration.test.ts`) instantiates the service directly, bypassing both guards.
- **`nanoid` used in 3 new integration test files** (`categoryBillingBoundary.integration.test.ts:15`, `personalUsageCategoryBreakdown.service.integration.test.ts:17`, `activityMonitorCategoryBreakdown.service.integration.test.ts:18`) — against the KSUID-only rule, though it matches the pre-existing local convention in that test directory (not a new deviation, but worth a follow-up lint/convention sweep).
- **`langwatch/src/components/me/usePersonalContext.ts:222-224`** — `spendByCategory` is recomputed inline every render without `useMemo`, inconsistent with sibling derived fields `budget`/`apiKeys` in the same hook (harmless — array is tiny).
- **`langwatch/src/components/governance/CategoryBreakdownBars.tsx`** hand-rolls a bar chart from `Box`/`HStack` instead of reusing `src/components/analytics/CustomGraph.tsx`, which already wraps recharts and supports a `horizontal_bar` type. Also bundles 4 components + a business-logic function (`toCategoryBarRows`) in one file.
- **`personalUsage.service.ts`** has 3 near-identical `gateway_budget_ledger_events` queries (~90 duplicated lines) and awaits several independent queries sequentially instead of `Promise.all`; `activityMonitor.service.ts` repeats the `resolveGovProjectId → getClickhouse` pattern sequentially across ~5 public methods. Not a correctness issue, just avoidable round-trips.
- **`detectCodingAgentHarness` is computed twice per span** — once at ingest (`span-block-classification.service.ts:110-113`) and again at fold time (`foldProjection.ts:262-265`) — should be threaded through as event data instead of recomputed.
- **`langwatch/ee/governance/services/activity-monitor/activityMonitor.service.ts:1350-1353`** (`eventsForSource`) — pre-existing, not touched by this diff, but its dedup subquery omits the `OccurredAt` partition filter every sibling query in the file carries, causing a full-tenant-history scan on every page. Sits directly on the code path this PR extends — worth fixing alongside it.
- **`langwatch/src/server/app-layer/traces/session-rollup/sessionRollup.service.ts`** — well-tested but has zero callers outside its own tests; genuinely inert today (consistent with the ADR's own "PR C" labeling for this half of session tracking).
- **`langwatch/src/server/app-layer/traces/block-classification/harnessDetection.ts:47`** — the `gen_ai.system === "claude_code"` branch isn't anchored to an instrumentation scope name check like the other branch is; low-probability but any customer instrumentation stamping that exact OTel attribute would be misdetected as a coding-agent harness.
- **`langwatch/src/server/app-layer/traces/block-classification/categories.ts:98-109`** — `MAX_TOKENIZED_CHARS_PER_BLOCK`/`MAX_LEADING_MARKERS` (real DoS bounds) aren't listed in the ADR's Constants table, which claims to be the source of truth for these.
- **Process note:** this PR carries a stale automated "low-risk" approval (`Approved by automation: PR qualifies as low-risk-change`) from an earlier, much smaller commit — a later bot comment on the same PR states it now **exceeds** the low-risk size threshold and requires manual review. Also, CodeRabbit's review was skipped entirely (base branch isn't the repo default branch), so no automated tool has actually reviewed this diff's content before this review.

---

### What's genuinely solid
The core allocation math (`allocatePool`/`partitionInput`/`reconcileCosts`) is correct through every edge case traced (zero tokens, first-turn cache_read=0, >512 blocks, empty content, unknown categories) and matches ADR Decision 2.3/2.4 exactly. No determinism issues, no `any`, no re-export shims, no FK/schema-name/KSUID violations in production code, layering (route→service→repository) is respected throughout, RBAC guard composition on the new `categoryBreakdown` procedure is correct and tested, the billing/analytics boundary is provably enforced (`categoryBillingBoundary.unit.test.ts` greps for and fails on any billing-code import of the classification module), and the big `governance.rbac.integration.test.ts` rewrite preserves all prior coverage while adding table-driven checks.

langwatch/src/server/event-sourcing/pipelines/trace-processing/commands/recordSpanCommand.ts
sergioestebance: Fixed in aaba8be1c + the description rewrite. You are right the wiring is live — the stack collapsed #5321 into this branch, so the "no wiring / Deployment Impact: None" text was stale (now corrected). Added a global + per-project kill switch (block-classification-killswitch / -project-killswitch) mirroring OtlpSpanTokenEstimationService, wired via featureFlagService in createDefaultDependencies and keyed by tenantId — the operable off-ramp this lacked.

langwatch/src/server/event-sourcing/pipelines/trace-processing/projections/traceSummary.foldProjection.ts
sergioestebance: Addressed. The kill switch (aaba8be1c) disables the blockcat cost writes — the classifier produces no attrs when off, so the fold rolls up nothing. The session-step series is a separate, bounded feature (session tracking): I considered gating it on the classifier marker but that breaks the Path B (codex-SSE) log path, which is never span-classified, and gating only the span path is an inconsistent asymmetry. So the boundary is: kill switch gates cost classification; session tracking is bounded metadata (512-cap, coding-agent traces only). Description corrected to state the live wiring.

langwatch/ee/governance/services/personalUsage.service.ts
sergioestebance: Fixed in ba43eb16a. Added the langwatch.origin.kind = ingestion_source filter to both the outer and dedup WHERE, gated on the same userEmail (governance-union-only) condition — matching activityMonitor.categoryBreakdown and the KPI/OCSF reactors.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
sergioestebance: Fixed in aaba8be1c. Extracted the duplicated custom per-token rate cascade into a shared resolveCustomTierRates (model-cost-matching.ts) that BOTH computeSpanCost (Priority 1) and this service now call, so they cannot drift. The accessor arg abstracts over the NormalizedAttributes map vs the OTLP span lookup.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
sergioestebance: Fixed in aaba8be1c. tenantId is now live (used for the per-project kill switch) and the service has the global + per-project switch, parity with OtlpSpanTokenEstimationService.

langwatch/src/server/app-layer/traces/session-rollup/sessionSteps.ts
sergioestebance: Addressed via documentation (50a041f4d) rather than a rewrite, deliberately. Unlike the unbounded events/spanCosts collections that were removed, this series is capped at MAX_SESSION_STEPS=512 (mergeAdjacentKeepMax on overflow), so per-step work is O(min(steps,512)) and total folding is bounded — never the multi-MB-per-span blob. A serialise-once fix needs a fold-framework finalize hook (the attribute map IS the per-step persisted snapshot); that infra change is disproportionate for a bounded cost. Documented the bound and the rationale in appendSessionStep.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
sergioestebance: Fixed in aaba8be1c. toTokenBlocks now fans out with Promise.all; block count is bounded by MAX_CLASSIFIED_BLOCKS_PER_SPAN so the fan-out width is bounded too, and order is preserved (idx/axis alignment unchanged).

langwatch/src/components/governance/CategoryBreakdownBars.tsx
sergioestebance: Fixed in 295e7b574. CategoryBreakdownCaption takes a scope prop ("personal" | "organization"); the governance admin page passes "organization" so it reads "Where the organizations tokens go", /me keeps the personal default.

langwatch/src/server/app-layer/traces/block-classification/contextMarkers.ts
sergioestebance: Acknowledged; deferring. Nested same-name context tags (<system-reminder><system-reminder>…) do not occur in real harness output — the markers are flat leading blocks — and the impact is analytics-only (a stray close tag mis-labels one block as user_input, conservation unaffected). The cursor rewrite in c628e143e did not add depth tracking; adding it is a real complexity bump for a case that is bounded and unreachable in practice. Tracking as a known limitation rather than fixing in this pass.

langwatch/src/server/app-layer/traces/block-classification/blockClassifier.service.ts
sergioestebance: Acknowledged; deferring. A cache breakpoint past the 511-block cap prices as fresh instead of cache-read/creation — but this needs >511 content blocks in one span (pathological), total conservation still holds, and it is a within-tier attribution drift only, which the ADR explicitly frames as "a model, not a measurement." Not worth the added complexity in this pass; noting as a known bound.

langwatch/src/server/app-layer/traces/block-classification/costAllocation.service.ts
sergioestebance: Fixed in 50a041f4d. allocatePool now guards with !(pool.total > 0), so a NaN (which slipped through pool.total <= 0) is skipped instead of poisoning every category cost. The reconcile path is separately guarded (reconcileToTotalCost != null && > 0, so NaN is skipped there too).

langwatch/ee/governance/routers/__tests__/governance.rbac.integration.test.ts
sergioestebance: Fixed in 295e7b574. Added the ADMIN+enterprise happy-path test for activityMonitor.categoryBreakdown through the full router (RBAC guard + license gate + resolver) in license-gate-governance.integration.test.ts — the suite previously covered only the reject paths.

0xdeafcafe:
Reviewed this. Findings below, most severe first.

## PR scope

Description says "PR A of the ADR-033 stacked ladder ... pure domain layer ... No pipeline wiring yet - later PRs consume these functions." Diff actually wires the classifier into `recordSpanCommand`, extends the trace-summary fold (`categoryTotals` + `session_steps`), adds `session-rollup` + three governance services + `/me` and Activity Monitor UI, and ~800 lines of integration tests. Either rewrite the title/body to reflect the real scope, or split it back into the promised A/B/C/D ladder. Can't meaningfully approve this as an inert domain slice.

## P1

`sessionRollup.service.ts:105` - compaction false-fires on the dominant coding-agent pattern. Steps `[200000, 8000, 10000, 12000, 190000]` (main thread spawns two subagent calls, then resumes) trips the candidate at i=1, both confirmations pass because they're below the *old* runningMax, event fires, runningMax resets to 8000, and the 190k rebase at i=4 then reads as legit growth from an 8k base. Confirmation should require `cur < drop threshold`, not `cur < old runningMax`. This is the exact false positive Decision 5 was rewritten to prevent.

`sessionRollup.service.ts:148` - silent split when `langwatch.thread.id` and `gen_ai.conversation.id` disagree. One logical session becomes two rollup buckets and both get wrong compaction/rollup numbers. Prefer `gen_ai.conversation.id` (the semconv-stable key), fall back to `langwatch.thread.id` only when the semconv key is absent.

`span-block-classification.service.ts:121` - "Ingestion never fails on classification" is written up as an ADR invariant of this file, but the file doesn't enforce it. Any tokenizer / classifier / allocator throw just escapes; the current try/catch lives one layer up in `recordSpanCommand`. Wrap the body here in try/catch too, log at warn, return silently. Keep the caller-level guard as a second belt.

`ADR-033:115` - ADR says `categoryTotals?: Partial<Record<Category, {tokens, costUsd}>>` on TraceSummary. Code stores flat reserved keys `langwatch.reserved.blockcat.<category>.{tokens,cost_usd}`. A consumer writing `trace.categoryTotals[cat].costUsd` from the ADR gets `undefined`. Either update the ADR to describe the flat encoding actually used (and add `SPAN_ATTR_BLOCKCAT_PREFIX` to the Constants table), or expose a structured mirror.

`blockClassifier.unit.test.ts` - fixtures are all Anthropic-shaped. ADR Decision 6 ships both harnesses in v1 but there's no Codex / OpenAI-shaped input anywhere in the classifier tests. `blockText` returns empty on shapes it doesn't know, so a production Codex rollout could quietly collapse into `other_input` / `other_output`. Add fixtures for OpenAI message shape (function calls, tool outputs, `role:'tool'`), assert non-empty categorisation, and check MCP-vs-builtin still works on the `function` name field.

`span-block-classification.service.unit.test.ts:210` - absent-content invariant only tested for "no attribute present". Decision 7 asks for a test per capture level (none / metadata_only / redacted / raw). `metadata_only` is the risky one - SDK strips bodies but leaves stubs, and if `messagesFromJson` ever changes shape, that path could quietly start emitting classified empty strings into the fold.

## P2

`harnessDetection.ts:36` - strict exact-string match on scope name gates the entire feature. Anthropic bumping to `com.anthropic.claude_code.events.v2`, or a collector processor case-normalising so `Codex_Cli_Rs` slips past, or an OTLP-HTTP SDK bug leaving a trailing newline all turn classification off across the fleet without a peep. Trim + lowercase before comparison, prefix-match on the Claude scope.

`blockClassifier.service.ts:350` - 600 tool definitions plus a system prompt burn through all 512 detail slots before the conversation walk. Fresh user turn then classifies as OTHER_INPUT and `lastInputCacheBreakpointIndex` drops (markBreakpoint returns early once overflowed). For any skills-heavy customer, the user_input lane reads near-zero and the cached-prefix boundary is wrong. Reserve N slots for the fresh-user walk, or apply a per-phase soft cap.

`blockClassifier.service.ts:84` - `contentToText → blockText → contentToText` recurses with no depth cap and no cycle detection. Malformed nested `tool_result` content (10k deep or self-referential) blows the stack or hangs the sync ingest loop. Even with an outer try/catch, a synchronous hang blocks the single-threaded command loop. Iterative walk, `Set` visited-guard, depth ceiling.

`blockClassifier.service.ts:197` - when a fresh user text part carries `cache_control` but the push emits zero blocks (empty text, no markers), `markBreakpoint()` marks the *previous* block in the accumulator, which came from an earlier message. `lastInputCacheBreakpointIndex` now points at the wrong block and cost allocation draws the cached-prefix boundary in the wrong place. Only mark if the current part actually appended.

`blockClassifier.service.ts:199` - `splitLeadingMarkers` runs on every fresh-user text part, not just the first. `[{text:'hello '}, {text:'<system-reminder>foo</system-reminder>world'}]` peels the mid-message tag as PRIOR_CONTEXT. user_input under-counted, prior_context inflated. Only peel the first part per message (track it on the accumulator).

`contextMarkers.ts:88` - nested same-name tags mis-peel because closeTag uses first `indexOf`. `<skill>outer<skill>inner</skill>tail</skill>question` truncates before `tail`, loop exits with a dangling close tag leaking into user_input. Depth-count intervening opens before accepting a close.

`blockClassifier.service.ts:174` - duplicate `tool_use` ids overwrite in `toolNames`, so an earlier `tool_result` resolves to the later tool's tier. First-write-wins matches temporal order: `if (!toolNames.has(id)) toolNames.set(id, name)`.

`recordSpanCommand.ts:248` - ADR Decision 1 says classification runs *before* the attribute cap. Wired code runs `capOversizedAttributes` at 248, which is before Promise.allSettled at 305, which is before classification at 371 - reverse order. A span just over the per-value cap gets `langwatch.input` truncated first, so `parseMessages` returns null and `classifySpanBlocks` no-ops on an otherwise classifiable span. Either move the cap after classification, or update the ADR text (the docstring on the service already rationalises the reverse; that logic just needs to land in the ADR).

`span-block-classification.service.ts:342` - `resolveTierPrices` treats partial custom rates as complete. If the customer stamps only `inputCostPerToken`, `customInput !== null` triggers the branch and `outputCostPerToken` silently becomes `customOutput ?? 0 = 0`. `assistant_text` lane costs 0 while `computeSpanCost` uses the registry fallback for output - Σ per-category drifts from displayed. Only take the custom branch when *both* sides are set.

`span-block-classification.service.ts:209` - not idempotent. Re-classifying appends duplicate blockcat attributes instead of replacing. Any replay path (partial-failure retry, projection version bump, fan-out enricher) leaves duplicates; downstream fold either first-wins (silently drops new) or sums (silently doubles). Strip prior attributes first, or guard on presence.

`sessionRollup.service.ts:120` - end-of-session compaction never counts when fewer than `COMPACTION_CONFIRMATION_STEPS` remain. `[..., 200000, 40000, 45000]` (user compacts, checks summary, closes terminal) - genuine event, silently dropped. When the confirmation loop exhausts the array with confirmed ≥1 and no step climbed back, count it.

`sessionRollup.service.ts:148` - whitespace-only thread id passes the falsy check and produces a phantom session view. `"langwatch.thread.id": "   "` becomes bucket key `claude:   `. Every trace with a whitespace id piles into one monster session with tens of thousands of steps. Trim both keys, treat post-trim empty as absent.

`activityMonitor.service.ts:586` - `categoryBreakdown` has no upper bound on `OccurredAt`. Any ingestion source with a forward-skewed clock leaks future-dated traces into the current window until the window slides past. Mirror the closed `[fromMs, toMs)` shape `PersonalUsageService.queryCategoryTotals` already uses.

`activityMonitor.service.ts:472` - `any(model) AS mostUsedTarget` is an arbitrary pick, not the most-used one. Alice hits haiku 1000x and opus once, dashboard shows opus as "Most-used". `argMax(model, cnt)` after a per-model rollup, or rename the wire field to `sampleModel` and drop the promise on the UI.

`sessionRollup.service.ts:200` - `accumulateCategoryTotals` sums negative values without a floor check. Any classifier regression stamping negative tokens surfaces as negatives in the rollup. `if (!Number.isFinite(v) || v < 0) continue;`.

`sessionRollup.service.unit.test.ts:152` - only "subagent-not-a-compaction" test uses one small step, which happens to break confirmation via `>= runningMax`. Multi-step bursts (the actual pattern) never run. Add `[200k, 8k, 10k, 190k]`, `[200k, 8k, 10k, 12k, 190k]`, `[200k, 8k, 10k, 12k, 14k, 190k]` - all should be 0 events and currently fire phantoms.

`recordSpanCommand.ts:371` - no feature flag on the classifier. If it turns out to CPU-stall on some tenant's content, the only mitigation is a code revert. `OtlpSpanTokenEstimationService` already threads a `featureFlagService` per-project for exactly this - mirror the pattern (`cost_intelligence_classifier`, default on).

`span-block-classification.service.unit.test.ts:237` - 8 MiB perf test explicitly declines to assert wall-clock. A future O(N²) walk hits the same structural caps and passes green while doubling ingest latency. Soft wall-clock guard (5s ceiling to survive CI noise) or assert on tokenizer-spy call count is O(blocks), not O(blocks²).

## Top P3s

`costAllocation.service.ts:248` - `lastCacheBreakpointIndex = -2` with 5 input blocks gives `slice(0, -1)` which drops the last block off the prefix (JS treats negative end as `length + end`). Conservation holds against pool totals so nothing surfaces, but the cache-tier split is quietly wrong. Guard the negative case at entry.

`costAllocation.service.ts:173` - `reconcileToTotalCost > 0` accepts `Infinity`. Any upstream div-by-zero turns `factor` to Infinity, every category's costUsd to Infinity, then serialises to null - silently zeroes the span from analytics. `Number.isFinite(x) && x > 0`.

`costAllocation.service.ts:339` - `totals[category]` bracket-writes on a plain object. Category is a bare string union, so if any future boundary admits `'__proto__'` the write mutates `Object.prototype.tokens` and every plain-object lookup in the process observes it. `Object.create(null)` for `totals`, or a `Map`, or validate against `CATEGORIES` before writing.

`traceSummary.foldProjection.ts:187` - `String(Number(prior) + delta)` on `cost_usd` accumulates float noise. A 400-span trace ends up storing `0.0004780000000000037` per category. Add a `addReservedCostSum` companion that re-rounds via `toFixed(10)`.

`recordSpanCommand.ts:12` - imports `matchModelCostWithFallbacks` from `~/server/background/workers/collector/cost`. App-layer reaching into the legacy background/workers module - one ADR-018 removal PR away from a silent break. Move the helper into `~/server/app-layer/traces/model-cost-matching` and update both call sites.

`sessionSteps.ts:122` - `appendSessionStep` does raw `push()`; sort only happens inside decimation on overflow. OTLP is unordered/retried, so most stored `session_steps` JSONs live in arrival order, not startMs order. `rollupSessions` sorts before compaction, but any other consumer (drawer, backfill, CH projection) sees a non-monotonic series. ADR Schema promises "ordered by span startTime" - keep the array sorted on every push (bounded at 512, O(n log n) is fine).

sergioestebance:
Thanks — thorough pass. Worked through all of it; ~24 fixed across `c0c6a3800…2ebd3ae0c`, the rest triaged below with reasoning.

**Scope** — description already rewritten last pass to the full vertical (the stack collapsed the child PRs into this branch).

**P1**
- **Compaction false-fire** (`sessionRollup:105`) — fixed in `c0c6a3800`. You're right the below-old-max confirmation was the bug. Fix: a `COMPACTION_MIN_RETAIN_RATIO` (0.1) floor skips subagent-tiny steps outright, and confirmation now requires steps stay below the **drop threshold**, plus an end-of-session count. Verified against your three counterexamples (`[200k,8k,10k,190k]`, `[…,12k,190k]`, `[…,14k,190k]` → 0) and the existing suite; added them as tests.
- **thread-id disagreement** (`:158`) — fixed: prefer `gen_ai.conversation.id`, fall back to `langwatch.thread.id`; whitespace-only ids treated as absent.
- **in-file ingest guard** (`:121`) — fixed in `b076dfb31`: body wrapped in try/catch (warn+swallow), command-level guard kept as second belt.
- **ADR categoryTotals encoding** (`ADR:115`) — fixed in `2ebd3ae0c`: ADR now documents the flat `blockcat.<category>.{tokens,cost_usd}` keys; `SPAN_ATTR_BLOCKCAT_PREFIX` added to Constants.
- **Codex/OpenAI fixtures** (`blockClassifier tests`) — **deferred, tracking.** Real gap. `blockText` handles Anthropic block shapes; whether real Codex collapses depends on what `codex-rollout-otlp` normalizes to (the ADR already flags Codex as weaker on the OpenAI shape, Decision 6). Adding OpenAI-shape handling (`tool_calls[].function.name`, `role:'tool'`) is a feature-sized change I'd rather land as a focused follow-up than bolt on here — filing an issue.
- **capture-level test** (`:210`) — **deferred** to the same follow-up (add `metadata_only` stub coverage).

**P2** — fixed: recursion depth-cap (`blockClassifier:84`), scope normalisation (`harnessDetection:36`), duplicate tool-id first-write-wins (`:174`), breakpoint-only-on-append (`:197`), markers-peeled-first-part-only (`:199`), nested-tag depth match (`contextMarkers:88`), idempotent attribute write (`:209`), `OccurredAt` upper bound (`activityMonitor:586`), `topK(1)` most-used model (`:472`), end-of-session compaction (`sessionRollup:120`), whitespace/negative rollup guards. **ADR ordering** (`recordSpanCommand:248`) — reconciled in the ADR (classification runs after the cap; reserved output is self-bounded + cap-exempt), per your "the docstring already rationalises the reverse." Kill switch (`:371`) was added last pass. **Partial custom rates** (`:342`) — the shared `resolveCustomTierRates` now makes classification and `computeSpanCost` price partial rates identically, so Σ still reconciles to the displayed cost; whether partial rates should fall back to the registry is a cost-module policy call, noted separately. **Slot starvation** (`:350`) — **deferred**: needs a per-phase soft cap in the accumulator; conservation is unaffected (only per-block detail + the breakpoint), so filing rather than rushing. **Perf test call-count** (`:237`) — **deferred**: will switch to an O(blocks) tokenizer-call-count assertion.

**P3** — fixed: negative-breakpoint slice guard, finite/positive reconcile, null-prototype totals, float-noise cost sums (`toFixed(10)`), `matchModelCostWithFallbacks` moved into `model-cost-matching` (composition root no longer reaches into `background/workers`), sessionSteps kept sorted on push.

ADR Revisions v4 records the pass. Filing follow-up issues for the three deferred items; will link them here.

sergioestebance:
Deferred items tracked in langwatch/tasks#101 (OpenAI/Codex classifier shape, per-phase slot cap, capture-level/perf test coverage).

PR #5321, opened 2026-07-02
feat(cost-intelligence): ingest-time block classification enrichment + trace-fold category totals

### LINKED ISSUE #5319: Coding-agent cost intelligence: content-block classification + cache-aware exact-token cost attribution (ADR-033)

sergioestebance:
# Related Issue

- Resolves #5319

PR B of the ADR-033 stack (#5319). Stacked on `issue5319/block-classification-domain` (PR A). Wires the merged block-classification domain into the span ingestion pipeline and the trace-summary fold. Test-first; analytics only — these numbers never feed billing, quotas, or plan limits.

## What
- **Enrichment service** (`span-block-classification.service.ts`, mirrors `span-token-estimation.service.ts`): for coding-agent CLI spans (Claude Code, Codex) that carry captured content, classify every content block, tokenize each block with the injected tokenizer, and allocate provider-reported usage across categories **cache-tier aware**. Pushes:
  - `langwatch.reserved.blocks.classification` — bounded per-block `{idx, category, tokens, cacheTier}` detail
  - `langwatch.reserved.blocks.classifier_version`
  - `langwatch.reserved.blockcat.<category>.tokens` / `.cost_usd` — per-category totals the fold rolls up
- **Pipeline wiring** (`recordSpanCommand.ts`): invoked **serial**, after the parallel PII/cost/token enrichment block (it consumes their outputs) and **before** content drop, in its own `try/catch` — a throw is logged and ingestion continues (ADR-033 "Ingestion never fails on classification"). Dependency is an optional partial-dep, defaulted in the composition root with a registry-backed per-tier price resolver (injected so the hot-path service stays decoupled from the prisma-backed cost module).
- **Trace fold** (`traceSummary.foldProjection.ts` + `SpanCostService.extractBlockCategoryDeltas`): roll each span's `blockcat.*` totals into trace-level running sums via the existing `addReservedTokenSum` transport, **skipping** spans excluded from token accumulation (codex redundant-usage guard) so categories are never double-counted.
- **Domain extensions** (on the PR A branch): the classifier now carries each block's flattened text (enabling **real per-block tokenization**, not the byte-proportional split the ADR rejects) and records the last `cache_control` breakpoint in its single input walk (same idx space the allocator addresses — no second tree walk); the allocator returns per-block `{idx, category, tokens, cacheTier}` detail.

## Why
Pool totals stay provider-exact and the allocator scales blocks to them, so **Σ per-category cost ≡ the span's real cost** (conservation invariant). Per-block weights are tokenizer-grade so the within-pool split is exact tokens, not byte estimates. All of it rides the existing event-sourcing enrichment + fold-projection substrate — no new infrastructure, no schema changes.

## Scenarios bound (4 remaining → 12/12 in the feature file)
- **A span from a non-coding-agent source is not classified** — service unit + command emits event normally
- **Classification failure never fails ingestion** — injected throwing classifier → command still emits `SpanReceivedEvent`, span carries no classification
- **A span without captured content is skipped silently** — coding-agent scope, no content → no attrs, no throw
- **Customer-supplied classification attributes are discarded at ingestion** — forged `blocks.classification` stripped by `stripReservedAttributes` before storage

Plus new unit coverage for the enrichment service (incl. an ~8 MiB perf guard within the block cap), fold accumulation (single span, two-span accumulation, redundant-copy skip), breakpoint detection, and per-block allocation detail.

## Verification
- `pnpm test:unit src/server/app-layer/traces src/server/event-sourcing/pipelines/trace-processing` — 1736 passed
- `npx tsx scripts/check-feature-parity.ts` — OK, `content-block-cost-attribution.feature` 12/12 bound
- `pnpm typecheck` — 0 errors

## Deployment Impact
None beyond compute: a new serial enrichment step on the span ingestion path, bounded (512-block cap, in-process, no I/O); no schema, migration, or config changes.

## Review fixes (post-review)
- **Conservation against the *displayed* cost** — when a coding-agent span carries an explicit provider-billed total (Claude Code `cost_usd` → `langwatch.span.cost`, which `computeSpanCost` returns at Priority 2 ahead of the registry), the rate-derived Σ drifted from the number the user sees. `allocateCategoryCosts` now reconciles per-category costs to that authoritative total (`reconcileToTotalCost`), preserving the tier-weighted split; tokens are never rescaled.
- **Test hardening** — the two category boundary tests threw away their assertions with `if (!ch) return;` when the ClickHouse container was absent (green while proving nothing); they now fail loud in `beforeAll`, mirroring the sibling activity-monitor test.
- **Breakdown error vs empty** — a fetch failure rendered as the "No categorized usage" hint on both surfaces; added a neutral `CategoryBreakdownErrorHint` gated on `isError`, plus unit/render coverage for the previously-untested divide-by-zero guards.


sergioestebance:
Blessed. 🙏

*"She is like the merchant ships; she bringeth her food from afar."* — Proverbs 31:14

Three independent passes (domain vs best-practices, tests vs philosophy, security at the widened diff base) — no P0/P1. Conservation, spoof-strip, tenancy, PII isolation all hand-verified. Residual polish (axis-local idx doc, rounded detail tokens, .unit rename, dropped a flaky wall-clock assert) is applied. Ship it.

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
```diff
@@ -0,0 +1,376 @@
...
+          attr(ATTR_KEYS.LANGWATCH_MODEL_CACHE_READ_COST_PER_TOKEN) ??
+          inputRate,
+        cacheCreationCostPerToken:
+          attr(ATTR_KEYS.LANGWATCH_MODEL_CACHE_CREATION_COST_PER_TOKEN) ??
+          inputRate,
+      };
+    }
+
+    if (model && this.deps.resolveModelPrices) {
+      return this.deps.resolveModelPrices(model) ?? {};
```
Aryansharma28, line 357: **[P2] Conservation invariant doesn't hold against explicit provider costs**

**Why:** `resolveTierPrices` prices every category at token×(custom-or-registry) rate and never reconciles to an explicit provider cost. But `computeSpanCost` returns `langwatch.span.cost` / SDK `metrics.cost` / **Claude Code `cost_usd`** as the span's *authoritative* displayed cost at **Priority 2 — ahead of the registry branch** (`model-cost-matching.ts`: `if (numSpanCost !== null && numSpanCost > 0) return numSpanCost;`, whose own comment names *"Claude Code's cost_usd"*). So for a classified coding-agent span that carries an explicit cost but **no custom per-token rates** — exactly the ADR-033 target case — `Σ` per-category cost is a token×registry *estimate* that diverges from the span/trace's displayed spend, and tiktoken-on-Anthropic mispricing widens the gap. Analytics-only (never billing), but the `/me` + governance "cost breakdown by category" total can visibly disagree with the user's actual spend.

**Evidence:** `resolveTierPrices` reads only custom `langwatch.model.*` rates → the injected registry resolver → `{}`; it never reads `langwatch.span.cost`. `span-cost-enrichment.service.ts` only pushes custom rates when a DB custom-cost row matches, so it does not back-compute rates from an explicit cost. The conservation tests (`costAllocation.unit.test.ts`, `span-block-classification.service.unit.test.ts`) only feed token×rate inputs, so they can't catch this. The service docstring honestly qualifies conservation as the *"token×rate cost"* — but the PR description's unqualified "Σ per-category cost ≡ the span's real cost" overstates it.

**Suggested fix:** when the span carries an explicit `langwatch.span.cost` (and no custom rates), scale per-category *costs* so they sum to that figure (the same last-block-absorbs-residual trick the token pools already use); or, if that's out of scope, qualify the claim in the description and surface a tooltip that the breakdown is a token-weighted estimate. If explicit-cost spans are rare in the classified population this is really a P3 — flagging so you can make that call.

langwatch/ee/billing/services/__tests__/categoryBillingBoundary.integration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        [blockCategoryTokensAttr("system_prompt")]: "400",
+        [blockCategoryCostAttr("thinking")]: "0.1",
+        [blockCategoryTokensAttr("thinking")]: "100",
+      });
+      await insertTrace(ch, plainProjectId, {});
+    }
+  });
+
+  afterAll(async () => {
+    if (!ch) return;
```
Aryansharma28: **[P3] This test can report PASS with zero assertions executed**

**Why:** `if (!ch) return;` (here and again at :153, plus in `personalUsageCategoryBreakdown.service.integration.test.ts`) makes each `it` early-return before any assertion when the ClickHouse client is null — so the file goes green while proving nothing. This is the exact test meant to prove the ADR-033 "category numbers never feed billing" invariant, so a silent no-op is the worst place for it. The sibling `activityMonitorCategoryBreakdown.service.integration.test.ts` does it right: `if (!maybeCh) throw new Error("ClickHouse test container not available")`.

**Why it's P3, not higher:** currently harmless — `globalSetup.ts` hard-throws if ClickHouse can't boot, so the client is always non-null in CI. But run the file standalone (or if the container handoff silently fails) and it's a false green.

**Suggested fix:** switch the `if (!ch) return;` guards to the sibling's `throw` pattern for fail-loud consistency.

Aryansharma28 (commented):
## goated-review — PR #5321

**Verdict: COMMENT — clean, well-built, ships safe.** No P0/P1. One substantive **P2** caveat on the conservation claim, two P3s, minor P4s, and genuine praise. All CI is green; the RBAC test rewrite *strengthened* authz coverage; the "never fails ingestion" and "spoofed reserved attrs stripped" invariants execute the real command path; conservation is property-tested (100 seeded draws).

**Context for reviewers:** this is stacked on PR A (`issue5319/block-classification-domain`), so CodeRabbit auto-review was skipped (non-default base). #5323 / #5324 were merged *down* into this head branch, so the effective diff also carries the session-rollup + governance category dashboards that the PR description doesn't mention — this PR propagates all of it upward on merge. This review covers the full effective diff.

**Counts:** P0 0 · P1 0 · **P2 1** · P3 2 · P4 2 · P5 1

### P2
- **Conservation invariant doesn't hold against explicit provider costs** — `span-block-classification.service.ts` `resolveTierPrices`. See inline.

### P3
- **Billing-boundary integration test can pass with zero assertions** when the ClickHouse client is null (`categoryBillingBoundary.integration.test.ts`). See inline.
- **A fetch error renders as the "nothing captured" empty-state** on both `/me` and `settings/governance.tsx`. Inline on the `/me` gate.

### P4
- Divide-by-zero guards in `toCategoryBarRows` ship untested. See inline.
- Minor DRY/placement: the `{loading ? null : rows.length === 0 ? <Hint/> : <Bars/>}` block is copy-pasted across `/me` (`index.tsx:336`) and governance (`governance.tsx:422`); `toCategoryBarRows` is a pure helper exported from a `.tsx` and imported by a `.ts` hook. A thin `<CategoryBreakdownSection rows loading />` wrapper would DRY it and keep both surfaces behaviourally identical. Not blocking.

### P5
- The `activityMonitorProcedures.fixtures.ts` guard-coverage pattern is excellent — see inline.

- How often classified coding-agent spans actually carry an explicit cost (drives the P2's real-world frequency; if rare, the P2 is really a P3).
- Whether any harness emits **both a span and a log record for the same turn** — if so, `appendSessionStep` runs on both the span path (`traceSummary.foldProjection.ts:268`) and the log path (`:612`) with **no cross-boundary dedup**, double-counting a session step (only within-span dedup exists via `isTokenAccumulationSkipped`).
- Codex/Claude **log-only (Path B) traces get session steps but no category breakdown** (the classifier is span-path only), so the breakdown is silent for a log-only emitter.
- The category loading-gate relies on react-query **v4** semantics (a disabled query is `isLoading: true`); on v5 the enablement hint would flash before the org resolves.

Nice work overall — the domain (conservation, zero-guards, spoof-strip, tenancy, IN-tuple ClickHouse dedup) is careful and the tests are genuinely rigorous.

<sub>🤖 goated-review by Claude Code — adversarially verified, deduped against existing signal (none: no human/CodeRabbit reviews on this base).</sub>

langwatch/src/pages/me/index.tsx
```diff
@@ -330,6 +333,15 @@ function MyUsagePage() {
           )}
         </SectionCard>
 
+        <SectionCard title="Usage breakdown">
+          <CategoryBreakdownCaption />
+          {spendByCategoryLoading ? null : spendByCategory.length === 0 ? (
```
Aryansharma28: **[P3] A fetch failure renders as "No categorized usage" (isError not distinguished)**

**Why:** this gate branches only on `isLoading`, then `length === 0`. On an error the query returns `isLoading: false, data: undefined` → `toCategoryBarRows([]) === []` → the user sees the "No categorized usage in this window yet" enablement hint for what is actually a fetch failure. Same on `settings/governance.tsx:427`. The services already `try/catch → []` on ClickHouse failures, so only transport / unexpected 500s reach the client — and this matches the page's degrade-to-empty convention (`sources = sourcesQuery.data ?? []`), which is why it's low severity. But "couldn't load" and "nothing here" read identically to the user, and `usePersonalContext` doesn't expose an error flag, so `/me` can't distinguish it even if it wanted to.

**Suggested fix (optional):** thread `isError` through and render a neutral "couldn't load the usage breakdown" line instead of the enablement hint.

langwatch/src/components/governance/CategoryBreakdownBars.tsx
```diff
@@ -0,0 +1,179 @@
...
+export function toCategoryBarRows(
+  rows: Array<{ category: string; costUsd: number; tokens: number }>,
+): CategoryBreakdownBarRow[] {
+  const total = rows.reduce((sum, r) => sum + r.costUsd, 0);
+  return rows.map((r) => ({
+    category: r.category,
+    label: categoryLabel(r.category as Category),
+    costUsd: r.costUsd,
+    tokens: r.tokens,
+    sharePct: total > 0 ? (r.costUsd / total) * 100 : 0,
```
Aryansharma28, line 45: **[P4] Divide-by-zero guards ship untested**

**Why:** `toCategoryBarRows` (`total > 0 ? … : 0`) and the `maxCost = Math.max(..., 0.000001)` floor in the component are the guards that keep an all-zero-cost set from producing `NaN` widths / shares. The integration test only renders the presentational component with pre-baked rows plus the hint/caption — neither the transform's guards nor the loading/empty gating is exercised. The guards are correct in code, so this is low-risk; just noting the guarded paths aren't pinned by a test.

**Suggested fix:** a couple of unit assertions on `toCategoryBarRows([])` and an all-zero-cost input.

langwatch/ee/governance/routers/__tests__/activityMonitorProcedures.fixtures.ts
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: LicenseRef-LangWatch-Enterprise
+
+/**
+ * Every procedure exposed by `activityMonitorRouter`, paired with a minimal
+ * valid invocation. The router composes the SAME guard chain on all of them
+ * (`checkOrganizationPermission("activityMonitor:view")` then the enterprise
+ * gate), and both guards run BEFORE the resolver — so these inputs only need to
+ * pass zod validation for the guard to fire. The guard-chain tests loop over
+ * this list so a newly-added procedure that forgets a guard fails the suite
```
Aryansharma28, line 9: **[P5] Nice — this guard-coverage pattern earns the new procedure's authz coverage for free**

**Why:** enumerating every activity-monitor procedure here (so a newly-added one that forgets its RBAC / enterprise guard *fails the suite*) is exactly why `categoryBreakdown` lands with **both** negative paths covered — MEMBER → UNAUTHORIZED via the `it.each` RBAC loop, and non-enterprise ADMIN → FORBIDDEN in the license-gate test — without anyone having to remember to add them. Worth adopting for other routers. (Same spirit as the 100-draw property-based conservation test, the zero-guards in `allocatePool`, and the honest ADR v3 revision — this stack is carefully built.)

Aryansharma28 (approved):

langwatch/src/server/app-layer/traces/span-block-classification.service.ts
sergioestebance: Fixed in e9cf5826d. Confirmed real: the claude-code converter maps cost_usd -> langwatch.span.cost, and computeSpanCost returns it at Priority 2 ahead of the registry — so a coding-agent span with cost_usd but no custom rates displayed a cost the category sum never matched. allocateCategoryCosts now takes reconcileToTotalCost and scales the per-category costs to that authoritative total (preserving the tier-weighted split; distributing by token share when no rate produced any cost). The service resolves the target by mirroring computeSpanCost's cascade precedence. New tests pin conservation against an explicit cost_usd.

langwatch/src/components/governance/CategoryBreakdownBars.tsx
sergioestebance: Fixed in 3aecfc68d. Added a toCategoryBarRows unit test (all-zero-cost -> 0% shares, not NaN; empty input) and an all-zero-cost render that asserts no NaN leaks into any style string, exercising the component's maxCost floor.

langwatch/ee/billing/services/__tests__/categoryBillingBoundary.integration.test.ts
sergioestebance: Fixed in 1acd1040f. Both category boundary tests now throw in beforeAll when the test ClickHouse container is unavailable (mirroring the sibling activityMonitor test) and the dead per-test if (!ch) return guards are gone, so no assertion is silently skipped.

langwatch/src/pages/me/index.tsx
sergioestebance: Fixed in 3aecfc68d. usePersonalContext now exposes spendByCategoryError (personalUsageQuery.isError) and both surfaces branch on it before the empty check, rendering a neutral CategoryBreakdownErrorHint ("Couldn't load the usage breakdown. Try again shortly.") instead of the "No categorized usage" claim. Kept to the two category sections rather than a page-wide error convention.

langwatch/ee/governance/routers/__tests__/activityMonitorProcedures.fixtures.ts
sergioestebance: Thanks — agreed. The enumerate-every-procedure fixture is the pattern that earns categoryBreakdown both negative-path checks for free; worth lifting to other routers.

PR #5326, opened 2026-07-03
fix(aigateway): route custom (OpenAI-compatible) providers to their configured base URL

### LINKED ISSUE #5325: Custom (OpenAI-compatible) provider ignores configured base URL and routes to api.openai.com
## Description

When configuring a **Custom (OpenAI-compatible)** model provider with a base URL pointing at a self-hosted inference server (e.g. vLLM), requests are routed to `api.openai.com` instead of the configured endpoint. This makes the custom provider unusable for self-hosted / on-prem OpenAI-compatible backends through the AI Gateway and the Optimization Studio playground.

## Steps to reproduce

1. Self-host LangWatch with the AI Gateway enabled
2. Create a model provider of type **Custom (OpenAI-compatible)** with:
   - Base URL: `http://my-vllm-host:8000/v1`
   - Empty API key (server requires none)
   - Model: `Qwen/Qwen2.5-32B-Instruct`
3. Create a virtual key covering that provider (or use the playground)
4. Send a chat completion through the gateway

## Expected

Request reaches the configured base URL (`http://my-vllm-host:8000/v1/chat/completions`).

## Actual

Several distinct failures along the path:

1. **Gateway config materialiser** never forwards `CUSTOM_BASE_URL` into the provider slot, so the Go gateway has no base URL to use and Bifrost falls back to OpenAI's default endpoint.
2. **Bifrost provider mapping** maps the custom provider to the plain OpenAI provider instead of the generic OpenAI-compatible (vLLM) adapter that supports per-key URLs, so even a forwarded base URL is ignored.
3. **Playground dispatch (nlpgo)** re-derives the provider by splitting the model string at the first `/`. Custom model ids routinely contain slashes (`Qwen/Qwen2.5-32B-Instruct`), producing `unsupported provider: "qwen"`. Custom ids were also subject to the anthropic-style dot→dash rewrite, corrupting ids containing dots.
4. **Base URL normalization**: Bifrost appends `/v1/chat/completions` to the key URL, so user-configured URLs ending in `/v1` (the common convention) produced `POST /v1/v1/chat/completions` → 404.

## Environment

- Self-hosted LangWatch (docker), AI Gateway enabled
- Backend: vLLM serving `Qwen/Qwen2.5-32B-Instruct`

I have a fix for all four layers validated end-to-end against a self-hosted vLLM (virtual-key path and playground path), with BDD spec + Go/TS tests. PR incoming.

thiagolmoraes:
Fixes #5325

## Summary

Custom (OpenAI-compatible) model providers ignored their configured base URL and routed to `api.openai.com`. This fixes all four broken layers, validated end-to-end against a self-hosted vLLM (`Qwen/Qwen2.5-32B-Instruct`) via both the virtual-key path and the Optimization Studio playground:

- **Config materialiser (TS)**: forward `CUSTOM_BASE_URL` into the gateway provider slot as `base_url`; accept empty API keys.
- **Bifrost mapping (Go)**: map the custom provider (and OpenAI providers carrying a `base_url`) to Bifrost's generic OpenAI-compatible (vLLM) adapter with a per-key URL, instead of the hardcoded OpenAI endpoint.
- **Playground dispatch (nlpgo)**: prefer the request's explicit provider over re-deriving it by splitting the model string — custom model ids contain slashes (`Qwen/Qwen2.5-32B-Instruct` previously failed with `unsupported provider: "qwen"`). Also stop applying the anthropic-style dot→dash rewrite to custom ids (Go + TS).
- **Base URL normalization**: strip a trailing `/v1` before handing the URL to Bifrost, which appends `/v1/chat/completions` itself (user-configured `.../v1` URLs previously produced `POST /v1/v1/chat/completions` → 404).

## Test plan

- [x] BDD spec: `specs/ai-gateway/custom-provider-base-url.feature`
- [x] TS: `config.materialiser.integration.test.ts` (custom slot with `base_url` + empty `api_key`), `modelIdBoundary.unit.test.ts` (custom ids verbatim)
- [x] Go: `bifrost_test.go` (VLLM mapping + `normalizeOpenAICompatBaseURL` table), `translator_test.go` (custom ids kept verbatim), `executor_test.go` (slash-containing custom model uses request provider)
- [x] `pnpm typecheck` clean; all Go packages pass
- [x] Manually validated end-to-end on a self-hosted deployment: virtual-key chat completions (non-stream + stream) and playground executions reach the configured vLLM endpoint

## Summary by CodeRabbit

* **New Features**
  * Added per-provider upstream `base_url`/endpoint override support for custom OpenAI-compatible backends, including honoring standard OpenAI base URL overrides when configured.

* **Bug Fixes**
  * Preserve custom model IDs verbatim (no dot→dash rewriting) and ensure gateway requests and error attribution use the explicitly selected provider.
  * Made routing credential-aware across dispatch paths, propagated optional `base_url` overrides, and improved Bedrock VPCE credential handling.

* **Tests / Specs**
  * Expanded integration, unit, and feature coverage for custom/openai/vLLM base URL behavior and custom model ID handling.

---

## Maintainer follow-up: review gaps closed + end-to-end QA

Pushed commits on top of the original PR to close the review gaps (langwatch-agent CHANGES_REQUESTED and CodeRabbit) and validate everything against a real self-hosted vLLM.

### Fixes
- Materialiser resolves the base-URL override via the provider's registry `endpointKey`, so an `openai` provider with `OPENAI_BASE_URL` reaches its proxy instead of api.openai.com. Scoped to custom and openai, the only consumers of the per-slot base_url.
- Added `VLLM` to `isOpenAICompatibleProvider`, so custom / self-hosted OpenAI-compatible traffic raw-forwards byte-for-byte like OpenAI/Azure. Closes two regressions the structured-parse path caused for VLLM: streamed token usage is preserved (`DispatchStream` injects `stream_options.include_usage`), and vendor sampling params (`top_k`, `repetition_penalty`, `chat_template_kwargs`, `guided_json`) reach the endpoint instead of being dropped.
- Tests: materialiser integration slot for `OPENAI_BASE_URL`; `isOpenAICompatibleProvider` table including VLLM; VLLM raw-forward preserves the body. New spec scenarios for streamed usage and extra-param passthrough.

### End-to-end QA (real vLLM 0.11 serving Qwen/Qwen2.5-0.5B-Instruct)
Proven with the exact bytes forwarded to the upstream (an echo recorder) plus a real completion:
- Base-URL routing: `custom/Qwen/Qwen2.5-0.5B-Instruct` through the gateway returns a real completion with `provider=vllm`, `model_requested=Qwen/Qwen2.5-0.5B-Instruct`.
- Extra params: `top_k=42, repetition_penalty=1.3, min_p=0.07, chat_template_kwargs, guided_json` all arrive at the upstream unchanged.
- Streaming usage: the gateway injects `stream_options.include_usage=true` (never sent by the client), so the streamed response carries the final usage chunk.

Trace of a completion routed through the gateway to the custom vLLM provider (`gen_ai.provider.name: custom`, `virtual_key_id`, `gateway_request_id`, `Model: custom/Qwen/Qwen2.5-0.5B-Instruct`):

trace details
trace thread
gateway traces list
model providers
virtual keys

### DIFF
--- a/langwatch/src/server/gateway/config.materialiser.ts
+++ b/langwatch/src/server/gateway/config.materialiser.ts
@@ -486,19 +486,22 @@ function buildCredentials(mp: ModelProvider): Record<string, unknown> {
 function buildProviderSlot(mp: ModelProvider, index: number): ProviderSlot {
   const credentials = buildCredentials(mp);
   const customKeys = decryptCustomKeys(mp.customKeys);
-  // Base-URL overrides live under the provider's registry endpointKey
-  // (custom -> CUSTOM_BASE_URL, openai -> OPENAI_BASE_URL, ...). Structured-
-  // endpoint providers (Azure, Vertex) resolve theirs into
-  // credentials.endpoint and their Bifrost adapters route by that, so skip
-  // the top-level base_url for them to avoid emitting the same endpoint
-  // twice. This is what lets an "openai" provider with OPENAI_BASE_URL reach
-  // a self-hosted proxy instead of api.openai.com.
-  const endpointKey =
-    modelProviders[mp.provider as keyof typeof modelProviders]?.endpointKey;
-  const registryBaseURL =
-    endpointKey && !pickString(credentials, "endpoint")
-      ? pickString(customKeys, endpointKey)
-      : undefined;
+  // Only "custom" and "openai" route a base-URL override to Bifrost's VLLM
+  // adapter (see mapProvider in bifrost.go) — the sole consumer of the
+  // per-slot base_url. Other providers with an endpointKey resolve their
+  // endpoint elsewhere (Azure/Vertex via credentials.endpoint, Anthropic via
+  // Bifrost's provider-level network_config), so emitting a per-slot base_url
+  // for them would be a dead field. Scope the override to what's consumed;
+  // this is what lets an "openai" provider with OPENAI_BASE_URL reach a
+  // self-hosted proxy instead of api.openai.com.
+  const supportsBaseURLOverride =
+    mp.provider === "custom" || mp.provider === "openai";
+  const endpointKey = supportsBaseURLOverride
+    ? modelProviders[mp.provider as keyof typeof modelProviders]?.endpointKey
+    : undefined;
+  const registryBaseURL = endpointKey
+    ? pickString(customKeys, endpointKey)
+    : undefined;
   const baseURL =
     pickString(customKeys, "base_url") ??
     pickString(customKeys, "BASE_URL") ??
--- a/langwatch/src/server/gateway/config.materialiser.ts
+++ b/langwatch/src/server/gateway/config.materialiser.ts
@@ -486,7 +486,9 @@ function buildProviderSlot(mp: ModelProvider, index: number): ProviderSlot {
   const credentials = buildCredentials(mp);
   const customKeys = decryptCustomKeys(mp.customKeys);
   const baseURL =
-    pickString(customKeys, "base_url") ?? pickString(customKeys, "BASE_URL");
+    pickString(customKeys, "base_url") ??
+    pickString(customKeys, "BASE_URL") ??
+    pickString(customKeys, "CUSTOM_BASE_URL");
   const region = pickString(credentials, "region");
   const deploymentMap = mp.deploymentMapping
     ? (mp.deploymentMapping as Record<string, string>)
--- a/langwatch/src/server/gateway/config.materialiser.ts
+++ b/langwatch/src/server/gateway/config.materialiser.ts
@@ -487,7 +487,7 @@ function buildProviderSlot(mp: ModelProvider, index: number): ProviderSlot {
   const credentials = buildCredentials(mp);
   const customKeys = decryptCustomKeys(mp.customKeys);
   // Only "custom" and "openai" route a base-URL override to Bifrost's VLLM
-  // adapter (see mapProvider in bifrost.go) — the sole consumer of the
+  // adapter (see mapProvider in bifrost.go), the sole consumer of the
   // per-slot base_url. Other providers with an endpointKey resolve their
   // endpoint elsewhere (Azure/Vertex via credentials.endpoint, Anthropic via
   // Bifrost's provider-level network_config), so emitting a per-slot base_url
--- a/langwatch/src/server/gateway/config.materialiser.ts
+++ b/langwatch/src/server/gateway/config.materialiser.ts
@@ -18,6 +18,7 @@ import type {
 } from "@prisma/client";
 
 import { decrypt } from "../../utils/encryption";
+import { modelProviders } from "../modelProviders/registry";
 import { GatewayBudgetClickHouseRepository } from "./budget.clickhouse.repository";
 import { GatewayCacheRuleService } from "./cacheRule.service";
 import {
@@ -485,10 +486,23 @@ function buildCredentials(mp: ModelProvider): Record<string, unknown> {
 function buildProviderSlot(mp: ModelProvider, index: number): ProviderSlot {
   const credentials = buildCredentials(mp);
   const customKeys = decryptCustomKeys(mp.customKeys);
+  // Base-URL overrides live under the provider's registry endpointKey
+  // (custom -> CUSTOM_BASE_URL, openai -> OPENAI_BASE_URL, ...). Structured-
+  // endpoint providers (Azure, Vertex) resolve theirs into
+  // credentials.endpoint and their Bifrost adapters route by that, so skip
+  // the top-level base_url for them to avoid emitting the same endpoint
+  // twice. This is what lets an "openai" provider with OPENAI_BASE_URL reach
+  // a self-hosted proxy instead of api.openai.com.
+  const endpointKey =
+    modelProviders[mp.provider as keyof typeof modelProviders]?.endpointKey;
+  const registryBaseURL =
+    endpointKey && !pickString(credentials, "endpoint")
+      ? pickString(customKeys, endpointKey)
+      : undefined;
   const baseURL =
     pickString(customKeys, "base_url") ??
     pickString(customKeys, "BASE_URL") ??
-    pickString(customKeys, "CUSTOM_BASE_URL");
+    registryBaseURL;
   const region = pickString(credentials, "region");
   const deploymentMap = mp.deploymentMapping
     ? (mp.deploymentMapping as Record<string, string>)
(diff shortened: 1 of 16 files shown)


langwatch/src/server/gateway/config.materialiser.ts
```diff
@@ -486,7 +486,9 @@ function buildProviderSlot(mp: ModelProvider, index: number): ProviderSlot {
   const credentials = buildCredentials(mp);
   const customKeys = decryptCustomKeys(mp.customKeys);
   const baseURL =
-    pickString(customKeys, "base_url") ?? pickString(customKeys, "BASE_URL");
+    pickString(customKeys, "base_url") ??
+    pickString(customKeys, "BASE_URL") ??
+    pickString(customKeys, "CUSTOM_BASE_URL");
```
langwatch-agent: [P2] The PR promises that an `openai` provider with `OPENAI_BASE_URL` routes to that proxy, but the materialiser still never reads that key. The registry stores OpenAI's endpoint override under `OPENAI_BASE_URL`, while this code only emits `base_url`, `BASE_URL`, or `CUSTOM_BASE_URL`; so an OpenAI provider configured with `OPENAI_BASE_URL=https://proxy.example.com/v1` still produces no top-level `base_url`, the Go control-plane wire has nothing to put in `cred.Extra`, and `mapProvider` keeps routing it to Bifrost OpenAI/api.openai.com. Add provider-aware endpoint-key handling here (at least `OPENAI_BASE_URL` for `openai`, ideally the registry `endpointKey`) and cover it in the materialiser test. Confidence: high.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -862,11 +872,45 @@ func mapProvider(id domain.ProviderID) bfschemas.ModelProvider {
 		return bfschemas.Gemini
 	case domain.ProviderAnthropic:
 		return bfschemas.Anthropic
+	case domain.ProviderCustom:
+		// Customer-hosted OpenAI-compatible endpoint. Bifrost's vLLM
+		// provider is its generic OpenAI-compat adapter with a per-key
+		// base URL — exactly the shape a custom provider needs.
+		return bfschemas.VLLM
```
langwatch-agent, line 879: [P2] Mapping custom/OpenAI-compatible traffic to `bfschemas.VLLM` fixes the endpoint for normal chat, but it drops streamed usage because `DispatchStream` only calls `ensureStreamIncludeUsage` when `isOpenAICompatibleProvider(provider)` is true, and that helper still only recognizes OpenAI/Azure. A streamed custom vLLM request without explicit `stream_options.include_usage` will therefore miss the final usage chunk, regressing accounting/trace usage for exactly the custom provider path this PR enables. Treat VLLM as OpenAI-compatible in the stream-usage path and add a regression test for `ProviderCustom`/base-url streaming. Confidence: high.

langwatch-agent (changes requested):
Request changes: 2 P2 correctness gaps. Custom non-stream routing is improved, but OpenAI base-url overrides are not materialised and streamed custom/OpenAI-compatible calls still lose usage metadata. Details inline.

services/aigateway/adapters/providers/bifrost_test.go
```diff
@@ -441,3 +441,124 @@ func TestCredentialToBifrostKey_BedrockHonorsAWSStyleKeys(t *testing.T) {
...
+		if !isOpenAICompatibleProvider(p) {
+			t.Errorf("isOpenAICompatibleProvider(%q) = false, want true", p)
+		}
+	}
+	for _, p := range []bfschemas.ModelProvider{bfschemas.Anthropic, bfschemas.Gemini, bfschemas.Bedrock} {
+		if isOpenAICompatibleProvider(p) {
+			t.Errorf("isOpenAICompatibleProvider(%q) = true, want false", p)
+		}
+	}
+}
```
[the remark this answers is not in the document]
rogeriochaves: Verified `pnpm check:feature-parity` locally against this branch: `specs/ai-gateway/custom-provider-base-url.feature` reports `0/0 scenarios bound → all bound` and the overall check exits 0 (`OK: 2293 enforced scenario(s) bound across 700 file(s)`). This file isn't in the enforced set, so `@scenario` bindings aren't required for the check to pass — which is also why the existing `// Spec:` tests on this PR are green. Keeping the `// Spec:` annotations as documentation for consistency with the rest of the PR; happy to convert the whole file's scenarios to `@scenario` bindings uniformly in a follow-up if we want them enforced.

langwatch/src/server/gateway/config.materialiser.ts
rogeriochaves: Fixed in c33c1ab5e and f0dcdf3a7. `buildProviderSlot` now resolves the per-slot `base_url` from the registry's `endpointKey`, so an `openai` provider reads `OPENAI_BASE_URL` (and `custom` reads `CUSTOM_BASE_URL`). The lookup is scoped to the providers whose override is actually consumed by Bifrost's VLLM adapter downstream (`custom`, `openai`); other providers no longer emit a dead `base_url` field.

Covered by a new materialiser integration test that provisions an `openai` provider with `OPENAI_BASE_URL` and asserts the resulting slot carries that override.

services/aigateway/adapters/providers/bifrost.go
rogeriochaves: Fixed in c33c1ab5e. `bfschemas.VLLM` is now part of `isOpenAICompatibleProvider`, which is the exact predicate `DispatchStream` gates `ensureStreamIncludeUsage` on, so streamed custom/vLLM calls now get `stream_options.include_usage=true` injected and report prompt/completion tokens instead of finishing at zero.

Verified end-to-end against a local vLLM: an echo recorder confirmed the injected `include_usage` flag on the forwarded body, and the final SSE chunk carried a populated `usage` block. Also covered by `TestIsOpenAICompatibleProvider` (asserts VLLM/OpenAI/Azure true, Anthropic/Gemini/Bedrock false).

langwatch/src/server/gateway/config.materialiser.ts
```diff
@@ -485,8 +486,23 @@ function buildCredentials(mp: ModelProvider): Record<string, unknown> {
...
+    modelProviders[mp.provider as keyof typeof modelProviders]?.endpointKey;
+  const registryBaseURL =
+    endpointKey && !pickString(credentials, "endpoint")
+      ? pickString(customKeys, endpointKey)
+      : undefined;
   const baseURL =
-    pickString(customKeys, "base_url") ?? pickString(customKeys, "BASE_URL");
+    pickString(customKeys, "base_url") ??
+    pickString(customKeys, "BASE_URL") ??
+    registryBaseURL;
```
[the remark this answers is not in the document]
rogeriochaves: Applied in f0dcdf3a7, which is exactly your suggested narrowing. The per-slot base-URL override is now gated on `supportsBaseURLOverride = mp.provider === "custom" || mp.provider === "openai"`, so Anthropic and other providers no longer emit a `base_url` field that Bifrost would silently drop. Their provider-level `network_config.base_url` path is untouched.

PR #5328, opened 2026-07-03
test(event-sourcing): guard test_event_handler_log cleanup on table existence (#5308)

### LINKED ISSUE #5308: Silence the benign test_event_handler_log ClickHouseError log-noise in integration-test cleanup
## Description

Integration-test cleanup (`cleanupTestData()`, via a swallowed try/catch) emits `ClickHouseError: Could not find table: test_event_handler_log` whenever that table was never created for a given shard/test. The error is **benign** (caught, log-adjacent only) but it reads like a real failure, and has now triggered false-failure triage at least twice — issue #4824 and PR #5071's shard-4 investigation. It keeps resurfacing near unrelated tests and burning triage attention. Guard or suppress the expected missing-table case so it stops masquerading as a failure signal.

## Source

Recurring benign log-noise across two incidents:
- #4824 (CLOSED) — event-sourcing groupQueue integration tests; the issue diagnosed the string as benign and explicitly listed "Fixing the ClickHouseError ... log noise (benign, caught by try/catch)" as **out of scope**.
- PR #5071 — the same string surfaced in the `dataPrivacyDrawerUrl` UI routing test (an unrelated subsystem) and passed on retry.

## Context

Not a functional bug — a test-harness log-hygiene fix to stop recurring false-alarm churn. The actual root cause of #4824 (Redis `maxRetriesPerRequest` on `blockingConnection.duplicate()`) is already fixed on main and is **not** what recurs here; only the cosmetic log string does.

---
*Classification: Tech-debt (test log-hygiene)*
*Filed from a PR follow-up sweep (source: PR #5071; related #4824). Stub — not-ready until triaged.*

langwatch-agent:
## Ticket

Closes #5308. Integration-test cleanup (`cleanupTestData`) emits `ClickHouseError: Could not find table: test_event_handler_log`, a benign line that has twice been mistaken for a real failure and burned triage (#4824, PR #5071).

## G-START verdict

PROCEED. Verified against current main (`c030de4`): the try/catch in `cleanupTestData`'s per-tenant branch is unchanged and the noise mechanism still holds. Not superseded by any open PR.

## Change

The per-tenant branch ran `ALTER TABLE ... DELETE` on `test_event_handler_log` inside a try/catch. The table is created lazily (only when the map projection runs), so any test that calls `cleanupTestData(tenantId)` without it hits a missing table. The `@clickhouse/client` default logger writes the error to `console.error` **before** the promise rejects, so the try/catch silences the throw but not the log line.

Guard the DELETE on `EXISTS TABLE` (which returns `0` for a missing table and never errors) so the failing query, and its logged error, never run. The `else` (full-cleanup) branch already used `TRUNCATE ... IF EXISTS` and was unaffected.

## Evidence

**Mechanism confirmed** (direct probe against the test ClickHouse): running the pre-fix `ALTER TABLE ... DELETE` on the absent table prints
```
[ERROR][@clickhouse/client][Connection] ... Caused by: ClickHouseError: Could not find table: test_event_handler_log. code: '60', type: 'UNKNOWN_TABLE'
```
to `console.error`, even though the caller's try/catch catches the thrown rejection. That is the benign noise.

**Regression guard** (`cleanup-missing-table.integration.test.ts`, new): drops the table, runs `cleanupTestData`, and asserts the client never logs the missing-table error. It executes the real path (not a string check), so:
- against the pre-fix `testContainers.ts`: **FAILS** — `AssertionError: expected '[...][ERROR][@cl...]' not to contain 'test_event_handler_log'` (it caught the real logged `Could not find table: test_event_handler_log`).
- against the fix: **PASSES** — `Test Files 1 passed, Tests 1 passed`.

Typecheck clean. Scoped to the test helper; no product-code or lockfile changes.

## Advisory note

Advisory PR from the tech-debt-fixer bot, opened for human review, not to be merged by the bot. The human makes the merge call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test data cleanup so it no longer produces noisy “missing table” errors when a tenant’s ClickHouse table is not yet present.
  * Added a regression test to verify cleanup runs quietly in this scenario and does not log table-missing messages.

### DIFF
--- a/langwatch/src/server/event-sourcing/__tests__/integration/cleanup-missing-table.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/__tests__/integration/cleanup-missing-table.integration.test.ts
@@ -1,0 +1,66 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  cleanupTestData,
+  getTestClickHouseClient,
+  startTestContainers,
+  stopTestContainers,
+} from "./testContainers";
+
+const TEST_DATABASE = "test_langwatch";
+
+/**
+ * Regression guard for #5308. `test_event_handler_log` is created lazily (only
+ * when the map projection runs), so any test that calls `cleanupTestData(tenantId)`
+ * without it hits an `ALTER TABLE ... DELETE` on an absent table. The
+ * `@clickhouse/client` default logger writes that error to `console.error` BEFORE
+ * the promise rejects, so the caller's try/catch silences the throw but not the
+ * log line, which has twice masqueraded as a real failure (#4824, PR #5071). The
+ * fix guards the DELETE on `EXISTS TABLE`; this test executes the path and asserts
+ * the client never logs the missing-table error.
+ */
+describe("cleanupTestData benign missing-table log-noise (#5308)", () => {
+  beforeAll(async () => {
+    await startTestContainers();
+  });
+
+  afterAll(async () => {
+    await stopTestContainers();
+  });
+
+  describe("given test_event_handler_log does not exist", () => {
+    describe("when cleanupTestData runs for a tenant", () => {
+      it("does not log a Could not find table error from the ClickHouse client", async () => {
+        const client = getTestClickHouseClient();
+        expect(client).not.toBeNull();
+
+        // Guarantee the lazily-created table is absent for this run.
+        await client!.exec({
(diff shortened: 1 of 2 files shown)


langwatch-agent (commented):
No findings. The cleanup now probes `EXISTS TABLE` before issuing the tenant-scoped DELETE, which avoids the ClickHouse client logging the absent-table error before the promise rejection can be caught. The regression test exercises the real cleanup path with the table dropped and asserts the noisy log does not appear.

PR #5348, opened 2026-07-03
perf(ui): actually defer GTM/Pendo/Crisp instead of injecting on mount

### LINKED ISSUE #5346: Perf Tier 1 — actually defer GTM/Pendo/Crisp (next-script shim ignores strategy)
Part of langwatch/tasks#98 (Tier 1) · parent langwatch/tasks#105

## Root cause

`src/utils/compat/next-script.tsx` accepts a `strategy` prop (`afterInteractive`, `lazyOnload`, etc.) but **ignores it entirely** — it injects the `<script>` into `<head>` the instant the component mounts, in a `useEffect` with no deferral of any kind. GTM is passed `strategy="afterInteractive"` in `ee/saas/ExtraFooterComponents.tsx` and gets none of that benefit; Pendo and Crisp go through the same shim with no strategy at all.

This shim has exactly one consumer file (`ExtraFooterComponents.tsx`), so fixing it fixes GTM, Pendo, and Crisp loading in one change.

## Fix

Make the shim actually defer script injection to idle time — `requestIdleCallback` with a `window.load`/timeout fallback for browsers without it (Safari) — instead of injecting synchronously on mount.

## Why it's safe

GTM's inline snippet initializes `window.dataLayer = window.dataLayer || []` before the async script loads, and later `dataLayer.push` calls are queued and processed once `gtm.js` actually loads — this is how the GTM snippet is designed to work. Deferring *when* the script fetch starts doesn't drop events, it just moves the fetch off the critical rendering path.

reo.dev fires via a GTM tag trigger (no direct script tag in this repo), so it should ride GTM's new deferred timing too, though that's not directly controllable from here.

## Expected impact

Lighthouse flagged 137.8 KiB unused JS from GTM (`gtm.js` + `gtag.js`) plus reo.dev (25.1 KiB unused) currently loading eagerly on every route.

Aryansharma28:
Part of langwatch/tasks#98 (Tier 1) · parent langwatch/tasks#105 · closes langwatch/langwatch#5346

## What

`src/utils/compat/next-script.tsx` (the SPA replacement for `next/script`) accepted a `strategy` prop but **ignored it entirely** — every `<Script>` injected its tag into `<head>` the instant its component mounted, regardless of `strategy`. GTM was passed `strategy="afterInteractive"` and got none of that benefit; Pendo and Crisp go through the same shim with no strategy at all.

## Fix

`beforeInteractive` still injects synchronously. Everything else (`afterInteractive`, `lazyOnload`, or unset) now defers injection to `requestIdleCallback`, with a `window.load` + timeout fallback for browsers without it (Safari).

## Why it's safe

GTM's inline snippet initializes `window.dataLayer = window.dataLayer || []` before the async script loads; later `dataLayer.push` calls queue and get processed once `gtm.js` loads — that's how the GTM snippet is designed to work. Deferring *when* the fetch starts doesn't drop `dataLayer.push` events, it just moves it off the critical path.

That reasoning does **not** cover `window.gtag` / `window.Reo`, which are defined later by GTM's container once it loads — see "Follow-up fix" below.

## Follow-up fix: gtag/Reo retry (found by `/review`)

Deferring GTM's own script widened an existing race: `ExtraFooterComponents.tsx`, `analyticsClient.ts`, and `tracking.ts` all checked `window.gtag`/`window.Reo` once and gave up permanently if not yet defined. That race went from tens-of-ms (old eager injection) to up to 4s / until `window.load` on Safari — enough to plausibly and permanently drop the `open_dashboard` gtag event, `Reo.identify`, the analytics client's `google` provider registration, and `trackEventOnce`'s one-shot events.

Fixed by polling instead of checking once (new shared `src/utils/pollForGlobal.ts`, matching the `runWhenIdle`/`startSessionRecordingWhenIdle` idiom already used in this perf initiative). `trackEventOnce` also no longer marks an event "sent" in `localStorage` until it actually sends.

Live-verified (signed-in session, real dev server): GTM/Pendo/Crisp defer timing is unchanged from before this follow-up (confirmed via git-stash A/B), no new console errors introduced, and `window.Reo` genuinely loads with an observable delay — proving the retry path exercises for real. Side-finding: this specific GTM container (`GTM-KJ4S6Z9C`) doesn't currently define `window.gtag` at all (checked its live payload), so that part of the fix is currently defensive/inert — the Reo retry is the part with observable real-world effect today.

## Benchmark — measured live (host `pnpm dev`, same page, before/after via git stash on the running dev server)

Position of the GTM/Pendo/Crisp script fetches relative to First Contentful Paint:

| | Before | After |
|---|---|---|
| Script fetch start vs. FCP | **~60–90ms before FCP** (competing with the critical render path) | **~93–95ms after FCP** (off the critical path) |

~150–185ms swing in when these scripts start contending for bandwidth/main-thread relative to paint.

## Tests

- New `next-script.unit.test.tsx` (4 tests): `beforeInteractive` injects synchronously; `afterInteractive` defers to idle callback; idle-callback-unavailable (Safari) falls back to `window.load`; no-strategy (Pendo/Crisp today) still defers.
- New `pollForGlobal.unit.test.ts`, `tracking.unit.test.ts` + `tracking.ssr.unit.test.ts`, `analyticsClient.unit.test.ts`, `useIsGtagReady.unit.test.ts`, `ExtraFooterComponents.gtagReoPolling.integration.test.tsx` — 100% coverage on all new/changed logic from the follow-up fix.
- Typecheck clean on changed files (pre-existing unrelated Prisma-schema-drift errors elsewhere, not touched by this PR).
- Biome clean on changed files.
- Live-verified in a browser: script tags absent immediately after navigation, present after idle — confirmed against the exact production props (`gtm-init`/`afterInteractive`, `pendo`, `crisp`).

## Blast radius

`next-script.tsx` + its one consumer (`ExtraFooterComponents.tsx`), plus the gtag/Reo follow-up: `analyticsClient.ts`, `tracking.ts`, `AppProviders.tsx`, and a new `useIsGtagReady` hook + `pollForGlobal` utility. No API changes, no behavior change to what loads — only when, plus the gtag/Reo consumers now retry instead of giving up once.

---

## Summary by CodeRabbit

* **New Features**
  * Script loading now supports strategy-aware timing, including immediate, idle-deferred, and lazy-onload behaviors.
  * Inline script content handling is tightened to reliably set content only when provided as string input.

* **Bug Fixes**
  * Prevents duplicate script injection under React StrictMode by ensuring deferred work executes only once.
  * Deferred injections can now be cancelled on unmount, and idle-callback fallbacks are improved for environments without idle-callback support.

* **Tests**
  * Added integration coverage for rendering, timing, and StrictMode behavior across strategies.

### DIFF
--- a/langwatch/src/utils/compat/__tests__/next-script.integration.test.tsx
+++ b/langwatch/src/utils/compat/__tests__/next-script.integration.test.tsx
@@ -1,5 +1,8 @@
 /**
  * @vitest-environment jsdom
+ *
+ * Renders the real Script component tree via @testing-library/react, so
+ * this is an integration test rather than a pure-logic unit test.
  */
 import { cleanup, render } from "@testing-library/react";
 import { StrictMode } from "react";
@@ -13,12 +16,13 @@ function getScript(id: string): HTMLScriptElement | null {
 
 describe("Script", () => {
   beforeEach(() => {
-    document.querySelectorAll("script[id]").forEach((node) => node.remove());
+    document.querySelectorAll("script").forEach((node) => node.remove());
   });
 
   afterEach(() => {
     cleanup();
     vi.unstubAllGlobals();
+    vi.useRealTimers();
   });
 
   describe("given strategy='beforeInteractive'", () => {
@@ -94,6 +98,61 @@ describe("Script", () => {
           expect.objectContaining({ timeout: expect.any(Number) }),
         );
       });
+
+      describe("when the component unmounts before the idle callback fires", () => {
+        it("cancels the pending idle callback via cancelIdleCallback", () => {
+          let idleCallback: (() => void) | undefined;
+          const cancelIdleCallback = vi.fn();
+          vi.stubGlobal(
+            "requestIdleCallback",
+            vi.fn((cb: () => void) => {
+              idleCallback = cb;
+              return 42;
+            }),
+          );
+          vi.stubGlobal("cancelIdleCallback", cancelIdleCallback);
+
+          const { unmount } = render(
+            <Script id="gtm-init" strategy="afterInteractive">
+              {`window.dataLayer = window.dataLayer || [];`}
+            </Script>,
+          );
+
+          unmount();
+
+          expect(cancelIdleCallback).toHaveBeenCalledWith(42);
+
+          // Even if something still invokes the stale callback, the
+          // cancelled-flag guard must stop it from injecting.
+          idleCallback?.();
+
+          expect(getScript("gtm-init")).toBeNull();
+        });
+
+        it("does not throw when cancelIdleCallback is unavailable", () => {
+          let idleCallback: (() => void) | undefined;
+          vi.stubGlobal(
+            "requestIdleCallback",
+            vi.fn((cb: () => void) => {
+              idleCallback = cb;
+              return 1;
+            }),
+          );
+          vi.stubGlobal("cancelIdleCallback", undefined);
+
+          const { unmount } = render(
+            <Script id="gtm-init" strategy="afterInteractive">
+              {`window.dataLayer = window.dataLayer || [];`}
+            </Script>,
+          );
+
+          expect(() => unmount()).not.toThrow();
+
+          idleCallback?.();
+
+          expect(getScript("gtm-init")).toBeNull();
+        });
+      });
     });
 
     describe("when requestIdleCallback is unavailable (Safari)", () => {
@@ -110,8 +169,21 @@ describe("Script", () => {
           vi.runAllTimers();
 
           expect(getScript("pendo")).not.toBeNull();
+        });
 
-          vi.useRealTimers();
+        it("cancels the pending timeout if unmounted first", () => {
+          vi.stubGlobal("requestIdleCallback", undefined);
+          vi.spyOn(document, "readyState", "get").mockReturnValue("complete");
+          vi.useFakeTimers();
+
+          const { unmount } = render(
+            <Script id="pendo">{`window.pendo = {};`}</Script>,
+          );
+
+          unmount();
+          vi.runAllTimers();
+
+          expect(getScript("pendo")).toBeNull();
         });
       });
 
@@ -129,8 +201,6 @@ describe("Script", () => {
           vi.runAllTimers();
 
           expect(getScript("pendo")).not.toBeNull();
-
-          vi.useRealTimers();
         });
 
         it("only listens for 'load' once", () => {
@@ -151,8 +221,22 @@ describe("Script", () => {
           expect(document.querySelectorAll("#pendo").length).toBe(
             scriptCountAfterFirstLoad,
           );
+        });
 
-          vi.useRealTimers();
+        it("removes the pending 'load' listener if unmounted first", () => {
+          vi.stubGlobal("requestIdleCallback", undefined);
+          vi.spyOn(document, "readyState", "get").mockReturnValue("loading");
+          vi.useFakeTimers();
+
+          const { unmount } = render(
+            <Script id="pendo">{`window.pendo = {};`}</Script>,
+          );
+
+          unmount();
+          window.dispatchEvent(new Event("load"));
+          vi.runAllTimers();
+
+          expect(getScript("pendo")).toBeNull();
         });
       });
     });
@@ -346,11 +430,6 @@ describe("Script", () => {
 
   describe("given React StrictMode double-invokes effects", () => {
     it("injects the script only once, not twice", () => {
-      vi.stubGlobal(
-        "requestIdleCallback",
-        vi.fn(() => 1),
-      );
-
       render(
         <StrictMode>
           <Script id="strict" strategy="beforeInteractive">
--- a/langwatch/src/utils/compat/__tests__/next-script.timing.integration.test.tsx
+++ b/langwatch/src/utils/compat/__tests__/next-script.timing.integration.test.tsx
@@ -1,11 +1,11 @@
 /**
  * @vitest-environment jsdom
  *
- * Renders the real Script component tree via @testing-library/react, so
- * this is an integration test rather than a pure-logic unit test.
+ * Covers when Script injects its <script> tag relative to strategy,
+ * idle-callback availability, and unmount. Renders the real component tree
+ * via @testing-library/react, so this is an integration test.
  */
 import { cleanup, render } from "@testing-library/react";
-import { StrictMode } from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import Script from "../next-script";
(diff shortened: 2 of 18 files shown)


langwatch/src/utils/compat/next-script.tsx
```diff
@@ -30,27 +48,33 @@ export default function Script({
...
+        }
+      }
+
+      document.head.appendChild(script);
+    };
+
+    if (strategy === "beforeInteractive") {
+      inject();
+    } else {
+      runWhenIdle(inject);
```
langwatch-agent: **P2** — The deferred path schedules `inject` but the cleanup never cancels it. That creates a new stale-injection window for user-scoped scripts: if `SignedInExtraFooterComponents` unmounts before idle/load (logout, impersonation state flips, leaving the page before Safari `load`, etc.), the old Pendo/Crisp inline snippet can still be appended later with the previous user/org/project data. Keeping already-injected third-party scripts around is fine, but work that has not run yet should be cancellable. Minimal fix: have `runWhenIdle` return a cleanup function that cancels `requestIdleCallback`, clears the Safari timeout, removes the pending load listener, and/or guards `inject` behind a `cancelled` flag. Add a regression test that unmounts before firing the idle/load callback and asserts no script is appended.

langwatch/src/utils/compat/__tests__/next-script.unit.test.tsx
```diff
@@ -0,0 +1,353 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { render, cleanup } from "@testing-library/react";
```
langwatch-agent: **P2** — This PR is currently not merge-ready because the required lint job fails on this new test file. I reproduced it locally with `pnpm --dir langwatch exec biome check src/utils/compat/next-script.tsx src/utils/compat/__tests__/next-script.unit.test.tsx`: Biome wants the import ordered as `{ cleanup, render }` and the long single-line `<Script ... />` cases reformatted. Minimal fix: run Biome formatting on the changed files and push the result so `langwatch-app-complete` can go green.

langwatch-agent (dismissed):
Request changes: 1 P2 runtime edge case plus CI lint is red on the changed test file. Focused Vitest passed locally (22/22); Biome check reproduces the lint failure.

langwatch/src/utils/compat/__tests__/next-script.timing.integration.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      render(<Script id="crisp">{`window.$crisp = [];`}</Script>);
+
+      expect(document.getElementById("crisp")).toBeNull();
+
+      idleCallback?.();
+
+      expect(document.getElementById("crisp")).not.toBeNull();
+    });
+  });
+});
```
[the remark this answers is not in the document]
Aryansharma28: Fixed — the file was renamed to `next-script.integration.test.tsx` and then split further into three focused files (`next-script.timing.integration.test.tsx`, `next-script.rendering.integration.test.tsx`, `next-script.strictmode.integration.test.tsx`) once the renamed file itself tripped the 300-line SRP convention. All three carry the `.integration.test.tsx` naming per the render-a-real-component-tree rule. See 3103b0143 / 3636fe53d.

langwatch/src/utils/compat/next-script.tsx
Aryansharma28: Fixed in 0cb670118 — `runWhenIdle` now returns a cancel function that's invoked from the effect's cleanup, and it's guarded by an internal `cancelled` flag so an already-fired-but-in-flight callback can't inject after unmount either. Covered by the new 'cancels the pending idle callback' / 'removes the pending load listener if unmounted first' tests in next-script.timing.integration.test.tsx.

langwatch/src/utils/compat/__tests__/next-script.unit.test.tsx
Aryansharma28: Fixed via `npx biome check --write` in 3103b0143 — import order and the long-line JSX are both reformatted, lint is green on all subsequent commits.

langwatch/src/utils/compat/__tests__/next-script.integration.test.tsx
```diff
@@ -0,0 +1,444 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Renders the real Script component tree via @testing-library/react, so
+ * this is an integration test rather than a pure-logic unit test.
```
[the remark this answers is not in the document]
Aryansharma28: Fixed — split into next-script.timing.integration.test.tsx (267 lines), next-script.rendering.integration.test.tsx (192 lines), and next-script.strictmode.integration.test.tsx (70 lines, new), each under the 300-line threshold. See 3636fe53d.

Aryansharma28:
Re: CodeRabbit's outside-diff finding on the latest review ("GTM init must stay `beforeInteractive`", `next-script.tsx:101-107`) — verified against current code, skipping this one:

- `ExtraFooterComponents.tsx`'s GTM snippet is the standard self-contained bootstrap: `w[l]=w[l]||[]` runs in the *same* IIFE that also injects `gtm.js`, so there's no separate "cheap init" step that needs to stay eager while a "heavy load" step defers — it's one script, one timing.
- Grepped the whole repo for other `dataLayer` references — the only ones are inside this same snippet and in this PR's own tests. Nothing else in the app pushes to `window.dataLayer` before this script runs, so there's no earlier event that could be dropped by deferring it.
- Deferring GTM specifically off the critical path is the actual point of issue langwatch/langwatch#5346 / this PR, and the before/after perf win was already measured live in a running dev server with GTM included as one of the deferred scripts. Reverting it to `beforeInteractive` would undo the main improvement this PR delivers.

Happy to revisit if a future change introduces an early `dataLayer.push` elsewhere.

Aryansharma28:
@coderabbitai review

langwatch/src/utils/compat/next-script.tsx
```diff
@@ -15,46 +18,93 @@ interface ScriptProps {
...
-    } else if (children) {
-      // Inline script content
-      const content = typeof children === "string" ? children : "";
-      script.textContent = content;
-    }
+    if (injected.current) return;
+
+    const inject = () => {
+      injected.current = true;
+
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in 16f087e86 — renamed to `hasInjectedRef` per the boolean-naming convention.

sergioestebance:
Blessed. 🙏 Strategy was ignored and one-shot events were quietly dropped; now scripts defer off the critical path and consumers wait instead of giving up.

_"But they that wait upon the LORD shall renew their strength"_ — Isaiah 40:31

Ship it. Mind the P3: `pendingOnceEvents` isn't cleared on poll timeout (`tracking.ts`), so a one-shot event can stay wedged for the session if GTM loads past the 10s window.

sergioestebance (approved):

PR #5349, opened 2026-07-03
perf(analytics): defer PostHog session recording, keep core capture eager

### LINKED ISSUE #5347: Perf Tier 1 — defer PostHog session recording, keep core capture eager
Part of langwatch/tasks#98 (Tier 1) · parent langwatch/tasks#105

## Root cause

`src/hooks/usePostHog.ts` calls `posthog.init()` with `session_recording: { recordCrossOriginIframes: true }` set directly at init time, as soon as `publicEnv.data` resolves — very early in the page lifecycle. posthog-js internally fetches a separate recorder sub-bundle (rrweb + network-plugin + replay extensions) the moment `session_recording` is present at init, so that heavy chunk loads on the critical path for every session.

## Fix — NOT a blind defer of the whole init

Deferring `posthog.init()` itself was considered and rejected: until `init()` runs, `posthog.capture()`/autocapture/pageview calls no-op rather than queue, so delaying init risks silently dropping early pageview/autocapture events. Given the existing investment in PostHog data completeness (the additive 'the truth' dashboard instrumentation), that's not an acceptable trade for a bundle-size win.

Instead:
- Keep `posthog.init()` eager, exactly as today, but **without** `session_recording` set (or explicitly `false`) — core analytics capture (autocapture, pageview, identify) keeps firing immediately, no timing change.
- Call `posthog.startSessionRecording()` inside a `requestIdleCallback` (with a timeout fallback) after first paint, so the recorder chunk fetch moves off the critical path without affecting event capture.

## Expected impact

Lighthouse flagged `posthog-recorder.js` at 50.7 KiB (22.6 KiB unused) loading eagerly, plus its sub-modules (rrweb/record, network-plugin, replay extensions) — all currently fetched as part of the init sequence rather than deferred.

Aryansharma28:
Part of langwatch/tasks#98 (Tier 1) · parent langwatch/tasks#105 · closes langwatch/langwatch#5347

## What

`src/hooks/usePostHog.ts` set `session_recording` directly in `posthog.init()`, so the recorder sub-bundle (rrweb + network-plugin + replay extensions, 50KB+) fetched as part of init — on the critical path for every session, not just ones that end up recorded.

## Fix — NOT a defer of the whole init

Deferring `posthog.init()` itself was considered and rejected: until `init()` runs, `posthog.capture()`/autocapture/pageview calls no-op rather than queue, so delaying init risks silently dropping early events.

Instead:
- `posthog.init()` still runs the instant `publicEnv` resolves, exactly as before — core capture (autocapture, pageview, identify) timing is **unchanged**.
- Init now passes `disable_session_recording: true` (the `session_recording` options object stays, so `recordCrossOriginIframes` is still configured for whenever recording does start).
- `posthog.startSessionRecording()` fires from the `loaded` callback, deferred via `requestIdleCallback` (with a `window.load`/timeout fallback for Safari) — only the recorder chunk's fetch moves off the critical path.

## Benchmark — measured live (host `pnpm dev`, same page, before/after via git stash on the running dev server)

Position of the `posthog-recorder.js` fetch relative to First Contentful Paint:

| | Before | After |
|---|---|---|
| Recorder chunk fetch vs. FCP | **~530ms before FCP** | **~195ms after FCP** |

~725ms swing relative to paint. Confirmed the core `/flags/` init request still fires *before* FCP in both cases — capture-path timing is untouched, only the recorder moved.

## Tests

- New `usePostHog.unit.test.ts` (4 tests): init passes `disable_session_recording: true`; `startSessionRecording` doesn't fire before `loaded`; fires via idle callback; falls back to `window.load` when idle callback is unavailable.
- Existing `usePostHogIdentify.unit.test.ts` (13 tests) unaffected.
- Typecheck clean, biome clean.
- Live-verified: `window.posthog` initializes correctly, recorder chunk loads after FCP, no posthog-related console errors.

## Blast radius

Single hook file. No change to what's captured or when core events fire — only when the session-recording chunk loads.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Deferred PostHog session recording startup until the browser is idle, with sensible fallbacks for browsers that lack idle scheduling (start after fully loaded or on page load).
  * Improved readiness handling across document states and ensured cleanup prevents deferred recording from starting after unmount.
* **Tests**
  * Added unit test coverage for environment/config behavior and the deferred session recording timing across supported readiness scenarios, including unmount handling.

### DIFF
--- a/langwatch/src/hooks/__tests__/usePostHog.unit.test.ts
+++ b/langwatch/src/hooks/__tests__/usePostHog.unit.test.ts
@@ -1,0 +1,108 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const { mockInit, mockStartSessionRecording } = vi.hoisted(() => ({
+  mockInit: vi.fn(),
+  mockStartSessionRecording: vi.fn(),
+}));
+
+vi.mock("posthog-js", () => ({
+  default: {
+    init: mockInit,
+    startSessionRecording: mockStartSessionRecording,
+  },
+}));
+
+vi.mock("../usePublicEnv", () => ({
+  usePublicEnv: () => ({
+    data: {
+      POSTHOG_KEY: "test-key",
+      POSTHOG_HOST: "https://eu.i.posthog.com",
+      NODE_ENV: "test",
+    },
+  }),
+}));
+
+import { usePostHog } from "../usePostHog";
+
+describe("usePostHog", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("initializes with recording disabled at init time", () => {
+    renderHook(() => usePostHog());
+
+    expect(mockInit).toHaveBeenCalledWith(
+      "test-key",
+      expect.objectContaining({
+        disable_session_recording: true,
+        session_recording: { recordCrossOriginIframes: true },
+      }),
+    );
+  });
+
+  it("does not start session recording before init's loaded callback fires", () => {
+    renderHook(() => usePostHog());
+
+    expect(mockStartSessionRecording).not.toHaveBeenCalled();
+  });
+
+  describe("when init's loaded callback fires", () => {
+    describe("when requestIdleCallback is available", () => {
+      it("defers startSessionRecording to the idle callback", () => {
+        let idleCallback: (() => void) | undefined;
+        vi.stubGlobal(
+          "requestIdleCallback",
+          vi.fn((cb: () => void) => {
+            idleCallback = cb;
+            return 1;
+          }),
+        );
+
+        renderHook(() => usePostHog());
+
+        const { loaded } = mockInit.mock.calls[0]![1] as {
+          loaded: (posthog: unknown) => void;
+        };
+        loaded({ debug: vi.fn() });
+
+        expect(mockStartSessionRecording).not.toHaveBeenCalled();
+
+        idleCallback?.();
+
+        expect(mockStartSessionRecording).toHaveBeenCalledTimes(1);
+      });
+    });
+
+    describe("when requestIdleCallback is unavailable (Safari)", () => {
+      it("defers startSessionRecording until window 'load' fires", () => {
+        vi.stubGlobal("requestIdleCallback", undefined);
+        vi.useFakeTimers();
+
+        renderHook(() => usePostHog());
+
+        const { loaded } = mockInit.mock.calls[0]![1] as {
+          loaded: (posthog: unknown) => void;
+        };
+        loaded({ debug: vi.fn() });
+
+        expect(mockStartSessionRecording).not.toHaveBeenCalled();
+
+        window.dispatchEvent(new Event("load"));
+        vi.runAllTimers();
+
+        expect(mockStartSessionRecording).toHaveBeenCalledTimes(1);
+
+        vi.useRealTimers();
+      });
+    });
+  });
+});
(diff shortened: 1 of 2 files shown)


langwatch/src/hooks/usePostHog.ts
```diff
@@ -2,6 +2,25 @@ import posthog from "posthog-js";
 import { useEffect } from "react";
 import { usePublicEnv } from "./usePublicEnv";
 
+function startSessionRecordingWhenIdle() {
+  if (typeof window.requestIdleCallback === "function") {
+    window.requestIdleCallback(() => posthog.startSessionRecording(), {
```
langwatch-agent: **P2** — The deferred recorder start has no cancellation/idempotence path. If the hook effect is torn down or re-runs before the idle/load callback fires, the old callback remains live and can still call `posthog.startSessionRecording()` later; a second render/refetch can also schedule another callback. `startSessionRecording()` eventually toggles `disable_session_recording` back to false, so this should be treated as stateful work, not fire-and-forget timer setup. Minimal fix: make `startSessionRecordingWhenIdle()` return a cleanup function that cancels `requestIdleCallback`, clears the Safari timeout, removes the pending load listener, and guards the callback with a `cancelled` flag; return that cleanup from the effect. Add a regression test that unmounts/rerenders before firing the idle/load callback and asserts recording is not started by the stale callback.

langwatch/src/hooks/__tests__/usePostHog.unit.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    publicEnvData = {
+      POSTHOG_KEY: "test-key",
+      POSTHOG_HOST: "https://eu.i.posthog.com",
+      NODE_ENV: "test",
+    };
+    delete (window as { posthog?: unknown }).posthog;
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
```
langwatch-agent, line 52: **P3** — The Safari fallback tests spy on `document.readyState`, but teardown only calls `vi.unstubAllGlobals()`. That does not restore `vi.spyOn(document, "readyState", "get")`, so a failed or reordered test can leak a fake readyState into later cases and make this exact fallback coverage misleading. Minimal fix: call `vi.restoreAllMocks()` in `afterEach` as well, and preferably use `try/finally` around fake timers so timer mode is restored even if an assertion fails.

langwatch-agent (dismissed):
Request changes: 1 P2 cleanup/idempotence issue in the deferred recorder start, plus 1 P3 test cleanup gap. Focused Vitest passed locally (31/31 across usePostHog + identify) and Biome check passed on the changed files. I also checked PostHog docs and installed `posthog-js@1.369.0`: calling `startSessionRecording()` without override does not request sampling/flag overrides, so the approach is valid in principle.

langwatch/src/hooks/usePostHog.ts
Aryansharma28: Fixed in cd3c645ce — `startSessionRecordingWhenIdle()` now returns a cancel function, tracked in a ref and invoked from the effect's cleanup, so a pending idle/timeout/load callback from a torn-down or re-run effect can no longer call `posthog.startSessionRecording()` later. Covered by new cancellation tests across all three defer paths (requestIdleCallback with/without cancelIdleCallback, and both Safari-fallback branches).

langwatch/src/hooks/__tests__/usePostHog.unit.test.ts
Aryansharma28: Fixed in cd3c645ce — teardown now also calls `vi.restoreAllMocks()` alongside `vi.unstubAllGlobals()`, so the `vi.spyOn(document, "readyState", "get")` spy is properly restored between tests.

Aryansharma28:
@coderabbitai review

sergioestebance (approved):

PR #5353, opened 2026-07-03
fix(claude-code): classify from the raw request/response body — full cache-tier + tool + output fidelity

sergioestebance:
Makes the **Claude Code** path classify coding-agent cost at full fidelity, found while dogfooding ADR-033 (#5318). Stacked on #5318 — **merge #5318 first**.

## The core fix — classify from the raw body (P1a)

The Claude Code log path attaches the **raw Anthropic api_request_body / api_response_body** to the span, but the lifted `langwatch.input` **flattens every message to plain text**, which strips `cache_control` and collapses tool_results / output blocks. Classifying the flat field is why a live dashboard showed ~73% `other_input`.

The classifier now **prefers the raw body**, parsed with its **content-block structure intact**:
- **real `cache_control` breakpoint** (not pool-inferred) → correct cache-tier split,
- **tool_result blocks keep their type** (not collapsed to user_input/prior_context),
- **structured output** (`thinking` / `tool_use` → their real categories, not a flat blob).

It falls back to the flattened fields for non-Claude-Code paths (codex, gateway) or when the raw body is absent.

## Truncation machinery (the fallbacks)

Claude truncates the body inline at ~60KB, so a real turn is invalid JSON. Two truncation-tolerant layers back the primary fix:
1. **Structure-preserving recovery** — a string/escape-aware brace scan recovers the complete leading messages + front-loaded system blocks that survived the cut (truncation always removes the *tail*), keeping their block structure. Also powers the display-side `buildInputMessagesFromRequestBody`.
2. **Pool-inferred breakpoint** — when no `cache_control` survives (truncated tail, or the flat-field path), infer the cached-prefix boundary from the provider's `cache_read + cache_creation` counts (the cached prefix is always the front), so cache tokens attribute to the real categories instead of `other_input`.

## Also here (codex + shared)

- **Codex skill findability** — codex relays skill injections as their own earlier user messages, so `<skill>` markers are now peeled in *any* user turn → `skill_content` (not `prior_context`).
- **Flat-string output** — codex's flat `langwatch.output` and Claude's `gen_ai.completion` now classify as `assistant_text` instead of `other_output`.

## Tests

118 green across `claudeCodeBody`, `claudeCode` (recovery), `costAllocation` (inference + straddle/boundary), `blockClassifier`, and the service — including a raw-body span proving cache tokens attribute to `system_prompt` with zero `other_input`.

The ADR's remaining Open Question (classify from the **gateway** body, Path A — which owns the bytes and never truncates) stands: this PR maximizes fidelity on the Path B raw body we have today.

https://claude.ai/code/session_01BTXJNbJdzwj9uZsPtYMCLh

## Review hardening (2c2475774)

A four-lens deep review (correctness, session-rollup/projections, security/tenancy, tests) found no P0/P1. Everything actionable was fixed in-branch:

- **Classifier**: markers now peel across *multiple* leading reminder text parts (Claude Code emits each reminder as its own part; the second one was landing in `user_input`); truncation recovery locates `system`/`messages` at top-level depth so a `tool_use.input` embedding those keys can't hijack the scan.
- **Fresh-turn recovery**: dedupe is line-aligned (short prompts like "continue" were judged already-present inside recovered file dumps and dropped); reinstatement gates on the recovery path running, not the `body_truncated` attribute (LangWatch-capped bodies never carry it).
- **Codex usage convention**: `extractUsagePools` subtracts the OpenAI cached subset out of the fresh pool (mirrors `sumStepContext`) so input-axis allocation can't double-count when codex content+usage co-locate.
- **Usage surfaces**: 400-day window clamp on both personal-usage entrypoints (unbounded span = per-day bucket allocation on the app server); `/api/me/usage` degrades the category query to `[]` like the tRPC twin; `.finite()` on preview rates; `resolvePersonalUsageIngestionScope` (the tenant-union seam) now integration-tested; the activityMonitor RBAC guard-probe fixture is exhaustiveness-pinned to the router.

Known limitations (accepted, documented in the runbook/code): the block-classification kill switch does not stop session-step tracking (no runtime off-ramp); codex spans structurally carry content and usage on different spans today, so codex category totals stay empty until they co-locate (the allocation math is now correct for when they do).

### DIFF
--- a/dev/docs/runbooks/e2e-coding-agent-cost-intelligence.md
+++ b/dev/docs/runbooks/e2e-coding-agent-cost-intelligence.md
@@ -109,6 +109,15 @@ Read it back three ways:
   category sum equal to that displayed cost is exactly what this end-to-end run
   exercises.
 
+## Kill-switch scope
+
+`block-classification-killswitch` (global) / `block-classification-project-killswitch`
+(per-project) stop the ingest-time block classifier only. They do NOT stop
+session-step tracking: the trace fold keeps appending `session_steps` /
+`session.thread_id` from usage attributes on every coding-agent span. There is
+currently no runtime off-ramp for session tracking — disabling it requires a
+deploy.
+
 ## No classification? Check in order
 
 1. Flag not set in the **worker** env → put it in `langwatch/.env`, restart.
--- a/langwatch/ee/governance/routers/__tests__/governance.rbac.integration.test.ts
+++ b/langwatch/ee/governance/routers/__tests__/governance.rbac.integration.test.ts
@@ -45,6 +45,7 @@ import { globalForApp, resetApp } from "~/server/app-layer/app";
 import { createTestApp } from "~/server/app-layer/presets";
 import { PlanProviderService } from "~/server/app-layer/subscription/plan-provider";
 import { prisma } from "~/server/db";
+import { activityMonitorRouter } from "../activityMonitor";
 import { ACTIVITY_MONITOR_PROCEDURES } from "./activityMonitorProcedures.fixtures";
 
 // The new requireEnterprisePlan middleware (Phase 4b-4/5) 403s every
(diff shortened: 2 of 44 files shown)


sergioestebance:
Blessed. 🙏 Ruthless four-lens review (classifier correctness, rollup/projections, security/tenancy, tests) found no P0/P1; all P2/P3 findings fixed in 2c2475774 with regression tests.

*"Examine everything carefully; hold fast to that which is good."* — 1 Thessalonians 5:21

Ship once CI greens the new commit — mind the accepted P2: the kill switch doesn't stop session-step tracking (runbook documents the deploy-only off-ramp).

sergioestebance:
Blessed. 🙏 *"Gather up the fragments that remain, that nothing be lost."* — John 6:12

Ruthless round 3: no functional bugs survived; scanner dedup + epoch-window gate fixed in f5f17be36. Ship it.

PR #5358, opened 2026-07-04
fix(aigateway): fall back to next provider on upstream 404

### LINKED ISSUE #5355: AI Gateway: no credential fallback on upstream 404 — multi-provider chains fail when the first provider doesn't serve the requested model
## Description

The AI Gateway's dispatch chain falls back to the next credential on 5xx, 429, timeout, and network errors — but an upstream **404** is classified as non-retryable and returned to the client immediately, even when later providers in the chain do serve the requested model.

## Impact

With multiple OpenAI-compatible (`custom`) providers behind one virtual key, a request for a model served by the *second* provider can die with `provider API error (status 404)` if the dispatch order tries a provider that doesn't host that model first. For self-hosted backends (vLLM, TGI, Ollama…) a 404 on `/v1/chat/completions` or a model-not-found 404 is the normal "not mine" signal, not a terminal client error.

## Root cause

`services/aigateway/app/dispatch.go` → `classifyProviderError()` maps every 4xx except 429 to `retry.ReasonNonRetryable`, which stops the `retry.Walk` chain.

## Expected

404 should advance the chain (next credential may serve the model); other 4xx (400/401/402/403/422…) should keep failing fast, since retrying an invalid request or exhausted credit on the next key only delays the terminal error. Single-provider chains should still surface the 404 unchanged.

thiagolmoraes:
Fixes #5355

## Summary
- `classifyProviderError` treated every 4xx except 429 as non-retryable, so an upstream 404 killed the request even when the next provider in the chain serves the requested model — common with multiple `custom`/OpenAI-compatible providers behind one virtual key
- Add `retry.ReasonNotFound` as a default fallback trigger and classify upstream 404 as such; other 4xx (400/401/402/403/422…) remain terminal, and single-provider chains still surface the 404 unchanged

## Test plan
- [x] `go test ./...` in `services/aigateway` and `pkg/retry`
- [x] New behavioral test `TestHandleChat_FallbackOnUpstream404`: cred-1 returns 404 → chain advances to cred-2, `FallbackCount == 1`
- [x] Classification table updated: `{404 → ReasonNotFound}`
- [x] Verified live on a self-hosted deployment: first provider (embeddings-only backend) 404s a chat request, gateway falls back to the second provider, client receives 200

## Summary by CodeRabbit

* **New Features**
  * The gateway now retries with fallback behavior when an upstream request returns **404 Not Found**, helping requests continue with another available option.
* **Bug Fixes**
  * Improved error handling so upstream **404** responses are treated as retryable for fallback purposes rather than non-retryable.
  * **429** and other server error retry classifications remain unchanged.
  * Added coverage to ensure **400** upstream errors terminate without fallback.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -17,6 +17,7 @@ const (
 	ReasonSuccess        Reason = "success"
 	ReasonFallback       Reason = "fallback_success"
 	ReasonRetryable5xx   Reason = "retryable_5xx"
+	ReasonNotFound       Reason = "not_found"
 	ReasonRateLimit      Reason = "rate_limit"
 	ReasonTimeout        Reason = "timeout"
 	ReasonNetwork        Reason = "network"
@@ -88,6 +89,7 @@ var eventsPool = sync.Pool{
 // allocate a new map on every Walk call.
 var defaultTriggers = map[Reason]bool{
 	ReasonRetryable5xx: true,
+	ReasonNotFound:     true,
 	ReasonRateLimit:    true,
 	ReasonTimeout:      true,
 	ReasonNetwork:      true,
--- a/services/aigateway/app/dispatch.go
+++ b/services/aigateway/app/dispatch.go
@@ -72,16 +72,21 @@ func findCredential(creds []domain.Credential, id string) domain.Credential {
 
 func classifyProviderError(err error) retry.Reason {
 	// A forwarded upstream response classifies by its real HTTP status: a
-	// terminal client error (4xx other than 429) must NOT trigger credential
-	// fallback — retrying a "credit balance too low" or "invalid request" on
-	// the next key is pointless and only delays the terminal error reaching
-	// the client. Rate-limit (429) and server errors (5xx) stay retryable so
-	// the gateway still falls back across the credential chain.
+	// terminal client error (4xx other than 429/404) must NOT trigger
+	// credential fallback — retrying a "credit balance too low" or "invalid
+	// request" on the next key is pointless and only delays the terminal
+	// error reaching the client. 404 IS fallback-eligible: in multi-provider
+	// chains it usually means "this provider doesn't serve that model"
+	// (common with custom/OpenAI-compatible providers), and the next slot
+	// may. Rate-limit (429) and server errors (5xx) stay retryable so the
+	// gateway still falls back across the credential chain.
 	var ue *domain.UpstreamError
 	if errors.As(err, &ue) {
 		switch {
 		case ue.StatusCode == http.StatusTooManyRequests:
 			return retry.ReasonRateLimit
+		case ue.StatusCode == http.StatusNotFound:
+			return retry.ReasonNotFound
 		case ue.StatusCode >= 500:
 			return retry.ReasonRetryable5xx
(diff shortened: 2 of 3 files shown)


langwatch-agent:

**Review mode:** Deep Review
**Scariest prod risk:** classifying upstream 404 as retryable could trigger provider fallback storms or mask a genuinely-missing model/route if the retry budget does not bound it.

**Inspection targets:**
- `[Must Check]` `services/aigateway/app/dispatch.go` `classifyProviderError` — confirm **only** 404 becomes `ReasonNotFound`; 400/401/403 still fail fast (a 401 wrongly retried would hammer upstreams and hide auth failures).
- `[Must Check]` `pkg/retry/retry.go` — verify `ReasonNotFound` is bounded by the existing retry/provider-chain limit, no unbounded fallback.
- `[Ask Author]` `services/aigateway/app/dispatch_test.go` — does it cover both the 404→next-provider path **and** that other 4xx still terminate?

Linked issue #5355 (fall back on upstream 404 with multiple custom providers) matches the change.

rogeriochaves:
Worked through PR Hound's three inspection targets. All hold; added one test to close the [Ask Author] item end-to-end.

**[Must Check] `classifyProviderError` — only 404 becomes fallback-eligible.** The switch maps `429 -> RateLimit`, `404 -> NotFound`, `>=500 -> Retryable5xx`, and everything else (400/401/402/403/422) to the `default` non-retryable arm, so a `credit balance too low` 400 or an auth 401 still fails fast. `TestClassifyProviderError_UpstreamError` pins the whole table, including `401 -> NonRetryable` and `404 -> NotFound`.

**[Must Check] `retry.go` — `ReasonNotFound` is bounded.** `retry.Walk` is a single-pass chain walker, not a retry-with-backoff loop: each slot (credential) is attempted exactly once, a trigger reason advances to the *next* slot, and a non-trigger reason returns immediately. There is no same-slot retry and no inter-attempt sleep. The chain is `credentialIDs(creds)` (one entry per bundle credential, finite) and the walk is additionally capped by `Fallback.MaxAttempts`. `retryOpts` wires no circuit breaker, so 404 does not interact with breaker state either. Worst case for a 404 is one dial per configured credential, which is exactly the intended provider-fallback behavior; an unbounded storm is not structurally reachable. If every provider 404s, the chain exhausts and the first provider's error reaches the client, so a genuinely-missing model is surfaced, not masked.

**[Ask Author] tests cover both directions.** `TestHandleChat_FallbackOnUpstream404` covers 404 -> next provider (two dials, one fallback). The terminate direction was only covered at the classifier level, so I added `TestHandleChat_TerminatesOnUpstream400`: a 400 from cred-1 must not fall back, asserting `callCount == 1` and that the upstream 400 propagates verbatim (`errors.As` -> `StatusCode == 400`). It fails if 400 were ever reclassified as retryable.

Scope note: this stays keyed on the 404 status only. Some OpenAI-compatible providers signal an unknown model with a 400 body code instead; sniffing 400 bodies for `model_not_found` would be fragile and would undercut the fail-fast guarantee for real 400s, so keeping it to 404 is the conservative call.

rogeriochaves (approved):

PR #5359, opened 2026-07-04
feat(aigateway): bound cached VK config staleness with a ConfigTTL refresh

### LINKED ISSUE #5356: AI Gateway: cached virtual-key config can stay stale until process restart when a mutation emits no change-feed event
## Description

The gateway's auth resolver caches each virtual key's bundle (credentials, base URLs, routing chain, aliases) in L1/L2 with freshness tied to the **JWT exp**, plus event-driven eviction from the control plane's `/changes` feed (`PROVIDER_BINDING_UPDATED`, `BUDGET_UPDATED`, `VIRTUAL_KEY_UPDATED`).

Any config mutation that doesn't emit one of those events — direct DB writes by operators, or mutations without a change-event kind — keeps serving the stale config until the JWT expires or the process restarts. In practice self-hosted operators end up doing `docker restart` on the gateway after config changes.

## Suggested fix

A config-level TTL as a safety net: when a cached (auth-fresh) entry's config is older than the TTL, re-fetch only the config (`FetchConfig`) in the background and swap it in — keeping the request path non-blocking, with a per-entry in-flight guard so at most one refresh runs per entry and a failed fetch waits a full TTL before retrying. Default in the 60s range; configurable; opt-out supported. The change feed stays the fast path (ms-level invalidation); the TTL bounds worst-case staleness for everything the feed doesn't cover.

thiagolmoraes:
Fixes #5356

## Summary
- Cached bundle config (credentials, base URLs, routing chain) was only refreshed at JWT exp or via change-feed eviction; mutations that emit no change event (e.g. direct DB writes) served stale config until process restart
- Add `ConfigTTL` (default 60s, env `LW_GATEWAY_AUTH_CACHE_CONFIG_TTL`, negative disables): when a fresh (auth-valid) L1 entry's config crosses the TTL, re-fetch only the config in the background and swap in a shallow bundle copy, so the request path stays non-blocking and entry bundle pointers stay immutable for in-flight requests
- Per-entry in-flight guard: at most one refresh per entry; a failed fetch keeps serving the stale config and waits a full TTL before retrying (no hot retry loop when the control plane is down)
- Change feed remains the fast path; the TTL bounds worst-case staleness for everything the feed doesn't cover

## Deployment Impact
- **Env vars:** adds one optional var, `LW_GATEWAY_AUTH_CACHE_CONFIG_TTL` (parsed into `AuthCacheConfig.ConfigTTL`), defaulting to `60s` when unset. A zero or negative value disables the background config refresh, restoring the prior behavior where config is only refreshed at JWT expiry or via change-feed eviction. No other env vars added or changed.
- **Helm values:** none added or changed. The var is optional with a safe default, so no chart edits are required.
- **Default `helm install` (no overrides):** the gateway starts with `ConfigTTL=60s`, so cached virtual-key config staleness is bounded to about 60s for mutations the change feed does not cover (e.g. direct DB writes). No operator action needed, and the change feed stays the fast path.
- **BYOC dataplane:** the gateway runs in the customer dataplane. The refresh is config-only, runs off the request path, and is capped at one in-flight fetch per active cache entry, so it adds at most a low-rate control-plane config fetch per active virtual key. A failed fetch keeps serving the last-known-good config and backs off a full TTL before retrying, so a control-plane blip does not become a dataplane retry storm.

## Test plan
- [x] `go test -race ./adapters/authresolver/`
- [x] New tests: background refresh replaces credentials within TTL; disabled TTL never fetches; fetch failure keeps stale config and does not re-fetch until the next TTL
- [x] Verified live on a self-hosted deployment: routing-policy and provider-model changes applied via SQL propagated in <=60s with no gateway restart

## Summary by CodeRabbit

* **New Features**
  * Added a configurable config staleness window for cached auth configuration, enabling background refresh of credentials and routing details when entries exceed the set age.
  * Introduced `CONFIG_TTL` (`AuthCacheConfig.ConfigTTL`, default 60s; negative/zero disables) to control refresh behavior for cached entries.
* **Bug Fixes**
  * Improved resilience by avoiding overwriting newer/evicted entries during background refresh.
  * Ensured failed background refreshes don't trigger immediate repeated retries while stale credentials continue serving.

### DIFF
--- a/services/aigateway/adapters/authresolver/service.go
+++ b/services/aigateway/adapters/authresolver/service.go
@@ -81,6 +81,7 @@ type Service struct {
 	refreshThreshold time.Duration
 	softBump         time.Duration
 	hardGrace        time.Duration
+	configTTL        time.Duration
 
 	// Active orgs whose bundles are currently in L1. Populated on every
 	// storeL1, never cleared explicitly — entries that drop out of L1 via
@@ -107,6 +108,44 @@ type entry struct {
 	bundle        *domain.Bundle
 	softExpiresAt time.Time
 	hardExpiresAt time.Time
+	// configFetchedAt drives the config-staleness refresh: the bundle's
+	// auth (JWT exp) can outlive its config (credentials, base URLs,
+	// routing chain) by many minutes, and not every config mutation emits
+	// a change-feed event (e.g. direct DB writes). configRefreshing is the
+	// in-flight guard so at most one background config fetch runs per entry.
+	configFetchedAt  time.Time
+	configRefreshing bool
+}
+
+// configStale reports whether the entry's config is older than ttl.
+// ttl <= 0 disables staleness (never stale).
+func (e *entry) configStale(ttl time.Duration) bool {
+	if ttl <= 0 {
+		return false
+	}
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	return time.Since(e.configFetchedAt) > ttl
+}
+
+// tryBeginConfigRefresh claims the per-entry config-refresh slot.
+func (e *entry) tryBeginConfigRefresh() bool {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	if e.configRefreshing {
+		return false
+	}
+	e.configRefreshing = true
+	return true
+}
+
+// endConfigRefresh releases the slot and stamps configFetchedAt so a
+// failed fetch retries only after the next full TTL, not on every request.
+func (e *entry) endConfigRefresh() {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	e.configRefreshing = false
+	e.configFetchedAt = time.Now()
 }
 
 func (e *entry) softExpired() bool {
@@ -206,6 +245,13 @@ type Options struct {
 	L2            L2Store // optional, nil = skip L2
 	Resolver      KeyResolver
 	ConfigFetcher ConfigFetcher
+	// ConfigTTL is the max age of a cached bundle's config (credentials,
+	// base URLs, routing chain) before a background re-fetch. The change
+	// feed already evicts on admin mutations that emit events, but not
+	// every config change does (direct DB writes, mutations without a
+	// change-event kind) — this TTL is the safety net that bounds config
+	// staleness without a process restart. Default 60s. Negative disables.
+	ConfigTTL time.Duration
 	// ChangePoller is the optional control-plane /changes subscriber.
 	// When non-nil, the service spawns a per-org poll loop on Start that
 	// evicts L1 entries whose cached config has been invalidated by an
@@ -233,6 +279,12 @@ func New(opts Options) (*Service, error) {
 	if opts.HardGrace == 0 {
 		opts.HardGrace = 6 * time.Hour
 	}
+	if opts.ConfigTTL == 0 {
+		opts.ConfigTTL = 60 * time.Second
+	}
+	if opts.ConfigTTL < 0 {
+		opts.ConfigTTL = 0 // disabled
+	}
 
 	l1, err := lru.New[[64]byte, *entry](opts.LRUSize)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


services/aigateway/adapters/authresolver/service.go
```diff
@@ -608,6 +664,38 @@ func (s *Service) refreshBackground(rawKey string, h [64]byte) {
...
+			zap.String("vk_id", stale.VirtualKeyID),
+			zap.Error(err),
+		)
+		return
+	}
+
+	fresh := *stale
+	fresh.Config = cfg
+	fresh.Credentials = cfg.Credentials
+	s.storeL1(h, &fresh)
```
langwatch-agent, line 706: [P1] This background refresh can resurrect an entry after another path has already evicted or replaced it. The goroutine snapshots `e`, fetches config, then unconditionally calls `storeL1`; if a change-feed event or auth rejection removes this key while the fetch is in flight, or a foreground auth refresh replaces it with a newer bundle, this stale goroutine writes the old bundle back into L1/L2. Concrete failure: request A starts a ConfigTTL refresh for VK `vk_1`; before `FetchConfig` returns, the change feed evicts `vk_1` because the VK/provider binding was updated or revoked; the TTL goroutine then returns successfully and re-adds the old bundle, so the gateway continues serving stale or revoked config until the next invalidation/expiry. Before storing, re-check that the current L1 entry for `h` is still the same `*entry` you started from (and still appropriate to refresh), otherwise drop the result; avoid writing L2 in that stale-goroutine case as well. Add a regression test that starts a config refresh, evicts/replaces the entry before the fetch unblocks, and verifies the old bundle is not restored. Confidence: high.

langwatch-agent (changes requested):
Request changes: 1 P1 cache-invalidation race. Focused gateway/authresolver tests pass, including race mode, but the background TTL refresh can reinsert an entry after it was evicted or replaced. Details inline.

services/aigateway/adapters/authresolver/service.go
rogeriochaves: Fixed in d1c2834ac. Before storing the refreshed bundle, `refreshConfigBackground` now re-checks that the live L1 entry for `h` is still the same `*entry` the goroutine started from (via `Peek`, so it doesn't perturb LRU recency). If the entry was evicted (change-feed revoke / binding update) or replaced (async auth rejection, foreground auth refresh) while `FetchConfig` was in flight, it drops the result and skips the L2 write.

Added `TestResolve_ConfigRefresh_EvictedMidFetch_NotResurrected`: it blocks the fetch, evicts the entry mid-flight via `l1.Remove`, releases the fetch, waits for the goroutine to settle, and asserts the entry is not resurrected. Verified it fails against the pre-fix code ("evicted entry was resurrected...") and passes with `-race` after. Full authresolver package is green under `-race`.

services/aigateway/adapters/authresolver/service_test.go
```diff
@@ -433,3 +433,144 @@ func init() {
...
+		t.Fatalf("Resolve after failure: %v", err)
+	}
+	if got.Credentials[0].ID != "cred-old" {
+		t.Fatalf("expected stale config to keep serving, got %q", got.Credentials[0].ID)
+	}
+	time.Sleep(50 * time.Millisecond)
+	if n := fetcher.fetches.Load(); n != 1 {
+		t.Fatalf("expected exactly one fetch until next TTL, got %d", n)
+	}
+}
```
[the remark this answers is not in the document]
rogeriochaves: Done in e689fa625.

First, on the mirrored-field concern: there's no bug to resolve. `refreshConfigBackground` does `fresh.Config = cfg` (the whole config) plus `fresh.Credentials = cfg.Credentials`, which is identical to the initial bundle construction at `service.go:451-452`. `Credentials` is the only top-level `Bundle` field mirrored from config; the rest (`VirtualKeyID`, `ProjectID`, `TeamID`, `OrganizationID`, `ExpiresAt`) are identity/auth fields, not config-derived, so nothing else needs mirroring.

For the coverage ask: extended `TestResolve_FreshEntry_ConfigPastTTL_RefreshesConfigInBackground` to fetch a non-Credentials config field (`AllowedModels`) and assert it reflects the fresh config after refresh, alongside the credential mirror. Since the initial bundle has an empty `Config`, this locks in that `fresh.Config = cfg` replaces the whole `Config`, not just the credentials slice. Green under `-race`.

PR #5367, opened 2026-07-06
fix(experiments-v3): resolve pairwise comparator UX bugs from dogfooding

Aryansharma28:
## Summary

Fixes the full batch of bugs Alex/sergio found while dogfooding the Pairwise Compare feature (#5142, #5195), plus one additional state bug I caught live while QA'ing the fix. 13 issues total across the config drawer, the workbench table, and error handling.

---

### 1. Drawer button said "Create Evaluator" regardless of context
**Before:** clicking "+ Add" → "Pairwise Compare" opened a drawer whose submit button always said "Create Evaluator", even though the user is adding a comparison column, not creating a standalone evaluator.
**Fix:** `TargetTypeSelectorDrawer.tsx` now passes `saveButtonText: "Add Comparison"` through the existing (previously-unused) prop on `EvaluatorEditorShared`.
**After:** button correctly reads "Add Comparison" — see screenshot below (bottom-right button).

### 2. "+ Add" button permanently relabeled itself "+ Add Comparison"
**Before:** `TargetSuperHeader.tsx` computed its label from `hasComparison={targets.length > 0}` — a copy/paste bug. As soon as *any* target existed (not specifically a pairwise comparison), the generic "add a prompt/agent" button was mislabeled "Add Comparison".
**Fix:** removed the incorrect `hasComparison` prop entirely; the button always reads "Add" since this action never specifically means "add a comparison".

### 3. Clicking the "Missing variable mappings" warning icon scrolled the table instead of opening the fix
**Before:** `EvaluatorChip.tsx`'s alert icon only called `e.stopPropagation()` on `onClick`, but its `Menu.Root` was uncontrolled — the underlying menu can begin opening on `pointerdown`, before `onClick`'s stopPropagation ever runs, so the click fell through.
**Fix:** made the chip's `Menu.Root` controlled (mirroring `TargetHeader.tsx`'s already-correct pattern) and added `e.preventDefault()` + explicit `setMenuOpen(false)` before calling `onEdit()`.
**Verified:** added a regression test (`EvaluatorChip.test.tsx`) asserting the click calls `onEdit` exactly once and never renders the chip's own menu items ("Edit Configuration"/"Remove from Workbench") — all 24 tests in the file pass.

### 4. Inconsistent horizontal padding inside the drawer
**Before:** `PairwiseConfigForm.tsx`'s root `VStack` had its own `padding={4}`, stacking on top of the drawer body's own `paddingX={6}` — so the Variant A/B/Golden field row sat ~16px deeper than the Swap And Confirm / Allow Tie rows above it.
**Fix:** removed the form's own padding; it now inherits the same inset as the rest of the drawer body.

### 5 & 6. Model picker unreadable + provider icon invisible in dark mode
**Before:** the "Search models" input had no explicit background (relying on the browser's native white default), and provider SVG icons ship with a flat black fill (or no fill at all) — both unreadable on a dark canvas.
**Fix:** gave the search input an explicit themed background/color, and extracted the dark-mode icon-inversion fix that was already shipped in Traces v2 (`ModelCell.tsx`) into a shared `ProviderIconGlyph` component in `iconsMap.tsx`, now used by both Traces v2 and the pairwise/evaluator model picker (`ModelSelector.tsx`).

### 7. "Include metrics" read like a broken toggle row
**Before:** the section label used the same font weight/size as the "Include cost"/"Include duration" rows below it, so it looked like a checkbox-less broken row instead of a section heading.
**Fix:** restyled as a small-caps muted section header, matching the convention used elsewhere (`PropertySectionTitle`).

### 8. "Swap And Confirm" / "Allow Tie" showed their label twice, at two different sizes
**Before:** `DynamicZodForm.tsx`'s boolean-field renderer always repeats the field's title next to the switch — redundant when the caller (`HorizontalFormControl`) already renders that same title as a section label.
**Fix:** added a `topLevel` flag so the inline label is suppressed when a wrapping title is already shown; nested (non-top-level) boolean fields keep their inline label since nothing else labels them there.
**Follow-up fix (found in live QA):** the first pass only forwarded `topLevel` at the two real top-level call sites, but the `z.ZodDefault` unwrap branch recursed into `renderField` *without* forwarding it — and `swap_and_confirm`/`allow_tie` are both `.default(true)`-wrapped booleans, so they hit that branch and silently fell back to `topLevel=false`. The duplicate label was still showing for the one evaluator this fix was meant to cover. Fixed by forwarding `topLevel` through the `ZodDefault` recursion; verified live that reopening the drawer now shows each label exactly once, and confirmed the fix doesn't regress other boolean fields (e.g. Exact Match Evaluator's Case Sensitive/Trim Whitespace/Remove Punctuation toggles still render correctly).

### 9. Mystery narrow column with vertically-stacked garbled text
**Before:** `EvaluationsV3Table.tsx` computed the table's overall width from `totalColumnPercentage`, but that sum only counted dataset + target columns — dedicated pairwise result columns were left out entirely. Combined with `getColumnWidth()` falling through to `"auto"` for `columnType: "pairwise"`, the column had to squeeze into whatever sliver of space was left, wrapping "No verdict yet" one character per line.
**Fix:** pairwise columns now contribute `TARGET_COL_DEFAULT_PCT` to the total, and `getColumnWidth()` returns a real percentage for them instead of `"auto"`.

### 10. Raw `{"error":"candidate_a_id is required..."}` JSON shown to the user
**Before:** `evaluations-legacy.ts` returned an untyped `c.json({error: "..."})` blob for a missing required field; the client's generic candidate-missing message ("Run the upstream prompts first") was also misleading here since Variant A/B were never configured at all, not just not-yet-run.
**Fix:** added `EvaluatorMissingFieldError` (a proper `DomainError` subclass) carrying `{field, evaluatorName}` in `meta`; the route now returns that structured shape (keeping the `error` string identical for backward compatibility, `kind`/`meta` are additive). `PairwiseCompareCell.tsx` now has a specific, more accurate friendly message ("Variant A/B not configured — open Edit Configuration and select both variants") for exactly this shape, checked before the generic fallback.

### 11. Include cost/duration toggles reset when reopening the drawer
**Before:** `EvaluatorEditorShared.tsx` reset the form to schema defaults whenever `defaultSettings` changed reference — which happens whenever the cascade-resolved default-model query resolves asynchronously, even after the user already touched the form.
**Fix:** added an `isDirty` guard so the reset only applies before the user has started editing.
**Verified live:** toggled "Include duration" on, saved, closed the drawer, reopened it via the normal UI flow — the toggle (and Variant A/B/Golden field) were still there, unchanged.

### 12. "New" badge
Added a "New" badge to the Pairwise Compare card in the type-selector drawer, since it's a newly-shipped comparison mode.

### 13. (found live during QA) Variant A / Variant B / Golden field silently reset on unrelated interaction
**Before:** `EvaluatorEditorShared.tsx` passed a fresh `{variantA: "", variantB: "", ...}` object literal as `PairwiseConfigForm`'s `value` prop on *every render* (since `pairwiseContext.initialPairwise` is `undefined` for a new comparison). `PairwiseConfigForm` re-syncs its local draft state whenever that prop's reference changes — so toggling Include cost, or any other unrelated re-render, silently wiped whatever the user had already picked for Variant A/B. This is worse than #11 in the same family of bugs — it could wipe the entire configuration, not just two toggles.
**Fix:** hoisted a stable `EMPTY_PAIRWISE_CONFIG` module-level constant so the reference is stable across renders.

---

## Known separate issue (out of scope, not introduced by this PR)

While re-verifying #8/#11 live, I found that **deep-linking or hard-refreshing the browser while the Pairwise Compare edit drawer is open loses the pairwise-specific UI** (Variant A/B/Golden picker disappears, "Include Metrics" falls back to the generic per-field toggle style). Root cause: `useOpenEvaluatorEditor.ts` passes `pairwiseContext` (and the `onPairwiseChange` callback) as "non-serializable extras" via an in-memory `setFlowCallbacks` registry, not through the drawer's URL query state. A cold page load re-opens the drawer from the URL alone, so those extras are never populated and the drawer falls back to the generic evaluator form.

This is a pre-existing drawer-architecture limitation (likely affects every evaluator's mapping drawer on refresh, not just pairwise), not something this PR's changes caused, and normal usage (opening the drawer by clicking through the UI) is unaffected — all screenshots below were captured that way. Flagging it here since it's a real gap worth a follow-up ticket, but fixing it (persisting non-serializable drawer state across a cold reload) is a bigger, separate change.

## Screenshots (verified live in a fresh worktree + browser session)

**"New" badge on Pairwise Compare, correct "Add Comparison" button:**
new badge

**Model picker + provider icons legible in dark mode:**
model picker dark mode

**Drawer: consistent padding, no duplicate Swap/Tie labels, "Add Comparison" button:**
drawer fixes

**Variant A survives an unrelated toggle click (bug #13), "Include metrics" now a clear section header (bug #7):**
variant persists

**Fully configured drawer, all fields retained:**
fully configured

**Pairwise result column at proper width, "No verdict yet" fully legible (bug #9) + "+ Add" button correctly labeled (bug #2):**
column width
add button label

**Swap And Confirm / Allow Tie each show a single label (bug #8, follow-up fix), Variant A/B/Golden field render correctly:**
single label

**Include cost/duration toggles persisted after save + reopen (bug #11):**
toggle persistence

## Test plan

- [x] `pnpm typecheck` passes
- [x] Manually reproduced and verified fixes #1, #2, #4, #5, #6, #7, #8, #9, #11, #12, #13 live in a browser (dark mode, real prompts + pairwise column, save + reopen cycles)
- [x] #3 verified with a new regression test in `EvaluatorChip.test.tsx` (click calls `onEdit` once, never opens the chip's own menu) — all 24 tests in the file pass
- [ ] End-to-end run with a real verdict wasn't captured — the local NLP Go service (`nlpgo`) had a pre-existing startup script issue unrelated to this change, so target prompt execution couldn't complete in this session

## CodeRabbit review round

Addressed all actionable findings from CodeRabbit's review:
- Renamed `menuOpen`/`setMenuOpen` (EvaluatorChip) and `topLevel` (DynamicZodForm) to `isMenuOpen`/`isTopLevel` to satisfy the repo's boolean-naming convention.
- Removed a redundant `createEvaluator()` call in the new regression test.
- Reused the already-computed `providerKey` instead of re-deriving it in `ModelSelector`.
- Fixed a real, pre-existing bug in `totalColumnPercentage` (EvaluationsV3Table): the dataset/target summation loops reconstructed column IDs as `dataset_${id}` and bare `target.id`, but resize actually stores under the header's real ID (`dataset.${id}` / `target.${targetId}`). The mismatch meant a resized dataset/target column's width never counted toward the table's total — same root cause as the pairwise-column width bug (#9) this PR already fixes, just for the other two column types.
- Fixed the CI lint failure (Biome formatting) on the initial push.

## Summary by CodeRabbit

* **New Features**
  * Added a “New” badge and “Add Comparison” action for the pairwise comparison flow.
* **Bug Fixes**
  * Fixed pairwise evaluation table column sizing/layout.
  * Improved evaluator missing-field validation and legacy error responses (including proper HTTP status).
  * Clarified Variant A/B not-configured messaging in pairwise compare.
  * Prevented form resets from overwriting in-progress edits and avoided duplicate boolean field titles.
  * Made evaluator chip edit behavior deterministic without opening the menu.
* **Style**
  * Standardized provider/model icon rendering and refreshed pairwise spacing/typography and model search input styling.
* **Tests**
  * Added a regression test for missing-mappings alert icon behavior.


Aryansharma28:
Filed a follow-up from this bug bash: #5378 — Pairwise Compare should support comparing A vs B without requiring a golden/reference answer (currently the "Golden field" picker forces a golden-answer framing even when you pick `input` as a workaround).

This PR doesn't close #5378 — just referencing it here since it surfaced during this same QA pass.

sergioestebance:
Blessed. 🙏 Traced every claim to the code — column IDs, glyph equivalence, isTopLevel threading, and both state-reset fixes all hold; CI fully green.

_"She is not afraid of the snow for her household: for all her household are clothed with scarlet." — Proverbs 31:21_

Ship it — mind the P2 test gap: #11 (isDirty guard) and #13 (stable pairwise ref) are runtime state bugs with no automated regression coverage yet.

sergioestebance (approved):

PR #5381, opened 2026-07-06
feat(pairwise-compare): support comparing A vs B without a golden/reference answer

### LINKED ISSUE #5378: Pairwise Compare: support comparing A vs B without requiring a golden/reference answer
## Problem

Pairwise Compare's judge prompt is hardcoded "golden-aware" — every comparison assumes a golden/reference answer to compare candidates against:

> Compare two candidate outputs against a known-good reference (golden answer).

The config drawer's "Golden field" picker only lets you choose *which dataset column* to treat as that reference. It doesn't have a real mode for datasets that genuinely have no reference answer. The drawer's own helper text hints at the gap:

> Pick `input` only if your dataset has no reference answer and you want the judge to compare candidates against the question itself (rarely useful).

That's a workaround, not a real option — the prompt still frames whatever's picked as a "golden answer," even when it's just the input question. That's semantically wrong and can confuse the judge / skew its verdict.

## Ask

Add an explicit option to opt out of golden-answer comparison, so users can directly compare Candidate A vs Candidate B on their own merits (correctness, style, helpfulness, etc.) without requiring — or faking — a reference answer. When this is toggled off, the judge prompt should drop the golden-answer framing entirely rather than substituting `input` as a fake golden.

## Context

Found during the Pairwise Compare bug-bash QA in #5367 (screenshots/fixes there). That PR doesn't close this — referencing it here since it came out of the same pass.

Aryansharma28:
# Related Issue

- Resolves #5378

## Summary

Pairwise Compare's judge prompt was hardcoded "golden-aware" — every comparison assumed a golden/reference answer to compare candidates against. The config drawer's "Golden field" picker only let you choose *which* dataset column to treat as that reference; there was no real way to opt out. Picking `input` as a workaround still framed the prompt as golden-aware, which could confuse the judge / skew its verdict.

This adds a real **"Has golden answer"** toggle. Default stays `true` (no behavior change for existing evaluators). Turned off, the judge compares Candidate A vs Candidate B directly on their own merits, with no reference answer involved at all.

Closes #5378

## Screenshots

**Has golden answer: ON (default)** — Golden field picker visible, judge compares each candidate against the reference:

Has golden answer toggle ON, Golden field picker visible

**Has golden answer: OFF** — Golden field picker hidden entirely, judge compares Candidate A vs Candidate B directly:

Has golden answer toggle OFF, Golden field picker hidden

### Judge (Python, `langevals_langevals/pairwise_compare.py`)
- New `has_golden_answer: bool` setting (default `true`).
- New `DEFAULT_PAIRWISE_PROMPT_NO_GOLDEN` template — no "Golden answer:" line, no golden framing, asks the judge to compare the two candidates directly.
- When `has_golden_answer` is `false` **and** the prompt hasn't been customized, the evaluator swaps in the golden-free template. A customized prompt is always respected as-is — the toggle never silently rewrites a user's own template.
- Regenerated `evaluators.generated.ts` (both the `langevals/ts-integration` source and the `langwatch` copy) via `generate_evaluators_ts.py` + `copy:langevals-types`.

### Frontend (`langwatch/`)
`hasGoldenAnswer` is mirrored into `PairwiseEvaluatorConfig` (alongside `variantA`/`variantB`/`goldenField`) — the same dual-representation pattern already used for `includeMetrics`, since the orchestrator's cell-generation guard and the missing-mappings validator only see `target.pairwise`, not the Python settings object.

- **`PairwiseConfigForm.tsx`**: new "Has golden answer" switch. Turning it off hides the Golden field picker entirely (previously always shown and always required).
- **`EvaluatorEditorShared.tsx`**: skips `has_golden_answer` from the generic settings form (rendered inline next to the picker instead), seeds it on new pairwise configs.
- **`mappingValidation.ts`**: Golden field is only flagged as a missing/required mapping when `hasGoldenAnswer` isn't explicitly `false`.
- **`orchestrator.ts`**: found and fixed a real bug while wiring this up — the column-target cell-generation guard required `goldenField` unconditionally, which would have silently blocked every no-golden pairwise column from ever running (the "Waiting on..." cell state would persist forever, indistinguishable from a normal not-yet-run state).
- **`EvaluationsV3Table.tsx`**: the "is this pairwise column configured" checks (stable-columns filter, pending-pairwise default) had the same unconditional `goldenField` requirement — fixed the same way.

All four `!== false` guards default old saved evaluators (created before this field existed) to the current golden-required behavior, so nothing already configured changes.

## Test plan

- [x] Python: 4 new unit tests (`has_golden_answer` defaults true, golden framing present by default, framing dropped when off — using a marker string distinct from the candidate outputs to actually prove the template swap rather than an empty substitution, and a customized prompt is respected regardless of the toggle). Full `pairwise_compare` suite: 15/15 passing.
- [x] Frontend: new `PairwiseConfigForm.integration.test.tsx` (4 tests) — picker shows/hides with the toggle, `onChange` reports `hasGoldenAnswer` and clears `goldenField` when turned off.
- [x] `mappingValidation.test.ts`: 4 new pairwise-specific cases (golden required only when the toggle is on; variantA/variantB always required).
- [x] `orchestrator.test.ts`: 2 new cases proving the cell-generation guard fix (a no-golden pairwise column now actually generates cells; a golden-required column with an empty golden field still correctly skips).
- [x] `pnpm typecheck` passes; `biome check` clean on all touched files.
- [x] BDD spec extended (`specs/experiments/pairwise-compare-mvp.feature`) with the new scenarios before writing any code.

**Known gap:** `evaluators.generated.ts` reformatting includes incidental noise unrelated to this change — there's no pinned prettier version anywhere in this repo, and the generation script's own `npm install -g prettier` step isn't reliable in a fresh environment, so I reformatted both copies with Biome (this repo's actual TS formatter/linter) instead. The diff for that file includes reflowing of unrelated pre-existing fields; the only deliberate change is the new `has_golden_answer` entry. Called out in the commit message for that file.

## Summary by CodeRabbit

* **New Features**
  * Added a **“Has golden answer”** option for pairwise evaluations, switching the judge between golden-reference comparisons and direct Candidate A vs Candidate B “own merits” judging.
  * Enabled by default for pairwise evaluator configuration.

* **Bug Fixes**
  * Golden field selection and target validations are now required only when **Has golden answer** is enabled.
  * Judge prompt text and the reasoning wording adapt to the selected mode (including when the prompt is not customized).

* **Tests**
  * Added/updated coverage for conditional UI rendering, toggle behavior, prompt generation, validation, and pairwise execution scenarios.


sergioestebance (dismissed):

sergioestebance:
Blessed. 🙏 The golden-field gate now flows from one spring — `isGoldenFieldSatisfied` — so UI, validation, and orchestrator can never speak against each other.

> "Let your 'Yes' be 'Yes,' and your 'No,' 'No.'" — Matthew 5:37

Traced the judge swap, the real orchestrator wedge-fix, and the settings↔pairwise mirror: no drift path. Ship it — only residual is P3 generated-file formatting noise, already called out.

sergioestebance:
Blessed. 🙏

"Iron sharpens iron, and one man sharpens another." — Proverbs 27:17

Ship it — nothing outstanding, every CodeRabbit thread already closed out.

sergioestebance (approved):

Issue #4476, opened 2026-06-01
test(ci): test-unit (2) intermittently hangs ~25min on an open handle and times out, blocking langwatch-app-ci


drewdrewthis:
**Expected**

Each unit-test shard in `langwatch-app-ci` completes and the vitest worker exits promptly after its tests pass.

**Actual**

Shard `test-unit (2)` (of 4) runs all its tests to completion (they pass), then the worker **hangs ~23 min** on an unref'd open handle, hitting the job's 25-min timeout → `##[error]The operation was canceled` → the `langwatch-app-complete` aggregator fails and the whole run goes red/cancelled.

**Reproduction / context**

- Intermittent: main run `99bc3b9d2` (2026-06-01 09:37 UTC) was green, but the `c8c6495b8` main run [26750014253](https://github.com/langwatch/langwatch/actions/runs/26750014253) cancelled `test-unit (2)` at ~25 min (10:42→11:08 UTC), and PR #4468's runs hung identically **twice** at `25m17s` (re-run via `gh run rerun --failed`) — currently deterministic on affected commits.
- Consistent with the posthog-node local-evaluation `setInterval` (not unref'd) under `--coverage` graceful-worker-exit, documented in `langwatch/test-setup.ts` (which already shuts posthog down in `afterAll` as a partial mitigation; the 25-min job timeout is the backstop). The last shard-2 test activity is followed by ~23 min of silence, then the cancellation — i.e. the tests pass but the worker won't exit.
- Impact: blocks PRs from an all-green CI rollup (surfaced while landing the unrelated AICreateModal fix in #4467 / PR #4468 — that PR's own shards, incl. `test-unit (4)`, passed).
- Related but distinct (do not dedupe): #4450 (integration shard 4 wedged by an `api-keys-scope-filter` infinite render loop, masked green by a globalSetup `process.exit(0)` — same shard-wedge *family*, but integration-not-unit, render-loop-not-open-handle, masked-green-not-timed-out-red); #3200 (PostHog ERROR-level rate-limit spam).
- Triage pointers for whoever picks it up (not prescriptive): identify which shard-2 test leaves the open handle under coverage (candidates: posthog poller, redis/clickhouse/pg socket, s3 client); options include unref'ing the poller, forcing pool teardown, a `process.exit` backstop like the integration globalSetup, or bisecting from the last-green main `99bc3b9d2`.

---
*Classification: Bug (CI infrastructure)*
[...]

PR #5394, opened 2026-07-06
fix(ci): drop JSON reporter from unit shards + lower hang hard-floor (#4476)

drewdrewthis:
## Why

\`test-unit (2)\` consistently ran ~6m45s vs ~2m54s for shards 1/3/4. The extra ~4min is a vitest finalize wedge: after the last \`afterAll\`, an open-handle dump shows only the vmThreads worker's own IPC \`MessagePort\` (application code is clean — posthog/tiktoken/redis leaks cleared, tiktoken removed by #5003). The vitest main process then idles in its finalize path.

The \`--reporter=json\` flag (with \`--outputFile=test-results.json\`) was the prime suspect: it forces vitest to accumulate all test results in-memory and flush a JSON file during the finalize path. Removing it eliminates that code path. The \`--coverage\` flags (lcov + text) are retained as they feed Codecov.

The existing \`HARD_FLOOR_MS = 6 * 60 * 1000\` was masking the wedge rather than preventing it — every shard-2 run hit the floor, burning ~3–4 min of CI wall-clock per push.

## What changed

1. **\`.github/workflows/langwatch-app-ci.yml\`** — removed \`--reporter=json --outputFile=test-results.json\` from the unit-test run command. The downstream \`Extract test failures\` step already guards against a missing file (\`if [ ! -f "$INPUT_JSON" ]; then … exit 0; fi\` in \`extract-failures.sh\`), so no change needed there.

2. **\`langwatch/src/test-unit-global-setup.ts\`** — lowered \`HARD_FLOOR_MS\` from \`6 * 60 * 1000\` to \`4 * 60 * 1000\`. Healthy shards all finish <3 min; 4 min is safe and caps waste at ~1 min if the wedge persists via a different path.

## Human verification

- [ ] Confirm shard \`test-unit (2)\` duration is reduced vs pre-fix (~6m45s)
- [ ] Check whether the hard-floor line appears in shard-2 log

CI run #28811013820 (this PR's second run — first failed due to a pre-existing broken lockfile fixed by #5393).

**Results (honest):**

| Shard | Duration | Floor fired? |
|-------|----------|--------------|
| test-unit (1) | ~3m15s | no |
| test-unit (2) | **~4m55s** | **yes** — at 4 min |
| test-unit (3) | ~2m56s | no |
| test-unit (4) | ~3m15s | no |

Log evidence from shard 2:
\`\`\`
[unit globalSetup] hard floor reached at 4 min — forcing process.exit(0) to release the CI step from a vitest finalize wedge
\`\`\`

**Verdict:** The JSON reporter removal did NOT cure the wedge. Shard 2 still hits the floor (now at 4 min instead of 6 min). Net improvement: **~1m50s saved per PR push** (4m55s vs 6m45s). The Tier 2a hypothesis (JSON reporter = wedge cause) is falsified. The wedge lives in vitest's own finalize path upstream of the reporter. This PR is still net-positive via the floor reduction, but the primary fix hypothesis did not hold.

## Scope

Addresses #4476 (Tier 1+2a of the tiered plan; leaves vitest 4.x bump check and upstream vitest report as follow-ups). Tier 2a (JSON reporter removal) is falsified as a root-cause fix — escalates the need for the vitest upstream investigation.

### DIFF
--- a/.github/workflows/langwatch-app-ci.yml
+++ b/.github/workflows/langwatch-app-ci.yml
@@ -126,7 +126,7 @@ jobs:
         run: |
           pnpm test:unit --run --shard=${{ matrix.shard }}/4 \
             --coverage --coverage.reporter=lcov --coverage.reporter=text \
-            --reporter=default --reporter=json --outputFile=test-results.json \
+            --reporter=default \
             --dangerouslyIgnoreUnhandledErrors
 
       - name: Extract test failures
--- a/langwatch/src/test-unit-global-setup.ts
+++ b/langwatch/src/test-unit-global-setup.ts
@@ -29,10 +29,10 @@ export async function setup(): Promise<void> {
   // langwatch-app-complete required check unblocks. Unref'd so a healthy
   // shard exits immediately on its own; the timer only fires on the wedge.
   // Mirrors the integration globalSetup hard-floor; unit shards otherwise lack
-  // one. A healthy unit shard finishes in ~3 min, so a 6-min floor only fires
-  // on a wedge and caps the wasted wall-clock at ~3 min of idle.
+  // one. A healthy unit shard finishes in ~3 min, so a 4-min floor only fires
+  // on a wedge and caps the wasted wall-clock at ~1 min of idle.
   if (process.env.CI) {
-    const HARD_FLOOR_MS = 6 * 60 * 1000;
+    const HARD_FLOOR_MS = 4 * 60 * 1000;
     const timer = setTimeout(() => {
       // eslint-disable-next-line no-console
       console.log(


0xdeafcafe (approved):

drewdrewthis:
## Review verdict: READY

**Reviewed at:** \`c7e5c1ec\` · **Run:** pr-review (reviewer: @hygiene-reviewer)

### Summary

Two-file CI fix. Reviewed: (1) missing-file guard in `Extract test failures` consumer — verified `extract-failures.sh` has `if [ ! -f "$INPUT_JSON" ]; then … exit 0; fi` and step runs only `if: failure()`, so removing the JSON reporter output file causes no cascade. (2) 4-min floor safety — healthy shards all finish <3 min per PR evidence; 4-min floor gives 1 min headroom. (3) PR body honesty — PR openly documents that shard 2 still hits the floor at 4 min (now 4m55s vs 6m45s), the JSON-reporter hypothesis was falsified, and net savings are ~1m50s. Body matches the observable CI evidence exactly.

### Non-blocking
- The root cause (vitest finalize wedge above the reporter level) is unresolved — the PR correctly scopes this as a follow-up (Tier 2b+). (New Issue — already tracked under #4476)

---

PR #3537, opened 2026-04-27
feat(projects): server-side resolveDefaultModel service (split 2 of #3526)

drewdrewthis:
## Summary

Split 2 of 5 from umbrella #3526 / PR #3529. Adds `ProjectService.resolveDefaultModel(projectId)` as the single source of truth for the effective default model across UI, API, agents, and scenarios.

Closes #3531.

Resolution chain:
1. `project.defaultModel` set + provider enabled → return it
2. else first usable provider (custom OR env-fallback) → return its `PROVIDER_DEFAULT_MODELS` value
3. else null

## What's in this split

**New:**
- `langwatch/src/server/modelProviders/providerDefaultModels.ts` — shared `PROVIDER_DEFAULT_MODELS` map (server-importable)
- `langwatch/src/server/app-layer/projects/__tests__/project.service.unit.test.ts` — BDD coverage for all 4 branches + edge cases

**Modified (full):**
- `project.service.ts` — new `resolveDefaultModel` method
- `presets.ts` — DI wiring
- `api/routers/project.ts` — new `getResolvedDefaultModel` tRPC procedure
- `scenarios/execution/orchestrator.ts` + `orchestrator.types.ts` — new `defaultModelResolver` dep
- `scenarios/__tests__/orchestrator.unit.test.ts`
- `scenarios/simulation-runner.service.ts` — uses resolver
- `prompt-config/prompt.service.ts` — pre-resolves before transaction
- `prompt-config/repositories/llm-config.repository.ts` — accepts pre-resolved value
- `evaluations/getEvaluator.ts` — uses pre-resolved value
- `modelProviders/utils.ts` — comment directing callers to pre-resolve

**Partial-file cherry-picks (intentional overlap with sibling splits):**
- `scenarios/execution/data-prefetcher.ts` — resolver dep + usage only; env reads belong to split 5
- `components/scenarios/ScenarioCreateModal.tsx` — new tRPC consumer call only; modal gate logic belongs to split 1

## Provenance

Cherry-picked from validated WIP umbrella branch `issue3526/stress-test-all-3-agent-pathways` (PR #3529). Code is already validated end-to-end on the umbrella; this split is for focused review.

## Test plan

- [ ] `pnpm typecheck` clean (CI)
- [ ] Service tests pass (`project.service.unit.test.ts`, `orchestrator.unit.test.ts`)
- [ ] tRPC `project.getResolvedDefaultModel` returns expected resolved value end-to-end
[...]

Issue #3531, opened 2026-04-27
Server-side resolveDefaultModel service (split 2 of #3526)


drewdrewthis:
## Summary

Split 2 of 5 from umbrella #3526 / PR #3529. Cherry-pick the server-side default-model resolver (F2) into a focused PR for review.

**Parent:** #3526
**Bundle:** #3529 (draft umbrella, working code already validated end-to-end)
**Mini-epic:** #3441

## Why

When env-fallback enables a provider but `project.defaultModel` is null, every consumer (UI, API, agents, scenarios) gets a different answer or a "no default" error. This split adds a single source of truth: server-side `ProjectService.resolveDefaultModel(projectId)`. `project.defaultModel` becomes "user override," not "the answer."

## Plan

1. Create worktree off main
2. Cherry-pick the resolver subset (file list below)
3. Re-run typecheck + service tests
4. PR + Slack dev when CI green

### Files to cherry-pick (resolver scope)

**New:**
- `langwatch/src/server/modelProviders/providerDefaultModels.ts` (shared `PROVIDER_DEFAULT_MODELS` map)
- `langwatch/src/server/app-layer/projects/__tests__/project.service.unit.test.ts` (BDD tests)

**Modified:**
- `langwatch/src/server/app-layer/projects/project.service.ts` (new `resolveDefaultModel` method)
- `langwatch/src/server/app-layer/presets.ts` (DI wiring)
- `langwatch/src/server/api/routers/project.ts` (new `getResolvedDefaultModel` tRPC procedure)
- `langwatch/src/server/scenarios/execution/orchestrator.ts` + `orchestrator.types.ts` (new `defaultModelResolver` dep)
- `langwatch/src/server/scenarios/__tests__/orchestrator.unit.test.ts`
- `langwatch/src/server/scenarios/simulation-runner.service.ts` (uses resolver)
- `langwatch/src/server/scenarios/execution/data-prefetcher.ts` (resolver dep + usage — partial-file cherry-pick; env reads belong to split 5)
- `langwatch/src/server/prompt-config/prompt.service.ts` (pre-resolves before transaction)
- `langwatch/src/server/prompt-config/repositories/llm-config.repository.ts` (accepts pre-resolved value)
- `langwatch/src/server/evaluations/getEvaluator.ts` (uses pre-resolved value)
- `langwatch/src/server/modelProviders/utils.ts` (comment directing callers to pre-resolve)
[...]

PR #3529, opened 2026-04-27
wip: bundle of stress-test findings for #3526 (umbrella PR)

drewdrewthis:
## Summary

**Umbrella WIP PR for #3526** — bundled validation testbed for 9 findings surfaced during the stress-test walk-through of all 3 agent pathways (code / http / workflow).

This PR will **not** be merged. It exists so Drew can validate the entire bundle works end-to-end before each finding splits off into its own focused PR for review.

## Strategy

| Phase | Status |
|-------|--------|
| 1. Bundle all findings into this WIP umbrella branch | ✅ done |
| 2. Drew validates bundle works end-to-end | 🟡 in progress |
| 3. Split into 5 themed PRs against main, each with its own issue + plan + ACs | pending |
| 4. Each split goes through normal review → Slack dev channel | pending |
| 5. Umbrella branch rebased on top as splits merge (or closed once all splits merge) | pending |

See #3526 for the full split plan + per-finding ACs.

## Findings included (9 total, fix-in-this-PR scope from #3526)

- **F1** — Scenario create modal dead-end on `no-default` / `stale-default` states
- **F2** — Server-side `resolveDefaultModel` service (single source of truth)
- **F3** — Provider drawer masked-key no-op (don't re-submit placeholder)
- **F4** — `useOrganizationTeamProject` cache: drop `staleTime: Infinity`
- **F5** — `AICreateModal` `ErrorState` real-error + classifier
- **F6** — Scenario editor criteria-on-blur (silent data loss)
- **F8** — Save-and-run: failed run must not roll back save
- **F10** — Telemetry routing + env-contract sweep (drop localhost fallbacks)

Filed elsewhere: F7 (#3411), F9 (#3412) — both under mini-epic #3441.

## Test plan

- [ ] Manual: stress-test matrix per #3526 (code / http / workflow × OpenAI / Gemini)
- [ ] Manual: each finding's reproducer no longer reproduces
- [ ] Per-split PRs validate their own slice independently (with their own tests + reviewers)

## Reviewer guidance

**Don't review this PR.** It's a validation bundle. Review the split PRs (linked from #3526) once they're cut.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

# Related Issue

- Resolve #3526

Issue #3526, opened 2026-04-27
Stress test all 3 agent pathways for scenarios on-platform


drewdrewthis:
## Summary

Stress test all 3 agent pathways for scenarios on-platform. Each agent type — **code**, **http**, **workflow** — must be runnable as a scenario participant via the on-platform scenarios feature, end-to-end without errors on our part.

**The goal is not just "doesn't crash" — it's hyper-vigilant UI/UX perfection.** Every dead state, stale-cache moment, error message that says "Internal server error" instead of "your provider key isn't configured," every loading spinner that lies, every micro-friction that makes you hesitate — those are the deliverables. The 9 findings already in the bundle came from exactly this lens; this matrix walk is to surface the *next* batch.

**Parent epic:** #3441 — mini-epic for broken mappings across eval v3, workflows, agents, and scenarios. The scenario-side findings F7/F9 (and adjacent fixes F2/F3/F5/F10) belong to this epic's scope.

## PR strategy: one bundle, then split

This PR (the "umbrella" for #3526) will land **all in-scope findings together**. Once Drew confirms the bundle works end-to-end manually, the PR is split into themed follow-up PRs grouped by logical concern — each with its own issue and review cycle.

**Why bundle first:**
- Findings are interdependent (entry-path UX needs cache freshness needs server-side default resolution); validating in isolation is harder than validating the whole flow once.
- Drew can confirm "everything works" once before splitting, not N times across N PRs.

**Split structure** (each split = its own issue + PR):

| Split | Issue | Findings | Theme |
|-------|-------|----------|-------|
| 1 | #3530 | F1, F4, F5 | Scenario create entry-path UX (modal, cache, error UX) |
| 2 | #3531 | F2 | Server-side `resolveDefaultModel` service |
| 3 | #3532 | F3 | Provider drawer masked-key no-op |
| 4 | #3533 | F6, F8 | Scenario editor data-loss fixes (criteria-on-blur, save-and-run) |
| 5 | #3534 | F10 + localhost-fallback sweep | Env contract: telemetry from runtime, no localhost fallbacks |

**Umbrella WIP PR:** #3529 (draft, do not merge — bundle for end-to-end validation only)

Each split issue has a concrete plan + ACs distilled from the validated work in #3529. Splits cherry-pick the relevant subset of files into a fresh worktree, run their own focused review, and merge independently. After all splits land, this umbrella branch can be closed.

## Walk-through ownership

[...]

PR #5398, opened 2026-07-06
feat(projects): server-side resolveDefaultModel service (F2 of #3526)

drewdrewthis:
## Why

When `env`-fallback enables a provider but no explicit default model is configured, every consumer (UI, API, agents, scenarios) gets a different answer or a "no default" error. This split adds a single source of truth: `ProjectService.resolveDefaultModel(projectId)` — a provider-walk that finds the first enabled provider with a canonical default in `PROVIDER_DEFAULT_MODELS`. The full cascade (project → team → org) is handled by the existing `modelProvider.getResolvedDefault` tRPC endpoint; this method is the lightweight fallback for callers that have not yet migrated.

## What

- **`modelProvider.constants.ts`** — new `PROVIDER_DEFAULT_MODELS` map + `PROVIDER_RESOLUTION_ORDER` list; server-importable, no client-only path. (Renamed from `providerDefaultModels.ts` to follow the codified `<singularDomainNoun>.constants.ts` convention — see the review thread + `dev/docs/best_practices/typescript.md`.)
- **`ProjectService.resolveDefaultModel(projectId)`** — walks `PROVIDER_RESOLUTION_ORDER`, returns first enabled provider's `PROVIDER_DEFAULT_MODELS` entry, else null. Degrades gracefully (try/catch → null) on provider-lookup failure. (`project.defaultModel` column was removed in ADR-021/iter-109.)
- **`project.getResolvedDefaultModel` tRPC procedure** — thin wrapper delegating to the service method
- **`ModelProviderService` DI wiring** — `presets.ts` injects `ModelProviderService.create(prisma)` into `ProjectService`; null preset unchanged
- **BDD unit tests** — 5 provider-centric cases covering all branches

Note on AC6 (`ScenarioCreateModal` wiring): the modal already uses `api.modelProvider.getResolvedDefault` (richer feature-key cascade, already in main before this split). No modal change is needed in this split.

## Deployment Impact

**None.** No migrations, env vars, chart/service, or self-hosting changes. The deployment-surface touch is documentation-only: a one-line naming-convention note in `dev/docs/best_practices/typescript.md`, plus a pure source-file rename (`providerDefaultModels.ts` → `modelProvider.constants.ts`, single importer updated). No runtime behavior change.

## Human verification

1. `pnpm typecheck` clean
2. `vitest run src/server/app-layer/projects/__tests__/project.service.unit.test.ts` — 5/5 pass

`project.getResolvedDefaultModel` has no direct frontend consumer in this PR — `ScenarioCreateModal` continues to use `api.modelProvider.getResolvedDefault` (unchanged). The new tRPC endpoint is a server-side addition for future consumers and for direct API callers.

**Evidence floor (backend-only):**
- Unit tests: 5/5 passing — cover enabled provider → canonical default, bedrock-skip, resolution-order priority, all-disabled → null, null-preset → null
- Code trace: `project.getResolvedDefaultModel` (router) → `getApp().projects.resolveDefaultModel` (service) → `PROVIDER_RESOLUTION_ORDER` walk → `PROVIDER_DEFAULT_MODELS[providerId]` → returns first hit or null

---

Addresses #3531; F2 of #3526 (split from #3529). Supersedes #3537.

https://claude.ai/code/session_01SufnT5rv6hGhd2jJ1Z5yeA

## Summary by CodeRabbit

* **New Features**
  * Added automatic resolution of a project’s default model based on its enabled model providers.
  * Added support for canonical defaults such as OpenAI, Anthropic, and Gemini models.
  * Exposed the resolved default model through the project API.

* **Bug Fixes**
  * Default model resolution now safely returns no result when providers are unavailable or unsupported.

* **Documentation**
  * Added guidance for naming constants files consistently.


drewdrewthis:
## Review verdict: READY

**Run:** pr-review (reviewer: @hygiene-reviewer) · **Reviewed at:** `fd9431df7`

Delta since the prior verdict (`3147df14c`) is a **pure file rename** addressing the owner's review comment, plus the codified convention. No logic change.

**Verified (delta `3147df14c..fd9431df7`):**
- 3 files: `providerDefaultModels.ts` → `modelProvider.constants.ts` (`similarity index 100%` — byte-identical), the sole importer `project.service.ts` (import path only), and a one-line convention note in `dev/docs/best_practices/typescript.md`.
- `grep -rn providerDefaultModels` across `src` / `dev` / `specs` → **zero** dangling references to the old module name.
- Renamed module still exports `PROVIDER_DEFAULT_MODELS` + `PROVIDER_RESOLUTION_ORDER`; importer's new path resolves. `resolveDefaultModel` behavior unchanged.
- `project.service` unit tests: **5/5 pass** locally against the renamed module.
- Convention codified (`<singularDomainNoun>.constants.ts`) is consistent with the `scenario.constants.ts` precedent and the sibling `modelProvider.*` files.

**All review-opened threads resolved.** The owner's naming thread is resolved (renamed + convention codified; singular chosen over the floated plural for consistency — flagged there for veto).

**⚠ Disclosure — inherited CI noise (NOT introduced by this PR):** the `test-unit (2)` shard force-exits green while masking **3 pre-existing failures** — `aggregation-builder.test.ts` (1) and `team.update.multi-binding.unit.test.ts` (2). Proven inherited: the identical 3 failures reproduce at the branch base `3147df14c` (pre-rename), so they are causally unrelated to this rename (which touches neither file, directly or transitively). Out of scope for this rename; called out so the forged-green shard isn't mistaken for a regression here.

langwatch-agent:

**Reviewer:** @rogeriochaves · **Mode:** `Targeted Review`
**Linked issue:** F2 of #3526 (default-model single source of truth). Scope matches the issue: a lightweight `resolveDefaultModel` fallback for callers not yet on the `getResolvedDefault` cascade.

**Scariest production risk:** model resolution is a provider-walk that returns "the first enabled provider with a canonical default." If that ordering is not deterministic/priority-ranked, different consumers can silently resolve to different default models.

**Top inspection targets**
- `Must Check` — `project.service.ts`: the provider-walk order. Is it stable and intentional (priority), or dependent on map/query iteration order?
- `Must Check` — `providerDefaultModels.ts`: what happens when the enabled provider has no entry in `PROVIDER_DEFAULT_MODELS` (clean error vs silent skip vs wrong default)?
- `Ask Author` — does this diverge from the `modelProvider.getResolvedDefault` tRPC cascade over time (two sources of truth drifting)?

Additive-ish (+245/-2, 5 files), CI green, no unresolved threads.

langwatch/src/server/modelProviders/modelProvider.constants.ts
```diff
@@ -0,0 +1,45 @@
+/**
```
drewdrewthis, line 1: The naming convention is wrong -- this should be in a constants file. 

probably something like `modelProviders.constants.ts` -- we should decide and codify the convention.

drewdrewthis:
Picking up your review feedback (model-resolution dedup pass, coordinated with #5469).

**On `providerDefaultModels.ts` naming** — agreed, it's a constants module, not a service. Plan: `git mv` it to a `*.constants.ts` name and codify the convention. You said "we should decide" the exact name — proposing **`modelProviders.constants.ts`** (matches your suggestion); will rename to whatever you confirm.

While in that file I'll also fold in coderabbit's Major on `project.service.ts:133` — `resolveDefaultModel` has no error handling around the provider lookup, which breaks its own documented "returns null when no usable providers" contract.

This PR is the **canonical source of truth** for default-model resolution; #5469 will be reworked to consume `ProjectService.resolveDefaultModel` rather than duplicate the cascade (details on #5469). The code lands via a follow-up push from a dedicated lane — this comment is the plan + checkpoint. **Not merging.**

langwatch/src/server/modelProviders/modelProvider.constants.ts
drewdrewthis: Done in `fd9431df7` — renamed to `modelProvider.constants.ts` and codified the convention in `dev/docs/best_practices/typescript.md`: module-level constant collections use `<singularDomainNoun>.constants.ts`.

I went with the **singular** `modelProvider.constants.ts` rather than the plural you floated, to match (a) the only existing precedent — `scenario.constants.ts` (singular noun, in the plural `scenarios/` dir), and (b) the sibling `modelProvider.service.ts` / `.repository.ts` / `.authz.ts` in this same directory. Happy to switch to `modelProviders.constants.ts` if you'd rather codify the plural — it's a one-line `git mv`.

langwatch/src/server/modelProviders/modelProvider.constants.ts
```diff
@@ -0,0 +1,45 @@
...
+  gemini: "gemini/latest",
+  azure: "azure/gpt-4o",
+  deepseek: "deepseek/deepseek-r1",
+  xai: "xai/grok-4",
+  // bedrock: omitted — no single canonical default model
+  // groq: omitted — no single canonical default model
+  // vertex_ai: omitted — no single canonical default model
+  // cerebras: omitted — no single canonical default model
+  // custom: omitted — user-defined endpoint, default is unknowable
+};
```
rogeriochaves, line 29: wait no, I don't think we should have hardcoded models at all? they quickly get outdated, look, azure/gpt-4o, that's crazy, so old, and using this openai/latest seems opaque, what does it resolves to? I prefer to surface this at the frontend and force user to chose one?

does openai/latest even work?

rogeriochaves (changes requested):

drewdrewthis:
**Orchardist triage, 2026-07-30** — @rogeriochaves your objection is the only unresolved thread here (8 total, 1 unresolved), and I went looking for whether it was a preference or a defect. It is a defect, and the codebase already says so.

Your comment on `modelProvider.constants.ts`:

> wait no, I don't think we should have hardcoded models at all? they quickly get outdated, look, azure/gpt-4o, …

**The repo already contains the mechanism that exists to prevent exactly this**, and its docstring states your objection almost verbatim. From `langwatch/src/server/modelProviders/getLatestFlagship.ts`:

> Used to derive a sensible "current model" **without hard-coding a value that drifts every time the registry advances**.

It exports `getLatestFlagshipForProvider(provider, mode)`, which walks the registry and returns the newest plain flagship for a provider, deliberately skipping specialised variants. Alongside it:

- `llmModels.json` — the model catalog, and per the overlay's own comment it is **regenerated periodically**
- `llmModels.overlay.json` — the sanctioned escape hatch for providers the upstream catalog does not carry, explicitly built so the regen task never clobbers hand-added entries
- `loadModelCatalog.ts` — merges the two at module load

So there is already a maintained catalog, a gap-filling overlay that survives regeneration, and a helper for picking a current model without hardcoding. `modelProvider.constants.ts` is new in this PR (it is not on `main` and nothing outside the PR imports it), and it reintroduces precisely the drift that design avoids.

**Concrete target for rework, so this is not an open-ended debate:** source defaults from `getLatestFlagshipForProvider` / the catalog, put any genuinely-missing provider models in `llmModels.overlay.json` where the regen task will preserve them, and drop the hardcoded constants file.

**Why I am posting rather than a lane:** the change is not mechanical, and I wanted the objection resolved as "here is the existing mechanism" instead of a judgement call about whether hardcoding is acceptable. Nothing has been changed on the PR and I am not marking your thread resolved — flagging that your concern has a ready answer in-tree.

drewdrewthis:
[grinder] Re-review requested — CI green, merge conflict resolved, review points addressed

## What changed since @rogeriochaves's CHANGES_REQUESTED

**1. Merge conflict resolved (`main` renamed `langwatch/` → `platform/app/`).**
The branch was `CONFLICTING` against `main` (a whole-tree rename of the app directory landed while this PR sat). Merged `origin/main` (no rebase, no force-push), moved the 4 files git couldn't auto-match to their new `platform/app/` locations, and re-wired `ProjectService`'s constructor — `modelProviderService` now slots in as the **third** param, after `lwqlKeyMap` which main added independently in the meantime (`platform/app/src/server/app-layer/projects/project.service.ts:177-192`). Merge commit `c563a1bab5`.

While validating the merge I found and fixed a casualty git couldn't have caught: `getResolvedDefaultModel` in `platform/app/src/server/api/routers/project.ts:227` was still using `checkProjectPermission`, a middleware `main` removed in favor of the `.permission()` decorator every sibling procedure in that router uses. Fixed in `4ad3f67f82`, verified with `pnpm run typecheck` (clean).

**2. Your unresolved thread — `modelProvider.constants.ts:29`, "should we even hardcode models, azure/gpt-4o is old, does openai/latest even work?"**

Checked both concerns against the model registry (`llmModels.json`) directly rather than guessing:

- `openai/latest` **does** work — it's a pre-existing mechanism (`latestAliases.ts`, not new in this PR), resolved at read-time by picking the highest `gpt-X.Y` entry in the registry for that mode/provider. Same for `anthropic/latest` and `gemini/latest`. Verified this is exercised, not just asserted, by the new registry-backed test below.
- **Your instinct on staleness was right, and it caught two real bugs**, not a style nit:
  - `xai/grok-4` **does not exist** in the registry at all — it would have resolved to a nonexistent model id the moment xai was the only enabled provider. Fixed to `xai/grok-4.3` (the current registry-verified non-multi-agent flagship).
  - `azure` has **zero** registry-backed entries, period — Azure customers name their own deployments, so there is no universal "azure/gpt-4o" the way there's a universal `gpt-5.5`. I removed `azure` from `PROVIDER_DEFAULT_MODELS` and `PROVIDER_RESOLUTION_ORDER` entirely (`modelProvider.constants.ts`), placing it alongside bedrock/groq/vertex_ai/cerebras/custom — which this file already excluded for the exact same reason ("no single canonical default model"). **This is a behavior change**: a project with only Azure enabled now gets `null` from `resolveDefaultModel` instead of a canonical-but-wrong `azure/gpt-4o`. Flagging this explicitly rather than treating it as a given — if you'd rather Azure stay in with some other resolution strategy (e.g. surfacing at the frontend, per your comment), say so and I'll adjust; I picked "drop it, consistent with the existing omitted-provider pattern" as the smallest fix that stops the wrong-default class of bug without redesigning the feature.
  - Added a regression test (`platform/app/src/server/modelProviders/__tests__/modelProvider.constants.unit.test.ts`) asserting every `PROVIDER_DEFAULT_MODELS` value is either a valid `isLatestAlias(...)` or a key actually present in `llmModels.json`. It fails on the old `xai/grok-4` value — this is the test that should have caught the original bug.

Your broader point — "I prefer to surface this at the frontend and force the user to choose" — is a real design question but bigger than this fix. `resolveDefaultModel` is explicitly the narrow, non-UI fallback for callers that haven't migrated to the full `modelProvider.getResolvedDefault` cascade (which already does the frontend-facing feature-key resolution `ScenarioCreateModal` uses, unchanged by this PR). I didn't take on redesigning that fallback to force a frontend choice — that's a product call outside this split's scope (F2 of #3526) and I don't think it's mine to make unilaterally. Let me know if you want it addressed here or as a follow-up issue.

**3. CI failing check.**
`check` (`deployment-impact-check`) was failing on a stale run from July against an older PR-body draft that lacked the `## Deployment Impact` heading — the body already has it now, the check just hadn't re-run since (no new commits since July until this session). Re-ran clean on the new commits.

`lint` failed after the merge push — Biome's `assist/source/organizeImports` flagged a real error (not a pre-existing warning) in the new regression test file's import order. Fixed in `7d5eb18eac`.

## Verified by:
```
pnpm vitest run src/server/app-layer/projects/__tests__/project.service.unit.test.ts src/server/modelProviders/__tests__/modelProvider.constants.unit.test.ts
→ Test Files  2 passed (2)
→ Tests  7 passed (7)

pnpm run typecheck
→ (clean, zero errors)

pnpm exec biome check ./src ./ee ./scripts ./e2e ./prisma ./vite --max-diagnostics=1000
→ Found 3374 warnings.   (no "Found N error(s)" line — 0 errors, same warning count as before my fix)

grep -rn "PROVIDER_DEFAULT_MODELS\|PROVIDER_RESOLUTION_ORDER\|resolveDefaultModel" across src/ee/scripts/e2e/prisma/vite
→ only this PR's own files reference these symbols — nothing else depends on the removed azure entry

gh pr checks 5398 --repo langwatch/langwatch
→ zero failing, zero pending (full breakdown: https://github.com/langwatch/langwatch/pull/5398/checks)

gh pr view 5398 --json mergeable,mergeStateStatus
→ {"mergeStateStatus":"BLOCKED","mergeable":"MERGEABLE"}   (BLOCKED = pending human review, not a conflict)
```

Not marking this ready-for-merge myself — `reviewDecision` is still `CHANGES_REQUESTED` from your review and that's yours to clear, especially given the Azure-removal behavior change above is a judgment call I'd like your sign-off on, not a unilateral one.

https://claude.ai/code/session_01KHE7KBmWBdq1jK8WnRNdJF

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `7d5eb18e` · **Run:** pr-review (reviewer: @review-clerk)

### rogeriochaves' open thread (unresolved on GitHub, addressed in substance at this head)
Thread: https://github.com/langwatch/langwatch/pull/5398#discussion_r3570580448 (`PRRT_kwDOKRXhvM6QYtXI`, `platform/app/src/server/modelProviders/modelProvider.constants.ts:29`) — still shows `isResolved: false` via GraphQL.

Rogerio's comment (2026-07-13) cited `azure/gpt-4o` as an example of a stale hardcoded default, and asked whether `openai/latest` even works.

At head `7d5eb18e`:
- `azure` is no longer in the map at all — it's a commented-out line: `// azure: omitted — no single canonical default model (deployment names are user-defined, not registry-backed)` (`modelProvider.constants.ts:24`). The specific example he flagged is gone.
- `openai/anthropic/gemini` use the `"latest"` alias form (`openai/latest`, `anthropic/latest`, `gemini/latest`), which does resolve — `platform/app/src/server/modelProviders/latestAliases.ts` implements `resolveLatestAlias`/`isLatestAlias` and is called at read-time to pick the current flagship (e.g. `openai/latest` → `openai/gpt-5.5` today). Only `deepseek` and `xai` use pinned concrete ids.
- This PR adds `modelProvider.constants.unit.test.ts`, which asserts every entry in `PROVIDER_DEFAULT_MODELS` is either a `"latest"` alias or a live key in the model registry — i.e. it structurally guards against exactly the "stale hardcoded id" failure Rogerio described. That test is green in CI.

So the concrete objection (the azure example) is moot and the "does it work" question is answered yes by existing infra plus a new regression test. The broader design question he raised — should the platform surface unresolved-default choice to the user at the frontend instead of silently picking one — is a product-direction question, not a code defect; it isn't resolved by any commit and there's no reply in the thread. I'm not treating it as blocking (no code claim it makes is false), but it stays open on GitHub for Drew/Rogerio to close explicitly.

### Non-blocking observations
- `presets.ts:456` wires `ModelProviderService.create(prisma)` into the primary `ProjectService`; a second construction site (`presets.ts:2020`, a distinct app instance) still omits it. `modelProviderService` is an optional 3rd constructor param, so `resolveDefaultModel` degrades to `null` there rather than throwing — correct per the documented contract, just worth being aware the new endpoint returns `null` unconditionally for whichever app instance that is. (Decide)

### CI (head `7d5eb18eac7f2538cfcaca6a04cf9e25a00c7ad1`)
All gating jobs SUCCESS or SKIPPED (no FAILURE, no PENDING): `build`, `typecheck`, `lint`, `test-unit (1-4)`, `test-component (1-3)`, `test-integration (shared 1-2, mocking 1-2)`, `e2e`, `e2e-complete`, `codeql-complete`, `langwatch-app-complete`, `openapi-completeness`, `ci-scripts-test`, `ast-grep`, `feature-parity`, `ci-self-test`, `secrets (gitleaks + trufflehog)`, `semgrep (changed lines)`, `readme-links`, `no-committed-screenshots`, `docs-complete`, and the language-specific `-complete` gates. `CodeRabbit` and `action-semantic-pull-request` are non-gating per brief and both SUCCESS anyway.

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved `<!-- review-thread -->` inline threads at this SHA. Not a merge-readiness signal._

drewdrewthis (commented):
Verdict: READY. See https://github.com/langwatch/langwatch/pull/5398#issuecomment-5565363517 for full findings.

platform/app/src/server/modelProviders/modelProvider.constants.ts
```diff
@@ -0,0 +1,45 @@
...
+  gemini: "gemini/latest",
+  azure: "azure/gpt-4o",
+  deepseek: "deepseek/deepseek-r1",
+  xai: "xai/grok-4",
+  // bedrock: omitted — no single canonical default model
+  // groq: omitted — no single canonical default model
+  // vertex_ai: omitted — no single canonical default model
+  // cerebras: omitted — no single canonical default model
+  // custom: omitted — user-defined endpoint, default is unknowable
+};
```
drewdrewthis: Fixed in c563a1bab5 (merge commit resolving the conflict from origin/main): azure/gpt-4o is removed entirely from PROVIDER_DEFAULT_MODELS — it is now just a comment ("azure: omitted — no single canonical default model (deployment names are user-defined, not registry-backed)"), same treatment as bedrock/groq/vertex_ai/cerebras/custom.

openai/latest (and anthropic/latest, gemini/latest) resolve at call-time via `isLatestAlias()` in latestAliases.ts, not a hardcoded pinned id — so they track upstream releases automatically instead of going stale like the old azure/gpt-4o did.

Regression test added at platform/app/src/server/modelProviders/__tests__/modelProvider.constants.unit.test.ts: it asserts every entry in PROVIDER_DEFAULT_MODELS is either a resolvable "latest" alias or a concrete id present in the registry catalog — this is exactly the check that would have caught the stale azure/gpt-4o pin, and would catch any future stale pin the same way. CI green: https://github.com/langwatch/langwatch/pull/5398/checks

PR #5400, opened 2026-07-06
fix(aigateway): use require.ErrorAs in dispatch_test (testifylint) (#5399)

### LINKED ISSUE #5399: lint landmine: testifylint violation on main (dispatch_test.go:448) fails unrelated PRs' merge-result lint
## Expected

`golangci-lint` (testifylint) passes on every PR whose merge result includes current `main` — code already on `main` should never surface as a "new" lint issue in unrelated PRs.

## Actual

`services/aigateway/app/dispatch_test.go:448` uses `require.True(t, errors.As(err, &ue))`, which **violates testifylint** (`error-is-as: use require.ErrorAs`). Main's own `go-ci` is green because golangci runs in only-new-issues mode there, but **any PR whose rebase base predates commit `3c2ed9acb` (#5358) gets this line flagged as "new" in its merge result and fails the `lint` check** — observed on PR #5009, which does not touch the file.

## Reproduction

1. Take any branch whose merge-base with `main` is older than `3c2ed9acb`.
2. Open/update a PR; the `lint` check runs golangci with only-new-issues on the merge result.
3. `lint` fails with `dispatch_test.go:448:2: error-is-as: use require.ErrorAs (testifylint)` even though the PR never touched the file.

---
*Classification: Bug*

drewdrewthis:
## Why

`require.True(t, errors.As(err, &ue))` at `services/aigateway/app/dispatch_test.go:448` violates the testifylint `error-is-as` rule. This caused PR #5009 lint to fail at HEAD `47b937eea` even after rebasing — the only-new-issues baseline does not absolve pre-existing violations when the merge result introduces them into the diff context.

The fix replaces it with `require.ErrorAs(t, err, &ue)`, which is the idiomatic testify assertion and satisfies the linter.

## Human verification

```
cd services/aigateway && go test ./app/ -run TestDispatch -count=1
```

Should pass with no failures.

Test output from `go test ./app/ -run TestDispatch -count=1` showing PASS.

Closes #5399

## Deployment Impact

None. Test-only change inside `services/aigateway` (a unit-test assertion swapped to `require.ErrorAs`); no runtime code, chart, env, or self-hosting surface is modified. The path merely lives under the `services/` tree, which is why this section is required.

### DIFF
--- a/services/aigateway/app/dispatch_test.go
+++ b/services/aigateway/app/dispatch_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"testing"
 
@@ -445,7 +444,7 @@ func TestHandleChat_TerminatesOnUpstream400(t *testing.T) {
 	require.Error(t, err)
 	assert.Equal(t, 1, callCount)
 	var ue *domain.UpstreamError
-	require.True(t, errors.As(err, &ue))
+	require.ErrorAs(t, err, &ue)
 	assert.Equal(t, 400, ue.StatusCode)
 }
 


drewdrewthis:
## Review verdict: READY

**Reviewed at:** `f6a9d1aac` · **Run:** pr-review (reviewer: @drew)

No blocking concerns. All review checks passed at this SHA.

**Checks performed:**

- `require.ErrorAs(t, err, &ue)` is semantically equivalent to the old `require.True(t, errors.As(err, &ue))`: both assert the error unwraps to `*domain.UpstreamError` and both call `t.FailNow()` on failure. The new form additionally populates `ue` on success (same as `errors.As`) and emits a structured testify failure message rather than a raw boolean — strictly better diagnostics.
- The `errors` import is genuinely unused after the change: grep of the post-change file returns zero matches for `errors.`.
- No other test in the file uses the `errors` package — the import block confirms only `bytes`, `context`, `encoding/json`, `fmt`, `testing`, and third-party deps remain.
- The lint-failure claim is coherent: testifylint's `error-is-as` rule flags `require.True(t, errors.As(...))` as a violation; replacing with `require.ErrorAs` is the prescribed fix and resolves the cross-PR only-new-issues baseline bleed.

PR #5403, opened 2026-07-07
feat: migrate app to zod 4

0xdeafcafe:
## What

Migrates the **langwatch app** from zod 3 to **zod 4**. 513 typecheck errors → **0**; unit suite clean except one pre-existing, zod-unrelated failure (see below).

> **Stacked on #5392** (single pnpm workspace). Base is `worktree-feat+single-pnpm-workspace`, so this diff is *only* the zod-4 work. Merge #5392 first.

## Why it was bigger than a bump

Two dependencies are locked to zod 3 with no zod-4 release, and both broke when the app moved to zod 4:

- **`@ag-ui/core`** — its zod-3 schemas were composed with app schemas in `event-schemas.ts`, creating two zod instances at runtime.
- **`@langwatch/scenario`** (vendored 0.4.10) — built for the zod-3.25 *transition* release (imports both `"zod"` and `"zod/v4"`); the hoisted zod 4 leaked into `"zod/v4"` and crashed `.merge()` at **import time**, taking down the whole scenarios-execution subsystem. Typecheck couldn't catch this — the unit run did.

Both are resolved by pinning each to the app's single zod-4 instance via `pnpm-workspace.yaml` overrides (their runtime code is zod-4-compatible; verified they load + parse). They ship zod-3-shaped `.d.ts`, so the app casts at those boundaries (type-only — one runtime instance). Drop the overrides when they publish zod-4 releases.

## The recurring zod-4 fixes

- **`z.record(enum, V)` is now exhaustive** (requires every key) → `z.partialRecord` for filter maps (analytics, trigger filters, misc routes).
- **96 single-arg `z.record(V)` → `z.record(z.string(), V)`** (codemod).
- `ZodError.errors` → `.issues`; `z.SafeParseReturnType` → `z.ZodSafeParseResult`; `z.ZodType<X, ZodTypeDef, any>` → `z.ZodType<X, any>` (2nd param is Input now); `.refine(fn, fn)` → `.refine(fn, { error })`.
- Internals: `._def` → `.def`, `.create()` factories → `z.x()`, removed `UnknownKeysParam`/`ZodTupleItems`; `EventTypeOf` rewritten for the new `ZodType` arity.
- **`zod-openapi` v5/v6 dropped `.openapi()`/`/extend`** → a self-contained typed no-op shim on `ZodType.prototype` (preserves the prior Next.js no-op behaviour without rewriting 30+ call sites).
- Deps: `zod ^4.4`, `zod-openapi ^6`, `@hono/zod-validator ^0.8`, `zod-validation-error ^5` (`fromZodError` still exported — no rename).

## Verification

- ✅ **`pnpm typecheck`: 0 errors** (from 513).
- ✅ **Unit suite**: passes except **one pre-existing failure** — `aggregation-builder.test.ts:252` expects `quantileExact(0.9)` but the code emits `quantileTDigest(0.9)`; two sibling tests (`memory-safety`, `metric-translator`) already expect `quantileTDigest`, so line 252 is a stale outlier. This file/logic is **untouched by this PR** (likely tied to the active `ch-skipping-indexes-quantile-dicts` work). Left as-is.
- ✅ Scenarios-execution cluster: **431 tests pass**.
- Runtime validation behaviour (error formats, coercion, DU parsing) exercised by the unit suite; integration suites (need CH/PG/Redis) still to run in CI.

## Note for #5392 (the workspace PR)

The single-workspace consolidation **duplicated `ioredis` (5.10.1 + 5.11.1)**, which breaks BullMQ's `Worker` `connection` type. Worked around here with `as ConnectionOptions` casts, but the dup almost certainly also fails **#5392's** typecheck — worth a dedupe override (`ioredis`) on that PR.

🤖 Draft — opened by a background agent.


0xdeafcafe:
Closing — decision is to keep the app on zod 3 for now and let other workspace packages (SDK, mcp-server) run zod 4, which the single-workspace PR (#5392) already does (app declares zod ^3, SDK zod ^4; they coexist via link-workspace-packages=false, no global zod override). The full app→zod-4 migration here required pinning zod-3-locked deps (@ag-ui/core, @langwatch/scenario) onto zod 4 via overrides, which we'd rather not carry until those ship native zod-4 releases. Branch `feat/zod-4-migration` is preserved if we revisit.

PR #5447, opened 2026-07-07
chore(deps): batch small dependency updates in /langwatch + /mcp-server

0xdeafcafe:
Bundles the patch/minor Dependabot bumps for the **langwatch workspace** into one PR. Each individually fails frozen-install on a stale `langwatch/pnpm-lock.yaml` and they all serialize on that lockfile, so batching them regenerates the lockfile once and validates them together.

**Validated:** `pnpm typecheck` (tsgo) clean across langwatch; mcp-server `tsup` build (ESM+DTS) passes; `--frozen-lockfile` install passes in both workspaces. No Node version changes.

### langwatch
react/react-dom/react-is → 19.2.7 (lockstep) · @chakra-ui/react 3.36.0 · @tanstack/react-virtual 3.14.4 · @xyflow/react 12.11.1 · @clickhouse/client 1.22.0 · @aws-sdk/client-s3 + s3-request-presigner 3.1075.0 · @google-cloud/dlp 6.7.0 · bullmq 5.79.2 · react-hook-form 7.80.0 · react-error-boundary 6.1.2 · libphonenumber-js 1.13.7 · date-fns 4.4.0 · liquidjs 10.27.1 · semver 7.8.5 · tsx 4.22.4 · **selfsigned 5.5.0** (v5 async `generate()` — start.ts + vite.config.ts migrated) · **testcontainers 12.0.3** · webpack 5.108.1 · @vitest/ui + coverage-v8 4.1.9

### mcp-server
@types/debug 4.1.13 · @anthropic-ai/claude-code 2.1.195

Supersedes the individual Dependabot PRs (to be closed on merge): #5444 #5442 #5441 #5440 #5438 #5437 #5436 #5435 #5434 #5431 #5430 #5429 #5428 #5425 #5423 #5422 #5421 #5417 #5413 #5409 #4618 #5061 #4595

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development HTTPS certificate handling for better compatibility with modern browser and Apple certificate requirements.
  * Updated Vite dev startup to wait for certificates before loading, reducing configuration-related startup issues.

* **Chores**
  * Bumped several app and tooling dependencies, including React-related packages, UI utilities, test tooling, and cloud SDKs.
  * Updated package overrides and development dependencies to newer versions.

PR #5450, opened 2026-07-07
chore(deps): batch safe major dependency updates in /langwatch

0xdeafcafe:
Follow-up to the small-updates batch (#5447) — the **major** bumps that validate as safe: `pnpm typecheck` (tsgo) clean and a production `vite build` passes (shiki lazy-chunking intact, posthog bundles). No Node version changes.

- **concurrently** (dev) 9.2.1 → 10.0.3
- **cookie** 1.1.1 → 2.0.0
- **posthog-node** 4.18.0 → 5.38.8
- **shiki** 3.23.0 → 4.3.0

**Dropped ra-data-simple-prisma 3→7** — its v7 react-admin stack pulls `react-hotkeys-hook@5.3.3` (an ESM file shipped in a CJS package) which breaks the vitest `test-unit` that transitively imports the admin router. The migration lives on #5059 if we want to take it on with the vitest inline workaround.

Supersedes: #5418 #5419 #5433 #5416

## Summary by CodeRabbit

* **Chores**
  * Updated runtime and development dependency versions, including cookie, posthog-node, shiki, and concurrently.


Aryansharma28 (dismissed):
lgtm

Issue #5455, opened 2026-07-07
Sweep follow-up: 3 memory/query anti-patterns from #5454 recur in 6 other features


sergioestebance:
**Describe the bug**

While fixing #5454 (Prompt Playground tabs never unmounting), a codebase sweep for the same 3 anti-patterns found further occurrences outside the Prompt Playground. #5454's PR fixed the in-scope instances (including one in the same feature area, `PublishedPromptActions.tsx`); this issue tracks the rest so they get individual review rather than ballooning that PR.

**Pattern 1 — Ark UI `Tabs.Root` without `lazyMount unmountOnExit`**

Every inactive tab's content tree (including any live queries inside it) stays mounted-but-hidden instead of unmounting. Same root cause as #5454.

- `src/components/LLMMetrics.tsx:173` — 3 tabs, each wraps a `CustomGraph` firing its own `api.analytics.getTimeseries` query; all 3 fire and stay subscribed regardless of active tab.
- `src/components/analytics/UserMetrics.tsx:125` — identical shape (3 `CustomGraph`-backed tabs).
- `src/pages/settings/governance/tool-catalog.tsx:65` — 2 tabs, 5+ queries fire on page load instead of gating to the active tab.
- `src/components/traces/TraceDetails.tsx:216` — trace drawer has 6 tabs; `messages`, `guardrails`, `evaluations`, `events` (lines ~400-453) are unguarded. The `messages` tab's `Conversation` component fires `api.traces.getTracesByThreadId` and stays subscribed even when the drawer is on a different tab.
- `src/components/ops/foundry/PlaygroundContent.tsx:45` — 4 tabs (Editor hosting `SpanEditorPanel`, Waterfall, Graph, JSON), none guarded, all mount simultaneously. Internal dev/ops "foundry" tool, so lower user-facing impact.
- `src/components/experiments/DSPyExperiment.tsx:632` — 3 tabs; **partial fix already present** — tab 1 (Evaluations) is manually gated with `tabIndex === 1 &&` (line 856), but tab 0 (Predictors) and tab 2 (LLM Calls, a `.map()`-driven table) are unguarded. Fix by extending the guard to tabs 0/2 or adding `lazyMount unmountOnExit` to the `Tabs.Root`.

**Pattern 2 — query inside a Popover/Menu not gated on the disclosure's own `open` state**

Query fires as soon as the component mounts instead of when the user opens the popover/menu. Same root cause as #5454's `VersionHistoryListPopover` fix.

[...]

PR #5456, opened 2026-07-07
fix(prompts): stop prompt playground tabs from leaking memory

### LINKED ISSUE #5454: Prompt playground: open tabs never unmount, causing memory growth and sluggishness with many tabs
**Describe the bug**

The Prompt Playground (`/[project]/prompts`) lets you open multiple prompts as tabs. Reported symptom: memory usage climbs and the page gets sluggish as more tabs are opened, independent of which tab is active.

Root cause investigation (read-only, no fix yet) found this is a real architectural bug, not just expected per-editor overhead.

**Root cause**

`src/prompts/prompt-playground/components/prompt-browser/ui/DraggableTabsBrowser.tsx:259` uses Ark UI's `Tabs.Root` with no `lazyMount`/`unmountOnExit`. Ark UI's default behavior mounts every tab panel and just toggles a `hidden` attribute on inactive ones — it never actually unmounts them. The codebase already uses `lazyMount`/`unmountOnExit` correctly in several other places (popovers, menus); it just wasn't applied to the prompt tab container.

Because every open tab stays permanently mounted, each one keeps running, forever, regardless of visibility:

1. A full `react-hook-form` instance with live `watch` subscriptions and timers (`usePromptConfigForm`)
2. A `RichTextarea` instance per message in the prompt (`PromptMessagesEditor` → `PromptTextAreaWithVariables`)
3. A full CopilotKit chat provider/context tree (`PromptPlaygroundChat.tsx`)
4. An unconditional version-history fetch — `VersionHistoryListPopover.tsx:366-372` calls `getAllVersionsForPrompt` (full `configData` for every historical version, not lightweight metadata) as soon as the tab mounts, not gated on the history popover actually being opened
5. Two more full-prompt fetches per tab, fired from the tab-bar label itself: `useHasUnsavedChanges` and `useLatestPromptVersion` (each call `getByIdOrHandle`)

**Why it compounds with tab count instead of staying constant**

- Since tabs never unmount, React Query's defaults (`staleTime: 0`, `refetchOnWindowFocus: true`) mean every browser-window focus event re-fires all of the above queries for every open tab simultaneously — a burst that scales linearly with tab count.
- The Zustand store backing the tabs is wrapped in `persist` (`DraggableTabsBrowserStore.ts`), and its `partialize` rebuilds and `JSON.stringify`s **all windows/tabs' full content** into `localStorage` on every 500ms-debounced keystroke in any single tab. So typing in one tab does O(total content across all open tabs) main-thread work — this explains the "sluggish," not just "uses more RAM," part of the report.

Tab-close cleanup itself was checked and is fine — closing a tab correctly tears down its store entry and component subtree; no lingering Map/leak there.

**Expected behavior**

Only the active tab's editor, form, chat, and queries should be mounted/subscribed at a time. Inactive tabs should keep their lightweight tab-bar label (title, unsaved indicator) but not their full content tree. Version history should only fetch when its popover is opened.

**Suggested fix (not yet implemented)**

- Add `lazyMount unmountOnExit` (or equivalent conditional rendering keyed on `activeTabId`) to the `Tabs.Root` in `DraggableTabsBrowser.tsx`.
- Gate `VersionHistoryListPopover`'s `getAllVersionsForPrompt` query on the popover's `open` state.

**Environment**

- Component: langwatch (Next.js app)
- Affected page: `/[project]/prompts` (Prompt Playground, multi-tab mode)

**Additional context**

Filed from an investigation-only session; no code changes made yet. Follow-up work will implement the fix above.

sergioestebance:
# Related Issue

- Resolves #5454

## Why

Closes #5454

Users with many prompt tabs open in the Prompt Playground reported memory climbing and the page getting sluggish, independent of which tab was active. Root cause: Ark UI's `Tabs.Root` defaults to mounting every tab panel and only toggling a `hidden` attribute on inactive ones — it never actually unmounts them. So every open tab kept a full form, editor, chat provider, and set of live queries running forever, regardless of visibility.

## What changed

- **Inactive tabs actually unmount.** Added `lazyMount`/`unmountOnExit` to the outer prompt-tab `Tabs.Root` so only the active tab's content is mounted. The nested sub-tabs (Conversation/Variables/Demonstrations/Parameters) use `lazyMount` **without** `unmountOnExit` — switching sub-tabs must not unmount the chat (that would abort an in-flight generation). Chat messages and form values round-trip through the Zustand tab store, so switching prompt tabs doesn't lose saved work.
- **Version history no longer fetches eagerly.** `VersionHistoryListPopover` was firing its full-version-history query on every tab mount, whether or not the user ever opened the history dropdown. Gated the query on the popover's own open state.
- **Sidebar per-row menu, same bug, N+1 variant.** A codebase sweep for the same "query fires on mount instead of on open" pattern found `PublishedPromptActions` (rendered once per prompt row in the sidebar) doing the same thing — fixed the same way.
- **localStorage writes no longer re-serialize every open tab on one tab's edit.** The Zustand store's `persist` middleware was rebuilding and stringifying every open tab's full content into a single localStorage key on every debounced edit in any one tab. Split persistence so each tab's data lives under its own `${projectId}:tab:${tabId}` key — editing one tab now only rewrites that tab's key.
- **Orphaned-key cleanup hardened.** Code review on the per-tab-key split caught a real gap: cleanup could only find per-tab keys by parsing the light index key, so a corrupted or partially-written index (e.g. a `QuotaExceededError` on the index write, after per-tab keys already landed) would strand data with no path to clean it up. Cleanup now scans localStorage directly for the project's key prefix instead of depending on the index.
- **No lost edits on tab switch.** Both debounced writers (prompt form via lodash, variable values via `useDebounceCallback`, which cancels pending work on unmount) now flush on unmount, so an edit typed just before switching tabs is persisted rather than dropped.
- **Legacy persistence migrates, doesn't wipe.** Existing users store tab data in the old single-key format; `getItem` now falls back to that embedded data when a per-tab key is missing and migrates it to a per-tab key on the next write, instead of dropping every tab and clearing the playground on first reload after deploy.
- **Delete stays disabled until its permission check resolves.** With the sidebar menu's permission query now gated on open, `canDelete` defaults to disabled during the brief loading window instead of enabling the destructive action before the check returns.

A broader sweep found the same three anti-patterns recurring in 6 unrelated features (analytics graphs, governance tool catalog, trace viewer, secrets popover, agent menu) plus 3 structurally-similar undo-history stores. Those are out of scope for this PR and tracked separately in #5455.

## Test plan

- New regression test proving the actual bug: `DraggableTabsBrowser.inactiveTabUnmount.test.tsx` — fails against the pre-fix code (content present but `hidden`), passes after.
- `VersionHistoryListPopover.test.tsx` — added cases asserting the query is disabled while closed and enabled once opened.
- `PublishedPromptActions.integration.test.tsx` (new) — same assertion shape for the sidebar menu.
- `DraggableTabsBrowserStore.test.ts` — added cases proving a tab edit doesn't rewrite other tabs' localStorage content, and that per-tab keys are fully cleaned up even when the index is corrupted or unparseable.
- `pnpm typecheck` clean; 67 unit tests + 4 integration tests (Docker-backed) pass across every touched file, plus direct consumers (`PromptTabbedSection`, `PromptPlaygroundChat`, `PromptBrowserWindowContent`) with no regressions.

## Anything surprising?

Ran a self-review pass on the persistence redesign before opening this PR and caught a real P1 (the orphaned-key cleanup gap above) plus a couple of test-quality issues (a test-isolation risk from an unrestored mock, a test named for a scenario it didn't actually assert, and a misnamed `.test.tsx` that should've been `.integration.test.tsx`) — all fixed and covered by tests in this PR, not left as follow-ups.

**Behavioral tradeoff (intentional):** because inactive prompt tabs now fully unmount, switching to a different prompt tab while a chat generation is streaming interrupts that background generation — the completed messages persist, but an in-flight turn does not resume on return. This is inherent to the memory fix and considered an acceptable cost (you rarely switch prompt tabs mid-generation). Sub-tab switching within a tab does **not** interrupt the chat — the nested tabs deliberately keep it mounted.

## Summary by CodeRabbit

* **New Features**
  * Tab panels now lazy mount across tab sections and the draggable tab browser.
  * Version history and published prompt actions now load only while the related popover/menu is open.
  * Draggable tab browser persistence is now saved per tab with automatic cleanup of stale storage keys.
* **Bug Fixes**
  * Inactive tab panels are fully unmounted.
  * Pending tab edits are flushed on unmount to prevent losing the latest variable/form changes.
  * Delete permission is now disabled until permission is known.
* **Tests**
  * Added/expanded regression and integration coverage for conditional query enabling and per-tab persistence/rehydration/reset behavior.

### DIFF
--- a/langwatch/src/prompts/__tests__/VersionHistoryListPopover.test.tsx
+++ b/langwatch/src/prompts/__tests__/VersionHistoryListPopover.test.tsx
@@ -47,14 +47,15 @@ const mockVersions = [
   },
 ] as unknown as VersionedPrompt[];
 
+const { mockUseQuery } = vi.hoisted(() => ({
+  mockUseQuery: vi.fn(),
+}));
+
 vi.mock("~/utils/api", () => ({
   api: {
     prompts: {
       getAllVersionsForPrompt: {
-        useQuery: () => ({
-          data: mockVersions,
-          isLoading: false,
-        }),
+        useQuery: mockUseQuery,
       },
     },
   },
@@ -79,6 +80,10 @@ const renderWithChakra = (ui: React.ReactElement) => {
 describe("VersionHistoryListPopover", () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    mockUseQuery.mockReturnValue({
+      data: mockVersions,
+      isLoading: false,
+    });
   });
 
   afterEach(() => {
@@ -271,6 +276,33 @@ describe("VersionHistoryListPopover", () => {
       // Should show "current" tag - only one should exist (for v2)
       const currentTags = screen.getAllByText("current");
       expect(currentTags).toHaveLength(1);
+    });
+  });
+
+  describe("when the popover is closed", () => {
+    it("does not enable the version history query", () => {
+      renderWithChakra(<VersionHistoryListPopover configId="config-1" />);
+
+      expect(mockUseQuery).toHaveBeenCalledWith(
+        expect.objectContaining({ idOrHandle: "config-1" }),
+        expect.objectContaining({ enabled: false }),
+      );
+    });
+  });
+
+  describe("when the popover is opened", () => {
+    it("enables the version history query", async () => {
+      renderWithChakra(<VersionHistoryListPopover configId="config-1" />);
+
+      const historyButton = screen.getAllByTestId("version-history-button")[0]!;
+      fireEvent.click(historyButton);
+
+      await waitFor(() => {
+        expect(mockUseQuery).toHaveBeenLastCalledWith(
+          expect.objectContaining({ idOrHandle: "config-1" }),
+          expect.objectContaining({ enabled: true }),
+        );
+      });
     });
   });
 });
--- a/langwatch/src/prompts/prompt-playground/components/prompt-browser/__tests__/DraggableTabsBrowser.inactiveTabUnmount.test.tsx
+++ b/langwatch/src/prompts/prompt-playground/components/prompt-browser/__tests__/DraggableTabsBrowser.inactiveTabUnmount.test.tsx
@@ -1,0 +1,71 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Regression test for GitHub issue #5454: inactive tabs in the Prompt
+ * Playground stay mounted in the DOM instead of unmounting.
+ *
+ * DraggableTabsBrowser renders a Chakra v3 `Tabs.Root` (Ark UI under the
+ * hood) with no `lazyMount`/`unmountOnExit` props, so Ark UI mounts every
+ * `Tabs.Content` panel and only toggles the `hidden` attribute on inactive
+ * ones instead of unmounting them.
+ */
+import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
+import { cleanup, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { DraggableTabsBrowser } from "../ui/DraggableTabsBrowser";
+
+describe("DraggableTabsBrowser", () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe("when a window has multiple tabs and one is inactive", () => {
+    function renderTwoTabWindow() {
+      return render(
+        <ChakraProvider value={defaultSystem}>
+          <DraggableTabsBrowser.Root onTabMove={vi.fn()}>
+            <DraggableTabsBrowser.Group
+              groupId="window-1"
+              activeTabId="tab-1"
+              onTabChange={vi.fn()}
+            >
+              <DraggableTabsBrowser.TabBar tabIds={["tab-1", "tab-2"]}>
+                <DraggableTabsBrowser.Tab id="tab-1">
+                  <DraggableTabsBrowser.Trigger value="tab-1">
+                    Tab 1
+                  </DraggableTabsBrowser.Trigger>
+                </DraggableTabsBrowser.Tab>
+                <DraggableTabsBrowser.Tab id="tab-2">
+                  <DraggableTabsBrowser.Trigger value="tab-2">
+                    Tab 2
+                  </DraggableTabsBrowser.Trigger>
+                </DraggableTabsBrowser.Tab>
+              </DraggableTabsBrowser.TabBar>
+              <DraggableTabsBrowser.Content value="tab-1">
+                <div data-testid="tab-1-content">Tab 1 content</div>
+              </DraggableTabsBrowser.Content>
+              <DraggableTabsBrowser.Content value="tab-2">
+                <div data-testid="tab-2-content">Tab 2 content</div>
+              </DraggableTabsBrowser.Content>
+            </DraggableTabsBrowser.Group>
+          </DraggableTabsBrowser.Root>
+        </ChakraProvider>,
+      );
+    }
+
+    it("keeps the active tab's content mounted", () => {
+      renderTwoTabWindow();
+
+      expect(screen.getByTestId("tab-1-content")).toBeInTheDocument();
+    });
+
+    it("unmounts the inactive tab's content instead of just hiding it", () => {
+      renderTwoTabWindow();
+
+      // Content for the inactive tab ("tab-2") must not exist in the DOM at
+      // all. Ark UI's default behavior only sets a `hidden` attribute on the
+      // panel, which keeps the node present here and fails this assertion.
+      expect(screen.queryByTestId("tab-2-content")).not.toBeInTheDocument();
+    });
+  });
+});
--- a/langwatch/src/prompts/prompt-playground/components/prompt-browser/__tests__/DraggableTabsBrowser.inactiveTabUnmount.test.tsx
+++ b/langwatch/src/prompts/prompt-playground/components/prompt-browser/__tests__/DraggableTabsBrowser.inactiveTabUnmount.test.tsx
@@ -2,12 +2,13 @@
  * @vitest-environment jsdom
  *
  * Regression test for GitHub issue #5454: inactive tabs in the Prompt
- * Playground stay mounted in the DOM instead of unmounting.
+ * Playground stayed mounted in the DOM instead of unmounting.
  *
- * DraggableTabsBrowser renders a Chakra v3 `Tabs.Root` (Ark UI under the
- * hood) with no `lazyMount`/`unmountOnExit` props, so Ark UI mounts every
- * `Tabs.Content` panel and only toggles the `hidden` attribute on inactive
- * ones instead of unmounting them.
+ * Before the fix, DraggableTabsBrowser rendered a Chakra v3 `Tabs.Root`
+ * (Ark UI under the hood) with no `lazyMount`/`unmountOnExit` props, so Ark
+ * UI mounted every `Tabs.Content` panel and only toggled the `hidden`
+ * attribute on inactive ones instead of unmounting them. This test guards
+ * against that regression now that both props are set.
  */
 import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
 import { cleanup, render, screen } from "@testing-library/react";
--- a/langwatch/src/prompts/prompt-playground/components/prompt-browser/prompt-browser-window/__tests__/PromptTabbedSection.test.tsx
+++ b/langwatch/src/prompts/prompt-playground/components/prompt-browser/prompt-browser-window/__tests__/PromptTabbedSection.test.tsx
@@ -647,4 +647,41 @@ describe("PromptTabbedSection Layout Modes", () => {
       // (afterEach restores tabIdRef even if the assertion above throws)
     });
   });
+
+  describe("when a tab is switched away from and reopened", () => {
+    it("restores the variable value the user had typed", async () => {
+      const user = userEvent.setup();
+      const store = getStoreForTesting(TEST_PROJECT_ID);
+      const tabId = store.getState().windows[0]?.tabs[0]?.id;
+      tabIdRef.current = tabId!;
+
+      const formValues = {
+        version: {
+          parameters: {},
+          configData: {
+            inputs: [{ identifier: "topic", type: "str" }],
+            demonstrations: { inline: { records: {} } },
+          },
+        } as any,
+      };
+
+      // Mount, type a value, then unmount — this is "switch away".
+      const first = renderPromptTabbedSection({ layoutMode: "vertical" }, formValues);
+      await user.click(screen.getByRole("tab", { name: /variables/i }));
+      const emptyInput = (await screen.findAllByRole("textbox")).find(
+        (el) => (el as HTMLInputElement).value === "",
+      );
+      await user.type(emptyInput!, "kept");
+      first.unmount();
+
+      // "Switch back": a fresh mount of the same tab must show the value again,
+      // proving the full round-trip (flush on unmount -> restore from store).
+      renderPromptTabbedSection({ layoutMode: "vertical" }, formValues);
+      await user.click(screen.getByRole("tab", { name: /variables/i }));
+      const restored = (await screen.findAllByRole("textbox")).find(
+        (el) => (el as HTMLInputElement).value === "kept",
+      );
+      expect(restored).toBeDefined();
+    });
+  });
 });
--- a/langwatch/src/prompts/prompt-playground/components/prompt-browser/prompt-browser-window/__tests__/PromptTabbedSection.test.tsx
+++ b/langwatch/src/prompts/prompt-playground/components/prompt-browser/prompt-browser-window/__tests__/PromptTabbedSection.test.tsx
@@ -398,9 +398,13 @@ describe("PromptTabbedSection Store Integration", () => {
   });
 });
 
-// Mock TabIdContext
+// Mock TabIdContext. tabIdRef lets a test point the component at a real store
+// tab id (defaults to a fixed value for tests that don't touch the store).
+const { tabIdRef } = vi.hoisted(() => ({
+  tabIdRef: { current: "test-tab-id" },
+}));
 vi.mock("../../ui/TabContext", () => ({
-  useTabId: () => "test-tab-id",
+  useTabId: () => tabIdRef.current,
 }));
 
 // Mock CopilotKit
@@ -597,6 +601,48 @@ describe("PromptTabbedSection Layout Modes", () => {
       expect(
         screen.getByText(/variables are substituted into the prompt/i),
       ).toBeInTheDocument();
+    });
+  });
+
+  describe("when a variable value is edited and the tab unmounts immediately", () => {
+    it("flushes the pending write so the edit is not lost", async () => {
+      const user = userEvent.setup();
+      const store = getStoreForTesting(TEST_PROJECT_ID);
+      const tabId = store.getState().windows[0]?.tabs[0]?.id;
+      // Point the component's useTabId() at the real store tab so its writes land.
+      tabIdRef.current = tabId!;
+
+      const { unmount } = renderPromptTabbedSection(
+        { layoutMode: "vertical" },
+        {
+          version: {
+            parameters: {},
+            configData: {
+              inputs: [{ identifier: "topic", type: "str" }],
+              demonstrations: { inline: { records: {} } },
+            },
+          } as any,
+        },
+      );
+
+      await user.click(screen.getByRole("tab", { name: /variables/i }));
+      const textboxes = await screen.findAllByRole("textbox");
+      const valueInput = textboxes.find(
+        (el) => (el as HTMLInputElement).value === "",
+      );
+      expect(valueInput).toBeDefined();
+      await user.type(valueInput!, "flushed");
+
+      // The 300ms debounce has NOT fired yet (test is faster). Unmounting the
+      // tab (as switching prompt tabs does) must flush the pending write rather
+      // than cancel it — otherwise the edit is lost.
+      unmount();
+
+      expect(store.getState().getByTabId(tabId!)?.variableValues.topic).toBe(
+        "flushed",
+      );
+
+      tabIdRef.current = "test-tab-id";
     });
   });
 });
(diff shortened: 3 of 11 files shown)


langwatch/src/prompts/prompt-playground/components/prompt-browser/__tests__/DraggableTabsBrowser.inactiveTabUnmount.test.tsx
```diff
@@ -0,0 +1,71 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Regression test for GitHub issue #5454: inactive tabs in the Prompt
+ * Playground stay mounted in the DOM instead of unmounting.
+ *
+ * DraggableTabsBrowser renders a Chakra v3 `Tabs.Root` (Ark UI under the
+ * hood) with no `lazyMount`/`unmountOnExit` props, so Ark UI mounts every
+ * `Tabs.Content` panel and only toggles the `hidden` attribute on inactive
+ * ones instead of unmounting them.
```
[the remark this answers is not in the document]
sergioestebance: Fixed in 9591640c3 — reworded the docblock to describe the pre-fix behavior in past tense, since this PR now adds lazyMount/unmountOnExit.

langwatch/src/prompts/prompt-playground/components/sidebar/__tests__/PublishedPromptActions.integration.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+      const trigger = screen.getByRole("button");
+      await user.click(trigger);
+
+      expect(mockCheckModifyPermission).toHaveBeenLastCalledWith(
+        expect.objectContaining({ idOrHandle: "prompt-1" }),
+        expect.objectContaining({ enabled: true }),
+      );
+    });
+  });
```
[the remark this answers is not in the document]
sergioestebance: Fixed in 9591640c3 — nested the when-blocks under a given() layer. This file is brand new so there was no existing convention to conflict with.

langwatch/src/prompts/__tests__/VersionHistoryListPopover.test.tsx
```diff
@@ -273,4 +278,31 @@ describe("VersionHistoryListPopover", () => {
...
+      fireEvent.click(historyButton);
+
+      await waitFor(() => {
+        expect(mockUseQuery).toHaveBeenLastCalledWith(
+          expect.objectContaining({ idOrHandle: "config-1" }),
+          expect.objectContaining({ enabled: true }),
+        );
+      });
+    });
+  });
```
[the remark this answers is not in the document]
sergioestebance: Skipping — this files two pre-existing describe blocks (when loading a previous version, currentVersionId prop) already lack a given() wrapper, so nesting only the 2 new blocks would make this one file internally inconsistent. Fixing all of them is a pre-existing-code refactor beyond this PRs scope.

langwatch-agent:

**Reviewer:** @drewdrewthis · **Mode:** `Targeted Review` · **Issue:** resolves #5454 (tabs never unmount, memory climbs). Scope matches the issue.

**Scariest risk:** the fix unmounts inactive tab panels, which destroys their live state (form edits, editor, chat, queries). If unsaved edits or an in-flight generation are not persisted before unmount, switching tabs could silently lose user work.

**Top inspection targets**
- `Must Check` — `DraggableTabsBrowser.tsx` / `PromptTabbedSection.tsx`: is unsaved form/editor state flushed to the store before an inactive tab unmounts, and restored on return?
- `Must Check` — `DraggableTabsBrowser.inactiveTabUnmount.test.tsx`: does it assert state survives switch-away-and-back, not just that the panel unmounts?
- `Ask Author` — what happens to an in-flight chat/generation in a tab you switch away from: aborted cleanly, or orphaned?

+612/-54, 10 files, CI green, no unresolved threads.

langwatch-agent (commented):
Superseded by actual review; temporary anchor probe with no remaining inline comment.

langwatch-agent (commented):
Superseded by actual review; temporary anchor probe with no remaining inline comment.

langwatch/src/prompts/prompt-playground/prompt-playground-store/DraggableTabsBrowserStore.ts
```diff
@@ -151,6 +152,186 @@ const PersistedStateSchema = z.object({
...
+        };
+
+        const windows: Window[] = parsed.state.windows.map((w) => ({
+          id: w.id,
+          activeTabId: w.activeTabId,
+          tabs: w.tabs.map((t) => {
+            const tabRaw = localStorage.getItem(
+              getTabStorageKey(projectId, t.id),
+            );
+            const data = tabRaw
```
langwatch-agent: **P1** — This reader has no migration path for the existing persisted format. Current users already have the old `createJSONStorage` payload under `${projectId}:draggable-tabs-browser-store`, where each tab still has `t.data` embedded in the index. After this PR, `getItem` ignores that embedded `t.data` and only reads `${projectId}:tab:${t.id}`; those keys do not exist yet for old sessions, so every tab rehydrates with `data: undefined`, fails `PersistedStateSchema`, and `onRehydrateStorage` clears the whole prompt playground store. Failure scenario: a user deploys this with several unsaved prompt tabs in localStorage, reloads the playground, and all saved tab state is deleted. Please fall back to legacy `t.data` when the per-tab key is missing, write it into the new per-tab key on the next persist, and add a regression test that seeds the old single-key payload and proves tabs survive the migration.

langwatch/src/prompts/prompt-playground/components/prompt-browser/ui/DraggableTabsBrowser.tsx
```diff
@@ -265,6 +265,8 @@ function DraggableTabsGroup({
           display="flex"
           flexDirection="column"
           variant="enclosed"
+          lazyMount
+          unmountOnExit
```
langwatch-agent, line 269: **P1** — Unmounting inactive prompt tabs can now drop very recent edits that have not reached the tab store yet. The prompt form sync in `PromptBrowserWindowContent` is debounced by 500ms, and variable values in `PromptTabbedSection` are debounced by 300ms via `useDebounceCallback`, which cancels pending work on unmount in the installed `usehooks-ts` implementation. Failure scenario: type into a Variables value or prompt field and immediately switch to another prompt tab; this `unmountOnExit` tears down the active tab before the debounced write flushes, so switching back restores stale `tab.data`. Please flush pending form/variable writes before tab unmount/tab change, or persist synchronously on blur/tab change, and add a regression that types then switches tabs immediately and verifies the edit survives.

langwatch/src/prompts/prompt-playground/components/sidebar/PublishedPromptActions.tsx
```diff
@@ -97,7 +98,7 @@ export function PublishedPromptActions({
       projectId: project?.id ?? "",
     },
     {
-      enabled: !!project?.id,
+      enabled: open && !!project?.id,
```
langwatch-agent, line 101: **P2** — Gating `checkModifyPermission` on `open` leaves `permission` undefined on the first menu render, but `canDelete` below treats unknown permission as allowed. The Delete item is enabled until the query resolves, and if the user opens the confirmation dialog during that window the dialog can still call `handleDelete` even after the permission result arrives. The server rejects unauthorized deletes, so this is not an auth bypass, but it is a destructive-flow regression and confusing error path for org-level prompts owned by another project. Please default to disabled while the permission query is unresolved (`permission?.hasPermission === true`) and cover the loading state in the integration test instead of only asserting that the query is enabled.

langwatch-agent (dismissed):
Request changes: 2 P1s, 1 P2. The lazy-unmount direction is right, and the focused tests I ran pass, but this still has deploy-time persisted-tab data loss plus a fast-switch state-loss path. Details inline.

sergioestebance:
## Performance evidence — the many-tabs degradation is gone

Live A/B profile with a headless-Chromium (Playwright) script: same account, same prompt, opened as 1→16 tabs against two running servers — **pre-fix `main` (65d7cab82)** and **this PR branch** — sampling after a forced GC at each step (so numbers are retained memory, not nodes awaiting collection).

| metric · broken → fixed | 1 tab | 4 tabs | 8 tabs | 12 tabs | 16 tabs |
|---|---|---|---|---|---|
| Live DOM nodes | 2074 → 2021 | 6453 → 2050 | 12289 → 2086 | 18125 → 2122 | 23961 → 2158 |
| Mounted tab panels | 4 → 2 | 16 → 2 | 32 → 2 | 48 → 2 | 64 → 2 |
| Hidden mounted-but-inactive panels | 2 → 0 | 11 → 0 | 23 → 0 | 35 → 0 | 47 → 0 |
| Prompt editors mounted | 2 → 2 | 8 → 2 | 16 → 2 | 24 → 2 | 32 → 2 |
| JS heap MB (after forced GC) | 76.1 → 75.1 | 117.7 → 79.7 | 171.2 → 84.8 | 216.2 → 91.4 | 269.6 → 96.8 |
| Typing latency ms (23-char burst) | 397 → 406 | 520 → 423 | 627 → 380 | 788 → 386 | 858 → 388 |

**Reading it:** every metric grows linearly with tab count on pre-fix and stays **flat** on the fix.

- **Live DOM nodes** — 23,961 → 2,158 at 16 tabs (**~11× smaller**). Inactive tabs now unmount instead of lingering.
- **Hidden mounted-but-inactive panels** — 47 → 0. This is the root cause made visible: Ark UI mounted every tab and only set `hidden`; `lazyMount`/`unmountOnExit` now actually unmounts them.
- **JS heap** — 269.6 → 96.8 MB at 16 tabs (**~2.8× less**); ~13 MB/tab → ~1.4 MB/tab.
- **Typing latency** (fixed 23-char burst) — 858 → 388 ms at 16 tabs (**~2.2× faster**); this is the "sluggishness" users reported, flat again on the fix.

The keystroke burst also emitted a Chrome trace per run: **20 MB** of main-thread work on pre-fix vs **4.2 MB** on the fix — the same story in raw trace weight.

_Methodology: forced `HeapProfiler.collectGarbage` before each sample; live DOM via `document.querySelectorAll('*')`; hidden panels via `[role=tabpanel][hidden]`. Profiling script is not committed (local dev artifact)._

langwatch-agent (dismissed):
Requested changes are not addressed yet. Re-checked latest head `fc4d958`: (1) legacy single-key persisted tab data is still not migrated into per-tab keys, so existing sessions can still be cleared on upgrade; (2) pending debounced prompt form writes are still not flushed before `lazyMount`/`unmountOnExit` unmounts inactive tabs; (3) Delete still defaults enabled while the gated permission query is unresolved (`permission?.hasPermission ?? true`). I did not rerun the focused tests because the requested fixes are visibly absent in the current code.

langwatch/src/prompts/prompt-playground/prompt-playground-store/DraggableTabsBrowserStore.ts
sergioestebance: Fixed in 1d0759090 — getItem now falls back to the legacy embedded t.data when a per-tab key is absent, then migrates it to a per-tab key on the next persist. Regression test seeds the old single-key payload and proves tabs survive + migrate. Good catch, this would have wiped existing users' open tabs on deploy.

langwatch/src/prompts/prompt-playground/components/prompt-browser/ui/DraggableTabsBrowser.tsx
sergioestebance: Fixed in e6e607ad4 — both debounced writers (form via lodash, variables via useDebounceCallback which cancels on unmount) now flush on unmount so a just-typed edit is persisted before the tab tears down. Regression test types a variable value then unmounts and asserts it survives.

langwatch/src/prompts/prompt-playground/components/sidebar/PublishedPromptActions.tsx
sergioestebance: Fixed in e6e607ad4 — canDelete now defaults to false until the permission query resolves as allowed (permission?.hasPermission === true), so Delete stays disabled during the loading window on first open. Integration tests cover both the loading and resolved states.

langwatch/src/prompts/prompt-playground/prompt-playground-store/__tests__/DraggableTabsBrowserStore.test.ts
```diff
@@ -474,6 +590,32 @@ describe("DraggableTabsBrowserStore", () => {
...
+      // store instance wrote a per-tab key that this instance's in-memory
+      // ref map never saw. reset() must still clear it via the prefix scan.
+      const orphanKey = `${TEST_PROJECT_ID}:tab:orphan-from-other-instance`;
+      localStorage.setItem(orphanKey, JSON.stringify(createTabData()));
+
+      store.getState().addTab({ data: createTabData() });
+      store.getState().reset();
+
+      expect(localStorage.getItem(orphanKey)).toBeNull();
+    });
```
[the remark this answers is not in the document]
sergioestebance: Skipping — DraggableTabsBrowserStore.test.ts is structured flat by store method (describe("addTab"), describe("removeTab"), describe("reset"), …), not given/when. Wrapping only the new reset cases in a given layer would make the one file internally inconsistent; converting the whole file is out of scope for this PR.

langwatch/src/prompts/prompt-playground/components/prompt-browser/prompt-browser-window/__tests__/PromptTabbedSection.test.tsx
```diff
@@ -598,4 +603,46 @@ describe("PromptTabbedSection Layout Modes", () => {
...
+      // The 300ms debounce has NOT fired yet (test is faster). Unmounting the
+      // tab (as switching prompt tabs does) must flush the pending write rather
+      // than cancel it — otherwise the edit is lost.
+      unmount();
+
+      expect(store.getState().getByTabId(tabId!)?.variableValues.topic).toBe(
+        "flushed",
+      );
+
+      tabIdRef.current = "test-tab-id";
```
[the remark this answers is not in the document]
sergioestebance: Fixed in 862f98d27 — restore tabIdRef.current in the describe afterEach so it always runs even if an assertion throws, rather than only at the end of the one test. Covers every test using the mock, not just this one.

drewdrewthis:
@sergioestebance — re-review at head `fc4d958` finds the requested changes are still not addressed. Three items block this:

1. **Legacy tab migration** — single-key persisted tab data is still not migrated into per-tab keys, so existing sessions can be cleared on upgrade.
2. **Debounced-write flush** — pending debounced prompt-form writes are still not flushed before `lazyMount`/`unmountOnExit` unmounts inactive tabs (data loss on tab switch).
3. **Delete permission default** — Delete still defaults enabled while the gated permission query is unresolved (`permission?.hasPermission ?? true`); should default disabled until resolved.

Please address all three and re-request review — this PR is off the merge list until it re-passes.

drewdrewthis:
Correction to my previous comment — it was stale. Re-checking the live HEAD (`862f98d`), all three requested changes were in fact addressed by the author's 11:08–11:09 push, which landed just before I commented. Verified:

1. **Legacy tab migration** — `a8f4fae`: `LightTab.data` reads the legacy single-key format and migrates it into per-tab keys (test: "adopts the embedded tab data and migrates it to per-tab keys"). Existing sessions are no longer wiped on upgrade. ✅
2. **Debounced-write flush** — `158fae1`: `updateTabDataDebounced.flush()` added to the unmount cleanup in both `PromptBrowserWindowContent` and `PromptTabbedSection` (tests added). No stale `tab.data` on tab switch. ✅
3. **Delete permission default** — `158fae1`: `canDelete = permission?.hasPermission === true` (was `?? true`) — Delete now stays disabled until the permission query resolves. ✅

CI is green. Ready for re-review. Apologies for the noise on the prior comment.

drewdrewthis:
@sergioestebance  question for you: https://github.com/langwatch/langwatch/pull/5456#issuecomment-4912041293

sergioestebance:
Thanks @drewdrewthis — answering the attention map directly:

**① Unsaved state flushed before unmount + restored on return — yes.** Both debounced writers now flush on unmount: the prompt form (`PromptBrowserWindowContent.tsx:156`, lodash debounce) and variable values (`PromptTabbedSection.tsx:108`, `useDebounceCallback`, which otherwise cancels pending work on unmount). On remount the form re-reads `tab.data.form.currentValues` and variables re-read `variableValues` from the store. So an edit typed right before a tab switch is persisted, not dropped.

**② Test now asserts switch-away-and-back, not just unmount.** You were right that `inactiveTabUnmount.test.tsx` only proves the panel unmounts. Added a round-trip test (`PromptTabbedSection.test.tsx` → "restores the variable value the user had typed"): type a value → unmount (switch away) → remount (switch back) → assert it's shown again. Verified it fails without the flush fix, so it actually catches the "lose user work" regression.

**③ In-flight generation on switch-away: aborted cleanly, not orphaned.** The `CopilotKit` provider lives inside the tab content, so unmounting the prompt tab tears it down and aborts the run — no orphaned background request. Completed messages persist via the store snapshot; the in-flight turn does not resume. This is a deliberate tradeoff of the memory fix (documented in the description). Note the *nested* sub-tabs use `lazyMount` **without** `unmountOnExit`, so switching Conversation→Parameters mid-stream does **not** abort — only switching to a different prompt tab does.

sergioestebance:
@drewdrewthis answered the brief here → https://github.com/langwatch/langwatch/pull/5456#issuecomment-4916221369

TL;DR on the three targets:
1. **Flush before unmount + restore on return** — done: `.flush()` on unmount in both writers (`PromptBrowserWindowContent.tsx:156`, `PromptTabbedSection.tsx:108`); remount re-reads from the store.
2. **Test asserts switch-away-and-back** — you were right it was missing. Added `restores the variable value the user had typed` (`PromptTabbedSection.test.tsx`, commit `51a5d6f6e`): type → unmount → remount → value restored on screen. Verified it fails without the flush.
3. **In-flight generation on switch-away** — aborted cleanly, not orphaned (CopilotKit provider unmounts with the tab); completed messages persist; sub-tab switches deliberately keep the chat mounted.

And thanks for the self-correction on the earlier stale re-review — no worries.

Aryansharma28 (approved):

PR #5464, opened 2026-07-07
fix(model-providers): use flat provider list for ID-based lookup in edit drawer

### LINKED ISSUE #5463: fix(model-providers): edit drawer shows blank form for multi-instance custom providers
## Bug

When multiple model providers share the same `provider` string (e.g. four rows all with `provider = "custom"`), opening the edit drawer for any non-winning row shows a completely blank form — no models, no base URL, no API key, no extra headers.

## Root Cause

`useModelProvidersSettings` calls `getAllForProjectForFrontend` which returns a `Record<providerKey, config>`. When multiple rows share the same provider string, the record collapses them into a single entry (narrowest-scope wins). Both `EditModelProviderDrawer` and `EditModelProviderForm` then do:

```ts
Object.values(providers).find((p) => p.id === modelProviderId)
```

On a record with only one `"custom"` entry, the IDs of the three non-winning rows are never found → `find` returns `undefined` → form falls back to a blank default object. Every field that reads from `provider` (customModels, customKeys, extraHeaders, etc.) renders empty.

## Steps to Reproduce

1. Add 4+ providers all using the "Custom" provider type (e.g. four local vLLM endpoints)
2. Go to Settings → Model Providers
3. Click **Edit** on any provider that is not the first one added
4. All fields in the drawer are blank (no models, no URL, no API key)

## Fix

Add a parallel `listAllForProjectForFrontend` query to `useModelProvidersSettings` that returns one entry per stored row. Expose it as `providersList` and use it for the ID lookup in both `EditModelProviderDrawer` and `EditModelProviderForm`. The `providers` Record is preserved for all other consumers that only need enabled-state or model metadata.

thiagolmoraes:
Fixes #5463

## Summary

- `useModelProvidersSettings` now fires a parallel `listAllForProjectForFrontend` query and exposes the result as `providersList` (one entry per stored row)
- `EditModelProviderDrawer` and `EditModelProviderForm` use `providersList?.find(p => p.id === id)` instead of `Object.values(providers).find(...)` for the ID lookup
- The `providers` Record shape is preserved unchanged for all other consumers (enabled-state checks, model metadata, `hasEnabledProviders`)

## Test plan

- [ ] Add 4+ providers with `provider = "custom"` (e.g. local vLLM endpoints with different base URLs and models)
- [ ] Open edit drawer for each — all fields (models, base URL, API key, extra headers) populate correctly
- [ ] Edit and save each — changes persist correctly
- [ ] Single-instance providers (Anthropic, OpenAI, etc.) still work normally
- [ ] `ModelSelector`, `LLMConfigPopover`, scenario components unaffected

## Summary by CodeRabbit

* **New Features**
  * Added a GitHub Actions workflow to build HML deployments from the `deploy/hml` branch.
  * Enhanced model discovery and `/v1/models` responses with allowlist, alias, policy deny, and deterministic sorting behavior (deduped by model ID).
* **Bug Fixes**
  * Improved model provider selection/editing using ID-based matching for existing entries.
  * Enabled OpenAI-compatible `/v1/models` discovery for self-hosted providers and better handling of Anthropic-style custom base URLs.
* **Documentation**
  * Added a manual guide for Anthropic custom base URL testing.
* **Chores**
  * Pinned front-end dependency versions and reduced runtime image contents.


.github/workflows/hml.yml
```diff
@@ -0,0 +1,12 @@
+name: hml
+on:
+  push:
+    branches: [deploy/hml]
+
+jobs:
+  build:
+    runs-on: [self-hosted, hml]
```
langwatch-agent, line 8: **P1** — This bugfix PR adds a new `push` workflow for `deploy/hml` that runs arbitrary branch contents through a self-hosted `hml` runner and a mutable host script (`$HOME/.deploy/run.sh`). That is a separate deployment/security surface from the model-provider drawer fix, and it can be triggered by future pushes to this branch before this code has been reviewed as deploy automation. Please remove this workflow from this PR and ship it separately with the normal runner/branch-protection/permissions/concurrency review.

langwatch/src/hooks/useModelProvidersSettings.ts
```diff
@@ -14,8 +14,17 @@ export function useModelProvidersSettings(params: {
     { enabled: Boolean(projectId) },
   );
 
+  // Full list (one entry per stored row) used for ID-based lookups in the
+  // edit drawer. The Record shape above collapses multiple rows with the same
+  // provider string (e.g. four "custom" rows) into one, so ID lookups for
+  // non-winning rows return undefined and the form renders blank.
+  const providersList = api.modelProvider.listAllForProjectForFrontend.useQuery(
```
langwatch-agent, line 21: **P2** — Adding this second query broke the existing focused hook test suite because `src/hooks/__tests__/useModelProvidersSettings.unit.test.ts` still mocks only `getAllForProjectForFrontend`. I ran `pnpm run test:unit src/hooks/__tests__/useModelProvidersSettings.unit.test.ts --run` and all 5 tests fail with `TypeError: Cannot read properties of undefined (reading 'useQuery')` at this line. Please update the test mock to include `listAllForProjectForFrontend.useQuery` (and ideally add coverage for `providersList` loading/data) so the regression is caught in the app test suite.

langwatch-agent (changes requested):
Request changes: 1 P1, 1 P2. The model-provider lookup fix is directionally right, but this PR currently adds an unrelated self-hosted deploy workflow and breaks the existing focused hook unit test. Details inline.

thiagolmoraes:
Closing: the underlying bug was fixed upstream by #5383 (`bb123148a`, merged 2026-07-13), which removed the collapsed-record ID lookup this PR targeted — same root cause reported in #5463, resolved with a dedicated edit-target resolver and integration coverage. This branch has since accumulated unrelated fork-deployment commits, so there is nothing left to salvage here. Superseded and no longer needed.

PR #5469, opened 2026-07-08
fix(evaluations-legacy): pass resolved model to getEvaluatorDefaultSettings (#5468)

### LINKED ISSUE #5468: fix(evaluations-legacy): legacy path hardcodes DEFAULT_MODEL, bypassing the project model cascade
## Expected

Evaluations executed via the legacy evaluations path use the model resolved by the project's model cascade (project setting → resolved default), like the rest of the evaluation surfaces.

## Actual

The legacy path ignores the resolved model argument and hardcodes `DEFAULT_MODEL`, bypassing the model cascade entirely — projects configured with a different default get evaluations run on the wrong model.

## Reproduction

Configure a project default model different from the global `DEFAULT_MODEL`; run an evaluation through the legacy evaluations path; observe the evaluation executes with `DEFAULT_MODEL` instead of the project's resolved model. (Flagged during the 2026-07-04 evaluations-legacy review pass — the call site takes no resolved arg.)

---
*Classification: Bug*

drewdrewthis:
## What

Fixes #5468.

The legacy evaluations route called `getEvaluatorDefaultSettings(evaluatorDefinition as any)` without a `resolved` model argument. When the arg is absent the function falls through to `DEFAULT_MODEL` (`getLatestOpenAIChatFlagship() ?? "openai/gpt-5"`), bypassing the project model cascade entirely.

## Why

Every external API caller (`/api/evaluations`, `/api/guardrails`) was silently getting evaluations run on the global flagship model, even if the project was configured with a different default. The three UI call sites (`CheckConfigForm`, `EvaluatorPropertiesPanel`, `EvaluatorEditorShared`) and the server-side evaluator-create path (`app/api/evaluators/.../app.v1.ts`) were already correct — the legacy REST route was the sole gap.

## How

After the project is loaded in `handleEvaluatorCall`, resolve the project's cascade-configured models and pass them as the second arg to `getEvaluatorDefaultSettings`. The resolution is lazy — `resolveEvaluatorSettingsDefaults(project.id)` is called only inside the `!workflowEvaluatorDef` branch where its result is consumed, so `custom/` evaluators and workflow-evaluator paths skip the two Prisma-backed queries entirely.

Extracted a small exported helper `resolveEvaluatorSettingsDefaults(projectId)` so the resolve+map logic is unit-testable without the Hono/prisma/auth harness.

## Deviations from the original issue sketch (deliberate)

The issue comment suggested `getResolvedDefaultForFeature(project.id, "evaluations")` returning a string. The **actual** contract is different, and following the sketch verbatim would have silently no-op'd:

1. **Signature:** `getResolvedDefaultForFeature(ctx: ReadCtx, { projectId, featureKey })` — takes a `{ prisma, session }` ctx + params object, returns `DefaultModelEffective | null` (an object with `.model`), not a bare string.
2. **Feature key:** there is **no `"evaluations"` key** in the feature registry — `featureByKey("evaluations")` returns `undefined`, so the resolver would return `null` and the fix would do nothing. The correct DEFAULT-role key for the evaluator model is **`evaluator.create_default`** (the same key `app.v1.ts` uses on evaluator create). Embeddings resolve via `analytics.topic_clustering_embeddings`, matching the UI.
3. **Second-arg shape:** `getEvaluatorDefaultSettings(evaluator, resolved?)` expects `{ defaultModel?, embeddingsModel? }`, so the resolved `.model` values are mapped into that shape.

## Human verification

1. Configure a project with a non-default model (e.g. `openai/gpt-4o` while global flagship is `openai/gpt-5`).
2. POST to `/api/evaluations` via the REST API.
3. Observe evaluator runs with the project's configured model, not `openai/gpt-5`.
4. Repeat with a project with no custom model — confirm it still uses `DEFAULT_MODEL` (no regression).

**Fails-on-main signature (cascade-bypass case):** on `main`, no resolved arg is passed, so `getEvaluatorDefaultSettings` always returns `model: DEFAULT_MODEL` regardless of project config. The new test stubs `getResolvedDefaultForFeature` to return `"openai/gpt-4o"` and asserts the merged settings carry `openai/gpt-4o` (and `!== DEFAULT_MODEL`) — this passes on this branch and would fail on main. A companion test pins the no-config path (resolver → `null` → `DEFAULT_MODEL`) so there is no regression for projects without a custom default (AC#2).

```
# grep DEFAULT_MODEL in evaluations-legacy.ts — only the JSDoc mention, no code/fallback site:
668: * `DEFAULT_MODEL` (`getLatestOpenAIChatFlagship()`), bypassing the project's

# tsc (tsgo) — 0 errors in changed files; the 2 reported errors pre-exist on main (start.ts, vite.config.ts):
src/start.ts(47,7): error TS2769: No overload matches this call.
vite.config.ts(63,7): error TS2769: No overload matches this call.

# vitest — new cascade test + existing legacy coercion test:
 Test Files  2 passed (2)
      Tests  15 passed (15)
```

Backend-only — no UI surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model selection for legacy evaluation requests by resolving project-level defaults for both the default model and embeddings model.
  * Ensured correct fallback to standard default settings when no cascade configuration is available.
* **Tests**
  * Added unit tests for legacy evaluator default-resolution behavior, covering configured defaults and the unset/no-config fallback cases.

### DIFF
--- a/langwatch/src/server/routes/__tests__/evaluations-legacy-model-cascade.unit.test.ts
+++ b/langwatch/src/server/routes/__tests__/evaluations-legacy-model-cascade.unit.test.ts
@@ -1,0 +1,128 @@
+/**
+ * Regression: the legacy evaluations REST route (POST /api/evaluations/...)
+ * used to call `getEvaluatorDefaultSettings(evaluatorDefinition)` WITHOUT the
+ * resolved-model argument, so every API-triggered evaluation fell through to
+ * the hardcoded global `DEFAULT_MODEL` and ignored the project's model cascade
+ * configuration (issue #5468).
+ *
+ * The fix threads `resolveEvaluatorSettingsDefaults(project.id)` — which wraps
+ * the same cascade resolver the UI and server-side create path use — into the
+ * settings-merge, so a project with a custom default model gets that model.
+ *
+ * These tests pin:
+ *  - AC#3/#4/#6: `resolveEvaluatorSettingsDefaults` returns the cascade model
+ *    for the correct project + the `evaluator.create_default` feature key.
+ *  - AC#2:       when the cascade has nothing configured (resolver -> null),
+ *    `getEvaluatorDefaultSettings` still falls back to `DEFAULT_MODEL`.
+ */
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// The route module imports prisma (db) and the app layer at load time; stub
+// the resolver so no DB is touched. `prisma` itself is only handed to the
+// resolver as `ctx.prisma`, which the mock ignores.
+const getResolvedDefaultForFeature = vi.fn();
+vi.mock("~/server/modelProviders/modelDefaults.read", () => ({
+  getResolvedDefaultForFeature: (...args: unknown[]) =>
+    getResolvedDefaultForFeature(...args),
+}));
+
+import { getEvaluatorDefaultSettings } from "~/server/evaluations/getEvaluator";
+import { DEFAULT_MODEL } from "~/utils/constants";
+import { resolveEvaluatorSettingsDefaults } from "../evaluations-legacy";
+
+// A minimal evaluator definition whose settings carry a `model` field — the
+// exact shape `getEvaluatorDefaultSettings` maps the resolved default onto.
+const evaluatorWithModel = {
+  name: "LLM Judge",
+  requiredFields: ["input", "output"],
+  optionalFields: [],
+  settings: {
+    model: { default: "openai/gpt-5" },
+    max_tokens: { default: 8192 },
+  },
+} as any;
+
+describe("resolveEvaluatorSettingsDefaults", () => {
+  beforeEach(() => {
+    getResolvedDefaultForFeature.mockReset();
+  });
+
+  describe("when the project has a custom default model configured", () => {
+    it("maps the cascade-resolved model into { defaultModel, embeddingsModel }", async () => {
+      getResolvedDefaultForFeature.mockImplementation(
+        async (_ctx: unknown, params: { featureKey: string }) => {
+          if (params.featureKey === "evaluator.create_default") {
+            return {
+              model: "openai/gpt-4o",
+              source: "feature_override",
+              scope: "project",
+            };
+          }
+          return {
+            model: "openai/text-embedding-3-large",
+            source: "feature_override",
+            scope: "project",
+          };
+        },
+      );
+
+      const resolved = await resolveEvaluatorSettingsDefaults("proj-1");
+
+      expect(resolved).toEqual({
+        defaultModel: "openai/gpt-4o",
+        embeddingsModel: "openai/text-embedding-3-large",
+      });
+    });
+
+    it("resolves for the given project id and the evaluator.create_default feature key (AC#3/#4)", async () => {
+      getResolvedDefaultForFeature.mockResolvedValue({
+        model: "anthropic/claude-3-5-sonnet",
+        source: "role_default",
+        scope: "team",
+      });
+
+      await resolveEvaluatorSettingsDefaults("proj-42");
+
+      expect(getResolvedDefaultForFeature).toHaveBeenCalledWith(
+        expect.anything(),
+        { projectId: "proj-42", featureKey: "evaluator.create_default" },
+      );
+      // Never called with an undefined project id (AC#4).
+      for (const call of getResolvedDefaultForFeature.mock.calls) {
+        expect(call[1].projectId).toBe("proj-42");
+      }
+    });
+
+    it("feeds getEvaluatorDefaultSettings the custom model, not DEFAULT_MODEL (AC#6)", async () => {
+      getResolvedDefaultForFeature.mockResolvedValue({
+        model: "openai/gpt-4o",
+        source: "feature_override",
+        scope: "project",
+      });
+
+      const resolved = await resolveEvaluatorSettingsDefaults("proj-1");
+      const settings = getEvaluatorDefaultSettings(
+        evaluatorWithModel,
+        resolved,
+      );
+
+      expect((settings as any).model).toBe("openai/gpt-4o");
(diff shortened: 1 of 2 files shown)


drewdrewthis:
## Review verdict: READY

**CI:** 55 checks pass, 0 failing (CodeQL in-progress at post time, langwatch-app-ci unit/integration/build/typecheck running — all non-skipped gates green).

**Implementation verified:**
- `resolveEvaluatorSettingsDefaults` calls `getResolvedDefaultForFeature(ctx, { projectId, featureKey })` — exact signature match (confirmed against `modelDefaults.read.ts:93`).
- Feature keys `evaluator.create_default` and `analytics.topic_clustering_embeddings` match `featureRegistry.ts:50` and `featureRegistry.ts:121` and the parallel `app.v1.ts:199/203` create path.
- `getEvaluatorDefaultSettings` second-arg shape `{ defaultModel, embeddingsModel }` maps correctly to `model` (line 37) and `embeddings_model` (line 40) in `getEvaluator.ts`.
- UI call sites (tRPC `getResolvedDefault`) use the same object-shape second arg — no inconsistency.
- `resolveEvaluatorSettingsDefaults` is new; no pre-existing duplicate to reuse.
- Null-cascade path (resolver returns `null` → `{ defaultModel: null, embeddingsModel: null }`) preserves `DEFAULT_MODEL` fallback — AC#2 clean.

**Tests:** Unit tests stub `getResolvedDefaultForFeature`, assert correct projectId + featureKey call, confirm cascade model flows through `getEvaluatorDefaultSettings`, and confirm null-cascade falls back to `DEFAULT_MODEL`. Not tautologies.

No behavior change for projects without custom defaults. No hygiene issues. One-line fix at the call site with a well-scoped extracted helper. No comments needed.

drewdrewthis (commented):
Applied CodeRabbit suggestion in commit 0ea1e039d — moved `await resolveEvaluatorSettingsDefaults(project.id)` inside the `!workflowEvaluatorDef` branch. `custom/` evaluators and workflow-evaluator paths now skip the two Prisma-backed cascade queries. No behavior change on the corrected path.

drewdrewthis:
## Review verdict: READY

**Delta from 2014dc66a (CodeRabbit quick-win):** `resolveEvaluatorSettingsDefaults(project.id)` moved inline into the `!workflowEvaluatorDef` branch — lazy, so workflow-evaluator paths and schema-less requests skip both Prisma-backed cascade queries entirely. The fix is strictly better: tighter scope, no wasted I/O on the paths that never consumed the result.

**CodeRabbit thread:** already resolved (applied + `drewdrewthis` replied in 0ea1e039d).

**CI:** green (55 pass, 0 failing at HEAD).

All prior verification still holds: signature match, feature keys, `getEvaluatorDefaultSettings` mapping, null-cascade fallback, no pre-existing duplicate. No new issues introduced by the delta.

rogeriochaves (approved):

drewdrewthis:

drewdrewthis:
Reworking this to stop duplicating default-model resolution and consume #5398's service instead (per owner review). Checkpoint of the findings before the code lands:

**1. This branch is independent of #5398** — `fix/5468-legacy-eval-model-cascade` branches from `main`, not from `fix/3531-resolve-default-model-trpc` (verified: #5398's head is not an ancestor of this one). So this PR **cannot import `resolveDefaultModel` until #5398 lands on main** (or this branch is rebased to stack on #5398). We will **state that dependency and hold** rather than copy the service in here — copying is the duplication we're removing.

**2. The duplication is the cascade walk, not the whole function.** This PR's `resolveEvaluatorSettingsDefaults` (in `evaluations-legacy.ts`) resolves **two feature-scoped models** — a default (`evaluator.create_default`) and an embeddings model (`analytics.topic_clustering_embeddings`) — via `getResolvedDefaultForFeature`. #5398's `resolveDefaultModel(projectId)` resolves the **single project default**. The shared logic is the underlying project→org→global→null cascade. Dedup options: (A) generalize #5398's service to take a `featureKey`; (B) extract the shared cascade helper and have both consume it. Leaning (B) for smaller blast radius on #5398's already-reviewed surface — will confirm after diffing `modelDefaults.read` against #5398's internal cascade (they may already be the same helper).

**Dependency:** needs #5398 first. **Holding review-ready, not merging.** Code via a dedicated lane; full plan checkpointed.

drewdrewthis:
**On the "rework #5469 to consume #5398's resolution instead of duplicating" ask — verified against the code: there is no duplication to remove. #5469 is already the single-source-of-truth consumer, so it stays as-is (no code change).**

The two functions resolve *different things via different mechanisms*:

| | #5469 `resolveEvaluatorSettingsDefaults` | #5398 `resolveDefaultModel` |
|---|---|---|
| Mechanism | feature-key **DB cascade** (project→team→org) via the shared `getResolvedDefaultForFeature` → `resolveModelForFeature` | **provider-order fallback** over `PROVIDER_DEFAULT_MODELS` / `PROVIDER_RESOLUTION_ORDER` |
| Source of truth | `ModelDefaultConfig` rows | enabled-provider canonical defaults |
| Returns | two feature-scoped models (`evaluator.create_default` + `analytics.topic_clustering_embeddings`) | one project-level model string, or `null` |
| Touches `ModelDefaultConfig`? | yes | no — by design (`resolveModelForFeature`: *"intentionally no global system fallback"*) |

`getResolvedDefaultForFeature` / `resolveModelForFeature` already live on **`main`** (not introduced by #5398), and #5469 **re-implements no cascade** — it calls that shared helper twice and maps the result shape. So #5469 is already deduplicated against the one cascade source.

Reworking it to call #5398's `resolveDefaultModel` would **regress #5468**: it would bypass the `ModelDefaultConfig` cascade entirely and cannot produce the embeddings model.

The two branches are also **independent** — both branch from `main`, and `fix/3531…` is not an ancestor of `fix/5468…` (`git merge-base --is-ancestor` confirms). So #5469 carries no dependency on #5398; the #5398 file rename lands separately.

PR #5551, opened 2026-07-09
fix(invite): surface acceptInvite failures instead of dead-ending on the loading screen

### LINKED ISSUE #5550: Invite accept page dead-ends on the loading screen when acceptInvite fails (error only visible in console)
## Symptom

Opening an invite link while signed in as a different account than the invite was sent to leaves the user on the bare LangWatch loading screen forever. No error message, no way forward. The `organization.acceptInvite` mutation correctly fails with `FORBIDDEN` ("The invite was sent to X, but you are signed in as Y") — but the message is only visible in the browser console. Reported from a real invite flow on app.langwatch.ai.

The page *has* an error UI (`src/pages/invite/accept.tsx` renders an Alert with the server message + a "Log Out and Try Again" button), but it never renders.

## Root cause — two interacting defects

**1. `InnerProviders` remounts the whole page subtree when PostHog initializes** (`src/AppProviders.tsx`):

```tsx
{postHog ? (
  <PostHogProvider client={postHog}>{children}</PostHogProvider>
) : (
  children
)}
```

`usePostHog()` returns `undefined` until the `publicEnv` tRPC query resolves, then flips to the posthog client. That changes the element *type* at this tree position (page → `PostHogProvider`), so React unmounts and remounts the entire routed page shortly after boot — on every SaaS page load.

**2. `useAcceptInviteOnce` cannot recover state after a remount** (`src/hooks/useAcceptInviteOnce.ts`):

The hook fires the mutation at most once per invite code via a module-scoped `Set` guard (added in #3325 to prevent double-submits — the guard deliberately survives remounts). But all *outcome* state lives in the per-instance `useMutation`. Sequence:

1. Page mounts, session resolves, effect fires `mutate({ inviteCode })` — guard records the code.
2. Mutation fails (403 logged to console).
3. PostHog init flips the provider tree → page subtree remounts.
4. The remounted instance has a fresh **idle** mutation; the guard blocks re-firing; `deriveStatus` returns `"loading"` → `LoadingScreen` forever.

Whether the user sees the error UI or a permanent spinner is a race between the `publicEnv` query (drives the remount) and the session query + mutation dispatch. The existing regression test asserts the remounted instance does not re-submit, but never asserts what status it reports — which is a permanent `loading`.

The success path is exposed to the same race (remount mid-flight can drop the redirect/toast).

## Fix

1. **Stable provider tree**: always render `PostHogProvider` so PostHog init no longer remounts the page subtree.
2. **Remount-proof invite state**: keep the accept outcome in a module-scoped store (alongside the existing one-shot guard) so any remounted instance resolves to the real status instead of dead-ending on `loading`.
3. **Error UI**: keep showing the server message explaining the situation, and add a **Go to Dashboard** action next to "Log Out and Try Again" so the user is never stranded.

sergioestebance:
# Related Issue

- Resolves #5550

## Why

Accepting an invite while signed in as a different account (or with an expired/invalid code) left the user on the bare LangWatch loading screen forever — the `FORBIDDEN`/`NOT_FOUND` error from `organization.acceptInvite` was only visible in the browser console. Reported from a real invite flow on app.langwatch.ai.

Closes #5550

## What changed

- **`AppProviders.tsx`**: `InnerProviders` now always wraps the routed page in `PostHogProvider` (with the `posthog-js` module singleton). Previously the wrap was conditional on `usePostHog()` resolving, and that element-type flip made React unmount + remount the entire routed page subtree shortly after every SaaS boot.
- **`useAcceptInviteOnce`**: mutation outcomes are now recorded in a module-scoped store (read via `useSyncExternalStore`), the counterpart of the existing module-scoped one-shot submit guard from #3325. A remounted page instance now resolves to the real terminal state instead of reporting `loading` forever.
- **`/invite/accept` error UI**: adds a primary **Go to Dashboard** action (hard navigation, busting stale pre-invite caches) next to **Log Out and Try Again**, so the user always sees what happened and has a way out.

## How it works

Two defects had to interact to produce the dead-end:

1. The PostHog conditional wrap remounted the page subtree once the `publicEnv` query resolved.
2. The one-shot guard (correctly) blocked a re-submit after the remount, but the outcome lived only in the unmounted instance's `useMutation` state — the fresh instance stayed idle, so `deriveStatus` returned `"loading"` and the page rendered `LoadingScreen` indefinitely.

Whether the user saw the error UI or the permanent spinner was a race between the `publicEnv` query (drives the remount) and the session query + mutation dispatch. Both sides are fixed: the tree no longer remounts on posthog init, and the hook now survives any remount (error boundaries, HMR, future provider changes) because react-query keeps the original mutation's callbacks alive even after the dispatching component unmounts — they write the outcome to the store and every subscribed instance re-derives its status.

## Test plan

- `src/hooks/__tests__/useAcceptInviteOnce.integration.test.tsx`: two new regression tests — remount after the error, and remount while the mutation is still in flight with the error arriving after. **Both fail without the fix** (status stuck on `loading`), pass with it. All 10 tests green.
- `src/pages/invite/__tests__/accept.integration.test.tsx` (new): error state shows the server's explanation + both actions; "Go to Dashboard" hard-navigates to `/`; "Log Out and Try Again" signs out; loading state shows no alert. 5 tests green.
- `pnpm typecheck` green.
- **Browser-verified end-to-end**: signed in locally, hit `/invite/accept?inviteCode=bogus-code-5550` → error card renders with message and both buttons; "Go to Dashboard" navigates home. Screenshot below.

## Anything surprising?

The provider fix removes a boot-time remount of the whole page tree that every SaaS page paid for — the invite page was just the only surface with one-shot semantics that made it fatal. Passing the uninitialized posthog singleton to `PostHogProvider` is inert when no `POSTHOG_KEY` is configured; `usePostHog()` still owns initialization.

---

### Screenshot — error state after the fix (was: permanent blank loading screen)

invite accept error UI

## Summary by CodeRabbit

* **New Features**
  * Improved invite acceptance handling so progress and error state can survive page reloads or remounts without getting stuck.
  * Added a new **Go to Dashboard** action on invite error screens for quicker recovery.

* **Bug Fixes**
  * Fixed invite acceptance errors from disappearing or resetting when the page rerenders.
  * Preserved the displayed error message and status after failures.
  * Updated the error screen to better handle both failed and in-progress invite acceptance states.

### DIFF
--- a/langwatch/src/AppProviders.tsx
+++ b/langwatch/src/AppProviders.tsx
@@ -1,4 +1,5 @@
 import { ChakraProvider } from "@chakra-ui/react";
+import posthog from "posthog-js";
 import { PostHogProvider } from "posthog-js/react";
 import type { ReactNode } from "react";
 import { AnalyticsProvider } from "react-contextual-analytics";
@@ -55,11 +56,15 @@ export function InnerProviders({ children }: { children: ReactNode }) {
             isGtagReady,
           })}
         >
-          {postHog ? (
-            <PostHogProvider client={postHog}>{children}</PostHogProvider>
-          ) : (
-            children
-          )}
+          {/* Always wrap in PostHogProvider with the module singleton —
+              `usePostHog()` initializes it in an effect once publicEnv
+              resolves, so conditionally wrapping on that flip changes the
+              element type at this position and React unmounts + remounts
+              the ENTIRE routed page subtree shortly after boot. That
+              remount wiped in-flight page state (#5550: /invite/accept
+              dead-ended on the loading screen). The uninitialized
+              singleton is inert when no POSTHOG_KEY is configured. */}
+          <PostHogProvider client={posthog}>{children}</PostHogProvider>
         </AnalyticsProvider>
         <Toaster />
       </CommandBarProvider>
--- a/langwatch/src/hooks/__tests__/useAcceptInviteOnce.integration.test.tsx
+++ b/langwatch/src/hooks/__tests__/useAcceptInviteOnce.integration.test.tsx
@@ -34,8 +34,14 @@ const {
     captureExceptionSpy: vi.fn(),
     mockState: {
       handlers: {} as {
-        onSuccess?: (data: unknown) => void;
-        onError?: (error: { message: string }) => void;
+        onSuccess?: (
+          data: unknown,
+          variables: { inviteCode: string },
+        ) => void;
+        onError?: (
+          error: { message: string },
+          variables: { inviteCode: string },
+        ) => void;
       },
       mutation: {
         isLoading: false,
@@ -197,9 +203,10 @@ describe("useAcceptInviteOnce()", () => {
           mockState.mutation.error = {
             message: INVITE_ALREADY_ACCEPTED_MESSAGE,
           };
-          mockState.handlers.onError?.({
-            message: INVITE_ALREADY_ACCEPTED_MESSAGE,
-          });
+          mockState.handlers.onError?.(
+            { message: INVITE_ALREADY_ACCEPTED_MESSAGE },
+            { inviteCode: "invite-abc" },
+          );
         });
         rerender();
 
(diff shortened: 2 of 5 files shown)


sergioestebance:
Blessed. 🙏 *"For you were once darkness, but now you are light in the Lord. Live as children of light."* — Ephesians 5:8. The silent console error now walks in the light with a way home. Ship it — mind the P2: the provider-tree stability fix itself has no regression test.

0xdeafcafe (approved):

PR #5577, opened 2026-07-09
feat(prompts): add Duplicate prompt action to playground menu

Aryansharma28:
## Summary

Adds a **Duplicate prompt** action to the prompt playground's row overflow menu (`⋮`), right after "Replicate to another project" — per the Slack thread asking for a duplicate option next to that existing action.

A new `api.prompts.duplicate` tRPC mutation duplicates a prompt within the **same project**, distinct from `api.prompts.copy` (cross-project replicate). Duplicates are numbered `handle-1`, `handle-2`, … , incrementing past whatever is already taken.

> Naming originally used `handle(1)` parentheses. `handleSchema` forbids that format, and a handle that fails it is silently forced into "draft" mode (`handle: null`) when the prompt is reopened for editing — see `isHandleValid` in `llmPromptConfigUtils.ts`. A hyphen suffix is schema-safe.

## Where the logic lives

Both `duplicate` and `copy` now delegate to `PromptService`, so the router only does transport work (permissions, license limit, error mapping):

- `PromptService.duplicatePrompt()` — same-project, numbered `-1`, `-2`, …
- `PromptService.copyPrompt()` — cross-project, tries the bare handle first, then `_copy1`, `_copy2`, …
- Shared, private to the service: `generateUniqueHandle()`, `deriveBaseHandle()`, `buildCreateParamsFromSource()`
- `LlmConfigRepository.setCopiedFromPrompt()` — new. `copyPrompt` previously wrote the copy-source link through raw Prisma with no `projectId` in the `where` clause; every write against a project-scoped model must carry it.
- `hoistSystemMessage()` — extracted to `~/server/prompt-config/hoist-system-message.ts`, shared by `copy`, `duplicate`, `syncFromSource`, and `pushToCopies`. It is load-bearing: `createPrompt` rejects input carrying both a `prompt` and a system message (`SystemPromptConflictError`), so a prompt read back from storage cannot be handed straight back to it.
- `toHandleSlug()` — new. `deriveBaseHandle` falls back to `name`, which is free-form text the handle format rejects. Guarantees a schema-valid handle for prompts that never had one.

This removes the three module-level helpers the previous revision added to the router (`normalizePromptData`, `generateUniqueHandle`, `buildCreatePromptParamsFromSource`), addressing @sergioestebance's "we can use its own service for this function". The router keeps only transport concerns; `rethrowAsTRPCError` translates the service's `NotFoundError` into a tRPC `NOT_FOUND` at the boundary.

`copy`'s candidate sequence is unchanged — `base`, `base_copy1` … `base_copy100`, then throw.

**One deliberate behavior change to `copy`:** it previously derived the base handle as `handle ?? id`. For a prompt with no handle (drafts and legacy configs carry `handle: null`) that produced a base like `prompt_1h5icu8XRkHHbaQlrOgwq` — uppercase, which `handleSchema` rejects — so the copy silently landed with an invalid handle and got forced into draft mode on reopen. It now derives `handle ?? name` and slugifies anything the schema would reject. Prompts that already have a valid handle are unaffected.

## Tests

`specs/prompts/duplicate-prompt.feature` describes the behavior; every scenario is bound to a test.

- `prompts.duplicate.integration.test.ts` (9) — real tRPC + Prisma. Covers numbering (`-1` then `-2`), config carry-over, own version history, staying in-project, not-found, and that the plan's prompt allowance is enforced before duplicating. Also pins that a structured output's `responseFormat` survives: it is derived from `outputs` at read time rather than stored, so nothing else would catch the mapper dropping `outputs`.
- `prompts.copy.integration.test.ts` (8) — **new**. `copy` had no integration coverage at all, while this PR both refactors it and changes its base-handle derivation. Covers handle preservation, `_copy1` suffixing, `copiedFromPromptId`, source left untouched, the prompt allowance, and not-found.
- `llm-config.repository.setCopiedFromPrompt.unit.test.ts` (2) — pins the `{ id, projectId }` tenancy filter.
- `prompt.service.duplicate-and-copy.unit.test.ts` (15) — both suffix contracts, slug fallback for handle-less prompts, and the max-attempts throw (`HandleGenerationError` after 101 candidates).
- `handle-slug.unit.test.ts` (15) / `hoist-system-message.unit.test.ts` (6) — pure functions. The slug suite asserts the invariant that matters: whatever goes in, what comes out satisfies `handleSchema`.
- `PublishedPromptActions.integration.test.tsx` (+5) — clicking "Duplicate prompt" calls the mutation with `{ idOrHandle, projectId }`, refreshes the list, names the duplicate in the toast, surfaces failures, and stays quiet when the global prompt-limit dialog has already told the user.

Each suite was mutation-tested: the source was deliberately broken (off-by-one in the numbering, dropped loop bound, removed `enforceLicenseLimit`, removed the menu item, dropped the `projectId` tenancy filter, dropped `outputs`, …) and every mutation was confirmed to turn the intended test red.

Two of those mutations caught tests that were passing for the wrong reason. The first `enforceLicenseLimit` assertion was satisfied by the `prompts.create` call in the test's own setup, so it stayed green even with the limit check deleted from `duplicate` — now fixed with `mockClear()` plus `toHaveBeenCalledTimes(1)`. And the new `projectId` tenancy filter was initially unreachable from the service test, which stubs the repository — hence the dedicated repository test.

## Known limitations (all pre-existing, none introduced here)

- **`checkHandleUniqueness` re-resolves the organization on every candidate.** Each probe costs a `project.findUnique` with two joins plus the uniqueness lookup. Hoisting the org lookup would halve it, but `checkHandleUniqueness` reads `params.organizationId` inside its `OR` clause, so passing it flips that branch's semantics — for a legacy row with a null `organizationId` a graceful "handle taken" would become a Prisma `P2002`. Left alone rather than smuggle a behavior change into a refactor.
- **Handle uniqueness is check-then-insert.** `generateUniqueHandle` probes for a free handle and `createPrompt` then inserts it; `LlmPromptConfig` has `@@unique([handle])`, so two concurrent duplicates of the same prompt can both see `-1` free and one will lose with a Prisma `P2002`. Fixing it properly means insert-and-retry on `P2002`. `copy` has had this race since it was written; this PR does not widen it. Filed as #5592.
- **The REST route (`app.v1.ts`) never calls `enforceLicenseLimit`.** The prompt allowance is enforced on the tRPC path only, so a free plan can exceed it via the SDK. Filed as #5593.

## Test plan

- [x] `pnpm typecheck`
- [x] `prompt-config` unit suite
- [x] `prompts.duplicate.integration.test.ts` + `prompts.copy.integration.test.ts` against real Postgres
- [x] `PublishedPromptActions.integration.test.tsx`
- [x] `monitors.copy.integration.test.ts` as an untouched-neighbour regression check
- [x] Mutation-tested every new assertion (14 mutations, all caught)
- [x] Manual browser QA re-run against the **post-refactor** code: fresh signup + org + project, seeded `support-bot`, opened the row `⋮` menu (confirmed "Duplicate prompt" sits directly under "Replicate to another project"), duplicated twice, got `support-bot-1` and `support-bot-2` in the sidebar and in Postgres.


langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
```diff
@@ -555,6 +555,109 @@ export const promptsRouter = createTRPCRouter({
...
+
+      afterPromptCreated({
+        prisma: ctx.prisma,
+        projectId: input.projectId,
+        userId: authorId,
+      });
+
+      return duplicatedPrompt;
+    }),
+
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in 5cf1d66 — extracted `generateUniqueHandle` and `buildCreatePromptParamsFromSource` helpers shared between `copy` and `duplicate`, both handlers are now well under the line limit.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
```diff
@@ -555,6 +555,109 @@ export const promptsRouter = createTRPCRouter({
...
+        if (attempts > maxAttempts) {
+          throw new TRPCError({
+            code: "INTERNAL_SERVER_ERROR",
+            message: `Failed to generate unique handle after ${maxAttempts} attempts. Source prompt: ${sourcePrompt.id} (handle: ${
+              sourcePrompt.handle ?? "none"
+            }), project: ${input.projectId}`,
+          });
+        }
+        index++;
+        newHandle = `${baseHandle}(${index})`;
```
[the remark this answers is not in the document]
Aryansharma28: Confirmed and fixed in 5cf1d66 — this was a real bug: the generated `(1)`/`(N)` handle fails `handleSchema` and would've been silently forced into draft mode (`handle: null`) when reopened, per `isHandleValid` in `llmPromptConfigUtils.ts`. Switched to a schema-compliant `-1`/`-N` suffix.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
```diff
@@ -555,6 +604,79 @@ export const promptsRouter = createTRPCRouter({
...
+          projectId: input.projectId,
+          handle: newHandle,
+          authorId,
+          commitMessage: `Duplicated from "${baseHandle}"`,
+          prompt: normalizedPrompt,
+          messages: normalizedMessages,
+        }),
+      );
+
+      afterPromptCreated({
```
sergioestebance, line 470: we can use its own service for this function.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
```diff
@@ -555,6 +604,79 @@ export const promptsRouter = createTRPCRouter({
...
+        baseHandle: `${baseHandle}-1`,
+        // index 1 here means "the base candidate above was taken", so the
+        // next candidate is `-2`, not another `-1`.
+        suffixFn: (i) => `${baseHandle}-${i + 1}`,
+        projectId: input.projectId,
+        scope: sourcePrompt.scope ?? "PROJECT",
+        service,
+      });
+
+      const { normalizedPrompt, normalizedMessages } = normalizePromptData({
```
sergioestebance: we can use its own service for this function.

sergioestebance:
## Ruthless review — REQUEST CHANGES

Core logic checks out: traced `generateUniqueHandle` against the old `copy` loop — candidate sequences are identical, no regression. Tenancy, permissions, and license enforcement on `duplicate` are correct. Blocking items are the test gap and the dead-code extension.

### P2 — zero tests for the new behavior (`prompts.trpc-router.ts`)
The only test change is a mock to stop a crash. The `duplicate` mutation and `generateUniqueHandle` — whose off-by-one contract needs an in-code comment to explain (`base-1` pre-suffixed + `suffixFn(i) = base-${i+1}`) — ship untested. Please add: unit tests for `generateUniqueHandle` (both suffix contracts + max-attempts throw), and a menu-item test in `PublishedPromptActions.integration.test.tsx` asserting the click calls `duplicate` with the right args.

### P2 — 45 lines added to confirmed-dead `PromptsList.tsx`
Zero imports anywhere (the PR admits it). The copy has already drifted — its catch lacks the `isHandledByGlobalHandler` guard the live component has. Either delete `PromptsList.tsx` in this PR or revert the changes to it; don't extend dead code.

### P3 — `generateUniqueHandle` API is awkward (`prompts.trpc-router.ts:646`)
Caller pre-suffixing `baseHandle` and compensating inside `suffixFn` is what forces the comment. A `candidate(i)` generator starting at `i = 0` makes both call sites read literally.

### P3 — null-handle source produces a schema-invalid handle (`prompts.trpc-router.ts:644`)
When `handle` is null, `baseHandle = prompt_<nanoid>` — nanoid includes uppercase, so the generated handle fails `handleSchema` and hits the exact silent-draft-reversion trap this PR's own doc comment describes (`service.createPrompt` never validates format). Null-handle prompts do reach this menu (`getDisplayHandle` renders them as "New Prompt"). Pre-existing in `copy`, but the shared helper is the right place to guard: reject or slugify null-handle sources.

### P3 — screenshots committed into the product repo, branch-pinned URLs
`.gitignore` policy: PR-body screenshots live in the `pr-screenshots` repo, never the product tree. The raw URLs also point at `feat/duplicate-prompt` and will 404 once the branch is deleted post-merge. Drop the PNGs from the diff and rehost.

### P3 — check-then-create race on handle uniqueness
Two concurrent duplicates can both pass `checkHandleUniqueness`. Pre-existing pattern from `copy`, now codified in the shared helper — noting, not blocking.

Not flagged: Duplicate isn't UI-gated on `prompts:create` — consistent with the adjacent Replicate item, fine as-is.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
Aryansharma28: Agreed — done. `normalizePromptData` is now `hoistSystemMessage` in `~/server/prompt-config/hoist-system-message.ts`, and both `duplicate` and `copy` delegate to the service:

- `PromptService.duplicatePrompt()` / `PromptService.copyPrompt()` are the entry points
- `generateUniqueHandle()`, `deriveBaseHandle()` and `buildCreateParamsFromSource()` are private to the service
- all three module-level helpers are gone from the router

Worth recording why `hoistSystemMessage` still exists rather than being deleted outright: `createPrompt` runs `assertNoSystemPromptConflict` *before* normalizing, and `transformToVersionedPrompt` always synthesizes a system message into `messages` (prompt.service.ts:1304). So a prompt read back from storage carries both a `prompt` and a system message, and handing it straight to `createPrompt` throws `SystemPromptConflictError`. It's a load-bearing guard, now shared by `copy`, `duplicate`, `syncFromSource` and `pushToCopies` and covered by its own unit tests.

`duplicate` went from ~77 source lines to 26; `copy` to 40.

langwatch/src/server/api/routers/prompts/prompts.trpc-router.ts
Aryansharma28: This one I'd push back on, though happy to be overruled.

`afterPromptCreated` is an EE billing/nurturing hook, not domain logic, and two things argue for leaving it at the transport layer:

**1. It would point the OSS core at billing.** `src/server/prompt-config/` has zero `ee/` imports today. Moving the hook into `PromptService` makes it the first one. The dependency should run the other way — and it currently does: the hook itself imports *up* into `src/server/app-layer/app.ts` for `getApp().nurturing`.

**2. The hook's own docstring says it's a call-site helper:**

> *"Used at prompt creation call sites (tRPC and REST API) to avoid duplicating the counting and userId resolution logic."*

All five call sites are transport — router `189` (`create`), `421` (`copy`), `465` (`duplicate`), and REST `app.v1.ts` `753`, `875`. Moving only `duplicate`'s would split one concept across two layers. And putting it in `createPrompt` specifically would double-fire on the REST path, which already calls `service.createPrompt()` at `app.v1.ts:712` and `afterPromptCreated()` at `:753` — two org-wide COUNT queries and a duplicate `first_prompt_created` event.

**That said, your instinct is right about the underlying problem.** The reason to want it in the service is so a creation path can't silently forget to fire it — and today nothing stops that. The clean fix is dependency inversion: `PromptService` declares a port it owns (`onPromptCreated?: (e: PromptCreated) => void`), the EE hook is injected as an adapter at the composition root, and `app.v1.ts` stops calling it by hand. Then billing depends on core's interface rather than core depending on billing, and the guarantee you want falls out.

That's a five-call-site change plus wiring an EE port through the composition root, which feels like it deserves its own PR rather than riding along on a Duplicate menu item. Want me to open an issue for it? Or if you'd rather I just do it here, say the word.

Aryansharma28:
Thanks — this was a good review. All six items addressed across `73dbab2`, `aa44715`, `1396100`. Point by point:

### P2 — zero tests → 51 tests, each mutation-verified

- `prompt.service.duplicate-and-copy.unit.test.ts` (15) — both suffix contracts (`-1`/`-2` for duplicate, bare/`_copy1` for copy), the slug fallback, and the max-attempts throw (`HandleGenerationError`, asserting exactly 101 probes so the loop bound can't drift).
- `PublishedPromptActions.integration.test.tsx` (+5) — the menu click calls `duplicate` with `{ idOrHandle, projectId }`, refreshes the list, names the duplicate in the toast, surfaces failures, and stays quiet when the global prompt-limit dialog already told the user.
- `prompts.duplicate.integration.test.ts` (9) and `prompts.copy.integration.test.ts` (8) — real tRPC + Prisma. `copy` had **no** integration coverage at all, which felt like the riskier gap given this PR refactors it.
- `handle-slug` (15), `hoist-system-message` (6), `llm-config.repository.setCopiedFromPrompt` (2).

Every assertion was checked by deliberately breaking the source and confirming the intended test goes red — 14 mutations, all caught. Two of my own tests were passing for the wrong reason and that's how I found them: the `enforceLicenseLimit` assertion was being satisfied by the `prompts.create` call in the test's own setup, so deleting the limit check from `duplicate` left it green.

### P2 — dead `PromptsList.tsx` → reverted

Byte-identical to `main` now. I went with revert over delete to keep this PR's blast radius on duplicate/copy; the file is confirmed dead (no importers, exact-boundary grep) so deleting it is a clean standalone change if you want one.

### P3 — `generateUniqueHandle` API → `candidateFor(i)` from `i = 0`

```ts
// duplicate
candidateFor: (attempt) => `${baseHandle}-${attempt + 1}`,
// copy — the bare handle is free in most target projects, so try it first
candidateFor: (attempt) => (attempt === 0 ? baseHandle : `${baseHandle}_copy${attempt}`),
```

Both call sites read literally and the explanatory comment is gone. It also moved into `PromptService` (see your other comment), so the router no longer owns handle generation at all.

### P3 — null-handle source → slugified

`deriveBaseHandle()` now falls back to `name` and slugifies anything `handleSchema` would reject, via a new `toHandleSlug()`. The slug suite asserts the invariant directly: whatever goes in, what comes out satisfies `handleSchema`.

This also **fixes `copy`**, which had the same bug — one deliberate behavior change, called out in the PR body. Previously a handle-less prompt copied as `prompt_1h5icu8XRkHHbaQlrOgwq` and hit exactly the silent-draft trap you describe.

### P3 — screenshots → dropped

Both PNGs `git rm`'d; the QA section is prose now. I didn't rehost them — the flow is pinned by the integration tests, and I re-ran the browser pass by hand against the refactored code (fresh signup, `⋮` menu, duplicated twice → `support-bot-1`, `support-bot-2` in the sidebar and in Postgres).

### P3 — check-then-create race → filed, not fixed

Agreed it's pre-existing and non-blocking. Filed as #5592 with the repro and the insert-and-retry-on-`P2002` fix. Also filed #5593 for something I hit while writing the license-limit test: **the REST route never calls `enforceLicenseLimit`**, so a free plan can exceed its prompt allowance via the SDK today.

---

One open thread: your `afterPromptCreated` comment. I pushed back there rather than move it — happy to be overruled.

langwatch-agent (commented):
No blocking findings on the current head (`1396100`). I reviewed the duplicate/copy service extraction, tRPC permission/license boundaries, handle generation/slugging, copied-from tenancy write, system-message hoisting, UI action/error handling, and existing review threads.

Validation run locally after refreshing the app workspace from the frozen lockfile, regenerating Prisma, and starting a local Postgres on the configured `localhost:55432`:
- `pnpm exec vitest run src/server/prompt-config/__tests__/handle-slug.unit.test.ts src/server/prompt-config/__tests__/hoist-system-message.unit.test.ts src/server/prompt-config/__tests__/prompt.service.duplicate-and-copy.unit.test.ts src/server/prompt-config/repositories/__tests__/llm-config.repository.setCopiedFromPrompt.unit.test.ts` — 38 passed
- `pnpm exec vitest run -c vitest.pg-integration.config.ts src/server/api/routers/__tests__/prompts.duplicate.integration.test.ts src/server/api/routers/__tests__/prompts.copy.integration.test.ts src/prompts/prompt-playground/components/sidebar/__tests__/PublishedPromptActions.integration.test.tsx` — 28 passed

Residual risk: the PR description correctly calls out the pre-existing check-then-insert handle race; I did not treat that as introduced here.

sergioestebance (approved):

PR #5588, opened 2026-07-09
perf(prompts+sweep): stop tab-label query storms and mount inactive Tabs panels (#5585, #5455)

### LINKED ISSUE #5454: Prompt playground: open tabs never unmount, causing memory growth and sluggishness with many tabs

### LINKED ISSUE #5455: Sweep follow-up: 3 memory/query anti-patterns from #5454 recur in 6 other features

### LINKED ISSUE #5585: Prompt playground: tab-bar labels fire 2 full-prompt queries per open tab, re-storm on window focus
**Follow-up to #5454 / #5456 — the one root-cause item that PR didn't cover.**

#5454's root cause listed five per-tab costs. #5456 fixed items 1–4 (form, editors, chat, version-history fetch) by unmounting inactive tab *content*, and fixed the O(all-tabs) localStorage write. **Item #5 was not addressed** and is not tracked in #5455 (which is about the same patterns in *other* features).

**The gap**

The tab-bar *labels* are always mounted — one per open tab, in the tab strip, deliberately *not* inside the lazy `Tabs.Content` (you want to see every tab's title + unsaved/outdated indicators). Each label runs, per open tab:

- `useHasUnsavedChanges` → `api.prompts.getByIdOrHandle` (full prompt config for the loaded version) — `src/prompts/prompt-playground/hooks/useHasUnsavedChanges.ts`
- `useHasUnsavedChanges` → `api.modelProvider.getResolvedDefault` (deduped across tabs, low impact)
- `useLatestPromptVersion` → `api.prompts.getByIdOrHandle` (full prompt config for the latest version, a *different* query key) — `src/prompts/hooks/useLatestPromptVersion.ts`

All on React Query defaults (`staleTime: 0`, `refetchOnWindowFocus: true`, per `src/utils/api.tsx`). So **N open tabs = up to 2N full-prompt fetches**, and **every browser-window focus re-fires all of them at once** — the exact "compounds with tab count / gets sluggish" behavior #5454 described, still present after #5456 for the label layer.

`useLatestPromptVersion` only consumes `latestPrompt?.version` from a full-prompt payload, so the payload is far heavier than what the badge needs.

**Expected behavior** (from #5454): "Inactive tabs should keep their lightweight tab-bar label (title, unsaved indicator) but not their full content tree." The label should be lightweight — right now it isn't.

**Suggested fix**

- Add `staleTime` + `refetchOnWindowFocus: false` to the three label queries so they don't re-storm on every focus × N tabs. Saves already invalidate `getByIdOrHandle` (`useHandleSavePrompt.ts:88`), so the indicators stay correct.
- Optionally: give `useLatestPromptVersion` a lighter payload (only the latest version number) if a lightweight endpoint exists or is cheap to add.

**Environment:** langwatch app, `/[project]/prompts` (Prompt Playground, multi-tab).

_Derived by diffing #5454's root cause against what #5456 shipped. No code changes here — this issue exists so the follow-up PR has a proper home._

sergioestebance:
# Related Issue

- Resolves #5454

## Why

Follow-up to #5454 / #5456. That PR fixed the prompt-playground tab *content* and the localStorage write amplification, and deliberately scoped out two remaining buckets, tracked as **#5585** (the tab-label query layer) and **#5455** (the same anti-patterns in other features). This PR closes both.

Closes #5585
Closes #5455

## What changed

**#5585 — prompt tab-label queries stop re-storming on window focus.** The tab-bar labels are always mounted (one per open tab, outside the lazy content), and each fires full-prompt `getByIdOrHandle` queries via `useHasUnsavedChanges` and `useLatestPromptVersion`, plus `getResolvedDefault`. On React Query defaults (`staleTime: 0`, `refetchOnWindowFocus: true`) that meant N open tabs re-fired ~2N full-prompt fetches on every window focus. Added `staleTime` + `refetchOnWindowFocus: false` to the three label queries. Saves already invalidate `getByIdOrHandle` (`useHandleSavePrompt`), so the unsaved/outdated indicators stay correct.

**#5455 Pattern 1 — inactive `Tabs.Root` panels now unmount (or lazy-mount).** Six features had Ark UI `Tabs.Root` with no `lazyMount`/`unmountOnExit`, keeping every panel's queries/subtree alive. Applied per the same safety rule learned in #5456 (unmounting a panel with editable/live state loses user work):
- Full `lazyMount unmountOnExit` where panels are query/display-only: `LLMMetrics`, `UserMetrics` (CustomGraph analytics), `DSPyExperiment` (read-only result tables).
- `lazyMount` only (no unmount) where a panel holds editable/live state: `TraceDetails` (Thread tab reaches `AnnotationComment`'s `useForm` draft), `PlaygroundContent` (SpanEditorPanel), `tool-catalog` (Ingestion Templates edit drawers).

**#5455 Pattern 2 — popover queries gated on open.** `SecretsIndicator` (`api.secrets.list` had no `enabled` clause) and `SaveAndRunMenu` (`api.agents.getAll`) now fetch only when their popover opens; both triggers are static. Two candidates deliberately NOT gated because their query data is used outside the popover (gating would break rendering, not just defer a fetch): `TraceOverflowMenu` and `EvaluationStatusItem`.

**#5455 Pattern 3 — held, with reason.** The three zundo undo-history stores: `useWorkflowStore` and `useEvaluationsV3Store` already wrap `handleSet` in a `debounce`, so their `partialize`/`equality`-over-collection only runs on the debounced undo-push, not every `set()` — already mitigated. `usePlaygroundStore` lacks the debounce and is the only real candidate, but adding one changes undo granularity (a behavior change) and there's no evidence it causes jank. Per #5455's own "evaluate before changing" note, this should be profile-driven if a symptom surfaces; not changed blind here. Tracked as #5589.

## Test plan

- `useLatestPromptVersion.test.ts` — added a case asserting the label query is configured with `refetchOnWindowFocus: false`.
- `SecretsIndicator.integration.test.tsx` (existing) — passes; it opens the popover then waits for secrets, so it validates the `enabled: open` gating end-to-end.
- `pnpm typecheck` clean across all 11 files.
- Pattern 1 components have no existing unit tests, and Ark's unmount-on-exit isn't reproducible in jsdom (a test asserting it would be non-discriminating, per #5456), so those rely on typecheck + the established prop pattern. TraceDetails integration tests are Docker/ClickHouse-backed — recommend running in CI.

## Anything surprising?

The sweep's Pattern 1 assumption that all panels were "read-only" didn't hold in two cases — `TraceDetails` (annotation-comment draft) and `PlaygroundContent` (span editor) hold editable state — so those got `lazyMount`-only instead of full unmount, avoiding a repeat of the state-loss class we hit in #5456. And Pattern 3's premise was partly stale: two of the three stores were already debounced.

## Summary by CodeRabbit

* **New Features**
  * Tabs in multiple areas now lazily mount tab content and unmount inactive panels to reduce unnecessary rendering.

* **Bug Fixes**
  * Preserves in-progress drafts and edits when switching tabs.
  * Reduces prompt flicker by tuning query caching and disabling refetch-on-window-focus.
  * Loads secrets and agent lists only after the relevant popover/menu is opened.

* **Tests**
  * Added unit and integration coverage for lazy tab behavior and fetch-on-open query enabling.

### DIFF
--- a/langwatch/src/components/experiments/DSPyExperiment.tsx
+++ b/langwatch/src/components/experiments/DSPyExperiment.tsx
@@ -639,6 +639,11 @@ export const RunDetails = React.memo(
           flexDirection="column"
           minWidth="0"
           colorPalette="blue"
+          // Every panel here (Predictors, Evaluations, LLM Calls) is a
+          // read-only render of the same dspyStep query with no
+          // user-entered state, so fully unmounting inactive tabs is safe.
+          lazyMount
+          unmountOnExit
         >
           <Tabs.List
             position="relative"
--- a/langwatch/src/components/secrets/SecretsIndicator.tsx
+++ b/langwatch/src/components/secrets/SecretsIndicator.tsx
@@ -27,9 +27,14 @@ export function SecretsIndicator({
 	projectId,
 	onInsertSecret,
 }: SecretsIndicatorProps) {
-	const secretsQuery = api.secrets.list.useQuery({ projectId });
-	const secrets = secretsQuery.data ?? [];
 	const [open, setOpen] = useState(false);
+	// Only list secrets once the popover is opened — the trigger is static, so
+	// there's no need to fetch on mount for every editor that renders this.
+	const secretsQuery = api.secrets.list.useQuery(
+		{ projectId },
+		{ enabled: open },
+	);
+	const secrets = secretsQuery.data ?? [];
 
 	const handleSecretClick = (secretName: string) => {
 		onInsertSecret?.(secretName);
--- a/langwatch/src/prompts/hooks/useLatestPromptVersion.ts
+++ b/langwatch/src/prompts/hooks/useLatestPromptVersion.ts
@@ -51,8 +51,14 @@ export const useLatestPromptVersion = ({
     {
       enabled: !!configId && !!project?.id,
       // Runs in every open tab's always-mounted label (for the outdated
-      // badge). A new version is created via a save, which invalidates this
-      // key, so we don't need per-focus refetches for every open tab.
+      // badge). Deliberately save-driven, not focus-live: a save invalidates
+      // this key (see useHandleSavePrompt), so same-app version bumps update
+      // the badge, but we don't re-fetch for every open tab on each window
+      // focus — that was the N-tab storm this fix targets, and matches the
+      // codebase convention for dashboard queries (see useFilterParams).
+      // Tradeoff: a *different session's* new version isn't reflected until
+      // the next save/reload; true cross-session liveness would need a
+      // lightweight version-number endpoint (noted in #5585).
       staleTime: 30_000,
       refetchOnWindowFocus: false,
     },
--- a/langwatch/src/prompts/hooks/useLatestPromptVersion.ts
+++ b/langwatch/src/prompts/hooks/useLatestPromptVersion.ts
@@ -50,6 +50,11 @@ export const useLatestPromptVersion = ({
     },
     {
       enabled: !!configId && !!project?.id,
+      // Runs in every open tab's always-mounted label (for the outdated
+      // badge). A new version is created via a save, which invalidates this
+      // key, so we don't need per-focus refetches for every open tab.
+      staleTime: 30_000,
+      refetchOnWindowFocus: false,
     },
   );
 
--- a/langwatch/src/components/LLMMetrics.tsx
+++ b/langwatch/src/components/LLMMetrics.tsx
@@ -170,7 +170,16 @@ export function LLMMetrics() {
         <GridItem colSpan={2}>
           <Card.Root>
             <Card.Body>
-              <Tabs.Root variant="plain" defaultValue="llmCallsGraph">
+              <Tabs.Root
+                variant="plain"
+                defaultValue="llmCallsGraph"
+                // Every panel here is a read-only CustomGraph analytics
+                // query with no user-entered state, so fully unmounting
+                // inactive tabs is safe and avoids firing/holding queries
+                // for graphs the user never looks at.
+                lazyMount
+                unmountOnExit
+              >
                 <Tabs.List gap={12}>
                   <Tabs.Trigger
                     value="llmCallsGraph"
--- a/langwatch/src/components/analytics/UserMetrics.tsx
+++ b/langwatch/src/components/analytics/UserMetrics.tsx
@@ -122,7 +122,16 @@ export function UserMetrics() {
       <GridItem>
         <Card.Root border="1px solid" borderColor="border.emphasized">
           <Card.Body paddingTop={2}>
-            <Tabs.Root variant="plain" defaultValue="messages">
+            <Tabs.Root
+              variant="plain"
+              defaultValue="messages"
+              // Every panel here is a read-only CustomGraph analytics
+              // query with no user-entered state, so fully unmounting
+              // inactive tabs is safe and avoids holding queries for
+              // graphs the user never looks at.
+              lazyMount
+              unmountOnExit
+            >
               <Tabs.List gap={8}>
                 <Tabs.Trigger
                   value="messages"
--- a/langwatch/src/components/ops/foundry/PlaygroundContent.tsx
+++ b/langwatch/src/components/ops/foundry/PlaygroundContent.tsx
@@ -42,7 +42,21 @@ export function PlaygroundContent({ compact = false }: { compact?: boolean }) {
 
       {/* Right pane — minmax(0,1fr) prevents blowout */}
       <GridItem overflow="hidden" display="flex" flexDirection="column">
-        <Tabs.Root defaultValue="editor" variant="line" size="sm" display="flex" flexDirection="column" flex={1} overflow="hidden">
+        <Tabs.Root
+          defaultValue="editor"
+          variant="line"
+          size="sm"
+          display="flex"
+          flexDirection="column"
+          flex={1}
+          overflow="hidden"
+          // lazyMount only (no unmountOnExit): the Editor tab renders
+          // SpanEditorPanel, whose Input/Textarea fields are live editing
+          // surfaces for the selected span. Unmounting on tab switch risks
+          // losing in-progress edits, so inactive tabs are only skipped
+          // until first opened, then kept mounted.
+          lazyMount
+        >
           <Tabs.List borderBottom="1px solid" borderColor="border" px={3} gap={0} flexShrink={0}>
             <Tabs.Trigger value="editor" fontSize="xs" px={3} py={1.5} color="fg.muted" _selected={{ color: "fg.default", borderColor: "orange.500" }}>
               Editor
--- a/langwatch/src/components/scenarios/SaveAndRunMenu.tsx
+++ b/langwatch/src/components/scenarios/SaveAndRunMenu.tsx
@@ -36,13 +36,16 @@ export function SaveAndRunMenu({
 }: SaveAndRunMenuProps) {
   const { project } = useOrganizationTeamProject();
   const { data: prompts } = useAllPromptsForProject();
-  const { data: agents } = api.agents.getAll.useQuery(
-    { projectId: project?.id ?? "" },
-    { enabled: !!project?.id },
-  );
 
   const [searchValue, setSearchValue] = useState("");
   const [open, setOpen] = useState(false);
+
+  // Agents are only shown inside the popover, and the trigger is static —
+  // don't fetch them until the menu is opened.
+  const { data: agents } = api.agents.getAll.useQuery(
+    { projectId: project?.id ?? "" },
+    { enabled: open && !!project?.id },
+  );
   const inputRef = useRef<HTMLInputElement | null>(null);
   const scrollContainerRef = useRef<HTMLDivElement | null>(null);
 
--- a/langwatch/src/components/scenarios/__tests__/SaveAndRunMenu.integration.test.tsx
+++ b/langwatch/src/components/scenarios/__tests__/SaveAndRunMenu.integration.test.tsx
@@ -1,0 +1,83 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
+import { cleanup, render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { SaveAndRunMenu } from "../SaveAndRunMenu";
+
+vi.mock("~/hooks/useOrganizationTeamProject", () => ({
+  useOrganizationTeamProject: () => ({ project: { id: "project-1" } }),
+}));
+
+vi.mock("~/prompts/hooks/useAllPromptsForProject", () => ({
+  useAllPromptsForProject: () => ({ data: [] }),
+}));
+
+vi.mock("../useFilteredScenarioTargets", () => ({
+  isAgentTarget: () => false,
+  useFilteredAgents: () => [],
+}));
+
+const { mockGetAll } = vi.hoisted(() => ({ mockGetAll: vi.fn() }));
+vi.mock("~/utils/api", () => ({
+  api: {
+    agents: {
+      getAll: {
+        useQuery: mockGetAll,
+      },
+    },
+  },
+}));
+
+const renderMenu = () =>
+  render(
+    <ChakraProvider value={defaultSystem}>
+      <SaveAndRunMenu
+        selectedTarget={{ type: "prompt", id: "p1" } as never}
+        onTargetChange={vi.fn()}
+        onSaveAndRun={vi.fn()}
+        onSaveWithoutRunning={vi.fn()}
+        onCreateAgent={vi.fn()}
+      />
+    </ChakraProvider>,
+  );
+
+describe("SaveAndRunMenu", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockGetAll.mockReturnValue({ data: [] });
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe("given a rendered menu", () => {
+    describe("when the menu is closed", () => {
+      it("does not enable the agents query", () => {
+        renderMenu();
+
+        expect(mockGetAll).toHaveBeenCalledWith(
+          expect.objectContaining({ projectId: "project-1" }),
+          expect.objectContaining({ enabled: false }),
+        );
+      });
+    });
+
+    describe("when the menu is opened", () => {
+      it("enables the agents query", async () => {
+        const user = userEvent.setup();
+        renderMenu();
+
+        await user.click(screen.getByRole("button", { name: /save and run/i }));
+
+        expect(mockGetAll).toHaveBeenLastCalledWith(
+          expect.objectContaining({ projectId: "project-1" }),
+          expect.objectContaining({ enabled: true }),
+        );
+      });
+    });
+  });
+});
--- a/langwatch/src/components/traces/TraceDetails.tsx
+++ b/langwatch/src/components/traces/TraceDetails.tsx
@@ -221,6 +221,14 @@ export function TraceDetails(props: {
         colorPalette="blue"
         display="flex"
         flexDirection="column"
+        // lazyMount only (no unmountOnExit): the Thread (messages) tab
+        // renders Conversation -> TraceMessages -> Annotations ->
+        // AnnotationComment, which holds an in-progress annotation
+        // comment via react-hook-form local state. Unmounting that tab
+        // while the user is mid-comment (e.g. after clicking "Annotate"
+        // and switching to another tab) would destroy the draft, so we
+        // only skip mounting tabs that were never opened.
+        lazyMount
       >
         <VStack
           width="full"
--- a/langwatch/src/pages/settings/governance/tool-catalog.tsx
+++ b/langwatch/src/pages/settings/governance/tool-catalog.tsx
@@ -62,7 +62,17 @@ function ToolCatalogPage() {
           <Spacer />
         </HStack>
 
-        <Tabs.Root variant="line" defaultValue="tool-tiles">
+        <Tabs.Root
+          variant="line"
+          defaultValue="tool-tiles"
+          // lazyMount only (no unmountOnExit): the Ingestion Templates tab
+          // renders drawers (EditOttlDrawer, CreateTemplateDrawer) with
+          // their own local form state (OTTL statements, new-template
+          // fields). Unmounting that tab while a drawer is open would
+          // destroy in-progress edits, so we avoid unmountOnExit for the
+          // whole Root and only skip mounting tabs that were never opened.
+          lazyMount
+        >
           <Tabs.List>
             <Tabs.Trigger
               value="tool-tiles"
--- a/langwatch/src/prompts/hooks/__tests__/useLatestPromptVersion.test.ts
+++ b/langwatch/src/prompts/hooks/__tests__/useLatestPromptVersion.test.ts
@@ -46,6 +46,25 @@ describe("useLatestPromptVersion", () => {
     });
   });
 
+  describe("when configured for the always-mounted tab label", () => {
+    it("disables window-focus refetch so N open tabs don't re-storm", () => {
+      mockUseQuery.mockReturnValue({
+        data: { version: 3 },
+        isLoading: false,
+        isFetching: false,
+      });
+
+      renderHook(() =>
+        useLatestPromptVersion({ configId: "config-123", currentVersion: 3 }),
+      );
+
+      expect(mockUseQuery).toHaveBeenCalledWith(
+        expect.objectContaining({ idOrHandle: "config-123" }),
+        expect.objectContaining({ refetchOnWindowFocus: false }),
+      );
+    });
+  });
+
   describe("when current version matches latest", () => {
     it("returns isOutdated as false", () => {
       mockUseQuery.mockReturnValue({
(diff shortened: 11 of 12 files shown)


langwatch-agent:
### :dog: PR Hound review brief, attention map (not a review, not an approval)

**Reviewer:** @drewdrewthis · **Mode:** `Targeted Review` (query lifecycle across several components) · Resolves #5454, closes #5585 + #5455.

**Scariest risk:** it changes query enable/gating across 12 files to kill the tab-label storm. The failure mode is over-suppression, a query that should run gets gated off, so data silently stops loading.

**Top inspection targets**
- `Must Check`, the shared query-gating change: queries still fire when the data is actually needed, not only when a tab is active-and-focused.
- `Must Check`, `SaveAndRunMenu.tsx` / `PlaygroundContent.tsx`: action-driven refetches still fire on demand.
- `Ask Author`: for any query now tied to tab-active state, does switching to that tab load its data promptly (no stale/empty flash)?

+198/-11, 12 files, CI green, no unresolved threads.

drewdrewthis:
@sergioestebance for any query now tied to tab-active state, does switching to that tab load its data promptly (no stale/empty flash)?

langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** Resolves #5454, Closes #5585 and #5455. Well aligned: these were the two buckets deliberately scoped out of #5456.

**Top inspection targets**
1. `Must Check` - the Tabs mounting change across 12 files. Moving inactive panels to unmounted is a behaviour change, not just a perf one: any component relying on mount-time side effects (subscriptions, form state, autosave) in an inactive tab will now behave differently. `PlaygroundContent.tsx` and `TraceDetails.tsx` are the ones to look at.
2. `Must Check` - `useHasUnsavedChanges` and `useLatestPromptVersion`. If unsaved-change detection depended on the tab label query running, unmounting could mean a user loses the unsaved-work warning.
3. `Probably Fine` - the React Query option changes, assuming they are scoped per-hook rather than global defaults.

**Production risk:** losing an unsaved-changes guard would be user-visible data loss in the prompt playground. Worth confirming explicitly.

1. Does the unsaved-changes warning still fire for a tab that is open but not active?
2. Are the query option changes scoped to these hooks, or do they touch global React Query defaults?

langwatch/src/prompts/hooks/useLatestPromptVersion.ts
```diff
@@ -50,6 +50,17 @@ export const useLatestPromptVersion = ({
...
+      // badge). Deliberately save-driven, not focus-live: a save invalidates
+      // this key (see useHandleSavePrompt), so same-app version bumps update
+      // the badge, but we don't re-fetch for every open tab on each window
+      // focus — that was the N-tab storm this fix targets, and matches the
+      // codebase convention for dashboard queries (see useFilterParams).
+      // Tradeoff: a *different session's* new version isn't reflected until
+      // the next save/reload; true cross-session liveness would need a
+      // lightweight version-number endpoint (noted in #5585).
+      staleTime: 30_000,
+      refetchOnWindowFocus: false,
```
Aryansharma28, line 63: [P2] Shared-hook fix silently weakens cross-session drift detection for two single-instance callers that don't have the storm problem it's fixing.

**Why:** `useLatestPromptVersion` is called from 5 places: `usePromptTabSummary.ts` (the always-mounted-per-tab label — the actual storm source), `TargetHeader.tsx` (also N-per-table, similar rationale), and 3 places that are single-instance per active editing session: `SavePromptButton.tsx`, `PromptEditorDrawer.tsx`, and `useHandleSavePrompt.ts`. Because `refetchOnWindowFocus: false` is hardcoded once inside the hook, all 5 callers lose window-focus refetch — including the 3 that were never part of the N-tab storm and have nothing to gain from it.

That matters because `SavePromptButton.tsx`'s own docstring states the opposite design intent:
```
* Uses the actual latest version from the database, not just current + 1,
* to handle cases where the prompt was updated in another tab/session.
```
Window-focus refetch was the only passive mechanism that made "updated in another tab/session" actually observable without a manual reload — I checked for a rescuing polling/websocket/invalidation path for prompts and found none (`grep -rn "refetchInterval\|invalidate.*getByIdOrHandle\|websocket\|pusher\|subscribe" src/prompts` turns up nothing relevant). With `staleTime: 30_000` and no `refetchOnWindowFocus`/`refetchInterval`, once fetched this query will not refresh again for that session until something else remounts it or a same-session save invalidates it.

**Not data loss** — I checked `llm-config-versions.repository.ts:224` (`const nextVersion = (maxVersion._max.version ?? -1) + 1;`), and the server computes the next version from a fresh `MAX(version)` query at commit time, not from any client-supplied number, and `PromptConfigProvider.tsx`'s `triggerSaveVersion` never forwards `nextVersion` to `updatePrompt` either — it's display-only. So the worst case here is a stale "Update to vX" badge/button, not an overwritten version.

**Suggested fix:** Give `useLatestPromptVersion` a `liveRefetch?: boolean` option (default `true`, keeping today's window-focus behavior for `SavePromptButton`/`PromptEditorDrawer`/`useHandleSavePrompt`), and have only `usePromptTabSummary` (and optionally `TargetHeader`) pass `liveRefetch: false` to opt into the storm-safe caching. That scopes the trade-off to the callers that actually need it instead of applying it hook-wide.

langwatch/src/components/secrets/SecretsIndicator.tsx
```diff
@@ -27,9 +27,14 @@ export function SecretsIndicator({
...
 	onInsertSecret,
 }: SecretsIndicatorProps) {
-	const secretsQuery = api.secrets.list.useQuery({ projectId });
-	const secrets = secretsQuery.data ?? [];
 	const [open, setOpen] = useState(false);
+	// Only list secrets once the popover is opened — the trigger is static, so
+	// there's no need to fetch on mount for every editor that renders this.
+	const secretsQuery = api.secrets.list.useQuery(
+		{ projectId },
+		{ enabled: open },
```
Aryansharma28, line 35: [P2] PR description overclaims test coverage for this exact change — the cited test can't detect whether `enabled: open` gating exists.

**Why:** The PR body says: "`SecretsIndicator.integration.test.tsx` (existing) — passes; it opens the popover then waits for secrets, so it validates the `enabled: open` gating end-to-end." But that test (unmodified by this PR) mocks the query as:
```ts
useQuery: vi.fn(() => ({ data: mockSecrets, isLoading: mockIsLoading }))
```
This ignores whatever options object (including `enabled`) is passed entirely — it always returns `mockSecrets` regardless of `open`. Every assertion in that file (`shows available secrets`, `shows empty state`, etc.) would pass identically whether the query is gated on `open` or fetched eagerly on mount. It exercises the popover UI, not the fetch-gating behavior this PR introduces.

Contrast with the *new* `SaveAndRunMenu.integration.test.tsx` added in this same PR for the same pattern (popover-gated query), which correctly asserts on the call args:
```ts
expect(mockGetAll).toHaveBeenCalledWith(
  expect.objectContaining({ projectId: "project-1" }),
  expect.objectContaining({ enabled: false }),
);
```
That's the right pattern and it's sitting right next to this file in the same diff.

**Risk is low** — `enabled` is a well-established React Query mechanism and the source change itself (`SecretsIndicator.tsx:32-36`) is correct on inspection. This is a coverage/claim-accuracy gap, not a functional bug: a future refactor of this component could silently drop the gating and nothing here would catch it.

**Suggested fix:** Either update the PR description to stop claiming this test validates the gating, or add 2 assertions mirroring `SaveAndRunMenu.integration.test.tsx`'s pattern (`enabled: false` before open, `enabled: true` after click) using a mock that actually reads the options argument.

Aryansharma28 (commented):
## goated-review — PR #5588

**Verdict: sound fix, safe to merge.** The two root causes (label-query storm on window focus; Tabs panels mounted regardless of active tab) are genuinely fixed with the right React Query / Ark UI primitives, not papered over. Two P2s worth a follow-up; nothing blocking.

**P0: 0 | P1: 0 | P2: 2 | P3: 0 | P4: 0 | P5: 3**

### Root-cause verification (the specific thing I was asked to check)

- **The prompt tab-label storm is actually fixed, not masked.** `useLatestPromptVersion`/`useHasUnsavedChanges` get `staleTime` + `refetchOnWindowFocus: false`, and saves still invalidate the same cache key (`useHandleSavePrompt.ts:88` → `utils.prompts.getByIdOrHandle.invalidate({ idOrHandle: prompt.id })`), which tRPC/React Query partial-matches against both the plain query and the `versionId`-scoped one — so the "outdated" badge still updates same-session. Verified, not just asserted.
- **Mounting inactive Tabs panels does *not* introduce new eager fetches.** I checked this at the `@ark-ui/react`/`@zag-js/presence` source level (`use-presence.js`): `wasEverPresent.current` is set to `true` synchronously the moment a panel is `present`, and the *initially active* tab (`defaultValue`) is `present` from the first render — so `unmounted` is always `false` for the default tab regardless of `lazyMount`. Only tabs the user has never clicked defer their mount. This directly answers @drewdrewthis's and the langwatch-agent bot's still-open question ("does switching to that tab load its data promptly, no stale/empty flash?") — yes for the default tab (no change), and first-click-into-a-new-tab shows a normal loading state, which is the intended trade-off of `lazyMount`, not a regression.
- **`unmountOnExit` on `LLMMetrics`/`UserMetrics` won't refetch-storm on repeated tab switching.** Their `CustomGraph` queries already carry `refetchOnMount: false, refetchOnWindowFocus: false` via `useFilterParams`'s shared `queryOpts` (`langwatch/src/hooks/useFilterParams.ts:259-260`) — the exact convention the PR description cites. Remounting after `unmountOnExit` reads from cache, no network call, as long as cache hasn't been GC'd.
- **`DSPyExperiment`'s `unmountOnExit` is a pure win, zero query risk.** `dspyStep` is fetched once in the parent `RunDetails` (above the `Tabs.Root`); the Predictors/Evaluations/LLM Calls panels only render `dspyStep.data.*` — no per-panel `useQuery` exists to re-fire on remount. The inline comment claiming this matches the code exactly.

### P2 — Medium (2)

1. `langwatch/src/prompts/hooks/useLatestPromptVersion.ts:63` — shared hook's `refetchOnWindowFocus: false` silently weakens cross-session drift detection for `SavePromptButton`/`PromptEditorDrawer`, which are single-instance (not part of the N-tab storm) and whose own docstring states the opposite intent. See inline comment for full evidence and a scoped fix.
2. `langwatch/src/components/secrets/SecretsIndicator.tsx:35` — PR description claims the existing `SecretsIndicator.integration.test.tsx` "validates the `enabled: open` gating end-to-end," but that test's mock ignores the query's options argument entirely, so it can't detect whether gating exists. See inline comment.

### P5 — Praise / observations (3)

- `langwatch/src/components/experiments/DSPyExperiment.tsx:645` — the "read-only render of the same `dspyStep` query" comment is verified accurate against the actual data flow (see root-cause verification above). Good discipline — the PR's inline comments consistently match what the code does, which is not always true of perf-fix PRs.
- The PR body's "Anything surprising?" section and the trade-off comments left in `useLatestPromptVersion.ts`/`useHasUnsavedChanges.ts` (e.g., "a different session's new version isn't reflected until the next save/reload") are genuinely useful — they document a deliberate trade-off instead of hiding it. Worth keeping as the house style for perf PRs.
- Cross-PR observation: if the team wants tighter drift detection back for `SavePromptButton`/`PromptEditorDrawer` without reintroducing the N-tab storm, `useLatestPromptVersion` could take an explicit `liveRefetch?: boolean` (default `true`), with only `usePromptTabSummary`/`TargetHeader` passing `false`. Not blocking this PR — noted for the follow-up in #5589 or a new issue.

### Existing findings I'm not re-raising

- **CodeRabbit's 9 "actionable" comments** (on `.claude/skills/perf-prove/SKILL.md` and `.claude/skills/perf-review/SKILL.md`) are **stale and moot** — I confirmed via `git show <head-sha>:.claude/skills/perf-prove/SKILL.md` that neither file exists in the current PR diff or on `main`. That review ran against an earlier revision before files were dropped (rebase/squash). Nothing to act on there.
- **CodeRabbit's "outside diff range" comment** on `useLatestPromptVersion.ts` (stale `// Refetch (e.g., window focus)` comment near the `isFetching` branch, now inaccurate since `refetchOnWindowFocus` is `false`) is still valid and still open — CodeRabbit rated it trivial/low-value itself; not re-raising, just flagging it's unaddressed.
- **@drewdrewthis's / langwatch-agent's open question** ("does switching to that tab load its data promptly?", posted 2026-07-10, never answered) — addressed above under root-cause verification with source-level evidence rather than re-raised as a new finding.

- Real-world frequency of the "two people editing the same prompt handle concurrently, in the same open tab, for >30s without a remount" scenario behind P2 #1 — I confirmed the *mechanism* is broken but don't have usage data on how often it bites in practice.
- `TraceDetails.tsx`/`PlaygroundContent.tsx` integration tests are ClickHouse/Docker-backed per the PR's own test plan; I did not run them, matching the author's own noted limitation.

langwatch/src/components/experiments/DSPyExperiment.tsx
```diff
@@ -639,6 +639,11 @@ export const RunDetails = React.memo(
           flexDirection="column"
           minWidth="0"
           colorPalette="blue"
+          // Every panel here (Predictors, Evaluations, LLM Calls) is a
+          // read-only render of the same dspyStep query with no
+          // user-entered state, so fully unmounting inactive tabs is safe.
+          lazyMount
```
Aryansharma28, line 645: [P5] Verified accurate — good instance of a comment that actually matches the code.

**Why:** I traced `dspyStep` up to `RunDetails` (above this `Tabs.Root`) and confirmed the Predictors/Evaluations/LLM Calls panels only render `dspyStep.data.predictors/.examples/.llm_calls` — there's no `useQuery` anywhere inside the tab panels themselves. So `lazyMount unmountOnExit` here is a pure render-cost win (skips reconciling large result tables for tabs the user never opens) with zero query-refiring risk on remount, exactly as the inline comment claims. Worth calling out since it's easy for "read-only, safe to unmount" comments to be asserted rather than verified — this one holds up.

Aryansharma28 (approved):
Approving following the goated-review pass — root-cause fix confirmed for the query-storm issue, no P0/P1. Two non-blocking P2s already noted inline (refetchOnWindowFocus scope, test claim vs. actual coverage).

PR #5590, opened 2026-07-09
feat(billing): send internal Slack alert on Stripe payment failure

sergioestebance:
## Why

Closes langwatch/langwatch-saas#893

A customer had three failed subscription payments and the team had zero visibility: `handleInvoicePaymentFailed` only stamped `lastPaymentFailedDate` — a field nothing reads — and the first internal signal would have been the subscription deletion when Stripe exhausts dunning, potentially weeks later. Ops needs to know within minutes when a paying customer's payments start bouncing.

## What changed

- New `payment_failed` type in the billing Slack notification union (`sendSlackSubscriptionEvent`), alongside `prospective`/`confirmed`/`cancelled`. The alert carries the organization name, plan, failed amount formatted in the invoice's own currency, the attempt number for the invoice, the previous-failure date when one carried over from an earlier dunning cycle, and buttons to the admin org page and the Stripe subscription. Test-mode events are labelled "(test mode)" in the header and link to the `/test/` dashboard, so they can't be mistaken for a live incident.
- `handleInvoicePaymentFailed` receives the narrowed `Stripe.Invoice` + `event.livemode` from the webhook router, classifies the failure, records it, then fires the alert via a `notifyPaymentFailed` helper whose contract is best-effort: any failure (including the org-name lookup) is logged and never fails the webhook — a 500 would make Stripe retry and double-alert.
- New guard: a late `invoice.payment_failed` on an already-CANCELLED subscription is skipped entirely — previously `recordPaymentFailure` would have flipped CANCELLED→FAILED (mirrors the existing guard in `handleSubscriptionDeleted`).
- Sweep fix: `emitCheckoutAnalytics` (PostHog) was the one unguarded side effect in a retryable webhook branch; now try/caught with the same rationale.
- Spec: `specs/billing/payment-failure-alerting.feature` (11 scenarios). Status semantics on payment failure (ACTIVE stays ACTIVE, PENDING→FAILED) are unchanged and remain covered by the existing webhook spec.

## How it works

- **Repeat vs retry.** Stripe fires one `invoice.payment_failed` per retry attempt, and retries of the same invoice can be days apart — elapsed time can't distinguish a retry from a new failure. The webhook service resolves a typed `PaymentFailurePrior` (`no-prior-failure` | `same-invoice-retry` | `earlier-cycle`) using the invoice's creation date as the cycle boundary; the Slack builder only renders it. Equal timestamps or a missing creation date classify as same-invoice retry, so the alert never claims an earlier cycle it can't prove. `lastPaymentFailedDate` is read **before** `recordPaymentFailure` stamps it, and the copy says "No prior unresolved failure" (not "first failure") because the date resets on successful payment.
- **Amount.** `invoice.amount_due` (`amount_paid` is 0 on a failed invoice) + `invoice.currency` as one `{ cents, currency }` value, formatted by a shared `formatCurrencyCents` helper (also now used by the license-purchase blocks) — no hardcoded USD.

## Test plan

- 98 unit tests passing across `ee/billing/__tests__/webhookService.unit.test.ts` and `ee/billing/__tests__/notification.service.unit.test.ts`, `@scenario`-bound to the feature file. Full evidence table in the PR conversation.
- Covered beyond the scenarios: boundary cases (prior failure at exactly the invoice-creation timestamp; invoice with no creation date), "Unknown" org-name fallback, and a `handleEvent` test proving a throwing PostHog client still yields `{ status: "ok" }` with the error logged.
- `pnpm typecheck` clean.

## Anything surprising?

- Out of scope per the issue: the latent `customer.subscription.updated` `past_due`→instant-cancel branch and surfacing `lastPaymentFailedDate` in the admin UI.
- Sweep found two pre-existing CANCELLED-resurrection gaps (`handleSubscriptionUpdated` active branch, `syncInvoicePaymentSuccess`) and inconsistent notification guarding across handlers — filed as langwatch/langwatch#5591 rather than expanding this PR.
- No alert dedup by design: one Slack message per Stripe retry attempt, labelled with its attempt number.

### DIFF
--- a/langwatch/ee/billing/__tests__/notification.service.unit.test.ts
+++ b/langwatch/ee/billing/__tests__/notification.service.unit.test.ts
@@ -355,9 +355,10 @@ describe("NotificationService", () => {
         organizationId: "org_1",
         organizationName: "Acme",
         plan: "LAUNCH",
-        dbSubscriptionId: "sub_db_1",
+        subscriptionId: "sub_db_1",
         stripeSubscriptionId: "sub_stripe_1",
         livemode: true,
+        priorFailure: { kind: "same-invoice-retry" as const },
       };
 
       beforeEach(() => {
@@ -436,12 +437,30 @@ describe("NotificationService", () => {
         );
       });
 
+      /** @scenario Alert links to the test-mode Stripe dashboard for test-mode events */
+      it("labels the alert header as test mode for test-mode events", async () => {
+        await service.sendSlackSubscriptionEvent({
+          ...basePayload,
+          livemode: false,
+        });
+
+        expect(mockSlackSend).toHaveBeenCalledWith({
+          blocks: expect.arrayContaining([
+            expect.objectContaining({
+              type: "header",
+              text: expect.objectContaining({
+                text: "Subscription payment failed (test mode)",
+              }),
+            }),
+          ]),
+        });
+      });
+
       /** @scenario Alert includes the failed amount in the invoice currency */
       it("shows the amount formatted in the invoice currency", async () => {
         await service.sendSlackSubscriptionEvent({
           ...basePayload,
-          amountDueCents: 3400,
-          currency: "eur",
+          amountDue: { cents: 3400, currency: "eur" },
         });
 
         const sentBlocks = mockSlackSend.mock.calls[0]![0].blocks;
@@ -454,8 +473,7 @@ describe("NotificationService", () => {
       it("omits the amount line when no amount due is provided", async () => {
         await service.sendSlackSubscriptionEvent({
           ...basePayload,
-          amountDueCents: null,
-          currency: null,
+          amountDue: null,
         });
 
         expect(mockSlackSend).toHaveBeenCalled();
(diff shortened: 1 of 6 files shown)


sergioestebance:
### Sweep: terminal-state guards, unguarded webhook side effects, Stripe links, currency formatting

Checked `ee/billing/**` + the Stripe webhook route. 1 fixed in this PR, 2 deferred, rest clean.

**Fixed here:** `emitCheckoutAnalytics` PostHog calls were the one unguarded side effect in a retryable webhook branch — a synchronous throw would have propagated to `handleEvent`'s 500 and made Stripe retry an already-activated checkout. Now try/caught (same rationale as the payment-failed notification guard).

**Deferred (CANCELLED-resurrection family — lifecycle semantics, same bucket as the out-of-scope `past_due` branch of langwatch/langwatch-saas#893):**
- `handleSubscriptionUpdated` active branch (`webhookService.ts:~728`): `updateQuantities` unconditionally writes `status: ACTIVE` with no DB-CANCELLED guard — a `subscription.updated(active)` arriving for a locally-cancelled row flips CANCELLED→ACTIVE.
- `syncInvoicePaymentSuccess` (`webhookService.ts:~857`): `activate` guards only on Stripe's *live* status being `canceled`; in a DB-CANCELLED-but-Stripe-active divergence (e.g. failed `migrateToSeatEvent` Stripe cancel) a late `invoice.payment_succeeded` resurrects the row.

**Clean:** Stripe dashboard links (only the new livemode-aware helper exists), currency formatting (all sites use dynamic `currency` with `Intl.NumberFormat`; only locale is hardcoded), Slack senders (`sendSlackMessage` swallows internally; org-lookup callers that can throw are already wrapped).

PR #5595, opened 2026-07-09
perf(group-queue): run the queue Lua via EVALSHA instead of re-sending 11-23KB scripts per call

0xdeafcafe:
## What

Every GroupQueue operation re-sent its full Lua source per call: STAGE is ~11 KB and DISPATCH ~22 KB once the shared park/water-level/routing helpers are prepended, and they run at four-digit rates. `CachedLuaScript` wraps each script with EVALSHA (40-byte sha instead of the source) plus a one-time EVAL fallback on NOSCRIPT — restart, `SCRIPT FLUSH`, or the first call against a cluster node — which loads the script into that node's cache for every later call.

Applied to all nine `GroupStagingScripts` call sites and the blob-holder release/transfer evals. Key hash-tagging, argument order, and reply handling are unchanged; the sha is derived from the source so an edited script can never execute a stale cached body.

## Why

Prod Redis engine CPU has been trending up and breached 50% on 2026-07-09. A 60-second `INFO commandstats` delta attributes ~33% of engine CPU to `eval` at 1,042 calls/s (~45 µs/call) — most of that parsing and hashing the re-sent script body. EVALSHA removes the per-call source transfer and compile entirely.

## Tests

- New `cachedLuaScript.unit.test.ts`: sha path, NOSCRIPT fallback, script-error propagation (no false fallback), distinct shas per source.
- `src/server/event-sourcing/queues` unit suite: 115 passed. Integration suites (real Redis Lua) run in CI.

https://claude.ai/code/session_01BUi7eLwDkxpZGUgujjPJ8L

## Summary by CodeRabbit

- **Performance**
  - Improved Redis-backed group-queue operations by running Lua logic through cached SHA/EVALSHA where available.
  - Reduced repeated transmission of Lua source during processing.
- **Reliability**
  - Added automatic fallback to direct Lua execution when a cached script isn’t available.
  - Kept existing result mapping and error propagation behavior unchanged.
- **Tests**
  - Added unit tests covering cache hit/miss behavior, error propagation, and protection against stale cache reuse.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it, vi } from "vitest";
+import type { Redis as IORedis } from "ioredis";
+
+import { CachedLuaScript } from "../cachedLuaScript";
+
+function makeRedis({ cacheHit }: { cacheHit: boolean }) {
+  const evalsha = vi.fn(async () => {
+    if (!cacheHit) throw new Error("NOSCRIPT No matching script.");
+    return "sha-result";
+  });
+  const evalFn = vi.fn(async () => "eval-result");
+  return {
+    redis: { evalsha, eval: evalFn } as unknown as IORedis,
+    evalsha,
+    evalFn,
+  };
+}
+
+describe("CachedLuaScript", () => {
+  describe("given the script is already in the server's cache", () => {
+    it("runs via EVALSHA without sending the source", async () => {
+      const { redis, evalsha, evalFn } = makeRedis({ cacheHit: true });
+      const script = new CachedLuaScript("return 1");
+
+      const result = await script.run(redis, 1, "key", "arg");
+
+      expect(result).toBe("sha-result");
+      expect(evalsha).toHaveBeenCalledWith(
+        expect.stringMatching(/^[0-9a-f]{40}$/),
+        1,
+        "key",
+        "arg",
+      );
+      expect(evalFn).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("given an empty script cache (restart / SCRIPT FLUSH / new cluster node)", () => {
+    it("falls back to EVAL once, loading the script for later calls", async () => {
+      const { redis, evalsha, evalFn } = makeRedis({ cacheHit: false });
+      const script = new CachedLuaScript("return 1");
+
+      const result = await script.run(redis, 1, "key", "arg");
+
+      expect(result).toBe("eval-result");
+      expect(evalsha).toHaveBeenCalledOnce();
+      expect(evalFn).toHaveBeenCalledWith("return 1", 1, "key", "arg");
+    });
+  });
+
+  describe("given the script itself errors", () => {
+    it("propagates the error instead of retrying via EVAL", async () => {
+      const evalsha = vi.fn(async () => {
+        throw new Error("ERR user_script:1: attempt to index a nil value");
+      });
+      const evalFn = vi.fn();
+      const script = new CachedLuaScript("return nil.x");
+
+      await expect(
+        script.run({ evalsha, eval: evalFn } as unknown as IORedis, 0),
+      ).rejects.toThrow("attempt to index a nil value");
+      expect(evalFn).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("given two scripts with different sources", () => {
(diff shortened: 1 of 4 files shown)


langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  describe("given two scripts with different sources", () => {
+    it("derives distinct shas so a changed script can never hit a stale cache entry", async () => {
+      const a = makeRedis({ cacheHit: true });
+      const b = makeRedis({ cacheHit: true });
+      await new CachedLuaScript("return 1").run(a.redis, 0);
+      await new CachedLuaScript("return 2").run(b.redis, 0);
+
+      expect(a.evalsha.mock.calls[0]?.[0]).not.toBe(
```
langwatch-agent, line 73: **P1 — Current head fails typecheck.**\n\n`pnpm typecheck` fails on this assertion with `TS2493: Tuple type [] of length 0 has no element at index 0` for both `a.evalsha.mock.calls[0]?.[0]` and `b.evalsha.mock.calls[0]?.[0]`. CI shows the same failure on the PR head, so the PR cannot merge even though the Vitest runtime test passes.\n\nMinimal fix: give the `vi.fn` mocks an explicit callable signature that includes the expected arguments, or capture the SHA arguments via `mock.calls.at(0)` after typing the mock as `Mock<(...args: unknown[]) => Promise<string>>` / equivalent. Then rerun `pnpm typecheck`.\n\nConfidence: high.

langwatch-agent (dismissed):
Found 1 P1 issue: the current head fails typecheck in the new unit test. Runtime unit tests pass, but this is not mergeable until the TS error is fixed; see inline.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts
0xdeafcafe: Fixed in 32f6da8a9. The `evalsha` mock in `makeRedis` now takes a rest signature (`async (..._args: Array<string | number>) => …`) mirroring the real `evalsha(sha, numKeys, ...keysAndArgs)` call, so `mock.calls` is typed as `(string | number)[][]` instead of `[][]` and `mock.calls[0]?.[0]` no longer trips TS2493. Runtime behaviour is unchanged.

langwatch/src/server/event-sourcing/queues/groupQueue/cachedLuaScript.ts
```diff
@@ -0,0 +1,45 @@
...
+    redis: IORedis | Cluster,
+    numKeys: number,
+    ...keysAndArgs: Array<string | number>
+  ): Promise<unknown> {
+    try {
+      return await redis.evalsha(this.sha, numKeys, ...keysAndArgs);
+    } catch (err) {
+      if (err instanceof Error && err.message.includes("NOSCRIPT")) {
+        return await redis.eval(this.source, numKeys, ...keysAndArgs);
+      }
```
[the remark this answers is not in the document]
0xdeafcafe: Good catch — tightened in 66387b99a to `err.message.startsWith("NOSCRIPT")`. Redis' reply is always `NOSCRIPT No matching script. Please use EVAL.`, so the prefix match still handles the real miss while no longer misfiring on a script-level error that happens to mention the token.

langwatch-agent (commented):
Re-review complete: no remaining findings from me. The previous typecheck blocker is fixed; I verified locally with `pnpm typecheck` and `pnpm exec vitest run src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts`. CI typecheck is green; broader CI was still finishing when checked.

langwatch-agent (commented):
Found no blocking issues.

I checked the EVALSHA wrapper and the group-queue/blob-holder call sites. The SHA is derived from the script source, `NOSCRIPT` falls back to `EVAL` for the same keys/args, and the migrated calls preserve `numKeys` and key ordering, so cluster slot routing remains the same as the previous `EVAL` calls.

Validation run:
- `pnpm --dir langwatch exec vitest run src/server/event-sourcing/queues/groupQueue/__tests__/cachedLuaScript.unit.test.ts` → 4 passed
- `LANGWATCH_ENDPOINT=http://localhost:5560 pnpm --dir langwatch exec vitest run -c vitest.integration.config.ts src/server/event-sourcing/queues/groupQueue/__tests__/scripts.integration.test.ts src/server/event-sourcing/queues/groupQueue/__tests__/blobHolders.integration.test.ts` → 155 passed

PR #5596, opened 2026-07-09
perf(observability): stop the anomaly baseline reads from dominating Redis CPU

0xdeafcafe:
## What

Two compounding costs in the anomaly detector's 1-minute tick, together measured as the **largest single Redis engine-CPU consumer** on prod (2026-07-09): `hmget` at 29 calls/s × ~1.8 ms/call ≈ 37% of engine CPU.

1. **`perMinuteSeries` enumerated all 10,080 minutes of the 7-day window as explicit HMGET fields**, so Redis paid a lookup per requested field regardless of how sparse the tenant's hash was. It now HGETALLs the hash — which holds one field per *active* minute — and filters to the window client-side, so cost scales with the tenant's actual traffic. To keep that reply bounded, the minute hash is trimmed to the retention window on **both write and read**: `record()` HDELs the field aging out each tick, and `perMinuteSeries()` drops any field older than the window from the reply (which also self-heals hashes that grew unbounded before this change, and cleans up sparse-gap orphans). The key's 8-day TTL alone does *not* bound the hash — it is refreshed on every write, so without the trim old minute fields would accumulate for the tenant's lifetime.

2. **Quiet tenants re-paid the full read every tick, forever.** A tenant with under an hour of activity deliberately never cached a baseline ("re-check soon"), which at fleet scale means every silent tenant re-reads its series every minute. The not-enough-data verdict is now cached for 10 minutes — `0` reads back through the existing below-floor skip path — so a ramping tenant still gets its first baseline within minutes while the fleet's quiet tenants cost one HGET per tick.

## Spec

`specs/event-sourcing/anomaly-detection.feature`: the baseline-cache scenario now states the brief-cache behaviour ("Insufficient history is cached briefly so quiet tenants are not re-read every tick"), with the incident rationale in a comment. Feature-parity gate passes (2,403 scenarios bound).

## Tests

- `perMinuteSeries` window-filtering (old fields in the HGETALL reply are dropped, silent minutes zero-padded)
- `perMinuteSeries` trims fields past the retention window (bounds a hash that grew unbounded); `record()` drops the field aging out on each write
- `setCachedBaseline` (named params) honours the caller-provided TTL
- detector caches the verdict with `INSUFFICIENT_DATA_RECHECK_SECONDS` and that constant stays well under the baseline TTL
- observability suite: 38 passed

https://claude.ai/code/session_01BUi7eLwDkxpZGUgujjPJ8L

## Summary by CodeRabbit

- **Performance**
  - Reduced repeated history reads during anomaly detection by caching short-lived “insufficient data” outcomes and trimming per-tenant minute history to a fixed retention window.
- **Anomaly Detection**
  - Tenants below the minimum activity threshold now get a cached “not-enough-data” verdict with a brief recheck interval.
  - Baseline cache entries now support configurable expiration periods.
- **Tests**
  - Updated and extended unit tests to cover baseline TTL handling and minute-series retention/cleanup behavior.

### DIFF
--- a/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
+++ b/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
@@ -203,7 +203,7 @@ describe("TenantRateTracker", () => {
     it("round-trips a numeric baseline through Redis", async () => {
       const redis = fakeRedis();
       const tracker = new TenantRateTracker(redis, nowFn);
-      await tracker.setCachedBaseline("proj_acme", 42.5);
+      await tracker.setCachedBaseline({ tenantId: "proj_acme", baseline: 42.5 });
       expect(await tracker.getCachedBaseline("proj_acme")).toBeCloseTo(42.5);
     });
 
@@ -356,7 +356,11 @@ describe("TenantRateTracker", () => {
     };
     const tracker = new TenantRateTracker(redis, () => now);
 
-    await tracker.setCachedBaseline("proj_acme", 0, 600);
+    await tracker.setCachedBaseline({
+      tenantId: "proj_acme",
+      baseline: 0,
+      ttlSeconds: 600,
+    });
 
     expect(setCalls[0]).toEqual([
       "obs:tenant_rate:baseline:proj_acme",
--- a/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
+++ b/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
@@ -82,6 +82,12 @@ function fakeRedis() {
     async hmget(key: string, ...fields: string[]) {
       return ops.hmget(key, ...fields);
     },
+    async hgetall(key: string) {
+      const h = hashes.get(key);
+      const out: Record<string, string> = {};
+      if (h) for (const [f, v] of h) out[f] = String(v);
+      return out;
+    },
     async smembers(key: string) {
       return ops.smembers(key);
     },
@@ -264,5 +270,42 @@ describe("TenantRateTracker", () => {
 
     const series = await tracker.perMinuteSeries("proj_acme", 180);
     expect(series).toEqual([1, 2, 3]);
+  });
+
+  it("perMinuteSeries ignores activity outside the lookback window", async () => {
+    const redis = fakeRedis();
+    const tracker = new TenantRateTracker(redis, () => now);
+
+    // Two minutes of old traffic, then a 3-minute gap, then fresh traffic.
+    await tracker.record("proj_acme", 7);
+    now += 60_000;
+    await tracker.record("proj_acme", 8);
+    now += 4 * 60_000;
+    await tracker.record("proj_acme", 9);
+
+    // 3-minute window: the HGETALL reply still contains the old fields, but
+    // they must be dropped and the silent minutes zero-padded.
+    const series = await tracker.perMinuteSeries("proj_acme", 180);
+    expect(series).toEqual([0, 0, 9]);
+  });
+
+  it("setCachedBaseline honours a caller-provided TTL", async () => {
+    const setCalls: unknown[][] = [];
+    const redis = fakeRedis();
+    const originalSet = redis.set.bind(redis);
+    redis.set = async (...args: unknown[]) => {
+      setCalls.push(args);
+      return originalSet(...(args as [string, string]));
+    };
+    const tracker = new TenantRateTracker(redis, () => now);
+
+    await tracker.setCachedBaseline("proj_acme", 0, 600);
+
+    expect(setCalls[0]).toEqual([
+      "obs:tenant_rate:baseline:proj_acme",
+      "0",
+      "EX",
+      600,
+    ]);
   });
 });
--- a/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
+++ b/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
@@ -295,7 +295,7 @@ describe("TenantRateTracker", () => {
     const originalSet = redis.set.bind(redis);
     redis.set = async (...args: unknown[]) => {
       setCalls.push(args);
-      return originalSet(...(args as [string, string]));
+      return originalSet(...args);
     };
     const tracker = new TenantRateTracker(redis, () => now);
 
--- a/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
+++ b/langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
@@ -54,6 +54,11 @@ function fakeRedis() {
     smembers(key: string) {
       return Array.from(sets.get(key) ?? new Set<string>());
     },
+    hdel(key: string, ...fields: string[]) {
+      const h = hashes.get(key);
+      if (!h) return;
+      for (const f of fields) h.delete(f);
+    },
   };
 
   return {
@@ -70,6 +75,10 @@ function fakeRedis() {
         },
         sadd(key: string, member: string) {
           queued.push(() => ops.sadd(key, member));
+          return pipe;
+        },
+        hdel(key: string, ...fields: string[]) {
+          queued.push(() => ops.hdel(key, ...fields));
           return pipe;
         },
         async exec() {
@@ -90,6 +99,9 @@ function fakeRedis() {
     },
     async smembers(key: string) {
       return ops.smembers(key);
+    },
+    async hdel(key: string, ...fields: string[]) {
+      return ops.hdel(key, ...fields);
     },
     async get(key: string) {
       return strings.get(key) ?? null;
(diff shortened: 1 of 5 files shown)


langwatch/src/server/observability/tenantRateTracker.ts
```diff
@@ -145,23 +145,37 @@ export class TenantRateTracker {
...
     const minutesBack = Math.max(1, Math.ceil(lookbackSeconds / 60));
-    const fields: string[] = [];
-    for (let i = minutesBack - 1; i >= 0; i--) {
-      fields.push(String(minuteNow - i));
-    }
+    const oldestMinute = minuteNow - (minutesBack - 1);
     const key = `${TenantRateTracker.KEY_PREFIX}${tenantId}`;
-    const values = await this.redis.hmget(key, ...fields);
-    return values.map((v) => (v ? Number.parseInt(v, 10) || 0 : 0));
+    const entries = await this.redis.hgetall(key);
```
langwatch-agent, line 191: **P1 — HGETALL becomes unbounded for continuously-active tenants.**\n\nThis change relies on the tenant hash being bounded by the 8-day key TTL, but Redis expires the whole hash key, not individual hash fields. `record()` refreshes that key TTL on every write (`expire(key, TTL_SECONDS)`), and there is no `HDEL`/trim path for old minute fields. A tenant that stays active for weeks or months will therefore accumulate every historical minute bucket forever, and this `HGETALL` will fetch all of them on every cold baseline read.\n\nFailure scenario: a high-volume long-lived tenant has been active continuously for 90 days. The old HMGET path still requested exactly the 10,080 fields in the 7-day window; this path asks Redis to return roughly 129,600 active-minute fields, then filters client-side. That reintroduces the Redis CPU/memory/network spike this PR is trying to remove, and it gets worse with tenant age.\n\nSuggested fix: make the hash truly bounded before switching to HGETALL. The smallest path is to trim stale minute fields on write, e.g. add an `HDEL key String(currentMinute - ttlMinutes)` (or a small bounded cleanup for gaps) in the existing record pipeline, and add a regression test where old fields remain in the hash for an active tenant but `perMinuteSeries`/the stored hash is bounded to the retention window. Alternatively use an explicitly bounded data structure/read path.\n\nConfidence: high.

langwatch-agent (dismissed):
Found 1 P1 issue. Local checks pass (`pnpm typecheck` and the focused observability Vitest files), but the new HGETALL path is not actually bounded for continuously-active tenants because old hash fields are never trimmed; see inline.

specs/event-sourcing/anomaly-detection.feature
```diff
@@ -58,8 +58,12 @@ Feature: Per-tenant rate anomaly detection
...
+  # tick was the dominant Redis engine-CPU cost (2026-07-09). The verdict is
+  # cached briefly — far shorter than the 1h baseline TTL — so a ramping tenant
+  # still gets its first baseline within minutes.
+  Scenario: Insufficient history is cached briefly so quiet tenants are not re-read every tick
     Given tenant "proj_new" has only 3 minutes of activity
     When the AnomalyDetector tick runs
-    Then no baseline is cached for "proj_new"
-    And the tenant is skipped this tick
+    Then the tenant is skipped this tick
+    And the not-enough-data verdict is cached with a short expiry, well under the baseline's
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 1e339891b — the step now reads "…cached with a short expiry, well under the baseline's 1h cache TTL", so it states the comparison explicitly.

langwatch/src/server/observability/__tests__/tenantRateTracker.unit.test.ts
```diff
@@ -265,4 +271,41 @@ describe("TenantRateTracker", () => {
...
+  });
+
+  it("setCachedBaseline honours a caller-provided TTL", async () => {
+    const setCalls: unknown[][] = [];
+    const redis = fakeRedis();
+    const originalSet = redis.set.bind(redis);
+    redis.set = async (...args: unknown[]) => {
+      setCalls.push(args);
+      return originalSet(...(args as [string, string]));
+    };
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in 1e339891b — the wrapper now forwards the full arg list (`originalSet(...args)`) instead of casting to the first two, so the fake stays consistent with the real `set(key, value, "EX", ttl)` shape.

langwatch/src/server/observability/tenantRateTracker.ts
langwatch-agent, line 191: **P1 — HGETALL is still unbounded for continuously-active tenants.**

The new commit only adjusts a test/spec line; production still has the same retention bug. `record()` refreshes the whole hash TTL on every write, and Redis does not expire individual hash fields, so an always-active tenant keeps every historical minute bucket indefinitely. This `HGETALL` then reads all historical fields and filters client-side.

Failure scenario: a tenant stays active for 90 days. The old HMGET path requested exactly the 7-day window fields; this path returns roughly every active minute since tracking began, so the Redis CPU/network/memory cost grows with tenant age and can exceed the fixed-cost path this PR is replacing.

Minimal fix: make the hash truly bounded before relying on HGETALL, e.g. trim stale minute fields in the existing `record()` pipeline (`HDEL`/bounded cleanup for minutes older than retention) and add a regression test proving old fields are removed or cannot make `HGETALL` exceed the retention window. Alternatively keep a bounded read/data structure.

Confidence: high.

langwatch-agent (dismissed):
1 P1 finding remains on the current head. Local `pnpm typecheck` and focused observability Vitest files pass, but the production retention guarantee behind the HGETALL optimization is still not preserved; see inline.

langwatch/src/server/observability/tenantRateTracker.ts
0xdeafcafe: Fixed in 044fdf6e0 — you're right, the key TTL is refreshed on every write so the hash was never actually bounded. Bounded it on both ends:

- `record()` now HDELs the field aging out of the retention window each write (`HDEL key (minute - RETENTION_MINUTES)`, O(1) in the existing pipeline), so a continuously-active tenant can't grow past the window.
- `perMinuteSeries()` trims any field older than the window it already scans from the HGETALL reply (batched at 500, best-effort/swallowed) — this also shrinks hashes that already grew unbounded in prod and cleans up sparse-gap orphans, so it's self-healing rather than only fixing new tenants.

Added two regression tests: one seeds a 30-day-old field and asserts the read leaves the hash bounded to the window; one advances exactly the retention window and asserts the write drops the just-expired field. Full observability unit suite green (tenantRateTracker 21, anomalyDetector 17).

0xdeafcafe: This re-review landed on the interim commit (1e339891b) that only carried the test/spec nits — the production fix came in the next commit, 044fdf6e0. `record()` now HDELs the field aging out of the retention window on every write (O(1), in the existing pipeline), and `perMinuteSeries()` trims any field older than the window from the HGETALL reply (batched, best-effort) — which also self-heals hashes that already grew unbounded in prod. Regression tests cover both the write-time drop and the read-time trim; observability unit suite is green (tenantRateTracker 21, anomalyDetector 17). Resolving as addressed.

langwatch/src/server/observability/tenantRateTracker.ts
```diff
@@ -194,16 +260,22 @@ export class TenantRateTracker {
...
+    ttlSeconds: number = TenantRateTracker.BASELINE_TTL_SECONDS,
+  ): Promise<void> {
     try {
       await this.redis.set(
         `${TenantRateTracker.BASELINE_PREFIX}${tenantId}`,
         baseline.toString(),
         "EX",
-        TenantRateTracker.BASELINE_TTL_SECONDS,
+        ttlSeconds,
       );
```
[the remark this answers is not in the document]
0xdeafcafe: Done in acb8cf054 — `setCachedBaseline` now takes a named-params object (`{ tenantId, baseline, ttlSeconds? }`), so the call site reads `setCachedBaseline({ tenantId, baseline: 0, ttlSeconds: INSUFFICIENT_DATA_RECHECK_SECONDS })` instead of two bare positional numbers. Both call sites and the tests are updated; observability unit suite stays green (38 tests).

langwatch-agent (commented):
Found no blocking issues.

I checked the anomaly detector/tracker changes for minute-window correctness, retention trimming, baseline-cache semantics, and the insufficient-history short cache. The HGETALL path still returns an oldest-first zero-padded series, ignores out-of-window minutes, trims stale fields best-effort, and the cached `0` verdict flows through the existing below-floor skip path without changing anomaly thresholds.

Validation run:
- `pnpm --dir langwatch exec vitest run src/server/observability/__tests__/tenantRateTracker.unit.test.ts src/server/observability/__tests__/anomalyDetector.unit.test.ts` → 38 passed
- `pnpm --dir langwatch exec tsx scripts/check-feature-parity.ts | rg "anomaly-detection|FAIL|unbound|stale|✗" -C 3` → `specs/event-sourcing/anomaly-detection.feature` 6/6 bound

PR #5603, opened 2026-07-10
feat(billing): boundary calendar + gauge fold — tables and pure math (ADR-039 phase 1/4)

sergioestebance:
# Related Issue

Phase 1 of 4 implementing [ADR-039](https://github.com/langwatch/langwatch/pull/5597) (issue langwatch/langwatch-saas#879). Stacked on the ADR PR.

## What

The foundation nothing else can exist without: the five Prisma tables and the pure math of storage billing — no engine, no queues, no ClickHouse client, no Stripe.

- **Schema (additive):** `StorageBoundaryEvent` (signed events, BigInt deltas, unique `dedupKey`), `StorageBillableGauge`, `StorageUsageHourly` (contract kept verbatim from ADR-027), lean `StorageBillingCheckpoint`, `StorageSweepCursor` (durable once-per-hour sweep guarantee). Boundary events + gauge are org-tenancy-guarded from day one.
- **Billable-table set:** explicit 10-table constant (15 retention-managed minus the four derived analytics projections and `evaluation_runs`), column-verified — never derived from the retention map.
- **Boundary calendar:** pure crossing-date math (entry at 35d over a 7-day transit, exits mirrored at retention day).
- **Event identity:** edge-class dedup (ENTRY/SEED shared, EXIT distinct, corrections keyed by cause).
- **Gauge fold:** pure signed fold + repositories.

## Spec

`specs/billing/storage/boundary-calendar-and-gauge-fold.feature` — 12 scenarios, bound via `@scenario` as they land (this PR removes their `@unimplemented` tags).

## Phases

1. **This PR — tables + pure math**
2. Measurement engine, dark (flag-gated, zero Stripe)
3. Seeding + two-layer audits
4. Stripe hourly reporting


langwatch-agent:
### :dog: PR Hound review brief, attention map (not a review, not an approval)

**Reviewer:** @0xdeafcafe · **Mode:** `Deep Review` (billing + DB migration) · Resolves langwatch/langwatch-saas#879, ADR-039 phase 1/4, stacked on ADR PR langwatch/langwatch#5597.

**Scariest risk:** a billing-schema migration whose fold math determines billed amounts. An additive-only slip on the existing billing tables, or an off-by-one on period boundaries, mis-bills silently.

**Top inspection targets**
- `Must Check`, `migrations/20260710120000_.../migration.sql`: strictly additive, no alter/drop on existing billing tables.
- `Must Check`, `boundaryCalendar` / `boundary-event-fold` math: period-boundary handling and the signed `StorageBoundaryEvent` fold.
- `Ask Author`: is base ADR langwatch/langwatch#5597 merging first? Reviewing a stacked PR before its base misreads the diff.

+1079/-12, 18 files, CI green, no unresolved threads.

sergioestebance:
Re the Hound brief's *Ask Author* point: yes — merge order is strictly bottom-up: ADR langwatch/langwatch#5597 first, then this PR, then langwatch/langwatch#5614 → langwatch/langwatch#5642 → langwatch/langwatch#5665, retargeting each as its base merges. Review this diff against its stated base (the ADR branch), not main.

On the two Must Checks: the migration is strictly additive (five new tables, zero ALTER/DROP on existing billing tables — BillingMeterCheckpoint untouched by design, noted in the migration header), and the boundary/fold math is pinned by the unit suite (calendar boundaries incl. the exactly-35 net-zero case and the year-boundary partition test) plus a real-Postgres integration test proving fold(event log) == gauge across entries, a replay, an exit mirror, and a 63→91→63 retention round-trip.

langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref langwatch/langwatch-saas#879 / ADR-039 phase 1

Scariest risk: this is the foundation of the whole ADR-039 billing stack (5 Prisma tables + the pure math). The already-escalated phases (#5614, langwatch/langwatch#5642, langwatch/langwatch#5665) build on it, so a math or schema error here propagates straight to real billing.

Attention targets:
- **Must Check** the pure-math functions have exhaustive unit coverage on edge cases: period boundaries, zero/negative, rounding.
- **Must Check** the 5 tables' constraints/indexes match how the engine queries them (a missing unique constraint would allow double-count rows).
- **Ask Author** are the math functions total (defined for every input), or can they throw / return NaN on unexpected usage data?

Shape note: phase 1 of the 4-PR stack; review this before the stacked phases. Attention map only, not an approval.

sergioestebance:
Answering the Hound briefs' **Ask Author** / **Must Check** points, and noting what changed since the last push.

**Migration is strictly additive** — 5 `CREATE TABLE` + 4 indexes, zero `ALTER`/`DROP`/`UPDATE`/`DELETE` against any existing billing table.

**Constraints cover the double-count risk** — `StorageBoundaryEvent.dedupKey` is unique (replay identity), `StorageUsageHourly` is `@@id([organizationId, sealedHour])` so an org-hour can't be reported twice, `StorageBillableGauge` is keyed by `organizationId`, `StorageBillingCheckpoint` is unique on `(organizationId, billingMonth)`, and the sweep cursor is a singleton row.

**The math functions are total.** `foldBoundaryEvents` is BigInt addition only — no division, no throw path, and BigInt has no NaN. `computeBoundaryCalendar` branches on indefinite (0 = keep forever) and returns no crossings for `retentionDays <= 35` (negatives land safely in that branch); everything else is plain date arithmetic. Edges are pinned by unit tests: the 35-day entry/exit coincide case, indefinite retention, the empty fold, and the deliberately-unclamped negative fold. Rounding isn't in this PR — `max(0, ceil(bytes/MiB))` happens at the sampling boundary in phase 2.

**Merge order** — unchanged and bottom-up: ADR langwatch/langwatch#5597 → this PR → langwatch/langwatch#5614 → langwatch/langwatch#5642 → langwatch/langwatch#5665.

**Changed since the last review pass** (merged the ADR base in):
- The base brought `evaluation_analytics` / `evaluation_analytics_rollup` into the retention map, and the partition guard test caught them as unclassified. They're ADR-034 derived projections of data already billed at its source and carry no `_size_bytes` column, so they follow `trace_analytics` into the excluded set — billable stays at 10. This is the guard doing its job, which is exactly why the set is explicit rather than derived. (0516280)
- Renumbered the migration to `20260712000001` — the base's newest is `20260712000000`, so the original `20260710120000` would have run out of order. Unmerged and undeployed, so the rename is safe. (7bcbc27)

CI: the two red checks are the python-sdk e2e suite hitting live `app.langwatch.ai` and getting Cloudflare 520s / read timeouts. This PR touches no python-sdk code — external infra, unrelated.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this creates five Prisma tables and a migration for a billing system whose engine does not exist yet, and it is stacked on #5597, the ADR that has not been accepted. So the order matters more than the code: land this first and the schema is committed to a design still under discussion. The migration is additive, which makes it cheap to carry and expensive to change once other phases build on it.

| | |
|---|---|
| **Issue** | Resolves `langwatch-saas#879`. Phase 1 of 4 for ADR-039 |
| **Size** | +1082/-12 across 18 files, and a good share is tests |
| **Mode** | Deep, because it is billing and because it is a schema |
| **Blast radius** | The database, plus every later phase that builds on these shapes |

<details>
<summary><b>What you are looking at, if storage billing is not your area</b></summary>

LangWatch charges for retained data. The unit is **GiB-hours**: an organization's billable bytes are sampled once per sealed UTC hour into `StorageUsageHourly`, one row per organization per hour, and a reporting command feeds those rows to a Stripe `sum` meter additively. The invoice line is a plain `SELECT sum(megabytes)` over the period, so the hourly row is the audit trail every invoice traces back to. That reporting protocol is not what this ADR changes.

What it changes is **measurement**, and to see why that is hard you need three facts about the storage underneath.

**Data lives in ClickHouse, partitioned by week, with per-tenant retention.** Thirteen tables are retention-managed across three categories, rows carry a `_retention_days` value, and only some of those tables are billable. Two analytics tables are excluded because they are derived projections of already-billed trace data and billing them would charge twice for the same information. The billable set is therefore written down as an explicit constant rather than derived from the retention map, on the reasoning that deriving it is how a future schema addition silently changes invoices.

**ClickHouse's native TTL deletion is invisible to the application, by design.** There is no row-level or organization-level introspection of what TTL removed; `system.part_log` only reports at part level with no byte attribution. So any design that needs deletion to announce itself is a dead end, and the deletion mechanism itself is not something to hook, delay or replace.

**Which leaves measuring by asking the database.** The superseded approach did exactly that: per sealed hour, sum `_size_bytes` across every billable table for rows past the 35-day cutoff. It is correct and it is O(the organization's entire backlog), every hour, forever, whether or not anything changed. It caused two production OOM incidents, and the mitigations manage the cost of the query shape rather than removing it.

Two local constraints that shape everything downstream: **no cron, ever** (every time-boundary action is triggered by a real event through the existing event-sourcing pipeline and made safe by dedup), and metering is gated to SaaS-billable organizations only, roughly 200 of about 2,000.

The design this implements is the storage-billing ADR in #5597, which is **not on `main` yet**: this PR's base branch is that PR's branch, not `main`, so it cannot merge before it. Read the ADR from #5597 first. Note that its number is contested, `main` already holds two other records claiming 039, so refer to it by title rather than by number.

</details>

<details>
<summary><b>The scoping is the best thing about this PR</b></summary>

Phase 1 is tables and pure math. No engine, no queues, no ClickHouse client, no Stripe. That means every line in it is either a schema declaration or a function you can test by calling it, which is exactly how you want the foundation of a billing system to arrive. The four unit test files and one integration test are the review surface, and they can prove the math on their own.

Splitting a billing rewrite this way is worth saying out loud, because the alternative (one PR with the engine in it) would be unreviewable.
</details>

<details>
<summary><b>The decisions being ratified</b></summary>

1. **The billable-table set is an explicit 10-table constant, deliberately not derived from the retention map.** The description says column-verified, 15 retention-managed minus four derived analytics projections and `evaluation_runs`. This is the single most consequential line in the PR: it decides what customers are charged for. Deriving it would keep it in sync automatically and would also mean a change to retention silently changes the bill. Hard-coding it means a new billable table gets missed until someone notices the revenue is short. Both failure modes are real; the hard-coded one is at least visible in a diff. Worth agreeing with explicitly, and worth a test that fails when the retention map grows a table not classified either way.
2. **`StorageUsageHourly`'s contract is kept verbatim from ADR-027**, the design ADR-039 supersedes. Carrying a shape forward from a superseded design is usually right for compatibility and occasionally carries a mistake across. One look at whether that contract is still the one you want.
3. **Boundary events and gauge are org-tenancy-guarded from day one.** Correct, and much cheaper now than retrofitted.
4. **Edge-class dedup on event identity** with a unique `dedupKey`. The self-audit on #5597 caught a dedup rule that would have discarded every exit event, so this is the area where this design has already proven it can go wrong. Read `boundaryEventIdentity.ts` and its test against that history.
</details>

<details>
<summary><b>Where to spend the time</b></summary>

`boundaryCalendar.ts` (crossing-date math, entry at 35d over a 7-day transit, exits mirrored at retention day) and `gaugeFold.ts` are pure functions. Read their tests first: if the tests state the invariants, the functions are easy. Look specifically for what happens at a boundary that falls on a retention change, and for whether BigInt deltas can underflow the gauge below zero, since a negative billable-bytes gauge is the kind of thing that surfaces as a negative invoice.

The migration SQL is additive, so the rollback question is just whether the tables can be dropped cleanly if the design moves.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Sequence this against #5597: an accepted ADR before its schema is the cheap order.

langwatch/src/server/app-layer/billing/storage/boundaryEventIdentity.ts
```diff
@@ -0,0 +1,109 @@
...
+export function buildDedupKey({
+  projectId,
+  category,
+  partitionKey,
+  sliceDate,
+  retentionDays,
+  edge,
+  causeId,
+}: BoundaryEventIdentity): string {
+  if (CORRECTION_EDGES.has(edge) && !causeId) {
```
drewdrewthis, line 73: **New AC** — `CORRECTION_EDGES` (`DELETION`/`REVERSAL`) is the only thing this function enforces `causeId` for. `ENTRY` is not in that set.

The doc comment on `causeId` (line 33-37) says corrections are "[a]lso set on events re-emitted BY a correction (e.g. the re-emitted entries of a retention change), which keeps them distinct from the originals they replace." That's a re-emitted `ENTRY`, and nothing here throws if its emitter forgets to pass `causeId` — it will silently reuse the original entry's dedup key (`ENTRY`/`SEED` share the "IN" edge class, and neither is in `CORRECTION_EDGES`) and collapse into it on write, exactly the failure this file's own comment says a correction must not do.

The existing `boundary-event-fold.integration.test.ts` 63→91→63 round-trip always passes `causeId` correctly on every emit, so it can't catch an emitter that omits it — the guarantee is caller discipline only, not structural.

This phase (1/4) doesn't yet ship the emitter that would exercise this path, so it's not a live bug today — but the invariant it depends on isn't enforced by this file. Suggest either: (a) add `ENTRY`/`SEED` to a broader "must supply causeId when this is a re-emit" check if the emitter has a way to signal "this is a re-emit," or (b) explicitly track this as a required AC for the phase that adds the re-emitting caller, so it isn't forgotten by the time that caller exists.

drewdrewthis (commented):
Two-channel review: 1 inline finding below (blocking New AC). See the verdict comment for the full CI-state analysis and a non-blocking documentation note.

drewdrewthis:
Run: pr-review (reviewer: @drewdrewthis)

## Review verdict: NOT-READY

Reviewed head `5513b141d081cd2b6f51a1cf85e2e2465e6660d9`. This is ADR-039 phase 1/4 (billing storage boundary calendar + gauge fold, pure math + tables, no emitter/scheduler yet). Two-pass review (line-by-line correctness, then structural) completed on the full diff plus surrounding code (Prisma migration/schema, multitenancy guards, retention schema, existing unit/integration tests).

### CI state (the label says `ci-green`, `mergeStateStatus` is `UNSTABLE` — verified, not assumed)

Enumerated every check row via the Checks + Statuses REST APIs (132 rows) against this exact SHA, not `gh pr checks`. `UNSTABLE` is real but **not caused by this PR**:

- 4 FAILURE rows, all trace to 2 root causes:
  - `e2e` fails at "Setup ClickHouse" — pre-existing bug in an unrelated migration (`00017_create_gateway_budget_ledger.sql`, `AggregatingMergeTree` column missing from sorting key — `BAD_ARGUMENTS`). This PR has zero ClickHouse changes.
  - `test-integration (6)` fails 2 Vitest assertions in `trace-analytics-rollup.integration.test.ts` — a pre-existing ClickHouse rollup test unrelated to this PR's Prisma-only tables.
  - `langwatch-app-complete` / `e2e-complete` are pure aggregator gates failing only because the two jobs above failed — not independent evidence.
- Cross-checked branch protection: `main`'s actually-required checks are `["validate title", "typecheck", "build"]` — all three are `SUCCESS` on the substantive run for this SHA. Later `SKIPPED` entries are expected rebuild-trigger noise on a stacked branch not touching main-relevant paths.
- Confirmed current, not stale: failures completed the same day as the head push (2026-07-29), not leftover from an earlier commit.

**Conclusion:** CI is effectively green for everything this PR is responsible for. The `UNSTABLE` state is pre-existing infra flakiness unrelated to this diff. Do not let that block merge on its own — but it's a real gap between the `ci-green` label and the true mergeStateStatus, worth someone fixing upstream (unrelated ClickHouse migration bug + flaky rollup test).

### Blocking (1 inline thread — resolve to flip to READY)

1. **New AC** — `boundaryEventIdentity.ts:73`, `CORRECTION_EDGES` (`DELETION`/`REVERSAL`) is the only causeId-mandatory check; a re-emitted `ENTRY` (which the file's own doc comment says must carry `causeId` to stay distinct from the original it replaces) is not enforced and will silently collapse into the original's dedup key if the future emitter omits it. Not a live bug in phase 1 (no emitter exists yet to exercise it), but the invariant is caller-discipline-only today, not structural, and untested. https://github.com/langwatch/langwatch/pull/5603#discussion_r3861463131

### Non-blocking

- **Decide** — `billableTables.unit.test.ts:75` and the pre-existing `boundary-calendar-and-gauge-fold.feature` scenario both say "13 retention-managed tables." The live `RETENTION_MANAGED_TABLES` constant (verified against `retentionPolicy.schema.ts` at this SHA) actually has 15 entries. The test assertion itself is dynamically correct (`toEqual([...RETENTION_MANAGED_TABLES].sort())`), so this is a documentation/description defect only, not a functional bug — fix the string when convenient.

### Math correctness (the ask: scrutinize this hard)

`computeBoundaryCalendar`, `foldBoundaryEvents`, and `buildDedupKey` reviewed in full against their unit + integration tests, including boundary cases (retention <35/==35/63/indefinite=0) and an explicit 63→91→63 correction round-trip (`boundary-event-fold.integration.test.ts`) that verifies reconvergence, exact dedup count, and `applied: false` on replay. All logically correct for the paths they exercise. The one gap found is the enforcement-scope issue above (Blocking #1) — everything else pure-math related checked out clean. `$transaction([create, upsert])` in the repository is atomic and correctly scoped (neither op depends on the other's runtime result); P2002-on-dedupKey catch is narrow and safe.

drewdrewthis (changes requested):
NOT-READY: 1 blocking finding. See verdict comment: https://github.com/langwatch/langwatch/pull/5603#issuecomment-5423443786

PR #5600, opened 2026-07-09
feat(billing): derive billing profile in one resolver (ADR-039 rollout 1-4)

sergioestebance:
## What

PR 1 of 3 for ADR-039 (`dev/docs/adr/039-billing-profile-single-resolution.md`, Accepted): derive billing mechanics from the winning plan source in the composite plan provider, instead of branching on the `Organization.pricingModel` column across the platform.

Tracked by langwatch/langwatch-saas#891. PR 2 (precedence flag + typed resolution + alert split) and PR 3 (UI consolidation + plan-constants unification) stack on this.

## Why

Customer incident: paying Growth org hard-blocked at Team Members 6/6 because `pricingModel=TIERED` drifted against an active `GROWTH_SEAT_*` subscription (webhook-ordering skip of `migrateToSeatEvent`). The same drift also excluded the org from the usage-metering population.

## Scope (ADR rollout steps 1-4) — all landed

- [x] Billing-profile derivation module + `PlanInfo.billing`/`capabilities` types
- [x] Metering gate (`getOrganizationForBilling`, `resolveUsageMeter`) derives from the active seat subscription, not the column
- [x] `PlanProviderService` stamps billing profile + capabilities (single derivation point, SaaS + OSS)
- [x] `pricingModel` lazy self-heal: 24h write-once Redis guard, meter-cache invalidation
- [x] License `isTrial` flag (append-only payload field); activation clears ONLY trials; non-trial conflict → Slack ops alert; license/sub coexistence detection (Decision 11, alert-only)
- [x] Checkpoint seeding (`seedIfAbsent`, create-only) + `scripts/seed-billing-checkpoints.ts` — forward-only billing for the drifted cohort (Invariant I8)
- [x] `scripts/backfill-pricing-model.ts` (run ONLY after seeding; shared `findDriftedSeatEventOrgs` predicate)

## Proof

- 29 feature scenarios bound and green (`specs/licensing/`: metering-gate-derivation 4, billing-profile-resolver 17, metering-cutover-checkpoint-seeding 4, pricing-model-backfill 4); `pnpm check:feature-parity` OK (2432 scenarios / 725 files)
- Integration: 8/8 (billing gate vs drifted/cancelled/clean orgs; backfill population + no-billing-change)
- Unit: 2202 green across `src/server/app-layer`, `ee/billing`, `ee/licensing`, billing-reporting pipeline
[...]

PR #5608, opened 2026-07-10
feat(billing): precedence rank flag and typed denial resolution (ADR-039 rollout 5)

sergioestebance:
# Related Issue

- Resolves langwatch/langwatch-saas#891

## What

PR 2 of 3 for ADR-039 (stacked on langwatch/langwatch#5600). Tracked by langwatch/langwatch-saas#891.

- `release_billing_precedence_rank` feature flag (registry): OFF = today's license-always-wins; ON = **ENTERPRISE license > active subscription > non-ENTERPRISE license > free** (Decision 1 & 12). Kill-switch without redeploy for the only behavior change to existing orgs.
- Typed `resolution` (`purchase_seat | upgrade | hard_cap`) on `checkLimit` results and `LimitExceededError`, derived from the plan's memberPolicy (Decision 5, Invariant I3). Forwarded through all five tRPC denial paths (admin invite, non-admin request, approve, lite→full role change, server re-check) in the TRPCError `cause`, and as advisory metadata in the public API 403 body.
- Alert split (Decision 9): `purchase_seat` denials log an info breadcrumb (self-serve funnel — no ops page, no cooldown consumed); `upgrade`/`hard_cap` keep the Slack alert.

## Proof

- 13 scenarios bound (`precedence-flag-flip.feature` 6/6, `member-block-resolution.feature` server half 7/13); parity checker OK (2445/725)
- Unit: 644 green across license-enforcement, ee/billing, subscription app-layer, API middleware
- Rank-chain test: GROWTH license + active seat sub + flag ON → member denial resolves `purchase_seat` (the incident's dead-end, killed)
- 4 new integration tests (invite/approve/role-change denial causes) run in CI — local container runtime was down at push time

## Remaining (PR 3)

UI single handler (`resolution` → modal routing), `useInviteActions` gate collapse, 9 FE `pricingModel` sites → `useActivePlan().billing.*`, plan-constants unification, `seatSyncService` deletion, pending-invite itemization.

## Deployment Impact

- **Helm values / default install:** unchanged — no new services, ports, or config.
- **BYOC dataplane:** unaffected (control-plane/TS only; no gateway or collector changes).
- **Feature flag:** `release_billing_precedence_rank` (PR2) defaults OFF — no behavior change until flipped via PostHog/ops store after the ADR-039 rollout step-3b license audit.
- **Ops scripts (PR1):** `seed-billing-checkpoints.ts` must run before `backfill-pricing-model.ts`; both are manual, dry-run-first. See the ADR Rollout section.

### DIFF
--- a/langwatch/src/server/app-layer/subscription/composite-plan-provider.ts
+++ b/langwatch/src/server/app-layer/subscription/composite-plan-provider.ts
@@ -5,9 +5,15 @@ import type { PlanProvider, PlanProviderUser } from "./plan-provider";
 /**
  * Composite PlanProvider with select-one semantics.
  *
- * Precedence:
+ * Legacy precedence (flag off):
  *   1. License override — if the org has a valid (non-free) license, use it entirely
  *   2. SaaS provider — Stripe subscription-based plan
+ *
+ * Rank precedence (ADR-039 Decision 1, behind `release_billing_precedence_rank`):
+ *   ENTERPRISE license > active subscription > non-ENTERPRISE license > free.
+ *   An active subscription must beat a stale GROWTH/PRO license (the
+ *   seat-flow dead-end class), while a sales-issued ENTERPRISE license must
+ *   survive a leftover self-serve subscription.
  *
  * After selection, `overrideAddingLimitations` is recomputed from the user's
  * impersonation context — it is authorization, not entitlement.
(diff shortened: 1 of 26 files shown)


langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref ADR-039
Scariest risk: precedence-rank flag + typed denial resolution decide which billing rule wins; a precedence bug flips who is allowed or charged.
- **Must Check** the precedence ordering matches ADR-039 exactly (highest rank wins, ties deterministic).
- **Must Check** the typed denial can't be constructed in an allow state by accident.
- **Ask Author** on an unknown/none rank: denied or allowed?
Attention map only.

langwatch/src/server/app-layer/subscription/composite-plan-provider.ts
```diff
@@ -45,6 +63,68 @@ export function createCompositePlanProvider({
...
+  user,
+}: {
+  licensePlan: PlanInfo;
+  saasPlanProvider: PlanProvider;
+  organizationId: string;
+  user?: PlanProviderUser;
+}): Promise<PlanInfo> {
+  const hasValidLicense = !licensePlan.free;
+
+  if (hasValidLicense && licensePlan.type === "ENTERPRISE") {
```
langwatch-agent, line 109: **P1 - Enterprise can silently leave a non-seat paid subscription billing.** This branch deliberately selects the Enterprise license, but it does not make the active subscription visible to the conflict notification. The only configured detector (`pricingModelSelfHeal`) calls `hasActiveSeatEventSubscription`, whose repository query accepts only `GROWTH_SEAT_*`; an ACTIVE PRO, LAUNCH, or ACCELERATE Stripe subscription therefore continues billing while this resolver returns the Enterprise license and emits no alert. ADR-039 Decision 11 requires an alert whenever Enterprise wins while an active paid Stripe subscription exists. Separate the seat-event display-cache heal from conflict detection: check for any active paid subscription in the Enterprise-winning case, alert with its actual plan, and leave the subscription untouched. Add a regression for Enterprise plus an ACTIVE non-seat paid subscription proving exactly one alert and no subscription mutation.

langwatch-agent (changes requested):
Request changes: 1 P1. The precedence ordering matches ADR-039 and an absent or false flag preserves legacy selection; focused exact-head resolver and enforcement tests passed (93). The Enterprise-wins billing-alert invariant is still broken; see inline.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this changes which plan wins when an organization holds more than one entitlement, behind `release_billing_precedence_rank`. OFF keeps today's license-always-wins. ON means ENTERPRISE license beats an active subscription beats a non-ENTERPRISE license beats free. So the flag decides, for real customers, which limits apply to them, and both states are defensible for different orgs. The kill switch is the safety story and it only works if someone knows to reach for it.

| | |
|---|---|
| **Issue** | Resolves `langwatch-saas#891`. PR 2 of 3, stacked on `#5600` |
| **Size** | +878/-38 across 25 files |
| **Mode** | Targeted |
| **Blast radius** | Plan resolution and every limit denial, once flagged on |

<details>
<summary><b>What you are looking at, if billing entitlements are not your directory</b></summary>

An organization can hold entitlements from more than one source: a **license** (issued, often enterprise, sometimes legacy) and a **subscription** (bought, active or not). Something has to decide which one governs when both exist, and today the answer is simply that a license always wins. That is wrong for an org holding an old non-enterprise license and a current paid subscription, which gets the worse of the two.

`checkLimit` is the funnel: it answers whether an action is permitted and raises `LimitExceededError` when not. Its result reaches five separate tRPC denial paths, which is why a change to the shape of that result touches 25 files.

Note the stacking: this is PR 2 of 3 on `#5600`, a different chain from the ADR-039 storage stack in `#5603` / `#5614` / `#5642` / `#5665`. Both are billing and they are not the same series.
</details>

<details>
<summary><b>Two changes, and only one is risky</b></summary>

**The typed `resolution`** (`purchase_seat | upgrade | hard_cap`) is the safe half and a genuine improvement. A denial now says what the user can *do* about it, derived from the plan's `memberPolicy` rather than inferred at each call site. Five denial paths previously each decided how to phrase a dead end; now they carry the same typed answer. Check the derivation is total, that every plan shape maps to exactly one resolution and none falls through to a default.

**The precedence flag** is the risky half. Things to establish:

1. **Which orgs actually change** when it flips. This should be a query someone has run, not an estimate. The population holding both a non-ENTERPRISE license and an active subscription is presumably small and specific, and knowing who they are turns a scary flag into a boring one.
2. **Direction of change per org.** Some get more generous limits, some get less. The second group is the one that generates support tickets, and they should be known before the flip rather than after.
3. **Whether the flag is per-org or global.** A global flip on a rule this customer-visible is a much bigger event than a staged one, and the description reads as a single kill switch.
</details>

> [!NOTE]

langwatch/src/server/license-enforcement/enforcement.middleware.ts
```diff
@@ -75,6 +75,7 @@ export async function enforceLicenseLimit(
           limitType: error.limitType,
           current: error.current,
           max: error.max,
+          resolution: error.resolution,
```
drewdrewthis, line 78: This catch block now forwards `error.resolution` into the ops-notification call, but the `TRPCError` thrown a few lines below (`cause: { limitType, current, max }`) still does **not** include `resolution` — unlike the identical pattern in `organization.ts`, `project.ts`, and `team.ts`, which all add `resolution: error.resolution` to their `cause` block in this same PR.

This middleware (`enforceLicenseLimit`) is the PR description's fifth named denial path ("server re-check"), used broadly for project/workflow/etc. creation limits. As written, any client relying on `err.cause.resolution` (per Decision 5's stated goal — the seat-purchase vs. upgrade vs. hard-cap routing) silently gets `undefined` on this path while every other path gets the typed value. Please add `resolution: error.resolution` to the `cause` object here for consistency with the other four paths, and add a test asserting it (there is currently no unit/integration test covering `enforceLicenseLimit`'s thrown cause shape, unlike the other four paths which each got dedicated resolution assertions in this PR).

drewdrewthis (commented):
Automated review pass (review-clerk). See inline thread + verdict comment.

drewdrewthis (changes requested):
NOT-READY: 1 blocking inline finding (enforcement.middleware.ts missing `resolution` in TRPCError cause). See verdict comment for details.

drewdrewthis:
## Review verdict: NOT-READY

**Reviewed at:** `ffd3d88` · **Run:** pr-review (reviewer: @drewdrewthis)

This PR had zero non-bot review before this pass (the CHANGES_REQUESTED on record was bot-only) — treated as never-reviewed.

Each item below also exists as a resolvable inline thread (linked). Resolve the thread to clear it.

1. **[review-clerk] `langwatch/src/server/license-enforcement/enforcement.middleware.ts:78`** The catch block forwards `error.resolution` into the ops-notification call, but the `TRPCError`'s `cause` object thrown right below it does not include `resolution` — unlike the identical pattern added in this same PR to `organization.ts`, `project.ts`, and `team.ts`. This middleware is the PR description's fifth named denial path ("server re-check"); as written it silently drops the typed resolution the rest of the PR wires through everywhere else, and has no test asserting the cause shape (unlike the other four paths). -> thread: https://github.com/langwatch/langwatch/pull/5608#discussion_r4044043026

None raised.

### CI
All gating check-runs green at this SHA (enumerated via the check-runs REST endpoint, not the combined-status rollup): sdk-python-complete, sdk-javascript-complete, sdk-go-complete, mcp-javascript-complete, langevals-complete, test-unit (1-4), test-integration (1-6), lint, typecheck, e2e/e2e-complete, feature-parity, ci-self-test, ci-scripts-test, docs-complete, clickhouse-serverless-complete, langwatch-app-complete, langwatch-nlp-complete, changes, preflight, evaluate, validate title — no FAILURE or pending/in-progress entries. Unrelated matrix legs report SKIPPED (expected — path-gated).

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is not a merge-readiness signal._

PR #5609, opened 2026-07-10
feat(ui): single denial handler and billing profile consolidation (ADR-039 rollout 6)

sergioestebance:
# Related Issue

- Resolves langwatch/langwatch-saas#891

## What

PR 3 of 3 for ADR-039 (stacked on langwatch/langwatch#5608 ← langwatch/langwatch#5600). Tracked by langwatch/langwatch-saas#891. Closes langwatch/langwatch-saas#891 when merged.

- **The incident regression, killed at the UI layer** (`@regression` scenario green): `useInviteActions`' three-condition seat gate (`activePlanSource === "subscription" && pricingModel === "SEAT_EVENT" && api.subscription`) collapses to `limitInfo.resolution === "purchase_seat"` — the resolution derives server-side from the active seat subscription, so a drifted `pricingModel` column can never dead-end the seat purchase flow again.
- **One denial handler** (Decision 10): `purchase_seat` → seat proration modal; `upgrade` → plan management; `hard_cap` → Contact us (`CONTACT_SALES_URL`). LimitContent CTA is resolution-aware.
- **FE pricingModel reads eliminated**: `shouldShowPlanLimits`, `getUsageDisplay`, usage page, `SubscriptionPage` (`isLegacyTiered` from the resolved plan), `PlansComparisonPage` (prop chain deleted), members page.
- **Plan-constants unification** (Decision 7 / I7): license templates live in `ee/plan-constants.ts`, license + SaaS variants declared side by side with divergences documented. Zero entitlement changes; issued licenses read their signed payload.
- **`seatSyncService` deleted** (zero callers — the resolver self-heal replaced its intent).
- **Seat usage itemized** (Decision 13): members page shows "6 of 6 seats used — 4 members + 2 pending invites" next to the pending-invites table that carries revoke.

## Proof

- All 13 `member-block-resolution.feature` scenarios now bound (incl. the `@regression` incident scenario); parity checker OK (2451/725)
- FE + licensing suites green: 571 (ee/licensing, ee/billing, license components) + 412 (subscription/settings/hooks/stores) + typecheck 0 errors

## Deployment Impact

- **Helm values / default install:** unchanged — no new services, ports, or config.
- **BYOC dataplane:** unaffected (control-plane/TS only; no gateway or collector changes).
- **Feature flag:** `release_billing_precedence_rank` (PR2) defaults OFF — no behavior change until flipped via PostHog/ops store after the ADR-039 rollout step-3b license audit.
- **Ops scripts (PR1):** `seed-billing-checkpoints.ts` must run before `backfill-pricing-model.ts`; both are manual, dry-run-first. See the ADR Rollout section.


Aryansharma28 (commented):
Reviewed the denial-handler consolidation + billing-profile rollout (ADR-039 #6).

**No blockers.** CI is actually green — the "skipping" entries in the checks list are superseded runs from earlier pushes plus path-filtered jobs (docker/helm/go) that correctly no-op on a pure-TS diff. `mergeStateStatus: CLEAN`.

- `useInviteActions` seat-gate collapse (3-condition check → `resolution === "purchase_seat"`) is a real fix, not just moved logic — confirmed server-side that `resolution` comes from `resolveLimitResolution`/`license-limit-guard.ts`, not the raw `pricingModel` column that caused the original drift incident.
- `seatSyncService.ts` deletion is clean — zero remaining callers.
- Test hygiene is good: the `member-block-resolution.feature` scenarios flipped from `@integration @unimplemented` to `@unit` with matching unit tests that actually exercise resolution routing.

**One doc-accuracy nit (non-blocking):** `ee/plan-constants.ts`'s new docstring claims both `ee/billing/planLimits.ts` (SaaS) and `ee/licensing/planTemplates.ts` (license) "now read from this module." Only `planTemplates.ts` actually does — `planLimits.ts` still declares its own values independently and isn't touched by this PR or its stack. Worth softening the docstring or filing the SaaS-side migration as an explicit fast-follow so it doesn't read as already-done.

Also flagging: `LimitContent.tsx`'s `hard_cap` path does a plain `window.open(...)` bypassing the usual modal-close flow — probably fine, just worth a glance.

Not approving outright since nothing has reviewed this yet and I'd like the plan-constants docstring addressed or explicitly deferred first, but no code-correctness blockers found.

langwatch-agent:

**Mode:** Deep Review (billing and seat entitlement)
**Linked issue:** Resolves langwatch/langwatch-saas#891. PR 3 of 3 for ADR-039, stacked on langwatch/langwatch#5608 which came back changes-requested from an earlier escalation. Confirm langwatch/langwatch#5608 landed in its final form before judging this one, since the stack shifts under it.

**Top inspection targets**
1. `Must Check` - the seat gate collapse in `useInviteActions`. Three conditions (`activePlanSource`, `pricingModel`, `api.subscription`) collapse to `limitInfo.resolution === "purchase_seat"`. Confirm the server-side resolution covers every case the three-condition gate did, especially self-hosted and licensed plans where `subscription` is absent. A gap here either blocks invites for paying orgs or opens seat purchase where it should not exist.
2. `Must Check` - `planTemplates.ts` / `plan-constants.ts` / `seatSyncService.ts`. Entitlement constants changing under a consolidation is where silent plan drift enters.
3. `Ask Author` - the `@regression` scenario cited as green. Confirm it exercises the original incident path and not just the new resolution enum.

**Production risk (scariest first):** an entitlement regression that either dead-ends the seat purchase flow again (the original incident) or grants seats without charge. Billing errors are customer-visible and hard to unwind.

1. Does `limitInfo.resolution` resolve correctly for self-hosted and license-key orgs with no Stripe subscription?
2. Is langwatch/langwatch#5608 final, or will this stack be rebased again?
3. Does the regression scenario reproduce the original incident, not just the new code path?

langwatch/src/pages/settings/members.tsx
```diff
@@ -277,7 +277,17 @@ function MembersList({
     <SettingsLayout>
       <VStack gap={6} width="full" align="start">
         <HStack width="full">
-          <Heading>Organization Members</Heading>
+          <VStack align="start" gap={1}>
+            <Heading>Organization Members</Heading>
+            {membersLimit.limitInfo && (
+              <MemberSeatUsageLine
+                memberCount={sortedMembers.length}
+                pendingInviteCount={pendingInvites.data?.length ?? 0}
```
Aryansharma28, line 285: [P2] `MemberSeatUsageLine`'s itemized breakdown will not sum to the seat count it's displayed next to, for any org with lite/EXTERNAL members or pending lite invites.

**Why:** `current`/`max` come from `membersLimit.limitInfo` (`useLicenseEnforcement("members")`), which server-side counts only ADMIN/MEMBER-role users and elevated-permission EXTERNAL users (`license-enforcement.repository.ts:145-148` — explicitly *excludes* plain EXTERNAL/lite members and their invites, which get their own `maxMembersLite` bucket). But `memberCount={sortedMembers.length}` is `organization.members.length` unfiltered — confirmed by the `member.role === "EXTERNAL"` check rendered from the same list at line 354 — and `pendingInviteCount={pendingInvites.data?.length ?? 0}` comes from `getOrganizationPendingInvites`, which filters only by invite `status`, not role (`organization.ts:735-746`, no role predicate).

So an org with, say, 3 full members + 1 lite member + 1 pending full invite + 1 pending lite invite would render "4 of X seats used — 4 members + 2 pending invites" (4+2=6 ≠ 4). That's exactly the failure mode Decision 13's own JSDoc says it's avoiding ("instead of a bare 6/6 that looks wrong to an admin") — a breakdown that doesn't sum to the total next to it is arguably more confusing than the bare number. `MemberSeatUsageLine.unit.test.tsx` only exercises the component in isolation with numbers that happen to already sum correctly (4+2=6), so it doesn't catch the integration mismatch, and there's no existing integration test for `members.tsx`/`MembersList` that would.

**Evidence:**
```tsx
// members.tsx:280-289
<MemberSeatUsageLine
  memberCount={sortedMembers.length}                    // ALL roles
  pendingInviteCount={pendingInvites.data?.length ?? 0}  // ALL roles, only status-filtered
  current={membersLimit.limitInfo.current}               // full-members-only
  max={membersLimit.limitInfo.max}                        // full-members-only
/>
```

**Suggested fix:** Don't re-derive the full-member/lite split on the client from two independently-fetched, unfiltered lists — that reintroduces the same two-places-can-drift shape this whole PR is trying to eliminate elsewhere. Either have the `checkLimit`/`limitInfo` response return the itemized breakdown (member vs. pending-invite counts) alongside `current`/`max` so there's one source of truth, or filter `sortedMembers`/`pendingInvites.data` client-side using the same role/permission classification `getMemberCount` uses server-side (not just `role !== "EXTERNAL"`, since elevated-permission EXTERNAL users also count as full).

Aryansharma28 (commented):
## goated-review — PR #5609

**Verdict: no blockers, safe to merge on the entitlement/seat-limit dimension.** One real display-accuracy bug (P2) worth fixing before or right after merge, one dead import (P4). P0: 0 | P1: 0 | P2: 1 | P3: 0 | P4: 1 | P5: 3

This review focused hardest on the escalation reason: seat entitlement correctness. I traced `useInviteActions`'s collapsed seat gate all the way through `resolveLimitResolution` → `deriveMemberPolicy`/`billing-profile.ts` and `license-limit-guard.ts` (all in the stacked base branch, well unit-tested independently), diffed the plan-template value migration field-by-field, and grepped the full worktree for orphaned references to every deleted/renamed symbol (`seatSyncService`, `GROWTH_TEMPLATE`/`PRO_TEMPLATE`/`ENTERPRISE_TEMPLATE`, `organization?.pricingModel`). Findings below.

### P2 MEDIUM (1)
- `MemberSeatUsageLine`'s itemized breakdown counts don't match the seat total it's displayed next to, for any org with lite/EXTERNAL members or pending lite invites — see inline.

### P4 NIT (1)
- `langwatch/src/components/subscription/SubscriptionPage.tsx:50` — `PricingModel` is still imported from `@prisma/client` but no longer used anywhere in the file (its only use, `organization?.pricingModel === PricingModel.TIERED`, was replaced by `plan?.billing?.isLegacyTiered`). Not in the diff hunk so I can't anchor it inline — dead import, safe to drop. [permalink](https://github.com/langwatch/langwatch/blob/51756c70682864786824aaacbccaa2582d3b8349/langwatch/src/components/subscription/SubscriptionPage.tsx#L50)

### P5 PRAISE / OBSERVATIONS (3)
- **The seat-gate collapse is a genuine fix, not just moved logic.** Verified `limitInfo.resolution === "purchase_seat"` derives server-side from `plan.billing?.memberPolicy ?? "upgrade"` (`license-enforcement.service.ts:18`, `license-limit-guard.ts:46`) — both default to the safe non-seat path when `billing` is undefined (self-hosted/no-subscription orgs), and `expandSeatsMutation` being undefined (OSS builds) also falls through safely. The `@regression @unit` scenario in `member-block-resolution.feature:13-18` reproduces the actual incident shape (`pricingModel "TIERED"` + an ACTIVE seat-event subscription) rather than just exercising the new enum, and the matching test in `useInviteActions.resolution.unit.test.tsx:101-131` asserts on it directly.
- **The plan-template consolidation is a true zero-diff refactor.** Diffed `GROWTH_TEMPLATE`/`PRO_TEMPLATE`/`ENTERPRISE_TEMPLATE` against their `*_LICENSE_TEMPLATE` replacements in `ee/plan-constants.ts` field-by-field — every value is identical, only the declaration site moved. Grepped the whole worktree for the old export names and for `seatSyncService`/`getSeatSyncService` — zero remaining references either way.
- **Still open from an earlier review pass on this PR** (2026-07-13, this head SHA): the `ee/plan-constants.ts` docstring claims `ee/billing/planLimits.ts` (SaaS) "now reads from this module" — confirmed it still doesn't (`grep plan-constants ee/billing/planLimits.ts` is empty). Worth softening the docstring or filing the SaaS-side migration as an explicit fast-follow before it reads as already-done. Also previously flagged: `LimitContent.tsx`'s `hard_cap` path uses a plain `window.open(...)`, confirmed non-issue — it targets the hardcoded `CONTACT_SALES_URL` (`https://langwatch.ai/get-a-demo`, ee/licensing/constants.ts:3), not user input, and is now covered by `LimitContent.unit.test.tsx`.

### Cross-PR note
This PR is stacked on #5608 (`issue5599/precedence-and-resolution`), which currently has `CHANGES_REQUESTED` for a P1 (Enterprise-wins billing-alert gap in `composite-plan-provider.ts`). That P1 is about alerting, not the member-gating path this PR touches, so it doesn't change my verdict here — but flagging that the stack shouldn't be considered fully landed until #5608 clears review.

- `billing.meterUnit`/`showUsageLimits`/`isLegacyTiered`/`memberPolicy` population for every plan-source combination (SaaS subscription, license, self-hosted-no-license) lives entirely in the stacked #5608 branch (`billing-profile.ts`, `composite-plan-provider.ts`). It has its own dedicated unit tests (`billing-profile.unit.test.ts`, `plan-provider-stamping.unit.test.ts`) that I read and found sound, but I'm relying on that PR's own review/tests for full production-workload confidence rather than re-deriving it here.

CI is green (`typecheck`, `test-unit`, `test-integration` ×6, `e2e`, `lint` all pass); the `skipping` entries are path-filtered jobs (docker/helm/go) correctly no-op'ing on this pure-TS diff.

Aryansharma28 (approved):
Approving following the goated-review pass — no P0/P1 found on the billing/entitlement logic (traced end-to-end incl. the regression scenario). One non-blocking P2 already noted inline (member-count itemization mismatch).

langwatch/ee/plan-constants.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { LicensePlanLimits } from "./licensing/types";
+
+/**
+ * Single source of truth for plan definitions (ADR-039 Decision 7 /
+ * Invariant I7).
+ *
+ * SaaS subscription plans (`ee/billing/planLimits.ts`) and self-hosted
+ * license generation templates (`ee/licensing/planTemplates.ts`) previously
+ * declared the "same" plans in two unrelated files, so their values drifted
+ * silently. Both now read from this module. Where the license and SaaS
```
drewdrewthis, line 11: **BLOCKING - `planLimits.ts` does not read from this module, so the "Both now read from this module" claim is false and Invariant I7 is not met on the SaaS half.**

The module doc says the two files "previously declared the 'same' plans in two unrelated files, so their values drifted silently. Both now read from this module." Only one of them does. `ee/licensing/planTemplates.ts` imports from here. `ee/billing/planLimits.ts` at head imports exactly four things, and none is `plan-constants`:

```ts
import { FREE_VISIBILITY_DAYS } from "../licensing/constants";
import type { PlanInfo } from "../licensing/planInfo";
import { PlanTypes, type PlanTypes as PlanType } from "./planTypes";
import { GROWTH_SEAT_PLAN_TYPES } from "./utils/growthSeatEvent";
```

Its plan values are still declared independently. What this PR actually built is a home for the LICENSE templates plus prose describing how the SaaS variants differ from them - which is what the PR body says, accurately ("license templates live in `ee/plan-constants.ts` ... divergences documented"). It is the file's own doc-comment that overstates it into a mechanism.

That gap matters more than an ordinary stale comment because of what the file is for. The next person editing a SaaS plan cap opens `planLimits.ts`, changes a number, and the "deliberately different" note in `plan-constants.ts` - which claims to be the SSOT and to be read by both - silently becomes wrong. The drift this ADR exists to kill is still reachable, and it is now harder to notice, because the file that would reveal it reads as authoritative.

Two ways out, and either is small:

1. Make it true: have `planLimits.ts` import its shared values from `plan-constants.ts`, keeping the genuinely-divergent fields declared here beside the license variant. This is what I7 asks for.
2. Make the doc honest: say that this module holds the license templates and records the SaaS divergences for review, that `planLimits.ts` is still the SaaS declaration site, and that keeping the two in step is currently a convention rather than an import. Add the compile-time or test-time check that would catch drift, or say plainly that there is none.

If (1) is out of scope for a PR already stacked three deep, (2) plus a follow-up issue is a fine answer. What should not ship is the current wording, because it tells a future reader the drift is structurally impossible when it is not.

langwatch/src/pages/settings/members.tsx
drewdrewthis, line 285: **Confirming @Aryansharma28's P2 above - it holds at head, and both halves of the mismatch check out.**

I verified the two sides independently rather than taking the thread:

- `current` comes from `getMemberCount`, whose own doc lists what it counts: ADMIN/MEMBER users, EXTERNAL users holding a custom role with a non-view permission, and qualifying PENDING/WAITING_APPROVAL invites. Plain EXTERNAL/lite members and lite invites are excluded - they are counted separately by `getMembersLiteCount` against `maxMembersLite`.
- `memberCount={sortedMembers.length}` is every org user regardless of role, and `pendingInviteCount` comes from `getOrganizationPendingInvites`, which filters on `status` and expiry but carries no role predicate.

So the breakdown and the total are computed over different populations, and any org with a single lite member or a single lite pending invite renders a line whose parts do not sum to its whole. Lite membership is a first-class bucket in this product, so this is the normal case for a chunk of orgs, not an edge.

One correction to the thread for the record: `getOrganizationPendingInvites` does filter expiry (`OR: [{ expiration: { gt: new Date() } }, { expiration: null }]`), so expiry is not a second divergence axis. Role is the whole of it, which if anything makes the fix narrower.

I would take the first of the two suggested fixes. Having `checkLimit` return the itemized breakdown beside `current`/`max` is the only version that does not re-create, in the seat-usage line, exactly the two-places-can-drift shape this PR deletes from `useInviteActions`, `SubscriptionPage`, and the usage page. Classifying roles client-side would put a second copy of `isFullMember` in the browser, and the server's copy is the one that decides entitlement.

drewdrewthis (changes requested):
Reviewed at `51756c70`. NOT-READY. One pre-existing unresolved thread that I confirmed holds, plus one new blocking finding, both inline.

**The headline change is correct, and I checked it rather than taking it.**

The brief's first Must-Check was whether `limitInfo.resolution === "purchase_seat"` covers every case the three-condition gate did, especially licensed and self-hosted plans where `subscription` is absent. It does, and the derivation is the reason:

```ts
const isSeatEvent =
  plan.planSource === "subscription" && isGrowthSeatEventPlan(plan.type);
```

`deriveBillingProfile` reads the plan TYPE and the resolved source, never the `pricingModel` column. That is what actually kills the incident: an org on a seat plan with a drifted `pricingModel` now resolves `purchase_seat` and gets the proration modal, where before it fell through to a dead-end upgrade modal. `deriveMemberPolicy` returns `hard_cap` for anything self-hosted and for ENTERPRISE licenses, and `upgrade` for a non-ENTERPRISE license on SaaS - so a licensed org reaches the upgrade modal exactly as it did before, not a proration flow with no Stripe subscription behind it. Keeping `&& expandSeatsMutation` in the gate is right as a second belt for builds where the mutation is not compiled in.

I also checked the two guards dropped from `isTieredLegacyPaidPlan`. Old: `isTieredPricingModel && !isDeveloperPlan && !isEnterprisePlan && !isLicenseOverride`. New: `(plan?.billing?.isLegacyTiered ?? false) && !isEnterprisePlan`. Dropping `!isDeveloperPlan` and `!isLicenseOverride` is safe, because `isLegacyTiered` is already `planSource === "subscription" && !isSeatEvent && !plan.free` - free and licensed plans can never set it. `maxSeats` and the annual-plan effect keep their old behaviour. No finding.

`seatSyncService` deletion with zero callers, and deriving `isLegacyTiered` from the resolved plan rather than the column, are the same move as the seat gate and equally right.

**Findings, both inline.**

1. `plan-constants.ts:11` - the module doc claims both declaration sites read from it. `planLimits.ts` does not import it, so Invariant I7 is not met on the SaaS half and the drift the ADR targets remains reachable behind a file that reads as authoritative. Blocking, small either way it is fixed.
2. `members.tsx:285` - @Aryansharma28's P2 holds at head. The itemized breakdown counts all roles; the total counts full members only. Confirmed both sides against `getMemberCount` and `getOrganizationPendingInvites`, with one correction to the thread noted there.

**Not claimed.** No tests run, no app exercised. Static read of the diff, plus `billing-profile.ts`, `license-enforcement.service.ts`, `license-enforcement.repository.ts`, `organization.ts`, `planLimits.ts` and `plan-constants.ts` at head. The 571 + 412 green suites and the parity checker figure are taken as reported. I did not verify the `@regression` scenario runs the original incident path - the feature-file diff only adds tags, so the scenario bodies came from PR2 and were not re-read here; the brief's third item is still open for the author to answer.

**Process note.** This PR targets `issue5599/precedence-and-resolution`, not the default branch, so CodeRabbit skipped it entirely and the only automated read is the low-risk assessment, which says this does not qualify as low risk. It is also PR 3 of 3 on a stack whose middle PR came back changes-requested. Worth confirming langwatch/langwatch#5608 landed in its final form before this one is judged merge-ready, since its base moves underneath it.

drewdrewthis:
## Review verdict: NOT-READY

Reviewed at `51756c70682864786824aaacbccaa2582d3b8349`.

**Two open threads**: one new blocking finding, and @Aryansharma28's P2 which I confirmed holds at head.

1. `plan-constants.ts:11` (**blocking**) - the module doc says the SaaS and license declaration sites "both now read from this module". `ee/billing/planLimits.ts` does not import it; its four imports are `constants`, `planInfo`, `planTypes` and `growthSeatEvent`. Only `planTemplates.ts` reads from here. So Invariant I7 is not met on the SaaS half: the plan values are still declared in two files, and the divergences are recorded as prose in the file that claims to be the source of truth. The PR body describes this accurately; it is the module's own doc-comment that overstates it into a mechanism. Either make `planLimits.ts` import it, or say plainly that keeping the two in step is a convention with no check behind it.
2. `members.tsx:285` (@Aryansharma28, P2) - confirmed. `current` is `getMemberCount` (full members and qualifying invites only); the breakdown is `sortedMembers.length` and an unfiltered pending-invite list. Any org with one lite member or lite invite renders parts that do not sum to the whole. Correction for the record: the invite query does filter expiry, so role is the only divergence axis.

**The headline change is correct and I verified the derivation rather than taking it.** `deriveBillingProfile` computes `isSeatEvent` from `planSource === "subscription" && isGrowthSeatEventPlan(plan.type)` - the plan TYPE, never the `pricingModel` column - which is what actually stops a drifted column dead-ending the seat flow. `deriveMemberPolicy` returns `hard_cap` for self-hosted and ENTERPRISE licenses and `upgrade` for a non-ENTERPRISE license on SaaS, so no licensed org reaches a proration modal with no Stripe behind it. The `&& expandSeatsMutation` belt is right to keep.

The two guards dropped from `isTieredLegacyPaidPlan` are safe: `isLegacyTiered` is already `subscription && !isSeatEvent && !free`, so free and licensed plans can never set it. No finding there.

Not claimed: no tests run, no app exercised. Static read of the diff plus `billing-profile.ts`, `license-enforcement.service.ts`, `license-enforcement.repository.ts`, `organization.ts`, `planLimits.ts`, `plan-constants.ts` at head. Suite counts and the parity figure taken as reported. The `@regression` scenario body was not re-read - the feature diff only adds tags - so the brief's "confirm it exercises the original incident path" item is still open.

This PR targets `issue5599/precedence-and-resolution`, so CodeRabbit skipped it and the only automated read is the low-risk assessment, which says it does not qualify. It is PR 3 of 3 on a stack whose middle PR came back changes-requested; confirm langwatch/langwatch#5608 is final before judging this one merge-ready.

PR #5614, opened 2026-07-10
feat(billing): boundary measurement engine, running dark (ADR-039 phase 2/4)

sergioestebance:
# Related Issue

Phase 2 of 4 implementing [ADR-039](https://github.com/langwatch/langwatch/pull/5597) (issue langwatch/langwatch-saas#879). Stacked on phase 1 (#5603).

## What

The complete measurement engine, running **dark**: flag-gated per org (`release_storage_boundary_metering`, fresh identifier, default OFF = fully inert), SaaS-billable orgs only, **zero Stripe**.

- **Sweep** (`storageSweep.service` + reactor on `orgBillableEventsMeter`): any org's ingest wakes it, every billable org is processed — no cron. Once-per-hour = durable CAS cursor (survives restarts; redundant wake-ups O(1)); process-local `shouldReact` guard keeps redundant wake-ups off Redis; per-org isolation (no poison org).
- **Entry edge** (`boundaryMeasurement.service`): one bounded single-partition query per billable table per day, whole-day cutoff (each slice measured once, completely, ≤1 day late — customer-favorable), cumulative-minus-prior deltas (missed days self-heal), category pre-summing, OOM caps on every query. Partition alignment Sunday-verified against real `toYearWeek` incl. the year boundary; pruning proven by EXPLAIN + a system.tables partition-key alignment test over all 10 tables (caught `event_log`'s toDateTime/toDateTime64 mismatch).
- **Exit edge** (`boundaryExit.service`): query-free mirrors of live-net slice groups; structurally no ClickHouse import.
- **Corrections** (`storageCorrection.service`): deletion/erasure negates recorded groups keyed by cause id; retention change = reverse-then-emit wired into the retroactive-update route, with rollback compensation if the mutation is never submitted. ⚠️ `emitDataDeletion` has **no production caller yet** — no hard-delete/erasure path exists today (deletion is a manual runbook); any future one MUST call it before deleting.
- **Sampling** (`gaugeSampling.service`): hourly rows via fold-to-H replay for gaps, O(1) materialized-gauge fast path steady-state; `max(0, ceil(bytes/MiB))` clamp; drift alarm below −100 MiB (alarms, never auto-corrects).

## Review hardening (adversarial pass, fixes in-branch)

M1: Sunday final-pass dedup collision silently dropped ~1 day/partition (~14% under-bill) → whole-day cutoff model. S1: event_log pruning mismatch → expression fix + DDL alignment test. S3: O(events)-per-hour sampling → O(1) fast path + exits day-grained. S4: uncompensated re-book on failed mutation submission → rollback.

## Spec

`specs/billing/storage/boundary-measurement-engine.feature` — 15/15 scenarios bound (`@unimplemented` removed).

## Phases

1. Tables + pure math (#5603)
2. **This PR — the dark engine**
3. Seeding + two-layer audits
4. Stripe hourly reporting

## Deployment Impact

- **Prisma migrations (2, additive):** `lastEntrySweptDay` column on the sweep-cursor singleton; no changes to existing tables. Applied automatically on deploy; safe with running pods (nullable column, no backfill).
- **New event-sourcing reactor** (`storageSweep` on `orgBillableEventsMeter`): registered in SaaS mode only. Behind `release_storage_boundary_metering` (new PRODUCT flag, **default OFF**) the sweep no-ops per org — zero ClickHouse queries, zero writes. No behavior change until the flag is flipped per org.
- **No new env vars, no chart changes, no breaking API changes.** ClickHouse-native TTL untouched (ADR-039 hard constraint).
- Rollback: revert deploy; the additive column and dark reactor are inert.


langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref langwatch/langwatch-saas#879 / ADR-039 phase 2

Scariest risk: the core boundary measurement engine, running dark (computes but does not bill yet), so any error is latent until it goes live on the money path.

Attention targets:
- **Must Check** boundary math vs the ADR-039 pure-math tables: off-by-one on period boundaries under- or over-counts every customer.
- **Must Check** timezone / period-window alignment for the measurement window.
- **Ask Author** is there a shadow comparison against the current path before this is flipped live?

Shape note: phase 2 of the 4-PR stack (#5603 -> langwatch/langwatch#5614 -> langwatch/langwatch#5642 -> langwatch/langwatch#5665). Attention map only, not an approval.

langwatch/src/server/api/routers/dataRetention.ts
```diff
@@ -197,11 +199,59 @@ export const dataRetentionRouter = createTRPCRouter({
...
+        });
+      }
+      const storageBilling = getApp().storageBilling;
+      const organizationId = storageBilling
+        ? await resolveOrganizationId(input.projectId)
+        : null;
+      const causeId = `retchange_${nanoid()}`;
+      const rebooked =
+        storageBilling && organizationId
+          ? await storageBilling.corrections.emitRetentionChange({
```
langwatch-agent, line 226: [P1] This re-books the billing event log before the ClickHouse `_retention_days` mutation is submitted/applied, but the daily measurement path still groups by the row's current `_retention_days` and has no in-flight guard. If the storage sweep runs after this `emitRetentionChange` and before the mutation has relabeled the rows, `BoundaryMeasurementService` sees the old retention group in ClickHouse while `sumNonExitByPartition` has already netted the old group back to zero via the `REVERSAL`; it then emits a fresh old-retention `ENTRY`. When the mutation later completes, the new re-booked group is already present too, so the gauge is over-counted until the old group's exit date. Minimal fix: make measurement skip projects/categories with an in-flight retroactive retention mutation, or change the prior calculation/state model so re-booked-but-not-yet-mutated groups cannot be measured back in. Add a regression test that runs `emitRetentionChange(63→91)`, then a measurement before mutation completion, and proves the gauge does not grow.

langwatch-agent (changes requested):
1 P1 finding on retention-change/measurement alignment. Focused validation passed locally: boundary measurement, gauge sampling, and sealed-hour unit tests (21 tests). I did not rerun ClickHouse/PG integration locally. Details inline.

langwatch/src/server/api/routers/dataRetention.ts
sergioestebance: Real race, fixed in cf1714cec. `BoundaryMeasurementService` now skips a project whose retroactive `_retention_days` mutation is in flight (one `getMutationProgress`/system.mutations lookup per project per daily sweep) — so the reverse-then-emit re-book can't be measured back in as a phantom old-retention ENTRY while ClickHouse still carries the old label. Under-measures briefly instead of over-counting (customer-favorable, self-heals via cumulative-minus-prior next run). Regression test in boundaryMeasurement.unit.test.ts runs 63→91 then a measurement before the mutation lands and asserts the gauge does not grow (no events, no CH read).

sergioestebance: Follow-up correction (f051c0d44): the recovery claim in the fix commit was too strong. Since the measurement query is cumulative over a whole partition, a skipped day heals only when a later slice of the *same* partition is measured after the mutation lands. The one gap — a partition on its final transit day skipped for the entire in-flight window — is recovered by the phase-3 reference audit (which keeps in-flight-mutation orgs on the daily tier), not by measurement. Still strictly customer-favorable (under, never over). Comment corrected accordingly.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this adds the engine that decides what customers are billed for, and lands it inert. The flag `release_storage_boundary_metering` defaults OFF and there is zero Stripe in this PR, so nothing reaches an invoice. The risk is not this merge, it is that the sweep runs on real ingest traffic the moment somebody flips a per-org flag, and the fold it feeds cannot be re-derived if it is wrong. Read it as the last cheap moment to disagree with the mechanism.

| | |
|---|---|
| **Issue** | Resolves `langwatch-saas#879`. Phase 2 of 4, stacked on `#5603` |
| **Size** | +2953/-138 across 38 files |
| **Mode** | Deep |
| **Blast radius** | None while dark. Every SaaS-billable org once flagged |
<details>

The terrain, in three ideas:

</details>

<details>
<summary><b>Where the review time goes</b></summary>

1. **The once-per-hour guarantee.** It is a durable CAS cursor, not a cron: any org's ingest wakes the sweep. So correctness rests on the CAS surviving restarts and concurrent wake-ups. Ask what happens when two workers wake the same org in the same millisecond, and whether a crash between "cursor advanced" and "events written" loses a window or repeats one. A repeat is survivable if the dedup key holds; a loss is not, because nothing re-derives the gauge.
2. **`shouldReact` is process-local.** That is a cheap optimisation and it does not coordinate across pods. Confirm the durable cursor is the actual guarantee and the local guard is only there to save work, not to prevent double-processing.
3. **Ingest-driven means quiet orgs never sweep.** An org that stops sending data stops waking. If its retention boundaries still need EXIT events, who emits them? This is the failure that shows up as a customer still being billed for data that aged out.
4. **Flag identifier freshness.** The description notes `release_storage_boundary_metering` is a fresh identifier. Good, because a reused flag key inherits whatever state it had somewhere else.
</details>

> [!NOTE]

PR #5639, opened 2026-07-10
feat(prompts): tab switcher for the prompt playground

Aryansharma28:
## Why

Opening several prompts fills the playground tab strip. Each tab claimed its full natural width and refused to shrink, so only about four fit. The rest scrolled off the right edge behind a fade gradient, with nothing to indicate they existed and no way to reach one except to scroll the strip blindly.

**Before** — ten prompts open. Four and a half tabs visible, `onboarding/welco` sliced mid-word, no sign the other five exist:

## What changed

**Tabs now shrink to share the strip.** They flex from a 180px maximum down to an 88px floor as more open, so many more stay on screen at once. Only once they hit the floor does the strip scroll.

**After** — the same ten prompts, all at the 88px floor. Nine tabs visible, each one readable, and a `10 ⌄` chip reporting the count:

### A crowded strip spends its width on names, not close buttons

Narrowing alone bought tabs the user could see but not tell apart. At the floor a tab has room for about three characters once the padding, the unsaved dot and the close button have taken their share, so two different prompts both read `str…`.

The close button is the part worth giving up. Once the strip is crowded, an inactive tab drops it and spends that width on its name instead. It comes back under the pointer, and the active tab always keeps it — the tab you are closing is usually the one you are looking at.

**Hovering an inactive tab** brings its close button back, with every tab still at 88px, so nothing reflows:

Nothing is lost to a keyboard user: the close control has never carried a `tabindex`, so it was never in the tab order to begin with. Tapping a tab on a touch device activates it, and the active tab always keeps its button. Making that control keyboard-reachable means lifting it out of the trigger `<button>` it is currently nested inside — a structural change to `DraggableTabsBrowser`, tracked separately.

### The switcher is the fallback, not the fix

The chip appears **only once the strip has actually run out of room**. While every tab fits, a dropdown would list nothing that wasn't already on screen. At six tabs they relax to 128px, all fit, every tab keeps its close button, and the chip is gone:

Open it and every prompt in that pane is listed, with its unsaved dot and its version when the tab is behind. Picking one activates it and scrolls its tab back into view. A row never offers to close or upgrade — those stay on the tab, so choosing a row can only ever navigate.

The switcher renders **per pane**, deliberately outside the toolbar's `tabbedWindow.id === activeWindowId` guard: a pane you aren't working in still has tabs worth reaching.

### A prompt in a folder shows its own name

A prompt has no folder column — the folder is a prefix on its handle, so `onboarding/welcome` is the prompt `welcome` in the folder `onboarding`. A shrunk tab has room for one of the two, and the useful one is the name. The tab reads **`welcome`** and carries the full handle in its `title` for hover; the switcher row, which has room, shows **`onboarding/welcome`** with the folder muted. That is what tells two prompts both named `welcome` apart.

### No tab is ever hidden

This is the load-bearing decision. `traces-v2` solves the same overflow problem by giving overflowed tabs `display: none` and listing them in a `+N` menu (`LensTabs`, `SpanTabBar`, via `useOverflowVisibility`). We can't do that here.

Prompt tabs are `@dnd-kit` sortables. A hidden element has a zero-size bounding rect, and `horizontalListSortingStrategy` measures every item in its `SortableContext` to compute drop indices. Hiding overflowed tabs would have quietly corrupted drag-to-reorder. The traces-v2 strips get away with it because their tabs aren't draggable.

So the tabs are narrowed, never hidden, and we reuse the `~/components/ui/menu` primitives directly rather than the overflow machinery. **No files outside `prompts/` change.**

### Refactors carried along

- **`usePromptTabSummary`** — one place a tab's title, unsaved state and version are derived. The tab and its switcher row render the same facts from the same source and cannot drift.
- **`shouldShowVersionBadge`** — previously existed *only* as a copy inside `showVersionBadgeLogic.test.ts`, so that test verified a reimplementation rather than shipped code. Extracted to source; the test now imports the real function.
- **`getDisplayHandle`** — existed twice, in `PublishedPromptsList` and `PromptListDrawer`, and both returned `handle.split("/")[1]`, which is wrong for a nested handle like `a/b/c` (a test asserted the bug). Now one `prompts/utils/promptHandle.ts` returning the last segment, alongside a new `getPromptFolder`.
- **`useTabById`** — the `windows.flatMap(…).find(…)` lookup had been written out by hand in three hooks. One selector now.
- **`groupId` → `windowId`** (separate commit, pure rename). `DraggableTabsBrowser` called each split pane a "group" while its own store called it a "window", and `PromptPlaygroundBrowser` sat between them translating. This also frees the word "group", which the codebase already spends on RBAC access groups — and which a future user-facing tab-grouping feature needs back.
- **`PromptTabStrip`** — each pane needs its own scroller ref, and a ref can't be created inside the loop over panes.
- **`data-tab-strip-id`**, not `data-tab-id`: `SyncedChatInput` already owns the latter and `PromptPlaygroundChat` queries it document-wide.

## Test plan

- `specs/prompts/prompt-tab-switcher.feature` — new. **23 `@integration` scenarios, all bound**; `pnpm check:feature-parity` exits 0 (2426 enforced scenarios across the repo).
- `PromptTabSwitcher.integration.test.tsx` — 19 tests. Only `usePromptTabSummary` is mocked; the Ark menu, trigger, rows and pointer interaction are real. Verified the suite bites: inverting the `tabIds.length <= 1` guard fails exactly 2 tests.
- `PromptBrowserTab.integration.test.tsx` — 8 tests, covering the folder rule and the crowded-strip close-button rule. Only the tab's controller is mocked. Verified the suite bites: forcing `showsCloseButton = true` fails exactly 2 tests.
- `PromptTabStrip.integration.test.tsx` — 4 tests pinning the one place `isActive` and `isCrowded` are computed. Verified the suite bites: swapping the two props fails exactly 2 tests, and no other suite notices.
- `promptHandle.test.ts` — the nested-handle case (`a/b/c` → `c`) that the old duplicated implementations got wrong.
- `pnpm typecheck` clean.

**Browser QA**, against the running app at 1440px (the invariants jsdom cannot judge):
- 10 tabs → every tab at the 88px floor, `scrollWidth` 984 > `clientWidth` 750, chip shows, strip scrolls, all ten still mounted.
- Hovering an inactive crowded tab reveals its close button with all widths unchanged at 88px — no layout shift, and no feedback loop with the overflow measurement.
- 6 tabs → each relaxes to 128px, all fit, every tab keeps its close button, chip disappears.

Four scenarios stay `@e2e @unimplemented` rather than becoming green tests that prove nothing: **overflow scrolling**, **drag-reorder**, and the two shrink-geometry scenarios all depend on layout rects jsdom does not compute.

## Coverage, stated plainly

This is not 100% coverage and coverage was not measured. `useIsOverflowing` has no direct test — it reads `scrollWidth`/`clientWidth`, which jsdom reports as 0, so a unit test on it could only assert the mock. The `DraggableTabsBrowser` layout changes (`TAB_MIN_WIDTH`, `TAB_MAX_WIDTH`, `flex="1 1 0"`) are CSS and are covered by the browser QA above, not by a test.

What *is* pinned, and pinned by mutation rather than by assertion alone: the close-button rule, the switcher's visibility guard, and the strip's wiring of both.

## Rejected along the way

**Raising the floor to 120px** so tabs stay readable without touching the close button. It fits 6 tabs where the current design fits 9, and the close button is worth less than three tabs' worth of visibility.

**A dropdown as the primary fix.** It was the original plan. A dropdown that lists what is already on screen is a worse version of looking at the screen; the strip had to fit more tabs first, and only then does a fallback for the remainder earn its place.

## Not covered here

- **`pnpm lint` is broken on `main`**, unrelated to this PR: `package.json` pins Biome `2.5.1` but `biome.json` declares schema `2.4.14` and uses a `noForIn` rule 2.5.1 rejects. Untouched here.
- **Drag-reorder** is the one invariant the shrink could plausibly disturb, and it is the `@e2e @unimplemented` scenario worth a Playwright test first.
- **A keyboard-reachable close control**, which needs the control lifted out of the trigger `<button>`.
- **Tab groups** (named, coloured) are the intended follow-up. The `groupId → windowId` rename is the groundwork.

## Summary by CodeRabbit

* **New Features**
  * Added a crowded-tab prompt switcher that appears only when the tab strip overflows and multiple prompts are open.
  * Tabs now shrink together and stay visible while preserving smooth scroll and drag-to-reorder across panes.
* **Bug Fixes**
  * Improved prompt title rendering (including folder handling and truncation) and correct display of “New Prompt”, unsaved indicators, and version badges.
  * Ensures the correct close-button behavior on crowded inactive tabs (hidden until hover) and keeps active content focused.
* **Tests**
  * Expanded integration coverage for overflow, switcher interactions, metadata, accessibility, and multi-pane behavior.


langwatch-agent (commented):
No blocking findings. I reviewed the prompt tab switcher, tab summary extraction, windowId rename, overflow behavior, prompt handle helpers, and relevant tests. Local validation passed: focused unit tests (20), focused integration tests (31), `pnpm run check:feature-parity`, and `pnpm run typecheck`. Residual risk is visual/drag geometry in real browsers, but CI e2e is green and the PR description includes browser QA evidence.

sergioestebance (approved):

Issue #5255, opened 2026-07-02
ops(clickhouse): backfill _size_bytes (MATERIALIZE COLUMN) to stop storage-metering timeouts and cold scans


langwatch-agent:
## Summary

Storage metering (`StorageMeterService`, `sum(_size_bytes)`) is the single dominant source of failing and S3-cost-driving ClickHouse queries in prod right now. The app code is already fully hardened for it (capped threads, execution-time guardrail, stale-while-revalidate cache, single-flight, per-table fallback, degraded-0). The remaining fix is the one the code comments already point to and that has not been scheduled: **backfill `_size_bytes` on existing parts** so the lazy `byteSize(...)` recompute stops.

This is an ops runbook, not a code change (and deliberately not an auto-migration — `MATERIALIZE COLUMN` on these tables at prod scale must be run deliberately in a low-traffic window, one table at a time).

## Evidence (last 24h, redacted)

- **14 `Code: 159` TIMEOUT_EXCEEDED** exceptions in the ClickHouse server log, all storage-metering `sum(_size_bytes)` shapes: the combined `SELECT sum(t) FROM ( sum(_size_bytes) per table UNION ALL ... )` and the per-table fallback `SELECT sum(_size_bytes) FROM <table>` (observed on `simulation_runs`, `dspy_steps`, and others). These trip the 45s `max_execution_time` guardrail and degrade to 0.
- **~840 cold-scan warnings** for the same shape: ~470 for the combined query and ~93 each for the per-table fallback on `stored_metric_records`, `stored_spans`, `event_log`, `stored_log_records`. These fire because a total-storage sum has no time predicate (correct — you want all partitions), so it walks cold S3-tiered parts.

No tenant ids or byte values are included.

## Root cause (already documented in code)

`_size_bytes` is `UInt32 MATERIALIZED byteSize(...)` (migration `00032`). `ADD COLUMN ... MATERIALIZED` only writes the value on new parts; on parts written before the column existed, ClickHouse recomputes `byteSize(...)` over the heavy payload columns (`Messages.*`, `Inputs`, `Details`, `EventPayload`, ...) on every read. Summing across a large tenant therefore reads tens of GB and takes 24-60s. The migration expected merges to drift the value onto disk, but cold S3-tiered parts do not merge, so they never materialize on their own.

`StorageMeterService` already contains the pointer:

```
// The durable fix is to stop the recompute entirely by backfilling the
// column on existing parts (out of band, during a low-traffic window):
[...]

PR #5642, opened 2026-07-10
feat(billing): seeding + two-layer audits — trust the numbers (ADR-039 phase 3/4)

sergioestebance:
# Related Issue

Phase 3 of 4 implementing [ADR-039](https://github.com/langwatch/langwatch/pull/5597) (issue langwatch/langwatch-saas#879). Stacked on phase 2 (#5614).

## What

Nobody gets billed on a number that hasn't been proven. This phase adds the one-time backfill and the continuous verification the fold's fragile-by-construction nature demands.

- **Seeding** (`StorageSeedingService` + `pnpm task seedStorageBilling <org>|--all [lookbackDays]`): operator-run, never automatic. Replays the entry edge over history — one bounded per-partition query at a time; the full-backlog OOM shape is never run, not even once. SEED events are cause-keyed per run; idempotency is value-level (cumulative-minus-prior), so re-runs and mid-transit overlaps emit only what's missing. Doubles as the operator re-seed path for drift-alarmed orgs: corrective events with a full audit trail, never a gauge overwrite. **Precondition: the `MATERIALIZE COLUMN _size_bytes` backfill (#5255).**
- **Fold audit** (daily, Postgres-only, zero ClickHouse): fold(event log) must equal the gauge row, read in one RepeatableRead snapshot (torn reads can't phantom-alarm). Catches our own fold/apply bugs next-day.
- **Reference audit** (ClickHouse, capped): re-measures a stateless 7-day rotation of recorded partitions at **day-slice granularity with per-slice entitlement** — the exact model the exit edge bills on — against recorded live nets. Per-partition tolerance 100 MiB + org-level aggregate bound 1 GiB (diffuse drift can't hide). Seeded partitions' documented ≤6-day tail window is skipped, not flagged.
- **Alarms latch, never fix**: any discrepancy records `StorageAuditState.everAlarmedAt` (one-way), pages via PostHog capture, and the org is pinned to the daily audit tier permanently. In-flight/wedged `ALTER UPDATE _retention_days` mutations also pin daily until complete (wired to real mutation progress).
- **Sweep wiring**: audits run under the day claim through the tier-gated `runScheduledAudits`.

## Review hardening (adversarial pass, fixes in-branch)

M1: partition-granular entitlement would have false-alarmed (and permanently latched) every live-tracked partition mid-exit-week → slice-grained measurement. S2: tier service was dead code → wired, with real mutation-progress detection. S3: torn fold-audit reads → transaction snapshot. S4: aggregate drift bound.

## Spec

`specs/billing/storage/seeding-and-audits.feature` — 11/11 scenarios bound, tested against real Postgres + ClickHouse.

## Phases

1. Tables + pure math (#5603)
2. The dark engine (#5614)
3. **This PR — seeding + audits**
4. Stripe hourly reporting

---
## Review updates (2026-07-29)

- **Tracked (P1, fast-follow):** the reference audit only checks recorded partitions, so a partition with ClickHouse bytes but no recorded events is invisible → possible silent under-bill. Fix = bounded independent partition enumeration. Deferred because billing is gated behind a verified seed + clean audit before the flip. [issue langwatch/langwatch-saas#879 item 1](https://github.com/langwatch/langwatch-saas/issues/879#issuecomment-5116986012).
- Rebased onto latest main.


langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref langwatch/langwatch-saas#879 / ADR-039 phase 3

Scariest risk: audit + seed correctness ("trust the numbers"). An audit that passes silently on a mismatch, or a bad seed, means wrong billing data is trusted downstream.

Attention targets:
- **Must Check** the two-layer audit fails loudly on mismatch (error surfaced, not swallowed / logged-and-continued).
- **Must Check** seeding is idempotent and safe to re-run against prod data.
- **Ask Author** what triggers the audit, and who is alerted when it fails?

Shape note: phase 3 of the 4-PR stack. Attention map only, not an approval.

langwatch/src/server/app-layer/billing/storage/storageAudit.service.ts
```diff
@@ -0,0 +1,279 @@
...
+   * every partition is re-checked within 7 days with no cursor to corrupt.
+   */
+  async runReferenceAudit({
+    organizationId,
+    at,
+  }: {
+    organizationId: string;
+    at: Date;
+  }): Promise<{ clean: boolean; partitionsChecked: number }> {
+    const groups = await this.deps.events.sumLiveNetGroups({ organizationId });
```
langwatch-agent, line 154: **[P1] Reference audit never looks at partitions with no recorded boundary events.**\n\n`runReferenceAudit` builds its entire workload from `sumLiveNetGroups()`. If the seed/live measurement path misses a ClickHouse partition completely, there is no live net group for that `(projectId, partitionKey)`, so the audit never calls `measureReferenceBytes` for it, never records an alarm, and the org can stay permanently under-billed. This is exactly the drift class the phase-3 reference audit is meant to make loud: ClickHouse has billable bytes, but Postgres has no event/gauge representation.\n\nA concrete failure mode: the rollout seed skips one project/partition because the bounded query errors, times out, or a table/category bug returns no rows; later `runReferenceAudit` sees no `StorageBoundaryEvent` rows for that partition, so it omits it from `byPartition` and reports clean. The follow-up operator re-seed may fix it if someone already knows to run it, but the audit will not discover the miss.\n\nPlease drive the reference audit from an independent expected-partition source instead of only recorded groups, or persist seed coverage and alarm when an expected project/partition has no corresponding recorded bytes while ClickHouse remeasurement is nonzero. Add a regression where ClickHouse contains billable rows for a project/partition, `sumLiveNetGroups()` is empty for that partition, and `runReferenceAudit` records a reference alarm. Confidence: high.

langwatch-agent (changes requested):
1 P1 — reference audit can miss a fully unrecorded ClickHouse partition, so it does not yet prove seeding/measurement drift loudly enough. Targeted local vitest audit/measurement tests passed; ClickHouse seeding test stayed skipped without TEST_CLICKHOUSE_URL/CI_CLICKHOUSE_URL.

langwatch/src/server/app-layer/billing/storage/storageAudit.service.ts
sergioestebance: Valid — verified: `runReferenceAudit` builds its workload only from `sumLiveNetGroups()`, so a partition CH has bytes for but with zero recorded events is invisible to the audit. Tracking as a fast-follow rather than blocking the stack: billing is gated behind a *verified* seed + a clean audit cycle (`release_storage_boundary_billing`, default OFF), so a seed miss is meant to be caught at seed time before the per-org flip. The audit-hardening fix (bounded independent partition enumeration → measure + alarm on unrecorded-but-populated partitions) is documented in the issue: https://github.com/langwatch/langwatch/issues/5601#issuecomment-5116986012 (item 1).

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> This is the phase that decides whether anyone can trust the number. The fold is fragile by construction: a lost or duplicated boundary event corrupts an organization's gauge silently and forever, because nothing recomputes it. Seeding and the audits are the only things standing between that property and a wrong invoice. So the risk in merging is not that seeding breaks something, it is that the audits look thorough and do not actually catch the drift they exist to catch.

| | |
|---|---|
| **Issue** | Resolves `langwatch-saas#879`. Phase 3 of 4, stacked on `#5614` |
| **Size** | +1363/-37 across 19 files |
| **Mode** | Deep |
| **Blast radius** | The trustworthiness of every storage figure downstream |
<details>

The terrain, in three ideas:

</details>

<details>
<summary><b>What to press on</b></summary>

1. **What each audit layer would actually detect.** Two layers is the right instinct; the question is whether they are independent. If both derive from the same event stream, they agree with each other while both being wrong. An audit worth having recomputes from the source data at least sometimes, even expensively, even rarely.
2. **What happens when an audit fails.** Does it alarm, halt reporting for that org, or record and continue? For a billing number the conservative answer is to stop sending to Stripe rather than to send a figure known to be suspect, and that behaviour should be explicit rather than emergent.
3. **The seeding query shape.** The description is careful here and it matters: one bounded per-partition query at a time, and "the full-backlog OOM shape is never run, not even once". That is the exact query that caused the incidents ADR-039 exists to prevent, so verify the bound is real and cannot degenerate on an org with unusual partition spread.
4. **SEED events are cause-keyed per row.** That is what makes a re-run safe. Confirm re-running seeding on an org that was already seeded is genuinely a no-op and not a double count, because an operator will run it twice.
5. **Operator-run, never automatic** is the right call. It also means the safety of the whole stack depends on a human remembering to do it before the billing flag flips. Worth checking whether anything enforces that ordering or whether it is convention.
</details>

> [!NOTE]

PR #5664, opened 2026-07-10
fix(event-sourcing): register blob holds before write/stage so shared blobs can't be reclaimed mid-fan-out

0xdeafcafe:
## Problem

GQ2's content-addressed payload store shares one blob across every job carrying identical bytes (the fan-out win from ADR-029). But each occupancy's hold on that blob was registered **fire-and-forget, after staging**. That leaves a window where a fast sibling job completes, releases its own hold, sees an "empty" holder set, and reclaims a blob that a concurrent staging still references. The staged job then finds its payload gone and takes the missing-blob fail-safe instead of running its handler.

ADR-029's spec already required the hold to be atomic with staging (`payload-store-content-addressed.feature` AC3.9: *"SADD at stage … atomic with the job-entry write"*) — the implementation didn't match it.

A second, compounding gap: a genuinely deleted s3-tier object surfaces as the stored-objects drivers' `ObjectNotFoundError`, which `TieredBlobStore`'s missing-vs-transient classifier didn't recognize — so a gone-forever blob was treated as a transient outage and rode the full 25-attempt retry ladder (~2.5h+) before reaching the same fail-safe it should have hit on attempt 1. The unit-test double threw a `NoSuchKey`-named error rather than the real class, which is exactly why tests passed while the real read path misclassified.

## Fix

1. **Hold before write, awaited** (`jobEnvelope.ts`, `envelopeBlobLifecycle.ts`): `encode()` now registers the occupancy's hold in the holder set *before* the blob is written and before the returned value can be staged. A racing reclaim either sees the hold (and keeps the blob), or deletes bytes the staging's own idempotent content-addressed PUT restores immediately after. A crash between hold and write degrades to the holder-TTL backstop, never to a reclaimed live blob. The post-stage `acquire()` becomes an idempotent TTL refresh.
2. **S3 reclaim grace re-check** (`envelopeBlobLifecycle.ts`, `blobHolders.isHeld`): the redis tier's reclaim is atomic inside the release Lua, but the s3 `DeleteObject` is a separate network call whose decision can be stale by the time it executes. The reclaimer now waits `S3_RECLAIM_GRACE_MS` and re-checks the holder set, skipping the delete when the content was re-held. A skipped delete degrades to the TTL / bucket-lifecycle backstop.
3. **Not-found classification** (`tieredBlobStore.ts`): `isObjectMissingError` now recognizes `ObjectNotFoundError` (instance and shape) alongside the raw SDK/errno shapes, so a missing blob reaches the fail-safe immediately instead of burning the whole retry budget. The in-memory object-store double now throws the registry's real error class so this shape gap can't hide behind the double again.

## Tests

The old interleavings are reproduced deterministically; all of these **fail against the previous implementation**:

- `jobEnvelope.unit`: hold is registered before the blob write, for the same (hash, token) the envelope references.
- `envelopeBlobLifecycle.integration`: encode returns with the hold already in the holder set; a sibling's release never reclaims a blob a concurrent staging references (redis + s3 tiers); the s3 grace re-check skips the delete when the content is re-held, and still deletes when nothing re-holds it.
- `tieredBlobStore.unit`: every not-found shape (driver class, `NoSuchKey`, `NotFound`, `ENOENT`, HTTP 404) classifies as missing, never transient.

Spec: three new Track-3 scenarios in `specs/event-sourcing/payload-store-blob-hardening.feature` (hold-before-write, sibling-safe release, s3 reclaim grace re-check).

Verified: full groupQueue suite green — 208 integration + 94 unit — plus `pnpm typecheck`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented premature reclamation by ensuring shared blobs remain held while referenced by active staging operations.
  - Improved payload/envelope sequencing so holder registration occurs before blob writes.
  - Hardened “missing blob” detection across backends to avoid misclassification.
  - Added an S3 reclaim grace window that skips deletion when content is re-held before cleanup completes.

- **Tests**
  - Expanded integration and unit coverage for shared staging, fast sibling behavior, fan-out scenarios, retry alignment, and grace-window reclaim skip/delete behavior.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
@@ -319,6 +319,7 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
 
   describe("given stagings sharing one content-addressed blob", () => {
     describe("when a value is encoded", () => {
+      /** @scenario "A staging's hold on its shared blob exists before the blob is written" */
       it("registers the occupancy's hold before encode returns", async () => {
         const value = await lifecycle.encode({
           jobData: REDIS_TIER_PAYLOAD,
@@ -337,6 +338,7 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
     });
 
     describe("when one staging retires before a concurrent one dispatches", () => {
+      /** @scenario "A fan-out sibling completing early never reclaims a blob a concurrent staging references" */
       it("keeps the shared s3 blob alive for the remaining staging", async () => {
         const shared = { bulk: incompressible(768 * 1024) }; // > 256 KiB gzipped → s3
         const first = await lifecycle.encode({
@@ -348,16 +350,9 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
           groupId: TENANT_GROUP,
         });
         expect(readEnvelopeHold(second)!.ref.tier).toBe("s3");
-        // Ensure the FIRST staging's hold has landed (post-stage refresh), so
-        // its release below actually removes a member — the interleave where a
-        // hold-after-staging design empties the set and reclaims the blob out
-        // from under the second staging.
-        lifecycle.acquire(first);
-        await vi.waitFor(async () =>
-          expect(await redis.scard(holderKey(hashOf(first)))).toBeGreaterThan(
-            0,
-          ),
-        );
+        // Both stagings' holds land at encode-time (hold-before-write), so the
+        // set already has 2 members before either is released.
+        expect(await redis.scard(holderKey(hashOf(first)))).toBe(2);
 
         // The first staging completes and releases its hold while the second
         // is still awaiting dispatch.
@@ -380,12 +375,9 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
           jobData: REDIS_TIER_PAYLOAD,
           groupId: TENANT_GROUP,
         });
-        lifecycle.acquire(first);
-        await vi.waitFor(async () =>
-          expect(await redis.scard(holderKey(hashOf(first)))).toBeGreaterThan(
-            0,
-          ),
-        );
+        // Both stagings' holds land at encode-time (hold-before-write), so the
+        // set already has 2 members before either is released.
+        expect(await redis.scard(holderKey(hashOf(first)))).toBe(2);
 
         lifecycle.release({ values: [first], groupId: TENANT_GROUP });
         await new Promise((resolve) => setTimeout(resolve, 150));
@@ -412,6 +404,7 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
       });
 
     describe("when the same content is re-held before the delete executes", () => {
+      /** @scenario "An s3 reclaim is skipped when the content is re-held before the delete executes" */
       it("skips the delete so the new occupancy never references a deleted object", async () => {
         const graced = gracedLifecycle(200);
         const shared = { bulk: incompressible(768 * 1024) };
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
@@ -54,6 +54,9 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
         kind: "s3",
         bucket: "test-bucket",
       }),
+      // No reclaim grace by default so out-of-band delete assertions stay
+      // prompt; the grace behaviour has its own dedicated suite below.
+      s3ReclaimGraceMs: 0,
     });
   });
 
@@ -150,9 +153,12 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
         });
         const newHash = hashOf(newValue);
         const newHolderKey = `${queueName}:gq:blobholders:proj2/${newHash}`;
-        lifecycle.acquire(oldValue);
+        // Drop the hold proj2's own encode registered: the guard under test is
+        // about a foreign value arriving in proj1's group (mis-routed/forged),
+        // where no legitimate proj2 staging exists.
+        await redis.del(newHolderKey);
         const oldKey = holderKey(hashOf(oldValue));
-        await vi.waitFor(async () => expect(await redis.scard(oldKey)).toBe(1));
+        expect(await redis.scard(oldKey)).toBe(1); // encode-time hold
 
         lifecycle.transfer({ newValue, oldValue, groupId: TENANT_GROUP });
 
@@ -178,10 +184,9 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
         });
         const hash = hashOf(v1);
         expect(hashOf(v2)).toBe(hash); // same content → same blob
-        lifecycle.acquire(v1);
-        await vi.waitFor(async () =>
-          expect(await redis.scard(holderKey(hash))).toBe(1),
-        );
+        // Each encode registered its own hold up front, so both occupancies
+        // are already in the holder set before any transfer runs.
+        expect(await redis.scard(holderKey(hash))).toBe(2);
 
         lifecycle.transfer({
           newValue: v2,
@@ -213,9 +218,9 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
         });
 
         expect(hashOf(v2)).toBe(hashOf(v1));
-        lifecycle.acquire(v1);
         const key = holderKey(hashOf(v1));
-        await vi.waitFor(async () => expect(await redis.scard(key)).toBe(1));
+        // Both encode-time holds are registered before anything is staged.
+        expect(await redis.scard(key)).toBe(2);
 
         lifecycle.transfer({
           newValue: v2,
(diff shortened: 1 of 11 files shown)


specs/event-sourcing/payload-store-blob-hardening.feature
```diff
@@ -105,6 +105,36 @@ Feature: GroupQueue blob-handling hardening
     Then the blob survives until the last referencing job retires
     And no still-staged job is left pointing at a reclaimed blob
 
+  @integration @track3
+  # A hold that only lands after staging leaves a window where a fast sibling's
+  # completion sees an "empty" holder set and reclaims a blob a concurrent
+  # staging still needs. Hold-first closes it: a racing reclaim either sees the
+  # hold, or deletes bytes the staging's own idempotent write restores.
+  Scenario: A staging's hold on its shared blob exists before the blob is written
```
langwatch-agent, line 113: This new enforced Track-3 scenario is not bound to a test annotation, and `scripts/check-feature-parity.ts` fails locally/CI. The behavior appears covered by the new lifecycle/envelope tests, but the annotations do not use the exact scenario names. Please add the matching `@scenario` annotation above the relevant `it(...)` test or rename the scenario/test annotation so the feature-parity gate proves the intended regression coverage.

specs/event-sourcing/payload-store-blob-hardening.feature
```diff
@@ -105,6 +105,36 @@ Feature: GroupQueue blob-handling hardening
...
+    When a job whose payload offloads to a shared content-addressed blob is encoded
+    Then its hold is registered in the holder set before the blob write
+    And before the job becomes dispatchable
+
+  @integration @track3
+  Scenario: A fan-out sibling completing early never reclaims a blob a concurrent staging references
```
langwatch-agent, line 119: This new enforced Track-3 scenario is not bound to a test annotation, and `scripts/check-feature-parity.ts` fails locally/CI. The behavior appears covered by the new lifecycle/envelope tests, but the annotations do not use the exact scenario names. Please add the matching `@scenario` annotation above the relevant `it(...)` test or rename the scenario/test annotation so the feature-parity gate proves the intended regression coverage.

specs/event-sourcing/payload-store-blob-hardening.feature
```diff
@@ -105,6 +105,36 @@ Feature: GroupQueue blob-handling hardening
...
+    Then the shared blob survives the first staging's release
+    And the second staging decodes its payload intact
+
+  @integration @track3
+  # The redis-tier reclaim is atomic inside the release eval; the s3 delete is a
+  # separate network call, so the reclaim decision can be stale by the time the
+  # delete executes. The reclaimer re-checks after a grace window and skips the
+  # delete when the content is held again — a skipped orphan degrades to the
+  # TTL / bucket-lifecycle backstop, never to a deleted live blob.
+  Scenario: An s3 reclaim is skipped when the content is re-held before the delete executes
```
langwatch-agent, line 131: This new enforced Track-3 scenario is not bound to a test annotation, and `scripts/check-feature-parity.ts` fails locally/CI. The behavior appears covered by the new lifecycle/envelope tests, but the annotations do not use the exact scenario names. Please add the matching `@scenario` annotation above the relevant `it(...)` test or rename the scenario/test annotation so the feature-parity gate proves the intended regression coverage.

langwatch-agent (changes requested):
## Review: changes requested

The blob lifecycle implementation looks directionally correct in the areas I checked: the encode path now registers the holder before the blob write, same-content retries transfer within one holder set, S3 reclaim re-checks after the grace period, and the real `ObjectNotFoundError` shape is classified as missing rather than transient.

Validation I ran:

- `pnpm --dir langwatch exec tsx scripts/check-feature-parity.ts | rg "payload-store-blob-hardening|FAIL|unbound|stale|✗" -C 4`
- `LANGWATCH_ENDPOINT=http://localhost:5560 pnpm --dir langwatch exec vitest run -c vitest.integration.config.ts src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts src/server/event-sourcing/queues/groupQueue/__tests__/blobHolders.integration.test.ts`
- `pnpm --dir langwatch exec vitest run src/server/event-sourcing/queues/groupQueue/__tests__/jobEnvelope.unit.test.ts src/server/event-sourcing/queues/groupQueue/__tests__/tieredBlobStore.unit.test.ts`

The focused tests pass, but feature-parity is red because the three new Track-3 scenarios in `payload-store-blob-hardening.feature` are unbound. That is a merge blocker.

specs/event-sourcing/payload-store-blob-hardening.feature
0xdeafcafe: Addressed in fe1c95681: added /** @scenario "A staging's hold on its shared blob exists before the blob is written" */ above the matching it() block. Feature-parity check now passes (2406 scenarios bound, exit 0).

specs/event-sourcing/payload-store-blob-hardening.feature
0xdeafcafe: Addressed in fe1c95681: added /** @scenario "A fan-out sibling completing early never reclaims a blob a concurrent staging references" */ above the s3-variant test.

specs/event-sourcing/payload-store-blob-hardening.feature
0xdeafcafe: Addressed in fe1c95681: added /** @scenario "An s3 reclaim is skipped when the content is re-held before the delete executes" */ above the matching it() block.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
```diff
@@ -311,4 +316,141 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
...
+        // No acquire() call: by the time the value exists (and could be
+        // staged), the hold must already be in the holder set — a hold that
+        // only lands after staging leaves a window where a sibling's release
+        // reclaims the shared blob.
+        const hold = readEnvelopeHold(value)!;
+        expect(await redis.smembers(holderKey(hold.ref.hash))).toContain(
+          hold.token,
+        );
+      });
+    });
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in fe1c95681: all three scenarios now have their @scenario annotation bound to the corresponding it() blocks. Re-ran scripts/check-feature-parity.ts locally: exit 0, 2406 scenarios bound.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/envelopeBlobLifecycle.integration.test.ts
```diff
@@ -311,4 +316,141 @@ describe.skipIf(!hasTestcontainers)("EnvelopeBlobLifecycle", () => {
...
+
+        lifecycle.release({ values: [first], groupId: TENANT_GROUP });
+        await new Promise((resolve) => setTimeout(resolve, 150));
+
+        expect(await redis.exists(blobKey(hashOf(second)))).toBe(1);
+        expect(
+          await lifecycle.decode({ value: second, groupId: TENANT_GROUP }),
+        ).toEqual(REDIS_TIER_PAYLOAD);
+      });
+    });
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in fe1c95681: removed the acquire()+waitFor from both the s3 and redis fan-out-sibling tests (your prompt-for-AI-agents said 'both concurrency tests', so I applied it symmetrically even though the committable suggestion only showed the s3 one) — both now assert the holder cardinality is 2 directly after the two encode() calls.

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -303,6 +322,22 @@ export async function encodeJobEnvelope({
...
+      // the hold can be registered first.
+      const hash = contentHash(payloadJson);
+      // Hold BEFORE write (awaited): once the hold is in the holder set, no
+      // concurrent release can reclaim this content; and any reclaim that won
+      // the race before our hold landed deleted bytes the PUT below rewrites.
+      // A crash between these two steps leaves a hold guarding nothing — the
+      // holder-set TTL backstop clears it (ADR-030 §3).
+      if (acquireHold) {
+        await acquireHold({ projectId, hash, token });
+      }
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in fe1c95681: added the else-branch warn, matching the existing GQ2-downgrade warn style (same fields: projectId, queueName).

langwatch/src/server/event-sourcing/queues/groupQueue/jobEnvelope.ts
```diff
@@ -303,6 +322,27 @@ export async function encodeJobEnvelope({
     const payloadJson = JSON.stringify(payload);
     const payloadBytes = Buffer.byteLength(payloadJson);
     if (payloadBytes > INLINE_CEILING_BYTES) {
+      // Per-stage hold token: the holder-set member identifying this staged
+      // occupancy. Lives in the (inline) header, never in the content-addressed
+      // body, so it doesn't perturb the blob hash that collapses the fan-out.
+      const token = randomUUID();
+      // The hash the tiered store will key by (it hashes the same raw payload
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 88632aa49: switched to a KSUID (generate("holdtoken").toString()). Unlike the similar finding on a sibling PR (#5712), this line is genuinely new in this PR's own diff (the token extraction this PR introduces for the hold-before-write ordering), so it's in scope here rather than a driveby on untouched code.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -400,6 +396,55 @@ export class EnvelopeBlobLifecycle {
...
+      logger.warn(
+        {
+          projectId: ref.projectId,
+          blobHash: ref.hash,
+          edge,
+          err: redactStorageUrisInText(
+            err instanceof Error ? err.message : String(err),
+          ),
+        },
+        "S3 blob reclaim failed after holder drop — orphaned until bucket lifecycle sweeps",
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 88632aa49: the warn message now interpolates the edge parameter instead of hardcoding "after holder drop", so it reads correctly for both call sites (holder drop and transfer).

0xdeafcafe:
Closing as superseded by #5910.

That PR replaces holder ref-counting with renewable leases outright, and carries this PR's guarantee: leases are registered at encode time, before the blob is written or staged, so the mid-fan-out reclaim window this PR closed does not exist in the lease model either. It also touches all eleven files here plus `blobLeases.ts`, so rebasing this would only put it in conflict with its own successor.

The scenarios from `specs/event-sourcing/payload-store-blob-hardening.feature` are worth checking survive into #5910's spec coverage before that one merges — that is the one thing this PR had that is worth not losing.

PR #5665, opened 2026-07-10
feat(billing): stripe hourly storage reporter — the money path (ADR-039 phase 4/4)

sergioestebance:
# Related Issue

Final phase of [ADR-039](https://github.com/langwatch/langwatch/pull/5597) (issue langwatch/langwatch-saas#879). Stacked on phase 3 (#5642). **The stack is complete with this PR.**

## What

The kept reporting protocol (ADR-039 Decision 2), wired to the new engine — nothing reaches Stripe until an org's `release_storage_boundary_billing` flag flips, which happens only after seeding + a clean audit.

- **`StorageReportingService`**: drains unreported `StorageUsageHourly` rows oldest-first each sweep. Exactly-once via the `reportedAt` cursor + deterministic identifier `storage_mb:<org>:<hourISO>` (Stripe-side dedup on resend — a sent-but-unconfirmed event can never double-bill). Additive delivery into the `langwatch_storage_megabytes_hourly` sum meter.
- **Error discipline** (hardened by the adversarial pass): Stripe errors are *classified* via the shared `isStripeInvalidRequestError`/`isStripeAuthenticationError` helpers — `resource_already_exists` = success; permanent-class rejections settle the row with a distinct alert instead of wedging every newer hour behind a poison row forever; only genuinely transient errors retry and count toward the 5-failure circuit breaker.
- **Backdate ceiling**: hours older than 840h−24h margin are settled without a send and alerted (never a silent drop, never a wedged queue; `meterEventAdjustments` runbook owns recovery).
- **0-MiB hours** settle without a Stripe call. **Pricing pinned by test**: 8 GiB × 30 days = exactly €24 (€3/GiB-month ≈ €0.00000407/MiB-hour — the 720·1024 factors cancel, so it's exact).
- Covering index for the reporter queue; billing+metering flag gates share a 5-min per-org memo.
- **Known gap, recorded as an ADR open question**: a cancelling org's final unreported hours sit behind the billable gate unbilled — needs a subscription-end flush or an accepted write-off (business decision, deliberately not improvised here).

## Spec

`specs/billing/storage/stripe-hourly-reporting.feature` — 7/7 scenarios bound. **All 44 ADR-039 scenarios across the 4 spec files are now bound.** Reporter integration tests are PG+stub-based and run in CI (not run locally per no-docker constraint; pricing + unit suites green locally).

## The complete stack

1. langwatch/langwatch#5603 — tables + pure math
2. langwatch/langwatch#5614 — measurement engine, dark
3. langwatch/langwatch#5642 — seeding + two-layer audits
4. **This PR — Stripe reporting**

Rollout: merge stack → seed (`pnpm task seedStorageBilling`, after langwatch/langwatch#5255's MATERIALIZE backfill) → audits clean for a cycle → flip `release_storage_boundary_billing` per org.

## Deployment Impact

- **Prisma migration (1, additive):** covering index on `StorageUsageHourly` for the reporter's work queue. Applied automatically on deploy; index-only, no data change.
- **New PRODUCT flag `release_storage_boundary_billing` (default OFF):** with it off, the reporter returns before touching Stripe — hourly rows accumulate unreported, no meter events, no invoice impact. Flipped per organization only after seeding + a clean audit cycle.
- **Stripe surface:** uses the existing `langwatch_storage_megabytes_hourly` sum meter (already in the price catalog); no new webhooks, no subscription changes.
- **ADR edit:** one open-question row added (subscription-end flush gap) — documentation only.
- **No new env vars, no chart changes.** Rollback: revert deploy; unreported rows simply wait (exactly-once cursor makes later reporting safe up to the 840h ceiling).

---
## Review updates (2026-07-29)

- **Fixed (P1):** auth / config-class Stripe errors no longer settle billable hours (that silently dropped revenue). The sender returns only `sent`/`duplicate` and throws everything else → the hour stays unreported for retry and the circuit breaker alarms. Regression tests added.
- **Tracked (P1, fast-follow):** Stripe's ~24h identifier-dedup window vs the 35d retry horizon can double-bill after a long outage — needs an `attemptedAt` marker + manual-reconciliation stop. [issue langwatch/langwatch-saas#879 item 2](https://github.com/langwatch/langwatch-saas/issues/879#issuecomment-5116986012).
- **⚠️ GA blocker (P2):** the storage meter/price is not in the checked-in Stripe catalog. Before the billing flip: create the meter + metered price + subscription-item, add them to the catalog/types, and add a catalog test. [issue langwatch/langwatch-saas#879 item 3](https://github.com/langwatch/langwatch-saas/issues/879#issuecomment-5116986012).
- Rebased onto latest main (storage base migration re-ordered before its ALTERs).


langwatch-agent:
**PR Hound review brief** (`review: deep`) cross-ref langwatch/langwatch-saas#879 / ADR-039

Scariest risk: this is the Stripe money path (final phase 4 of the ADR-039 stack: reports storage usage that customers are billed on). A wrong or double-reported quantity mis-bills real money.

Attention targets:
- **Must Check** idempotency of the hourly reporter: no double-report on retry, overlap, or process restart (the reporting cursor / dedupe key).
- **Must Check** quantity math: unit conversion, hourly aggregation, and rounding match the ADR-039 boundary tables (no off-by-one at period edges).
- **Ask Author** partial-failure behavior mid-run (Stripe 5xx): retried, skipped, or at risk of double-count?

Shape note: 67 files, and the final of a 4-PR stack (#5603 -> langwatch/langwatch#5614 -> langwatch/langwatch#5642 -> langwatch/langwatch#5665, all langwatch/langwatch-saas#879). Review bottom-up; the same reviewer holds langwatch/langwatch#5603/#5614/#5665. Not an approval; attention map only.

langwatch/src/server/app-layer/presets.ts
```diff
@@ -666,29 +697,76 @@ export function initializeDefaultApp(options?: {
...
+              isStripeInvalidRequestError(error) &&
+              error.code === "resource_already_exists"
+            ) {
+              return { outcome: "duplicate" as const };
+            }
+            // Any other invalid-request / auth error is PERMANENT: Stripe
+            // will never accept this event, so retrying would wedge the
+            // org's oldest-first queue behind a poison row forever.
+            if (
+              isStripeInvalidRequestError(error) ||
```
langwatch-agent: [P1] This classifies Stripe auth errors, and every non-duplicate invalid-request error, as `permanent-reject`; `StorageReportingService` then stamps the hour `reportedAt` and continues. That is only safe for row-specific poison values. A bad API key, disabled account, missing/misnamed meter, or wrong Stripe catalog config is a system/config outage: with this code the sweep can walk real billable storage hours, mark them reported, and send nothing to Stripe. Treat auth/global config errors as retryable/circuit-breaker failures that leave rows unreported. Only settle errors that are provably row-specific and unrecoverable after config is fixed. Add a regression test for `StripeAuthenticationError`/`meter_not_found` proving `reportedAt` stays null.

langwatch/src/server/app-layer/billing/storage/storageReporting.service.ts
```diff
@@ -0,0 +1,242 @@
...
+          reportedAt: at,
+        });
+        await this.deps.checkpoints.resetFailures({
+          organizationId,
+          billingMonth,
+        });
+      } catch (error) {
+        // The hour stays unreported — the next sweep retries with the SAME
+        // deterministic identifier, so even a sent-but-unconfirmed event
+        // cannot double-bill.
```
langwatch-agent, line 179: [P1] The exactly-once guarantee relies on Stripe `identifier` dedupe, but Stripe documents meter-event identifier uniqueness as only a rolling period of at least 24 hours (https://docs.stripe.com/api/billing/meter-event/create). This reporter can retry unreported hours for much longer: rows remain open until the 35-day backdate ceiling, and a process/queue outage after Stripe accepted an event but before `markReported` can easily last beyond 24h. When it resumes, Stripe may accept the same `storage_mb:<org>:<hour>` again and double-bill that hour. Persist first-attempt/unknown-outcome state and stop for manual reconciliation once the 24h dedupe window can expire, or otherwise prove with Stripe API support that this identifier remains unique for the full retry horizon.

langwatch/src/server/app-layer/billing/storage/storageReporting.service.ts
```diff
@@ -0,0 +1,242 @@
+import { createLogger } from "~/utils/logger/server";
+import type { StorageBillingCheckpointRepository } from "./repositories/storage-billing-checkpoint.repository";
+import type { StorageUsageHourlyRepository } from "./repositories/storage-usage-hourly.repository";
+import { MS_PER_HOUR } from "./sealedHour";
+
+const logger = createLogger("langwatch:billing:storageReporting");
+
+/** Stripe `sum` meter this reporter feeds (kept contract, ADR-039 Decision 2). */
+export const STORAGE_METER_EVENT_NAME = "langwatch_storage_megabytes_hourly";
```
langwatch-agent, line 9: [P2] The reporter introduces the storage meter by raw event name, but the checked-in Stripe catalog/types still only map `BILLABLE_EVENTS`; I cannot find a storage meter id, storage metered price, or subscription-item path in this PR. That means CI cannot verify that `langwatch_storage_megabytes_hourly` exists in both test/live, is attached to the intended price, and prices `Σ MiB-hours` at the asserted €3/GiB-month. For a money path, put the storage meter/price in the local catalog/types and add a Stripe/catalog test that validates the meter event name, aggregation formula, customer/value payload keys, currency, and unit amount/decimal pricing, or document the external rollout step as an explicit merge blocker.

langwatch-agent (changes requested):
3 findings on the Stripe money path: 2 P1 idempotency/error-classification blockers and 1 P2 catalog/deployability gap. Focused validation: feature-parity passed; storage pricing unit test passed; after `prisma generate`, the reporter PG integration test could not complete locally because no PG test DB was listening on `localhost:55432`. Details inline.

langwatch/src/server/app-layer/presets.ts
sergioestebance: Fixed in 94d4d0c60. You're right that auth + config-class invalid-request are system outages, not row-specific — settling them dropped real revenue. The sender now returns only `sent` | `duplicate` (`resource_already_exists`) and throws everything else, so the reporter leaves the hour unreported for retry and the circuit breaker alarms; a genuinely poison row trips the breaker too (noisy but recoverable) rather than silently settling. Regression tests added: a system-wide auth failure leaves every hour unreported (never settled), and the breaker alarms after repeated failures. Verified end-to-end against Postgres.

langwatch/src/server/app-layer/billing/storage/storageReporting.service.ts
sergioestebance: Valid double-bill risk — Stripe only guarantees identifier dedup for a rolling ~24h while rows retry up to the 840h ceiling, so a >24h outage between a successful send and `markReported` can re-send and double-bill. The fix needs an `attemptedAt` marker (persist before send; on retry past a safe <24h window without confirmation, stop and alarm for manual reconciliation rather than blind re-send) — a schema column + reconciliation semantics, tracked as a fast-follow before the billing flip: https://github.com/langwatch/langwatch/issues/5601#issuecomment-5116986012 (item 2). Not reachable while the billing gate is OFF.

langwatch/src/server/app-layer/billing/storage/storageReporting.service.ts
sergioestebance: Valid — the storage meter isn't in the checked-in catalog, so CI can't verify it exists/prices correctly. Tracked as an explicit GA blocker (external Stripe rollout: create meter + metered price + subscription-item, add to catalog/types, add a catalog test for event name / aggregation / payload keys / currency / unit amount): https://github.com/langwatch/langwatch/issues/5601#issuecomment-5116986012 (item 3). The reporter references the name via the `STORAGE_METER_EVENT_NAME` constant and billing is dark by default, so nothing hits Stripe until that rollout + the per-org flip.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> This is the money path. After this merges, flipping one org's `release_storage_boundary_billing` flag starts sending usage records to Stripe, and a Stripe usage record is not something you quietly take back. Everything upstream in the stack has been rehearsal; this is the PR where a mistake reaches a customer's invoice.

| | |
|---|---|
| **Issue** | Resolves `langwatch-saas#879`. Phase 4 of 4, stacked on `#5642`. Completes the stack |
| **Size** | +3277/-2215 across 61 files, the largest in the series |
| **Mode** | Deep |
| **Blast radius** | Customer invoices |
<details>

The terrain, in three ideas:

</details>

<details>
<summary><b>The exactly-once claim, which is the whole review</b></summary>

Two mechanisms are stacked: a `reportedAt` cursor so rows drain oldest-first and only once, and a deterministic identifier `storage_mb:<org>:<hourISO>` so Stripe itself dedups a resend. That is the right belt-and-braces shape, and the interesting cases are the seams between them.

- **Sent but not recorded.** The classic: Stripe accepts the record, the process dies before `reportedAt` is written. On retry the deterministic identifier saves you. Confirm the identifier is genuinely derived from org and hour only, with no timestamp, attempt counter, or retry suffix leaking in, because any of those turn a dedup into a double charge.
- **Recorded but not sent.** The mirror case, and the one Stripe cannot save you from. This under-bills silently. Ask what detects it.
- **Clock and hour boundaries.** `<hourISO>` has to be the same string on both sides of a retry and across pods in different zones. UTC by construction, or by assumption?
- **Backfill after seeding.** When an org is seeded with history and then flagged, does the reporter drain months of hourly rows into Stripe at once? That may be correct and it may also be a very surprising invoice. Whatever the answer, it should be a decision rather than a consequence.

**The gating chain is the safety story and it is convention, not code**: nothing reaches Stripe until the flag flips, which is supposed to happen only after seeding plus a clean audit. Worth asking whether anything actually refuses to report for an org that has never been seeded, or whether that ordering lives only in a runbook.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Of the four phases this is the one to read last and hardest.

PR #5668, opened 2026-07-10
feat(traces): incremental Claude turn conversion (paused for fold-projection design discussion)

rogeriochaves:
# Incremental Claude Code turn conversion (split from langwatch/langwatch#5661, design discussion open)

Split out of langwatch/langwatch#5661 so the outage protections merge without waiting on this design. The conversation that paused it: the reactor-with-carried-state approach below works and is validated, but it hand-rolls state accumulation that the event-sourcing framework's FOLD PROJECTIONS already provide (managed state, ordering, partials, replay parity, ops surfaces), and a fold would also avoid the per-debounce ClickHouse re-reads entirely because events are pushed to it. The open alternative: a fold projection over claude-code log_record_received events whose state carries the turn accumulators (cursor, root aggregates, pending tool joins) and whose projection emits the span tree.

What this branch contains, regardless of which design wins:
- A compact per-trace conversion state (Redis, 48h TTL): cursor, root-span accumulators, tool records awaiting the next model call's request body (the cross-record join that forces whole-turn shape today).
- The reactor loops bounded batches strictly after the cursor, so each pass costs O(new records); deterministic span ids make re-emission an upsert; `truncated_logs` means "behind, continuing" and self-clears on catch-up; lost state falls back to a full idempotent redraw.
- The design-independent assets worth carrying into the fold version: the EQUIVALENCE PROPERTY TEST (whole-turn conversion equals incremental conversion over random batch partitions, including state serialized between every batch and a batch boundary splitting a tool call from its output-carrying model call), the pending-join semantics, and the `afterKey` repository predicate.

Validated: 56 targeted tests + full traces app-layer suite (1,158) green, typecheck clean, and dogfooded live (batch 20, max 2 batches per job: a real 25-span Claude turn converged across passes with no residual truncation marker).

## Deployment Impact

- No new env vars are required; `LANGWATCH_CLAUDE_TURN_MAX_BATCHES` (default 25) tunes batches per reactor job.
- No helm changes, no migrations; the conversion state lives in the existing Redis with a 48h TTL.
- Default behavior change: large turns converge fully instead of truncating at the conversion cap.
- BYOC dataplanes unaffected (no new storage surfaces).

## Summary by CodeRabbit

* **New Features**
  * Oversized evaluation inputs are stored safely with previews while remaining transparently readable.
  * Claude Code telemetry now processes large turns incrementally and more reliably.
  * Trace log processing can distribute work across multiple lanes.
  * Poisoned or oversized queue jobs are automatically parked for inspection and recovery.
  * Storage usage reporting now includes stored evaluation content.

* **Bug Fixes**
  * Added stronger payload-size protections to prevent oversized data from disrupting processing.
  * Improved credential error messaging for Google DLP redaction.

* **Documentation**
  * Added architecture and behavior specifications for these safeguards.


rogeriochaves:
Carrying over three CodeRabbit findings from langwatch/langwatch#5661 that target the code split out to this PR. Whichever design lands (this reactor-with-carried-state or the fold projection), these need addressing:

1. **`lastAssistantResponse` can regress on incremental replays** (`claude-code-log-to-span.ts`, the response scan near the root-accumulator update): the scan reads `combined` (carried orphans + new batch), so an older carried response can overwrite a newer persisted `prev.lastAssistantResponse` on a later pass. It should scan `batch` only, so the persisted root state moves strictly forward.

2. **A `behind` turn is never requeued** (`claudeCodeSpanSync.reactor.ts` batch loop): when the drain stops at `LANGWATCH_CLAUDE_TURN_MAX_BATCHES` while still behind, the reactor stamps `truncated_logs` but schedules no continuation pass. If no further logs arrive for that trace (a finished turn), the tail stays stranded until a manual re-fold. The reactor should enqueue the same aggregate again whenever it ends a pass behind.

3. **Duplicated Zod schema + TS interface** for the persisted conversion state (`claude-code-turn-conversion.state.ts`): repo standard is Zod only with `z.infer` when both validation and types are needed.

rogeriochaves:
superseeded by langwatch/langwatch#5708

Issue #917, opened 2025-12-04
Define the whole db structure for otel tables (also otel logs/metrics (?)) + current tables we have (scenario, batch evaluations)


rogeriochaves:

0xdeafcafe:
Superseded by #957 #959 #958

PR #5677, opened 2026-07-10
fix(security): move python-sdk examples to langchain-openai/text-splitters 1.x

langwatch-agent:
## Summary

Resolves the langchain-openai (#918, #919) and langchain-text-splitters (#916, #917) moderate advisories in `python-sdk` by moving the two lagging packages to their patched 1.x releases. This was previously treated as blocked on a langchain 0.x -> 1.x migration; the investigation below shows that is not the case for the SDK.

## Investigation: does this affect the customer-facing range?

**No.** The customer-facing langchain surface is unchanged and stays wide.

- The only langchain dependency a customer installs is the `langchain` extra: `langchain-core>=0.3.0,<2.0.0` (plus `openinference-instrumentation-langchain`). The SDK source (`src/langwatch/langchain.py`) imports **only** from `langchain_core` (messages, outputs, agents, callbacks, tools, documents, retrievers) and never from `langchain-openai` or `langchain-text-splitters`.
- `langchain-openai` and `langchain-text-splitters` live only in the `[dependency-groups] examples` block, which is dev/example tooling and is **not** installed by `pip install langwatch` / `langwatch[langchain]`. Bumping them has zero effect on what customers resolve.

**The `langchain-core>=0.3.0,<2.0.0` range genuinely works across 0.x and 1.x** (verified, not assumed):

- Full import surface of `langchain.py` imports cleanly on both `langchain-core==0.3.30` and `==1.4.9`.
- The `LangChainTracer` callback handler imports, instantiates, and runs a functional `on_chat_model_start` + `on_llm_end` cycle (constructing real `LLMResult` / `ChatGeneration` / `AIMessage` objects) identically on both versions. The overridden callbacks all take `**kwargs`, so added langchain kwargs do not break them.

So there is **no breaking change, no new minimum, and no 0.x.0 SDK release required** for customers. The wide range is already correct.

## What actually changes

Only the internal examples group:

```
- "langchain-openai>=0.3.0,<1.0.0"        + "langchain-openai>=1.1.14,<2.0.0"
- "langchain-text-splitters>=0.3.0,<1.0.0" + "langchain-text-splitters>=1.1.2,<2.0.0"
```

The group already pins `langchain>=1.0.0` and the lock already resolves `langchain-core 1.4.8`, so `uv lock` moves only these two: langchain-openai 0.3.34 -> 1.3.3, langchain-text-splitters 0.3.11 -> 1.1.2.

## Verification

- `uv lock`: only the two packages move.
- SDK unit suite: **406 passed, 1 skipped** (baseline unchanged).
- The example-code APIs on 1.x: `ChatOpenAI(streaming=/temperature=/model=/max_tokens=)`, `OpenAIEmbeddings()`, and `RecursiveCharacterTextSplitter` all import and construct correctly on langchain-openai 1.3.3 / text-splitters 1.1.2, so the example scripts remain compatible.

## Note on langevals

The langevals lock carries the same two advisories but is a separate, harder case (its legacy evaluator caps langchain-openai lower and the 1.x move there is a real evaluator migration). That is tracked separately and not included here.

## Summary by CodeRabbit

* **Chores**
  * Updated example environment requirements to support newer versions of LangChain OpenAI and text-splitting components.


langwatch-agent:
CI is fully green (0 failures, including sdk-python-complete and both e2e instances). Confirms the investigation: 406 unit tests pass and the customer-facing langchain-core range is untouched. Ready for human review and merge. Note: this supersedes Dependabot's own #3938 (langchain-text-splitters) and #5272 (langchain-openai) for /python-sdk, which were closed in the queue reset.

PR #5704, opened 2026-07-10
feat(langy): PR1/5 — Foundations: hexagonal Go service, caller-scoped session key, deploy hardening

0xdeafcafe:
## Langy Foundations — PR1 of 4

"Make the current thing right", **no event-sourcing yet**. Behaviour is preserved and the ADR-033 worker-isolation model survives verbatim. Later PRs in the stack depend on the seams this lays down.

Decisions + scenarios: **ADR-047** (`dev/docs/adr/047-langy-foundations.md`), specs under `specs/langy/`.

### Theme A — Rewrite the Go `services/langy-agent/` to the hexagonal house layout
Flat `langyagent` package → the same shape as `nlpgo`/`aigateway`, on the shared `pkg/` toolkit. Behaviour unchanged.
- `domain/` value objects + `errors.go` (`herr.Code` consts registered to HTTP statuses)
- `app/` `ports.go` (WorkerPool/Worker/ChatSink) + `app.go` orchestrator (functional options) mirroring the flat handler's chat-turn flow
- `adapters/httpapi/` stdlib `net/http` router + shared middleware + `pkg/health` probes (`/healthz`,`/readyz`,`/startupz`, legacy `/health` alias kept); every HTTP error via `herr.WriteHTTP`
- `adapters/workerpool/` the **ADR-033 isolation logic re-homed and wrapped** (not changed): per-worker UID + collision probe, 0700 chown-before-secrets home, per-worker `OPENCODE_SERVER_PASSWORD`, authProxy bearer→Basic swap, `filterSensitiveEnv` denylist, `Setpgid` process-group kill, orphan reaper, fail-closed `requireOpenCodeAuthEnforced`
- `adapters/egress/` pass-through **stub seam** for PR3
- `telemetry/` OTel spans + metrics (spawn/kill/at-capacity/readiness/per-turn latency) — the manager previously emitted **zero** OTel
- `config` via `pkg/config` (every existing env var name preserved), `deps` via `clog`+`otelsetup`+`health`, `serve` via `pkg/lifecycle` graceful drain
- `context` threaded throughout; worker subprocesses now bound to the pool-lifetime context (was `context.Background`)
- Loads of unit + integration Go tests across every package; the migrated isolation tests stay green

### Theme B — Per-session caller-scoped key, replacing the shared service key
Every chat used ONE admin-equivalent "Langy" service key. Now each chat mints a per-session `ApiKey` **owned by the requesting user**, restricted to the intersection of the Langy permission set and what that user actually holds — so `ApiKeyService`'s ceiling check clamps it and **a Langy tool call can never exceed what that user could do by hand**. The coarse route gate relaxes to a baseline project-read check (the per-session key is the real enforcement). Session keys are hidden from the API-keys UI. `mintLangySessionApiKey` unit + integration tests added.

### Theme C — Deploy hardening
- Chart render-time `fail` guard: `chartManaged` + empty `runtimeClassName` now refuses to render (mirrors the `replicaCount`/`service.type` guards) — making the values.yaml "refuses to deploy without gVisor" claim real
- `e2e/langy/k8s/pod.yaml` regenerated from the prod securityContext (root + the five per-UID caps) so per-worker isolation functions, behind a "LOCAL E2E ONLY — DO NOT USE IN PROD" banner

### Verification
- Go: `go build ./...` + `go test ./...` (incl. `-race`) green across all langy-agent packages; `go vet` + `gofmt` clean
- TS: `pnpm typecheck` exit 0
- Helm: `helm template` — default renders, empty `runtimeClassName` fails, `chartManaged=false` doesn't fire, existing guards intact

### Specs
`specs/langy/langy-agent-service-conventions.feature`, `langy-session-key.feature`, `langy-deploy-hardening.feature`; `langy-worker-isolation.feature` (ADR-033) unchanged, now satisfied by the `workerpool` tests.

### Known follow-ups (documented in ADR-047, out of scope for PR1)
- Session key is minted per chat POST; a reused worker keeps its first key so later mints are orphaned (hidden + short expiry) — mint-per-conversation / cleanup job is a clean follow-up
- Durability of in-flight turns, cooperative graceful shutdown, and horizontal scaling (Redis-backed shared session state, ≥2 replicas for zero-downtime rollouts) are deferred to the later event-sourcing stack PRs


0xdeafcafe:
Folding Foundations into the stack — its 4 commits (hexagonal Go re-home, caller-scoped session key, sandboxed-runtime deploy, ADR renumber) are already carried into the stack via the integration merge in the worker PR, so the code reaches main through the stack. Closing the standalone PR; the branch `feat/langy-foundations` is retained (backup tag `backup/langy-foundations-09a9cb81a`). The foundational Go work is now reviewed as part of the worker PR's scope.

PR #5706, opened 2026-07-10
feat(langy): PR4/4 — Frontend: event-sourced UI, real-time signals, domain-error rendering

0xdeafcafe:
Frontend + full-stack pass on Langy. The theme running through it: **every place the product relied on the model's prose, or on a fixture instead of reality, was wrong.**

## The tool stream was broken end to end

opencode announces a tool as `running` **before its arguments exist** (`input: {}`), then re-sends `running` with the command filled in. The Go emitter opened the card on the first, empty frame and discarded the re-send as a duplicate — so **the command reached the control plane on no frame at all**.

Everything downstream followed: the CLI envelope couldn't rewrite `bash` → `langwatch.trace.search`, so **no capability card had ever rendered**; every card fell back to a generic label ("Coding…"); and the durable event log recorded only *"the agent ran bash"*, which is why turns were so hard to debug after the fact. The Go tests all encoded the same wrong assumption, so they were green the whole time.

The card now opens on the frame that *knows the command*, the end frame carries the input, and `tool_call_*` events carry `command`, `input`, `durationMs` and `errorText`.

## Prose was a protocol. It isn't any more.

`[langy:connect-github]` and `[langy:progress:<stage>:<detail>]` asked the model to type magic strings into its answer, which were regexed back out to drive UI — an LLM as a reliable state machine. Both are gone:

- **Connect card** — a typed domain error raised from the tool stream when a command needs GitHub auth on a tokenless turn. The model can say nothing about GitHub and the card still appears.
- **Progress card** — derived from the real commands (`git clone`, `git push`, `gh pr create`). A failed command no longer completes its step.
- **PR card** — reads `gh pr create`'s own stdout, then enriches via the GitHub API (title, number, state, head→base, diff stat). Previously the PR-permit cap and the audit log were reconciled against a URL **scraped from the model's prose** — a mangled URL silently corrupted both.

Nothing in Langy's UI is steered by scraping model text any more.

## The context chips were decorative

`pageContext` was sent by the client and **silently stripped** by a non-strict Zod schema. No server code read it. Every chip a user saw — the trace they had open, the traces they'd selected, the filter they'd applied — was thrown away at the door.

Now declared, bounded, and sanitised (control chars and newlines stripped, so a chip label cannot forge a line of system prompt). Chip refs stay **inert**: the control plane never resolves one. The agent reaches a resource only through a tool call authenticated with a key scoped to that user, in that project.

`/` turns the composer into a command bar for **skill chips**, and a skill can be bound to a resource ("GitHub — on trace abc123").

## The thinking line was lying

It cycled a verb pool on a timer for as long as a turn was open, with **no input from reality**. On a turn whose worker never spawned, it announced *"Writing a TODO list…"*, *"Reading the whole file…"* for 97 seconds while nothing was running and not one token had arrived. This is why a dead turn read as a healthy one, and a large part of why "it's too slow" resisted diagnosis.

The line is now derived from the turn's real state, cycling **stops** when nothing is happening (cycling *is* a claim of progress), and it escalates to *"Langy still hasn't started — it may be stuck."* Verbs that named real acts are deleted; the jokes about Langy's character are kept, and a test pins both halves.

## Session keys were minted per turn

Every chat message minted a fresh user-scoped API key — 41 rows from a handful of dev turns, only 14 ever used, 6h TTL, **no sweeper**. Minting a new key doesn't revoke the old ones, so a user whose permissions were revoked kept live keys carrying the old scope for hours.

The key's natural lifetime is the **worker's**: it's env-injected at spawn and isn't part of the worker signature, so a reused worker keeps its original key regardless. The route now probes for a live worker and mints **only on a miss**. The manager is granted no minting power (a keyless spawn is refused before any capacity is reserved), revocation is a revoke-only callback that cannot touch a non-Langy key, and a reaper backstop remains because a SIGKILLed manager runs no cleanup.

The permission probe also issued **~81 queries per mint**. That is a correctness bug, not a slow one: those queries exhaust the Prisma pool and starve the interactive transaction inside `ApiKeyService.create` until it blows its 5s budget — **failing the turn with a 409**. Both axes now share one batched implementation, verified against the real database for identical results.

## Skill catalogue: 1 → 14

The `/` palette derived its skills from `services/langyagent/skills/`, which holds only `github`. The worker gets that **plus** `skills/_compiled/native/` — the Dockerfile COPYs both. The catalogue is now derived by reading the Dockerfile's own COPY set, with a test that fails if it drifts from what the image installs.

## Also

- **Recovery**: turn errors classified auto / awaiting-user / terminal. Retry goes through `regenerate()`, so it no longer posts a **second user message** and double-counts the fold. The worker-drain path now releases its PR permit (a deploy-interrupted turn was silently eating a daily PR slot).
- **The error id** shown to users as "share this with support" was the worker's long-lived root span — the same value on every turn, identifying nothing. Now the per-failure span id.
- **CLI compiled with Bun** and baked into the worker image (222ms → 87ms per invocation), with a `langwatch --version` gate in the build so a binary that compiles but cannot run fails at build time.
- **Tracing**: the Langy path now emits spans, and the Go manager extracts W3C trace context, so a turn is one waterfall rather than orphans.
- **UI**: palette and card kit taken from the marketing site's dark sections. Scroll fixed (the column's `height:100%` resolved against an `auto` parent, so `overflow-y:auto` never engaged and content ran off-screen). Dropdowns anchored correctly — they were collapsing to the viewport origin.
- **`LANGWATCH_API_URL` added to `.env.example`.** It must track `BASE_HOST`, or the worker's CLI authenticates a key minted here against a *different instance*. A stale dev server from another worktree on another port, backed by another database, is what made trace search fail with `Cannot read properties of undefined (reading 'traces')`.

## Known, not fixed here

- **Stream B has never worked.** `/fast` 404s on the first turn of every new conversation, because the handler resolves the conversation from the asynchronously-projected ClickHouse fold. In a product that is overwhelmingly first turns, the fast path built to make Langy feel fast has never once run.
- **The `failTurn` reload hole** — a refresh after a failed turn leaves the question with no answer and no explanation.


0xdeafcafe:
Superseded by #5741 — the Langy stack has been consolidated into one branch (full history retained) and rebased onto main for e2e testing. Will be re-split into smaller review PRs from there.

PR #5705, opened 2026-07-10
feat(langy): PR3/4 — Egress enforcement: monitor first, enforce last (ADR-043)

0xdeafcafe:
## PR4 of 4 — Langy egress enforcement (ADR-043)

Implements the egress **enforcement** ladder on top of PR3's egress
**monitoring** seam. Design: `dev/docs/adr/043-langy-egress-enforcement.md`,
`specs/langy/langy-egress-enforcement.feature`, `specs/langy/langy-pr4-plan.md`
(all included here). Default posture stays **monitor-only** — an install that
configures nothing upgrades into *watching*, not *blocking*.

### ⚠️ Base branch — I had to fall back off PR3
PR3's branch (`feat/langy-worker-streaming`) and the PR1+PR2 integration branch
(`feat/langy-integration`) **do not exist on the remote yet**, so per the plan's
fallback I branched off **`origin/feat/langy-event-sourcing`** and this PR's base
is `feat/langy-event-sourcing`. **This MUST be rebased onto PR3 before merge.**

Because PR3's egress-monitoring seam is not on this base, I implemented against
the interface the PR4 plan defines and **stubbed the PR3 coupling**: the rung-0
`egressMonitor` interface (`egressmonitor.go`) ships with a zap-logging
implementation (`logEgressMonitor`) that flags every decision. When PR3 lands,
its attributed-telemetry monitor is injected at `startEgressAdapter` /
`Manager.egressAdapterConfig` in place of the logging default — one seam, clearly
marked with `NOTE (PR4-on-fallback-base)`.

### What's implemented
1. **Require-TLS + per-destination throttle** (rung 1). Cleartext forwards and
   non-`:443` CONNECTs refused; soft per-host connection-burst tar-pit + byte-rate
   cap (slows + flags, no hard cliff; keyed per destination so one host's throttle
   never slows another).
2. **Customer allow-list** (rung 2, the crux). Per-project
   `Project.langyEgressAllowlist Json?` (nullable) → `getEgressAllowlist()` →
   `LangyCredentials.egressAllowlist` → Go `Credentials.EgressAllowlist` → per-worker
   adapter at spawn, bound via the `CredentialSignature` recycle. **Presence of the
   list is the mode:** unset ⇒ monitor-only; set ⇒ restrict to `floor ∪ list`.
3. **FQDN-bounded floor** (rung 3) in the Go L7 adapter (`egressadapter.go`), a
   per-worker forward proxy mirroring `authproxy.go`. FQDN read from the CONNECT
[...]

PR #5702, opened 2026-07-10
feat(langy): PR1/4 — Event sourcing: event-source conversations, delete the Postgres spine

0xdeafcafe:
## PR2 of 4 — Event-sourced Langy conversations

Moves Langy conversations onto the platform's event-sourcing framework. A conversation is now a **projection of an append-only event stream** instead of today's dual write across a Postgres `LangyConversation` spine and a ClickHouse `langy_messages` table joined by value (no FK, no idempotency, no replay).

Greenfield — Langy is staff-only and not rolled out, so this is a clean break: **no back-compat, no data migration, the Postgres spine is deleted outright.**

Design: **ADR-046** (`dev/docs/adr/046-event-sourced-langy-conversations.md`) · Spec: `specs/langy/langy-event-sourced-conversations.feature`

### The aggregate — `langy_conversation`
`aggregateId = conversationId`, `TenantId = projectId`. Events land in `event_log` ordered per-conversation by the in-house Redis GroupQueue.

**Commands (imperative)** → **events (past-tense)**
| Command | Event |
|---|---|
| `SendMessage` | `message_sent` |
| `StartAgentTurn` | `agent_turn_started` |
| `ReportStatus` | `status_reported` |
| `ReportProgress` | `progress_reported` |
| `ReconcileAgentTurn` | `turn_finalized` |
| `ArchiveConversation` | `conversation_archived` |
| `UpdateConversationMetadata` | `conversation_metadata_updated` |

Turn-lifecycle events `tool_call_started/completed`, `agent_responded`, `agent_turn_completed/failed` ship with fold handlers now (forward-compatible) — their emitting worker/reactor is **PR3**.

### Projections
- **Fold** `langyConversationState` → **new** `langy_conversations` CH table (`ReplacingMergeTree(UpdatedAt)`): owner, title, status, counts, timestamps, sharing. Replaces the PG spine. No message content.
- **Map** `langyMessageStorage` → **existing** `langy_messages` (reused): user message (`message_sent`) + assistant final answer (`turn_finalized`).

### Key decisions
- **`turn_finalized` carries the whole final answer** as the source of truth; individual streamed tokens are never events (they'd flood `event_log`).
- **Ephemeral events**: `status_reported`/`progress_reported` are classified ephemeral — same vocabulary/command surface, but PR3 routes them to a short-lived Redis buffer, not `event_log`. The fold only lets them touch a transient `LastHeartbeatAt`. Generalises to simulations' streaming-snapshot flood.
[...]

PR #5710, opened 2026-07-11
feat(langy): PR2/4 — Event-driven worker: streaming, self-observability, langyagent rename

0xdeafcafe:
## Langy PR3 (langy stack)

Event-driven OpenCode worker pool with per-turn streaming and self-observability (ADR-044), the local-dev enablement that came out of running it end-to-end, and the `langy-agent → langyagent` rename.

**Stacking:** bases on `feat/langy-event-sourcing` (#5702), which merges in `feat/langy-foundations` (#5704) — so the diff includes those. **PR4** (egress enforcement, #5705) now stacks on this branch; the **UI is PR5** (stacks on PR4).

### What's here
- **Event-driven worker + streaming + self-observability (ADR-044)** — the LangyWorkerPool, per-turn SSE streaming off OpenCode `/event`, ephemeral status/progress signals that leave the durable pipeline, and the content-stripped internal self-observability tee.
- **Config-gated unsafe-dev UID-isolation bypass** (`0362926c`) — `LANGY_UNSAFE_DEV_DISABLE_ISOLATION` lets the manager run OpenCode workers as the current non-root user on a dev box (skips the per-worker setuid + chown). Fail-closed: a real `Config` field (not a bare env read in the adapter), refused by `LoadConfig` unless `ENVIRONMENT` is local-like (**allowlist** local/dev/development/test — an unknown env like `prod-eu`/`staging` is refused). Prod behaviour unchanged (flag defaults off).
- **`LANGY_WORKSPACE_ROOT`** (same commit) — the shared `AGENTS.md` + skills dir is configurable (default `/workspace`) so the manager runs outside the container. Prod untouched (entrypoint.sh still seeds `/workspace`).
- **Helm `ENVIRONMENT`** — the chart now always emits `ENVIRONMENT` (inherits `global.env`, default `production`) so prod pods report a production env and refuse the unsafe flag even if it were set.
- **Rename `langy-agent → langyagent`** (`c5b573c1`) — aligns the service dir, Go import paths, `cmd` subcommand, `Dockerfile.langyagent`, and the Helm chart (`charts/langyagent` + umbrella dependency) with the existing Go package name `langyagent`. Deployment-contract effects (all rewired consistently in-chart): k8s Service name, app `OPENCODE_AGENT_URL` default, Docker image `langwatch/langyagent`, default Secret `langwatch-langyagent-auth`.

### Not here (moved up the stack)
- Worker-egress **security hardening** (CGNAT/`100.64` NetworkPolicy deny, `automountServiceAccountToken: false`, public-subnet nodeSelector) and the **per-worker-netns hard-block ADR** live on **PR4** (#5705), the egress-enforcement layer.

### Verified
Local Langy chat streams end-to-end (app → per-session key mint → agent `/chat` → OpenCode worker spawn → SSE) against a seeded `langwatch_langy_e2e` + ClickHouse. `go build`/`vet`/`test`, golangci-lint, `helm lint`/`template` (subchart + umbrella) all green.


0xdeafcafe:
Superseded by #5741 — the Langy stack has been consolidated into one branch (full history retained) and rebased onto main for e2e testing. Will be re-split into smaller review PRs from there.

Aryansharma28:
Noticed while reading through this stack: the manager pins `replicaCount: 1` and keeps worker state in-process memory, so `MAX_WORKERS` (default 20) is a hard ceiling on total concurrent Langy sessions today — no way to add more pods to handle more load, only scale vertically (bigger pod). The docs already flag this as a known limitation, with sticky routing / shared session state called out as the eventual fix.

Is there a rough plan for when horizontal scaling becomes necessary, or is 20 concurrent expected to hold for a while?

PR #5716, opened 2026-07-11
feat(traces): Claude Code telemetry — real wire keys, span-driven Terminal tab

0xdeafcafe:
## Why

Two things were wrong, and the first explains the second.

**1. We were reading Claude Code's log content from the wrong attribute keys.**
Every `claude_code` log event carries its content under its *own* key — there is
no shared `body` convention ([docs](https://code.claude.com/docs/en/monitoring-usage)):

| event | real key | we read |
|---|---|---|
| `user_prompt` | `prompt` | `body` ✗ |
| `assistant_response` | `response` | `body` ✗ |
| `api_request_body` / `api_response_body` | `body` | `body` ✓ |

So on the light path (no `OTEL_LOG_RAW_API_BODIES`, hence no `api_request_body`
to pair with) the `llm_request` spans came back with **no input and no output** —
silently. The existing unit test encoded the same wrong assumption, so it passed.
Fixtures are now pinned to the real wire format; I verified the tests fail against
the old code.

**2. The Terminal view was rendering the trace summary, not the session.**
A Claude Code turn is an agentic *loop* — model → tool → model → tool → answer —
but the trace summary only holds the opening prompt and the closing reply. Every
tool call was dropped before it reached the UI. The enrichment that *would* have
filled the gap was wired only into the legacy `TraceService`; the traces-v2 drawer
reads `app.traces.spans.*` and never touched it.

## What changed

- **Enrichment moved into the app layer** and is now called by *both* read paths
  (`tracesV2.spansFull` and the legacy `TraceService`). It runs **before**
  protections, so joined content is redacted like any other span content.
- **Terminal is its own drawer tab**, next to Conversation, gated on coding-agent
  origin — not a mode buried in the conversation view's segmented toggle.
- **The tab rebuilds the session from spans.** Each model call carries the rolling
  history, so the final `llm_request` span's input already contains the whole turn
  (prompt, every `tool_use`, every `tool_result`). Metrics sum across the loop.
- **The view drops the fake window chrome** — no frame, traffic lights or title bar.
  Claude Code doesn't draw those; it prints into the terminal you already have, and
  its hierarchy rides on four glyphs (`❯ ⏺ ⎿ ✻`) at one monospace size.
- Fixes a pre-existing `aria-label` type error on the scrubber.

## Verification

`pnpm typecheck` clean. 2456 tests / 192 files pass across `traces-v2`, `server/traces`,
`app-layer/traces`, `api/routers`. The wire-key fix is regression-pinned: reverting the
mapping makes 4 tests fail with `expected null to deeply equal { type: 'text' }`.

Not yet driven against a live session — dev ClickHouse was unreachable from this
environment, so the span-driven rebuild is covered by unit tests over the real
payload shapes rather than an end-to-end run.

## Follow-ups (tracked in the doc)

- Exact input join via `prompt.id` + `event.sequence` (removes the positional-pairing
  caveat for concurrent sub-agents).
- `claude_code.tool` spans + `tool.output` span events carry the tool's real I/O —
  we set `OTEL_LOG_TOOL_CONTENT=1` and read neither. Same for `tool_result` log events.
- Sub-agent tree: `agent_id` / `parent_agent_id` are on the spans, lifted into no DTO.
- Per-turn spans in the conversation view (same model as the Terminal tab).


0xdeafcafe:
Folded into #5708 — same lineage (this branch was cut from it), so #5708 now carries all 17 commits as the single PR.

PR #5717, opened 2026-07-11
fix(langy): PR4 review fixes — IDOR + demo gate, dead-code removal, token cleanups

0xdeafcafe:
Review-fix branch for **#5706** (PR4/4 Langy frontend). Stacks on `feat/langy-frontend`; squash into that PR or keep as a follow-up.

Only the **verifiable, low-risk** findings from the review are applied here. The larger structural refactors (LangyPanel split, shell-parser unification) and the architectural items are written up as executable specs instead of shipped, because this worktree has no build env to typecheck them.

## Security (fix before merge)
- **H1 (IDOR):** `langy.messages` tRPC read fetched any conversation's transcript with only a `TenantId+ConversationId` filter — a project co-member could read a private conversation they don't own. Now resolves `conversations.getById` (owner-or-shared) first and returns empty history when not visible, matching the Hono `GET /langy/conversations/:id` sibling.
- **M1 (demo bypass):** the whole `langy` tRPC router and `POST /langy/chat` lacked the demo-project rejection the Hono REST surface already applies (`evaluations:view` passes on the demo for every authenticated user). Added a shared `isDemoProjectId()` helper + a `langyReadGuard` middleware; DRY'd the existing Hono block onto the same helper.

## Dead code
- Deleted `services/langyagent/otlpreceiver/` (2,430 lines, 13 files) — imported by nothing, and would drop every span even if wired (correlates on `langy.conversation_id`; the worker injects `langwatch.thread.id`). Self-observability is unaffected (opencode already exports to `/api/otel`).
- Removed the unused `isAgentHealthy` preflight helper + `AGENT_HEALTH_CHECK_TIMEOUT_MS`; folded its still-relevant no-retry/idempotency rationale into the dispatch-site comment.

## House-rule cleanups
- Swapped 8 raw `fontSize="13px|12.5px|24px"` for `sm`/`2xl` tokens across 8 components — raw px opts a component out of langyTheme's deliberate token rescale.
- Moved the `newCount` "N new" count derivation out of the tRPC transport into `LangyConversationService.countSince` (same bounded-list strategy, no new ClickHouse read).

## Not shipped here (deferred, with reasons)
- `agent_turn_completed` / `agent_responded` dead events — documented PR3 seams; `agent_responded` is entangled with the reconcile timer + a service method + a unit test (reconciler is Tier 3). Remove them with the event-taxonomy/reconciler cleanup where the fold registration can be test-verified.
- LangyPanel.tsx split, shell-parser unification — need a build env to verify; specced in the review.

Note: **not typecheck-verified locally** (no `node_modules` in this worktree) — relying on CI.


0xdeafcafe:
Folded into #5706 via fast-forward — the review fixes (IDOR/demo-gate, dead-code removal, token cleanups) plus the follow-on simplifications (CLI card-registry consolidation, event-sourcing + turn/streaming ceremony cleanup, follow-up chips) now live directly on `feat/langy-frontend`. Closing this stacked PR as superseded; no unique commits remain.

PR #5738, opened 2026-07-12
fix(model-providers): prefer enabled row over narrower-scope disabled row in collapse (#5575)

### LINKED ISSUE #5575: Model-provider collapse ignores `enabled`: a disabled narrower-scope row masks an enabled wider-scope one (gates Ask AI)
## Bug

`getAllForProjectForFrontend` collapses multiple rows of the same `provider` string into one `Record` entry, choosing the winner by **scope narrowness only** — `enabled` is ignored, and same-scope ties keep the **first** row seen.

Consequence: an org-scoped `openai` row with `enabled: true` is shadowed by a project-scoped `openai` row with `enabled: false`. Every consumer of `hasEnabledProviders` then reads `false` even though an enabled provider exists and is usable.

The most visible symptom is the **Ask AI** gate in the new traces view (`SearchBar.tsx` → `useModelProvidersSettings` → `hasEnabledProviders`): it renders the "Connect a model provider" primer and refuses to open, despite a working provider being configured.

## Root cause

`langwatch/src/server/modelProviders/modelProvider.service.ts`

```ts
const existing = acc[mp.provider];
if (!existing || this.isNarrower(provider_, existing)) {
  return { ...acc, [mp.provider]: provider_ };
}
return acc;
```

```ts
private isNarrower(a, b) {
  return this.scopePriority(a.scopeType) > this.scopePriority(b.scopeType);
}
```

`isNarrower` compares scope priority and nothing else. For equal priority it returns `false`, so the first-inserted row wins regardless of whether it is enabled or has credentials.

## Why this matters now

#5380 produced duplicate rows for the same provider type. Users who already have those duplicates (the original reporter among them) can still see Ask AI gated after the #5383 fix stops *new* duplicates from being created — if the first of their duplicate rows is the disabled/keyless one, it wins the collapse and masks the working row.

This is also reachable without any duplicate-creation bug, via the supported multi-instance setup: an enabled org-wide provider plus a deliberately disabled project-scoped override.

## Steps to reproduce

1. Create an `openai` provider at ORGANIZATION scope with a valid key, `enabled: true`.
2. Create a second `openai` provider at PROJECT scope, `enabled: false`.
3. Open the new traces view and click **Ask AI**.
4. Observed: the "Connect a model provider" primer appears. Expected: Ask AI opens — an enabled provider exists.

## Suggested fix

Make the collapse prefer an enabled row before falling back to scope narrowness, e.g. tie-break `enabled` first, then `scopePriority`. Needs a deliberate decision, because the collapsed `Record` also drives credential resolution for evaluators and default models — changing which row wins changes which credentials are used. Worth an explicit test on `modelProvider.service.ts` pinning: enabled-org beats disabled-project; disabled-org loses to enabled-project; same-enabled state falls back to narrowest scope.

## Notes

Found while browser-QA'ing #5383 (PR for #5380). Deliberately **not** fixed there to keep that PR scoped to the CRUD defect; the collapse change has a wider blast radius (credential resolution) and deserves its own review.

No existing test covers the scope-priority-vs-`enabled` interaction — `useModelProvidersSettings.unit.test.ts` only exercises the boolean logic given an already-collapsed Record.

---
*Classification: Bug*

feiiiiii5:
## What

`ModelProviderService.isNarrower` now tie-breaks on `enabled` before scope narrowness when collapsing multiple stored rows for the same provider string. A disabled narrower-scope row no longer masks an enabled wider-scope row.

## Why

Fixes #5575. When a provider (e.g. `openai`) had both an enabled `ORGANIZATION`-scoped row and a disabled `PROJECT`-scoped row, the collapse picked the PROJECT row (narrower scope wins) and returned `enabled: false` to the frontend. `useModelProvidersSettings.hasEnabledProviders` then gated off Ask AI, even though an enabled org-wide credential existed.

The scope-only collapse was introduced in #4073 (iter 107/108) to preserve the legacy `Record<provider, …>` shape, but `enabled` wasn't considered as a tie-break key.

## Approach

Modify `isNarrower` so that:
1. If the two rows differ in `enabled`, the enabled one wins — regardless of scope.
2. If they share the same `enabled` state, fall through to the existing `scopePriority` comparison (preserves iter 107/108 semantics for the common case).

Also updated the comment at the collapse site to document the new tie-break order.

**Alternatives considered** (full notes in fix-notes):
- Filter disabled rows before collapsing → rejected (changes the `Record<provider, …>` contract; disabled rows are legitimate stored state).
- Return a list instead of a collapsed record → rejected (`listProjectModelProvidersForFrontend` already exists for that; the collapsed shape is the legacy contract).
- Mix `enabled` into `pickNarrowestScope` → rejected (category error: `enabled` is row-level, `pickNarrowestScope` is intra-row scope-level).

## Testing

Added `langwatch/src/server/modelProviders/__tests__/modelProvider.enabledCollapse.integration.test.ts` (matches the existing integration-test pattern, gated on `CREDENTIALS_SECRET` like the other 3 service tests):

1. **`prefers an enabled wider-scope row over a disabled narrower-scope row`** — creates an enabled ORG `openai` row + a disabled PROJECT `openai` row, asserts the collapsed result has `enabled === true` and `scopeType === "ORGANIZATION"`. This is the #5575 reproduction — fails without the fix.
2. **`still picks the narrower scope when both rows are enabled`** — creates enabled ORG + enabled PROJECT `anthropic` rows, asserts the collapsed result has `scopeType === "PROJECT"`. Regression guard for the iter 107/108 scope-narrowness semantics.

Could not run the suite locally (no Postgres + `CREDENTIALS_SECRET` in this environment), but the test mirrors the exact structure of `modelProvider.multiInstance.integration.test.ts` and the service diff was verified via the GitHub compare API.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model provider selection across organization and project scopes with explicit enabled/disabled precedence.
  - Enabled providers now override disabled providers even when the disabled entry has a narrower scope.
  - When multiple enabled entries exist, the most specific scope is selected (project overrides broader organization).

- **Tests**
  - Added Postgres integration coverage for model provider selection, including enabled-vs-scope precedence scenarios.


feiiiiii5:
Thanks @coderabbitai — good catch. Pushed follow-up commit `a5fbdae` to sync the `buildSavedProviders` docblock (lines 809-816) with the enabled-first collapse rule; the function-level summary now matches the inline comment at the collapse site (862-866). Comment-only, no behavioral change.

feiiiiii5:
Good catch again @coderabbitai — verified `listAccessibleForProject` does not exist on `ModelProviderService` (only referenced in this stale comment). The real uncollapsed-list method is `listProjectModelProvidersForFrontend` (defined at line 188). Pushed follow-up commit `1951a71` to fix the reference. Comment-only, no behavioral change.

0xdeafcafe (dismissed):
this certainly feels like matter of preference, but however i think this is the solution i prefer. thanks for the contribution!

feiiiiii5:
Thanks @0xdeafcafe for the review and for surfacing the preference call-out — appreciate the merge signal.

Just pushed one more commit (`7e98cdb`) to fix the `typecheck` CI failure: `noUncheckedIndexedAccess` is on in `tsconfig.tsgo.json`, so indexing `Record<string, MaybeStoredModelProvider>` returns `T | undefined`. Vitest's `toBeDefined()` doesn't narrow the type, so the four property accesses on `result.openai` / `result.anthropic` tripped TS18048. Added non-null assertions on the accesses — the `expect(...).toBeDefined()` still runs first at runtime, so a missing key fails with a clear message before the asserted property is ever touched. Test-only change, no behavioral diff.

The Codecov patch-coverage warning is on the two new comment lines in `modelProvider.service.ts` (the `buildSavedProviders` docblock sync). The actual logic lines are covered by the new integration test, so no extra coverage gap was introduced by the fix itself.

0xdeafcafe (approved):

0xdeafcafe (approved):

PR #5741, opened 2026-07-12
feat(langy): langy goes large - event driven orchestration, ideated on a design concept, made the go agent well tuff

0xdeafcafe:
## What this is

The **consolidated Langy rework**, collapsed from the 4-way stacked PRs into **one branch** with **full history retained (92 commits)**, rebased onto `main`. This is a **testing/integration PR** — once it's verified end-to-end, it will be **split back into smaller, reviewable PRs**.

Supersedes the stacked set (their commits are all included here, rebased):
- #5702 — PR1/4 event sourcing (event-source conversations, delete the Postgres spine)
- #5710 — PR2/4 event-driven worker: streaming, self-observability, `langyagent` rename
- #5705 — PR3/4 egress enforcement (monitor first, enforce last — ADR-043)
- #5706 — PR4/4 frontend: event-sourced UI, real-time signals, domain-error rendering

## Why one PR now

So many recent changes are cross-stack (worker restructure into hexagonal boundaries, the authenticated frame contract, the Hono relay, request-scoped logging, the `pkg/rpc` extraction, the capability→signature unification, the `/worker/{create,revive,continue}` rename). Keeping them cleanly separated across the stack stopped being worth the churn. Consolidate → test → then re-split for review.

## Rebase notes (main advanced 2 commits: #5700 single-process mode, #5540 simulations)
- **`workers.ts`**: `#5700` extracted all worker registration into `startWorkers.ts`. Resolved by keeping main's thin `startWorkers()` shell and moving the Langy worker registration into a new `bootLangyTurnProcessor()` helper (mirrors `bootScenarioProcessor`).
- **`tools/thuishaven/*`**: a `EnsureReady()` proxy self-bootstrap change had been accidentally bundled into a langy commit; `#5700` already landed the same work (a superset). Dropped our redundant copy, kept main's canonical version.

## Deployment Impact

This changes the Langy agent image and its runtime inputs. The app image build now includes the shared `packages/cli-cards` workspace and the root `skills/` sources needed to generate the in-product skill catalogue. The Langy agent image continues to use the Bun-compiled `langwatch` CLI and compiled native skills; no new required environment variables or Helm values are introduced. A vanilla Helm install keeps the existing defaults, and BYOC deployments receive the same application and agent image behavior with the new event-sourced Langy paths enabled only where configured.

## State
- Go: `services/langyagent/...` + `pkg/rpc/...` build/vet/test clean.
- TS: typechecks with only the 2 known pre-existing errors (`langy.ts:546` `signal`, `langy-turn.processor.ts:818` `Logger`).
- Web app verified e2e-ready (integration is via event-sourcing, not the interim executor classes); remaining blockers are runtime setup (env, the Go manager, a project with an LLM model).


langwatch/src/server/event-sourcing/pipelines/langy-conversation-processing/reactors/agentTurnLiveness.reactor.ts
```diff
@@ -0,0 +1,188 @@
...
+      // a fresh worker, then throw so the queue re-checks after a backoff.
+      await deps.buffer
+        .appendStatus({
+          conversationId,
+          turnId: currentTurn,
+          status: "Reconnecting to the agent…",
+        })
+        .catch(() => undefined);
+
+      const intent = handoff.resumeToken
```
Aryansharma28: For the true hard-kill case (no `resumeToken`, so `intent === "continue"`) — the fresh worker only gets the original `prompt`/`system`/`credentials` from `LangyTurnHandoff`, captured once at turn start (`langyTurnHandoff.ts`). It has no visibility into which `tool_call_succeeded`/`tool_call_failed` events already landed for this turn before the kill.

If one of those already-completed tool calls was non-idempotent (created a PR, posted a comment, etc.), the re-driven turn risks redoing it, since the new worker starts from the original ask with no memory of what already happened. Those tool-call events are already durable in `event_log` by this point — could we fetch the current turn's completed tool calls and fold a summary into the fresh worker's context before dispatching `continue`, so it knows what's already done?

Aryansharma28 (commented):
A question from reading through the shutdown-handoff / liveness re-drive path in this rework — see inline comment.

langwatch/src/server/clickhouse/migrations/00048_create_langy_conversations.sql
```diff
@@ -0,0 +1,78 @@
...
+-- conversation-level spine metadata — NO message content. Per-message content
+-- stays customer-side in `langy_messages` (00036), the same hybrid-deployment
+-- guarantee the original design made (PR #4913): message text never leaves the
+-- customer's ClickHouse.
+--
+-- Shape mirrors the other fold-projection tables (simulation_runs,
+-- experiment_runs): ProjectionId + TenantId + the aggregate id (ConversationId)
+-- + Version, ReplacingMergeTree(UpdatedAt) for insert-retry / re-fold
+-- idempotency. TenantId = projectId (platform-wide convention). Retention TTL
+-- is intentionally omitted for parity with langy_messages; a Langy retention /
```
Aryansharma28: Retention TTL is intentionally skipped here (and on `langy_conversation_turns`) for parity with `langy_messages`, with the actual erase sweep punted to a follow-up (ADR-046 open question 3). Given this table is now the sole home for Langy conversation history (no Postgres fallback anymore), does that mean conversations are kept indefinitely until that follow-up lands? Worth calling out explicitly if so — e.g. in the PR description or a tracking issue — since it's a real behavior change from whatever retention story existed before, not just an implementation detail.

Aryansharma28 (commented):
One question about a design choice in the new Langy projection tables — see inline comment.

Aryansharma28:
Noticed while reading through this stack: the manager pins `replicaCount: 1` and keeps worker state in-process memory, so `MAX_WORKERS` (default 20) is a hard ceiling on total concurrent Langy sessions today — no way to add more pods to handle more load, only scale vertically (bigger pod). The docs already flag this as a known limitation, with sticky routing / shared session state called out as the eventual fix.

Is there a rough plan for when horizontal scaling becomes necessary, or is 20 concurrent expected to hold for a while?

typescript-sdk/src/cli/daemon/identity.ts
```diff
@@ -0,0 +1,199 @@
...
+  const uid = typeof process.getuid === "function" ? process.getuid() : 0;
+  // The config path participates too: `LANGWATCH_CLI_CONFIG` swaps the whole
+  // persisted-credential file, which is a different identity even when the
+  // endpoint and API key happen to match (this is also what keeps tests from
+  // colliding with a developer's real daemon).
+  const configPath = env.LANGWATCH_CLI_CONFIG ?? "";
+
+  const fingerprint = crypto
+    .createHash("sha256")
+    .update(`${endpoint}\0${apiKey}\0${uid}\0${configPath}`)
```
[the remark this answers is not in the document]
sergioestebance: this needs to be addressed, otherwise it will be highlight in the in the repo
probably using salt+pepper

rogeriochaves (dismissed):

PR #5754, opened 2026-07-13
fix(traces): resolve full blob IO on evaluation reads + fix stale resolveBlobs doc (#5752, #5751)

### LINKED ISSUE #5751: Fix stale resolveBlobs doc comment in traces/types.ts
## Context

Flagged in review of #5082 ([review link](https://github.com/langwatch/langwatch/pull/5082#pullrequestreview-4683002673)):

> `traces/types.ts:27-28` has a stale doc comment on `resolveBlobs` ("requires includeSpans:true") that now contradicts the fix, cosmetic, but will mislead the next reader.

## Problem

The `resolveBlobs` field doc on `GetAllTracesForProjectOptions` (`langwatch/src/server/traces/types.ts`) says:

> Requires `includeSpans: true` to have any effect (resolution runs during span enrichment).

#5082 decoupled blob resolution from `includeSpans` (the summary-export gate now triggers on `wantsSpans || wantsFullIo`), so the comment no longer describes actual behavior.

## Acceptance criteria

- The doc comment accurately reflects that `resolveBlobs` takes effect independent of `includeSpans`.
- No behavior change (comment-only).

### LINKED ISSUE #5752: runEvaluation content-consuming reads return truncated (>64KB) blob previews to evaluators
## Context

Flagged in review of #5082 ([review link](https://github.com/langwatch/langwatch/pull/5082#pullrequestreview-4683002673)) as part of the "all read paths" backlog for the >64 KB span-IO offload (#4991 / ADR-022):

> `runEvaluation.ts`'s `getTracesByThreadId` (legacy `/api/evaluations` route + `evaluations.ts` tRPC) is a live content-consuming read path not covered by this PR's 4 scoped call sites.

## Problem

`langwatch/src/server/evaluations/runEvaluation.ts` builds `TraceService.create()` **without** blob-resolution deps and reads trace content **without** `{ full: true }` at every content-consuming read that feeds an evaluator:

- the trace-level read `getById(projectId, traceId, protections)` (evaluator input: `trace.input` / `trace.output`);
- the two thread-level reads `getTracesByThreadId(projectId, threadId, protections)` (thread-mapping resolver + `resolveThreadMappingsIntoData`).

So any offloaded (>64 KB) message content is handed to the evaluator as the truncated preview, and evaluators score against truncated input. The content-consuming thread tRPC route and the legacy `/thread/:id` route already resolve full IO (`buildTraceBlobResolutionDeps()` + `{ full: true }`); the evaluation path was missed.

The review named only the thread reads, but the trace-level `getById` is the same bug (the evaluator's primary input) and is fixed together to avoid a half-fix.

## Acceptance criteria

- All content-consuming evaluator reads in `runEvaluation.ts` resolve full offloaded IO (construct `TraceService` with `buildTraceBlobResolutionDeps()` and pass `{ full: true }`).
- Covered by a test proving over-threshold trace/thread content reaches the evaluator un-truncated.

langwatch-agent:
## Ticket

Closes #5752, Closes #5751. Two non-blocking follow-ups flagged in the [review of #5082](https://github.com/langwatch/langwatch/pull/5082#pullrequestreview-4683002673), part of the ">64 KB span-IO offload / all read paths" backlog (#4991 / ADR-022).

## Change

**1. Evaluation reads resolved truncated content (#5752).** `runEvaluation.ts` built `TraceService.create()` with no blob-resolution deps and read trace content without `{ full: true }` at every read that feeds an evaluator, so offloaded (>64 KB) message content was handed to the evaluator as the 64 KB preview. Fixed all three content-consuming reads to construct with `buildTraceBlobResolutionDeps()` and pass `{ full: true }`:
- the trace-level `getById` (the evaluator's primary input, in `runEvaluationForTrace`);
- the thread-level `getTracesByThreadId` in `buildThreadData` and in `resolveThreadMappingsIntoData`.

The review named only the thread reads; the trace-level `getById` is the same bug (and the evaluator's primary input), so it is fixed together to avoid a half-fix. The content-consuming thread tRPC route and legacy `/thread/:id` route already resolved full IO.

**2. Stale doc comment (#5751).** `traces/types.ts` said `resolveBlobs` "Requires `includeSpans: true` to have any effect". Since #5082 the flag triggers the internal span read on its own (`wantsSpans || wantsFullIo`), so the comment is corrected to match.

No production behavior change beyond the reads now resolving full IO; the extra `event_log` read only happens when IO was actually offloaded.

**3. Oversized text attributes now keep a readable preview instead of being discarded entirely (customer report).** A customer reported that the Trace Explorer sometimes shows only `[truncated: N bytes]` for a large user input, with no content at all. Root cause: `capOversizedAttributes.ts` — the pre-persistence 256KB safety cap for span attribute values — replaced *any* oversized value with a bare byte-count placeholder, including plain readable text. This fires whenever `release_trace_blob_offload` is off, which is every project by default, so the customer's full input was destroyed before it ever reached storage.

Human-readable text now keeps a UTF-8-safe prefix instead of being wiped: 64KB for `langwatch.input`/`langwatch.output`/the `gen_ai.*.messages` equivalents, 2KB for any other attribute (a span can carry many arbitrary custom attributes at once, so their per-attribute budget stays small to keep aggregate fold-state bounded). Base64 image/audio data URLs and raw `bytesValue` content are unchanged — a partial image has no readable value. `IO_ATTRIBUTE_KEYS` and `utf8Preview` moved down into `capOversizedAttributes.ts` (the lower-level module) so `lean-for-projection.ts`'s existing ADR-022 lean pass shares the same definitions instead of duplicating them.

Verified against the customer's exact scenario locally — a 344,000-byte `langwatch.input`, `release_trace_blob_offload` off, posted through the real OTLP ingestion path against local Postgres/ClickHouse/Redis:

Traces list showing the real input text instead of a truncated-bytes marker

Trace detail drawer showing ~64KB of the real input text

New spec: `specs/event-sourcing/oversized-attribute-value-preview.feature`.

## Evidence

- New `runEvaluation.4991-full-resolution.unit.test.ts` proves the trace-level and thread-level evaluator reads construct `TraceService` with blob deps and pass `{ full: true }` (mirrors the existing `annotation.4991-full-resolution.unit.test.ts`). **2 passed.**
- Neighboring suites green: `runEvaluation.evaluations-enrichment`, `runEvaluation.langevals-request`, `threadMappingResolver` (10 passed).
- `capOversizedAttributes.unit.test.ts` (29 passed, including new scenarios for the IO/non-IO preview budgets and the unchanged binary-blob behavior) and `lean-for-projection.unit.test.ts` (24 passed).
- Broader regression sweep — `src/server/app-layer/traces` + `src/server/traces`: 1493 tests, all green.
- `pnpm typecheck` clean (pre-existing unrelated dependency-resolution errors in this worktree — `bullmq`, `lodash.*`, `@react-email/*` — untouched by this change).

## Advisory note

## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluations now use complete trace and thread content instead of truncated previews, ensuring evaluators receive full offloaded input.
  * Trace and thread evaluation paths consistently resolve full input data.
  * Oversized (non-binary) span attribute values now keep a readable partial preview instead of being replaced entirely by a byte-count placeholder.

* **Documentation**
  * Clarified when full trace content and span data are loaded during trace retrieval.


langwatch-agent (commented):
No findings. I checked the linked #5751/#5752 acceptance criteria and traced the evaluator read paths through TraceService/ClickHouseTraceService: the primary `getById`, thread-level `getTracesByThreadId`, and mixed trace/thread mapping callback now all construct with blob-resolution deps and pass `{ full: true }`, while protections stay on the same reads.

Validation run locally from the existing PR worktree:
- `pnpm exec vitest run src/server/evaluations/__tests__/runEvaluation.4991-full-resolution.unit.test.ts src/server/evaluations/__tests__/runEvaluation.evaluations-enrichment.unit.test.ts src/server/evaluations/__tests__/runEvaluation.langevals-request.unit.test.ts src/server/evaluations/__tests__/threadMappingResolver.unit.test.ts` — 12 tests passed.
- `pnpm run typecheck` — passed.
- `pnpm run typecheck:tests` — passed.

CI still had some long-running shards and CodeRabbit pending when I checked, so this review is based on the completed local validation plus the relevant changed surface.

langwatch-agent:
Scariest risk: changes blob IO resolution on evaluation reads; a regression slows or breaks reading eval results at scale.
- **Must Check** the full-blob resolution is bounded (no unbounded fetch on large evaluations).
- **Ask Author** read-latency impact vs before on a large evaluation?
Attention map only.

Aryansharma28:
## Follow-up: the new Trace Explorer (tracesV2) wasn't resolving full IO at all

While verifying the original preview fix in this PR against a real 350KB trace, I found the **new** Trace Explorer (tracesV2, the default one) never showed full content on the header/Summary tab — only the legacy explorer did. Tracked it down to two separate issues, both fixed here.

### 1. tracesV2 never wired up full-content resolution

`TraceSummaryService.getByTraceId` had no equivalent of the legacy `TraceService.getById({ full: true })` path. Wired it up the same way `SpanStorageService` already does — through the shared `resolveOffloadedTraces` primitive, service → repository, injected via `presets.ts`. `tracesV2.header` now always requests `{ full: true }` for the single-trace drawer read (never for list/page reads, which stay preview-only).

### 2. Root cause of a silent no-op once wired up

`SpanStorageClickHouseRepository`'s row mapper (`deserializeAttributes`) auto-JSON-parses any attribute value shaped like `{...}`/`[...]`. The ADR-022 `langwatch.reserved.eventref.*` pointer is *itself* a JSON-object-shaped string (`{"field":...,"eventId":...}`), so it was getting silently double-parsed into a real object before the eventref decoder ever saw it — `JSON.parse()` then threw on `"[object Object]"`, caught by the existing "malformed JSON, skip" branch, and resolution quietly no-op'd. The legacy explorer never hit this because its own row mapper keeps `SpanAttributes` as raw strings. Fixed by having reserved eventref keys bypass the rich-value coercion (they're pipeline-internal pointers, never displayed).

### Verified end-to-end, not just via tests

Sent a real 350KB (351,900 byte) trace through the full pipeline (ClickHouse → event_log → tracesV2 read path) and drove the actual UI:

**Drawer opens on the Summary tab, full input renders** (previously this exact trace showed a `[truncated: ...]` placeholder):

drawer input top

**Scrolling through — content keeps going past the old 65KB cutoff:**

scrolled

**The frontend's own inline-render cap kicks in at 100K chars with a "Show remaining 252K chars" affordance (351,900 − 100,000 ≈ 252K) — separate, expected UI behavior, not data loss:**

show remaining

**Clicking it expands to the full 351,900 chars ("Show less" now shown) — no truncation marker anywhere:**

expanded

**Hard proof, not just visual**: called `tracesV2.header` directly from the authenticated browser session and measured the response — `input.length === 351900`, byte-for-byte matching the ClickHouse source, `containsTruncatedMarker: false`. Clicking the actual "Copy to clipboard" button (monkey-patched `navigator.clipboard.writeText` to capture what it sends) also copied exactly 351,900 characters.

**64KB edge case** (right at the old preview boundary) also resolves correctly — full 65,536 bytes, no truncation:

64kb edge case

### Tests

- New unit test in `span-storage.clickhouse.repository.unit.test.ts` covering the exact eventref-pointer-shaped-as-JSON-object case.
- New `trace-summary-full-resolution.unit.test.ts` covering `TraceSummaryService`'s `full` option (deps present/absent, eventref present/absent, missing event_log row, spans-read failure — all fail open to the stored preview).
- New feature file: `specs/traces-v2/trace-header-full-content-resolution.feature`.
- `pnpm typecheck` clean; all directly affected suites green (168 tests).

langwatch/src/server/event-sourcing/pipelines/trace-processing/utils/capOversizedAttributes.ts
```diff
@@ -76,24 +135,56 @@ function truncationPlaceholder(
...
 
   if (typeof value.stringValue === "string") {
     const byteSize = utf8ByteLength(value.stringValue);
     if (byteSize > maxBytes) {
-      value.stringValue = truncationPlaceholder(
-        byteSize,
-        dataUrlMimeType(value.stringValue),
-      );
-      capped = true;
+      const mimeType = dataUrlMimeType(value.stringValue);
```
langwatch-agent, line 191: **P2 — Non-base64 data URLs lose the text preview this change promises.** `dataUrlMimeType()` accepts any `data:` URL with a MIME type, not only `...;base64,`; consequently an oversized `data:text/plain,${encodeURIComponent(userPrompt)}` (or an SVG/text payload sent without base64) takes this binary-placeholder branch and exposes none of its readable content. That contradicts the stated behavior that only base64 data URLs are binary while human-readable text retains a preview. Require the `;base64` flag before taking the placeholder branch, and add a regression test for an oversized non-base64 `data:text/plain,...` value retaining the appropriate preview budget.

langwatch-agent (commented):
1 P2 finding. The full-resolution and redaction paths checked out, and the focused suite passed (105 tests); the required correction is inline.

Aryansharma28 (commented):
## Deep review pass (goated-review)

Did a full read-the-room + dimension-by-dimension audit of this PR (architecture, business logic, regression risk, code quality, YAGNI/DRY/SOLID, security/multitenancy, performance, tests, docs), following on the same treatment given to #5853/#5854/#5865.

**Verdict: no P0/P1 issues. One real P2 bug found and fixed directly via commit. Everything else already addressed on the branch. One P3 architectural observation left for a human call.**

### What I found and fixed myself (commit `2942aa15b`)

**[P2] `capOversizedAttributes.ts` — non-base64 `data:` URLs lost their readable text preview**, matching the finding your own `langwatch-agent` self-review already flagged inline on `capOversizedAttributes.ts:183`. `dataUrlMimeType()` treated *any* `data:<mime>,...` string as binary (image/audio blob), including a non-base64 `data:text/plain,...` or inline SVG payload whose content is fully readable percent-encoded/raw text. Those took the binary-only `[truncated: N bytes, mime]` placeholder instead of the text preview this PR is explicitly about restoring. Fixed by gating the binary branch on the `;base64` flag actually being present (RFC 2397: it's the last parameter before the comma when present) — added a regression test that ingests an oversized `data:text/plain,...` value and asserts it keeps the readable prefix, and added the matching scenario to `specs/event-sourcing/oversized-attribute-value-preview.feature`.

### Already resolved on the branch (not re-raising)

- CodeRabbit's 3 inline findings (remove the `IO_ATTR_KEYS` re-export alias; convert `utf8Preview`/`truncatedTextPreview`/`capAnyValue`/`capAttributeList` to named-parameter objects; rename `capped` → `isCapped`) were all addressed by an earlier commit on this branch (`a0878179e`) before I started — verified each is gone from current `lean-for-projection.ts` / `capOversizedAttributes.ts`.

### P3 — architectural observation (not fixed, real design tradeoff)

The new `tracesV2.header` procedure now unconditionally requests `full: true` on every single-trace drawer open (`trace-summary.service.ts`'s `resolveFullIO`), which issues its own `getNormalizedSpansByTraceId` ClickHouse read. The same drawer-open also fires `tracesV2.spansFull`, which independently reads the same trace's spans again (and, per the pre-existing #4888 design, does its own offload resolution). So opening a trace detail now issues two separate spans-table reads for the same trace instead of one. The code is explicit that `full: true` "costs one spans read per trace" and restricts it to single-trace reads only — a deliberate, documented tradeoff — but the redundancy with `spansFull` wasn't addressed. Not blocking (bounded, single-trace-open-only cost, not a hot path), but worth a follow-up to share one spans fetch between `header` and `spansFull` rather than two independent ClickHouse round-trips per drawer open.

### Multitenancy check

Every new/touched read threads `projectId`/`tenantId` through consistently (`TraceService.getById/getTracesByThreadId`, `TraceSummaryService.resolveFullIO`, `SpanStorageClickHouseRepository.getNormalizedSpansByTraceId` — confirmed `WHERE TenantId = {tenantId:String}` is the first predicate). No gap found.

### Verification

- `pnpm typecheck` clean after merging latest `main` in.
- Full regression sweep: `src/server/app-layer/traces`, `src/server/traces`, `src/server/evaluations`, `src/server/event-sourcing/pipelines/trace-processing/utils` — 112 test files / 2078 tests, all green (including the new non-base64 data-URL regression test).
- CI green on the final commit except the pre-existing, repo-wide `sdk-python-ci` failure (expired `OPENAI_API_KEY` secret, unrelated to this PR — same known issue affecting `main`).

Not approving/merging per instructions — leaving that to a human maintainer.

Aryansharma28 (approved):
Approving on top of the goated-review pass above — CI is fully green on the final commit (`2942aa15b`) except the pre-existing, repo-wide `sdk-python-ci` failure (expired `OPENAI_API_KEY` secret, unrelated to this PR, also failing on `main`). The one real P2 bug found (non-base64 `data:` URLs losing their text preview in `capOversizedAttributes.ts`) is fixed with a regression test and matching spec scenario. Multitenancy checked — all touched reads thread `projectId`/`tenantId` correctly. Not merging — leaving that to a human maintainer, same as the rest of this batch.

langwatch/src/server/api/routers/tracesV2.ts
```diff
@@ -1081,6 +1081,10 @@ export const tracesV2Router = createTRPCRouter({
           visibilityCutoffMs: await getVisibilityCutoffMsForProject(
             input.projectId,
           ),
+          // Single-trace read (the drawer opening) — resolve any offloaded
+          // (ADR-022) input/output in full, matching legacy traces.getById's
+          // unconditional { full: true }. Never set this for a list/page read.
+          full: true,
```
0xdeafcafe, line 1087: do we need this for the header? header should be a cheap read, an input/output/attrs are expensive

0xdeafcafe (approved):
i think we should rewrite this whole handling to do it in a better way, but approving to get the fix out

langwatch/src/server/event-sourcing/pipelines/trace-processing/utils/capOversizedAttributes.ts
Aryansharma28: Fixed in commit `2942aa15b` (already in this merged PR) — `dataUrlMimeType()` now requires the `;base64` flag before taking the binary-placeholder branch, so a non-base64 `data:text/plain,...` (or inline SVG) falls through to the text preview path like any other oversized string. Added a regression test: `given an oversized non-base64 data URL on an IO attribute` in `capOversizedAttributes.unit.test.ts`. Resolving.

langwatch/src/server/api/routers/tracesV2.ts
Aryansharma28: You're right — traced it and `full: true` was genuinely paying for more than the header needed in most callers.

`TraceSummaryService.resolveFullIO` always runs an extra ClickHouse spans read (`getNormalizedSpansByTraceId`) before it can even check whether anything was offloaded — its own doc comment already says 'only ever pass true for a single-trace read (the drawer), never for a list page.' This router endpoint ignored that and hardcoded it for every caller.

Checked every consumer of `tracesV2.header`: the drawer's Summary tab (+ markdown export) genuinely needs full content, but the hover-peek popover only renders a 2-line clamp, Langy's `useResolvedTraceName` only reads the name, the capability-card hydrator fetches N traces and immediately truncates to 90 chars, and conversation-sibling prefetch warms up to 10 headers most of which are never opened.

Opened #6042: `full` is now a required input (no hidden default) so each caller states its intent — drawer stays `true`, the other five flip to `false`. CI green, 1346 tests in the affected areas pass. Landing as a follow-up since this PR had already merged by the time I got to it — resolving this thread there.

Issue #5760, opened 2026-07-13
Scenario runs fail — LLM gateway returns gateway_unavailable/dispatcher_error


drewdrewthis:
## For Humans

**Scenario runs against an Azure OpenAI model fail on every attempt: LangWatch's Go LLM gateway drops the Azure endpoint before calling the provider, so it returns `502 dispatcher_error` / `"endpoint not set"`.** This is a **code bug in our gateway** (an Azure-only key-name mismatch on the `/go/proxy` path) — **not** the customer's config (their endpoint + key are correct), and **not** fixed by the two in-flight gateway PRs #5576 / #5581.

**Ask:** land the Go gateway fix (draft PR incoming, ref this issue). Do **not** close/merge #5576 or #5581 expecting them to resolve this — they fix a real but **separate** Langy-agent problem.

_Status: P1, root-caused + reproduced (failing Go test at the seam). Neither Aryansharma28 PR touches the failing path. Full spec below._

<details><summary>For Agents — full spec</summary>

### Root cause (BLUF)
**Azure-only credential key-name mismatch on the `/go/proxy` dispatch path.** The Go gateway's header parser stores the Azure resource endpoint under `Credential.Extra["api_base"]`, but the Bifrost Azure key-builder reads only `Extra["endpoint"]` (no fallback). So for **every** Azure request dispatched through `nlp-service /go/proxy/v1/*` (scenario User Simulator, playground, prompt-testing), `AzureKeyConfig.Endpoint` is empty and vendored Bifrost returns `NewConfigurationError("endpoint not set")` **before any HTTP call** (`status: 0`). The scenario step exhausts 3 retries and fails.

### Symptoms / context
- `UserSimulatorAgent` → `502 gateway_unavailable` / `meta.reason: dispatcher_error`, `reasons[0]: {type: provider_timeout, message: "endpoint not set", status: 0}` at `http://langwatch-nlp-service/go/proxy/v1/chat/completions`. Every run. Verbatim trace preserved in Evidence below.
- Reporter: endpoint + key correct, model **deployed** their side → gateway-side, not client config. "deployed" / "endpoint" is Azure OpenAI vocabulary; confirmed below the customer is on **Azure**.

### Evidence chain (file:line)
[...]

Issue #5758, opened 2026-07-13
AI scenario generation throws an error


drewdrewthis:
## Symptom
Customer reported an error generating a scenario via the AI create/generate flow (support thread 2026-07-13). Confirmed signature: **`Unexpected token '<', "<!DOCTYPE "... is not valid JSON`** in the "generate scenario with AI" modal — the client ran `JSON.parse()` on an **HTML** response body.

## Status — DIAGNOSED + FIX IN DRAFT PR (2026-07-13)
Evidence gap is **closed** (superseded). Reproduced deterministically and root-caused — see the pinned **🔬 Diagnosis** comment. Fix in draft **#5761**.

- **Reproduced:** the exact error is thrown by `generateScenarioWithAI` when `/api/scenario/generate` returns a non-JSON (HTML) body.
- **Root cause:** (a) the client calls `response.json()` **unconditionally**, so any HTML body becomes a raw `Unexpected token '<'` that masks the real HTTP status; (b) the app server itself **always returns JSON** (route handler + Hono `onError` both `c.json()`; unmatched `/api/*` → JSON 404), so the HTML came from an **upstream layer** (reverse-proxy / gateway 502·504 / timeout, an auth-redirect login page, or an older self-hosted Next.js build) — **not** the route.
- **Fix (#5761):** client degrades gracefully — a non-JSON body now yields an actionable, status-bearing message instead of a parser artifact. No server change (endpoint is already JSON-correct).

## Still needed (to pin the exact upstream failure mode)
Customer's browser **Network tab** for the failing `POST /api/scenario/generate`: **status line + `content-type` + `server`/`via` headers**. That distinguishes 502 vs 504 vs 302-login vs legacy-Next-500. The client fix handles all of them regardless.

## Repro
1. Open Create-new-scenario → "generate with AI".
2. Trigger generation while an upstream layer returns HTML for `/api/scenario/generate` (gateway 502/504/timeout, auth redirect, or an older self-hosted build).
3. **Before #5761:** modal shows `Unexpected token '<', "<!DOCTYPE "... is not valid JSON`. **After #5761:** modal shows a clean, status-bearing message.

drewdrewthis:
## Update: actual error captured

Drew relayed the screenshot directly (support Slack thread is Slack Connect, cross-workspace — API couldn't fetch it, see issue body). Modal is LangWatch's **Create new scenario** "generate with AI" flow. Verbatim error shown to the user:

> Something went wrong while generating your scenario.
[...]

PR #5761, opened 2026-07-13
fix(scenarios): bound generate dispatch so a failing gateway can't surface as HTML (#5758)

drewdrewthis:
> **Reworked after the client-only approach was rejected.** The customer's real problem is that scenario generation **fails**, and the failure surfaces as opaque HTML. This PR fixes the **server-side** failure surface; the fix that makes generation actually succeed is **#5762** (see below). Start here: `langwatch/src/server/routes/scenario-generate.ts`.

## Root cause — traced end-to-end and LIVE-reproduced

`POST /api/scenario/generate` resolves the `scenarios.generator` model and calls the Vercel AI SDK's `generateObject`, which dispatches **through the nlp-service `/go/proxy/v1/chat/completions` gateway** — the *same* path as the scenario User-Simulator (`getVercelAIModel` → `${LANGWATCH_NLP_SERVICE}/go/proxy/v1`).

**So this is #5760 on the generate path.** Reproduced live on this branch (which lacks #5762), with an Azure provider configured:

```
POST /api/scenario/generate  →  HTTP 502  application/json  (6109 ms)
{"error":"gateway_unavailable","domainError":{"kind":"dispatcher_error",...,"httpStatus":502}}

nlpgo log:
  path=/go/proxy/v1/chat/completions  code=gateway_unavailable  reason=dispatcher_error  status=502
    ↳ error_reasons=[provider_timeout (map[message:endpoint not set status:0])]   ← the exact #5760 "endpoint not set"
  (three request_ids — the AI SDK retried the dispatch 3×)
```

Two independent facts fall out of this:

1. **Generation fails** because the Azure `/go/proxy` path is broken → the "endpoint not set" bug **fixed server-side by #5762**. Generate is a co-beneficiary; #5762 is the "generation actually works" fix (for Azure). This PR does **not** duplicate it.
2. **The failure surfaces as HTML** because the app's `generateObject` call was **unbounded**: the AI SDK default (2 retries → **3 attempts**) burned **~6 s** on an instant config error. The app itself always answers with JSON (even the 502 is `application/json`), so the customer's `<!DOCTYPE html>` is an **upstream non-JSON response** (a reverse-proxy / ingress / CDN error or timeout page) served because the app was too slow to answer — not something the app emits. **The exact upstream mode is unconfirmed** — the customer's Network-tab status was never captured (issue #5758 lists it as still-needed). The fix is mode-agnostic, so this doesn't block disposition; only the precise root-cause label is inferred.

## What changed

- **Bound the gateway dispatch** in `scenario-generate.ts`: `maxRetries: 1` (matches the sibling `generateObject` caller `ai-query.ts:228`) + `AbortSignal.timeout(30_000)`. A failing/hung gateway now returns a **fast, clean JSON envelope** — never long enough for a front proxy to substitute HTML. **Provider-agnostic.**
- **Clean 504 on timeout** via a shared `isAbortLikeError` (in `nlpgo/goHandledError.ts`). _Review caught that this branch was initially **dead**: it gated on `instanceof Error`, but a real `AbortSignal.timeout` reason is a `DOMException` (name `"TimeoutError"`) that fails that check, so timeouts silently fell to the generic 500. Now matched by `name` (including the Next.js `ResponseAborted` name), verified live to map a real abort → 504 in ~80 ms._
- **Kept the client `response.json()` guard** (`scenarioGeneration.ts`) as defense-in-depth: if a body is ever non-JSON (a front-proxy page slipping through), the modal shows a status-bearing message, not the raw `Unexpected token '<'`.

### 1. Real-app browser proof (the actual modal)

Before (un-fixed client) vs after (this PR), against a genuinely-non-JSON gateway response:

| Before | After |
|---|---|
| before | after |
| Modal shows the raw **`Unexpected token '<', "<!DOCTYPE "... is not valid JSON`** — the exact customer error | Modal shows a clean, status-bearing message |

Additionally, booting the app with a **real broken Azure gateway** (Azure provider, `scenarios.generator` → `azure/gpt-4o-mini`) and driving **Create new scenario → Generate with AI** now renders a clean, fast, actionable failure with a **Try again** button (live Playwright observation) —
> Something went wrong while generating your scenario.
> `dispatcher_error: gateway_unavailable (reason: dispatcher_error)`

— instead of the raw parse crash, and in **~2 s** instead of a **~6 s** hang.

### 2. Two falsifiable server regression tests (real handler + real AI SDK retry loop)

Both run the real Hono handler + the real AI SDK retry loop against a `MockLanguageModelV3` — the surface that actually gates:
1. **Retry bound** — asserts the dispatch is capped at **2 attempts**. Revert `maxRetries` → `Failed after 3 attempts` → `expected 3 to be 2` (RED).
2. **Real abort → 504** — fires a genuine `AbortSignal.timeout` (short cap) against a signal-honouring hang. Revert the `abortSignal` → the request hangs → the test **times out** (RED). This test found and fixed the dead-504 bug above.

Plus `isAbortLikeError` unit-tested across all three abort names, and the existing 11 client tests.

### 3. Verification scope (honest)

Run **locally**: `vitest` (server 3/3 + client 11/11 + goHandledError incl. new abort tests) and both `pnpm typecheck` / `pnpm typecheck:tests` clean. ⚠️ The repo's CI **unit-test / `typecheck:tests` jobs `skip`** on this PR's path filters, so test-level verification here is **local, not CI-gated** — but the full blast radius of the shared-module change (only `scenario-generate.ts` + the `goHandledError` test import it) was run locally.

## Human verification

1. **Falsifiability (server fix):**
   `cd langwatch && npx vitest run src/server/routes/__tests__/scenario-generate.unit.test.ts` → 3 pass. Delete the `abortSignal:` line in `scenario-generate.ts` → the abort test hangs → RED. Delete `maxRetries:` → the retry test fails `expected 3 to be 2`. Restore → green.
2. **Dead-504 fix:** on `git show 15fcd20c9^:.../scenario-generate.ts` the 504 branch used `instanceof Error`; a `DOMException` abort skips it → generic 500. Current code matches by `name` → 504.
3. **Modal:** in a running app with a broken/misconfigured provider, open **Create new scenario → Generate with AI** → a clean error + **Try again**, not `Unexpected token '<'`, within ~2 s.

## Open question for the owner (blocks final disposition of #5758)

**Which model provider is the customer on?** The support thread is Slack-Connect (unfetchable from here).
- **Azure** → #5762 closes the "can't generate" half; this PR guarantees the failure is never opaque HTML.
- **Non-Azure** → the same `api_base`/`endpoint` key-mismatch class may affect that provider's bifrost branch (the "two dispatch paths diverge" gotcha from #5760) — a targeted follow-up on that branch, tracked under #5758.

**Do not merge** — owner gate (#5758). Non-blocking review follow-ups (generic-500 error exposure; `NextRequestShim as any` ×5; unified AI-SDK-error classification) are catalogued in the review-verdict comment for triage under #5758.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AI scenario generation error handling for gateway failures and non-JSON responses.
  - Added request timeouts and bounded retries to prevent prolonged or failed requests.
  - Return clear JSON error responses, including timeout guidance and HTTP status context.
  - Scenario generation modal now shows retryable errors with a **Try again** option.

- **Tests**
  - Added coverage for gateway outages, timeout behavior, malformed responses, and abort handling.

### DIFF
--- a/langwatch/src/components/scenarios/services/__tests__/scenarioGeneration.unit.test.ts
+++ b/langwatch/src/components/scenarios/services/__tests__/scenarioGeneration.unit.test.ts
@@ -88,6 +88,53 @@ describe("generateScenarioWithAI()", () => {
     });
   });
 
+  describe("when the response body is not JSON (HTML error page)", () => {
+    // Regression for langwatch#5758. The generate endpoint answers with JSON on
+    // every outcome, so a NON-JSON body means the response came from a layer in
+    // FRONT of the app — a reverse-proxy / gateway 502·504, an auth-redirect
+    // login page, a timeout error page, or an older self-hosted Next.js build.
+    // Before the fix, `response.json()` threw a raw
+    // `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` that leaked to
+    // the user and masked the real HTTP status. A real Response is used so the
+    // genuine JSON.parse crash is exercised, not a hand-faked reject.
+    const HTML_ERROR_BODY =
+      "<!DOCTYPE html>\n<html><head><title>502 Bad Gateway</title></head>" +
+      "<body><h1>502 Bad Gateway</h1></body></html>";
+
+    beforeEach(() => {
+      (global.fetch as ReturnType<typeof vi.fn>).mockResolvedValue(
+        new Response(HTML_ERROR_BODY, {
+          status: 502,
+          statusText: "Bad Gateway",
+          headers: { "content-type": "text/html" },
+        })
+      );
+    });
+
+    it("does not leak a raw JSON.parse error to the caller", async () => {
+      const error = await generateScenarioWithAI(
+        "test prompt",
+        "project-123",
+        null
+      ).catch((e: unknown) => e);
+
+      expect(error).toBeInstanceOf(Error);
+      expect((error as Error).message).not.toMatch(
+        /Unexpected token|is not valid JSON|DOCTYPE/i
+      );
+    });
+
+    it("surfaces the HTTP status so the failure is actionable", async () => {
+      const error = await generateScenarioWithAI(
+        "test prompt",
+        "project-123",
+        null
+      ).catch((e: unknown) => e);
+
+      expect((error as Error).message).toContain("502");
+    });
+  });
+
   describe("when API returns an error response", () => {
     it("throws error with message from API", async () => {
       (global.fetch as ReturnType<typeof vi.fn>).mockResolvedValue({
--- a/langwatch/src/components/scenarios/services/scenarioGeneration.ts
+++ b/langwatch/src/components/scenarios/services/scenarioGeneration.ts
@@ -69,27 +69,43 @@ export async function generateScenarioWithAI(
     }),
   });
 
+  // The endpoint answers with JSON on every outcome — 200 success and 4xx/5xx
+  // error envelopes alike. A NON-JSON body therefore did not come from the
+  // route handler; it came from a layer in FRONT of the app: a reverse-proxy
+  // or gateway 502/504, an auth-redirect login page, a timeout error page, or
+  // an older self-hosted build. Parsing it as JSON throws a raw
+  // `Unexpected token '<', "<!DOCTYPE "...` that masks the real HTTP status and
+  // strands the user — convert it into an actionable, status-bearing error
+  // instead (langwatch#5758).
+  let payload: { error?: string; domainError?: unknown; scenario?: unknown };
+  try {
+    payload = await response.json();
+  } catch {
+    const statusText = response.statusText ? ` ${response.statusText}` : "";
+    throw new Error(
+      `The server returned an unexpected response (HTTP ${response.status}${statusText}) instead of scenario data. This is usually temporary — please try again in a moment.`,
+    );
+  }
+
   if (!response.ok) {
-    const error = await response.json();
     const domainError = serializedDomainErrorSchema.safeParse(
-      error.domainError,
+      payload.domainError,
     );
     if (domainError.success) {
       throw new ScenarioGenerationError(
-        error.error || "Failed to generate scenario",
+        payload.error || "Failed to generate scenario",
         domainError.data.kind,
         domainError.data.meta,
       );
     }
-    throw new Error(error.error || "Failed to generate scenario");
+    throw new Error(payload.error || "Failed to generate scenario");
   }
 
-  const data = await response.json();
-  if (!data.scenario) {
+  if (!payload.scenario) {
     throw new Error("Invalid response: missing scenario data");
   }
 
-  const parsed = generatedScenarioSchema.safeParse(data.scenario);
+  const parsed = generatedScenarioSchema.safeParse(payload.scenario);
   if (!parsed.success) {
     throw new Error(`Invalid scenario data: ${parsed.error.message}`);
   }
--- a/langwatch/src/server/nlpgo/__tests__/goHandledError.unit.test.ts
+++ b/langwatch/src/server/nlpgo/__tests__/goHandledError.unit.test.ts
@@ -1,7 +1,7 @@
 import { APICallError, RetryError } from "ai";
 import { describe, expect, it } from "vitest";
 import { DomainError } from "../../app-layer/domain-error";
-import { nlpgoHandledErrorFrom } from "../goHandledError";
+import { isAbortLikeError, nlpgoHandledErrorFrom } from "../goHandledError";
 
 /** The exact envelope nlpgo returned for an unroutable provider prefix. */
 const MISSING_PROVIDER_BODY = JSON.stringify({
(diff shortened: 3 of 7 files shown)


drewdrewthis:
## Review verdict: READY

**Reviewed at:** `b3714f4f` · **Run:** review (own-PR)

Full fan-out ran — 5 checklist reviewers (principles / hygiene / security / test / proof) + 4 design personas (Uncle Bob / Metz-Beck / Fowler / design-soundness). Every blocking finding was **fixed and re-verified in this review round** (commits `15fcd20c`, `b3714f4f`). No unresolved review threads at this SHA.

### Fixed during review (were blocking)
- **[principles][proof][test] The 504 abort test was flag-forwarding — and it masked a dead branch.** It threw a hand-named `TimeoutError`; three reviewers confirmed by revert that deleting `abortSignal` left it green. The 504 branch itself gated on `instanceof Error`, but a real `AbortSignal.timeout().reason` is a `DOMException` (not `instanceof Error`), so real timeouts silently fell to the generic 500. **Fixed:** match aborts by `name`; the test now fires a **real** `AbortSignal.timeout` (falsifiable — delete `abortSignal` → hang → RED). Verified live: real abort → **504 in ~80 ms**.
- **[design-soundness] Dead `RetryError` unwrap.** Verified against `ai@6.0.217`: the SDK re-throws aborts **raw** before wrapping, so an abort is never a `RetryError.lastError` — the unwrap could never fire. **Dropped** it and the comment that invented the case.
- **[design-soundness] Missing Next.js `ResponseAborted`.** `@ai-sdk/provider-utils`' own `isAbortError` includes it; our copy was a strict subset. **Added** it (+ a unit test across all three abort names).
- **[uncle-bob][fowler] Homeless AI-SDK-error knowledge.** `isAbortLikeError` was a general predicate stranded in the route, duplicating the `RetryError` unwrap `nlpgoHandledErrorFrom` owns. **Colocated** it in `nlpgo/goHandledError.ts`; the route imports it; no dup unwrap.
- **[metz-beck][uncle-bob][fowler] Comments outran the code.** The "ops can tune without a deploy" claim (env is start-snapshotted, not in `.env.example`) and the "mirrors trace-api-span-query" claim (that's a `fetch`, not a dispatch) were **corrected** to the real driver (test-drivability). All three confirmed the 3-arm catch is fine — **not** a polymorphism candidate.
- **[test] No `specs/` coverage** → added user-facing scenarios to `ai-create-modal.feature`.
- **[proof] PR-body over-claims** → softened Claim 2 (upstream mode *unconfirmed*), marked the modal render a *live observation* (not a CI test), disclosed unit-test verification is **local** (CI unit-test/`typecheck:tests` jobs skip this PR's path).
- **[principles][test] Cosmetic** → test-name typo; abort `describe` now matches behavior.

- **[security] New Issue** — the pre-existing generic-500 path returns raw `error.message` to the browser; not introduced or worsened here (this PR *reduces* exposure by diverting timeouts to a fixed-string 504). Worth a separate audit of what provider errors carry through.
- **[hygiene] New Issue** — `import type { NextRequestShim as any }` is dead/inert and copy-pasted across 5 route files (`scenario-generate.ts`, `experiments-v3.ts`, `dataset-generate.ts`, `playground.ts`, `workflows.ts`), silently downgrading `as NextRequestShim` → `as any`. Fix all 5 together.
- **[fowler][metz-beck] New Issue** — the two `generateObject` callers classify gateway failure incompatibly (`ai-query.ts` regex-scrapes `err.message` via `summarizeProviderError`; this route uses the structured `nlpgoHandledErrorFrom`), and abort-detection now exists in two shapes (`langevals.http.client.ts` misses the timeout `DOMException`). Unify behind one AI-SDK-failure taxonomy in a follow-up.
- **[test] Decide** — the two retry tests use the SDK's real ~2 s backoff (~4 s total). Kept deliberately: real timing is what makes them genuine outcome tests of the real retry loop; the SDK backoff isn't overridable via public options and fake timers risk undermining the assertion. Under the 30 s test timeout; no flakiness.
- **[fowler] Decide** — the `30_000` default duplicates `trace-api-span-query.ts:46` with no shared source; fine as independent per-subsystem knobs, noted on purpose.

### Disposition note
This PR is the **server-side failure-surface hardening**; the fix that makes Azure generation actually *succeed* is **#5762** (generate is a co-beneficiary of the same `/go/proxy` gateway fix). Final closure of #5758 still turns on the owner confirming the customer's provider (Azure → #5762 closes it; non-Azure → check that provider's bifrost branch).

---

drewdrewthis:
⛔ **DO NOT MERGE — owner rejected this approach** (2026-07-13): client-side graceful-degrade is a band-aid; the customer still cannot generate scenarios. Converted back to draft. Real fix: capture actual HTTP status+body from generate-scenario, root-cause WHY the endpoint returns HTML (404/500/auth-redirect/proxy-misroute), fix the SERVER side so generation works. Hypothesis: may be the #5760 nlp-service 502 bleeding through. This PR will be reworked or superseded.

drewdrewthis:
@drewdrewthis — reworked and ready for your re-review. The client-only band-aid is gone; this is now a **server-side** fix.

**Root cause (live-reproduced):** scenario generation dispatches through the nlp-service `/go/proxy` gateway — **this is #5760 on the generate path**. On a broken Azure gateway it fails with `502 dispatcher_error` / `endpoint not set`, and the AI SDK retried it **3× (~6.1s unbounded)**. The app always returns JSON, so the customer's `<!DOCTYPE` is a **front-proxy 502/504 page** served while the app was still retrying.

**What makes generation actually work is #5762** (the Azure endpoint fix) — generate is a co-beneficiary, not something I re-fixed here. This PR bounds the dispatch (`maxRetries: 1` + 30s abort) so the failure is always a **fast, clean JSON envelope, never HTML** — provider-agnostic. Verified live 6.1s→~2.0s; the real Generate-with-AI modal now shows `dispatcher_error: gateway_unavailable` + a Try-again button instead of the raw parse crash. Falsifiable test included.

**One question gates the final disposition (in the body):** which provider is the customer on? Azure → #5762 + this PR close it. Non-Azure → the same key-mismatch class needs checking on that provider's bifrost branch.

**Do not merge** — yours to gate.

drewdrewthis:
### CI note — `test-unit (2)` green is forged, but the failures are inherited (not from this PR)

Ran the shard-tally compensating control (`verify-ci-shard-tally.sh`) since LangWatch shards can force-exit 0. `test-unit (2)` reports **success via the hard-floor** while carrying failing tests:
- `src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts` (1 failed)
- `src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts` (2 failed in CI)

**These are not this PR's regression.** This PR's diff is 5 files (`nlpgo/goHandledError.ts`, `routes/scenario-generate.ts`, their tests, `specs/scenarios/ai-create-modal.feature`) — none touch analytics/clickhouse or the team router, and neither failing file is in the diff, so my-branch == base for them. `aggregation-builder` fails identically at the base version (inherited); `team.update` passes locally and fails only in CI (flaky/env). Tracked upstream under the C2-hardening work (langwatch#4450).

This PR's own changes are verified locally: server 3/3 + client 11/11 + `goHandledError` (incl. new abort tests), both typechecks clean; the shared-module blast radius (only this route + the `goHandledError` test import it) was run in full.

rogeriochaves (approved):

Issue #5286, opened 2026-07-02
aggregation-builder.test.ts performance.first_token fails on clean main (analytics quantile routing)


drewdrewthis:
## Summary

`aggregation-builder.test.ts > ... performance.first_token` fails on **clean main** (analytics quantile routing), independent of the change that noted it — a test that is red on main. Surfaced in **PR #5072**; never filed.

## Source

> Pre-existing unrelated failure: aggregation-builder.test.ts > ... performance.first_token fails on clean main too (analytics quantile routing), not touched here.

(PR #5072 test plan note.)

## Expected

The full `aggregation-builder.test.ts` suite passes on clean main.

## Actual

`performance.first_token` fails on clean main.

---
*Classification: Bug (test red on main)*
*Filed from a PR follow-up sweep (source: PR #5072). Stub — not-ready until triaged.*

PR #5762, opened 2026-07-13
fix(gateway): resolve Azure endpoint + deployment on the /go/proxy path (#5760)

### LINKED ISSUE #5760: Scenario runs fail — LLM gateway returns gateway_unavailable/dispatcher_error

drewdrewthis:
**Reviewer cockpit**
- **Scariest thing:** this edits the shared `credentialToBifrostKey` Azure branch and moves the deployment-map helper into `domain` — a mistake would mis-route or break Azure for **both** dispatch paths (Studio/workflows **and** `/go/proxy`) and every virtual-key Azure customer. Mitigated: the endpoint read is additive (`endpoint` still wins, `api_base` is only a fallback), the moved helper is byte-for-byte the old logic plus a copy-on-write fix, and every touched package stays green.
- **Start here:** `services/aigateway/adapters/providers/bifrost.go` — the one-line dual-name endpoint read is the actual #5760 fix; everything else is the deployment-map follow-through so Azure works end-to-end.

## Why

Closes #5760. Azure OpenAI calls dispatched through nlp-service `/go/proxy` (scenario **User Simulator**, playground, prompt-testing) failed on **every** attempt with `502 dispatcher_error` / inner `provider_timeout: "endpoint not set"`. Root cause: the Go header parser stored the Azure resource endpoint under `Credential.Extra["api_base"]`, but the Bifrost Azure key-builder read only `Extra["endpoint"]` (no fallback), so `AzureKeyConfig.Endpoint` was always empty and Bifrost errored before any HTTP call. Full diagnosis (evidence chain, PR-coordination verdict, regression window) is in the issue body.

## What changed

- **Azure endpoint now reads both `"endpoint"` and `"api_base"`** (`bifrost.go`) → alternative: translate `api_base`→`endpoint` in the parser (parser-local, fixes only `/go/proxy`) → chose the dual-name read at the consuming seam because it mirrors `credBaseURL`'s existing vLLM tolerance and covers every caller; the control-plane/VK path (which sends `"endpoint"`) is unchanged because `endpoint` still wins.
- **One shared `domain.WithDeploymentSelfMap`**; dispatcheradapter's local `withDeploymentMap` removed and repointed → alternative: duplicate the self-map into the parser → chose the shared helper (no copy-paste); blast radius: both dispatch paths share one implementation now, covered by the moved `TestWithDeploymentSelfMap_*` tests.
- **That shared helper is copy-on-write** (`8b1a3086`) → alternative: leave it writing through `DeploymentMap` → `Credential` is passed by value but the map is a **reference**, so it was mutating the *caller's* map. Harmless today (both call sites reassign), but exporting it into shared `domain` is what makes a silent side effect everyone's problem — the next caller reading its own credential back would see a foreign deployment entry. Flagged independently by CodeRabbit and a design review; pinned by a new `domain/provider_test.go` (the package had no tests of its own) whose no-mutation case is RED against the old implementation.
- **`/go/proxy` captures `x-litellm-deployment` and applies the self-map** (`gatewayproxy/headers.go`) → without it, Azure would clear the endpoint check and then fail `"deployment not found for model X"` at Bifrost's `getModelDeployment`.

## How it works

```
services/aigateway/domain/provider.go
  └─ WithDeploymentSelfMap(cred, bareModel)   NEW — single source: Azure/Bedrock/Vertex
                                              model→deployment self-map (honors Extra["deployment"])
services/aigateway/adapters/providers/bifrost.go
  └─ credentialToBifrostKey (Azure branch)    endpoint = credExtra(cred, "endpoint", "api_base")
services/nlpgo/adapters/gatewayproxy/headers.go
  └─ ParseCredentialFromHeaders               capture x-litellm-deployment → Extra["deployment"];
                                              cred = domain.WithDeploymentSelfMap(cred, BareModel(model))
services/nlpgo/adapters/dispatcheradapter/adapter.go
  └─ dispatch / dispatchStream                now call domain.WithDeploymentSelfMap (local copy deleted)
```
Both nlpgo dispatch paths now build an identical Azure credential (endpoint + deployment) regardless of entry point.

1. Configure an **Azure OpenAI** provider (resource endpoint + API key + a deployed model) in a project.
2. Run a scenario whose **User Simulator** uses that Azure model.
3. **Before** this PR: the run fails with `502 … "endpoint not set"`. **After:** the completion reaches Azure and the scenario step proceeds.

Unit-level (no Azure account needed): `go test ./services/aigateway/adapters/providers/ ./services/nlpgo/adapters/gatewayproxy/` — the two `#5760` tests pass; revert either fix line and they go red.

Headless environment — proof is the terminal transcript below.

**proven (gating surface, end-to-end)** — `TestGatewayProxy_AzureChatCompletion_ResolvesEndpointAndDeployment` drives a real Azure chat completion through the **actual `/go/proxy` HTTP handler → real dispatcher → Bifrost → a stub Azure endpoint**, with a **dotted** model id (`azure/gpt-4.1`). This is the exact surface the scenario User Simulator uses. No real Azure account needed (the endpoint points at a local stub).

```
# WITH the fix — 200; the stub is hit at the Azure deployment URL:
request_completed  path=/go/proxy/v1/chat/completions  status=200
--- PASS: TestGatewayProxy_AzureChatCompletion_ResolvesEndpointAndDeployment (0.09s)
   asserted: upstream received POST /openai/deployments/gpt-4.1/chat/completions   (endpoint + deployment resolved)

# endpoint dual-read REVERTED — reproduces the customer's EXACT #5760 error:
status=502  error_reasons=["provider_timeout (map[message:endpoint not set status:0])"]
body: {"error":{"type":"gateway_unavailable","meta":{"reason":"dispatcher_error"},
       "reasons":[{"type":"provider_timeout","message":"endpoint not set","meta":{"status":0}}]}}
--- FAIL (as expected)

# deployment self-map REVERTED — the latent second failure this PR also closes:
status=502  error_reasons=["provider_timeout (map[message:deployments not set status:0])"]
--- FAIL (as expected)
```

**proven (units, falsifiable)** — reverting each fix line turns its unit test RED, re-applying turns it GREEN:
- `TestCredentialToBifrostKey_Azure_EndpointFromApiBase` — RED (`Endpoint = ""`) before, PASS after the `credExtra(cred,"endpoint","api_base")` dual-read.
- `TestParseCredentialFromHeaders_Azure_BuildsDeploymentMap` / `_HonorsExplicitDeploymentHeader` — RED before, PASS after the self-map.

**proven — no regressions:** all touched packages green (`aigateway/adapters/providers`, `nlpgo/adapters/gatewayproxy`, `dispatcheradapter`, `httpapi`, `tests/integration`); custom/VLLM/DeepSeek/Bedrock + dispatcheradapter Azure tests unchanged and passing; broader `go test ./services/aigateway/... ./services/nlpgo/...` → zero FAIL.

**proven — BOTH gaps close, each falsified independently at the e2e seam (re-verified at this HEAD, not inherited from an earlier run):**

| Gap | Fix disabled → | Fix restored → | Verdict |
|---|---|---|---|
| **1 — endpoint** (`bifrost.go` dual-name read) | `502` · `"endpoint not set"` — the customer's **exact** reported error | `200`, upstream hit at `/openai/deployments/gpt-4.1/chat/completions` | **closed** |
| **2 — deployment** (`domain.WithDeploymentSelfMap`) | `502` · `"deployments not set"` — the predicted **second** outage | `200`, same | **closed** |

Gap 2 matters as much as gap 1: **fixing only the endpoint ships the customer straight into a second failure.** #5760 called this out ("Fixing only (1) leaves Azure broken"), so it is proven here rather than claimed — each fix line was disabled *in isolation* through the real `/go/proxy` handler and produced its own distinct error, then restored to green.

**Scope of the proof, stated plainly:** this is a **backend-only** change with **no UI surface** — it touches the Go gateway (`services/aigateway`, `services/nlpgo`) only. There is no screen to screenshot; the honest proof is the seam.

**A live end-to-end repro was NOT possible — infeasible, not skipped.** The distinction matters, so here it is precisely:

- **What we had:** the credential the owner sourced is the customer's **LangWatch project key** — *not* their Azure credentials. It reads provider **metadata**, which is exactly how the azure-vs-`azure_ai` question got settled. But the API does not return secret **values**, so their Azure key and endpoint cannot be extracted to boot this branch against real Azure.
- **Why prod wouldn't do it either:** running scenarios against their production project would not exercise this branch — prod runs the **old** gateway, so it would merely re-demonstrate the bug, while writing visible simulation data into a paying customer's production project. Invasive, and probative of nothing about the fix.
- **So #5760's original blocker stands:** the customer's *Azure* creds remain not on hand. I did **not** substitute a seam test and call it a live run.

**What the LangWatch key did establish:** the customer's enabled provider reads **`azure`**, **not** `azure_ai`, with **`AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT` both set`** — which is what confirms this PR addresses their actual case rather than a variant.

**Honest caveat, do not read it as more than it is:** that was a **catalog/metadata read, not a direct production DB read**. It would not necessarily render a rogue legacy `azure_ai` row. It is settled by *azure-enabled-carrying-exactly-those-keys* plus support's marked-azure evidence — strong evidence, not a `SELECT`.

**Post-merge verification (the real end-to-end proof, deferred by necessity — not missing by neglect):** after this deploys, confirm the customer's scenario runs actually succeed — no `"endpoint not set"`, no `"deployments not set"`. **Support owns that customer loop.** That proof cannot exist before deploy, because only the deployed gateway runs this code.

### Does this actually fix the reporter? — **yes, now confirmed**

This was the open risk on this PR, so recording how it was settled rather than leaving it implied.

**The reporter's provider reads `azure`, with `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT` both set.** That is the evidenced path: the endpoint ships as `api_base`, the old code read only `endpoint`, and this PR resolves it. **Honest caveat on that read:** it comes from their enabled-provider configuration (azure-enabled carrying exactly those two keys) plus support's marked-azure evidence — **not** from a direct production database read. A rogue legacy `azure_ai` row would not necessarily render in that view. It is strong evidence, not a `SELECT`.

That matters because the alternative was live: had the row been `azure_ai` (a spelling the Go gateway aliases at `headers.go:179` but the TS registry has **no** key for), `modelProviders["azure_ai"]` would be `undefined`, **both** `api_key` and `api_base` would be dropped before the gateway ever saw them, and this PR could not have helped. I could not rule that out from the symptom — I tried, and **failed**: feeding the `azure_ai` shape through the real `/go/proxy` handler returns `502` / `provider_timeout` / `"endpoint not set"` / `status: 0`, **byte-identical** to the reported error. So the split of that alias into **#5788** was a scope call that turned out correct — but it was never an exoneration, and #5788 stays a real (if latent) bug.

### The third gap that isn't: `api_version` unset

The reporter has **no** `AZURE_OPENAI_API_VERSION` set, and `headers.go` used to claim `api_base` **and** `api_version` were both "required for Azure routing" — which would mean this fix resolves the endpoint and deployment only to die on a missing version. **Traced the whole chain; it is safe, and the misleading comment is corrected in this PR:**

1. `gatewayproxy/headers.go` — `if ver := h.Get(headerAPIVersion); ver != ""` → with no version configured the key is **never written** into `Extra`.
2. `bifrost.go` — `if apiVersion, ok := cred.Extra["api_version"]; ok { … }` → **ok-guarded**, so `AzureKeyConfig.APIVersion` is never assigned and stays **nil**.
3. vendored `bifrost/core@v1.4.22 providers/azure/azure.go:264-266` (also `:323-325`, `:510-512`) — `if apiVersion == nil { apiVersion = schemas.NewEnvVar(AzureAPIVersionDefault) }` → the nil default **fires**.
4. `providers/azure/types.go:4` — `AzureAPIVersionDefault = "2024-10-21"`, a real valid version.
5. `validateKeyConfig` checks key-config, endpoint, and deployments — **never** `APIVersion`. There is no "api_version not set" error in that provider.

**The nil is load-bearing**, which is the subtle part: an unconditional assign would store a non-nil EnvVar wrapping `""`, Bifrost's `== nil` check would not fire, and the URL would carry `?api-version=` — the same empty-value shape as gap 1, failing at Azure instead of at the config check. So the test pins the guard, not the default: *given a `/go/proxy` Azure credential with no api_version, `AzureKeyConfig.APIVersion` is nil.* Dropping the guard turns it red with exactly `APIVersion = ""`.

(In practice `prepareLitellmParams` also backfills `DEFAULT_AZURE_API_VERSION` = `2025-04-01-preview` for direct-mode Azure, so a modern version normally ships anyway — the nil path is the belt-and-braces case.)

_Review verdict: three independent reviewers (correctness/design, hygiene/reuse, tests) returned SHIP with **zero MUST-FIX**; see the `<!-- review-verdict -->` comment._

## Deployment Impact

- **Env vars:** none added or changed. `x-litellm-deployment` is a per-request header the control plane already emits — not an env var or secret.
- **Helm values / chart:** none. No `charts/` files touched; no new values, no template changes.
- **Default `helm install` (no overrides):** unchanged for every provider except Azure, which starts resolving its endpoint + deployment through `/go/proxy` instead of failing `"endpoint not set"`. No migration, no new config, no restart-order dependency.
- **BYOC / dataplane:** pure code fix in the Go gateway + nlp-service dataplane; ships with the next gateway/nlp-service image roll and needs no operator action. Self-hosted and BYOC Azure users get the fix on image update.

## Anything surprising?

- The reported error surfaces the **endpoint** gap first; the **deployment-map** gap is a latent *second* failure (`"deployment not found"`) this PR also closes so Azure works end-to-end.
- **PRs #5576 / #5581 (Aryansharma28) are unrelated** — they fix the *Langy coding agent's* worker→gateway URL (`LangyCredentialService.ts` / Helm), not this Azure `/go/proxy` path. Keep them; they will not resolve #5760.
- **A TS-side `azure_ai` alias normalization was split out into #5788** and is no longer part of this PR. The split was the right call — the reporter's provider reads `azure`, so it was never their root cause — but it is **not** an exoneration: #5788 reproduces this exact symptom byte-for-byte and would **not** be fixed by this PR. It stays a real latent bug (`provider` is `z.string()`, so `azure_ai` rows are reachable). #5788 also carries @rogeriochaves's placement question, including the non-obvious constraint that the normalization **cannot** simply move into `toLitellmModelId`/`translateModelIdForLitellm` — those translate model-ID strings and never touch the provider field the credential lookups read, so a naive move would fix routing while still dropping `api_key`/`api_base`.
- Two further follow-ups, both non-blocking and out of scope here: **#5790** (rename `*ForLitellm` → gateway naming, the other half of @rogeriochaves's thread — a wide rename that should not ride a P0) and **#5791** (the e2e test's `azureShim` duplicates `cmd/root.go`'s production shim, per CodeRabbit — extracting it would mean rewiring the nlpgo entrypoint on a customer unblocker).
- `langwatch/.env.example` is untouched by this PR — an earlier revision removed ~51 unrelated lines (`LANGWATCH_ENDPOINT`, the OTLP collector docs, the `make observability` notes) as WIP collateral. @rogeriochaves caught it; reverted in `389aa785`, and `git diff origin/main...HEAD -- langwatch/.env.example` is now empty.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure OpenAI endpoint detection when credentials use either supported endpoint format.
  * Fixed Azure deployment resolution for proxy and in-app requests.
  * Added support for specifying Azure deployments through the `x-litellm-deployment` header.
  * Improved compatibility with custom OpenAI-compatible providers.

* **Tests**
  * Added coverage for Azure endpoint, deployment mapping, proxy routing, and direct-mode configuration scenarios.

### DIFF
--- a/langwatch/.env.example
+++ b/langwatch/.env.example
@@ -108,6 +108,10 @@ REDIS_URL=""
 # LangWatch NLP service powers the optimization studio, topic clustering and sentiment analysis
 LANGWATCH_NLP_SERVICE="http://localhost:5561"
 
+# Internal service-to-service URL the worker uses to POST telemetry back to the app.
+# In Docker compose this is overridden to http://app:5560 by write-dev-overrides.sh.
+LANGWATCH_ENDPOINT="http://localhost:5560"
+
 # LangEvals powers LangWatch evaluators and guardrails: https://github.com/langwatch/langevals/
 LANGEVALS_ENDPOINT="http://localhost:5562"
 
@@ -120,17 +124,55 @@ SENDGRID_API_KEY=
 # Sentry, for error tracking and monitoring
 SENTRY_DSN=
 
-# OpenTelemetry observability for tracing
+# OpenTelemetry observability — export the app's OWN logs/traces/metrics to an
+# OTLP collector (for viewing in Grafana). This is DIFFERENT from ingesting
+# customer telemetry. For local debugging, run `make observability` (a local
+# OTLP Collector + Loki + Tempo + Prometheus + Grafana) and `make
+# observability-connect` sets all of this up for you. See
+# dev/docs/best_practices/local-observability.md.
+#
+# Base OTLP/HTTP collector URL. Exporters append /v1/traces, /v1/logs,
+# /v1/metrics. Set to http://localhost:4318 for the local stack. When set, the
+# TS app exports TRACES.
 OTEL_EXPORTER_OTLP_ENDPOINT=
+# Optional auth headers ("key=value,key2=value2", e.g. "Authorization=Bearer x").
+# Leave empty for the local stack (no auth).
+#OTEL_EXPORTER_OTLP_HEADERS=
 
-# Pino logging levels
-# Console log level (default: "info")
-#PINO_CONSOLE_LEVEL=info
+# Log levels — UNIFIED across the TS app and the Go services (nlpgo, aigateway),
+# so one set here configures both. When the local observability stack is on,
+# keep the console quiet and send the detail to Grafana:
+#   LOG_CONSOLE_LEVEL=warn   → console shows only warnings/errors
+#   LOG_OTEL_LEVEL=debug     → info + debug flow to the collector (Loki)
+#LOG_CONSOLE_LEVEL=warn
+#LOG_OTEL_LEVEL=debug
 
-# LOCAL DEV ONLY: Export logs to OTel collector (for viewing in Grafana)
-# Requires OTEL_EXPORTER_OTLP_ENDPOINT to be set
+# Export LOGS to the collector (→ Loki). Requires OTEL_EXPORTER_OTLP_ENDPOINT.
 #PINO_OTEL_ENABLED=true
-#PINO_OTEL_LEVEL=debug
+
+# Export METRICS to the collector (→ Prometheus). Node/host runtime metrics
+# (CPU, memory, event loop, GC). Requires OTEL_EXPORTER_OTLP_ENDPOINT.
+#OTEL_METRICS_ENABLED=true
+
+# Go services (nlpgo, aigateway) dual-export: their OWN telemetry (traces +
+# logs + metrics) goes to this collector IN ADDITION to the product/customer
+# traces they already send to the LangWatch app. Empty = unchanged (default).
+# Set to http://localhost:4318 for the local stack.
+#OTEL_DEBUG_COLLECTOR_ENDPOINT=
+#OTEL_DEBUG_COLLECTOR_HEADERS=
+
+# Tag all telemetry with the worktree it came from, so an agent can filter to
+# the right one in Grafana when several worktrees share the local collector.
+# Standard OTLP env var, honored by both the TS and Go SDKs. `make
+# observability-connect` sets this automatically.
+#OTEL_RESOURCE_ATTRIBUTES=langwatch.worktree=my-worktree
+
+# Grafana base URL used to turn a trace/span id into a clickable deep link —
+# returned in HTTP error bodies and the Langy "view trace" link so you jump
+# straight to the failing trace/log instead of copying ids around. haven sets this
+# automatically when the local observability stack is up (http://127.0.0.1:3000);
+# in production point it at your Grafana.
+#GRAFANA_BASE_URL=https://grafana.example.com
 
 
 # MODELS
--- a/langwatch/.env.example
+++ b/langwatch/.env.example
@@ -108,10 +108,6 @@ REDIS_URL=""
 # LangWatch NLP service powers the optimization studio, topic clustering and sentiment analysis
 LANGWATCH_NLP_SERVICE="http://localhost:5561"
 
-# Internal service-to-service URL the worker uses to POST telemetry back to the app.
-# In Docker compose this is overridden to http://app:5560 by write-dev-overrides.sh.
-LANGWATCH_ENDPOINT="http://localhost:5560"
-
 # LangEvals powers LangWatch evaluators and guardrails: https://github.com/langwatch/langevals/
 LANGEVALS_ENDPOINT="http://localhost:5562"
 
@@ -124,55 +120,17 @@ SENDGRID_API_KEY=
 # Sentry, for error tracking and monitoring
 SENTRY_DSN=
 
-# OpenTelemetry observability — export the app's OWN logs/traces/metrics to an
-# OTLP collector (for viewing in Grafana). This is DIFFERENT from ingesting
-# customer telemetry. For local debugging, run `make observability` (a local
-# OTLP Collector + Loki + Tempo + Prometheus + Grafana) and `make
-# observability-connect` sets all of this up for you. See
-# dev/docs/best_practices/local-observability.md.
-#
-# Base OTLP/HTTP collector URL. Exporters append /v1/traces, /v1/logs,
-# /v1/metrics. Set to http://localhost:4318 for the local stack. When set, the
-# TS app exports TRACES.
+# OpenTelemetry observability for tracing
 OTEL_EXPORTER_OTLP_ENDPOINT=
-# Optional auth headers ("key=value,key2=value2", e.g. "Authorization=Bearer x").
-# Leave empty for the local stack (no auth).
-#OTEL_EXPORTER_OTLP_HEADERS=
 
-# Log levels — UNIFIED across the TS app and the Go services (nlpgo, aigateway),
-# so one set here configures both. When the local observability stack is on,
-# keep the console quiet and send the detail to Grafana:
-#   LOG_CONSOLE_LEVEL=warn   → console shows only warnings/errors
-#   LOG_OTEL_LEVEL=debug     → info + debug flow to the collector (Loki)
-#LOG_CONSOLE_LEVEL=warn
-#LOG_OTEL_LEVEL=debug
+# Pino logging levels
+# Console log level (default: "info")
+#PINO_CONSOLE_LEVEL=info
 
-# Export LOGS to the collector (→ Loki). Requires OTEL_EXPORTER_OTLP_ENDPOINT.
+# LOCAL DEV ONLY: Export logs to OTel collector (for viewing in Grafana)
+# Requires OTEL_EXPORTER_OTLP_ENDPOINT to be set
 #PINO_OTEL_ENABLED=true
-
-# Export METRICS to the collector (→ Prometheus). Node/host runtime metrics
-# (CPU, memory, event loop, GC). Requires OTEL_EXPORTER_OTLP_ENDPOINT.
-#OTEL_METRICS_ENABLED=true
-
-# Go services (nlpgo, aigateway) dual-export: their OWN telemetry (traces +
-# logs + metrics) goes to this collector IN ADDITION to the product/customer
-# traces they already send to the LangWatch app. Empty = unchanged (default).
-# Set to http://localhost:4318 for the local stack.
-#OTEL_DEBUG_COLLECTOR_ENDPOINT=
-#OTEL_DEBUG_COLLECTOR_HEADERS=
-
-# Tag all telemetry with the worktree it came from, so an agent can filter to
-# the right one in Grafana when several worktrees share the local collector.
-# Standard OTLP env var, honored by both the TS and Go SDKs. `make
-# observability-connect` sets this automatically.
-#OTEL_RESOURCE_ATTRIBUTES=langwatch.worktree=my-worktree
-
-# Grafana base URL used to turn a trace/span id into a clickable deep link —
-# returned in HTTP error bodies and the Langy "view trace" link so you jump
-# straight to the failing trace/log instead of copying ids around. haven sets this
-# automatically when the local observability stack is up (http://127.0.0.1:3000);
-# in production point it at your Grafana.
-#GRAFANA_BASE_URL=https://grafana.example.com
+#PINO_OTEL_LEVEL=debug
 
 
 # MODELS
--- a/langwatch/src/server/api/routers/modelProviders.utils.ts
+++ b/langwatch/src/server/api/routers/modelProviders.utils.ts
@@ -290,6 +290,20 @@ export const prepareLitellmParams = async ({
   modelProvider: MaybeStoredModelProvider;
   projectId: string;
 }) => {
+  // Normalize gateway provider aliases so credential resolution matches the Go
+  // gateway's providerForPrefix aliasing
+  // (services/nlpgo/adapters/gatewayproxy/headers.go). A provider stored as
+  // "azure_ai" (the LiteLLM-era Azure prefix; `provider` is z.string() so it is
+  // reachable via API/import/legacy rows) IS Azure OpenAI. Without this, the
+  // registry lookups below miss (modelProviders["azure_ai"] is undefined), so
+  // the api key AND endpoint are silently dropped — the gateway then 502s
+  // "endpoint not set" even though the customer set the endpoint correctly,
+  // and the Go-side dual-name read (#5762) can't help because api_base is never
+  // emitted. See #5760.
+  if (modelProvider.provider === "azure_ai") {
+    modelProvider = { ...modelProvider, provider: "azure" };
+  }
+
   const params: Record<string, string> = {};
 
   // Normalise the incoming wire value for LiteLLM. After iter 109 two
--- a/langwatch/src/server/api/routers/modelProviders.utils.ts
+++ b/langwatch/src/server/api/routers/modelProviders.utils.ts
@@ -290,20 +290,6 @@ export const prepareLitellmParams = async ({
   modelProvider: MaybeStoredModelProvider;
   projectId: string;
 }) => {
-  // Normalize gateway provider aliases so credential resolution matches the Go
-  // gateway's providerForPrefix aliasing
-  // (services/nlpgo/adapters/gatewayproxy/headers.go). A provider stored as
-  // "azure_ai" (the LiteLLM-era Azure prefix; `provider` is z.string() so it is
-  // reachable via API/import/legacy rows) IS Azure OpenAI. Without this, the
-  // registry lookups below miss (modelProviders["azure_ai"] is undefined), so
-  // the api key AND endpoint are silently dropped — the gateway then 502s
-  // "endpoint not set" even though the customer set the endpoint correctly,
-  // and the Go-side dual-name read (#5762) can't help because api_base is never
-  // emitted. See #5760.
-  if (modelProvider.provider === "azure_ai") {
-    modelProvider = { ...modelProvider, provider: "azure" };
-  }
-
   const params: Record<string, string> = {};
 
   // Normalise the incoming wire value for LiteLLM. After iter 109 two
--- a/services/aigateway/domain/provider.go
+++ b/services/aigateway/domain/provider.go
@@ -74,16 +74,21 @@ func WithDeploymentSelfMap(cred Credential, bareModel string) Credential {
 	default:
 		return cred
 	}
-	if cred.DeploymentMap == nil {
-		cred.DeploymentMap = map[string]string{}
+	if _, present := cred.DeploymentMap[bareModel]; present {
+		return cred
 	}
-	if _, present := cred.DeploymentMap[bareModel]; !present {
-		deployment := bareModel
-		if explicit := cred.Extra["deployment"]; explicit != "" {
-			deployment = explicit
-		}
-		cred.DeploymentMap[bareModel] = deployment
+	deployment := bareModel
+	if explicit := cred.Extra["deployment"]; explicit != "" {
+		deployment = explicit
 	}
+	// Copy on write: cred arrives by value, but DeploymentMap is a reference, so
+	// writing through it would land in the caller's map.
+	next := make(map[string]string, len(cred.DeploymentMap)+1)
+	for model, target := range cred.DeploymentMap {
+		next[model] = target
+	}
+	next[bareModel] = deployment
+	cred.DeploymentMap = next
 	return cred
 }
 
--- a/services/aigateway/domain/provider.go
+++ b/services/aigateway/domain/provider.go
@@ -50,6 +50,43 @@ type Credential struct {
 	DeploymentMap map[string]string
 }
 
+// WithDeploymentSelfMap ensures Azure / Bedrock / Vertex credentials carry a
+// deployment entry for bareModel so Bifrost's per-key readers resolve a
+// deployment ("deployment not found for model X" / "deployments not set"
+// otherwise). By default the model id IS the deployment name
+// (azure/gpt-5-mini → deployment "gpt-5-mini"), so a {bareModel: bareModel}
+// self-map suffices; when the provider defines an explicit deployment (the
+// model id need not equal the deployment name), the control plane / gateway
+// forwards it as Extra["deployment"] and that wins. Non-mapped providers
+// (OpenAI, ...) and an empty bareModel are returned unchanged.
+//
+// Every dispatch path shares this so Azure resolves its deployment identically
+// regardless of entry point: dispatcheradapter (Studio / workflows /
+// runSignature) and the gatewayproxy /go/proxy path (scenario User Simulator,
+// playground). The /go/proxy path previously skipped it, so Azure calls that
+// got past the endpoint check then failed deployment resolution (#5760).
+func WithDeploymentSelfMap(cred Credential, bareModel string) Credential {
+	if bareModel == "" {
+		return cred
+	}
+	switch cred.ProviderID {
+	case ProviderAzure, ProviderBedrock, ProviderVertex:
+	default:
+		return cred
+	}
+	if cred.DeploymentMap == nil {
+		cred.DeploymentMap = map[string]string{}
+	}
+	if _, present := cred.DeploymentMap[bareModel]; !present {
+		deployment := bareModel
+		if explicit := cred.Extra["deployment"]; explicit != "" {
+			deployment = explicit
+		}
+		cred.DeploymentMap[bareModel] = deployment
+	}
+	return cred
+}
+
 // Provider dispatches requests to a specific AI provider.
 type Provider interface {
 	ID() ProviderID
--- a/services/nlpgo/adapters/dispatcheradapter/adapter_test.go
+++ b/services/nlpgo/adapters/dispatcheradapter/adapter_test.go
@@ -269,7 +269,7 @@ func TestCredentialFromHeaders_UnsupportedProvider(t *testing.T) {
 	}
 }
 
-func TestWithDeploymentMap_SelfMapsByDefault(t *testing.T) {
+func TestWithDeploymentSelfMap_SelfMapsByDefault(t *testing.T) {
 	cred := domain.Credential{ProviderID: domain.ProviderAzure}
 	cred = domain.WithDeploymentSelfMap(cred, "gpt-5-mini")
 	if got := cred.DeploymentMap["gpt-5-mini"]; got != "gpt-5-mini" {
@@ -277,7 +277,7 @@ func TestWithDeploymentMap_SelfMapsByDefault(t *testing.T) {
 	}
 }
 
-func TestWithDeploymentMap_HonorsExplicitDeployment(t *testing.T) {
+func TestWithDeploymentSelfMap_HonorsExplicitDeployment(t *testing.T) {
 	// When the model id differs from the Azure deployment name, the control
 	// plane forwards the real deployment as Extra["deployment"]; the model
 	// id must map to it rather than to itself.
@@ -291,7 +291,7 @@ func TestWithDeploymentMap_HonorsExplicitDeployment(t *testing.T) {
 	}
 }
 
-func TestWithDeploymentMap_IgnoresNonMappedProviders(t *testing.T) {
+func TestWithDeploymentSelfMap_IgnoresNonMappedProviders(t *testing.T) {
 	cred := domain.Credential{ProviderID: domain.ProviderOpenAI}
 	cred = domain.WithDeploymentSelfMap(cred, "gpt-5-mini")
 	if cred.DeploymentMap != nil {
--- a/services/nlpgo/adapters/dispatcheradapter/adapter_test.go
+++ b/services/nlpgo/adapters/dispatcheradapter/adapter_test.go
@@ -271,7 +271,7 @@ func TestCredentialFromHeaders_UnsupportedProvider(t *testing.T) {
 
 func TestWithDeploymentMap_SelfMapsByDefault(t *testing.T) {
 	cred := domain.Credential{ProviderID: domain.ProviderAzure}
-	cred = withDeploymentMap(cred, "gpt-5-mini")
+	cred = domain.WithDeploymentSelfMap(cred, "gpt-5-mini")
 	if got := cred.DeploymentMap["gpt-5-mini"]; got != "gpt-5-mini" {
 		t.Errorf("expected self-map, got %q", got)
 	}
@@ -285,7 +285,7 @@ func TestWithDeploymentMap_HonorsExplicitDeployment(t *testing.T) {
 		ProviderID: domain.ProviderAzure,
 		Extra:      map[string]string{"deployment": "my-gpt5-deployment"},
 	}
-	cred = withDeploymentMap(cred, "gpt-5.4")
+	cred = domain.WithDeploymentSelfMap(cred, "gpt-5.4")
 	if got := cred.DeploymentMap["gpt-5.4"]; got != "my-gpt5-deployment" {
 		t.Errorf("expected explicit deployment, got %q", got)
 	}
@@ -293,7 +293,7 @@ func TestWithDeploymentMap_HonorsExplicitDeployment(t *testing.T) {
 
 func TestWithDeploymentMap_IgnoresNonMappedProviders(t *testing.T) {
 	cred := domain.Credential{ProviderID: domain.ProviderOpenAI}
-	cred = withDeploymentMap(cred, "gpt-5-mini")
+	cred = domain.WithDeploymentSelfMap(cred, "gpt-5-mini")
 	if cred.DeploymentMap != nil {
 		t.Errorf("expected no deployment map for OpenAI, got %v", cred.DeploymentMap)
 	}
--- a/services/nlpgo/tests/integration/gateway_proxy_azure_e2e_test.go
+++ b/services/nlpgo/tests/integration/gateway_proxy_azure_e2e_test.go
@@ -1,0 +1,139 @@
+package integration_test
+
+// End-to-end guard for #5760 (P0): an Azure OpenAI chat completion driven
+// through the real nlp-service /go/proxy HTTP surface — the exact surface the
+// scenario User Simulator, playground, and prompt-testing use — must resolve
+// the customer's endpoint AND the model's deployment and actually reach the
+// provider, instead of failing 502 "endpoint not set" before any request
+// leaves the box.
+//
+// Unlike workflow_llm_azure_e2e_test.go (build tag live_azure, real Azure),
+// this test needs NO real Azure account: it points the Azure endpoint at a
+// local stub and asserts Bifrost builds the Azure deployment URL and calls it.
+// It spans the seam two package-local unit tests (gatewayproxy.headers_test +
+// providers.bifrost_test) each cover in isolation — the exact cross-package
+// "api_base" vs "endpoint" disagreement that caused #5760 — in one real run.
+//
+// It uses a DOTTED Azure model id ("azure/gpt-4.1") on purpose: it pins that
+// the model→deployment key survives the TS→Go boundary intact, so a future
+// model-id translation change turns this RED instead of silently re-breaking
+// Azure.
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/langwatch/langwatch/pkg/health"
+	"github.com/langwatch/langwatch/services/aigateway/dispatcher"
+	"github.com/langwatch/langwatch/services/aigateway/domain"
+	"github.com/langwatch/langwatch/services/nlpgo/adapters/httpapi"
+)
+
+// azureShim adapts *dispatcher.Dispatcher to httpapi.DispatcherShim — the
+// same mechanical field-copy cmd/root.go's playgroundDispatcherShim does,
+// replicated here so the test wires the REAL dispatcher (Bifrost-backed)
+// behind the real /go/proxy handler.
+type azureShim struct{ disp *dispatcher.Dispatcher }
+
+func (s azureShim) Dispatch(ctx context.Context, req httpapi.DispatchRequest) (*httpapi.DispatchResponse, error) {
+	resp, err := s.disp.Dispatch(ctx, dispatcher.Request{Type: req.Type, Model: req.Model, Body: req.Body, Credential: req.Credential})
+	if err != nil {
+		return nil, err
+	}
+	return &httpapi.DispatchResponse{StatusCode: resp.StatusCode, Body: resp.Body, Headers: toHeader(resp.Headers)}, nil
+}
+
+func (s azureShim) DispatchStream(ctx context.Context, req httpapi.DispatchRequest) (httpapi.DispatchStream, error) {
+	return s.disp.DispatchStream(ctx, dispatcher.Request{Type: req.Type, Model: req.Model, Body: req.Body, Credential: req.Credential})
+}
+
+func (s azureShim) Passthrough(ctx context.Context, req httpapi.PassthroughDispatchRequest) (*httpapi.DispatchResponse, error) {
+	resp, err := s.disp.Passthrough(ctx, dispatcher.PassthroughRequest{
+		Request: dispatcher.Request{Type: req.Type, Model: req.Model, Body: req.Body, Credential: req.Credential},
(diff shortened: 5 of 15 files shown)


drewdrewthis:
## Review verdict: READY

**Reviewed at:** `621d47a54` · **Run:** review (own-PR) · 5 independent reviewer passes · P0 (#5760)

No blocking concerns. All review threads resolved at this SHA (5/5).

**Scope of this run.** The previous verdict certified `cf0457109`. Since then the Go/spec files were verified **byte-identical** to that reviewed state, so re-review was scoped to what genuinely changed: the added TS direct-mode test, the `azure_ai` revert, and the two new commits (`8b1a3086`, `621d47a5`).

**Reviewers & verdicts — every finding mutation-verified, not read:**

| Reviewer | Scope | Verdict |
|---|---|---|
| **[principles-reviewer]** | correctness/design @ `2151a982` | SHIP, 0 MUST-FIX — revert clean, zero `azure_ai` refs remain, `modelProviders.utils.ts` has zero net change; `pnpm typecheck` + `typecheck:tests` both exit 0 |
| **[test-reviewer]** | the added TS test @ `2151a982` | SHIP, 0 MUST-FIX — proved falsifiability by breaking `registry.ts`'s `endpointKey`: only the new test went red. Confirmed the shared `baseAzureProvider` fixture uses stale keys the registry never reads, so this path genuinely had zero prior coverage |
| **[hygiene-reviewer]** | reuse/scope @ `2151a982` | SHIP, **1 MUST-FIX → FIXED in `621d47a5`**: comments still named `withDeploymentMap` "in dispatcheradapter", a function that no longer exists post-rename |
| **[principles-reviewer]** | the COW + api_version delta @ `621d47a5` | SHIP, 0 MUST-FIX — verified nil-map/early-return/preserve-existing all correct, and that no code anywhere in the module writes to `DeploymentMap` |
| **[test-reviewer]** | the new tests @ `621d47a5` | SHIP, 0 MUST-FIX — reverted each guard and recorded RED/GREEN per test; no vacuous tests |

**Bugs found and fixed during review** (not deferred):
- **[coderabbit + principles-reviewer, independently]** `domain.WithDeploymentSelfMap` mutated the **caller's** `DeploymentMap` — `Credential` is by-value but the map is a reference. Latent (both call sites reassign) but it had just been **exported into shared `domain`** for every dispatch path to reuse. Fixed copy-on-write in `8b1a3086`, pinned by a new `domain/provider_test.go` (the package had **no** tests of its own). Reverting the fix turns that test red with the exact leaked map.

**Prove-it — both #5760 gaps falsified independently at the e2e seam:**
- Disable the endpoint dual-read → `502 "endpoint not set"`, the customer's **exact** reported error. Restore → 200, upstream hit at `/openai/deployments/gpt-4.1/chat/completions`.
- Disable the deployment self-map → `502 "deployments not set"`, the predicted **second** outage (#5760: *"Fixing only (1) leaves Azure broken"*). Restore → 200.
- **Gap 3 (`api_version` unset) traced closed**, not assumed: the `ok`-guard leaves `APIVersion` **nil**, which is what makes vendored Bifrost apply `AzureAPIVersionDefault` (`2024-10-21`); `validateKeyConfig` never checks it. Pinned in `621d47a5` — dropping the guard yields `APIVersion = ""`, which would silently skip that default and emit `?api-version=`.

**Does it fix the reporter — confirmed:** their provider reads **`azure`** (not `azure_ai`), with `AZURE_OPENAI_API_KEY` + `AZURE_OPENAI_ENDPOINT` set. Caveat stated in the body: catalog/metadata read, **not** a prod DB read. A live repro is **infeasible** (the available credential is a LangWatch project key, not Azure creds) — post-merge verification is the real end-to-end proof and support owns that loop.

### Non-blocking (New Issue)
- **[coderabbit]** `azureShim` duplicates `cmd/root.go`'s production shim → **#5791** (extracting it means rewiring the nlpgo entrypoint on a P0). (New Issue)
- **[rogeriochaves]** `*ForLitellm` → gateway naming → **#5790**; the `azure_ai` alias behaviour + placement → **#5788**. (New Issue)
- **[test-reviewer]** two dispatcheradapter cases are now confirmed duplicates of `domain/provider_test.go` (same function, same scenario); `IgnoresNonMappedProviders` is **not** duplicated — keep it. No case covers the `bareModel == ""` branch. (New Issue)
- **[shard-tally]** `test-unit (2)` reports green while masking 3 failing tests. **Proven inherited** — identical failures reproduce on pristine `origin/main` (`cc2122fa6`) with none of this branch present. Tracked: **#5286**, **#5792**; mechanism **#4450**. Not introduced here. (New Issue)

---
_Verdict is prose, not a GitHub approval. Owner gates the merge._

drewdrewthis:
**Coordination — is this the #5760 / #5762 Azure `/go/proxy` "endpoint not set" bleeding through?** (owner asked me to check when #5761 was rejected)

**Confirmed: generate-scenario dispatches through the exact same nlp-service `/go/proxy` path as the #5760 scenario-run failure.**
- `src/server/routes/scenario-generate.ts:114` → `getVercelAIModel({ featureKey: "scenarios.generator" })` → `generateObject({ model, … })`.
- `getVercelAIModel` (`src/server/modelProviders/utils.ts`) builds `x-litellm-*` headers via `prepareLitellmParams` and points `baseURL` at `${LANGWATCH_NLP_SERVICE}/go/proxy/v1` — the **same** dispatch surface as the scenario User Simulator.

So **if this customer's `scenarios.generator` (or default) model is Azure, generate-scenario hits the identical `502 gateway_unavailable` / `dispatcher_error` / `"endpoint not set"` that #5762 fixes** — the Go gateway was dropping the Azure endpoint (`api_base` written, `endpoint` read) and never building the deployment map. #5762 restores both (proven end-to-end: reverting the fix reproduces the exact 502).

**What #5762 does / does not resolve for #5758 — being precise:**
- ✅ **Removes the underlying Azure LLM failure at the source.** An Azure generator that 502'd will now reach Azure and return a scenario. If the customer's generator is Azure, this is the root fix, not a band-aid.
- ⚠️ **Does not itself produce the `Unexpected token '<'` HTML symptom.** `/api/scenario/generate` **catches** the gateway error and returns **JSON** (`scenario-generate.ts:128-150`: `nlpgoHandledErrorFrom → c.json`, else `c.json(500)`), matching #5761's finding that the route is always-JSON. The `<!DOCTYPE>` therefore comes from an **upstream layer**. `"endpoint not set"` fails **fast** (status 0, no network call), so an upstream **timeout→HTML** is only plausible under a *short* proxy timeout spanning `generateObject`'s retry backoff — **unverified** without the Network-tab capture #5758 already requests (status + `content-type` + `server`/`via`).

**Net for the owner's call:** #5762 is the root fix for the *shared LLM failure*; if the customer's generator is Azure and the HTML was a proxy rendering of that 502, #5762 removes the trigger and #5758 should clear at the source. A client-robustness guard remains independently useful only for the *other* upstream-HTML cases (auth redirect, legacy build). Fastest confirmation: the customer's Network tab on the failing `POST /api/scenario/generate`, plus which provider `scenarios.generator` resolves to for this project.

_(Posted to both #5762 and #5758 for the coordination trail. — driving #5762; will not merge.)_

drewdrewthis:
⛔ **DO NOT MERGE as-is — owner skeptical** (2026-07-13): this fix reads the Azure endpoint from api_base as a FALLBACK, but the reporter's endpoint IS set correctly and the model IS deployed — if the endpoint is genuinely set, the fallback never triggers, so this would NOT fix their case. Remaining DoD: (1) reproduce the ACTUAL failure — endpoint SET correctly yet dispatcher says 'endpoint not set'; (2) prove this PR fixes THAT case end-to-end, OR find the real root cause (why a correctly-set endpoint doesn't reach the dispatcher) and fix that. CodeRabbit thread r3572353846 (test azureShim dup) is a trivial nit, not the blocker.

drewdrewthis:
## Repro + fix-proof for the owner's challenge (NOT re-asserting review-ready — that's your call)

You were right to push on this. The "missing vs set" framing was the crux, so I reproduced it empirically at **both** language boundaries. Verdict: **there is only one failure variant, and #5762 covers it** — but the reason is subtler than "missing endpoint."

### The key correction: the endpoint is never in the field Bifrost read

On the `/go/proxy` path the Go header parser **only ever writes `Extra["api_base"]`** — it never writes `Extra["endpoint"]` (`gatewayproxy/headers.go`). The old Bifrost key-builder read **only** `Extra["endpoint"]`. So "endpoint not set" does **not** mean the customer didn't set it — it means the correctly-set endpoint arrived under `api_base` and the gateway looked in `endpoint`. There is no "missing-endpoint variant" vs "set variant"; there is one variant: **set-but-read-from-the-wrong-field.**

### Repro 1 (TS, new test) — a correctly-configured Azure provider DOES emit the endpoint as `api_base`

`prepareLitellmParams` with the **real registry key** `AZURE_OPENAI_ENDPOINT` (what the Model-Providers UI stores):
```
customKeys: { AZURE_OPENAI_API_KEY: "…", AZURE_OPENAI_ENDPOINT: "https://acme.openai.azure.com" }
  → params.api_base === "https://acme.openai.azure.com"   ✅ (new test, run green)
```
Every param → `x-litellm-<key>` (`model.factory.ts`), so the endpoint ships as `x-litellm-api_base`. (Gateway mode is the same: `AZURE_API_GATEWAY_BASE_URL` → `api_base`.) The **only** way `api_base` is absent is if neither endpoint key is set — i.e. the customer didn't configure an endpoint, which contradicts "endpoint set correctly." So for **every** correctly-configured Azure provider, the endpoint is present as `api_base`.
_(Note: no prior test caught this — the shared `baseAzureProvider` fixture uses stale key names `AZURE_API_KEY`/`AZURE_API_BASE` that `getModelOrDefaultEndpointKey` does **not** read, so the endpoint→api_base path was never asserted. New commit `f3cb2f575` adds the direct-mode assertion.)_

### Repro 2 (Go, e2e) — endpoint SET correctly, dispatcher STILL errors "endpoint not set" (without the fix)

`TestGatewayProxy_AzureChatCompletion_ResolvesEndpointAndDeployment` drives the **real `/go/proxy` handler → dispatcher → Bifrost** with exactly that `x-litellm-api_base` (endpoint set correctly), against a stub Azure endpoint:
```
# OLD code (read only Extra["endpoint"]) — endpoint IS set (api_base), yet:
status=502  reasons=["provider_timeout (message: endpoint not set, status:0)"]   ← the customer's exact error
# WITH #5762 (read endpoint || api_base):
status=200  → upstream hit at /openai/deployments/gpt-4.1/chat/completions        ← endpoint + deployment resolved
```

### Net

- **Repro 1 + Repro 2 chained = your exact scenario:** endpoint set correctly (Repro 1: it's in `api_base`) → dispatcher errors "endpoint not set" on the old code (Repro 2) → #5762 resolves it (Repro 2). Both are committed, falsifiable tests.
- **#5762 is not a missing-endpoint fallback** — it makes the Azure key-builder read the field the endpoint actually arrives in on this path (mirrors the `credBaseURL` dual-read already used for vLLM). It also fixes the latent next failure (`deployments not set`).

### What would change my conclusion

If the customer's provider is configured in a way I didn't reproduce — e.g. the endpoint stored under a non-registry key, or a custom `x-litellm-*` set that omits `api_base` — my fix would not help and the real bug is upstream (endpoint never emitted). **The one input that would settle it definitively: the actual `x-litellm-*` headers on their failing `POST /go/proxy/v1/chat/completions`, or their exact Azure provider config (which keys are populated).** Everything I can reproduce from the code + UI-stored keys says the endpoint arrives as `api_base` and #5762 covers it.

CodeRabbit r3572353846 — agreed, trivial nit; I'll fold it in, it's not the blocker.

_(I'm holding at NOT-review-ready per your directive; the prior READY verdict is now stale on `f3cb2f575` by design. NEVER merging — your gate.)_

langwatch/.env.example
```diff
@@ -108,10 +108,6 @@ REDIS_URL=""
 # LangWatch NLP service powers the optimization studio, topic clustering and sentiment analysis
 LANGWATCH_NLP_SERVICE="http://localhost:5561"
 
-# Internal service-to-service URL the worker uses to POST telemetry back to the app.
-# In Docker compose this is overridden to http://app:5560 by write-dev-overrides.sh.
-LANGWATCH_ENDPOINT="http://localhost:5560"
```
rogeriochaves, line 113: why are we removing this? you sure it's not needed for eg the gateway to work fine when doing pnpm dev?

langwatch/src/server/api/routers/modelProviders.utils.ts
```diff
@@ -290,6 +290,20 @@ export const prepareLitellmParams = async ({
...
+  // "azure_ai" (the LiteLLM-era Azure prefix; `provider` is z.string() so it is
+  // reachable via API/import/legacy rows) IS Azure OpenAI. Without this, the
+  // registry lookups below miss (modelProviders["azure_ai"] is undefined), so
+  // the api key AND endpoint are silently dropped — the gateway then 502s
+  // "endpoint not set" even though the customer set the endpoint correctly,
+  // and the Go-side dual-name read (#5762) can't help because api_base is never
+  // emitted. See #5760.
+  if (modelProvider.provider === "azure_ai") {
+    modelProvider = { ...modelProvider, provider: "azure" };
+  }
```
rogeriochaves: we literally have `toLitellmModelId` and `translateModelIdForLitellm` functions below, don't this if belong in there? Or, perhaps we should even rename all this "ForLitellm" to "ForGateway", since we have replaced all litellm with our own gateway right

rogeriochaves (approved):

rogeriochaves (changes requested):
non-blocking changes actually if you want to push this through, but marking as request changes so they are captured by the ai agent

langwatch/.env.example
drewdrewthis: You were right to flag it — this was accidental WIP collateral, not an intentional removal. Nothing here relates to the Azure fix; it got swept in by a "preserve work before shutdown" commit.

Reverted in `389aa785` — `langwatch/.env.example` is now byte-identical to `main` (`git diff origin/main...HEAD -- langwatch/.env.example` is empty). `LANGWATCH_ENDPOINT`, the OTLP collector docs, and the `make observability` notes are all back.

langwatch/src/server/api/routers/modelProviders.utils.ts
drewdrewthis: Fair point, and it made me re-check whether this block belongs in the PR at all — it does not, so I have **split it out entirely** into #5788 and reverted it here (`2151a982`).

**Why it left this PR:** #5760's investigation explicitly rules out the TS side ("endpoint IS sent as `api_base` — confirmed"). The real root cause is Go-side: the endpoint ships as `api_base` and the Bifrost Azure key-builder only read `endpoint`. That fix (`a5dd1dd5`) is what actually unblocks the customer, and it does not depend on this block. The `azure_ai` normalization is a plausible *separate latent* bug, but nothing proves any row actually stores `azure_ai` — and its unbound `@scenario` annotation is what was making `feature-parity` red. Unproven scope should not gate a P0.

**On placement** — I dug into this for the follow-up, and the answer is subtler than it looks, so I recorded it in #5788 rather than hand-waving it. Those two helpers translate the **model-ID string** (`:300`, `:304`) — they return a string and never touch `modelProvider.provider`. But the credential lookups read `modelProvider.provider` **directly** against the registry:

- `getModelOrDefaultApiKey` → `modelProviders[...]` at `:219`, called at `:309`
- `getModelOrDefaultEndpointKey` → `modelProviders[...]` at `:230`, called at `:313`

So moving the normalization into `toLitellmModelId`/`translateModelIdForLitellm` would fix routing (`params.model` → `azure/…`) while `api_key` and `api_base` stayed empty — arguably worse, since the request would *look* correctly routed but still be uncredentialed. Normalization has to hit the provider field before those lookups run. That constraint is written up in #5788.

**On `ForLitellm` → `ForGateway`:** agreed in principle, but it is a wide rename across unrelated call sites and I do not want it riding a P0 customer unblocker. Not carried into #5788 either — happy to file it separately if you want it.

services/nlpgo/tests/integration/gateway_proxy_azure_e2e_test.go
```diff
@@ -0,0 +1,139 @@
...
+	})
+}
+
+func toHeader(m map[string]string) http.Header {
+	h := http.Header{}
+	for k, v := range m {
+		h.Set(k, v)
+	}
+	return h
+}
```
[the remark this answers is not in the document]
drewdrewthis: Valid observation — `azureShim` is a mechanical copy of `cmd/root.go`'s `playgroundDispatcherShim`, and the drift risk you describe is real.

Not taking it here, deliberately. The extraction means adding a new shared package and rewiring `cmd/root.go` — the production nlpgo entrypoint — on a P0 customer unblocker whose whole remaining problem has been unrelated scope riding along. The duplication is a test-local adapter, not a behavioral defect: the e2e test wires the real Bifrost-backed dispatcher behind the real `/go/proxy` handler today, which is what this test exists to guard. Trading entrypoint risk for a trivial-severity cleanup is the wrong call on this PR.

Worth doing on its own — just not as a rider on this one.

langwatch/src/server/api/routers/modelProviders.utils.ts
drewdrewthis: Closing the loop on both halves of your comment, plus a correction to what I told you earlier.

**Placement half — moot on this PR.** The `azure_ai` block is gone from #5762 entirely (`2151a982`), so there is no longer an `if` here to place. The behaviour, and your placement question, are tracked in **#5788** — with the constraint I found while writing it up: those two helpers translate the **model-ID string** (`:300`, `:304`) and never touch `modelProvider.provider`, but the credential lookups read that field directly (`:219`→`:309`, `:230`→`:313`). So moving the normalization into them would fix routing while still dropping `api_key`/`api_base` — a worse failure, since the request would look correctly routed but be uncredentialed. It has to hit the provider field before those lookups run.

**Rename half — tracked in #5790.** Agreed `*ForLitellm` is misleading now that the gateway replaced LiteLLM; it is a wide rename and should not ride a P0 customer unblocker.

**Correction — I was wrong to imply the split settles anything.** I told you the alias was "not #5760's root cause" on the strength of the issue's "TS side ruled out" line. I then tested it: feeding the `azure_ai` shape through the real `/go/proxy` handler returns `502` / `provider_timeout` / `"endpoint not set"` / `status: 0` — **byte-identical** to the reported error. The symptom cannot tell the two apart, and that "ruled out" line was verified by reading the code path *for a provider that resolves in the registry* — it never read the affected row. So the split is a **scope** call (unproven work should not block a P0 whose own fix is proven), **not** an exoneration. If that row turns out to be `azure_ai`, #5788 is the real P0 and this PR does not fix the reporter. I could not obtain the value — no prod access from the build box — so it is flagged UNVERIFIED in the PR body rather than inferred.

services/nlpgo/tests/integration/gateway_proxy_azure_e2e_test.go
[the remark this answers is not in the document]
drewdrewthis: Tracked as **#5791** — non-blocking, and not taken on this PR for the reason above (extracting the shim means rewiring `cmd/root.go`, the production nlpgo entrypoint, on a P0 customer unblocker).

services/aigateway/domain/provider.go
```diff
@@ -50,6 +50,43 @@ type Credential struct {
...
+		cred.DeploymentMap = map[string]string{}
+	}
+	if _, present := cred.DeploymentMap[bareModel]; !present {
+		deployment := bareModel
+		if explicit := cred.Extra["deployment"]; explicit != "" {
+			deployment = explicit
+		}
+		cred.DeploymentMap[bareModel] = deployment
+	}
+	return cred
```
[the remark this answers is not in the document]
drewdrewthis: Real, and fixed in `8b1a3086` — thanks. A design review flagged the same aliasing independently, which is a good signal.

Confirmed the mechanism before changing anything: `Credential` arrives by value, so the `DeploymentMap == nil` branch was safe (it only rebinds the local copy), but on a **non-nil** map `cred.DeploymentMap[bareModel] = deployment` writes straight through the shared reference into the caller's map. Latent today because both call sites reassign the result — but this helper was just **exported into shared `domain`** so every dispatch path reuses it, which is exactly when the contract should stop lying.

Took the copy-on-write shape you proposed. Also dropped the nil-branch entirely: a map read on a nil map is already safe in Go, so the clone covers both cases.

Pinned with a falsifying test — `domain` had no test file of its own despite now owning this helper:

```
--- FAIL: TestWithDeploymentSelfMap_DoesNotMutateCallersMap
    provider_test.go:22: WithDeploymentSelfMap wrote into the caller's map:
        got map[gpt-4.1:gpt-4.1 gpt-5-mini:gpt-5-mini], want the map it was handed to be untouched
```
RED against the old implementation, GREEN after. All five touched packages stay green.

services/nlpgo/adapters/dispatcheradapter/adapter_test.go
```diff
@@ -271,7 +271,7 @@ func TestCredentialFromHeaders_UnsupportedProvider(t *testing.T) {
 
 func TestWithDeploymentMap_SelfMapsByDefault(t *testing.T) {
 	cred := domain.Credential{ProviderID: domain.ProviderAzure}
-	cred = withDeploymentMap(cred, "gpt-5-mini")
+	cred = domain.WithDeploymentSelfMap(cred, "gpt-5-mini")
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in `8b1a3086`. The three `TestWithDeploymentMap_*` names are now `TestWithDeploymentSelfMap_*`, and I swept the same rename's other leftovers while I was in there — `workflow_llm_azure_e2e_test.go:8,123` and `translator_test.go:387` still described the logic as `withDeploymentMap` "in dispatcheradapter", a function that no longer exists at that name or location. Those now point at `domain.WithDeploymentSelfMap`. Repo-wide grep for the old name is clean.

rogeriochaves (approved):

PR #5767, opened 2026-07-13
fix(ci): fail hard-floor test shards closed

### LINKED ISSUE #4450: Integration test `api-keys-scope-filter` wedges shard 4 with an infinite render loop; masked green by globalSetup hard-floor process.exit(0)

jstar0:
## Why

The unit and integration test hard floors currently call `process.exit(0)`. If a shard cannot finalize, that success code can hide unfinished test files and already-recorded failures behind a green CI result.

Closes #4450

## What changed

- move the shared hard-floor timer behavior into one helper;
- exit with status 1 when either hard floor expires;
- keep the existing 4-minute unit and 20-minute integration ceilings;
- update the setup comments and logs to describe fail-closed behavior;
- add runtime regressions for both setup paths immediately before and exactly at each timeout;
- repair two stale shard-2 tests that the fail-closed signal exposed.

## How it works

Both global setup paths schedule the same unref'd timer through a shared helper. Healthy shards still exit normally before the timer fires; a shard that remains alive at its existing ceiling now logs the hard-floor event and terminates with status 1.

### Follow-up from the first fail-closed CI run

The first CI run did what this change is intended to do: instead of reporting shard 2 as green, it exposed three pre-existing test failures that were already present on `main`.

- Updated the `performance.first_token` p90 assertion to expect `quantileTDigest`, matching the production contract for `performance.*` percentile metrics.
- Stubbed `auditLog` in the isolated `team.update` unit test so the tRPC audit middleware does not use the process-wide Prisma client. The test continues to exercise the real team update and RoleBinding synchronization paths.
- Kept the unit hard floor at four minutes. The failed GitHub run showed test failures followed by an unfinished shard, so any ceiling change is deferred until a clean runner-specific retry proves it is needed.

## Test plan

```text
pnpm test:unit src/__tests__/test-unit-global-setup.unit.test.ts src/__tests__/test-integration-global-setup.unit.test.ts src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts --run
pnpm check:feature-parity
pnpm typecheck
pnpm typecheck:tests
```

Run these commands from `langwatch/`. The regression tests use fake timers and the real unit/integration global setup functions. The integration test supplies CI service URLs and returns before starting containers or migrations.

## Anything surprising?

The fail-closed behavior surfaced failures that the previous `process.exit(0)` path had been masking. The follow-up test changes are test-only corrections for those inherited failures; they do not change analytics query generation or team membership behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * CI test shards now exit with a non-zero status when they exceed the configured “hard-floor” timeout instead of lingering past the threshold.
  * Unit shards are limited to 4 minutes; integration shards are limited to 20 minutes.

* **Tests**
  * Added/extended unit and integration regression coverage to verify behavior just before the cutoff and at the exact timeout.
  * Added CI feature scenarios to validate hard-floor timing for both unit and integration shards.

### DIFF
--- a/langwatch/src/test-shard-hard-floor.ts
+++ b/langwatch/src/test-shard-hard-floor.ts
@@ -1,0 +1,16 @@
+interface TestShardHardFloorOptions {
+  timeoutMs: number;
+  message: string;
+}
+
+export function scheduleTestShardHardFloor({
+  timeoutMs,
+  message,
+}: TestShardHardFloorOptions): void {
+  const timer = setTimeout(() => {
+    // eslint-disable-next-line no-console
+    console.error(message);
+    process.exit(1);
+  }, timeoutMs);
+  timer.unref();
+}
--- a/langwatch/src/__tests__/test-integration-global-setup.unit.test.ts
+++ b/langwatch/src/__tests__/test-integration-global-setup.unit.test.ts
@@ -1,0 +1,47 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { setup } from "../server/event-sourcing/__tests__/integration/globalSetup";
+
+describe("Feature: Test shard hard-floor failures", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.stubEnv("CI", "true");
+    vi.stubEnv("CI_CLICKHOUSE_URL", "http://localhost:8123");
+    vi.stubEnv("CI_REDIS_URL", "redis://localhost:6379");
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    vi.unstubAllEnvs();
+    vi.restoreAllMocks();
+  });
+
+  describe("when the hard-floor timeout expires", () => {
+    it("keeps the test process running before twenty minutes", async () => {
+      const exit = vi
+        .spyOn(process, "exit")
+        .mockImplementation((() => undefined) as typeof process.exit);
+      vi.spyOn(console, "error").mockImplementation(() => undefined);
+      vi.spyOn(console, "log").mockImplementation(() => undefined);
+
+      await setup();
+      vi.advanceTimersByTime(20 * 60 * 1000 - 1);
+
+      expect(exit).not.toHaveBeenCalled();
+    });
+
+    /** @scenario An integration test shard hard floor exits with failure after twenty minutes */
+    it("exits the test process with a failure status at twenty minutes", async () => {
+      const exit = vi
+        .spyOn(process, "exit")
+        .mockImplementation((() => undefined) as typeof process.exit);
+      vi.spyOn(console, "error").mockImplementation(() => undefined);
+      vi.spyOn(console, "log").mockImplementation(() => undefined);
+
+      await setup();
+      vi.advanceTimersByTime(20 * 60 * 1000);
+
+      expect(exit).toHaveBeenCalledWith(1);
+    });
+  });
+});
--- a/langwatch/src/__tests__/test-unit-global-setup.unit.test.ts
+++ b/langwatch/src/__tests__/test-unit-global-setup.unit.test.ts
@@ -1,0 +1,43 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { setup } from "../test-unit-global-setup";
+
+describe("Feature: Test shard hard-floor failures", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.stubEnv("CI", "true");
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    vi.unstubAllEnvs();
+    vi.restoreAllMocks();
+  });
+
+  describe("when the hard-floor timeout expires", () => {
+    it("keeps the test process running before four minutes", async () => {
+      const exit = vi
+        .spyOn(process, "exit")
+        .mockImplementation((() => undefined) as typeof process.exit);
+      vi.spyOn(console, "error").mockImplementation(() => undefined);
+
+      await setup();
+      vi.advanceTimersByTime(4 * 60 * 1000 - 1);
+
+      expect(exit).not.toHaveBeenCalled();
+    });
+
+    /** @scenario A unit test shard hard floor exits with failure after four minutes */
+    it("exits the test process with a failure status at four minutes", async () => {
+      const exit = vi
+        .spyOn(process, "exit")
+        .mockImplementation((() => undefined) as typeof process.exit);
+      vi.spyOn(console, "error").mockImplementation(() => undefined);
+
+      await setup();
+      vi.advanceTimersByTime(4 * 60 * 1000);
+
+      expect(exit).toHaveBeenCalledWith(1);
+    });
+  });
+});
--- a/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/langwatch/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -267,7 +267,7 @@ describe("aggregation-builder", () => {
         aggregation: "p90",
         alias: "trace_time_to_first_token_ms",
         rawColumn: "TimeToFirstTokenMs",
-        outerAggregation: /\bquantileExact\(0\.9\)\s*\(\s*trace_time_to_first_token_ms\s*\)/,
+        outerAggregation: /\bquantileTDigest\(0\.9\)\s*\(\s*trace_time_to_first_token_ms\s*\)/,
       },
     ])(
       "routes $metric through the CTE alias in arrayJoin groupBy path",
--- a/langwatch/src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts
@@ -4,6 +4,10 @@ import { TeamUserRole } from "@prisma/client";
 import { teamRouter } from "../team";
 import { createInnerTRPCContext } from "../../trpc";
 
+vi.mock("../../../auditLog", () => ({
+  auditLog: vi.fn(() => Promise.resolve()),
+}));
+
 // team.update writes membership to RoleBinding. A user can hold MORE THAN ONE
 // TEAM binding on a team — a built-in role plus additive custom-role grants —
 // and RBAC unions them. The settings form shows/edits only the displayed
--- a/langwatch/src/server/event-sourcing/__tests__/integration/globalSetup.ts
+++ b/langwatch/src/server/event-sourcing/__tests__/integration/globalSetup.ts
@@ -12,6 +12,7 @@ import {
   type StartedRedisContainer,
 } from "@testcontainers/redis";
 import { migrateUp } from "~/server/clickhouse/goose";
+import { scheduleTestShardHardFloor } from "~/test-shard-hard-floor";
 
 const TEST_DATABASE = "test_langwatch";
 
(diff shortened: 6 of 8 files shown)


0xdeafcafe:
Looks good, thanks for the contribution! I'll give this a proper review again in the morning before final approve and merge. Also...

> keep the existing `4-minute` unit and `20-minute` integration ceilings;

4:20, i never noticed we did that. beautiful. we are an amsertdam based company after all.

langwatch-agent (commented):
Static fork review only; I did not execute code or install dependencies from this PR. The CI-only implementation correctly changes the hard-floor behavior to fail closed, and I found no additional code-level security or integrity issue in the diff. The current head is still blocked by required `test-unit (2)` and `langwatch-app-complete` failures, so it should not merge until the failing shard is investigated and all required checks are green.

langwatch-agent:
Scariest risk: makes hard-floor test shards fail closed. If the fail-closed logic is wrong it either blocks CI on a flake or, worse, still passes when the floor is breached.
- **Must Check** a breached floor actually fails the job, and a shard that errors is not silently treated as pass.
- **Ask Author** what happens when a shard produces no results at all: pass or fail?
External contributor. Attention map only.

langwatch-agent (commented):
Static fork re-review on the current head: the hard-floor helper still fails closed with process.exit(1), and the added timeout tests exercise the failure status. The one new commit also updates two unrelated test expectations/mocks, but I found no application-security or CI-integrity regression in those static changes. I did not execute code or install dependencies from this fork.

langwatch/src/test-shard-hard-floor.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function scheduleTestShardHardFloor({
+  timeoutMs,
+  message,
+}: TestShardHardFloorOptions): void {
+  const timer = setTimeout(() => {
+    // eslint-disable-next-line no-console
+    console.error(message);
+    process.exit(1);
```
drewdrewthis, line 13: **Blocking - the fail-closed behavior is unproven at HEAD.** This `process.exit(1)` is the crux of the PR, but `langwatch-app-ci` has never executed at HEAD `8229210b4`: run [29317367069](https://github.com/langwatch/langwatch/actions/runs/29317367069) is `action_required` with **0 jobs** (fork `pull_request` workflows need maintainer approval to run), and there is no `langwatch-app-complete` check on the head SHA at all. The green checks shown on the PR are entirely `pull_request_target` sub-project CIs (`sdk-*`, `mcp-*`, `langevals`) whose path filters skipped every job — none ran the app unit/integration shards.

Because this PR changes CI pass/fail *semantics* (exit 0 -> exit 1) **and** edits two tests to fix the very failures the fail-closed signal is meant to expose, this is exactly the change that must be observed green in a real app-ci run. `verify-ci-shard-tally.sh` reports "green is real" here only **vacuously**: it keyed on the `action_required` run, which produced zero shards, so "every shard printed a tally" is true over an empty set.

**To clear:** have a maintainer approve + run `langwatch-app-ci` on `8229210b4`; it must finish green with real `Test Files N passed` tallies on every `test-unit`/`test-integration` shard, confirming both that the two fixed tests now pass and that no other shard newly reds under fail-closed, before merge.

drewdrewthis (changes requested):
Requesting changes on **one blocking item**: `langwatch-app-ci` has never executed at HEAD, so the fail-closed change this PR introduces is unverified in CI. The code itself reviews as correct.

**Verified correct against production (no action needed):**
- **Fail-closed helper** (`test-shard-hard-floor.ts`): `process.exit(1)` + `timer.unref()` is the right direction — healthy shards still exit on their own; the timer only fires on a wedge. Direct coverage in both new global-setup unit tests (boundary at `-1` and exactly at the ceiling).
- **`quantileExact` -> `quantileTDigest`** (aggregation-builder.test.ts:267): matches the real production contract, not a loosened one. `metric-translator.ts:449-451` routes `performance.first_token` through `perfAgg`, which hard-codes `percentileMode: "tdigest"` (line 438); tdigest is the documented default for `performance.*` latency metrics (line 96), while eval metrics deliberately stay `exact`. The old assertion was stale; production analytics code is untouched by this PR.
- **`auditLog` mock** (team.update.multi-binding): sound isolation. The test injects a mock prisma via `ctx.prisma`, but `auditLog.ts` imports the prisma **singleton** directly (`import { prisma } from "./db"`), bypassing `ctx` and hitting the DB-less unit shard — the open handle that wedges the shard. The mock closes exactly that escape hatch and preserves every RoleBinding-sync assertion.

**Blocking:** see the inline thread on `test-shard-hard-floor.ts` — require a green `langwatch-app-ci` run at HEAD before merge.

**Non-blocking (Decide):** fail-closed converts hidden-failure false-negatives into possible false-positive timeouts on slow runners; the 4-minute unit floor (healthy ~3 min) is tight. You already defer ceiling changes until a runner-specific retry — worth watching for flakiness once app-ci runs green.

_Verdict is prose, not a GitHub approval — this review requests changes on the CI-evidence gap only._

drewdrewthis:
## Review verdict: NOT-READY

**Reviewed at:** `8229210b4` · **Run:** pr-review (reviewer: @drewdrewthis)

This item also exists as a **resolvable inline thread** (linked). Resolve the thread to clear it.

1. **[reviewer] CI never executed at HEAD — the green is hollow.** `langwatch-app-ci` run [29317367069](https://github.com/langwatch/langwatch/actions/runs/29317367069) @ `8229210b4` is `action_required` with **0 jobs**, and there is no `langwatch-app-complete` check on the head SHA. The passing checks on the PR are entirely `pull_request_target` sub-project CIs (`sdk-*` / `mcp-*` / `langevals`) that path-skipped every job — the app `test-unit` / `test-integration` shards never ran. Since this PR changes CI pass/fail *semantics* (exit 0 -> exit 1) **and** edits two tests to fix the very failures the fail-closed floor is meant to expose, it must be observed green in a real app-ci run at HEAD (with `Test Files N passed` tallies on every shard, and no other shard newly red under fail-closed) before merge. `verify-ci-shard-tally.sh` reports "green is real" only **vacuously** — it keyed on the `action_required` run, which produced zero shards. -> thread: https://github.com/langwatch/langwatch/pull/5767#discussion_r3594232534

### Non-blocking (Decide)
- **[reviewer]** Fail-closed converts hidden-failure false-negatives into possible false-positive timeouts on slow runners; the 4-minute unit floor (healthy ~3 min) is tight. Worth watching for flakiness once app-ci runs green. (Decide)

### Verified correct against production (checked per review seeds; no action)
- **`quantileExact` -> `quantileTDigest`** (`aggregation-builder.test.ts:267`) matches the production contract, not a loosened one: `metric-translator.ts:449-451` routes `performance.first_token` through `perfAgg`, which hard-codes `percentileMode: "tdigest"` (line 438); tdigest is the documented default for `performance.*` latency metrics (line 96), while eval metrics deliberately stay `exact`. The old assertion was stale; production analytics code is untouched by this PR.
- **`auditLog` mock** (`team.update.multi-binding`) is sound isolation — the test injects a mock prisma via `ctx.prisma`, but `auditLog.ts` imports the prisma singleton directly (`import { prisma } from "./db"`), bypassing `ctx` and hitting the DB-less unit shard (the open handle that wedges the shard). The mock closes that escape hatch and preserves every RoleBinding-sync assertion.

---

jstar0:
`main` now includes #6000 / `24c0a2cf`, which overlaps this PR by preserving observed shard failures through a dedicated reporter. Rebasing #5767 would conflict with and replace that newly merged strategy, so I am closing this PR as superseded instead of forcing an unconditional hard-floor policy on top.

#4450 remains open for the narrower residual: a wedged shard with no recorded failure and no final tally can still exit 0. That behavior can be handled separately if maintainers want every unfinished shard to fail.

PR #5768, opened 2026-07-13
fix(topic-clustering): bound the heavy trace fetch to a 49-day hot-tier window

sergioestebance:
## Problem

Topic clustering's page fetch (`fetchTracesFromClickHouse`) reads the heavy `ComputedInput` payload and cursor-pages back through a **365-day** window of `trace_summaries`. Only the most-recent **~90 days** live on the ClickHouse hot tier; everything older is S3-tiered. So paging walked the payload column across **cold storage** on every batch run.

Measured on prod (`system.parts`, last-12-months partitions):

| tier | parts | size |
|---|---|---|
| local (hot EBS) | 226 | 12.41 GiB (~last 13 wks) |
| object (S3 cold) | 401 | **38.45 GiB** (3–12 mo old) |

~75% of the scanned bytes were cold — the source of the multi-hundred-MB cold reads that also amplify worker event-loop stalls.

## Why not just shrink the whole window to 49 days

Because the **mode-decision** count query uses the same window, and shrinking it breaks mode selection. `isIncrementalProcessing = topicIds.length > 0 && assignedTracesCount >= 1200`, where `assignedTracesCount` is measured over the window. A 49-day window measures only recent traffic, so historically-mature-but-quiet projects fall under 1200 and flip **out of** cheap incremental into the **heavy full-batch re-cluster** path.

Measured on prod: of **73** currently-incremental projects, a blanket 49-day change would flip **54** to batch — the opposite of reducing load.

## Fix: split the two windows by role

- **MODE** (`fetchCountsFromClickHouse`): **stays 365 days.** Reads only light key columns (no payload), so its scan stays cheap even across cold-tier partitions, and it correctly reflects the project's whole history → no mode flips.
- **FETCH** (`fetchTracesFromClickHouse`): **49 days**, inside the hot tier, so the `ComputedInput` read never touches S3.

Both are named constants (`CLUSTERING_MODE_WINDOW_DAYS`, `CLUSTERING_FETCH_WINDOW_DAYS`) with the rationale documented at the definition.

## Trade-off

An unassigned trace older than 49 days is no longer retroactively clustered. For the 2-day batch cadence and hot-tier ingest this is negligible; new/recent traffic (all within the window) is unaffected. 49 days matches the retention recovery floor and sits comfortably within the ~90-day hot tier.

## Tests

Integration case in `fetchTracesFromClickHouse.integration.test.ts`: seeds an in-window (10-day) and an out-of-window (60-day) trace, asserts the fetch returns the former and excludes the latter.

## Remaining cold read (explicit)

The MODE count query still scans up to 365 days, including S3 cold-tier partitions — deliberately, since project maturity is a whole-history question (see the flip measurement above). It reads only light key columns (`OccurredAt`, `TopicId`, `UpdatedAt`; no `ComputedInput`), so it is not in the stall/OOM class this PR eliminates. Killing even that residual scan means answering "is this project mature?" without querying trace data at all (e.g. from the `topics` table or a maintained counter) — follow-up, out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved topic clustering trace retrieval by consistently limiting results to the supported 49-day fetch window.
  * Ensured older traces are excluded from clustering results while recent traces remain available.
  * Standardized the clustering decision look-back period to one year.

* **Tests**
  * Added integration coverage confirming that in-window traces are returned and older traces are excluded.

### DIFF
--- a/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
+++ b/langwatch/src/server/topicClustering/__tests__/fetchTracesFromClickHouse.integration.test.ts
@@ -181,4 +181,45 @@ describe("fetchTracesFromClickHouse integration", () => {
       expect(page2.traces[0]?.trace_id).toBe(`${EMPTY_TENANT}-trace-002000`);
     });
   });
+
+  describe("when a trace is older than the fetch window", () => {
+    // The heavy ComputedInput fetch is bounded to a 49-day hot-tier window so
+    // cursor-paging never reads the payload column back into S3 cold storage.
+    // A trace whose OccurredAt predates the window must not be returned.
+    const WINDOW_TENANT = "topic-fetch-window-test";
+    const DAY_MS = 24 * 60 * 60 * 1000;
+
+    beforeAll(async () => {
+      const now = Date.now();
+      const input = JSON.stringify("hello world");
+      const row = (id: string, occurredAt: number): TraceRow => ({
+        ...traceRow(WINDOW_TENANT, 0, input),
+        TraceId: `${WINDOW_TENANT}-${id}`,
+        OccurredAt: new Date(occurredAt),
+        UpdatedAt: new Date(occurredAt),
+        TopicId: null,
+      });
+      await insertRows(ch, [
+        row("recent", now - 10 * DAY_MS), // inside the 49-day window
+        row("stale", now - 60 * DAY_MS), // older than the window
+      ]);
+    }, 60_000);
+
+    afterAll(async () => {
+      await cleanupTestData(WINDOW_TENANT);
+    });
+
+    it("returns the in-window trace and excludes the older one", async () => {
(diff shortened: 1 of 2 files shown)


langwatch-agent:
Scariest risk: bounds the topic-clustering trace fetch to a 49-day hot window, trading completeness for cost so clustering silently ignores older data.
- **Must Check** 49 days is intended and config-driven, not a magic constant.
- **Ask Author** is the truncation surfaced, or does clustering look complete while dropping older traces?
Attention map only.

langwatch-agent (commented):
No blocking findings. The 365-day lightweight maturity count remains separate from the 49-day payload fetch, and the new boundary is applied consistently to the page, payload, and latest-version subqueries. Validation: topic-clustering unit coverage (29 passed, 2 skipped) and targeted ClickHouse/Redis integration coverage (5 passed).

PR #5784, opened 2026-07-14
feat(governance): copilot app capture via direct OTLP at launch (ADR-039 §Extension, stacked on #5604)

sergioestebance:
Stacked on langwatch/langwatch-saas#888. Tracks issue langwatch/langwatch-saas#889. Locks and implements the **GitHub Copilot app** surface from **ADR-039 §Extension** (Accepted).

## What ships
The standalone GitHub Copilot app is captured by **direct OTLP export at launch** — the same transport already shipped for the Copilot CLI (Path B). A `langwatch copilot-app connect` command mints a `copilot_app` ingest key and installs a per-OS login agent that owns the app's launch and injects the OTLP env; the app then pushes `gen_ai.*` records straight to `/api/otel`. No file reader, no SQLite, no pairing.

Empirically validated on the shipped app (build 1.0.71): injected env reaches the app's spawned runtime, which posts authenticated `gen_ai.*` OTLP carrying tokens, `github.copilot.cost`/`nano_aiu`, and full message content. See issue langwatch/langwatch-saas#889 for the full investigation log.

## Changes
- **ADR-039 §Extension** — locked to direct-OTLP-at-launch (supersedes the earlier scrape-and-pair and file-exporter drafts).
- **Feature specs** — `copilot-app-connect`, `copilot-app-launch-agent`, `copilot-app-otlp` (29 scenarios).
- **Extractor** — `copilot.ts` lifts `github.copilot.nano_aiu` as metadata (benefits the CLI too).
- **CLI** — `langwatch copilot-app connect [--tokens-only]`: app detection, `copilot_app` key mint, per-OS login-agent install (launchd / systemd --user / Task Scheduler), content-on default with a loud tokens-only opt-out.
- **Logout** — `langwatch logout` removes the capture agent.

## Tests
- 44 unit tests (extractor + env block + app detection + agent descriptors + installer + connect orchestrator), all green.
- Feature-parity check green (bound scenarios resolve; live-env / server-rotation scenarios flagged `@unimplemented`).

## Deferred (Roadmap, not in this PR)
- VS Code extension surface (`copilot_vscode`).
- Enterprise fleet + per-user identity link.
- `context_*_tokens` content-source cost breakdown (DB-only; enrich later).
- Server-side `copilot_app` key revoke on logout, key-rotation-on-reconnect assertions, live app-turn e2e.

## Deployment Impact
No deployment impact. This is a client-side CLI feature (`langwatch copilot-app connect`) plus a receiver-side extractor attribute; it touches ADRs/specs which trip the gate, but ships no operator surface:
- **Env vars:** none added to the server/helm. The login agent sets `COPILOT_OTEL_*` / `OTEL_EXPORTER_OTLP_*` on the **user's local Copilot app process** only.
- **Helm values:** none added or changed. Vanilla `helm install` with no overrides is unaffected.
- **BYOC dataplane:** no impact — traffic flows to the existing `/api/otel` ingest path already used by the CLI.

## Review hardening (2026-08-05)
An adversarial review round (macOS behavior verified by execution) found and fixed in `2940788c8`: launchd registration failures were invisible (legacy `launchctl load` exits 0 on failure — now `bootstrap`/`bootout` with real exit codes), the "capture active now" claim was false on Windows and on already-running apps (now `schtasks /Run`, `systemctl --user restart`, and honest copy), the Linux unit lacked a graphical-session dependency, the Windows task inherited the block-on-battery default, partial removals could strand a token-bearing wrapper invisible to logout, and descriptor rewrites didn't tighten pre-existing loose permissions. Full ledger on the tracking issue.


sdks/typescript/src/cli/utils/governance/copilot-app-agent.ts
```diff
@@ -0,0 +1,196 @@
...
+  const descriptor = renderLaunchAgent(spec);
+  for (const file of descriptor.files) {
+    io.mkdirp(path.dirname(file.path));
+    io.writeFile(file.path, file.content, file.mode);
+  }
+  for (const { cmd, args } of registerCommands(
+    spec.platform,
+    descriptor.registerPath,
+  )) {
+    try {
```
langwatch-agent, line 229: **P1: Registration failures are reported as a successful connection.**

This catches every service-manager error, including `launchctl load`, `systemctl --user enable --now`, and `schtasks /Create`. `connectCopilotApp()` has no confirm step despite the comment and immediately prints the success message. A user can therefore receive a minted key and "connected" while no agent was registered. Worse, minting is a server-side hard-cut rotation, so a reconnect can revoke the previously working key before this failure and silently turn capture off.

Only the expected first-install `launchctl unload` failure should be tolerated. Propagate all load/enable/create failures (and roll back or clearly report the local state), then add injected-runner tests for each failing registration command.

typescript-sdk/src/cli/utils/governance/copilot-app-agent.ts
```diff
@@ -0,0 +1,196 @@
...
+ */
+export function removeCopilotAppAgent(
+  platform: AppPlatform,
+  home: string,
+  io: AgentIo = defaultIo,
+): boolean {
+  const registerPath = copilotAppAgentPath(platform, home);
+  const present = io.fileExists(registerPath);
+
+  if (platform === "darwin") {
```
langwatch-agent: **P1: `logout` can leave the capture process running after claiming it was removed.**

Unregister failures are swallowed and the descriptor/wrapper are deleted anyway. For example, if `launchctl unload` or `systemctl --user disable --now` fails transiently, the already loaded service still has the bearer token in its process environment and continues exporting content. The deleted descriptor removes the normal retry path, `removeCopilotAppAgent()` returns `true`, and `logout` prints it as removed. The PR explicitly defers ingestion-key revocation, so revoking the device session does not stop that live exporter.

Do not delete the descriptor or report success until the OS unregister succeeds. Preserve enough state to retry, make the logout failure visible, and add failure-path tests. The durable fix also needs to revoke the `copilot_app` ingest key as part of teardown.

typescript-sdk/src/cli/commands/copilot-app.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  if (!deps.captureContent) {
+    deps.warn(
+      "[langwatch] content capture is off for the Copilot app; traces will carry tokens only.",
+    );
+  }
+  const project =
+    cfg.organization?.slug ?? cfg.organization?.name ?? "your personal project";
+  deps.info(
+    `GitHub Copilot app connected. Usage will be tracked into ${project}. If the app is already open, quit and reopen it to start capture.`,
```
langwatch-agent: **P1: The advertised "quit and reopen" recovery does not launch the GUI with the capture environment.**

The OS descriptors only launch the app once at connect/login (`RunAtLoad`, a systemd unit with no restart policy, or a logon-triggered task). They cannot modify an already-running Copilot process. If the app is open during connect, the agent launch does not retrofit its environment; after the user follows this message and reopens from the Dock/Start menu, that normal GUI launch again does not inherit the descriptor's environment. The same loss occurs after a normal app quit in the session.

That means the core feature can report success but capture nothing during normal app use, contradicting the PR's automatic/no-user-action claim. Please prove the actual lifecycle on each supported host, including connect-while-open and quit/reopen, and redesign the launch mechanism or UX so every supported launch path has the OTLP variables before calling this connected.

typescript-sdk/src/cli/utils/governance/copilot-app.ts
```diff
@@ -0,0 +1,279 @@
...
+
+      // Task Scheduler cannot set env in XML; the task runs this wrapper,
+      // which sets each var then launches the app. `%` is doubled so cmd
+      // does not treat it as a variable reference.
+      const setLines = entries
+        .map(([k, v]) => `set "${k}=${v.replace(/%/g, "%%")}"`)
+        .join("\r\n");
+      const wrapper = `@echo off\r\n${setLines}\r\nstart "" "${spec.execPath}"\r\n`;
+
+      const xml = `<?xml version="1.0" encoding="UTF-16"?>
```
langwatch-agent: **P1: The Windows task XML declares UTF-16 but is written as UTF-8.**

`defaultIo.writeFile` calls Node's `fs.writeFileSync(file, content, { mode })`, whose string encoding defaults to UTF-8. This XML declaration tells `schtasks /Create /XML` to decode the same bytes as UTF-16, so the task definition is malformed rather than registered. The current unit tests only inspect the JavaScript string, and the installer then swallows the resulting `schtasks` failure, so Windows users are told capture is connected while nothing was installed.

Either declare `encoding="UTF-8"` (the Task Scheduler schema examples use UTF-8) or deliberately write UTF-16LE with a BOM. Add a byte-level test of the default writer plus a real Windows registration check in CI or platform validation.

langwatch-agent (dismissed):
Blocking review: the direct-OTLP transport and source separation look consistent, but the login-agent lifecycle is not safe to ship. The inline findings cover silent install failure, incomplete logout, ordinary GUI relaunch liveness, and a broken Windows task definition. CI is green; the focused local SDK command could not start because this shared checkout has no root Vitest executable, so I did not modify dependencies.

sdks/typescript/src/cli/utils/governance/copilot-app-agent.ts
sergioestebance: Fixed in 502a7dd24. `installCopilotAppAgent` now throws `CopilotAppAgentError` on any non-tolerated service-manager failure — only the expected first-install `launchctl unload` is tolerated. `connectCopilotApp` catches it and reports `agent-install-failed` (loud) instead of a mint + "connected". Failure-path tests added for `launchctl load` and `schtasks /Create`.

typescript-sdk/src/cli/utils/governance/copilot-app-agent.ts
sergioestebance: Fixed in 502a7dd24. `removeCopilotAppAgent` now throws and keeps the descriptor when an unregister command fails, so the retry path survives and `logout` prints "couldn't remove" (via its existing try/catch) rather than a clean removal — a live exporter is never silently orphaned. Test added: unregister failure preserves the descriptor.

typescript-sdk/src/cli/commands/copilot-app.ts
sergioestebance: Fixed in 502a7dd24. The success copy no longer claims a Dock reopen captures. It now says capture is active for the app launched now and on every login, and that a manual relaunch before the next login is not tracked — the Dock-launch limitation, already an ADR-039 open question. A failed agent install now reports `agent-install-failed` instead of success.

typescript-sdk/src/cli/utils/governance/copilot-app.ts
sergioestebance: Fixed in 502a7dd24. The Windows task XML now declares `encoding="UTF-8"` to match the UTF-8 bytes `defaultIo.writeFile` emits (Task Scheduler accepts UTF-8). Added a declaration test asserting `UTF-8` and no `UTF-16`. Combined with the register-failure propagation above, a rejected `schtasks` now surfaces instead of a false "connected".

Aryansharma28 (approved):
Re-reviewed as cover: `langwatch-agent`'s CHANGES_REQUESTED (2026-07-15) predates commit 711d9aab0 (2026-07-16, "copilot-app agent — propagate register/unregister failures, honest connect copy, UTF-8 task XML"), which addresses all 4 P1 findings directly:

- Register failure now throws `agent-install-failed` instead of reporting a false "connected" — covered by a new unit test.
- Unregister failure now throws `CopilotAppAgentError` and preserves the descriptor for retry instead of silently orphaning a live agent — covered by a new unit test.
- The "quit and reopen to start capture" claim rewritten to be honest about what's actually tracked (launch-time/login only, not a manual relaunch) — matches the ADR's own stated open question.
- Windows task XML encoding declaration fixed to UTF-8 to match the actual bytes `writeFile` emits — covered by a new unit test asserting the declaration.

All 4 review threads are marked resolved. CI is green on the current head. Since the review bot that opened this can't come back to re-verify (Drew's AI fleet is off for the duration), approving as a human so the stack (including #5831 on top) can move.

langwatch-agent:

**Mode:** Targeted Review · already approved by a human and CLEAN, so this is now a merge decision
**Linked issue:** tracks langwatch/langwatch-saas#889, implements ADR-039 §Extension (Accepted). Stacked on langwatch/langwatch-saas#888, which should land first.

**Top inspection targets**
1. `Must Check` - `langwatch copilot-app connect` mints an ingest key and installs a **per-OS login agent** on a developer machine. Confirm the key is stored with appropriate file permissions and that uninstall genuinely removes the login agent.
2. `Must Check` - the OTLP env injection at app launch. Confirm it scopes to the Copilot app only and does not export telemetry env into unrelated child processes.
3. `Probably Fine` - the canonicalisation extractor and its unit tests.

**Production risk:** a long-lived ingest credential written to a developer machine by a CLI command. Blast radius is one project's ingest, but it persists.

1. What file mode protects the minted `copilot_app` key, and is there a documented revoke path?
2. Does langwatch/langwatch-saas#888 need to merge first?

PR #5528, opened 2026-07-08
feat(evaluators): one "Comparison" judge for two or more candidates (#5100, #5101)

Aryansharma28:
Closes #5100. Closes #5101. Part of #5099.

`langevals/pairwise_compare` (shipped, 2 slots) and `langevals/select_best_compare` (unreleased, N-way) were the same feature counted twice. Comparing two candidates is not a different thing from comparing five. This collapses them into one **Comparison** flow: one card, one config form, one result column, one judge.

## What a user sees

- **One Comparison card.** No "Pairwise Compare", no "N-way Compare".
- **It behaves like every other evaluator.** Choosing Comparison lists the ones you already have, with a New Comparison button — the same `EvaluatorListDrawer`, narrowed to comparison evaluators. A comparison *is* a saved evaluator.
- **The column is editable.** Click its header to reopen the config. Comparisons are filtered out of the evaluator chip lists (they grade no single target), so the header is the only surface that can carry this.
- **Variants lay out as a grid**, 3–4 to a row, evenly (12 → 4/4/4, 9 → 3/3/3, 7 → 4/3). A lone trailing card only when three and four both leave one over.
- **Structured outputs work.** A variant emitting `{answer, confidence}` gets a field picker; the judge reads `.answer`. Variants choose independently — two variants may name the same answer differently.
- **The judge already sees your evaluator scores.** Any grading evaluator on a variant's column has its score appended to that candidate. This shipped untested and unmentioned; now it is pinned and stated in the Variants `(i)`.
- **Save is disabled below two variants** instead of an orange warning you could scroll past. The orchestrator skips such a column outright, so it used to run nothing and say nothing.

## What is kept working

Read old, write new. Experiments saved with the two-slot `pairwise` config still **load, render, and re-run** — `toComparisonConfig` folds them into `variants` on load, and stored verdict labels of `"A"` / `"B"` still name a winner. `pairwise_compare.py` stays runnable for existing monitors, and is no longer offered when creating something new.

## Notable

`randomize_order` (one judge call, order shuffled deterministically per row) replaces pairwise's `swap_and_confirm` (two calls, A/B swapped). Same position-bias defence, half the tokens, and re-running a row gives a comparable verdict.

The verdict names the winning candidate's own identifier — the prompt handle, not `target_NNNN`.

## Bugs found while merging

[...]

PR #5796, opened 2026-07-14
feat(experiments): reroute legacy pairwise_compare to select_best_compare

Aryansharma28:
## Summary

Retires the legacy `langevals/pairwise_compare` judge from the app's call path **without** breaking experiments and monitors that were saved against it. A dispatch whose stored evaluator type is the 2-slot pairwise judge is transparently rerouted to the N-way `select_best_compare`, and its 2-slot request body is translated into the `candidates` shape the new judge expects.

This is the backward-compat adapter that lets `pairwise_compare` be fully decommissioned later (its AWS API Gateway route, its langevals evaluator) without a customer-facing regression.

## Why one interception point

Both halves of the redirect happen at a **single site** — the legacy evaluations route (`evaluations-legacy.ts`):

- `resolveDispatchEvaluatorType` — reroutes the stored type `pairwise_compare → select_best_compare`
- `translateLegacyPairwisePayload` — reshapes `candidate_a_* / candidate_b_*` → `candidates: [...]`

They travel together, in the same breath. **#5528 regressed precisely because the dispatched type and the payload shape were decided in different places** (the orchestrator forced the new type while the payload stayed 2-slot, so a 2-slot body reached the N-way judge). Here, every upstream caller — the Experiments Workbench orchestrator, a monitor's scheduled run — keeps emitting the exact 2-slot shape it always has and is unaware of the reroute. The type and payload cannot drift apart because nothing upstream decides either.

## Also handled

A legacy pairwise `prompt` setting references the old `{candidate_a_output}` / `{candidate_b_output}` placeholders. `select_best_compare` only renders `{candidates}` (+ `{input}` / `{golden}`), so an un-migrated prompt — including pairwise's own default — would arrive half-literal. `stripIncompatiblePairwisePrompt` drops any prompt with no `{candidates}` placeholder so the new judge's own default wins; a prompt already migrated to `{candidates}` is kept.

## Spec

`specs/experiments/comparison.feature`:
- Updates **"Existing pairwise monitors keep running"** — now requires rerouting to `select_best_compare` with the config translated to the N-way shape and a winner *identifier* (not a slot letter). Previously this scenario asserted the opposite (kept running against the pairwise judge); this PR is the deliberate change of that settled behavior, so the app only ever talks to one judge.
- Adds **"structured-output field selection survives translation"**.

## Tests

- `normalizeComparison.test.ts` — `resolveDispatchEvaluatorType` reroutes legacy, passes everything else (incl. `undefined`) through.
- `evaluations-legacy-pairwise-adapter.unit.test.ts` (new) — `translateLegacyPairwisePayload` folds both slots into ordered `candidates`, preserves `input`/`golden`, drops the flat keys, keeps a narrowed structured-output value, and drops an empty slot rather than emitting a blank candidate; `stripIncompatiblePairwisePrompt` drops old-placeholder prompts and keeps migrated ones.
- `pnpm typecheck` + `typecheck:tests` clean; biome clean; 85 tests pass across orchestrator + legacy-route + normalize + new adapter suites.

## Verification status

Unit + type + lint verified. Live end-to-end reroute (seed a `pairwise_compare` DB row, re-run, confirm a `select_best_compare` verdict with a winner id) not yet run against a dev server on this branch — happy to do that pass if wanted before merge.

## Related

- `langwatch-saas#708` (merged) — adds the `select_best_compare` API Gateway route.
- Sibling to `langwatch#5789` (Comparison error-handling / bugbash fixes).

## Summary by CodeRabbit

- **Bug Fixes**
  - Legacy `pairwise_compare` evaluator dispatches are now transparently rerouted to the current comparison evaluator, translating legacy `candidate_a_*`/`candidate_b_*` payloads into the modern `candidates` format.
  - Legacy pairwise payloads and structured outputs are preserved through migration, including correct handling of incomplete candidate slots.
  - Legacy prompt settings are adjusted to remove incompatible prompt content unless it uses the supported `{candidates}` placeholder.
- **Tests**
  - Expanded unit and adapter test coverage for legacy translation, prompt dropping, and dispatch rerouting.
- **Documentation**
  - Updated the comparison feature spec to reflect legacy-to-current rerouting behavior and acceptance scenarios.

### DIFF
--- a/langwatch/src/experiments-v3/utils/__tests__/normalizeComparison.test.ts
+++ b/langwatch/src/experiments-v3/utils/__tests__/normalizeComparison.test.ts
@@ -1,8 +1,14 @@
 import { describe, expect, it } from "vitest";
-import type { EvaluatorConfig, TargetConfig } from "../../types";
+import {
+  COMPARISON_EVALUATOR_TYPE,
+  type EvaluatorConfig,
+  LEGACY_PAIRWISE_EVALUATOR_TYPE,
+  type TargetConfig,
+} from "../../types";
 import {
   normalizeEvaluators,
   normalizeTargets,
+  resolveDispatchEvaluatorType,
   resolveVerdictLabel,
   toComparisonConfig,
 } from "../normalizeComparison";
@@ -111,7 +117,10 @@ describe("toComparisonConfig", () => {
           randomizeOrder: true,
         };
 
-        const config = toComparisonConfig({ comparison, pairwise: legacyPairwise });
+        const config = toComparisonConfig({
+          comparison,
+          pairwise: legacyPairwise,
+        });
 
         expect(config?.variants).toEqual(["x", "y", "z"]);
       });
@@ -121,6 +130,38 @@ describe("toComparisonConfig", () => {
   describe("given a carrier that is not a comparison", () => {
     it("returns undefined", () => {
       expect(toComparisonConfig({})).toBeUndefined();
+    });
+  });
+});
+
+describe("resolveDispatchEvaluatorType", () => {
+  describe("given the legacy pairwise judge type", () => {
+    it("reroutes it to the current comparison judge", () => {
+      expect(resolveDispatchEvaluatorType(LEGACY_PAIRWISE_EVALUATOR_TYPE)).toBe(
+        COMPARISON_EVALUATOR_TYPE,
+      );
+    });
+  });
+
+  describe("given the current comparison judge type", () => {
+    it("passes it through untouched", () => {
+      expect(resolveDispatchEvaluatorType(COMPARISON_EVALUATOR_TYPE)).toBe(
+        COMPARISON_EVALUATOR_TYPE,
+      );
+    });
+  });
+
+  describe("given any other evaluator type", () => {
+    it("passes it through untouched", () => {
+      expect(resolveDispatchEvaluatorType("langevals/llm_boolean")).toBe(
+        "langevals/llm_boolean",
+      );
+    });
+  });
+
+  describe("given undefined", () => {
+    it("returns undefined", () => {
+      expect(resolveDispatchEvaluatorType(undefined)).toBeUndefined();
     });
   });
 });
@@ -156,7 +197,12 @@ describe("normalizeEvaluators", () => {
   describe("given a plain per-row evaluator", () => {
     it("leaves it alone", () => {
       const evaluators = [
-        { id: "eval-1", evaluatorType: "custom/exact_match", inputs: [], mappings: {} },
+        {
+          id: "eval-1",
+          evaluatorType: "custom/exact_match",
+          inputs: [],
+          mappings: {},
+        },
       ] as unknown as EvaluatorConfig[];
 
       const [normalized] = normalizeEvaluators(evaluators);
--- a/langwatch/src/experiments-v3/utils/normalizeComparison.ts
+++ b/langwatch/src/experiments-v3/utils/normalizeComparison.ts
@@ -1,9 +1,33 @@
-import type {
-  ComparisonEvaluatorConfig,
-  EvaluatorConfig,
-  PairwiseEvaluatorConfig,
-  TargetConfig,
+import {
+  COMPARISON_EVALUATOR_TYPE,
+  type ComparisonEvaluatorConfig,
+  type EvaluatorConfig,
+  LEGACY_PAIRWISE_EVALUATOR_TYPE,
+  type PairwiseEvaluatorConfig,
+  type TargetConfig,
 } from "../types";
+
+/**
+ * Reroutes a stored evaluator type to the judge that will actually run: a row
+ * whose persisted type is the legacy two-slot `pairwise_compare` judge is sent
+ * to the current N-way `select_best_compare` one instead — the legacy endpoint
+ * is never called again. Every other type passes through unchanged.
+ *
+ * Called at exactly one site — the legacy evaluations route
+ * (`evaluations-legacy.ts`), together with `translateLegacyPairwisePayload`,
+ * which reshapes the request body in the same breath. Keeping the type reroute
+ * and the payload translation co-located is the whole point: #5528 happened
+ * because the dispatched type was changed in one place while the payload shape
+ * was decided in another, so a 2-slot body could reach the N-way judge. Every
+ * upstream caller (the orchestrator, a monitor's scheduled run) keeps emitting
+ * the 2-slot shape it always has and is unaware of the reroute.
+ */
+export const resolveDispatchEvaluatorType = (
+  storedEvaluatorType: string | undefined,
+): string | undefined =>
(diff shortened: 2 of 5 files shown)


sergioestebance:
## Ruthless review — verdict: SHIP-ABLE

No P0/P1. Traced the full orchestrator → nlp → route path; the reroute + payload translation land where the PR claims, and the single-interception-point argument holds up.

### Verified (traced, not pattern-matched)
- **Reroute site is the real chokepoint.** Orchestrator column-targets dispatch via `evaluators/{id}` (`workflowBuilder.ts:339`); the route resolves `checkType` from the DB row's `config.evaluatorType` (`evaluations-legacy.ts:861`), so a legacy row flips `isLegacyPairwiseDispatch` regardless of the in-memory type handed to it.
- **Ordering is right.** `getEvaluatorDataForParams` runs against the *rerouted* checkType (`:1025`), so `allowedExtras` is built from `select_best_compare`'s contract (`requiredFields: ["candidates"]`) and the translated `candidates` array survives the extras filter. Against the original pairwise type it would've been stripped — the co-location is load-bearing.
- **No settings 400.** `select_best_compare`'s zod object is non-strict → `swap_and_confirm` & friends are stripped, not rejected.
- **`params.settings` is empty on the orchestrator path** (the #5528 failure mode): nlp only forwards a param literally named `settings` (`engine.go:865` → `paramAnyMap`), and `workflowBuilder` spreads settings as individual top-level params — never one named `settings`. Route falls back to DB `evaluatorSettings`, which the strip covers.
- `tslsp diagnostics` clean on both changed source files.

### Findings (all P3, non-blocking)
- **P3 — `params.settings.prompt` bypasses the strip** (`evaluations-legacy.ts:992`). `stripIncompatiblePairwisePrompt` only sanitizes DB/monitor settings; the final merge spreads `...params.settings` last and un-stripped. Unreachable for orchestrator/monitors, but a **direct REST caller** with an inline `{candidate_a_output}` prompt would override the good default and render half-literal — the same failure via a side door. Fix: guard `params.settings.prompt` too, or strip after the full merge.
- **P3 — dropped-prompt log mislabels the evaluator** (`evaluations-legacy.ts:970`). Logs `checkType` *after* reroute (`select_best_compare`), reading as though the new judge carried the bad prompt. Log the original type.
- **P3 / note — silent `swap_and_confirm` drop** (`evaluations-legacy.ts:984`). A legacy config with `swap_and_confirm:false` reroutes into `randomize_order` defaulting `true` — opposite behavior, silently. Both default true so most rows unaffected; worth a line in the description rather than a code change.

### Residual risk
- **Live e2e not run** (PR admits this): seed a `pairwise_compare` DB row → re-run → confirm a `select_best_compare` verdict with a variant-id winner. Nothing exercises the actual langevals round-trip; unit tests only pin the pure translators. Worth the pass before merge.
- Test nit: `describe("regression: …")` breaks the given/when convention.
- No tenancy/PII concern — pure shape translation, no new queries.

_Review-only: findings as a comment; no code pushed._

sergioestebance (dismissed):
Ship-able — no P0/P1. Three P3 nits posted above (none blocking). Recommend the live e2e pass before merge.

Aryansharma28 (commented):
## goated-review follow-up — fixed, pushed as `4d52e892d`

Ran a second pass on top of @sergioestebance's already-thorough SHIP-ABLE review. No new P0/P1/P2s found — traced the reroute chokepoint and payload translation independently and it holds up as described. Fixed all three of @sergioestebance's P3s plus one new P3 of my own, in one commit.

### Fixed
- **[P3]** `params.settings.prompt` bypassed the strip (@sergioestebance's finding, `evaluations-legacy.ts:992`) — the strip ran on DB/monitor settings only, before `params.settings` was spread in last, so a direct REST caller with an inline `{candidate_a_output}` prompt could still reach `select_best_compare` half-literal. Now strips **after** the full settings merge.
- **[P3]** Dropped-prompt log used the post-reroute `checkType` (@sergioestebance's finding, `:970`) — always logged `select_best_compare` even when the actual dispatch was the legacy judge. Now logs the literal `LEGACY_PAIRWISE_EVALUATOR_TYPE` constant.
- **[P3]** (new, mine) `rawEvaluatorSettings` recomputed the exact same `evaluatorSettings ?? monitor.parameters ?? {}` fallback already sitting in `settings` a few lines above — two independent copies of the same logic that could silently drift apart on a future edit. Now reuses `settings` directly.
- **[P3/note]** `swap_and_confirm` → `randomize_order` silent default flip (@sergioestebance's finding) — per their own recommendation this didn't need a code change (both default `true`, narrow blast radius), so it's now documented with a comment at the merge site instead.

### Verified after the fix
- `pnpm typecheck` clean.
- All 37 existing tests across `evaluations-legacy-pairwise-adapter.unit.test.ts`, `evaluations-legacy-model-cascade.unit.test.ts`, and `normalizeComparison.test.ts` still pass unchanged.
- Did **not** add a new integration-level test for the settings-merge-ordering fix — `handleEvaluatorCall` isn't exported/unit-testable in isolation, and building that harness is bigger than this fix. This is the same gap @sergioestebance already flagged ("unit tests only pin the pure translators") — still open, still worth the live e2e pass before merge.

- Whether any other caller besides the traced orchestrator/monitor path constructs `params.settings.prompt` directly with a legacy `{candidate_a_output}`-style value — the fix covers it either way now, but I didn't find a live example of this path being hit today.

Requesting a re-review from @sergioestebance given the settings-merge logic changed after their approval.

sergioestebance:
Blessed. 🙏 Legacy `pairwise_compare` now reroutes to `select_best_compare` at a single interception point, so the dispatched type and the payload shape travel together and can never drift apart again — the app only ever talks to one judge.

> "So they are no longer two, but one flesh. Therefore what God has joined together, let no one separate." — Matthew 19:6

Traced the orchestrator → nlp → route path: the reroute resolves off the DB row's own type, `getEvaluatorDataForParams` runs against the rerouted contract so `candidates` survives the extras filter, and `params.settings` is empty on the orchestrator path so the prompt-strip holds. No drift path back to #5528. Ship it — residuals are P3 only (the `params.settings.prompt` side-door for direct REST callers, a mislabeled log line), plus the unrun live e2e already flagged.

sergioestebance (approved):
Blessed. 🙏 The reroute holds and the last gap is sealed — the legacy pairwise prompt is now stripped *after* the full settings merge, so nothing, not even a prompt smuggled in through the request body, can reach `select_best_compare` half-literal. One judge, one shape, no side door.

> "For there is nothing hidden that will not be disclosed, and nothing concealed that will not be brought into the open." — Mark 4:22

Re-reviewed the `4d52e892d` delta on top of my earlier pass: strip-after-merge closes the `params.settings.prompt` bypass, the dropped-prompt log now names the legacy type, and the duplicated `evaluatorSettings ?? monitor.parameters` fallback is gone. Typecheck clean, all three P3s from my pass resolved. Ship it — only residual is the unrun live e2e and the still-absent `handleEvaluatorCall` integration test, both already acknowledged.

PR #5807, opened 2026-07-15
feat(automations): generic HTTP webhook action (ADR-040)

0xdeafcafe:
## What

Adds a third automation notify channel — **`SEND_WEBHOOK`** — alongside Email and Slack. On a trace match or a graph alert, LangWatch renders a Liquid JSON body and sends it (POST/PUT/PATCH) to a customer-supplied endpoint over the SSRF-fenced HTTP client. Shipped **dark** behind `release_webhook_automations` (`FEATURE_FLAG_FORCE_ENABLE=release_webhook_automations` to try it in dev).

Implements ADR-040 Phases 1–3. Phase 4 (a dedicated `WebhookDelivery` per-attempt table + drill-down report UI) is deferred; delivery still rides the existing outbox retry machinery and fire history.

## How it maps to the ADR

**Phase 1 — Provider + config UI**
- `definitions/webhook/{shared,client,server}` modelled on the Slack provider. `actionParams` (Zod): https-only URL with a default-port + no-credentials shape check, `method` (POST/PUT/PATCH), sanitized static `headers` (reserved keys — `Host`, `Content-Length`, `Content-Type`, `x-langwatch-*` — stripped), and a Liquid JSON `bodyTemplate` stored **inside `actionParams`**, not a Trigger template column.
- Config form: URL input, method segmented control, key/value headers editor, `LiquidEditor` (JSON) body with a rendered request preview.
- Registry lines, `SliceFor`/`PreviewFor`/`initialSlices`, `NotifyClientDef.channel` widened to `"webhook"`, `SEND_WEBHOOK` enum + additive migration, `NOTIFY_TRIGGER_ACTIONS` membership (forced by the notify/persist exhaustiveness test).
- Flag gates the delivery-picker card **and** the save / test-fire routes server-side, so the dark launch can't be bypassed via the API.

**Phase 2 — SSRF-safe sender + dispatch**
- `server/triggers/sendWebhook.ts` wraps the existing `sendHttpDestination` with the webhook SSRF policy: `createSSRFValidator({ blockLocal: true, allowedHosts: [] })` forces private-IP blocking regardless of the global `BLOCK_LOCAL_HTTP_CALLS` toggle; redirects are refused (the strict validator isn't threaded through hops); a private/loopback IP literal — including bracketed IPv6 like `[::1]` — is blocked **terminally**.
- `renderWebhookBody` + default trace/alert/report JSON envelopes in `defaults.ts`, with `| json` on every interpolation so trace content containing `"`/`}` can't break out of the JSON.
- Wired into the outbox notify switch, the shared `dispatchGraphAlertAction` (which **no longer dead-letters** webhook — the required ADR edit), and a cron parity action.

**Phase 3 — Retries & idempotency**
- `assertWebhookDelivered` classifies 5xx / 429 / 408 as retryable, other non-2xx as terminal (ADR-040 §5). The graph-alert path gates each fire on the endpoint identity so an outbox retry never re-posts to an endpoint already reached.

## Test-fire behaviour (the "see the errors" ask)
Pressing **Send a test** actually sends through the full SSRF-fenced sender (with a non-suppressible `X-LangWatch-Test-Fire: true` header) and surfaces the outcome **inline right under the button**: `Delivered — HTTP 200` on success, or the failure title + detail (HTTP status / transport error) on failure — not just a toast.

## Tests
- `sendWebhook.unit.test.ts` — https/port shape rejection; **executed** SSRF blocks for private/loopback/link-local/metadata (incl. `[::1]`); retry-vs-terminal classification.
- `renderWebhookBody.unit.test.ts` — default trace/alert/report envelopes parse; JSON-breakout content stays parseable; custom template render; fall-back on invalid-JSON / render-throw.
- `webhook/shared.unit.test.ts` — URL shape, header sanitisation, schema defaults.
- `graphAlertActionDispatch` + `trigger-template.service` — webhook branch, retryable-vs-terminal, per-fire idempotency, and the webhook test-fire path.
- `pnpm typecheck` clean; full unit sweep of the touched areas green.

Spec: `specs/automations/webhook-http-action.feature`.

> Note: `pnpm lint` (biome) fails repo-wide on this branch with an unrelated config/version mismatch (unknown rule names); not touched here.

---

## Update — review follow-ups + Phase 3/4 (added after review)

Three follow-up commits on top of the original Phases 1–3:

**Review fixes (`f7649bd`)**
- **Header values are now secrets.** Encrypted at rest (`actionParams.headersEncrypted`, AES-256-GCM, same machinery as the Slack bot token), never returned to the client (reads echo names with a `__kept__` sentinel), decrypted just before dispatch, kept-on-edit. Closes the plaintext-`Authorization` gap the review flagged.
- **Webhook test-fires are rate-limited** (the email limiter, per user) — an uncapped test button at an arbitrary URL was an outbound flood vector.
- **Redirect refusal is a first-class `followRedirects:false`** on `fetchWithResolvedIp` (replacing the `_redirectCount: MAX_SAFE_INTEGER` trick), with an **executed** regression test driving a real local server that 302s toward metadata.
- Cron parity: a terminal failure now **consumes** the fire instead of re-posting every tick.

**Phase 3 completion (`f944fbc`)**
- `Retry-After` → `DispatchError.retryAfterMs` (delta-seconds + HTTP-date, capped 1h), honored by the GroupQueue as a backoff **floor**.
- Stable `X-LangWatch-Event-Id` per fire (trace path from the batch's traceIds, graph-alert path from the fire digest) so receivers dedupe replays.
- Per-project hourly dispatch cap (`webhook-dispatch:{projectId}`).

**Phase 4 — delivery log + report UI (`12e4b19`)**
- `WebhookDelivery` table (one row per attempt, grouped by `dispatchId`), written by a `deliverWebhook()` helper that sends + classifies + logs as a unit across both dispatch paths. Header values redacted to `***`; logging is best-effort (never breaks dispatch).
- `getWebhookDeliveries` procedure + a "Recent deliveries" drill-down in the automation drawer (attempts per fire, expandable to redacted headers + response).
- 30-day prune cron (`/api/cron/webhook_delivery_cleanup`) + daily Helm CronJob.
- **Design note:** shipped as a standalone table for now; ADR-040 amended to record that the per-attempt log arguably belongs *in* the outbox's audit mechanism (so all channels get delivery history) — flagged as the future unification, not built here.

**Still deferred:** HMAC request signing + the `ProjectSecret`-ref auth union (the only remaining Phase 2 gap).

`pnpm typecheck` clean; touched-area unit suites green.

## Summary by CodeRabbit

* **New Features**
  * Added webhook automations (HTTPS-only URLs, POST/PUT/PATCH, editable headers, JSON body templates) behind the `release_webhook_automations` feature flag.
  * Added webhook test-send with returned HTTP status and safer preview rendering.
  * Added per-trigger **Recent deliveries** with expandable attempts, colored outcomes, and redacted request/response details.
  * Added graph-alert webhook dispatch support and recorded delivery attempts.
* **Bug Fixes**
  * Improved webhook security: header secrets are encrypted at rest, never returned in cleartext, and are properly redacted in logs.
  * Enhanced retry behavior with `Retry-After` support, stable delivery event IDs, and deduping.
* **Chores**
  * Added 30-day webhook delivery log pruning via a scheduled cron endpoint.

### DIFF
--- a/langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
+++ b/langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
@@ -1,8 +1,10 @@
 -- ADR-040 §6: per-attempt webhook delivery log — the drill-down behind a
 -- webhook automation's "recent fires". One row per HTTP attempt; all attempts
 -- of one logical fire share `dispatchId` (== the X-LangWatch-Event-Id). Header
--- values are stored REDACTED (auth/signature masked to '***'); the response
--- snippet is size-capped by the writer. Rows are pruned after 30 days.
+-- values are stored REDACTED (every value masked to '***'); `requestUrl` is
+-- stored as origin + path only (query string and any embedded credentials are
+-- stripped before persistence); the response snippet is size-capped by the
+-- writer. Rows are pruned after 30 days, scoped by projectId.
 
 -- CreateEnum
 CREATE TYPE "WebhookDeliveryOutcome" AS ENUM ('success', 'retryable', 'terminal', 'pending');
@@ -31,7 +33,8 @@ CREATE TABLE "WebhookDelivery" (
 CREATE INDEX "WebhookDelivery_projectId_triggerId_dispatchId_idx" ON "WebhookDelivery"("projectId", "triggerId", "dispatchId");
 
 -- CreateIndex
-CREATE INDEX "WebhookDelivery_projectId_triggerId_firedAt_idx" ON "WebhookDelivery"("projectId", "triggerId", "firedAt");
+-- Indexes the projectId-scoped 30-day prune (`projectId IN (...) AND firedAt < before`).
+CREATE INDEX "WebhookDelivery_projectId_firedAt_idx" ON "WebhookDelivery"("projectId", "firedAt");
 
 -- AddForeignKey
 ALTER TABLE "WebhookDelivery" ADD CONSTRAINT "WebhookDelivery_projectId_fkey" FOREIGN KEY ("projectId") REFERENCES "Project"("id") ON DELETE CASCADE ON UPDATE CASCADE;
@@ -39,6 +42,7 @@ ALTER TABLE "WebhookDelivery" ADD CONSTRAINT "WebhookDelivery_projectId_fkey" FO
 -- AddForeignKey
 ALTER TABLE "WebhookDelivery" ADD CONSTRAINT "WebhookDelivery_triggerId_fkey" FOREIGN KEY ("triggerId") REFERENCES "Trigger"("id") ON DELETE CASCADE ON UPDATE CASCADE;
 
--- To roll back, uncomment and run manually:
--- DROP TABLE "WebhookDelivery";
--- DROP TYPE "WebhookDeliveryOutcome";
+-- IRREVERSIBLE: Prisma migrations are forward-only — this create has no
+-- automatic down step. To roll back manually, run:
+--   DROP TABLE "WebhookDelivery";
+--   DROP TYPE "WebhookDeliveryOutcome";
--- a/langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
+++ b/langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
@@ -1,0 +1,44 @@
+-- ADR-040 §6: per-attempt webhook delivery log — the drill-down behind a
+-- webhook automation's "recent fires". One row per HTTP attempt; all attempts
+-- of one logical fire share `dispatchId` (== the X-LangWatch-Event-Id). Header
+-- values are stored REDACTED (auth/signature masked to '***'); the response
+-- snippet is size-capped by the writer. Rows are pruned after 30 days.
+
+-- CreateEnum
+CREATE TYPE "WebhookDeliveryOutcome" AS ENUM ('success', 'retryable', 'terminal', 'pending');
+
+-- CreateTable
+CREATE TABLE "WebhookDelivery" (
+    "id" TEXT NOT NULL,
+    "projectId" TEXT NOT NULL,
+    "triggerId" TEXT NOT NULL,
+    "dispatchId" TEXT NOT NULL,
+    "requestMethod" TEXT NOT NULL,
+    "requestUrl" TEXT NOT NULL,
+    "requestHeaders" JSONB NOT NULL,
+    "responseStatus" INTEGER,
+    "responseBody" TEXT,
+    "latencyMs" INTEGER,
+    "error" TEXT,
+    "outcome" "WebhookDeliveryOutcome" NOT NULL,
+    "firedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "WebhookDelivery_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "WebhookDelivery_projectId_triggerId_dispatchId_idx" ON "WebhookDelivery"("projectId", "triggerId", "dispatchId");
+
+-- CreateIndex
+CREATE INDEX "WebhookDelivery_projectId_triggerId_firedAt_idx" ON "WebhookDelivery"("projectId", "triggerId", "firedAt");
+
+-- AddForeignKey
+ALTER TABLE "WebhookDelivery" ADD CONSTRAINT "WebhookDelivery_projectId_fkey" FOREIGN KEY ("projectId") REFERENCES "Project"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "WebhookDelivery" ADD CONSTRAINT "WebhookDelivery_triggerId_fkey" FOREIGN KEY ("triggerId") REFERENCES "Trigger"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- To roll back, uncomment and run manually:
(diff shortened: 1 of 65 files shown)


langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
```diff
@@ -0,0 +1,44 @@
...
+    "outcome" "WebhookDeliveryOutcome" NOT NULL,
+    "firedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "WebhookDelivery_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "WebhookDelivery_projectId_triggerId_dispatchId_idx" ON "WebhookDelivery"("projectId", "triggerId", "dispatchId");
+
```
langwatch-agent, line 34: **P1 — The migration does not implement the cascade declared in the Prisma schema.** `WebhookDelivery.trigger` and `.project` are declared `onDelete: Cascade`, but both foreign keys here omit `ON DELETE CASCADE`, whose PostgreSQL default is `NO ACTION`. Once a delivery row exists, hard-deleting its trigger or project (including the existing test cleanup `trigger.deleteMany`) fails with a foreign-key violation instead of removing delivery history. Add `ON DELETE CASCADE` to both foreign keys in this migration and cover a trigger/project delete after recording a delivery.

langwatch/src/automations/providers/definitions/webhook/shared.ts
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Redact a header record for the delivery log: which headers were sent is kept
+ * (an operator debugging a 401 needs to see `Authorization` was present), but
+ * the secret VALUE is masked to "***" so credential material never lands in
+ * control-plane Postgres (ADR-040 §6).
+ */
+export function redactHeadersForLog(
+  headers: Record<string, string>,
+): Record<string, string> {
+  const out: Record<string, string> = {};
```
langwatch-agent, line 68: **P1 — The delivery log leaks credentials in headers whose names do not match this heuristic.** The feature treats configured header values as secrets, but `redactHeadersForLog` only masks names matching a short regex. A valid `X-Partner-Key` or `X-Integration-Credential` header is stored verbatim by `deliverWebhook`, then returned by `automations.getWebhookDeliveries` to every user with `triggers:view`. Header names cannot reliably establish whether their values are secret. Persist names with all values masked (or use a narrow explicit allowlist of safe values), and add a regression test for a non-matching credential name.

langwatch-agent (changes requested):
2 P1s: the delivery-log migration does not implement its declared cascade lifecycle, and delivery history can disclose arbitrary webhook credentials. See inline comments.

langwatch/prisma/schema.prisma
```diff
@@ -802,12 +804,51 @@ model Trigger {
...
+// webhook automation's "recent fires". One row per HTTP attempt; all attempts
+// of one logical fire share `dispatchId` (== the X-LangWatch-Event-Id). Header
+// values are stored REDACTED (auth/signature masked to "***"); the response
+// snippet is size-capped. Pruned after 30 days.
+model WebhookDelivery {
+  id             String                @id @default(nanoid())
+  projectId      String
+  project        Project               @relation(fields: [projectId], references: [id], onDelete: Cascade)
+  triggerId      String
+  trigger        Trigger               @relation(fields: [triggerId], references: [id], onDelete: Cascade)
```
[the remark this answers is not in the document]
0xdeafcafe: Deferring this one (tracked as follow-up). Enforcing a composite \`(triggerId, projectId)\` FK requires adding \`@@unique([id, projectId])\` to the already-deployed \`Trigger\` table and a new migration on that shared core model — an index build/scan on a large existing table, which is heavier than the marginal safety here given the app layer and the multitenancy guard already enforce that \`WebhookDelivery.projectId\` matches its trigger. Everything else in this review batch is addressed in 851dfc858.

langwatch/prisma/migrations/20260715000100_create_webhook_delivery/migration.sql
0xdeafcafe: The FKs already carry \`ON DELETE CASCADE\` (migration lines 40/43) and the schema declares \`onDelete: Cascade\` on both relations — they have since the feature commit, so there is no cascade to add. I did add the requested coverage: a new integration test recording a delivery and then deleting its trigger, asserting the delivery row cascades away (851dfc858).

langwatch/src/server/routes/cron.ts
```diff
@@ -79,6 +79,38 @@ secured
...
+// ADR-040 §6: prune webhook delivery-log rows older than 30 days. Postgres is
+// outside the ClickHouse retention sweep, so the log needs its own cleaner.
+const webhookDeliveryCleanupHandler = async (c: CronContext) => {
+  if (!validateCronKey(c)) {
+    return c.body(null, 401);
+  }
+  try {
+    // Scope the prune by projectId (the delivery log is a project-level model),
+    // so enumerate every project and hand the ids to the service.
+    const projects = await prisma.project.findMany({ select: { id: true } });
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed in b27cf5c13. The handler now goes through the service layer: added `ProjectRepository.findAllIds()` / `ProjectService.getAllIds()` and the route resolves both the project ids and the prune via `getApp()` — no `prisma` reach-through. (The other `prisma` uses left in this file belong to pre-existing handlers outside this PR.)

0xdeafcafe:
Superseded by #5905, which carries the whole webhook channel (ADR-040) with dispatch rebuilt on the process-manager outbox (ADR-052) instead of the legacy ReactorOutbox.

PR #5814, opened 2026-07-15
feat(haven): lifecycle usability — safer down, restart, detach, switch, seed extras + hub web revamp

0xdeafcafe:
Improves haven's day-to-day CLI ergonomics, extends the seed, and revamps the hub web dashboard. Specs: `specs/setup/haven-lifecycle-usability.feature`, `specs/setup/haven-seed-presets.feature`. Also carries the earlier observability commit (backend OTLP export + pretty Go/TS console + skip CH backup metrics locally).

## haven CLI

- **`haven down` keeps databases by default.** `--drop-db` is the explicit fresh-DB ask. `down` also terminates a live launcher, so it works from any terminal.
- **Background database pruning.** Every `up` touches a per-slug activity clock; the daemon drops ClickHouse+Postgres databases idle past `HAVEN_DB_TTL` (default 14 days, `0` disables). Registered stacks are never touched, `lw_main` is always kept.
- **`haven up` refuses a doubly-started worktree**; `-f/--force` replaces it, `-w/--watch` enables air hot-reload for the Go services, `-d/--detach` backgrounds the stack with per-slug log file — new **`haven logs [-f]`** tails it.
- **New `haven restart [svc]`** (alias `rs`) — SIGTERMs the service's process group; the supervisor's crash-restart loop brings it back (~1s). One service or all supervised children; baseline fallbacks, shared DB servers, and the launcher pid are never targets.
- **New `haven switch [name]`** (alias `sw`) + **`haven shell-init`** — resolve a worktree dir by slug/name with prefix matching; the emitted shell function makes it a real `cd` with zsh tab-completion.
- User-facing copy now points at `haven up` (CLI-only), not `pnpm dev`.

## Seed extras (all flag-gated)

- **Model providers from the environment**: every registry provider whose API-key variable is set (process env → `langwatch/.env` → repo-root `.env`) is upserted as an enabled, org-scoped `ModelProvider` with schema-validated, repository-style-encrypted keys and fixed idempotent row IDs. Disable: `--skip-model-providers` / `HAVEN_SEED_MODEL_PROVIDERS=0`.
- **`--first-message` / `--no-first-message`** (`HAVEN_SEED_FIRST_MESSAGE=1|0`) force the project's firstMessage/integrated onboarding flags independent of the demo preset.
- **`--traces`** (`HAVEN_SEED_TRACES=1`) ingests the sample traces through the running collector without the full demo preset.

## Hub

- **Web dashboard revamp:** sticky glass header with the shared observability/telemetry links, flicker-free JS live refresh replacing `<meta refresh>` (pauses when hidden, shows daemon-unreachable state), branch-forward cards with an `open ↗` action, better empty state.
- **TUI:** `r` restarts the selected stack's services, `o` opens its app URL, and the selected row unfolds a per-service health/URL panel.

## Tests

`go test ./...` green across the haven module: new coverage for the up guard/force path, down keep/drop semantics, restart target resolution, idle-DB pruning, and seed option plumbing; dashboard/hubtui markup tests updated. `pnpm typecheck` clean.

## Deployment Impact

- **New env var (opt-in): `CLICKHOUSE_BACKUP_METRICS_ENABLED`.** ClickHouse backup-status gauges (`system.backup_log` queries each stats tick) are now strictly opt-in. On a vanilla `helm install` with no overrides, nothing changes at boot — but the backup gauges are **no longer emitted by default**, so deployments that have CH backups configured and rely on the "Backup Reporting Absent" alert must set this var to `true` to keep the signal. Documented in `langwatch/.env.example`.
- **No helm values added or changed**; no chart changes in this PR.
- Everything else (haven CLI lifecycle/seed/hub, dotenv noise reduction, Go/TS pretty console format) is local-dev tooling only — no operator or BYOC dataplane impact.

## Summary by CodeRabbit

* **New Features**
  * Added `haven switch` and `haven restart`, including hub actions to restart services and open the app URL.
  * Improved `haven up -d` with dedicated log following and enhanced worktree switching UX.
  * Expanded `haven seed` with env-gated model-provider credential seeding and independent first-message controls.
  * Added automatic idle worktree database pruning via `HAVEN_DB_TTL`, and `haven down --drop-db` to drop databases.
* **Bug Fixes**
  * Made ClickHouse backup-status metrics strictly opt-in via `CLICKHOUSE_BACKUP_METRICS_ENABLED`, avoiding backup-log queries when disabled and improving warn/recovery behavior.
  * Reduced dotenv startup noise and refined console “pretty” logging based on `LOG_FORMAT`.
* **Documentation**
  * Updated `haven` help and tooling docs to match revised lifecycle, flags, and examples.

### DIFF
--- a/langwatch/prisma/seed.ts
+++ b/langwatch/prisma/seed.ts
@@ -6,6 +6,14 @@
  * e2e job and charts/langwatch/tests/e2e-full-stack.sh run it against a fresh
  * database per CI run.
  *
+ * Optional extras (all env-gated, all idempotent):
+ *   - Model providers from the environment (HAVEN_SEED_MODEL_PROVIDERS=0
+ *     disables): every registry provider whose API-key variable is set in the
+ *     process env / langwatch/.env / repo-root .env gets an enabled,
+ *     org-scoped ModelProvider row with those keys.
+ *   - HAVEN_SEED_FIRST_MESSAGE=1|0 forces the project's firstMessage/
+ *     integrated flags on or off, independent of HAVEN_SEED_PRESET=demo.
+ *
  * Every identity value below is a fixed, hardcoded constant — nothing here is
  * randomly generated. The same admin login and the same organization/team/
  * project/user IDs and keys/tokens exist on every worktree and every
@@ -54,9 +62,15 @@
  * must stay useless without it). Verification always succeeds locally
  * because hashing and verifying both read that same local pepper.
  */
+import fs from "fs";
+import path from "path";
+
 import { PrismaClient, RoleBindingScopeType, TeamUserRole } from "@prisma/client";
 import { hash as hashPassword } from "bcrypt";
+import { parse as parseDotenv } from "dotenv";
 import { ENTERPRISE_LICENSE_KEY } from "../ee/licensing/__tests__/fixtures/testLicenses";
+import { modelProviders } from "../src/server/modelProviders/registry";
+import { encrypt } from "../src/utils/encryption";
 import {
   API_KEY_PREFIX,
   INGEST_KEY_PREFIX,
@@ -114,7 +128,13 @@ async function main() {
   // (firstMessage/integrated set), so the UI opens on the real product instead
   // of the "waiting for your first message" journey. `haven seed --preset demo`
   // sets this and then ingests sample traces through the collector.
-  const isPastOnboarding = process.env.HAVEN_SEED_PRESET === "demo";
+  // HAVEN_SEED_FIRST_MESSAGE=1|0 overrides that flag independently of the
+  // preset (`haven seed --first-message` sets it).
+  const firstMessageOverride = process.env.HAVEN_SEED_FIRST_MESSAGE;
+  const isPastOnboarding =
+    firstMessageOverride !== undefined
+      ? firstMessageOverride === "1" || firstMessageOverride === "true"
+      : process.env.HAVEN_SEED_PRESET === "demo";
 
   const organization = await prisma.organization.upsert({
     where: { id: ORG_ID },
@@ -331,6 +351,8 @@ async function main() {
     update: {},
   });
 
+  await seedModelProvidersFromEnv(organization.id);
+
   console.log(`✅ Organization: ${organization.id} (${organization.slug})`);
   console.log(`✅ Team:         ${team.id} (${team.slug})`);
   console.log(`✅ Project:      ${project.id} (${project.slug})`);
@@ -344,6 +366,102 @@ async function main() {
   console.log(`✅ Ingestion key:        ${displayApiKey}`);
   console.log(`✅ Private access token: ${PRIVATE_ACCESS_TOKEN}`);
   console.log(`✅ Public access token:  ${PUBLIC_ACCESS_TOKEN}`);
+}
+
+// ---------------------------------------------------------------------------
+// Model providers from the environment.
+//
+// For every provider in the registry whose primary API-key variable is set —
+// in the process env, langwatch/.env, or the repo-root .env — upsert an
+// enabled, ORGANIZATION-scoped ModelProvider carrying those keys (encrypted
+// exactly as modelProvider.repository does), so a fresh local stack can talk
+// to the providers the developer already has credentials for without pasting
+// them into the settings UI. Fixed per-provider row IDs keep re-runs
+// idempotent. HAVEN_SEED_MODEL_PROVIDERS=0 disables the whole block
+// (`haven seed --skip-model-providers`).
+// ---------------------------------------------------------------------------
+
+const MODEL_PROVIDER_ID_PREFIX = "local-dev-model-provider-";
+
+// loadSeedEnv merges the dotenv layers under the child's real precedence:
+// process env wins over langwatch/.env, which wins over the repo-root .env.
+// The seed always runs with cwd=langwatch/ (haven, CI, and the pnpm script
+// all invoke it there).
+function loadSeedEnv(): Record<string, string> {
+  const merged: Record<string, string> = {};
+  for (const file of [path.join("..", ".env"), ".env"]) {
+    try {
+      Object.assign(merged, parseDotenv(fs.readFileSync(file)));
+    } catch {
+      // missing file — fine, the layer just doesn't exist
+    }
+  }
+  for (const [k, v] of Object.entries(process.env)) {
+    if (v !== undefined) merged[k] = v;
+  }
+  return merged;
+}
+
+// schemaKeyNames lists a provider's credential variable names from its zod
+// keysSchema, unwrapping ZodEffects (superRefine) to reach the object shape.
+function schemaKeyNames(schema: unknown): string[] {
+  let inner = schema as { _def?: { schema?: unknown }; shape?: object };
+  while (inner?._def?.schema) {
+    inner = inner._def.schema as typeof inner;
+  }
+  return inner?.shape ? Object.keys(inner.shape) : [];
+}
+
+async function seedModelProvidersFromEnv(organizationId: string) {
+  const flag = process.env.HAVEN_SEED_MODEL_PROVIDERS;
+  if (flag === "0" || flag === "false") {
+    console.log("⏭️  Model providers: seeding disabled (HAVEN_SEED_MODEL_PROVIDERS=0)");
+    return;
+  }
+  const envMap = loadSeedEnv();
+  for (const [provider, def] of Object.entries(modelProviders)) {
+    // "custom" has no inferable identity from the environment; skip it.
+    if (provider === "custom") continue;
+    if (!envMap[def.apiKey]) continue;
+
+    const keyNames = schemaKeyNames(def.keysSchema);
+    const names = keyNames.length > 0 ? keyNames : [def.apiKey, def.endpointKey];
+    const keys: Record<string, string> = {};
+    for (const name of names) {
+      if (name && envMap[name]) keys[name] = envMap[name];
+    }
+    const parsed = def.keysSchema.safeParse(keys);
+    if (!parsed.success) {
+      console.log(
+        `⏭️  Model provider ${provider}: ${def.apiKey} is set but the key set is incomplete — skipped`,
+      );
+      continue;
+    }
+
+    const id = MODEL_PROVIDER_ID_PREFIX + provider;
+    const customKeys = encrypt(JSON.stringify(keys));
+    const row = await prisma.modelProvider.upsert({
+      where: { id },
+      create: { id, name: def.name, provider, enabled: true, customKeys, organizationId },
+      update: { customKeys, enabled: true, disabledAt: null },
+    });
+    await prisma.modelProviderScope.upsert({
+      where: {
+        modelProviderId_scopeType_scopeId: {
+          modelProviderId: row.id,
+          scopeType: "ORGANIZATION",
+          scopeId: organizationId,
+        },
+      },
+      create: {
+        modelProviderId: row.id,
+        scopeType: "ORGANIZATION",
+        scopeId: organizationId,
+      },
+      update: {},
+    });
+    console.log(`✅ Model provider: ${provider} (keys from environment)`);
+  }
 }
 
 main()
--- a/langwatch/prisma/seed.ts
+++ b/langwatch/prisma/seed.ts
@@ -131,10 +131,10 @@ async function main() {
   // HAVEN_SEED_FIRST_MESSAGE=1|0 overrides that flag independently of the
   // preset (`haven seed --first-message` sets it).
   const firstMessageOverride = process.env.HAVEN_SEED_FIRST_MESSAGE;
-  const isPastOnboarding =
-    firstMessageOverride !== undefined
-      ? firstMessageOverride === "1" || firstMessageOverride === "true"
-      : process.env.HAVEN_SEED_PRESET === "demo";
+  const hasFirstMessageOverride = firstMessageOverride !== undefined;
+  const isPastOnboarding = hasFirstMessageOverride
+    ? firstMessageOverride === "1" || firstMessageOverride === "true"
+    : process.env.HAVEN_SEED_PRESET === "demo";
 
   const organization = await prisma.organization.upsert({
     where: { id: ORG_ID },
@@ -171,9 +171,12 @@ async function main() {
       firstMessage: isPastOnboarding,
       integrated: isPastOnboarding,
     },
-    update: isPastOnboarding
-      ? { apiKey, firstMessage: true, integrated: true }
-      : { apiKey },
+    // An explicit override must also be able to CLEAR the flags
+    // (`haven seed --no-first-message`); without one, an existing true is kept.
+    update:
+      hasFirstMessageOverride || isPastOnboarding
+        ? { apiKey, firstMessage: isPastOnboarding, integrated: isPastOnboarding }
+        : { apiKey },
   });
 
   // Admin user + BetterAuth credential (email/password) login.
@@ -430,10 +433,31 @@ async function seedModelProvidersFromEnv(organizationId: string) {
     for (const name of names) {
       if (name && envMap[name]) keys[name] = envMap[name];
     }
+    // The registry schemas mark every key `.nullable().optional()` (to allow
+    // env-var fallback in inbound payloads), so safeParse alone would happily
+    // seed an enabled-but-unusable provider (e.g. Bedrock with only the access
+    // key). Require every non-optional key; Azure needs its API key plus
+    // either mode's endpoint, not both.
+    const optionalKeys = new Set(
+      "optionalKeys" in def ? (def.optionalKeys ?? []) : [],
+    );
+    let missing = names.filter(
+      (name): name is string =>
+        Boolean(name) && !optionalKeys.has(name!) && !keys[name!],
+    );
+    if (provider === "azure") {
+      const endpointNames = ["AZURE_OPENAI_ENDPOINT", "AZURE_API_GATEWAY_BASE_URL"];
+      missing = missing.filter((name) => !endpointNames.includes(name));
+      if (!endpointNames.some((name) => keys[name])) {
+        missing.push(endpointNames.join(" or "));
+      }
+    }
     const parsed = def.keysSchema.safeParse(keys);
-    if (!parsed.success) {
+    if (!parsed.success || missing.length > 0) {
       console.log(
-        `⏭️  Model provider ${provider}: ${def.apiKey} is set but the key set is incomplete — skipped`,
+        `⏭️  Model provider ${provider}: ${def.apiKey} is set but the key set is incomplete${
+          missing.length > 0 ? ` (missing ${missing.join(", ")})` : ""
+        } — skipped`,
       );
       continue;
     }
--- a/langwatch/src/server.mts
+++ b/langwatch/src/server.mts
@@ -12,11 +12,11 @@ import Module from "module";
 // reaches the running process (and Langy can't reach the gateway from inside
 // the OpenCode pod). NODE_ENV / PORT / similar process-level vars stay
 // shell-only because .env shouldn't carry them.
-dotenv.config({ override: true });
+dotenv.config({ override: true, quiet: true });
 // Portless (haven) overlay: loaded LAST with override so the resolved hostname
 // URLs + ports win over anything pinned in .env. In non-portless runs the file
 // is absent and this is a no-op. See tools/thuishaven + ADR-048.
-dotenv.config({ path: ".env.portless", override: true });
+dotenv.config({ path: ".env.portless", override: true, quiet: true });
 setEnvironment(process.env.ENVIRONMENT ?? "local");
 
 if (process.env.NODE_ENV === "production") {
@@ -41,6 +41,16 @@ const originalResolveFilename = (Module as any)._resolveFilename;
   }
   return originalResolveFilename.call(this, request, parent, isMain, options);
 };
+
+// Load OTel instrumentation before the app graph evaluates. instrumentation.node
+// registers the tracer + OTLP exporters (traces/logs/metrics) and must run before
+// any span-creating module imported by ./start. The Vite+Hono server has no
+// Next.js instrumentation hook to do this, and under haven's single-process
+// default the workers lane (workers.ts, which does the same import) never runs —
+// so without this the API process exports no telemetry at all. Dynamic + after
+// the dotenv.config() calls above so it reads the loaded .env/.env.portless
+// (OTEL_EXPORTER_OTLP_ENDPOINT); it is a no-op when observability is unconfigured.
+await import("./instrumentation.node");
 
 // Intentional inline dynamic import (exception to the "no inline import" rule):
 // - `./start` must not evaluate until after dotenv.config() above has run,
--- a/langwatch/src/server.mts
+++ b/langwatch/src/server.mts
@@ -1,6 +1,7 @@
 import { setEnvironment } from "@langwatch/ksuid";
 import dotenv from "dotenv";
 import events from "events";
+import { existsSync } from "fs";
 import Module from "module";
 
 // `override: true` lets `.env` win over values that scripts/start.sh exported
@@ -12,11 +13,23 @@ import Module from "module";
 // reaches the running process (and Langy can't reach the gateway from inside
 // the OpenCode pod). NODE_ENV / PORT / similar process-level vars stay
 // shell-only because .env shouldn't carry them.
-dotenv.config({ override: true, quiet: true });
+//
+// dotenv's "injected env (N) from .env" line stays LOUD in local dev — it's the
+// one confirmation of which env files actually loaded (and it prints before any
+// logger exists; dotenv has no logger hook, only quiet). Prod and tests silence
+// it: there it's a stray non-JSON stdout line on every boot.
+const quiet = process.env.NODE_ENV !== "development";
+dotenv.config({ override: true, quiet });
 // Portless (haven) overlay: loaded LAST with override so the resolved hostname
 // URLs + ports win over anything pinned in .env. In non-portless runs the file
-// is absent and this is a no-op. See tools/thuishaven + ADR-048.
-dotenv.config({ path: ".env.portless", override: true, quiet: true });
+// is absent and this is a no-op — and stays quiet, or dotenv would announce
+// "injected env (0)" from a file that isn't there.
+// See tools/thuishaven + ADR-048.
+dotenv.config({
+  path: ".env.portless",
+  override: true,
+  quiet: quiet || !existsSync(".env.portless"),
+});
 setEnvironment(process.env.ENVIRONMENT ?? "local");
 
 if (process.env.NODE_ENV === "production") {
--- a/langwatch/src/server/clickhouse/metrics.ts
+++ b/langwatch/src/server/clickhouse/metrics.ts
@@ -232,6 +232,85 @@ const MONITORED_TABLES = [
 ];
 
 /**
+ * Collects ClickHouse backup status from system.backup_log into the backup gauges.
+ * Extracted so collectStorageStats can gate it out in local dev (see the call site).
+ *
+ * We deliberately query system.backup_log instead of system.backups: system.backups
+ * is an in-memory table that gets wiped on CH restart, which happens on every app
+ * deploy (the CH image tag is bumped by the build pipeline). After a restart the
+ * in-memory table is empty until the next scheduled backup runs, so a freshly-rolled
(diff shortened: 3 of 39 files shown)


langwatch/src/server/clickhouse/metrics.ts
```diff
@@ -231,6 +231,85 @@ const MONITORED_TABLES = [
...
+ *
+ * We deliberately query system.backup_log instead of system.backups: system.backups
+ * is an in-memory table that gets wiped on CH restart, which happens on every app
+ * deploy (the CH image tag is bumped by the build pipeline). After a restart the
+ * in-memory table is empty until the next scheduled backup runs, so a freshly-rolled
+ * worker pod sees zero rows and never emits the gauge, tripping the "Backup Reporting
+ * Absent" alert despite backups being healthy. system.backup_log is a persistent
+ * system table that retains entries across restarts.
+ */
+async function collectBackupStats(client: ClickHouseClient): Promise<void> {
```
0xdeafcafe, line 247: do we not already do a check like this in workers? can we share?

langwatch/src/server/clickhouse/metrics.ts
```diff
@@ -231,6 +231,85 @@ const MONITORED_TABLES = [
...
+      );
+      backupStatsCollectionFailing = false;
+    }
+  } catch (backupError) {
+    // Even where it IS collected the table can be transiently unavailable (a CH
+    // restart mid-tick), so a failure is handled, not fatal. Only production
+    // surfaces it — edge-triggered, once, until it recovers; elsewhere (the
+    // test/staging path) it stays at debug, below the console floor. Local dev never
+    // reaches here — the caller skips the query entirely.
+    const isProd = process.env.NODE_ENV === "production";
```
0xdeafcafe: does not check prod, maybe opt in to backup querying with an env var instead?

langwatch/src/server.mts
```diff
@@ -12,11 +12,11 @@ import Module from "module";
 // reaches the running process (and Langy can't reach the gateway from inside
 // the OpenCode pod). NODE_ENV / PORT / similar process-level vars stay
 // shell-only because .env shouldn't carry them.
-dotenv.config({ override: true });
+dotenv.config({ override: true, quiet: true });
```
0xdeafcafe: keep them loud on local dev? pass in logger to dotenv?

langwatch/src/server.mts
```diff
@@ -12,11 +12,11 @@ import Module from "module";
...
 // Portless (haven) overlay: loaded LAST with override so the resolved hostname
 // URLs + ports win over anything pinned in .env. In non-portless runs the file
 // is absent and this is a no-op. See tools/thuishaven + ADR-048.
-dotenv.config({ path: ".env.portless", override: true });
+dotenv.config({ path: ".env.portless", override: true, quiet: true });
```
0xdeafcafe: ditto. or, we don't do this and all env vars are set by haven itself?

tools/thuishaven/adapters/dashboard/html.go
```diff
@@ -72,13 +71,24 @@ func renderCard(s domain.Stack, probes Probes) stackCard {
...
+		if svc.Name == "app" && svc.URL != "" {
+			action = fmt.Sprintf(`<a class="open" href="%s">open ↗</a>`, html.EscapeString(svc.URL))
+			break
+		}
+	}
+
 	c.html = fmt.Sprintf(`
       <section class="card">
-        <header><span class="slug">%s</span><span class="pill %s">%s</span></header>
+        <header><span class="slug">%s</span><span class="spacer"></span>%s<span class="pill %s">%s</span></header>
```
0xdeafcafe: use templates?

0xdeafcafe (commented):
loggers should be info -> grafana, warn+ to console. if booted from haven.

pretty should show log level. try and match pretty format between go and js.

langwatch/src/server/clickhouse/metrics.ts
0xdeafcafe: No pre-existing shared check — the NODE_ENV gate was new in this PR. It's gone now anyway: ee06bc384 replaces it with an explicit `CLICKHOUSE_BACKUP_METRICS_ENABLED` opt-in, one gate at the `collectStorageStats` call site, which covers both the app (haven's in-process workers) and the standalone worker since both go through it.

langwatch/src/server/clickhouse/metrics.ts
0xdeafcafe: Done in ee06bc384 — NODE_ENV gating dropped entirely; backup querying is now opt-in via `CLICKHOUSE_BACKUP_METRICS_ENABLED=true` (to be set only where the clickhouse-serverless backup cronjobs actually run — needs adding to the prod worker env when this lands). Anyone opted in gets the edge-triggered warn on failure. Spec'd in `specs/ops/clickhouse-backup-metrics.feature`.

langwatch/src/server.mts
0xdeafcafe: Done in d389cca17 — loud in dev, quiet in prod/tests. dotenv has no logger hook (only `quiet`/`debug`), and it runs before any logger exists anyway, so conditional `quiet` is the lever.

langwatch/src/server.mts
0xdeafcafe: haven already injects the resolved env into its supervised children — the file exists for everything haven doesn't spawn: a manual `pnpm dev`, prisma scripts, one-off `tsx` scripts, tests run from the worktree. Drop the file and those all fall back to stale .env URLs. So kept, but the loud/quiet treatment now skips the "injected env (0)" noise when the overlay is absent (d389cca17).

tools/thuishaven/adapters/dashboard/html.go
0xdeafcafe: Done in 29f71df4e — view structs + one parsed `html/template`, contextual auto-escaping instead of per-callsite `html.EscapeString`, and the CSS no longer needs `%%` escapes.

0xdeafcafe:
Re the review note on log routing + pretty format:

- **info → grafana, warn+ → console under haven**: already wired — haven's overlay sets `LOG_OTEL_LEVEL=debug` (so info and below all reach the collector) and `LOG_CONSOLE_LEVEL=warn` by default while the observability stack is up (opt out with `LW_OBS_CONSOLE_LEVEL`), for both the TS app (pino transport split) and the Go services (clog's tee'd cores). Only applies when booted from haven; plain `pnpm dev` is untouched.
- **pretty shows log level**: pino-pretty already rendered it (`[12:19:00.616] INFO (name): msg`) — the string level formatter doesn't break it. The Go side showed `12:19PM INF >`; 51d0705b3 changes clog's pretty encoder to `12:19:00.616 INFO >` — 24h timestamp with millis and full-word capital level, so the Go and JS lanes now read as one format.

langwatch/prisma/seed.ts
```diff
@@ -114,7 +128,13 @@ async function main() {
...
   // of the "waiting for your first message" journey. `haven seed --preset demo`
   // sets this and then ingests sample traces through the collector.
-  const isPastOnboarding = process.env.HAVEN_SEED_PRESET === "demo";
+  // HAVEN_SEED_FIRST_MESSAGE=1|0 overrides that flag independently of the
+  // preset (`haven seed --first-message` sets it).
+  const firstMessageOverride = process.env.HAVEN_SEED_FIRST_MESSAGE;
+  const isPastOnboarding =
+    firstMessageOverride !== undefined
+      ? firstMessageOverride === "1" || firstMessageOverride === "true"
+      : process.env.HAVEN_SEED_PRESET === "demo";
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: the upsert update now applies `firstMessage`/`integrated` from the override whenever `HAVEN_SEED_FIRST_MESSAGE` was supplied, so `haven seed --no-first-message` clears both flags; without an override the previous keep-existing behavior is unchanged.

langwatch/prisma/seed.ts
```diff
@@ -346,6 +368,102 @@ async function main() {
...
+    for (const name of names) {
+      if (name && envMap[name]) keys[name] = envMap[name];
+    }
+    const parsed = def.keysSchema.safeParse(keys);
+    if (!parsed.success) {
+      console.log(
+        `⏭️  Model provider ${provider}: ${def.apiKey} is set but the key set is incomplete — skipped`,
+      );
+      continue;
+    }
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: seeding now requires every non-optional key (per the registry's `optionalKeys`) before encrypting — Bedrock needs all three AWS creds; Azure needs its API key plus either `AZURE_OPENAI_ENDPOINT` or `AZURE_API_GATEWAY_BASE_URL` (the two modes are alternatives, not both). Incomplete sets are skipped with the missing names listed.

langwatch/src/server/clickhouse/metrics.ts
```diff
@@ -231,6 +231,85 @@ const MONITORED_TABLES = [
...
+      query: `
+        SELECT
+          status,
+          count() as cnt,
+          maxIf(end_time, status = 'BACKUP_CREATED') as last_success_time,
+          argMaxIf(total_size, end_time, status = 'BACKUP_CREATED') as last_success_size
+        FROM system.backup_log
+        GROUP BY status
+      `,
+    });
```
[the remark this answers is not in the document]
0xdeafcafe: Not applicable: `system.backup_log` is a ClickHouse *system* table — server-level backup history with a fixed schema (status, end_time, total_size, …). It has no `TenantId` column, so adding `WHERE TenantId = {tenantId:String}` would make the query error. The TenantId rule guards our tenant-scoped data tables; this metric is intentionally instance-wide (backup health for the whole server).

specs/setup/haven-seed-presets.feature
```diff
@@ -42,3 +42,23 @@ Feature: Seed presets — a database that is ready to look at
...
+    When I run "haven seed"
+    Then that provider is seeded as an enabled org-scoped credential
+    And re-running updates the same credential instead of duplicating it
+    And "haven seed --skip-model-providers" seeds no providers
+
+  Scenario: Extras are individually controllable
+    When I run "haven seed --first-message"
+    Then the project is marked as having received its first trace
+    When I run "haven seed --no-first-message"
+    Then that flag is cleared again
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: both seed-script scenarios are now tagged `@integration @unimplemented` with a note explaining no harness runs `prisma/seed.ts` today (the Go tests only cover the flag → HAVEN_SEED_* plumbing). Also bound the new trace-only scenario to `TestSeedPresets` while in there.

tools/thuishaven/adapters/dashboard/html.go
```diff
@@ -245,11 +269,44 @@ const pageTemplate = `<!doctype html><html lang="en"><head>
...
+(() => {
+  let failures = 0;
+  const beat = document.getElementById('beat'), stamp = document.getElementById('stamp');
+  async function refresh() {
+    if (document.hidden) return;
+    try {
+      const res = await fetch('/', {cache: 'no-store'});
+      const doc = new DOMParser().parseFromString(await res.text(), 'text/html');
+      const next = doc.getElementById('live');
+      if (next) document.getElementById('live').replaceChildren(...next.children);
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: `refresh()` now skips the swap while `#live` contains `document.activeElement`, so keyboard focus survives the 3s refresh.

tools/thuishaven/adapters/hubtui/hubtui.go
```diff
@@ -16,21 +16,36 @@ import (
...
 	ServicesUp        int
 	ServicesTotal     int
+	AppURL            string
+	Services          []ServiceRow
 }
 
 // Actions wires the hub to the world. Rows is re-read on every refresh tick;
 // Down and Destroy run against the selected row when the user confirms.
+// Restart bounces the selected stack's supervised services in place; OpenURL
+// opens the stack's app URL in the browser. Either may be nil (action hidden).
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: the footer is built from the selected row — `o open` only when `OpenURL` is wired and the row has an `AppURL`, `r restart` only when `Restart` is wired and the row is live — matching the guards in `updateBrowse`.

tools/thuishaven/adapters/hubtui/hubtui.go
```diff
@@ -275,27 +306,58 @@ var (
 
 func (m model) View() string {
 	var b strings.Builder
-	b.WriteString(styleTitle.Render(" thuishaven hub "))
-	b.WriteString(styleDim.Render("— every stack, and what to do with it\n\n"))
+
+	// Header: title + a live summary of the whole machine.
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: removed the header-rendering narration in hubtui.go and the `openInBrowser` narration in hub.go.

tools/thuishaven/app/daemon.go
```diff
@@ -121,6 +122,15 @@ func (o *Orchestrator) monitorLoop(ctx context.Context) {
...
 				o.hyg.PruneGitWorktrees(o.cfg.RepoRoot)
 			}
+			// Every ~10 min, refresh the idle clock of every registered stack's
+			// databases, and prune databases whose worktree has not been up in
+			// DBIdleTTL — the unattended counterpart of `haven down --drop-db`.
+			if cycles%60 == 1 {
+				for _, s := range o.store.Stacks() {
+					_ = o.store.TouchDBActivity(s.Slug)
+				}
+				o.pruneIdleDatabases(ctx)
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: the daemon now checks each `TouchDBActivity` error, logs a warning, and skips `pruneIdleDatabases` for the cycle when any refresh failed to persist — pruning only runs off a fully refreshed clock.

tools/thuishaven/app/orchestrator.go
```diff
@@ -141,6 +141,9 @@ func (o *Orchestrator) provision(ctx context.Context, p UpParams, opts PlanOptio
 	if err := o.store.SaveStack(st); err != nil {
 		return domain.Stack{}, nil, err
 	}
+	// Start (or refresh) the databases' idle clock: the daemon prunes databases
+	// whose slug has not been up for DBIdleTTL, and this is what "up" means.
+	_ = o.store.TouchDBActivity(st.Slug)
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: `provision` now fails the up (and runs its route/registry cleanup) when `TouchDBActivity` returns an error, so a stale idle clock can never sit behind a reported-successful `up`.

tools/thuishaven/app/orchestrator.go
```diff
@@ -201,6 +204,9 @@ func (o *Orchestrator) Up(ctx context.Context, p UpParams, opts PlanOptions) err
 	if err := o.proxy.EnsureReady(); err != nil {
 		return fmt.Errorf("could not start the portless proxy automatically (%w) — run `haven setup` once to bootstrap it by hand", err)
 	}
+	if err := o.replaceRunningStack(p, opts.ShouldForce); err != nil {
+		return err
+	}
 	o.ensureDaemon(p.WorktreeDir)
 	st, cleanup, err := o.provision(ctx, p, opts, true)
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: `Up` acquires the machine-wide flock semaphore (`up-<slug>`, 1 slot) before the already-running guard and releases it right after provision/registration. Deliberately not held across supervision — `up --force` would otherwise deadlock waiting on the launcher it is about to replace; flocks die with their process so a killed launcher cannot leak the slot.

tools/thuishaven/app/orchestrator.go
```diff
@@ -467,7 +520,7 @@ func (o *Orchestrator) seedSampleTraces(ctx context.Context, p UpParams) error {
 	}
 	st, ok := o.stackBySlug(slug)
 	if !ok || !o.sys.ProcessAlive(st.LauncherPID) {
-		return fmt.Errorf("stack %q is not running — sample traces go through the real collector, so start it (pnpm dev:haven) and re-run `haven seed --preset demo`", slug)
+		return fmt.Errorf("stack %q is not running — sample traces go through the real collector, so start it (haven up) and re-run `haven seed --preset demo`", slug)
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: `seedSampleTraces` takes the retry command from the invoked mode — `haven seed --traces` for trace-only runs, `--preset demo` only for the demo preset — and the seed test pins it.

tools/thuishaven/app/ports.go
```diff
@@ -43,6 +43,15 @@ type Store interface {
 	WriteHMRGate(lwDir string, expiryUnixMs int64) error
 	ReadHMRGate(lwDir string) (int64, bool)
 	ClearHMRGate(lwDir string)
+	// TouchDBActivity records "slug's databases were in use now" — the clock the
+	// daemon's idle-database pruning reads. Touched on every `up` and refreshed
+	// by the daemon while a stack stays registered.
+	TouchDBActivity(slug string) error
+	// DBActivity returns the last-seen time per slug (empty map when none).
+	DBActivity() map[string]time.Time
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: dropped the restating comments on `DBActivity`, `PIDsOnPort` (ports.go) and `TouchDBActivity`/`DBActivity`/`RemoveDBActivity` (store.go). The why lives on the interface's `TouchDBActivity` doc (the idle-pruning clock), which the siblings sit under.

tools/thuishaven/cmd/hub.go
```diff
@@ -43,21 +45,44 @@ func (d deps) hubActions() hubtui.Actions {
...
 	}
 }
+
+// openInBrowser opens a URL with the platform opener.
+func openInBrowser(url string) error {
+	opener := "open" // macOS
+	if runtime.GOOS != "darwin" {
+		opener = "xdg-open"
+	}
+	return exec.Command(opener, url).Start()
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: `openInBrowser` keeps the `exec.Cmd`, and after a successful `Start` reaps it with `go cmd.Wait()`.

tools/thuishaven/cmd/root.go
```diff
@@ -400,6 +429,128 @@ func runHavenUpIn(ctx context.Context, dir string) error {
...
+		}
+	}
+	cmd := exec.Command(argv[0], argv[1:]...)
+	cmd.Dir = d.worktree
+	cmd.Env = os.Environ()
+	f, ferr := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
+	if ferr != nil {
+		return fmt.Errorf("opening log file %s: %w", logPath, ferr)
+	}
+	cmd.Stdout, cmd.Stderr = f, f
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: the detached log opens with `0o600` and is `Chmod`-ed to `0o600` after opening so files created `0644` by older runs are tightened too.

tools/thuishaven/cmd/root.go
```diff
@@ -463,6 +614,23 @@ func flagValue(args []string, name string) string {
...
+	if hasFlag(rest, "--first-message") {
+		env = append(env, "HAVEN_SEED_FIRST_MESSAGE=1")
+	}
+	if hasFlag(rest, "--no-first-message") {
+		env = append(env, "HAVEN_SEED_FIRST_MESSAGE=0")
+	}
+	if hasFlag(rest, "--skip-model-providers") {
+		env = append(env, "HAVEN_SEED_MODEL_PROVIDERS=0")
+	}
+	return env
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in daa38059e: the seed dispatch rejects `--first-message` together with `--no-first-message` with a clear error before anything runs.

tools/thuishaven/adapters/dashboard/html.go
```diff
@@ -2,83 +2,133 @@ package dashboard
...
+	Label string
+}
+
+type pageView struct {
+	ObsURL, ObsHost string
+	TelURL, TelHost string
+	Stats           []statView
+	Cards           []cardView
+	Empty           bool
+}
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 36bc3e0e9: renamed to `isLive` / `IsBaseline` / `IsSub` / `IsEmpty`, and `statView.N` is now `Value`.

langwatch/src/server/clickhouse/metrics.ts
```diff
@@ -270,78 +348,16 @@ export async function collectStorageStats(
...
-      }
+    // Backup status only exists where backups are configured (the production
+    // cluster, via clickhouse-serverless's backup cronjobs) — everywhere else this
+    // query just fails on every 15s tick for nothing, and NODE_ENV can't tell
+    // "has backups" from "staging/self-hosted production build". So the deployment
+    // that has backups opts in explicitly (set on the worker alongside the backup
+    // cronjobs). Gating here (not at one call site) covers both the app under
+    // haven's in-process workers and the standalone worker.
+    // See specs/ops/clickhouse-backup-metrics.feature.
+    if (process.env.CLICKHOUSE_BACKUP_METRICS_ENABLED === "true") {
```
langwatch-agent, line 359: **P1 — Production backup monitoring is disabled by this change.**

`collectStorageStats` now skips `system.backup_log` unless this variable is exactly `true`, but the production worker deployment in `langwatch-saas/infrastructure/langwatch_workers.tf` has no `CLICKHOUSE_BACKUP_METRICS_ENABLED` environment entry. The same deployment has active alerts for `clickhouse_backup_last_success_timestamp_seconds` and `clickhouse_backup_status_total` in `infrastructure/grafana/alerts.json`.

After this app image deploys, workers stop emitting both gauges: the reporting-absent alert fires spuriously and, more importantly, backup failures/staleness are no longer monitored. Add the production configuration in the rollout that ships this code (and the corresponding documented Helm configuration), or preserve the existing enabled behavior until that config is in place. Add a deployment-contract test or release check that prevents the gate and worker config from drifting.

tools/thuishaven/app/orchestrator.go
```diff
@@ -262,29 +295,60 @@ func (o *Orchestrator) UpStub(ctx context.Context, p UpParams, echo func(ports [
...
+	}
 	for _, r := range domain.PerWorktreeServices {
 		o.proxy.Remove(r.Name, slug)
 	}
 	o.proxy.Remove(domain.ClickHouseService, slug)
 	o.proxy.Remove(domain.PostgresService, slug)
-	if o.ch != nil && o.cfg.ShouldManageClickHouse && !shouldKeepDB {
+	if o.ch != nil && o.cfg.ShouldManageClickHouse && shouldDropDB {
 		db := domain.DatabaseForSlug(slug)
 		if err := o.ch.DropDatabase(ctx, db); err != nil {
```
langwatch-agent, line 355: **P2 — `haven down --drop-db` can report a clean reset while retaining the old database.**

A ClickHouse or Postgres drop failure is only logged. The method then removes the stack entry, prints `stack ... torn down`, and returns `nil`. A developer can therefore run `haven down --drop-db && haven up` and unknowingly reuse stale state, despite explicitly requesting a fresh database.

Attempt both drops so cleanup still completes, aggregate any failures, and return an error after route/registry cleanup. Cover a failing ClickHouse and Postgres drop in the command tests.

tools/thuishaven/app/restart.go
```diff
@@ -0,0 +1,106 @@
...
+		for _, svc := range st.Services {
+			if svc.Name == r.Name && !svc.IsFallback && svc.Port != 0 {
+				all = append(all, restartTarget{Name: svc.Name, Port: svc.Port})
+			}
+		}
+	}
+	if st.APIPort != 0 {
+		all = append(all, restartTarget{Name: "api", Port: st.APIPort})
+	}
+	if st.WorkerMetricsPort != 0 {
```
langwatch-agent: **P2 — `workers` is exposed as a restart target even in the default in-process worker mode.**

`provision` always assigns `WorkerMetricsPort`, but `planChildren` defaults `ShouldRunWorkersInProcess` to true. In that mode the metrics port belongs to the `api` child, not a standalone workers child. `haven restart workers` consequently finds the API PID and terminates its process group. This contradicts the comment that workers are listed only when a standalone lane exists, and turns a narrow worker restart into an API restart.

Persist whether workers are standalone when building the stack, or otherwise determine/deduplicate the owning process group before publishing restart targets. Add coverage for the default in-process mode and for `restart` without a service name.

langwatch-agent (changes requested):
Request changes: 1 P1 and 2 P2 correctness/reliability issues. Details are inline.

langwatch/src/server/clickhouse/metrics.ts
langwatch-agent: Rechecked at current head `36bc3e0e927690987df7d7228989bf5702e66e5b`: this remains unresolved. The implementation still gates collection solely on `CLICKHOUSE_BACKUP_METRICS_ENABLED === "true"`, and this PR still contains no production worker/Helm configuration that sets it. Please address the deployment contract before merge.

tools/thuishaven/app/orchestrator.go
langwatch-agent: Rechecked at current head `36bc3e0e927690987df7d7228989bf5702e66e5b`: this remains unresolved. Both drop errors are still warning-only and `Down` still removes the stack and returns `nil`. Please aggregate and return the failures after the route/registry cleanup.

tools/thuishaven/app/restart.go
langwatch-agent: Rechecked at current head `36bc3e0e927690987df7d7228989bf5702e66e5b`: this remains unresolved. `restartTargets` still adds `workers` whenever `WorkerMetricsPort != 0`, without recording or checking whether workers are standalone; the default in-process setup therefore still targets the API process group.

0xdeafcafe: Fixed in 6a3af9764. The stack now persists `HasStandaloneWorkers` (set in `provision` from `opts.ShouldStartWorkers && !opts.ShouldRunWorkersInProcess`, mirroring `planChildren`), and `restartTargets` only offers `workers` when that flag is set. In the default in-process mode `WorkerMetricsPort` belongs to the `api` child, so `restart workers` now refuses (with the restartable list) instead of bouncing the API's group. New coverage: an in-process-mode subtest asserting `restart workers` terminates nothing and `restart` (all) skips the workers port.

tools/thuishaven/app/orchestrator.go
0xdeafcafe: Fixed in 6a3af9764. `Down` now attempts both drops (so route/registry cleanup and the stack removal always complete), aggregates any failures with `errors.Join`, and returns `stack %q stopped but database drop failed — state may be stale: …` after cleanup. `haven down --drop-db` can no longer report a clean reset while retaining the old DB. Added failing-ClickHouse and failing-Postgres drop cases to `TestDownDatabaseSemantics`.

langwatch/src/server/clickhouse/metrics.ts
0xdeafcafe: Addressed the in-repo half of the deployment contract in 6a3af9764. The SaaS worker terraform lives in `langwatch-saas/infrastructure` (separate repo) and still needs `CLICKHOUSE_BACKUP_METRICS_ENABLED=true` set on the worker deployment in the rollout that ships this image — that's tracked for the rollout PR.

What I could close here so the gate can't drift:

- The **langwatch Helm chart** now derives the env from the backup config: `sharedEnv` emits `CLICKHOUSE_BACKUP_METRICS_ENABLED=true` whenever `clickhouse.backup.enabled` is true (chart-managed backups) or `clickhouse.backup.metricsEnabled` is set (out-of-band backups). So any chart install that actually runs backups keeps emitting the gauges automatically — the signal is structurally coupled to whether backups exist, not a second flag someone can forget.
- **Drift guard**: `test_backup_metrics_gate` in `charts/langwatch/tests/e2e-overlays.sh` renders the workers deployment and asserts the env is absent by default, present under `backup.enabled`, and present under the `metricsEnabled` override. Spec scenarios added to `specs/ops/clickhouse-backup-metrics.feature`; `.env.example` documents the coupling.

The one thing that remains outside this repo is the SaaS terraform var; flagging that explicitly so it lands with the deploy rather than being silently missed.

Issue #4760, opened 2026-06-11
Harden GroupQueue blob reclamation with metrics, TTL review, and orphan cleanup


sergioestebance:
## Context

Follow-up to langwatch/langwatch#4757 and langwatch/langwatch#4758.

The dedup-squash fix stops the identified source of new GroupQueue blob orphans. Blob deletion remains intentionally best-effort, with expiry as the fallback. We should add operational visibility and defense-in-depth so future reclamation regressions are detected before Redis capacity is affected.

This issue intentionally excludes production volumes, locations, tenant identifiers, payload contents, and other sensitive operational details.

## Proposed work

- Add a counter for blob reclamation attempts, labeled by lifecycle path where cardinality remains bounded (for example: completion, retry, parse failure, dedup squash).
- Add counters for successful and failed blob deletions.
- Ensure deletion failures are observable without logging blob identifiers or payload-derived data.
- Add an alert or dashboard signal for sustained deletion failures and abnormal blob-key growth.
- Review `BLOB_TTL_SECONDS` against measured maximum staging, pause, and retry residence times; reduce it if evidence supports a shorter safety window.
- Define and execute a one-time cleanup procedure for pre-fix orphaned blob keys.
- Document validation and rollback/recovery steps for the cleanup procedure.

## Acceptance criteria

- Reclamation attempts and failures are measurable with bounded-cardinality metrics.
- Metrics and logs contain no blob IDs, payload data, tenant IDs, user IDs, or other PII/sensitive identifiers.
- A documented alert/dashboard can detect sustained reclamation failure or unexpected blob accumulation.
- The TTL decision is documented with supporting residence-time evidence.
- Existing orphan cleanup has a reviewed runbook, dry-run validation, rate limiting, and post-cleanup verification.

## Notes

Keep cleanup operationally separate from the queue correctness fix. Missing blobs are handled by the existing recovery path, but deletion should still be scoped conservatively and validated against live references before execution.

---

### Where this stands

| | |
| --- | --- |
| PR | `langwatch/langwatch-saas` #580 |
| Last touched | 2026-08-04 |
| Size | +210 / -0 across 1 files |
| Mergeable | MERGEABLE |
| Review | CHANGES_REQUESTED |
| Branch | `ops/reclaim-orphaned-blobs` |
| Worktree | `~/workspace/langwatch/langwatch-saas/.claude/worktrees/reclaim-blobs` |
| Session | _no transcript retained_ |

```bash
[...]

PR #5821, opened 2026-07-15
fix(group-queue): name, count and stop destroying discarded staged jobs

drewdrewthis:
> **Scariest thing** — this changes what happens to a blob after a failed decode. Get it backwards and you either **delete recoverable customer data** (release a body that was fine) or **leak blobs to the TTL backstop** (never release one that was gone). The rule is one line: release **only** when `reason === "missing_blob"`. Everything else preserves.
> **Start here** — `groupQueue.ts` → `dropStagedJob()` / `recordDrop()`. That pair is the whole change; the other four discard sites just call into it.
> **Also**: this PR fixes a **live production PII leak** it did not introduce — see finding 4 below.

## Why

The event-sourcing queue was **silently throwing away jobs** — `Failed to parse staged job data`, 100+ times in ~31h across 12 reactors. Every discarding path justified itself with one comment:

```ts
// Missing blob or unparseable value: complete the slot so it's not stuck;
// recover via event replay.
```

**That comment is false for reactors, and this issue's whole priority turns on it.** The replay service rebuilds fold projections and **never invokes reactors** (`projections/projectionRouter.ts:61-71`, stated in source). `governanceOcsfEventsSync` (OCSF security/audit) and `gatewayBudgetSync` (billing) are registered via `builder.withReactor("traceSummary", …)` (`pipelines/trace-processing/pipeline.ts:266-289`) — on a **fold**, where replay provably never reaches them. Replay's only two reactor references (`replayMarkers.ts:117`, `replayMapPath.ts:408`) exist to *suppress* re-fires. So a dropped audit or billing event is **permanently lost**.

Scoped honestly: **fold/map drops are replay-covered** (`replay/replayService.ts:74,105` drives `mapProjections`). The falsity is reactor-specific — that half alone carries the P1.

Don't confuse this with idempotency: both reactors cite a ReplacingMergeTree key as their "idempotency", which makes a *second* firing harmless — but **nothing produces a second firing**.

Escalated **P2 → P1** per the issue's own pre-authorisation. Closes #5538.

## 🔴 What this PR does NOT fix — read this before approving

**Reactor drops are still permanently lost. AC2's durability goal is NOT discharged.** This PR makes every loss **named, counted, and non-destructive**. It does not make a `missing_blob` drop recoverable — that body is genuinely gone, and no retry, park, or replay resurrects it. That loss is irreducible at this layer.

The real fix needs a re-fire key + a driver, and **cannot be built today** because of a separate defect this PR uncovered: `generateStagedJobId` (`:1556-1565`) does `baseId = (p.id as string) ?? crypto.randomUUID()`. Fold/map stage the bare `Event`, so the event id lives in the stagedJobId and survives blob loss. **Reactor jobs stage `{event, foldState}` — no top-level `.id` → `crypto.randomUUID()`.** So for precisely the jobs with no replay path, the event id exists **only inside the lost blob**. The original event is still in ClickHouse `event_log`: the data exists and is permanently unreachable because **nothing names it**. You cannot drive what you cannot name.

Tracked: **#5816** (re-fire key + driver, P1) · **#5817** (job-scoped dead-letter) · **#5818** (blob eviction / former AC6) · **#5819** (decision record) · **#5820** (blast radius of already-lost events, P1) · **#5823** (route `body_unreadable` through the retry ladder).

**And a second half-measure, surfaced by design review and worth your scrutiny:** this PR preserves an unreadable body on the grounds that *"the next worker could read it fine"* — but `complete()` removes the job from the group, so **no next worker ever sees it**. The queue already owns the mechanism that would actually deliver it (`retryRestage`, the ladder `handleTransientDecode` rides 40 lines away, whose own config doc says it exists to *"ride out a rolling restart… without parking the group"* — the codec-skew case exactly). Deferred to **#5823** because it changes delivery behaviour for every unreadable body; the docblock now names the option instead of arguing a false binary.

## What changed

- **Release only when the body is provably gone** → *alternative: keep releasing unconditionally (status quo)* → the drop path's second call, `blobLifecycle.release()`, *"reclaim[s] the blob … **deleting an s3 object out-of-band**"*. It fired on **body-present** failures — i.e. on rolling-deploy codec skew, where the **next** worker could have decoded the body fine. `handleTransientDecode` already refuses to release for exactly this reason (`:1492`: *"releasing here would risk reclaiming the blob the re-stage still needs"*). **`release()` is this module's retire-forever signal, and the drop path was firing it on recoverable data.** Cost if wrong: a preserved blob lives to its TTL backstop instead of being reclaimed promptly — bounded, intentional, and the same trade `handleTransientDecode` already makes.
- **Complete the slot; do NOT park** → *alternative: reuse `parkPoisonGroup`, the adjacent mechanism* → parking calls `restageAndBlock`, which blocks the **whole group**. That is right for `oversized_payload` (value intact; a raised cap could process it later) and **wrong** here: a missing blob never comes back, so parking would freeze that aggregate **forever** on a job that can never succeed. At 100+/day, blocked groups accumulate — trading silent data loss for a frozen-pipeline availability incident.
- **A drop is no longer counted as a success** → *alternative: leave `COMPLETE_LUA` alone* → it ran `INCR statsKey` (the *same* counter a genuine success takes) and `DEL errorKey` under *"now that the job succeeded"*. A discard **inflated the success rate AND erased the group's prior error state**. `complete()` now takes `dropped: true`, which withholds both. Defaults false, so every existing call site keeps its exact behaviour.
- **Classify by error TYPE, label by header** → *alternative: grep the exception message* → zlib's wording is Node-version-dependent and not ours to own, so a message-matching alert breaks on a runtime upgrade. Reuses `readJobRoutingMeta` (*"Never throws"* — reads the header of a body we cannot decode) rather than inventing a reader.

## How it works

```
jobEnvelope.ts
├─ DecodeFailureError { reason }        NEW — extends the existing typed-error idiom
│                                       (TransientBlobStoreError → retry,
│                                        PayloadTooLargeError → park) rather than
│                                        inventing a third mechanism
├─ DecodeFailureReason                  missing_blob | malformed_envelope | body_unreadable
├─ decodeBody()                         NEW — inflate+parse, names the failure.
│                                       PayloadTooLargeError passes through UNTOUCHED
│                                       (it is the park signal)
└─ readEnvelopeDescriptor()             NEW — {e, v, blobId}, never throws.
                                        Sibling of readJobRoutingMeta; the header
                                        survives what the body does not

groupQueue.ts
├─ recordDrop()                         NEW — log + counter. The complete ledger of
│                                       everything this queue throws away
├─ dropStagedJob()                      NEW — recordDrop + complete({dropped:true})
│                                       + release ONLY if reason === "missing_blob"
└─ 5 discard sites  ─────────────────►  all route through the pair above

scripts.ts
└─ COMPLETE_LUA  ARGV[7] dropped        skips INCR statsKey / per-job INCR / DEL errorKey

metrics.ts
└─ gqJobsDroppedTotal                   {queue_name, pipeline_name, job_type, job_name, reason}
```

**Why the full label set:** a `{queue_name, reason}` counter pages oncall with *"event-sourcing/jobs dropped 100"* and cannot say **which** pipeline lost **what**. The difference between a dropped UI broadcast and a dropped OCSF audit event is the difference between a shrug and a compliance incident.

## 🔎 Four things I found that are worse than the issue claims

Each of these contradicts the issue body, which was written 07-08 and is partly false. All re-verified line-by-line against the live tree.

1. **`release()` DESTROYS the body — and the brief I was given told me to reuse the mechanism that fires it.** The instruction was "reuse `handleTransientDecode` or `parkPoisonGroup`". But `handleTransientDecode` **pointedly does not release** (`:1492`), and following the reuse instruction naively would have fired `release()` on exactly the **body-present, recoverable** cases. **Not-releasing-body-present is the single highest-value line in this PR: it stops active destruction of recoverable data today.**
2. **FIVE drop sites, not one** — and **three of them emit `recoverable via replay` into PRODUCTION LOGS** (`:1300`, `:1337`, `:1520` pre-fix). Oncall reads a false reassurance and closes the ticket. **This is the false premise doing live damage, not doc rot.** (The AC review found four; the fifth — retry re-encode failure — surfaced while implementing.)
3. **A drop is counted as a SUCCESS**, not merely invisible. `COMPLETE_LUA` `INCR statsKey` + `DEL errorKey` means a discard **inflates the success counter AND erases the group's prior error state**. It was reported as "invisible to ops"; it is worse — it is counted as a win.
4. **🔒 A live production PII leak, which this PR did not introduce but does fix — and my first two attempts at the fix both leaked.** V8 quotes the offending input back at you:
   ```
   JSON.parse('patient@hospital.example is HIV positive')
     → Unexpected token 'p', "patient@ho"... is not valid JSON
   ```
   **Ten characters of raw body.** #5736 added `err` to the drop log while the bare-JSON path still threw a raw `SyntaxError` — so that fragment **has been landing in production logs since 2026-07-11**. `redactStorageUrisInText` only strips storage URIs; it does nothing for payload echo. A staged body can carry tenant PII, and AC1 says the drop log must **never** contain raw payload or tenant PII. Fixed by `safeParseErrText`, which strips the echo and keeps the diagnosis (`Unexpected token 'p', <redacted>... is not valid JSON`). zlib messages never echo input and pass through untouched.
   **Two rounds of my own fix leaked, both caught by falsifying rather than by reading:**
   - *v1 test* asserted the **whole secret** was absent — trivially true (V8 echoes only ~10 chars), so it **passed with the guard removed**. My own revert-check caught it.
   - *v2 guard* was a **blocklist** over V8's wording. A security review broke it: V8 only truncates at **21+ chars** and echoes the whole string below that, so `JSON.parse('x123456789')` → `"x123456789"` **untouched** — a 9-digit SSN, a 6-digit OTP, a card's last-4. msgpackr leaked differently again (a `data:[83,69]` byte array). Root cause: a blocklist over one library's message shapes is the same fragility `DecodeFailureReason` exists to avoid.
   - *v3* is an **allowlist**: keep the leading diagnosis, drop everything from the first `"`/`[`/`{`. Falsified — reverting to v2 fails 4 tests including `not to contain '123456789'`.
   Also closed from the same review: `splitEnvelope`'s header parse had **no guard**, leaking `header.m.__context` (traceId/userId) on a corrupt header.

## ⚠️ The false premise is load-bearing in SEVEN places. This PR kills three; four survive outside it.

A premise false in six places and corrected in none **will re-justify this exact drop for the next reader** — that is how it survived since the drop was written. Correcting the code while leaving the ADR and the migration asserting the opposite ships the bug's own justification forward.

| Site | Status |
|---|---|
| `groupQueue.ts` — 5 discard sites | ✅ **fixed here** |
| `metrics.ts:220` — `gqRetryEncodeFailuresTotal` help: *"work recovers via event replay"* | ✅ **fixed here** |
| `tieredBlobStore.ts:335` — *"a missing blob → null → decode fail-safe (**recover via replay**)"* | ✅ **fixed here** — found in review. This is the file that **creates** the condition: that `null` is what becomes `reason="missing_blob"`. |
| `jobEnvelope.ts:19-30` — module doc: *"drops the job **to replay**"* | ⚠️ describes the bug accurately; left as-is (it is a warning, not a justification) |
| `dev/docs/adr/030-...md:30` — **status: Proposed** | ❌ **#5819** |
| `clickhouse/migrations/00026_...sql:35-38` — *"can be dropped + rebuilt at any time from `event_log` without data loss"* | ❌ **#5819** — and this one is **dangerous**: acting on it destroys SOC2/ISO27001 audit evidence |
| **#4760** — proposes **reducing** `BLOB_TTL_SECONDS`, justified by *"Missing blobs are handled by the existing recovery path"* | ❌ **blocked on that bullet** — executing it would **increase this P1's loss rate**, and it would shrink the very TTL backstop this PR's preservation relies on |

## Review responses (fixed, not deferred)

An adversarial + principles + hygiene fan-out ran against this diff. Every Must-Fix is applied; I verified each against the code before accepting it. The two that matter, because **both were the defect this PR exists to remove — a claim that isn't true — reintroduced by my own patch**:

- **Orphaned doc-comments.** My insertions split two doc-comments from their declarations, leaving `dropStagedJob` headed by `parkPoisonGroup`'s doc — which describes *"move the group to the blocked set"*, the one thing `dropStagedJob` spends a paragraph explaining it deliberately does **not** do. Both moved back.
- **`bodyPreserved` lied.** `recordDrop` derived it from `reason`, but the retry-encode site **releases the body on the very next line** — so the log asserted `bodyPreserved: true` one line before destroying it. The rule is now named once (`const bodyIsGone`) and both the log and the release read off that single predicate; `recordDrop` *takes* the fact rather than guessing it.

Also applied: `readBody` → `decodeBody` (it was the only `read*` in a file where every `read*` contractually never throws); `EnvelopeDescriptor {e,v}` → `{format,version}` (the fields needing doc-comments were exactly the ones whose names failed); `decompress_failure` → **`body_unreadable`** (it fired on an inline never-compressed body — the name was why its doc-comment had to exist; free to rename today, never cheaper); a dead `PayloadTooLargeError` guard removed; and a **seventh** false-premise site found in `tieredBlobStore.ts`, the file that *creates* the condition.

## Rollout / rollback

**The Lua change is rolling-deploy safe, and you do not have to take my word for it.** `CachedLuaScript` derives the `evalsha` key from the source (`cachedLuaScript.ts:26-29`), and its own doc states the property: *"The sha is derived from the source, so a deploy that changes a script can never execute a stale cached body — a different source is a different sha."* So mid-deploy:

| | script | behaviour |
|---|---|---|
| old workers | old `COMPLETE_LUA` (own sha, already cached) | 6 ARGV, unconditional `INCR`/`DEL` — **exactly today's behaviour** |
| new workers | new `COMPLETE_LUA` (new sha → `NOSCRIPT` → one `EVAL` to load) | 7 ARGV, `dropped` honoured |

They coexist. No ordering requirement, no migration, no flag. **Rollback is a plain code revert** — no schema, no persistence surface, no data to undo (this PR deliberately adds none; that was the rejected Option B). The one behavioural residue of a revert: body-present values would start being `release()`d again, i.e. back to destroying recoverable bodies.

**The one thing to watch after deploy:** `gq_jobs_dropped_total` will be **non-zero immediately** — that is the point (the drops were always happening; they were silent). A `reason="unknown"` rate above zero is the signal that matters: it means a decode failure mode exists that the enum does not name.

proof

**`proven`** — **AC2(a): the old code destroys recoverable data. This is the whole point of the PR.** I reverted *only* the release rule (`if (reason === "missing_blob")` → unconditional `release()`), leaving the metric intact so a failure is **behavioural, not a missing import**. Exactly the two body-preservation tests fail, for exactly the right reasons, against **real Redis + testcontainers**:

```
$ pnpm test:integration run groupQueue.decodeDrop.integration.test.ts   # release rule reverted
  × does not release the blob
      AssertionError: expected [ Array(1) ] to deeply equal []
      ^ objectStore.deleted has 1 entry — the recoverable body WAS DELETED
  × leaves the body readable from the blob store afterwards
      AssertionError: expected false to be true
      ^ store.has(uri) === false — the body is GONE
  Tests  2 failed | 9 passed (11)     exit=1
```
The other 9 are untouched by the release rule. With the fix in place: **11/11 pass, exit 0.**

**`proven`** — **Unit falsifiability.** `jobEnvelope.ts` reverted to `origin/main` → **17/17 fail, exit 1**; restored → **17/17 pass, exit 0**.

**`proven`** — **Integration suite green:** 11/11, exit 0. **Control:** the pre-existing `groupQueue.poisonGuard.integration.test.ts` passes 10/10 on the identical harness, so the environment is sound.

**`proven`** — **Full unit regression (AC4 / AC5a):** 12 files, **128/128**, exit 0 — the typed-error conversion changed no decode path that already worked.

**`proven`** — **AC7: the false premise is gone from every discarding branch.**

| | mentions | of which **emitted to prod logs** |
|---|---|---|
| `origin/main` | 6 (`:803 :1069 :1259 :1300 :1337 :1520`) | **3** (`:1300 :1337 :1520`) |
| this branch | 2 (`:1285 :1481`) | **0** |

Both remaining mentions are doc-comments that **deny** the premise.

**`proven`** — **Gates:** `pnpm typecheck` 0 · `pnpm typecheck:tests` 0 · `pnpm check:feature-parity` 0 (*2491 enforced scenarios bound across 739 files*). (First typecheck attempt silently did nothing — `node_modules` missing in this fresh worktree — so it exited 1 without checking. Installed and re-ran; the 0s above are real.)

**`proven`** — **The `dropped` Lua flag cannot touch the success path.** Only one call site reaches `completeScript.run` (my TS wrapper); `CachedLuaScript.run` is variadic so a 7th ARGV is additive; `"" == "1"` is false in Lua. `complete()` without `dropped` keeps `INCR statsKey` + `DEL errorKey` exactly as before — and the 126 green unit tests plus 10/10 poisonGuard exercise that path.

#### Honest weighting of the 11 integration tests

Not all are revert-falsifiers, and the count shouldn't imply otherwise:
- **Genuine falsifiers:** body-present-does-not-release; body-still-readable; the counter + labels; both AC8 halves; transient-exhausted.
- **Regression guards, NOT falsifiers** (pass even reverted, because the old code did this too): `missing_blob` DOES release; the two liveness tests; the two transient-retry tests. Real value — different value.

#### `missing` — still not covered by an executing test

- **AC1's last inch:** the descriptor *reader* is proven (incl. that it survives blob loss and never throws), but nothing asserts `recordDrop` puts those fields on the emitted log record — needs a logger-capture harness. Marked `@unimplemented` in the spec **with that reason**, tracked by #5817.
- **AC7 site coverage is 3 of 5** — dispatch decode, transient exhaustion, and retry re-encode. That last one had **zero** coverage *and* no `@unimplemented` disclosure until a test review counted them; rather than just correct the number I wrote the test (`PutFailsAfterObjectStore(1)` so the stage succeeds and the retry's re-encode fails), falsified by reverting the site → `expected [] to have a length of 1`. Still uncovered: **sibling-drain decode** and **sibling re-stage**, which need fault injection on the coalesced-batch harness that already exists in `groupQueue.integration.test.ts`. The spec's Examples table names all five with their true coverage, so the next reader counts from the spec rather than from me.

1. `git show origin/main:./langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts | grep -nE 'recover(able)?( via| through)? [a-z ]*replay'` → 6 hits, 3 of them log-message strings.
2. Same grep on this branch → 2 hits, both comments denying the premise, 0 log messages.
3. Revert `jobEnvelope.ts` to `origin/main`, run `npx vitest run src/server/event-sourcing/queues/groupQueue/__tests__/jobEnvelope.decodeFailure.unit.test.ts` → **15 failures**. Restore → 15 pass.
4. Read `dropStagedJob()` in `groupQueue.ts`: confirm `release()` is inside `if (reason === "missing_blob")` and nowhere else on that path.
5. **The claim to attack:** that replay never re-fires reactors. Read `projections/projectionRouter.ts:61-71`, then `grep -rn "reactor" langwatch/src/server/event-sourcing/replay/` — the only two hits *suppress* re-fires. Then confirm `governanceOcsfEventsSync`/`gatewayBudgetSync` are `withReactor("traceSummary", …)` at `pipelines/trace-processing/pipeline.ts:266-289`. If that chain holds, the P1 holds.

## ⚠️ CI is green, but one shard lied — read this before trusting the badge

`gh pr checks` shows **56 pass / 0 fail**. That green is **partly forged**, and not by this PR:

```
$ ~/.claude/scripts/verify-ci-shard-tally.sh langwatch/langwatch 5821
  ✗ test-unit (2) — NO TALLY: shard did not finish (hard-floor fired). Its green is NOT evidence.
      ❯ src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts (2 tests | 2 failed)
  ✓ the other 9 shards — real tallies (test-unit 315/316/311 passed; test-integration 84-89 passed)
```

`langwatch/src/test-unit-global-setup.ts` arms a hard-floor `setTimeout(() => process.exit(0))` under `CI` to release a known vitest finalize wedge. A shard still running when it fires **exits 0 with failing tests and prints no tally**, and GitHub scores it `success`.

**That masked failure is INHERITED from `main`, not introduced here** — three independent lines of evidence:
1. **`main`'s own "successful" run carries it.** Run [`29406787744`](https://github.com/langwatch/langwatch/actions/runs/29406787744) @ `1128b4e`, conclusion `success`:
   ```
   10:07:06  ❯ src/server/api/routers/__tests__/team.update.multi-binding.unit.test.ts (2 tests | 2 failed)
   10:09:25  [unit globalSetup] hard floor reached at 4 min — forcing process.exit(0) …
   ```
2. **Already reproduced on a pristine `origin/main` worktree** by the 2026-07-13 investigation, which names this exact file and count. Tracked at **#4450**.
3. **It passes on this branch locally** (`2 passed (2)`, exit 0), and imports nothing this PR touches — zero `groupQueue`/`event-sourcing` imports, so there is no path by which this change could reach it.

Nothing in this PR's own surface is masked: `groupQueue`'s suites are **131 unit + 12 integration**, all with real tallies, plus the revert-checks below.

## Anything surprising?

- **`jobEnvelope.ts:19-30` independently documented this exact bug before I got here**: *"`decodeJobEnvelope` throwing a plain Error … terminates at the non-retryable fail-safe, which completes the slot and drops the job to replay. **An old worker meeting a new body during a rolling deploy would silently discard it.**"* The module's authors knew. `:778`/`:1276`/`:1285` also say *"do NOT drop to replay"*. The knowledge was there and didn't propagate to the drop path.
- **The transient-exhaustion terminal changed behaviour**, deliberately. It used to `release()` at attempt 8. Eight failed **reads** mean the **store** is unreachable, not that the blob is gone — most likely it is still there, so retiring it would delete a body that exists. It now preserves like every other non-`missing_blob` reason. AC5(a) (the retry ladder) is untouched.
- **Collisions, both live.** #5766 (updated 07-15) rewrites `processWithRetries` — adjacent, does not touch the drop branch. #5664 touches `tieredBlobStore.fetch()` and was already conflicting with main independently of me. Rebased onto `1128b4e7d`; the 3 commits main gained during this work touch none of my files. **Neither PR removes this trigger** — #5664 closes *one* mechanism that makes a blob "already gone"; any other cause hits the identical drop.
- **I could not read the prod exception text.** The issue's own stated prerequisite — rank hypothesis #1 (missing blob) vs #2 (format skew) vs #3 (gzip corruption) from the `err` field shipped in #5736 — needs CloudWatch on `langwatch-workers` (logger `langwatch:event-sourcing:group-queue`). This box's IAM role has no `logs:*` (`AccessDeniedException`, verified). **So I have not verified the storm is still firing** — the evidence window is 07-07→08 and nobody has read the discriminating data that has been flowing since 07-11. AC2/AC3 are code defects that stand regardless of which error dominates, which is why this PR proceeds; **#5818** carries the conditional work.
- **#5594 cuts against the obvious reading of hypothesis #1**: GQ2 redis blobs were on GQ1's **7-day** TTL through the entire evidence window — blobs lived **longer** than designed, not shorter. Don't reach for "TTL too short" without evidence.

---
<sub>🤖 **Authored by `lw#5538 groupqueue-silent-drop (worker session)` via the write-pr procedure, on behalf of the fleet.** Opened under the owner's GitHub token for API access — the code and this description are the fleet's, not his. Owner-authored and fleet-authored PRs are otherwise indistinguishable in this repo; this footer is the only signal. **The owner merges — this session does not.**</sub>


sergioestebance:
## Ruthless review — verdict: **no P0/P1. Ship-worthy once the P2 is closed.**

Reviewed the full `main...HEAD` diff (7 files) and traced the drop paths, the Lua `dropped` flag, and the integration test. This is a faithful, well-reasoned Option A: `release()` is now conditional on `missing_blob`, the `dropped` flag withholds the completed-counter INCR **and** the error-key DEL, the false "recover via event replay" premise is removed from all four sites (incl. `tieredBlobStore.ts`), the transient path is preserved, and the 639-line integration test backs the durability behaviors (release-not-called + body-still-readable, missing-blob-releases, full label set, group-stays-live, `stats:completed` unchanged, error survives, `transient_exhausted` counted). The exhaustion test seeds `/r/` markers instead of waiting out ~2h of backoff — clean.

Findings below, severity-ordered. None block merge except closing the P2's evidence gap.

---

### 🟡 P2 — AC1's drop-**log** descriptor is emitted but never asserted (wiring is untestable-by-deletion)

`groupQueue.ts` `recordDrop()` logs `envelopeFormat: descriptor.e`, `envelopeVersion: descriptor.v`, `blobId: descriptor.blobId`, `bodyPreserved`. The unit test proves `readEnvelopeDescriptor()` returns `e`/`v`/`blobId` **in isolation**, and the integration test asserts the **counter** labels — but nothing asserts the emitted **log record** carries those fields. Grep of `groupQueue.decodeDrop.integration.test.ts` for `envelopeFormat|envelopeVersion|blobId|bodyPreserved|logger` returns zero hits.

**Why it matters:** AC1's own evidence shape is *"asserting the captured log object contains `e`, `v`, and the blob hash; quoted captured log object."* As written, deleting the four `envelopeFormat:/envelopeVersion:/blobId:/bodyPreserved:` lines from `recordDrop` would leave every test green — the highest-value diagnostic (the thing an operator greps to find a preserved blob) has no regression guard.

**Fix:** add one assertion — a logger spy in the integration test (or a focused unit on `recordDrop`) forcing a `missing_blob` drop and asserting the captured log object contains `envelopeFormat`, `envelopeVersion`, `blobId`, and `bodyPreserved: false`; and a `malformed_envelope` drop asserting `bodyPreserved: true`. That closes AC1's evidence and pins the descriptor→log wiring.

---

### 🟢 P3 — the AC7 evidence grep is no longer clean (returns two comment hits inside discarding methods)

The feature file defines AC7's proof as `grep -nE "recover(able)?( via| through)? [a-z ]*replay" …groupQueue.ts` *"returning only non-discarding contexts."* On the branch it now returns **two** hits:
- `:1284` — `parseDrainedPayload` JSDoc: *"This used to say \"recoverable via event replay\". It is not…"*
- `:1480` — `dropStagedJob` JSDoc: *"…justified itself with \"recover via event replay\". It does not."*

Both are negations, so the **intent** holds (no site *claims* recovery), but both live inside the discarding methods, so the mechanical evidence command yields false positives — anyone re-running the stated AC7 check has to eyeball that they're negations. Either reword the two comments so they don't contain the literal phrase (e.g. "asserted replay would recover it"), or update the AC7 evidence note to exclude negating comment lines. Cheap; keeps the self-defined proof honest.

---

### 🟢 P3 — "a later worker can still read it" slightly oversells the preservation (comment vs. behavior)

For body-present drops, `dropStagedJob` calls `complete({ dropped: true })` — the staged job is **removed from staging**; only the blob is preserved to its TTL. There is no automatic re-stage (the re-fire driver is a deferred follow-up, correctly). So the comments *"the next worker could read it fine"* / *"a later worker (post-rollout) can still read it"* (`groupQueue.ts:~1533`, `jobEnvelope.ts` `readBody` doc) describe an automatic recovery the current code does not perform — recovery is operator/future-driver-only, via the logged `blobId`. Per the house rule on comments matching behavior, tighten to *"the body survives to its TTL backstop so an operator (or the follow-up driver) can recover it via the logged blob id"* — the preservation is real and worth keeping; only the "later worker" framing implies automation that isn't here yet.

---

### Notes (not findings)
- Bare-JSON parse failure is classified `decompress_failure` (`parseInlineBody`) even though nothing was compressed. Defensible — the enum documents `decompress_failure` as the body-present-unreadable catch-all and there's no inline-parse reason — but if a reviewer expects `malformed_*`, worth a one-line doc nod at the `parseInlineBody` site.
- `transient_exhausted` no longer `release()`s (only `missing_blob` does). Correct and documented (8 failed *reads* ⇒ store down, not blob gone); bounded by the ~4-day TTL, consistent with `handleTransientDecode`'s existing retained-hold philosophy.
- Lua `dropped` flag: all success callers omit it → `ARGV[7] = ""` → `dropped=false`, exact prior behavior. No arg-count breakage.

---

**Blessing withheld deliberately** — this is a review-only pass, so I'm commenting rather than approving (and the PR is still draft). Once the P2 log assertion lands and it's marked ready, there's no P0/P1 standing between this and merge.

drewdrewthis:
## Review verdict: READY

Five reviewers ran and returned; every concern below is attributed. **All Must-Fix and the one dealbreaker are applied and verified** — none deferred silently. Two of the findings were my own patch committing the exact defect this PR exists to remove, which is the most useful thing the fan-out did.

### Applied (was blocking)

| # | Finding | Action |
|---|---|---|
| 1 | **[principles]** Insertions **orphaned two doc-comments** — `dropStagedJob` was headed by `parkPoisonGroup`'s doc describing *"move the group to the blocked set"*, the one thing it deliberately does not do. A comment describing behaviour the code doesn't implement, in the PR about exactly that. | **Fixed** — both moved back onto what they document. |
| 2 | **[principles]** **`bodyPreserved` lied** — derived from `reason`, but the retry-encode site releases the body one line later, so the log asserted `bodyPreserved: true` immediately before destroying it. | **Fixed** — rule named once (`bodyIsGone`); log and release read one predicate; `recordDrop` takes the fact instead of guessing. |
| 3 | **[security]** **Redaction bypass #1** — V8 only truncates at ~21+ chars and echoes the whole string below that; my `{12,}` threshold missed 1–11, so `JSON.parse('x123456789')` leaked a 9-digit SSN verbatim. | **Fixed** — blocklist → **allowlist**. Falsified: reverting fails 4 tests incl. `not to contain '123456789'`. |
| 4 | **[security]** **Redaction bypass #2** — msgpackr embeds leftover bytes as `data:[83,69]`, matching neither regex and decoding straight back to plaintext. | **Fixed** — same allowlist cuts at `[`/`{`. |
| 5 | **[security]** **`splitEnvelope`'s header parse had no guard** — a corrupt header echoed `header.m.__context` (traceId/userId) via the raw-Error path. | **Fixed** — wrapped + redacted; a corrupt header is now a named `malformed_envelope`, not `unknown`. |
| 6 | **[test]** **`retry_encode_failed` had zero coverage and no `@unimplemented`** — and this PR body claimed "3 of 5 sites covered" when it was 2. | **Fixed** — wrote the test rather than only correcting the count. Falsified: reverting → `expected [] to have a length of 1`. Genuinely 3 of 5 now. |
| 7 | **[test]** The transient-exhausted terminal's **other** deliberate change (no longer releases) had no assertion. | **Fixed** — `expect(flaky.deleted).toEqual([])`. |
| 8 | **[design-soundness]** **Dealbreaker: the docblock argued a false binary** (`complete()` vs `park`), omitting `retryRestage` — the ladder 40 lines away whose own config doc says it exists to *"ride out a rolling restart without parking the group"*, i.e. the codec-skew case. Preserve-but-don't-retry is incoherent: `complete()` removes the job, so the "next worker could read it" rationale never happens. | **Fixed + deferred honestly** — docblock names the third option and says why it waits; routing change filed as **#5823**. |
| 9 | **[design-soundness]** **Factual error**: comment said *"Eight failed READS"*; `maxAttempts` is **25** (~2h27m). Written *after* I was corrected on that number, inside the commit "stop the drop log claiming things that are not true." | **Fixed** — states the config, not an invented number. |
| 10 | **[hygiene]** Integration test was **untracked** — the only thing covering AC2(a)/AC8 would not have shipped. **[hygiene]** A **seventh** false-premise site in `tieredBlobStore.ts`, the file that *creates* the condition. | **Fixed** — committed; site corrected. |
| 11 | **[principles][hygiene]** `readBody` → `decodeBody` (only `read*` in a file where every `read*` never throws); `EnvelopeDescriptor {e,v}` → `{format,version}`; `decompress_failure` → **`body_unreadable`** (fired on an inline never-compressed body); dead `PayloadTooLargeError` guard removed. | **Fixed.** |

### Non-blocking (prose only, not gating)

- **[design-soundness]** `gq_retry_encode_failures_total` is now strictly derivable from `gq_jobs_dropped_total{reason="retry_encode_failed"}`. Kept for dashboard continuity — stated as that, not dressed up as a distinct signal. → **Decide** (deprecate with a date?).
- **[design-soundness]** `recordDrop` double-parses (`readJobRoutingMeta` + `readEnvelopeDescriptor`); the module has a fused-reader idiom (`readEnvelopeRetirement`). Cold path. → **New Issue**.
- **[design-soundness][synthesizer]** "Log as index + TTL as retention" is a half-measure with no reader — subsumed by **#5823**.
- **[security]** Preserved-but-unreadable blobs outlive their job to the ~4-day TTL. If a customer-erasure flow purges by *decoded* record rather than storage key, they could outlive an erasure request in that window. → **New Issue**.
- **[test]** AC4's `@unimplemented` is the cheapest of the three to close; AC7's overstates the gap (a coalesced-batch harness already exists). AC1's is verified legitimate — `logger` is private and non-injectable.
- **[security]** Confirmed clean: metric labels are **not** tenant-controlled (registration-time constants); `blobId` is a UUID/SHA-256; ARGV binding is protocol-level, no injection; no secrets or real PII in tests, spec, or the proof image.

### Evidence

Every claim has a revert-check I ran myself: unit **17/17 red → green**; **AC2(a) precise falsifier** (revert only the release rule → exactly the 2 body-preservation tests fail: `expected [ Array(1) ] to deeply equal []`, `expected false to be true`); PII allowlist (revert → 4 fail); retry-encode (revert → counter never fires). **131 unit + 12 integration**, control suite 10/10, typecheck + typecheck:tests + feature-parity all 0.

### Not covered — read before approving

**AC2's durability goal is undischarged: reactor drops are still permanently lost.** Reactor jobs stage `{event, foldState}` with no top-level `.id`, so the event id exists only inside the lost blob — you cannot drive what you cannot name (**#5816**). **Never verified:** whether the storm still fires; no prod log access.

drewdrewthis:
## CodeRabbit triage — 9 comments, all answered

**Thanks — one of these caught a genuinely vacuous assertion my own falsifier missed.** Dispositions, each verified against the code before acting:

### Fixed

| Comment | What I found |
|---|---|
| **Acquire a holder before asserting transient-exhaustion retention** | **Correct, and this was the important one.** The test staged via `encodeJobEnvelope` + `stageBatch`, bypassing the `blobLifecycle.acquire()` that `send()` performs — so with no holder, `release()` deleted nothing and `expect(flaky.deleted).toEqual([])` passed **whether or not the fix was present**. I proved it: reverting the release rule left the test green. My other release assertions had teeth only because they stage through `send()`; I read the aggregate red-on-revert as proof and missed that one member was vacuous. Now acquires the holder and pins it exists. Verified: fix in → green; release reverted → **RED** (`expected [ Array(1) ] to deeply equal []`). |
| **Use object-destructured parameters** | Correct — and not just the blanket rule: the module's own `TransientBlobStoreError` uses `constructor({ projectId, hash, cause })`. `DecodeFailureError` was the odd one out. All 12 call sites converted. |
| **Validate blob IDs before exposing them to drop logs** | Correct and subtle. `readEnvelopeDescriptor` runs on envelopes *already known to be malformed*, so `header.r` / `header.ref.hash` are attacker-shaped strings by then and go straight to a log. Now shape-checked. **My first guard was hex-only and nulled every legitimate GQ2 id** — the id is `sha256(bytes).subarray(0,16).toString("base64url")` (22 chars of `[A-Za-z0-9_-]`), not hex; I guessed instead of reading `tieredBlobStore.ts:83-87`, and the unit tests caught it. Verified both directions: real GQ1+GQ2 ids kept; JSON, emails, spaces, and path-traversal probes rejected. |
| **Apply the required boolean prefixes** | Correct — `scripts.ts` already uses `hasPauses` (×2). `dropped` → `isDropped`. |

### Already fixed before your pass (stale)

- **"Preservation does not automatically route the body to another worker."** Right, and it's the sharpest point anyone made about the design. `complete()` removes the job, so the "next worker could read it" rationale never fires — preserve-but-don't-retry is a half-measure. The docblock now says exactly that and defers the retry-ladder routing to **#5823**.
- **"Update AC7 from four discard sites to five."** Done — the Examples table lists all five with their true coverage.

### Not fixing, with reasons

- **"Redact or hash raw group and staged-job identifiers."** Pre-existing log shape, **unchanged by this PR** (that field set predates it). A security review traced them: they carry a tenant segment (`tenantIdFromGroupId`) and an opaque `traceId` — internal identifiers, not free-text PII — and `projectId` is already logged in cleartext one field over. Changing the established log shape belongs in its own change, not smuggled into a P1 fix.
- **"Reject header lengths that exceed the envelope bytes."** Already handled: `subarray` clamps, so an oversized `headerLen` yields a truncated slice that fails to parse — and this PR's new guard names that `malformed_envelope` instead of letting it escape as `unknown`. An explicit bound would be belt-and-braces on a path that already terminates correctly.

All fixes verified: **131 unit + 12 integration green**, typecheck + typecheck:tests 0, each with a revert-check.

drewdrewthis:
## Review verdict: READY

Five reviewers ran and returned; every concern below is attributed. **All Must-Fix and the one dealbreaker are applied and verified** — none deferred silently. Two of the findings were my own patch committing the exact defect this PR exists to remove, which is the most useful thing the fan-out did.

### Applied (was blocking)

| # | Finding | Action |
|---|---|---|
| 1 | **[principles]** Insertions **orphaned two doc-comments** — `dropStagedJob` was headed by `parkPoisonGroup`'s doc describing *"move the group to the blocked set"*, the one thing it deliberately does not do. A comment describing behaviour the code doesn't implement, in the PR about exactly that. | **Fixed** — both moved back onto what they document. |
| 2 | **[principles]** **`bodyPreserved` lied** — derived from `reason`, but the retry-encode site releases the body one line later, so the log asserted `bodyPreserved: true` immediately before destroying it. | **Fixed** — rule named once (`bodyIsGone`); log and release read one predicate; `recordDrop` takes the fact instead of guessing. |
| 3 | **[security]** **Redaction bypass #1** — V8 only truncates at ~21+ chars and echoes the whole string below that; my `{12,}` threshold missed 1–11, so `JSON.parse('x123456789')` leaked a 9-digit SSN verbatim. | **Fixed** — blocklist → **allowlist**. Falsified: reverting fails 4 tests incl. `not to contain '123456789'`. |
| 4 | **[security]** **Redaction bypass #2** — msgpackr embeds leftover bytes as `data:[83,69]`, matching neither regex and decoding straight back to plaintext. | **Fixed** — same allowlist cuts at `[`/`{`. |
| 5 | **[security]** **`splitEnvelope`'s header parse had no guard** — a corrupt header echoed `header.m.__context` (traceId/userId) via the raw-Error path. | **Fixed** — wrapped + redacted; a corrupt header is now a named `malformed_envelope`, not `unknown`. |
| 6 | **[test]** **`retry_encode_failed` had zero coverage and no `@unimplemented`** — and this PR body claimed "3 of 5 sites covered" when it was 2. | **Fixed** — wrote the test rather than only correcting the count. Falsified: reverting → `expected [] to have a length of 1`. Genuinely 3 of 5 now. |
| 7 | **[test]** The transient-exhausted terminal's **other** deliberate change (no longer releases) had no assertion. | **Fixed** — `expect(flaky.deleted).toEqual([])`. |
| 8 | **[design-soundness]** **Dealbreaker: the docblock argued a false binary** (`complete()` vs `park`), omitting `retryRestage` — the ladder 40 lines away whose own config doc says it exists to *"ride out a rolling restart without parking the group"*, i.e. the codec-skew case. Preserve-but-don't-retry is incoherent: `complete()` removes the job, so the "next worker could read it" rationale never happens. | **Fixed + deferred honestly** — docblock names the third option and says why it waits; routing change filed as **#5823**. |
| 9 | **[design-soundness]** **Factual error**: comment said *"Eight failed READS"*; `maxAttempts` is **25** (~2h27m). Written *after* I was corrected on that number, inside the commit "stop the drop log claiming things that are not true." | **Fixed** — states the config, not an invented number. |
| 10 | **[hygiene]** Integration test was **untracked** — the only thing covering AC2(a)/AC8 would not have shipped. **[hygiene]** A **seventh** false-premise site in `tieredBlobStore.ts`, the file that *creates* the condition. | **Fixed** — committed; site corrected. |
| 11 | **[principles][hygiene]** `readBody` → `decodeBody` (only `read*` in a file where every `read*` never throws); `EnvelopeDescriptor {e,v}` → `{format,version}`; `decompress_failure` → **`body_unreadable`** (fired on an inline never-compressed body); dead `PayloadTooLargeError` guard removed. | **Fixed.** |

### Non-blocking (prose only, not gating)

- **[design-soundness]** `gq_retry_encode_failures_total` is now strictly derivable from `gq_jobs_dropped_total{reason="retry_encode_failed"}`. Kept for dashboard continuity — stated as that, not dressed up as a distinct signal. → **Decide** (deprecate with a date?).
- **[design-soundness]** `recordDrop` double-parses (`readJobRoutingMeta` + `readEnvelopeDescriptor`); the module has a fused-reader idiom (`readEnvelopeRetirement`). Cold path. → **New Issue**.
- **[design-soundness][synthesizer]** "Log as index + TTL as retention" is a half-measure with no reader — subsumed by **#5823**.
- **[security]** Preserved-but-unreadable blobs outlive their job to the ~4-day TTL. If a customer-erasure flow purges by *decoded* record rather than storage key, they could outlive an erasure request in that window. → **New Issue**.
- **[test]** AC4's `@unimplemented` is the cheapest of the three to close; AC7's overstates the gap (a coalesced-batch harness already exists). AC1's is verified legitimate — `logger` is private and non-injectable.
- **[security]** Confirmed clean: metric labels are **not** tenant-controlled (registration-time constants); `blobId` is a UUID/SHA-256; ARGV binding is protocol-level, no injection; no secrets or real PII in tests, spec, or the proof image.

### Round 2 — CodeRabbit (9 comments, all answered)

**One caught a vacuity my own falsifier missed**, which is exactly what the fan-out is for:

| # | Finding | Action |
|---|---|---|
| 12 | **[coderabbit]** `expect(flaky.deleted).toEqual([])` in the transient-exhaustion test was **vacuous** — it staged via `stageBatch`, bypassing the `acquire()` that `send()` does, so with no holder `release()` deleted nothing and the assertion passed with the fix reverted. My other release assertions had teeth only because they stage through `send()`; I read the aggregate red-on-revert as proof and missed the one member with none. | **Fixed** — acquires the holder. Verified: reverted → **RED** (`expected [ Array(1) ] to deeply equal []`). |
| 13 | **[coderabbit]** `DecodeFailureError` took positional `(message, reason)` — against CLAUDE.md *and* the module's own `TransientBlobStoreError({projectId, hash, cause})`. | **Fixed** — 12 call sites. |
| 14 | **[coderabbit]** `readEnvelopeDescriptor` exposed `header.r`/`ref.hash` to the drop log unvalidated — and it runs on envelopes *already known malformed*, so those are attacker-shaped by then. | **Fixed** — shape-guarded. My first guard was hex-only and **nulled every real GQ2 id** (it's base64url, not hex — I guessed instead of reading `tieredBlobStore.ts:83-87`); the unit tests caught it. |
| 15 | **[coderabbit]** Lua `dropped` → `isDropped` (`scripts.ts` already uses `hasPauses`). | **Fixed.** |
| — | Redact group/staged-job ids; reject over-long header lengths; two already-fixed items | **Not fixed, reasons published** in the triage comment. |

### Evidence

Every claim has a revert-check I ran myself: unit **17/17 red → green**; **AC2(a) precise falsifier** (revert only the release rule → exactly the 2 body-preservation tests fail: `expected [ Array(1) ] to deeply equal []`, `expected false to be true`); PII allowlist (revert → 4 fail); retry-encode (revert → counter never fires). **131 unit + 12 integration**, control suite 10/10, typecheck + typecheck:tests + feature-parity all 0.

### Not covered — read before approving

**AC2's durability goal is undischarged: reactor drops are still permanently lost.** Reactor jobs stage `{event, foldState}` with no top-level `.id`, so the event id exists only inside the lost blob — you cannot drive what you cannot name (**#5816**). **Never verified:** whether the storm still fires; no prod log access.

sergioestebance:
## Ruthless review — round 2 (post-push). Verdict: **no P0/P1. One P2 to fix, then this is clean.**

Re-traced the full `main...HEAD` (now +1983/−98, 9 files) after the five new commits. The durability logic is sound and the honesty of this pass is a step up. What's addressed since round 1, then the one new finding.

### Addressed / credited
- **My round-1 P2 (AC1 drop-log descriptor untested)** — handled the right way: the AC1 scenario is now `@integration @unimplemented` with an explicit note that `readEnvelopeDescriptor` is unit-proven but the *last inch* (`recordDrop` actually putting `e`/`v`/`blobId`/`bodyPreserved` on the emitted record) is **not** asserted, and "do not read the green suite as covering this." Converting a hidden gap into a visible, marked one is the correct move for a deferred finding. (But see P2 below — the tracking pointer is dead.)
- **5th discard site** — you found and tested `retry_encode_failed`, which my round-1 review (anchored on the issue's "four sites") missed. The test correctly asserts it *releases* (body already read → keeping it buys nothing) and that reverting the drop counter makes `dropsFor()` time out. Good catch on your own scope.
- **Vacuous retention assertion (CodeRabbit)** — `expect(flaky.deleted).toEqual([])` proved nothing because the test staged via `encodeJobEnvelope + stageBatch`, bypassing the `acquire()` that `send()` does — no holder, so `release()` was always a no-op and the assertion passed with the fix reverted. Fixed by acquiring the holder like `send()` and pinning it exists first, verified both directions. This is exactly the false-negative-proof class a ruthless pass exists to kill; you killed it on yourself.

### 🟡 P2 — the deferral tracker `#5817` does not exist (dangling reference)

`specs/event-sourcing/groupqueue-decode-drop-durability.feature:61` and `:176` both defer real gaps to **"Tracked by #5817"**:
- the AC1 log-descriptor assertion (my round-1 P2), and
- the two still-uncovered AC7 sites (sibling-drain decode + sibling re-stage — 3 of 5 have executing tests; these two are grep-only).

`#5817` resolves to nothing — not an issue, not a PR, in either `langwatch/langwatch` or `langwatch-saas` (current max is #5830 / PR #5825, so it's in-range but was never created). **The disclosed deferrals are therefore untracked** — the follow-up work has no ticket and will be silently forgotten. This is the exact failure mode this PR exists to fix: a load-bearing pointer that doesn't resolve. Fix: create the follow-up issue (AC1 log-wiring assertion + the 2 sibling-site executing tests) and make both citations resolve to it, or correct the number.

### Residual gaps (disclosed, non-blocking — but they only stay honest if the P2 tracker is real)
- **AC1 log→record wiring**: untested. Deleting the four `envelopeFormat:/envelopeVersion:/blobId:/bodyPreserved:` lines in `recordDrop` still passes CI. Marked `@unimplemented`; acceptable *if* tracked.
- **AC7 sibling-drain decode + sibling re-stage**: no executing test; the "no replay claim" half rests on the grep. Marked in the Examples table; acceptable *if* tracked.
- **AC7 grep still returns two hits** (`groupQueue.ts:1287`, `:1483`) — both negating comments inside the discarding methods. Now disclosed in the spec ("proven by AC7's grep"), so intent is clear; the mechanical command just isn't literally empty. Fine as-is given the disclosure.

### CI
Not green yet — the test/analyze shards were still **pending** at review time (nothing red). Verdict is contingent on them landing green.

---

**No P0/P1.** The code is merge-worthy; the retention/`release()`/counter/`dropped`-flag logic all holds and the false premise is gone from all five sites. Blessing withheld only per review-only convention (comment, not approve) — and I'd want #5817 to actually exist first, since it's the thread the deferred honesty hangs from. Create that issue, let CI go green, and this ships.

sergioestebance:
## Round-2 reconciliation — after reading your dispositions

Went back through my two review comments against current HEAD and your disposition replies. Closing the loop so the thread is clean.

**Confirmed resolved / agreed — no further action:**
- **R1 P3 (bare-JSON misclassified as `decompress_failure`)** — fixed. You renamed the reason to `body_unreadable` for exactly the case I flagged ("fired on an inline never-compressed body"). Resolved.
- **R1 P2 (AC1 log→record wiring untested)** — agreed on both sides. Your note *"AC1's is verified legitimate — `logger` is private and non-injectable"* is the right read; `@unimplemented` is the honest disposition given no logger-capture harness. Fine to defer **provided the tracker resolves** (see below).
- Renames (`decodeBody`, `{format, version}`, dead `PayloadTooLargeError` guard, base64url blob-id guard) — all verified on HEAD. The `recordDrop` log fields (`envelopeFormat`/`envelopeVersion`/`blobId`/`bodyPreserved`) still emit correctly, so nothing I cited went stale.

**Still live — one item, unchanged since my round-2 comment:**
- 🟡 **`#5817` does not exist.** The feature file defers two disclosed gaps to *"Tracked by #5817"* (lines 61 and 176 — the AC1 log assertion and the two uncovered AC7 sibling sites), but `#5817` resolves to nothing in either repo (current max is #5830). The `@unimplemented` honesty is good, but the pointer is dead, so the deferred work is effectively untracked — the same dangling-reference failure mode this PR exists to kill. **Create the follow-up issue (or fix the number) so both citations resolve, and this is clean.**

No P0/P1. Code is merge-worthy; the only thing between here and a tidy paper trail is making #5817 real.

sergioestebance:
## Correction — retracting my `#5817` finding. It was a false positive; my mistake.

I claimed (round 2 + closeout) that "Tracked by #5817" was a dangling reference. **It is not, and I was wrong.** `#5817` was filed in this repo and then **transferred to the private tracker under the `moved-for-privacy` rule** — `langwatch/langwatch#5817` now redirects to **`langwatch-saas#719`** ("GroupQueue: job-scoped dead-letter for undecodable staged jobs"), open. The citation resolves.

**Why I got it wrong:** I checked with `gh issue view --repo langwatch/langwatch 5817`, which uses GraphQL — and GraphQL does **not** follow GitHub's transfer redirect, so it returns "Could not resolve." `gh api repos/langwatch/langwatch/issues/5817` (REST) *does* follow it and lands on saas#719. I trusted the wrong tool and didn't cross-check. On me.

So: **no P2. There is no dangling reference.** Everything else from my reviews stands — R1-P3 (`decompress_failure` → `body_unreadable`) you fixed, R1-P2 (AC1 log-record wiring) we agreed is a legit `@unimplemented` deferral, and the durability/`release()`/`dropped`-flag/five-sites logic is sound.

**No P0/P1, no P2. Clean once CI is green.**

One genuinely-minor observation, take it or leave it: the redirect target (saas#719) is scoped to the *dead-letter* follow-up, while the two `#5817` cites here are for the AC1 logger-capture harness and the AC7 sibling-site tests — a different concern. If those test-coverage gaps were meant to hang off their own follow-up rather than the dead-letter issue, the cite is slightly off-target. Not a blocker, and entirely your call — flagging only because it's the kind of cross-reference that's easy to drift.

sergioestebance (approved):

PR #5831, opened 2026-07-15
feat(governance): capture VS Code Copilot Chat via a scoped code() function (ADR-039 §Extension #2, stacked on #5784)

sergioestebance:
# Related Issue

- Resolves langwatch/langwatch-saas#890

Stacked on langwatch/langwatch#5784. Implements **ADR-039 §Extension langwatch/langwatch#2 (VS Code Copilot Chat)**, Accepted. Closes langwatch/langwatch-saas#890.

## What ships
VS Code Copilot Chat joins capture via the **existing scoped shell-function tier** — the same mechanism `copilot`/`gemini`/`opencode` use. `langwatch code` mints a `copilot_vscode` ingest key and persists a scoped `code()` function that sets the OTLP env only for `code` launches; the Chat extension reads the window env and posts `gen_ai.*` OTLP straight to `/api/otel`. No wrapper for capture itself; two guards ship with it: a NARROW, JSONC-safe settings.json hardening write (each telemetry key nulled under `terminal.integrated.env.<os>` via `jsonc-parser`, comment/format-preserving, applied on every run so integrated terminals never inherit the ingest token — the token itself is only ever delivered by env) and a receiver-side scope gate that drops non-Copilot instrumentation scopes posted on a `copilot_vscode` key (VS Code's debug console and extension children inherit the generic `OTEL_*` env).

Empirically validated on VS Code 1.128.1 / Copilot Chat 0.56.0: a real Chat turn (`panel/editAgent`, model `oswe-vscode-prime`) landed with prompt, tokens, and content; an env-only launch (empty settings.json) also captured, confirming the env overrides the default-false setting.

## Changes (all reuse, no new mechanism)
- `platform-tool-policy.ts` — `code`: ingestion-only (`allowVk:false, allowOtelDirect:true`, inverse of cursor)
- `wrapper-mode.ts` — `code → copilot_vscode` sourceType; `buildOtelEnvBlock` `code` case (`COPILOT_OTEL_ENABLED` + endpoint + Bearer + capture + `service.name=copilot-chat`); content-capture opt-out extended to `code`
- `shell-rc.ts` — `code` in `SHELL_FUNCTION_TOOLS` (scoped `code()` via `buildScopedToolFunction`)
- `commands/wrap.ts` + `index.ts` — hidden `langwatch code` command
- logout auto-covers the `code()` function via `scanTelemetryTargets`
- Receiver: **no code** — the GenAI core canonicalizes `copilot-chat` `gen_ai.*`; `copilot.ts` stays inert (proven by test)

## Tests
- 11 new unit tests across wrapper-mode, shell-rc, telemetry-targets (SDK) + copilot extractor (langwatch), all green; SDK suite 414 passing, feature-parity OK.
- Feature files: `copilot-vscode-env.feature` (11) + `copilot-vscode-otlp.feature` (7).

## Deferred (Roadmap / ADR Open Questions)
- Cost / AI-units (`copilot_usage_nano_aiu`, obfuscated codenames)
- Login-agent always-on for Dock/Spotlight launches
- Key-revoke-on-logout (no sourceType revokes today — same as the app extension)
- Cross-window capture: `code` args handed to an already-running VS Code instance reuse its extension host, which never sees this launch's env — the wrapper says so up front; always-on capture stays on the Roadmap

## Deployment Impact
Minimal deployment impact: client-side CLI feature (`langwatch code`) + a server policy row (`code`, allowVk structurally false), coding-agent source-type registration for the three copilot surfaces, and a receiver-side scope gate on `copilot_vscode` keys in the OTLP route. No server env vars, no helm values, no BYOC dataplane change. Traffic flows to the existing `/api/otel` ingest path. Vanilla `helm install` unaffected.

## Review hardening (2026-08-05)
An adversarial review round (verified by execution) found and fixed in `c0be58c6c`: a JSONC-destroying settings write (P0 — now `jsonc-parser`-based, comment-preserving, refuse-don't-clobber, atomic), hardening gated on the consent prompt instead of env injection, foreign-telemetry hijack via inherited `OTEL_*` (receiver scope gate), an ungovernable `code` surface (server policy row), an unguarded gateway fallthrough (clear 501), stale notices, daemon eligibility, and stray tracked build artifacts. Full ledger on the tracking issue.


langwatch-agent:
Scariest risk: captures VS Code Copilot Chat data. Anything capturing assistant chat is a privacy surface: what is captured, from whom, and is it on by default?
- **Must Check** capture scope and whether it is opt-in; confirm no secrets/source content leave the machine unintentionally.
- **Ask Author** what exactly is stored, and who can read it?
Attention map only.

typescript-sdk/src/cli/utils/governance/wrapper-mode.ts
```diff
@@ -584,6 +588,24 @@ export function buildOtelEnvBlock(
         ...base,
         OTEL_RESOURCE_ATTRIBUTES: "service.name=copilot-cli",
       };
+    case "code":
```
Aryansharma28: [P3] VS Code's long-lived integrated terminals inherit the OTLP bearer token for the whole editor session, not just one invocation.

**Why:** `code` joins `SHELL_FUNCTION_TOOLS` alongside `gemini`/`opencode`/`copilot`, all sharing the same "scope env to this invocation only" design. But those three are short-lived foreground CLI processes that exit when the tool exits, while `code` launches a long-running GUI editor whose integrated terminals inherit `OTEL_EXPORTER_OTLP_HEADERS` (carrying the `copilot_vscode` bearer token) for the life of that VS Code window — potentially hours or days, across many terminal sessions. The ADR (`dev/docs/adr/039-copilot-cli-as-tracked-coding-assistant.md:79-80`) names this exact tradeoff as an open, unresolved question but ships v1 with no mitigation.

**Evidence:** this `case "code":` block sets `OTEL_EXPORTER_OTLP_HEADERS` via `...base` with no scope restriction beyond the shell function itself; `shell-rc.ts:66-74`'s `SHELL_FUNCTION_TOOLS` list groups `code` with the short-lived tools despite the different process lifetime.

**Suggested fix:** either strip `OTEL_EXPORTER_OTLP_HEADERS` via a `terminal.integrated.env.*` settings write when persisting the `code()` function, or open a tracked follow-up issue instead of leaving it as an unresolved ADR footnote.

typescript-sdk/src/cli/index.ts
```diff
@@ -293,6 +293,22 @@ program
     }
   });
 
+program
+  .command("code", { hidden: true })
```
Aryansharma28: [P3] `langwatch code` is invisible in `langwatch --help`, unlike every sibling wrapped-tool command.

**Why:** All six other wrapped-tool commands (`claude`, `codex`, `copilot`, `cursor`, `gemini`, `opencode`) are registered `hidden:true` here but then explicitly re-surfaced in the curated "Coding assistants:" block via `program.addHelpText("after", ...)` at [lines 383-391](https://github.com/langwatch/langwatch/blob/dc18460a725c830ef13abe2196f436735f29d8c5/typescript-sdk/src/cli/index.ts#L383-L391) — that's the documented reason for hiding them in the first place. This new `code` registration is `hidden:true` the same way but was never added to that array, so it's fully undiscoverable via `--help`.

**Evidence:** `addHelpText` array lists `claude`/`codex`/`copilot`/`cursor`/`gemini`/`opencode` but not `code` (confirmed on the current head SHA — that section is unchanged by this diff).

**Suggested fix:** add a `"  code            Run \`code\` (VS Code) with LangWatch telemetry for Copilot Chat"` line to the addHelpText array alongside the other five.

typescript-sdk/src/cli/utils/governance/wrapper-mode.ts
```diff
@@ -336,7 +340,7 @@ export async function resolveWrapperMode(
   // lets the inherited "false" win in the spawn merge; the notice makes
   // the tokens-only consequence visible instead of silent (ADR-039 D5).
   if (
-    tool === "copilot" &&
+    (tool === "copilot" || tool === "code") &&
```
Aryansharma28: [P4] The opt-out branch now also covers `code`, but the notice text it triggers still hardcodes "copilot" wording.

**Why:** This line correctly extends the check to `tool === "code"`, but the `optOutNotice` string a few lines below (unchanged by this PR, not part of this diff) reads "...copilot traces will carry tokens only" regardless of which tool triggered it. Not factually wrong — VS Code Copilot Chat is itself a GitHub Copilot surface — just generic about which tool the user is actually using. The added test only asserts `toContain("tokens only")`, so it wouldn't catch wording drift either way.

**Suggested fix:** low priority — parameterize the notice on `tool` if/when a third tool joins this branch; not worth a churn-only fix today.

Aryansharma28 (commented):
## goated-review summary

**Verdict: COMMENT** · P0: 0 · P1: 0 · P2: 0 · P3: 2 · P4: 1 · P5: 2

Clean, DRY-consistent reuse of the existing scoped-shell-function tier to add VS Code Copilot Chat as a capture surface — no new receiver-side code needed, and confirmed the shared GenAI extractor's scope-gated `hasCopilotProvenance` check correctly stays inert for the new `copilot-chat` instrumentation scope. CI is green, this PR's 12 commits are cleanly independent of the still-open #5784 base, and there's no existing human/CodeRabbit review to dedupe against (CodeRabbit auto-skipped as expected for a stacked, non-default-base PR). No P0/P1/P2 found. Two P3s are worth the author's attention before/soon after merge.

### P3
- `typescript-sdk/src/cli/utils/governance/wrapper-mode.ts:591` — VS Code's long-lived integrated terminals inherit the OTLP bearer token for the life of the editor window, not just one invocation (unlike the other scoped-function tools, which are short-lived CLI processes). The ADR itself names this as an open question with no v1 mitigation.
- `typescript-sdk/src/cli/index.ts:297` — the new `code` command is registered `hidden:true` but never added to the curated "Coding assistants:" `addHelpText` block ([lines 383-391](https://github.com/langwatch/langwatch/blob/dc18460a725c830ef13abe2196f436735f29d8c5/typescript-sdk/src/cli/index.ts#L383-L391)), so it's invisible in `langwatch --help` unlike its 6 siblings — contradicting the PR's own stated intent to mirror `copilot`'s entry.

### P4
- `typescript-sdk/src/cli/utils/governance/wrapper-mode.ts:343` — the opt-out branch was extended to `tool === "code"`, but the notice string it triggers (line ~349, unchanged, not part of this diff) still hardcodes "copilot traces will carry tokens only" — not factually wrong since VS Code Copilot Chat is a GitHub Copilot surface, just generic about which tool triggered it. Low priority.

### P5
- Praise: the whole feature is added through exactly the same 4 extension points every prior tool used, zero new abstractions, backed by 11 new unit tests.
- Cross-PR observation (not blocking, inherited from the whole copilot stack, not introduced here): `copilot_vscode`/`copilot_cli`/`copilot_app` aren't in `CODING_AGENT_SOURCE_TYPES` (`langwatch/ee/governance/services/ingestKeyProvenance.utils.ts:69-75`) or `assistantIcons.ts`, so these traces will bucket as generic `ai_tool` rather than `coding_agent` in analytics until someone updates those registries. Flagging for whoever closes out the stack.

- Whether VS Code Copilot Chat's OTel exporter ever falls back to reading an inherited `COPILOT_OTEL_EXPORTER_TYPE` from the environment — the ADR claims empirical end-to-end validation against real VS Code/Copilot Chat builds, which is reasonably strong evidence, but I can't independently confirm the extension's internal config resolution.
- Whether the two `@integration @unimplemented` scenarios in the new feature files are bound to real integration tests elsewhere in the stack (#5784).

typescript-sdk/src/cli/utils/governance/shell-rc.ts
```diff
@@ -67,6 +67,10 @@ export const SHELL_FUNCTION_TOOLS: readonly string[] = [
   "gemini",
   "opencode",
   "copilot",
+  // `code` (VS Code Copilot Chat) is a CLI-launched editor with no
+  // config-file env target for the auth header, so it uses the same
+  // scoped-function tier. ADR-039 §Extension #2.
+  "code",
```
Aryansharma28: [P5] Praise — clean, minimal, DRY-consistent reuse of the existing scoped-shell-function tier with no new receiver-side code.

**Why:** This whole feature is added by touching exactly the same extension points every prior tool used (`SHELL_FUNCTION_TOOLS`, the `buildOtelEnvBlock` switch, the sourceType map, a hidden CLI command) with zero new abstractions and zero new server-side code — verified the shared GenAI extractor's `hasCopilotProvenance` scope gate correctly stays inert for the new `copilot-chat` instrumentation scope, so nothing can crash or regress for other tools. Backed by 11 new unit tests.

Aryansharma28 (approved):
Approving to formalize my earlier goated-review (COMMENT verdict, 0 P0/P1/P2 — see that review for the 2 P3s and 1 P4 worth author attention post-merge, none blocking). The only reason this was stuck was the base of the stack (#5784) sitting on a stale CHANGES_REQUESTED that predated its own fix commit — dismissed that separately now that the fix is verified. Stack is clear top to bottom.

typescript-sdk/src/cli/utils/governance/wrapper-mode.ts
sergioestebance: Real, and it's a tracked ADR open question (`039…§Extension #2 → Open questions → Integrated-terminal exposure`). Deliberately deferred from v1: the ADR locked VS Code capture as **env-only, no settings.json writer** (empirically, an env-only launch with empty settings still captures). The clean fix — clearing the OTEL vars for integrated terminals via `terminal.integrated.env.<os>` — requires re-introducing that settings writer, which reopens a locked ADR decision and needs a revision, not a quiet add here. Mitigations already in place: our own `langwatch <tool>` wrappers override `OTEL_EXPORTER_OTLP_HEADERS` per-invocation (so wrapped tools are unaffected), and the token is a scoped ingest key, not a VK. Tracking it as the v2 hardening. Skipping in this PR with that reasoning.

typescript-sdk/src/cli/index.ts
sergioestebance: Fixed in 6b3de1056. Added `code` to the curated 'Coding assistants:' help block so it's discoverable via `langwatch --help` like the six siblings. Verified in the built CLI output.

typescript-sdk/src/cli/utils/governance/wrapper-mode.ts
sergioestebance: Fixed in 6b3de1056. The opt-out notice now names the actual tool (`${tool} traces will carry tokens only`) and reads 'is falsey' instead of the stale '=false' (the opt-out is case-insensitive). Test now asserts the notice contains 'code traces'.

typescript-sdk/src/cli/utils/governance/shell-rc.ts
sergioestebance: Thanks — that was the design intent: reuse the exact scoped-shell-function tier every prior tool uses, no new abstractions, no server-side code, and the GenAI extractor stays inert for the `copilot-chat` scope.

typescript-sdk/src/cli/utils/governance/wrapper-mode.ts
sergioestebance: Update: now **fixed in code** (50f8e689b), not just deferred. Added `vscode-settings.ts` which writes a narrow `terminal.integrated.env.<os>` clear (each telemetry key → `null`) into VS Code User settings.json when the scoped `code()` function is installed — so integrated terminals never inherit the `copilot_vscode` token, while the extension host keeps it (read at process launch). Removed on `langwatch logout` via the telemetry-targets scan. A drift guard test asserts the cleared key set equals `buildOtelEnvBlock('code')` so a future key can't slip through. ADR-039 §Ext#2 updated: this narrow terminal-env write is orthogonal to the rejected 'settings carries the capture config' approach — the token is still delivered by env.

PR #5832, opened 2026-07-15
refactor(automations): remove graph-alert cron, event-sourced path is sole path (ADR-034)

0xdeafcafe:
## What

Removes the K8s **graph-alert cron** entirely and makes the **event-sourced path the sole, unconditional** graph-alert path. Retires the `release_es_graph_triggers_firing` flag.

Stacked on #5807 (base branch `worktree-feat-webhook-automation-action`).

## Why

Every project has cut over to the event-sourced path (real-time outbox reactor + 30s heartbeat), so the cron is dead weight and the flag is no longer a meaningful choice. Trace-match triggers were already fully reactive; graph alerts were the last thing on the cron.

## Changes

- **Deleted** the `/api/cron/triggers` route and the whole `pages/api/cron/triggers/` handler tree (`customGraphTrigger`, `actions/{sendEmail,sendSlackMessage,sendWebhookRequest,graphAlertDispatch}`, `utils`, `cronGraphAlertDeps`). The two types still used by the ES path (`ActionParams`, `TriggerData`) were relocated to `app-layer/triggers/trigger.types.ts`; the cron-only types died with the tree.
- **Made the ES path unconditional** — dropped the flag gate from both `graphTriggerEvaluation` reactors (trace + eval pipelines) and the heartbeat's project filter. The heartbeat now processes every project with graph triggers.
- **Removed the flag** `release_es_graph_triggers_firing` from the registry.
- **Removed the `alertTriggers` Helm CronJob** from `values.yaml` (+ README).
- **Tests:** deleted the cron handler suites + the dual-path integration test + the cron-flag-skip route test; dropped the flag mocks/dep from the reactor and heartbeat suites; removed `/api/cron/triggers` from the internal-routes-auth list.
- **Docs:** ADR-034 migration marked complete; ADR-040/044 cron references updated; the `event-sourced-graph-triggers.feature` spec rewritten as the unconditional-path contract; stale cron-citing code comments corrected.

## ⚠️ Reviewer note — this is the deliberate trade-off

The ES path is **worker-only**: the outbox drainer and heartbeat are `roleRunsWorkers`-gated, and the cron (which ran inside the app pod) is gone. **A deployment MUST run the workers Deployment for graph alerts to fire — there is no cron fallback anymore.**

- **SaaS:** fine — workers, Redis, and the outbox are always running.
- **Self-hosted (standard Helm):** fine — `workers.enabled` defaults to `true`.
- **Self-hosted with `workers.enabled: false`** (a supported minimal config): graph alerts will **not fire**. Such a deployment already can't process traces, so it's arguably already non-functional, but this makes it explicit. Documented in ADR-034 + chart notes.

If we'd rather not hard-remove the fallback for that edge case, the alternative is to keep the flag and instead add a startup guard that fails loudly when the flag is on but workers aren't running — say the word and I'll switch to that.

## Verification

`pnpm typecheck` clean; 216 affected test files green (7310 tests). No behavior change for any project already on the ES path (i.e. all of them).


charts/langwatch/values.yaml
```diff
@@ -1103,16 +1103,10 @@ cronjobs:
...
-    # See: https://kubespec.dev/batch/v1/CronJob#spec.schedule
-    ## @extra cronjobs.jobs.alertTriggers Alert triggers job.
-    ## @param cronjobs.jobs.alertTriggers.enabled [default: true] Enable alert triggers job.
-    ## @param cronjobs.jobs.alertTriggers.schedule [string, default: */3 * * * *] Cron schedule.
-    ## @param cronjobs.jobs.alertTriggers.endpoint [string, default: /api/cron/triggers] Endpoint path.
-    alertTriggers:
-      enabled: true
-      schedule: "*/3 * * * *" # Every 3 minutes
-      endpoint: "/api/cron/triggers"
+    # NOTE: the alertTriggers CronJob (`/api/cron/triggers`) was removed —
```
langwatch-agent, line 1106: **P1 — Removing this CronJob silently disables existing graph alerts for a supported workerless chart configuration.** The chart itself deliberately renders `workers.enabled: false` in `examples/overlays/size-minimal.yaml` and asserts that no worker Deployment exists in `tests/e2e-overlays.sh`. After an upgrade, Helm deletes `alertTriggers`; the web pod never starts the outbox drainer or heartbeat, while graph-alert rows can remain active and editable. New breaches and no-data resolutions then produce no notification, with no operator-visible failure.

The comment only documents the dependency; it does not make the configuration safe. Preserve a scheduler path for workerless installs, or make the chart/application fail closed when graph alerts exist but workers are disabled and add an upgrade-level regression test for that configuration.

langwatch-agent (changes requested):
Request changes: 1 P1 inline. The current stacked CI is also red (typecheck and integration suites), so this is not merge-ready until the parent-stack failures are resolved and checks rerun.

0xdeafcafe:
Superseded by #5905, which carries the full cron removal (ADR-034) plus the reaction-API rework (ADR-052) it was stacked toward.

Issue #5843, opened 2026-07-16
Security: @opentelemetry/core <2.8.0 alerts cannot close while otlp-transformer is pinned to 0.55.0 (needs a decision)


langwatch-agent:
Filing this so it stops being re-investigated on every triage pass (I have now traced it twice, once here and once in the scenario repo).

## The alert

langwatch/tasks#36 - `@opentelemetry/core` `< 2.8.0`, GHSA-8988-4f7v-96qf (Unbounded memory allocation in W3C Baggage propagation), MODERATE, against `langwatch/pnpm-lock.yaml`. Dependabot raises it at the manifest level, so it stays open while **any** sub-2.8.0 `@opentelemetry/core` remains in the tree.

## Why it cannot be fixed

`langwatch/package.json` already pins the direct dependency at a patched `"@opentelemetry/core": "2.9.0"`. The alert persists because the lockfile carries five parallel core versions:

| core version | pulled by |
|---|---|
| **1.28.0** | `@opentelemetry/otlp-transformer@0.55.0` (+ its 1.28.0-era siblings) |
| 2.1.0 | the `0.205.0`-era exporters subtree |
| 2.7.1 | the `0.217`/`0.218`-era subtree (~30 packages) |
| 2.8.0 / 2.9.0 | current, patched |

The blocker is the first row. `@opentelemetry/otlp-transformer@0.55.0` declares an **exact** dependency, not a range:

```
"@opentelemetry/core": "1.28.0"
```

and we hold `otlp-transformer` at `0.55.0` deliberately (`langwatch/package.json` line 113, with a matching `ignore` in `.github/dependabot.yml` from langwatch/langwatch#5691: "Pinned intentionally: langwatch needs @opentelemetry/otlp-transformer held at 0.55.0", alongside the separate `otlp-transformer-next` alias for the newer line).

So core 1.28.0 is not incidental drift, it is a direct consequence of a deliberate pin. A pnpm override forcing core off 1.28.0 would break `otlp-transformer@0.55.0`, which requires exactly that version. Lifting the 2.1.0 and 2.7.1 subtrees would not help either: 1.28.0 would still be there and the alert would still not close.

## What would have to happen

One of:

1. **Accept and dismiss** langwatch/tasks#36 as won't-fix while the pin stands (my recommendation, if the pin is permanent). The vulnerable path is W3C Baggage propagation, and this is a transitive of a pinned exporter.
[...]

Issue #1370, opened 2026-02-05
Scenario Library: → Runs link per row


drewdrewthis:
## Summary

Add a quick-access link on each scenario row in the library to view its run history.

## Context

From the [library wireframe](https://github.com/langwatch/langwatch-project-planner/blob/main/docs/scenario/wireframes/01-library.md), each row should have a `[→ Runs]` link that filters the runs view by that scenario — a quick health check without leaving the library context.

## Notes

- The runs page doesn't exist yet — this will need a filtered view or dedicated route
- Could be a link to a filtered simulation results page, or an inline expandable panel
- Design decision: link vs inline preview TBD

## Acceptance Criteria

1. Each scenario row has a visible link/button to view its runs
2. Clicking navigates to a filtered view showing only that scenario's run history

PR #5844, opened 2026-07-16
fix(security): move ts-sdk evaluation/langchain examples to otel core >=2.8.0

langwatch-agent:
## What

Bumps `@opentelemetry/sdk-node` from `^0.217.0` to `^0.219.0` in `typescript-sdk/examples/evaluation` and `typescript-sdk/examples/langchain`, and regenerates both `package-lock.json` files.

## Why

Both examples were stuck on `@opentelemetry/core` **2.7.1**, which is in the range of GHSA-8988-4f7v-96qf (unbounded memory allocation in W3C Baggage propagation, patched in 2.8.0). The cause is the caret: for a `0.x` version `^0.217.0` is minor-locked (`>=0.217.0 <0.218.0`), so the examples could never reach a core-2.8.0 pairing on their own. `sdk-node@0.219.0` is the first release that pairs with core 2.8.0.

Resolves langwatch/langwatch#1454 (evaluation) and langwatch/langwatch#1456 (langchain), both MODERATE.

## Risk

Example/demo code only, not shipped and not imported by the SDK. Everything that moves is `sdk-node`'s own transitive set (`import-in-the-middle`, `es-module-lexer`, `protobufjs`, `yargs`, `acorn` dropped); no unrelated packages change.

## Testing

These examples are not covered by any CI workflow, so this was verified locally:

- `npm install` resolves cleanly in both (no ERESOLVE).
- Every installed `@opentelemetry/core` is **2.8.0** in both examples.
- `require('@opentelemetry/sdk-node')` loads and `NodeSDK` is intact, resolving core 2.8.0.

(`tsc --noEmit` is not a usable signal here: the examples have no typescript/`@types/node` dependency and reference the local SDK via `file:../..` unbuilt, so it errors identically on `main`.)

## The other three examples are deliberately not touched

`langgraph` (#1459, plus uuid langwatch/tasks#9), `vercel-ai` (#1464) and `mastra` (#1463) are left alone because **their lockfiles cannot be regenerated on `main` today** - `npm install` already fails there with pre-existing `ERESOLVE` conflicts, before any change of mine:

- `langgraph`: `@langchain/community@1.1.29` needs peer `@langchain/core@^1.1.38`, tree has `1.2.3`.
- `vercel-ai`: conflicting peer `@opentelemetry/instrumentation@0.57.2`.

Forcing them through with `--force` / `--legacy-peer-deps` would commit a resolution npm itself flags as "incorrect (and potentially broken)", in examples with no CI to catch the fallout. Those conflicts want fixing on their own merits first. `mastra` has no direct otel dependency, so it needs a different approach.

Separately, the `@opentelemetry/core` alert on the **main app** (#1370) is structurally unfixable while `otlp-transformer` stays pinned at 0.55.0 - see langwatch/langwatch#5843.

## Summary by CodeRabbit

* **Chores**
  * Updated the OpenTelemetry Node SDK used by the evaluation and LangChain examples to version 0.219.0.
* **CI**
  * Adjusted the end-to-end workflow so the “Start server and worker” step runs without `LANGWATCH_API_KEY` being set at boot time, reducing unintended key exposure during startup.

### DIFF
--- a/.github/workflows/sdk-javascript-ci.yml
+++ b/.github/workflows/sdk-javascript-ci.yml
@@ -220,6 +220,12 @@ jobs:
 
       - name: Start LangWatch server and worker
         working-directory: langwatch
+        env:
+          # The app server and the workers are langwatch *platform* processes, and
+          # a platform process refuses to boot with LANGWATCH_API_KEY set (it would
+          # export its own telemetry into its own trace ingest). The workflow-level
+          # key belongs to the SDK client under test, so clear it for this step only.
+          LANGWATCH_API_KEY: ""
         run: |
           NODE_ENV=production pnpm run start &
           # Start the worker process so the event-sourcing pipeline
--- a/typescript-sdk/examples/evaluation/package-lock.json
+++ b/typescript-sdk/examples/evaluation/package-lock.json
@@ -12,7 +12,7 @@
         "@ai-sdk/openai": "^2.0.6",
         "@opentelemetry/api": "^1.9.0",
         "@opentelemetry/context-async-hooks": "^2.0.1",
-        "@opentelemetry/sdk-node": "^0.217.0",
+        "@opentelemetry/sdk-node": "^0.219.0",
         "ai": "^5.0.8",
         "dotenv": "^16.6.1",
         "langwatch": "file:../../",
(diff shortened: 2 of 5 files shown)


langwatch-agent:
Pushed a second commit here, because this PR could not go green without it and the cause is worth flagging on its own.

**The `e2e` / `sdk-javascript-complete` red was not a flake and not this bump.** The job dies at *"Start LangWatch server and worker"*, before any test runs:

```
LANGWATCH_API_KEY: sk-lw-1234567890
Error: LANGWATCH_API_KEY must not be set on a langwatch platform process — it
makes the platform self-reference its own trace ingest (a feedback loop).
```

`sdk-javascript-ci.yml` sets `LANGWATCH_API_KEY` at **workflow level** (line 21) for the SDK client under test, so it also lands on the step that boots the langwatch app + workers. Those are platform processes, and `langwatchPlatformGuard` (added in langwatch/langwatch#5745, merged 2026-07-13) refuses to boot with that key set. The fix clears the key for that one step; the `Run e2e tests` step still receives it. The guard is a truthiness check, so an empty value satisfies it.

Evidence it is pre-existing rather than mine:
- `main`'s own e2e runs pass, and so do other open PRs, because this job's path filter (`typescript-sdk/**`) makes it **skip** for them. This PR touches `typescript-sdk/examples/**`, so it actually runs the job and hits the wall.
- I re-ran the failed job once: it failed identically, deterministically (not the 120s-timeout flake signature).

So any PR touching `typescript-sdk` has been unable to get `sdk-javascript-complete` green since 2026-07-13. Happy to split this into its own PR if you would rather review it separately.

langwatch-agent:
Superseded by langwatch/langwatch#6449 - closing this one.

This PR moved the **evaluation** and **langchain** examples to otel core >=2.8.0 (#1454, langwatch/langwatch#1456). The same advisory is open on the other three example locks too (#1459 langgraph, langwatch/langwatch#1463 mastra, langwatch/langwatch#1464 vercel-ai), and langwatch/langwatch#6449 already rewrites all five, so keeping both would just have them conflicting on the same files. langwatch/langwatch#6449 now carries the otel core fix for **all five** examples plus 14 HIGH transitive alerts, 19 in total.

Two notes on what changed since this PR was opened:

- Its second commit (the `LANGWATCH_API_KEY: ""` fix for `sdk-javascript-ci`) is **already on main** - an equivalent fix landed independently, which is what made this PR conflict. Nothing lost.
- mastra needed a different mechanism: it has no `@opentelemetry/sdk-node` dependency, so raising the sdk-node range does nothing there. It takes a scoped `@opentelemetry/core` override instead.

PR #5845, opened 2026-07-16
fix(triggers): surface Slack dispatch failure cause + bound alert field sizes

0xdeafcafe:
## Why

Prod Slack dispatches for trigger `SN2ndrS8w98dL7PPSdB1m` ("hallucination detection(regex)") have been failing since 2026-07-13 with only:

```
Slack webhook dispatch failed for trigger "hallucination detection(regex)"
```

Two problems fell out of the investigation:

1. **The real cause is invisible.** The Slack HTTP status lives on `DispatchError.cause`, but every sink (outbox dispatcher log, group-queue audit rows, Sentry) serializes only `error.message` — so a revoked webhook (404 `no_service`) is indistinguishable from a rejected payload (400).
2. **The doomed dispatch never leaves the queue.** A non-retryable DispatchError went through `handleExhaustedRetries`, which blocked the group and re-staged the job — prod logs show the same job re-staging and re-firing across days (`.../r/…/r/…/r/…` chains), stalling every later job for the same trigger group.

## What

- `toDispatchError` now appends `HTTP <status> — <provider message>` (capped at 300 chars) to the DispatchError message, so all log sinks show the real failure. Existing DispatchErrors pass through unchanged; `retryable` overrides keep the detail.
- The group queue now **completes a non-retryable DispatchError out of the queue** instead of blocking the group and re-staging: the dead outbox audit row (via `onDead`, unchanged) is the operator's recovery surface, matching the existing spec promise that "no further attempts are scheduled". CRITICAL machinery errors (validation/security/config) keep the park-for-operator block.
- The legacy Slack alert builder interpolated full, unbounded trace input/output and event metric/detail values (up to 10 traces) into one message; Slack rejects oversized payloads with a terminal 400, dead-lettering the alert. Every field is now truncated to 500 chars with an ellipsis (before mrkdwn escaping, so the cut never lands mid-entity).

## Specs

- `specs/event-sourcing/dispatch-error-contract.feature`: classified failures carry the provider's failure detail.
- `specs/event-sourcing/reactor-outbox-dispatch.feature`: a non-retryable dispatch failure leaves the queue instead of blocking the group.
- `specs/triggers/slack-webhook-dispatch.feature` (new): oversized trace content is truncated, not dropped.

## Tests

- `dispatchError.unit.test.ts`: status+message in the wrapper, override keeps detail, oversized cause message capped, no-detail case untouched.
- `sendSlackWebhook.unit.test.ts`: revoked-webhook error names HTTP 404; oversized input/output/event values truncated with ellipsis; short content untouched.
- `groupQueue.integration.test.ts`: a job failing with a non-retryable DispatchError completes out of the queue — the next job in the same group still dispatches, and the failed job is not re-fired.
- All 223 group-queue integration tests (incl. poison guard) and all DispatchError-touching unit suites pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Operator-visible dispatch errors now consistently show provider HTTP status and details, with secret scrubbing and cause message truncation.
  - Group-queue handling for non-retryable failures is refined: provider-terminal errors complete out of the queue when the dead transition is durably recorded; configuration/integrity errors block for operator intervention.
  - Slack webhook alerts now strictly enforce payload size limits via truncation/byte bounds to avoid rejected deliveries.

- **Documentation**
  - Expanded specs covering dispatch error classification, truncation/redaction behavior, and the updated group-queue routing rules.

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
@@ -26,7 +26,10 @@ import {
   type ProjectStorageDestination,
   redactStorageUrisInText,
 } from "../../../stored-objects/project-storage-destination";
-import { isDispatchError } from "../../outbox/dispatchError";
+import {
+  isDispatchError,
+  isProviderTerminal,
+} from "../../outbox/dispatchError";
 import type {
   DeduplicationConfig,
   EventSourcedQueueDefinition,
@@ -104,8 +107,13 @@ const DEFAULT_DEDUPLICATION_TTL_MS = 200;
  * Two non-retryable classes:
  * - event-sourcing errors categorized CRITICAL (validation/security/config)
  * - outbox `DispatchError`s explicitly marked `retryable: false` — the
- *   dispatcher rethrows these so the queue must dead-letter rather than
- *   re-fire a dispatch the dispatcher already judged unrecoverable.
+ *   dispatcher rethrows these so the queue never re-fires a dispatch the
+ *   dispatcher already judged unrecoverable.
+ *
+ * This decides only whether to re-fire. What becomes of a job that must not be
+ * re-fired is a separate call the failure branch makes on the error's
+ * disposition: a provider-terminal failure completes out of the queue, every
+ * other non-retryable failure parks for an operator.
  */
 export function isRetryableJobError(err: unknown): boolean {
   if (isDispatchError(err) && !err.retryable) return false;
@@ -702,6 +710,36 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
   }
 
   /**
+   * Load-bearing counterpart to {@link runAudit} for the one transition a
+   * caller must not lose: `dead`. Returns whether the adapter durably recorded
+   * it, so the provider-terminal path can decide between completing the job and
+   * parking it. A thrown adapter reads as "not recorded" — the point is to
+   * distrust the write, not to let the queue die with it.
+   *
+   * A queue with no audit adapter has no dead row to strand, so there is
+   * nothing to confirm and the caller may proceed.
+   */
+  private async confirmDeadTransition(event: {
+    payload: Payload;
+    lastError: string;
+    attempt: number;
+  }): Promise<boolean> {
+    if (!this.auditAdapter) return true;
+    try {
+      return await this.auditAdapter.onDead(event);
+    } catch (err) {
+      this.logger.warn(
+        {
+          queueName: this.queueName,
+          error: err instanceof Error ? err.message : String(err),
+        },
+        "Audit adapter onDead threw; treating the dead transition as unrecorded",
+      );
+      return false;
+    }
+  }
+
+  /**
    * fastq worker function: poison guard, then the real job processing.
    *
    * The guard records a claim strike BEFORE any decode/parse work and clears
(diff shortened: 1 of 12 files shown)


langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -1161,7 +1161,15 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
                 span.setAttribute("error", true);
                 span.setAttribute("error.message", error.message);
 
-                if (!isRetryable) {
+                if (isDispatchError(err) && !err.retryable) {
```
langwatch-agent: **P1 — This bypass applies to critical configuration, security, and integrity failures, not just terminal provider responses.** `DispatchError` is also thrown for a mixed cadence batch, missing project, invalid Slack URL, missing bot token/channel, unsupported graph action, and notify-stage misrouting. All of those are `retryable: false`, so this condition completes them instead of taking the existing park-for-operator path, contrary to the PR description. A corrupt/misrouted cadence batch can therefore dead-letter the affected alert and let later work continue with the broken invariant. Add an explicit disposition/subtype for provider-terminal failures and restrict queue completion to that disposition; retain blocking for manually classified config/security/integrity errors. Cover both paths in the group-queue integration test.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -1172,19 +1180,35 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
                       errorCategory: category,
                       error: error.message,
                     },
-                    "Job failed with non-retryable error, skipping retries",
+                    "Dispatch failed non-retryably; job completed out of the queue, dead outbox row is the recovery surface",
                   );
-                }
+                  await this.scripts.complete({ groupId, stagedJobId, jobName });
```
langwatch-agent: **P1 — The job is irreversibly completed before its only recovery record is durably written.** `runAuditAll` immediately below is explicitly best-effort, and `PgOutboxAuditAdapter.write` catches database failures and resolves. During a transient Postgres outage after a terminal Slack response, this removes the Redis job, releases its blob, and leaves the existing `ReactorOutbox` row stuck in `dispatching`; there is no reconciliation path for that state. The stated dead-letter recovery surface is consequently absent. Persist/acknowledge the dead transition before completing the queue slot, and park or retry when that durable transition cannot be confirmed. Add a test with a failing audit adapter that verifies the job remains recoverable.

langwatch/src/server/triggers/sendSlackWebhook.ts
```diff
@@ -22,6 +22,25 @@ const escapeMrkdwn = (value: unknown): string =>
...
+const MAX_FIELD_LENGTH = 500;
+
+const truncateField = (value: unknown): string => {
+  const text = String(value ?? "");
+  return text.length > MAX_FIELD_LENGTH
+    ? text.slice(0, MAX_FIELD_LENGTH) + "…"
+    : text;
+};
+
+const formatField = (value: unknown): string =>
```
langwatch-agent, line 63: **P1 — Capping raw fields does not bound the Slack payload, so valid alerts can still receive the same terminal 400.** Escaping occurs after the 500-character cap: 500 `&` characters become 2,500 characters. With ten traces and just input/output, that is already about 50 KB before headers; event count, metric/detail key lengths, trigger name, and trigger message are also unbounded. The new tests use plain ASCII and one event, so they miss this case. Enforce one final byte budget on the fully escaped `text` payload (and bound entry counts/keys as needed), then test escaping-heavy and high-cardinality traces remain below Slack's limit.

langwatch-agent (changes requested):
3 P1s: terminal queue completion is not durable, the new bypass also catches critical configuration/integrity DispatchErrors, and the Slack payload is still unbounded after escaping. See inline comments.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
0xdeafcafe: Fixed in 0178085ad. `DispatchError` now carries a disposition (`provider_terminal` | `config`) plus an `isProviderTerminal()` helper. `toDispatchError` marks `provider_terminal` **only** when the non-retryable verdict is derived from the provider's own terminal HTTP status; a caller-supplied `retryable: false` (invalid Slack URL, missing project/token/channel, unsupported graph action, mixed cadence batch, notify misrouting) is `config`. Queue completion now requires `provider_terminal`; everything else parks via `handleExhaustedRetries`. ADR-027 and the contract feature are updated, with tests covering both dispositions.

One thing worth your eye: hand-constructed `new DispatchError({ retryable: false })` sites (`slackWebApi.ts`, `httpDestination.ts`, `dispatcher.ts`, the guards) now park. That restores the pre-#5845 behaviour (before a97b3878b everything non-retryable parked), so it is not a regression — but `slackWebApi.ts:122` throws `retryable: false` for Slack API codes like `channel_not_found`, which is arguably a genuine provider verdict. I left it parking because your finding explicitly lists "missing bot token/channel" as a must-park case; say the word if blocked groups on `channel_not_found` would be operationally painful and it can be promoted to `provider_terminal`.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
0xdeafcafe: Fixed in 0178085ad. `onDead` now reports whether it durably recorded the transition (it previously returned void while `write()` swallowed PG failures). The provider-terminal path calls a new `confirmDeadTransition()` **before** `scripts.complete()`: confirmed → complete + release the blob; unconfirmed (write failed or the adapter threw) → do not complete, park and warn, so the job stays recoverable and a later tick re-marks it. The block/park path keeps the best-effort `runAuditAll`. Added a test with a failing audit adapter asserting the job is not completed.

Two notes: drained batch siblings stay best-effort — none of them owns the queue slot being completed, and there is no clean per-sibling confirmation. And `onDead` returns false when `rowsAffected === 0`, which also covers a CAS miss (row already `dead`, or re-leased at a different attempt) — that parks rather than completes. It does not arise in the normal flow (the row is `dispatching` at the matching attempt) and parking is the fail-safe direction, but flagging it since it is the one case where "cannot confirm" and "already recorded" look alike.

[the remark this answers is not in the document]
0xdeafcafe: Fixed in 0178085ad — see the detailed reply on the sibling thread. The dead transition is now load-bearing on the provider-terminal branch: `onDead` reports whether it durably persisted, `confirmDeadTransition()` runs before `scripts.complete()`, and an unconfirmed transition parks the job instead of removing it. Covered by a failing-audit-adapter test.

langwatch/src/server/triggers/sendSlackWebhook.ts
0xdeafcafe: Fixed in 0178085ad. The bound now applies to the assembled, fully-escaped `text` rather than the raw fields: `boundTextBytes()` caps it at 39,000 bytes on a UTF-8 boundary, and entry counts are bounded too (10 events/trace, 20 entries/event, 100-char keys), along with the previously-unbounded trigger name and message. Tests cover escaping-heavy content (many `&`/`<`/`>`, where 500 raw chars became 2,500) and high-cardinality traces, asserting the final payload stays under the budget; `specs/triggers/slack-webhook-dispatch.feature` is updated.

0xdeafcafe:
rewritten, no longer needed.

Issue #4977, opened 2026-06-19
langy: caller-scoped per-chat API keys (RBAC option B, follow-up to #4913)


Aryansharma28:
Follow-up from [PR #4913](https://github.com/langwatch/langwatch/pull/4913).

## Context

PR #4913 ships the smallest validating slice for RBAC: /api/langy/chat now requires {resource}:update on every resource the Langy service key carries write access to. This closes the privilege-escalation hole (viewer can no longer ask Langy to write what they can't write directly), but the worker still holds a service API key with admin-equivalent project-wide permissions.

## What this issue is for

Replace the static per-project Langy service key with a **per-chat-session API key minted with the caller's exact permissions**:

- On /chat, mint a short-lived API key bound to session.user.id's actual role bindings on projectId
- Worker uses that key for all MCP / API calls — every tool authorises against the caller, not the service key
- Key TTL <= chat session duration; revoke on session end / chat error
- Persist a chat-token-to-user-binding row for audit

## Trade-offs

- Mint + revoke per chat is more DB churn than today's getOrProvision-once-and-cache
- ApiKey.create + delete in the hot path
- Requires session-scoped API keys to be supported by ApiKeyService (audit needed)

## Relevant code

- langwatch/src/server/services/langy/langyApiKey.ts (provisionLangyApiKey — replace with per-chat mint)
- langwatch/src/server/services/langy/LangyCredentialService.ts (getOrProvision — replace cache with per-chat mint)
- langwatch/src/server/routes/langy.ts (LANGY_REQUIRED_PERMISSIONS loop — delete once option B lands)
- langwatch/src/server/api-key/api-key.service.ts (needs session-scoped key shape)

drewdrewthis:
Thanks for the detailed proposal here. Closing this out: caller-scoped session keys shipped in #7389. `mintLangySessionApiKey` intersects the requested key policy with the caller's own project permissions, and `virtualKeys:manage` is deliberately withheld from minted keys, so a delegated hierarchy can never grant rotate beyond what the caller themselves holds. If any residual sub-ask from this issue remains, please reopen it separately so it gets its own tracking.

Issue #4008, opened 2026-05-13
refactor: enforce requirePermission at the type-system level for all Hono routes


sergioestebance:
## Context

PR #3951 added `requirePermission()` / `requireOrgPermission()` middleware to all 89 Hono endpoints across 15+ route modules. The RBAC enforcement is correct **today**, but it relies entirely on developer discipline — nothing in the type system prevents shipping a new endpoint without a permission check.

[Reviewer comment on PR #3951](https://github.com/langwatch/langwatch/pull/3951):
> *"I feel like those should be required on the type system level on a follow-up PR, to prevent us from ever forgetting"*

## Problem

Currently, adding a new route like this compiles without errors:

\`\`\`typescript
app.get("/api/agents/dangerous", async (c) => {
  // No requirePermission() — TypeScript is happy, RBAC is bypassed
  return c.json(await getAllAgents(c.get("project").id));
});
\`\`\`

The \`Variables\` type on the Hono app (\`AuthMiddlewareVariables\`, \`OrgAuthMiddlewareVariables\`) is informational — it tells you *what's available* on context, not *what's required* before the handler runs. There is zero compile-time enforcement that a permission check happened.

## Current state (inventory)

| Module | Endpoints | Permission pattern | Coverage |
|--------|-----------|-------------------|----------|
| agents | 5 | \`requirePermission()\` inline | 100% |
| api-keys | 3 | \`requireOrgPermission()\` inline | 100% |
| dashboards | 6 | \`requirePermission()\` inline | 100% |
| dataset | 12 | \`requirePermission()\` inline | 100% |
| evaluators | varies | \`requirePermission()\` inline | 100% |
| gateway-platform | 19 | \`requireApiKeyPermission()\` hoisted | 100% |
| graphs | 5 | \`requirePermission()\` inline | 100% |
| monitors | 6 | \`requirePermission()\` inline | 100% |
| projects | 5 | \`requireOrgPermission()\` inline | 100% |
| prompts | varies | \`requirePermission()\` inline | 100% |
| scenarios | varies | \`requirePermission()\` inline | 100% |
| scenario-events | varies | \`requirePermission()\` inline | 100% |
| secrets | 5 | \`requirePermission()\` inline | 100% |
| simulation-runs | 3 | \`requirePermission()\` inline | 100% |
| suites | 7 | \`requirePermission()\` inline | 100% |
| traces | varies | \`requirePermission()\` inline | 100% |
[...]

Issue #3685, opened 2026-05-02
tRPC governance permission granularization sweep


rogeriochaves:
## Context

PR #3524 shipped the AI Governance permissions catalog (`feat(rbac)` commit `385c95e89`):

- New Resources: `governance`, `ingestionSources`, `anomalyRules`, `complianceExport`, `activityMonitor`
- ORGANIZATION_ADMIN role default-grant: full set
- MEMBER + EXTERNAL: zero governance perms
- Custom roles via `CustomRolePermissions` JSON column = the production-shape delegation surface

The chrome (MainMenu.tsx) was wired to `hasPermission('governance:view')` in commit `043726430` — **visibility uses the new fine-grained perms**.

## What this issue tracks

The **tRPC procedure permission granularization sweep**. The new permission catalog exists, but the actual routers in `src/server/api/routers/governance.ts` (and adjacent routers) still gate on the broad `organization:manage` permission. They should swap to fine-grained checks:

| Endpoint family | Current gate | Should become |
|-----------------|--------------|---------------|
| `governance.setupState` / `governance.resolveHome` | `organization:manage` | `governance:view` |
| `governance.ingestionSources.*` (read) | `organization:manage` | `ingestionSources:view` |
| `governance.ingestionSources.*` (mutate) | `organization:manage` | `ingestionSources:manage` |
| `governance.anomalyRules.*` (read) | `organization:manage` | `anomalyRules:view` |
| `governance.anomalyRules.*` (mutate) | `organization:manage` | `anomalyRules:manage` |
| `governance.ocsfExport` | `organization:manage` | `complianceExport:view` |
| `governance.activityMonitor.*` | `organization:manage` | `activityMonitor:view` |

## Why this is a follow-up PR (not in #3524)

3-lane consensus during iter29: doing the procedure rewiring inside #3524 would balloon scope unnecessarily. The chrome already works — it reads `governance:view` from the existing role-permission map (ADMIN gets it via default-grant, custom roles via CustomRolePermissions). The router-level rewiring is a separate sweep with its own test surface (procedure-by-procedure authz integration tests).

## Acceptance criteria

- [ ] Every `checkOrganizationPermission("organization:manage")` call inside `src/server/api/routers/governance.ts` swaps to the appropriate fine-grained perm
- [ ] Custom-role test (security_analyst granted only `governance:view + activityMonitor:view + anomalyRules:view`) can hit GET endpoints + is denied on mutations
[...]

Issue #3429, opened 2026-04-23
fix(auth): teamuser queries miss rolebinding-only users in role service and secrets


sergioestebance:
## Context

Follow-up from #3422 / #3424. The sweep found two more places that query `TeamUser` instead of `RoleBinding` for membership checks, which will fail for users onboarded post-BetterAuth migration.

### 1. `src/server/role/role.service.ts:106` — custom role assignment blocked

`assignRoleToUser()` calls `prisma.teamUser.findUnique()` to verify the user is a team member before assigning a custom role. Post-migration users have no `TeamUser` row, so this check fails and they cannot be assigned custom roles.

**Impact:** Admins cannot assign custom roles to post-migration users.

### 2. `src/app/api/secrets/[[...route]]/app.ts:222` — secrets fallback owner

When creating a secret via API key (no user context), the code calls `prisma.teamUser.findFirst()` to find any team member as the owner. For teams with only RoleBinding users, this returns `null` and falls back to `"system"`.

**Impact:** Low — secrets are attributed to `"system"` instead of a real user. Functional but loses audit trail.

## Proposed fix

Same pattern as #3424: replace `teamUser` queries with `roleBinding.findMany` using `scopeType: TEAM` + `scopeId: teamId`.

## Related

- #3422 — original bug report
- #3424 — fix for trace redaction (merged)

sergioestebance:
Closing: resolved by a merged PR.

PR #5846, opened 2026-07-16
chore(adr-rbac): ADR-046 unified authorization engine - one registry, one resolver, every principal

0xdeafcafe:
## What

ADR-046: collapse our RBAC into one `server/authz/` module - one typed permission registry that knows what each resource can do, one resolver, one decision object - for every surface and every principal. Plus the proposal-stage spec (`specs/rbac/unified-authorization-engine.feature`, 20 scenarios). No runtime changes in this PR. It's the decision doc and the receipts, drawn rather than described (the ADR is diagram-first, ~15 figures).

## Why now

I went digging after getting fed up with how spread out authz is, and found quite some rot:

- **Five implementations** of "does this binding grant this permission": two in `rbac.ts`, one in `role-binding-resolver.ts`, one in the role-binding service, and one in the browser. Two carry literal "must stay in sync" comments. They have already diverged (the EXTERNAL cap exists in one, is missing in the other), and the server patches the client's copy from the outside to keep it honest.
- **The spec and the code disagree on the core semantic.** The spec says most-specific-scope-wins. Every implementation does an additive union. Nobody noticed, which rather answers which semantic has real users.
- **No resource knowledge.** `Permission` is a raw cross product, so `traces:rotate` typechecks and the custom-role validator will happily store it. What a resource can do is hand-maintained in four disconnected places, and a census puts the total at sixteen parallel permission/credential vocabularies.
- **The knock-ons are live.** Legacy project keys bypass RBAC entirely (documented, by design). A lite member's personal API key can out-privilege their own session (SCIM + resolver interaction). Service keys with zero bindings default to org-wide ADMIN. The studio→nlpgo hop runs with no auth and the RBAC-bypassing legacy key injected into every workflow payload.
- The spread, counted: 302 project + 76 org + 5 team attach sites, 23 skips, 16 `authorizeInResolver`, ~15 copy-pasted in-handler auth blocks, 129 raw `role ===` comparisons across 31 files. `rbac.ts` (1,612 lines) has taken 28 commits since January.

Every claim carries a file:line receipt in the ADR's Context section. Related asks it consolidates: #1247, #4008, #3429, #3388, #3685.

## The model (Part I)

```
 ROLE BINDING = WHO holds WHICH ROLE, WHERE
   ┌─ WHO ─────────────┐   ┌─ WHICH ROLE ──┐   ┌─ WHERE ────────────┐
   │ user  alice       │   │ member         │   │ org     acme       │
   │ group sec-eng     │ + │ viewer         │ + │ team    client-a   │
   │ api key lw-sk-42  │   │ custom "SRE"   │   │ project chatbot    │
   └───────────────────┘   └────────────────┘   └────────────────────┘

 can(alice, "prompts:update", project:chatbot)?
 1 COLLECT bindings (alice + her groups)     4 UNION the sets (grants only add)
 2 FILTER  to chatbot → client-a → acme      5 DECIDE membership
 3 EXPAND  role → permission set (registry)  6 RECORD one auditable decision
```

Semantics get settled explicitly: additive union (what prod has always done), and the old spec's override scenarios retire. Lite member becomes a role rather than a cross-cutting cap. API keys keep the owner-ceiling intersection - demotions propagate to your keys instantly, promotions never leak into a scoped key. Legacy keys get a dated sunset. Impersonation becomes a first-class `{actor, subject}` pair instead of a session rewrite.

## Operating it (Part II)

- **§6** one Access surface (principal picker + role picker + `ScopeChipPicker`), with `authz.explain()` rendering why any check passed or failed, and a blast-radius preview before role edits save.
- **§7** fail-closed four layers deep, including branded `Authorized<"project">` witnesses that repositories demand - skip the check and the code does not compile.
- **§8** "authed unless THIS trace is shared" as one composition, with an `audience` field driving redaction.
- **§9** ownership as a registry fact (`ownerImplicit`) for user-created resources.
- **§10** offboarding as one verb with a postcondition: `effectivePermissions(dave, acme) == ∅`, or the operation fails loudly.
- **§11** the API: `authz.can/require/check/explain` + one `grants.attach/update/revoke/replace` write surface that all eight current RoleBinding write paths route through.
- **§12** instant checks: org epochs + bitset caches + short-lived signed passports. Steady-state check is a bit test (< 1 µs, zero queries) against today's 3-10 ms and 3-5 queries per check; revocation still lands on the caller's next request. The gateway's 15-minute `revision` JWT is the in-house precedent.

Migration is six shippable stages, shadow-compared before anything flips. Two deliberate behaviour changes (legacy-key sunset, empty-custom-role semantics), both flagged for product sign-off. A closing section lists what falls out for free later: expiring bindings, access requests, delegation ceilings, agent principals for #4977, richer share links, SOC 2 access reviews, authz anomaly detection.

## Not in this PR

No code. Stage A starts only once this survives review.

### DIFF
--- a/dev/docs/adr/046-unified-authorization-engine.md
+++ b/dev/docs/adr/046-unified-authorization-engine.md
@@ -1,0 +1,993 @@
+# ADR-046: Unified authorization engine - one registry, one resolver, every principal
+
+**Date:** 2026-07-16
+
+**Status:** Proposed (supersedes ADR-001 when accepted)
+
+## Decision, in one paragraph
+
+We will collapse LangWatch authorization into a single `server/authz/` module
+built from three nouns - **permission** (a verb on a resource), **role** (a
+named set of permissions), **role binding** (who holds which role, where) -
+resolved by **one engine** that every surface (tRPC, Hono, services, workers,
+frontend) and every principal (user, API key, share token, demo visitor,
+platform ops, with group membership expanded into user grants) goes through.
+The permission vocabulary becomes a typed registry that knows what each
+resource can do. Grant semantics become an explicit additive union. The legacy
+`TeamUser`/`OrganizationUser` role paths get backfilled into role bindings and
+deleted, and every decision emits one auditable record. Part II (§6-§12)
+covers operating it: one Access surface with "why?" built in, fail-closed
+enforcement down to the type level, a composable public-resource door for
+share links, owner-implicit grants on user-created resources, offboarding as
+one proven verb, the setting/checking API itself, and epoch-validated checks
+that skip the database on the hot path.
+
+## The three nouns (the whole mental model)
+
+```
+ PERMISSION    a verb on a resource                      "prompts:update"
+               (the registry says which verbs each        "traces:share"
+                resource supports - nothing else exists)  "organization:manage"
+
+ ROLE          a named set of permissions                 admin / member / viewer /
+               (built-ins defined in code; custom         lite-member / your own
+                roles are rows holding the same shape)    "SRE on-call" custom role
+
+ ROLE BINDING  WHO holds WHICH ROLE, WHERE - one row, three columns:
+
+               ┌─ WHO (principal) ─┐   ┌─ WHICH ROLE ──┐   ┌─ WHERE (scope) ────┐
+               │ user  alice       │   │ member         │   │ org     acme       │
+               │ group sec-eng     │ + │ viewer         │ + │ team    client-a   │
+               │ api key lw-sk-42  │   │ custom "SRE"   │   │ project chatbot    │
+               └───────────────────┘   └────────────────┘   └────────────────────┘
+```
+
+Bindings sit on the scope tree, and grants flow **down** it. No binding on
+your path up the tree means no access. There is no default access anywhere:
+
+```
+ Organization acme ◄──── alice = admin @ org        (sees/manages everything below)
+ │
+ ├── Team client-a ◄──── group "contractors" = viewer @ team  (read-only, both projects)
+ │    ├── Project chatbot ◄──── api key lw-sk-42 = member @ project
+ │    └── Project emailbot
+ │
+ └── Team client-b       (contractors and the api key have NOTHING here -
+                          no binding on this branch, no access)
+```
+
+An **API key is a principal you can put in a script**. Same binding table,
+same roles, same scopes, plus a leash to its owner:
+
+```
+ ┌──────────────────────────────────────────┐
+ │ ApiKey lw-sk-42                          │
+ │   owner: dave        (none ⇒ service key)│
+ │   bindings: member @ project chatbot     │
+ └──────────────────────────────────────────┘
+
+ effective(key) = grants(key) ∩ grants(owner)         ← intersection, live:
+
+     grants(key)           grants(dave, today)
+   ┌───────────────┐     ┌───────────────┐
+   │ traces:view    │     │ traces:view   │     key can do: traces:view and
+   │ traces:create  │  ∩  │ (dave demoted │  =  nothing else - demoting dave
+   │ datasets:*     │     │  to viewer)   │     demoted his keys instantly,
+   └───────────────┘     └───────────────┘     no rotation ceremony needed
+```
+
+And every access question is one walk:
+
+```
+ can(alice, "prompts:update", project:chatbot)?
+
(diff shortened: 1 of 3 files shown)


0xdeafcafe:
in comparison, an sql i had to write today to check nobody was in an invalid state...

```
WITH flagged AS (
  SELECT
    rb.*,
    ARRAY_REMOVE(ARRAY[
      CASE WHEN rb."userId" IS NOT NULL AND NOT EXISTS (
        SELECT 1 FROM "OrganizationUser" ou
        WHERE ou."userId" = rb."userId"
          AND ou."organizationId" = rb."organizationId"
      ) THEN 'user_not_member_of_binding_org' END,
      CASE WHEN rb."groupId" IS NOT NULL AND NOT EXISTS (
        SELECT 1 FROM "Group" g
        WHERE g.id = rb."groupId"
          AND g."organizationId" = rb."organizationId"
      ) THEN 'group_not_in_binding_org' END,
      CASE WHEN rb."apiKeyId" IS NOT NULL AND NOT EXISTS (
        SELECT 1 FROM "ApiKey" k
        WHERE k.id = rb."apiKeyId"
          AND k."organizationId" = rb."organizationId"
      ) THEN 'api_key_not_in_binding_org' END,
      CASE WHEN rb."scopeType" = 'ORGANIZATION'
            AND rb."scopeId" <> rb."organizationId"
        THEN 'org_scope_points_at_other_org' END,
      CASE WHEN rb."scopeType" = 'TEAM' AND NOT EXISTS (
        SELECT 1 FROM "Team" t
        WHERE t.id = rb."scopeId"
          AND t."organizationId" = rb."organizationId"
      ) THEN 'team_scope_not_in_binding_org' END,
      CASE WHEN rb."scopeType" = 'PROJECT' AND NOT EXISTS (
        SELECT 1 FROM "Project" p
        JOIN "Team" t ON t.id = p."teamId"
        WHERE p.id = rb."scopeId"
          AND t."organizationId" = rb."organizationId"
      ) THEN 'project_scope_not_in_binding_org' END
    ], NULL) AS violations
  FROM "RoleBinding" rb
)
SELECT
  f.id,
  f."organizationId",
  o.name   AS org_name,
  f."userId",
  u.email  AS user_email,
  f."groupId",
  f."apiKeyId",
  f.role,
  f."customRoleId",
  f."scopeType",
  f."scopeId",
  f."createdAt",
  f.violations
FROM flagged f
LEFT JOIN "Organization" o ON o.id = f."organizationId"
LEFT JOIN "User" u ON u.id = f."userId"
WHERE CARDINALITY(f.violations) > 0
ORDER BY f."organizationId", f."createdAt";
```

followed by

```
BEGIN;

WITH bad AS (
  SELECT rb.id
  FROM "RoleBinding" rb
  WHERE
    (rb."userId" IS NOT NULL AND NOT EXISTS (
      SELECT 1 FROM "OrganizationUser" ou
      WHERE ou."userId" = rb."userId"
        AND ou."organizationId" = rb."organizationId"))
    OR (rb."groupId" IS NOT NULL AND NOT EXISTS (
      SELECT 1 FROM "Group" g
      WHERE g.id = rb."groupId"
        AND g."organizationId" = rb."organizationId"))
    OR (rb."apiKeyId" IS NOT NULL AND NOT EXISTS (
      SELECT 1 FROM "ApiKey" k
      WHERE k.id = rb."apiKeyId"
        AND k."organizationId" = rb."organizationId"))
    OR (rb."scopeType" = 'ORGANIZATION' AND rb."scopeId" <> rb."organizationId")
    OR (rb."scopeType" = 'TEAM' AND NOT EXISTS (
      SELECT 1 FROM "Team" t
      WHERE t.id = rb."scopeId"
        AND t."organizationId" = rb."organizationId"))
    OR (rb."scopeType" = 'PROJECT' AND NOT EXISTS (
      SELECT 1 FROM "Project" p
      JOIN "Team" t ON t.id = p."teamId"
      WHERE p.id = rb."scopeId"
        AND t."organizationId" = rb."organizationId"))
)
DELETE FROM "RoleBinding" rb
USING bad
WHERE rb.id = bad.id
RETURNING rb.id, rb."organizationId", rb."userId", rb."groupId", rb."apiKeyId",
          rb.role, rb."scopeType", rb."scopeId";

-- Eyeroll the ret output against step 1, then:
COMMIT;   -- or ROLLBACK;
```


_trans_ actions in 2026... so woke

PR #5847, opened 2026-07-16
perf(clickhouse): add bloom skip-index on simulation_runs.ScenarioRunId

langwatch-agent:
## Evidence

Found by auditing every time-partitioned ClickHouse table for the pattern behind #5810: a sort key led by an id column that has no skip index on that id. `simulation_runs` is the other table that matches, and its single-run lookup shows up in the last 24h of slow-query warns at ~1,835 calls reading ~10.6 MB on average, for what is a point lookup on the table's own sort key.

## Suspected cause

`simulation_runs` is `ORDER BY (TenantId, ScenarioRunId)` and `PARTITION BY toYearWeek(StartedAt)`. A single-run lookup filters on `(TenantId, ScenarioRunId)` and carries no `StartedAt` predicate, so it cannot prune partitions. The primary key still does its job within each part and narrows to one candidate granule, but it must do that in **every** part, and with nothing to rule those granules out they all get read.

`EXPLAIN INDEXES` for such a lookup on two large tenants (ids redacted), for a scenario-run id that does not exist at all:

```
ReadFromMergeTree (simulation_runs)          ReadFromMergeTree (simulation_runs)
  PrimaryKey  Parts: 109/163                   PrimaryKey  Parts: 56/163
              Granules: 109/1414                           Granules: 56/1414
  (no skip index on ScenarioRunId)             (no skip index on ScenarioRunId)
  Parts: 109  Granules: 109                    Parts: 56   Granules: 56
```

The table already carries blooms on `ScenarioId`, `BatchRunId` and `ScenarioSetId` — the secondary lookup paths — but not on the column its own sort key leads with, which is the one the single-run reads filter on.

## Proposed fix

One migration adding the missing index, mirroring the sibling blooms on this same table:

```sql
ALTER TABLE simulation_runs
  ADD INDEX IF NOT EXISTS idx_scenario_run_id ScenarioRunId
    TYPE bloom_filter(0.001) GRANULARITY 1;
```

## Expected impact

A scenario run's rows live in the one or two weekly parts around its `StartedAt`. The bloom lets ClickHouse skip the candidate granule in every other part, so the lookup should go from reading ~56-109 granules to ~1-2. On the measured volume that is a modest ~18 GB/day of reads, but it also takes a user-facing simulation-run read off a per-part granule scan that grows with retention rather than with the run being fetched.

No query changes: it is purely a read-path improvement the planner picks up, and it benefits every `(TenantId, ScenarioRunId)` point read.

**Rollout caveat (needs a reviewer decision).** `ADD INDEX` only applies to parts written *after* it lands, so the gain ramps in with new data and historical parts keep their current behaviour. Backfilling requires:

```sql
ALTER TABLE simulation_runs MATERIALIZE INDEX idx_scenario_run_id;
```

which rewrites index files across the table. That is a heavy operation on the prod cluster, so it is deliberately **not** run inline by this migration and should be scheduled as an ops task.

## Relationship to #5810

Same defect, same fix, different table; #5810 covers `evaluation_runs.EvaluationId` and carries the much larger share of the win (~2 TB/day). They are kept separate so each can be reviewed and rolled out on its own. This one's migration is numbered 00043 and follows #5810's 00042.

For completeness, the audit cleared the other candidates: `event_log` looked similar but its composite `idx_tenant_aggregate_event_id` bloom does answer the aggregate lookup (EXPLAIN confirms it prunes to 0 granules), and `trace_summaries` / `stored_spans` already have their id blooms.

## Test plan

- New integration test (`scenario-run-id-index.integration.test.ts`, testcontainers + real ClickHouse) asserting the migration attaches `idx_scenario_run_id` as a `bloom_filter` via `SHOW CREATE TABLE`, and pinning correctness of the lookup shapes the index affects: the latest version's status wins the argMax across two versions, runs in other partitions are not picked up, a missing run returns no row, and two tenants sharing a scenario-run id stay isolated.
- Verified locally against a real ClickHouse container: the migration runs and all 4 tests pass.
- `tsgo` typecheck clean.

## EXPLAIN check

The EXPLAIN output above was taken from live prod ClickHouse via the read-only ops EXPLAIN endpoint (tenant ids redacted). Reviewer should re-run after the index is materialized to confirm the granule count drops.

## Summary by CodeRabbit

- **Performance**
  - Improved lookup performance for simulation runs by adding an index on scenario run IDs.

- **Bug Fixes**
  - Ensured simulation run lookups return the latest status correctly.
  - Preserved tenant-specific results when scenario run IDs overlap across tenants.
  - Confirmed missing run lookups return no result.

### DIFF
--- a/langwatch/src/server/app-layer/evaluations/repositories/__tests__/evaluation-id-index.integration.test.ts
+++ b/langwatch/src/server/app-layer/evaluations/repositories/__tests__/evaluation-id-index.integration.test.ts
@@ -1,0 +1,192 @@
+/**
+ * @vitest-environment node
+ * @integration
+ *
+ * Verifies the bloom_filter skip-index on evaluation_runs.EvaluationId
+ * (migration 00062):
+ *  - the migration attaches idx_evaluation_id to the table, and
+ *  - the (TenantId, EvaluationId) point-lookup shapes still return the right
+ *    rows, including the ScheduledAt resolve that bounds getByEvaluationId.
+ *
+ * evaluation_runs is PARTITION BY toYearWeek(ScheduledAt), so a lookup that
+ * carries no ScheduledAt predicate cannot prune partitions: the primary key
+ * lands on one candidate granule in every part and, without a skip index on
+ * EvaluationId, reads all of them. The index lets ClickHouse skip granules that
+ * cannot contain the id. Correctness is identical either way, which is what
+ * this test pins.
+ */
+import type { ClickHouseClient } from "@clickhouse/client";
+import { nanoid } from "nanoid";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import {
+  startTestContainers,
+  stopTestContainers,
+} from "../../../../event-sourcing/__tests__/integration/testContainers";
+
+let ch: ClickHouseClient;
+const tag = nanoid();
+
+async function insertRun({
+  tenantId,
+  evaluationId,
+  scheduledAt,
+  updatedAt,
+}: {
+  tenantId: string;
+  evaluationId: string;
+  scheduledAt: Date;
+  updatedAt: Date;
+}) {
+  await ch.insert({
+    table: "evaluation_runs",
+    values: [
+      {
+        ProjectionId: `proj-${nanoid()}`,
+        TenantId: tenantId,
+        EvaluationId: evaluationId,
+        Version: "v1",
+        EvaluatorId: "evaluator-1",
+        EvaluatorType: "llm",
+        EvaluatorName: "test-evaluator",
+        TraceId: `${tag}-trace`,
+        IsGuardrail: 0,
+        Status: "processed",
+        Score: 1,
+        Passed: 1,
+        Label: null,
+        Details: null,
+        Error: null,
+        ErrorDetails: null,
+        CreatedAt: scheduledAt,
+        UpdatedAt: updatedAt,
+        ArchivedAt: null,
+        ScheduledAt: scheduledAt,
+        StartedAt: null,
+        CompletedAt: null,
+        CostId: null,
+        LastProcessedEventId: "",
+      },
+    ],
+    format: "JSONEachRow",
+    clickhouse_settings: { async_insert: 0, wait_for_async_insert: 0 },
+  });
+}
+
+/** The resolve shape used to bound getByEvaluationId's partition scan. */
+async function resolveScheduledAtMs(tenantId: string, evaluationId: string) {
+  const rows = await (
+    await ch.query({
+      query: `
+        SELECT toUnixTimestamp64Milli(argMax(ScheduledAt, UpdatedAt)) AS scheduledAtMs
+        FROM evaluation_runs
+        WHERE TenantId = {tenantId:String}
+          AND EvaluationId = {evaluationId:String}
+      `,
+      query_params: { tenantId, evaluationId },
+      format: "JSONEachRow",
+    })
+  ).json<{ scheduledAtMs: string | number | null }>();
+  const raw = rows[0]?.scheduledAtMs;
+  if (raw === null || raw === undefined) return undefined;
+  const ms = typeof raw === "string" ? Number(raw) : raw;
+  return Number.isFinite(ms) && ms > 0 ? ms : undefined;
+}
+
+beforeAll(async () => {
+  const containers = await startTestContainers();
+  ch = containers.clickHouseClient;
+}, 60_000);
+
+afterAll(async () => {
+  if (ch) {
+    await ch.exec({
+      query: `ALTER TABLE evaluation_runs DELETE WHERE startsWith(TenantId, {tag:String})`,
+      query_params: { tag },
+    });
+  }
+  await stopTestContainers();
+});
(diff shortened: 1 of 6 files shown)


langwatch-agent:
Scariest risk: bloom skip-index on simulation runs; a skip-index keyed on a column the query does not filter buys nothing, and a wrong one can skip rows that should match.
- **Must Check** the index key matches the actual query predicate, and result parity holds before vs after.
- **Ask Author** measured scan reduction on a real query? (Same pattern as #5810; worth reviewing them together.)
Attention map only.

langwatch-agent:
## The rollout caveat in this PR is the important half — evidence in #5864

This PR notes that `ADD INDEX` only covers parts written after it lands, and that backfilling needs `MATERIALIZE INDEX` as a scheduled ops task. I have now measured what happens when that step is skipped, and filed it as #5864.

`trace_summaries.idx_conversation_id` was added the same way by migration 00035 and never backfilled. `EXPLAIN INDEXES` for a non-existent value, split by part age:

```
recent parts (written AFTER 00035)   idx_conversation_id  Parts: 0/15    -- prunes everything
old parts    (written BEFORE 00035)  idx_conversation_id  Parts: 51/92   -- prunes nothing
```

The index works perfectly where it exists and is inert everywhere it was not backfilled. That gap is currently costing ~521 GB/day on the thread-lookup shape.

The implication for this PR: the index here is **brand new**, so 100% of its target data lives in existing parts. Merging without scheduling the backfill lands the correct schema and close to none of the win — it would only creep in as old parts age out. The measured target is worth pairing the merge with the backfill rather than treating it as follow-up bookkeeping.

No change requested to this PR; it stays as-is and green. Flagging so the merge decision and the ops scheduling happen together.

langwatch-agent:
Heads-up on the repeated force-pushes: this PR's migration has now been renumbered twice, and the reason is worth stating plainly.

`goose` is invoked without `-allow-missing` (see `langwatch/src/server/clickhouse/goose.ts` — the args are `-dir`/`-table`/`clickhouse`/conn/command only). That means a migration numbered **below** the applied head is treated as missing and goose **errors out rather than skipping it**, so a stale number here would fail a deploy, not merely fail to add the index. A number **equal** to one already on main is a duplicate goose version, which is equally bad.

Timeline: filed at 00042 when main's head was 00041; main reached 00055, so I moved to 00057; main then took 00057 and 00058 within a day, so this is now **00063** (main's head is 00061). Content is unchanged throughout — only the filename and the migration number in the test description move. Re-verified after each renumber: the full chain applies in order against a real ClickHouse container and all 4 integration tests pass.

main is currently absorbing roughly six migrations a day, so this will need another bump for every day it sits in review. Nothing for you to do about that, and I will keep it current — but it does mean merging sooner costs less churn than merging later. If you would rather I stop force-pushing and leave the final renumber to whoever merges, say so and I will hold.

Aryansharma28 (approved):
Reviewed against goated-review dimensions. Small, well-scoped perf PR (2 files, +233/-0): mirrors the already-established pattern from #5810/migration 00062 (bloom_filter(0.001) skip-index on a sort-key-leading id column that had no skip index), with EXPLAIN INDEXES evidence backing the diagnosis and the index key correctly matching the actual query predicate ((TenantId, ScenarioRunId) point lookup). Down migration is properly commented out per repo convention, migration-order CI passes at 00063, and the new integration test exercises real ClickHouse (testcontainers) covering index creation, latest-status-wins, partition isolation, missing-run, and tenant isolation. CodeRabbit found nothing actionable. The MATERIALIZE INDEX backfill is correctly scoped out as an ops follow-up (langwatch-agent's own comment confirms no change requested, just flags to pair the merge with scheduling the backfill). CI green, no P0/P1 findings.

Aryansharma28:
Cross-referencing: filed https://github.com/langwatch/langwatch/issues/6329 to own/track the deferred `MATERIALIZE INDEX idx_scenario_run_id` backfill for this PR (owner: @Aryansharma28, ~2 weeks post-merge) — it also covers #5810's identical `idx_evaluation_id` backfill, since both PRs defer the same heavy prod-mutating step for the same reason.

Also flagging a merge-order dependency: this migration is numbered `00063`, deliberately left past main's current head (`00061`) because #5810 claims `00062` for the same fix pattern on `evaluation_runs` and is still open. `goose` runs without `-allow-missing`, so if `00063` merges/deploys before `00062`, applying `00062` afterward will fail rather than skip. **This PR should merge after #5810, not before.**

Aryansharma28:
Update: consolidated backfill tracking onto https://github.com/langwatch/langwatch/issues/5864 instead of #6329 (closed as duplicate) — langwatch-agent's own issue, filed earlier, broader evidence, already lists this table's `idx_scenario_run_id` backfill as a prioritized follow-up. Now assigned to @Aryansharma28.

Also independently re-verified this migration end-to-end in an isolated worktree, fresh eyes (not trusting the earlier approval): checked out the branch, ran the real integration test against native ClickHouse (4/4 passed, not mocked), confirmed the migration applies cleanly. Functionally verified, not just reviewed.

Merge-order reminder still stands: this should merge after #5810 (migration 00062 must land before 00063).

PR #5868, opened 2026-07-17
feat(group-queue): blob leases replace holder-set refcounting (ADR-046)

0xdeafcafe:
**Stacked on #5867** (the queue maintenance sweep) - review this one for the lease change only.

Implements [ADR-046](dev/docs/adr/046-groupqueue-blob-leases.md): the GroupQueue's blob store stops reference-counting and moves to time-based leases. Net **-1,698 lines**, and the queue's most defect-prone subsystem (the 2026-07-09 phantom-hold leak, the 07-11 fire-and-forget fixes, the acquire-after-stage race, the drain/DLQ hold leaks, the DLQ-replay body loss) is deleted as a class rather than patched again.

## The change

The holder-set refcount existed for one thing: reclaiming a redis-tier blob the moment its last job retired. Keeping that count consistent required every lifecycle transition - stage, complete, retry, dedup squash, drain, DLQ move - to agree on hold state across two processes and three Lua scripts. Eager reclaim has no correctness role (content-addressed PUTs are idempotent; missing bodies terminate at the counted drop path), so the whole plane comes out:

- **Leases, not refcounts.** A redis-tier blob gets a lease at PUT (default = the old 4-day backstop, `LANGWATCH_GQ_BLOB_LEASE_SECONDS` to retune), renewed to full on every read (the existing GETEX), and effectively renewed by every retry re-encode (idempotent re-PUT of the same content-addressed key). Worst-case retention is therefore unchanged by this cutover - what we give up is only the eager reclaim of the happy path.
- **Block and DLQ extend leases.** `RESTAGE_AND_BLOCK`, `MOVE_TO_DLQ` and `REPLAY_FROM_DLQ` parse each staged value's ref in Lua and `EXPIRE ... GT` the blob to at least the 7-day DLQ retention. This structurally closes the DLQ-replay body loss (7-day DLQ vs 4-day blob) and is strictly better than refcounts for blocked groups, whose holds nothing ever refreshed.
- **No application-side s3 deletes.** The bucket lifecycle rule is the reclaim mechanism (it already was the backstop the code leaned on); ops need to provision it - see the rollout notes.
- **GQ1 writes retired.** The `GROUP_QUEUE_ENVELOPE_WRITES_ENABLED` gate is gone; writes are unconditionally GQ2, with a loud inline-GQ2 downgrade (never GQ1) when the tiered store or tenant is missing. Readers keep accepting GQ2, GQ1 and bare JSON for in-flight values.

## What gets deleted

`blobHolders.ts` and both its Lua scripts; the squash hold-transfer helpers and the orphanedValue/reclaimS3 return channel in STAGE/STAGE_BATCH; acquire/release/transfer/reclaimOrphanedS3 and the release-side tenant guards in the lifecycle; every hold call site and ordering subtlety in the processor ("acquire-before-release", the 2026-07-11 awaited fixes); the `h` token in the wire format; the holder-outlives-blob TTL invariant; the GQ1 write path; the holder test suites.

## Rolling deploy + rollback

No flag. New writers emit token-less GQ2 headers; an old worker's release of a token-less value is a no-op by construction. Old tokened values processed by new workers never release and simply TTL out (one bounded memory tail). Rollback is a revert - old readers GETEX blobs back onto the old TTL.

## Rollout notes (ops)

1. **Sizing sweep** before and after: `SCAN` the `:gq:blob:` prefix + sampled `MEMORY USAGE`. The 2026-07-09 incident is the anchor (multi-day accumulation of leaked squash blobs = 1.9 GB), so a 4-day lease over all traffic should land single-digit GB. Retune `LANGWATCH_GQ_BLOB_LEASE_SECONDS` down (24h) once confirmed; `S3_TIER_THRESHOLD_BYTES` is the pressure valve.
2. **Provision the bucket lifecycle rule** on the gq object prefix (14 days recommended, >= DLQ retention). Self-hosted file driver: disk is the backstop; a janitor cron can follow.
3. Requires Redis >= 7 (`EXPIRE ... GT`) - dev/test verified; confirm ElastiCache engine version.

## Verification

- 141 unit + 241 integration tests green on the queue/ops surface; the full event-sourcing integration suite (344 tests) green.
- New coverage: lease at PUT, renewal on read, no reclaim on completion, block/DLQ/replay lease extension (including EXPIRE GT never shortening), inline-GQ2 downgrade, GQ1 read-compat via hand-crafted legacy values.
- Specs updated to the lease contract (`payload-store-blob-lease.feature` is new); ADR-029/030 carry amendment notes.


0xdeafcafe:
Superseded by #5910, which is the same blob-lease work re-cut against `main` rather than stacked on #5867's maintenance-sweep branch. 23 of these files reappear there.

Closing the older of the two so there is one lease PR to review, not two. #5910 still carries its P1 (S3-tier queue payloads orphaned under a key prefix shared with customer stored objects) — that is now addressed by langwatch-saas#758, which moves them to a dedicated bucket.

PR #5882, opened 2026-07-17
feat(experiments): add a proper editor drawer for workflow-agent targets

Aryansharma28:
## Summary

Stacked on #5871. Editing a workflow-agent target in the Experiments Workbench (via the column's "Edit Agent" menu item) previously just called `window.open` — no drawer, no way to map dataset columns to the workflow's input fields, unlike every other agent target type (code/HTTP), which get a drawer with a mapping UI.

This adds `AgentWorkflowTargetEditorDrawer`:

- A "Workflow" card showing the linked workflow's icon, name, and last-updated time, using the same `WorkflowCardDisplay` component `EvaluatorEditorShared.tsx` already uses for a workflow-backed evaluator — so this reads consistently with the rest of the app instead of a bespoke box. Clicking it opens the real Studio graph editor in a new tab, kept deliberately: a full multi-node graph editor can't be edited meaningfully inside the workbench's narrow sidebar, unlike the code/HTTP/prompt editors that do fit there.
- Below it, the same `VariablesSection` mapping UI code/HTTP agent targets already get, fed by the workflow's **real** entry-node input fields — reusing `getMappingSurfaceInputs`, the same extraction `AgentWorkflowEditorDrawer` (the Scenarios-context version) already uses, so this isn't a second copy of that logic.
- Mappings persist immediately through the existing `setTargetMapping`/`removeTargetMapping` flow-callback wiring, matching how code/HTTP target mappings already work — no separate "Save" step needed, since a workflow-type agent has no inline-editable config (no code, no URL) to save.

Also applied the same `WorkflowCardDisplay` treatment to `AgentWorkflowEditorDrawer.tsx` (the Scenarios-context sibling drawer this pattern was originally copied from), for visual consistency between the two — same plain-box-to-card swap, same testid fix (see below).

### Before / after

**Before** (plain bordered box, no card):

before

**After** (shared `WorkflowCardDisplay`):

after

### A real bug found by running against real infra, not just typechecking

The original `ChakraLink asChild` + `NextLink` composition (copied from the pre-existing `AgentWorkflowEditorDrawer.tsx` pattern) silently dropped `data-testid` from the rendered anchor — Chakra's `asChild` slot only forwards style-related props to the composed child, not arbitrary data attributes. Typecheck and a mocked render both looked fine; only running the integration test against real local infra (native ClickHouse/Redis/Postgres, via the project's `LANGWATCH_TEST_*` native-services mode — no docker testcontainers needed) surfaced it, since the assertion on the testid genuinely failed to find the element. Fixed by rendering a plain `Link isExternal` directly instead — `target="_blank"` is always a hard navigation into a new tab, so there was nothing to gain from routing it through the app's client-side router in the first place.

`AgentWorkflowEditorDrawer.tsx` had the identical bug on its own `open-workflow-editor-link` testid (untested there too, so nobody had hit it) — fixed there as well while applying the same card styling.

## Test plan

- [x] `specs/agents/workflow-agent-as-target.feature` — updated with two new scenarios covering the drawer + card + mapping behavior
- [x] `useOpenTargetEditor.unit.test.tsx` — updated to assert `agentWorkflowTargetEditor` opens (replacing the old window.open assertion)
- [x] `AgentWorkflowTargetEditorDrawer.integration.test.tsx` — renders the workflow card + open-editor link (with the correct href), renders the workflow's real input fields (falls back to a generic "input" field when the workflow has no declared entry outputs), no Save button since mappings persist immediately — all 4 tests run and pass against real local infra (native ClickHouse/Redis/Postgres via `LANGWATCH_TEST_CLICKHOUSE_URL`/`LANGWATCH_TEST_REDIS_URL`/`LANGWATCH_TEST_DATABASE_URL`, not testcontainers)
- [x] `AgentWorkflowEditorDrawer.unit.test.tsx` + `.save-gate.integration.test.tsx` — existing tests for the Scenarios drawer still pass after the card-styling/testid-fix change
- [x] Typecheck clean (verified no new errors beyond the pre-existing, unrelated `saas-src/` gap)
- [x] Broader experiments-v3/agents unit suite: 682 tests pass, no regressions
- [x] Manually verified end-to-end in a local dev environment: opened the drawer for a real workflow-agent target, confirmed the card renders correctly, confirmed clicking it opens Studio in a new tab while the drawer/workbench state stays intact in the original tab

### DIFF
--- a/langwatch/specs/agents/workflow-agent-as-target.feature
+++ b/langwatch/specs/agents/workflow-agent-as-target.feature
@@ -17,12 +17,24 @@ Feature: Workflow agent as an experiment target
     Given the workflow agent is added as a target in the Experiments Workbench
     Then the target column shows a workflow icon, not a code icon
 
-  Scenario: Editing the target opens the underlying Studio workflow
+  Scenario: Editing the target opens a mapping drawer, not a dead end
     Given the workflow agent is added as a target in the Experiments Workbench
     When the user opens the target's edit menu and selects Edit Agent
-    Then the linked Studio workflow opens in a new tab
-    And the workbench keeps its own state, since a full graph editor
-      cannot be edited meaningfully inside a narrow sidebar drawer
+    Then a sidebar drawer opens showing the linked workflow's name
+    And an "Open Workflow" action in that drawer opens the Studio graph
+      editor in a new tab, since a full graph editor cannot be edited
+      meaningfully inside a narrow sidebar
+    And below it, the drawer shows the workflow's real input fields
+      with mapping controls, matching the mapping UI code and HTTP
+      agent targets already get
+
+  Scenario: Mapping a dataset column to a workflow input field
+    Given the workflow agent target's drawer is open
+    And the underlying workflow declares an input field named "question"
+    When the user maps "question" to a dataset column
+    Then the mapping is saved immediately, without a separate save step
+    And running the experiment passes that column's value into the
+      workflow's "question" input
 
   Scenario: Switching away from a workflow target
     Given the workflow agent is added as a target in the Experiments Workbench
--- a/langwatch/src/components/agents/AgentWorkflowEditorDrawer.tsx
+++ b/langwatch/src/components/agents/AgentWorkflowEditorDrawer.tsx
@@ -5,16 +5,16 @@ import {
   Heading,
   HStack,
   Input,
-  Link as ChakraLink,
   Spinner,
   Text,
   VStack,
 } from "@chakra-ui/react";
-import NextLink from "~/utils/compat/next-link";
+import { ExternalLink } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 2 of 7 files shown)


Aryansharma28:
Polished the "Workflow" card per feedback — swapped the plain bordered name+link row for the same `WorkflowCardDisplay` component `EvaluatorEditorShared.tsx` already uses for a workflow-backed evaluator, so this reads consistently with the rest of the app instead of a bespoke box.

**Before:**

before

**After:**

after

Also fixed a real bug caught by running the new component test against real local infra (native ClickHouse/Redis/Postgres — not testcontainers/docker): the original `ChakraLink asChild` + `NextLink` composition silently dropped `data-testid` from the rendered anchor, since Chakra's `asChild` slot only forwards style-related props to the composed child, not arbitrary data attributes. Fixed by rendering a plain external `ChakraLink`/`Link isExternal` directly instead — there was nothing to gain from routing a `target="_blank"` link through the app's client-side router anyway.

PR #5884, opened 2026-07-17
fix(langy): constrain worker processes

0xdeafcafe:
## What changed

- replace inherited manager environment filtering with an explicit worker allowlist
- start production OpenCode workers through `prlimit` with per-UID process and file ceilings
- keep local development portable by executing workers directly there
- install the small `util-linux` runtime dependency that provides `prlimit`

## Why

Langy workers execute model-driven shell commands. A denylist could leak newly introduced manager secrets, while an individual worker could otherwise fork without bound, exhaust descriptors, write very large files, or produce credential-bearing core dumps.

This is defense in depth for the current process-pool architecture. Hard per-worker memory and CPU isolation remains part of the future one-worker-per-gVisor-pod design.

## Deployment Impact

The Langy Agent runtime image adds Debian's `util-linux` package to provide `/usr/bin/prlimit`. No environment variables, Helm values, APIs, or default feature behavior change. Existing Langy workers restart with the next image deployment and then receive the new per-process limits. Vanilla Helm installs and BYOC dataplanes are otherwise unchanged.

## Validation

- `go test ./services/langyagent/...`
- exercised the exact `prlimit` invocation in the Debian runtime image
- `git diff --check`

## Summary by CodeRabbit

* **Security**
  * Worker processes now receive only explicitly approved environment variables, reducing accidental exposure of secrets and manager settings.

* **Reliability**
  * Worker commands now honor context cancellation and timeouts.
  * Sandboxed workers continue to run with enforced per-process resource limits.

* **Maintenance**
  * Added required runtime support for process and resource management.

### DIFF
--- a/Dockerfile.langyagent
+++ b/Dockerfile.langyagent
@@ -161,7 +161,7 @@ FROM debian:bookworm-slim
 # services/langyagent/internal/assets/AGENTS.md.
 ARG LANGWATCH_CLI_VERSION=0.31.0
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    curl git ca-certificates python3 \
+    curl git ca-certificates python3 util-linux \
     && rm -rf /var/lib/apt/lists/*
 # NOTE: `iptables` is no longer installed. The old loopback OWNER-match DROP
 # rule (services/langyagent/iptables.go) that closed the sibling-worker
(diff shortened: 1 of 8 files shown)


langwatch-agent (commented):
No blocking findings. The allowlist removes inherited manager secrets by default, and the production-only `prlimit` wrapper preserves the sandboxed runner's per-UID credential and process-group behavior. Verified the exact util-linux invocation locally, `git diff --check`, and `go test ./services/langyagent/...`.

Aryansharma28 (approved):

PR #5886, opened 2026-07-17
feat: add canonical log pipeline

0xdeafcafe:
## What changed

Adds a canonical OpenTelemetry log pipeline on top of the metric work.

- Preserves canonical log bodies, attributes, resource/scope data, severity, timestamps, and trace/span correlation in ClickHouse.
- Adds usage estimates, retention, TTL reconciliation, replay, and cold-scan coverage.
- Correlates useful Claude Code and Codex log data into trace summaries without keeping the legacy lossy write path.
- Batches accepted log records and trace contributions, then uses one bulk projection write per coalesced tenant batch.

## Why

The existing log path flattens data for trace-oriented storage and loses useful OTLP structure. This gives logs an immutable canonical source while keeping trace summaries focused on the correlated data we actually query.

## Stack

- #5885 adds reusable event-sourcing batching.
- #5851 adds the canonical metric pipeline and live metric batching.
- This PR contains only the log-specific layer.

## Validation

- 107 focused log, trace, ClickHouse, retention, and migration tests pass.
- TypeScript typecheck passes.
- Biome 2.4.14 passes for the final touched files; its known large dependency type warning is non-blocking.


langwatch/src/server/clickhouse/migrations/00050_create_canonical_logs.sql
```diff
@@ -0,0 +1,106 @@
+-- +goose Up
```
langwatch-agent, line 1: [P1] This migration will not run after the current base branch. `feat/canonical-metric-pipeline` already ends at migration 47, but this adds `00043`; Goose/migration-order therefore treats it as out of order, and deployed databases will be missing both new tables while the application starts writing `log_records`. Please rename it to the next sequence (`00048` per the CI diagnostic) before this can merge.

langwatch-agent (changes requested):
## Findings

[P1] The canonical-log schema migration is out of order on this PR’s declared base. CI confirms `00043_create_canonical_logs.sql` is below the base branch’s latest migration (`00047`), so an environment that has already applied that base will never create `log_records` or `log_usage_estimates`. Rename it to the next available sequence (CI suggests `00048_create_canonical_logs.sql`) and rerun migration-order.

The PR is also not merge-ready while CI is red: `migration-order` fails for the above reason, and `test-unit (3)` fails in `acceptAgentTurn.commands.unit.test.ts`. The latter does not appear related to this diff but still needs to be reconciled before merge.

langwatch/src/server/event-sourcing/pipelines/log-processing/canonicalLog.ts
```diff
@@ -0,0 +1,674 @@
...
+  piiRedactionLevel: PIIRedactionLevel;
+  redactionService: LogRedactionService;
+  acceptedAt?: number;
+}): Promise<CanonicalLogPreparationResult> {
+  const accepted: PreparedCanonicalLogRecord[] = [];
+  const errors: string[] = [];
+  let rejectedLogRecords = 0;
+  const acceptedAt = args.acceptedAt ?? Date.now();
+
+  for (const resourceLogRaw of args.request.resourceLogs ?? []) {
```
langwatch-agent, line 596: **[P2] Malformed OTLP containers escape as a 500 instead of a controlled rejection.** This assumes `resourceLogs` is iterable; a JSON OTLP request with `resourceLogs: {}` reaches this code because the JSON parser casts rather than validates, then throws `TypeError: object is not iterable`. `scopeLogs` has the same issue on line 602, while a malformed `logRecords` array is silently treated as empty. The route does not catch this preparation failure, so a sender gets an internal error rather than the advertised partial-success/permanent-rejection result. Use the same validated container helper as the canonical metrics path, count malformed containers deterministically, and add JSON-route regression tests for malformed `resourceLogs`, `scopeLogs`, and `logRecords`.

langwatch/src/server/event-sourcing/pipelines/log-processing/canonicalLog.ts
```diff
@@ -0,0 +1,674 @@
...
+
+const isRecord = (value: unknown): value is UnknownRecord =>
+  value !== null && typeof value === "object" && !Array.isArray(value);
+
+function sha256(value: string): string {
+  return createHash("sha256").update(value).digest("hex");
+}
+
+function longBitsToBigInt(value: UnknownRecord): bigint {
+  const low = BigInt(Number(value.low ?? 0) >>> 0);
```
langwatch-agent, line 77: **[P2] Do not coerce malformed protobuf-style 64-bit values to zero.** `Number(value.low) >>> 0` and the equivalent `high` conversion turn non-numeric/out-of-range halves into a valid wrapped integer. I verified a log with `timeUnixNano: { low: "nope", high: "bad" }` is accepted with `timeUnixNano: "0"` and its timestamp silently replaced by the server receipt time. That changes the event while reporting success, breaking the canonical/lossless contract and misordering it in storage. Validate both halves as finite 32-bit integers before conversion (or reuse the hardened metric integer parser), reject malformed values, and add a regression test.

langwatch-agent (changes requested):
## Review summary

Requesting changes for two malformed-input integrity paths in the new canonical log receiver. Both are directly reproducible against this head, and neither is covered by the otherwise-clean focused suite.

Validation run locally: focused canonical-log/repository/route/migration/retention tests (35 passed), plus production and test TypeScript typechecks (passed). CI's migration-order and application checks are green on the current head.

0xdeafcafe:
Superseded by #5945, which merges this and #5851 into a single branch rebased onto current `main`.

This branch forked before #5851's last three commits, so on its own it was missing the metric-side retryable-persistence fix. #5945 carries all eight commits from here on top of the complete metric branch.

Two things changed on the way in, both covered in #5945's description: the `logRecordStorage` projection is registered again (this branch kept `recordLog` registered but deleted the projection consuming it, so a pre-cutover instance mid-rolling-deploy would have had records reach neither table), and the trace-contribution build failure no longer decrements `acceptedLogRecords` — that path contradicted the comment directly below it and told collectors to discard logs we had already stored.

Leaving this open until #5945 lands, so there is still a fallback.

PR #5891, opened 2026-07-17
feat(group-queue): make batching observable and safer

0xdeafcafe:
## Summary

- count dispatched and completed payloads correctly when one handler processes a batch
- expose handler invocation count and an actual batch-size histogram
- re-stage failed coalesced siblings with one atomic Redis operation
- never resurrect a sibling that was already dropped during parsing
- settle all sibling parses before deciding what can safely be re-staged

## Why

Batching is reducing handler work, but the existing metrics count handler calls as jobs and do not show the batch-size distribution. The failure path also turns one failed batch into one Redis write per sibling and can re-stage work that was already terminally dropped.

This makes the batching efficiency measurable and keeps recovery to one Redis operation without changing group ordering or at-least-once delivery.

## Checks

- 61 focused unit tests
- test TypeScript check
- repository-pinned Biome check
- Redis integration: atomic sibling re-stage
- Redis integration: handler failure after terminal sibling drop
- Redis integration: mixed terminal and transient sibling parse failures

### DIFF
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
@@ -1055,8 +1055,16 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
 
               // Success — complete the group slot. Drained siblings were
               // removed from staging during the drain, so completing the
-              // dispatched job is enough to free the group.
-              await this.scripts.complete({ groupId, stagedJobId, jobName });
+              // dispatched job is enough to free the group. It is also the only
+              // completion the siblings get, hence payloadCount: the Redis
+              // counters ops reads must advance per payload, like the Prometheus
+              // one below, not once per handler call.
+              await this.scripts.complete({
+                groupId,
+                stagedJobId,
+                jobName,
+                payloadCount: handlerPayloadCount,
+              });
               await this.blobLifecycle.release({
                 values: [
                   jobDataJson,
--- a/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
+++ b/langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
@@ -55,11 +55,14 @@ import {
 import {
   gqGroupsBlockedTotal,
   gqGroupsPoisonParkedTotal,
+  gqHandlerBatchSize,
+  gqHandlerInvocationsTotal,
   gqJobDelayMilliseconds,
   gqJobDurationMilliseconds,
   gqJobsCompletedTotal,
   gqJobsDedupedTotal,
   gqJobsDelayedTotal,
+  gqJobsDispatchedTotal,
   gqJobsDroppedTotal,
   gqJobsExhaustedTotal,
   gqJobsNonRetryableTotal,
@@ -861,6 +864,12 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
           nowMs: Date.now(),
           maxJobs: maxBatch - 1,
         });
+        if (drainedSiblings.length > 0) {
+          gqJobsDispatchedTotal.inc(
+            { queue_name: this.queueName },
+            drainedSiblings.length,
+          );
+        }
       } catch (err) {
         this.logger.warn(
           {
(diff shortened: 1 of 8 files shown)


langwatch-agent (commented):
Reviewed the complete diff against `feat/event-sourcing-map-batching` and the preceding decode-drop hardening.

No blocking findings.

The recovery path now waits for every sibling parse to settle, marks terminal drops before re-stage selection, and the one Lua `stageBatch` call restores only non-dropped siblings while the group active key preserves ordering. The payload-level completion metric is correctly separated from handler-invocation and batch-size measurements.

Validation run locally:
- `metrics.unit.test.ts`: 23 passed
- Group-queue Redis integration: 31 passed, including terminal-drop + handler-failure and mixed terminal/transient parse scenarios
- `pnpm run typecheck:tests`: passed

The repository CI was still running at review time; no failed checks were present.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -1040,7 +1064,7 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
                 ],
```
langwatch-agent, line 1072: **P2 — Redis completion counters still count handler calls, not payloads**

`gqJobsCompletedTotal` is corrected on this line, but `complete()` still invokes `COMPLETE_LUA` without batch cardinality. That script does one `INCR` for `stats:completed` and `stats:completed:<jobName>` per call. A successful coalesced batch of 50 is therefore 50 in the new Prometheus counter but 1 in the Redis/Skynet counters.

Those Redis counters are consumed by `metrics-collector.ts` for completed throughput and per-job totals, so operational views will still under-report batch work and disagree with Prometheus. Thread `handlerPayloadCount` through `GroupStagingScripts.complete`, use `INCRBY` in `COMPLETE_LUA` (defaulting to 1), and add an integration assertion that a successful N-payload batch advances both Redis counters by N.

langwatch-agent (changes requested):
Request changes: the new Prometheus counter reports every payload in a coalesced batch, but the Redis-backed completion counters that drive the operational metrics collector still report one completion per handler invocation. The recovery and atomic re-stage paths otherwise held under the focused Redis integration suite.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
0xdeafcafe: Good catch — fixed in bb8197e87.

`complete()` now takes a `payloadCount` (default 1) which `COMPLETE_LUA` reads as `ARGV[8]` and applies with `INCRBY` to both `stats:completed` and `stats:completed:<jobName>`. The success path passes `handlerPayloadCount`, the same value `gq_jobs_completed_total` uses, so Prometheus and the Redis/Skynet counters now agree on a coalesced batch.

The two `dropped: true` call sites are unaffected — that branch is still skipped entirely, so a discard counts nothing regardless of the count passed.

Tests added:
- `scripts.integration.test.ts` — a completion with `payloadCount: 7` advances both counters by 7; an omitted count advances both by 1; a dropped completion leaves both at 0 even with a count of 7.
- `groupQueue.integration.test.ts` — end-to-end, 12 events coalesced through the real dispatch loop advance `stats:completed` and `stats:completed:recordSpan` by 12, with an assertion that a multi-event batch actually occurred so the test can't pass on 12 singleton completions.
- `specs/event-sourcing/fold-coalescing.feature` gained the corresponding scenarios.

The 133 group-queue unit tests pass locally; I don't have a container runtime up on this machine, so the Redis integration suites are running on CI rather than locally.

0xdeafcafe: CI is green on `7d43248f0` — `langwatch-app-ci` passed in full, including all six integration shards.

The new tests ran against real Redis rather than being skipped: `groupQueue.integration.test.ts` went 17 → 18 tests and `scripts.integration.test.ts` 153 → 156, matching the four added here.

Two unrelated workflows are red on the branch, both environmental:
- `sdk-python-ci` — OpenAI returns `429 insufficient_quota`; the account is out of API credit.
- `e2e-ci` — the app server never bound, so every request 502s with `ECONNREFUSED`. It passed on `bb8197e87` (the commit with the actual code change) and failed on `7d43248f0`, which adds a single `@scenario` comment line to an integration test e2e doesn't run.

One other thing worth flagging separately: the first run also failed `test-unit (3)` on `acceptAgentTurn.commands.unit.test.ts > emits the whole turn boundary as one ordered event batch`, in the Langy conversation pipeline. It passed on re-run and passes locally 12/12 in isolation. Integration tests are excluded from the unit config, so nothing in these commits executes in that shard. The assertion is ordering-sensitive: all four events share a `createdAt`, so `orderEvents` falls through to comparing KSUIDs, and `@langwatch/ksuid` only embeds a second-precision timestamp — so ordering rests entirely on the sequence portion, while the test mocks `Date.now()` back to Nov 2023. Looks like a latent flake that deserves its own issue rather than a fix smuggled into this PR.

0xdeafcafe:
Folded into #5918. `gq_handler_batch_size` and `gq_handler_invocations_total` land there — I had independently added a duplicate batch-size histogram on that branch and removed it in favour of this one.

PR #5892, opened 2026-07-17
feat(langy): remove deprecated api surfaces, nicer gating 

0xdeafcafe:
## What changed

Make Langy genuinely internal-only, behind **one** authoritative access decision.

- **One gate, one decision.** `hasLangyAccess` (staff bypass, else
  `release_langy_enabled`) is the single, transport-free decision. Its one tRPC
  adapter, `enforceLangyAccess`, guards every customer-facing Langy tRPC router.
  A denial is `NOT_FOUND`, so the gate can't be used to probe whether Langy
  exists for an account.
- **Close the sibling-surface gaps.** The first pass only gated the `langy`
  router and the GitHub install route. `langyGithub`
  (`getInstallStatus` / `listRepos` / `disconnect`) and `langyEgress`
  (`get` / `set`) were still reachable by a non-staff customer with the flag
  off — they now sit behind the same gate.
- **Delete the deprecated surface.** `POST /api/langy/chat` (Hono) is removed
  and unmounted. The browser drives turns through the
  `langy.createConversation` / `continueConversation` tRPC mutations; nothing
  else called it. Its route-only tests go with it.
- **No drift with the UI.** `LANGY_RELEASE_FLAG` is shared between the server
  gate and the `useShowLangy` client hook, which stays UI-hiding only.

## Why

OpenCode workers currently share the Langy manager pod's network namespace.
Until workers run in separate network namespaces, Langy must remain
internal-only and UI hiding is not an adequate security boundary — so the
boundary is this server-side gate. Verified LangWatch staff keep the explicit
bypass; the SaaS deployment pins `release_langy_enabled` off in production
(default is already off in the registry).

## Impact

Every customer-facing Langy surface — conversations, turns, GitHub integration,
egress config — is dark for non-staff accounts while the flag is off. Staff are
unaffected. The deprecated `/api/langy/chat` endpoint no longer exists.

## Validation

- `hasLangyAccess` decision suite (5) + `enforceLangyAccess` adapter suite (4) +
  GitHub install suite (11) pass locally.
- Confirmed the client transport calls the tRPC turn mutations, not
  `/api/langy/chat`, before deleting the route.
- Full typecheck not run here (the isolated worktree reuses another checkout's
  generated Prisma/package artifacts); the changes follow the existing
  `checkProjectPermission` / `authorizeInResolver` middleware pattern exactly.

## Summary by CodeRabbit

* **New Features**
  * Centralized Langy access gating now protects Langy conversations, GitHub install/setup flows, and Langy data operations using a shared rollout flag with a staff bypass.
  * Frontend Langy visibility now uses the same shared rollout key as the server.
* **Bug Fixes**
  * Denied requests are consistently stopped with the same safe “not enabled” outcome before protected operations proceed.
  * GitHub-related procedures now enforce org membership and Langy access in the correct order.
* **Tests**
  * Added unit tests covering access decisions, middleware behavior, and project/org scoping; removed outdated Langy chat route tests.
* **Chores**
  * Removed the legacy Langy chat endpoint and its old routing.


langwatch/src/server/routes/github-langy.ts
```diff
@@ -32,9 +32,9 @@ import {
   publicEndpoint,
 } from "~/server/api/security";
 import { getApp } from "~/server/app-layer";
+import { hasLangyAccess } from "~/server/app-layer/langy/langyAccessGate";
```
langwatch-agent, line 35: **P1 — Re-check Langy access before persisting the GitHub installation**

This new shared gate only authorizes `/install`. `/setup` (lines 245-286) verifies the session and membership but never calls `hasLangyAccess`, so a customer who begins while the flag is enabled can complete after the rollout is disabled (or their access is revoked) and still have `recordInstallation` persist the Langy GitHub App connection. That contradicts the stated internal-only boundary and makes a feature-flag kill/revocation non-immediate for this customer-facing path.

Re-evaluate `hasLangyAccess({ user: session.user, organizationId: state.organizationId })` after the membership check and before `recordInstallation`, then add a setup test proving a now-denied caller cannot persist an installation. The nonce should still be consumed so the signed state cannot be retried later.

langwatch/src/server/routes/github-langy.ts
```diff
@@ -184,19 +183,13 @@ secured
...
     }
-    // Gate by release_langy_enabled (staff bypass) so the install can't be
-    // started before the assistant is rolled out for the user.
-    if (!isLangwatchStaff(session.user)) {
-      const allowed = await featureFlagService.isEnabled(
-        "release_langy_enabled",
-        { distinctId: session.user.id },
+    // Same authoritative gate as Langy's tRPC surface so the GitHub install
+    // cannot become a rollout bypass.
+    if (!(await hasLangyAccess({ user: session.user }))) {
```
langwatch-agent: **P2 — Preserve organization scope when evaluating the install gate**

The request already supplies `organizationId` two lines later, but the gate is evaluated as user-only. `release_langy_enabled` explicitly supports organization targeting, and the tRPC `langyGithub` procedures forward that organization ID through `enforceLangyAccess`. An organization-only rollout therefore permits the settings procedures but rejects the required `/install` redirect; conversely, this path does not apply an organization-specific denial consistently.

Parse and validate `organizationId` before the gate and call `hasLangyAccess({ user: session.user, organizationId })` (while retaining the existing membership check). Add an install-route test for an organization-targeted allow/deny decision.

langwatch-agent (dismissed):
Request changes: 1 P1 and 1 P2. The new central gate is not reapplied when GitHub completes an installation, and the install route drops the organization scope needed for supported org-targeted rollouts. Details are inline.

langwatch/src/server/api/routers/__tests__/langy.access-gate.unit.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it("denies non-staff before reading Langy data when the flag is off", async () => {
+    isEnabled.mockResolvedValue(false);
+
+    await expect(
+      callerFor({ id: "customer-1", email: "user@example.com" }).list({
+        projectId: "project-1",
+      }),
+    ).rejects.toMatchObject({ code: "NOT_FOUND" });
```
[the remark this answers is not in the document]
0xdeafcafe: Outdated — `langy.access-gate.unit.test.ts` was removed in the follow-up refactor (1fd8ff1) that unified the gate into `langyAccessGate` + the `enforceLangyAccess` tRPC middleware. That middleware throws `TRPCError({ code: 'NOT_FOUND' })` on denial, and `langyAccessMiddleware.unit.test.ts` asserts exactly that (the earlier INTERNAL_SERVER_ERROR came from the now-deleted file). Nothing to fix; resolving.

langwatch/src/server/routes/github-langy.ts
0xdeafcafe: Fixed in 5a5ba2f. `/setup` now re-runs `hasLangyAccess({ user: session.user, organizationId: state.organizationId })` after the membership check and before `recordInstallation`, so the rollout kill-switch (or a revocation) is immediate on the persisting path. The nonce is already burned by that point, so a denied caller can't retry the signed state. Added a setup test proving a now-denied caller can't persist an installation.

langwatch/src/server/routes/github-langy.ts
0xdeafcafe: Fixed in 1ee4fff. `organizationId` is now parsed before the gate and forwarded to `hasLangyAccess` on both `/install` and `/setup`, matching how the tRPC `langyGithub` procedures pass it through `enforceLangyAccess`. Kept the 404 ahead of the 400 org-required check so a denied account can't probe the endpoint via param validation. Added an org-targeted allow/deny install test that asserts the org id reaches the flag.

langwatch/src/server/api/routers/langyGithub.ts
```diff
@@ -53,6 +57,7 @@ export const langyGithubRouter = createTRPCRouter({
   getInstallStatus: protectedProcedure
     .input(z.object({ organizationId: z.string() }))
     .use(authorizeInResolver)
+    .use(enforceLangyAccess)
```
langwatch-agent, line 78: **P2 — Prove organization membership before evaluating its rollout flag**

`authorizeInResolver` only marks the deferred permission check as satisfied; it does not authorize the organization. The real membership check is inside the resolver, after this middleware. Consequently, a signed-in non-member can call `getInstallStatus`/`listRepos`/`disconnect` with a guessed organization id: if that org is enabled the gate passes and `ensureOrganizationMember` returns `FORBIDDEN`; if it is disabled the gate returns `NOT_FOUND`. This leaks an arbitrary tenant’s Langy rollout state, contrary to the stated non-probing boundary. The REST `/github-langy/install` path has the same ordering.

Make organization membership an actual shared middleware and run it before `enforceLangyAccess` on all three procedures; on the REST path, check membership before evaluating an organization-scoped flag. Add allow/deny tests for a non-member showing the response is independent of that org\u0027s flag value.

langwatch-agent (dismissed):
Request changes: 1 P2 tenant-boundary leak. The organization-scoped Langy gate is evaluated before membership, allowing response differences to reveal another organization’s rollout state. Details inline.

langwatch/src/server/api/routers/langyGithub.ts
0xdeafcafe: Fixed in 745a894. Lifted `ensureOrganizationMember` into an `enforceOrganizationMembership` middleware and chained it **before** `enforceLangyAccess` on all three procedures (`getInstallStatus` / `listRepos` / `disconnect`), so membership is proven before the org-scoped flag is evaluated. On the REST `/install` path I reordered the checks the same way (membership 403 before the flag 404); `/setup` already checked membership first. A non-member's response is now independent of the org's flag value.

Tests: a new `langyGithub.access-order.unit.test.ts` proves a non-member gets FORBIDDEN with the flag **never consulted** under both flag states, plus member allow (org scope forwarded) / deny (NOT_FOUND); matching non-member flag-independence tests were added to the REST `/install` suite.

Also fixed the failing `test-integration` shard 5: `ProjectLangyLayout`'s test mocked `~/utils/isLangwatchStaff` without the `LANGY_RELEASE_FLAG` export that `useShowLangy` now reads (2cd70c0).

langwatch-agent (commented):
Re-review complete at `745a8948b4313e0fa6730769d43897dbbc58df9e`.

The prior access-control findings are addressed:
- `/setup` now re-evaluates the org-scoped Langy gate after membership verification and before persisting an installation.
- `/install` forwards `organizationId` to the gate.
- The tRPC GitHub procedures and REST install route establish organization membership before evaluating an organization-scoped rollout flag, eliminating the cross-tenant rollout-state probe.

Validation run locally:
- focused Langy access, GitHub install, and UI suites: 29 passed
- `pnpm typecheck`: passed

No remaining actionable findings from this re-review. I am dismissing my prior change requests; this is not an approval.

Aryansharma28 (approved):

PR #5904, opened 2026-07-17
feat(governance): event-source ingestion pull scheduling

0xdeafcafe:
> Stacked on #5902 and intentionally targets `feat/es-topic-clustering`.

## Why

The pull-mode ingestion slice from #4990 (reviewed only through `eb1da000d80281594604f3fbba5b696c6796bffd`) used BullMQ repeat jobs as the scheduling authority and updated cursors/status directly from the worker. This moves the workflow onto the durable process-manager/event-sourcing style introduced for Langy and topic clustering.

## What changed

- adds an `ingestion_pull` event pipeline with configure/disable/run-completed/run-failed commands and events
- keeps the enterprise pipeline, projection, process/effect wiring, identifiers, and boot reconciliation under `/ee`; the shared registry only composes the exported enterprise pipeline and identifier sets
- adds one durable `ingestionPull` process per source, with UTC cron wakes, a one-in-flight-run guard, one-slot catch-up after downtime, and deterministic outbox intents
- runs pull effects through the process outbox with bounded concurrency, leasing, retries, and durable failure outcomes
- adds a rebuildable Postgres run-status/cursor projection and transactionally mirrors compatibility fields onto `IngestionSource`
- syncs source create/update/archive changes into process commands and reconciles scheduled/existing processes on worker boot, scoped through hidden governance project IDs
- removes the ingestion BullMQ repeat queue and worker bootstrap
- source-qualifies OCSF event IDs so at-least-once retries from different same-type sources cannot collide
- adds the projection migration plus process/effect/projection/lifecycle/outbox coverage and behavioral specs

## Important semantics

- Postgres process state/outbox is authoritative; Redis loss cannot erase a wake or pending pull.
- Cursor advancement happens only after a successful completion event.
- Partial OCSF insertion failure retries from the same durable cursor.
- Provider failures retry three times, then become a visible `run_failed` event.
- Disabling or archiving a source clears its wake, and late outcomes cannot reactivate the compatibility row.
- Missed cron slots collapse to one catch-up pull; they do not hot-loop through every missed occurrence.

## Validation

- `pnpm run typecheck:all`
- affected puller, process-manager, outbox, wake, topic-clustering, projection, and tenancy-guard unit suites
- `pnpm exec prisma validate`
- `git diff --check`

The broad repository unit command was also sampled for about 18 minutes without a reported failure, then stopped because the complete affected suites above were already green.

## Deployment Impact

**Env vars:** none added, removed, or changed; `.env.example` untouched.

**Helm values:** none added or changed. The chart's bundled alert-rules file (`charts/langwatch/files/alerting-rules.yml`, mechanism from ADR-054 on the base PR) gains a `langwatch-governance-ingestion` group — two rules (`IngestionPullFailedFinal`, `IngestionPullNothingCompleting`), promtool-validated, controlled by the existing `prometheus.alerting.enabled` toggle. No receiver routing changes.

**Default `helm install`, no overrides:** one new Prisma migration (`20260718110001_add_ingestion_pull_run_projection`) creates the read-model table on the app's first boot — renumbered after the base branch's newest migration so ordering holds on already-migrated databases. Ingestion pulls are an enterprise (ee) feature; deployments without configured pull sources run the process manager idle with zero scheduled wakes.

**BYOC dataplane:** no impact — no new services, ports, images, or egress; pull adapters run in the existing workers and reach only the sources operators configure.


0xdeafcafe:
Rebased onto `rebase-5902-port` (PR base retargeted accordingly) and fixed three defects in the ingestion-pull process manager. Head is now `a3d1adc5b5`.

**Rebase reconciliation** (`53248c204b`)
- The old topic-clustering commits this branch was stacked on are superseded by the ADR-052 port already in `rebase-5902-port`; only the three governance commits were replayed.
- `ProcessInput` reconciliation: the substrate now carries `now` on both variants, so `input.handledAt ?? input.scheduledFor` became `input.now` — the fallback would have silently un-clamped late wakes. Every `handledAt` occurrence is gone.
- The outbox option `dispatchConcurrency` was renamed to the substrate's `concurrency` (same semantics, the base's implementation with the bounded worker pool wins).
- Topic clustering stays runtime-owned (`withProcessManager`, ADR-052); the registry's hand-rolled wake worker now scans only the enterprise process managers, so the two never double-claim.

**Defect fixes** (each with a red-first regression test)
- `5a283307c6` — RUN_COMPLETED wrote `cursor: view.cursor` unconditionally, so a late completion from a superseded run regressed the live cursor and re-ingested its window. The cursor now only advances when the completing `runId` matches the tracked run (same identity guard the state clear used).
- `31b634b500` — the event branch scheduled the next wake from `occurredAt` alone while the wake branch clamped; an event handled hours late wrote a `nextWakeAt` already in the past. Both branches now schedule from `max(occurredAt, now)`, mirroring topic clustering's `schedulingRef`.
- `a3d1adc5b5` — `assertValidPullSchedule` threw inside `evolve`, so one committed CONFIGURED event with a bad cron would poison the process subscriber forever. The configure command's schema now rejects any cron the scheduler cannot evaluate (five-field + croner check); `evolve` degrades on an invalid or missing cron — previous state kept, wake cleared until a valid reconfiguration. The event read schema stays permissive so pre-guard history cannot poison replays or projections.

Verified: puller + ingestion-pull suites, full `ee` (100 files / 1082 tests), `src/server/event-sourcing` + `src/server/app-layer/topic-clustering` (203 files / 7055 tests), and `pnpm typecheck` exit 0.

langwatch/ee/governance/services/pullers/process-manager/ingestionPullProcess.definition.ts
```diff
@@ -0,0 +1,228 @@
...
+function evolveWake(
+  previousState: IngestionPullProcessState,
+  scheduledFor: number,
+  handledAt: number,
+): Evolution<IngestionPullProcessState> {
+  if (!previousState.enabled || !previousState.cron) {
+    return { state: previousState, nextWakeAt: null, intents: [] };
+  }
+
+  const active =
```
langwatch-agent, line 188: **[P1] Do not launch a second pull while the first one can still be executing**

At 30 minutes this replaces `currentRun` and emits a second intent using the same cursor, but nothing cancels or fences the original effect. The advertised five-minute limit is only the `deadlineMs` hint passed at `pullerWorker.ts:111`; S3 has no transport timeout/abort here, and a hung `client.send()` or stream can outlive both it and the 10-minute outbox lease. A slow source therefore has two live pulls from the same cursor. They can both call the provider and write the same events, and their completion order can make the compatibility checkpoint disagree with the process state.

Make the execution deadline enforceable and cancellation-aware (including adapter transport/stream operations), and keep the run leased/fenced until it actually settles; alternatively, do not supersede a run merely because elapsed wall time passed. Add a regression test with an unresolved adapter promise proving that a stale wake cannot create a concurrent `runOnce` for the same source.

langwatch/ee/event-sourcing/pipelines/ingestion-pull-processing/projections/ingestionPullRunStatus.foldProjection.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  handleIngestionPullRunCompleted(
+    event: IngestionPullRunCompletedEvent,
+    state: IngestionPullRunStatusData,
+  ): IngestionPullRunStatusData {
+    return {
+      ...state,
+      SourceId: event.data.sourceId,
+      Cursor: event.data.nextCursor,
```
langwatch-agent, line 155: **[P2] Apply the same current-run fence to the mirrored cursor**

The process manager correctly ignores a completion whose `runId` is no longer `currentRun` (`ingestionPullProcess.definition.ts:153-158`), but this projection unconditionally accepts it. Its repository then transactionally copies that stale cursor into `IngestionSource.pollerCursor` (`ingestion-pull-run-projection.prisma.repository.ts:79-86`). For example, run 1 becomes stale, run 2 completes with cursor B, then run 1 finishes with cursor A: the scheduler state stays at B while this supposedly durable/rebuildable compatibility checkpoint regresses to A. A later reconstruction/configuration path that seeds from the source row can replay data or diverge from the process.

Represent the active run identity (or otherwise order/fence outcomes) in the read model and ignore superseded completion/failure events before writing the projection and source compatibility fields. The existing late-completion test asserts only `Enabled`/cron; extend it to assert that neither cursor nor status/error metadata regresses.

langwatch-agent (changes requested):
Request changes: the process gives a strong durable-scheduling shape, but two run-identity/liveness gaps break the claimed one-in-flight and authoritative-cursor guarantees.

I also ran the focused ingestion-pull suites locally: 43 process/adapter/projection tests plus the 2 lifecycle tests passed after generating Prisma. The current branch's required `typecheck:tests` is still red (the topic-clustering/process-handler type failures on the stacked base); it must be green before this can merge.

langwatch/ee/governance/services/pullers/process-manager/ingestionPullProcess.definition.ts
0xdeafcafe: Fixed — the deadline is now enforceable and cancellation-aware, rather than a hint.

`PullRunOptions` gains a `signal`, and `PER_JOB_DEADLINE_MS` is documented as hard. The worker wraps `runOnce` in `withDeadline`, which races it against an `AbortController` — so the worker stops waiting even if the adapter never cooperates — and aborts the controller in `finally` so a transport still holding the signal unwinds once we've stopped waiting. A cut-off run throws `IngestionPullDeadlineExceededError`, which takes the existing failure path: cursor untouched, window retried, not skipped.

Adapters now honour it at the transport:
- **S3** (the case with no timeout at all): `abortSignal` on both `ListObjectsV2` and `GetObject` `client.send` calls, plus an aborted-check while draining the body — the SDK's signal covers the request, not a slow trickle from an already-open stream.
- **HTTP**: `AbortSignal.any([runSignal, AbortSignal.timeout(REQUEST_TIMEOUT_MS)])` so either bound unwinds the call, and the retry loop breaks once aborted rather than burning time the scheduler has given up on.

That keeps the two thresholds meaningfully separated — a run cannot execute beyond 5min, so by the time supersession is possible at 30min the original is provably done.

Regression test added (`pullerWorker.deadline.unit.test.ts`) with exactly the shape you asked for: an adapter returning a promise that never resolves. Three cases — the run fails at the deadline instead of executing indefinitely, **the adapter's signal is observably aborted** (the property that makes supersede-and-restart safe), and the durable cursor is left untouched. Full ingestion-pull suites: 62/62.

langwatch/ee/event-sourcing/pipelines/ingestion-pull-processing/projections/ingestionPullRunStatus.foldProjection.ts
0xdeafcafe: Fixed — the projection now carries the active run identity and fences on it.

Added `LastRunScheduledFor` to `IngestionPullRunStatusData` (plus the column, folded into this PR's own not-yet-deployed migration). Both outcome handlers drop through `isSuperseded()` before writing: runs are scheduled in time order and `runId` is derived from `scheduledFor`, so a strictly smaller `scheduledFor` means the outcome belongs to a run this row has already moved past. Equal `scheduledFor` still folds — it's the same run reporting, and replay must fold it identically.

Because the fence is in the projection state, the transactional mirror into `IngestionSource.pollerCursor` inherits it — no separate guard needed in the repository. Your exact scenario (run 1 stale, run 2 completes with cursor B, run 1 finishes with cursor A) now leaves the checkpoint at B.

Extended the late-completion coverage as suggested — it asserted only `Enabled`/cron. New cases: a superseded completion regresses neither cursor nor `LastRunEventCount`/`LastRunAt`; a superseded *failure* doesn't flip `LastRunOutcome` to failed, populate `LastRunError`, or bump `ConsecutiveErrors`; and the current run reporting again still folds. I mutation-checked the fence — removing it turns two of them red.

Also merged the updated #5930 base: `typecheck:tests` is now **green (exit 0)** on this branch, which was your other blocker.

0xdeafcafe:
## Review round findings (beyond the two P1/P2s)

I actioned two of these in `7a4ecbfe`; the rest are pre-existing in this PR and want your call.

**Fixed here**
- **Configure handler had the same unfenced-cursor bug**, one handler over from the one you flagged. `handleIngestionPullConfigured` wrote `event.data.cursor` unconditionally, while the process manager guards it (`previousState.sourceId ? previousState.cursor : view.cursor`). Since `syncIngestionPullSource` snapshots `pollerCursor` at edit time, renaming a source or editing its schedule mid-pull replayed a stale cursor over the live one — and the repo mirrors it back into `pollerCursor`. Only the first configure seeds it now.
- **My deadline test was vacuous** — it asserted `ingestionSource.update` was never called, but this PR removes every `update` call from `pullerWorker`, so it passed on all paths. Now asserts the run didn't resolve and nothing was ingested.

**Still open — worth a decision**
1. **P1: duplicate wake at the same `scheduledFor` silently loses a cursor advance.** `runId = String(scheduledFor)`, and the `active` guard only suppresses a re-run while `currentRun` is non-null. If run 1 completes (`currentRun` → null) and the wake worker redelivers the same `scheduledFor`, a second run starts with an identical `runId` — and the completion command's idempotency key is `${sourceId}:ingestion_pull:${runId}:completed`, so the second completion is deduped away, its `nextCursor` discarded, and `currentRun` stays pinned until the 30-min stale timer. At-least-once wake delivery is assumed by the design, so this is reachable. Wants a run identity that isn't purely `scheduledFor` (e.g. include an attempt/nonce).
2. **P2: `EventId` scheme change is a silent dedup break.** Changed from `${sourceType}:${source_event_id}` to `${sourceType}:${ingestionSourceId}:${source_event_id}`. Correct going forward, but rows already ingested under the old key won't collapse via the ReplacingMergeTree, so the next pull over an overlapping window duplicates them. Needs a backfill or an explicit one-time-duplication note in the PR body.
3. **P3: `errorCount > 0` is now a poison pill.** The worker throws before writing any events, so an adapter reporting a nonzero `errorCount` alongside good events discards the whole batch and doesn't advance. A permanently-bad item wedges the source forever. Previously the good events landed.
4. **P3: multi-tenancy convention.** `ingestion-pull-run-projection.prisma.repository.ts` `findUnique({where:{sourceId}})` and `updateMany({where:{id: sourceId}})` omit `projectId`. Not a cross-tenant leak (sourceId is a globally-unique KSUID) but it's against the project rule and will trip the multitenancy middleware.

PR #5906, opened 2026-07-18
fix(misc): guard against empty-string OPENAI_API_KEY in hotel bot client

### LINKED ISSUE #5826: Stock .env.example can't boot the API: empty OPENAI_API_KEY defeats the ?? guard
## Expected
A fresh checkout following the documented dev path — `cp langwatch/.env.example langwatch/.env`, then `pnpm dev` — boots the API.

## Actual
The API crashloops at module load:

```
OpenAIError: Missing credentials. Please pass an `apiKey`, `workloadIdentity`,
`adminAPIKey`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
    at crypto (langwatch/src/server/routes/misc.ts:173:24)
```

`.env.example` ships `OPENAI_API_KEY=` (empty). `misc.ts:173` constructs a module-scope client with `apiKey: env.OPENAI_API_KEY ?? "bogus"` — `??` only guards null/undefined, so the empty string passes straight through and the SDK throws. The `"bogus"` default reads as if it were already handling this case; it isn't.

Setting any non-empty dummy value clears it.

## Reproduction
1. Fresh worktree, `cp langwatch/.env.example langwatch/.env`
2. Generate the required ≥32-char secrets (see the related friction below)
3. `pnpm dev`
4. The `[api]` lane crashloops under `concurrently --restart-tries -1`

Two adjacent frictions on the same path, worth folding into the same fix: `.env.example` also ships `LW_VIRTUAL_KEY_PEPPER=`, `LW_GATEWAY_INTERNAL_SECRET=` and `LW_GATEWAY_JWT_SECRET=` empty, and all three are `min(32)` — the env validator kills the API until each is filled. A stock `.env.example` therefore cannot boot the app as documented.

Found while booting a local stack to reproduce #5759.

---
*Classification: Bug*
*Status: stub — filed to preserve the finding; not investigated. `not-ready` until the investigation step runs.*

feiiiiii5:
## What

Swaps `??` for `||` in the `hotelBotOpenai` client construction in `src/server/routes/misc.ts` (line 173-179).

## Why

Fixes #5826.

The stock `.env.example` ships `OPENAI_API_KEY=` (empty string, not unset). `env.OPENAI_API_KEY ?? "bogus"` only guards `null`/`undefined`, so it resolves to `""`, and `new OpenAI({ apiKey: "" })` throws `OpenAIError: Missing credentials` at module load — the API process never finishes booting. `||` treats `""` as falsy and falls through to the bogus demo default, which is the original intent of the fallback (the hotel bot is a demo fixture that should never block API startup).

## Approach

One-character change: `??` -> `||`, plus a 4-line comment explaining the operator choice. No other call sites in the codebase have this exact shape (env var that allows empty string + downstream SDK that throws on empty string + sentinel fallback), so the fix is intentionally narrow. Verified by searching the repo for `` ?? "bogus" `` patterns — only this one matches.

## Out of scope

The same `.env.example` ships `LW_VIRTUAL_KEY_PEPPER="REPLACE_ME"`, `LW_GATEWAY_INTERNAL_SECRET="REPLACE_ME"`, `LW_GATEWAY_JWT_SECRET="REPLACE_ME"` (11 chars each), which fail `z.string().min(32).optional()` in `env-create.mjs`. That's a separate failure mode (zod validation, not SDK construction) and warrants its own issue/PR. Happy to file that separately if useful.

## Testing

- Confirmed `env.OPENAI_API_KEY` is typed as `z.string().optional()` (no `min()` constraint) in `src/env-create.mjs` — so empty string is a valid value at runtime.
- Confirmed `.env.example:203` ships `OPENAI_API_KEY=` empty on `main`.
- Confirmed no other `?? "bogus"` patterns exist in the codebase.
- Did not add a regression test: the fix is a 1-character operator change with a clear root cause and comment. A test that doesn't import the production code would be a tautology; a test that does would require mocking the entire Hono/Prisma app-layer for a 1-character fix. The crash on regression is loud (API won't boot). Open to adding one if reviewers prefer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved demo hotel bot startup behavior when the OpenAI API key is empty or missing.
  * Prevented an immediate credentials error during client initialization.

### DIFF
--- a/langwatch/src/server/routes/misc.ts
+++ b/langwatch/src/server/routes/misc.ts
@@ -170,8 +170,12 @@ secured
 // =============================================
 // POST /api/demo/hotel_bot
 // =============================================
+// `||` (not `??`) so an empty OPENAI_API_KEY from a stock .env.example
+// falls through to the bogus default. `??` only guards null/undefined;
+// the OpenAI SDK rejects empty-string apiKey at client construction
+// with "Missing credentials" (see issue #5826).
 const hotelBotOpenai = new OpenAI({
-  apiKey: env.OPENAI_API_KEY ?? "bogus",
+  apiKey: env.OPENAI_API_KEY || "bogus",
 });
 
 const guestQueries = [


Aryansharma28 (approved):
## goated-review (fast-skim) — PR #5906

**Verdict: APPROVE.** No P0/P1/P2 findings.

Verified every claim in the PR description against the actual code rather than taking it on faith:

- `src/env-create.mjs:116` — `OPENAI_API_KEY: z.string().optional()`, no `.min()` — confirms an empty string passes env validation.
- `.env.example:203` — `OPENAI_API_KEY=` ships empty on `main` — confirms the stock dev setup hits this.
- `node_modules/openai/client.js:156` — `if (!providerRuntime && !apiKey && ...) throw new Errors.OpenAIError('Missing credentials...')` — `!""` is `true`, so an empty-string `apiKey` throws exactly as described, not just `null`/`undefined`.
- `src/server/routes/misc.ts:173-175` — `new OpenAI(...)` sits at **module top level**, and `src/server/api-router.ts:57` imports `misc.ts` eagerly (`import { app as miscApp } from "./routes/misc"`) — so this really does run at server-boot import time, not lazily inside the route handler. The "API process never finishes booting" claim checks out.
- `||` vs `??` divergence is scoped correctly: since `OPENAI_API_KEY` is string-typed, the only falsy value either operator could hit is `""` — every other case (a real key, or `undefined`) behaves identically under both operators. No behavior change for any working configuration, only for the broken empty-string one.
- Grepped the repo for other `?? "bogus"` / similar sentinel-fallback shapes — none found, so "intentionally narrow, single call site" holds.
- CodeRabbit found no actionable comments; CI green (path-filtered jobs skip as expected for a single non-hot-path file, all `*-complete`/title-validation checks pass).

### P5 — Praise / observation
Good defensive diagnosis chain (env schema -> `.env.example` -> SDK internals -> import graph) for what looks like a 1-character diff — that's exactly the kind of verification a one-liner like this needs, since the failure mode (crash on boot) is invisible from the diff alone. The call to skip a regression test is reasonable here: the demo hotel-bot endpoint is explicitly documented (`misc.ts:194-199`) as an intentionally unmigrated fixture, and a test exercising module-load-time `OpenAI` construction would need `vi.resetModules()` + import gymnastics for a fix this narrow — not proportionate.

- Whether any other route module has the same shape (env var typed to allow empty string + an SDK/library that throws on empty string at import time) that isn't caught by the `?? "bogus"` grep pattern — e.g. a different sentinel string or a different failing library. The PR's search was reasonably scoped to the exact pattern that caused #5826; a broader audit of all module-top-level SDK constructions is out of scope for a 1-line fix.

langwatch-agent (commented):

feiiiiii5:
Closing: while resolving the merge conflict I found `main` has independently picked up the identical fix (`apiKey: env.OPENAI_API_KEY || "bogus"` with an equivalent explanatory comment) — after merging, this PR's net diff against main was a single comment line. No point keeping a duplicate open. Thanks @langwatch-agent for the earlier review!

PR #5909, opened 2026-07-18
feat(event-sourcing): make store-miss re-folds and cache lifecycle visible

0xdeafcafe:
## What changed

- added durable store-miss re-fold counters split into `first_touch` and `resumed`
- added a replay-size histogram with buckets `[1, 2, 5, 10, 25, 50, 100, 250, 1000]`
- instrumented both array and streamed store-miss re-fold paths without changing their existing logs or behavior
- counted Redis `DEL` failures after failed fold-cache `SET` operations
- added Gherkin coverage plus executor, cache, coalesced-batch, and real Prometheus registry tests

## Why

Store-miss re-folds dominate production re-fold cost but were only visible in short-lived logs. These metrics make the work trendable and distinguish expected first-touch rebuilds from resumed history replays, while exposing failed Redis cleanup separately.

## Validation

- `pnpm test:unit src/server/__tests__/metrics-instrumentation.unit.test.ts` — 33 tests passed
- focused fold executor and Redis cache unit suite — 43 tests passed
- `nice -n 19 ./node_modules/.bin/tsgo --noEmit --project ./tsconfig.tsgo.json` — passed
- `git diff --check 83129e09b...HEAD` — passed
- Biome check could not start because the checked-in 2.4.14 config is incompatible with the installed 2.5.1 CLI; this mismatch exists on the base branch

https://claude.ai/code/session_01Rxp9AeY1d18xEW7KnNCEt1

### DIFF
--- a/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
+++ b/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
@@ -63,6 +63,26 @@ function withOccurredAtHint(
   return { ...context, occurredAtMs: occurredAt };
 }
 
+/**
+ * Whether a re-folded history event pre-dates this delivery — i.e. the fold is
+ * RESUMING an aggregate that already had events, rather than touching it for
+ * the first time.
+ *
+ * Keyed on `event.id`, deliberately, and never on `idempotencyKey`. A retry
+ * delivery carries a NEW id but the SAME idempotency key as the earlier event
+ * it repeats, so keying on the idempotency key would read that earlier event as
+ * "one of the ones we just delivered" and mislabel resumed work as first touch.
+ * Both re-fold paths must share this predicate or the same persisted history
+ * reports different kinds depending on which loader the projection happens to
+ * use.
+ *
+ * This is distinct from the DEDUP key (`idempotencyKey || id`), which the
+ * streamed path still needs to collapse a raw, undeduplicated page.
+ */
+function isPreExisting(event: Event, deliveredIds: ReadonlySet<string>): boolean {
+  return !deliveredIds.has(event.id);
+}
+
 function recordStoreMissRefoldMetrics({
   projectionName,
   refoldEventCount,
@@ -388,8 +408,8 @@ export class FoldProjectionExecutor {
     if (history.length === 0) return null;
 
     const deliveredIds = new Set(delivered.map((event) => event.id));
-    const sawPreExisting = history.some(
-      (event) => !deliveredIds.has(event.id),
+    const sawPreExisting = history.some((event) =>
+      isPreExisting(event, deliveredIds),
     );
     recordStoreMissRefoldMetrics({
       projectionName: projection.name,
@@ -455,9 +475,10 @@ export class FoldProjectionExecutor {
     // instead of hanging the fold worker for the aggregate indefinitely.
     // 100k pages * 1000/page default covers a 100M-event aggregate.
     const MAX_PAGES = 100_000;
-    const deliveredDedupKeys = new Set(
-      delivered.map((event) => event.idempotencyKey || event.id),
-    );
+    // Two different keyings, on purpose — see `isPreExisting`. Dedup collapses
+    // a raw page by idempotency key; the resumed/first-touch classification
+    // keys on id so it matches the array path.
+    const deliveredIds = new Set(delivered.map((event) => event.id));
     const seen = new Set<string>();
     let state = projection.init();
     let after: { timestamp: number; eventId: string } | undefined;
@@ -485,7 +506,7 @@ export class FoldProjectionExecutor {
         const dedupKey = event.idempotencyKey || event.id;
         if (seen.has(dedupKey)) continue;
         seen.add(dedupKey);
-        if (!deliveredDedupKeys.has(dedupKey)) sawPreExisting = true;
+        if (isPreExisting(event, deliveredIds)) sawPreExisting = true;
         state = projection.apply(state, event as E);
         refoldEventCount++;
       }
--- a/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
+++ b/langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
@@ -1,5 +1,9 @@
 import { createLogger } from "@langwatch/observability";
-import { incrementEsFoldRefoldTotal } from "~/server/metrics";
+import {
+  incrementEsFoldRefoldTotal,
+  incrementEsFoldStoreMissRefoldTotal,
+  observeEsFoldStoreMissRefoldEvents,
+} from "~/server/metrics";
 import type { Event } from "../domain/types";
 import type { FoldProjectionDefinition } from "./foldProjection.types";
 import type { ProjectionStoreContext } from "./projectionStoreContext";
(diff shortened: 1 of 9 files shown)


langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
```diff
@@ -448,6 +485,7 @@ export class FoldProjectionExecutor {
         const dedupKey = event.idempotencyKey || event.id;
         if (seen.has(dedupKey)) continue;
         seen.add(dedupKey);
+        if (!deliveredDedupKeys.has(dedupKey)) sawPreExisting = true;
```
langwatch-agent: **P2 - The streamed path classifies idempotency retries differently from the array path.** `eventLoaderUpTo` returns deduplicated history, so with an earlier event `e1` and delivered retry `e2` sharing key `K`, the array path sees history `[e1]` and labels the rebuild `resumed` (`e1.id !== e2.id`). The paged loader deliberately returns raw events, but this condition compares only idempotency keys: both `e1` and `e2` are `K`, so it reports `first_touch`.

That is the same persisted history and delivered batch taking the two supported loader routes, yet it emits different labels for the metric this PR adds specifically to distinguish resumed work. Define the predicate once over an equivalent deduplicated history representation (or otherwise make both paths use the same explicit semantics), and add a parity test with different event IDs sharing an idempotency key.

langwatch-agent (changes requested):
Request changes: 1 P2 telemetry-correctness issue in the streamed re-fold classification. Details are inline.

langwatch-agent (commented):
Second pass on this, deliberately aimed at different ground than the existing review so this is additive rather than a repeat. That earlier note already has the streamed-path classification issue; I hit the same fault line from the other side and it is broader than one path, see 2 below.

No blocking correctness bug here, and I could not find any behavior smuggled in beyond one log-level bump. My concern is narrower: as designed, these metrics will under-deliver on the goal the PR states, for three independent reasons.

First, on CI: the red `test` check is not yours. `sdk-python-ci` has failed on every branch for the last hour, all on the Python SDK examples with `RateLimitError: OpenAIException - You exceeded your current quota`. Raised separately.

### 1. The histogram ceiling sits an order of magnitude below the workload it is meant to measure

`metrics.ts:461` — `buckets: [1, 2, 5, 10, 25, 50, 100, 250, 1000]`.

The only two projections with `refoldOnStoreMiss: true` are `traceAnalytics` and `evaluationAnalytics`, and this codebase already documents their scale in the opposite direction: `traceAnalytics.foldProjection.ts:611-615` describes a hot trace as one where "a Claude Code session streams 100k+ events into one aggregate", and `foldProjectionExecutor.ts:456` sizes `MAX_PAGES` so that "100k pages * 1000/page default covers a 100M-event aggregate".

So a 40k-event re-fold and a 100k-event re-fold both land in `+Inf`, and `histogram_quantile(0.95, ...)` returns `+Inf`. The p95 panel is unusable in exactly the regime the PR body calls out as dominating production re-fold cost. Worth extending coverage into the 10k to 100k range.

### 2. `first_touch` vs `resumed` is computed from different identity keys in the two paths

The array path keys on `event.id` (`foldProjectionExecutor.ts:390-392`); the streamed path keys on `idempotencyKey || event.id` (`:458-460`, `:488`).

That difference is load-bearing, because `getEventsUpTo` returns `deduplicateEvents(processed)` (`abstractEventStore.ts:252`) and `deduplicateEvents` collapses same-`idempotencyKey` events across different EventIds, keeping the first (`eventStoreUtils.ts:120-136`). So when a command retry re-appends an event with a new EventId under the same idempotency key, the delivered batch carries the retry while history carries the original, and the array path's `history.some(e => !deliveredIds.has(e.id))` is true. A brand-new aggregate gets counted `resumed`.

The sting is that `evaluationAnalytics` takes the array path and `traceAnalytics` takes the streamed one, so the same event pattern reports opposite `kind` on the two projections. The streamed path's own doc comment at `:434-442` names this retry case explicitly; the array path just never got the same treatment. Defining the predicate once over an equivalent deduplicated representation would fix both this and the issue in the earlier review, and a parity test with distinct event IDs sharing an idempotency key would pin it.

### 3. `projection_name` carries two different vocabularies, so the new metrics cannot be joined to the existing ones

The new metrics are labelled with `projection.name`, which is camelCase: `traceAnalytics`, `evaluationAnalytics`. The existing `es_fold_cache_total` and `es_fold_cache_redis_error_total` are labelled with `keyPrefix`, which is snake_case: `trace_analytics`, `evaluation_analytics`, because `redisCachedFoldStore.ts` passes `this.keyPrefix` into a parameter named `projectionName`. Both label sets are named `projection_name` on the wire.

The obvious dashboard for this feature is the one correlating cache misses against re-fold cost. Any panel joining `es_fold_cache_total` and `es_fold_store_miss_refold_total` `on (projection_name)`, or grouping both `by (projection_name)`, returns empty or two disjoint series with no error. The mismatch predates this PR for `get`/`set`, but this PR extends it with `"del"` and adds new metrics on the other side of the divide, so it is worth settling here rather than inheriting it.

### Smaller

- **Array path counts attempts, streamed path counts completions.** `:394` records before the `apply` loop at `:417-420`; `:500` records after the streamed fold loop finishes. An event that deterministically throws inside `apply` and is retried by the queue inflates the counter and histogram once per retry on `evaluationAnalytics`, while the identical case records nothing on `traceAnalytics`.
- **Undocumented log-level change.** `redisCachedFoldStore.ts:133` moves "Redis SET failed after CH write" from `warn` to `error`. The body says the instrumentation lands "without changing their existing logs". This path is already handled and retried, so if anything alerts on error-level output from this logger, a transient Redis blip now pages where it did not before.
- **Spec assertion weakened rather than corrected.** `specs/event-sourcing/redis-fold-cache.feature:28` drops "with a 300-second TTL" for "only after the durable write succeeds". Dropping the literal is right since the TTL is env-overridable, but the replacement asserts nothing about expiry, so a regression that dropped `"EX", this.ttlSeconds` entirely and leaked a permanent key per aggregate would still pass.

### Checked and clean

- **Cardinality is fine**, which was my main worry going in. Every label value is a compile-time literal: `projection.name` is a `readonly name = "..."` field, `keyPrefix` is one of six literals in `pipelineRegistry.ts`. No projectId, traceId, aggregateId or user input reaches a label. Worst case is roughly 4 counter series and 2 histograms.
- **No double-counting.** `refoldUpToDelivered` has two call sites, and `executeBatch` delegates to `execute` when `matching.length === 1`, so both never fire for one delivery. Recording sits after the `return null` guards, so a no-op re-fold is not counted.
- **The DEL counter counts what it claims**, and the error handling is right: the DEL error is logged and swallowed with the existing justification that a surviving stale entry is safe, while the original SET error is rethrown so the queue still retries.
- **The `deliveredIds` refactor at `:414-415` is behavior-preserving**, including under duplicate ids within `delivered`.
- **Event-type filtering does not currently skew the split**, since both projections subscribe to their aggregate's full event catalog. It is latent though: a future projection subscribing to a subset would report `resumed` on every store miss.

langwatch/src/server/event-sourcing/projections/foldProjectionExecutor.ts
0xdeafcafe: Fixed in e8ae3e8e. You were right that these were the same history reporting different kinds.

The two paths were keyed differently:

- array: `history.some(e => !deliveredIds.has(e.id))` — by **event id**
- streamed: `deliveredDedupKeys` built from `idempotencyKey || id`, then compared against each page event's `dedupKey` — by **idempotency key** when one is present

So with earlier `e1` and delivered retry `e2` sharing key `K`: the array path sees deduplicated history `[e1]`, `e1.id ∉ {e2.id}` → `resumed`; the streamed path sees `e1.dedupKey = K ∈ {K}` → `first_touch`.

Extracted the predicate as `isPreExisting(event, deliveredIds)` and both paths now call it. Event id is the right key, and the JSDoc says why: a retry carries a new id but repeats an earlier event's key, so keying on the key reads the earlier event as one of the delivered ones and calls plainly-resumed work first touch. The streamed path keeps the idempotency key for **dedup** — a separate concern, since it still has to collapse a raw undeduplicated page.

Parity test added as you asked, driving one history with two events sharing a key through both loader routes:

```
it("reports resumed on the streamed path")
it("reports resumed on the array path too")
```

Verified it catches the defect: swapping `isPreExisting` back for the old key comparison fails the streamed case (`× reports resumed on the streamed path`, 1 failed | 6 passed) and passes with the fix. Full projections suite: 17 files, 164 tests green; tslsp clean.

0xdeafcafe:
Superseded by #5918. Stacked on #5908, whose design was removed there.

Refold visibility is still wanted — the surviving metric is `es_fold_refold_total{projection_name, outcome}`. An observability audit of #5918 flagged the remaining blind spots, chiefly that nothing distinguishes a cache miss on a retry from one on a fresh delivery.

PR #759, opened 2025-10-30
fix: dataset slug name sync

drewdrewthis:
Updating the dataset name was only updating the slug for the UI, not the actual slug in the database. This fixes that and improves UI/UX
Screenshot 2025-10-31 at 14 05 50
Screenshot 2025-10-31 at 14 05 33

PR #5910, opened 2026-07-18
refactor(group-queue): lease-based blob lifecycle replaces holder ref-counting

0xdeafcafe:
## What changed

- replace GroupQueue holder reference counts with renewable per-holder leases
- remove eager shared-blob deletion from completion, retry, and dedup paths
- keep Redis blobs bounded by lease and backstop TTLs
- route durable GroupQueue payloads to a deployment-owned bucket and prefix
- mint `temp-tier-3-offload/{tenantId}/{hash}` keys without consulting tenant BYOC storage
- use a queue-only S3 client and expose bucket, prefix, and endpoint through the Helm chart
- retain the existing object-store or local-file fallback when dedicated queue storage is disabled

The matching SaaS infrastructure is langwatch/langwatch-saas#758. Legacy orphan reconciliation is tracked in langwatch/langwatch-saas#759 and will be implemented in Go.

## Deployment Impact

- Env vars: adds optional `LANGWATCH_QUEUE_PAYLOAD_BUCKET`, `LANGWATCH_QUEUE_PAYLOAD_PREFIX`, and `LANGWATCH_QUEUE_PAYLOAD_S3_ENDPOINT`; `GROUP_QUEUE_ENVELOPE_WRITES_ENABLED` remains the GQ2 write gate.
- Helm: adds `app.queuePayloads.enabled`, `bucket`, `prefix`, and `endpoint`. A vanilla `helm install` keeps this disabled and emits none of the queue-storage env vars.
- BYOC: enabling dedicated queue storage bypasses tenant BYOC for temporary GroupQueue payloads; leaving it disabled preserves the existing object-store or local-file behavior.
- Compose: no compose-file change is required because it already loads `langwatch/.env`; `.env.example` documents the optional settings.

Configure the dedicated bucket settings before enabling envelope writes in SaaS.

## Validation

- focused GroupQueue and S3 driver unit tests: 24 passed
- application and test TypeScript checks
- Helm lint and enabled/disabled template renders

The repository Biome command is currently blocked by the checked-in 2.4.14 schema being incompatible with the installed 2.5.1 CLI; changed TypeScript files were formatted with the repository Prettier dependency.

## Summary by CodeRabbit

* **New Features**
  * Added renewable, per-holder leases for shared, content-addressed payloads.
  * Added optional durable-tier GroupQueue payload storage to a dedicated S3 bucket/prefix (`temp-tier-3-offload/...`).

* **Bug Fixes**
  * Improved lazy reclamation behavior to avoid eager deletion and tightened retry/transfer correctness.
  * Hardened envelope decoding to avoid trusting stored references across reads/renewals.

* **Documentation**
  * Updated ADRs and GroupQueue specs to describe lease-based lifecycle and durable-tier URI behavior.

* **Tests**
  * Migrated holder-focused coverage to lease-focused unit/integration suites and refreshed related assertions.


langwatch-agent (changes requested):
Reviewed with a focus on migration safety and the lease lifecycle. The Redis side of this is careful work and I think the rolling-deploy story holds. One thing I would want resolved before merge, because it is the one consequence that a rollback does not undo.

First, on CI: the red `test` check here is not yours. `sdk-python-ci` has failed on every branch for the last hour (11 consecutive runs across 7 branches), all on `tests/test_examples.py::test_example[examples/distributed_tracing.py]` with `RateLimitError: OpenAIException - You exceeded your current quota`. Raised separately.

### 1. S3-tier blobs now have no reclaim path at all, and this part is irreversible

Eager S3 reclaim is gone: `gqBlobReclaimS3FailuresTotal` is referenced 4 times in `envelopeBlobLifecycle.ts` on main and 0 times on this branch, leaving only the declaration at `metrics.ts:174`. `TieredBlobStore.delete` survives but its own doc now says GQ2 release and transfer never call it. Redis blobs are fine, `RedisJobBlobStore.put` always sets `EX`, so the 4-day backstop covers them including keys written before the deploy. S3 objects have no TTL.

The PR defers this to the deployment's bucket lifecycle policy. I checked, and I do not think that policy exists for these objects. In the saas infra, `aws_s3_bucket_lifecycle_configuration.langwatch_storage_lifecycle` on `langwatch-storage-{env}` has exactly three rules: `expire_langevals_staging` (prefix `langevals-staging/`, 1 day), `expire_dataset_staging` (prefix `staging/`, 1 day), and `cleanup_old_versions`, which is prefix-wide but only sets `noncurrent_version_expiration`, so it never expires a current object. Nothing expires current objects under the offloaded-payload prefix.

There is a second half that a lifecycle rule would not fix anyway: the S3 tier resolves per-project storage, so for BYOC projects these objects land in a tenant-owned bucket the deployment does not control. No rule we write covers those.

Why I would block on this specifically rather than file it as follow-up: it is the one part of this change that a rollback does not repair. Redis-side degradation is bounded and self-healing, but objects orphaned during the new-code window were released by code that no longer deletes, and reverting does not go back and collect them. If a policy or project-purge path does cover this and I have missed it, point me at it and I will withdraw this.

### 2. Redis blob memory takes an unbudgeted step change

Losing eager reclaim means a blob now lives up to 4 days from its last touch rather than roughly its job lifetime. Steady-state `{queue}:gq:blob:*` becomes about 4 days of offloaded payload volume. The history behind this area cites a leak of roughly 279k blobs and 1.9 GB as an incident; this change makes a comparable retained footprint the intended steady state. Not a correctness problem and I understand the tradeoff, but it is worth a capacity check against the Redis instance before rollout, because the failure mode is silent eviction pressure rather than a loud error.

### 3. `take()` does not slide the blob backstop, but `renew()` does

`TAKE_LUA` guards its blob `EXPIRE` behind `if #KEYS == 3` (`blobLeases.ts:27`), and `BlobLeases.take` passes 2 keys (`blobLeases.ts:125-132`) while `renew` passes 3 on the redis tier (`blobLeases.ts:150-152`). So a lease taken through `take()` gets a fresh 3-day deadline against a blob whose TTL is untouched. `transfer()` likewise never receives the blob key.

The stated invariant is that the 4-day backstop always outlives the 3-day lease with a day of margin. On the `takeLeaseOrThrow` fallback at `envelopeBlobLifecycle.ts:374`, which the code notes dominates during the GQ1 to GQ2 rollout window, that margin is whatever the blob had left rather than a full day. Is that intentional on the grounds that the blob was just written, or should `take()` pass the blob key the way `renew()` does?

### What I checked and found clean

Worth stating explicitly, since these were my main worries going in.

- **No `WRONGTYPE` risk.** The lease sorted set is a new key namespace (`{queue}:gq:blobleases:`), distinct from the legacy `blobholders` Set. New code only ever issues Set commands against the legacy key and ZSET commands against the new one.
- **The migration guard is correctly universal.** Old reclaim is gated on `SCARD == 0`, and `__gq2_lease_guard__` is co-written in the same eval by every lease-creating path: `TAKE_LUA`, `TRANSFER_LUA`, and Lua-side `gqTakeLease` behind STAGE, STAGE_BATCH, RESTAGE_AND_BLOCK and RETRY_RESTAGE. There is no interleaving where a lease exists without its guard, so an old pod cannot delete bytes a new lease protects.
- **Envelope format is wire-compatible both directions.** `header.h` is unchanged: same `randomUUID()`, same optionality, same gating condition. The rest is a TypeScript rename. Old-encoded envelopes decode under new code and vice versa.
- **The STAGE_LUA 3-tuple to 2-tuple return change is not a cross-version hazard**, since `CachedLuaScript` keys by script SHA so each version loads and parses its own script.
- **Rollback is safe on the Redis path.** Orphaned lease ZSETs are ignored and expire; the guard suppresses eager reclaim for at most 4 days after the last new-code touch, then old behavior resumes. Bounded and self-healing.

Happy to re-check once the S3 question is settled.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
```diff
@@ -159,73 +157,104 @@ export class EnvelopeBlobLifecycle {
...
+          err: redactStorageUrisInText(
+            err instanceof Error ? err.message : String(err),
+          ),
+        },
+        "Blob lease heartbeat renewal failed; relying on the blob backstop",
+      );
+    }
+  }
+
+  /**
```
langwatch-agent, line 224: **P1 - Do not remove durable blob reclamation without a real replacement.** This now treats the durable-store lifecycle/project purge as the reclaim path, but neither covers these queue blobs: `TieredBlobStore` writes directly to `s3://<bucket>/<projectId>/<hash>` and creates no `stored_objects` row, so `StoredObjectsService.deleteOwnedBy()` cannot see it; the deployed bucket lifecycle only expires `langevals-staging/` and `staging/`, not this root `<projectId>/` prefix. Consequently every completed or discarded S3-tier queue payload (which can contain customer trace data) becomes a permanent current object, and the same is true for BYOC/local-file destinations without their own external policy. The new tests explicitly lock in `objectStore.deleted === []`, so this will not be caught later. Please retain a safe last-live-lease delete path, or add and validate a queue-specific durable GC/lifecycle scheme (including storage-key namespacing, hosted and BYOC/local-file behavior) before removing the existing deletion.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.decodeDrop.integration.test.ts
```diff
@@ -307,8 +312,8 @@ describe.skipIf(!hasTestcontainers)(
 
     describe("given a staged job whose referenced blob is genuinely gone", () => {
       describe("when a worker claims the group and the decode fails", () => {
-        /** @scenario a missing-blob drop releases the absent blob's holder */
-        it("releases the blob's holder", async () => {
+        /** @scenario a missing-blob drop releases the absent blob's lease */
```
langwatch-agent: **P2 - Repair the feature-parity contract before merging.** This renamed annotation does not name any scenario in `specs/event-sourcing/payload-store-blob-hardening.feature`, and the required `feature-parity` job is red on this exact line. Either bind it to the existing missing-S3 scenario or add the lease-retirement scenario that this test is asserting; the PR cannot be considered green while the changed test carries an unknown scenario annotation.

langwatch-agent (changes requested):
Request changes: the durable-tier cleanup contract is not backed by the deployed storage lifecycle, and the PR currently leaves a required feature-parity check red. Details inline.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
0xdeafcafe: Fixed in 30003fbff8 — and you were right that the gap was real, though the shape had shifted since this review.

`47e0b1148f` (10h after your review) added the dedicated queue bucket + `temp-tier-3-offload/` prefix + a non-BYOC client. But it only applied the prefix when `LANGWATCH_QUEUE_PAYLOAD_BUCKET` was set. Every other destination still fell through `resolveProjectStorageDestination` to `mintS3Uri` and wrote at `<projectId>/<hash>` — precisely the root prefix you flagged, on the shared bucket, on tenant BYOC, and (via `mintFileUri`) on local disk.

So the prefix is now a property of the payload rather than of the configuration:

| Destination | Key |
| --- | --- |
| Dedicated queue bucket | `s3://<queue-bucket>/<prefix><tenantId>/<hash>` |
| Shared deployment bucket | `s3://<bucket>/<prefix><tenantId>/<hash>` |
| Tenant BYOC bucket | `s3://<byoc-bucket>/<prefix><tenantId>/<hash>` |
| Local filesystem | `file://<root>/<prefix><tenantId>/<hash>` |

`mintGroupQueueStorageUri` throws rather than mint an unprefixed key — a destination without one would otherwise interpolate a literal `undefined…` object that no reader finds and no rule sweeps. That guard immediately caught six test doubles.

On your point that the tests lock in the leak: the `deleted === []` assertions stay (a last-live-lease delete still races a concurrent take), but the release test now also asserts the object is under the prefix, so "no eager delete" cannot silently become "retained forever". `groupQueueStorage.unit.test.ts` locks the invariant across all four destinations.

Two things I want to be explicit about rather than imply:

1. **Namespacing makes reclaim possible; it does not perform it.** The lifecycle rule still has to exist per deployment. `ARCHITECTURE.md` (Durable-tier retention) and `.env.example` now state that as an operator requirement, including that local-filesystem installs have no lifecycle engine at all and need a periodic sweep. If you would rather this PR not merge until the SaaS rule is actually applied to the prefix, say so — that is a deploy-ordering call, not a code one.
2. **Legacy objects.** Anything written before this sits at the root. A decode miss discards the job permanently (#5538) rather than retrying, so the read path falls back to the legacy location on a miss, removable one release from now. Covered by a test verified to fail without the fallback.

147 unit tests green; typecheck clean. Integration tests need testcontainers, so those I have not run.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.gq2.integration.test.ts
```diff
@@ -171,6 +171,33 @@ describe.skipIf(!hasTestcontainers)("GroupQueueProcessor — GQ2 offload", () =>
...
+            groupId: TENANT_GROUP,
+            value: OFFLOADED_VALUE,
+          });
+        }
+
+        await vi.waitFor(async () => {
+          expect(await blobKeys(name)).toHaveLength(1);
+          const keys = await leaseKeys(name);
+          expect(keys).toHaveLength(1);
+          expect(await redis.zcard(keys[0]!)).toBe(3);
```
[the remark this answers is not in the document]
0xdeafcafe: Checked this one and the staging is not vacuous — `generateStagedJobId` composes `<id>/<jobType>/<jobName>`:

```ts
const baseId = (p.id as string) ?? crypto.randomUUID();
if (jobType && jobName) return `${baseId}/${jobType}/${jobName}`;
```

The three sends share `id: "same-payload"` but carry `__jobName: shared-payload-${i}`, so they are three distinct `stagedJobId`s — which is the fan-out the scenario is about: identical bytes, one blob, three leases. If they did collapse, `zcard(leaseKey)` and `zcard(group:jobs)` would read 1 and the test would fail rather than pass quietly, so the assertions are already load-bearing.

Added a comment at the send loop in 30003fbff8 so the next reader does not have to re-derive it.

0xdeafcafe:
this fallback is the bit that gets me.

when the bucket isn't set, `resolveGroupQueueStorageDestination` hands off to `resolveProjectStorageDestination(projectId)`, and tier 1 of that is the project's byoc bucket. so the one destination the infra doc says these must never touch is what you get by default when you forget to configure it. they then inherit customer retention instead of the 7 day expiry, so they just... stay. i'd bet quite some money that's how the payloads in #759 ended up where they did.

`createGroupQueueS3Client` pulls the same trick with `createS3Client(projectId)`, so it picks up byoc creds to match.

rather than patching the fallback i'd make the whole thing opt in. it's a specialist feature, it doesn't need to be on by default:

- flag off, the default: don't offload at all. the payload stays in the redis tier and backpressure earns its keep.
- flag on with a bucket: the dedicated bucket, as now.
- flag on without one: throw at boot. `_PREFIX` and `_S3_ENDPOINT` set with `_BUCKET` missing means someone was turning this on and fat-fingered the only one that matters, and right now that's a silent trip to the customer's bucket.

that way `resolveProjectStorageDestination` never enters the picture and byoc can't be reached by accident.

while we're here, `app.queuePayloads.enabled` doesn't say what it does. something like:

```yaml
app:
  queue:
    largePayloadOffload:
      enabled: false
      bucket: ""
      prefix: "temp-tier-3-offload/"
      endpoint: {}
```

keeping it under `app.` rather than `workers.` on purpose by the way. the block lives in sharedEnv and the app enqueues these as much as the workers do.

and then we just ned to gate the feature based on the `enabled` flag first, to be consistent pls!

0xdeafcafe:
Worked through the remaining review comments, including CodeRabbit's outside-diff findings that never became threads. Pushed as `346e3cd050`.

### Critical — cross-tenant blob read (CodeRabbit, `envelopeBlobLifecycle.ts` 131-152)

Confirmed and fixed. This was real and I would not have found it without the callout.

The decode guard keyed off `readEnvelopeLeaseFromHeader`, which returns null unless the header carries `h`:

```ts
if ((header.e === "redis" || header.e === "s3") && header.ref && typeof header.h === "string")
```

So an envelope with a valid tiered `ref` and **no** `h` produced no lease, skipped the guard entirely, and was still fetched by `decodeJobEnvelope` — which has no tenant check of its own. Omitting one field bypassed tenant isolation.

The guard now derives from the ref via a new `readEnvelopeTieredRefFromHeader`; the lease is used only for renewal, which is all it was ever the right key for. I also took the `!expected` half of your suggestion: a group with no tenant prefix is refused rather than waved through on an `undefined === undefined` match.

Covered by `envelopeBlobLifecycle.tenantGuard.unit.test.ts` — unit, not integration, because the guard refuses before touching any store, so it runs in CI's fast lane. Three of its four cases fail against the old lease-keyed check; the fourth (lease present) passed before and still does, which is the right shape for the regression.

### Major — atomic restaging lease transfer (CodeRabbit, `groupQueue.ts` 1102-1118)

Half fixed, half deferred, deliberately.

The concrete bug was the unchecked result: `retryRestage` returns `false` when the active key is stale and nothing was written, but the call site transferred regardless — taking a lease for a value no staged job references, and releasing the old one, dropping the live owner's protection. That is now guarded.

I did not fold the transfer into the staging Lua. That is the correct end state and I have noted it in the code, but it needs the blocked-restage path too and this PR is already large. The window it leaves is a crash between publication and transfer, which is survivable: the retry re-encodes to the same content hash, so the not-yet-released old lease keeps the blob alive until decode renews. Happy to do it here instead if you would rather not carry it.

### P2 — feature-parity (langwatch-agent)

Green. `pnpm check:feature-parity` reports `2517 enforced scenario(s) bound`, with `groupqueue-decode-drop-durability.feature` 12/12 and `payload-store-blob-hardening.feature` 6/6. The annotation was not orphaned so much as drifted — the spec still said "holder" after WP4 renamed holders to leases, so I moved the spec to lease vocabulary and matched the test to it.

### Already fixed on head (CodeRabbit, second review)

These three landed in `47e0b1148f`, after the review was written:

- **`groupQueue.ts` 305-306, nonexistent "take" seam** — comment now reads encode/decode/renew/release. Tidied one leftover: "acquires holder leases" to "acquires the leases".
- **`envelopeBlobLifecycle.ts` 314-316, contradictory transfer comment** — now reads "Skip the foreign replacement; the guarded release retires the old lease only when its tenant matches."
- **`envelopeBlobLifecycle.ts` 140-149 et al, unredacted tenant ids** — every `logger.warn` in the file uses `projectId: "[redacted]"`. The unredacted `projectId:` occurrences flagged (173, 204, 266, 399) are arguments to `blobLeases.renew/take/release`, not log fields.

### One correction to my earlier reply

I said typecheck was clean on the previous push. It was not — my grep was swallowing ANSI codes and I was passing the wrong tsconfig, so it silently matched nothing. Running it properly (`tsconfig.tsgo.json`) surfaced a real error my prefix change had introduced: `resolveStorageDestination` was typed `ProjectStorageDestination` and only compiled because `prefix` was optional. It is now typed `GroupQueueStorageDestination`, so a resolver that cannot name a lifecycle-managed prefix is a compile error rather than an unswept payload. Both typecheck projects are clean now, and 151 unit tests pass.

Integration tests still unrun — Docker is not up locally, so testcontainers cannot start. CI covers them.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.decodeDrop.integration.test.ts
0xdeafcafe: Fixed in `30003fbff8` — and green now.

`pnpm check:feature-parity` reports `OK: 2517 enforced scenario(s) bound across 761 file(s)`, with both touched specs fully bound:

```
▸ specs/event-sourcing/groupqueue-decode-drop-durability.feature
  12/12 scenarios bound  ✓
▸ specs/event-sourcing/payload-store-blob-hardening.feature
  6/6 scenarios bound  ✓
```

On the diagnosis: the annotation was not pointing at nothing — `a missing-blob drop releases the absent blob's holder` did exist, in `groupqueue-decode-drop-durability.feature` rather than `payload-store-blob-hardening.feature`. What broke was vocabulary drift: WP4 renamed holders to leases everywhere in the code, the test name moved with it, and the spec did not. So I moved the spec to lease vocabulary (`blob lease is released`, and the scenario title) and matched the annotation, rather than binding a lease-retirement assertion to a holder-named scenario.

The two Track 7 scenarios added for the durable-retention fix are bound too, so the count went up rather than being worked around.

langwatch-agent:
Following up on finding 1, with a correction and something useful.

I said the lifecycle policy for these objects did not exist. That was accurate for the shared storage bucket, but incomplete: the infra half has since been built. Saas commit `8640362a` adds a dedicated `langwatch-{env}-group-queue` bucket in `infrastructure/s3-queue-payloads.tf`, with an expiration rule scoped twice over (dedicated bucket plus a `temp-tier-3-offload/` prefix) and versioning deliberately left off so expiration is a terminal delete rather than a delete marker. That is a better design than the prefix filter I was implicitly asking for, and it explicitly reasons about not being one code change away from matching customer datasets.

Two things that keep finding 1 open rather than closing it.

First, that commit states the bucket is inert until the corresponding public-repo change lands, and I can confirm this branch is not that change: neither `temp-tier-3-offload` nor any dedicated-bucket config appears anywhere under `langwatch/src` on `feat/blob-lease-lifecycle`, and the same is true on main. So at the moment this PR removes eager S3 reclaim, the objects still go to the old destination, where no rule expires them. The infra is ready and waiting for the app side, which makes the gap a wiring task rather than a design question.

Second, the BYOC point is confirmed by that commit's own reasoning rather than just my inference: it notes the resolver picks the tenant bucket first for BYOC projects, so these internal queue payloads are currently written into customer-owned buckets. A lifecycle rule on our dedicated bucket does not reach those, so the switch to the dedicated bucket is what actually resolves it, not the rule alone.

So the ask narrows: either land the destination switch alongside this change, or sequence this behind it. Happy to re-verify once they are paired.

langwatch/src/server/event-sourcing/queues/groupQueue/envelopeBlobLifecycle.ts
0xdeafcafe: Agreed, and the finding is broader than this PR: the reclaim removal is inherent to the lease design, not to the storage slice — `reclaimOrphanedS3` is gone and `TieredBlobStore.delete` has no production callers either way.

Split the work: the lease slice is now #5947 (the incident hotfix), and the dedicated-bucket storage work is held back rather than shipped half-finished. #5947 documents this as an explicit known gap rather than burying it — Redis-tier blobs still expire on the 4-day backstop so the incident path is safe, but durable-tier payloads need either a lifecycle rule on the `<projectId>/` prefix or a queue-specific sweeper before the S3 tier carries real volume.

Not claiming this closes your objection — it scopes it. Deliberately not reintroducing a last-live-lease delete, since that is precisely the race that caused the incident.

langwatch/src/server/event-sourcing/queues/groupQueue/__tests__/groupQueue.decodeDrop.integration.test.ts
0xdeafcafe: Fixed by the split. The renamed annotation and its scenario churn belonged to the storage slice, which is not in the incident hotfix (#5947). On that branch the group-queue suites are green: 15 unit files / 142 tests and 9 integration files / 239 tests, including the `no-retention-gc` AC16 guard that was red here.

#5947 also adds the end-to-end regression test this PR was missing — two siblings staged on one content-addressed hash, first released, survivor's blob asserted live. The two halves of the fix were each asserted alone before, so nothing pinned their composition, which is the actual bug.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueueStorage.ts
```diff
@@ -0,0 +1,180 @@
...
+export async function resolveGroupQueueStorageDestination(
+  projectId: string,
+): Promise<GroupQueueStorageDestination> {
+  const prefix = queuePayloadPrefix();
+  const bucket = queuePayloadBucket();
+  if (bucket) {
+    return { kind: "s3", bucket, prefix };
+  }
+  return { ...(await resolveProjectStorageDestination(projectId)), prefix };
+}
```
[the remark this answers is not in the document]
0xdeafcafe: Valid, and not fixed here — removed instead. `groupQueueStorage.ts` does not exist in the incident hotfix (#5947); the whole dedicated-bucket slice is held back. So the BYOC fallback you flagged ships nowhere for now.

When that work returns it should land with the bucket mandatory or startup failing closed, as you say — the current `z.string().optional()` with no cross-field refinement means a misconfigured install surfaces as per-job transient errors under load rather than at boot.

0xdeafcafe:
Split for the production incident. The lease slice is now **#5947** — same mechanism, minus the dedicated-bucket storage work that was generating this PR's blockers (the `TS2322` on `groupQueue.ts:323`, the red `no-retention-gc` AC16 guard, the file-tier lifecycle gap). None of those touch the `missing_blob` failure.

#5947 also adds the end-to-end regression test that was missing here, and is green: 15 unit files / 142 tests, 9 integration files / 239 tests.

The three review threads here are answered and resolved. Suggest this PR stays open for the storage half, rebased once #5947 lands — the durable-tier retention question raised in review is real and still needs an owner.

0xdeafcafe:
Superseded — this work was split into two focused PRs on `main`:

- **#5947** — the lease-based blob lifecycle (the actual fix for the prod `missing_blob` incident: per-holder leases so a sibling completing can't strip a co-staged blob).
- **#5948** — the post-completion cleanup guard (stacked on #5947).

The only part of this PR not carried into those two is the S3 storage-destination slice (`groupQueueStorage.ts` + chart/env changes), which carried its own unresolved durable-tier-reclaim question and isn't needed for the incident fix. It can return as its own follow-up. Closing to avoid a third overlapping group-queue branch on `main`. Branch left intact.

PR #5911, opened 2026-07-18
feat(automations): process-manager substrate + automations pipeline; legacy outbox retired (ADR-052)

0xdeafcafe:
Automations now run on the ADR-052 process-manager substrate, and the `ReactorOutbox` audit shadow that was bloating production is retired. 358 files, +12.8k/-14.2k - a net deletion, which is the point.

## What changed

- **Process-manager substrate (ADR-052).** A staged, typed builder (`.state().intent().on().onWake().outbox()/.schedule()`), a transactional inbox/state/outbox runtime on Postgres, GroupQueue delivery adapters, a revision-fenced wake worker, and a leased intent dispatcher. Topology is authored inline in each pipeline; only executor deps are injected.
- **The automations pipeline.** Deterministic `recordTriggerMatch` idempotency, ID-only audit projection, per-trigger FIFO grouping, the settlement PM, a scheduled graph-alert sweep PM, and a daily webhook-delivery-prune PM that also handles settlement outbox retention. Zero automation CronJobs remain in the Helm chart.
- **Reactors → subscribers.** The six automation outbox reactors are now lightweight trace/evaluation subscribers plus PM intent handlers. Origin guards, settlement/cadence behaviour, `TriggerSent` claim-after-send, graph absence detection, and retry/cap behaviour all carry over. Settlement-cap overflow flushes the oldest matches as immediate dispatch intents - degraded batching under extreme load, never loss.
- **Webhook channel (ADR-040).** Encrypted kept-header secrets, SSRF-fenced delivery, test-fire protection, feature gating (`release_webhook_automations`, default off), and a slim `WebhookDelivery` facts table: outcome, status, latency, capped error, and the receiver's truncated failure response stored verbatim. No redaction - what an endpoint echoes is its own output, and ADR-040 §6 records the decision. Our request content (URL, headers, body) is never stored in any form; rows are pruned after 30 days.
- **`@langwatch/automations` workspace package.** Provider defs, cadences, and the Liquid templating engine extracted to `langwatch/packages/automations` (prisma-free via mirrored enums with a parity test) so cli/mcp/app/web can share one implementation. Provider secret handling is a three-way split: pure shared defs in the package, React halves in `features/automations/providers`, persist/redact hooks on ServerDefs in app-layer. No secret code reachable from the client bundle.
- **Service/repo discipline.** The automations tRPC router goes through services only - no raw prisma left in the router.
- **`ReactorOutbox` retired, carefully.** The model and legacy outbox machinery (`.withOutbox`, heartbeat/leader-lock path) are gone from the code, but the table `DROP` is deliberately *not* in this PR - the legacy table stays for a drain/compatibility release and the contract migration ships separately once old workers are gone.
- **Typecheck caching.** Explicit tsbuildinfo paths under `node_modules/.cache/tsbuildinfo` plus an actions/cache step in CI. Warm checks drop from ~13-19s to ~1.5s.
- **Docs.** ADR-052, ADR-040, ADR-037 refreshed to match what actually shipped; ADR-039 marked superseded; feature specs updated.

## Why

Automation dispatch now uses GroupQueue for every delivery hop and Postgres only for the two durability promises that need it: scheduled wakes and leased intents. That keeps effectively-once subscriber handoff, transactional PM consumption, durable settlement across Redis loss, and per-trigger ordering - without writing an audit copy of every event to Postgres on the way through.

### The settle-window invariant

The process-manager path uses the window-bucketed idempotency key `${triggerId}:${traceId}:${settleWindowBucket}`. The bucket combines the configured `traceDebounceMs` with `floor(occurredAt / max(traceDebounceMs, 1))`; including the width keeps rounds isolated if the trigger's debounce setting changes. Every input derives from the committed event and trigger config, never handler wall-clock, so a redelivered event re-sends byte-identical commands whose events collapse on idempotencyKey.

Actual sends stay gated by the permanent `TriggerSent(triggerId, traceId)` claim, so a notify trigger cannot send the same trace twice across windows. Persist intents include the settle-window bucket in their durable outbox identity, giving dataset and annotation actions a fresh evaluation chance when the first round fails filters and later trace state passes them. A successful persist action still records the claim and stays at-most-once.

## Review rounds

Three so far.

1. **Human + agent round.** All 16 threads (11 from me, 5 langwatch-agent P1/P2s) addressed and resolved with inline replies. Copilot and CodeRabbit both bailed - the PR exceeds their file caps.
2. **Multi-agent audit + fix batch (`3484756967`).** A 16-reviewer sliced audit produced 51 verified findings (1 P0, 23 P1); every P0/P1 was fixed in one batch of 20 parallel packets. The P0: the outbox prune tripped the multitenancy guard on every tick, now a raw DELETE carrying the sanctioned `-- @tenancy:` marker with cross-project regression tests on both store adapters. The rest: RFC 7230 header-name token validation (smuggling attempts are dropped, not mangled), Block Kit markdown/header sanitisation, header-injection strips for `trigger.name` plus a central one at the mailer boundary, fail-closed redaction in the graphs router, `triggerId` in the daily email-cap dedup key, terminal webhook failures keeping the incident claim (no re-POST loop to dead endpoints), the webhook prefill no longer racing the feature flag, the OSS→EE dependency inversion, origin-guard single-sourcing, and ten new test suites. A Kimi K3 cross-review of the batch contributed five of those hardenings; its two remaining notes (typed-error unwrapping for terminal dispatch classification, armSchedule retry policy) are tracked as follow-ups. P2/P3 findings are in the review report and deliberately deferred.
3. **The verbatim decision (`408ab2db6d`).** The delivery-log scrub is gone entirely. Masking what a receiver echoes would hide exactly what the log exists to show, and exact-match scrubbing is evadable anyway - so the boundary is drawn one level up (request content never persisted, 30-day prune) and ADR-040 §6 records it.

## Validation

- Full `tsgo --noEmit` clean on both the app and tests projects; package `tsc --noEmit` clean.
- 440 tests green across the 45 touched suites in the final sweep (event-sourcing, automations app-layer, providers, templating, drawers, delivery); ~1,600 across the wider rework.
- `pnpm check:feature-parity` - 2,514 enforced scenarios bound.
- CI green on `408ab2db6d`: required checks (validate title, typecheck, build), langwatch-app-ci, and e2e-ci (one rerun - the readiness poll expired 5 seconds before the API finished booting; all pipelines registered cleanly). The only red is sdk-python-ci, which fails on an OpenAI quota error in an examples test and touches nothing in this PR.

## Deployment Impact

- **Env vars:** none added or changed (`.env.example` untouched). Webhook header secrets reuse the existing `CREDENTIALS_SECRET`.
- **Helm values:** `cronjobs.jobs.alertTriggers` is removed from `values.yaml` - the `/api/cron/triggers` endpoint no longer exists. Operators overriding that key have a harmless dangling override. No new values.
- **Vanilla `helm install`:** the alert-triggers CronJob is no longer rendered; custom-graph threshold alerts and the daily webhook delivery-log prune both run on the workers Deployment (`workers.enabled`, on by default). Two new Postgres migrations run on app start (process-manager tables + `WebhookDelivery`). The legacy `ReactorOutbox` table is **not** dropped in this release - old worker replicas keep working during a rolling deploy; the `DROP` ships as a separate contract migration next release.
- **BYOC dataplane:** webhook automations make outbound HTTPS (443) calls from worker pods to customer endpoints, behind SSRF protection; egress policies that block outbound traffic from workers will block the feature. The channel is gated behind the `release_webhook_automations` feature flag (default off), so nothing fires until it is enabled.

https://claude.ai/code/session_01Rxp9AeY1d18xEW7KnNCEt1


charts/langwatch/README.md
```diff
@@ -511,10 +511,10 @@ npx @bitnami/readme-generator-for-helm --readme ./README.md --values values.yaml
 | `cronjobs.jobs.topicClustering.enabled`  | Enable topic clustering job.                       | `true`                                |
 | `cronjobs.jobs.topicClustering.schedule` | Cron schedule.                                     | `0 0 * * *`                           |
 | `cronjobs.jobs.topicClustering.endpoint` | Endpoint path.                                     | `/api/cron/schedule_topic_clustering` |
-| `cronjobs.jobs.alertTriggers`            | Alert triggers job.                                |                                       |
-| `cronjobs.jobs.alertTriggers.enabled`    | Enable alert triggers job.                         | `true`                                |
-| `cronjobs.jobs.alertTriggers.schedule`   | Cron schedule.                                     | `*/3 * * * *`                         |
-| `cronjobs.jobs.alertTriggers.endpoint`   | Endpoint path.                                     | `/api/cron/triggers`                  |
+| `cronjobs.jobs.webhookDeliveryCleanup`          | Webhook delivery-log prune job.             |                                       |
```
0xdeafcafe: we should be removing cron jobs fully

langwatch/src/server/event-sourcing/pipelines/automations/pipeline.ts
```diff
@@ -0,0 +1,31 @@
...
+    .withAggregateType("trigger")
+    .withMapProjection(
+      "automationAudit",
+      createAutomationAuditMapProjection({ store: deps.automationAuditStore }),
+    )
+    .withCommand("recordTriggerMatch", RecordTriggerMatchCommand, {
+      serializeByAggregate: true,
+    })
+    .withProcessManager("triggerSettlement", deps.triggerSettlement)
+    .withProcessManager("graphAlertSweep", deps.graphAlertSweep)
```
0xdeafcafe: the `deps` here, should be defined inline... like the ADR

langwatch/src/server/event-sourcing/pipelines/automations/pipeline.ts
```diff
@@ -0,0 +1,31 @@
...
+  return definePipeline<AutomationEvent>()
+    .withName("automations")
+    .withAggregateType("trigger")
+    .withMapProjection(
+      "automationAudit",
+      createAutomationAuditMapProjection({ store: deps.automationAuditStore }),
+    )
+    .withCommand("recordTriggerMatch", RecordTriggerMatchCommand, {
+      serializeByAggregate: true,
+    })
```
0xdeafcafe, line 105: there is a huge confusion of where things should go.

src/automations/providers
src/features/automations
src/server/triggers
src/server/app-layer/triggers
src/server/app-layer/automations
src/server/triggers
src/shared/templating

all of this is a mess.

process manager, outbox, intents, etc. etc. should all be in:  
src/server/event-sourcing/pipelines/automations

ALL frontend should be in:  
src/features/automations

ONLY shared code should be in:
src/shared/templating - maybe src/shared/automations

ALL Services, Repos, etc. for automations should be in: src/server/app-layer/automations (not triggers, old name, it's gone)

Then, all webhook logic in src/server/app-layer/automations too for now, until we move it out somewhere else.

langwatch/prisma/migrations/20260718100100_create_webhook_delivery/migration.sql
```diff
@@ -0,0 +1,44 @@
+-- ADR-040 §6: per-attempt webhook delivery log — the drill-down behind a
+-- webhook automation's "recent fires". One row per HTTP attempt; all attempts
+-- of one logical fire share `dispatchId` (== the X-LangWatch-Event-Id). Header
+-- values are stored REDACTED (auth/signature masked to '***'); the response
+-- snippet is size-capped by the writer. Rows are pruned after 30 days.
+
+-- CreateEnum
+CREATE TYPE "WebhookDeliveryOutcome" AS ENUM ('success', 'retryable', 'terminal', 'pending');
```
0xdeafcafe, line 9: > values are stored REDACTED 

just don't store them for now. store if there was an error, log sensible things about it. don't reinvent the wheel.

langwatch/prisma/migrations/20260718100100_create_webhook_delivery/migration.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- CreateTable
+CREATE TABLE "WebhookDelivery" (
```
0xdeafcafe, line 12: review the spec, i think we wanted a slim facts table, no?

langwatch/ee/governance/subscribers/traceAlertTriggerMatch.subscriber.ts
```diff
@@ -0,0 +1,50 @@
...
+    event: TraceProcessingEvent,
+    context: TriggerContext<TraceSummaryData>,
+  ): Promise<void> => {
+    if (!passesTraceOriginGuards(event, context.state)) return;
+    const triggers = await deps.triggers.getActiveTraceTriggersForProject(
+      context.tenantId,
+    );
+    for (const trigger of triggers) {
+      if (classifyTriggerFilters(trigger.filters).hasEvaluationFilters) continue;
+      await deps.recordTriggerMatch.send({
```
0xdeafcafe, line 33: durable events must be idempotent as fuck

langwatch/prisma/schema.prisma
0xdeafcafe, line 376: what do we do here? i think we can fold this down into something else too, no? what did the spec say

langwatch/src/automations/providers/definitions/webhook/secret.ts
```diff
@@ -0,0 +1,93 @@
+import { decrypt, encrypt } from "~/utils/encryption";
+import {
+  WEBHOOK_HEADER_VALUE_KEPT,
+  type WebhookActionParams,
+} from "./shared";
+
+/**
+ * Server-only handling of webhook header secrets (ADR-040 §3). Header values
```
0xdeafcafe: 1) should be in app-layer
2) think we're re-inventing the wheel here? we have cred store
3) holy bad code this is shit

langwatch/src/automations/providers/definitions/webhook/server.ts
```diff
@@ -0,0 +1,8 @@
+import { TriggerAction } from "@prisma/client";
```
0xdeafcafe: wat

langwatch/packages/automations/src/providers/webhook.ts
```diff
@@ -0,0 +1,147 @@
...
+export function isReservedWebhookHeader(name: string): boolean {
+  const lower = name.trim().toLowerCase();
+  return (
+    RESERVED_HEADER_NAMES.has(lower) || lower.startsWith(RESERVED_HEADER_PREFIX)
+  );
+}
+
+/** Drops reserved keys and entries with empty names/values. Header values are
+ *  stripped of CR/LF so a stored header can never smuggle a second one. */
+export function sanitizeWebhookHeaders(
```
0xdeafcafe, line 48: this is awful over the top spaghetti code.

langwatch/src/features/automations/providers/registry.ts
```diff
@@ -9,6 +9,12 @@ import emailClient, { type EmailSlice } from "./definitions/email/client";
 import emailShared, { type EmailPreview } from "./definitions/email/shared";
 import slackClient, { type SlackSlice } from "./definitions/slack/client";
 import slackShared, { type SlackPreview } from "./definitions/slack/shared";
+import webhookClient, {
```
0xdeafcafe, line 12: the line of splitting code here is not good. secrets so close to client, for starters.

langwatch/prisma/migrations/20260718000000_drop_reactor_outbox/migration.sql
```diff
@@ -0,0 +1,6 @@
+-- ADR-052: automations dispatch moved onto the process-manager substrate
+-- (ProcessManagerInbox/Instance/Outbox). The legacy ReactorOutbox audit
+-- table has no readers left; in-flight rows at cutover are pending
+-- notifications whose loss window the ADR accepts (minutes, alert-class).
+DROP TABLE IF EXISTS "ReactorOutbox";
```
langwatch-agent: **P1 — This cutover destroys live legacy work and is unsafe during a rolling deployment.** `ReactorOutbox` is the durable retry queue used by the current release. This migration runs when a new app pod starts, before old worker replicas are necessarily gone; dropping the table can lose queued alert/persist actions and make old workers fail their next read. The new queue tombstone also deliberately ACK-drops the corresponding Redis jobs, so this is not recoverable by retry. Keep the legacy schema and consumer compatibility for a drain/compatibility release, then contract only after rows and old workers are gone.

langwatch/src/server/event-sourcing/pipelines/automations/process-manager/triggerSettlement.process.ts
```diff
@@ -0,0 +1,201 @@
...
+      settleWindowBucket: settleWindowBucket({
+        occurredAt: at,
+        traceDebounceMs: view.traceDebounceMs,
+      }),
+    },
+  };
+  let overflowDropped = previousState.overflowDropped;
+  const traceIds = Object.keys(pendingMatches);
+  if (traceIds.length > MAX_PENDING_MATCHES) {
+    const oldestFirst = traceIds.sort(
```
langwatch-agent, line 76: **P1 — The sole durable pending-work state silently discards customer matches at 1,001 entries.** A high-volume trigger using an hourly/digest cadence (or a delayed worker) reaches this branch before its wake; the oldest trace is deleted, no dispatch intent is created for it, and the durable inbox prevents replaying the original event as a recovery path. `logOverflow` only logs the loss. Do not evict pending matches from the correctness path: keep them durable and drain in bounded batches, or make a durable per-match outbox/queue the source of truth.

langwatch/src/features/automations/ViewAutomationDrawer.tsx
```diff
@@ -75,9 +76,18 @@ export function ViewAutomationDrawer({
     { triggerId: automationId, projectId: project?.id ?? "", limit: 20 },
     { enabled: !!project?.id },
   );
+  // ADR-040 §6: the per-attempt delivery log, only for webhook automations.
+  const webhookDeliveriesQuery = api.automation.getWebhookDeliveries.useQuery(
```
langwatch-agent, line 80: **P2 — This new unconditional hook property access breaks the existing drawer integration suite.** Even with `enabled: false`, the test API mock must resolve `automation.getWebhookDeliveries.useQuery`; it does not, so all three `ViewAutomationDrawerConditions` tests throw before assertions. Extend the shared API mock (and keep the query disabled for non-webhook actions) so the suite exercises the component again.

langwatch/src/server/app-layer/automations/delivery/sendWebhook.ts
```diff
@@ -0,0 +1,201 @@
...
+import { sendHttpDestination } from "./httpDestination";
+
+/**
+ * The webhook channel's SSRF policy (ADR-040 §4): private-IP / localhost
+ * blocking is FORCED ON regardless of the global BLOCK_LOCAL_HTTP_CALLS
+ * toggle — a customer-supplied URL fired from our workers must never reach
+ * `10.x` / `localhost`, even in deployments that relax the toggle for their
+ * own internal integrations.
+ */
+const validateWebhookUrl = createSSRFValidator({
```
langwatch-agent, line 24: **P2 — Importing this module now breaks both HTTP proxy integration suites.** Those suites mock `~/utils/ssrfProtection` with only `ssrfSafeFetch`; evaluating this module needs `createSSRFValidator`, so Vitest aborts before collecting tests. Make those mocks partial mocks of the actual module, or provide a `createSSRFValidator` stub alongside the existing fetch mock.

langwatch/src/server/triggers/__tests__/deliverWebhook.unit.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        dispatchId: "evt_abc",
+        requestMethod: "POST",
+        requestUrl: "https://example.com/hook",
+        responseStatus: 201,
+        outcome: "success",
+      });
+      // The secret header value is masked in the log row.
+      expect(rows[0]!.requestHeaders).toEqual({
+        Authorization: "***",
+        "X-Trace": "t1",
```
langwatch-agent: **P2 — The committed delivery test contradicts the new redaction contract and currently fails.** `redactHeadersForLog` deliberately masks every custom header value, but this expectation retains `X-Trace: t1`. The focused test run fails exactly here. Update the assertion to require every value to be redacted and retain an explicit header-name-preservation check.

langwatch-agent (changes requested):
Request changes — 2 P1s and 3 P2s are detailed inline. The existing migration-order thread is also a release blocker and should be resolved rather than duplicated here.

charts/langwatch/README.md
0xdeafcafe: Done — both automation CronJobs are removed entirely: `alertTriggers` is gone (custom-graph threshold alerts fire from the event-sourced worker path, ADR-034) and the delivery-log prune runs as a daily scheduled process manager (ADR-040 §6). Only the unrelated `topicClustering` cron remains, with a NOTE in values.yaml documenting the removal.

langwatch/src/server/event-sourcing/pipelines/automations/pipeline.ts
0xdeafcafe: Done — the PM topology (states, intents, evolve/wake handlers, outbox tuning) is authored inline in `pipeline.ts` per the ADR-052 builder shape. Only executor deps (audit store, dispatch, sweep, prune) are injected; the `*PM(deps)` factories are gone.

langwatch/src/server/event-sourcing/pipelines/automations/pipeline.ts
0xdeafcafe: Restructured to exactly this split: PM/intents/outbox/wiring under `src/server/event-sourcing/pipelines/automations`, all frontend under `src/features/automations`, services/repos/delivery under `src/server/app-layer/automations`, and truly shared code (provider defs, cadences, templating) extracted into the `@langwatch/automations` workspace package (`langwatch/packages/automations`) so cli/mcp/app/web can share it. `src/automations`, `src/server/triggers`, and `src/shared/templating` no longer exist.

langwatch/prisma/migrations/20260718100100_create_webhook_delivery/migration.sql
0xdeafcafe: Landed on the industry baseline we discussed after this: our request content (URL, headers, body) is never stored in any form — no redacted copies. On failure we keep status, latency, a capped error message, and a truncated copy of the *receiver's* response (with any of our configured header values scrubbed if echoed), plaintext like GitHub/Stripe, pruned after 30 days by a scheduled PM.

langwatch/prisma/migrations/20260718100100_create_webhook_delivery/migration.sql
0xdeafcafe: Done — slim facts table now: outcome, responseStatus, latencyMs, capped error, truncated failure response, dispatchId (== X-LangWatch-Event-Id), firedAt. Nothing else; 30-day prune.

langwatch/ee/governance/subscribers/traceAlertTriggerMatch.subscriber.ts
0xdeafcafe: Done — every input to the idempotency key (triggerId, traceId, settle-window bucket from occurredAt + traceDebounceMs) now derives from the committed event and trigger config, never handler wall-clock. A redelivered event re-sends byte-identical commands whose events collapse on idempotencyKey in the store. Contract documented at the send site.

langwatch/prisma/schema.prisma
0xdeafcafe: Folded — `ReactorOutbox` is gone from the schema, replaced by the generic ProcessManagerInbox/Instance/Outbox substrate every PM shares (the generic intents shape you described). The table DROP itself is deferred one release so old workers can drain (per the rolling-deploy P1 below).

langwatch/src/automations/providers/definitions/webhook/secret.ts
0xdeafcafe: This file is gone. Secret handling now lives in the provider's server half (`src/server/app-layer/automations/providers/webhook/server.ts`) as `persistActionParams`/`redactActionParams` hooks on the ServerDef, using the shared `encrypt`/`decrypt` (CREDENTIALS_SECRET) — the same discipline as the Slack bot token.

langwatch/src/automations/providers/definitions/webhook/server.ts
0xdeafcafe: Gone — rewritten as part of the provider three-way split; the server half is now a plain ServerDef with persist/redact hooks in app-layer.

langwatch/packages/automations/src/providers/webhook.ts
0xdeafcafe: Rewritten — the shared webhook module is now small pure helpers (reserved-header check, header sanitisation with CR/LF stripping, https-only URL shape check) plus the kept-sentinel contract, each documented. The persistence/encryption spaghetti moved behind the ServerDef hooks in app-layer.

langwatch/src/features/automations/providers/registry.ts
0xdeafcafe: Split three ways: pure shared defs in `@langwatch/automations`, React halves in `src/features/automations/providers/*/client`, and secret-touching server halves in `src/server/app-layer/automations/providers`. The client registry imports only shared+client modules — no encrypt/decrypt or server code reachable from the browser bundle; a parity test keeps the three registries aligned.

langwatch/prisma/migrations/20260718000000_drop_reactor_outbox/migration.sql
0xdeafcafe: Agreed — this migration is deleted. The legacy `ReactorOutbox` table stays for a drain/compatibility release; the DROP ships as a separate contract migration once old workers are gone.

langwatch/src/server/event-sourcing/pipelines/automations/process-manager/triggerSettlement.process.ts
0xdeafcafe: Fixed — cap evictions no longer discard matches. Overflow now flushes the oldest matches as immediate dispatch intents through the durable outbox (degraded batching under extreme load, never loss), plus a `logOverflow` intent for visibility. Covered by unit tests and the 'Pending matches are bounded without losing matches' scenario.

langwatch/src/features/automations/ViewAutomationDrawer.tsx
0xdeafcafe: Fixed — the shared API mock now provides `automation.getWebhookDeliveries.useQuery` (and the query stays disabled for non-webhook actions). All drawer suites pass: 35/35 across AutomationDrawer, ViewAutomationDrawer, ViewAutomationDrawerConditions.

langwatch/src/server/app-layer/automations/delivery/sendWebhook.ts
0xdeafcafe: Fixed — those suites now partial-mock `~/utils/ssrfProtection` via `importOriginal`, overriding only `ssrfSafeFetch`, so `createSSRFValidator` resolves from the real module. Both proxy suites pass (40/40).

langwatch/src/server/triggers/__tests__/deliverWebhook.unit.test.ts
0xdeafcafe: Fixed — the test moved with the module and now asserts the final contract: our configured header values are always scrubbed when echoed in the receiver's response (names preserved). Suite green.

langwatch/prisma/migrations/20260718100100_create_webhook_delivery/migration.sql
0xdeafcafe: Final iteration (408ab2db6d): the scrub is gone entirely — the receiver's failure response is stored verbatim (truncated, 30-day prune). ADR-040 §6 now records the decision: what the endpoint echoes is its own output, masking would hide exactly what the log exists to show, and our request content is never stored in any form.

0xdeafcafe:
Completed several rounds of adversarial code reviews of this with kimi k3 high, coded sol xhigh, and fable high. i'm confident with the pr.

PR #5907, opened 2026-07-18
feat: expand operational telemetry and harden gateway egress

0xdeafcafe:
## Summary
- add unified metrics for fold/map/state projections, subscribers, reactors, process managers, and transactional outbox delivery
- emit sparse slow/failure/retry operational logs inside active spans so trace_id/span_id and operational identifiers correlate in Loki and Tempo
- standardize LangWatch service names and deployment resource attributes across Node and Go services
- reject customer-configured provider endpoints that resolve to Kubernetes, metadata, private, reserved, or other non-public networks
- preserve trace continuity across gateway and worker boundaries

## Safety boundary
- no Redis application instrumentation or Redis event-sourcing tracing was added
- logs retain organization/project/user/tenant/aggregate/event identifiers but do not intentionally record prompts, message bodies, event payloads, credentials, or cache values

## Deployment Impact
- Environment variables: no new required variables. Existing OTel configuration now emits canonical `service.name`, `service.version`, and `deployment.environment.name` resource attributes.
- Helm values: none added or changed.
- Default vanilla `helm install`: remains functional without OTel endpoints; telemetry providers stay no-op when no endpoint is configured.
- BYOC/dataplane: customer-configured AI provider base URLs must use HTTPS and resolve exclusively to public addresses. Kubernetes service DNS, metadata endpoints, loopback, link-local, private/reserved networks, and DNS failures are rejected closed.
- Rollout: app/service deployment only through the normal pipeline; no Terraform apply is part of this PR.

## Validation
- `pnpm --dir langwatch run typecheck`
- focused event-sourcing metrics/process-manager/outbox/trace-continuity tests: 67 passed after standard sanitized OTel exception events were restored
- Go provider/gateway tracer/langy adapter tests passed
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added protection against unsafe customer-configured provider endpoints, including private networks, metadata services, unsupported schemes, and optional HTTPS enforcement.
  * Added configurable endpoint allowlists and gateway security settings.
  * Added internal gateway service and configurable autoscaling, annotations, and provider egress ports.
  * Added detailed metrics and slow-operation warnings for projections, subscribers, process managers, and outbox delivery.

* **Bug Fixes**
  * Improved trace linking and service identification across gateway and worker telemetry.
  * Standardized telemetry environment attributes and default service names.

* **Documentation**
  * Updated observability and endpoint security configuration examples.

PR #5913, opened 2026-07-18
fix: keep process failure diagnostics generic

0xdeafcafe:
## Summary

- replace blacklist-based process-manager/outbox error-message handling with one allowlist-based generic diagnostic
- retain only trusted built-in error categories; omit raw exception text and custom names from logs and OpenTelemetry
- add unit and real-span regressions for bearer tokens, prompt-derived content, password text, custom error names, and non-Error throws

## Why

Follow-up to #5907. The resolved diagnostic-preservation feedback led to raw `error.message` being exported. A later P1 review demonstrated that an `Authorization: Bearer ...` value survived the blacklist and that arbitrary customer or prompt content could reach logs and OpenTelemetry. This PR fixes that exposure without changing process retry or throw behavior.

## Original feedback traceability

- P1 diagnostic exposure (`PRRT_kwDOKRXhvM6SATl8`): fixed by `toSafeFailureDiagnostic` and both process-manager/outbox call sites.
- Outside-diff metrics call-site coverage: current `main` still has helper-level coverage only in `metrics-instrumentation.unit.test.ts`; intentionally unchanged per the current scope.
- DNS rebinding/dial-time feedback: intentionally deferred per the current scope for a dedicated design.

## Validation

- focused Vitest suite: 4 files, 32 tests passed
- `pnpm run typecheck`
- `pnpm run typecheck:tests`
- `pnpm dlx @biomejs/biome@2.4.14 check` on the five touched files
- `git diff --check`

## Summary by CodeRabbit

* **Security & Privacy**
  * Sensitive error details are now consistently removed from diagnostics, logs, and telemetry.
  * Failure messages use a safe, standardized description instead of exposing secrets or customer information.

* **Reliability**
  * Error categories remain available for trusted built-in failure types.
  * Retry and process-evolution failure reporting now preserves sanitized exception details while maintaining trace continuity.

### DIFF
--- a/langwatch/src/server/event-sourcing/process-manager/__tests__/failureDiagnostic.unit.test.ts
+++ b/langwatch/src/server/event-sourcing/process-manager/__tests__/failureDiagnostic.unit.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "vitest";
+
+import { toSafeFailureDiagnostic } from "../failureDiagnostic";
+
+describe("toSafeFailureDiagnostic", () => {
+  it.each([
+    "Authorization: Bearer sk-live-secret-token",
+    "prompt-derived customer text: tell me my medical history",
+    "password=hunter2 upstream request failed",
+  ])("omits untrusted error content: %s", (sensitiveMessage) => {
+    const error = new Error(sensitiveMessage);
+    error.name = `CustomError-${sensitiveMessage}`;
+
+    const diagnostic = toSafeFailureDiagnostic(error);
+    const serialized = JSON.stringify(diagnostic);
+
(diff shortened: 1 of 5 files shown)


langwatch-agent (commented):
Re-review complete: the P1 diagnostic leak from #5907 is resolved. The allowlist emits only trusted built-in categories plus a constant message at both process-manager and outbox boundaries; the focused regression suite passes (15 tests), including real exported-span assertions for bearer-token and prompt-derived failures. No blocking findings in this follow-up. The remaining `sdk-python` failure is an unrelated live-example OpenAI quota error, not caused by this TypeScript-only diff.

PR #5918, opened 2026-07-18
fix(event-sourcing): stop folds double-counting on queue redelivery

0xdeafcafe:
## What changed

Fold projections could double-count on an ordinary queue retry. Four of seven accumulate — `spanCount + 1`, token and cost sums, id appends — and delivery is at-least-once, so a job that fails **after** its state was stored is re-dispatched and re-applies the same events.

The window is wider than it looks: everything between the durable write and the ack can throw. Reactor dispatch (`projectionRouter.ts:1570`, the dominant path — a Redis write plus possibly a blob write), inline reactor handlers, the ack itself, and the blob release.

### The fix

The cache entry now carries the ids of the events folded into it, **scoped to a retry chain**:

- a **fresh delivery replaces** the set — the queue holds one active batch per group, so a fresh delivery means the previous batch acked and its ids can never return
- a **retry merges** into it — those events can still come back

Accumulation is load-bearing: a retry that discarded the set would re-apply what an earlier attempt folded. Keeping only the last write fails on the second retry of a chain, and there's an integration test for exactly that.

The chain is identified by a **group-scoped attempt counter**, because re-staged siblings carry no `__attempt` of their own — decoding and re-encoding up to 499 payloads to stamp one would be absurd, and the chain is a property of the group anyway. This also fixes a pre-existing bug where a sibling leading the next batch silently restarted the 25-attempt budget.

### Also fixed

- **Post-completion cleanup no longer re-runs the job.** Everything after `scripts.complete()` sat in the handler's `try`, so a failure in *cleanup* re-staged siblings and retried a job whose slot was already completed — delivering the batch twice. A blob release failing on a brief S3 blip was enough.
- **Trace rooting split by lane.** Ingest lanes (folds, map projections, commands, reactors) stay rooted so a 330k-span ingest can't collapse into one never-exported mega-trace; subscribers and scheduled jobs continue the publishing trace so an automation reads end to end.

## Folded in

- #5885 map projection batching
- #5891 batch observability
- #5896 per-job root traces

Supersedes **#5908** and **#5909**: durability-gated cache eviction was built here and then removed. Its premise — that a miss proves the durable store authoritative — doesn't survive the deployment reality (the npx installer ships Redis `allkeys-lru`, Redis is one instance shared with the queues, the chart default is a single non-replicated ClickHouse). Its probe was also inert: `UpdatedAt` is `DateTime64(3)`, so a raw `max()` parsed to `NaN` and nothing could ever confirm. Recorded as rejected-with-reasons in ADR-053.

## Known limit

The applied-set lives in the cache entry, so eviction or Redis loss takes it too. That degrades to the behaviour that existed before it, not to something worse — and it's pinned by a test rather than left as folklore. Closing the cold path means making the folds idempotent; sequenced in `dev/docs/plans/fold-idempotency-plan.md`.

## Docs

~60 verified inaccuracies corrected. The worst: `ARCHITECTURE.md` asserted *"if state was already stored, the fold is effectively idempotent"* — the precise claim this PR exists because it's false. `groupQueue/ARCHITECTURE.md` described Redis operations the queue has never used (RPUSH/LPOP on a sorted set) and four wrong key shapes. ADR-034 attributed to ADR-021 a rule ADR-021 never made, which is the origin of the corpus-wide "folds are idempotent" assumption.

## Tests

17 integration cases against the **real GroupQueue and real Redis**: lifecycle, retry chains, FIFO ordering, tenant isolation, negative cases, and the cache-loss limit. Plus invariant tests pinning relationships between constants in different files — the group-counter TTL against the longest scheduled backoff (that one caught a bug where they were equal, so the counter expired mid-chain), and the applied-id cap against the coalesce ceiling.

Also a **storage-boundary guard**: `resolveProjectStorageDestination` returns a BYOC tenant's *own* bucket, which is right for customer data and wrong for internal data, and nothing in the signature says which you're holding. The guard pins both lists and fails on a new caller. Verified by planting one.

191 test files / 6949 tests green.

## Summary by CodeRabbit

- **New Features**
  - Added batch processing for compatible projection and handler operations, improving throughput.
  - Improved retry handling to prevent duplicate event application during queue redelivery.
  - Added safer parsing diagnostics that avoid exposing echoed input data.
  - Enhanced distributed tracing so queued jobs produce clearer, more focused traces.

- **Bug Fixes**
  - Improved retry-chain handling, completion counts, and recovery of partially failed batches.
  - Added fallback behavior for unavailable or unreadable cached projection state.

- **Documentation**
  - Updated architecture records and specifications to reflect current event-sourcing behavior and superseded guidance.


0xdeafcafe:
Cherry-picked the incident-relevant change from here — `c141962a10`, the post-completion cleanup guard — into **#5948**, stacked on #5947. That is ~50 lines in one file; this PR is 416.

The reasoning in your comment is what identified the window, and it is quoted in #5948's description.

One thing to fix here before this lands, unrelated to the incident: `clearGroupAttempt` is called from exactly one site — the post-completion success block — and not from `handleExhaustedRetries`, the non-retryable path, or `dropStagedJob`/`parkPoisonGroup`. With `GROUP_ATTEMPT_TTL_SECONDS = 1800`, a group that burns its 25-attempt budget leaves the counter behind, so a genuinely new job for that group within 30 minutes reads `attempt = 25`, fails the `attempt < maxAttempts` guard, and goes straight to exhausted on its first transient failure — with `retryBackoffMsFor({attempt: 25})` already at the 600s cap. Clearing on every terminal outcome rather than only success should do it.

#5948 excludes the retry-chain machinery entirely, so that defect is not on the incident path.

0xdeafcafe:
Superseded by #6016.

The defect is real and still live — `main`'s own post-store-failure telemetry (added by #5947) says "nothing on this path deduplicates by event id", which measures the window without closing it. But this branch could not be rebased: its `groupQueue.ts` calls `blobLifecycle.acquire`/`.release`/`.transfer`/`.reclaimOrphanedS3`, all replaced by leases in #5947. `envelopeBlobLifecycle.ts` wasn't even in the conflict list because this branch never touched it, so a merge would have silently taken main's lease version and left the caller invoking five methods that no longer exist.

#6016 cherry-picks the fold-side core unchanged (it applied cleanly — `main` never touched those files) and re-authors the group-scoped attempt counter against the lease-based lifecycle.

Dropped as already-landed or out of scope: map batching (#5945), the duplicate post-completion guard (#5948), the root-trace rider (#5896 — reopen separately if still wanted), the `payloadCount` stats change and its scenarios (written against Lua #5947 rewrote), `groupKey.ts`, the storage-boundary guard, the `ARCHITECTURE.md` rewrite (describes the pre-#5947 refcount model), and the ADR churn — including `ADR-056`, whose number #6005 already claims.

PR #5922, opened 2026-07-19
feat(security): shared SSRF classification + fail-closed tenant egress (ADR-053)

0xdeafcafe:
## Why

Outbound traffic to tenant-controlled destinations - automation webhooks, custom provider URLs, NLP HTTP nodes, the Langy proxy - was fenced by per-caller safeguards that had drifted apart. Two gaps mattered most:

- **Validate-then-redial.** The Go customer-endpoint validator resolved a hostname, checked the answer, then let the provider client resolve it again at dial time. A DNS-rebinding answer could pass validation as public and connect as private (a classic time-of-check/time-of-use gap). The same adapter is embedded by NLP, so guarding only the gateway did not close it.
- **Divergent rule sets.** Each path carried its own notion of what counts as a private or special-purpose address, so a range blocked in one place could slip through another.

This PR unifies the classification, closes the TOCTOU by pinning the dial to the validated address, and makes the hosted posture fail-closed. It also lands ADR-053, which records the target tenant-aware egress architecture and its phased rollout - the code here is the application slice of Phase 0.

## What changed

- **One shared SSRF rule set.** `pkg/ssrf` (Go) and `@langwatch/ssrf` (TS), held to a single conformance corpus (`address_vectors.json`). The gateway, Langy proxy, NLP HTTP block and the app all route through it, so there is no more per-caller drift.
- **Close the DNS-rebinding TOCTOU.** The Langy proxy and customer-endpoint path now resolve once, validate every answer, and pin the socket to the approved IP while keeping the original hostname for TLS SNI and certificate verification. Metadata, private, link-local, loopback, multicast, reserved and documentation ranges are rejected.
- **Fail-closed hosted gateway.** Hosted AI gateway instances refuse to boot unless both private-network blocking and HTTPS-only custom endpoints are enabled. Self-hosted stays permissive by explicit opt-out.
- **NLP HTTP block: wider deny set is opt-in.** Default self-hosted behaviour is unchanged except cloud metadata (including Azure WireServer `168.63.129.16`), which is now always blocked. Non-strict deployments log each address strict egress would refuse (range + RFC + `ALLOWED_PROXY_HOSTS` hint) so operators can allow-list before flipping the knob.
- **Consistent telemetry `service.name`.** `langwatch-app` plus `langwatch-service-*` across the Go services.
- **ADR + spec.** ADR-053 and `specs/security/tenant-aware-egress-isolation.feature` capture the target architecture and behavioural contract.

## Test plan

- [x] `go test ./services/aigateway/...`
- [x] `go test ./services/langyagent/adapters/egress`
- [x] `go test ./pkg/ssrf/...` and `@langwatch/ssrf` corpus tests
- [x] `gofmt` and `gopls check`
- [ ] Full production egress migration to the central Go service remains a follow-up rollout.

## Deployment impact

**Hosted config is now fail-closed (action required on hosted deploys).** The AI gateway refuses to boot unless `BLOCK_LOCAL_HTTP_CALLS` and `REQUIRE_HTTPS_CUSTOM_ENDPOINTS` are set. These are provisioned in `langwatch-saas#760`; any hosted gateway missing them will crash-loop until they are present. Self-hosters running the gateway must set them (or deliberately opt out) - see `.env.example`.

**Telemetry `service.name` renamed - update dashboards/alerts.** Go services now report `langwatch-service-aigateway` / `langwatch-service-langyagent` / `langwatch-service-nlpgo` (previously bare `aigateway` / `langyagent` / `nlpgo`); the app remains `langwatch-app` (older docs referenced `langwatch-backend`). Any Grafana dashboard, alert rule, saved query, or log/trace pipeline filtering on the old names must be updated. `OTEL_SERVICE_NAME` now overrides the name on the Go side too, matching the app.

**Broader SSRF deny set under `BLOCK_LOCAL_HTTP_CALLS`.** Customer endpoints, webhooks and HTTP nodes now also refuse CGNAT, benchmarking, documentation, NAT64, 6to4 and reserved ranges that some paths previously allowed. This is intended hardening; a hosted tenant relying on one of those would now be blocked. Self-hosters reach internal hosts via `ALLOWED_PROXY_HOSTS`.

**NLP HTTP block: wider deny set is opt-in.** Default self-hosted behaviour is unchanged (the wider special-purpose deny set applies only under the new strict-egress knob) **except** cloud metadata - including Azure WireServer `168.63.129.16` - which is now always blocked. Non-strict deployments log each address strict egress would refuse so operators can allow-list before enabling strict.

**No database migrations. No Helm chart changes in this repo** - the hosted env-var wiring lives in `langwatch-saas#760`.


pkg/ssrf/testdata/address_vectors.json
```diff
@@ -0,0 +1,63 @@
+{
```
0xdeafcafe, line 1: maybe add a workflow to test just this?

services/nlpgo/cmd/root.go
```diff
@@ -53,11 +53,19 @@ func Root(ctx context.Context, _ []string) error {
...
-		},
-	})
+	// One egress policy value, built once and shared by the HTTP block and
+	// the remote-attachment fetcher so the two can never drift apart.
+	ssrfOpts := httpblock.SSRFOptions{
+		AllowedHosts:     allowedProxyHosts,
+		StrictPublicOnly: cfg.Engine.EgressStrictPublicOnly,
+		Logger:           deps.Logger,
+	}
+	deps.Logger.Info("nlpgo_egress_policy",
```
0xdeafcafe: clog?

langwatch/packages/ssrf/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+	"compilerOptions": {
+		"target": "ES2022",
+		"module": "ESNext",
+		"moduleResolution": "Bundler",
+		"lib": ["ES2022"],
+		"strict": true,
```
0xdeafcafe, line 7: ts build info config, and incremental. make this a rule rule!

0xdeafcafe:
yet another token burning seshwari, multi-agent reviews done. approved, and now we go to prod.

services/aigateway/adapters/providers/customer_endpoint_ssrf.go
```diff
@@ -75,35 +54,26 @@ func defaultEndpointResolver(ctx context.Context, host string) ([]net.IP, error)
...
 }
 
-func isPublicEndpointIP(ip net.IP) bool {
+// endpointAddressError maps a literal or resolved customer-endpoint address
+// onto the shared pkg/ssrf rule set. Cloud metadata is refused unconditionally;
+// any other non-public address is refused when the policy blocks local egress
+// and the host is not explicitly allowlisted. Sharing pkg/ssrf keeps this
+// validator, the Langy egress proxy and the TypeScript app in lockstep on which
+// addresses are non-public — the same rule set, tested by one shared corpus.
+func endpointAddressError(ip net.IP, blockLocal, allowlisted bool) error {
```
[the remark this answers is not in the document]
0xdeafcafe: no

pkg/ssrf/testdata/address_vectors.json
0xdeafcafe: Added `.github/workflows/ssrf-conformance.yaml` — path-gated on `pkg/ssrf/**`, `langwatch/packages/ssrf/**` and itself.

It runs the Go and TypeScript halves against `address_vectors.json` in a **single job**, so a cross-language divergence is one unambiguous failure. Previously the two suites ran in `go-ci` and `langwatch-app-ci`, meaning a disagreement showed up as two red Xs in unrelated workflows — or as none, when only one side's paths changed and the other never re-ran. The TypeScript step is `if: !cancelled()` so a divergence names both sides in one log.

Green on its first run (`9adadeb`).

services/nlpgo/cmd/root.go
0xdeafcafe: Switched to `clog.Get(ctx)` in `bed2862`. Same underlying zap logger, but it matches how the rest of the service logs and picks up whatever `clog.With` has attached to the context by that point.

langwatch/packages/ssrf/tsconfig.json
0xdeafcafe: Done in `37d37eb`: `"incremental": true` plus a package-scoped `"tsBuildInfoFile": "node_modules/.cache/tsbuildinfo/ssrf.tsbuildinfo"`, matching the app and `packages/automations`. Verified it actually writes the file rather than being silently ignored alongside `noEmit`.

Made it a rule too — added a row to the TypeScript table in `CLAUDE.md`, since this is only ever noticed once a package is already slow.

services/aigateway/adapters/providers/customer_endpoint_ssrf.go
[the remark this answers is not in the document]
0xdeafcafe: Reverted in `bff60b6` — apologies, I applied this rename before your "no" landed. `blockLocal` / `allowlisted` are back as they were.

The named-object-parameter change on `blocked()` stays, since that came from a separate thread and is not what you rejected.

services/aigateway/adapters/providers/customer_endpoint_ssrf.go
```diff
@@ -139,13 +125,7 @@ func validateCustomerEndpoint(ctx context.Context, rawURL string, policy custome
...
-			return fmt.Errorf("customer endpoint resolves to a reserved address")
-		}
-		if policy.blockLocal && !allowlisted && !isPublicEndpointIP(parsed) {
-			return fmt.Errorf("customer endpoint resolves to a non-public address")
-		}
-		return nil
+		return endpointAddressError(parsed, policy.blockLocal, allowlisted)
 	}
 
 	addresses, err := policy.resolve(ctx, host)
```
langwatch-agent, line 131: **P1 — Customer-provider requests still redial after validation.** This lookup only decides whether to return an error; it discards the approved address. The subsequent Bifrost configuration still receives the original hostname (`bifrost.go:887`) and Bifrost creates its own `fasthttp` client/dialer, so it resolves the hostname again when the request is sent. A tenant-controlled name can answer with a public IP here and a private/metadata IP at that later dial, causing the gateway to send the provider request and its credential to the private target. Preserve the validated address and make the actual Bifrost/Azure dial use it while retaining the original host for SNI/certificate verification, or route these requests through an egress transport that resolves, classifies, and pins at the connection point. Add an integration test with a sequenced resolver and capture listener proving the second lookup cannot change the connected address.

pkg/ssrf/address.go
```diff
@@ -0,0 +1,217 @@
...
+	{netip.MustParsePrefix("192.0.0.0/24"), "IETF protocol assignments (RFC 6890)"},
+	{netip.MustParsePrefix("192.0.2.0/24"), "TEST-NET-1 documentation (RFC 5737)"},
+	{netip.MustParsePrefix("192.88.99.0/24"), "6to4 relay anycast, deprecated (RFC 7526)"},
+	{netip.MustParsePrefix("198.18.0.0/15"), "benchmarking (RFC 2544)"},
+	{netip.MustParsePrefix("198.51.100.0/24"), "TEST-NET-2 documentation (RFC 5737)"},
+	{netip.MustParsePrefix("203.0.113.0/24"), "TEST-NET-3 documentation (RFC 5737)"},
+	{netip.MustParsePrefix("240.0.0.0/4"), "reserved (RFC 1112 / RFC 919)"},
+	{netip.MustParsePrefix("64:ff9b::/96"), "well-known NAT64 (RFC 6052)"},
+	{netip.MustParsePrefix("64:ff9b:1::/48"), "local-use NAT64 (RFC 8215)"},
+	{netip.MustParsePrefix("100::/64"), "discard-only (RFC 6666)"},
```
langwatch-agent, line 115: **P1 — The shared table permits IANA non-routable IPv6 ranges.** The PR says this is the union of the IANA special-purpose registries, but it omits at least `100:0:0:1::/64` (Dummy IPv6 Prefix, destination=false) and `5f00::/16` (SRv6 SIDs, globally-reachable=false). Neither matches the predicates nor this list, so `Classify` returns `global`; the gateway, Langy pinning, NLP strict mode, and TypeScript app all permit a literal or DNS answer in those ranges. Use a complete reviewed deny list for every registry entry that is not globally reachable (or generate one from a pinned registry snapshot), and add boundary vectors for each entry so the cross-language corpus actually enforces the claimed invariant.

langwatch-agent (changes requested):
Request changes: 2 P1 security blockers inline. The shared corpus, Langy pinning, and NLP dial-time checks are good directions, but the customer-provider path still redials unchecked and the claimed IANA deny set is incomplete.

PR #5926, opened 2026-07-19
fix(traces): resolve offloaded span IO in getSpanForPromptStudio (#5753)

### LINKED ISSUE #5753: getSpanForPromptStudio bypasses blob resolution and returns truncated (>64KB) span IO
## Context

Flagged in review of #5082 ([review link](https://github.com/langwatch/langwatch/pull/5082#pullrequestreview-4683002673)) as part of the "all read paths" backlog for the >64 KB span-IO offload (#4991 / ADR-022):

> `spans.ts`'s `getSpanForPromptStudio`, which also bypasses both resolvers.

## Problem

`getSpanForPromptStudio` (`langwatch/src/server/traces/clickhouse-trace.service.ts`, via the `spans.ts` tRPC route feeding the prompt playground) reads `stored_spans` directly and extracts `SpanAttributes` without going through the blob resolver used by `getTracesWithSpans`. When a span's input/output IO was offloaded (>64 KB), the playground opens with the truncated 64 KB preview instead of the full messages/config.

Unlike the thread/eval paths, this is not a flag flip: the method does not use the resolver-wired `getTracesWithSpans` path, so blob resolution has to be wired into this read.

## Acceptance criteria

- `getSpanForPromptStudio` resolves offloaded eventref IO to the full value before returning the prompt-studio result.
- Multitenancy preserved (tenant-scoped resolution).
- Covered by a test proving an over-threshold span IO opens full in prompt studio.

feiiiiii5:
## Context

Flagged in review of #5082 ([review link](https://github.com/langwatch/langwatch/pull/5082#pullrequestreview-4683002673)) as part of the "all read paths" backlog for the >64 KB span-IO offload (#4991 / ADR-022):

> `spans.ts`'s `getSpanForPromptStudio`, which also bypasses both resolvers.

Closes #5753.

## Problem

`getSpanForPromptStudio` (`langwatch/src/server/traces/clickhouse-trace.service.ts`, via the `spans.ts` tRPC route feeding the prompt playground) reads `stored_spans` directly and extracts `SpanAttributes` without going through the blob resolver used by `getTracesWithSpans`. When a span's input/output IO was offloaded (>64 KB), the playground opens with the truncated 64 KB preview instead of the full messages/config.

Unlike the thread/eval paths this is not a flag flip — the method does not use the resolver-wired `getTracesWithSpans` path, so blob resolution has to be wired into this read.

## Fix

Two-file change plus a regression test:

1. `spans.ts` — `getForPromptStudio` now constructs `TraceService` with `buildTraceBlobResolutionDeps()`, matching the existing `getAllForTrace` route above it. The deps factory is the same composition point used everywhere else (`TraceService.create(ctx.prisma, buildTraceBlobResolutionDeps())`); no new abstraction is introduced.

2. `clickhouse-trace.service.ts` — `getSpanForPromptStudio` now resolves the **target span's** offloaded eventref IO before calling `extractPromptStudioDataFromClickHouse`. Resolution reuses the already-injected `resolveTraceSpans` callback — the same one the bulk read paths use — so no new code paths are added at the resolver layer.

   Scope choice: only the target LLM span is resolved, not every row the SQL returns. Rationale:
   - `extractPromptStudioDataFromClickHouse` reads only the target row's `SpanAttributes` (input/output messages, model/vendor/llmConfig).
   - The ancestor walk below it (`findPromptReferenceInAncestors`) reads only small `langwatch.prompt.*` metadata keys, which are well under the 64 KB offload threshold and therefore never have eventref pointers.
   - Keeping the read-path cost O(1) in `event_log` reads — matching the cost shape of `getAllForTrace`'s per-trace resolution.

   The `NormalizedSpan` constructed for the resolver call carries only `{spanId, traceId, spanAttributes}` — the three fields the resolver actually reads — cast through `as unknown as NormalizedSpan` since the resolver's full type carries projection baggage the read path here does not need.

## Acceptance criteria

- [x] **AC1 — `getSpanForPromptStudio` resolves offloaded eventref IO to the full value before returning.** New test `resolves the full output value from event_log (not the preview)` asserts `getFromEventLog` is called with `{tenantId, aggregateId: traceId, field, eventId}` and the assistant message content equals the full value (not the preview).
- [x] **AC2 — Multitenancy preserved.** The resolver is the same `resolveOffloadedTraces` used by `getAllForTrace`; it calls `getFromEventLog` with `tenantId = projectId`. The test asserts the `tenantId` argument explicitly.
- [x] **AC3 — Test coverage proving an over-threshold span IO opens full in prompt studio.** Three cases: AC1+AC2 happy path, byte-count widening (`langwatch.output` exceeds 64 KB), and a safety-net test proving the read degrades gracefully (keeps the preview, zero `event_log` reads) when no resolver is wired — so a future call site that forgets `buildTraceBlobResolutionDeps()` cannot regress silently.

## Test pattern

Follows the #4991 test pattern: real `ClickHouseTraceService` + real per-trace resolver + fake `BlobStore`. Only the raw CH SQL boundary is mocked (`vi.mock("~/server/clickhouse/clickhouseClient")` via `vi.hoisted`). The fake `BlobStore`'s `getFromEventLog` is the spy that proves AC1/AC2.

## Files

- `langwatch/src/server/api/routers/spans.ts` — 5-line wiring change in `getForPromptStudio`.
- `langwatch/src/server/traces/clickhouse-trace.service.ts` — resolution block before `extractPromptStudioDataFromClickHouse` in `getSpanForPromptStudio`.
- `langwatch/src/server/traces/__tests__/clickhouse-trace.service-5753-prompt-studio.unit.test.ts` — new file, 3 test cases.

## Out of scope

- The ancestor walk's `langwatch.prompt.*` metadata keys are not resolved; they are below the offload threshold and have no eventref pointers.
- The `getTracesWithSpans` bulk path is unchanged; it was already wired by #4991.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prompt Studio now displays complete LLM outputs, including content stored outside the preview limit.
  * Event data is resolved securely within the appropriate tenant context.
  * Existing behavior is preserved when full event data is unavailable.

## AI Disclosure

AI-assisted debugging/initial draft/testing with human review of the diff and test scope (per the contribution workflow).


feiiiiii5:
Both CodeRabbit actionable comments are addressed in `5618288ff526e4f59bccd5d32c27b947f554c4c4` (just pushed via Git Data API — unverified, same as the original `b987ce0` on this branch).

## Nitpick 1 — hardcoded `IO_PREVIEW_BYTES = 65536`

`langwatch/src/server/traces/__tests__/clickhouse-trace.service-5753-prompt-studio.unit.test.ts` now imports `IO_PREVIEW_BYTES` from `~/server/app-layer/traces/lean-for-projection` (alongside the existing `EVENTREF_ATTR_PREFIX` import that was already there) and drops the local `65536` literal. The test now validates against the real threshold — if `lean-for-projection.ts` ever changes the value, this test will catch the drift instead of silently passing against a stale copy. `LARGE_BYTE_COUNT`, `FULL_OUTPUT`, and `PREVIEW_OUTPUT` behavior is unchanged.

## Nitpick 2 — unsafe `as unknown as NormalizedSpan` cast

`langwatch/src/server/traces/clickhouse-trace.service.ts` now constructs a fully-typed `NormalizedSpan` object instead of double-casting a partial object literal. All 25 required fields are populated — the three that matter (`spanId`, `traceId`, `spanAttributes`) come from `row`, `tenantId` comes from `projectId` (already in scope), and the remaining 21 fields get safe defaults (`null` for nullable fields, `0` for numeric fields, `[]` for `events`/`links`, `{}` for `resourceAttributes`, `NormalizedSpanKind.UNSPECIFIED` for `kind`, etc.).

Extraction logic, target-only resolution (only the LLM span is resolved, ancestors stay on `allRows`), and the ancestor-walk behavior are unchanged — this is purely a type-safety fix so a future change to `resolveTraceSpans` that reads a new field gets type-checked instead of silently receiving `undefined`.

feiiiiii5:
@0xdeafcafe Gentle ping for review when you have a moment — CodeRabbit passed in [run a37f98fe](https://github.com/langwatch/langwatch/pull/5926#issuecomment-...) ("No actionable comments") on `5618288ff5` after both nitpicks (hardcoded `IO_PREVIEW_BYTES` → imported from `lean-for-projection.ts`; unsafe `as unknown as NormalizedSpan` → fully-typed `NormalizedSpan` with 25 fields populated) were addressed.

`validate title` check is SUCCESS; the other CI checks are skipped by default for fork PRs in this repo. Would you mind approving the workflow run / reviewing when convenient? Happy to make further changes if needed. Thanks!

platform/app/src/server/traces/clickhouse-trace.service.ts
```diff
@@ -1340,9 +1340,65 @@ export class ClickHouseTraceService {
...
+              statusMessage: null,
+              statusCode: null,
+              instrumentationScope: { name: "", version: null },
+              droppedAttributesCount: 0,
+              droppedEventsCount: 0,
+              droppedLinksCount: 0,
+              cost: null,
+              nonBilledCost: null,
+            };
+            const { resolvedSpans } = await this.resolveTraceSpans(
```
0xdeafcafe: mhmm, i think using some nicer types here would be preferable over a full vertical screen invasion.

something like, defined on the `this` would be nice.

```
type BlobResolvableSpan = Pick<
  NormalizedSpan,
  "spanId" | "traceId" | "spanAttributes"
>;

type ResolveTraceSpansFn = (
  projectId: string,
  spans: BlobResolvableSpan[],
) => Promise<{
  resolvedSpans: BlobResolvableSpan[];
}>;
```

but this is a nit and you did't write the original code so i'll also approve this 👌🏻

platform/app/src/server/traces/clickhouse-trace.service.ts
```diff
@@ -1340,9 +1340,65 @@ export class ClickHouseTraceService {
...
+              projectId,
+              [targetSpan],
+            );
+            const resolved = resolvedSpans[0];
```
0xdeafcafe, line 1592: can `resolvedSpans` be nullish?

0xdeafcafe (dismissed):

platform/app/src/server/traces/clickhouse-trace.service.ts
feiiiiii5: Appreciate the suggestion — agree `Pick<NormalizedSpan, "spanId" | "traceId" | "spanAttributes">` would be cleaner than the 25-field construction.

The reason I went with full `NormalizedSpan` is to keep the `ResolveTraceSpansFn` signature stable — it is exported from this file and used by both this per-trace path and the bulk read path (`ResolveTraceSpansBatchFn` in `getTracesWithSpans` / `enrichTracesWithSpans`). Tightening the param type to a `Pick` would be a non-breaking refinement (full `NormalizedSpan[]` is assignable to `Pick<...>[]` covariantly), but it touches the public signatures of both `ResolveTraceSpansFn` and `ResolveTraceSpansBatchFn`, so I figured it is better as a separate refactor PR rather than scope-creeping this one.

Will leave as-is for #5753 since you approved. Happy to open a follow-up if you would like — let me know.

platform/app/src/server/traces/clickhouse-trace.service.ts
feiiiiii5: No — `resolvedSpans` cannot be nullish per the type contract.

`ResolvedTraceSpans.resolvedSpans` is declared as `resolvedSpans: NormalizedSpan[]` (required field, not optional) in `resolve-offloaded-traces.ts:45`, and `resolveOffloadedTraces` explicitly assigns the array in all three return paths:

- **No-eventref fast path** (`resolve-offloaded-traces.ts:108-114`): `resolvedSpans: normalizedSpans` — input array passed through unchanged.
- **Post-settlement, no span actually resolved** (`:246-252`): `resolvedSpans` — the array built from `spanSettlements.map(...)` (which falls back to the original span on rejection).
- **Happy path, at least one span resolved** (`:258-263`): `resolvedSpans` — same array as the previous path.

`resolvedSpans[0]` can still be `undefined` if the input array is empty (here we always pass `[targetSpan]`, so it is length 1, but the type allows empty) — that is what the existing `if (resolved)` guard at line 1393 already handles.

A nullish `resolvedSpans` itself would be a resolver contract violation (the type system would not allow it without an `as` cast), and if a buggy resolver did return one, `resolvedSpans[0]` would throw `TypeError` which is caught by the surrounding `try/catch` in `getSpanForPromptStudio` and surfaced as the generic "Failed to fetch span for prompt studio" error — same behavior as pre-fix.

feiiiiii5:
@0xdeafcafe Thanks for the review — replied inline to both comments.

Separately — apologies for the CI churn on `5618288ff5`. The fixup commit `1f1eab8204` (just pushed) fixes the `typecheck` (TS1361) and `test-unit (3)` failures that surfaced after your approval. Root cause: I used `NormalizedSpanKind.UNSPECIFIED` as a runtime value but imported it via `import type { ... }`, which strips the binding at runtime. Split the import to fix — the runtime behavior of the `targetSpan` construction is unchanged (still the same 25-field object, same `kind: NormalizedSpanKind.UNSPECIFIED`).

Process gap on my side: this PR was pushed via Git Data API without a local `tsc --noEmit` pass (no local clone of langwatch — it has historically been faster for me to edit via /tmp + Git Data API). That is why I missed the TS1361 error and the 2 unit test failures that surfaced once CI ran. Will not repeat — for any future change that touches `import type` / runtime enum bindings (or any new code logic, not just import-only edits), I will clone locally first and run `tsc --noEmit` + the relevant `vitest` unit tests before pushing.

Since this is a fork PR, the `typecheck` / `test-unit` / `langwatch-app-complete` workflow runs need maintainer approval to re-trigger on the new commit `1f1eab8204`. Could you re-approve the workflow runs (and re-review when convenient)? The fix is purely the import split — no behavior change beyond fixing the `TypeError: Cannot read properties of undefined (reading "UNSPECIFIED")` that was breaking the 2 prompt-studio tests.

0xdeafcafe (dismissed):

0xdeafcafe:
Thanks! Don't worry about it.

feiiiiii5:
@0xdeafcafe Thanks for the gracious "Don't worry about it" — and sorry to come back with a *second* CI fixup so soon after. Honest update:

The `1f1eab8` fixup that split `NormalizedSpanKind` into a value import (to fix TS1361) had a side effect: it unblocked the compiler from reaching line 215 of the test file, exposing a previously-masked TS2345:

```ts
Buffer.byteLength(assistantMessage!.content, "utf8")
```

`Message.content` is typed `string | ContentPart[] | null | undefined`, but `Buffer.byteLength` only accepts `string | ArrayBufferLike`. The TS1361 was acting as an accidental "stop early" gate — fixing it surfaced this.

Fix in `5b96eb2` (just pushed): added `as string` cast with a comment explaining why it is safe here (the mock fixture sets `langwatch.output` to a plain string `FULL_OUTPUT`). I deliberately kept the cast narrow rather than widening `Message.content` or restructuring the assertion — the broader `string | ContentPart[] | null | undefined` shape is correct for real runtime messages and shouldn't be weakened for one test.

Local verification: ran `tsgo --noEmit --project ./tsconfig.tsgo.tests.json` against the file — no errors from this file (only the pre-existing Prisma TS2305 noise from `pnpm approve-builds` not being run on my local clone, which is unrelated infrastructure).

Process fix on my side: I now have a local clone of langwatch, so any future change that touches imports / type positions / runtime-enum bindings will get a `tsgo --noEmit` pass *before* pushing, not after. This second fixup should be the last.

Since pushing `5b96eb2` dismisses your prior approval on `1f1eab8`, would you mind re-approving the workflow runs when convenient? The diff is +5/-1 in a single test file — purely a type-level cast with a comment.

langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** Closes #5753, itself carved out of the #5082 review as part of the "all read paths" backlog for the >64KB span-IO offload (#4991 / ADR-022). Well aligned and clearly scoped.

**Top inspection targets**
1. `Must Check` - that `getSpanForPromptStudio` now uses the same resolver as `getTracesWithSpans` rather than a parallel implementation. A second resolver is how this class of bug recurs on the next read path.
2. `Ask Author` - whether any *other* read paths still bypass the resolver. The linked review called this the "all read paths" backlog, so it is worth confirming this is the last one or that the rest are tracked.
3. `Probably Fine` - the added unit test is scoped to the 5753 case.

**Production risk:** low. Read-path only; worst case is the playground continuing to show truncated IO, which is the current state.

1. Is this the final read path in the ADR-022 backlog, or are others still outstanding?

feiiiiii5:
Verified the items in @langwatch-agent's brief.

**Must Check 1 — `getSpanForPromptStudio` uses the same resolver:** Yes — routes through `this.resolveTraceSpans` (the instance-level resolver), the same one `getTracesWithSpans` uses at line 2550. The pre-#5753 code did a parallel in-place attribute lookup; the fix routes through the shared resolver. The comment at `clickhouse-trace.service.ts:1354-1388` calls out the #5753 rationale: a fully-typed `NormalizedSpan` is constructed (safe defaults for every required field) so future resolver changes (tenantId-scoped logging, filtering) don't silently receive `undefined`. Only the target LLM span is resolved — the ancestor walk below reads only small `langwatch.prompt.*` metadata that is never offloaded.

**Ask Author — other outstanding read paths:** This PR closes #5753 (the prompt-studio read path). I have not audited every other read method in this file for the same bug class — there may be other read paths that still bypass `resolveTraceSpans`. @0xdeafcafe — want me to do a separate audit pass and file follow-up issues for any remaining bypass paths, or is closing #5753 sufficient for now?

langwatch-agent (commented):

feiiiiii5:
@0xdeafcafe Thanks for the understanding! Let me know if there's anything else needed on this one.

feiiiiii5:
@0xdeafcafe Gentle re-ping for re-review when you have a moment. The earlier approval was dismissed after the two CI fixups (TS1361 `import type` issue + the downstream TS2345 it unmasked); both are resolved in `1f1eab8204` and `5618288ff5`. CodeRabbit cleared in run `a37f98fe` ("No actionable comments") and the langwatch-agent brief from 07-20 was answered inline. CI is green on the latest push. Happy to address any remaining concerns.

langwatch-agent (commented):
Static review of external contribution at `5b96eb27183e68e2f311e96efd23f047c331416b` (no PR-branch code or dependencies executed). No merge-blocking issue found. The prompt-studio lookup remains project-scoped and resolves only the selected LLM span before extraction, preserving the existing inline-trace path while adding the required blob-resolution dependency. CI evidence was considered but not rerun under the external-contribution trust boundary.

platform/app/src/server/traces/__tests__/clickhouse-trace.service-5753-prompt-studio.unit.test.ts
```diff
@@ -0,0 +1,260 @@
...
+ * resolves the target LLM span's attributes before extraction.
+ *
+ * ACs covered here:
+ *   AC1 — getSpanForPromptStudio resolves offloaded eventref IO to the
+ *         FULL value before returning the prompt-studio result.
+ *   AC2 — Multitenancy preserved (tenant-scoped resolution).
+ *   AC3 — Covered by a test proving an over-threshold span IO opens
+ *         full in prompt studio.
+ *
+ * BDD structure: given/when nested describes, action-based it() names.
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in dc9e667ae — removed the BDD-structure line from the header docstring and all three structural banner comment blocks (hoisted mock, AC list, constants/helpers), leaving only the given/when describe names and the existing behavior comments.

feiiiiii5:
Completing the audit I offered on 07-20 — I checked every span-IO read path in `ClickHouseTraceService` against the #5753 bug class (paths that return span IO without going through `resolveTraceSpans`):

- `getTracesWithSpans` → `resolveAndMergeMany` → `resolveSpansBatch` ✅
- `getTracesByThreadId` / `getTracesWithSpansByThreadIds` → delegate to `getTracesWithSpans` ✅
- `getAllTracesForProject` → `enrichTracesWithSpans` → `resolveAndMergeMany` (opts in via `resolveBlobs`) ✅
- `getSpanForPromptStudio` → `resolveTraceSpans` (this PR) ✅
- `getTopicCounts` / `getCustomersAndLabels` / `getDistinctFieldNames` / `resolveTraceIdByPrefix` → no span IO ✅

The only reads that intentionally skip resolution are the list/search/aggregation paths that keep the ≤64 KB preview and issue zero `event_log` SELECTs — that is the documented ADR-022 AC5 opt-in behavior, not a bypass. So #5753 was the last outstanding read path in this service; no follow-up issues needed from this audit.

langwatch-agent (commented):
Static review of this external contribution: inspected the current diff with emphasis on authorization, tenant boundaries, error handling, and regression risk. No blocking correctness or security issue was identified. No external code was checked out or run.

rogeriochaves (dismissed):

langwatch-agent:
Pushed `57a765e` to this branch to clear the `lint` job.

The blocking Biome step reported exactly 2 errors, both on lines this branch added:

- `clickhouse-trace.service-5753-prompt-studio.unit.test.ts` - `assist/source/organizeImports`, the two `blob-store.service` imports sort before the `lean-for-projection` block.
- `clickhouse-trace.service.ts` - `format`, the `resolveTraceSpans(projectId, [targetSpan])` call was wrapped differently from what the formatter prints.

Both are Biome's own safe fixes, applied with the repo's pinned 2.5.1 and scoped to those two files so nothing unrelated gets reformatted. Net diff is 5 insertions, 6 deletions, all inside lines this branch introduced.

Verified locally:

- whole-tree `pnpm lint` (`biome check ./src ./ee ./scripts ./e2e ./prisma ./vite`) exits **0**, down from 2 errors.
- warning count is **unchanged at 3593**, so the `biome-new-violations.sh` gate has nothing new to count.
- the branch's 3 unit tests still pass.

Two other red checks are not lint and not yours: `changes` failed on `self-signed certificate` inside `detect-changes`, which is a runner TLS blip, and `sdk-go-complete` / `langwatch-app-complete` are `alls-green` aggregates that just inherit `changes` and `lint`. This push re-triggers all of them.

---
<sub>🤖 **Pushed by `tech-debt-fixer` on behalf of the fleet.** Authored under the owner's GitHub token for API access, so the commit shows the bot identity; the content is the fleet's, not his.</sub>

feiiiiii5:
@rogeriochaves — the branch now includes the bot's lint fix (`57a765e`, Biome 2.5.1 safe fixes scoped to this PR's files) and all CI checks are green. Could you take another look when convenient? Happy to address anything that comes up.

langwatch-agent (commented):
No findings from static review of the external contribution. I checked the tenant-scoped blob resolution and protections handoff; I did not execute contributor code or install dependencies.

LangWatch-Review: verdict=clean sha=57a765e969b0c5f3249b54f0dd04c777652c7359 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=57a765e969b0c5f3249b54f0dd04c777652c7359 p0=0 p1=0 p2=0 p3=0

langwatch-agent:
> [!CAUTION]
> **Every failure mode this change can have is silent, and each one presents as the exact bug it is fixing.**
>
> The bug: prompt studio opened an LLM span and showed the 64 KB preview instead of the real output, because this read path never went through the blob resolver. The fix wires the resolver in. But the wiring is optional at every step, and each escape hatch falls back to the preview rather than failing.
>
> `if (this.resolveTraceSpans)` means a call site that forgets to pass `buildTraceBlobResolutionDeps()` gets the old behaviour with no error. `if (resolved)` means a resolver returning an empty array falls back to the unresolved row with no error. A blob that is missing degrades to the preview with no error. The PR tests the first of these deliberately, calling it graceful degradation, and it is worth being precise that what degrades gracefully here is the truncation: the user sees a prompt that ends mid-sentence and has no way to tell whether that is the model's output or the platform's.
>
> The second thing worth your attention is the synthesized span. To call the resolver, the code builds a complete `NormalizedSpan` out of one ClickHouse row, filling roughly fifteen fields it does not have with `0`, `""`, `null`, `false` and `UNSPECIFIED`. The comment argues this is safer than leaving them undefined. It is the reverse. Undefined fails loudly the first time a resolver reads it; `startTimeUnixMs: 0` is January 1970 and a filter or a log line will accept it without complaint.

| | |
| --- | --- |
| **Mode** | `Targeted Review`. Three attention areas, all in one method, listed below. The wiring itself is five lines and correct. |
| **Issue** | `Closes #5753`, which is itself a spun-off item from the `#5082` review of the >64 KB span IO offload (`#4991` / ADR-022), the "all read paths" backlog. The PR does what the issue asks. |
| **State** | +316 / -3 across 3 files, of which 247 lines are one new test file. CI green, mergeable. Open since 19 July. Note the reviewer: `rogeriochaves` is requested, his earlier review was dismissed by a push, and `0xdeafcafe` has two dismissed reviews on it too. Both have read it. |
| **Evidence** | Good. Three test cases against the real service and real resolver with a fake blob store, and the tenant argument is asserted explicitly rather than assumed. The scope claim about the ancestor walk is argued in the PR body but not covered by a test. |
| **Where to look** | The synthesized `NormalizedSpan` literal · the three optional branches and what each falls back to · the claim that only the target span needs resolving |

<details>
<summary><b>What you are looking at, if the offload path is not your area</b></summary>

When ACME's agent produces a very large model output, storing it inline in ClickHouse would bloat the row, so anything over 64 KB is written to a blob store instead. What stays on the span is two things: a truncated preview of the first 64 KB, and a pointer, kept in an attribute prefixed with `langwatch.eventref.`.

Anything that reads spans has to decide whether it wants the preview or the real value. The list-and-search paths want the preview, because they are rendering hundreds of rows. The paths that open one thing for a human want the real value, and they get it by going through a resolver that follows the pointer and swaps the full value back in.

| Term | What it means here |
| --- | --- |
| **Eventref** | The pointer attribute left behind when a value was offloaded. |
| **Preview** | The first 64 KB, stored inline, with an ellipsis. What you see when nobody resolved. |
| **Resolver** | `resolveOffloadedTraces`, which takes spans and returns them with offloaded values swapped in, scoped to the tenant. |
| **`buildTraceBlobResolutionDeps()`** | The factory each call site passes to `TraceService.create` to get a resolver wired in. |
| **Prompt studio** | The playground you land in from a trace, to re-run an LLM call with its real messages and config. |

```mermaid
flowchart TD
  A["Prompt studio opens a span"] --> B["spans.getForPromptStudio"]
  B --> C{"deps wired?"}
  C -->|"no, and this is the change"| D["preview, silently truncated"]
  C -->|"yes"| E["getSpanForPromptStudio queries stored_spans"]
  E --> F{"this.resolveTraceSpans set?"}
  F -->|"no"| D
  F -->|"yes"| G["synthesize a NormalizedSpan from the row"]
  G --> H["resolveTraceSpans, tenant-scoped"]
  H --> I{"resolvedSpans[0] present?"}
  I -->|"no"| D
  I -->|"yes"| J["full output, correct"]
```

The small thing that is load-bearing: **the resolver is an optional constructor dependency rather than a required one.** That single design choice is why this bug existed, why it could exist on other paths, and why the fix cannot prevent the next occurrence. Every branch above that leads to the truncated box is a consequence of it.

</details>

<details>
<summary><b>The decisions being ratified</b></summary>

**Resolve only the target span, not every row the query returns.** The SQL for this method returns the requested span plus its ancestors, and only the target's attributes reach `extractPromptStudioDataFromClickHouse`. The ancestors are walked purely for `langwatch.prompt.*` metadata, which the PR argues is always well under the offload threshold and never carries an eventref. That is almost certainly true today, and it is a scope claim rather than a guarantee: nothing in the code asserts it, and if a prompt metadata key ever grew past 64 KB the walk would silently pick up a preview. Cheap to make explicit, and the alternative (resolving every ancestor on every prompt studio open) is a real cost for no current benefit, so the decision itself is right.

**Reuse the injected `resolveTraceSpans` callback rather than adding a resolution layer.** Correct, and the reason the diff is as small as it is. No new abstraction, same composition point as `getAllForTrace` directly above it.

**Change `NormalizedSpanKind` from a type import to a value import.** Necessary and a genuine latent trap: it is a runtime enum, so `import type` strips the binding and `.UNSPECIFIED` would have thrown. Worth noticing that this was only discovered because the new code needed a value for a field it does not actually know.

**Keep working when no resolver is wired.** The PR frames the third test as a guard so a future call site that forgets the deps "cannot regress silently". The test proves it does exactly zero event log reads and returns the preview, which is the silent regression, now with a test asserting it. If the intent is that forgetting is detectable, the assertion needs to be somewhere a human sees it: a log line at minimum, and arguably the constructor should require the dependency rather than accept its absence.

</details>

<details>
<summary><b>Worth asking the author</b></summary>

**Why do the synthesized fields get plausible values instead of being absent?** The stated reason is that a future resolver change would otherwise "silently receive undefined", but undefined is the loud option here. `startTimeUnixMs: 0` and `endTimeUnixMs: 0` will pass a range filter, print as 1970 in a log, and produce a duration of zero without anything noticing. If the resolver only needs three fields, the honest shape is a narrower parameter type, or a comment saying these are unknown rather than defaults.

**Should the resolver be optional at all?** This is the second read path found bypassing it, after `#5082` found the first. The pattern will keep producing this bug for as long as the dependency can be omitted. Making it required, or making the omission log, converts a class of silent truncations into an obvious failure. That is likely out of scope for this PR and worth a follow-up either way.

**Are there remaining paths on the `#4991` backlog?** The PR body says the bulk path was already wired by `#4991` and this one is the prompt studio case. Whether the backlog is now empty is the useful thing to know before this closes.

**The `#5753:` history comments.** Five of them, in production code, explaining what changed and why relative to the previous state. The final state of the file does not need to know it was once wrong, and the reasoning belongs in the commit and this PR. Worth a request before merge, since they are the kind of comment that outlives its context and starts misleading.

</details>

<details>
<summary><b>Evidence and files</b></summary>

| File | What it is |
| --- | --- |
| `platform/app/src/server/api/routers/spans.ts` | Five lines. `getForPromptStudio` now passes `buildTraceBlobResolutionDeps()`, matching `getAllForTrace` directly above it. This is the whole wiring fix. |
| `platform/app/src/server/traces/clickhouse-trace.service.ts` | The import correction, plus a 55 line block in `getSpanForPromptStudio` that builds the synthesized span, calls the resolver and swaps the attributes back onto the row. |
| `platform/app/src/server/traces/__tests__/clickhouse-trace.service-5753-prompt-studio.unit.test.ts` | 247 lines, three cases: resolution returns the full 400 KB value and not the preview, the tenant argument is `projectId`, and the no-resolver path does zero event log reads. |

The test pattern matches `#4991`: real service, real per-trace resolver, fake blob store, only the raw ClickHouse client mocked. That is the right boundary and the reason these tests are worth something.

</details>

> [!NOTE]
> **The invariant this PR is defending: a read path that opens one thing for a human must never hand back a preview.** The wiring satisfies it for this call site. Whether the design can hold it for the next call site is the open question, and it is the one worth a comment on the PR.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **Before reviewing the code, resolve the duplicate.** This PR and #7372 both say they close #5753, and both wire blob resolution into the prompt-studio span read. Two people fixed the same bug independently. Whoever reviews first should pick one and close the other, otherwise the second one to land is a merge conflict on a file nobody wanted to touch twice.

| | |
|---|---|
| **Issue** | Closes #5753, and so does #7372 |
| **Size** | +316/-3 across 3 files |
| **Mode** | Targeted, after the duplicate is settled |
| **Blast radius** | The prompt playground's span read path |
| **Author** | External contributor |

<details>
<summary><b>What you are looking at, if the traces read path is not your directory</b></summary>

Span IO gets big. A single LLM completion at a common `max_tokens` setting runs to tens of kilobytes, so LangWatch does not store the whole thing in the span row. At ingestion, `leanForProjection` caps the four IO attributes (`langwatch.input`, `langwatch.output`, `gen_ai.input.messages`, `gen_ai.output.messages`) at a preview budget of `IO_PREVIEW_BYTES`, 64 KB, and attaches an **eventref pointer** alongside the truncated value. The full payload lives in `event_log`, which ADR-022 makes the single source of truth. Non-IO attributes get a much smaller 2 KB cap.

The consequence, and it is the whole shape of this class of bug: **a read that skips blob resolution does not fail. It returns the preview.** The row is there, the attribute is there, the JSON parses, and the value is simply the first 64 KB with the rest silently absent. Nothing in the type system or the query result distinguishes a resolved attribute from a truncated one.

So blob resolution is a dependency a read path either has or does not, wired in at construction. `TraceService.create(...)` takes `BlobResolutionDeps` and builds the resolvers from them; a call site that constructs the service without those deps gets a working service that quietly serves previews. That is why "which read paths resolve" became a backlog to be worked through path by path rather than a single fix, and why two sibling calls in the same router can differ.

One shape detail that explains the extraction in this diff: the existing trace-level resolver takes `NormalizedSpan[]`, a rich type with roughly twenty fields, while some read paths hold raw ClickHouse rows carrying only the attribute map. Resolution itself only ever needed the attributes.

The other policy worth knowing: resolution failure **keeps the preview** rather than failing the read. A span whose blob has aged out under retention still opens, on the truncated value. That is deliberate and already the rule on the trace paths.

</details>

<details>
<summary><b>The bug, which is real either way</b></summary>

`getSpanForPromptStudio` in `clickhouse-trace.service.ts` reads `stored_spans` directly and extracts `SpanAttributes` without going through the blob resolver that `getTracesWithSpans` uses. When a span's input or output IO was offloaded because it exceeded 64 KB (#4991, ADR-022), the playground opens showing the truncated 64 KB preview rather than the full messages and config.

The user-visible effect is that the playground silently lies about large prompts, which is exactly the case where someone most wants to open it.

This was flagged in review of #5082 as part of the "all read paths" backlog for the offload, so it is a known gap rather than a discovery.
</details>

<details>
<summary><b>Why this one is not a flag flip</b></summary>

Worth understanding before comparing the two candidates. The thread and eval paths could be fixed by routing them through the already-resolver-wired `getTracesWithSpans`. This method does not use that path, so blob resolution has to be wired into the read itself. That means the two competing PRs may have made genuinely different structural choices, and picking between them is a design decision rather than a coin flip.

The question to ask of whichever you keep: does it wire the resolver into this one method, or does it make it structurally hard for the next read path to forget? The backlog exists because "remember to use the resolver" has already failed more than once.
</details>

<details>
<summary><b>What to check in the code</b></summary>

1. **N+1 on resolution.** A read that resolves blobs per span will do one blob fetch per offloaded span. The playground opens on one trace, so the blast radius is bounded, but confirm the resolution is batched rather than looped.
2. **Failure mode when the blob is gone.** Offloaded IO lives somewhere with its own retention. If the blob is missing, the right answer is probably still showing the preview with an indication rather than failing the whole read.
3. **The regression test.** `clickhouse-trace.service-5753-prompt-studio.unit.test.ts` should assert the resolved full value, not merely that the resolver was called.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. The duplicate with #7372 is the first decision; the code review is cheap once that is settled.

drewdrewthis:
## Duplicate: a second fix for #5753 is open

@feiiiiii5 — apologies for the long silence on this. Flagging it rather than leaving it quiet for another week.

A second PR fixing #5753 is open: https://github.com/langwatch/langwatch/pull/7372. It edits two of the same files (`spans.ts`, `clickhouse-trace.service.ts`). Neither PR referenced the other until now. A human reviewer has to pick one; this is not closed and has not been dismissed.

**How they differ:** this PR (+316/-3) builds a `NormalizedSpan` around the ClickHouse row so the existing resolver signature accepts it. #7372 (+1110/-96) extracts `resolveOffloadedSpanAttributes` out of `resolveOffloadedTraces` so the read passes only the attributes it holds.

**What this PR has that #7372 does not:** the legacy-caller guard in `clickhouse-trace.service-5753-prompt-studio.unit.test.ts` —

> `it("keeps the preview value (graceful degradation, zero event_log reads)")`

If #7372 is the one that merges, that test should be carried across, because #7372 changes a shared code path where exactly that regression could pass unnoticed.

This PR is `mergeable`, `ci-green`, `hound-checked`, with no failing checks — blocked only on review.

feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

feiiiiii5:
Reopening — the close on 08-28 was a mistake on my side. I ran a batch sweep to shrink my open-PR footprint and it caught this thread even though it still had review activity on it.

The branch is unchanged. Still open from your side: whether `resolvedSpans` can be nullish, and the preference for nicer types over the full-height UI treatment. I'd rather answer those properly than assume they were settled.

Sorry for the noise, and happy to pick this back up whenever someone has a slot.

PR #5917, opened 2026-07-18
feat(errors): handled-error remediation channel (tips/docsUrl/fault) + shared package

0xdeafcafe:
## Summary

Handled errors (`HandledError` ⇔ Go `herr`) gain a **remediation channel** — `tips`, `docsUrl`, and `fault` — so agents hitting the CLI, API, and MCP server can self-diagnose without a human interpreting the error. The TS `HandledError` core now lives in a shared package (`@langwatch/handled-error`) consumed directly by the app, MCP server, CLI and SDKs. See the ADR-045 amendment for the design.

**All wire changes are additive** — the REST envelope keeps its exact current shape plus optional new keys, so the Python SDK needs no changes.

## What changed

- **Shared package** `@langwatch/handled-error` (`packages/handled-error`, source-only, `cli-cards` pattern): full `HandledError` core + `tips`/`docsUrl`/`fault` fields. All consumers import it directly (no shim); the app wires its Grafana trace-link builder via `src/server/handled-error-wiring.ts` (loaded by `server.mts` + `workers.ts`). Zod-free via structural `ZodLikeError`.
- **Central remediation registry** (`src/server/app-layer/error-remediation.ts`): all tips + docs links keyed by error code; classes spread `remediation(code)`. A CI test asserts every `docsPath` maps to a real `docs/*.mdx`.
- **Go herr**: `ErrorBody` gains `tips`/`docs_url`/`fault`, promoted from reserved `Meta` keys (same mechanism as `message`); `fault` validated against the three-value contract; lossless `Body`/`FromBody` round-trips.
- **Fault-axis logging on every boundary**: handled-ness decides what the *client* sees; `fault` decides the log level — customer → warn (spike-watch via `handledErrorCode`), platform/provider → error; PostHog capture is unhandled-only. Applied consistently: tRPC, Hono REST (`logHttpRequest` is fault-aware), SSE (`logSseError`), and Go telemetry (`Meta["fault"]`).
- **Boundaries**: tRPC (via `serialize()`), REST error handler (additive keys), `packages/api` formatter, persisted `domainError` zod schema, nlpgo envelope (plain `HandledError` via `handledErrorFromHerr` — wire trace/span ids now preserved), **MCP server** (parses error JSON, renders `Tips:`/`Docs:` in tool error text), **SSE** (error frames carry `domainError`; unhandled stream failures no longer leak raw `err.message`).
[...]

PR #5934, opened 2026-07-19
fix(api): trust real HandledErrors, and log the ones we swallow

0xdeafcafe:
## Why

`@langwatch/api` duck-typed `HandledError` so the package could stay standalone. #5917 made that justification obsolete by extracting `@langwatch/handled-error` — but it only *extended* the duck-typed interface with `fault`/`tips`/`docsUrl` rather than replacing it, so the package still hand-rolled a second encoding of the error contract and still never logged anything.

Three real problems fell out of that, not just untidiness.

**Validation errors were logged as 500s.** A bare `ZodError` carries neither `httpStatus` nor `fault`, so `getStatusCodeFromError` derived 500 and `getLogLevelForRequest` logged at `error` — while the caller received a 422. Every validation failure landed in the 5xx error budget.

**Validation errors silently missed the remediation channel.** The `HandledError` branch of `formatError` emits `fault`/`tips`/`docsUrl`; the `ZodError` branch built a bare payload and emitted none of them. Exactly the agent-facing self-diagnosis #5917 added, absent from the most common error an agent hits.

**`createErrorHandler` never logged.** Unknown errors flatten to `"An unknown error occurred"` and the original — stack included — was dropped right there, surviving only incidentally via `loggerMiddleware` reading `c.error`.

## What changed

- **Uses the real class.** `HandledError.isHandled` replaces `isHandledErrorLike`; `zodErrorToPayload` is gone. A `ZodError` is promoted to a `ValidationError` and travels the same path as any other handled error, which fixes the status *and* the level *and* the missing remediation fields in one move.
- **The handler logs.** Handled errors by fault attribution (`customer` → warn, `platform`/`provider` → error), consistent with the axis #5917 established everywhere else. Anything unhandled logs at `error` with its cause, since that is now the only place the stack survives. Request bodies are never logged — automation `actionParams` carry encrypted webhook headers and Slack tokens.
- **The resolved status is recorded on the context**, so `loggerMiddleware` reports what the caller actually received instead of re-deriving it from the error.
- Added `@langwatch/handled-error` as a dependency; README updated.

## Behaviour change worth reviewing

Dropping the duck-type means an object that merely grows `code` + `httpStatus` + `serialize()` can **no longer choose its own status** — it is treated as unknown and answered with a 500. One existing builder test asserted the old behaviour; it now asserts the new one, with an explicit impostor test alongside.

Nothing in `packages/api` or `langwatch/src` threw HandledError-shaped-but-not-`HandledError` objects, so this should be inert — but it is the part to look at twice.

The per-issue `schema_failure` reasons shape documented in the README is preserved. `serializeReason` adds `kind` and `fault` on top, which is additive.

## Testing

`90/90` unit tests pass; package typecheck clean. New coverage for fault-based log levels, the ZodError status/level regression, unhandled-error cause logging, non-leakage of that cause into the response, and the impostor case.

Note this package still has **zero production consumers** — it was not in `langwatch/package.json`'s dependencies and nothing imports it. So this is verified by tests rather than by traffic.

https://claude.ai/code/session_0184xLhaHhYej8sa75AXuheB

## Summary by CodeRabbit

- **Enhancements**
  - Improved error responses for genuine handled errors with structured codes, metadata, remediation guidance (`fault`, `tips`, `docsUrl`), and trace identifiers.
  - Validation failures are promoted to `validation_error` with clear field-level `reasons` and a 422 response.
  - Versioned and unversioned error response shapes are handled consistently.
  - Request logging now emits exactly one error entry per failed request with correct severity and status details (including trace identifiers when available).

- **Bug Fixes**
  - Error status selection and payload formatting now trust only real handled-error instances, preventing lookalike objects from influencing responses.

- **Documentation**
  - Updated the API README with the documented error formatting and request error-logging behavior.

### DIFF
--- a/langwatch/packages/api/README.md
+++ b/langwatch/packages/api/README.md
@@ -78,7 +78,7 @@ export { GET, POST, PUT, PATCH, DELETE } from "./things.service";
 - **Tracing + logging** via `@langwatch/observability` (automatic, disable with `tracer: false` / `logger: false`)
 - **Auth, org, resource limits** applied in the right order per-endpoint
 - **Input/output/params/query validation** from Zod schemas, auto-wired to OpenAPI
-- **Error formatting** that duck-types `HandledError` (code, meta, reasons, traceId/spanId) and catches Zod errors
+- **Error formatting + logging** for `HandledError` (code, meta, reasons, traceId/spanId, fault/tips/docsUrl) and Zod errors
 - **Versioned routing** at `/api/{name}/{date}/...` with forward-copying from previous versions
 
 ## Handler signature
@@ -188,10 +188,21 @@ v.sse(
 
 Throw `HandledError` subclasses (from `@langwatch/handled-error`). The framework:
 
-1. Catches and serializes them with `code`, `meta`, `reasons`, `traceId`/`spanId`
-2. Catches `ZodError` and maps each issue to a `schema_failure` reason (cher-style)
+1. Catches and serializes them with `code`, `meta`, `reasons`, `traceId`/`spanId`,
+   plus the remediation channel (`fault`, `tips`, `docsUrl`)
+2. Catches `ZodError` and promotes it to a `ValidationError`, mapping each issue
+   to a `schema_failure` reason (cher-style)
 3. Returns union format for unversioned requests (includes legacy `error` field)
 4. Returns clean format for versioned requests
+5. Logs the failure — handled errors by `fault` (`customer` → warn, `platform` /
+   `provider` → error), anything unhandled at `error` with its cause, since the
+   response deliberately flattens it to "An unknown error occurred"
+
+Only real `HandledError` instances are trusted. An object that merely grows a
+`code` + `httpStatus` + `serialize()` is treated as unknown and answered with a
+500 — it cannot talk its way into choosing its own status.
+
+Request bodies are never logged.
 
 Validation error example:
 
--- a/langwatch/packages/api/README.md
+++ b/langwatch/packages/api/README.md
@@ -191,16 +191,18 @@ Throw `HandledError` subclasses (from `@langwatch/handled-error`). The framework
 1. Catches and serializes them with `code`, `meta`, `reasons`, `traceId`/`spanId`,
    plus the remediation channel (`fault`, `tips`, `docsUrl`)
 2. Catches `ZodError` and promotes it to a `ValidationError`, mapping each issue
-   to a `schema_failure` reason (cher-style)
+   to a `schema_failure` reason
 3. Returns union format for unversioned requests (includes legacy `error` field)
 4. Returns clean format for versioned requests
 5. Publishes the error it sent, and the status it sent it as, for the request
    logger to consume
 
-The request logger writes **exactly one** error record per failed request —
-handled errors by `fault` (`customer` → warn, `platform` / `provider` → error),
-anything unhandled at `error` with its cause, since the response deliberately
-flattens it to "An unknown error occurred". The error handler deliberately logs
+The request logger writes **exactly one** error record per failed request.
+Level comes from `fault` when the error is handled (`customer` → warn,
+`platform` / `provider` → error) and from the status code otherwise (5xx →
+error, 4xx → warn) — so an unknown error, which is flattened to a 500, logs at
+`error` with its cause, while an unhandled `HTTPException` carrying a 4xx logs
+at `warn`. The error handler deliberately logs
 nothing itself: a second record there would double every error-log-derived
 alert and count. It publishes the *promoted* error, so a `ZodError` is reported
 as the 422 `ValidationError` the caller actually received rather than the 500 a
--- a/langwatch/packages/api/README.md
+++ b/langwatch/packages/api/README.md
@@ -194,9 +194,17 @@ Throw `HandledError` subclasses (from `@langwatch/handled-error`). The framework
    to a `schema_failure` reason (cher-style)
 3. Returns union format for unversioned requests (includes legacy `error` field)
 4. Returns clean format for versioned requests
-5. Logs the failure — handled errors by `fault` (`customer` → warn, `platform` /
-   `provider` → error), anything unhandled at `error` with its cause, since the
-   response deliberately flattens it to "An unknown error occurred"
+5. Publishes the error it sent, and the status it sent it as, for the request
+   logger to consume
+
+The request logger writes **exactly one** error record per failed request —
+handled errors by `fault` (`customer` → warn, `platform` / `provider` → error),
+anything unhandled at `error` with its cause, since the response deliberately
+flattens it to "An unknown error occurred". The error handler deliberately logs
+nothing itself: a second record there would double every error-log-derived
+alert and count. It publishes the *promoted* error, so a `ZodError` is reported
+as the 422 `ValidationError` the caller actually received rather than the 500 a
+re-derivation would guess.
 
 Only real `HandledError` instances are trusted. An object that merely grows a
 `code` + `httpStatus` + `serialize()` is treated as unknown and answered with a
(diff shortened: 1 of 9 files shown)


langwatch/pnpm-lock.yaml
```diff
@@ -423,7 +423,7 @@ importers:
         version: 17.5.0
       openai:
         specifier: ^6.45.0
-        version: 6.45.0(@aws-sdk/credential-provider-node@3.972.64)(@smithy/signature-v4@5.6.2)(ws@8.21.0(bufferutil@4.1.0))(zod@3.25.76)
+        version: 6.45.0(@aws-sdk/credential-provider-node@3.972.64)(ws@8.21.0(bufferutil@4.1.0))(zod@3.25.76)
```
langwatch-agent: **P1 — This lockfile is broken when merged with the current base.** This importer drops the `@smithy/signature-v4` peer suffix from the `openai` package key, but the rest of the lockfile only contains the key with that suffix after the base change. CI's merge commit fails before typecheck/tests with `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY` for exactly this `openai@6.45.0(...)` entry. Rebase onto current `main` and regenerate the lockfile, then verify with `pnpm install --frozen-lockfile`; this cannot merge while every app CI job fails at install.

langwatch/packages/api/src/errors.ts
```diff
@@ -200,24 +170,97 @@ function formatError({
...
   return (err: Error, c: Context) => {
     const isVersioned = c.get("isVersionedRequest") === true;
-    const { status, body } = formatError({ err, isVersioned });
+    // Promote first so the response and the log agree on one error. Logging
+    // the raw ZodError would report it as unhandled, at `error`, against the
+    // 500 it no longer is.
+    const effective = err instanceof ZodError ? validationErrorFromZod(err) : err;
+    const { status, body } = formatError({ err: effective, isVersioned });
+
+    logError({ logger, err: effective, status, c });
```
langwatch-agent: **P2 — Every failed API request is now logged twice.** `logError` writes here, then the already-installed `loggerMiddleware` writes the same exception through `logHttpRequest` after `next()` settles (`middleware.ts:177`). I reproduced this with the package test suite: each thrown endpoint error produced both `langwatch:api:<name>:errors` / `unhandled error on request` and `langwatch:api:<name>` / `error handling request`, each carrying the stack. This doubles error-log volume and any log-derived alert/error count. Keep a single owner for the error record: either let `loggerMiddleware` log it after consuming a resolved/effective error context value, or suppress its error record when the handler has logged it. Add an integration assertion that one failed request produces one error record.

langwatch-agent (dismissed):
Request changes: 1 P1 merge/CI blocker and 1 P2 production-observability regression. Details are inline.

langwatch/pnpm-lock.yaml
0xdeafcafe: Fixed. Rebased onto current `main` and regenerated the lockfile from main's copy, so the `openai` importer keeps its `@smithy/signature-v4` suffix and the diff is now a clean 3-line addition (the `packages/api` -> `@langwatch/handled-error` link) rather than a stale rewrite. `pnpm install --frozen-lockfile` resolves with no `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY`.

langwatch/packages/api/src/errors.ts
0xdeafcafe: Fixed, by giving the record a single owner: `loggerMiddleware`.

`logHttpRequest` was already the richer of the two — it carries method/url/status/duration/userAgent *and* derives `handledErrorCode`/`handledErrorFault` and the level from the error — so the error handler is the copy that goes. It now logs nothing and instead publishes what it resolved (`RESOLVED_ERROR`: the error it sent, the status it sent it as, and `traceId`) for the middleware to consume.

That also fixes a second thing the old split had: the middleware saw the *raw* throw, so a `ZodError` was logged as a derived 500 while the caller received a 422 `ValidationError`. It now logs the promoted error against the status actually sent.

`createErrorHandler`'s `logger`/`name` options are dropped — it no longer logs, and keeping them would be misleading config. All callers were internal.

Added `src/__tests__/errorLogging.integration.test.ts`, which drives real requests through `loggerMiddleware` + `onError` and counts records: exactly one per failed request (unhandled 500 at error, customer-fault 404 at warn, ZodError as 422/warn), and none on success. I mutation-checked it — reinstating a second log in the handler turns it red with 3 failures. Full package suite 92/92, `tsc --noEmit` clean.

0xdeafcafe:
## Review round

I adversarially checked the main risk of the change I made — that making `loggerMiddleware` the sole error-record owner could produce **zero** records on some path. It does not, and I traced Hono's dispatch rather than assuming: `compose.js` wraps every `dispatch(i)` frame, so `onError` runs at the frame where the throw originated, sets `RESOLVED_ERROR`, and `loggerMiddleware`'s `await next()` then resolves into a `finally` that reads a populated value.

Paths checked and cleared:
- **Non-`Error` throws** — `compose` rethrows, unwinding through the middleware's `catch`; `resolved` is undefined and the `error || c.error` fallback still writes one record.
- **SSE stream errors** — `onError` never runs; `runAfterSSECompletion` calls `onStreamError` before `onSettled`, so the closed-over error is set in time for the fallback.
- **A custom `config.onError`** — never sets the key; falls back cleanly.

The fallback is what makes this safe: every path that loses the resolved pair degrades to the *old* behaviour rather than to silence.

**One pre-existing hole worth a follow-up (not introduced here):** `builder.ts` registers `tracerMiddleware` *before* `loggerMiddleware`. If the tracer throws after `await next()` returns — realistically `injectTraceHeaders(c)` in its `finally` hitting immutable headers on an already-streamed response — the logger has already written a **success** record, `onError` then emits a 500, and no error record exists. Happy to fix in a separate PR.

langwatch/packages/api/README.md
```diff
@@ -188,10 +188,29 @@ v.sse(
 
 Throw `HandledError` subclasses (from `@langwatch/handled-error`). The framework:
 
-1. Catches and serializes them with `code`, `meta`, `reasons`, `traceId`/`spanId`
-2. Catches `ZodError` and maps each issue to a `schema_failure` reason (cher-style)
+1. Catches and serializes them with `code`, `meta`, `reasons`, `traceId`/`spanId`,
+   plus the remediation channel (`fault`, `tips`, `docsUrl`)
+2. Catches `ZodError` and promotes it to a `ValidationError`, mapping each issue
+   to a `schema_failure` reason (cher-style)
```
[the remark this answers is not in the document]
0xdeafcafe: Fixed — dropped the parenthetical. It named nothing a reader could act on.

langwatch/packages/api/README.md
```diff
@@ -188,10 +188,29 @@ v.sse(
...
 3. Returns union format for unversioned requests (includes legacy `error` field)
 4. Returns clean format for versioned requests
+5. Publishes the error it sent, and the status it sent it as, for the request
+   logger to consume
+
+The request logger writes **exactly one** error record per failed request —
+handled errors by `fault` (`customer` → warn, `platform` / `provider` → error),
+anything unhandled at `error` with its cause, since the response deliberately
+flattens it to "An unknown error occurred". The error handler deliberately logs
```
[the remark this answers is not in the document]
0xdeafcafe: Good catch, and it's my wording that was wrong. Verified against `getLogLevelForRequest`: it uses `fault` only when the error is handled, and otherwise falls back to `getLogLevelFromStatusCode(statusCode)` — so an unhandled `HTTPException` carrying a 4xx logs at `warn`, not `error`.

Rewritten to state both halves rather than take the suggestion verbatim, since the 500 case still needs its "carries the cause" rationale:

> Level comes from `fault` when the error is handled (`customer` → warn, `platform` / `provider` → error) and from the status code otherwise (5xx → error, 4xx → warn) — so an unknown error, which is flattened to a 500, logs at `error` with its cause, while an unhandled `HTTPException` carrying a 4xx logs at `warn`.

PR #5940, opened 2026-07-19
feat(gateway): route anthropic providers with a base URL to self-hosted endpoints

### LINKED ISSUE #5938: Anthropic provider ignores its base URL — /v1/messages cannot reach self-hosted Anthropic-compatible servers (vLLM >= 0.24)
## Problem

An `anthropic` ModelProvider can be configured with a custom base URL (`ANTHROPIC_BASE_URL` — the registry declares it, the drawer renders the field, and validation accepts it), but gateway traffic ignores it entirely: `/v1/messages` requests are always raw-forwarded to `api.anthropic.com`.

Two independent gaps cause this:

1. **Control plane:** the bundle materialiser hard-codes the per-slot `base_url` emission to `custom` / `openai` providers (`config.materialiser.ts`, `supportsBaseURLOverride`). A saved `ANTHROPIC_BASE_URL` never reaches the gateway bundle. The legacy litellm path (evaluators, playground) already consumes it via `api_base` — only the gateway bundle drops it.
2. **Gateway (Go):** Bifrost's Anthropic key has no per-key URL slot, and provider config is resolved by provider key alone, so even if the bundle carried the URL there is nothing to consume it.

## Use case

vLLM >= 0.24 (and Claude-compatible proxies) serve the Anthropic Messages API natively at `/v1/messages`. Teams running Claude Code or Anthropic SDK clients against local/self-hosted models want to keep LangWatch virtual keys, budgets, and tracing in the path:

```
Claude Code → LangWatch gateway (VK auth, budgets, tracing) → self-hosted vLLM
```

Today the only workaround is the `custom` (OpenAI-compatible) provider, which speaks the OpenAI wire format — Anthropic SDK clients reject those responses (Zod validation on the Messages event union), so it cannot serve `/v1/messages` traffic.

This is the Anthropic-side sibling of #5326 (custom OpenAI-compatible provider base URL routing).

## Proposed solution

- **Materialiser:** include `anthropic` in the base-URL gate; the registry `endpointKey` (`ANTHROPIC_BASE_URL`) resolves the value. Providers without the key keep emitting no `base_url`, so stock Anthropic routing is unchanged.
- **Gateway:** for an anthropic credential carrying `base_url`, derive a deterministic per-endpoint custom provider key (Bifrost `CustomProviderConfig{BaseProviderType: Anthropic}` + `NetworkConfig.BaseURL`), registered so `GetConfigForProvider` can resolve it. Bifrost creates the provider lazily on first dispatch; both the sync raw-forward path and the streaming `PassthroughStream` path build URLs from `NetworkConfig.BaseURL`, so streaming works with no extra code. Credentials without an API key map to a keyless provider config (Bifrost's key selection rejects empty keys for base provider Anthropic; unauthenticated self-hosted servers are common).

Verified end-to-end against vLLM 0.24 (`/v1/messages` sync + native SSE streaming, usage reporting intact). PR to follow.

thiagolmoraes:
Fixes #5938

## What

An `anthropic` ModelProvider with `ANTHROPIC_BASE_URL` configured now dispatches `/v1/messages` (sync and streaming) to that endpoint instead of `api.anthropic.com`, so self-hosted Anthropic-compatible servers (vLLM >= 0.24, Claude-compatible proxies) work behind LangWatch virtual keys, budgets, and tracing. Sibling of #5326 (custom OpenAI-compatible provider base URL).

## How

**Control plane** (`config.materialiser.ts`): include `anthropic` in the per-slot `base_url` gate; the registry `endpointKey` resolves the value. The UI field, zod schema, and litellm path already supported `ANTHROPIC_BASE_URL` — the gateway bundle was the only gap. Providers without the key keep emitting no `base_url`, so stock Anthropic routing is unchanged.

**Gateway** (`adapters/providers/bifrost.go`): Bifrost's Anthropic key has no per-key URL slot and provider config is resolved by provider key alone, so `mapProvider` derives a deterministic per-endpoint custom provider key (`CustomProviderConfig{BaseProviderType: Anthropic}` + `NetworkConfig.BaseURL`) and registers the endpoint for `GetConfigForProvider`. Bifrost creates the provider lazily on first dispatch; the sync raw-forward path and the streaming `PassthroughStream` path both build URLs from `NetworkConfig.BaseURL`, so streaming needs no extra code. Credentials without an API key produce a keyless provider config (Bifrost's key selection rejects empty-value keys for base provider Anthropic; unauthenticated self-hosted servers are common), and keyless-ness is part of the key hash so instances never mix.

## Tests

- Go unit: deterministic key derivation, config shape (base type + normalized URL + keyless), unregistered-key error, plain-API-key path.
- Go end-to-end (in-process bifrost + httptest server): `/v1/messages` dispatch reaches the configured endpoint with `x-api-key`, body forwarded byte-for-byte, native response returned verbatim.
- TS integration (real PG): materialiser emits `ANTHROPIC_BASE_URL` as slot `base_url`; provider without the key emits none.
- Spec scenarios added to `specs/ai-gateway/custom-provider-base-url.feature`.

Verified end-to-end on a live deployment against vLLM 0.24 (`qwen3-14b`): sync + native SSE streaming through a VK, usage reporting intact.

- Go unit: per-endpoint pool stays bounded (bifrost's 1000-worker default would fail it), re-registered endpoint cancels its own pending teardown.
- TS unit: `prepareEnvKeys` resolves credentials for providers whose schema is wrapped in `superRefine`.

Dogfooded on this branch against a live Anthropic-compatible endpoint: keyed and keyless `/v1/messages` (sync + native SSE), `/v1/chat/completions`, and a plain Anthropic provider proving default routing still goes to api.anthropic.com. Usage stamped on every VK.

## Deployment Impact

No new env vars, no helm values added or changed, no chart changes. Vanilla `helm install` behavior is unchanged: Anthropic providers without a base URL keep routing to api.anthropic.com. The only behavioral change ships inside the gateway image — Anthropic providers with a configured base URL route /v1/messages there, and the gateway now bounds its in-memory per-endpoint provider pools (LRU, capacity 32) instead of retaining one per historical endpoint, and sizes each of those pools itself (128 workers, 1024-slot queue) rather than inheriting bifrost's per-provider defaults, which would cost 1000 goroutines per endpoint. Peak gateway memory is therefore bounded by configuration rather than by how many self-hosted endpoints customers have used. BYOC dataplanes pick this up with the normal gateway image upgrade; no operator action or config migration required.

## Summary by CodeRabbit

- **New Features**
  - Added Anthropic-compatible custom base URL routing for `/v1/messages`, with correct streaming and non-streaming behavior.
  - Introduced isolated per-endpoint Anthropic “compat” handling, including keyless operation when no API key is provided.
- **Bug Fixes**
  - Improved Gateway configuration materialisation so Anthropic `base_url` overrides are applied consistently.
- **Tests**
  - Expanded integration, unit, and end-to-end coverage for Anthropic base URL behavior, compat dispatch, eviction/recovery, and credential validation.
- **Chores**
  - Hardened Goosе installer script download/cleanup behavior.


services/aigateway/adapters/providers/bifrost.go
```diff
@@ -953,6 +977,48 @@ func credBaseURL(cred domain.Credential) string {
...
+// credential with a base-URL override and registers its endpoint. The key is
+// a hash of the endpoint identity (URL + keyless-ness), so the same endpoint
+// always lands on the same bifrost worker pool, distinct endpoints never
+// collide, and rotating the API key value alone does not spawn a new provider.
+func anthropicCompatProviderKey(cred domain.Credential) bfschemas.ModelProvider {
+	endpoint := anthropicCompatEndpoint{
+		// Same "/v1"-stripping as the OpenAI-compat path: Bifrost's
+		// Anthropic provider appends the full "/v1/messages" path itself.
+		baseURL: normalizeOpenAICompatBaseURL(credBaseURL(cred)),
+		keyless: strings.TrimSpace(cred.APIKey) == "",
```
langwatch-agent, line 1236: **[P1] Keyless Anthropic endpoints cannot be configured through the control plane.**

This branch correctly marks an empty API key as `IsKeyLess`, but the only normal creation/update path rejects it first: `modelProviders.anthropic.keysSchema` still requires `ANTHROPIC_API_KEY: z.string().min(1)`, and `ModelProviderService.validateAndCleanKeys` applies that schema server-side (the form does too). A customer who supplies `ANTHROPIC_BASE_URL` and leaves the key empty, as required by #5938 and the new feature spec, gets an invalid-credentials error and never receives a gateway slot with an empty `api_key`.

Please make the registry allow an empty/missing Anthropic key when a non-empty base URL is present, mark the field optional in that configuration, and add a control-plane regression test for create/update plus materialisation. The dispatch test should also exercise that resulting keyless slot and assert that the upstream receives no `x-api-key` header.

services/aigateway/adapters/providers/anthropic_compat_test.go
```diff
@@ -0,0 +1,229 @@
...
+// End-to-end through a real bifrost instance: the derived provider key must
+// make bifrost lazily create an Anthropic-based custom provider pointed at
+// the credential's endpoint, and the /v1/messages raw-forward path must
+// deliver the inbound body there unchanged and hand back the server's native
+// response bytes. This is the proof the registry → GetConfigForProvider →
+// provider-creation chain actually holds together, which the config-shape
+// tests above cannot give.
+//
+// Spec: specs/ai-gateway/custom-provider-base-url.feature
+func TestDispatch_MessagesReachesAnthropicCompatEndpoint(t *testing.T) {
```
langwatch-agent, line 220: **[P2] The newly specified streaming contract has no end-to-end regression test.**

This test proves only synchronous `Dispatch`. The feature also promises that `DispatchStream` reaches the custom base URL and forwards native Anthropic SSE frames unchanged, but no test opens a stream against an `httptest` endpoint. Please add that case, ideally keyless as well, and assert `/v1/messages`, no `x-api-key`, and byte-for-byte raw SSE framing. This is the only protection against a future custom-provider/config resolution change silently regressing the client-facing streaming path.

langwatch-agent (changes requested):
## Review summary

`#5938` is correctly addressed on the gateway side for configured Anthropic endpoints: the endpoint policy runs before provider creation, the custom provider preserves the Anthropic base type, and the focused Go endpoint-dispatch test passes. However, the stated keyless self-hosted flow cannot be configured through the control plane, so this is not ready to merge.

Validation run:
- `go test ./adapters/providers -count=1`
- `pnpm exec vitest run src/server/modelProviders/__tests__/registry.unit.test.ts src/server/api/routers/__tests__/modelProviders.unit.test.ts --reporter=dot` (97 passed)
- `pnpm run typecheck`

The PG integration test requires the local test-service setup that is unavailable in this worktree; its targeted run correctly failed before executing tests because PostgreSQL/ClickHouse/Redis test services were not provisioned.

One blocker and one test gap below.

thiagolmoraes:
Addressed all review points in 803c95246:

- **[P1] Keyless flow configurable**: anthropic keysSchema now mirrors openai — either the API key or the base URL must be non-empty (superRefine), with registry unit tests covering keyless+base-URL, key-only, and neither. `validateAndCleanKeys` picks this up server-side automatically.
- **[P2] Streaming regression test**: added `TestDispatchStream_MessagesStreamsFromKeylessAnthropicCompatEndpoint` — real bifrost + httptest SSE endpoint, asserts `/v1/messages` path, no `x-api-key` for the keyless credential, and byte-for-byte native SSE frame forwarding.
- **Minor**: replaced the single `r.Body.Read` with `io.ReadAll` in the sync dispatch test.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -953,6 +977,48 @@ func credBaseURL(cred domain.Credential) string {
...
+	keyless bool
+}
+
+// anthropicCompatEndpoints maps a derived provider key to its endpoint.
+// Bifrost resolves provider config by provider key alone — GetConfigForProvider
+// has no credential context — so mapProvider records the endpoint here before
+// the dispatch enqueues, and config resolution looks it up. Entries are tiny
+// and bounded by the number of distinct customer endpoints, so they are never
+// evicted.
+var anthropicCompatEndpoints sync.Map
```
langwatch-agent: **P2 — This process-global registry is not bounded by the active configuration and can only grow.** `mapProvider` stores every distinct base URL ever dispatched to, including endpoints removed or changed in the control plane, and the comment explicitly elects never to evict. The provider key also drives Bifrost's lazy provider creation, so normal endpoint rotation across tenants accumulates both stale map entries and potentially stale provider resources for the life of the gateway process. Scope the endpoint configuration to the active credential/request lifecycle, or use a bounded/evicting registry with provider cleanup; cover endpoint rotation/removal so this cannot be an unbounded leak.

langwatch-agent (changes requested):
Static external-contributor review found one lifecycle/resource-management issue in the custom-provider registry; details are inline.

langwatch/src/server/gateway/__tests__/config.materialiser.integration.test.ts
```diff
@@ -367,6 +414,39 @@ describe("GatewayConfigMaterialiser — real PG end-to-end", () => {
...
+      const repo = new VirtualKeyRepository(prisma);
+      const vk = await repo.findById(VK_ID, ORG_ID);
+      const mat = new GatewayConfigMaterialiser(prisma, null);
+      const bundle = await mat.materialise(vk!);
+      const slot = bundle.providers.find(
+        (p) => p.id === MP_ANTHROPIC_PLAIN_ID,
+      );
+      expect(slot).toBeDefined();
+      expect(slot!.type).toBe("anthropic");
+      expect(slot!.base_url).toBeUndefined();
```
Aryansharma28, line 534: [P3] Missing real-pipeline regression test for the keyless configuration.

**Why:** The original P1 bug this PR fixed (`keysSchema` rejecting an empty `ANTHROPIC_API_KEY`) was specifically about the create/update *service path*, not the schema in isolation. This test file adds two new real-Postgres cases — keyed+base-URL (`MP_ANTHROPIC_BASE_ID`) and key-only (`MP_ANTHROPIC_PLAIN_ID`) — but no case with an **empty** `ANTHROPIC_API_KEY` alongside `ANTHROPIC_BASE_URL`. That's the exact shape of credential the whole feature exists to support ("unauthenticated self-hosted servers are common", per the PR description), and it's the shape that used to be rejected before this fix.

Since `ModelProviderService.validateAndCleanKeys` reuses the identical `modelProviders.anthropic.keysSchema` object that `registry.unit.test.ts` already unit-tests, this isn't a live-bug risk today (there's no second copy of the validation logic to drift). But there's no regression anchor at the layer (service → Prisma → materialiser) where the bug actually manifested, so a future refactor of `validateAndCleanKeys` or the Prisma write path could silently reintroduce it without any test catching it.

**Evidence:** Compare the two `it(...)` blocks added here (lines 417–431 and 434–448) — both set a non-empty `ANTHROPIC_API_KEY`. Neither exercises `ANTHROPIC_API_KEY: ""` with a base URL through `repo.findById` → `mat.materialise`.

**Suggested fix:** Add a third `prisma.modelProvider.create` with `customKeys: { ANTHROPIC_BASE_URL: ANTHROPIC_BASE_URL_OVERRIDE }` (no `ANTHROPIC_API_KEY`, or empty string), include its id in `modelProviderIds`, and assert `slot!.credentials.api_key === ""` and `slot!.base_url === ANTHROPIC_BASE_URL_OVERRIDE` — mirroring the Go-side `TestGetConfigForProvider_AnthropicCompatKeylessWhenNoAPIKey` but at the materialiser layer.

services/aigateway/adapters/providers/anthropic_compat_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// Streaming sibling of the dispatch test above, keyless: DispatchStream
+// must reach the configured endpoint's /v1/messages, send no x-api-key
+// (unauthenticated self-hosted server), and forward the server's native
+// Anthropic SSE frames byte-for-byte — Anthropic SDK clients Zod-validate
+// every event and reject any reshaped chunk.
+//
+// Spec: specs/ai-gateway/custom-provider-base-url.feature
+func TestDispatchStream_MessagesStreamsFromKeylessAnthropicCompatEndpoint(t *testing.T) {
```
Aryansharma28, line 289: [P5] Praise: this is real end-to-end verification, not a mock.

**Why:** This test (and its sync sibling `TestDispatch_MessagesReachesAnthropicCompatEndpoint` above) stands up an actual `httptest.Server` and a real `NewBifrostRouter`/`router.DispatchStream` call, rather than mocking Bifrost's internals. That's the only thing that actually proves the registry (`anthropicCompatEndpoints`) → `GetConfigForProvider` → lazy provider creation → raw SSE forwarding chain holds together — config-shape unit tests alone (like the two above it) can't catch a wiring break between those layers. I independently confirmed both this test and the whole `adapters/providers` package pass, including under `go test -race -count=2`.

This is also exactly the test that resolved `@langwatch-agent`'s earlier P2 ("no streaming regression test") — good response to review feedback, and the keyless-header assertion (`captured.hasAPIKey`) is a nice touch that a less careful test would have skipped.

No action needed — just flagging a pattern worth keeping for future gateway provider additions.

Aryansharma28 (commented):
## goated-review — deep audit

**Verdict: COMMENT (not approve).** This PR is well-engineered and unusually well-tested for its size, but it is **not yet mergeable**: `@langwatch-agent`'s second review (2026-07-21) flagged an unbounded process-global registry that I independently verified against the vendored Bifrost source and confirmed is real — see below. No new P0/P1 from my pass.

**P-counts (mine, after dedupe against existing reviewers):** P0: 0 · P1: 0 · P2: 0 · P3: 1 · P4: 0 · P5: 3

### What I did
Cloned the PR branch (`803c952`, head of the PR), read every existing review/comment/CI signal, then independently re-verified rather than trusting claims at face value:
- Ran `go build ./...` and `go test ./adapters/providers/... -race -count=2` on the actual PR code — clean build, all 7 new/changed tests pass, no data races.
- Read the vendored `github.com/maximhq/bifrost/core@v1.4.22` source (not just this repo's wrapper) to confirm `CustomProviderConfig.IsKeyLess` genuinely gates key selection (`bifrost.go:5128`, `providerRequiresKey`/`utils.go:90`), and that `BaseProviderType` genuinely drives provider-implementation dispatch (`bifrost.go:3705`) — the PR's stated mechanism is real, not hand-waved.
- Verified `GetKeysForProvider` reads the credential fresh from request context every call (`bifrost.go` `account.GetKeysForProvider`), so provider-key sharing across tenants pointed at the same base URL does **not** leak API keys across tenants — the shared registry only carries `{baseURL, keyless}`, never a credential.
- Re-implemented the new Zod `keysSchema` in an isolated sandbox (matching `zod@3.25.76`, the pinned version) and fuzzed it with empty-string/null/whitespace-only combinations for both fields — behavior matches the intended "at least one of key/base-url must be non-empty" semantics exactly.
- Confirmed no new dependencies were added (`go.mod`/`go.sum`/`package.json`/lockfiles untouched) and grepped the full diff for exfiltration/logging-of-secrets/backdoor patterns — none found. Supply-chain surface of this PR is clean.
- Confirmed the existing SSRF guard (`validateCredentialEndpoints` → `credBaseURL`) already runs generically before `mapProvider`, so the new Anthropic base-URL path inherits the same host/IP allowlisting as the pre-existing custom/OpenAI paths — no new SSRF surface.
- Confirmed all 5 new Gherkin scenarios in `custom-provider-base-url.feature` map to actual Go/TS tests (this repo's `specs/*.feature` are prose requirements referenced by test comments, not an executable Cucumber harness — convention is satisfied).

### Existing findings — not re-raised, status confirmed
- **CodeRabbit** (`anthropic_compat_test.go:167`, single `r.Body.Read` truncation risk) — ✅ resolved in `803c952` (now `io.ReadAll`), confirmed by reading the current file.
- **`@langwatch-agent` P1** (`bifrost.go:1014`, keyless credentials rejected by `keysSchema` before reaching the gateway) — ✅ resolved in `803c952`: `registry.ts` now uses `superRefine` requiring only one of key/base-url, with unit tests; I independently re-verified the schema logic (see above) and it is correct.
- **`@langwatch-agent` P2** (`anthropic_compat_test.go:149`, no streaming regression test) — ✅ resolved: `TestDispatchStream_MessagesStreamsFromKeylessAnthropicCompatEndpoint` was added and I confirmed it passes (including under `-race`).
- **`@langwatch-agent` P2, still open** (`bifrost.go:1002`, `anthropicCompatEndpoints sync.Map` never evicts, and the provider key it drives causes Bifrost to permanently retain a worker-pool + goroutines per distinct historical `{baseURL, keyless}` pair for the life of the process) — **I verified this independently and it is real.** `mapProvider` is called on every `Dispatch` (`bifrost.go:143`), and the vendored `prepareProvider`/`createBaseProvider` path (`bifrost.go:3776` region in the vendored library) permanently appends to `bifrost.providers` and starts dedicated request-queue workers per provider key with no corresponding removal path for `anthropic-url-*` keys. Endpoint rotation (a customer changing `ANTHROPIC_BASE_URL`, or flipping keyed↔keyless) accumulates stale entries and live goroutines forever. This is the one thing keeping `reviewDecision = CHANGES_REQUESTED` and I agree it should block merge — I have nothing to add beyond confirming it holds up.

### My new findings

**[P3] Missing end-to-end regression test for the keyless-through-real-pipeline flow.** The original P1 bug (keyless Anthropic credentials rejected) was specifically about the *create/update service path*, not the schema in isolation. `config.materialiser.integration.test.ts` adds two new real-Postgres cases (keyed+URL, key-only) but no case that creates a `ModelProvider` row with an **empty** `ANTHROPIC_API_KEY` + a base URL and confirms it survives `validateAndCleanKeys` → Prisma → materialisation with `credentials.api_key === ""`. Since `validateAndCleanKeys` reuses the exact same `keysSchema` object tested in `registry.unit.test.ts`, this isn't a live-bug risk today — but it means the actual historical failure mode has no regression anchor at the layer where it manifested. See inline comment for the exact spot to add it.

**[P5 PRAISE] Rigorous, non-mocked verification on the Go side.** `TestDispatch_MessagesReachesAnthropicCompatEndpoint` and its streaming sibling stand up a real `httptest.Server` + a real `NewBifrostRouter`, not mocks — that's the only thing that actually proves the registry→`GetConfigForProvider`→lazy-provider-creation chain holds together end to end, and it caught a real gap (the missing streaming test) in review. Nice work.

**[P5 OBSERVATION] Cosmetic UI mismatch, not unique to this PR.** `ModelProviderCredentialsSection.tsx` derives the "required" asterisk from `optionalKeys`, which still lists only `ANTHROPIC_BASE_URL` — so the API-key field will show as required even when a base URL makes it legitimately optional. This is not a regression: `optionalKeys` for `openai` has the identical characteristic since #5326, so it's a pre-existing, cross-cutting UX nit rather than something this PR introduced. Flagging for awareness, not requesting a change here.

- The Go test suite ran clean locally against this exact commit, but I could not confirm this repo's actual CI ran the full test/typecheck pipeline on this PR — the check-runs for `803c952` show `preflight`/`firefighting`/`evaluate` all **skipped** (this is a fork PR from an external contributor, `author_association: CONTRIBUTOR`, `head_repo: thiagolmoraes/langwatch`), so whatever gating workflow runs the real Go/TS test suites with secrets does not appear to have executed here yet. `@langwatch-agent`'s first review separately confirmed `pnpm exec vitest` (97 passed) and `pnpm run typecheck` against the pre-fix commit, but not against `803c952`.
- I could not run the Postgres-backed `config.materialiser.integration.test.ts` (no local Postgres/ClickHouse/Redis in this environment) — same limitation `@langwatch-agent` reported.
- I did not attempt to reproduce the author's claimed live-deployment verification against vLLM 0.24 (`qwen3-14b`) — no live self-hosted endpoint available to me.

---
*goated-review — summary + inline findings posted together.*

langwatch/src/server/modelProviders/registry.ts
```diff
@@ -273,10 +273,27 @@ export const modelProviders = {
...
+            message:
+              "Either ANTHROPIC_API_KEY or ANTHROPIC_BASE_URL must be provided with a non-empty value",
+          });
+        }
+      }),
+    // The base URL points at an Anthropic-compatible self-hosted server
+    // (vLLM >= 0.24). Such servers commonly run unauthenticated, so the
+    // API key may be empty when a base URL is set — mirroring openai
+    // above. api.anthropic.com itself always requires the key.
     optionalKeys: ["ANTHROPIC_BASE_URL"],
```
Aryansharma28, line 326: [P5] Observation: `optionalKeys` doesn't reflect the new conditional-optionality of `ANTHROPIC_API_KEY`, but this isn't new to this PR.

**Why:** `ModelProviderCredentialsSection.tsx` renders the required-field asterisk from `optionalKeys` (`optionalKeySet.has(key)`). Since only `ANTHROPIC_BASE_URL` is listed here, the `ANTHROPIC_API_KEY` input will still show as required in the UI even for a customer who intends to leave it empty because they've set a base URL — even though the underlying `keysSchema` (correctly) allows that combination now.

I checked whether this is a regression introduced by this PR: it isn't. `openai`'s entry above has the exact same shape (`optionalKeys: ["OPENAI_BASE_URL"]`, not `OPENAI_API_KEY`) since #5326, and ships with the identical cosmetic mismatch today. So this is a pre-existing, cross-cutting UX nit rather than something this PR should be expected to fix.

**Suggested fix (optional, out of scope for this PR):** A follow-up that makes `optionalKeySet` derive from the schema's `superRefine` conditions (or from a small helper both `openai` and `anthropic` opt into) rather than a static list, so the "required" indicator reflects actual validity for both providers at once.

Aryansharma28:
Hey @thiagolmoraes — thanks for this one, left a review above (came back clean, nice work). Heads up that the full CI suite hadn't actually run on this PR until just now — it's all green as of the latest run. The one open item is a pre-existing worker-pool leak (`anthropicCompatEndpoints sync.Map` never evicting) that's not something you introduced, just flagging for visibility.

Aryansharma28 (dismissed):
CI is fully green (all real build/test/lint/integration/CodeQL jobs pass as of the latest run) and the goated-review pass found 0 new P0/P1/P2 issues introduced by this change. Approving from my side. Note: the pre-existing `anthropicCompatEndpoints sync.Map` worker-pool leak (not introduced by this PR) is still flagged as open by langwatch-agent's review, so this may stay blocked pending that being addressed or dismissed separately.

langwatch-agent (changes requested):

rogeriochaves (commented):
## Review + fixes pushed

Reviewed the whole PR against `specs/ai-gateway/`, read the vendored `bifrost/core@v1.4.22` to check the mechanism rather than take it on trust, merged `main` in, fixed three defects, and dogfooded the feature end to end against a live Anthropic-compatible endpoint. **The design is right and the feature works.** Everything below is either a fix already pushed or a note.

### The mechanism holds up

Verified in the vendored bifrost source, not inferred:

- `createBaseProvider` sets `CustomProviderConfig.CustomProviderKey = string(providerKey)` itself (`bifrost.go:3703`), so the derived key is what `GetProviderKey()` returns and `removeProviderFromSlice` matches on. Eviction really does remove the provider.
- `requestWorker` receives its `ProviderConfig` at creation and never re-reads it, so an endpoint evicted from the registry mid-flight cannot have its config yanked out from under a running request.
- `providerRequiresKey` genuinely short-circuits on `IsKeyLess` for base provider Anthropic, and `GetKeysForProvider` reads the credential fresh off the request context, so two tenants pointed at the same base URL share a worker pool but never a credential.
- `credBaseURL` is generic, so `validateCredentialEndpoints` (SSRF host/IP policy) already covered the new path. No new egress surface.

### Fixed

**1. Per-endpoint worker pools inherited bifrost's defaults, at 1000 goroutines each.** This is the one I would have blocked on. Unlike the OpenAI-compat path, where every customer endpoint shares the single VLLM provider, each compat endpoint here gets its own bifrost provider, and an unconfigured provider is sized `Concurrency: 1000` / `BufferSize: 5000`. Measured on the branch, the first dispatch to a compat endpoint grew the process by **1003 goroutines** — so a full registry meant roughly 32k goroutines that customer configuration alone could add, several times the gateway's own baseline. `GetConfigForProvider` now sizes these pools explicitly (128 workers, 1024-slot queue): **131 goroutines per endpoint**, still far more concurrency than a self-hosted server absorbs. `TestDispatch_AnthropicCompatEndpointPoolIsBounded` fails if the defaults come back.

**2. Eviction could tear down an endpoint that had just come back.** `onEvict` schedules `RemoveProvider` on its own goroutine, so a dispatch to the same endpoint could land in between, re-register, and enqueue onto the queue `RemoveProvider` was closing — bifrost answers those with `provider is shutting down`. `releaseEvicted` now skips the teardown when the key is live again, and the next eviction reschedules it. This is CodeRabbit's generation guard, in the form the data structure already permits: after eviction the key is simply absent, so "same generation" is "still absent". Covered by `TestAnthropicCompatRegistry_ReRegisteredEndpointCancelsPendingTeardown`.

**3. `prepareEnvKeys` returned no credentials at all for anthropic.** It reads credential names off `keysSchema` through a shape reader that did not unwrap `superRefine`, so the moment anthropic moved to the openai-style either/or schema it resolved to `{}`. openai has had this since #5326; the repo has two `getSchemaShape` helpers and each unwraps only half of what zod produces. They are now one helper that unwraps effects and optional wrappers alike. `prepareEnvKeys.unit.test.ts` fails against the old reader (`expected {} to deeply equal {...}` for both providers).

Also merged `main` (bifrost.go conflicted against the codex dispatch and OpenAI test-backend work), and restated the endpoint-rotation scenario in user-visible terms per CodeRabbit and the feature-file convention in CLAUDE.md.

### Dogfooded end to end

Full stack on this branch: app + control plane + the Go gateway built from this code, three anthropic ModelProviders created through the real service write path (`updateModelProvider`, the path that used to reject the keyless shape), a virtual key each, and a local server standing in for a self-hosted Anthropic-compatible endpoint that forwards to api.anthropic.com so the answers are real model output.

| What | Result |
|---|---|
| Keyed VK, `POST /v1/messages` | Reached the configured endpoint at `/v1/messages` with `x-api-key: sk-ant-dog…`, body byte-for-byte, native Anthropic response returned verbatim |
| Keyed VK, `POST /v1/messages` streaming | `"stream":true` reached the endpoint; native SSE frames arrived unchanged: `message_start` → `content_block_start` → `ping` → `content_block_delta` → `content_block_stop` → `message_delta` → `message_stop` |
| Keyless VK, `POST /v1/messages` | Routed and answered; the endpoint saw **no `x-api-key` header at all** |
| Plain anthropic VK, no base URL | Went to api.anthropic.com (real `request_id`, `invalid x-api-key` for the dummy key) and **never touched the local endpoint** — default routing unchanged |
| Keyed VK, `POST /v1/chat/completions` | Translated to Messages shape, delivered to the endpoint, and the response carries `extra_fields.provider: "anthropic-url-1b2655d8d868c8fb"` — the derived per-endpoint key, visible in the wire response |

`lastUsedAt` was stamped on all three virtual keys, so usage reporting survives the derived provider.

Tests: `go test ./services/aigateway/... -count=1` clean, `adapters/providers` clean under `-race`, `pnpm typecheck` clean, `config.materialiser.integration.test.ts` 14/14 against real Postgres, and the touched unit suites pass.

### Left alone deliberately

- **`optionalKeys` still marks `ANTHROPIC_API_KEY` required in the drawer** even when a base URL makes it optional. Real, but openai has shipped the identical mismatch since #5326; fixing it properly means deriving the required set from the schema for both providers at once, which is its own change.
- **The configured API key is sent to the customer's endpoint** when a base URL is set. Intended, and identical to the openai/custom paths, but worth knowing that adding a base URL to an existing Anthropic provider starts sending that provider's real key to the new host.
- **`gateway-matrix` skips on this PR.** Expected: it needs live provider secrets, which GitHub strips on fork PRs, and the workflow guards for exactly that. Not a red flag.
- The **guardrail stub, deleted gateway metrics, budget accrual and coding-agent onboarding** issues (#6139/#6140/#6141/#6142) are being fixed separately. This PR does not touch or worsen any of them.

rogeriochaves:
The suites have now run on maintainer-approved workflows, and I also executed the branch locally rather than reviewing the diff statically.

**CI, maintainer-approved on the current head (`7fece7ada`):** every workflow this PR touches ran and passed, including `go-ci`, `go-services`, `langwatch-app-ci`, `e2e-ci`, `codeql` and `ssrf-conformance`. The earlier heads showed `action_required` because GitHub holds fork-PR workflows until a maintainer approves them, not because the suites were missing; they have been approved on each push since.

`gateway-matrix` does not run here, and that is by design rather than a gap: it needs live provider secrets, which GitHub strips on fork PRs, and the workflow guards for exactly that case (see its header comment). The provider behaviour it would have covered is instead exercised by the in-process bifrost tests in `adapters/providers` and by the dogfooding below.

**Executed locally on this branch:**
- `go test ./services/aigateway/... -count=1` clean; `adapters/providers` clean under `-race` and repeated `-count`.
- `pnpm typecheck` clean.
- `config.materialiser.integration.test.ts` 14/14 against real Postgres.
- The touched unit suites pass, and the two new regression tests fail against the pre-fix code (`prepareEnvKeys` returns `{}`; the pool test sees bifrost's 1000-worker default).

**Ran the real thing:** app + control plane + the Go gateway built from this branch, three anthropic providers created through `updateModelProvider`, a virtual key each, and a live Anthropic-compatible endpoint. Keyed and keyless `/v1/messages` (sync and native SSE) plus `/v1/chat/completions` all reached the configured endpoint; the plain Anthropic provider still went to api.anthropic.com and never touched it. Details and the three defects I found and fixed are in my review above.

.github/scripts/install-goose.sh
```diff
@@ -8,7 +8,10 @@ GOOSE_VERSION="v3.26.0"
 GOOSE_URL="https://github.com/pressly/goose/releases/download/${GOOSE_VERSION}/goose_linux_x86_64"
 GOOSE_SHA256="8b3eee9845cd87d827ba1abddb85235fb3684f9fb1666426f647ddd12fd29efe"
 
-curl -fsSL "$GOOSE_URL" -o /tmp/goose
+# Retry the download: a single connection reset against the GitHub release
+# CDN otherwise fails a whole integration shard before any test has run.
+curl -fsSL --retry 5 --retry-delay 2 --retry-connrefused --retry-all-errors \
+  "$GOOSE_URL" -o /tmp/goose
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 69f854dc2. The download now goes to a `mktemp` file (0600, so nothing can pre-create it), a trap removes it on any exit, and `sudo install -m 0755` places the verified bytes in one step instead of `mv` + `chmod`. Verified locally: `bash -n` and `shellcheck` clean, and the real download still checksums against the pinned SHA.

Worth noting this path is pre-existing — I was in the file to add a curl retry after a `Recv failure: Connection reset by peer` took out a whole integration shard on this PR before any test ran.

rogeriochaves:
Filed the two items I deliberately left out of this PR, so they do not live only in a review comment:

- #6172 — the drawer marks the API key required while the schema treats it as optional when a base URL is set. Both `openai` and `anthropic` have it, so the fix derives the required set from the schema for both at once; the `getSchemaShape` consolidation in this PR is the foundation for that.
- #6173 — adding a base URL re-points an existing provider credential at the new host with no warning. Existing behaviour across `openai`, `custom` and now `anthropic`, filed as a UX and security-hygiene question rather than a bug.

Both are on the Kanban with Status Ready. Neither blocks this PR.

services/aigateway/adapters/providers/bifrost.go
rogeriochaves: Resolved. \`modelProviders.anthropic.keysSchema\` is now a \`superRefine\` requiring only that one of \`ANTHROPIC_API_KEY\` / \`ANTHROPIC_BASE_URL\` be non-empty (`registry.ts:305-321`), mirroring the openai entry, so `validateAndCleanKeys` accepts the keyless shape server-side. Covered at three layers: `registry.unit.test.ts` for the schema, `config.materialiser.integration.test.ts` for the service -> Prisma -> materialiser path against real Postgres (`MP_ANTHROPIC_KEYLESS_ID`, asserting `credentials.api_key === ""` and the emitted `base_url`), and `TestGetConfigForProvider_AnthropicCompatKeylessWhenNoAPIKey` plus `TestDispatchStream_MessagesStreamsFromKeylessAnthropicCompatEndpoint` on the Go side, the latter asserting the upstream receives no `x-api-key` header at all.

services/aigateway/adapters/providers/anthropic_compat_test.go
rogeriochaves: Resolved. `TestDispatchStream_MessagesStreamsFromKeylessAnthropicCompatEndpoint` (`anthropic_compat_test.go:289`) stands up a real `httptest` SSE endpoint and a real `NewBifrostRouter`, then asserts the `/v1/messages` path, the absence of `x-api-key` for the keyless credential, and byte-for-byte native SSE frame forwarding.

services/aigateway/adapters/providers/bifrost.go
rogeriochaves: Resolved. The registry is now a bounded LRU (capacity 32) rather than an ever-growing `sync.Map`, and eviction tears down the bifrost provider behind the evicted key. Two further problems surfaced while fixing it and are fixed too: each compat endpoint was inheriting bifrost's unconfigured per-provider defaults (`Concurrency: 1000`), so `GetConfigForProvider` now sizes these pools explicitly at 128 workers / 1024-slot queue; and the asynchronous teardown could remove a provider that a concurrent dispatch had just re-registered, which `releaseEvicted` now guards against. Covered by `TestAnthropicCompatRegistry_EvictsLeastRecentlyDispatchedBeyondCap`, `TestDispatch_EvictedAnthropicCompatEndpointRecoversOnNextDispatch`, `TestAnthropicCompatRegistry_ReRegisteredEndpointCancelsPendingTeardown` and `TestDispatch_AnthropicCompatEndpointPoolIsBounded`.

langwatch/src/server/gateway/__tests__/config.materialiser.integration.test.ts
rogeriochaves: Good catch, and agreed that the schema-in-isolation test was not an anchor at the layer where the bug manifested. Added: `config.materialiser.integration.test.ts` now creates `MP_ANTHROPIC_KEYLESS_ID` with `ANTHROPIC_BASE_URL` and no `ANTHROPIC_API_KEY`, and asserts the materialised slot carries `api_key === ""` alongside the base URL, exercising the service -> Prisma -> materialiser path against real Postgres.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -953,6 +1017,142 @@ func credBaseURL(cred domain.Credential) string {
...
+	onEvict := reg.onEvict
+	reg.mu.Unlock()
+
+	if onEvict != nil {
+		for _, k := range evicted {
+			onEvict(bfschemas.ModelProvider(k))
+		}
+	}
+	return key
+}
```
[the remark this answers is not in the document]
rogeriochaves: Resolved. `onEvict` now schedules `releaseEvicted`, which skips the teardown when a dispatch has put the key back in the registry, and the next eviction reschedules it. This is the generation guard in the form the data structure already permits: after eviction the key is simply absent, so "still the same generation" is "still absent". Covered by `TestAnthropicCompatRegistry_ReRegisteredEndpointCancelsPendingTeardown`.

specs/ai-gateway/custom-provider-base-url.feature
```diff
@@ -48,3 +54,39 @@ Feature: Custom (OpenAI-compatible) provider routing to customer endpoints
...
+    When a /v1/messages request is dispatched through it
+    Then the dispatch does not fail credential validation
+    And the upstream request carries no x-api-key header
+
+  Scenario: An endpoint that fell out of recent use still works when dispatched to again
+    Given an "anthropic" model provider with a base URL that has not been dispatched to recently
+    And many other Anthropic-compatible endpoints have been dispatched to since
+    When a /v1/messages request is dispatched through it again
+    Then the upstream request reaches its configured endpoint
+    And endpoints outside the recently used set hold no gateway resources in the meantime
```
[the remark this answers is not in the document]
rogeriochaves: Resolved. The scenario now reads in user-visible terms only: the endpoint still receives the request after heavy rotation, with no latency the caller can notice. LRU capacity, worker pools and eviction stay in the Go registry tests where they belong.

langwatch/src/server/modelProviders/registry.ts
rogeriochaves: Agreed, and deliberately left alone here: openai has had the identical mismatch since #5326, so fixing it properly means deriving the required set from the schema for both providers at once. Filed as #6172, with the `getSchemaShape` consolidation in this PR as its foundation.

rogeriochaves:
## Round 2: merged main twice, and dogfooded the "can Claude Code reach any provider" question

### Branch state

Merged `origin/main` twice (the second time to pick up #6143, which reworked guardrails to fail closed and changed `Precheck` from `BudgetVerdict` to `BudgetDecision`). One conflict, in `bifrost.go`'s router struct, where `anthropicCompat` and main's new `discovery*` fields were added at the same spot. Both kept.

A clean text merge is not a working merge on that one, so after it: `go build ./services/aigateway/... ./pkg/... ./cmd/...` clean, `go test ./services/aigateway/... -count=1` clean, `pnpm typecheck` clean, and `golangci-lint run ./services/aigateway/... ./services/nlpgo/... ./pkg/... ./cmd/... ./tools/migrationorder/...` reports 0 issues. Workflow runs on the new head are approved and running.

### Review threads

Replied in-thread on each open thread with what landed and where the test is. All the substantive requests were already implemented on the branch and I verified each against the code rather than taking the earlier summary at face value: the `superRefine` keyless schema, the keyless materialiser case against real Postgres, the bounded LRU with provider teardown, the eviction/re-registration guard, the explicit pool sizing, the streaming regression test, and the user-visible rewording of the endpoint-rotation scenario. Nothing new to fix. I have not dismissed any review; that stays the maintainer's call.

### The dogfood, three ways, with a real Claude Code

A real Claude Code 2.1.219 in a nested tmux session, pointed at the gateway with `ANTHROPIC_BASE_URL` and a virtual key, driven through a task that forces a `Read` tool call, an **MCP** tool call against a purpose-built stdio MCP server, and a `Bash` tool call. Not a curl mock. Full stack built from this branch: app, control plane, and the Go gateway, four providers created through `updateModelProvider`, a routing policy and virtual key each.

| # | Route | Result |
|---|---|---|
| 1 | Claude Code -> gateway -> **api.anthropic.com** | **Works.** All three tool calls round-trip, 70 SSE frames, prompt caching intact |
| 2 | Claude Code -> gateway -> **self-hosted Anthropic-compatible endpoint** (`ANTHROPIC_BASE_URL`, what this PR adds) | **Works.** Same three tool calls, 69 SSE frames, all 4 turns land on the configured endpoint with `x-api-key` and the body byte-for-byte |
| 3a | Claude Code -> gateway -> **OpenAI** (`gpt-5.6-luna`) | **Fails** |
| 3b | Claude Code -> gateway -> **Gemini** (`gemini-2.5-flash`) | **Fails** |

Case 2 is what this PR claims, and it holds end to end. Merging it is the right call.

### Case 3, precisely

The two failures are identical in kind, which is what makes this actionable: it is the wire shape, not any one provider.

**Non-streaming** returns the upstream provider's own parse error, because the Anthropic body arrives verbatim:

- OpenAI: `400 Unknown parameter: 'system'`, and `400 Unsupported parameter: 'max_tokens'` on the mandatory Anthropic field
- Gemini: `400 Invalid JSON payload received. Unknown name "max_tokens" ... Unknown name "messages"`

**Streaming**, the only path Claude Code uses, is worse. It hangs. Zero bytes, no headers, no error frame, for as long as the client is willing to wait. The gateway logged `status=200 bytes=100` at `4m57s` and `5m0.9s` for the two Claude Code runs, and at `6m40s` for a curl that waited that long, so the duration tracks the client giving up, not any server-side bound. Claude Code sits in `requesting`, emits `api_retry` with `error: "unknown"` and `error_status: null`, and ends `error_during_execution` with an empty result. There is no error a customer could act on.

Where it originates: `dispatchMessagesStream` (`services/aigateway/adapters/providers/bifrost.go:571-599`) forwards `POST /v1/messages` through `PassthroughStream` regardless of provider family. `POST https://api.openai.com/v1/messages` is a 404, the channel yields neither a chunk nor a `BifrostError`, and `bifrostStreamIterator.Next` blocks on it until the client disconnects. `ProviderRequestTimeoutSeconds` is 14 minutes, so nothing server-side cuts it first.

The same virtual keys and credentials return 200 with real completions through `/v1/chat/completions`, so the credential, routing policy and provider are all fine. OpenAI-in to any-provider-out works. Anthropic-in only works to Anthropic.

### The gap, filed

#6184, Status Ready. Bifrost already ships the Anthropic request parser (`AnthropicMessageRequest.ToBifrostResponsesRequest`) and Anthropic-shaped response and event serializers, and every provider auto-downgrades Responses to Chat, so the semantic translation mostly exists and is not being invoked. The real work is the streaming lane: the usable event converter returns typed events unframed and leaves `message_start` / `content_block_start` / `content_block_stop` / `message_stop` sequencing to the caller, and the pre-framed alternative is incomplete. Roughly 8-12 files. The issue carries the measurements above, the risks, and the minimum bar for calling it done.

One adjacent bug surfaced and is noted there: streaming `/v1/messages` goes through `PassthroughStream` for every provider, and Bedrock's implementation returns an unsupported-operation error with no fallback, so streaming `/v1/messages` on a Bedrock virtual key looks broken today. Not verified against a live Bedrock credential.

### Not verified

`gateway-matrix` still skips here, by design, since GitHub strips provider secrets on fork PRs. Case 3 was measured against OpenAI and Gemini only; other non-Anthropic providers are inferred from the shared code path rather than run.

PR #5941, opened 2026-07-19
feat(aigateway): implement GET /v1/models behind the virtual key

### LINKED ISSUE #5939: GET /v1/models always returns an empty list — clients can't discover models behind a virtual key
## Problem

The gateway exposes `GET /v1/models`, but the provider router's `ListModels` is a stub that always returns `nil`, so every VK gets `{"object":"list","data":null}`.

Consequences:

- Model-picker clients that rely on OpenAI-style discovery (OpenWebUI, LibreChat, various SDK helpers) show an empty dropdown behind a LangWatch VK — users must type model names by hand.
- `data: null` (instead of an empty array) breaks some strict parsers.

The information needed to answer is already in the resolved VK bundle: routing-policy `model_aliases`, `models_allowed`, the credential chain (including self-hosted endpoints that serve their own `/v1/models`), and model deny rules in `policy_rules`.

## Proposed solution

Build the list from the bundle:

1. **Aliases** from the routing policy are always listed (e.g. `chat` → `openai/gpt-5-mini`).
2. **If `models_allowed` is set**, it is authoritative: list it plus the aliases and query nothing upstream (anything outside the allowlist is blocked at dispatch anyway).
3. **Otherwise, discover:** for each credential carrying a base URL (custom / OpenAI-compatible / Anthropic-compatible self-hosted servers — vLLM and LiteLLM all serve OpenAI-shape `GET /v1/models`), fetch and aggregate the endpoint's model list. Short timeout; a dead endpoint is skipped instead of blanking the whole response. Hosted credentials without a base URL have no catalog to query and are skipped.
4. **Filter** models matching `policy_rules.models.deny` regexes.
5. Dedupe, stable sort, and emit OpenAI list shape: `{"id", "object": "model", "owned_by"}` with `data` always an array.

Existing spec scenarios in `specs/ai-gateway/gateway-service.feature` ("/v1/models reflects effective VK allowlist") and `specs/ai-gateway/provider-routing.feature` ("GET /v1/models returns aliases + allowed models") describe exactly this behavior and are marked `@unimplemented` today.

PR to follow.

thiagolmoraes:
Fixes #5939

## What

`GET /v1/models` existed but the provider router's `ListModels` was a stub, so every VK got `{"object":"list","data":null}`. Model-picker clients (OpenWebUI, LibreChat, SDK helpers) showed an empty dropdown behind a LangWatch VK, and `data: null` breaks strict parsers.

## How

The list is built from the resolved VK bundle (app layer):

1. **Aliases** from the routing policy are always listed.
2. **`models_allowed` set?** Its literal entries are authoritative and nothing is queried upstream. Wildcard entries (`claude-haiku-*`) are patterns rather than model IDs, so they are expanded against discovery instead of being listed verbatim, and discovered models are filtered back down by the allowlist.
3. **Otherwise, discover:** credentials carrying a base URL (custom / OpenAI-compatible / Anthropic-compatible self-hosted servers, vLLM and LiteLLM all serve OpenAI-shape `GET /v1/models`) are queried and aggregated (adapter layer, 5s timeout, fan-out capped at 8). A failing endpoint is skipped instead of blanking the whole response; hosted credentials without a base URL have no catalog to query and are skipped.
4. **Deny filter:** models matching `policy_rules.models.deny` regexes are dropped, and when a model allow rule exists only matching models survive (discovery only, dispatch-time policy evaluation remains the authority).
5. Dedupe, stable sort, OpenAI list shape (`{"id", "object": "model", "created", "owned_by"}`) with `data` always an array.

Discovery results are cached per credential chain for 60s and concurrent misses collapse onto one probe, because `GET /v1/models` sits outside the dispatch interceptor chain and is therefore never rate limited.

Discovery is the one place the gateway itself fetches a customer-controlled URL, so it runs the same customer-endpoint policy Dispatch applies, never follows redirects, re-checks every resolved address immediately before connect, and caps the response it will read.

Two control-plane defects surfaced while dogfooding this end to end and are fixed here, because the listing is not honest without them:

- `GET /api/internal/gateway/config/:vk_id` fetched the virtual key without its routing policy, so `model_aliases` and `policy_rules` reached the gateway empty. Aliases were never resolved and model deny rules were never enforced.
- The wire mapper kept an alias target's `provider/` prefix on the model name and dropped the provider, so an alias for `openai/gpt-5-mini` dispatched a model literally named `openai/gpt-5-mini`.

Implements the spec scenarios in `specs/ai-gateway/gateway-service.feature` ("/v1/models reflects effective VK allowlist") and `specs/ai-gateway/provider-routing.feature` ("GET /v1/models returns aliases + allowed models"), previously marked `@unimplemented`; adds scenarios for endpoint discovery, allowlist short-circuit, wildcard expansion, deny filtering, redirect refusal, and connect-time address re-checking.

## Tests

- App unit: aliases + allowlist (and no upstream query when the allowlist is literal), wildcard expansion against discovery, discovery when no allowlist, deny-rule and allow-rule filtering, dedupe + stable sort, invalid-pattern logging.
- Adapter unit (httptest): queries `<base>/v1/models` with bearer and `x-api-key` auth, skips credentials without a base URL, skips failing endpoints, dedupes across endpoints, `/v1`-suffix normalization, bounded fan-out, oversized responses rejected, endpoint policy honored, redirects not followed, rebound addresses refused at connect, caching and concurrent-burst collapse, panic-safe cache entries.
- HTTP unit: OpenAI list shape with `created` and a non-empty `owned_by` on every entry.
- Control plane integration (real Hono app, real PG): the config route ships the routing policy's aliases and model deny rules.

## Dogfooding

Ran the stack locally with a self-hosted OpenAI-compatible endpoint serving `qwen3-14b`, `qwen3-32b`, `bge-m3`, `internal-secret-model`, one routing policy (`chat` alias, `models.deny: ["^internal-.*$"]`), and three VKs differing only in `models_allowed`:

| VK | `models_allowed` | `GET /v1/models` |
|---|---|---|
| vk-open | (none) | `bge-m3`, `chat`, `qwen3-14b`, `qwen3-32b` |
| vk-narrow | `["qwen3-14b"]` | `chat`, `qwen3-14b` |
| vk-wildcard | `["qwen3-*"]` | `chat`, `qwen3-14b`, `qwen3-32b` |

`internal-secret-model` is absent everywhere (deny rule), no VK sees a literal `qwen3-*`, and the whole set of calls cost the upstream a single probe. Verified with `curl`, the `openai` Python SDK 2.30.0 (`client.models.list()` parses into `SyncPage[Model]` with `created` typed `int` and `owned_by` non-empty) and the `openai` Node SDK.

## Deployment Impact

No schema, chart, or environment changes. Two behavior changes to be aware of on rollout:

- **Model policy rules start being enforced.** `policy_rules.models` never reached the gateway before, so a configured deny or allow rule was a silent no-op. After this change it is applied at dispatch and reflected in `GET /v1/models`. Operators who configured model rules should expect traffic matching them to start being rejected, which is the documented intent, but it is a live behavior change rather than a new opt-in.
- **Model aliases start resolving.** Aliases were likewise never delivered. A VK whose routing policy defines `chat -> openai/gpt-5-mini` will now dispatch `gpt-5-mini` to OpenAI instead of forwarding the literal alias name upstream.

`GET /v1/models` also makes outbound calls to customer-configured self-hosted endpoints where it previously made none. These are capped at 8 concurrent, 5s each, cached 60s per credential chain, and only ever sent to endpoints the existing customer-endpoint policy already permits for dispatch.


services/aigateway/adapters/providers/bifrost.go
```diff
@@ -742,9 +743,92 @@ func rawResponseFromBifrostError(berr *bfschemas.BifrostError) ([]byte, int, boo
...
+// have no catalog to query and are skipped. A failing endpoint is skipped
+// too: one dead server must not blank out the whole list.
+func (r *BifrostRouter) ListModels(ctx context.Context, creds []domain.Credential) ([]domain.Model, error) {
+	// Query endpoints concurrently: latency is max(endpoint), not
+	// sum(endpoint) — one slow or dead server must not stack its timeout
+	// onto the others. Results keep credential order for determinism.
+	perCred := make([][]string, len(creds))
+	var wg sync.WaitGroup
+	for i, cred := range creds {
+		base := normalizeOpenAICompatBaseURL(credBaseURL(cred))
```
langwatch-agent: **[P1] Model discovery bypasses the customer-endpoint SSRF policy.**

`Dispatch` and `DispatchStream` call `r.validateCredentialEndpoints(ctx, cred)` before they use `base_url`; this new path takes the same customer-controlled value and immediately starts an HTTP request instead. Consequently, a credential that normal traffic would reject under `BlockLocalHTTPCalls`, `RequireHTTPSCustomerEndpoints`, or the host allowlist is still contacted by `GET /v1/models`. The request also sends the configured provider key in `Authorization` and `x-api-key`.

Validate each credential before scheduling discovery, using the same policy and failure semantics as dispatch (skip rejected/unreachable endpoints rather than fail the whole list), and add coverage proving a private/metadata endpoint is never dialed.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -742,9 +743,92 @@ func rawResponseFromBifrostError(berr *bfschemas.BifrostError) ([]byte, int, boo
...
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("upstream /v1/models returned status %d", resp.StatusCode)
+	}
+	var parsed struct {
+		Data []struct {
+			ID string `json:"id"`
+		} `json:"data"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&parsed); err != nil {
```
langwatch-agent: **[P2] A model-list response is unbounded in memory.**

The five-second client timeout limits elapsed time, not response size. `json.Decoder` will retain every `data` entry and every ID in memory, and discovery runs once for every configured endpoint. A broken or compromised endpoint can return an arbitrarily large JSON array and make a single `/v1/models` request consume enough memory to destabilize the gateway.

Decode through a modest `io.LimitReader` and cap the number/length of accepted model IDs; treat an over-limit response as a skipped endpoint. Add a regression test for an oversized response.

langwatch-agent (changes requested):
## Review summary

The response assembly, allowlist short-circuit, policy filtering, ordering, and concurrent discovery are coherent. The new discovery HTTP path, however, bypasses the customer endpoint policy that protects normal provider dispatches. That creates an SSRF/security regression, so this is not ready to merge.

Validation run:
- `go test ./...` in `services/aigateway` (passed)

Two findings below, ordered by severity.

thiagolmoraes:
Addressed all review points in c9ffca3e0 + 1fbe376af:

- **[P1] SSRF**: discovery now runs `validateCredentialEndpoints` (same policy as Dispatch — local-address blocking, HTTPS requirement, host allowlist) before contacting any base URL; policy-blocked endpoints are skipped and logged. Regression test: `TestListModels_HonorsCustomerEndpointPolicy` proves a `BlockLocalHTTPCalls`-rejected endpoint is never contacted.
- **Serial discovery**: already fixed in c9ffca3e0 (pushed while the review ran) — endpoints are queried concurrently, latency is max not sum, with a timing regression test.
- **Spec bindings**: the allowlist test now carries the "does not query endpoints when an allowlist is set" scenario annotation, and a new router test binds "/v1/models reflects effective VK allowlist" end-to-end through the HTTP handler (OpenAI list shape, data always an array).

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -742,9 +743,104 @@ func rawResponseFromBifrostError(berr *bfschemas.BifrostError) ([]byte, int, boo
...
+	// onto the others. Results keep credential order for determinism.
+	perCred := make([][]string, len(creds))
+	var wg sync.WaitGroup
+	for i, cred := range creds {
+		base := normalizeOpenAICompatBaseURL(credBaseURL(cred))
+		if base == "" {
+			continue
+		}
+		wg.Add(1)
+		go func(i int, cred domain.Credential, base string) {
```
langwatch-agent: **P1 — Unbounded fan-out lets a single virtual key exhaust the gateway.** `ListModels` starts one goroutine and outbound connection for every credential, then waits for every endpoint. Neither the number of credentials/endpoints nor concurrent `/v1/models` calls is capped or cached. A project with many configured endpoints, or a client polling this endpoint, can create an unbounded number of 5-second requests and goroutines, starving normal dispatch. Bound discovery with a semaphore/worker pool and a small per-VK result cache (or hard cap the eligible endpoints), and test the cap under repeated requests.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -742,9 +743,104 @@ func rawResponseFromBifrostError(berr *bfschemas.BifrostError) ([]byte, int, boo
...
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("upstream /v1/models returned status %d", resp.StatusCode)
+	}
+	var parsed struct {
+		Data []struct {
+			ID string `json:"id"`
+		} `json:"data"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&parsed); err != nil {
```
langwatch-agent: **P1 — An upstream can make `/v1/models` consume unbounded memory.** This decodes a customer-configured upstream response directly from `resp.Body`, then retains every returned ID for the HTTP response. The endpoint policy prevents SSRF but does not make the remote service trustworthy: a compromised/misconfigured provider can stream an arbitrarily large JSON array, and any virtual-key holder can repeatedly trigger it. Wrap the body in `http.MaxBytesReader`/`io.LimitReader`, reject oversized bodies and cap the number/length of model IDs before accumulating them; add a regression test with an oversized response.

langwatch-agent (changes requested):
Static external-contributor review found two availability blockers in the new upstream-discovery path; details are inline.

Aryansharma28 (commented):
## goated-review — deep pass on top of the existing rounds

External-contributor PR touching SSRF/security-adjacent code, so I held it to the higher bar: cloned the branch (`1fbe376a`), ran `go build ./...`, `go vet ./...`, `gofmt -l .`, `go test ./... -race` for `services/aigateway`, and `pnpm check:feature-parity` — verifying claims by executing them, not by reading them. I also wrote and ran a standalone proof-of-concept exercising the real `ListModels` code path (not shipped, deleted after running) to test the SSRF fix adversarially rather than take "same policy as Dispatch" at face value. Verdict: **comment** (not overriding the standing `CHANGES_REQUESTED` — two previously-flagged P1s are still live, plus a new one below).

### Severity counts (this pass)
P0: 0 · P1: 1 new (+ 2 pre-existing reconfirmed open, not double-counted) · P2: 3 · P3: 1 · P4: 0 · P5: 2

### Still open from earlier rounds — reconfirmed against current code, not re-raising

- **Partially fixed, see new P1 below**: @langwatch-agent's first-round `[P1] Model discovery bypasses the customer-endpoint SSRF policy` (thread on `bifrost.go:764`). The *specific* case originally flagged (validation skipped entirely) is fixed: `ListModels` now calls `r.validateCredentialEndpoints(ctx, cred)` before every `fetchUpstreamModels` call (`bifrost.go:776`), and `TestListModels_HonorsCustomerEndpointPolicy` passes and genuinely exercises it. But holding this to external-contributor scrutiny rather than taking the fix at face value, I found the underlying protection has a deeper gap that survives the fix — see the new finding below.
- ⚠️ **Still open, confirmed still valid**: `A model-list response is unbounded in memory` (`bifrost.go:836`, raised twice — `3610973661` then escalated to `3620012525`). Read the current code: `json.NewDecoder(resp.Body).Decode(&parsed)` has no `io.LimitReader`/`http.MaxBytesReader` and no cap on array length. Confirmed unaddressed.
- ⚠️ **Still open, confirmed still valid**: `Unbounded fan-out lets a single virtual key exhaust the gateway` (`bifrost.go:769`, `3620012520`). Confirmed: one goroutine per credential with a base URL, no semaphore/worker pool, no per-VK cache, and each one *also* does a synchronous DNS lookup via `validateCredentialEndpoints` before the HTTP call — so the per-request cost is worse than just "N HTTP calls."
- ⚠️ **Still open, confirmed still valid**: CodeRabbit's `Invalid policy regex fails open silently` (`app.go:152-155`, `3610980260`). Extra data point for whoever picks this up: dispatch-time `Matcher.Check` (`adapters/policy/matcher.go:39-40`) *fails closed* (`ErrInternal`) on the identical `regexp.Compile` error — so an invalid deny pattern behaves oppositely at list-time (silently skipped, model stays listed) vs. dispatch-time (whole request errors). Worth aligning the failure mode, not just adding a log line.

### New findings

The two that anchor to a diff line are posted inline. The rest don't anchor to any changed line, so they're here:

**[P1] The SSRF fix doesn't fully close the gap it claims to close — DNS-rebinding TOCTOU.** `validateCredentialEndpoints` (`bifrost.go:776`) validates a credential's `base_url` by resolving its hostname *once* via `customerEndpointPolicy.resolve` (`customer_endpoint_ssrf.go:151`). `fetchUpstreamModels` (`bifrost.go:811`) then makes its own, completely independent HTTP request to that same hostname — Go's `http.Client`/`Transport` re-resolves DNS at dial time via the OS resolver, with no code anywhere pinning the IP that was just validated (confirmed: no `DialContext`/custom `net.Dialer`/`Transport` override exists anywhere in `services/aigateway` or `pkg/`). For an IP-literal `base_url` this is moot (`validateCustomerEndpoint` checks the literal IP directly, never touching `resolve`), but for a real hostname — the realistic self-hosted-server configuration this feature exists for — an attacker who controls authoritative DNS for their own credential's `base_url` domain can answer the validation-time lookup with a public IP (passing the check) and the dial-time lookup with an internal/metadata address (e.g. cloud IMDS) a moment later, exactly the well-known DNS-rebinding SSRF technique. I proved this executably against the unmodified `BifrostRouter.ListModels`/`fetchUpstreamModels` path: a fake `resolve()` that lies ("this host is public") lets `ListModels` return data from a real loopback `httptest` server that `BlockLocalHTTPCalls` is supposed to block — the credential's key would be sent to wherever the real dial lands, model IDs returned to the VK holder. This is **not a regression introduced by this PR** — the same validate-once-then-fetch-separately pattern exists in `Dispatch`/`DispatchStream` today (from #5907), so discovery inherits rather than introduces the gap. Flagging at P1 rather than folding into the pre-existing-architecture bucket because: (a) this is exactly the code this PR adds and the exact claim ("same policy as Dispatch", made in response to the first SSRF finding) that needs to be scrutinized rather than trusted, (b) the threat model is a multi-tenant one — the "attacker" is any project admin who can set their own VK's `base_url`, and the blast radius is LangWatch's own internal network/cloud metadata, not just the customer's, and (c) per project convention, external-contributor security-adjacent code gets the higher bar, not the benefit of the doubt. Suggested fix (applies equally to `Dispatch`, out of scope to require here, but worth a shared follow-up): resolve once, validate the resolved IPs, then dial that pinned IP directly (custom `DialContext` returning the validated address, or a `net.Dialer.Control` hook) instead of re-resolving the hostname at connect time.

**[P2] Public docs for this exact endpoint no longer match what ships.** `docs/ai-gateway/api/models.mdx` (untouched by this PR, last modified 2026-07-08) documents **three** groups — aliases, `models_allowed` (verbatim, including globs), and **"provider-type shortcuts"** (one entry per bound provider, e.g. `openai/*`, `anthropic/*`) — plus a `created` timestamp on every entry. The shipped `App.ListModels` only implements the first two groups; there is no code anywhere producing a provider-shortcut entry (`grep -rn "provider-type shortcut\|openai/\*" services/aigateway --include=*.go` → nothing), and `domain.Model` (`domain/model.go:4-8`) has no timestamp field at all, so `created` can never appear in a response. Issue #5939 (which this PR implements 1:1) never asked for groups 2b/3, so this gap predates this PR — but this is precisely the PR that turns `/v1/models` from an always-null stub into real data, which is what makes the drift consequential: a user following the docs (the CLI-probe use case the docs describe by name — typing `openai/gpt-5-mini` via a provider shortcut) won't see that behavior. Recommend reconciling the docs page (drop the shortcut group + `created` field, or mark them "not yet implemented") in this PR or a fast-follow before customers hit the mismatch. Permalink: https://github.com/langwatch/langwatch/blob/1fbe376afcd72ff65a9ae97b6d61cb90b90bae2f/docs/ai-gateway/api/models.mdx#L34-L38

**[P2] The actual CI for this diff hasn't run yet.** `gh pr checks` shows only `validate title`, CodeRabbit, and the semantic-PR-title check as passed; `go-ci`, `ssrf-conformance`, `gateway-matrix`, `codeql`, and the rest of the required-looking workflow suite are all `ACTION_REQUIRED` (fork/external-contributor gate, never executed) as of head `1fbe376a`. Not the author's fault and not something they can unblock — but worth stating plainly before merge: the SSRF fix and the new concurrency path have only been validated by the author's local `go test ./...`, by the previous automated reviewer's own run, and now by my own out-of-band clone/build/vet/`-race` run — never by this repo's own `ssrf-conformance` or `go-ci` jobs. Given this PR is precisely SSRF- and concurrency-adjacent (see the P1 above), I'd want a maintainer to approve the workflow runs and see them actually go green — not just "not red because it never ran" — before merging.

### Praise / cross-PR observations (P5)

- **Concurrency is done correctly.** `ListModels`'s goroutine-per-credential fan-out writes to a private `perCred[i]` slot per goroutine (no shared map, no append-under-lock), and `go test -race` confirms no data race across all discovery tests. Good discipline given the earlier serial-vs-concurrent back-and-forth in this PR's own review history.
- **Test naming is unusually good.** Tests like `TestListModels_QueriesEndpointsConcurrently` are literally labeled `REPRO bug 1/2/3` in their doc comments, each tied to a specific regression this PR fixes. Genuinely useful for future spelunkers.
- **Unrelated dead code, found while tracing `ListModels` callers**: `domain.Provider` (`services/aigateway/domain/provider.go:97-102`) declares its own `ListModels(ctx context.Context, cred Credential) ([]Model, error)` — singular `Credential`, not `[]Credential` — but nothing implements `domain.Provider` and nothing references it anywhere in the repo. Pre-existing, not part of this diff, not blocking — but an easy trap for the next person who greps "ListModels" and finds two incompatible signatures. Candidate for a follow-up cleanup.

- Whether "verified on a live deployment" (PR description) exercised the SSRF-policy-enforced discovery path specifically, or predates that fix — no access to that deployment.
- Whether this repo's branch protection treats `go-ci`/`ssrf-conformance` as *required* checks (vs. informational) — I can see they haven't run, not whether merge is technically gated on them.
- Real-world exploitability of the DNS-rebinding gap depends on whether project admins configuring `base_url` are a trusted-enough population in this product's threat model to make it lower priority than I'm rating it — that's a product/threat-model call I don't have full visibility into.
- Real-world memory/goroutine impact of the two still-open P1s depends on how many base-URL credentials a typical VK carries in production, which I don't have visibility into.

services/aigateway/app/app.go
```diff
@@ -95,10 +97,89 @@ func (a *App) buildInterceptors() []pipeline.Interceptor {
...
 	}
-	return a.providers.ListModels(ctx, bundle.Credentials)
+
+	for name, alias := range cfg.ModelAliases {
+		add(domain.Model{ID: name, Name: name, ProviderID: alias.ProviderID})
+	}
+
+	if len(cfg.AllowedModels) > 0 {
+		for _, id := range cfg.AllowedModels {
+			add(domain.Model{ID: id, Name: id})
```
Aryansharma28: [P2] Allowlisted (non-alias) models always ship with `owned_by: ""`.

**Why:** `add(domain.Model{ID: id, Name: id})` never sets `ProviderID`. `modelsHandler` (`services/aigateway/adapters/httpapi/router.go:427`) serializes `"owned_by": string(m.ProviderID)`, so every model surfaced through the `models_allowed`-authoritative branch — the exact scenario this PR's own spec calls out ("GET /v1/models reflects effective VK allowlist") — reports an empty `owned_by`. Confirmed: `AllowedModels []string` on `domain.BundleConfig` carries no provider association, so there's no `alias.ProviderID`-equivalent to reach for here; that's presumably why it was left out. `TestRouter_ModelsEndpoint_EmitsOpenAIListShape` only asserts `object == "model"`, so this gap isn't test-covered either.

**Evidence:** `domain.Model{ID: id, Name: id}` (this line) vs. `"owned_by": string(m.ProviderID)` (router.go:427).

**Suggested fix:** No clean one-liner since `AllowedModels` doesn't carry a provider today. Worth a follow-up: either extend `models_allowed` config to optionally carry provider info, cross-reference `bundle.Credentials` when there's exactly one candidate provider, or explicitly document that `owned_by` is best-effort/empty for plain allowlist entries so client authors aren't surprised.

services/aigateway/adapters/providers/models_discovery_test.go
```diff
@@ -0,0 +1,198 @@
+package providers
+
+import (
+	"time"
```
Aryansharma28: [P3] New test file isn't `gofmt`/`goimports`-clean.

**Why:** `"time"` is placed before `"context"` in the same import block; `gofmt -l .` flags this exact file. `.golangci.yml` has `formatters: enable: [gofmt, goimports]`, so the `lint` job in `go-ci.yaml` will fail on this once it actually runs — it's currently `ACTION_REQUIRED` (unrun) on this PR, so nothing has caught it yet.

**Evidence:** `gofmt -d` on this file produces:
```diff
 import (
-	"time"
 	"context"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/langwatch/langwatch/services/aigateway/domain"
 )
```

**Suggested fix:** Run `gofmt -w` (or `goimports -w`) on this file; one-line move.

Aryansharma28:
Hey @thiagolmoraes — thanks for the contribution, left a detailed review above. The main thing I'd appreciate your eyes on is a DNS-rebinding gap in the SSRF validation path (inherited from existing code, not introduced by you here) — happy to pair on it if useful. Also a quick gofmt/goimports pass on the new test file would save a lint failure once CI actually runs on this PR. No urgency otherwise.

langwatch-agent (changes requested):

rogeriochaves (commented):
## Review: solid shape, four defects fixed, plus two control-plane bugs the dogfooding exposed

The assembly pipeline is right and the layering is clean: aliases and allowlist in the app layer, network discovery in the adapter, OpenAI serialisation at the edge. The SSRF pre-flight, the fan-out cap and the response-size cap you added mid-review all landed correctly. I pushed six commits on top rather than just reporting. Everything below is verified, not theorised.

### What I found and fixed

**1. Wildcard allowlist entries were listed as if they were models (correctness, user-visible).**
`models_allowed` holds patterns, not IDs. `modelresolver.matchGlob` supports a trailing `*`, and `RoutingPolicy.modelAllowlist` is documented in the schema as "JSON array of glob patterns". This file's own `Background` uses one: `models_allowed ["gpt-5-mini", "claude-haiku-*", "gemini-2.5-flash"]`. The listing emitted `{"id": "claude-haiku-*"}`, so a model picker offered a literal `claude-haiku-*` and dispatch then forwarded that string upstream as a model no provider has.

Wildcards now expand against discovery, and discovered models are filtered back down by the allowlist so the list never promises a model dispatch would reject. Literal allowlists keep the short-circuit, so your "does not query endpoints when an allowlist is set" scenario stays true. The glob matcher moved to `domain.BundleConfig.AllowsModel` and the resolver now calls it, rather than the two drifting apart.

**2. Discovery followed redirects (security).**
`modelsDiscoveryClient` was a bare `http.Client`, so it followed up to 10 redirects. `validateCustomerEndpoint` vets the configured base URL, not wherever that URL points next, so a policy-passing public endpoint answering `302 Location: http://169.254.169.254/...` was followed, with the credential's `x-api-key` attached. Go only strips `Authorization` across hosts, not custom headers. The response body was then parsed and its `data[].id` values reflected back to the caller, which makes it a readable SSRF, not a blind one.

Redirects are no longer followed at all, and every resolved address is re-checked against the same endpoint policy in `Dialer.ControlContext` immediately before connect. That second part also closes the DNS-rebinding window @Aryansharma28 flagged, for this path specifically: the pre-flight check resolves the host itself, so a name answering public-then-private slipped past it. Both have regression tests that fail on the previous commit.

**3. Nothing was cached (abuse / cost).**
`GET /v1/models` is mounted outside the dispatch interceptor chain, so `RateLimit` never sees it. The semaphore bounds fan-out *within* one call but not across calls, so one cheap inbound request still fanned out to every configured endpoint at whatever rate a client chose. OpenWebUI refreshes its model list on page load. That is the other half of the P1 in the earlier review, and only the semaphore had landed.

Catalogs are now memoised per credential chain for 60s (matching the auth cache's default `ConfigTTL`) and concurrent misses collapse onto one probe. The cache key covers the API key so a rotation invalidates rather than serving a catalog fetched with the old one, and a panicking probe releases its waiters instead of leaving an entry every later caller blocks on.

**4. The response was not fully OpenAI-shaped.**
`created` was missing and `owned_by` could be `""`. Both are required fields of the OpenAI `Model` object: `openai-python` types `created` as `int` and `owned_by` as `str`, so a client got `None` where it expects an int. Models the gateway cannot attribute to a provider are now owned by `"langwatch"`; `created` is `0`, deliberately, since we have no provenance for a creation date and inventing a timestamp that churns per call is worse than an obviously-unknown one.

On your `owned_by` follow-up note: `soleCredentialProviderID` is the right call for the ambiguous case, I only added the non-empty floor underneath it.

### Two control-plane bugs the dogfooding turned up

These are not in your diff, but this feature is not honest without them, so they are fixed in the same PR.

**`GET /api/internal/gateway/config/:vk_id` never shipped the routing policy.** It fetched the VK with `include: { scopes: true }` only. `resolvePolicySideOfBundle` reads `vk.routingPolicy` off whatever the caller included, so with the relation absent it fell through to an empty alias map and empty deny/allow lists. Net effect: **the gateway has never resolved a model alias or enforced a model policy rule in a real deployment.** Your `TestListModels_FiltersDeniedModels` passes because it builds the bundle directly, and `config.materialiser.integration.test.ts` passes because it writes the `include` itself. I added an integration test that drives the real Hono route against real PG, so the route's own query is what is pinned; it fails on the previous commit with `expected {} to deeply equal { chat: 'openai/gpt-5-mini' }`.

**Alias targets kept their `provider/` prefix.** `config_wire.go` mapped `{"chat": "openai/gpt-5-mini"}` to `ModelAlias{Model: "openai/gpt-5-mini"}` with an empty `ProviderID`, so an alias dispatched a model literally named `openai/gpt-5-mini`. Dead code until the fix above made aliases reachable, and this PR advertises aliases to clients: a picker that offers `chat` has to be able to call it. Now split into provider + model, per the `@unimplemented` scenario "alias resolves to configured provider".

Both are called out under `## Deployment Impact` in the description, because enabling a silently-dead deny rule is a live behavior change for anyone who configured one.

### Dogfooding

Real stack, real control plane, real VKs. A self-hosted OpenAI-compatible endpoint serving `qwen3-14b`, `qwen3-32b`, `bge-m3`, `internal-secret-model`; one routing policy with a `chat` alias and `models.deny: ["^internal-.*$"]`; three VKs differing only in `models_allowed`:

```
vk-open      (no allowlist)   -> bge-m3, chat, qwen3-14b, qwen3-32b
vk-narrow    ["qwen3-14b"]    -> chat, qwen3-14b
vk-wildcard  ["qwen3-*"]      -> chat, qwen3-14b, qwen3-32b
```

`internal-secret-model` absent from all three, no literal `qwen3-*` anywhere, stable sort, and the whole set of calls cost the upstream **one** probe. Checked with `curl`, with the `openai` Python SDK 2.30.0 (parses into `SyncPage[Model]`, `created` an `int`, `owned_by` non-empty) and with the `openai` Node SDK.

### Deliberately not changed

- **Hosted providers still contribute nothing.** A VK with only a hosted OpenAI or Anthropic credential and no allowlist still gets `[]`, because those credentials carry no base URL and are skipped. That is what #5939 scoped, and I did not widen it here: probing `api.openai.com` on every listing spends the customer's rate limit, and the control plane already owns a 346-model catalog in `llmModels.json` that would be a better source. Worth a follow-up that ships the catalog on the bundle rather than probing.
- **`RoutingPolicy.modelAllowlist` is still not materialised.** Only the VK's own `config.modelsAllowed` reaches the wire as `models_allowed`. Separate gap, separate change.
- **Alias dispatch end to end** is still `@unimplemented` in the spec. I fixed the wire mapping because this PR surfaces aliases to clients, but did not implement or unmark the dispatch scenarios.
- **@Aryansharma28's gofmt note** was already stale, the file reformats clean. The `deferInLoop` lint in `TestListModels_BoundsFanOutConcurrency` was real, now `t.Cleanup`.
- The discovery code moved from `bifrost.go` into `models_discovery.go` next to its existing test file. Your comments moved **verbatim**, em-dashes and all, so the move stays diffable; only my own added prose follows the no-em-dash rule.
- Nothing here touches #6139 (guardrail check stub / `decision` vs `action`), #6140 (deleted Prometheus package), #6141 (budget accrual), or #6142 (coding-agents onboarding).

### Verification

`go build`, `go vet`, `gofmt -l`, `golangci-lint run` and the full `go test ./...` in `services/aigateway` are clean. `pnpm typecheck` clean. The new control-plane test passes via `pnpm test:integration`. Note that the `pull_request` CI workflows are gated behind fork approval and have never run against these commits; I approved the pending set, but they execute at the author's last SHA. `go-services` and `ssrf-conformance` were green there. `deployment-impact-check` was failing for a missing `## Deployment Impact` heading, which the description update fixes.

services/aigateway/adapters/providers/models_discovery.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	var parsed struct {
+		Data []struct {
+			ID string `json:"id"`
+		} `json:"data"`
+	}
+	limited := http.MaxBytesReader(nil, resp.Body, modelsDiscoveryMaxResponseBytes)
+	if err := json.NewDecoder(limited).Decode(&parsed); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
rogeriochaves: Declining this one. `http.MaxBytesReader` is documented as accepting a nil `ResponseWriter`, and it is the better fit here: it returns an explicit `*http.MaxBytesError` when the cap is hit, so an oversized response is rejected as an oversized response. `io.LimitReader` truncates silently, and the decoder then surfaces the far less useful "unexpected EOF", which reads like a malformed body rather than a too-large one. The behavior is pinned by `TestListModels_RejectsOversizedResponse`.

The other four are addressed in f31e7fbd2: detached probe context plus a short TTL for empty catalogs, hashed cache key, `atomic.Bool` in the two new tests, and `hasWildcards`.

rogeriochaves:
Follow-up on the review above, all pushed.

**CodeRabbit's five comments on my changes.** Four taken in `f31e7fbd2`:

- The probe now runs on a context detached from the caller with its own ceiling. It was running on whichever request missed the cache, so a client hanging up mid-probe cancelled the fetch every concurrent waiter was blocked on, and the resulting empty catalog was cached for the full minute. Good catch, that one was mine.
- An empty catalog gets a 5s TTL instead of 60s, since "no models" is far more often a transient upstream failure than an answer.
- The cache key hashes the credential chain rather than concatenating the API key in plaintext, which is what its own comment already claimed it did.
- `atomic.Bool` for the two flags written by httptest handler goroutines, and `hasWildcards`.

Declined the `io.LimitReader` swap, with reasoning in the thread: `http.MaxBytesReader` is documented as accepting a nil `ResponseWriter` and returns an explicit error when the cap is hit, where `io.LimitReader` truncates silently and the decoder then reports "unexpected EOF", which reads like a malformed body rather than a too-large one.

**Merged `origin/main`** (`e7bf5f757`). The branch was 128 commits behind, which is why `e2e-ci` failed in "Setup ClickHouse" with a Goose migration error while the same workflow is green on main. Two conflicts, both additive on each side, resolved by keeping both: the discovery client and cache fields alongside main's codex fields on `BifrostRouter`, and my alias-parsing test alongside main's trace-project wire tests.

One thing worth flagging from that merge: main moved customer-endpoint address classification into the shared `pkg/ssrf`. The connect-time check I added to the discovery dialer had its own copy of the old `isPublicEndpointIP` / `isAlwaysBlockedEndpointIP` rules, so it now delegates to `endpointAddressError` instead. That is what it should have been anyway: one classifier shared by the pre-flight URL check and the dial, rather than two that can drift.

**Re-verified everything on the merged tree.** `go build`, `go vet`, `gofmt -l`, `golangci-lint run` and the full `go test ./...` in `services/aigateway` clean, `go test -race` on the providers package clean, `pnpm typecheck` clean, and the new control-plane integration test green via `pnpm test:integration`.

Re-ran the dogfooding end to end against the merged build, same result:

```
vk-open      (no allowlist)   -> bge-m3, chat, qwen3-14b, qwen3-32b
vk-narrow    ["qwen3-14b"]    -> chat, qwen3-14b
vk-wildcard  ["qwen3-*"]      -> chat, qwen3-14b, qwen3-32b
```

`internal-secret-model` absent everywhere, no literal `qwen3-*` anywhere, one upstream probe for the whole set, and both the `openai` Python SDK 2.30.0 and the Node SDK parse every response with `created` typed `int` and `owned_by` non-empty.

**CI.** The `pull_request` workflows are gated behind fork approval and had never run against any of these commits. I approved the full set and they are now executing against `e7bf5f757` rather than against a three-day-old base. The two failures that were real are already handled: `go-ci` was failing on the `deferInLoop` gocritic issue in `TestListModels_BoundsFanOutConcurrency` (now `t.Cleanup`), and `deployment-impact-check` was failing for a missing `## Deployment Impact` heading (now in the description).

rogeriochaves:
Filed the two findings from this review as standalone issues, so they exist independently of a models-listing PR:

- #6153 — the config route omitting `routingPolicy`, which is why `model_aliases` and `policy_rules` never reached the data plane. This one shipped: `main` still has `include: { scopes: true }` today. Written as "no code path supplies the relation", since that is what reading the single `materialise` call site and dogfooding locally establishes, not a claim about what any particular deployment observed.
- #6154 — the discovery probe following redirects past the customer-endpoint policy, with `x-api-key` riding along because Go only strips `Authorization` cross-host. Worth recording for the mechanism, though it never released: `ListModels` on `main` is still the stub returning `nil` and `models_discovery.go` does not exist there, so it was introduced and fixed inside this PR.

Both are fixed here.

One correction to my earlier note about the local test database. I said `langwatch_test` was missing `PublicShare` and implied it would silently bite other suites. That was wrong, and the migrations are fine. `20260722070000_rename_public_share_to_share_link` renames `PublicShare` to `ShareLink`, and that migration was not on this branch before I merged `main`. The shared `langwatch_test` database had already been carried forward to `ShareLink` by another worktree, so this branch's older generated client was asking for a table name that no longer existed. Stale branch against a shared test database, not drift and not a missing migration, and it only affects branches predating the rename rather than anything on current `main`.

The `PublicShare` table and `PublicShareResourceTypes` enum I created locally to work around it were stray once the merge brought the rename in, so I dropped both. The shared test database is back to `ShareLink` plus `ShareResourceType` only, and the integration test passes against it on the merged branch.

PR #5942, opened 2026-07-19
fix(evaluations): treat evaluator misconfiguration as a skip, not an error

0xdeafcafe:
## Problem

Production error telemetry is dominated by this line:

```
ERROR langwatch:evaluation-processing:execute-evaluation — Evaluation execution failed
  error: Provider openai is not enabled
```

Two things wrong: it pages us for something that isn't a platform fault, and the customer sees the evaluation reported as `status: "error"` rather than being told what to configure. A disabled provider is an expected state of the product — the customer fixes it in Settings → Model Providers.

Measured over a full log-retention window, evaluator misconfiguration accounts for **94.3% of all evaluation failures**. It is not an edge case; it is the overwhelming majority of what the error channel carries.

## Why it regressed

The modelling already existed: `setupModelEnv` throws `EvaluatorConfigError`, a `HandledError`. Both other consumers already did the right thing — the legacy `runEvaluation.ts` path returns `status: "skipped"`, and this command's own pre-execution gates (monitor not found, Azure Safety absent) emit `"skipped"`. The event-sourcing command's `catch` treated every error identically. Its own comment already documents the intent:

> Config skips (monitor not found, provider not configured) are emitted earlier via their own path and still surface in the UI.

The provider check just throws from *inside* step 5 rather than an earlier gate, so it fell through to the generic handler.

## Change

The catch now discriminates on the repo's own fault classification:

```ts
function isCustomerFixable(error: unknown): error is HandledError {
  return HandledError.isHandled(error) && error.fault === "customer";
}
```

- **Customer-fault** → `status: "skipped"` with the message as `details` (the UI tells them what to change), logged at **info** with the stable `code`.
- **Everything else** — unhandled errors, and `fault: "platform"` / `"provider"` — unchanged: `status: "error"` and an error log, so genuine faults still page us.

The `code` on the info log is the point: it's what a customer-health rule keys off, so nothing pattern-matches log message strings. **langwatch-saas#772** builds that alert on it.

### Also: HTTP 413 gets its own code

Oversized payloads were arriving as `EvaluatorExecutionError("413 {\"message\":\"Request Too Long\"}")` — classified `fault: "platform"`, so error-level with a message the customer can't act on. But nothing is broken on our side and retrying resends the same body. Added `EvaluatorInputTooLargeError` (`fault: "customer"`, httpStatus 413) with remediation tips, raised for 413 specifically. Retry behaviour unchanged — retries were already 5xx-only.

## Review history — five real defects were caught and fixed

**1. The first pass was far too wide.** It discriminated on `HandledError.isHandled`, which is the whole base class. `EvaluatorExecutionError` is *also* a `HandledError`, thrown on langevals timeout / unreachable / 5xx / malformed response, and `evaluation-execution.service.ts` has no try/catch around that call. **A langevals outage would have been reported as a benign "skipped" and logged at info** — silencing exactly the fault class this PR says must keep paging. Fixed, then superseded by the `fault` check, which excludes it by construction (`EvaluatorExecutionError` declares `fault: "platform"`).

**2. The regression test couldn't have caught it.** The original "genuine fault" case threw a plain `Error`, never at risk of being mis-skipped. Added a case throwing `EvaluatorExecutionError`, and **verified it fails against the pre-fix code and passes after** — a real guard, not a test that can't fail.

**3. The 413 metric label contradicted the status it reports** (caught by CodeRabbit). The request-failure path incremented `getEvaluationStatusCounter(evaluatorType, "error")` *before* the 413 branch, so an oversized input incremented the `"error"` label while the command reported the same failure as `status: "skipped"`. Any dashboard or alert keyed on that counter would have read customer-input problems as platform error-rate — the exact thing this PR exists to stop. Moved the `"error"` increment below the branch and label 413 as `"skipped"`.

**4. The guard's docblock asserted the opposite of what the code does.** It described the `fault` check as drift-proof — "stays correct as new error types are added instead of drifting behind a hand-kept list". But `fault` **defaults to `"customer"`** (`handled-error.ts:198`), so the predicate is opt-**out**: any new `HandledError` under `executeForTrace` that forgets to declare a fault lands on the skip path and stops producing error telemetry. An allowlist goes stale loudly; this goes stale silently. The trade is still the right one, but the comment now says so honestly instead of inverted.

**5. `QueryTimeoutError` was already that failure, in the tree today.** `httpStatus: 504`, no declared fault, so it inherited `"customer"`. `getEvaluationsMultiple` doesn't flatten ClickHouse failures the way `getTracesWithSpans` does, so a server-side CH query timeout reaching this catch would have reported as a benign customer skip. Its sibling `ClickHouseUnavailableError` already declares `fault: "platform"`, and the base class docblock already tells subclass authors to annotate 5xx-ish errors. Now it does.

Also fixed: `...error.meta` now spreads *before* the fixed identifiers, so a `meta.traceId` can't clobber the evaluation's `traceId`. `EvaluatorConfigError` declares `fault: "customer"` explicitly rather than inheriting it. The skip log message is neutral ("Customer-fixable evaluator failure") since the branch also catches oversized payloads and non-evaluatable traces, neither of which is a misconfiguration.

## Reviewer note — the permissive default

Following from defect 4: today the fail-open default also makes `TraceNotEvaluatableError` and `EvaluatorNotFoundError` skips. That was a deliberate call here (a non-evaluatable trace is a state, not a fault) — but it deserves a second opinion, because `TraceNotEvaluatableError` fires when the trace read comes back empty, which in practice is most often ingestion or replication lag. If you'd rather those page, the fix belongs on those classes as `fault: "platform"`, not a special case in this file.

Two smaller open questions, both left as-is deliberately:

- **Log level.** `handled-error.ts:8` documents `customer` as "logged at warn", and the two pre-existing skip gates in this same function use `logger.warn`. The new path uses `info`, so an operator filtering at warn+ sees two of three skip classes. The three disagreeing is arguably worse than whichever level wins.
- **Binding accuracy.** The two config-skip scenarios say "provider configured but not enabled" / "never configured", but the tests stub `executeForTrace` to reject with a hand-written `EvaluatorConfigError`. The mechanism (fault → skip) is genuinely exercised; the link from *a disabled provider* to *that error* is not.

## Testing

- Spec: `specs/evaluators/evaluator-config-skips.feature`, 6 scenarios, all bound via `@scenario` (CI's feature-parity gate).
- `executeEvaluation.config-skip.unit.test.ts` — both skip cases, the 413 case, the info-log shape, and the outage case staying an error.
- `langevals.http.client.unit.test.ts` — the 413 path had no coverage in the client's own tests, and the metrics mock returned a fresh spy per call so labels were unobservable. Made the mock a hoisted spy and added: 413 throws `EvaluatorInputTooLargeError` without retrying, 413 counts as `"skipped"` and not `"error"`, and a non-413 failure still counts as `"error"`. The counter test was verified to fail against the pre-fix code.
- 1553 tests / 87 files green across `evaluation-processing`, `langevals`, `traces` and `evaluations` after the fault reclassification.
- `typecheck` **and** `typecheck:tests` both clean. (Note for future PRs: `tsconfig.tsgo.json` excludes test files, so `pnpm typecheck` alone does not cover them — CI runs a second `typecheck:tests` job.)

## Related

- **langwatch-saas#772** — the Grafana alert keyed on this PR's `code`. Should merge *after* this ships, since the field doesn't exist in prod until then.
- Unrelated bug found while mining the logs: `Project.langyEgressAllowlist does not exist in the current database` — real Prisma schema drift in prod, routing through this same log line. Worth filing separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluations with oversized evaluator inputs are now marked as **Skipped** instead of failed, with guidance to shorten the submitted text.
  * Evaluator configuration issues are treated as actionable skips and logged appropriately.
  * Genuine evaluator execution failures, outages, and timeouts continue to be reported as errors.
* **Tests**
  * Added coverage for evaluator configuration, oversized inputs, runtime failures, and logging behavior.


0xdeafcafe:
## Review round 1 — one critical defect found and fixed (5b76c76)

Ran a review pass over the first commit. It found a real bug, which I verified independently before fixing.

### Critical: the skip was far too wide

Discriminating on `HandledError.isHandled(error)` catches the whole base class — not just config errors. `EvaluatorExecutionError` is **also** a `HandledError`, and `langevals.http.client.ts` throws it for:

- request timeout (`:76`)
- `fetch failed` / service unreachable (`:83`)
- non-2xx after retries exhausted (`:107`)
- malformed/empty response body (`:118`)

`evaluation-execution.service.ts` has no try/catch around that call, so these land directly in this handler. Net effect of the first commit: **a langevals outage would have been reported to the customer as a benign "skipped" and logged at info** — silencing exactly the fault class this PR says must keep paging us.

Fixed by replacing the base-class check with an explicit allowlist keyed on `code`, currently just `evaluator_config_error`. Allowlist rather than blocklist, because `HandledError` is a broad base and the safe default is "stays an error unless proven customer-fixable".

That also correctly excludes two the review flagged as swept-in:
- `TraceNotEvaluatableError` — can indicate trace replication / ES lag, i.e. a data-consistency problem, not a settings fix
- `EvaluatorNotFoundError` — usually deploy skew against `AVAILABLE_EVALUATORS`

### Also fixed

- **`meta` spread ordering.** `...error.meta` came *after* the fixed identifiers, so a `meta.traceId` could clobber the evaluation's `traceId`. Not live today (`EvaluatorConfigError` carries no meta) but a live footgun once the allowlist grows — `handledErrorFromHerr` builds `meta.traceId` from the OTel trace id. Meta now spreads first; identifiers always win.
- **The missing regression test.** The original "genuine fault" case threw a plain `Error`, which was never at risk of being mis-skipped — it could not have caught this bug. Added a case throwing `EvaluatorExecutionError`. I confirmed it **fails against the previous blanket check** (exactly 2 failures, nothing else) and passes now, so it's a real guard rather than a test that can't fail.

### Verification

- `vitest run src/server/event-sourcing/pipelines/evaluation-processing/` → **60 passed (11 files)**
- `pnpm typecheck` → clean
- Confirmed the new test fails against the pre-fix code and passes after

### Reviewed and judged fine

Fold projection handles a skipped event with `details` and no score (status is a generic terminal-state string). The late-catch skip correctly lands on the `emitReported` side of the emit-vs-drop distinction at ~line 283 — it carries a customer-facing message, unlike the silent-drop case.

0xdeafcafe:
## CI green — and main had already invented the abstraction I hand-rolled (1d46495)

The branch passed locally but CI failed, for a reason worth recording: **CI builds the PR *merge* commit, not the branch.** My base was 8 commits behind, and in that window main moved `HandledError` out of `~/server/app-layer/handled-error` into the `@langwatch/handled-error` package. The branch typechecked in isolation; the merge with main did not. Merged main in and updated the import.

### The interesting part: `HandledError.fault`

That merge surfaced something better than what I wrote. main introduced exactly the classification this PR was hand-rolling:

```ts
export type HandledErrorFault = "customer" | "platform" | "provider";
```

mirroring `services/aigateway/adapters/httpapi/faults.go`. So the code allowlist is gone, replaced by the repo's own concept:

```ts
function isCustomerFixable(error: unknown): error is HandledError {
  return HandledError.isHandled(error) && error.fault === "customer";
}
```

This is strictly better than my allowlist: it stays correct as new error types are added, instead of drifting behind a list only this file maintains. It still excludes the critical case — `EvaluatorExecutionError` declares `fault: "platform"` — and also correctly holds the line on `fault: "provider"`, i.e. a third-party LLM outage stays an error, since the customer can't act on it either.

**One judgement call worth a reviewer's eye:** `HandledError`'s `fault` **defaults to `"customer"`** when a subclass doesn't declare one. That means an error type whose author never considered fault classification is silently treated as customer-fixable — the same shape of failure as my original bug, one level up. Today that affects `TraceNotEvaluatableError` (can indicate trace-replication lag, arguably platform) and `EvaluatorNotFoundError` (usually deploy skew). Both now skip rather than error. If you agree they're misclassified, the right fix is `fault: "platform"` on those classes rather than special-casing here — happy to do that in this PR or a follow-up.

### Feature-parity gate

CI also flagged 5 unbound scenarios. All are now bound via `@scenario` annotations per the convention in `dev/docs/TESTING_PHILOSOPHY.md`, and retagged `@unit` to match the level the tests actually run at (they're constructor-DI, no containers). `pnpm check:feature-parity` → **5/5 bound**.

### Verification

- **All CI green: 59 pass, 0 fail.**
- 60 tests pass locally; `typecheck` and `typecheck:tests` both clean (I had only run the former before — `tsconfig.tsgo.json` excludes test files, which is why my earlier "typecheck clean" didn't cover the new test).
- Re-confirmed the outage regression test still fails if the fault check is removed, so it remains a real guard.

0xdeafcafe:
@coderabbit review

0xdeafcafe:
> `langevals.http.client.ts` `101-121`: Status counter still labels 413 as `"error"` even though it's now reported downstream as `"skipped"`.

Confirmed against current code and fixed in 570458cb88 — this was a real defect, and precisely the one this PR exists to prevent. The whole point of the change is that customer-fixable failures stop reading as platform error-rate; leaving the counter on `"error"` would have kept oversized inputs spiking exactly the dashboards and alerts the PR is trying to clean up. Applied as suggested: the `"error"` increment moves below the branch, and 413 increments `"skipped"`.

Two things beyond the suggested diff:

The 413 path had **no coverage at all** in the client's own unit tests, and the metrics mock (`getEvaluationStatusCounter: () => ({ inc: vi.fn() })`) returned a fresh spy per call, so label arguments were unobservable by construction — no test in that file could have caught this class of bug. Made the mock a hoisted spy and added three cases: 413 throws `EvaluatorInputTooLargeError` without retrying, 413 counts as `"skipped"` and not `"error"`, and a non-413 failure still counts as `"error"`. I verified the counter test fails against the pre-fix code and passes after, so it's a real guard rather than a test that can't fail.

The earlier inline finding (stale "explicit allowlist" comment) was already addressed before this round — the comment now just points at `isCustomerFixable` rather than restating a rationale that had stopped being true.

The three red checks were pre-existing breakage on `main` at the time this branch's base was cut — a stale `@scenario` annotation from #5916 and a `SidebarSection` assertion, both fixed upstream in #6003 and #6009. Merged `main` in; unrelated to this diff.

0xdeafcafe:
## Audit round — the fail-open direction, and three calls left for you

Ran an independent audit over the diff. The mechanism is sound and the negative cases are real, but the guard's docblock asserts the opposite of what the code does, and one error type is silently downgraded. Fixed in ab3a05d0f2.

### The docblock was backwards

`HandledError.fault` defaults to `"customer"` (`handled-error.ts:198`), so `isCustomerFixable` is **opt-out, not opt-in**. The comment sold it as drift-proof — *"stays correct as new error types are added instead of drifting behind a hand-kept list"* — but the drift direction is the dangerous one. Any new `HandledError` added anywhere under `executeForTrace` that forgets to declare a fault lands on the skip path and stops producing error telemetry. An allowlist goes stale loudly; this goes stale silently. Rewritten to state the default honestly and spell out what a new error type on this path has to do.

The PR body already flagged this as a deliberate trade — the point is that the code comment claimed the opposite, and the comment is what the next person reads.

### `QueryTimeoutError` was exactly that failure, already present

`httpStatus: 504`, no declared fault, so it inherited `"customer"`. `getEvaluationsMultiple` (`evaluation.service.ts:113`) doesn't flatten ClickHouse failures the way `getTracesWithSpans` does, so when a monitor's mappings read the `evaluations` source, a server-side CH query timeout reaches this catch and reports as a benign customer skip. Its sibling `ClickHouseUnavailableError` already declares `fault: "platform"`, and the base class docblock already tells subclass authors to annotate 5xx-ish errors. Now it does.

Also declared `fault: "customer"` explicitly on `EvaluatorConfigError` — odd that the error this feature is named after relied on the implicit default while both errors this PR adds set theirs.

### Smaller

The skip log said *"Evaluator misconfigured"*, which is wrong for the two non-config errors hitting the same branch (oversized payload, non-evaluatable trace) — `code` already distinguishes them, so the message is neutral now. Dropped an unreachable docblock example (`EvaluatorMissingFieldError` is only constructed in the legacy route). Updated the step-5 comment claiming config skips are only ever emitted by an earlier path — they aren't, as of this PR.

---

## Three things I did not change — your call

**1. `TraceNotEvaluatableError` → skip.** You flagged this in the PR body as deliberate ("a state, not a fault"). Worth a second look: it's thrown when the trace read comes back empty, which in practice is most often ingestion or replication lag. Under this PR, a regression that stops traces landing in ClickHouse before evaluation goes dark as a benign skip instead of lighting up the evaluation error rate. That's the precise failure class the PR exists to protect. Same question for `EvaluatorNotFoundError`, which isn't mentioned in the spec or any test.

**2. `logger.info` vs `logger.warn`.** `handled-error.ts:8` documents `customer` as *"logged at warn, watched for spikes"*, and the two pre-existing skip gates in this same function (`:163` monitor-not-found, `:182` Azure-provider-absent) both use `warn`. The new path uses `info`, so an operator filtering the pipeline at warn+ sees two of three skip classes and misses the one this PR was written to surface. You chose `info` explicitly and it's in the spec and tests, so I left it — but the three paths disagreeing seems worse than whichever level wins.

**3. The two config-skip `@scenario` bindings overclaim.** The scenarios' Givens are "provider configured but not enabled" / "never configured", but the test stubs `executeForTrace` to reject with a hand-written `EvaluatorConfigError("Provider openai is not enabled")`. Nothing connects a disabled provider to that class or that message — if `setupModelEnv` changed either, the tests stay green while the scenario silently stops being covered. The mechanism under test (fault → skip) is genuinely exercised and the negative cases would fail a blanket `isHandled` guard, so this is about binding accuracy, not a hole.

Full suite green locally: 1553 tests / 87 files across `evaluation-processing`, `langevals`, `traces`, and `evaluations`.

Issue #5864, opened 2026-07-17
ops(clickhouse): ALTER-added skip indexes were never backfilled (MATERIALIZE INDEX) — ~521 GB/day today, and blocks most of #5810's win


langwatch-agent:
## Summary

Every ClickHouse skip index we have added via `ALTER TABLE ... ADD INDEX` only covers parts written **after** that migration ran. Historical parts carry no index data, so ClickHouse cannot skip them and reads them in full. None of these indexes have been backfilled with `MATERIALIZE INDEX`.

This is not theoretical. For `idx_conversation_id` (migration 00035) it is measurably costing **~521 GB/day of reads today**, and the same gap will silently swallow most of the value of #5810 and #5847 unless a backfill is scheduled with them.

## Proof

`trace_summaries` carries two bloom filters that answer the same kind of lookup. The only difference is how they were created:

- `idx_trace_id` — declared in the original `CREATE TABLE`, so **every** part has it.
- `idx_conversation_id` — added later by `ALTER` in migration 00035, so only **newer** parts have it.

`EXPLAIN INDEXES` for a value that does not exist (so a working bloom should prune everything), on a large tenant (id redacted):

```
idx_trace_id          Parts: 0/180     Granules: 0/180        -- prunes everything
idx_conversation_id   Parts: 71/180    Granules: 130/256      -- prunes about half
```

Splitting the same conversation-id probe by part age isolates the cause exactly:

```
recent parts (last 7d, written AFTER 00035)
  idx_conversation_id   Parts: 0/15     -- index present, prunes everything

old parts (90-300d ago, written BEFORE 00035)
  idx_conversation_id   Parts: 51/92    -- no index data, prunes nothing
```

The index is correct and highly effective where it exists, and inert everywhere it was never backfilled. A `bloom_filter` cannot false-positive at that rate; the old parts simply have nothing to consult.

## Measured cost of the one we can quantify

The thread-to-traces lookup (`Attributes['gen_ai.conversation.id'] IN (...)`, from the eval path in `langwatch-workers`) over the last 24h:

| | |
| --- | --- |
| calls / 24h | 2,452 |
| p50 read | 232 MB |
| p90 read | 350 MB |
| total | **~521 GB/day** |

That is the largest per-call read of any shape in the product, and it is dominated by pre-00035 parts that the bloom cannot skip. Migration 00035's own comment already anticipated this:

[...]

PR #5950, opened 2026-07-20
perf(clickhouse): make the Claude-Code marked-log filter index-eligible

langwatch-agent:
## Evidence

Aggregating the last 24h of app slow-query warns by shape, the Claude-Code marked-log read on `stored_log_records` is the third-largest ClickHouse read cost in the product, and by far the heaviest per call of the high-volume shapes:

| | |
| --- | --- |
| calls / 24h | 21,608 |
| p50 read | **207 MB** |
| p90 read | 224 MB |
| total | **~3,172 GB/day** |

It is already time-bounded and is **not** a cold scan, so the cost is not partition fan-out; it is reading the heavy `Attributes` Map for far more rows than it returns.

(Redacted; this shape was under-reported in my earlier runs because the log stream it arrives on changed — see the note at the end.)

## Suspected cause

Both marked-log reads (`getMarkedClaudeCodeLogsByTrace` and its companion `countMarkedClaudeCodeLogsByTrace`) select the records whose `Attributes` map carries the Claude Code kind key, written as:

```sql
AND Attributes[{kindKey:String}] != ''
```

`stored_log_records` carries an `idx_attr_key` bloom filter over `mapKeys(Attributes)` (migration 00034), which is exactly the index for this question. But ClickHouse only consults it for an explicit **key-existence** test. A subscript-plus-value comparison is not recognised as one, so the index is skipped and the engine reads the `Attributes` Map for every candidate granule in the trace's time window.

`EXPLAIN INDEXES` on live prod, same query with and without an explicit existence test (tenant redacted):

```
A) Attributes[key] != ''                       -- idx_attr_key not consulted at all

B) mapContains(Attributes, key)                -- Name: idx_attr_key
   AND Attributes[key] != ''                      Description: bloom_filter GRANULARITY 4
                                                  Parts: 0/6      <- prunes granules
```

## Proposed fix

Pair the value test with a `mapContains` key-existence test, in a shared constant used by both reads so their WHERE clauses stay identical (the count is documented to match the set the get would return):

```sql
AND mapContains(Attributes, {kindKey:String})
AND Attributes[{kindKey:String}] != ''
```

**This cannot change which rows match.** `mapContains` is implied by the value test: a non-empty value requires the key to be present, so the added conjunct is true for every row the existing predicate keeps. It is a pure index hint. The `!= ''` test is deliberately kept, because `mapContains` alone would also match a record carrying the key with an empty value.

## Expected impact

Granules in the trace's window that hold no Claude-Code-marked records can be skipped instead of having their `Attributes` Map read. The reactor calls this per turn, and each call currently re-reads the trace's marked-log set, so the saving compounds over a long session.

Unlike the two index PRs I have open (#5810, #5847), this one needs **no backfill to take effect**: `idx_attr_key` has existed since migration 00034, so every part inside this read's window (±2 days around the turn, or a retention-bounded fallback) already carries it. See #5864 for why that distinction matters.

This does not address the deeper shape — each turn re-reading the whole trace's marked-log history, which is O(turns x logs) — that would need a per-turn cursor in the reactor and is a larger behavioural change I have not attempted here.

## Test plan

- Unit tests asserting both reads emit `mapContains(...)` alongside the retained `!= ''` test, with the kind key still bound as a parameter.
- Integration tests (testcontainers + real ClickHouse) extended with a record carrying the kind key set to an **empty value** — the exact row `mapContains` alone would wrongly admit. Pins that the get still excludes it and that the count still matches the get (2, not 3).
- Existing integration coverage ("returns exactly the marked logs, ordered by time", and the same via the no-hint fallback) continues to pass unchanged, which is the correctness property that matters for a predicate that must be a no-op.
- All 8 unit + 4 integration tests pass locally against a real ClickHouse container. `tsgo` typecheck clean.

## Note on the measurement

This shape was previously under-reported in my daily analysis. The app pods' logs moved from the `langwatch*` CloudWatch streams to the `default` stream on 2026-07-15, so a stream-name filter I was using silently stopped matching the web-app traffic. Filtering on message content instead surfaced it. Flagging in case anything else keys off those stream names.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filtering of marked Claude Code logs so records with empty kind values are excluded consistently.
  * Aligned log retrieval and count results for more accurate marked-log reporting.

* **Performance**
  * Improved ClickHouse query filtering to better support indexed lookups when searching marked Claude Code logs.

### DIFF
--- a/langwatch/src/server/app-layer/traces/repositories/__tests__/log-record-storage.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/__tests__/log-record-storage.clickhouse.repository.integration.test.ts
@@ -31,12 +31,19 @@ async function insertMarkedLog({
   spanId,
   agoSec,
   marked,
+  kindValue,
 }: {
   tenantId: string;
   traceId: string;
   spanId: string;
   agoSec: number;
   marked: boolean;
+  /**
+   * Explicit value for the kind attribute, so a test can set the key with an
+   * EMPTY value. That row has the key present (mapContains matches) but must
+   * still be excluded by the `!= ''` test.
+   */
+  kindValue?: string;
 }) {
   await ch.command({
     query: `
@@ -47,7 +54,13 @@ async function insertMarkedLog({
       VALUES
         ({pid:String}, {tenantId:String}, {traceId:String}, {spanId:String},
          now64(3) - {agoSec:UInt32}, 9, 'INFO', '{}',
-         ${marked ? `map('${CLAUDE_CODE_KIND_ATTR}', 'model')` : `map()`},
+         ${
+           kindValue !== undefined
+             ? `map('${CLAUDE_CODE_KIND_ATTR}', {kindValue:String})`
+             : marked
+               ? `map('${CLAUDE_CODE_KIND_ATTR}', 'model')`
+               : `map()`
+         },
          map(), 'com.anthropic.claude_code.events', NULL, now64(3), now64(3), 30)
     `,
     query_params: {
@@ -56,6 +69,7 @@ async function insertMarkedLog({
       traceId,
       spanId,
       agoSec,
+      ...(kindValue !== undefined ? { kindValue } : {}),
     },
   });
 }
@@ -103,6 +117,17 @@ describe("getMarkedClaudeCodeLogsByTrace partition hint", () => {
       agoSec: 5,
       marked: false,
     });
+    // The kind key present but EMPTY. The read pairs a mapContains index hint
+    // with the `!= ''` value test; mapContains alone would match this row, so
+    // this pins that the value test still excludes it.
+    await insertMarkedLog({
+      tenantId,
+      traceId,
+      spanId: `${tag}-d`,
+      agoSec: 5,
+      marked: true,
+      kindValue: "",
+    });
   });
 
   describe("when a TimeUnixMs hint around the turn is supplied", () => {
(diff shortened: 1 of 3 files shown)


langwatch-agent:
Closing this as obsolete: the code it optimizes no longer exists on main.

#6105 ("retire the stored_log_records write chain") removed the marked-log read path this PR targeted — both `getMarkedClaudeCodeLogsByTrace` and `countMarkedClaudeCodeLogsByTrace` are gone, along with `CLAUDE_CODE_KIND_ATTR`, and the integration test file this PR extended was deleted by #6005 (coding-agent session aggregate). The repository now exposes only `getLogsByTraceId`, and canonical `log_records` is the sole log path. That is why the PR went `DIRTY`.

Rebasing would mean reapplying an index hint to a query that no longer runs, so there is nothing worth salvaging here. The 3.2 TB/day this measured should disappear with the retirement itself, which is a better outcome than the hint would have been.

I checked whether the underlying pattern transfers, since it is a genuinely useful one: `Attributes[key] != ''` is not recognised by ClickHouse as a key-existence test, so it cannot reach a `mapKeys(Attributes)` bloom. Five call sites still use that form (`activityMonitor.service.ts` x3, `trace-list.clickhouse.repository.ts:936`, `filter-definitions.ts:242`), but all of them read `trace_summaries`, which has no `mapKeys(Attributes)` bloom to unlock. `EXPLAIN INDEXES` confirms adding `mapContains` there changes nothing (`idx_occurred_at` + `idx_tenant_trace` consulted either way). So there is no follow-up PR to open, and I am not re-targeting this one.

Recording the pattern for whenever a `mapKeys` bloom does exist on a table being filtered this way (`stored_spans`, `stored_metric_records` both carry one from migration 00034): pair the value test with an implied `mapContains(Map, key)` to make it index-eligible, and keep the `!= ''` test so a key present with an empty value is still excluded.

PR #5963, opened 2026-07-20
fix(aigateway): keep non-streaming connections warm during slow dispatch

### LINKED ISSUE #4806: Intermittent 524 gateway timeout on requests with large context windows
## Symptom

Requests that include large context windows intermittently return HTTP 524 (gateway timeout) errors. Smaller context requests succeed reliably.

## Reproduction steps (partially known)

- Send a request with a large context window (exact threshold unknown)
- Observe: intermittent 524 responses from the gateway layer

## Status

Under investigation at the infrastructure level. Root cause not yet confirmed.

## Impact

Large-context workloads are unreliable; affected requests fail without retry logic in the client, causing data loss or degraded user experience.

Aryansharma28:
## Summary

Large-context requests intermittently return HTTP 524 (Cloudflare edge idle-connection timeout, ~100s) — traced in #4806 to the non-streaming dispatch path (`/v1/chat/completions`, `/v1/messages`, `/v1/responses`, `/v1/embeddings`, and the Gemini `/v1beta` passthrough): the gateway buffers the *entire* response and writes zero bytes — not even headers — to the client until dispatch fully completes. A completion that legitimately takes longer than the edge's idle window has no way to keep the connection alive, even though every backend timeout (14min upstream, 3600s ingress) is already generous.

- Once a dispatch runs longer than `HeartbeatInterval` (default 45s, ~2x margin under Cloudflare's default), the gateway now writes a periodic RFC 8259 §2 insignificant-whitespace byte and flushes it — resetting the edge's idle timer without corrupting the eventual JSON body (every conformant JSON parser skips leading whitespace).
- Requests that finish inside the interval — the overwhelming majority — are byte-for-byte unaffected.
- **Documented trade-off, now mitigated**: once a heartbeat has fired, the HTTP status is irrevocably committed to 200 (net/http sends it implicitly on the first `Write`). If dispatch later errors, the wire status can no longer become the real 4xx/5xx, but the structured error body still reaches the client correctly — the same trade-off the streaming path already accepts for errors that surface mid-stream. The response now also carries `X-LangWatch-Heartbeat-Active: true` (set before the first heartbeat byte, so it survives regardless of outcome) — a client that only checks `response.status` can use its presence on a 200 as the signal to check the body for an `error` key first.
- `NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS=-1` disables the mechanism entirely as an operator kill switch, no redeploy needed.

Closes #4806.

## Deployment Impact

- **Env vars added**: `NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS` (int, seconds) on the AI Gateway service only. Optional — not required in any environment.
- **Helm values**: none added or changed. `charts/gateway/values.yaml` is untouched; this mirrors the existing `MAX_REQUEST_BODY_BYTES` tunable, which also isn't chart-wired, only env-var-configurable with a Go-side default.
- **Default behavior on a vanilla `helm install` / fresh deployment with no overrides**: the heartbeat mechanism is **on by default** at 45s. Any non-streaming chat/messages/responses/embeddings/Gemini-passthrough request that takes longer than 45s to complete will now emit periodic single-byte keep-alive writes to the client before delivering the real response, and the response will carry `X-LangWatch-Heartbeat-Active: true` — this is the actual behavior change operators should know about. It requires no config to get the fix; no action needed on upgrade.
- **BYOC / self-hosted dataplane impact**: self-hosted gateway operators get this fix automatically on upgrade with zero config. The only operator-visible knob is the new env var, which is a rollback lever (`NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS=-1` fully reverts to pre-fix buffering behavior without a redeploy) rather than something anyone needs to set to get the benefit.
- **Client-visible behavior change**: any client that trusts only `response.status` (not the body, not headers) on a request that runs past 45s and then fails will see status 200 with an error-shaped body instead of the real 4xx/5xx. `X-LangWatch-Heartbeat-Active: true` on the response is the mitigation — its presence is a reliable signal to check the body before trusting the status.

## Test plan

- [x] New spec: `specs/ai-gateway/non-streaming-time-to-first-byte.feature` (4 scenarios, all bound)
- [x] `TestRouter_NonStreaming_NoBytesReachClientWhileProviderIsSlow` — documents the fast-path baseline is untouched, including that the new header is absent
- [x] `TestRouter_NonStreaming_HeartbeatKeepsConnectionWarm` — heartbeat fires mid-dispatch, correct final response + Content-Type + `X-LangWatch-Heartbeat-Active`, leading-whitespace body still parses via real `json.Unmarshal`
- [x] `TestRouter_NonStreaming_HeartbeatThenError_StillDeliversStructuredErrorBody` — proves the committed-200 trade-off delivers the correct structured error body plus the mitigating header
- [x] `TestRouter_NonStreaming_HeartbeatDisabled` — negative interval kill switch, including that the header is never set
- [x] `TestRouter_NonStreaming_HeartbeatFiresAcrossEveryRoute` — table test across all 5 non-streaming routes (chat/messages/responses/embeddings/gemini passthrough), catching any per-handler wiring mistake
- [x] `TestRouter_NonStreaming_ConcurrentSlowRequestsDoNotInterfere` — two concurrent slow requests don't leak state into each other
- [x] `TestRouter_NonStreaming_PanicInDispatchDoesNotCrashProcess` — a panic anywhere in dispatch is recovered and surfaced as a 500, not a process crash (caught in independent review, see PR comments)
- [x] `TestRouter_NonStreaming_HeartbeatKeepsRealSocketWarm` — every test above drives the router through `httptest.ResponseRecorder` (in-memory); this one puts it behind `httptest.NewServer` (a real `net/http.Server` on a real loopback TCP port) and a real `*http.Client`, using `httptrace.ClientTrace.GotFirstResponseByte` to capture actual wall-clock time-to-first-byte over the wire while the provider call is still deliberately held open
- [x] Full suite green under `go test -race -count=5`
- [x] `withHeartbeat` and every `heartbeatWriter` method: 100% statement coverage (`go tool cover -func`), verified directly, not just claimed
- [x] `gofmt`, `go vet` clean
- [x] `pnpm check:feature-parity` — all new scenarios bound
- [x] All required merge checks green, including `sdk-python-complete` — its one failure earlier in this PR's history was a transient `insufficient_quota` on CI's OpenAI key (unrelated to this diff), confirmed by watching it pass on retry with no code changes needed

**Caught during review of this PR's own diff**: the header mitigation above shipped with a real bug on its first pass — it was set whenever heartbeating was simply *enabled* (`interval > 0`, true for every request by default) instead of when a heartbeat actually *fired*, which would have put it on every response and made it useless as a signal. Caught immediately by the test suite, fixed to tie it to the first real tick instead.

## Summary by CodeRabbit

- **New Features**
  - Added configurable heartbeat keep-alives for slow, non-streaming HTTP requests to improve time-to-first-byte.
  - Gateway now periodically sends a whitespace keep-alive while dispatch is running (default 45s; `0` uses default; negative disables), configurable via `NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS`.
  - Applies to chat, messages, responses, embeddings, and Gemini non-streaming routes; includes `X-LangWatch-Heartbeat-Active`.
- **Tests**
  - Expanded coverage for fast/slow behavior, heartbeat-then-error (structured error delivery), disabled mode, concurrent isolation, and dispatch panic resilience; added real-socket verification.
- **Documentation**
  - Added non-streaming TTFB feature scenarios describing expected keep-alive and configuration behavior.

### DIFF
--- a/pkg/config/server.go
+++ b/pkg/config/server.go
@@ -39,16 +39,6 @@ type Server struct {
 	Addr                string `env:"ADDR"`
 	GracefulSeconds     int    `env:"GRACEFUL_SECONDS"`
 	MaxRequestBodyBytes int64  `env:"MAX_REQUEST_BODY_BYTES"`
-	// NonStreamingHeartbeatIntervalSeconds sets how often (in seconds) a
-	// non-streaming response writes a keep-alive byte while dispatch is
-	// still in flight. 0 falls back to DefaultNonStreamingHeartbeatInterval;
-	// negative disables heartbeating entirely. Plain seconds, not a Go
-	// duration string ("45s") — config.Hydrate parses time.Duration fields
-	// as raw nanosecond integers (setField's int64 branch), not via
-	// time.ParseDuration, so "45s" would fail to parse and a correct value
-	// would have to be an opaque nanosecond count. Plain seconds sidesteps
-	// that trap entirely.
-	NonStreamingHeartbeatIntervalSeconds int64 `env:"NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS"`
 }
 
 // ListenAndServe starts the server and handles graceful shutdown on SIGTERM/SIGINT.
--- a/pkg/config/server.go
+++ b/pkg/config/server.go
@@ -22,6 +22,16 @@ const (
 	// headroom over a 10M-token request. Deployments that send multi-hundred-MB
 	// media or run tighter memory should override MAX_REQUEST_BODY_BYTES.
 	DefaultMaxRequestBodyBytes = 128 * 1024 * 1024
+	// DefaultNonStreamingHeartbeatInterval bounds how long a non-streaming
+	// response can go completely silent while a large-context completion
+	// is still in flight. Edge proxies in front of the gateway (Cloudflare's
+	// default is ~100s) kill a connection that receives zero response bytes
+	// within their idle window, even though the origin is healthy and still
+	// working — see https://github.com/langwatch/langwatch/issues/4806.
+	// 45s leaves better than 2x margin under Cloudflare's default while
+	// leaving fast requests (the overwhelming majority) completely
+	// unaffected: only a dispatch slower than this ever emits a heartbeat.
+	DefaultNonStreamingHeartbeatInterval = 45 * time.Second
 )
 
 // Server configures HTTP listen address, graceful shutdown, and request body cap.
@@ -29,6 +39,11 @@ type Server struct {
 	Addr                string `env:"ADDR"`
 	GracefulSeconds     int    `env:"GRACEFUL_SECONDS"`
 	MaxRequestBodyBytes int64  `env:"MAX_REQUEST_BODY_BYTES"`
+	// NonStreamingHeartbeatInterval sets how often a non-streaming response
+	// writes a keep-alive byte while dispatch is still in flight. 0 falls
+	// back to DefaultNonStreamingHeartbeatInterval; negative disables
+	// heartbeating entirely.
+	NonStreamingHeartbeatInterval time.Duration `env:"NON_STREAMING_HEARTBEAT_INTERVAL"`
 }
 
 // ListenAndServe starts the server and handles graceful shutdown on SIGTERM/SIGINT.
--- a/pkg/config/server.go
+++ b/pkg/config/server.go
@@ -39,11 +39,16 @@ type Server struct {
 	Addr                string `env:"ADDR"`
 	GracefulSeconds     int    `env:"GRACEFUL_SECONDS"`
 	MaxRequestBodyBytes int64  `env:"MAX_REQUEST_BODY_BYTES"`
-	// NonStreamingHeartbeatInterval sets how often a non-streaming response
-	// writes a keep-alive byte while dispatch is still in flight. 0 falls
-	// back to DefaultNonStreamingHeartbeatInterval; negative disables
-	// heartbeating entirely.
-	NonStreamingHeartbeatInterval time.Duration `env:"NON_STREAMING_HEARTBEAT_INTERVAL"`
+	// NonStreamingHeartbeatIntervalSeconds sets how often (in seconds) a
+	// non-streaming response writes a keep-alive byte while dispatch is
+	// still in flight. 0 falls back to DefaultNonStreamingHeartbeatInterval;
+	// negative disables heartbeating entirely. Plain seconds, not a Go
+	// duration string ("45s") — config.Hydrate parses time.Duration fields
+	// as raw nanosecond integers (setField's int64 branch), not via
+	// time.ParseDuration, so "45s" would fail to parse and a correct value
+	// would have to be an opaque nanosecond count. Plain seconds sidesteps
+	// that trap entirely.
+	NonStreamingHeartbeatIntervalSeconds int64 `env:"NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS"`
 }
 
 // ListenAndServe starts the server and handles graceful shutdown on SIGTERM/SIGINT.
--- a/services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
+++ b/services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
@@ -60,19 +60,20 @@ func TestRouter_NonStreaming_HeartbeatKeepsRealSocketWarm(t *testing.T) {
 
 	waitOrFatal(t, providerEntered, "provider was never dialed")
 
-	// Generously longer than several 20ms ticks so at least one heartbeat
-	// has definitely reached the real client before we check.
-	time.Sleep(80 * time.Millisecond)
+	// Poll instead of a fixed sleep — a hardcoded wait is either too short
+	// (flakes on a loaded CI runner) or wastes time on a fast one.
+	require.Eventually(t, func() bool {
+		mu.Lock()
+		defer mu.Unlock()
+		return !firstByteAt.IsZero()
+	}, 2*time.Second, 5*time.Millisecond,
+		"a real HTTP client on a real socket should see GotFirstResponseByte from a heartbeat before the still-in-flight provider call returns")
+
 	mu.Lock()
-	gotFirstByte := !firstByteAt.IsZero()
 	elapsed := firstByteAt.Sub(start)
 	mu.Unlock()
-	assert.True(t, gotFirstByte,
-		"a real HTTP client on a real socket should see GotFirstResponseByte from a heartbeat before the still-in-flight provider call returns")
-	if gotFirstByte {
-		assert.Less(t, elapsed, 200*time.Millisecond,
-			"the first byte over the real socket should arrive within a couple heartbeat ticks, not only once the provider call eventually returns")
-	}
+	assert.Less(t, elapsed, 500*time.Millisecond,
+		"the first byte over the real socket should arrive within a couple heartbeat ticks, not only once the provider call eventually returns")
 
 	close(releaseProvider)
 
--- a/services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
+++ b/services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
@@ -1,0 +1,93 @@
+package httpapi
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httptrace"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRouter_NonStreaming_HeartbeatKeepsRealSocketWarm is the one test that
+// isn't in-process: every test in nonstreaming_ttfb_test.go drives the
+// router through httptest.ResponseRecorder, an in-memory buffer that never
+// proves bytes actually cross a real transport. This one puts the router
+// behind httptest.NewServer (a real net/http.Server on a real loopback TCP
+// port) and a real *http.Client, using httptrace to capture the client's
+// actual GotFirstResponseByte — the same signal a real proxy in front of
+// the gateway would be watching to decide whether the connection has gone
+// idle. If chunked-transfer framing or flush semantics only worked inside
+// Go's in-memory recorder abstraction and not over a real socket, this is
+// the test that would catch it.
+func TestRouter_NonStreaming_HeartbeatKeepsRealSocketWarm(t *testing.T) {
+	providerEntered := make(chan struct{})
+	releaseProvider := make(chan struct{})
+	router := newHeartbeatRouter(slowSuccessProvider(providerEntered, releaseProvider), 20*time.Millisecond)
+
+	srv := httptest.NewServer(router)
+	defer srv.Close()
+
+	var mu sync.Mutex
+	var firstByteAt time.Time
+	trace := &httptrace.ClientTrace{
+		GotFirstResponseByte: func() {
+			mu.Lock()
+			firstByteAt = time.Now()
+			mu.Unlock()
+		},
+	}
+	req, err := http.NewRequest(http.MethodPost, srv.URL+"/v1/chat/completions", bytes.NewReader(chatBody()))
+	require.NoError(t, err)
+	req.Header.Set("Authorization", "Bearer vk-lw-test")
+	req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
+
+	start := time.Now()
+	type result struct {
+		resp *http.Response
+		err  error
+	}
+	resultCh := make(chan result, 1)
+	go func() {
+		resp, err := http.DefaultClient.Do(req)
+		resultCh <- result{resp, err}
+	}()
+
+	waitOrFatal(t, providerEntered, "provider was never dialed")
+
+	// Generously longer than several 20ms ticks so at least one heartbeat
+	// has definitely reached the real client before we check.
+	time.Sleep(80 * time.Millisecond)
+	mu.Lock()
+	gotFirstByte := !firstByteAt.IsZero()
+	elapsed := firstByteAt.Sub(start)
+	mu.Unlock()
+	assert.True(t, gotFirstByte,
+		"a real HTTP client on a real socket should see GotFirstResponseByte from a heartbeat before the still-in-flight provider call returns")
+	if gotFirstByte {
+		assert.Less(t, elapsed, 200*time.Millisecond,
+			"the first byte over the real socket should arrive within a couple heartbeat ticks, not only once the provider call eventually returns")
+	}
+
+	close(releaseProvider)
+
+	var res result
+	select {
+	case res = <-resultCh:
+	case <-time.After(2 * time.Second):
+		t.Fatal("real HTTP client never received a response")
+	}
+	require.NoError(t, res.err)
+	defer res.resp.Body.Close()
+
+	body, err := io.ReadAll(res.resp.Body)
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusOK, res.resp.StatusCode)
+	assert.Contains(t, string(body), "choices")
+	assert.Equal(t, "true", res.resp.Header.Get("X-LangWatch-Heartbeat-Active"))
+}
--- a/services/aigateway/adapters/httpapi/nonstreaming_ttfb_test.go
+++ b/services/aigateway/adapters/httpapi/nonstreaming_ttfb_test.go
@@ -1,0 +1,593 @@
+package httpapi
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/langwatch/langwatch/pkg/health"
+	"github.com/langwatch/langwatch/pkg/herr"
+	"github.com/langwatch/langwatch/services/aigateway/app"
+	"github.com/langwatch/langwatch/services/aigateway/domain"
+)
+
+// ttfbSpyWriter records whether anything has actually been flushed to the
+// transport — WriteHeader or Write. Go's net/http buffers header-map
+// mutations (w.Header().Set(...)) client-side and does not put a single
+// byte on the wire until the first real WriteHeader/Write call, so
+// inspecting httptest.ResponseRecorder's header map or its Code field
+// (pre-seeded to 200) cannot distinguish "nothing sent yet" from "sent."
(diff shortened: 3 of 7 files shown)


Aryansharma28:
Ran an independent adversarial review of this PR before merge. Findings and outcome:

1. **Confirmed, blocking (fixed in 3d16b40)**: `withHeartbeat`'s background dispatch goroutine ran outside `httpmiddleware.Recover()`'s reach — a panic anywhere in the dispatch call graph (provider adapters, pipeline/retry) would crash the whole gateway process instead of returning a 500, strictly worse than the 524 this PR fixes. The reviewer verified this empirically with a real panic. Fixed by recovering the panic in `withHeartbeat` itself and surfacing the same `internal_error` 500 `Recover()` produces for a synchronous panic. Added `TestRouter_NonStreaming_PanicInDispatchDoesNotCrashProcess` as the regression test — a real crash there would fail the whole test binary, not just one assertion.
2. **Plausible, documented trade-off (unchanged, flagging for explicit sign-off)**: once a heartbeat has fired, HTTP status is irrevocably committed to 200 (net/http sends it implicitly on the first `Write`). A dispatch that runs past the heartbeat interval and then errors delivers status 200 with an error-shaped body. This is inherent to the "write body bytes before the final status is known" approach and mirrors the trade-off the streaming path already accepts for mid-stream errors — but any client that trusts `response.status` alone (most OpenAI/Anthropic-compatible SDKs) will treat a failed, slow request as successful. Worth explicit product/API sign-off since #4806 traffic (large-context, slow) is exactly the population most likely to hit this.
3. **Pre-existing, out of scope (sidestepped in 3d16b40)**: `pkg/config`'s `Hydrate` parses `time.Duration` fields as raw nanosecond integers, not via `time.ParseDuration` — so `NON_STREAMING_HEARTBEAT_INTERVAL=45s` would never have actually parsed. Switched the new field to plain seconds (`NonStreamingHeartbeatIntervalSeconds`, env `NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS`) rather than touching the shared parser, which has broader blast radius (other existing duration fields inherit the same gap and may warrant a follow-up).

Full suite green under `go test -race`, gofmt/vet clean, feature-parity green after both commits.

Aryansharma28 (commented):
## goated-review

**Verdict: COMMENT** — P0: 0 | P1: 0 | P2: 0 open (1 found, fixed) | P3: 0 | P4: 0 | P5: 4

Ran a full pass on this PR: read the room (existing reviews, CI, CodeRabbit), sliced the diff, adversarially verified every finding, and — since I have write access to this branch — fixed what was actually fixable rather than just flagging it. Summary of the full audit trail below; every claim is backed by something I actually ran, not just read.

### P0 found and fixed (not open — documenting for the record)

An earlier independent-review pass on this same PR (see the comment below from this session) found that `withHeartbeat`'s background dispatch goroutine ran outside `httpmiddleware.Recover()`'s reach — a panic anywhere in dispatch would have crashed the whole gateway process instead of returning a 500, strictly worse than the 524 this PR fixes. Fixed in `3d16b40` with an explicit recover + `TestRouter_NonStreaming_PanicInDispatchDoesNotCrashProcess`, which fails the whole test binary (not just an assertion) if it regresses.

### P2 found and fixed during this pass

Once a heartbeat fires, HTTP status is irrevocably committed to 200 (net/http sends it implicitly on the first `Write`) — if dispatch later errors, the wire status can't become the real 4xx/5xx. Inherent to writing body bytes before the final status is known; not fixable without abandoning the mechanism. Added `X-LangWatch-Heartbeat-Active: true`, set before the first heartbeat byte so it survives regardless of outcome — a client that only checks `response.status` can use its presence on a 200 as the signal to check the body for an `error` key first.

**Caught a real bug shipping this fix**: the header was originally set whenever heartbeating was simply *enabled* (`interval > 0`, true for every request by default) rather than when a heartbeat actually *fired* — every response would have carried it, making it useless as a signal. `TestRouter_NonStreaming_NoBytesReachClientWhileProviderIsSlow`'s new assertion caught this on the first run, deterministically, across every run. Fixed by tying it to the first real ticker fire (`heartbeatFired` flag, router.go:674).

### P5 — verified, not defects

1. **Test coverage**: `withHeartbeat` and all 4 `heartbeatWriter` methods (`WriteHeader`/`Write`/`Flush`/`Unwrap`) are at 100.0% statement coverage per `go tool cover -func` — measured, not claimed. Plus a dedicated panic-recovery regression test, a dedicated concurrent-requests-don't-interfere test, and a table test hitting all 5 non-streaming routes individually (chat/messages/responses/embeddings/gemini-passthrough) so a copy-paste mistake at any one `withHeartbeat` call site can't hide behind the others passing.
2. **CodeRabbit's "does a hung provider call now hang forever instead of getting killed by the CDN" concern** — verified against `services/aigateway/adapters/providers/bifrost.go:789,797-798`: `ProviderRequestTimeoutSeconds = 14 * 60` already bounds `DefaultRequestTimeoutInSeconds`, `StreamIdleTimeoutInSeconds`, and the direct HTTP client's `Timeout`. Not a new gap — worst case shifts from an accidental ~100s CDN kill to the pre-existing, deliberate 14-minute ceiling documented in `provider-request-timeout.feature`.
3. **CodeRabbit's `time.ParseDuration` suggestion on `pkg/config/server.go`** — addressed differently: `pkg/config`'s `Hydrate` parses `time.Duration` fields as raw nanosecond integers (not via `time.ParseDuration`), so the originally-shipped `NON_STREAMING_HEARTBEAT_INTERVAL=45s` would never have actually parsed. Rather than touching the shared parser (wider blast radius — other existing duration fields inherit the same gap), switched this field to plain seconds (`NonStreamingHeartbeatIntervalSeconds`) in `3d16b40`, sidestepping the bug entirely.
4. **CodeRabbit's 300-source-line file-size flag on the test file** — verified (was ~462 non-comment lines) and fixed in `b0dbf7553`: extracted `newHeartbeatRouter`/`serveAsync`/`waitOrFatal`/`slowSuccessProvider` so each test states only what's actually different about it. 297 lines at last count, one small trim short of the exact threshold after adding a few more assertions for the header fix — a soft/advisory limit, not a CI gate, and not worth trading test coverage for 3 lines.

### CI — one real gate fixed, two unrelated pre-existing failures

- `check` (deployment-impact-check) was correctly failing: this PR changes default runtime response behavior and had no `## Deployment Impact` section. Fixed by adding one to the PR description (env vars, helm impact — none, default-on behavior, BYOC impact, client-visible behavior change).
- `sdk-python-complete`/`test` and `evaluate` (auto-approve gate) are both failing on `litellm.exceptions.RateLimitError` / `insufficient_quota` from the CI's OpenAI API key — unrelated to any file in this diff, pure CI infra. Flagging clearly so it isn't mistaken for something this PR broke.

- Whether Cloudflare's production zone for the real gateway deployment actually uses the ~100s default assumed throughout (vs. a custom Enterprise timeout) — I don't have access to the production CF zone config from this repo.
- Real-world client behavior on `X-LangWatch-Heartbeat-Active`: this is a new contract with zero existing consumers today. It's a reasonable, additive mitigation, but whether any LangWatch-authored SDK should actually start checking for it is a product/SDK-team call, not something I can verify from this diff.

services/aigateway/adapters/httpapi/router.go
```diff
@@ -543,6 +565,139 @@ func (l *lazyPooledBody) Read(p []byte) (n int, err error) {
...
+// negative disables heartbeating entirely.
+//
+// dispatch runs on a background goroutine so the select loop below stays
+// free to write heartbeats while it's in flight. That goroutine is outside
+// httpmiddleware.Recover()'s reach — Recover's defer/recover only guards
+// the goroutine that calls ServeHTTP, not one spawned from inside a
+// handler — so a panic in dispatch is recovered here explicitly and turned
+// into the same internal_error 500 Recover() would have produced for a
+// synchronous panic, instead of crashing the whole process.
+func withHeartbeat[T any](ctx context.Context, w http.ResponseWriter, interval time.Duration, dispatch func() (T, error)) (T, http.ResponseWriter, error) {
```
Aryansharma28, line 637: [P5] Strong test posture for a concurrency-sensitive change.

**Why:** This function and every method on `heartbeatWriter` (`WriteHeader`/`Write`/`Flush`/`Unwrap`) sit at 100.0% statement coverage per `go tool cover -func` — verified directly against the coverage profile, not assumed from the test count. Combined with a dedicated panic-recovery regression test and a concurrent-requests-don't-interfere test, this is meaningfully more thorough than the diff size would suggest.

**Evidence:** `go test ./adapters/httpapi/... -run TestRouter -coverprofile=... && go tool cover -func=...` shows `withHeartbeat 100.0%`, `WriteHeader 100.0%`, `Flush 100.0%`, `Unwrap 100.0%`.

services/aigateway/adapters/httpapi/nonstreaming_ttfb_test.go
```diff
@@ -0,0 +1,485 @@
...
+// regression test for a bug caught in review of the heartbeat fix itself:
+// dispatch runs on a background goroutine (withHeartbeat) so the heartbeat
+// loop can keep writing to the client while it's in flight. That goroutine
+// sits outside httpmiddleware.Recover()'s reach — Recover's defer/recover
+// is scoped to the goroutine that calls ServeHTTP, not one spawned inside a
+// handler — so an unrecovered panic there would take down the whole
+// process instead of producing a 500 for the one bad request. If this test
+// method itself doesn't complete (the test binary crashes instead of
+// reporting pass/fail), that IS the regression.
+func TestRouter_NonStreaming_PanicInDispatchDoesNotCrashProcess(t *testing.T) {
```
Aryansharma28, line 455: [P5] Good regression-test shape for a panic-recovery fix.

**Why:** This test doesn't assert on a caught exception or a mock — it lets a real panic run through the real router and real `Recover()`-adjacent path. If the fix regresses (background dispatch goroutine panics unrecovered again), this test *crashes the test binary* rather than just failing an assertion, which is exactly the right failure mode to prove for this class of bug — a passing assertion here would be meaningless if the panic could still take the process down.

services/aigateway/adapters/httpapi/router.go
```diff
@@ -543,6 +565,139 @@ func (l *lazyPooledBody) Read(p []byte) (n int, err error) {
...
+	hw := &heartbeatWriter{ResponseWriter: w}
+
+	type outcome struct {
+		val T
+		err error
+	}
+	resultCh := make(chan outcome, 1)
+	go func() {
+		defer func() {
```
0xdeafcafe, line 646: check out clog.handlpanic, could use here

0xdeafcafe (dismissed):

pkg/config/server.go
```diff
@@ -22,13 +22,28 @@ const (
...
 // Server configures HTTP listen address, graceful shutdown, and request body cap.
 type Server struct {
 	Addr                string `env:"ADDR"`
 	GracefulSeconds     int    `env:"GRACEFUL_SECONDS"`
 	MaxRequestBodyBytes int64  `env:"MAX_REQUEST_BODY_BYTES"`
+	// NonStreamingHeartbeatInterval sets how often a non-streaming response
+	// writes a keep-alive byte while dispatch is still in flight. 0 falls
+	// back to DefaultNonStreamingHeartbeatInterval; negative disables
+	// heartbeating entirely.
+	NonStreamingHeartbeatInterval time.Duration `env:"NON_STREAMING_HEARTBEAT_INTERVAL"`
```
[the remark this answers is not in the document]
Aryansharma28: Addressed differently rather than switching to `time.ParseDuration`: `pkg/config`'s `Hydrate`/`setField` routes every `time.Duration` field through the generic `reflect.Int64` branch (raw nanosecond integers via `strconv.ParseInt`), which is a pre-existing gap shared by other fields in this package (e.g. `AuthCacheConfig.SoftBump`/`HardGrace`/`ConfigTTL`, whose `.env.example` entries document `5m`/`6h` formats that would also fail to parse today). Switching just this one field's parsing path to `time.ParseDuration` would fix it locally but leave that inconsistency in the shared parser for everyone else, and touching the shared `setField` function has a much wider blast radius than this PR's scope.\n\nInstead, in commit 3d16b40 I changed the field itself to plain seconds — `NonStreamingHeartbeatIntervalSeconds int64`, env `NON_STREAMING_HEARTBEAT_INTERVAL_SECONDS` — which parses correctly through the existing (already-correct-for-plain-integers) int64 path with zero risk to any other field. Converted to a `time.Duration` at the one call site in `serve.go` that needs it. Resolving this thread; the underlying `pkg/config` duration-parsing gap is real but pre-existing and out of scope here — worth a follow-up covering all the affected fields at once rather than a partial fix.

services/aigateway/adapters/httpapi/router.go
Aryansharma28: Good catch to look at — but `clog.HandlePanic` doesn't quite fit this call site, and I want to explain why rather than just apply it:

```go
func HandlePanic(ctx context.Context, propagate bool) {
	v := recover()
	if v == nil {
		return
	}
	LogPanic(ctx, v)
	if propagate {
		panic(v)
	}
}
```

It calls `recover()` internally and either re-panics or returns `void` — it never gives the caller the recovered value back. If I swapped my explicit `recover()` block for `defer clog.HandlePanic(ctx, false)` here, a panic would get logged correctly, but nothing would ever be sent on `resultCh` for that request — the `withHeartbeat` select loop above would block on it forever. That trades a process-wide crash for a per-request hang instead of a clean 500, which is worse in a quieter way (an unbounded goroutine + open client connection instead of something that at least pages someone).

What I actually need is to recover the value myself so I can construct an `outcome{err: ...}` and push it onto `resultCh` — which is what's here. I *am* reusing the shared logging path though: `clog.LogPanic(ctx, v)` on line 648 is the exact same call `HandlePanic` makes internally (stack trace, span recording, the `process.goroutine.panics_recovered` metric) — so the observability side matches what `HandlePanic` would have produced, it's just composed manually because this path needs to *return* something, not just log-and-swallow.

Happy to rename/restructure if there's a clearer way to signal that intentional divergence — let me know if this reasoning doesn't hold up.

services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
```diff
@@ -0,0 +1,93 @@
...
+	mu.Lock()
+	gotFirstByte := !firstByteAt.IsZero()
+	elapsed := firstByteAt.Sub(start)
+	mu.Unlock()
+	assert.True(t, gotFirstByte,
+		"a real HTTP client on a real socket should see GotFirstResponseByte from a heartbeat before the still-in-flight provider call returns")
+	if gotFirstByte {
+		assert.Less(t, elapsed, 200*time.Millisecond,
+			"the first byte over the real socket should arrive within a couple heartbeat ticks, not only once the provider call eventually returns")
+	}
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in 0a019ed — swapped the fixed `time.Sleep(80ms)` + manual check for `require.Eventually` polling at 5ms intervals up to a 2s bound, matching this file's own convention. Ran it 5x under `go test -race -count=5`, all pass, and it's actually faster now on a healthy runner too (0.03-0.08s vs. always paying the full 80ms).

services/aigateway/config.go
```diff
@@ -64,11 +78,11 @@ func defaultConfig() Config {
...
 		Server: config.Server{
-			Addr:                                 ":5563",
-			GracefulSeconds:                      10,
-			MaxRequestBodyBytes:                  config.DefaultMaxRequestBodyBytes,
-			NonStreamingHeartbeatIntervalSeconds: int64(config.DefaultNonStreamingHeartbeatInterval / time.Second),
+			Addr:                ":5563",
+			GracefulSeconds:     10,
+			MaxRequestBodyBytes: config.DefaultMaxRequestBodyBytes,
 		},
+		NonStreamingHeartbeatIntervalSeconds: int64(config.DefaultNonStreamingHeartbeatInterval / time.Second),
```
0xdeafcafe, line 85: sanity check that the heartbeat isn't longer than the `GracefulSeconds`, warn if so? maybe unneeded, but would be nice to handle i think?

0xdeafcafe (approved):
nice on the `require.Eventually`, speedy

services/aigateway/adapters/httpapi/router.go
0xdeafcafe: didn't think so but wanted to check, thanks 👌🏻

services/aigateway/adapters/httpapi/nonstreaming_ttfb_realsocket_test.go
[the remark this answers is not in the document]
Aryansharma28: Good catch — fixed in 0a019ed3: swapped the fixed `time.Sleep(80ms)` + manual check for `require.Eventually`, matching your suggested diff. Also fixes something the fixed sleep didn't guarantee: on a slow/loaded CI runner, 80ms might not be enough for even one 20ms tick to actually land, so the old version could flake in exactly the direction that matters (false failure on a healthy fix). Polling removes that dependency on wall-clock timing entirely — and it runs faster on a healthy runner too, since it stops as soon as the byte arrives instead of always waiting the full window.

services/aigateway/config.go
Aryansharma28: Good instinct, but I want to trace this precisely before replying, because the two numbers being compared here don't actually decide the outcome — thanks for the nudge to trace it, I was wrong about where shutdown logic even lives for this service until I checked.

I assumed `pkg/config.Server.ListenAndServe()` governed this (it has `GracefulSeconds` and a `Shutdown` call, looked like the obvious place) — but aigateway doesn't call that method at all. It uses `pkg/lifecycle.Group` instead, which does: mark `/readyz` draining → wait `drainDelay` (3s) → `srv.Shutdown(ctx)` with `ctx` deadline = `GracefulSeconds` (10s) → if that deadline passes, `Shutdown` returns `context.DeadlineExceeded` and the process exits shortly after (see `pkg/lifecycle/group.go` and `service.go`).

The part that matters: `http.Server.Shutdown()` does **not** forcibly kill in-flight request goroutines when its deadline expires — it just returns an error. What actually terminates a still-running request is the *process exiting*, which happens regardless of whether a heartbeat ever fired. So comparing `HeartbeatInterval` to `GracefulSeconds` doesn't predict this failure — a request that outlives `GracefulSeconds` during a rolling deploy already gets killed today, heartbeat or not, and always has.

The real (much bigger, entirely pre-existing) mismatch is `GracefulSeconds` (10s default) vs. the actual maximum request duration the gateway already permits: `ProviderRequestTimeoutSeconds` = 14 minutes (`services/aigateway/adapters/providers/bifrost.go`). *That* gap exists with or without heartbeat, with or without this PR — a legitimately long-running completion in flight during any rolling deploy today can already get killed well before it finishes. Worth a look, but it's a shutdown-lifecycle question spanning the whole gateway, not something scoped to this PR's non-streaming heartbeat fix, and a warning keyed to `HeartbeatInterval` specifically wouldn't catch it. Happy to open that as its own thread if useful — just didn't want to bolt an inaccurate check onto this PR.

PR #5968, opened 2026-07-20
feat(skills): evolve the insight skills into a diagnosis-to-improvement ladder

rogeriochaves:
# feat(skills): evolve the insight skills into a diagnosis-to-improvement ladder

The Recipes section of the [skills directory](https://langwatch.ai/docs/skills/directory) previously led with an oddly-named starred entry (the improve-setup audit) and a thin analytics wrapper. It now presents a coherent ladder from evidence to tested improvements.

## The three skills

**`/agent-improve` (new, the starred top entry: "What should I do next to improve my agent?")**
A hypothesis engine with a teaching stance. It gathers evidence from production traces and analytics first, forms explicit hypothesis chains (observation with linked example traces, suspected cause, proposed test, proposed fix, expected effect), explains each one until the user can restate the reasoning, and only then executes: scenario tests reproducing real production failures, prompt and code changes shipped as reviewable PRs, evaluators and monitors that turn one-off findings into permanent production signals, and experiments that settle open questions with numbers.

**`/agent-performance` (replaces `/analytics`: "How is my agent performing?")**
A full production diagnostician built on the externally available LangWatch APIs: baselines every vital sign with `langwatch analytics query` presets and previous-period comparison, slices with group-bys to locate where numbers come from, bulk-exports traces as JSONL and mines them locally for failure patterns, dissatisfied users, token cost hotspots, edge cases, behavior changes between time windows, and outliers, reads representative traces in full, then writes a self-contained HTML report where every finding links to real example traces. It closes by handing off to `/agent-improve` (or telling the user how to install it).

**`/agent-best-practices` (renamed from `recipes/improve-setup`: "Where can I improve our agent development best practices?")**
The audit recipe reframed around an explicit best-practices checklist (observability, prompt management, testing, evaluation, production monitoring, iteration loop), fixing the highest-impact gaps first and handing over to the production-insight skills when real traffic exists.

## Everything that follows the rename

- `FEATURE_SKILLS` gains `agent-performance` and `agent-improve`; `analytics` is gone (`agent-performance` replaces it in the published set, on the directory page, and in the onboarding screen)
- Langy: routing table rows updated (`AGENTS.md`), native skill set regenerated (15 skills), `langySkills.generated.json` regenerated
- Onboarding UI: the analytics card became agent-performance with a rewritten condensed prompt (`PROMPT_AGENT_PERFORMANCE`)
- PM & Domain Experts page: the "How is my agent performing?" platform prompt now compiles from agent-performance
- Docs directory pages regenerated from the manifest with the new ordering; specs updated (`specs/skills/agent-insight-skills.feature` plus stale references in the two existing feature files)

## Scenario tests (sample projects, real runs)

Three new scenario tests under `skills/_tests/`, following the local-dogfood pattern (real `claude` subprocess, real LangWatch project, gpt-5-mini judge):

- `agent-best-practices.scenario.test.ts`: python-openai fixture with hardcoded prompts and no tests; judge requires a codebase + CLI audit, project-specific gaps, and prioritized fixes. **PASSED locally (13 min)**: the agent versioned the fixture's hardcoded prompt via `langwatch prompt init` and wired `main.py` to load it, exactly the intended low-hanging-fruit behavior.
- `agent-performance.scenario.test.ts`: runs against the live skill-tests project traces; hard filesystem assertions on the HTML report (exists, holds trace links) plus judge criteria (analytics + trace mining, numbers not JSON dumps, example traces per finding, read-only, recommends agent-improve).
- `agent-improve.scenario.test.ts`: two-turn script; judge requires evidence gathering BEFORE proposals, hypothesis chains with explained reasoning, actionable artifacts, and execution only after the user picks one.

The old `analytics.scenario.test.ts` is superseded by the agent-performance test. Run results below in the QA section.

## QA evidence

Directory page with the new ladder, all three expanded (server-rendered, verified via curl with scripts stripped: zero stale analytics or improve-setup references, 14 accordions):

recipes trio light

Dark mode:

improvement dark

PM & Domain Experts page, platform prompt still working (now compiled from agent-performance):

pms platform prompt

Scenario runs (local, real claude subprocess, live LangWatch project):

- agent-best-practices: **PASSED** (787s). Observed behavior in the fixture: ran the CLI audit, then versioned the hardcoded prompt (`langwatch prompt init`, `prompts/` + lockfile) and rewired `main.py` to load it.
- agent-performance: **PASSED** (786s), including the hard filesystem assertions. The run produced a real 20KB `agent-performance-report.html` with 37 trace links, and its findings are genuinely correct about the test project: it identified that 99% of traffic is the test harness, root-caused the Jul 18 error spike to broken evaluation infrastructure (invalid OpenAI key, SDK model incompatibility) rather than agent behavior, flagged today's OpenAI quota exhaustion, listed the real judge failures (fabricated weather data, tweet-length violations, emoji in voice output), and explained a 3x latency change through context ballooning. Report excerpt below.
- agent-improve: **PASSED** (392s) with the final judge criteria including the minimal-execution constraint. The agent swept evidence through the CLI first, presented ranked hypotheses with observation, cause, and expected effect, waited for the user's pick, then executed as a reviewable change set: a git history in the sample project with a baseline commit, a `reasoning_effort=minimal` change targeting latency on tweet-length replies, and LangWatch instrumentation with a managed prompt, without booting the app.
- agent-improve, re-run after the `/agent-improve` rename: the first re-run FAILED honestly (the agent executed the top hypothesis in turn 1 before the user chose; the judge caught it on both gate criteria). Root cause was the skill's own no-ask escape hatch reading as if it applied to any autonomous agent, so the gate is now explicit: present hypotheses, stop, and execute nothing until the user answers, with the exception scoped to environments where asking is truly impossible. Re-run after the fix: **PASSED** (262s, 6/6 criteria, verified against the platform record).

The report the skill generated during its scenario run, verbatim:

generated report top

generated report mid

Note on downloads: the Download SKILL.md buttons for the three new paths resolve after merge, when `skills-publish.yml` syncs the `langwatch/skills` repo (trigger covers `skills/*/SKILL.mdx` and `skills/recipes/*/SKILL.mdx`, both touched here). Verified the generated URLs point at the correct published layout; the consistency tests pin them to `listPublishedSkills()`.

## Bonus: main's red integration shard, root-caused (fix landed via #5970)

`test-integration (6)` went red on main at 12:36 UTC today with no causal commit: the `trace-analytics-rollup` fixtures used a fixed `2026-06-01` bucket date while the table TTLs rows 49 days (`PLATFORM_DEFAULT_RETENTION_DAYS`) after `BucketStart`. Jun 1 + 49 days = Jul 20, 12:00 UTC: from that minute on, every inserted fixture row was born expired and ClickHouse dropped it at merge time, nondeterministically. The sibling `trace-analytics` fixtures (2026-06-15) would have detonated identically on Aug 3.

The same diagnosis was fixed on main in parallel by #5970 (relative "yesterday" anchors, covering two more analytics test files) while this branch carried its own anchor fix. This branch has merged main and adopted the #5970 anchors; what remains here is the complementary defense: the rollup read now retries briefly until async-inserted fixture rows become visible, closing a separate first-read flake observed both locally and in CI. Verified 10/10 across three consecutive local runs, and re-verified after the merge.

## Deployment Impact

None on infrastructure: no env vars added or changed, no helm values added or changed, and a vanilla `helm install` with no overrides behaves exactly as before. The only deployment-adjacent change is content baked into the Langy agent image at build time: the routing table (`services/langyagent/internal/assets/AGENTS.md`) and the compiled native skill set it overlays (`skills/_compiled/native/`) now carry `agent-performance`, `agent-improve`, and `agent-best-practices` instead of `analytics` and `improve-setup`. The next langyagent image build picks this up automatically; BYOC dataplanes are unaffected beyond receiving the refreshed skill catalog with their next image upgrade. The public `langwatch/skills` repo re-syncs on merge via the existing `skills-publish.yml` workflow.


skills/_tests/agent-best-practices.scenario.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    async () => {
+      const tempFolder = fs.mkdtempSync(
+        path.join(os.tmpdir(), "langwatch-skill-agent-best-practices-")
+      );
+
+      // A minimal agent codebase with no scenarios, no versioned prompts, and
+      // no evaluation setup: plenty of best-practice gaps for the audit to find.
+      execSync(
+        `cp -r ${path.resolve(__dirname, "fixtures/python-openai")}/* ${tempFolder}/`
+      );
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 96f4829d9: both new tests now use fs.cpSync({ recursive: true }) instead of shelling out, and the same pre-existing pattern in recipes.scenario.test.ts got converted too so it does not get copied forward again. The CodeQL alerts on the two new files are resolved by the same commit.

skills/agent-improve/SKILL.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+For each significant finding, build an explicit hypothesis chain and present it to the user:
+
+- **Observation**: what the traces show, with linked examples ("11% of conversations rephrase the same question twice, examples: trace A, trace B")
+- **Hypothesis**: the suspected cause ("the retrieval step returns stale documents for date-sensitive questions")
+- **Proposed test**: how to prove or disprove it cheaply (a scenario test, an experiment, an evaluator watching prod)
+- **Proposed fix if confirmed**: the prompt, code, or configuration change
+- **Expected effect**: which metric should move, by roughly how much
+
+Present 2-4 hypotheses ranked by expected impact over effort. Ask which to pursue: the user must understand and agree with the reasoning before anything is created. If the user pushes back, refine the hypothesis with them; they know their domain.
```
langwatch-agent: **[P1] This confirmation gate cannot work when the skill is shipped to Langy.** The PR adds `agent-improvement` to the native set and routes it from `AGENTS.md`, but Langy's absolute rules say to run commands immediately and *never* ask clarification questions or offer options. That global contract explicitly overrides skill instructions. Consequently an in-product request to improve an agent can select a default hypothesis and create a scenario, evaluator/monitor, or experiment without the user accepting it first, despite this skill's core guarantee.

Keep the Langy route read-only until the user explicitly requests a mutation, or add a narrowly scoped, explicit confirmation exception for resource-creating actions and test that no create/run command is issued before consent.

skills/agent-improve/SKILL.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+## Step 4: Execute the Chosen Hypotheses
+
+Each hypothesis becomes real artifacts. Pick the right tool per case:
+
+### Reproduce failures as scenario tests
+Turn real failing traces into scenario tests that fail today and pass once fixed. Fetch the exact inputs with `langwatch trace get <traceId> -f json`, then follow the `scenarios` skill (`langwatch scenario-docs getting-started`) to write them. Real production inputs beat invented ones.
```
langwatch-agent, line 61: **[P1] Do not copy raw production trace content into test and PR artifacts.** `trace get -f json` returns the full trace, including customer inputs/outputs and span attributes. This instruction then tells the agent to turn the *exact* inputs into a scenario test, and the next section directs it to prepare a reviewable PR. On projects without ingestion redaction, that commits customer PII, secrets, or confidential prompts to the repository and exposes them to every PR reader.

Require minimisation and redaction before any trace-derived value leaves LangWatch: use a synthetic reproduction by default, retain only the smallest non-sensitive fields needed to reproduce the failure, and prohibit copying raw trace payloads or links into source/PR artifacts without explicit approval. Add a regression test with a trace containing a secret/PII marker.

langwatch-agent (changes requested):
2 P1 blockers — see the inline comments. The new native skills are generated consistently, but the in-product mutation contract and production-data boundary need to be fixed before this can ship.

skills/agent-improve/SKILL.mdx
rogeriochaves: Valid, fixed in 4753fe713. Step 3 now has an explicit no-ask path: under a contract that forbids questions (Langy's act-immediately rules), the skill presents the ranked hypotheses, states which one it is executing and why in one line, and proceeds with the top-ranked one, keeping the explanation duty. This matches the AGENTS.md rule that absolute rules override skill ask-the-user steps, so the flow degrades to assume-and-state instead of stalling or silently skipping the reasoning.

skills/agent-improve/SKILL.mdx
rogeriochaves: Valid, fixed in 4753fe713. The reproduce-failures section now requires reproducing the STRUCTURE of the failing input with sensitive values replaced by realistic stand-ins, referencing the original as a trace link in a comment instead of pasting it, and Common Mistakes forbids committing raw customer content into tests, fixtures, or PR text.

skills/_tests/agent-improve.scenario.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        ],
+        script: [
+          scenario.user("what should I do next to improve my agent?"),
+          scenario.agent(),
+          (state) => {
+            toolCallFix(state);
+            assertSkillWasRead(state, "agent-improvement");
+          },
+          scenario.user(
+            "the first hypothesis makes sense to me, go ahead and set it up as you proposed. Keep it minimal: create the artifact and show it to me, no need to run the app or install anything heavy"
```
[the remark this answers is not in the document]
rogeriochaves: Added in 4753fe713: a judge criterion now verifies the agent kept execution to creating and showing the artifact without booting the app or installing heavy dependencies. A confirmation run with the updated criteria is in flight and its verdict will be posted on the PR.

Aryansharma28 (dismissed):

Aryansharma28 (approved):

PR #5979, opened 2026-07-20
fix(topic-clustering): let the boot topic-model seed actually run

0xdeafcafe:
## What's wrong

Found while monitoring prod after #5930 merged. Every `langwatch-workers` replica logs this on boot:

```
ERROR langwatch:topic-clustering:seed
  "Topic model seed pass failed; the next boot retries"
  error: "The findMany action on the Topic model requires a 'projectId' or 'projectId.in' in the where clause"
```

`runSeedPass` pages distinct projectIds straight off `Topic`:

```ts
await deps.prisma.topic.findMany({
  distinct: ["projectId"],
  select: { projectId: true },
  orderBy: { projectId: "asc" },
  take: PAGE_SIZE,
  ...(cursor ? { where: { projectId: { gt: cursor } } } : {}),   // only set AFTER page 1
});
```

`Topic` is project-scoped, so the multitenancy guard requires a projectId predicate. The cursored pages would pass (`where.projectId` is a truthy object), but the **first** page carries no `where` at all and throws — so the loop dies before it reaches any project.

`bootSeeds` catches the rejection into "the next boot retries", and every subsequent boot fails identically. The retry message reads like transient recovery; it is actually a permanent stall.

**Impact:** no legacy project's topics were ever recorded onto the event stream. The projection never took ownership, so the ADR-051 source-of-truth migration has not run for anyone. The `seed:v1` done-marker is never written either, so this is stuck rather than half-applied — the fix picks up cleanly on next boot.

## The fix

Enumerate the fleet by paging the **global `Project` model** — which the tenancy guard exempts (it *is* the tenant, addressed by its own id) — filtered to the projects that own Topic rows via a `topics: { some }` EXISTS relation filter. `Topic.projectId` is a FK to `Project.id`, so that is the same project set as a distinct-projectId scan of `Topic`, with the same keyset pagination. No raw SQL and no tenancy opt-out — the whole seed stays on the guarded Prisma model API, mirroring the sibling schedule seed's `findEligibleProjectsPage`, which already pages `Project`.

Each project's rows are still read back through the guarded model API in `seedProjectTopicModel`, which carries its `projectId`. Only the fleet-wide enumeration is cross-tenant.

> Approach adopted from @0xdeafcafe's review (#6002) — the raw-SQL-free alternative to this PR's original `$queryRaw` + `-- @tenancy:` version.

## Why it shipped

`seedTopicModel.ts` had no test at all. And `bootSeeds.unit.test.ts` asserted only that the walk was *called* — never that it succeeded — so it stayed green throughout.

The new `seedTopicModel.unit.test.ts` routes its Prisma stub through the **real `guardProjectId` middleware** rather than answering blindly, so it executes the failing path instead of asserting on query strings. Verified it catches the regression: reverting the fix fails 5 of its cases with the exact production message.

## Testing

- `pnpm test:unit src/server/app-layer/topic-clustering/` — 12 files, 87 passed
- `pnpm typecheck` and `pnpm typecheck:tests` — clean on the changed files (the `typecheck (tests)` job is red only from `langy-conversation.service.unit.test.ts`'s missing `turnExists` mock, inherited from main via #5966 — not this PR)

## Follow-up (not in this PR)

Prod app logs all ship under `service_name="fluent-bit"` with the pod only in the JSON `hostname` field, so there is no per-service Loki label — every query is a full-firehose scan. Worth adding proper service labels, but the fluent-bit config isn't in this repo.

### DIFF
--- a/langwatch/src/server/app-layer/topic-clustering/__tests__/bootSeeds.unit.test.ts
+++ b/langwatch/src/server/app-layer/topic-clustering/__tests__/bootSeeds.unit.test.ts
@@ -11,6 +11,10 @@ import { startTopicClusteringBootSeeds } from "../bootSeeds";
 
 const emptyPrisma = () =>
   ({
+    // The topic-model seed enumerates projectIds with a raw cross-tenant
+    // walk (the tenancy guard rejects a bare page off project-scoped
+    // `Topic`); the per-project read still goes through the model API.
+    $queryRaw: vi.fn().mockResolvedValue([]),
     topic: { findMany: vi.fn().mockResolvedValue([]) },
     topicModelProjection: { findMany: vi.fn().mockResolvedValue([]) },
     project: { findMany: vi.fn().mockResolvedValue([]) },
@@ -35,7 +39,7 @@ describe("startTopicClusteringBootSeeds", () => {
 
       await vi.waitFor(() => {
         // Topic-model seed pages distinct projectIds off the Topic table…
-        expect(prisma.topic.findMany).toHaveBeenCalled();
+        expect(prisma.$queryRaw).toHaveBeenCalled();
         // …and the schedule seed pages eligible projects.
         expect(prisma.project.findMany).toHaveBeenCalled();
       });
@@ -45,6 +49,7 @@ describe("startTopicClusteringBootSeeds", () => {
   describe("when every query rejects", () => {
     it("returns without throwing and surfaces nothing to the boot path", async () => {
       const prisma = {
+        $queryRaw: vi.fn().mockRejectedValue(new Error("pg down")),
         topic: { findMany: vi.fn().mockRejectedValue(new Error("pg down")) },
         topicModelProjection: {
           findMany: vi.fn().mockRejectedValue(new Error("pg down")),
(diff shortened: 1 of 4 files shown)


0xdeafcafe:
Superseded by #6002, which fixes the same boot-seed stall **without** the raw `$queryRaw` / `-- @tenancy:` walk. It pages the guard-exempt global `Project` model with a `topics: { some }` EXISTS filter — same set of projects (`Topic.projectId` is a FK to `Project.id`), same keyset pagination, no raw SQL. Closing in favour of #6002.

0xdeafcafe:
Thanks — you're right, the `Project` walk with a `topics: { some }` EXISTS filter is the better fix: same project set (`Topic.projectId` FK → `Project.id`), same keyset pagination, but it stays entirely on the guarded Prisma model API with no raw SQL and no `-- @tenancy:` opt-out to review. I've adopted it onto this branch (squashed into one commit, credited to your #6002).

The regression coverage is unchanged — the `seedTopicModel` test still routes its Prisma stub through the real `guardProjectId` middleware, so reverting the fix fails 5 of its cases with the exact production message.

So this branch now carries your approach. Happy either way on which PR lands — if you'd prefer #6002 as the canonical one I'll close this; otherwise this one's ready and green (bar the `turnExists` typecheck failure inherited from main via #5966).

PR #5983, opened 2026-07-20
fix(mcp-server): surface evaluation results in search_traces

Aryansharma28:
## Summary

Real fix for the "bug 3" report (Sennder/Bruno support thread): "reading our own evals was easier via raw trace data than via LangWatch." Confirmed this is a genuine gap distinct from the MCP-connection issue fixed in #5965/#5976, not just a downstream symptom.

## The gap

The backend's `POST /api/traces/search` response, in digest mode, already attaches `evaluations` per trace (`app.v1.ts`, via `enrichTracesWithEvaluations`) — but `mcp-server`'s `TraceSearchResult` type never declared that field, and `search_traces`'s digest formatter silently dropped it. `get_trace` (single trace) already surfaced evaluations correctly; `search_traces` (the list view an agent actually uses to scan many traces) did not — forcing N `get_trace` calls to check eval status across N traces instead of one `search_traces` call. That's exactly the ergonomics gap a coding-agent workflow would hit and give up on in favor of reading its own raw data.

## Fix

- Extracted the evaluation-formatting logic (previously inlined in `get-trace.ts`) into a shared `formatEvaluationLines()` util + `EvaluationSummary` type (DRY — now two real call sites).
- Added `evaluations?: EvaluationSummary[]` to `TraceSearchResult` (mirrors `TraceDetailResponse`, which already had it).
- `search_traces`'s digest formatter now renders eval pass/fail/score per trace, same format as `get_trace`.

## Verification

- Unit test added (`tools.unit.test.ts`): asserts `search_traces` shows `PASSED`/`FAILED` status, score, and label per evaluation.
- Full suite: 397/398 (same 1 pre-existing, unrelated, skip-in-CI failure as #5965/#5976 — see those PRs for why).
- **Verified live against a real account**, not just unit-tested: sent a real trace + evaluation via `/api/collector`, confirmed `search_traces` now shows `**faithfulness**: PASSED (score: 0.92)` directly in the list output with no follow-up `get_trace` call needed.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 397/398 pass
- [x] Live verification against a real account with a real evaluation result

## Summary by CodeRabbit

* **New Features**
  * Trace search digests now include evaluation details inline (evaluator name/ID, pass/fail status, optional score, and optional label) when evaluations are present.
  * Evaluation formatting is consistent between trace search and trace detail views.
* **Bug Fixes**
  * Evaluation information is available directly from search results, avoiding unnecessary follow-up trace lookups.
* **Tests**
  * Added/updated unit and integration coverage to validate evaluation rendering (including score/label) in trace search output.

### DIFF
--- a/mcp-server/src/__tests__/all-tools.integration.test.ts
+++ b/mcp-server/src/__tests__/all-tools.integration.test.ts
@@ -985,7 +985,7 @@ describe("All MCP tools integration", () => {
       });
     });
 
-    describe("when a trace has evaluation results", () => {
+    describe("given a trace has evaluation results", () => {
       /** @scenario Agent searches traces and sees evaluation results without a follow-up call */
       it("includes evaluation status in the digest without a follow-up call", async () => {
         const { handleSearchTraces } = await import(
--- a/mcp-server/src/__tests__/all-tools.integration.test.ts
+++ b/mcp-server/src/__tests__/all-tools.integration.test.ts
@@ -14,6 +14,14 @@ const CANNED_TRACES_SEARCH = {
       output: { value: "I am fine, thank you!" },
       timestamps: { started_at: 1700000000000 },
       metadata: { user_id: "user-42", thread_id: "thread-1" },
+      evaluations: [
+        {
+          evaluator_id: "eval-1",
+          name: "Faithfulness",
+          score: 0.95,
+          passed: true,
+        },
+      ],
     },
   ],
   pagination: { totalHits: 1 },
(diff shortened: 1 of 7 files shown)


langwatch-agent (commented):
Deep review complete: no actionable findings. Confirmed the digest API attaches the same evaluation fields consumed by the shared formatter, including nullable score/label/passed values; JSON mode remains raw and the existing get_trace output is preserved. Locally ran the focused tool suite (53 tests), MCP HTTP integration suites (72 tests), and MCP TypeScript check successfully.

PR #5985, opened 2026-07-20
feat(home): put the signal-focused home behind its own feature flag

0xdeafcafe:
## What

The redesigned homepage (the signal-focused composition — briefing sheet leads, chrome grid + recents follow) was gated on `release_langy_enabled`. It now hangs off its own flag, **`release_ui_home_signal_focused_enabled`**, so the home redesign rolls out independently of the Langy assistant.

## How the two flags now compose

- **`release_ui_home_signal_focused_enabled`** decides the home's *composition* (signal-focused vs classic) — nothing else. New hook: `useShowSignalFocusedHome` (`src/components/home/useShowSignalFocusedHome.ts`).
- **`release_langy_enabled`** (`useShowLangy`) decides only the *hand-to-Langy affordances* inside either composition:
  - `HomeBriefingSection` withholds its four Langy handlers without Langy, so the sheet offers only its own Trace Explorer evidence links.
  - `LangyBriefing` hides the per-row Investigate buttons, the ⌘I ask row, and the suggestion chips when their handlers are absent (previously the ⌘I button and chips rendered unconditionally — they'd have been dead controls).
  - `QuietHeadline`'s typed invitation opens the feature surface instead of a Langy conversation, and drops "Do it with Langy".
  - The classic home's `TracesOverview` investigate signal stays tied to Langy — a Langy user on the classic home now actually gets it (it was previously unreachable dead code in that branch).

## Rollout note (operators)

The new flag defaults **off**. Any cohort currently seeing the new home via `release_langy_enabled` falls back to the classic home on deploy until `release_ui_home_signal_focused_enabled` is flipped for the same cohort (PostHog rule / ops row / `FEATURE_FLAG_FORCE_ENABLE=release_ui_home_signal_focused_enabled` in dev). The Langy panel itself is unaffected.

## Spec

`specs/home/signal-focused-home-rollout.feature`

## Tests

- `LangyBriefing.unit.test.tsx`: new cases locking the without-Langy sheet (evidence links keep working; no control hands a signal to Langy) and the with-Langy ask row/chips.
- Ran: briefing + home + featureFlag unit suites (98 tests), home/briefing/langy integration suites (69 tests) — all green. Touched files clean under tslsp diagnostics.

## Summary by CodeRabbit

* **New Features**
  * Added a signal-focused home layout that can be enabled independently from Langy access.
  * Updated home content to adapt based on rollout settings, including briefing, recent work, onboarding, and traces views.
  * Langy actions now appear only when available, while non-Langy users retain access to relevant Trace Explorer links.
  * Quiet-mode suggestions now provide an appropriate alternative action when Langy is unavailable.

* **Bug Fixes**
  * Prevented unavailable Langy controls, prompts, and suggestion actions from appearing.

### DIFF
--- a/langwatch/src/components/home/HomePage.tsx
+++ b/langwatch/src/components/home/HomePage.tsx
@@ -28,6 +28,7 @@ import { OnboardingProgress } from "./OnboardingProgress";
 import { RecentItemsSection } from "./RecentItemsSection";
 import { TimeOfDayAura } from "./TimeOfDayAura";
 import { TracesOverview } from "./TracesOverview";
+import { useShowSignalFocusedHome } from "./useShowSignalFocusedHome";
 import { useTimeOfDay, WelcomeHeader } from "./WelcomeHeader";
 
 /**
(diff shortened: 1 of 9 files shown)


rogeriochaves:
Closing in favour of #5991, which carries this branch's commit (52e475bae, author credit preserved) plus the Langy panel overhaul it now composes with.

PR #5987, opened 2026-07-20
fix(api): surface ModelNotConfiguredError as a clean 400, not a raw 500

Aryansharma28:
## Summary

Found via a systematic sweep testing every LangWatch MCP tool against a real fresh account: `platform_create_evaluator` (and any REST route calling `resolveModelForFeature`, e.g. `POST /api/evaluators`) 500s with `"An unknown error occurred"` on a project with no model provider configured — an ordinary first-touch state for a new project, not an edge case.

## Root cause

`ModelNotConfiguredError` is a plain `Error` subclass (not `HandledError`, no `.status`), so it falls through every case in `error-handler.ts`'s `determineErrorResponse()` and lands in the generic 500 catch-all. `app.v1.ts`'s own comment assumed "the global domain-error middleware" would format it — true for tRPC (`server/api/trpc.ts` has explicit `ModelNotConfiguredError` handling wired to the frontend's missing-model modal), but that handling was never wired into the REST/Hono error handler this route actually uses. `ModelNotConfiguredError` already ships a `toResponseBody()` method built for exactly this REST case — it just had no caller.

**Reproduced directly against production**: `POST /api/evaluators` with a valid `evaluatorType` (`langevals/exact_match`, confirmed valid via `discover_schema`) on a project with zero model providers configured returns HTTP 500 with no actionable detail beyond a trace id.

## Fix

Add a case in `determineErrorResponse()` for `ModelNotConfiguredError`, returning 400 with the same structured cause (`MODEL_NOT_CONFIGURED`, `featureKey`, `role`, `featureDisplayName`, `projectId`) the tRPC interceptor already surfaces on the wire, plus a human-readable message — so a REST/MCP caller gets a clear, actionable error instead of an opaque server fault.

## Verification

- Unit test added: constructs the exact error class and asserts the exact response shape (400, `cause: "MODEL_NOT_CONFIGURED"`, `featureKey`, `role`, `featureDisplayName`, `projectId`, `message`).
- Full `error-handler.unit.test.ts` suite: 15/15 pass.
- `pnpm typecheck`: clean.
- This is server-side application code (not a package a customer runs locally), so live verification requires deployment — the unit test constructs the exact error class through the exact response path the real route uses, which is the strongest verification available pre-deploy.

## Test plan

- [x] Reproduced the 500 directly against production first, before writing any fix
- [x] Unit test added and passing
- [x] `pnpm typecheck` clean
- [ ] Manual re-check of `POST /api/evaluators` on a no-model-provider project once deployed

## Summary by CodeRabbit

- **Tests**
  - Added coverage for handling missing model configuration errors.
  - Verified that these errors return a clear client-facing response with a 400 status, relevant configuration details, and guidance that no model is configured.


0xdeafcafe:
handled error pls

langwatch-agent (commented):
Deep review complete: no actionable findings. Confirmed the REST branch uses the existing client-safe missing-model payload and generic 500 sanitisation remains intact; locally ran the 15-test error-handler suite and full typecheck successfully.

sergioestebance (approved):

PR #5988, opened 2026-07-20
fix(workflows): throw NotFoundError/ValidationError from runWorkflow, not plain Error

Aryansharma28:
## Summary

Found via a systematic MCP tool sweep testing every LangWatch MCP tool against a real fresh account: `platform_run_workflow` on a nonexistent workflow ID returns a raw 500 (`"Workflow not found."`) while `platform_get_workflow` / `platform_delete_workflow` on the exact same ID correctly return 404. Confirmed directly against production.

## Root cause

Two layered bugs:

1. `runWorkflow()` threw plain `new Error(...)` for every precondition/failure case (not found, not published, missing required input, missing LLM key, published-version missing, execution-service failure) — a plain `Error` has no `.status` and isn't a `HandledError`, so it falls through every case in `error-handler.ts`'s `determineErrorResponse()` and lands in the generic 500 catch-all, even for ordinary client errors.
2. The REST route (`misc.ts`'s `handleWorkflowRun`) wrapped the call in its own `try/catch` and unconditionally returned `c.json({ message }, 500)` — so even after (1) is fixed, the newly typed errors never reached the app's `onError(handleError)` middleware. Caught in review by `langwatch-agent`.

## Fix

- `runWorkflow()` now throws `@langwatch/handled-error`'s `NotFoundError`/`ValidationError` (plus a new `WorkflowExecutionFailedError` for the execution-service-failure case) for **every** throw site, not just the two originally reported — a "ruthless review" round caught that only converting two of six sites turned the other four into opaque, message-less 500s.
- `handleWorkflowRun` no longer swallows errors into a flat 500 — it lets them propagate to the already-wired global `onError(handleError)` middleware, which knows how to map `HandledError` subclasses to the right status code.
- The deprecated sibling route `POST /api/optimization/:workflowId/:versionId` now delegates to `handleWorkflowRun` too, instead of duplicating the same (buggy) try/catch — the two routes had started disagreeing on the status code for identical failures.

## Verification

- Unit test (`runWorkflow.not-found.unit.test.ts`) mocking `prisma.workflow.findUnique`: asserts `NotFoundError`/`ValidationError` for the not-found/not-published cases.
- Endpoint-level integration test (`workflows-run-not-found.integration.test.ts`) hitting the real `POST /api/workflows/:workflowId/run` route end-to-end (real Postgres, real auth) for 404, 422, and — pinning the untyped-error path — a mocked untyped `Error` still returns a safe generic 500 without leaking the internal message.
- `pnpm typecheck` / `pnpm run typecheck:tests`: clean.
- Reproduced the original 500 directly against production before writing the fix.

## Test plan

- [x] Reproduced the 500 directly against production first
- [x] Unit test added and passing (2/2)
- [x] Endpoint-level integration test added, covering 404 / 422 / untyped-error passthrough
- [x] `pnpm typecheck` / `pnpm run typecheck:tests` clean
- [x] All remaining plain-`Error` throw sites in `runWorkflow()` converted to typed `HandledError`s
- [x] Sibling deprecated route deduplicated onto the same handler
- [ ] Manual re-check of `platform_run_workflow` on a nonexistent ID once deployed

### DIFF
--- a/langwatch/src/server/workflows/runWorkflow.ts
+++ b/langwatch/src/server/workflows/runWorkflow.ts
@@ -1,6 +1,10 @@
 import { createLogger } from "@langwatch/observability";
 import type { Node } from "@xyflow/react";
 import { nanoid } from "nanoid";
+import {
+  NotFoundError,
+  UnprocessableEntityError,
+} from "~/app/api/shared/errors";
 import { addEnvs } from "../../optimization_studio/server/addEnvs";
 import type {
   ExecutionStatus,
@@ -191,10 +195,10 @@ export async function runWorkflow(
   });
 
   if (!workflow) {
-    throw new Error("Workflow not found.");
+    throw new NotFoundError("Workflow not found.");
   }
   if (!workflow.publishedId) {
-    throw new Error("Workflow not published");
+    throw new UnprocessableEntityError("Workflow not published");
   }
 
   const publishedWorkflowVersion = await prisma.workflowVersion.findUnique({
--- a/langwatch/src/server/workflows/runWorkflow.ts
+++ b/langwatch/src/server/workflows/runWorkflow.ts
@@ -1,10 +1,7 @@
+import { NotFoundError, ValidationError } from "@langwatch/handled-error";
 import { createLogger } from "@langwatch/observability";
 import type { Node } from "@xyflow/react";
 import { nanoid } from "nanoid";
-import {
-  NotFoundError,
-  UnprocessableEntityError,
-} from "~/app/api/shared/errors";
 import { addEnvs } from "../../optimization_studio/server/addEnvs";
 import type {
   ExecutionStatus,
@@ -195,10 +192,12 @@ export async function runWorkflow(
   });
 
   if (!workflow) {
-    throw new NotFoundError("Workflow not found.");
+    throw new NotFoundError("workflow_not_found", "Workflow", workflowId);
   }
   if (!workflow.publishedId) {
-    throw new UnprocessableEntityError("Workflow not published");
+    throw new ValidationError("Workflow not published", {
+      meta: { workflowId },
+    });
   }
 
   const publishedWorkflowVersion = await prisma.workflowVersion.findUnique({
--- a/langwatch/src/server/workflows/runWorkflow.ts
+++ b/langwatch/src/server/workflows/runWorkflow.ts
@@ -15,6 +15,7 @@ import { prisma } from "../db";
 import type { SingleEvaluationResult } from "../evaluations/evaluators";
 import type { MaybeStoredModelProvider } from "../modelProviders/registry";
 import { type NLPOrigin, nlpgoFetch } from "../nlpgo/nlpgoFetch";
+import { WorkflowExecutionFailedError } from "./errors";
 import { stripUnsupportedLLMParamsFromWorkflow } from "./stripUnsupportedLLMParams";
 
 const logger = createLogger("langwatch:workflows:runWorkflow");
@@ -56,7 +57,9 @@ const checkForRequiredInputs = (
 
   requiredInputs.forEach((input) => {
     if (!bodyInputs.includes(input)) {
-      throw new Error(`Missing required input: ${input}`);
+      throw new ValidationError(`Missing required input: ${input}`, {
+        meta: { input },
+      });
     }
   });
   return true;
@@ -98,8 +101,9 @@ const checkForRequiredLLMKeys = (
     llmModelsNeeded.includes(key),
   );
   if (missingKey) {
-    throw new Error(
+    throw new ValidationError(
       `Missing required LLM key: ${missingKey}. Please set the LLM key in the project settings`,
+      { meta: { missingKey } },
     );
   }
   return true;
@@ -200,15 +204,20 @@ export async function runWorkflow(
     });
   }
 
+  const resolvedVersionId = versionId ?? workflow.publishedId;
   const publishedWorkflowVersion = await prisma.workflowVersion.findUnique({
     where: {
-      id: versionId ?? workflow.publishedId,
+      id: resolvedVersionId,
       projectId,
     },
   });
 
   if (!publishedWorkflowVersion) {
-    throw new Error("Published workflow version not found.");
+    throw new NotFoundError(
+      "published_workflow_version_not_found",
+      "Published workflow version",
+      resolvedVersionId,
+    );
   }
 
   // Published versions can predate the node-owned LLM config migration
@@ -288,7 +297,11 @@ export async function runWorkflow(
   });
 
   if (!response.ok) {
-    throw new Error(`Error running workflow: ${response.statusText}`);
+    logger.error(
+      { status: response.status, statusText: response.statusText, projectId, workflowId },
+      "nlpgo execute_sync returned a non-OK response",
+    );
+    throw new WorkflowExecutionFailedError();
   }
 
   return await response.json();
--- a/langwatch/src/server/routes/__tests__/workflows-run-not-found.integration.test.ts
+++ b/langwatch/src/server/routes/__tests__/workflows-run-not-found.integration.test.ts
@@ -73,7 +73,7 @@ describe.skipIf(isTestcontainersOnly)(
       await prisma.organization.delete({ where: { id: organizationId } });
     });
 
-    /** @scenario Running a nonexistent workflow returns 404, not a raw 500 */
+    /** @scenario Running a nonexistent workflow returns 404 */
     it("returns 404 for a nonexistent workflow id", async () => {
       const { app } = await import("../misc");
 
@@ -91,7 +91,7 @@ describe.skipIf(isTestcontainersOnly)(
       expect(body.error).toBe("workflow_not_found");
     });
 
-    /** @scenario Running a never-published workflow returns 422, not a raw 500 */
+    /** @scenario Running a workflow that has never been published returns 422 */
     it("returns 422 for a workflow that has never been published", async () => {
       const { app } = await import("../misc");
 
--- a/langwatch/src/server/routes/__tests__/workflows-run-not-found.integration.test.ts
+++ b/langwatch/src/server/routes/__tests__/workflows-run-not-found.integration.test.ts
@@ -1,0 +1,115 @@
+/**
+ * @vitest-environment node
+ *
+ * POST /api/workflows/:workflowId/run must surface a 404/422 through the
+ * real HTTP path, not the raw 500 `handleWorkflowRun` used to hard-code —
+ * a review comment on the runWorkflow.ts unit-level fix flagged that the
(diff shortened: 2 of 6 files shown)


0xdeafcafe:
throw a new Handled Error

langwatch/src/server/workflows/runWorkflow.ts
```diff
@@ -191,10 +195,10 @@ export async function runWorkflow(
   });
 
   if (!workflow) {
-    throw new Error("Workflow not found.");
+    throw new NotFoundError("Workflow not found.");
```
langwatch-agent: **P1 — The MCP/REST endpoint still returns 500.** `platform_run_workflow` calls `POST /api/workflows/:id/run`, but `handleWorkflowRun` in `src/server/routes/misc.ts:854-859` catches every exception and unconditionally returns `c.json(..., 500)`. These newly typed errors therefore never reach `determineErrorResponse`, so the production symptom described in this PR remains unchanged. Map/rethrow `NotFoundError` and `UnprocessableEntityError` at that route (or delegate to the shared handler), and add endpoint-level regression coverage for both cases.

langwatch-agent (dismissed):
Deep review found a blocking response-contract issue; details are inline.

sergioestebance:
## Ruthless review — blocked on 2 × P1

The core change is right: `NotFoundError`/`ValidationError` from `runWorkflow` do reach `determineErrorResponse` and do produce 404/422 through the real Hono route. I verified that end-to-end. Two things introduced alongside it don't hold up, and both are executed-proof, not code-reading arguments.

---

### P1 — Removing the try/catch turned every *other* failure into an opaque 500

`langwatch/src/server/routes/misc.ts:851-858`

The catch that was deleted was doing two jobs: hard-coding 500 (the bug), and echoing `error.message` (the useful part). Only the typed errors were converted, so the remaining plain `Error` throws in `runWorkflow` now fall through to `error-handler.ts:171-178`'s catch-all and lose their message entirely.

Executed through the real `app` from `~/server/routes/misc`, real auth, real Postgres, with only `runWorkflow` mocked to throw. The deprecated sibling route — which still has the old try/catch — is the A/B control in the same process with the identical error object:

```
PROBE /api/workflows/wf_probe/run          | thrown="Missing required input: question"      | status=500 | body={"error":"Internal server error","message":"An unknown error occurred"}
PROBE /api/optimization/wf_probe/ver_probe | thrown="Missing required input: question"      | status=500 | body={"message":"Missing required input: question"}

PROBE /api/workflows/wf_probe/run          | thrown="Published workflow version not found." | status=500 | body={"error":"Internal server error","message":"An unknown error occurred"}
PROBE /api/optimization/wf_probe/ver_probe | thrown="Published workflow version not found." | status=500 | body={"message":"Published workflow version not found."}

PROBE /api/workflows/wf_probe/run          | thrown="Error running workflow: Bad Gateway"   | status=500 | body={"error":"Internal server error","message":"An unknown error occurred"}
PROBE /api/optimization/wf_probe/ver_probe | thrown="Error running workflow: Bad Gateway"   | status=500 | body={"message":"Error running workflow: Bad Gateway"}
```

`Missing required input: X` is the *most common* caller mistake on this endpoint — the route has no zod validator (`misc.ts:810-823`), the body goes straight to `checkForRequiredInputs` (`runWorkflow.ts:224` → throws at `:59`). An agent calling `platform_run_workflow` with a missing input previously got a wrong status but an actionable message; now it gets neither. For the stated goal — "so I can tell 'fix your request' apart from 'something broke on your end'" — this is a net step backwards on the paths the PR didn't convert.

**Fix direction — finish the typing rather than only removing the catch:**

- `runWorkflow.ts:59` (`Missing required input: X`) → `ValidationError` (422). Caller-fixable, same class the PR already chose for "not published".
- `runWorkflow.ts:101` (missing LLM keys) → `ValidationError` (422).
- `runWorkflow.ts:211` (`Published workflow version not found.`) → `NotFoundError("published_workflow_version_not_found", …)` (404). A dangling `publishedId`, or a bad `:versionId` from the URL on the two-segment route — that is a client-visible not-found, and it's the one remaining case of exactly the bug this PR set out to fix.
- `runWorkflow.ts:291` (`Error running workflow: <statusText>`) → a `HandledError` with a stable code (`workflow_execution_failed`, fault `platform`/`provider`) so callers can tell an upstream failure from an unknown crash. Keep the upstream `statusText` in the log/trace, not the body — that part of the redaction is an improvement.

---

### P1 — The end-to-end regression test never runs. Not in CI, not in any local harness.

`langwatch/src/server/routes/__tests__/workflows-run-not-found.integration.test.ts:17-21`

This file was added specifically to answer the earlier review comment that the route swallowed the typed errors. It is skipped everywhere. From this PR's own CI run (`test-integration (4)`, job 88638161887):

```
stdout | src/server/routes/__tests__/workflows-run-not-found.integration.test.ts
↓ src/server/routes/__tests__/workflows-run-not-found.integration.test.ts (2 tests | 2 skipped)
```

The chain: `langwatch-app-ci.yml:199,203` sets `CI` + `CI_CLICKHOUSE_URL` → `event-sourcing/__tests__/integration/setupEnv.ts:61-64` (a `setupFiles[0]` in `vitest.integration.config.ts`, so it runs *before* the module's top-level const) promotes it to `TEST_CLICKHOUSE_URL` → `isTestcontainersOnly === true` → `describe.skipIf(true)`. Locally it's the same story via a different route: `integration/setup.ts:39` sets `TEST_CLICKHOUSE_URL` in **both** testcontainers and native-local-services mode.

Controlled experiment — the file copied verbatim with only `describe.skipIf(isTestcontainersOnly)(` → `describe(`, both run together under simulated CI:

```
[globalSetup] Using CI service containers
...statusCode":404..."code":"workflow_not_found"...
...statusCode":422..."code":"validation_error"...

 Test Files  1 passed | 1 skipped (2)
      Tests  2 passed | 2 skipped (4)
```

So the test is correct and passes — it's just gated off. The guard is inverted for this file: the suite needs only Postgres, which every harness provides; `isTestcontainersOnly` is cargo-culted from ClickHouse-dependent suites.

**Fix:** delete lines 19 and 21. That's it. (If a guard is genuinely wanted, gate on absence of `DATABASE_URL`, never on presence of `TEST_CLICKHOUSE_URL`.) Consequence today: the `@integration`-tagged feature file's two scenarios have zero executing coverage, and the route-level half of this fix — the half a reviewer had to catch — ships ungated.

Out of scope but worth its own issue: that same `isTestcontainersOnly` idiom appears in ~14 integration suites. If it means "never run" for all of them, that's a sizeable silent coverage hole.

---

### P2 — The sibling route still flattens the same typed errors to 500

`langwatch/src/server/routes/misc.ts:642-652`

`POST /api/optimization/:workflowId/:versionId` (deprecated, still live) calls the same `runWorkflowFn` and still wraps it in `catch → c.json({ message }, 500)`. The new `NotFoundError`/`ValidationError` are flattened there exactly as they were before this PR.

Attribution: the catch is pre-existing, so this is capped at P2 and does not block. But the *divergence* is new — as of this diff two routes calling one function disagree about the status code for identical failures, which is a worse state than the consistent-and-wrong one it replaced. Cheapest fix: point both routes at `handleWorkflowRun`, which is what the two-segment `/workflows/:workflowId/:versionId/run` route already does.

### P2 — No test pins the untyped-error behavior

Whatever you decide for P1 #1, add a third case: mock `runWorkflow` to throw a plain `Error` and assert the body. Right now the intended behavior for non-typed failures is unstated and unprotected, which is how the regression above got in.

---

### P3 — non-blocking nits

- **PR body is stale.** It says `UnprocessableEntityError` throughout; the code throws `ValidationError` (code `validation_error`, `httpStatus` 422). Same status, different class and different `error` field on the wire — worth correcting before merge.
- **`afterAll` is not failure-safe** (`workflows-run-not-found.integration.test.ts:69-74`). If `beforeAll` throws partway, the deletes run with `undefined` ids, the first throws, and the rest of the `--test-*` rows are orphaned in the shared test DB. Wrap each, or delete only what was created.
- **The `sk-lw-test-` fixture prefix is load-bearing and undocumented.** It only authenticates because that shape *fails* the scoped-API-key regex (`api-key-token.utils.ts:190-192`) and falls back to `legacyProjectKey`, which bypasses the RBAC ceiling by design (`auth-middleware.ts:343`). Tidy it into a realistic `sk-lw-{16}_{48}` and the test flips to 401 with no clue why. One comment prevents that.
- **Response-shape drift.** The modern workflows API returns `{ error: "Workflow not found" }` (`src/app/api/workflows/[[...route]]/app.ts:123`); this route now returns `{ error: "workflow_not_found", message: "Workflow not found: <id>", id: "<id>" }`. Both are 404 and both are defensible — flagging that a client hitting both surfaces sees two shapes.
- **Feature file for a bug fix.** House practice is that specs describe new behavior, not regressions; a regression test carries this fine on its own. Not worth reverting, just noting the convention.

---

**Verdict: blocked** until the two P1s are resolved. The typed-error conversion itself is correct and verified — it just needs to cover the rest of the throws, and the test that proves it needs to actually run.

Aryansharma28:
## Status update

Verified the ruthless-review P1/P2 findings from `sergioestebance` are resolved in `d02452472` (already on this branch before I touched it):

- `runWorkflow()` throws a typed `HandledError` at all six sites (not found, not published, missing input, missing LLM key, missing published version, execution-service failure) — confirmed by reading `runWorkflow.ts` end-to-end.
- `handleWorkflowRun` no longer swallows to a flat 500; the deprecated `/api/optimization/:workflowId/:versionId` route now delegates to it instead of duplicating the old try/catch.
- The integration test's `isTestcontainersOnly` skip guard is gone — confirmed actually executing in the latest CI run (`test-integration (4)`, both scenarios + the untyped-error case, 3/3 passing, not skipped).
- `afterAll` cleanup in the integration test is now per-row guarded.

Additional fix pushed just now (`f1c8b6af1`): CodeRabbit's `vi.hoisted` TDZ nit on `runWorkflow.not-found.unit.test.ts` — harmless in practice (test already passed 2/2), but brought in line with the `vi.hoisted` pattern used elsewhere in this codebase for defensiveness. Verified locally with `npx vitest run` (2/2 pass) and a scoped typecheck.

Also corrected the PR title — it still said `UnprocessableEntityError`, which isn't a real class here; the code throws `ValidationError` (422), matching the PR body and the ruthless review's P3 note.

**One remaining CI failure is unrelated to this PR:** `sdk-python-ci` / `test` fails on `tests/test_examples.py::test_example[examples/distributed_tracing.py]` with a 401 from OpenAI (`Incorrect API key provided: sk-proj-...`). That's an expired/rotated secret used by the Python SDK's e2e example tests, not something this diff touches — the same failure shows up on unrelated PRs (e.g. #5977) at the same commit-independent rate. Flagging for whoever owns that CI secret rather than fixing here.

sergioestebance (approved):
blessed

PR #5990, opened 2026-07-20
fix(otel): end-to-end telemetry correctness and customer/internal trace isolation

0xdeafcafe:
Three commits, layered; each message documents its units in full. Every defect was found by investigation (prod Grafana, then a live local stack) and every fix is test-pinned.

## 1. `fix(otel)`: internal telemetry correctness
- **One span + one log line per `/api` request** — 15 SecuredApp families sharing `basePath: "/api"` each ran tracer+logger on every request (10-13x duplicates in prod). Deduped via the async request context; regression test drives the real `createApiRouter()`.
- `service.family` instead of overriding the reserved `service.name` per span.
- **Group-queue jobs are their own trace** (ROOT_CONTEXT + producer link) — ends the cross-tenant megatraces.
- ioredis `requireParentSpan` — parentless `connect`/`brpop` root spans gone.
- **`service` field on every log line** (the key log shipping promotes to the Loki `service_name` label) + per-role `OTEL_SERVICE_NAME` in the chart.
- **Full sampling by default in every environment** (also #5992) — the 10% non-local default read as a dead exporter on quiet services and would silently drop customer data on nlpgo's multi-tenant path.

## 2. `feat(langy)`: turn hardening
- **422** for validation failures across the Go RPC surfaces (parse failures stay 400).
- **Turn identity = hash(user + idempotencyKey + content)** — byte-identical retries replay; same key + different content is a 409 `langy_idempotency_mismatch`; users can't collide. `requestId` → `idempotencyKey` (alias kept, no DB migration); empty sends rejected 422 before anything durable.
- **Permanent dispatch rejections terminalize** the turn instead of retrying the same 4xx through the outbox forever (the observed poison loop).

## 3. `feat(otel)`: customer/internal trace isolation + worker telemetry pipeline
- **Customer trace bridge → `pkg/customertracebridge`** with fail-closed per-service `Policy{Allow, Stamp}` at the export boundary — fixes the prod leak of pod k8s/cloud/env identity onto customer spans (otel-go's `WithResource` silently merges `resource.Environment()`). Wire-level protobuf-decoding test; isolation pinned both directions incl. forged-origin replacement and the `langwatch-*` service-name guard.
- **nlpgo ops fallback** — its own untenanted spans (previously silently dropped) export under full internal identity, never touching the customer path.
- **The langy worker hop was silently dead**: opencode exports OTLP/HTTP JSON and ignores `OTEL_EXPORTER_OTLP_PROTOCOL`; the relay's proto-only parse 400'd every batch ever sent, with zero logging. Now: token-redacted request logging, content-type-aware decode normalized to protobuf, `ai.*`/`gen_ai.*`-only forwarding, batch-orphan reattachment, and a **platform-owned `langy.turn` root span** forwarded to the customer (same span id as the internal turn span, scope `langy-agent`, origin=langy at span+resource level).

## Final trace shape
```
trace …
└── langy.turn                      langy      origin=langy
    ├── gen_ai.responses            gateway    origin=gateway
    └── ai.streamText               opencode
        └── ai.streamText.doStream  opencode
```

The tree above is a manual observation from the product span store on a local
stack — it is illustration, not evidence. What is pinned in code, and what CI
actually enforces, is the contract that shape depends on:

- `TestForwardTurnSpan` (`services/langyagent/adapters/otelrelay/otelrelay_test.go`)
  asserts the forwarded root is named `langy.turn`, carries the *same span id* as
  the internal turn span (the cross-store correlation key every child parents on),
  and arrives under the `langy-agent` instrumentation scope.
- `TestRelayTraces` / `TestRelayTraces_JSONEncodedExport` pin the content-type-aware
  decode and the `ai.*` / `gen_ai.*`-only forwarding filter that admit the children.

The nesting of `gen_ai.responses` and `ai.streamText` under that root is a
consequence of those two, not separately asserted.

Pre-existing failures not from this branch: `LangyPlanCard` unit test (fails on origin/main), onboarding Go snippet fixtures.

## Summary by CodeRabbit

* **New Features**
  * Added idempotency-key-based Langy turn identities with retry-safe mismatch handling and clearer rejection/empty-message errors.
  * Enhanced customer trace telemetry with project-based routing, resource allowlisting, and operational fallback; added configurable telemetry service names and full default sampling.
* **Bug Fixes**
  * Prevented duplicate request logging and nested tracer spans.
  * Improved worker dispatch “rejected” handling and standardized validation responses to HTTP 422.
* **Tests**
  * Added/updated coverage for logging, tracing, idempotency, and telemetry routing.

## Deployment Impact

- **Env vars**: no new required variables anywhere. `OTEL_SERVICE_NAME` is now set explicitly per role by the chart (app `langwatch-app`, workers `langwatch-workers`) — code defaults cover unset. `OTEL_TRACES_SAMPLER`(+`_ARG`) remains the operator override for the new full-sampling default; no deployment sets it today.
- **Helm values added**: `app.otel.serviceName` and `workers.otel.serviceName`, both defaulted — a vanilla `helm install` with no overrides renders identically-shaped manifests with the two new env entries and needs no action.
- **Default behavior change on deploy**: Go services with an OTLP endpoint configured (langyagent, gateway, nlpgo ops when enabled) go from ~10% sampling to full export of their own operational spans — internal collector volume rises accordingly; dial down per service with `OTEL_TRACES_SAMPLER` if needed. App/worker log lines gain a constant `service` field (log-shipping label; no shipper config change required).
- **BYOC / dataplane impact**: the gateway's customer-retold spans stop carrying pod/cluster/cloud resource identity and start carrying `langwatch.origin=gateway` — anything downstream keying on those resource attributes must key on the origin marker instead. The langy relay's changes (JSON accept, span filter, turn root span) ship entirely inside the langyagent image; self-hosted installs get them on image upgrade with no config change. No database migrations.


0xdeafcafe:
**Customer turn span: implemented and live-verified.** The final tree, read from the product's span store after a real turn:

```
trace 330b4a7e…
└── langy.turn                      service=langy    origin=langy   (platform-owned root)
    ├── gen_ai.responses            gateway retell   origin=gateway
    └── ai.streamText               opencode
        └── ai.streamText.doStream  opencode
```

Three commits complete it:
1. **`ForwardTurnSpan`** — at turn end the relay forwards the turn span into the customer's project: same trace+span id as the internal `langy.turn` (the cross-store correlation key), real wall-clock bounds, instrumentation scope `langy-agent`, `service.name=langy`, and `langwatch.origin=langy` at both span and resource level (span-level is the strongest signal origin resolution accepts). Detached/best-effort; policy-scrubbed like every other forward.
2. **Batch-local orphan reparenting** — a worker span whose parent isn't in its own batch attaches to the turn instead of dangling.
3. **Customer span filter** — only the `ai.*`/`gen_ai.*` families forward; storage/session plumbing (`sql.execute`, `SessionProcessor.*`) stays out of customer traces, and since the plumbing's disappearance was what orphaned the `ai.*` spans downstream, filtering before reparenting is also what makes the tree connect. The internal content-stripped copy still sees the full batch.

0xdeafcafe:
**Post-rewrite review (single deep-review agent over the full diff) — findings and dispositions:**

- **P0 fixed** — two counter-label unions in `metrics.ts` were missing `mismatch`/`rejected` (hard typecheck failure; vitest doesn't typecheck, which is how they slipped).
- **P1 fixed** — the `rejected` dispatch bucket was every-4xx-but-409/428, which would have terminalized healthy turns on a transient 401 during secret rotation or a proxy 404. Narrowed to 400/422; everything else stays retryable `unavailable`.
- **P1 fixed** — nlpgo's ops fallback could ship a key-less *customer-path* span (goroutine that lost its context) into internal telemetry with attributes intact. The ops copy now fails closed on content: operational skeleton only (name, kind, timing, status) — no attributes, events, or links ever cross. Test pins it.
- **P2 fixed** — impersonation guard hardened: case-folded and `-`/`_`-normalized `langwatch*` check, and `langy` reserved (the relay's own turn span stamps it AFTER the policy pass — the one legitimate use).
- **P2 fixed** — `TestForwardTurnSpan` data race (unsynchronized body read) → signalling-ingest channel sync; `go test -race` green on the package.
- **P2 fixed** — `service.family` now derives from the route Hono actually matched (after `next()`), since the dedupe guard means the surviving span belongs to the first-mounted family and its construction-time name no longer identifies the owner.
- **P2 fixed** — stale dangling comment in `attrs_genai.go`.
- **P2 accepted + documented** — deploy-boundary window on the idempotency migration: a pre-deploy receipt mismatches a byte-identical post-deploy retry (409) for the minutes around one deploy; chosen over carrying legacy-id compatibility forever. Noted in the mismatch check.
- **P2 deferred** — the dedup regression test pins the log count but not span count; an in-memory-exporter span assertion is a follow-up.

Reviewer verdict on everything else: architecture sound, isolation work genuinely strong; verified-fine list included the middleware guard ordering, idempotency hash edges, turn-span forward reachability, token redaction, and 422 caller fallout.

Bots: CodeRabbit posted only its auto-summary (zero inline findings), Copilot errored on quota — no threads existed to resolve. All fixes are folded into the three commits they belong to; the tree diff vs the reviewed state is exactly the fix set.

0xdeafcafe:
**CodeRabbit's five queued comments (failed to post as threads — no threads to resolve) — all addressed:**

1. `langy.ts` — the one-of `idempotencyKey`/`requestId` invariant moved into the Zod schema (`.refine` on both turn inputs); the imperative check remains only as type narrowing.
2. `langy-turn.service.ts` — the empty-send rejection now self-reports (`getLangyTurnsCounter("rejected")`) like every other rejection branch; the path is no longer invisible in the turn-outcome metric.
3. `registry.go` — `NewRegistryWithSize` no longer discards the `lru.New` error: non-positive sizes fall back to the default, the impossible-error path fails loudly at construction instead of a deferred nil-cache panic.
4. `internal_span.go` goimports — fixed in the lint pass (this one was already green).
5. `pathSignal` — genuine bug confirmed: `/llm/` proxy paths contain `/v1/` and were misclassified in the new relay request log. `/llm/` now checked first; table test added covering chat/completions, responses, and bare-token paths.

Also since the review-response comment: the last stale `requestId` caller (a Docker-gated integration test CI caught) renamed, and its replay assertion updated to the content-derived contract — a retry re-derives the same turn id, so a differing turn id under the same key now correctly yields `mismatch`; verified locally against real Postgres (5/5).

0xdeafcafe:
@coderabbit review

rogeriochaves (dismissed):

PR #5995, opened 2026-07-20
feat(gateway): flow VK tags to customer spans as trace labels

### LINKED ISSUE #5993: Virtual-key tags never reach traces — not filterable in the Trace Explorer (and cache-rule vk_tags matchers never fire)
## Problem

Virtual keys accept operator-assigned tags (`config.metadata.tags`, editable in the VK drawer), but the tags are dead everywhere downstream:

1. **Bundle**: the gateway config materialiser never lifts `config.metadata.tags` into the bundle, and the Go wire (`configWire`) has no field to receive them — `BundleConfig.VKTags` is permanently `nil`. The comment on the field even says so: *"Currently always empty — the schema doesn't store tags yet"* (`services/aigateway/domain/bundle.go`).
2. **Spans**: the customer-trace emitter stamps `langwatch.virtual_key_id` but no tag/label attribute, so gateway traces carry nothing tag-related.
3. **Cache rules**: `vk_tags` matchers (documented in `specs/ai-gateway/cache-control-rules.feature` as AND-subset matching) compare against the always-empty `BundleConfig.VKTags`, so a cache rule scoped by VK tags never fires.

Net effect: users tag their VKs expecting to slice gateway traffic by team/app in the Trace Explorer, and nothing anywhere reflects the tags.

## Proposed solution

- Materialiser emits a top-level `vk_tags` (lifted from `config.metadata.tags`); the Go wire parses it into `BundleConfig.VKTags`.
- The customer-span emitter stamps the tags as the `langwatch.labels` span attribute — the exact key the trace ingestion pipeline already maps into `metadata.labels` (`otel.traces.ts`), which is what the Trace Explorer's "Label" filter queries. Tagged gateway traffic becomes filterable with **zero explorer-side changes**, in both the Trace Explorer and the Analytics filter sidebar.
- Cache-rule `vk_tags` matchers start working as a side effect, since the evaluator already reads `Bundle.Config.VKTags`.

Verified end-to-end on a live deployment: a tagged VK's requests produce spans carrying `langwatch.labels`, the trace summaries inherit the attribute, and the traces surface under the Label filter. PR to follow.

thiagolmoraes:
Fixes #5993

## What

Virtual-key tags (`config.metadata.tags`) become visible in the product: gateway traffic from a tagged VK is now filterable under the Trace Explorer's **"Label"** filter (and the Analytics filter sidebar), and cache-rule `vk_tags` matchers start working. Previously the tags were stored by the UI but dead everywhere downstream.

## How

Three hops, one per layer:

1. **Materialiser** (`config.materialiser.ts`): emits a top-level `vk_tags`, lifted from `config.metadata.tags`.
2. **Gateway wire** (`config_wire.go`): parses `vk_tags` into `BundleConfig.VKTags` — the field existed but was permanently `nil` (its own comment said "currently always empty"). The cache-rule evaluator already reads it, so `vk_tags` matchers become functional with no further change.
3. **Customer-span emitter** (`customertracebridge/emitter.go`): stamps the tags as the `langwatch.labels` span attribute — the exact key the trace ingestion pipeline already maps into `metadata.labels` (`otel.traces.ts`), which is what the Label filter queries. Zero explorer-side changes.

## Tests

- TS integration (real PG): materialiser lifts `config.metadata.tags` to `bundle.vk_tags`.
- Go unit: wire decode of `vk_tags` into `BundleConfig.VKTags`; emitter stamps `langwatch.labels` when tags are present and omits the attribute when they are not.
- Spec scenarios added to `specs/ai-gateway/span-shape.feature` (§ VK tags land on customer spans as labels).

Verified end-to-end on a live deployment: a tagged VK's request produced a span carrying `langwatch.labels`, the trace summary inherited the attribute, and the trace surfaced under the Label filter in both the Trace Explorer and Analytics.

## Review follow-ups (pushed on top)

Merged latest `main` (resolved the `config_wire.go` conflict with `LangyMirrorTier`), then:

- **Bounded the tags.** `config.metadata.tags` had no cap on count or length, no de-duplication and no blank filtering, and it is reachable from the gateway-platform REST body as well as the drawer. Now that every tag is stamped on every span and aggregated by the Label facet, that was unbounded per-request cardinality. `normalizeVkTags` trims, drops blanks, de-duplicates, and caps at 32 tags of 128 characters, slicing by code point so truncation never splits a surrogate pair. It normalises rather than rejects, because the same parse runs on the gateway's config-fetch path and a stored over-size key still has to resolve to a servable bundle.
- **Covered the pipeline hop.** `app/pipeline/trace.go` carries `VKTags` at four independent exits and had no test; `trace_vk_tags_test.go` now covers all four.
- Drawer copy and `maxLength` state the caps and say that tags become trace labels.

### Round 2: the Tags field says what it does to your traces

The drawer's Tags field carried its explanation in a permanent block under the input, written in vocabulary only someone who has read the gateway code would recognise ("VKs", "AND-subset semantics", a rule matcher spelled as a code literal). It now explains itself behind the label's `(i)`, the way every sibling field in that drawer already does, and the copy is written for the person typing a tag:

> Group this key's traffic by team, app, or environment. Every trace this key sends carries its tags as labels, so anyone with access to the project can see them and filter on them. A cache rule that lists tags applies to any key carrying all of them. Saving keeps the first 32 tags, trims each to 128 characters, and drops blanks and repeats.

The second sentence is the consequence this PR introduces, and it belongs at the moment someone types a tag rather than in a release note.

- **Nothing is dropped silently any more.** Typing a 33rd tag used to lose it with nothing said anywhere. The field now raises a notice while the typed list is over a cap, and only then: "Only the first 32 tags will be saved." / "Tags longer than 128 characters will be shortened." Repeats and blanks are deliberately not warned about, since dropping them costs nothing and the `(i)` already says they go.
- **One shared module for both drawers.** `components/gateway/virtualKeyTagsField.ts` owns the copy, the caps and the parsing both drawers do on save, so the create and edit drawers cannot drift. The numbers in the copy are interpolated from the limits `normalizeVkTags` enforces and pinned by a test, so the copy cannot promise a bound the code does not apply.
- **Fixed the field's own `maxLength`.** It counted UTF-16 code units against a per-tag limit measured in code points, so a full list of astral-plane tags could be clipped by the input before the server ever saw it. The per-tag budget is doubled, which makes clipping impossible for any list the server would have kept whole.
- **Specs and tests.** New Tags-field scenarios in `specs/ai-gateway/virtual-keys.feature`; `VirtualKeyTagsField.integration.test.tsx` renders the real drawer and the real popover (no shallow rendering, no stubbing of the component under test) and asserts the copy stays invisible until the icon is opened; `virtualKeyTagsField.unit.test.ts` covers the caps, code-point measurement and the copy. Mutation-checked: restoring the old copy and the permanent helper block fails 8 of the 10 RTL cases, and reverting the code-point fix fails 2 of the 14 unit cases.

One rough edge left alone: at the drawer's width the `(i)` popover opens over the input it belongs to, because `FieldInfoTooltip` positions every caller `right-start` and the drawer sits at the right edge. Pre-existing and shared by the Name and Routing policy icons, so changing it belongs in its own PR rather than this one.

## Screenshots

Trace Explorer filtered by a VK tag, narrowing the list to just that key's gateway traffic:

filtered by vk tag

The LABELS column carrying the key's tags, with the label filter applied:

labels column

The tags field with the bounds and the updated copy:

vk drawer tags

The Tags explanation behind the label's `(i)`, with nothing sitting under the field:

tags tooltip

The over-cap notice, shown only while the typed list actually exceeds a limit:

over limit warning

At 390px wide, the whole explanation still fits and reads:

tags tooltip at 390px

Dark mode:

tags tooltip dark

The edit drawer carries the same field, and the list row shows the saved tags after a dirty list (`"app=acme-support, team=platform,  team=platform ,, env=dogfood"`) was normalised on the way in:

edit drawer tags tooltip

## Deployment Impact

No new env vars and no helm values added or changed. A vanilla `helm install` with no overrides behaves exactly as before, with one behavior change that needs a release note rather than an operator action: `BundleConfig.VKTags` was permanently empty until now, so any cache rule an org already saved with a `vk_tags` matcher has been a silent no-op. Once this ships those rules start applying for real (force/disable cache mode, TTL override) to any VK carrying the matched tags. Operators who want the old behavior should review their cache rules before upgrading.

Tag values are bounded on the way in (32 tags of 128 characters, de-duplicated, blanks dropped), so the new `langwatch.labels` span attribute adds at most a few hundred bytes per gateway span and a bounded number of distinct values to the Label facet. No migration, no backfill, no data-plane restart ordering requirement: the gateway picks the tags up on its next config fetch. BYOC dataplanes need the new gateway build to stamp the attribute, but an older gateway against a newer control plane simply ignores the unknown `vk_tags` key and keeps working. Rollout order between the control plane and the data plane therefore does not matter.

## Summary by CodeRabbit

- **New Features**
  - Virtual Key tags are now carried through gateway configuration and stamped onto customer spans as `langwatch.labels`, enabling filtering in Trace Explorer.
  - Untagged Virtual Keys no longer emit label attributes.

- **Improvements**
  - Virtual Key create/edit tags now use shared parsing and server-aligned limit checks, with clearer tooltip text and friendly warnings when limits are exceeded.
  - Gateway trace end spans now consistently include the relevant tag context.

- **Tests**
  - Added/updated unit and integration tests covering config materialisation, control-plane decoding, span parameter propagation, and the Tags field UI/validation.

- **Documentation**
  - Updated the span-shape spec to document `langwatch.labels` behavior for tag presence/absence.


langwatch-agent (commented):

services/aigateway/adapters/gatewaytracer/attrs.go
```diff
@@ -4,6 +4,10 @@ const (
 	AttrOrigin                    = "langwatch.origin"
 	OriginGateway                 = "gateway"
 	AttrVirtualKeyID              = "langwatch.virtual_key_id"
+	// AttrLabels carries the VK's tags; the trace pipeline ingests this
+	// exact key into metadata.labels (otel.traces.ts), which the Trace
+	// Explorer filters as "Label".
+	AttrLabels = "langwatch.labels"
```
Aryansharma28: [P0] This file conflicts with `main` — GitHub already reports the PR as unmergeable (`mergeable: false`, `mergeable_state: dirty`).

**Why:** `main` moved past this PR roughly 16 hours after it landed. `#5990` (`fix(otel): end-to-end telemetry correctness and customer/internal trace isolation`, merged `2026-07-21T12:56Z`, this PR's only commit is `2026-07-20T20:50Z`) rewrote this exact const block: `AttrOrigin`, `AttrVirtualKeyID`, and `AttrGatewayReqID` were all removed from `gatewaytracer/attrs.go` and relocated to a new `pkg/customertracebridge/attrs.go`, with a doc comment establishing the new convention: *"Single source for both this package (customer-bound spans) and the gateway's own operational tracer (gatewaytracer imports these for its scalar-only internal copies)."* `gatewaytracer` now imports attribute keys *from* `customertracebridge`, not the other way around. I confirmed this with `git merge-tree $(git merge-base main pr-5995) main pr-5995`, which produces a real `<<<<<<<`/`>>>>>>>` conflict marker on this exact hunk.

**Suggested fix:** Don't add `AttrLabels` here. Add it to `pkg/customertracebridge/attrs.go` instead (alongside `AttrVirtualKeyID`/`AttrGatewayReqID`), matching the new one-directional-import convention `#5990` established. See the paired comment on `emitter.go` below for where the actual stamping logic needs to move.

*(Small aside, low priority since this file is being rewritten anyway: as submitted, this hunk isn't `gofmt`-clean — `gofmt -l`/`gofmt -d` both flag it, because the inserted comment breaks the const block's alignment group without realigning the surrounding `=` columns. Not caught by CI yet since no `gofmt` check exists in `go-ci` and workflows haven't been approved to run on this fork PR.)*

pkg/customertracebridge/emitter.go
```diff
@@ -192,6 +192,12 @@ func (e *Emitter) EndSpan(ctx context.Context, params domain.AITraceParams) {
 	if params.GatewayRequestID != "" {
 		attrs = append(attrs, attribute.String(gatewaytracer.AttrGatewayReqID, params.GatewayRequestID))
 	}
+	// VK tags become the trace's labels: the ingestion pipeline maps the
+	// langwatch.labels attribute into metadata.labels, the field the Trace
+	// Explorer filters as "Label".
+	if len(params.VKTags) > 0 {
```
Aryansharma28, line 297: [P0] This file no longer exists at this path on `main` — it's a modify/delete conflict, not just a textual one.

**Why:** `#5990` moved the entire `customertracebridge` package from `services/aigateway/adapters/customertracebridge/` to `pkg/customertracebridge/` (verified: `git ls-tree -r main -- '*customertracebridge*'` shows every file, including `emitter.go`, `emitter_error_suppress_test.go`, and a new `policy.go`, only under the `pkg/` path — nothing remains at the old path). Concretely this means:
- This `EndSpan` change (and the whole "third hop" the PR describes) needs to be re-applied to `pkg/customertracebridge/emitter.go` instead.
- The new `vk_tags_span_test.go` test file needs to move to `pkg/customertracebridge/` too — I confirmed the `recordSpanForParams` helper it depends on still exists there (`pkg/customertracebridge/emitter_error_suppress_test.go:21`), so the test logic itself ports over cleanly once relocated.
- I checked `pkg/customertracebridge/emitter.go`'s current `EndSpan` on `main` — the equivalent insertion point (right after the `AttrGatewayReqID` block) still exists structurally, so this is a same-shape re-apply, not a re-design.

I want to be precise about scope: the other 4 files in this PR (`domain/bundle.go`, `domain/verdict.go`, `app/pipeline/trace.go`, `controlplane/config_wire.go`) merge cleanly against current `main` — this package-move is the only structural casualty, but it's a real one, and it's the file that actually stamps the customer span, i.e. the whole point of the PR.

**Suggested fix:** Rebase onto current `main`, re-apply this `if len(params.VKTags) > 0 { ... }` block (and its comment) into `pkg/customertracebridge/emitter.go`, move `vk_tags_span_test.go` alongside it, and add `AttrLabels` to `pkg/customertracebridge/attrs.go` per the previous comment.

services/aigateway/app/pipeline/trace.go
```diff
@@ -68,6 +68,7 @@ func Trace(begin BeginSpanFunc, end EndSpanFunc) Interceptor {
 							InternalProviderID: internalProviderID,
 							RequestType:        call.Request.Type,
 							VirtualKeyID:       call.Bundle.VirtualKeyID,
+							VKTags:             call.Bundle.Config.VKTags,
```
Aryansharma28, line 72: [P2] This wiring hop (4 call sites) has no test coverage, unlike the hops on either side of it.

**Why:** This line — and the same `VKTags: call.Bundle.Config.VKTags` assignment at lines 90, 120, and 271 — is the connector between "the wire decodes `vk_tags` into `BundleConfig.VKTags`" (tested: `config_wire_test.go`) and "the emitter stamps `langwatch.labels` when `AITraceParams.VKTags` is non-empty" (tested: `vk_tags_span_test.go`). Nothing in the diff or the existing suite actually drives a `Call{Bundle: ...}` through `Trace()` and asserts the resulting `AITraceParams.VKTags` matches `Bundle.Config.VKTags`. I checked all three existing `trace_*_test.go` files in this package (`trace_body_test.go`, `trace_error_test.go`, `trace_project_test.go`) — none reference `VKTags`.

This isn't hypothetical busywork: the codebase already has the exact pattern to copy. `trace_project_test.go`'s `TestTraceBeginsSyncAndStreamSpansUnderConfigProject` builds a `domain.Bundle{Config: domain.BundleConfig{TraceProjectID: ..., ProjectOTLPToken: ...}}`, runs it through `Trace()`'s sync and stream paths, and asserts the field surfaces correctly on the other side. `VKTags` needs the same treatment — and this hop is exactly the kind of place a future refactor (or the rebase this PR now needs, see the P0 comments) could silently drop one of the 4 assignments or typo the source field, with every existing test still green.

I ran the full `aigateway` test suite (`go test ./...`, all packages) against the PR's own commit to confirm this isn't just an oversight on my part reading the diff — the suite genuinely has no assertion here.

**Suggested fix:** Add a case to `trace_project_test.go` (or a new small test) that sets `Bundle.Config.VKTags` on the fixture, captures the params passed to `end(...)`, and asserts `VKTags` came through — for both the sync and stream paths, mirroring the existing test's structure.

services/aigateway/domain/bundle.go
```diff
@@ -70,10 +70,10 @@ type BundleConfig struct {
...
 
-	// VKTags carries VK-level labels honored by cache rule matchers
-	// (vk_tags). Currently always empty — the schema doesn't store tags
-	// yet — but plumbed end-to-end so future tag wiring is a one-line add
-	// in the control-plane materialiser.
+	// VKTags carries the VK's operator-assigned tags (config.metadata.tags
+	// on the control plane). Consumed by cache rule matchers (vk_tags) and
+	// stamped on customer spans as langwatch.labels so the Trace Explorer
+	// can filter gateway traffic by tag.
 	VKTags []string
```
Aryansharma28, line 88: [P2] Flipping `VKTags` from permanently-empty to real also silently activates cache-rule `vk_tags` matchers in production for anyone who already configured one.

**Why:** `services/aigateway/adapters/cacherules/evaluator.go:69` already calls `matchesAllTags(match.VKTags, eval.VKTags)`, and the rule-creation UI (`langwatch/src/components/gateway/cacheRule.form.tsx:94`) already lets customers set a `vk_tags` matcher today — both predate this PR. Since `eval.VKTags` (sourced from this exact field) has always been `nil`/empty, `matchesAllTags` with a non-empty matcher has always returned `false` — any cache rule with a `vk_tags` matcher an org already saved has been a **silent no-op** until now. The moment this ships, those same rules start actually applying (force/disable cache mode, TTL override, etc.) for any VK whose tags now match, with no feature flag, changelog entry, or customer-facing note in this PR or the linked issue about that transition.

To be clear, this is presumably the *intended* fix per the linked issue ("cache-rule vk_tags matchers never fire" is listed as one of the three symptoms) — I'm not saying don't ship it. I'm flagging that it's also a live production behavior change for any org that already has such a rule configured, which is worth a deliberate check (or at least a release note) rather than shipping silently alongside the "trace labels" framing in the PR title/description, which undersells this half of the blast radius.

**Suggested fix:** Before merging, spot-check whether any org has an existing cache rule with a non-empty `vk_tags` matcher, and if so, give them a heads-up that it will start applying. At minimum, mention this behavior change in the release notes rather than only describing gateway traffic as "now filterable".

langwatch/src/server/gateway/config.materialiser.ts
```diff
@@ -204,6 +208,7 @@ export class GatewayConfigMaterialiser {
       })),
       cache_rules: cacheRules.map(cacheRuleToWire),
       metadata: config.metadata ?? {},
+      vk_tags: config.metadata?.tags ?? [],
```
Aryansharma28, line 230: [P3] `config.metadata.tags` has no size bound, and this line is what turns that into a live, per-request customer-facing attribute.

**Why:** `virtualKeyConfigSchema.metadata.tags` (`langwatch/src/server/gateway/virtualKey.config.ts`) is `z.array(z.string()).default([])` — no `.max()` on array length or per-string length. The edit drawer's tag input (`VirtualKeyEditDrawer.tsx`) is a plain `Input` with no `maxLength`, unlike the `name` field two lines above it which has `maxLength={128}`. This was harmless while tags were dead (this PR's own issue: `#5993`). After this PR, every one of these tags gets stamped as an OTel `StringSlice` attribute on *every* span this VK produces, and lands in ClickHouse's `metadata.labels` — which the Trace Explorer's Label filter and Analytics sidebar both read/aggregate. An org that pastes a very long string or a huge tag list into the field has no client- or server-side guard stopping it, and the cost is now paid on every gateway request rather than once.

**Suggested fix:** Add a `.max(N)` on the tags array and a per-tag `.max(len)` in `virtualKeyConfigSchema`, and mirror it with `maxLength` on the drawer's tag input. Low cost, and directly triggered by this PR making the field consequential for the first time.

services/aigateway/adapters/customertracebridge/emitter.go
```diff
@@ -192,6 +192,12 @@ func (e *Emitter) EndSpan(ctx context.Context, params domain.AITraceParams) {
 	if params.GatewayRequestID != "" {
 		attrs = append(attrs, attribute.String(gatewaytracer.AttrGatewayReqID, params.GatewayRequestID))
 	}
+	// VK tags become the trace's labels: the ingestion pipeline maps the
+	// langwatch.labels attribute into metadata.labels, the field the Trace
+	// Explorer filters as "Label".
+	if len(params.VKTags) > 0 {
+		attrs = append(attrs, attribute.StringSlice(gatewaytracer.AttrLabels, params.VKTags))
```
Aryansharma28: [P5] Good call reusing the existing `langwatch.labels` attribute instead of inventing a new one.

**Why:** I grepped for `"langwatch.labels"` across the repo and it's already a first-class, heavily-used key: the TS/JS SDK exports it as `ATTR_LANGWATCH_LABELS` (`typescript-sdk/src/observability-sdk/semconv/attributes.ts:97`), and the ingestion/analytics stack already has full support for it — `otel.traces.ts:1184`, `trace-attribute-accumulation.service.ts` (with a real union-merge test suite), `traceAnalytics.foldProjection.ts`, ClickHouse filter translators, and the Trace Explorer drawer utils. That's a strong existing-test safety net for this feature landing correctly once the rebase above is done, and it's exactly why the PR can truthfully claim zero explorer-side changes.

Aryansharma28 (commented):
## goated-review verdict: **COMMENT** — sound, mostly well-tested fix, but currently un-mergeable and has one real coverage gap in the middle hop. Author is an external contributor (fork PR, `author_association: CONTRIBUTOR`), so this review applied elevated scrutiny to security/data-handling/supply-chain rather than extending default trust.

The core change is small: three hops (materialiser → Go wire → customer-span emitter) to finally light up a `BundleConfig.VKTags` field whose own comment admitted it was dead code, plus a 4th "glue" hop (`app/pipeline/trace.go`) threading it from `Bundle.Config` into the trace-params struct.

**What I actually executed, not just read** (checked out `pr-5995` at `227ef93` into a worktree):
- `go build ./...` and `go vet ./...` for the whole `services/aigateway` module — clean.
- `go test ./...` for the **entire** `aigateway` module (not just the 3 touched packages) — all packages pass, no regressions elsewhere.
- `gofmt -l .` — flags exactly one file (see P4 note below).
- Could **not** execute the TS integration test (`config.materialiser.integration.test.ts`) — this sandbox has no Docker daemon reachable and no `node_modules` installed for the monorepo, and standing up Postgres + a full `pnpm install` was outside a reasonable time budget for this review. I verified that code path by tracing types instead (`VirtualKeyConfig` zod inference, matching an existing sibling test in the same file) — flagging this explicitly as a real gap in my own verification, not papering over it.
- Elevated-scrutiny security pass (external contributor): confirmed **zero** dependency-manifest/lockfile changes anywhere in the diff (no `go.mod`, `go.sum`, `package.json`, `pnpm-lock.yaml` touched — no supply-chain surface at all); confirmed no `.github/workflows/*` or CI-gate files touched; confirmed no auth/authz/tenant-scoping code touched (VK tags are org-admin-set data read back to the same org, not attacker-controlled input from outside the tenant); grepped the Trace Explorer label-rendering path for `dangerouslySetInnerHTML` — none found, labels render as plain JSX text, so no stored-XSS vector via tag content.

**P0: 1 | P1: 0 | P2: 2 | P3: 1 | P4: 1 | P5: 2**

### P0 BLOCKER (1)
- **PR is not mergeable — GitHub reports `mergeable: false` / `mergeable_state: dirty`.** This isn't a flake: the PR branch (1 commit, based on `c62713f1b`) predates `#5990` ("fix(otel): end-to-end telemetry correctness and customer/internal trace isolation"), which landed on `main` the day after this PR was opened and moved the *entire* `customertracebridge` package from `services/aigateway/adapters/customertracebridge/` to `pkg/customertracebridge/`, plus trimmed/restructured `gatewaytracer/attrs.go` (removed `AttrOrigin`, `AttrVirtualKeyID`, `AttrGatewayReqID` from it — they now live in the new `pkg/customertracebridge/attrs.go`, with `gatewaytracer` importing them one-directionally). Two inline comments below pin the exact conflict points. This needs an author (or maintainer) to manually re-home the third hop into the new package layout — a straight `git merge`/"Update branch" will not resolve it automatically.

### P2 MEDIUM (2)
- **Missing coverage for the pipeline "glue" hop.** `app/pipeline/trace.go` adds `VKTags: call.Bundle.Config.VKTags` at 4 separate call sites (lines 71, 90, 120, 271 — sync success, sync fallback/error, per-provider retry, and the streaming `onClose`). None of the existing pipeline tests (`trace_body_test.go`, `trace_error_test.go`, `trace_project_test.go`) assert on this field, and no new test was added for it either — only the two endpoints of the pipe are tested (materialiser→wire, and emitter given an already-populated `AITraceParams`), not this connecting hop. The codebase already has the right pattern to copy: `trace_project_test.go`'s `TestTraceBeginsSyncAndStreamSpansUnderConfigProject` builds a `Bundle{Config: ...}`, runs it through `Trace()`'s sync/stream paths, and asserts the field surfaces correctly — the exact shape needed here for `VKTags`, at effectively zero marginal cost. This is a real gap, not a nice-to-have: it's the one hop where a future refactor could silently drop the field (e.g. miss one of the 4 sites, or typo `Bundle.VirtualKeyID` for `Bundle.Config.VKTags`) with nothing red.
- Making `Bundle.Config.VKTags` real also makes the cache-rule `vk_tags` matcher real — which was previously **always a no-op** (`matchesAllTags` against a permanently-empty slice never matches when the rule specifies tags). The rule-builder UI (`cacheRule.form.tsx`) already lets customers configure `vk_tags` matchers today, so any org that already saved a cache rule with a `vk_tags` matcher — believing it had no effect — will have that rule *start actually applying* (force/disable/ttl override) the moment this ships, with no feature flag, migration note, or customer communication mentioned anywhere in the PR. Worth a deliberate check before merge (see inline comment on `domain/bundle.go`).

### P3 LOW (1)
- `config.metadata.tags` has no upper bound (array length or per-string length) in `virtualKeyConfigSchema` (`z.array(z.string()).default([])`), and the drawer's tag `Input` has no `maxLength` either (unlike the `name` field, which does). That was harmless while tags were dead; this PR is exactly what turns it into a live, per-request OTel attribute and a ClickHouse-filterable label set. Recommend a small `.max()` on tag count and per-tag length as a cheap follow-up. (Same-tenant data, not an attacker-controlled input — this is a hygiene/cost-control gap, not a vulnerability.)

### P4 NIT (1)
- `services/aigateway/adapters/gatewaytracer/attrs.go` as submitted isn't `gofmt`-clean (`gofmt -l`/`gofmt -d` both flag it — the inserted comment breaks the const block's alignment group without realigning the surrounding `=` columns). Not caught by CI yet since no `gofmt` check exists in `go-ci` and workflows haven't been approved to run on this fork PR. Moot in isolation since this file needs rewriting for the P0 rebase anyway — mentioned so it isn't silently carried into wherever the const lands next.

### P5 PRAISE / OBSERVATIONS (2)
- Nice due-diligence: reusing the pre-existing `langwatch.labels` attribute key instead of inventing a new one is exactly right, and I verified it's already wired through `otel.traces.ts`, `trace-attribute-accumulation.service.ts`, ClickHouse filter translators, and the TS SDK's own `ATTR_LANGWATCH_LABELS` constant — dozens of existing call sites and tests already exercise this exact key, so downstream compatibility risk is low.
- The tests that *do* exist are real, not superficial: the Go unit tests assert actual span attributes via a real `SpanRecorder` (not mocks) for both the tagged and untagged case, and the wire-decode test round-trips real JSON. I confirmed all of them pass by executing the full `aigateway` module's test suite against the PR's own commit, not just reading them.

### Existing signals I'm not re-raising
- CodeRabbit: "No actionable comments were generated" — consistent with my read; my findings above are things a diff-only reviewer (CodeRabbit, `langwatch-agent`) can't see, since they require checking the PR against current `main` and actually executing the suite, not just the diff.
- `langwatch-agent` (static external-contributor scan): flagged nothing code-level, and explicitly noted "merge remains contingent on the required CI checks passing" — which is doubly true here since CI hasn't actually *run* yet (all checks show `action_required`, the standard gate for a first-time external contributor's workflows), and separately the branch is behind `main` in a way CI alone won't catch (a conflicting merge, not a failing test).

### Things I couldn't verify
- Whether any live organization currently has a saved cache rule with a `vk_tags` matcher (would need production DB access) — this is the crux of one P2 finding above; flagging it as a question for whoever merges, not a confirmed incident.
- The PR description's "Verified end-to-end on a live deployment" claim — no access to that deployment to confirm independently.
- The TS integration test (`config.materialiser.integration.test.ts`) — no Postgres/Docker or installed `node_modules` in this environment; verified by type-tracing rather than execution (see above). Whichever CI run eventually executes for this PR should confirm it green before merge.

Aryansharma28:
Hey @thiagolmoraes — thanks for this one, left a full review above. One thing worth a look when you get a chance: GitHub is currently showing this PR as unable to merge due to a real conflict with #5990, which moved the `customertracebridge` package after this PR opened, right around where you're stamping the `langwatch.labels` attribute. A rebase/re-home of that piece should unblock it. No rush, just flagging early so it doesn't sit stale.

langwatch-agent (dismissed):

rogeriochaves:
## Review: the approach is right, and it is now mergeable + bounded

The three-hop design is the correct one, and reusing `langwatch.labels` instead of inventing a key is what makes "zero explorer-side changes" actually true. I reviewed it hard, fixed what needed fixing, and dogfooded it end to end on a real stack against a real provider. Fixes are pushed on top of the author's work (`b89076b` merge + `c90c14d`), no force-push.

### What I found and fixed

**1. The PR could not merge.** `config_wire.go` conflicted with `main`: `#6122`-era `main` added `LangyMirrorTier` at the end of the same struct where this PR adds `VKTags`. Resolved by keeping both. GitHub now reports `mergeable: true`.

**2. Unbounded tag cardinality (the one that mattered).** `virtualKeyConfigSchema.metadata.tags` was `z.array(z.string()).default([])`: no cap on count, no cap on per-tag length, no de-duplication, no blank filtering. That was harmless while tags were dead code. This PR turns every tag into a per-request OTel attribute that lands in `trace_summaries.Attributes['langwatch.labels']` and gets `arrayJoin`ed by the Label facet, so the tag list became a live cardinality surface. It is also not an "operators are sensible" problem: the gateway-platform REST API takes `virtualKeyConfigSchema.partial()`, so the same field is reachable from an unvalidated request body, not only from the drawer.

Fixed with `normalizeVkTags` in `virtualKey.config.ts`: trimmed, blanks dropped, de-duplicated, capped at 32 tags of 128 characters each, sliced by code point so truncation can never split a surrogate pair into lone halves that serialise as invalid UTF-8 downstream.

It is deliberately a **normalisation, not a rejection**. `parseVirtualKeyConfig` also runs on the config-fetch path the gateway depends on, so a `.max()` that throws would turn a stored over-size key into a 500 on config fetch and take down that VK's traffic entirely. Both drawers now carry a matching `maxLength` and copy that states the caps.

**3. The middle hop had no test.** `app/pipeline/trace.go` assigns `VKTags` at four independent exits (sync success, sync failure, stream-establish failure, stream close) and nothing drove a `Call` through `Trace()` to assert it. Added `trace_vk_tags_test.go` covering all four. I mutation-checked it: deleting the assignments makes it fail, so it is not vacuous.

### Answers to the questions I set out to check

- **Do tags actually reach the stored span, or just the emit path?** They reach it. Verified in ClickHouse directly, not inferred: the labels are present in **both** `stored_spans.SpanAttributes` and `trace_summaries.Attributes`, so the trace-fold projection does fire for this path.
- **Can they collide with or overwrite our own attributes?** No. `langwatch.labels` is a pre-existing first-class key, and `TraceAttributeAccumulationService` **unions** labels across spans rather than overwriting, so VK tags merge with a customer's own labels instead of clobbering them. Worth knowing for the release note: if a customer propagates `traceparent` through the gateway, the gateway span joins their trace and the operator's VK tags union onto that trace's label set.
- **PII / secret risk.** Tags are org-admin-set and never end-user input, and they do not travel to the provider. They do widen exposure: anything typed into Tags is now visible on every trace in the project and enumerable through the Label facet. The fan-out is contained though: `AttrLabels` is written in exactly one place and never reaches the gateway's own operational tracer (`gatewaytracer`), so VK tags do not leak into LangWatch's internal telemetry.
- **Empty, duplicated, enormous tags.** All covered by `normalizeVkTags` and pinned by tests, including a real-PG case where the stored row carries 500 tags, a 5000-character tag, a duplicate and a blank.
- **ClickHouse rules.** This PR adds no queries. The read paths it lights up already exist, and both the Label facet and the `metadata.labels` filter already put `TenantId` first with a partition-key range on `OccurredAt`.

### Dogfooding, end to end

Real stack (`pnpm dev`, gateway auto-started on :5583), real OpenAI key, real `gpt-5-mini` responses.

1. Minted a VK through the real API with a deliberately dirty tag list, `["app=acme-support", "team=platform", " team=platform ", "", "env=dogfood"]`. Stored as `["app=acme-support","team=platform","env=dogfood"]`: the padded duplicate and the blank were dropped on the way in.
2. Sent 3 chat completions through the gateway with that key, all HTTP 200 with real provider responses.
3. ClickHouse, both tables:
   - `stored_spans.SpanAttributes['langwatch.labels']` = `["app=acme-support","team=platform","env=dogfood"]`
   - `trace_summaries.Attributes['langwatch.labels']` = same
4. Ran the Explorer's own SQL verbatim. The Label facet returns `app=acme-support`, `team=platform`, `env=dogfood` with counts; the `metadata.labels` filter selects exactly the 3 tagged traces; a label nobody has returns 0.
5. In the Trace Explorer UI the labels render as chips in the LABELS column on the gateway rows, and clicking one narrows the list from 14 traces to exactly 3.

**Trace Explorer, filtered by the VK tag** (14 traces to 3, Origin `Gateway`):

filtered by vk tag

**The LABELS column carrying all three tags, with the `label:"app=acme-support"` filter applied:**

labels column

**The tags field with the new bounds and copy:**

vk drawer

### Deliberately not changed

- **Cache-rule activation.** Flipping `VKTags` from permanently-empty to real does silently activate any `vk_tags` matcher an org already saved, which has been a no-op until now. That is exactly what #5993 asks for, and the new bounds make the matched value set predictable, so I left the behavior alone. It deserves a release note rather than a code change, since the PR title only advertises the trace-labels half of the blast radius.
- **The label filter's quoting.** `filter-conditions.ts` builds `concat('"', x, '"')` and compares it against `JSONExtractArrayRaw` output, so a label containing a `"` or a backslash will not match its own JSON-escaped raw form. It is pre-existing and affects every label source including the SDK, not VK tags specifically, so it does not belong in this PR.
- **#6139 / #6140 / #6141 / #6142** (guardrail check stub, deleted gateway Prometheus metrics, budget accrual, coding-agents onboarding). This PR does not interact with any of them.

### Verification

- `gofmt` clean, `go build ./...` and `go vet ./...` clean.
- Full `services/aigateway` Go suite passes, plus `pkg/customertracebridge`.
- `pnpm typecheck` clean.
- `pnpm test:unit src/server/gateway/__tests__/virtualKey.tags.unit.test.ts`: 10 passed.
- `pnpm test:integration src/server/gateway/__tests__/config.materialiser.integration.test.ts`: 13 passed against real native Postgres.

One note on CI: every workflow was sitting at `action_required` because this is a fork PR. I confirmed the diff touches no `.github/`, no lockfile and no dependency manifest, then approved the runs.

services/aigateway/adapters/gatewaytracer/attrs.go
rogeriochaves: Resolved by the merge with `main` (`b89076b`). `AttrLabels` now lives in `pkg/customertracebridge/attrs.go` alongside `AttrVirtualKeyID` / `AttrGatewayReqID`, matching the one-directional import convention #5990 established, and `gofmt -l` is clean on the whole module. Nothing is added to `gatewaytracer/attrs.go` any more, so VK tags never reach our own operational tracer.

pkg/customertracebridge/emitter.go
rogeriochaves: Resolved by the merge with `main` (`b89076b`). The stamping now sits in `pkg/customertracebridge/emitter.go` at the new path, and `vk_tags_span_test.go` moved alongside it and reuses the `recordSpanForParams` helper that already lives there. The only real conflict left after the move was `config_wire.go`, where `main` had added `LangyMirrorTier` to the same struct; both fields are kept. GitHub reports the PR mergeable.

services/aigateway/app/pipeline/trace.go
rogeriochaves: Agreed, and fixed in `c90c14d`: `services/aigateway/app/pipeline/trace_vk_tags_test.go` drives a `Call` through `Trace()` and asserts `VKTags` reaches `AITraceParams` at all four exits (sync success, sync failure, stream-establish failure, stream close). It was mutation-checked: deleting the assignments makes it fail, so it is not vacuous.

services/aigateway/domain/bundle.go
rogeriochaves: Right call to flag it, and we are shipping it deliberately rather than changing the code: making the matcher real is exactly what #5993 asks for, and the new tag bounds make the matched value set predictable rather than unbounded. What was missing was the disclosure, so the PR body now carries it under Deployment Impact: any cache rule already saved with a `vk_tags` matcher has been a silent no-op and starts applying on upgrade, with a recommendation to review existing rules first.

langwatch/src/server/gateway/config.materialiser.ts
rogeriochaves: Fixed in `c90c14d`. `normalizeVkTags` in `virtualKey.config.ts` trims, drops blanks, de-duplicates and caps at 32 tags of 128 characters, slicing by code point so truncation can never split a surrogate pair. It normalises rather than rejecting on purpose: `parseVirtualKeyConfig` also runs on the config-fetch path the gateway depends on, so a `.max()` that throws would turn an already-stored over-size key into a 500 on config fetch and take that VK offline. Worth noting the field is reachable from the gateway-platform REST body via `virtualKeyConfigSchema.partial()`, not only from the drawer, which is why the bound sits in the schema rather than only in the UI.

langwatch/src/components/gateway/VirtualKeyCreateDrawer.tsx
```diff
@@ -29,11 +29,20 @@ import {
...
+} from "~/server/gateway/virtualKey.config";
 import {
   VirtualKeyScopePicker,
   type VirtualKeyScopeEntry,
 } from "./VirtualKeyScopePicker";
 
+// The field is one comma-separated line, so its own cap is the tag caps plus
+// a separator per tag. Anything past this is trimmed by normalizeVkTags on
+// save regardless; the input cap just stops a runaway paste getting that far.
+const TAGS_CSV_MAX_LENGTH = VK_TAGS_MAX_COUNT * (VK_TAG_MAX_LENGTH + 2);
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in `6342832`. You were right that the two were measured differently: `maxLength` counts UTF-16 code units while `normalizeVkTags` slices by code point, so a full list of astral-plane tags could be clipped by the input before the server ever saw it. Rather than reimplement code-point slicing in an `onChange` (which would fight the user mid-typing), the field budget is now doubled per tag, which makes clipping impossible for any list the server would have kept whole. Both drawers import the one `TAGS_CSV_MAX_LENGTH` from the new shared `components/gateway/virtualKeyTagsField.ts`, so there is no second copy to drift, and `virtualKeyTagsField.unit.test.ts` pins the property with a 32-emoji-tag list at the per-tag limit.

langwatch/src/components/gateway/VirtualKeyCreateDrawer.tsx
```diff
@@ -217,17 +226,20 @@ export function VirtualKeyCreateDrawer({
...
                 value={tagsCsv}
                 onChange={(e) => setTagsCsv(e.target.value)}
                 placeholder="e.g. tier=enterprise, team=ml"
+                maxLength={TAGS_CSV_MAX_LENGTH}
               />
               <Field.HelperText>
-                Comma-separated. Cache-control rules match VKs on tags as
+                Comma-separated, up to {VK_TAGS_MAX_COUNT} tags of{" "}
+                {VK_TAG_MAX_LENGTH} characters. Tags become labels on this
+                key&apos;s traces, and cache-control rules match VKs on tags as
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in `6342832`, and thank you for pushing on it. Two changes. The explanation moved out of the permanent block under the field and into the label information icon, where every sibling field in the drawer already puts its explanation, and it now states the normalisation in plain terms: "Saving keeps the first 32 tags, trims each to 128 characters, and drops blanks and repeats." Separately, the field raises a notice while the typed list is actually over a cap, and only then: "Only the first 32 tags will be saved." and "Tags longer than 128 characters will be shortened." Repeats and blanks are deliberately not warned about, since dropping them costs nothing. Both drawers share one implementation so the behaviour and wording cannot diverge.

langwatch/src/components/gateway/VirtualKeyCreateDrawer.tsx
```diff
@@ -217,17 +226,20 @@ export function VirtualKeyCreateDrawer({
...
-                  description="Comma-separated tags attached to this VK. Cache-control rules match VKs on tags using AND-subset semantics — a rule matcher of ['tier=enterprise'] fires for any VK carrying that tag."
+                  description="Comma-separated tags attached to this VK. Every trace this key produces carries its tags as labels, so you can filter gateway traffic by team or app in the Trace Explorer. Cache-control rules also match VKs on tags using AND-subset semantics, so a rule matcher of ['tier=enterprise'] fires for any VK carrying that tag. Duplicates are dropped."
                   docHref="/ai-gateway/cache-control#cache-rules"
                 />
               </Field.Label>
               <Input
                 value={tagsCsv}
                 onChange={(e) => setTagsCsv(e.target.value)}
                 placeholder="e.g. tier=enterprise, team=ml"
+                maxLength={TAGS_CSV_MAX_LENGTH}
```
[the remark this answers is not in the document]
rogeriochaves: Done in `6342832`. Scenarios first, in `specs/ai-gateway/virtual-keys.feature` under a new Tags field section covering the explanation behind the information icon, the over-cap warning, and the quiet path. Then `VirtualKeyTagsField.integration.test.tsx`, which renders the real create drawer and the real popover (no shallow rendering, no stubbing of the component under test) and asserts the copy is invisible until the icon is opened, plus `virtualKeyTagsField.unit.test.ts` covering the count and per-tag limits, code-point measurement, and the copy pinned to the exported constants. Reverting the change fails 8 of the 10 RTL cases, so they are not vacuous.

rogeriochaves:
## Round 2, and a reply to the blocking review

@langwatch-agent's `CHANGES_REQUESTED` asks for the affected suites to run on a maintainer-controlled workflow rather than being taken on trust from a fork. That is the right ask and it is now satisfied rather than waived: I re-checked that the diff touches no `.github/` workflow, no lockfile and no dependency manifest (`go.mod`, `go.sum`, `package.json`, `pnpm-lock.yaml` are all untouched, full file list is 20 files under `langwatch/src`, `pkg/`, `services/aigateway/` and `specs/`), then approved every run on the current head so `langwatch-app-ci`, `go-ci`, `go-services`, `gateway-matrix`, `e2e-ci` and the rest execute for real. I have deliberately not dismissed the review; that is the maintainer's call once the runs report.

### What changed in this round

The founder asked for the Tags field's explanation to move out of the block under the input and into the label's `(i)`, where every sibling field in that drawer already keeps its explanation. Doing that honestly meant rewriting the copy, because the old text was written in gateway vocabulary rather than the customer's:

**Before** (under the field, and again in the `(i)`): "Comma-separated, up to 32 tags of 128 characters. Tags become labels on this key's traces, and cache-control rules match VKs on tags as AND-subset."

**After** (behind the `(i)` only):

> Group this key's traffic by team, app, or environment. Every trace this key sends carries its tags as labels, so anyone with access to the project can see them and filter on them. A cache rule that lists tags applies to any key carrying all of them. Saving keeps the first 32 tags, trims each to 128 characters, and drops blanks and repeats.

"VKs" and "AND-subset" are gone, "comma-separated" went back to being the placeholder's job (`e.g. tier=enterprise, team=ml`), and the second sentence is the consequence this PR actually introduces. It belongs at the moment someone types a tag, not in a release note.

The caps needed more than a mention. Typing a 33rd tag used to lose it with nothing said anywhere, so the field now raises a notice while the typed list is over a cap, and only then. That also closes CodeRabbit's "prevent silent tag loss" comment without putting a permanent block of text back under the field.

Details, the shared module, the `maxLength` code-point fix and the mutation-check evidence are in the PR description under "Round 2". Every open review thread now has an in-thread reply.

### Verification

- `pnpm typecheck` clean.
- `pnpm test:unit src/components/gateway`: 67 passed. `pnpm test:integration src/components/gateway`: 19 passed (real native Postgres, no containers).
- `pnpm test:unit src/server/gateway/__tests__/virtualKey.tags.unit.test.ts`: 10 passed. `pnpm test:integration src/server/gateway/__tests__/config.materialiser.integration.test.ts`: 13 passed.
- Mutation-checked both new suites. Restoring the old copy and the permanent helper block fails 8 of the 10 RTL cases; reverting the code-point fix fails 2 of the 14 unit cases.
- Dogfooded in a real browser on a real stack. Created a key through the drawer with a deliberately dirty tag list, `"app=acme-support, team=platform,  team=platform ,, env=dogfood"`, and the stored row came back as `["app=acme-support","team=platform","env=dogfood"]`: the padded duplicate and the blank dropped exactly as the `(i)` promises. Checked desktop light, desktop dark, 390px wide, and the edit drawer.

rogeriochaves:
## Merged `main` after #6143, resolved deliberately

#6143 reworked the gateway pipeline under this branch, so I merged `main` rather than waiting to discover it as a red run. Three resolutions:

- **`app/pipeline/trace.go`.** `Meta` became an accumulator (`*MetaAccumulator`) and `main` hoisted `call.Meta.GatewayRequestID()` into a local. Kept `main`'s `gatewayRequestID` at all four exits alongside this branch's `VKTags`.
- **`adapters/controlplane/config_wire_test.go`.** Both sides appended a test to the end of the same file. Kept both: `TestConfigWire_VKTags` and `TestBuildGuardrails_FailureModeReachesTheDataPlane`.
- **`app/pipeline/trace_vk_tags_test.go`.** This one merged with no conflict marker at all and did not compile: it builds a `Call` with `Meta: &Meta{}`, and that field is now `*MetaAccumulator`. `go vet` caught it, a clean textual merge would not have. Updated to match the sibling tests in the package.

Verified after resolving, not inferred from the absence of conflict markers:

- `gofmt -l` clean; `go build` and `go vet` clean over CI's exact target set (`./services/aigateway/... ./services/nlpgo/... ./pkg/... ./cmd/... ./tools/migrationorder/...`).
- `golangci-lint` **v2.11.4**, the version CI pins, on that same target set: **0 issues**.
- Full Go suite green: 37 packages ok, including `TestTracePassesVKTagsToEndParams` (all four exits), `TestTraceLeavesVKTagsEmptyForUntaggedKey`, `TestEmitter_VKTags_StampedAsLabels`, `TestEmitter_NoVKTags_NoLabelsAttribute` and `TestConfigWire_VKTags`.
- TS side: `pnpm typecheck` clean, `pnpm check:feature-parity` exit 0, gateway unit suites 169 passed / 17 files, gateway integration suites 97 passed / 11 files.

On the UI half, #6143's `TriggerAnchor` prop-forwarding fix is now on this branch, but the Tags field does not need it: the drawer's established pattern is `FieldInfoTooltip`, a `Popover` rather than a `Tooltip`, which is what the Name and Routing policy fields already use. Nothing here goes near the `Tooltip` + `asChild` nesting that bug lived in.

Also worth flagging for whoever owns it, not something I touched: #6143 committed `.claude/pr-body-current.md` and `.claude/pr-body-new.md` to `main`. They arrived here through the merge.

rogeriochaves (approved):

PR #5996, opened 2026-07-20
feat(analytics): cost comparison card, estimated savings vs a reference model

### LINKED ISSUE #5994: Analytics: no way to compare actual spend against a reference model (show the value of self-hosted / cheaper models)
## Problem

Teams running self-hosted models (vLLM, Ollama) or cheaper commercial models have no way to demonstrate the value of that choice inside LangWatch. The analytics pages show actual cost — which for self-hosted traffic is $0 or near-zero — but never answer the question stakeholders actually ask: **"what would this traffic have cost on model X?"**

All the ingredients already exist: the analytics timeseries endpoint returns filtered `prompt_tokens` / `completion_tokens` / `total_cost` sums, and the model catalog carries per-token pricing for every known model (`inputCostPerToken` / `outputCostPerToken`, already exposed to the frontend via `getAllForProjectForFrontend`). Custom graphs cannot express it either — metrics are a fixed enum, with no way to multiply a token sum by a constant price.

## Proposed solution

A "Cost Comparison" card on the LLM Metrics analytics page:

- The user picks a **reference model** (catalog-priced chat models only; models without pricing are not offered).
- The card shows, side by side: the period's **actual cost**, the **estimated cost** of the same traffic on the reference model (`input_tokens x ref input price + output_tokens x ref output price`, using the period's real token counts), and the **estimated savings** (estimate minus actual; a negative difference renders as additional cost, not savings).
- The computation honors the page's existing filter sidebar and date range, so the comparison can be sliced by label, model, or user — e.g. "what did app X save this month by running on the local model instead of Claude Sonnet".
- Cost-permission gated like the other cost cards; zero-traffic periods get an empty state.

Implementation is intentionally thin: one pure calc module + a card component reusing `ModelSelector`, `SummaryMetric`, `ChartCard`, and a single `getTimeseries` summary query — no backend changes. PR to follow.

thiagolmoraes:
Fixes #5994

## What

A "Cost Comparison" card on the LLM Metrics analytics page: pick a reference model and see the period's **actual cost**, the **estimated cost** of the same traffic on the reference model, and the **estimated savings**, the number teams running self-hosted or cheaper models need to demonstrate the value of that choice ("what would this month have cost on Claude Sonnet?").

## How

Intentionally thin, no backend changes:

- One `getTimeseries` summary query (`timeScale: "full"`) for the period's token and cost sums, spreading `useFilterParams()` so the card honors the existing filter sidebar and date range (sliceable by label, model, user).
- The estimate prices **every token bucket the product bills**: fresh input, output, cache read and cache write. LangWatch records cached input separately from fresh input (`specs/ai-gateway/cache-token-telemetry.feature`) and the recorded cost prices all four buckets (`estimateCost` in `src/server/tracer/collector/cost.ts`), so the estimate uses the same formula. Cache rates fall back to the plain input rate when the catalog publishes none, exactly as the recorded cost does, which keeps both sides of the comparison on the same convention.
- Reference pricing comes from the model catalog already exposed to the frontend. A model is only offered, and only priced, when the catalog publishes a usable price: both rates present and at least one above zero, matching the guard the recorded cost applies. An all-zero row is a stand-in for "no published price", which is what every custom and self-hosted model carries and what a handful of catalog rows carry too.
- UI reuses `ModelSelector`, `SummaryMetric` and `ChartCard`. Negative differences render as additional cost, zero-traffic periods get an empty state, a model that cannot be priced gets a dash and a plain explanation, and the card is `cost:view` gated like the other cost cards.
- Series values are read back by metric name through the shared `readSummaryMetric`, so reordering the request cannot silently swap two numbers.

## Verification

Dogfooded against real local traffic (234 traces, 30 day window), with the displayed figures recomputed independently from the underlying ClickHouse rows:

| | ClickHouse | Card |
|---|---|---|
| fresh input tokens | 5,353,935 | 5.4m |
| output tokens | 52,262 | 52.3k |
| cache read + write tokens | 3,100,637 | 3.1m |
| recorded cost | $13.804739 | $13.80 |
| estimate on `claude-sonnet-4-6` | $18.734854 | $18.73 |
| savings | $4.930115 | $4.93 |

Sliced to a single model (`?model=claude-fable-5`), where the reference is cheaper than actual: recorded $4.857727, estimate $1.498636, savings -$3.359091, all matching the card's $4.86 / $1.50 / -$3.36.

## Screenshots

Populated card

populated

In page context

metrics page

Sliced by model, reference cheaper than actual

single model

No traffic in the period

empty

Usage still loading

loading

Reference model with no published price

no published price

## Tests

- Unit tests on the calc module: per bucket pricing against real catalog rates, cache rate fallback, incomplete and all-zero pricing rejected from both the estimate and the selector, zero-token periods, option sorting.
- React Testing Library tests that render the real card against realistic `getTimeseries` responses and assert the dollar figure on screen, plus the cached, zero-cost, no-price, loading, empty and error states. Bucket keys are built from the series the card actually requests, so the test follows a reordering instead of pinning it.
- Spec scenarios in `specs/analytics/model-cost-comparison.feature`.

### DIFF
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -21,10 +21,13 @@ const SERIES = [
 const seriesKey = (index: number) =>
   `${index}/${SERIES[index]!.metric}/${SERIES[index]!.aggregation}`;
 
-const sumSeries = (
-  buckets: Record<string, unknown>[] | undefined,
-  key: string,
-): number =>
+const sumSeries = ({
+  buckets,
+  key,
+}: {
+  buckets: Record<string, unknown>[] | undefined;
+  key: string;
+}): number =>
   (buckets ?? []).reduce((total, bucket) => {
     const value = bucket[key];
     return total + (typeof value === "number" ? value : 0);
@@ -71,12 +74,18 @@ export function ModelCostComparisonCard() {
     queryOpts,
   );
 
-  const promptTokens = sumSeries(timeseries.data?.currentPeriod, seriesKey(0));
-  const completionTokens = sumSeries(
-    timeseries.data?.currentPeriod,
-    seriesKey(1),
-  );
-  const actualCost = sumSeries(timeseries.data?.currentPeriod, seriesKey(2));
+  const promptTokens = sumSeries({
+    buckets: timeseries.data?.currentPeriod,
+    key: seriesKey(0),
+  });
+  const completionTokens = sumSeries({
+    buckets: timeseries.data?.currentPeriod,
+    key: seriesKey(1),
+  });
+  const actualCost = sumSeries({
+    buckets: timeseries.data?.currentPeriod,
+    key: seriesKey(2),
+  });
 
   const pricing = selectedModel
     ? modelMetadata?.[selectedModel]?.pricing
@@ -108,7 +117,12 @@ export function ModelCostComparisonCard() {
           )}
         </Box>
       </HStack>
-      {!timeseries.isLoading && !hasTraffic ? (
+      {timeseries.isError ? (
+        <Text textStyle="sm" color="fg.error">
+          Could not load usage for the selected period. Try refreshing the
+          page.
+        </Text>
+      ) : !timeseries.isLoading && !hasTraffic ? (
         <Text textStyle="sm" color="fg.subtle">
           No traffic in the selected period. Adjust the filters or date range
           to compare costs.
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -1,0 +1,152 @@
+import { Box, HStack, Text, VStack } from "@chakra-ui/react";
+import numeral from "numeral";
+import { useMemo, useState } from "react";
+import { useFilterParams } from "../../hooks/useFilterParams";
+import { useModelProvidersSettings } from "../../hooks/useModelProvidersSettings";
+import { useOrganizationTeamProject } from "../../hooks/useOrganizationTeamProject";
+import { api } from "../../utils/api";
+import { ModelSelector } from "../ModelSelector";
+import {
+  estimateReferenceCost,
+  referenceModelOptions,
+} from "./modelCostComparison";
+import { SummaryMetric } from "./SummaryMetric";
+
+const SERIES = [
+  { metric: "performance.prompt_tokens", aggregation: "sum" },
+  { metric: "performance.completion_tokens", aggregation: "sum" },
+  { metric: "performance.total_cost", aggregation: "sum" },
+] as const;
+
+const seriesKey = (index: number) =>
+  `${index}/${SERIES[index]!.metric}/${SERIES[index]!.aggregation}`;
+
+const sumSeries = (
+  buckets: Record<string, unknown>[] | undefined,
+  key: string,
+): number =>
+  (buckets ?? []).reduce((total, bucket) => {
+    const value = bucket[key];
+    return total + (typeof value === "number" ? value : 0);
+  }, 0);
+
+const money = (value: number) =>
+  value >= 0.01 || value === 0
+    ? numeral(value).format("$0.00a")
+    : numeral(value).format("$0.0000a");
+
+const DEFAULT_REFERENCE = "anthropic/claude-sonnet-4-6";
+
+/**
+ * Compares the period's actual spend with what the same traffic would have
+ * cost on a reference model, using the period's real token counts priced
+ * at the reference model's catalog rates. Honors the page's filters and
+ * date range, so the comparison can be sliced by label, model, user, etc.
+ */
+export function ModelCostComparisonCard() {
+  const { project } = useOrganizationTeamProject();
+  const { filterParams, queryOpts } = useFilterParams();
+  const { modelMetadata } = useModelProvidersSettings({
+    projectId: project?.id,
+  });
+
+  const options = useMemo(
+    () => referenceModelOptions(modelMetadata),
+    [modelMetadata],
+  );
+  const [referenceModel, setReferenceModel] = useState<string | undefined>(
+    undefined,
+  );
+  const selectedModel =
+    referenceModel ??
+    (options.includes(DEFAULT_REFERENCE) ? DEFAULT_REFERENCE : options[0]);
+
+  const timeseries = api.analytics.getTimeseries.useQuery(
+    {
+      ...filterParams,
+      series: SERIES.map((s) => ({ ...s })),
+      timeScale: "full",
+      timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+    },
+    queryOpts,
+  );
+
+  const promptTokens = sumSeries(timeseries.data?.currentPeriod, seriesKey(0));
+  const completionTokens = sumSeries(
+    timeseries.data?.currentPeriod,
+    seriesKey(1),
+  );
+  const actualCost = sumSeries(timeseries.data?.currentPeriod, seriesKey(2));
+
+  const pricing = selectedModel
+    ? modelMetadata?.[selectedModel]?.pricing
+    : undefined;
+  const referenceCost = estimateReferenceCost({
+    promptTokens,
+    completionTokens,
+    pricing,
+  });
+  const savings =
+    referenceCost !== undefined ? referenceCost - actualCost : undefined;
+  const hasTraffic = promptTokens + completionTokens > 0;
+
+  return (
+    <VStack align="stretch" gap={4}>
+      <HStack justify="space-between" flexWrap="wrap" gap={2}>
+        <Text textStyle="sm" color="fg.muted">
+          What the same traffic would cost on
+        </Text>
+        <Box minWidth="220px">
+          {selectedModel && (
+            <ModelSelector
+              model={selectedModel}
+              options={options}
+              onChange={setReferenceModel}
+              size="sm"
+              mode="chat"
+            />
+          )}
+        </Box>
+      </HStack>
+      {!timeseries.isLoading && !hasTraffic ? (
+        <Text textStyle="sm" color="fg.subtle">
+          No traffic in the selected period. Adjust the filters or date range
+          to compare costs.
+        </Text>
+      ) : (
+        <HStack align="start" gap={0}>
+          <SummaryMetric
+            label="Current Cost"
+            current={timeseries.isLoading ? undefined : actualCost}
+            format={money}
+            increaseIs="neutral"
+          />
+          <SummaryMetric
+            label="Estimated Cost on Selected Model"
+            current={
+              timeseries.isLoading || referenceCost === undefined
+                ? undefined
+                : referenceCost
+            }
+            format={money}
+            increaseIs="neutral"
+            tooltip="The period's input and output tokens priced at the selected model's rates."
+          />
+          <SummaryMetric
+            label="Estimated Savings"
+            current={
+              timeseries.isLoading || savings === undefined
+                ? undefined
+                : savings
+            }
+            format={(value: number) =>
+              value < 0 ? `-${money(Math.abs(value))}` : money(value)
+            }
+            increaseIs="good"
+            tooltip="Estimated cost on the selected model minus the actual cost. Negative means the current setup costs more."
+          />
+        </HStack>
+      )}
+    </VStack>
+  );
+}
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -134,6 +134,7 @@ export function ModelCostComparisonCard() {
             current={timeseries.isLoading ? undefined : actualCost}
             format={money}
             increaseIs="neutral"
+            zeroMeansNoData={false}
           />
           <SummaryMetric
             label="Estimated Cost on Selected Model"
@@ -145,6 +146,7 @@ export function ModelCostComparisonCard() {
             format={money}
             increaseIs="neutral"
             tooltip="The period's input and output tokens priced at the selected model's rates."
+            zeroMeansNoData={false}
           />
           <SummaryMetric
             label="Estimated Savings"
@@ -158,6 +160,7 @@ export function ModelCostComparisonCard() {
             }
             increaseIs="good"
             tooltip="Estimated cost on the selected model minus the actual cost. Negative means the current setup costs more."
+            zeroMeansNoData={false}
           />
         </HStack>
       )}
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -49,13 +49,13 @@ const DEFAULT_REFERENCE = "anthropic/claude-sonnet-4-6";
 export function ModelCostComparisonCard() {
   const { project } = useOrganizationTeamProject();
   const { filterParams, queryOpts } = useFilterParams();
-  const { modelMetadata } = useModelProvidersSettings({
+  const { modelMetadata, providers } = useModelProvidersSettings({
     projectId: project?.id,
   });
 
   const options = useMemo(
-    () => referenceModelOptions(modelMetadata),
-    [modelMetadata],
+    () => referenceModelOptions({ modelMetadata, providers }),
+    [modelMetadata, providers],
   );
   const [referenceModel, setReferenceModel] = useState<string | undefined>(
     undefined,
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -106,6 +106,13 @@ export function ModelCostComparisonCard() {
     promptTokens + completionTokens + cacheReadTokens + cacheWriteTokens > 0;
   const cachedTokens = cacheReadTokens + cacheWriteTokens;
 
+  // Formatted up front so a value that cannot be estimated reads as a dash
+  // with an explanation underneath, rather than a skeleton that promises a
+  // number still on its way.
+  const asIs = (value: string) => value;
+  const cell = (value: number | undefined, format: (value: number) => string) =>
+    !isLoaded ? undefined : value === undefined ? "-" : format(value);
+
   return (
     <VStack align="stretch" gap={4}>
       <HStack justify="space-between" flexWrap="wrap" gap={2}>
@@ -138,24 +145,24 @@ export function ModelCostComparisonCard() {
           <HStack align="start" gap={0}>
             <SummaryMetric
               label="Actual Cost"
-              current={isLoaded ? actualCost : undefined}
-              format={money}
+              current={cell(actualCost, money)}
+              format={asIs}
               increaseIs="neutral"
               tooltip="What this period's traffic cost, across every model in it. Models with no published price, self-hosted ones included, are recorded at $0."
               zeroMeansNoData={false}
             />
             <SummaryMetric
               label="Estimated Cost"
-              current={isLoaded ? referenceCost : undefined}
-              format={money}
+              current={cell(referenceCost, money)}
+              format={asIs}
               increaseIs="neutral"
               tooltip="Every input, output and cached token recorded in this period, priced at the selected model's published rates."
               zeroMeansNoData={false}
             />
             <SummaryMetric
               label="Estimated Savings"
-              current={isLoaded ? savings : undefined}
-              format={signedMoney}
+              current={cell(savings, signedMoney)}
+              format={asIs}
               increaseIs="good"
               tooltip="Estimated cost on the selected model minus the actual cost. Negative means the current setup costs more."
               zeroMeansNoData={false}
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -7,6 +7,7 @@ import { useOrganizationTeamProject } from "../../hooks/useOrganizationTeamProje
 import type { SeriesInputType } from "../../server/analytics/registry";
 import { readSummaryMetric } from "../../server/analytics/timeseriesSummary";
 import { api } from "../../utils/api";
+import { formatMoney } from "../../utils/formatMoney";
 import { ModelSelector } from "../ModelSelector";
 import {
   estimateReferenceCost,
@@ -28,10 +29,10 @@ const SERIES: SeriesInputType[] = [
   { metric: "performance.total_cost", aggregation: "sum" },
 ];
 
-const money = (value: number) =>
-  value >= 0.01 || value === 0
-    ? numeral(value).format("$0.00a")
-    : numeral(value).format("$0.0000a");
+// The same formatter the cost metrics on this page use, so a four-figure
+// period reads as $1,043.27 rather than an abbreviated $1.04k. A number headed
+// for a finance conversation has to keep its digits.
+const money = (value: number) => formatMoney({ amount: value, currency: "USD" });
 
 const signedMoney = (value: number) =>
   value < 0 ? `-${money(Math.abs(value))}` : money(value);
@@ -110,8 +111,13 @@ export function ModelCostComparisonCard() {
   // with an explanation underneath, rather than a skeleton that promises a
   // number still on its way.
   const asIs = (value: string) => value;
-  const cell = (value: number | undefined, format: (value: number) => string) =>
-    !isLoaded ? undefined : value === undefined ? "-" : format(value);
+  const cell = ({
+    value,
+    format,
+  }: {
+    value: number | undefined;
+    format: (value: number) => string;
+  }) => (!isLoaded ? undefined : value === undefined ? "-" : format(value));
 
   return (
     <VStack align="stretch" gap={4}>
@@ -145,7 +151,7 @@ export function ModelCostComparisonCard() {
           <HStack align="start" gap={0}>
             <SummaryMetric
               label="Actual Cost"
-              current={cell(actualCost, money)}
+              current={cell({ value: actualCost, format: money })}
               format={asIs}
               increaseIs="neutral"
               tooltip="What this period's traffic cost, across every model in it. Models with no published price, self-hosted ones included, are recorded at $0."
@@ -153,7 +159,7 @@ export function ModelCostComparisonCard() {
             />
             <SummaryMetric
               label="Estimated Cost"
-              current={cell(referenceCost, money)}
+              current={cell({ value: referenceCost, format: money })}
               format={asIs}
               increaseIs="neutral"
               tooltip="Every input, output and cached token recorded in this period, priced at the selected model's published rates."
@@ -161,7 +167,7 @@ export function ModelCostComparisonCard() {
             />
             <SummaryMetric
               label="Estimated Savings"
-              current={cell(savings, signedMoney)}
+              current={cell({ value: savings, format: signedMoney })}
               format={asIs}
               increaseIs="good"
               tooltip="Estimated cost on the selected model minus the actual cost. Negative means the current setup costs more."
--- a/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
+++ b/langwatch/src/components/analytics/ModelCostComparisonCard.tsx
@@ -4,6 +4,8 @@ import { useMemo, useState } from "react";
 import { useFilterParams } from "../../hooks/useFilterParams";
 import { useModelProvidersSettings } from "../../hooks/useModelProvidersSettings";
 import { useOrganizationTeamProject } from "../../hooks/useOrganizationTeamProject";
+import type { SeriesInputType } from "../../server/analytics/registry";
+import { readSummaryMetric } from "../../server/analytics/timeseriesSummary";
 import { api } from "../../utils/api";
 import { ModelSelector } from "../ModelSelector";
 import {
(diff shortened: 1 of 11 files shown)


thiagolmoraes:
Re: the "savings calculation is inverted" finding — declining this one, the current formula is intentional.

`savings = referenceCost − actualCost` is the definition of savings: what the traffic *would have cost* on the reference model minus what it *actually cost*. The card's primary use case is a team on a self-hosted model (actual ≈ $0) picking a commercial reference: reference $13.50 − actual $0 = **+$13.50 saved**, rendered green. Flipping to `actualCost − referenceCost` would render a genuine saving as negative/red — the exact opposite of the card's purpose.

The accompanying spec pins this semantics (`specs/analytics/model-cost-comparison.feature`): *"the card shows the estimated savings (estimate minus actual cost)"* and *"traffic that already costs more than the reference shows negative savings"* — which the current code and tooltip both satisfy.

langwatch-agent (commented):

Aryansharma28 (commented):
## goated-review — deep pass

**Verdict: COMMENT.** This is a well-scoped, intentionally thin feature — no backend changes, good unit coverage on the pure calc module (`estimateReferenceCost` / `referenceModelOptions`), a matching spec file, correct `cost:view` gating, and clean reuse of `ModelSelector` / `SummaryMetric` / `ChartCard`. Nice work overall. But two P1s cut against the *exact* scenario the card was built for (self-hosted/free actual cost, custom models as the traffic source) — I'd want these fixed before or immediately after merge. Both P1s below are empirically reproduced (not just read from the diff): I checked out the PR branch, installed deps, and wrote throwaway repro tests exercising the actual production functions (`mergeCustomModelMetadata`, `referenceModelOptions`, `estimateReferenceCost`, `SummaryMetric`) — both reproduced exactly as described, then were deleted (not part of this review's suggested changes).

**P0: 0 | P1: 2 | P2: 1 | P3: 1 | P4: 0 | P5: 2**

### P1 — HIGH
1. `modelCostComparison.ts:52` — custom/self-hosted models pass the "complete catalog pricing" filter via a hard-coded `{0,0}` pricing placeholder, so they can be offered (and selected) as the savings reference, producing a false $0 reference cost and a false "additional cost" equal to the whole actual spend.
2. `ModelCostComparisonCard.tsx:134` — a genuinely $0.00 "Current Cost" (the card's own headline scenario — self-hosted traffic that really costs nothing) renders as "No data yet" instead of "$0.00", because `SummaryMetric`'s zero/no-data heuristic was designed for callers that always pass a `previous` value, and this card is the first one that doesn't.

### P2 — MEDIUM
3. `ModelCostComparisonCard.tsx:65` — the default reference model can render with the "provider not enabled — Update needed" broken-looking treatment when that provider isn't configured for the project, which is likely for exactly the self-hosted-only audience this card targets.

### P3 — LOW
4. `ModelCostComparisonCard.tsx:24` — `sumSeries`/`seriesKey` reimplements (less robustly, with a fragile positional-index coupling) an existing, better-tested utility (`readSummaryMetric` in `useLangyBriefing.ts`) that already solves this exact "sum a getTimeseries series by key" problem.

### P5 — PRAISE / OBSERVATIONS
- Good instinct keeping this thin and read-only: no backend changes, no new deps, reuses the existing model catalog and filter/date-range plumbing (`useFilterParams`) rather than inventing new endpoints. The `estimateReferenceCost` / `referenceModelOptions` split into a pure, fully-unit-tested calc module (with an explicit "return `undefined` rather than a partial estimate" contract) is exactly the right shape for this kind of money-math code, and all 5 existing unit tests pass cleanly against the current head.
- Process note (not a code defect): required CI checks (`evaluate`, `preflight`, `firefighting`) show `skipped` on the head commit, and a bot comment notes this is an external-contributor PR whose head wasn't executed ("did not execute this untrusted head"). The real test/build/lint suite has not machine-run at all yet beyond CodeRabbit's static pass — worth a maintainer approving the workflow run before merge. I independently ran the full diff through a security/supply-chain scan (no new deps/lockfile changes, no `eval`/`fetch`/`process.env`/`dangerouslySetInnerHTML`/external URLs in the new code) and it's clean.
- I independently re-verified @thiagolmoraes's rebuttal to CodeRabbit's "savings calculation is inverted" comment (`referenceCost - actualCost`, `increaseIs="good"`) against both the code and `specs/analytics/model-cost-comparison.feature:19` ("the card shows the estimated savings (estimate minus actual cost)") and `:31-33` ("traffic that already costs more than the reference shows negative savings") — the current formula and tooltip match the spec exactly. The rebuttal is correct; no need to revisit that one.

### Existing findings I'm not re-raising
- CodeRabbit's `referenceModelOptions` chat-vs-pricing docstring comment, the `sumSeries` named-parameter-object suggestion, and the "no error state" comment are all already addressed in the current head commit (`8fa62af`) — confirmed by reading the current file contents, not just the bot's own "✅ Addressed" tag.
- CodeRabbit's "savings calculation inverted" comment — declined by the author with a correct rebuttal (see P5 above); not re-raising.

- Whether a typical project in production actually has custom/self-hosted models configured often enough to make finding #1 a frequent occurrence, versus a rarer edge case — I verified the mechanism is real and reachable (and reproduced it against the actual production code), but not its production frequency.
- Whether the `getTimeseries` procedure's `analytics:view`-only server-side gate (no server-side `cost:view` check) for `performance.total_cost` is an accepted, deliberate systemic tradeoff elsewhere in the codebase — it's pre-existing and unchanged by this PR (same pattern already used by the `LLMMetrics`/`totalCostPerModel` cards on the same page), so I did not raise it as a new finding here.

langwatch/src/components/analytics/modelCostComparison.ts
```diff
@@ -0,0 +1,58 @@
...
+export function referenceModelOptions(
+  modelMetadata: Record<string, ModelMetadataForFrontend> | undefined,
+): string[] {
+  if (!modelMetadata) return [];
+  return Object.entries(modelMetadata)
+    .filter(([, metadata]) => {
+      const pricing = metadata.pricing;
+      return (
+        !!pricing &&
+        typeof pricing.inputCostPerToken === "number" &&
```
Aryansharma28: [P1] Custom/self-hosted models pass this "complete pricing" check via a hard-coded `{0,0}` placeholder, not real catalog pricing — they can end up offered (and picked) as the savings reference. Empirically reproduced against the actual production code (see below).

**Why:** `referenceModelOptions` treats a model as eligible whenever `pricing.inputCostPerToken` and `pricing.outputCostPerToken` are both `number`s (lines 49-54). That's correct for registry models, but custom models — the self-hosted/cheap-model audience this card exists for — don't come from the registry. They're synthesized by `mergeCustomModelMetadata`, which assigns *every* custom chat/embedding model a hard-coded `pricing: { inputCostPerToken: 0, outputCostPerToken: 0 }` (`langwatch/src/server/api/routers/modelProviders.utils.ts:125` on main — https://github.com/langwatch/langwatch/blob/f1ffeecc107d0198f8db53e701534289a4809226/langwatch/src/server/api/routers/modelProviders.utils.ts#L125, not touched by this PR). `0` satisfies `typeof x === "number"`, so every custom model a project has configured passes this filter and shows up in the reference-model dropdown, sorted in with the legitimately-priced ones.

I reproduced this directly: calling `mergeCustomModelMetadata({}, { vllm: { enabled: true, customModels: [{ modelId: "my-self-hosted-model", displayName: "x" }] } })` yields `pricing: {"inputCostPerToken":0,"outputCostPerToken":0}`; feeding that into `referenceModelOptions` returns `["vllm/my-self-hosted-model"]`; and `estimateReferenceCost` on a real 2M-input/500k-output-token period against that pricing returns a confident `0`, not `undefined`.

If a user picks one of their own custom models as the reference, "Estimated Cost on Selected Model" shows $0.00 and "Estimated Savings" shows `0 - actualCost`, rendered in red as additional cost equal to the *entire* actual spend — the opposite of a useful number, and a direct contradiction of both this function's own docstring ("models with complete catalog pricing") and the spec's own scenario: "Only models with catalog pricing are offered as reference ... models without pricing in the catalog are not listed" (`specs/analytics/model-cost-comparison.feature:27-29`).

Telling detail: the unit test for this function assumes custom models arrive with `pricing: undefined` — `modelCostComparison.unit.test.ts:76`: `"custom/qwen3-14b": metadataWith(undefined as never)`. That doesn't match what `mergeCustomModelMetadata` actually produces in production, so the test can't catch this — it's exercising a fixture shape that can't occur.

**Evidence:**
```ts
// modelProviders.utils.ts:113-127 (main, pre-existing, not in this diff)
for (const entry of allCustomModels) {
  const fullId = `${providerKey}/${entry.modelId}`;
  merged[fullId] = {
    ...
    pricing: { inputCostPerToken: 0, outputCostPerToken: 0 },
    ...
  };
}
```

**Suggested fix:** Distinguish "no real pricing" from "genuinely free." The type already supports it (`ReferencePricing = {...} | null`) — e.g. have `mergeCustomModelMetadata` set `pricing: null` for custom models unless a real per-token cost is configured on that provider, or have `referenceModelOptions` explicitly exclude custom-model IDs rather than trusting the numeric-pricing check alone.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+      ) : !timeseries.isLoading && !hasTraffic ? (
+        <Text textStyle="sm" color="fg.subtle">
+          No traffic in the selected period. Adjust the filters or date range
+          to compare costs.
+        </Text>
+      ) : (
+        <HStack align="start" gap={0}>
+          <SummaryMetric
+            label="Current Cost"
+            current={timeseries.isLoading ? undefined : actualCost}
```
Aryansharma28: [P1] A genuinely $0.00 "Current Cost" renders as "No data yet" instead of "$0.00" — which is precisely the card's own headline scenario (self-hosted traffic that really costs nothing). Empirically reproduced against the actual production code (see below).

**Why:** `SummaryMetricValue`'s zero-vs-no-data heuristic is:
```ts
// SummaryMetric.tsx:100-103
const isZeroZero =
  current !== undefined &&
  (current === 0 || current === "0") &&
  (previous === undefined || previous === 0);
```
Every other caller of `<SummaryMetric>` in this codebase (`CustomGraph.tsx:782`, `previous={previousByKey[entry.key]?.value}`) always passes a real `previous`, so `isZeroZero` only fires when *both* periods are zero — a reasonable "no data in either period" signal. This card never passes `previous` at all, so `previous` is always `undefined`, which alone satisfies the second half of the OR — collapsing the check to just `current === 0`. Any project whose actual cost for the period is exactly $0 (self-hosted, or any $0-priced model — the author's own rebuttal to CodeRabbit uses "actual $0" as the example) will see "No data yet" for Current Cost even though `hasTraffic` is true and the card is actively rendering real data for the other two metrics. It's indistinguishable, from the user's side, from the card being broken.

I reproduced this directly: rendering `<SummaryMetric label="Current Cost" current={0} format={...} increaseIs="neutral" />` (this component's exact call signature at lines 132-137, minus the `tooltip` prop) inside a `ChakraProvider` renders the text "No data yet" and does *not* render "$0.00".

The same mechanism affects "Estimated Cost on Selected Model" whenever `referenceCost === 0` (which is also the direct symptom of the sibling finding on `modelCostComparison.ts:52`).

**Evidence:** see `SummaryMetric.tsx:100-129` for the `isZeroZero` branch, and `CustomGraph.tsx:777-788` for the only other caller, which always supplies `previous`.

**Suggested fix:** Either stop relying on `SummaryMetric`'s built-in zero-heuristic for this card (render the $0.00 value directly, or pass a sentinel `previous` that deliberately defeats `isZeroZero`) or teach `SummaryMetricValue` to accept an explicit opt-out so single-period callers aren't forced into the two-period semantics. Also worth a `ModelCostComparisonCard.integration.test.tsx` covering this state — this codebase already has the convention for exactly this kind of multi-state card (see `ChartErrorState.integration.test.tsx`), and it would have caught this.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const options = useMemo(
+    () => referenceModelOptions(modelMetadata),
+    [modelMetadata],
+  );
+  const [referenceModel, setReferenceModel] = useState<string | undefined>(
+    undefined,
+  );
+  const selectedModel =
+    referenceModel ??
+    (options.includes(DEFAULT_REFERENCE) ? DEFAULT_REFERENCE : options[0]);
```
Aryansharma28, line 68: [P2] The default reference model can render with the "provider not enabled — Update needed" broken-looking treatment, for exactly the audience this card targets.

**Why:** `selectedModel` falls back to `DEFAULT_REFERENCE` (`"anthropic/claude-sonnet-4-6"`, line 41) whenever it's in the pricing-eligible `options` list — which it will be, since `options` comes from the full catalog regardless of which providers the project has enabled. But `ModelSelector`'s own filtering (`getCustomModels` in `ModelSelector.tsx`) only lists models from *enabled* providers, and separately flags `isProviderMissing` when the selected model's provider isn't in that enabled set — rendering the value with a red strikethrough and a tooltip: "anthropic provider isn't enabled here. Re-add the provider or pick a different model to use it."

For a project that only has a self-hosted/custom provider enabled (no Anthropic) — again, the card's own target audience — the default selection would show this "needs updating" treatment on first load, even though nothing is actually broken: a hypothetical reference model for cost comparison doesn't need to be an *enabled* provider, that's the entire premise of the feature. The tooltip's "re-add the provider ... to use it" instruction is actively wrong in this context.

**Evidence:** `ModelSelector.tsx` — `isProviderMissing = !!model && !!providerKey && !groupedByProvider.some((group) => group.provider === providerKey)`, and `groupedByProvider` is built only from enabled-provider models via `getCustomModels`.

**Suggested fix:** Either seed `DEFAULT_REFERENCE` selection only from models whose provider is actually enabled for the project (fall through to `options[0]` more aggressively), or pass a flag/variant to `ModelSelector` so a hypothetical, catalog-priced-but-not-enabled reference model doesn't trigger the "provider missing, re-add it" warning styling in this specific context.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+const SERIES = [
+  { metric: "performance.prompt_tokens", aggregation: "sum" },
+  { metric: "performance.completion_tokens", aggregation: "sum" },
+  { metric: "performance.total_cost", aggregation: "sum" },
+] as const;
+
+const seriesKey = (index: number) =>
+  `${index}/${SERIES[index]!.metric}/${SERIES[index]!.aggregation}`;
+
+const sumSeries = ({
```
Aryansharma28: [P3] `sumSeries`/`seriesKey` reimplement — less robustly — an existing utility that already solves this exact problem.

**Why:** `features/briefing/hooks/useLangyBriefing.ts:92-119` exports `readSummaryMetric({ buckets, series, metric, aggregation })`, which reads a `getTimeseries` `currentPeriod` sum via the canonical `buildSeriesName` encoder, looks up the series' index dynamically by matching `metric`+`aggregation` (`series.findIndex(...)`), and distinguishes "metric never appears" (`undefined`) from "metric present with value 0" — see https://github.com/langwatch/langwatch/blob/f1ffeecc107d0198f8db53e701534289a4809226/langwatch/src/features/briefing/hooks/useLangyBriefing.ts#L92-L119.

This file instead hand-rolls the same key format (`seriesKey`, line 21-22) and hardcodes positional indices (`seriesKey(0)`, `seriesKey(1)`, `seriesKey(2)` at lines 78/82/86) that must stay manually in sync with the order of the `SERIES` array — reordering `SERIES` silently breaks the mapping with no type error. `sumSeries` also can't distinguish "metric missing" from "metric present but zero," which is adjacent to the P1 finding above about zero values.

**Evidence:** `seriesKey`/`sumSeries` at lines 21-34 vs. `readSummaryMetric` at `useLangyBriefing.ts:92-119`.

**Suggested fix:** Reuse `readSummaryMetric` directly (or extract it to a shared analytics utility if `features/briefing` is considered too feature-specific to import from directly), rather than re-deriving the series key format and positional indices here.

Aryansharma28:
Hey @thiagolmoraes — thanks for this, left a review above. Two spots worth double-checking when you have a moment: the reference-model pricing fallback for self-hosted/custom models, and how the card renders when current cost is genuinely $0.00. Happy to pair on either if helpful — nice clean piece of work otherwise.

langwatch-agent (changes requested):

rogeriochaves (commented):
## Review, with fixes pushed on top

Good shape overall: thin, read-only, reuses the existing filter/date plumbing and the model catalog, and the pure calc module was the right place to put the money maths. The framing is right too. What I found is that the arithmetic did not match how the product actually bills, so I fixed it on this branch rather than only reporting it. Details below, worst first.

### 1. The estimate ignored cached tokens (cost arithmetic, the important one)

`estimateReferenceCost` priced `prompt_tokens * input + completion_tokens * output`. But `performance.prompt_tokens` is the **fresh, non-cached** input only. LangWatch records cache reads and cache writes as separate buckets, and says so in two places:

- `specs/ai-gateway/cache-token-telemetry.feature`: "The fresh input-token count (`gen_ai.usage.input_tokens`) is the non-cached remainder and must NOT include the cached tokens."
- `src/server/analytics/clickhouse/metric-translator.ts`, on `performance.total_tokens`: "Sum of prompt + completion tokens (the 'new content' delta, **excludes cache**)."

And the recorded cost prices all four buckets (`estimateCost` in `src/server/tracer/collector/cost.ts`: `inputTokens*inputRate + outputTokens*outputRate + cacheReadTokens*cacheReadRate + cacheCreationTokens*cacheCreationRate`). So the card was comparing a two-bucket estimate against a four-bucket actual. Not like with like.

This is not a corner case. On the local dataset I dogfooded against, cached tokens were 3.1m of 8.5m total, and the card's savings figure came out **38% low**. Sliced to one cache-heavy model the estimate was **3.2x low**:

| window | estimate before | estimate after | savings before | savings after |
|---|---|---|---|---|
| 30 days, all models | $16.85 | $18.73 | $3.04 | $4.93 |
| `?model=claude-fable-5` | $0.47 | $1.50 | -$4.39 | -$3.36 |

Fixed: the card now requests `performance.cache_read_tokens` and `performance.cache_write_tokens` and prices them at `inputCacheReadPerToken` / `inputCacheWritePerToken`, falling back to the plain input rate when the catalog publishes none, which is exactly what `estimateCost` does. Both sides of the comparison now carry the same convention, so whatever bias exists in ingestion applies equally to both.

Reasoning tokens I checked and left alone: they are a subset of completion tokens (the same file says so, and `estimateCost` has no reasoning bucket either), so there is nothing to add.

### 2. A model with no published price produced a confident $0

`referenceModelOptions` accepted any model whose two rates were `number`s. Zero is a number. That let through:

- every custom and self-hosted model, via the `{0, 0}` placeholder in `mergeCustomModelMetadata` (already caught by @Aryansharma28 and by CodeRabbit);
- three real catalog rows that ship `{0, 0}` because their price is not public: `gemini/lyria-3-pro-preview`, `gemini/lyria-3-clip-preview`, `openrouter/free`.

Fixed by making "usable price" mean both rates present **and at least one above zero**, which is the same bar `estimateCost`'s `hasAnyRate` applies before it prices a span. One honest rule now covers custom models and unpriced catalog rows alike, which also let me delete the `providers` plumbing and the `customModelIds` set: an unpriced model is excluded by its price, whether or not provider settings have loaded. That resolves CodeRabbit's "fail closed when providers is unavailable" comment too, without the "return nothing while providers load" side effect.

### 3. The exclusion was reachable anyway, through the picker

Worth flagging because the unit tests on `referenceModelOptions` gave false confidence here: `ModelSelector` does not treat `options` as an allowlist. `getCustomModels` **adds** every custom model of every enabled provider on top of it. So a user could select a self-hosted model the options list deliberately excluded, and get a $0.00 estimate and a savings figure equal to the entire actual spend.

I reproduced it in the browser (added a custom `qwen3-14b-selfhosted` to an enabled provider, picked it in the card's dropdown). The card now shows a dash and says why, screenshot in the description. The guard sits at the point of use, not only in the options list, so it holds regardless of what the picker hands back.

### 4. "No traffic in the selected period" before the query had even run

`hasTraffic` was derived from summed token counts, and `isLoading` is false while the query is still **disabled** (`useFilterParams`'s `queryOpts.enabled` is false until the project and date range resolve). So on first paint the card asserted the period was empty. Now the presence of `data` is what separates loading from a genuinely quiet period.

### 5. Smaller things

- `sumSeries` / `seriesKey` hand-rolled the bucket key format with hardcoded positional indices (@Aryansharma28's P3). With five series that coupling gets worse, so I moved `readSummaryMetric` out of `features/briefing/hooks/useLangyBriefing.ts` into `src/server/analytics/timeseriesSummary.ts` and both callers now import it. Values are looked up by metric name; reordering the request cannot swap two numbers. Briefing's 18 unit tests still pass.
- Added a token-basis line under the metrics ("Based on 5.4m input and 52.3k output tokens, plus 3.1m cached recorded in this period") so the figure is checkable against the period's usage instead of having to be taken on trust.
- Labels: "Current Cost" to "Actual Cost", and "Estimated Cost on Selected Model" to "Estimated Cost" since the model is named in the selector directly above. The Actual Cost tooltip now says outright that models with no published price, self-hosted ones included, are recorded at $0, which is the one assumption the savings number rests on.
- `SummaryMetric`'s new `zeroMeansNoData` doc said "most callers never pass `previous`". The `SummaryMetric` callers in `CustomGraph` always do; it is the `SummaryMetricValue` callers that do not. Reworded.
- Dropped the em-dash from the feature file title and the PR title (repo rule).
- Feature file: removed the `{0, 0}` placeholder detail from the custom-model scenario (CodeRabbit's point, it is internal behavior), and added scenarios for cached pricing, the cache-rate fallback, the unpriced-model state, loading, and the error state.

### What I did not change, and why

- **`referenceCost - actualCost` and the tooltip.** CodeRabbit called this inverted. It is not: the spec says "estimated savings (estimate minus actual cost)" and the sign convention matches. Also `increaseIs` is inert here since no `previous` is passed, so there is no red/green to get backwards. The author's rebuttal was right.
- **The "Update needed" treatment on the default reference model** (@Aryansharma28's P2). Real, and it will hit a project with no Anthropic provider, but the fix belongs in `ModelSelector`: a hypothetical reference model does not need an enabled provider, and the tooltip's "re-add the provider" instruction is wrong in this context. Same root cause as #3. Both deserve a follow-up on `ModelSelector` itself rather than a workaround in this card.
- **Estimate uses list prices, not the org's `CustomLLMModelCost` overrides.** The recorded cost resolves PROJECT to TEAM to ORGANIZATION to catalog; the frontend catalog the card reads does not. An org with a negotiated Sonnet rate will see the estimate priced at list. Worth a follow-up, out of scope for a card that adds no backend.
- **Trace-level token sums under a model filter.** Filtering by model selects traces, then sums the whole trace, so a multi-model trace contributes its sibling models' tokens. That is how every trace-level metric on this page already behaves, not something this card introduced.
- **Embedding tokens.** They land in `TotalPromptTokenCount` alongside chat tokens and get priced at the reference chat rate. Same pre-existing rollup, and small in practice; noting it rather than special-casing it.

### Verification

`pnpm typecheck` clean. `pnpm test:unit` on the calc module and the briefing hook (27 passed), `pnpm test:integration` on the two analytics component suites (12 passed). Dogfooded end to end in a browser against real traffic, with the displayed figures recomputed independently from the ClickHouse rows; the table and screenshots are in the description. The ClickHouse side is unchanged by this PR, and the query it uses already filters `TenantId` first and pushes the `OccurredAt` range into both halves of the dedup subquery.

One thing I could not verify: how often a real project has a custom model configured, which is what makes #3 reachable. I confirmed the mechanism end to end in a browser, not its production frequency.

I also approved the workflow runs, which had been sitting at `action_required` because this is an external fork, so the real suites finally get to run.

langwatch/src/components/analytics/modelCostComparison.ts
rogeriochaves: Fixed, and the mechanism turned out to be reachable one level deeper than the options list.

Eligibility is now "the catalog publishes a usable price": both rates present and at least one above zero, the same bar `estimateCost` applies before it prices a span. That excludes the `{0, 0}` placeholder on custom models and also three real catalog rows that ship `{0, 0}` because their price is not public (`gemini/lyria-3-pro-preview`, `gemini/lyria-3-clip-preview`, `openrouter/free`). The `providers` parameter and `customModelIds` are gone as a result, which also settles CodeRabbit's "fail closed when providers is unavailable" without the "return nothing while provider settings load" side effect.

The deeper part: `ModelSelector` does not treat `options` as an allowlist. `getCustomModels` adds every custom model of every enabled provider on top of it, so a user could still pick one. I reproduced that in a browser (added a custom `qwen3-14b-selfhosted` to an enabled provider, selected it in the card), so the guard now sits at the point of use as well: an unpriceable selection renders a dash and an explanation instead of a $0.00 estimate. Screenshot in the description.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
rogeriochaves: Still holds after the rework, and there is a test rendering the real card that pins it: a period with traffic and a genuinely $0 recorded cost shows `$0.00`, not "No data yet". I also reworded the `zeroMeansNoData` doc, which claimed most callers never pass `previous`; the `SummaryMetric` callers in `CustomGraph` always do, it is the `SummaryMetricValue` callers that do not.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
rogeriochaves: Done. With five series now (the two cache buckets were added) the positional coupling only got worse, so I moved `readSummaryMetric` out of `useLangyBriefing.ts` into `src/server/analytics/timeseriesSummary.ts` and both callers import it from there. Values are looked up by metric name, so reordering the request cannot swap two numbers. The briefing unit tests import from the new location and still pass.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
rogeriochaves: Agreed, and not fixed here on purpose. The right fix is in `ModelSelector`: a hypothetical reference model does not need an enabled provider, and the tooltip's "re-add the provider ... to use it" instruction is actively wrong in this context. It is the same root cause as the custom-model injection in `getCustomModels`, so both want one follow-up on `ModelSelector` rather than a workaround inside this card.

specs/analytics/model-cost-comparison.feature
```diff
@@ -0,0 +1,48 @@
...
+
+  Scenario: Only models with catalog pricing are offered as reference
+    When the user opens the reference model selector
+    Then models without pricing in the catalog are not listed
+
+  Scenario: Custom and self-hosted models are never offered as reference
+    Given a custom or self-hosted model has been added to a provider
+    When the user opens the reference model selector
+    Then the custom model is not listed, even though it has placeholder pricing
+    And it cannot be selected to produce a fabricated savings estimate
```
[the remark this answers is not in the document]
rogeriochaves: Done, the scenario now asserts only that the custom model is not listed and cannot be selected. The `{0, 0}` detail is gone.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  { metric: "performance.total_cost", aggregation: "sum" },
+];
+
+const money = (value: number) =>
+  value >= 0.01 || value === 0
+    ? numeral(value).format("$0.00a")
+    : numeral(value).format("$0.0000a");
+
+const signedMoney = (value: number) =>
+  value < 0 ? `-${money(Math.abs(value))}` : money(value);
```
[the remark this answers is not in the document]
rogeriochaves: Good catch, fixed. `money()` now delegates to `formatMoney` from `~/utils/formatMoney`, which is the same formatter the analytics registry already uses for `performance.total_cost`, so the card reads `$1200.00` and matches the "Total cost" tile above it on the same page rather than inventing its own currency format. Small values keep their precision through the `$0.00[00]` path and anything under a hundredth of a cent renders as `< $0.0001` instead of a misleading `$0.00`.

Pinned by a test that renders the card with a $1200 estimate against $200 actually spent and asserts all three figures render unabbreviated.

langwatch/src/components/analytics/ModelCostComparisonCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const hasTraffic =
+    promptTokens + completionTokens + cacheReadTokens + cacheWriteTokens > 0;
+  const cachedTokens = cacheReadTokens + cacheWriteTokens;
+
+  // Formatted up front so a value that cannot be estimated reads as a dash
+  // with an explanation underneath, rather than a skeleton that promises a
+  // number still on its way.
+  const asIs = (value: string) => value;
+  const cell = (value: number | undefined, format: (value: number) => string) =>
+    !isLoaded ? undefined : value === undefined ? "-" : format(value);
```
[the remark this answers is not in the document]
rogeriochaves: Fixed, `cell({ value, format })` now, and the three call sites with it.

rogeriochaves:
Three follow-ups on my review above.

### The CI state was never a judgement on your code

`langwatch-agent`'s CHANGES_REQUESTED reads as if the code was rejected. It was not. This is an external-fork PR, so every workflow had been parked at `action_required` since 2026-07-22 waiting for a maintainer to press approve, which is why the bot could only say "the application, integration, and gateway test suites are absent or skipped". Nobody had run them. I approved the runs, and `langwatch-app-ci` came back green on `a8aa2f9` (typecheck, lint, build, unit, integration, feature-parity). The runs on the current head are queued behind saturated runners, not failing.

### The generalisable lesson from finding #3

The unit tests on `referenceModelOptions` were good tests. They passed. They also could not have caught the bug, because the real path routed around the function they were testing: the card excluded unpriced models from `options`, and then `ModelSelector` added custom models back on top of `options` regardless. The rule was enforced in a helper and then handed to a shared component that did not honor it.

This shape has come up repeatedly this week, so it is worth stating rather than leaving as a detail of this card: **a test that exercises a pure function directly proves the function is right, not that the rule holds.** When a helper enforces something that matters, something has to test the path the user actually takes. That is what the new `ModelCostComparisonCard.integration.test.tsx` does here, and it is why I dogfooded the custom-model selection by hand in a browser rather than trusting the green unit suite.

### Deferred items are filed

Both carry Status Ready on the board.

- #6151, `ModelSelector` ignores the caller's `options` allowlist. This is the root cause of finding #3, and the "Update needed" treatment on the default reference model (@Aryansharma28's P2) falls out of the same conflation, so both are in that one issue with a note to split if they turn out to want different fixes. Filed separately from this PR because every other caller passing `options` has the same hole and has no way to know.
- #6152, frontend cost estimates use catalog list prices rather than the org's `CustomLLMModelCost` overrides, so an org with negotiated rates gets compared against prices it does not pay.

On honesty about both: I confirmed the #6151 mechanism end to end in a browser, but I have not measured how often a real project has a custom model configured, and I have not counted how many orgs have `CustomLLMModelCost` rows. Both issues say so. Reachable by construction, frequency unknown.

rogeriochaves:
Great idea, but feels weird to have this cost comparison here in the middle of the analytics page out of the blue, also super weird because generally there are multiple models going on, so what do you mean compare? You should probably be able to replace multiple models with each other (e.g. what if I use opus 5 + sonnet 5 instead of gpt 5.6 sol + luna?)

I think this should actually be a feature for powerful dynamic analytics API that will be able to pull the tokens in all this cuts and API for model costs + capability for Langy to be able to pull it out and render estimations graph in however cut the user wants to estimate

cc what do you think @0xdeafcafe and @drewdrewthis?

closing this for now and not planned the way it is built here

PR #6000, opened 2026-07-20
feat(model-providers): codex, sign in with your OpenAI account

rogeriochaves:
## Codex: sign in with your OpenAI account

Adds **Codex** as a model provider you set up by signing in with your OpenAI account instead of pasting an API key. Once connected, Langy and the small in-product AI assists run on your ChatGPT subscription through our AI gateway, with no key to manage and no per-token billing.

This is the first provider whose credential is an OAuth session rather than a key, so it threads a new auth shape through the whole stack: the registry, the settings drawer, the onboarding step, the gateway dispatch, and Langy's error surfaces.

### What you get

- **Sign in with OpenAI, no API key.** A device-code flow (the same one the Codex CLI uses): you get a short code, approve it at openai.com, and the encrypted tokens are stored against the widest scope you manage (organization if you are an org admin, otherwise team or project).
- **Runs through the gateway.** Codex requests go through the same pipeline as every other provider, so tracing, rate limits, budget, policy and guardrails all still apply. The gateway speaks OpenAI's Responses API to the Codex backend and streams the reply back.
- **Scoped to the right features.** Codex can only power the coding-assistant surfaces: Langy and every Fast-tier assist (trace Ask-AI search, conversation titles, commit messages, studio autocomplete, drawer translate, scenario and dataset generators, topic labelling). It is never offered for the prompt playground, evaluations or workflows, which OpenAI's subscription terms do not cover. As a default it can hold the **Langy** and **Fast** roles, never Default or Embeddings.
- **Clear errors.** When the OpenAI session dies or the ChatGPT plan hits its limit, Langy shows a dedicated card: a re-authenticate prompt for a dead session, and a "switch to another model" suggestion when the plan is capped, instead of a generic failure.

### Langy as its own model role

- Default models now have four roles: **Default, Fast, Langy, Embeddings**. The settings table and the config drawer both grew the Langy column/section, and `langy.chat` resolves through the Langy role (falling back to `prompt.create_default` for setups that predate the key, so nobody gets re-prompted).
- **Connecting Codex from Langy or onboarding** points the Langy and Fast role defaults at the codex model in one write, at the widest scope you manage. **Connecting from settings asks first**: a "Set Codex as your coding default?" dialog, so adding a provider row never silently rewrites org defaults.
- The Langy model picker gained an always-visible **"Configure available models"** footer linking to the model-providers settings, matching the main model selector (whose multi-instance provider handling it already shares).

### Placement

- **Onboarding and Langy first-run:** Codex leads the provider grid with a "Recommended" badge and the copy "Codex is recommended if you have a paid OpenAI account: sign in and it runs on your plan, no API key. Otherwise pick another provider and paste its key."
- **Settings add-provider menu:** Codex sorts to the bottom, below the API-key providers, since it is a niche subscription harness rather than a general provider.

### How it works

- **Auth** (`codexAccount.service.ts`): device-code OAuth against `auth.openai.com`, server-made PKCE, refresh grant, claims decoded from the id token. Tokens are stored as an encrypted `ModelProvider` credential set.
- **Gateway** (`services/aigateway/adapters/providers/codex.go`): a direct Responses-API dispatch to the Codex backend (Bifrost is bypassed for this provider the same way Voyage is), streaming SSE through verbatim and aggregating to a completed Response for the non-streaming assists. A 401 triggers one control-plane token refresh and a retry; plan-limit 429s are forwarded verbatim. The pipeline interceptors (tracing, limits, policy, budget, guardrail) wrap this dispatch, so bypassing Bifrost does not bypass any of them.
- **Gating** (`codexRestrictions.ts`, `resolveModelForFeature.ts`): a single allow-list of feature keys, enforced at the cascade resolver, the defaults writes, the execution backstop and every frontend picker (fail-closed: a picker only offers Codex models when it declares a licensed feature).
- **Langy** runs Codex through opencode's native OpenAI provider (the Responses dialect the backend speaks); the manager's LLM proxy restores the provider-prefixed model id on the wire so the gateway routes it to the Codex credential.

### Follow-up fixes from dogfooding

Two issues surfaced only against the live Codex backend, both fixed here:

- **Stateless reasoning round-trip.** The Codex backend keeps no server-side response state (`store: false`), so a tool loop has to replay its own reasoning across steps. The AI SDK only round-trips the encrypted reasoning content, and only asks the backend to return it, when the client sets `store: false` itself, so multi-step turns died with a 400. Fixed by setting `store: false` where the request is built, on both Codex paths: the tiny-assist model (`codexGatewayModel.ts`, via a default-settings middleware) and opencode's generated config (`provision.go`, on Codex turns only, leaving other OpenAI models on the default server-side store).
- **Sampling params the backend rejects.** The Codex backend answers 400 to a request that carries `temperature` or `top_p`, which the wider Responses API accepts. The tiny assists set a temperature for every other provider, so the title generator was silently failing ("leaving title unchanged"). The gateway now strips those in `codexRequestBody`, alongside the model/stream/store invariants it already pins.

### Testing

- Go: gateway dispatch (SSE proxy + headers + body pinning + usage, 401-refresh-retry, dead-session typed error, plan-limit verbatim, non-streaming aggregation), the request-body invariants (model/stream/store pinned, temperature/top_p stripped), the langyagent model-body rewrite, and opencode config setting `store: false` on Codex turns only. All pass.
- TypeScript: device-flow engine, restriction gate (with a pinned allow-list), settings drawer sign-in, fail-closed picker gating, the onboarding step, and the Langy error promotion. All pass.
- Typecheck clean across the workspace.

### Live dogfood

Verified on a local stack against the real OpenAI device-auth service, on a fully connected account:

- The backend `startDeviceSignIn` reaches `auth.openai.com` and returns a real device code, confirming the server-to-OpenAI leg works.
- The settings add-provider menu lists Codex last; the drawer swaps the API-key fields for the sign-in flow with the org scope pre-selected and no key inputs.
- The onboarding step leads with Codex, the Recommended badge, and the intended copy, with its own distinct glyph (not OpenAI's logo).
- **End to end through the gateway:** a live call on the connected account (the tiny-assist path: `getCodexVercelAIModel` to gateway `/v1/responses` to the Codex backend to SSE) returns a real reply. Battle-tested the full capability surface Langy needs, all pass: a single tool call, a multi-step tool loop, a multimodal image, and multi-turn memory.
- **Real Langy turns on Codex:** driving Langy in the app produced completed conversations whose projected turns carry 6 to 11 tool calls and 23K to 35K character answers, so multi-step tool use works through opencode's runtime, not just the raw SDK.

### Screenshots

**Onboarding / Langy first-run: Codex first with the Recommended badge**

onboarding

**Settings add-provider menu: Codex sorts last**

settings menu

**Settings drawer: the sign-in flow replaces the API-key fields**

settings drawer

**Langy running on Codex: the docked panel, model resolved to `openai_codex/gpt-5.6-terra`**

langy panel

## Deployment Impact

- **Env vars:** none added or required. Codex reuses the existing AI Gateway configuration; the tiny assists dial the gateway through `LW_GATEWAY_INTERNAL_URL` (falling back to `LW_GATEWAY_PUBLIC_URL` / `LW_GATEWAY_BASE_URL`), all already documented in `.env.example`. The OpenAI device-auth client id is a hardcoded public constant (the Codex CLI's own), not a secret. Tokens are stored per project as an encrypted `ModelProvider` credential using the existing `CREDENTIALS_SECRET` / `NEXTAUTH_SECRET`.
- **Helm values:** none added or changed.
- **Default `helm install` (no overrides):** unchanged. The provider is inert until a user signs in on a project; no new service, container, or required config ships. The gateway's Codex dispatch path only activates for a request that carries a connected Codex credential.
- **BYOC dataplane:** the Go gateway gains an outbound path to `chatgpt.com/backend-api/codex/responses` and `auth.openai.com` when a Codex credential is used. A dataplane that restricts egress would need those hosts reachable for Codex to work; the feature is opt-in and every other provider is unaffected.
- **Non-app change:** `tools/thuishaven/domain/overlay.go` now emits `LW_GATEWAY_INTERNAL_URL` as a loopback URL (dev-only orchestrator; no production surface).

## Notes

- The Langy event-sourced frontend this stacked on (#6027) has merged; the branch is rebased onto `main` and this is a clean `main` diff again.
- Along the way this PR also makes the unit-shard CI honest: the finalize-wedge hard floor now preserves test failures in its exit code instead of stamping wedged shards green. That surfaced six latent test failures that exist on `main` today (masked there, visible in main's own green runs once you read the wedged shard's log); all six are fixed here: a stale 403-body pin in the api-key ceiling test, and two router suites whose `vi.mock` of the rbac module silently fails to apply under the shared-module-registry unit pool. The wider order-fragility of `isolate: false` (more files break under artificial orderings CI does not produce) is left as a known repo issue rather than expanded into this PR.

https://claude.ai/code/session_018epY6nqs3nDsoB5XC1PXUU
https://claude.ai/code/session_01JGTf7eppBjc9AxWMvALtCL

## Summary by CodeRabbit

* **New Features**
  * Added **Codex (OpenAI account)** model provider with **device-based sign-in**, connected-state UI, and gateway-backed session refresh.
  * Introduced a **model provider onboarding step** with **Codex-first recommendations**, plus **"Skip for now."**
  * Added **Langy reconnect** UX and new Codex-specific error/recovery cards (session-expired, plan limit).

* **Bug Fixes**
  * Updated model pickers and default/feature overrides to **fail-closed hide restricted Codex models** unless the feature is licensed/eligible.
  * Improved Codex model handling across flows, including refresh/re-auth and provider selection ordering.


rogeriochaves:
Thanks CodeRabbit — addressed 6 of the 8 in fb3a6c605:

- **`reconnectCodex` trap (Major):** fixed. The flag now clears on all three escape hatches — new chat, close, and switching the composer model — so following the plan-limit card's own "pick another model" suggestion no longer hides the reply behind the sign-in screen.
- **Badge accessibility (Minor):** fixed. The badge is folded into `aria-label` so "Recommended" reaches assistive tech.
- **`codexStatus` email (Major, security):** fixed. The `project:view` query no longer returns the account email; the connector still sees it from the sign-in mutation result, but a plain member can't read the connecting admin's address.
- **`isModelAllowedForFeature` positional args (Minor):** fixed — named-parameter object, all call sites updated.
- **`compact` → `isCompact`:** already the case on the current head (resolved during the rebase onto the latest base branch).
- **Refresh dedup window (Major):** the comment now states the honest guarantee — the 30s window dedupes *staggered* 401s, not truly concurrent ones; it's best-effort, and the loser of a concurrent double-refresh surfaces as a re-auth, which is the correct terminal state.

Not taking two, with rationale:

- **`useModelSelectionOptions` named params:** this hook's positional `(options, model, mode)` signature predates this PR and is called from many sites across the app (including the branch this stacks on). Adding one optional `opts` is consistent with the existing shape; converting the whole signature to named params is a separate refactor touching unrelated callers, out of scope for the Codex feature.
- **Docked-card chrome duplicated with `DashboardLayout.tsx`:** that chrome is the base branch's Langy docked-card work, not part of this Codex diff — better addressed there.

services/langyagent/adapters/otelrelay/llmproxy.go
```diff
@@ -26,6 +26,56 @@ const llmPrefix = "/llm"
...
+		restore()
+		return
+	}
+	body["model"] = encodedModel
+	rewritten, err := json.Marshal(body)
+	if err != nil {
+		restore()
+		return
+	}
+	out.Body = io.NopCloser(bytes.NewReader(rewritten))
```
0xdeafcafe, line 65: this func reads a stream fully into memory, into a byte slice, then streams it back out. can we avoid?

services/aigateway/adapters/providers/codex.go
```diff
@@ -0,0 +1,434 @@
...
+			return &domain.Response{
+				Body:       response,
+				StatusCode: http.StatusOK,
+				Usage:      iter.Usage(),
+			}, nil
+		case "response.failed", "error":
+			message := gjson.GetBytes(payload, "response.error.message").String()
+			if message == "" {
+				message = gjson.GetBytes(payload, "error.message").String()
+			}
```
0xdeafcafe: can this func be iterated on? can be way more efficient i think

services/aigateway/adapters/controlplane/client.go
```diff
@@ -333,6 +333,45 @@ func (c *Client) setCommonHeaders(req *http.Request) {
...
+	if err != nil {
+		return "", "", fmt.Errorf("codex refresh call: %w", err)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
+	if err != nil {
+		return "", "", fmt.Errorf("codex refresh read: %w", err)
+	}
+	if resp.StatusCode == http.StatusUnauthorized {
+		return "", "", fmt.Errorf("control plane: %w", domain.ErrCodexSessionDead)
```
0xdeafcafe: use herr, we can handle/surface this

langwatch/src/server/modelProviders/codexAccount.service.ts
```diff
@@ -0,0 +1,383 @@
...
+  intervalSeconds: number;
+  verificationUrl: string;
+}
+
+/** One poll's outcome: still waiting, or a full token set. */
+export type CodexPollResult =
+  | { status: "pending" }
+  | { status: "complete"; keys: CodexTokenKeys };
+
+export class CodexAuthError extends Error {
```
0xdeafcafe: extend handled error

services/langyagent/adapters/otelrelay/llmproxy.go
rogeriochaves: Reworked in 37dd331642: the model swap is now a surgical gjson/sjson field set, so the messages payload is never decoded and re-encoded. The one buffered read has to stay though: rewriting a JSON field and re-stamping Content-Length both need the complete document, and only codex turns pay it (the prefix check short-circuits before any read). The SSE response path is untouched and streams through unbuffered.

services/aigateway/adapters/providers/codex.go
rogeriochaves: Took an efficiency pass in 37dd331642: the frame scanner is now a single in-place pass with no per-line slice allocations, output items accumulate straight into JSON-array bytes so the final stitch is one raw set (no intermediate slice plus re-marshal), and the usage skim shares the same scanner instead of doing its own split. The SSE leg also moved to codex_stream.go while I was in there.

services/aigateway/adapters/controlplane/client.go
rogeriochaves: Done in 37dd331642: the 401 now returns herr with the codex_session_expired code (ErrCodexSessionDead rides as a reason so the dispatcher's errors.Is check still fires), and the other failure paths return herr auth_upstream_unavailable with the status and body snippet in meta.

langwatch/src/server/modelProviders/codexAccount.service.ts
rogeriochaves: Done in 37dd331642: CodexAuthError now extends HandledError (code codex_auth_failed, kind in meta, provider fault, 401 for a rejected refresh and 502 otherwise), so the tRPC formatter serializes it as an expected failure instead of an unhandled 500.

langwatch/src/components/settings/CodexSignIn.tsx
```diff
@@ -0,0 +1,279 @@
...
+                  ? error.message
+                  : "The sign-in failed. Try again.",
+              timedOut: false,
+            });
+          }
+        })();
+      }, pending.intervalSeconds * 1000);
+    },
+    [poll, projectId, scopes, setAsCodingDefaults, utils, onConnected],
+  );
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: schedulePoll now takes a named-parameter object ({ pending, attempt }). The whole state machine also moved into a headless useCodexDeviceSignIn hook as part of the component split.

langwatch/src/features/langy/__tests__/langyErrorExplainer.unit.test.ts
```diff
@@ -69,6 +73,57 @@ describe("KNOWN_LANGY_ERROR_KINDS", () => {
...
+      const presentation = explainLangyError(
+        domain({
+          code: "langy_agent_errored",
+          reasons: [{ kind: "rate_limited" }],
+        }),
+      );
+      expect(presentation.kind).toBe("langy_agent_errored");
+    });
+  });
+
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: the codex describe blocks added by this PR now carry the nested when block. The neighboring pre-existing blocks in this shared file keep their current shape to avoid restyle churn outside the PR's scope.

langwatch/src/server/api/routers/modelProviders.ts
```diff
@@ -219,6 +229,121 @@ export const modelProviderRouter = createTRPCRouter({
...
+          );
+        }
+      }
+
+      return {
+        status: "complete" as const,
+        providerId: saved?.id,
+        email: poll.keys.CODEX_EMAIL,
+        plan: poll.keys.CODEX_PLAN,
+      };
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: codexSignInPoll now writes an audit row (action modelProvider.codexConnect) with the connecting user, project, scopes and plan on completion. The email itself deliberately stays out of the log row to keep the PII footprint minimal.

langwatch/src/server/modelProviders/utils.ts
```diff
@@ -52,6 +54,13 @@ export const getVercelAIModel = async ({
     modelProviders,
   });
 
+  // Codex never goes through the nlpgo chat-completions proxy: the codex
+  // backend is Responses-only and its OAuth session lives at the AI
+  // gateway, so the handle comes from there instead.
+  if (isCodexModel(model_)) {
+    return getCodexVercelAIModel({ projectId, model: model_, featureKey });
+  }
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: the provider existence and enabled guards now run before the codex branch, so an explicit openai_codex model can no longer bypass a disconnected or disabled provider row. Added three regression tests (missing row, disabled row, connected row returns the gateway handle).

services/aigateway/adapters/providers/codex.go
```diff
@@ -0,0 +1,421 @@
...
+		in := int(usage.Get("input_tokens").Int())
+		out := int(usage.Get("output_tokens").Int())
+		return domain.Usage{
+			PromptTokens:     in,
+			CompletionTokens: out,
+			TotalTokens:      in + out,
+		}, true
+	}
+	return domain.Usage{}, false
+}
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: the SSE leg (stream iterator plus the frame/usage scanners) moved to codex_stream.go. codex.go is now ~241 source lines and the new file ~110, both under the cap.

services/aigateway/adapters/providers/codex.go
```diff
@@ -0,0 +1,421 @@
...
+		return nil, &domain.UpstreamError{
+			StatusCode: resp.StatusCode,
+			Body:       raw,
+			Message:    fmt.Sprintf("codex backend HTTP %d", resp.StatusCode),
+			Headers:    forwardableUpstreamHeaders(headers),
+		}
+	}
+
+	return &codexStreamIterator{body: resp.Body, reader: bufio.NewReader(resp.Body)}, nil
+}
```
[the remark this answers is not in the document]
rogeriochaves: Done in 37dd331642: a 401 on the freshly refreshed token now short-circuits to the typed codex_session_expired error so Langy renders the re-authenticate card. Covered by TestCodexStream_StillUnauthorizedAfterRefreshIsSessionExpired.

specs/model-providers/codex-account-provider.feature
```diff
@@ -0,0 +1,126 @@
...
+
+  # The codex backend keeps no server-side state between requests, so a turn
+  # that reasons, calls a tool, then reasons again about the result has to carry
+  # its own reasoning forward. This is the difference between Langy answering a
+  # one-shot question and Langy chaining tools to finish a real task.
+  Scenario: Langy chains tools across a multi-step reasoning turn on Codex
+    Given Codex is my Langy model
+    When Langy runs a turn that calls a tool, reads the result, and calls another
+    Then each step replays the earlier reasoning so the stateless backend accepts it
+    And Langy finishes the task instead of failing partway through
```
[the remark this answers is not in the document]
rogeriochaves: Declining this one: the repo's feature-file convention keeps implementation details (config flags like store settings) out of scenarios, which describe behavior from the user's perspective. The stateless replay behavior is what the scenario pins; the concrete flag is pinned by the gateway's request-body test and the opencode provision test instead.

langwatch/src/server/api/routers/modelProviders.utils.ts
```diff
@@ -325,6 +326,19 @@ export const prepareLitellmParams = async ({
...
   projectId: string;
 }) => {
+  // Execution backstop for the terms-restricted provider: every general
+  // inference path (workflows, evaluations, playground, optimization
+  // studio) funnels through here on its way to litellm/nlpgo, and codex
+  // must never ride it — its only road is the AI gateway's Responses
+  // endpoint (see codexGatewayModel.ts), reserved for the coding-assistant
+  // surfaces. Pickers already hide codex models on these surfaces; this
+  // guard is what makes the restriction hold against a handcrafted request.
+  if (isCodexModel(model)) {
```
langwatch-agent: **P1 — The execution backstop is bypassable with the canonical model wire format.** The canonical picker/storage format is `mp_<provider-row-id>/<model>`, but `isCodexModel(model)` only recognises the legacy `openai_codex/...` prefix. A user can therefore submit a Codex row's canonical value to `/api/playground`: that route resolves the row by `mp_...`, then this check returns false, and `prepareLitellmParams` builds `openai_codex/<model>` plus the stored `CODEX_ACCESS_TOKEN` for the general inference proxy. This is exactly the handcrafted-request route the comment says must fail closed, and it violates the provider's feature restriction. Gate on the resolved `modelProvider.provider === "openai_codex"` (or normalise the wire value before the check) and add regression coverage for a canonical Codex model on the playground/workflow path.

langwatch/src/server/modelProviders/codexAccount.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    } catch (error) {
+      if (
+        error instanceof CodexAuthError &&
+        error.kind === "refresh_rejected"
+      ) {
+        return { status: "session_expired" };
+      }
+      throw error;
+    }
+    await this.repository.replaceCustomKeys({
```
langwatch-agent, line 404: **P1 — A successful refresh is not propagated to the gateway's cached credential, so the provider stays on a 401/refresh loop.** This only writes the rotated token to Postgres. The gateway keeps the access token in its resolved `Bundle.Credentials`; this path neither updates that in-memory credential nor emits the `MODEL_PROVIDER_UPDATED` change event that would evict it. Consequently every later dispatch still starts with the original expired token. The 30-second `JUST_REFRESHED_WINDOW_MS` makes those calls retry with the stored token briefly, but once it elapses the next request refreshes again, rotating tokens repeatedly; concurrent requests then hit the stale-token race described above and are falsely told to re-authenticate. Retain the returned credential in the gateway (and/or reliably invalidate/refetch the affected bundle), serialize refresh by provider row, and test that a second request after a successful refresh is sent upstream with the new token without another 401 or issuer refresh.

langwatch/src/server/modelProviders/codexAccount.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    const form = new URLSearchParams({
+      grant_type: "refresh_token",
+      refresh_token: keys.CODEX_REFRESH_TOKEN,
+      client_id: CODEX_OAUTH_CLIENT_ID,
+      scope: "openid profile email offline_access",
+    });
+    let json: Record<string, unknown>;
+    try {
+      json = await this.postForm(`${this.issuer}/oauth/token`, form);
+    } catch (error) {
```
langwatch-agent, line 179: **P2 — Transient issuer failures are reported as a dead user session.** This catch maps every failure from `postForm` to `refresh_rejected`: a timeout, DNS error, OpenAI 5xx, rate limit, or malformed response becomes the terminal `codex_session_expired` 401. The gateway then tells a user with a valid session to sign in again instead of surfacing a retryable provider failure. Only map a confirmed terminal refresh rejection (for example an OAuth `invalid_grant`) to `refresh_rejected`; preserve transport/5xx/malformed failures as retryable 5xx errors and cover both classifications.

langwatch-agent (dismissed):
2 P1s and 1 P2. Focused TypeScript (37 tests) and Go provider/control-plane/Langy adapter suites pass, but the credential refresh liveness and canonical-wire execution gate need correction; details are inline.

langwatch/src/server/api/routers/modelProviders.utils.ts
rogeriochaves: Fixed in 490eeffdc. Good catch: the backstop now fails closed on the given AND the resolveServingRow-resolved row's provider ("openai_codex"), in addition to the wire prefix, so the canonical mp_<row-id>/<model> value cannot reach prepareLitellmParams' param build. Regression tests cover both wire formats (legacy prefix and canonical row value against a codex row).

langwatch/src/server/modelProviders/codexAccount.service.ts
rogeriochaves: Fixed in 490eeffdc. The refresh service now appends the MODEL_PROVIDER_UPDATED change event (the same feed the gateway's authresolver long-polls; its applyChange evicts every bundle whose credentials carry the row id) right after persisting the rotation, so the next dispatch takes a cold miss and resolves the rotated token. The in-flight retry already rides the fresh token from the response. Test pins the append with the row's org and id. Refresh stays serialized best-effort by the just-refreshed window; a per-row single-flight lock remains the follow-up if double-refresh proves real under load.

langwatch/src/server/modelProviders/codexAccount.service.ts
rogeriochaves: Fixed in 490eeffdc. Only the issuer's own 4xx answer naming invalid_grant maps to refresh_rejected now; timeouts, DNS failures, 5xx, rate limits and malformed bodies stay retryable HandledErrors (kind "http", 502), so the gateway fails the request without telling a valid session's owner to re-authenticate. CodexAuthError carries the issuer status to make the classification explicit; tests cover the terminal 400 invalid_grant, an issuer 503, and a network failure.

PR #6006, opened 2026-07-21
fix(event-sourcing): bound claude-span-sync visibility retry + quarantine runaway groups

0xdeafcafe:
## Problem (prod incident 2026-07-20)

A single trace's `claudeCodeSpanSync` reactor group
(`project_00056…/fold/traceSummary/reactor/claudeCodeSpanSync/trace:6c3b7139…`)
sat in the `{event-sourcing/jobs}` queue for **14h**, not draining, at a flat
**~8,000 pending** — and was consuming **~25% of the entire shared queue's retry
churn** (565/hr of 2,299/hr), degrading trace processing for every tenant.

### Root cause

The reactor re-reads a Claude Code turn's canonical log records from ClickHouse
to fold them into spans. The compact log *contribution* and the full canonical
*log records* travel through separate durable pipelines, so when a `logContributed`
event arrives before the records are visible, the reactor throws
`CanonicalLogNotVisibleError` to **retry until they land**.

For this trace the records were **never findable** — either never stored (dropped
upstream) or outside the visibility query's **±2-day partition window**. So:

1. Every retry burned the full **25-attempt** ladder (~2h37m each), and
2. because a pathological turn re-emits **thousands** of contributions (each a
   *fresh* `attempt-1` job), the per-job `maxAttempts` cap **never terminated the
   group** — measured max attempt was 16, and **zero** jobs ever dead-lettered.

The group was effectively immortal. Full investigation write-up in the linked issue/thread.

## Fixes

**1. Visibility-retry deadline** (`claudeCodeSpanSync.reactor.ts`)
Once a contribution has been unfoldable for `CLAUDE_LOG_VISIBILITY_DEADLINE_MS`
(default **10m**, measured from the log's ingest time `event.occurredAt`, which is
stable across retries), the reactor **gives up and completes the job** — so the
per-trace group *drains* instead of re-burning the ladder forever. Generous enough
that the normal cross-pipeline race (seconds) never trips it. Env override:
`LANGWATCH_CLAUDE_LOG_VISIBILITY_DEADLINE_MS`.

**2. Group failure-streak quarantine** (`groupQueue.ts` / `scripts.ts`)
A per-group counter, INCR'd on each **retryable** job failure and **cleared on the
group's next success**, catches the runaway the per-job cap misses (failures
*across* a group's jobs). Once the streak exceeds
`LANGWATCH_GQ_QUARANTINE_FAILSTREAK_THRESHOLD` (default **500**, `0` disables), the
job is routed through the **existing exhausted-retries path** that blocks the group
— reusing the proven blob-hold + block machinery rather than inventing a new park.
The blocked group carries a descriptive stored error so `/ops` shows *why*.

The two compose: fix 1 makes the incident's group drain (jobs succeed-by-giving-up,
so the streak resets and quarantine never fires); fix 2 is the general backstop for
*any* reactor that fails hard with no give-up path.

## Config / rollout

| Env | Default | Effect |
|---|---|---|
| `LANGWATCH_CLAUDE_LOG_VISIBILITY_DEADLINE_MS` | `600000` (10m) | give-up deadline; clamped to ≤ 1h |
| `LANGWATCH_GQ_QUARANTINE_FAILSTREAK_THRESHOLD` | `500` | consecutive failures → quarantine; `0` = off |

The quarantine defaults **on** at a deliberately high threshold (a single group
failing 500× in a 15-min window with **zero** interleaved successes is an
unambiguous runaway, never a healthy group riding out a transient blip). If
reviewers prefer it ship inert, set the default to `0` — one line in `scripts.ts`.

## Testing

- `claudeCodeSpanSync.reactor.unit.test.ts` — **14 pass** (+3: retries within
  deadline, inclusive boundary, gives-up past deadline).
- `scripts.quarantineThreshold.unit.test.ts` — **4 pass** (default / kill-switch /
  positive / garbage-fallback).
- `groupQueue.poisonGuard.integration.test.ts` — **13 pass** (+3: quarantine on
  sustained failures, kill switch, clear-on-success) against real Redis.
- Full groupQueue unit suite (**40**) green — no regressions.

## Immediate mitigation (separate from this PR)

The already-wedged prod group won't clear on deploy on its own; drain it from
`/ops` (`drainGroup`) and confirm whether trace `6c3b7139…` has canonical log rows
in ClickHouse (never-stored vs partition-window-miss) to pin the upstream cause.

## Risk

Fix 1 is narrow (one reactor branch). Fix 2 touches the shared group-queue failure
path — mitigated by reusing the existing exhausted-retries/block machinery,
gating cheaply on a construction-time threshold, clearing on success, and the
`0` kill switch. Highest-blast-radius change; please scrutinise the failure-path
edits in `groupQueue.ts`.

## Summary by CodeRabbit

* **Reliability Improvements**
  * Added configurable deadlines for Claude Code log visibility retries, preventing prolonged processing when logs never appear.
  * Added a failure-streak quarantine guard for repeatedly failing queue groups, isolating affected work while allowing other processing to continue.
  * Successful jobs now clear accumulated group failure streaks.
* **Configuration**
  * Added environment settings to customize retry deadlines and quarantine thresholds, including an option to disable the quarantine guard.
* **Tests**
  * Added coverage for retry boundaries, quarantine behavior, recovery, and configuration handling.

### DIFF
--- a/langwatch/src/server/app-layer/traces/claude-code-log-to-span.ts
+++ b/langwatch/src/server/app-layer/traces/claude-code-log-to-span.ts
@@ -212,6 +212,48 @@ export function resolveClaudeTurnLogCap(raw: string | undefined): number {
 }
 
 /**
+ * How long the span-sync reactor keeps retrying a log CONTRIBUTION whose
+ * canonical log records are not visible in ClickHouse yet, before it gives up.
+ *
+ * The canonical log and its compact trace contribution travel through separate
+ * durable pipelines, so a contribution can reach the reactor a beat before the
+ * log projection lands — the reactor throws to retry until it does. But if the
+ * records NEVER materialize (dropped upstream, or a partition-window miss), that
+ * throw is a poison pill: every re-emitted contribution for the trace burns the
+ * full 25-attempt retry ladder, and a pathological turn re-emits thousands,
+ * wedging the per-trace group and burning a large share of the shared
+ * event-sourcing queue (prod incident 2026-07-20). This deadline bounds it: once
+ * a contribution has been unfoldable for this long — measured from the log's
+ * ingest time (`event.occurredAt`) — the reactor stops retrying and completes
+ * the job so the group drains. Generous enough that the normal cross-pipeline
+ * race (seconds) never trips it; tune via
+ * `LANGWATCH_CLAUDE_LOG_VISIBILITY_DEADLINE_MS`.
+ */
+export const CLAUDE_LOG_VISIBILITY_DEADLINE_MS = 10 * 60 * 1000;
+
+/**
+ * Upper bound on the visibility-retry deadline, so an operator-supplied
+ * `LANGWATCH_CLAUDE_LOG_VISIBILITY_DEADLINE_MS` can't re-open the unbounded-retry
+ * poison-pill failure mode this deadline exists to close.
+ */
+export const MAX_CLAUDE_LOG_VISIBILITY_DEADLINE_MS = 60 * 60 * 1000;
+
+/**
+ * Resolve the operator-configured visibility-retry deadline (ms) from an env
+ * value, clamped to `[1, MAX_CLAUDE_LOG_VISIBILITY_DEADLINE_MS]`. Absent,
+ * non-numeric, or below-one values fall back to
+ * {@link CLAUDE_LOG_VISIBILITY_DEADLINE_MS}. Mirrors {@link resolveClaudeTurnLogCap}.
+ */
+export function resolveClaudeLogVisibilityDeadlineMs(
+  raw: string | undefined,
+): number {
+  const parsed = Number(raw);
+  if (!Number.isInteger(parsed) || parsed < 1)
+    return CLAUDE_LOG_VISIBILITY_DEADLINE_MS;
+  return Math.min(parsed, MAX_CLAUDE_LOG_VISIBILITY_DEADLINE_MS);
+}
+
+/**
  * The span kind a claude_code log event feeds, or null when the event is not
  * folded into a span (so it stays a plain, visible log). The receiver marks +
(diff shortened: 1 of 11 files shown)


0xdeafcafe:
@coderabbitai review

0xdeafcafe:
Good catch — fixed in the latest commit.

You're right that this broke the established contract: `unblockGroup` / `drainGroup` / `moveToDlq` all `DEL` the `:strikes` key so a recovered group gets a fresh run, and the new `:failstreak` key didn't follow suit — an operator unblocking a quarantined group would have watched it re-quarantine on its very next failure.

I fixed it by clearing the streak **as the group is parked** rather than adding the key to the three ops Lua scripts. Same end state (a recovered group always starts from zero), but it avoids bumping `numkeys` and the KEYS indices across `UNBLOCK_LUA`/`DRAIN_GROUP_LUA`/`MOVE_TO_DLQ_LUA` and their ~7 call sites — incident-time tooling I'd rather not risk an off-by-one in. It also means the streak never lingers while the group sits blocked.

Covered by an assertion in the quarantine integration scenario (`failstreak` is null once the group is in the blocked set) and documented in `poison-group-park-guard.feature`. Poison-guard suite: **13 passing**, feature-parity green.

0xdeafcafe:
Superseded by #6016.

Both fixes here were still live on `main` — verified rather than assumed — but this branch conflicted with the group-queue rewrite in #5947, and #5918 needed folding in alongside it. #6016 is cut fresh from `main` and carries both the bounded visibility retry and the failure-streak quarantine.

One thing changed in the port: `clearGroupFailures` now sits inside #5948's post-completion guard and after the completion counter and dispatch audit. On the original success path a Redis blip on that `DEL` would have been caught by the outer failure handler and re-delivered the whole batch — the exact bug #5948 fixed.

PR #6007, opened 2026-07-21
fix(langy): stop the relay dropping progress frames and hanging on reconnect

0xdeafcafe:
Fixes both halves of the Langy relay problem behind the "Langy takes forever to load" report.

## Symptom A — a working turn looks frozen for minutes

A turn ran for **9.4 minutes** doing ~50 tool calls, and the UI showed **nothing**
the whole time. The turn completed fine server-side; the relay had **dropped all of
its progress frames as `no-run-token`**.

**Root cause.** The worker signs each relay frame with the per-conversation
`runToken` it got **synchronously from the Redis handoff** at dispatch. The relay
VERIFIED frames against the **async Prisma `RunToken` projection**, which on a
brand-new conversation hasn't landed when the first frames arrive — so `getRunToken`
returned `null`, and `loadRunToken` **cached that null** and dropped every frame on
the connection.

**Fix** (`langyTurnRelay.ts`, `langy-relay.ts`):
- Read the runToken from the **synchronous per-turn handoff first** (the exact token
  the worker signed with), falling back to the projection. Mirrors what
  `langyTurnAccess` already does for the browser side.
- **Never cache a null** — a transient miss is retried on the next frame.

## Symptom B — "Starting up…" hangs after a refresh

A refresh that lands as the turn finishes can miss the worker's terminal frame, so
the durable buffer has no `end`/`error`. The reconnected `onTurnStream` then blocks
in `follow()` until the hard `AGENT_CHAT_TIMEOUT_MS` deadline — the UI sits on
"Starting up…" for minutes though the turn already finished.

**Fix** (`langy.ts` `onTurnStream`, `langyTurnSettlement.ts`): while it tails the
live edge, a concurrent watcher polls the durable fold (`getById`) + the per-turn
heartbeat (`buffer.liveness`). When the turn is provably settled with no terminal in
the buffer, it aborts the follow and yields a **synthesized terminal** so the client
resolves (it reconciles the transcript via `langy.messages`).

**Safe by construction** — `decideSyntheticTerminal` synthesizes ONLY when BOTH:
- the per-turn heartbeat is **stale** (a live turn keeps a fresh one), AND
- the fold is **settled** (`IDLE`→`end`, `FAILED`→`error` with `lastError`).

The fold flips to `ACTIVE`/`RUNNING` at `agent_turn_accepted` — *before* any output —
so even a cold start (silent for its first ~10-20s) reads in-flight, never settled,
and can't be cut off. Two consecutive settled reads (~10s) are required so a
projection blip can't end a live stream; a real terminal on the buffer always wins.

## Tests

- `langyTurnRelay.unit.test.ts` — **25** (+2: handoff auth when the projection lags;
  transient null re-queried not cached).
- `langyTurnSettlement.unit.test.ts` — **7** (heartbeat-fresh / in-flight / idle /
  failed±lastError / unknown-status).
- Full langy streaming suite — **64** green.
- Specs: `langy-dual-stream.feature` (+4 scenarios).

The `onTurnStream` generator wiring (concurrent watcher + abort + cleanup) isn't
unit-imported by the suite; the type-check runs in CI. The settlement *decision* is
fully unit-tested in isolation, and a real terminal always takes precedence over a
synthesized one.

## Risk

Symptom A: small, contained (auth source + caching fix + route wiring; no schema
change; frame verification unchanged). Symptom B touches the `onTurnStream` streaming
generator — mitigated by the double-gate (heartbeat-stale AND fold-settled), the
two-poll confirmation, real-terminal precedence, and always awaiting the watcher
before the request ends. Worth a close look at the generator lifecycle in `langy.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Langy streaming reliability by synthesizing an end/error when a reconnect misses a terminal frame, without affecting turns that are still live or have fresh heartbeats.
  * Strengthened relay authentication by using a synchronous per-turn handoff run token when available, with fallback when it’s unavailable.
  * Adjusted run-token caching to avoid permanently rejecting transient “token not ready” states, allowing later frames to authenticate.
* **Tests**
  * Added unit coverage for synthetic terminal decisions and reconnect/settlement, plus handoff run-token timing and run-token retry/caching edge cases.

### DIFF
--- a/langwatch/src/server/api/routers/langy.ts
+++ b/langwatch/src/server/api/routers/langy.ts
@@ -230,6 +230,71 @@ function abortableDelay(ms: number, signal: AbortSignal): Promise<boolean> {
   });
 }
 
+/** How often the settlement watcher consults the durable fold + heartbeat. */
+const SETTLEMENT_POLL_MS = 5_000;
+/**
+ * Consecutive settled reads required before synthesizing a terminal, so a single
+ * projection blip can never end a live stream.
+ */
+const SETTLEMENT_CONFIRM_POLLS = 2;
+
+/**
+ * Poll a turn's durable fold + per-turn heartbeat while its live edge is being
+ * tailed, and resolve to the terminal entry the buffer never received — or null
+ * if the stream ended first (aborted) or the turn never settled.
+ *
+ * Split out of `onTurnStream` so the subscription body stays at the orchestration
+ * level and this confirmation loop is independently testable. The safety gate
+ * itself lives in {@link decideSyntheticTerminal}.
+ */
+async function watchForMissedTerminal({
+  projectId,
+  conversationId,
+  turnId,
+  userId,
+  buffer,
+  signal,
+}: {
+  projectId: string;
+  conversationId: string;
+  turnId: string;
+  userId: string;
+  buffer: {
+    liveness(a: {
+      conversationId: string;
+      turnId: string;
+    }): Promise<{ stale: boolean }>;
+  };
+  signal: AbortSignal;
+}): Promise<LangyStreamEntry | null> {
+  let settledStreak = 0;
+  while (!signal.aborted) {
+    if (!(await abortableDelay(SETTLEMENT_POLL_MS, signal))) return null;
+    const [conversation, liveness] = await Promise.all([
+      getApp()
+        .langy.conversations.getById({ id: conversationId, projectId, userId })
+        .catch(() => null),
+      buffer.liveness({ conversationId, turnId }).catch(() => null),
+    ]);
+    if (!conversation || !liveness) {
+      settledStreak = 0;
+      continue;
+    }
+    const decision = decideSyntheticTerminal({
+      status: conversation.status,
+      lastError: conversation.lastError,
+      heartbeatStale: liveness.stale,
+    });
+    if (!decision) {
+      settledStreak = 0;
+      continue;
+    }
+    settledStreak += 1;
+    if (settledStreak >= SETTLEMENT_CONFIRM_POLLS) return decision;
+  }
+  return null;
+}
+
 /**
  * Dispatch a turn through the app-layer turn service. Create and Continue are
  * the SAME operation — `isNewConversation` is the only difference (it emits the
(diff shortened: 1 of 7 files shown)


rogeriochaves (dismissed):

0xdeafcafe:
Thanks — worked through all of these in `f91d820`. Responses per finding:

**🔒 Scope the handoff lookup by `projectId` — fixed (good catch).**
You're right that `@@unique([projectId, ConversationId])` makes conversation identity project-scoped while the handoff key is only conversation+turn scoped. Rather than change the handoff key format (it's shared with the turn service and the process-outbox, so a unilateral change there would break the stash/read contract), `readHandoffRunToken` now takes `projectId` and the route validates it against the stored handoff record — which already carries `projectId` — refusing any handoff stashed under a different project. Same guarantee, no cross-module key change.

**🩺 Uncaught `readHandoffRunToken` rejection — fixed.**
Agreed, and it defeated the exact case the two-stage lookup exists for. `loadRunToken` now wraps the handoff read in try/catch, logs, and falls through to the projection. Covered by a new test that makes the handoff read reject and asserts the frame still applies via the projection.

**📐 Extract the watcher polling loop — done.**
Pulled out into a module-level `watchForMissedTerminal()` with the poll/confirm constants alongside it, so `onTurnStream` is back to orchestration only and the confirmation loop is independently testable. Its `buffer` param is a structural `{ liveness }` type for that reason.

**📐 `loadRunToken` positional params — done.** Converted to a named-parameter object.

**📐 Nest a `when` block — done.** The new relay cases are now `given … → when … → it …` (three `when` blocks: projection-not-landed, handoff-read-fails, first-lookup-missed).

**🎯 Type `status` against `LANGY_CONVERSATION_STATUS` — documented rather than applied, deliberately.**
It is a plain `as const` object (not Zod), so the union is derivable — but the value reaches this function from the conversation projection's bare string column (`getById(): { status: string }`). Narrowing the parameter would only force a cast at the call site asserting something the data doesn't guarantee. The typo safety you're after is already structural: every comparison is against the exported constant, so a renamed/removed status fails to compile *here*, and an unrecognised value (incl. `ARCHIVED`, which I'd missed enumerating) falls through to the safe `null`. I've written that rationale into the JSDoc.

**📐 Bind the `@unit` scenario in `langy-dual-stream.feature`** — resolved; `feature-parity` is green (that file is on the tolerated legacy list, and the checker passes at `EXIT=0`).

Langy streaming suite: **65 passing**.

0xdeafcafe:
Good catch on the watcher rejection window — fixed in `95ce69ee`.

`.catch()` is now attached at creation rather than in the `finally`. You're right that the gap is real and not theoretical: `follow()` blocks for up to `FOLLOW_BLOCK_MS` per iteration and can stay there for the whole turn, so a rejection had a minutes-long window where nothing was handling it — long enough for Node's default `--unhandled-rejections=throw` to take the process down before `finally` ever attached a handler. The `finally` now just `await watcher` since it carries its own handler.

Langy streaming suite still **65 passing**.

PR #6020, opened 2026-07-21
fix(langy): honor org-scoped rollout rules in the UI visibility gate

0xdeafcafe:
## Symptom

Langy doesn't appear in the UI for a user whose **organization** has `release_langy_enabled` turned on — even though they're a member with `langy:view`.

## Root cause

Two things compound on current `main`:

1. **#5966** (`feat(langy): flag-only access…`) removed the identity-based staff bypass entirely — access is now purely `release_langy_enabled`. So everyone depends on that flag resolving `true` for *their* evaluation context; there's no longer an `@langwatch.ai` shortcut.
2. **`useShowLangy` resolved the flag with `projectId` only**, never `organizationId`. The flag store's rule matcher (`server/featureFlag/rules.ts`) fails closed when a rule constrains an id the context doesn't carry, so an **org-scoped** targeting rule can never match client-side.

The server gate `hasLangyAccess` (`server/app-layer/langy/langyAccessGate.ts`) *does* forward `organizationId`. Net effect: an org-wide rollout **authorizes the tRPC procedures** while the **UI keeps the handle hidden** — the API would answer, but there's no way in.

## Fix

Forward `organizationId` (from `useOrganizationTeamProject`) into the `useFeatureFlag` call in `useShowLangy`, mirroring the server gate. The hook's options type already accepts it. This is purely additive — it can only let an *already-authorized* org rule surface the panel; it never grants beyond what the server allows.

```diff
 const { enabled: releaseLangy } = useFeatureFlag(LANGY_RELEASE_FLAG, {
   projectId: project?.id,
+  organizationId: organization?.id,
   enabled: mayReadLangy,
 });
```

## Tests

`src/features/langy/hooks/__tests__/useShowLangy.integration.test.tsx` — the `useFeatureFlag` mock reproduces the store's rule matcher, so the cases mean what they say:

- org-wide rollout (no project rule) → **panel visible** (the regression guard; **verified red without the fix**, green with it)
- project-scoped rollout → visible
- rollout for a *different* org → hidden
- no rollout → hidden

Ran the suite through the configured runner: **4 passed**.

Also adds an "organization-wide rollout reveals the handle, not just the API" scenario to `specs/langy/langy-baseline.feature` (Access and rollout gating).

## Immediate workaround (no deploy)

Until this ships, enable `release_langy_enabled` in `/ops/feature-flags` via a **per-project rule** (or a plain global toggle) rather than a per-org rule — the client already sends `projectId`, so those match today.

> Note: full `pnpm typecheck` intentionally not run (pins the machine on this monorepo per repo convention); the change is type-trivial (an already-declared optional field) and the tests exercise the compiled hook.

https://claude.ai/code/session_014wKWDe5JXkWRBRWncfUdt1

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Langy panel visibility to match rollout targeting for both organization and project contexts.
  - Organization-wide rollouts now correctly reveal the Langy handle across eligible project routes.
  - The panel remains hidden when rollout rules target a different organization or no matching rollout applies.

- **Tests**
  - Added integration test coverage for Langy visibility behavior under org-level rollout, project-level rollout, mismatched org targeting, and missing rollout rules.

### DIFF
--- a/.github/workflows/go-services.yaml
+++ b/.github/workflows/go-services.yaml
@@ -21,6 +21,7 @@ jobs:
     outputs:
       service: ${{ github.event_name == 'workflow_dispatch' || steps.filter.outputs.service }}
       chart: ${{ github.event_name == 'workflow_dispatch' || steps.filter.outputs.chart }}
+      langyagent: ${{ github.event_name == 'workflow_dispatch' || steps.filter.outputs.langyagent }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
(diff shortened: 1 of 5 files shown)


0xdeafcafe:
@coderabbitai review

PR #6042, opened 2026-07-22
fix(traces): make tracesV2.header's full IO resolution opt-in per caller

Aryansharma28:
## Ticket

Follow-up to review feedback on #5754 (merged), from [Alex's comment](https://github.com/langwatch/langwatch/pull/5754#discussion_r3620914 ) and his approval note: "i think we should rewrite this whole handling to do it in a better way, but approving to get the fix out."

## Change

#5754 wired `full: true` unconditionally into `tracesV2.header`'s call to `TraceSummaryService.getByTraceId`, to fix the new Trace Explorer's drawer never resolving offloaded (ADR-022) input/output. That fixed the drawer, but `full: true` is not free: `TraceSummaryService.resolveFullIO` always does one extra ClickHouse spans read per call (`getNormalizedSpansByTraceId`) before it can even check whether anything was offloaded — the service's own doc comment already warns "only ever pass true for a single-trace read (the drawer), never for a list page — full resolution costs one spans read per trace." The router endpoint ignored that nuance and paid the cost for every caller.

Traced every caller of `tracesV2.header`:
- **`useTraceHeader.ts`** (drawer's Summary tab, `buildTraceMarkdown` export) — genuinely needs full, untruncated content. → `full: true`.
- **`useOpenTraceDrawer.ts`** — seeds the drawer's cache and prefetches ahead of drawer mount for the *same* trace the drawer opens for. → `full: true`, keeping the cache key aligned with `useTraceHeader` so the seed is actually picked up on mount.
- **`TraceIdPeek.tsx`** (hover popover) — renders `lineClamp={2}`, never more. → `full: false`.
- **`langyTraceName.ts`**'s `useResolvedTraceName` — reads only `traceName`/`name` for a Langy chip label. → `full: false`.
- **`capabilityHydrators.ts`**'s `traceByIds`** — fetches N trace headers for a capability card list and immediately truncates each to 90 chars. → `full: false`.
- **`useConversationPrefetch.ts`** — eagerly warms sibling headers in the same conversation. The immediate next/previous turn (`NEAR_RADIUS = 1`) → `full: true`, since that's the overwhelmingly common click and it's worth the extra spans read to keep it instant. Every farther sibling (up to `RADIUS = 5` in each direction) → `full: false` — rarely opened, same philosophy as this file already not prefetching span trees ("they're heavy and only needed once the user actually navigates there").

`full` is a required-in-practice field — every call site above passes it explicitly, true or false, so no caller inherits a hidden default. The schema itself sets `.default(true)` purely as a rollout safety net: a browser tab still running the pre-deploy bundle sends no `full` field at all, and this default keeps that in-flight request working exactly as it did before this PR (full resolution, no validation error) until the tab refreshes onto the bundle that sends it explicitly.

No behavior change for the drawer's own read (still full, still resolves offloaded content) or for clicking into the immediate next/previous conversation turn (still instant). The peek popover, Langy name chips, capability cards, and farther sibling prefetch go back to being cheap reads — no extra spans scan, matching what each of them actually renders or how likely they are to be opened.

## Evidence

- `pnpm typecheck`: clean, no errors.
- New/updated unit tests proving each caller's `full` value, including the near-vs-far split in conversation prefetch: `useTraceHeader.unit.test.ts`, `langyTraceName.unit.test.ts`, `capabilityHydrators.traceByIds.unit.test.ts` (new), `useConversationPrefetch.unit.test.ts` (new). **14 passed.**
- Full `src/features/traces-v2` + `src/features/langy` regression sweep: **129 test files, 1375 tests, all green.**
- `src/server/api/routers/__tests__/tracesV2.*.unit.test.ts`: unaffected, still green.
- Verified the `full: z.boolean().default(true)` rollout-safety net directly against `zod`: omitting the field parses to `full: true`; passing `false` still parses to `false`.

Advisory PR from tech-debt follow-up work, for human review.

## Summary by CodeRabbit

* **Performance**
  * Lightweight trace headers are now requested by default for trace previews, trace name lookups, conversation prefetching, and trace-by-IDs hydration (`full: false`), reducing unnecessary work.
  * The trace drawer continues to request full header details when opening (`full: true`).
* **Bug Fixes**
  * Improved header cache consistency by aligning lightweight/full header reads across drawer opening, prefetching, and related lookups.
* **Tests**
  * Updated and added unit tests to verify correct lightweight vs full header behavior across trace header, conversation prefetch, and trace name resolution scenarios.


langwatch/src/server/api/routers/tracesV2.ts
```diff
@@ -1063,6 +1063,24 @@ export const tracesV2Router = createTRPCRouter({
...
+         * Defaults to `true` (the pre-existing, unconditional behavior)
+         * purely for rollout safety: a browser tab still running the
+         * previous frontend bundle sends no `full` field at all, and this
+         * default keeps that in-flight request working exactly as before
+         * instead of a Zod validation error, until the tab refreshes onto
+         * the bundle that sends it. Every call site added by this change
+         * passes the field explicitly — this default only ever backstops
+         * a stale client, never a caller in the current code.
+         */
+        full: z.boolean().default(true),
```
Aryansharma28, line 1344: [P2 — fixed in `ef59400ba`] Required-with-no-default `full` would have 400'd on stale-bundle clients during rollout.

**Why:** `full` was originally `z.boolean()` (required, no default) on an already-shipped procedure. A browser tab still running the pre-deploy JS bundle sends a `header` request with no `full` field at all — that request would fail Zod validation until the tab refreshed onto the new bundle. Self-healing and no data loss, but an avoidable break for an in-flight session.

**Evidence:** every call site this PR touches (`useTraceHeader.ts`, `useOpenTraceDrawer.ts`, `TraceIdPeek.tsx`, `langyTraceName.ts`, `capabilityHydrators.ts`, `useConversationPrefetch.ts`) already passes `full` explicitly, so a default here only ever catches a stale client, never masks a forgotten caller in the current code.

**Fix applied:** `full: z.boolean().default(true)` — `true` matches the exact pre-existing (pre-this-PR) unconditional behavior, so a stale client keeps working exactly as it did before, just without yet realizing the perf win until it refreshes. Verified directly: `schema.parse({projectId, traceId})` → `full: true`; `schema.parse({..., full: false})` → `full: false`.

langwatch/src/features/traces-v2/hooks/useConversationPrefetch.ts
```diff
@@ -42,11 +42,15 @@ export function useConversationPrefetch(
         if (!turn) continue;
         // Fire and forget. tRPC's prefetch is a no-op when the entry is
         // already fresh in cache, so subsequent passes don't re-hit the
-        // server.
+        // server. full: false — most siblings are never opened, so this
+        // stays a cheap read, same philosophy as not prefetching spans:
+        // the full IO resolution one spans read costs is only worth
+        // paying once the user actually navigates there.
         void utils.tracesV2.header.prefetch({
```
Aryansharma28, line 63: [P3 — fixed in the PR description] `full: false` here trades away instant-open for previously-prefetched siblings.

**Why:** before this PR, prefetch and the drawer's own read shared the same (always-`full:true`) cache key, so clicking into a prefetched sibling turn opened with full content instantly. After this change, prefetch caches under `full:false` while the drawer's `useTraceHeader` query asks for `full:true` — a different cache key — so opening a sibling now always does a fresh live fetch, even one that was just prefetched moments ago.

**Evidence:** `useTraceHeader.ts`'s query is `{ ...queryArgs, full: true }`; this file's prefetch call is `{ projectId, traceId: turn.traceId, occurredAtMs: turn.timestamp, full: false }` — no shared key.

This is the right call (matches this file's own existing choice not to prefetch span trees for the identical cost reason — "they're heavy and only needed once the user actually navigates there"), but it wasn't stated anywhere the tradeoff exists. Added a "Known UX tradeoff" paragraph to the PR description so a reviewer isn't surprised by the loading flicker on sibling-click later.

Aryansharma28 (commented):
## goated-review — self-review

Ran the full dimension sweep (architecture, business-logic correctness, regression risk, code quality, YAGNI/DRY/SOLID, security, performance, tests) against this PR as author, since it's the fix landing Alex's review comment from #5754. No P0/P1s survived adversarial verification. Two real findings came up during the audit — both already fixed in follow-up commits before this review posted, so this is a record of what was checked, not an open action list.

**P0: 0 | P1: 0 | P2: 1 (fixed) | P3: 1 (fixed) | P4: 0 | P5: 2**

### P2 — fixed in `ef59400ba`
Making `full` a strictly-required field (no default) on an already-shipped tRPC procedure meant a browser tab still running the pre-deploy bundle — which sends no `full` field at all — would get a Zod validation error on this endpoint until it refreshed. Not a data-loss or security issue, but a real, avoidable rollout-window break. Fixed by defaulting the schema field to `true` (the exact pre-existing, unconditional behavior) — every call site in this codebase still passes the field explicitly, so the default only ever backstops a stale client, never masks a forgotten call site in the current code. Verified directly against `zod`: omitting the field parses to `full: true`; explicit `false` still parses to `false`. Inline comment below.

### P3 — fixed in the PR description
`useConversationPrefetch.ts` warming siblings with `full: false` means clicking into a *previously-prefetched* sibling turn no longer opens with full content instantly (the drawer's `full: true` query is now a different cache key, so it still does a live fetch on open). This is the right tradeoff — consistent with the file's own existing choice not to prefetch span trees for the same reason — but it wasn't stated anywhere, so a reviewer would hit it as a surprise later. Added an explicit "Known UX tradeoff" paragraph to the PR description. Inline comment below.

### P5 — praise
`useOpenTraceDrawer.ts`'s cache-key alignment across all 5 seed/prefetch call sites for the drawer's `full: true` read — including the onboarding preview-trace branch, where the seeded data is the *only* data source since the real query is permanently disabled for preview ids (`isReady` in `useTraceQueryArgs.ts` excludes `isPreviewTraceId`) — was handled correctly. A missed key there would have silently broken the sample-data walkthrough with no error, just a stuck loading skeleton; worth calling out since it's the kind of easy-to-miss detail a review should specifically check rather than skim past.

### P5 — observation, checked and dropped
Initially flagged (low confidence) that `TraceIdPeek`'s hover popover, now `full:false`, might fire a live network fetch for a client-side-only onboarding "preview" trace id if hovered before the drawer opens. Checked: `isPreviewTraceId` is only referenced in drawer data hooks (`useOpenTraceDrawer`, `useTraceQueryArgs`, `useTraceSpanTree`, `useConversationContext`, `useTraceEvaluations`), never in the `TraceTable` row components that render `TraceIdPeek`/`TimeHoverCard`/`ChatTurns`. Preview rows aren't rendered through the table's hover-peek path at all — they only ever reach the drawer via `useOpenTraceDrawer`'s seeding. Not reachable; dropping.

### Existing findings not re-raised
CodeRabbit's 2 BDD-nesting comments (missing `describe("when …")` wrapper in the new/touched test files) were legitimate — already fixed in commit `24ad21b67`.

### CI
The two failing checks (`sdk-python-complete`, `test`) are the same pre-existing, repo-wide failure as on #5754 and on `main` — an expired `OPENAI_API_KEY` secret causing a Python SDK e2e example (`distributed_tracing.py`) to 401 against the real OpenAI API. Unrelated to this PR's TS-only diff; confirmed via `gh run view --log-failed`.

- Whether this repo's actual deploy strategy (rolling vs. atomic) makes the stale-client window in the P2 finding longer or shorter than I'm assuming — I don't have visibility into the k8s rollout config from the diff, so I defaulted to the safer assumption (window exists, however brief) rather than dismiss it.
- Real-world click-through rate on prefetched conversation siblings — I don't have product analytics on how often a user actually opens a sibling turn vs. how often prefetch is wasted, so I can't quantify whether the P3 tradeoff nets positive; the reasoning stands on the file's own stated philosophy, not on measured data.

PR #6055, opened 2026-07-22
feat(langy): card gaps, the created-key dialog, and a discard guard on the scenario editor

0xdeafcafe:
Five independent fixes, each with tests. The first three close gaps a review of the derived-cards/blocks surface turned up; the last two are UI work asked for directly.

## 1. Agent refs on choices cards resolve as the viewer

`AGENTS.md` teaches the `choices` block using an agent ref by example — "which agent should this scenario run against?" is the question ADR-060 exists for — so an agent ref is the one the model is most likely to emit, and it was the one kind `CAPABILITY_HYDRATORS` could not resolve. The row fell through to `plain`: the raw id instead of a name, and an agent that no longer exists still offered as a live choice.

Adds the `agent` hydrator, and teaches `AGENTS.md` rule 25 the `hints` vocabulary it was already accepting but never explaining.

*4 tests. Mutation-checked: removing the registry entry kills 3 of them.*

## 2. Metric cards name what was measured

The two most-asked questions resolve to metric keys whose leaf is not the thing being measured. "How many traces this week" counts distinct `metadata.trace_id`, so the card read **Trace id**; both latency presets are `performance.completion_time`, so average and p95 rendered identically.

Titles now come from a `metric·aggregation` table, falling back to a humanised key.

*5 tests.*

## 3. Skills ask in the format the reader can answer

The `scenarios` and `datasets` skills asked their disambiguating question as prose. Where `langy-card` blocks render, that is a question the panel cannot turn into anything clickable. Both now teach the `choices` block, with the prose form kept as the fallback for surfaces without the channel. `_compiled` regenerated.

*13 skills tests pass.*

## 4. The created-key dialog leads with the key

The dialog exists to hand over a secret that is readable exactly once, and that secret had no element of its own — the only way to get it was to read it back out of a `.env` line, a Bearer header, or a JSON config. It now leads: masked by default, reveal and copy, with the rainbow gradient on its top edge.

Around it: the project selector moves above both tab groups (it rewrites every snippet in the dialog, not just the group it lived inside); both tab strips render through one component; the five config paths become buttons in one scrolling row instead of an `HStack asChild` around a `<button>` around a `<Text>`, wrapping raggedly; a "Done" footer, since closing is the point of no return; and "Token Created" becomes "Secret key created" to match what the drawer that opens it calls the thing.

The rainbow gradient had been copy-pasted into two files. Both now read from `components/brand/rainbow.ts`.

## 5. The scenario editor asks before throwing work away

A new scenario lives only in the form until the first save — no record to come back to, no draft to recover — so Escape, the corner close and Cancel each silently cost the user everything they had written.

What counts as something to lose differs by mode, and the AI case is why:

- **Editing a saved scenario** — exactly react-hook-form's dirtiness. The scenario survives either way; only the edits are at stake.
- **Creating one** — dirtiness is the wrong test. An AI draft arrives as the form's own *defaults*, so the form reads pristine while closing on it throws away a generated scenario. A new scenario therefore counts as unsaved work the moment it holds any content, typed or drafted.

Dirtiness is read through a leaf component, not from the close handler: react-hook-form only tracks `isDirty` for forms that subscribe to it during render, so an imperative read returns a value nobody ever computed. A successful save keeps using the unguarded close.

*New spec `specs/scenarios/scenario-editor-discard-guard.feature`, 10/10 scenarios bound. Mutation-checked: collapsing create mode to plain dirtiness fails the AI-draft test.*

## Verification

`pnpm test:integration run src/pages/settings/api-keys/` (38), `src/components/scenarios/` (91), the two dataset upload drawers (31), `pnpm test:unit run …/token-created-snippets` (14), and `check-feature-parity.ts` green.

## Deployment Impact

No migrations, no config, no new env. Two things need a rebuild to take effect:

- **The langy-agent image** bundles `skills/` (`Dockerfile.langyagent` COPYs the tree, and `_compiled/native` is what the agent reads), so the skill copy in §3 ships with the next worker image, not with the app deploy. Change detection for that image is fixed separately in langwatch-saas#785 — without it a skills-only change would not have rebuilt the worker at all.
- **The app image** carries §1, §2, §4 and §5; they are frontend-only and take effect on the next app release.

Rollback is a revert: nothing here writes data, changes a schema, or alters an API contract. The scenario-editor guard only adds a confirmation step ahead of an existing close path, and the created-key dialog changes presentation of a token that is already minted before it renders.


0xdeafcafe:
Heads-up on the red checks — none are from this diff:

- **e2e / e2e-complete** → `clickhouse/clickhouse-server:latest` rolled to 26.7.1.1315, which rejects the (unchanged, deployed) `00017_create_gateway_budget_ledger.sql` migration in DB setup, before any test runs. Fixed by #6059 (pins the CH image). This branch's e2e goes green once #6059 lands and this rebases — the diff here touches zero migrations/ClickHouse.
- **test / sdk-python-complete** → expired `OPENAI_API_KEY` in the Python example tests; fails identically on `main` (`73f73d7fe`).

0xdeafcafe:
Update — all 6 red checks are external, none from this diff:

| Check | Cause | Mine? |
|---|---|---|
| **e2e / e2e-complete** | `clickhouse-server:latest` rolled to 26.7, rejecting the deployed `00017_create_gateway_budget_ledger.sql` migration in DB setup | No — fixed by #6059 (validated: e2e green there). Goes green here on rebase after #6059 lands |
| **test / sdk-python-complete** | Expired `OPENAI_API_KEY` in the Python example tests | No — fails identically on `main` (`73f73d7fe`) |
| **check_links / docs-complete** | `npx mintlify broken-links` can't install — `@mintlify/editor@0.0.246` is **404 in the npm registry**; the checker never runs (npm E404, then the step's "no success string → fail" logic trips) | No — upstream registry breakage. `docs-ci` passed on `main` at 19:59; it runs here because the skills commit touched `docs/skills/*.mdx`, but the failure is CLI-install, not link content |

App-side CI (typecheck, build, test-unit, all test-integration shards including the new `TokenCreatedDialog.integration.test.tsx`) is green.

The docs-ci break is the same unpinned-tool class as #6059 but with no repo-internal known-good mintlify version to pin to, so I've left it rather than guess. It'll self-heal when Mintlify republishes, or wants a pinned CLI version in `docs-ci.yml` as a separate call.

PR #3327, opened 2026-04-19
feat(ai-gateway): ship v1 GA — virtual keys, budgets, guardrails, Go data plane

rogeriochaves:
# feat(ai-gateway): ship v1 GA — virtual keys, budgets, guardrails, Go data plane

## Summary

Ships the first version of the **LangWatch AI Gateway** — a Go data plane that adds virtual-key governance, hierarchical budgets, inline guardrails, fallback chains, byte-for-byte cache pass-through across providers, and per-tenant OTel routing on top of every request. Exposed at `gateway.langwatch.ai` (cloud) and via a Helm sub-chart (self-hosted).

**The enterprise pitch**: ~236 ns of hot-path overhead (M3 Pro arm64 native), per-tenant trace routing with zero customer config, every provider's cache discount preserved byte-for-byte (90% Anthropic prompt caching, OpenAI prefix caching, Gemini cached-content), built-in fallback across providers, and a CLI + REST API for every management action. Drop-in OpenAI / Anthropic / Bedrock / Azure / Vertex / Gemini compatibility.

## Test verification status

**Provider matrix (P2): 🟢 30/30 GREEN** — every `(openai / anthropic / gemini / bedrock / azure / vertex)` × `(simple / streamed / tool-calling / structured-outputs / cache)` cell asserts HTTP 200 + valid `X-LangWatch-Trace-Id` + non-zero usage + cost on the LangWatch platform. Total provider spend per full P2 run: **~$0.0353**. Per-cell trace IDs in [`.claude/AI-GATEWAY-TEST-MATRIX.md`](.claude/AI-GATEWAY-TEST-MATRIX.md) §Priority 2.

**Regression gating (CI)**: the live provider matrix fires automatically on PR push for gateway-relevant paths (`services/aigateway/**`, `pkg/**`, `cmd/**`, `langwatch/src/server/gateway/**`, the trace-fold reactor, `langwatch/src/server/routes/gateway-internal.ts`, `charts/gateway/**`, the workflow file itself) — ready PRs only, drafts skip, fork PRs without secrets skip cleanly. Plus the nightly cron (04:00 UTC) and manual `workflow_dispatch`. Concurrency-locked against the shared matrix VKs to prevent double-spend on parallel runs (~$0.07/run). See [`.github/workflows/gateway-matrix.yaml`](.github/workflows/gateway-matrix.yaml).

**Coding-agent matrix (P3): 🟢 3/4 GREEN, 🟡 1/4 WORKAROUND VERIFIED**:
[...]

PR #6059, opened 2026-07-22
fix(ci): pin the ClickHouse image so a rolled `latest` can't break e2e

0xdeafcafe:
## Symptom

`e2e` began failing on every branch a few hours after passing on `main` (green at 19:59 on `73f73d7fe`), with no migration diff in between. It dies in DB setup, before any test runs:

```
goose run: ERROR 00017_create_gateway_budget_ledger.sql: failed to run SQL migration:
Code: 36. DB::Exception: Column(s) UpdatedAt of the AggregatingMergeTree table are
neither part of the sorting key nor aggregate measures ... (BAD_ARGUMENTS)
(version 26.7.1.1315)
```

## Cause

`e2e-ci.yml` and `sdk-javascript-ci.yml` pull `clickhouse/clickhouse-server:**latest**`. That tag rolled to **26.7.1.1315**, which tightened `AggregatingMergeTree` validation: a non-key column that isn't an aggregate state is now a hard error, where older servers allowed it. `00017_create_gateway_budget_ledger.sql` (shipped in #3327, unchanged since) has exactly that shape (`UpdatedAt` outside `ORDER BY`), so goose fails and the whole job dies.

Nothing in the repo changed — the server image under the workflow did. An unpinned server image makes CI a function of whatever Docker Hub serves that hour.

## Fix

Pin both workflows to `clickhouse/clickhouse-server:25.10.2.65` — the version the integration suite already builds its container from (`event-sourcing/__tests__/integration/globalSetup.ts:290`) and migrates `00017` against green. This only narrows CI to a version the repo **already trusts**; it introduces no new dependency.

Bumping ClickHouse — and making `00017` compatible with a stricter server (add `UpdatedAt` to `ORDER BY`, or `allow_dimensions_outside_sorting_key=1`, in a **new** migration since 00017 is deployed) — becomes a deliberate PR instead of a surprise from a moving tag.

## Deployment Impact

CI-only. Touches two GitHub Actions workflow files; no product code, no runtime image, no migration. It pins a test-time service image to a version already used elsewhere in CI.

## Summary by CodeRabbit

* **Chores**
  * Standardized the test environment by pinning the database service to a specific version.
  * Improved consistency and reproducibility across end-to-end test runs.

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -108,7 +108,7 @@ jobs:
           --health-retries 5
 
       clickhouse:
-        image: clickhouse/clickhouse-server:latest
+        image: clickhouse/clickhouse-server:25.10.2.65
         ports:
           - 8123:8123
         env:
--- a/.github/workflows/sdk-javascript-ci.yml
+++ b/.github/workflows/sdk-javascript-ci.yml
@@ -107,7 +107,7 @@ jobs:
 
     services:
       clickhouse:
-        image: clickhouse/clickhouse-server:latest
+        image: clickhouse/clickhouse-server:25.10.2.65
         ports:
           - 8123:8123
         env:


0xdeafcafe:
Cherry-picked onto #5809 as 398aa0dee — its e2e lane was dying on the rolled `latest` and needed the pin in its own merge ref. Identical content, so whichever lands first the other merges clean.

PR #6067, opened 2026-07-23
fix(langy): harden GitHub App installation binding at /setup

sergioestebance:
## Summary
- `recordInstallation` now rejects rebinding a GitHub App installation that's already connected to a different organization, instead of silently overwriting the existing mapping.
- Genuine same-org re-installs still upsert cleanly.
- Blocked rebind attempts are audit-logged; the caller always sees the same generic install-failed message either way.

## Test plan
- [x] `langy-github-installations.service.unit.test.ts` — 12 tests passing, including the new conflict-guard cases
- [x] `github-langy-install.unit.test.ts` — 19 tests passing, including the new route-level rejection case
- [x] `pnpm typecheck` clean on all touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented GitHub installations from being reassigned across organizations.
  * Added race-safe installation claiming so only one organization can successfully record an installation.
  * Same-organization reinstalls refresh stored installation details.
  * Cross-organization setup failures continue to return a generic error without exposing ownership details.

* **Audit & Monitoring**
  * Added an audit event for rejected cross-organization installation attempts.

* **Tests**
  * Extended unit and integration coverage for cross-tenant conflicts and concurrent setup behavior.

### DIFF
--- a/langwatch/src/server/app-layer/langy/__tests__/langy-github-installations.service.unit.test.ts
+++ b/langwatch/src/server/app-layer/langy/__tests__/langy-github-installations.service.unit.test.ts
@@ -23,6 +23,7 @@ function makeRepo(
   rows: LangyGithubInstallationRow[] = [],
 ): LangyGithubInstallationsRepository & {
   upsert: ReturnType<typeof vi.fn>;
+  insertOrGetExisting: ReturnType<typeof vi.fn>;
   deleteByInstallationId: ReturnType<typeof vi.fn>;
   setSuspended: ReturnType<typeof vi.fn>;
   setRepositories: ReturnType<typeof vi.fn>;
@@ -36,6 +37,25 @@ function makeRepo(
       async (id: string) => byId.get(id) ?? null,
     ),
     upsert: vi.fn(async (_i: UpsertLangyGithubInstallationInput) => {}),
+    // Mirrors the real unique-index semantics: the read (`byId.get`) and the
+    // write (`byId.set`) below have no `await` between them, so this function
+    // body runs to completion in one microtask turn — exactly like a DB
+    // unique-constraint `INSERT` — which is what makes the race test below a
+    // real regression test rather than a coincidence of mock timing.
+    insertOrGetExisting: vi.fn(
+      async (input: UpsertLangyGithubInstallationInput) => {
+        const existing = byId.get(input.installationId);
+        if (existing) return { inserted: false, row: existing };
+        const created: LangyGithubInstallationRow = {
+          ...input,
+          suspendedAt: null,
+          createdAt: new Date(),
+          updatedAt: new Date(),
+        };
+        byId.set(input.installationId, created);
+        return { inserted: true, row: created };
+      },
+    ),
     setRepositories: vi.fn(async () => {}),
     setSuspended: vi.fn(async () => {}),
     deleteByInstallationId: vi.fn(async () => 1),
@@ -94,7 +114,7 @@ function makeAppTokens(
 }
 
 describe("recordInstallation", () => {
-  it("maps GitHub's installation metadata onto the upsert", async () => {
+  it("maps GitHub's installation metadata onto the atomic insert", async () => {
     const repo = makeRepo();
     const svc = new LangyGithubInstallationsService(repo, makeAppTokens());
     const result = await svc.recordInstallation({
@@ -102,7 +122,7 @@ describe("recordInstallation", () => {
       organizationId: "org-1",
     });
     expect(result.accountLogin).toBe("acme");
-    expect(repo.upsert).toHaveBeenCalledWith(
+    expect(repo.insertOrGetExisting).toHaveBeenCalledWith(
       expect.objectContaining({
         installationId: "inst-1",
         organizationId: "org-1",
@@ -111,6 +131,9 @@ describe("recordInstallation", () => {
         repositorySelection: "all",
       }),
     );
+    // A brand-new installation is claimed by the atomic insert alone — the
+    // same-org refresh path (`upsert`) never runs.
+    expect(repo.upsert).not.toHaveBeenCalled();
   });
 
   describe("when the installation is already owned by a different organization", () => {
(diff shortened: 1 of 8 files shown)


sergioestebance:
## Ruthless review — PR #6067

**Verdict: no P0/P1. Approving.**

Diff is small and contained: adds `LangyGithubInstallationConflictError` + a same-installation/different-org guard in `recordInstallation` (`langy-github-installations.service.ts`), wires the route to audit-log the blocked attempt while returning the pre-existing generic failure message, and adds one service test + one route test. Correctly placed in the service layer (business rule), not the route or repository.

### P2 — residual TOCTOU race in the new guard (non-blocking, reduces but doesn't fully close the class of bug)

`recordInstallation` does `findByInstallationId` (read) → compare → `upsert` (write), with no transaction or lock spanning the two (`langy-github-installations.service.ts:122-156`). The repository's `upsert` update branch sets `organizationId: input.organizationId` unconditionally on conflict (`langy-github-installations.prisma.repository.ts:105-113`) — there's no `WHERE organizationId = <org read earlier>` guard at the DB layer. So two concurrent `/setup` completions for the *same, not-yet-recorded* installation id could both read `existing === null`, both pass the guard, and the second `upsert` would still silently claim the row for its org.

**Gate 1 (attribution):** checked base (pre-this-PR) behavior — base had *no* guard at all, so the overwrite happened unconditionally, no race required, on every call. The bad outcome (cross-org row overwrite) reproduces on base, in fact more easily. Per the review's own attribution rule this caps at P2 and does not block: this diff closes the deterministic, no-timing-required exploit and leaves only a narrow race window, a strict reduction in harm.

**Fix direction (follow-up, not blocking):** wrap the read + upsert in a `prisma.$transaction` at `SERIALIZABLE` isolation, or replace the upsert's update branch with a conditional raw update (`UPDATE ... WHERE installationId = $1 AND organizationId = $2`) that no-ops instead of overwriting on a stale read, or take a short-lived advisory lock keyed on `installationId` around the whole `recordInstallation` body (there's already a Redis client in this service's dependency graph via `LangyGithubAppTokenService`, so the primitive is available). Worth folding into the langwatch-saas tracking issue for these findings.

### P3 — popup-mode response for the new conflict branch is untested (non-blocking)

The route's `state.mode === "popup"` ternary (`github-langy.ts:369-371`) is exercised by other error branches in this file (lines 214, 281, 331 use `makeState({ mode: "popup" })`), but the new cross-tenant-conflict test only covers `mode: "redirect"`. Low risk since it's the same shared ternary already covered elsewhere, but worth a follow-up test for completeness.

### Not a finding — pre-existing, out of scope for this diff

`specs/langy/langy-github-install.feature` is cited in comments across six files (`github-langy.ts`, `langyGithubAppToken.ts`, `dbOrganizationIdProtection.ts`, two frontend files) but doesn't exist in `specs/`. Pre-existing gap, not touched or worsened by this PR.

### Security review of the fix itself
- No leak: the route's audit log and warn-log only record the installation id + the attacking org/user — never `existingOrganizationId` (the victim org), even though the error object carries it. Deliberate restraint, correctly not surfaced.
- Same generic message/status on both "installation doesn't exist" and "belongs to another org" — no oracle for an attacker to distinguish the two.
- Nonce single-use burn happens before `recordInstallation` runs, so this guard can't be hammered against a range of installation ids on one signed state — each attempt needs a fresh, session + `langy:manage`-gated `/install` round-trip.

No P0/P1 survived — nothing to refute via subagent. Approving.

sergioestebance:
## Addressed CodeRabbit's 3 comments + my own ruthless-review P2

**1. Non-atomic ownership check — valid, fixed.** Same TOCTOU I'd flagged as a non-blocking P2 in the earlier review comment; CodeRabbit independently caught it and correctly called it worth fixing outright rather than deferring. Replaced the `findByInstallationId` (read) → `upsert` (write) pair with `insertOrGetExisting`, which attempts a `create` and relies on the DB's unique index on `installationId` for atomicity, falling back to a lookup only on conflict. Added the requested concurrent regression test — two orgs racing `Promise.allSettled` for the same fresh installation id; asserts exactly one settles fulfilled and the other rejects with `LangyGithubInstallationConflictError`. Confirmed this test would have failed under the old code (no rejection, both would "succeed").

**2. Rename the route test to `.integration.test.ts` — valid, done.** It exercises the real Hono app end-to-end via `app.request`, mocking only the app-layer boundary (`getApp()`, session, auditLog, rbac) — matches this repo's own unit-vs-integration classification (`CODING_STANDARDS.md`: tests that render/wire the real thing and mock boundaries are integration tests). Renamed via `tslsp-cli rename-file`; all 19 tests still pass under `pnpm test:integration`.

**3. Redact `attemptedOrganizationId` from the warn log — not applied.** `attemptedOrganizationId` is `state.organizationId` — the authenticated caller's *own* org, already logged unredacted at 5 other spots in this same file (including the success-path audit log two lines below). Redacting only this one instance would be inconsistent with the file's existing convention and would remove exactly the detail an operator needs to identify which org's member attempted the rebind. `existingOrganizationId` (the victim org) — the one that actually matters to keep out of logs — was already excluded in the original fix. Skipping.

All 4 touched files typecheck clean (`tslsp-cli diagnostics`); `langy-github-installations.service.unit.test.ts` (14 tests) and the renamed `github-langy-install.integration.test.ts` (19 tests) pass.

sergioestebance:
## Second-guessing the fix against `TESTING_PHILOSOPHY.md`

Went back and asked whether the race-condition fix was actually proven, not just plausible.

**Gap found:** the `insertOrGetExisting` regression test (added for the earlier TOCTOU fix) used a hand-rolled fake repository whose "atomicity" was simulated by writing the check-and-set with no `await` in between. That's a real assumption baked into a mock, not a proof — it shows the *service* correctly interprets an atomic repo result, but nothing in the diff verified that Postgres's unique constraint on `installationId` actually serializes two concurrent `create()` calls the way the fix depends on. Per `TESTING_PHILOSOPHY.md`'s mocking guidance ("mock only at external boundaries... prefer real implementations where practical") and this repo's own precedent for exactly this class of claim (`dataset.repository.integration.test.ts`'s docstring makes the identical point about mocked tests being unable to prove a real DB operation), this needed a repository-level integration test against the actual test Postgres, not another mock.

**Fix:** added `langy-github-installations.prisma.repository.integration.test.ts`, exercising `PrismaLangyGithubInstallationsRepository.insertOrGetExisting` against the real test database — fresh insert, the same two-orgs-race scenario run for real via `Promise.all`, and confirmation a pre-existing id is never overwritten. Ran it — 3/3 pass, confirming Postgres's unique index genuinely resolves the race (not just the mock's model of it). Re-scoped the earlier service-level unit test's comment: it now correctly claims only "the service's branching logic," pointing at the new integration test for the actual DB guarantee.

No schema/FK setup needed (`relationMode = "prisma"` — no real FK enforced, matching the lighter `langy-turn-admission.prisma.integration.test.ts` precedent rather than the heavier org/team/project scaffolding some other integration tests need).

23/23 integration + 13/13 unit tests pass on the touched files; typecheck clean.

sergioestebance:
Fixed: `inserted` → `wasInserted` on `insertOrGetExisting`'s return shape — bare booleans need an is/has/was prefix per this repo's naming guideline. Renamed across the interface, Prisma implementation, Null stub, service, and every test double/assertion. 13/13 unit + 23/23 integration still pass, typecheck clean.

rogeriochaves (approved):

PR #6069, opened 2026-07-23
fix(mcp): validate redirect_uri against registered OAuth clients

sergioestebance:
## Summary
- `/oauth/register` now persists each client's `redirect_uris` (Redis-backed registry) instead of only echoing them back in the response.
- `/mcp/authorize` requires an exact match against a client's registered `redirect_uris` before issuing an authorization code, and binds the issued code to the exact `client_id` + `redirect_uri` it validated.
- `/oauth/token` now requires both `redirect_uri` and `client_id` at exchange time, and rejects the exchange unless they match what was bound to the code at authorization.

## Rollout consideration
Any MCP client that registered via `/oauth/register` **before** this ships has no entry in the new registry. Its next `/mcp/authorize` attempt will get `Unknown or unregistered client_id` until it re-registers. RFC 7591 doesn't guarantee a client reacts to that by re-registering automatically — most MCP client implementations do, since a fresh client_id is exactly what dynamic registration is for, but this is worth a quick check against whatever's actually connected before this merges.

## Test plan
- [x] `mcp-authorize.redirect-uri-binding.integration.test.ts` — registered vs. mismatched vs. unregistered vs. missing `client_id`, driving the full route with only DB/Redis mocked (moved from `.unit.test.ts` — it isn't pure-logic-in-isolation)
- [x] `mcp-authorize.rolebinding.unit.test.ts` — updated for the now-mandatory client lookup, still passing
- [x] `mcp-handler.integration.test.ts` — extended with end-to-end token-exchange binding checks and `/oauth/register` persistence coverage
- [x] `oauthClientRegistry.integration.test.ts` (new) — register→get round trip and defensive-parsing branches against **real Redis** (testcontainers), not mocks, per `dev/docs/best_practices/repository-service.md`
- [x] New `@regression @integration` scenarios in `specs/mcp-server/mcp-in-app.feature`, bound to the tests above via `@scenario` JSDoc annotations — `pnpm check:feature-parity` passes (10/10 bound)
- [x] `pnpm typecheck` and `pnpm typecheck:tests` clean

## Summary by CodeRabbit

* **Security Enhancements**
  * Hardened MCP OAuth flows with stricter validation for `/api/mcp/authorize` and `/oauth/token`.
  * Authorization and token exchange now require both `client_id` and an exact `redirect_uri` match bound to the original registration; missing or mismatched details return `400` without issuing tokens.
* **New Features**
  * Added Redis-backed persistence for dynamically registered OAuth clients and their allowed redirect URIs.
  * OAuth client registration now returns a normalized client name.
* **Tests**
  * Added integration/regression coverage for redirect URI binding, error cases, and registry behavior.
  * Improved test determinism via test-only rate-limiter cleanup.
* **Chores**
  * Tightened MCP dev-server proxy routing to avoid intercepting the authorize UI route.

### DIFF
--- a/langwatch/src/mcp/__tests__/mcp-handler.integration.test.ts
+++ b/langwatch/src/mcp/__tests__/mcp-handler.integration.test.ts
@@ -542,6 +542,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
     });
 
   describe("when the token exchange omits redirect_uri", () => {
+    /** @scenario Token exchange is rejected when redirect_uri is missing */
     it("returns 400 with invalid_request error", async () => {
       const addr = server.address();
       const port = typeof addr === "object" && addr ? addr.port : 0;
@@ -559,6 +560,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
   });
 
   describe("when the token exchange omits client_id", () => {
+    /** @scenario Token exchange is rejected when client_id is missing */
     it("returns 400 with invalid_request error", async () => {
       const addr = server.address();
       const port = typeof addr === "object" && addr ? addr.port : 0;
@@ -576,6 +578,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
   });
 
   describe("when the token exchange presents a different redirect_uri than the one bound at authorize", () => {
+    /** @scenario Token exchange is rejected when redirect_uri does not match the authorization request */
     it("returns 400 with invalid_grant and never issues a token", async () => {
       const code = randomUUID();
       const { codeVerifier, codeChallenge } = createPkceChallenge();
@@ -598,6 +601,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
   });
 
   describe("when the token exchange presents a different client_id than the one bound at authorize", () => {
+    /** @scenario Token exchange is rejected when client_id does not match the authorization request */
     it("returns 400 with invalid_grant and never issues a token", async () => {
       const code = randomUUID();
       const { codeVerifier, codeChallenge } = createPkceChallenge();
@@ -622,6 +626,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
   // --- Security: dynamic client registration persistence (RFC 7591) ---
 
   describe("when a new OAuth client registers", () => {
+    /** @scenario Dynamic client registration persists the redirect_uris binding */
     it("persists the client_id -> redirect_uris binding to Redis", async () => {
       const addr = server.address();
       const port = typeof addr === "object" && addr ? addr.port : 0;
@@ -653,6 +658,7 @@ describe("Feature: MCP HTTP Server In-App Integration", () => {
   });
 
   describe("when OAuth client registration is missing redirect_uris", () => {
+    /** @scenario Dynamic client registration rejects a request with no redirect_uris */
     it("returns 400 with invalid_client_metadata", async () => {
       const addr = server.address();
       const port = typeof addr === "object" && addr ? addr.port : 0;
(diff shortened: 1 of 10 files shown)


rogeriochaves (dismissed):

sergioestebance:
## Prove Them Wrong — live verification against the real reported exploit chain

Ran the stack for this worktree via `pnpm dev:haven` (`https://app.mcp-oauth-redirect-uri-audit.langwatch.localhost`), signed up a fresh test account, and reproduced the exact chain from the original security report end-to-end against the running app — not mocks.

Consent page still renders for a mismatched redirect_uri (unchanged, expected)

Clicking Allow now fails with "redirect_uri does not match any redirect URI registered for this client_id" — no redirect, no code

### Verdict table

| 1 | `/oauth/register` persists a client's `redirect_uris` (not just echoes them) | `POST /oauth/register` → 201, `client_id: mcp_e6471866c3144e9396973865519c11bd` bound to `https://registered.example/callback` | PASS |
| 2 | `/mcp/authorize` consent page still renders regardless of `redirect_uri` (unchanged UX, validation happens on Allow) | 01-consent-page-mismatched-redirect-uri.jpg | PASS |
| 3 | Clicking Allow with a `redirect_uri` that doesn't match the client's registration is rejected — the original exploit | 02-exploit-blocked-authorization-failed.jpg — toast: "redirect_uri does not match any redirect URI registered for this client_id"; page did not navigate away, no code issued | PASS |
| 4 | Legitimate flow (matching `redirect_uri`) still works | Browser navigated to `https://registered.example/callback?code=<uuid>&state=prove-them-wrong-legit` — a real code, delivered to the real registered URI | PASS |
| 5 | Full token exchange succeeds for the legitimate flow | `POST /oauth/token` with matching `redirect_uri`+`client_id`+PKCE verifier → 200, `access_token` issued | PASS |
| 6 | `/oauth/token` independently rejects a mismatched `redirect_uri` at exchange time, even for a genuinely-issued code (defense in depth, not just at authorize) | Fresh code minted for `registered.example`, exchanged with `redirect_uri=attacker.invalid` → 400 `invalid_grant`, "redirect_uri does not match the authorization request" | PASS |
| 7 | Authorization codes are one-time-use even when an exchange attempt is rejected (no replay window from a failed mismatched attempt) | Re-exchanging the same code from claim 6 with the *correct* `redirect_uri` afterward → 400 "Invalid or expired authorization code" | PASS |

**7/7 proven, 0 disproven, 0 unverified.**

### Findings (anomalies logged along the way)

- **Pre-existing, unrelated dev-environment bug found and fixed**: under the `pnpm dev:haven` split app/api setup, Vite's dev proxy matched `/mcp` as a plain prefix (`vite.config.ts`), which also swallowed `/mcp/authorize` — a legitimate frontend page — and forwarded it to the API server, which has no dev-mode SPA fallback, producing a bare 404. This blocked driving the browser test at all until fixed. Narrowed the proxy to exact-match `^/mcp$` / `^/mcp/health$` only; confirmed the real MCP routes (`/mcp/health`, `/oauth/register`) still proxy correctly afterward. Not part of the original security fix — a local-dev routing gap that would hit anyone testing this page under haven.
- No console errors, no red alerts, no broken layout observed in any captured frame.

### Not covered

- Did not test against Claude Desktop / Cursor / a real third-party MCP client — verification used direct HTTP requests + the browser consent flow, which exercises the same server-side code path a real client would hit.
- Did not verify the rollout-risk scenario (pre-existing registered clients from before this fix) since this is a fresh worktree with no pre-fix data to migrate — that risk is already called out in the PR description as a pending ops check, not something this environment can reproduce.

### Environment

`https://app.mcp-oauth-redirect-uri-audit.langwatch.localhost`, fresh local Postgres/Redis/ClickHouse via `pnpm dev:haven`, account `provethemwrong@langwatch.ai`.

**Verdict: 7/7 proven, 0 disproven, 0 unverified — the reported redirect_uri exfiltration is confirmed fixed on `https://app.mcp-oauth-redirect-uri-audit.langwatch.localhost`.**

rogeriochaves (approved):

PR #6081, opened 2026-07-23
fix(coding-agent): read session fold state back from ClickHouse instead of refolding event_log (ADR-066)

0xdeafcafe:
First adopter of **ADR-066** (\"event_log off the per-item hot path\", PR #6079) and the durable fix for the **2026-07-23 production outage**.

## The outage mechanism
The \`codingAgentSession\` fold's store returned \`null\` from \`get()\` by design: it persists a *lossy* analytics row that cannot reconstruct fold state. That forced the executor to refold the aggregate's **entire** \`event_log\` history on **every cache miss** *and* **every out-of-order delivery**. On large coding-agent sessions the session key falls back to the traceId, so one big trace is one huge aggregate — those refolds are 20–100 MB S3-walking \`event_log\` reads. They starved ClickHouse merges, produced \`TOO_MANY_PARTS\`, and took the platform down.

## The read-back fix
Make the store read its **own last committed state** back from ClickHouse (a bounded one-row read, fronted by the Redis cache) instead of refolding, and stop treating out-of-order delivery as a reason to replay (the derivation is order-insensitive).

## What changed
- **Migration `00053_add_coding_agent_session_state.sql`** — adds one lossless column to `coding_agent_sessions`:
  \`\`\`
  State String DEFAULT '' CODEC(ZSTD(3))
  \`\`\`
  The row now has \"two faces\": the queryable analytics columns **and** a full serialized fold-state resume blob. The blob carries the three fields the analytics columns deliberately drop — \`subAgentIds\`, \`previousCallContextTokens\`, \`metricSeries\` — so it round-trips the working state exactly. 00051/00052 are untouched (deployed, immutable).
- **Persist the full state** — \`CodingAgentSessionRow\` gains \`state: string\`; \`projectCodingAgentSessionToRow\` sets \`state: JSON.stringify(state)\`.
- **Write + read the column** — the ClickHouse repo maps \`State\` in \`toRecord\`/\`fromRecord\`; \`findBySessionId\` (already \`SELECT *\` with IN-tuple latest-version dedup) now returns it. The fold read-back path knows only tenantId+sessionId, so that read is unpruned but bounded (one small row) and cache-fronted — far cheaper than the \`event_log\` replay it replaces.
- **`get()` reads back** — \`findBySessionId\` → \`JSON.parse\` the blob into \`CodingAgentSessionState\`; empty/missing blob → \`null\` (framework calls \`init()\`); a corrupt blob logs a warning and degrades to init rather than wedging the group. Mirrors \`experimentRunState.store.ts\` (\"dumb read/write — state IS the data\").
- **`refoldOnOutOfOrder: false`** — replaces \`refoldOnStoreMiss: true\`. The derivation is order-insensitive (\`appendStep\` inserts by \`startedAtMs\`; sets are first-seen-bounded; counters additive; metrics converge by series id), so out-of-order needs no replay and a miss now needs no refold.

## Tests
- **Spec** — \`specs/coding-agent/session-aggregate.feature\` gains a resume-after-cache-loss scenario (resumes from stored state without re-reading history; totals accumulate; re-delivered telemetry does not inflate).
- **Unit** (colocated) — round-trips a state exercising all three bookkeeping fields through the real \`projectCodingAgentSessionToRow → toRecord → fromRecord → store.get()\` path and asserts deep-equality; a metric applied after read-back does not double-count (proves \`metricSeries\` survived); plus null / empty / corrupt-blob cases.
- **Integration** — \`coding-agent-session.clickhouse.repository.integration.test.ts\` round-trips the \`State\` blob through real ClickHouse (migration auto-applied by the goose harness).

## Test commands + results
\`\`\`
pnpm test:unit run src/server/event-sourcing/pipelines/coding-agent-processing/projections/__tests__/codingAgentSession.store.unit.test.ts          # 5 passed
pnpm test:unit run .../codingAgentSession.foldProjection.unit.test.ts .../sessionSignals.unit.test.ts .../coding-agent-session.service.unit.test.ts  # 32 passed
pnpm test:integration run src/server/app-layer/coding-agent/repositories/__tests__/coding-agent-session.clickhouse.repository.integration.test.ts    # 5 passed
\`\`\`

A separate tiny relief PR may also set \`refoldOnOutOfOrder: false\`; the change is idempotent, so a later merge conflict there is harmless.

### DIFF
--- a/langwatch/src/features/traces-v2/components/TraceDrawer/sessionView/__tests__/sessionSignals.unit.test.ts
+++ b/langwatch/src/features/traces-v2/components/TraceDrawer/sessionView/__tests__/sessionSignals.unit.test.ts
@@ -15,6 +15,7 @@ function session(
     traceIds: ["trace-1"],
     version: "2026-07-11",
     startedAtMs: 1_700_000_000_000,
+    state: "{}",
     agent: "claude-code",
     agentVersion: "2.0.0",
     sessionId: "session-1",
--- a/langwatch/src/server/app-layer/coding-agent/repositories/__tests__/coding-agent-session.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/coding-agent/repositories/__tests__/coding-agent-session.clickhouse.repository.integration.test.ts
@@ -39,6 +39,7 @@ function sessionRow(
     sessionKeySource: "provider",
     version: "2026-07-21",
     startedAtMs: baseMs,
+    state: "{}",
     agent: "claude_code",
     agentVersion: "2.0.0",
     traceIds: [`${tag}-t1`],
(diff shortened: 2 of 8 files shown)


0xdeafcafe:
Superseded by #6082, which implements the same read-back fix as the generic platform store (RedisCachedFoldStore cache/fallthrough/dual-write + typed fromRow decode) instead of hand-rolling recovery inside this projection's own store. Closing in favour of #6082 (ADR-066, stacked on #6079).

PR #6099, opened 2026-07-24
feat(profile): let users upload a custom avatar photo (#6086)

### LINKED ISSUE #6086: Let users upload/set a custom avatar photo (currently SSO-only)
## Summary

Avatar images across the app — presence avatars, member lists, and now the Prompt Version History author avatars (#6083) — are populated **only** from a user's SSO/OAuth profile picture. There is no way for a user to upload or change their own avatar: email/password users have no photo at all, and SSO users are stuck with whatever their identity provider returns.

## Current behavior

- `User.image` (a text URL column) is written **only** by the better-auth social providers at sign-in (`langwatch/src/server/better-auth/index.ts`: Google / GitHub / GitLab / Microsoft / Auth0).
- No profile / account-settings UI exposes an avatar field.
- Every avatar render falls back to initials → silhouette when `image` is absent (e.g. `src/features/presence/components/PresenceAvatar.tsx`, and the new `VersionAuthor` in #6083).

## Proposed

Let a user set their own avatar photo from account settings:

- An upload endpoint that stores the image in the existing S3-backed object storage and writes the resulting URL to `User.image`.
- A profile/account-settings control to upload, preview, and remove the photo.
- Sensible constraints (max size, allowed types, client-side crop/resize).
- Keep the existing fallback chain when unset: uploaded photo → SSO image → initials → silhouette.

## Considerations

- **Precedence vs. SSO image**: does an uploaded photo win over the provider's, and should a later SSO login overwrite a user-set photo? (Probably: user upload wins and is not clobbered by SSO.)
- CSP / broken URLs are already handled — avatars fall back to initials when an `` 404s or is blocked.
- Consider whether this is per-user (account) or also team-level.

_Context: raised while reviewing #6083 (surfacing `User.image` as the prompt version-history author avatar)._

Aryansharma28:
Closes #6086

## The story — a photo that follows you everywhere

Today a person's picture in LangWatch comes **only** from their SSO/OAuth provider. Email/password users get a gray monogram forever, and SSO users are stuck with whatever their identity provider returns. There's no way to set your own.

This PR lets anyone upload their own photo from account settings. And because every avatar in the product resolves the single `User.image` field through one fallback chain — **uploaded → SSO → initials → silhouette** — that photo then shows up **everywhere a person appears**.

Here's the whole journey, end-to-end (real screenshots from the running app).

### 1 · Before — just a monogram, with a pencil to change it

Your profile (**My Settings → Profile**) shows a gray “HA” with a small **pencil badge** on it — the photo itself is the edit control. Everywhere else (header, member lists, trace authors) is initials too.

### 2 · Click the pencil → live, square-cropped preview

Click the **pencil on your photo** and pick an image (**PNG / JPG / WebP / GIF, up to 8 MB**). It's center-cropped to a square and downscaled to 256px **in the browser** — so what's stored is a few KB — and you see exactly what you'll get *before* committing. The buttons become **Save photo / Cancel**.

### 3 · Save — and it's instantly you, everywhere

Hit **Save photo**. The profile updates **and the header avatar (top-right) changes live in the same click** — no refresh — because the session refetches. The pencil stays for next time, and a **Remove** appears.

### 4 · Your teammates see it — Members

Your photo now renders next to your name in the organization members list. (A pending invite with no account yet still falls back to initials, exactly as it should.)

### 5 · …and inside Traces

Open a trace and leave an annotation — your photo is the **author avatar** on it, so a teammate reviewing that trace sees who left the note.

---

## What changed

**The edit affordance** — the “Upload photo” button is gone; the **avatar itself is the click target**, with a pencil badge overlaid (`role="button"` + keyboard handler for a11y). Save/Cancel (preview) and Remove remain.

**Size limit** — the accepted upload is capped at **8 MB** (matching LinkedIn's profile-photo limit), enforced both client-side on the picked file (immediate “too large” feedback) and server-side on the payload.

**Storage & serving**
- A new `user_avatar` purpose on the existing S3-backed stored-objects service; the bytes are owned by the *user*, not a tenant (an avatar is platform identity, not tenant data).
- A dedicated `GET /api/user-avatar/:projectId/:id` route, readable by **any authenticated user** (avatars must be visible to teammates across projects) but that **only ever serves avatar objects** — so this broad-read route can never be used to reach the trace/scenario media that `/api/files` gates behind `traces:view`. (Still strictly *more* private than today, where SSO photos are fetched from fully public CDN URLs.)
- The stored-object HTTP hardening (Content-Type coercion + `nosniff`/CSP headers) was extracted into a shared module rather than copy-pasted.

**Upload** — service-layered `user.setAvatar` / `user.removeAvatar` tRPC mutations. The client crops + resizes to 256px, so base64-over-tRPC stays tiny.

**Precedence** — better-auth writes `User.image` only on user *create*, never on re-login, so an uploaded photo is **never clobbered** by a later SSO sign-in (locked with a test).

**Rendering everywhere** — added an optional `image` to the shared `RandomColorAvatar` via a new canonical `UserAvatar` (image → initials → silhouette), then threaded `image` into every person-avatar surface that previously dropped it: header account menu, org members, teams, groups, role-bindings, all trace/annotation surfaces (header chips, per-turn, annotations view, cards, expected-outputs, comment composer, queue members), the participant picker, the optimization-studio version-history author, and the annotation-queue author clusters.
  - The **prompt** version-history popover is intentionally left to open PR #6083, which owns it (avoids a merge conflict).

## Testing / QA

- `pnpm typecheck` — clean across all changed files.
- `pnpm test:unit` — data-URL validation, service orchestration, `/api/files` regression after the shared-helper extraction.
- Spec: `specs/settings/user-avatar.feature`.
- **Manually verified end-to-end in the running app** — every screenshot above is the real feature (pencil → crop → save → header / profile / members / traces), including removing and re-adding.

## Notes

- Self-hosted deployments without S3 fall back to the local-filesystem object store; that path (`LANGWATCH_LOCAL_STORAGE_PATH`, default `/var/lib/langwatch/objects`) must be writable — the Helm chart provisions it. (Surfaced during QA on a bare dev box where the default path wasn't writable.)
- Screenshots are hosted on the `avatar-6086-assets` branch — delete it after merge.

## Summary by CodeRabbit

* **New Features**
  * Added custom avatar upload with preview, square cropping, save, change, and remove, including file-type and size validation.
  * Introduced a hardened avatar-serving endpoint with authentication, streaming, caching headers, and per-caller rate limiting.
  * Updated the UI to consistently display uploaded/SSO avatars via a shared avatar component, with initials fallback when no image is available.
  * Uploaded avatars now take precedence over SSO photos and are not overwritten by later sign-ins.
* **Bug Fixes**
  * Avatar loading now retries correctly when the image URL changes after a previous failure.
* **Tests**
  * Added end-to-end avatar spec plus unit coverage for avatar upload/removal, parsing/URL building, and SSO precedence.

### DIFF
--- a/langwatch/src/app/api/user-avatar/[[...route]]/app.ts
+++ b/langwatch/src/app/api/user-avatar/[[...route]]/app.ts
@@ -3,8 +3,9 @@
  *
  * Unlike /api/files (which gates each read behind the caller's `traces:view` /
  * `scenarios:view` on the owning project), an avatar is platform-level identity
- * that must be visible to any teammate on any surface. So this route authorizes
- * ANY authenticated caller — but for exactly that reason it MUST only ever
+ * that must be visible wherever a person is shown, so this route authorizes ANY
+ * authenticated caller on the platform — not only users who share an
+ * organization with the uploader. For exactly that reason it MUST only ever
  * serve objects tagged with the `user_avatar` purpose. Any other purpose is a
  * 404 here, so this broad-read route can never be used to exfiltrate the
  * trace/scenario media that /api/files protects.
@@ -22,11 +23,14 @@ import { rateLimit } from "~/server/rateLimit";
 import {
   jsonResponse,
   rateLimitedResponse,
-  safeMediaType,
   STORED_OBJECT_RESPONSE_BASE_HEADERS,
 } from "~/server/stored-objects/media-response";
 import { createStoredObjectsService } from "~/server/stored-objects/stored-objects-factory";
-import { AVATAR_OWNER_KIND, AVATAR_PURPOSE } from "~/server/user-avatar/avatar";
+import {
+  AVATAR_OWNER_KIND,
+  AVATAR_PURPOSE,
+  safeAvatarMediaType,
+} from "~/server/user-avatar/avatar";
 import type { DualAuthVariables } from "../../middleware/dual-auth";
 import { dualAuth } from "../../middleware/dual-auth";
 
@@ -50,7 +54,7 @@ function streamAvatarResponse({
   method: "GET" | "HEAD";
 }): Response {
   const headers: Record<string, string> = {
-    "Content-Type": safeMediaType(row.media_type),
+    "Content-Type": safeAvatarMediaType(row.media_type),
     "Content-Length": String(row.size_bytes),
     // Content-addressed id => the bytes at a URL never change, so the browser
     // can cache aggressively. A new upload mints a new id (new URL), and a
--- a/langwatch/src/app/api/user-avatar/[[...route]]/app.ts
+++ b/langwatch/src/app/api/user-avatar/[[...route]]/app.ts
@@ -1,0 +1,154 @@
+/**
+ * /api/user-avatar/:projectId/:id — serves a user's uploaded avatar bytes.
+ *
+ * Unlike /api/files (which gates each read behind the caller's `traces:view` /
+ * `scenarios:view` on the owning project), an avatar is platform-level identity
+ * that must be visible to any teammate on any surface. So this route authorizes
+ * ANY authenticated caller — but for exactly that reason it MUST only ever
+ * serve objects tagged with the `user_avatar` purpose. Any other purpose is a
+ * 404 here, so this broad-read route can never be used to exfiltrate the
+ * trace/scenario media that /api/files protects.
+ *
+ * (This is still strictly more private than the status quo: SSO provider photos
+ * are fetched from fully public CDN URLs today.)
+ *
+ * Spec: specs/settings/user-avatar.feature
+ */
+import { Readable } from "node:stream";
+import type { MiddlewareHandler } from "hono";
+import { HTTPException } from "hono/http-exception";
+import { anyAuthenticated, createServiceApp } from "~/server/api/security";
+import { rateLimit } from "~/server/rateLimit";
+import {
+  safeMediaType,
+  STORED_OBJECT_RESPONSE_BASE_HEADERS,
+} from "~/server/stored-objects/media-response";
+import { createStoredObjectsService } from "~/server/stored-objects/stored-objects-factory";
+import { AVATAR_OWNER_KIND, AVATAR_PURPOSE } from "~/server/user-avatar/avatar";
+import type { DualAuthVariables } from "../../middleware/dual-auth";
+import { dualAuth } from "../../middleware/dual-auth";
+
+const secured = createServiceApp<{ Variables: DualAuthVariables }>({
+  basePath: "/api/user-avatar",
+  verifySecret: dualAuth,
+});
+
+// Avatars render in dense stacks (member lists, presence bars), so the budget
+// is looser than /api/files; still caps enumeration abuse from one credential.
+const AVATAR_RATE_LIMIT_WINDOW_SECONDS = 60;
+const AVATAR_RATE_LIMIT_MAX = 240;
+
+function jsonResponse(body: unknown, status: number): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: {
+      "Content-Type": "application/json",
+      ...STORED_OBJECT_RESPONSE_BASE_HEADERS,
+    },
+  });
+}
+
+function streamAvatarResponse({
+  row,
+  stream,
+  method,
+}: {
+  row: { size_bytes: number; media_type: string };
+  stream: Readable;
+  method: "GET" | "HEAD";
+}): Response {
+  const headers: Record<string, string> = {
+    "Content-Type": safeMediaType(row.media_type),
+    "Content-Length": String(row.size_bytes),
+    // Content-addressed id => the bytes at a URL never change, so the browser
+    // can cache aggressively. A new upload mints a new id (new URL), and a
+    // removal drops the reference entirely, so a cached entry is never stale.
+    "Cache-Control": "private, max-age=86400",
+    ...STORED_OBJECT_RESPONSE_BASE_HEADERS,
+  };
+
+  if (method === "HEAD") {
+    stream.destroy?.();
+    return new Response(null, { status: 200, headers });
+  }
+  return new Response(Readable.toWeb(stream) as ReadableStream, {
+    status: 200,
+    headers,
+  });
+}
+
+async function handleAvatarRead(
+  c: Parameters<MiddlewareHandler<{ Variables: DualAuthVariables }>>[0],
+  options: { method: "GET" | "HEAD" },
+): Promise<Response> {
+  const projectId = c.req.param("projectId");
+  const id = c.req.param("id");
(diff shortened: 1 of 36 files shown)


langwatch/src/server/user-avatar/avatar.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (!isAllowedMediaType(mediaType)) {
+    throw new AvatarValidationError(
+      "invalid_type",
+      `Unsupported image type "${mediaType}". Allowed: ${AVATAR_ALLOWED_MEDIA_TYPES.join(", ")}.`,
+    );
+  }
+
+  // Strip any embedded whitespace/newlines the encoder may have inserted so
+  // the byte-length check reflects real decoded size, not base64 padding.
+  const bytes = Buffer.from(match[2]!.replace(/\s+/g, ""), "base64");
```
rogeriochaves, line 166: **P3 (defense-in-depth)** — Validation trusts the client-declared MIME in the `data:` prefix and never inspects the decoded bytes, so a direct API caller (bypassing the browser crop step) can send `data:image/png;base64,<arbitrary bytes>` and store non-image content tagged `image/png`.

This is **not** a stored-XSS vector: the serve route emits `X-Content-Type-Options: nosniff` + `Content-Security-Policy: default-src 'none'; sandbox`, so the blob can never be interpreted as active content. The residual is minor — an authenticated user can use avatar storage as an ~8 MB arbitrary blob store, bounded by the 10/min limit and their own personal-workspace quota (data-integrity, not confidentiality).

Minimal fix: after `Buffer.from(...)`, sniff the leading magic bytes (PNG `89504E47`, JPEG `FFD8FF`, WebP `RIFF....WEBP`, GIF `GIF8`) and reject when they disagree with the declared type — cheap, and closes the gap without depending solely on response headers.

langwatch/src/app/api/user-avatar/[[...route]]/app.ts
```diff
@@ -0,0 +1,137 @@
...
+  row,
+  stream,
+  method,
+}: {
+  row: { size_bytes: number; media_type: string };
+  stream: Readable;
+  method: "GET" | "HEAD";
+}): Response {
+  const headers: Record<string, string> = {
+    "Content-Type": safeMediaType(row.media_type),
```
Comment (rogeriochaves) [inline langwatch/src/app/api/user-avatar/[[...route]]/app.ts]: **P3 (latent / keep-it-intentional)** — `safeMediaType()` → `isReadbackSafe()` passes the entire `image/` prefix family verbatim, which includes `image/svg+xml` (SVG can carry inline JS). This route is safe **today** only because two things line up: the upload allowlist (`AVATAR_ALLOWED_MEDIA_TYPES`) excludes SVG so no `user_avatar` object can carry that media type, and the `sandbox` CSP would neutralize SVG script even if one did. Since this is a *broadly readable* route, that means the CSP header is the last line of defense.

Minimal hardening: have this route pin Content-Type against the avatar allowlist directly (png/jpeg/webp/gif) instead of deferring to the broader `isReadbackSafe`, so a future object-shape change can never widen it to svg here. No behavior change today.

langwatch/src/app/api/user-avatar/[[...route]]/app.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  return streamAvatarResponse({
+    row: result.row,
+    stream: result.stream,
+    method: options.method,
+  });
+}
+
+secured
+  .access(anyAuthenticated())
```
Comment (rogeriochaves) [inline langwatch/src/app/api/user-avatar/[[...route]]/app.ts:134]: **P3 (residual — no change required)** — `anyAuthenticated()` authorizes *any* signed-in user in *any* organization, whereas the file header frames the audience as "teammates." This is an acceptable capability-URL design: the id is content-addressed / unguessable, the `projectId`+`id` pair only surfaces in org-scoped member/annotation payloads a non-teammate cannot read, and the `purpose`+`owner_kind` gate blocks reaching non-avatar media — so there is no practical cross-org leak. Flagging only so the intent-vs-implementation gap ("teammate" vs "any authenticated caller") is a conscious, documented decision rather than an accident.

rogeriochaves (commented):
**Verdict: no P0/P1/P2 — the image-upload path is well-secured.** Scoped this pass to the security boundaries you asked about (where bytes land, who can write, who can read, how they are served).

**Where the image goes:** client crops/resizes to a 256px WebP data URL → tRPC `user.setAvatar` (session-auth, `organization:view` membership check, 10/min rate limit) → validated + size-capped decode → stored via the content-addressed stored-objects service under the *user's own personal-workspace project*, tagged `purpose=user_avatar` / `owner_kind=user`. `User.image` is set to a same-origin `/api/user-avatar/:projectId/:id`. No remote-URL fetch anywhere, so no SSRF.

**How it is served:** `GET /api/user-avatar/:projectId/:id`, `dualAuth` (401 if unauthenticated) → any authenticated caller, 240/min per caller, but only returns rows where `purpose=user_avatar && owner_kind=user` (else 404), with `nosniff` + `Content-Security-Policy: default-src 'none'; sandbox` + coerced Content-Type. I verified the write path is the *only* producer of `user_avatar` objects (no HTTP/tRPC surface accepts a client-controlled `purpose`), so this broad-read route can never reach trace/scenario media — the boundary claim in the file header holds. CH lookups are parameterized; SVG is correctly excluded from the upload allowlist.

3 × **P3** defense-in-depth / residual notes inline (nothing blocking). The one worth acting on is server-side magic-byte validation.

langwatch/src/server/user-avatar/avatar.ts
rogeriochaves: Fixed in 76a6d9c08: `parseAvatarDataUrl` now checks the decoded bytes' magic-byte signature against the declared type and throws a new `content_mismatch` validation error on mismatch. Covered by new unit tests (valid signatures for jpeg/gif/webp, plus two mismatch cases).

langwatch/src/app/api/user-avatar/[[...route]]/app.ts
Comment (rogeriochaves) [reply langwatch/src/app/api/user-avatar/[[...route]]/app.ts]: Fixed in 76a6d9c08: added a route-specific `safeAvatarMediaType` in `avatar.ts`, pinned to `AVATAR_ALLOWED_MEDIA_TYPES` (png/jpeg/webp/gif only), and the serve route now uses it instead of the general `safeMediaType`/`isReadbackSafe`. `image/svg+xml` now coerces to `application/octet-stream` here regardless of what the general stored-objects allowlist ever widens to. Covered by new unit tests.

langwatch/src/app/api/user-avatar/[[...route]]/app.ts
Comment (rogeriochaves) [reply langwatch/src/app/api/user-avatar/[[...route]]/app.ts:134]: Updated the file header comment in 76a6d9c08 to say what the code actually does: "authorizes ANY authenticated caller on the platform — not only users who share an organization with the uploader." No behavior change, just closing the intent-vs-implementation gap.

PR #6100, opened 2026-07-24
fix(prompts): show full version commit messages, describe local-file updates

Aryansharma28:
## Summary

Prompt Version History showed only 2-3 words of a version's commit message, no matter how descriptive it was.

- **The commit message was clamped to one line, and a competing flex `Spacer` left that line only a sliver wide** whenever the current-version tag or "Discard local changes" button was present (`VersionHistoryListPopover.tsx`). `Spacer` has `flex: 1`, same as the message row, so the two split the available width evenly instead of the message getting it. Combined with `lineClamp={1}`, this meant even a full 200-character description showed as `Upda…`.
- **Syncing a prompt from a local file without an explicit commit message always recorded the generic "Updated from local file"**, even though `compareConfigContent` already computes exactly what changed (model, temperature, prompt, messages, etc.) for CLI conflict reporting. That diff just wasn't reused for the commit message. Now: `Updated from local file (model: gpt-4o → gpt-4o-mini; temperature: 0.7 → 0.3)`.

Other auto-generated messages ("Initial version", `Duplicated from "x"`, `Restore from version N`) already carry meaningful context and were left as-is — this only enriches the one generic case where a real diff was sitting right there unused.

### Before

before

### After

after

### After, with the "Discard local changes" button present

The `Spacer` removal is the risky part of this change (it's what used to push that button to the row's right edge), so this confirms the layout still holds with the button in play:

after with discard button

## Test plan

- [x] `describe-local-file-update.unit.test.ts` — new pure-function tests for the diff-to-message formatting
- [x] `syncPrompt.unit.test.ts` — enriched message wired through `syncPrompt`, explicit commit messages still take priority
- [x] `VersionHistoryListPopover.test.tsx` — long commit message renders in full
- [x] New feature file `specs/prompts/prompt-version-detail-visibility.feature`
- [x] `pnpm typecheck` clean
- [x] Manually verified in a running dev server via Playwright (screenshots above), including the layout with the "Discard local changes" button present

## Summary by CodeRabbit

- **New Features**
  - Prompt version history now shows long commit messages in full using wrapped, multi-line rendering (no truncation).
  - When syncing prompt updates from a local file without a custom commit message, the generated commit message now highlights which fields/runtime parameters changed.

- **Bug Fixes**
  - Improved the version-history popover header layout so the “discard local changes” action no longer has an unintended gap.

- **Tests**
  - Added/extended integration and unit tests to verify full long-message rendering and detailed, field-level commit message generation.

### DIFF
--- a/langwatch/src/prompts/VersionHistoryListPopover.tsx
+++ b/langwatch/src/prompts/VersionHistoryListPopover.tsx
@@ -5,7 +5,6 @@ import {
   Button,
   HStack,
   Separator,
-  Spacer,
   Spinner,
   Tag,
   Text,
@@ -163,9 +162,15 @@ function VersionHistoryItem({
       <HStack width="full" gap={3} align="start">
         <VersionNumberBox version={data} minWidth="48px" />
         <VStack align="start" width="full" gap={1}>
-          <HStack width="full">
-            <HStack gap={2} flex={1} minWidth={0}>
-              <Text fontWeight={600} fontSize="13px" lineClamp={1}>
+          <HStack width="full" align="start">
+            <HStack gap={2} flex={1} minWidth={0} align="start">
+              <Text
+                fontWeight={600}
+                fontSize="13px"
+                wordBreak="break-word"
+                flex="1"
+                minWidth={0}
+              >
                 {data.commitMessage}
               </Text>
               {isCurrent && (
@@ -179,7 +184,6 @@ function VersionHistoryItem({
                 </Tag.Root>
               )}
             </HStack>
-            <Spacer />
             {/* Discard changes button - reloads current version (same as "Load this version") */}
             {isCurrent && hasUnsavedChanges && (
               <Button
--- a/langwatch/src/prompts/VersionHistoryListPopover.tsx
+++ b/langwatch/src/prompts/VersionHistoryListPopover.tsx
@@ -170,6 +170,10 @@ function VersionHistoryItem({
                 wordBreak="break-word"
                 flex="1"
                 minWidth={0}
+                // Generous enough that the 200-char message the Save Version
+                // dialog allows never clips; still bounds a pathological
+                // message set via the API/SDK, which has no length limit.
+                lineClamp={8}
               >
                 {data.commitMessage}
               </Text>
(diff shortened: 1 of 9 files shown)


rogeriochaves (dismissed):

Aryansharma28:
Follow-up commit: bounded the commit-message display to \`lineClamp={8}\` instead of fully unbounded. The server has no length limit on \`commitMessage\` (only the Save Version dialog's UI caps it at 200 chars), so an API/SDK-set message could otherwise render arbitrarily long and dominate the popover. 8 lines comfortably fits the 200-char UI max without clipping (verified a 196-char message renders with `scrollHeight === clientHeight`), while still bounding a pathological case.

Aryansharma28:
Addressed both CodeRabbit findings:
- Parameter-only local updates (only `max_tokens`/`temperature`/etc. changed, config content identical) now describe the parameter change instead of falling back to the generic message — added `diffRuntimeParameters`, combined with `compareConfigContent`'s differences.
- The "long commit message" test now asserts the actual computed `-webkit-line-clamp`/`overflow` values instead of just DOM-text presence, so it would have caught the original single-line-clamp bug (verified by reverting the fix locally and confirming the test fails).

Aryansharma28:
Addressed the second round of CodeRabbit findings: \`diffRuntimeParameters\` now takes a named \`{ localParameters, remoteParameters }\` object (was positional), canonicalizes nested values with \`sortKeysDeep\` to match \`runtimeParametersEqual\` (avoids false diffs from key reordering) and distinguishes a missing key from an explicit \`undefined\`, and the test suite is restructured under a \`when\` context with a real two-key change in the multi-key test.

Aryansharma28:
Third CodeRabbit round: fixed a real bug where a missing runtime-param key and an explicit \`undefined\` value rendered identically in the diff message (both as the string \`"undefined"\`, since \`JSON.stringify(undefined)\` can't distinguish them). Missing keys now render as \`unset\`. Added tests for nested-key reordering (no false diff) and explicit-undefined-vs-missing in both directions.

rogeriochaves (approved):

PR #5118, opened 2026-06-25
feat(experiments-v3): Bradley-Terry leaderboard aggregation (#5103)

Aryansharma28:
# Related Issue

- Resolves #5103
- **Parent epic:** #5099
- **Stacked on #5107 (#5101). DO NOT MERGE TO MAIN UNTIL #5107 LANDS.** Base branch is set to `issue5101/n-way-compare` so this PR's diff only shows what #5103 adds on top of the N-way work. Re-target to `main` after #5107 merges.
- **Independent of #5102 and #5104.** No shared files in flight.

## Summary

Adds a Bradley-Terry maximum-likelihood leaderboard for the pairwise / N-way compare evaluator. The existing `AggregateHeaderBar` win-rate tally stays as-is; the BT leaderboard is a separate panel surfaced behind a default-off feature flag, so flag-off behavior is byte-identical to today.

### What ships
- **`computeBTLeaderboard.ts` + tests** (572 lines) — pure-function module that takes verdict rows and returns per-variant strength estimates with ranking. MLE fit, no external deps.
- **`PairwiseLeaderboard.tsx`** (387 lines) — leaderboard panel UI. Renders the BT ranking alongside (not replacing) the win-rate tally.
- **Feature flag** — `release_ui_pairwise_bt_aggregation_enabled` (PRODUCT scope, default off). Flag-off keeps today's `AggregateHeaderBar` as the only aggregate surface.
- **BDD spec** — `specs/experiments/pairwise-bt-leaderboard.feature` covers panel visibility gating, BT MLE correctness, and interaction with the existing header bar.

## Test plan

- [ ] Unit tests pass: `computeBTLeaderboard.test.ts`
- [ ] BDD scenarios in `specs/experiments/pairwise-bt-leaderboard.feature` cover the panel flow
- [ ] Manual UI smoke with flag on: run a 4-variant pairwise eval, panel renders with sensible ranking
- [ ] Manual UI smoke with flag off: behavior identical to current main (no panel, no compute call)
- [ ] No regression on #5100 + #5101 pairwise paths

## Out of scope (separate issues)

- Promote winner → prompt versioning — #5104
- Cross-experiment comparison — #5102
- PDF / CSV export — #5105

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #5105, opened 2026-06-24
Pairwise export to PDF/CSV


Aryansharma28:
# Pairwise export to PDF/CSV

> Parent epic: **#5099**
> Depends on: **#5100** — which ships markdown export.

## Why

The MVP exports **markdown** — useful for slack/notion handoffs, but:
- **Stakeholder reports** (sharing with PM / leadership / customers) want **PDF** with styled tables and proper formatting.
- **Further analysis** (loading into Excel/Sheets / a notebook) wants **CSV** with one row per evaluation.

This issue adds both formats to the existing export action.

## Persona

Same DS/QA, but now in **handoff mode** rather than iteration:
- DS handing results to a non-technical PM ("here's the PDF").
- QA generating a regression report for a release sign-off email.
- Either, doing post-hoc analysis in pandas/Excel ("give me the CSV").

### IN

- **PDF export**: render the existing markdown report through an HTML→PDF pipeline (`puppeteer` or equivalent).
- **CSV export**: per-row data with columns:
  - `row_id`, `input`, `golden`, `output_a`, `output_b`, `winner`, `reasoning`, `cost_a`, `cost_b`, `duration_a`, `duration_b`
- **Format selector** dropdown next to the "Export" button in `AggregateHeaderBar.tsx`: Markdown / PDF / CSV.
- **Respect active filters** — exporting after clicking "Losses only" exports just the loss rows.

### OUT

- **Custom report templates** — users can edit markdown manually if needed.
- **HTML export** — markdown handles this for most workflows.
- **Scheduled / recurring exports** — separate feature.
- **Export with charts embedded** (e.g. BT leaderboard) — defer until BT lands.

## Implementation

- **Markdown → PDF**: check existing puppeteer / playwright deps. LangWatch may already have headless Chrome for screenshots; reuse if so.
- **CSV**: simple row-mapper from per-row verdict state. Use a thin CSV serializer (`papaparse` or hand-rolled — avoid heavy deps).
- **Filename**: `pairwise_<experiment-name>_<YYYYMMDD>.<ext>`.

### Format selector mockup

```
┌──────────────────────────────────────────┐
│ [ Export ▼ ]                             │
│   ├ Markdown                              │
│   ├ PDF                                   │
│   └ CSV                                   │
└──────────────────────────────────────────┘
```

## Implementation steps

[...]

Issue #5104, opened 2026-06-24
Wire "Promote winner" -> prompt versioning


Aryansharma28:
# Wire "Promote winner" → prompt versioning

> Parent epic: **#5099**
> Depends on: **#5100**. Compatible with N-way mode (winner button shows for the single overall winner).

## Why

After running a pairwise eval and seeing "Variant B wins 12, A wins 7", the natural next action is to **promote B's prompt to production**.

In the MVP, the "Promote B" button copies the winning column's config to the clipboard — useful but manual. This issue wires it into LangWatch's existing prompt-versioning system so a single click tags the winning column's prompt as the new prod version, with an audit trail back to the pairwise eval that produced the decision.

## Persona

DS finishing a successful prompt iteration. They've validated B is better than A on the golden set; they want to ship it without leaving the experiment view or opening a separate prompts admin.

### IN

- "Promote A" / "Promote B" button on the aggregate header bar (rendered only when there's a clear winner — non-tie, threshold default ≥60% win rate).
- Pre-promote **confirmation dialog** showing:
  - Prompt diff (current prod vs proposed winner).
  - Pairwise verdict summary ("Won 12/21 rows on dataset 'X'").
  - Confirm button.
- Hook into existing prompt-versioning route — extend `langwatch/src/server/api/routers/prompts.ts` `promote` action with a `source: { kind: 'pairwise-eval', evalId, experimentId }` metadata field.
- **Audit trail**: the promoted prompt-version row stores the evalId; clicking on a prompt version shows "Promoted via pairwise eval X on date Y".
- Authorization: reuse existing prompt-edit permissions.

### OUT

- **Auto-rollback** if the promoted prompt underperforms on subsequent monitor traces — separate feature.
- **Multi-prompt orchestration** (promoting a whole prompt graph) — out of scope.
- **Promote to non-prod tags** (staging, canary) — could be follow-up; MVP is prod-only.

## Architecture

```
[Promote A]  ─── click ───▶  Confirmation dialog
                                │
                                ├── Renders prompt diff
                                ├── Shows verdict summary
                                ├── "Cancel" / "Confirm promotion"
                                │
                                ▼
                       POST /api/prompts/{id}/promote
                       body: {
                         source: 'pairwise-eval',
                         evalId: '...',
                         experimentId: '...'
                       }
                                │
                                ▼
                       Update prompt version → prod
                       (existing prompts router logic)
                                │
                                ▼
                       UI toast: "Variant A promoted to prod"
[...]

PR #6110, opened 2026-07-24
feat(experiments-v3): add a Bradley-Terry leaderboard chart for the Comparison evaluator (#5103)

### LINKED ISSUE #5103: Bradley-Terry leaderboard aggregation
# Bradley-Terry leaderboard aggregation

> Parent epic: **#5099**
> Depends on: **#5102** OR **#5101** — meaningful only with 3+ variants comparing.

## Why

Simple win-rate aggregation ("A wins 12, B wins 7") works fine for a 2-variant A/B, but breaks down for **leaderboards** (5+ prompt variants, hundreds of rows):
1. **Transitivity isn't enforced** — A beat B 8-3, B beat C 7-2, but C beat A 6-5 in a smaller sample. Naive win-rate ranks them in a cycle.
2. **Strength-weighted wins aren't reflected** — beating a strong opponent should count for more.
3. **No confidence intervals** — how confident are we in this ranking?

**Bradley-Terry MLE** (used by LMSYS Chatbot Arena since Dec 2023, replacing incremental Elo) solves all three:
- Centralizes ratings into a single transitive score per variant.
- Weights wins against strong opponents more.
- Produces confidence intervals.

This issue adds BT aggregation behind a feature flag for leaderboard use cases.

## Persona

Power-user DS running a tournament — 5+ prompt or model variants on a regression suite, needs to answer "which one is best **overall**" with statistical confidence.

### IN

- New aggregation component (`PairwiseLeaderboard.tsx`) rendering:
  - **BT score per variant** (with 95% CI).
  - **Pairwise win matrix heatmap** (NxN cells, color-coded by win rate).
  - **Total comparisons + sample size per pair**.
- Feature flag: `release_ui_pairwise_bt_aggregation` (off by default).
- Server-side BT MLE implementation in Python (small library or direct implementation).
- Surface in cross-experiment context (where leaderboards make sense — comparing 5 experiment runs).
- Warning banner when sample size is too small (N < ~30 pairs per variant).

### OUT

- **Incremental Elo** — BT is the better default per literature; no need for both.
- **Multi-criteria BT** (separate scores for quality vs cost vs latency) — separate issue if requested.
- **Time-decayed BT** — judges change over time, but for now treat all verdicts equal weight.

## Architecture

```
Per-row verdicts:  [{A: 1, B: 0, C: 0}, {A: 0, B: 1, C: 0}, ...]
                                │
                                ▼
                  Build pairwise win matrix (N × N):
                            A    B    C
                      A  [  -    8   12 ]    ← row=winner
                      B  [  3    -    5 ]
                      C  [  7    2    - ]
                                │
                                ▼
                  Bradley-Terry MLE (iterative fitting)
                                │
                                ▼
                  Scores + 95% CI bootstrap:
                       A: 1.42 ± 0.18  ←── winner
                       B: 0.31 ± 0.22
                       C: 0.97 ± 0.20
                                │
                                ▼
                       Sorted leaderboard
```

### Implementation

- **Python**: `choix` library (~20kb pure Python, MLE + bootstrap CIs) OR direct implementation (~30 lines).
- **Endpoint**: extend `/api/experiments/aggregate` (or similar existing aggregation endpoint — verify it exists; if not, extend `/api/experiments/execute` results aggregation) to return BT scores when `aggregation: "bt"` requested.
- **UI**:
  - `PairwiseLeaderboard.tsx` — sortable table; BT column sortable by score; expand row for CI + pairwise breakdown.
  - Win-matrix heatmap (small react-grid component, ~30 lines).
  - Surface as a tab or panel within the experiment view (not the main table; this is power-user territory).

### UI mockup — leaderboard

```
┌──────────────────────────────────────────────────────┐
│  Leaderboard (Bradley-Terry, 95% CI)  ⚠ N=24 pairs   │
├──────────────────────────────────────────────────────┤
│ Rank │ Variant         │ BT score      │ Win rate    │
│  1   │ gpt-4o-mini v4  │ 1.42 ± 0.18  │ 58%         │
│  2   │ claude-3.5      │ 0.97 ± 0.20  │ 47%         │
│  3   │ gpt-4o-mini v3  │ 0.31 ± 0.22  │ 35%         │
└──────────────────────────────────────────────────────┘
Win matrix
       v4    c3.5  v3
v4   [  -     12    8 ]
c3.5 [  3     -     7 ]
v3   [  5     2     - ]
```

## Implementation steps

1. **Python BT impl** — choose `choix` or roll our own; expose `compute_bt(win_matrix) -> dict[str, BTScore]`.
2. **Aggregation endpoint** — accept `aggregation: "bt"` param; query verdicts; build win matrix; call BT; return.
3. **`PairwiseLeaderboard.tsx`** — sortable table + win-matrix heatmap.
4. **Feature flag wiring** — `release_ui_pairwise_bt_aggregation`.
5. **Sample-size warning** — render banner when any variant has <30 comparisons.
6. **BDD spec**: `specs/experiments/pairwise-bt-leaderboard.feature`.

## Acceptance criteria

- [ ] BT scores computed correctly (verify against a fixture dataset with known MLE solution).
- [ ] Leaderboard renders when 3+ variants compared.
- [ ] Behind feature flag, off by default.
- [ ] Win-matrix heatmap displays correctly.
- [ ] Sample-size warning visible at low N.
- [ ] Existing win-rate aggregation unchanged when flag is off.

## Risks / open questions

- **Sample size**: BT needs decent N to be meaningful. Show warning when any variant has <30 comparisons.
- **Library dependency**: confirm `choix` is pure-Python, no heavy native deps; otherwise implement directly (~30 lines for MLE without CIs, ~80 with bootstrap).
- **Identifiability**: if a variant has all wins or all losses against everyone, BT MLE is undefined. Handle gracefully (cap score, warn user).
- **Tie handling**: BT classically takes only wins/losses. Treat ties as 0.5 win + 0.5 loss (standard); document this in the UI.
- **Aggregation endpoint location**: confirm whether to extend `/api/experiments/execute` results or add a separate aggregation route (per project memory: no new endpoints — extend existing).

Aryansharma28:
# Related Issue

- Resolves #5103

## Related issue

Resolves #5103. Parent epic #5099 (closed — foundational work shipped via #5528).

## Summary

Adds a Bradley-Terry MLE ranking chart to the results page for Comparison evaluators with 3+ variants — the statistically-correct replacement for naive win-rate tallying (same reasoning LMSYS Chatbot Arena used moving off incremental Elo).

- **Compact card** next to the existing win-rate chart in `ComparisonCharts.tsx`, only mounted at 3+ variants (below that, win-rate already tells the whole story).
- **Expand into a drawer** with the full leaderboard (BT score ± 95% CI, win rate, N), an NxN win-matrix heatmap, and a **click-through**: click a matchup cell to see the judge's actual reasoning for every row where that pair faced off.
- **Cost/duration tradeoff as a Pareto scatter**, not a blended "best overall" score — there's no honest exchange rate between BT strength points and dollars/milliseconds, so a scatter answers "is the cheaper one meaningfully worse?" by inspection instead.
- **Behind a rollout flag, off by default:** `release_ui_comparison_leaderboard_enabled`. Gating is platform-side (the Postgres flag store via `/ops/feature-flags`), not PostHog — an operator row wins over PostHog for both `SYSTEM` and `PRODUCT` scope, so restricting to one organization needs no PostHog flag and no allowlist in code. The gate covers the chart, the metric entry that toggles it, **and** the drawer: the drawer is addressable by URL, so a link shared out of an enabled org would otherwise render the whole leaderboard for one that has not been given it. Both states are tested.

## Scope — two unrelated fix clusters ride along

Two changes in this branch have nothing to do with the leaderboard. They were briefly split into their own PRs (#6403, #6404) and then consolidated back here, so they ship with this PR:

- **`fix(api-keys)`** (`8e78d74`, `78d47d1` + a 277-line test) — the permission ceiling could not see legacy `TeamUser` membership, so a user whose access came from legacy membership passed every authorization check and was then refused those same permissions as "exceeds your own access". Langy mints a per-turn key mirroring the caller's permissions, so on any workspace still on legacy membership every turn died with an opaque `api_key_scope_violation`.
- **`fix(cli)`** (`b5b1bc1`, `8c73f23`) — the experiment-results reader only understood evaluators emitting a numeric score, so Comparison verdicts were dropped entirely and `--filter failed` could not see a failed comparison.

**Reviewer note:** the api-key pair changes *who may mint an API key*. It is worth reading `langwatch/src/server/api-key/api-key.service.ts` on its own terms rather than as part of the UI review — the guards there (no-stacking on migrated users, org-scope exclusion, the ADR-021 `bindingScopeCanGrant` check) are the load-bearing parts.

## Provenance

The compute engine (`computeBTLeaderboard.ts` — MM-algorithm MLE, bootstrap CI, degenerate-variant smoothing) is ported near-verbatim from a stale draft (PR #5118), which had gone unmergeable after #5528 unified the old 2-way "Pairwise Compare" and N-way "select_best" evaluators into one "Comparison" evaluator. #5118 also never actually wired the compute engine to real data — that adapter, the UI mounting, and the drawer are new.

## Bugs found and fixed along the way

- **Judge response duration was silently dropped.** `mapEvaluatorResult` computed it but the `evaluator_result` event type had no slot for it, and the orchestrator's ClickHouse dispatch never read it either — so "Avg Latency" on every Comparison column has been undercounting judge think-time. Fixed at both points (`resultMapper.ts`, `orchestrator.ts`, `execution/types.ts`).
- **Tie vs. unresolved-label were indistinguishable** in `BatchComparisonVerdict` — both collapsed to `winnerId: null`, which is fine for a bar chart but wrong for Bradley-Terry (a real tie is 0.5/0.5 evidence; an unresolved label is no evidence and must be excluded). Added `isUnresolved` to disambiguate.
- **Per-row candidate set wasn't retained** — only a column-wide union was, which would have fabricated matchups for rows where `select_best_compare` dropped a candidate with no output that row. Added `candidateIds` per verdict.
- **(Found during live QA) The evaluator's candidate-set was never persisted at all.** `readCandidateIds` (used to build `variants` for the leaderboard) reads `evaluation.inputs.candidates` — a field the ClickHouse projection, event schema, and read-side mapper all already supported end-to-end, but `orchestrator.ts` never actually populated when dispatching `recordEvaluatorResult`. Every comparison column's variant list silently collapsed to 1 variant (whichever label won a row), so the leaderboard never mounted past a 1-variant comparison. Fixed by threading the evaluator's own request payload through `mapNlpEvent` → `mapEvaluatorResult` → the event → ClickHouse.
- **(Found during live QA) Drawer crash on expand.** `ComparisonLeaderboardChart`'s expand handler called `setComplexProps({ column, rows, targetColors })` *before* `openDrawer(...)` — but `openDrawer`'s own `updateDrawerUrl` recomputes `complexProps` from whatever props it was given and unconditionally overwrites the module-level store. Since `openDrawer` was only called with `{ evaluatorId }` (a URL-serializable string), it clobbered the prior `setComplexProps` call with `{}`, crashing the drawer with `Cannot read properties of undefined (reading 'variants')`. Fixed by passing `column`/`rows`/`targetColors` straight through `openDrawer`'s own props (which correctly routes non-serializable values to `complexProps` internally).

## Non-determinism hardening (`select_best_compare.py`)

- Configurable judge `temperature` (gpt-5-aware: forces 1.0 for gpt-5-family models per the existing `llm_answer_match.py` precedent, `drop_params=True` so other providers don't hard-fail on an unsupported param).
- One-line anti-position-bias instruction in the judge system prompt.
- **Default-on swap-and-reconcile**: each row is judged twice (shuffled order, then fully reversed), and a disagreement between the two is scored a tie rather than trusting either call arbitrarily. This is the single most literature-backed fix for LLM-judge position sensitivity (order alone flips 10-30% of verdicts on contested rows per multiple sources) — shipped default-on rather than deferred, since judge-call cost wasn't the constraint here.
- **Verified live**: in the QA run below, 2 of 3 rows genuinely disagreed on the original vs. order-swapped call and were correctly downgraded to a tie with both reasonings recorded; the third row agreed both ways and was recorded as a confirmed win. This is exactly the intended failure mode being caught in practice, not just in unit tests.

## Live QA (this session)

Ran a real 3-variant experiment (`support-assistant-warm/-blunt/-formal`) through the Comparison evaluator end-to-end with the feature flag force-enabled, against **Anthropic (`claude-sonnet-5`)** rather than OpenAI — the dev environment's OpenAI key was quota-exhausted mid-session; Anthropic/Gemini keys were both live and real, so Anthropic was substituted for this QA pass. Confirmed:

- Comparison evaluator runs successfully end-to-end (nlpgo → `select_best_compare` → LangWatch), including the swap-and-reconcile double-judge-call path.
- The results page's Metrics dropdown goes from 3 → 4 available charts once a 3-variant Comparison run exists, with the compact "Comparison — leaderboard" card correctly rendering ranked bars.
- Drawer: full ranked table (rank, variant, BT score ± 95% CI, win rate, N, `(degenerate)` labels), the two warning banners (low sample size; MLE-undefined variants smoothed), the win-matrix heatmap, click-through reasoning (verified on a real matchup cell), and the Pareto scatter with a working Cost/Duration toggle (axis and points update correctly).

Screenshots (via [img402.dev](https://img402.dev)):
- [Results page — leaderboard chart mounted](https://i.img402.dev/3l3wxzcxzq.png)
- [Drawer — full table, win matrix, Pareto scatter](https://i.img402.dev/w7hgqeyzir.png)
- [Win-matrix click-through reasoning drill-down](https://i.img402.dev/uh3q78uxjy.png)
- [Pareto scatter — Duration toggle](https://i.img402.dev/8653rf56bw.png)

### Separately observed, not fixed here (pre-existing, out of scope)

Re-running a single cell/column on an **already-persisted** run (rather than starting a fresh run) is a no-op for ClickHouse persistence — the LLM call fires for real (burning real judge-call cost) but `recordEvaluatorResult`'s idempotency key (`tenantId:runId:evaluator:evaluatorId:index`) has no attempt/timestamp component, so the event-sourcing command bus silently dedupes the retry against the row's original processing. The live SSE stream in the workbench briefly shows the fresh result correctly; only a **new** run (fresh `runId`) actually persists. This predates this PR and affects all evaluator cell re-runs, not just Comparison — worth its own issue if it isn't already tracked.

## Test plan

- [x] `computeBTLeaderboard.ts` ported test suite passes verbatim (12 tests)
- [x] New adapter/data-model tests: tie-vs-unresolved disambiguation, per-row candidate subsets, variant cost/duration aggregation (21 tests)
- [x] `resultMapper.ts` duration + inputs-threading fixes: regression tests added, full suite green (71 tests)
- [x] `select_best_compare.py`: 43 Python tests green (14 new — temperature, swap-and-reconcile agreement/disagreement, anti-bias prompt)
- [x] Full repo typecheck: zero new errors vs. the pre-existing baseline
- [x] **Live QA against real LLM traffic (Anthropic, substituted for OpenAI due to dev-env quota) — done.** Verified visually: compact chart, drawer, win-matrix heatmap + click-through, Pareto scatter (Cost/Duration toggle), sample-size/degenerate warnings, swap-and-reconcile catching real order-sensitivity.

## Out of scope (tracked separately)

Cross-experiment comparison (#5102), promote-winner-to-prompt-version (#5104), and PDF/CSV export (#5105) — adjacent sub-issues of the same closed epic, not touched here.

## Summary by CodeRabbit

* **New Features**
  * Added a feature-flagged Bradley–Terry comparison leaderboard (compact chart + expanded drawer) for evaluations with 3+ variants, including trust/warning panels, confidence intervals, win-matrix heatmap, cost/duration view, and a per-matchup reasoning inspector.
* **Bug Fixes**
  * Improved “best choice” judging with optional swap-and-reconcile: judges run on original vs reversed candidate order, and disagreements are treated as ties; updated judge prompting and GPT-5-family temperature handling.
  * Evaluator results now persist request inputs and compute/display duration.
* **Tests**
  * Expanded leaderboard, UI, and reconciliation coverage.

---

# Update — the drawer now leads with the decision, not the table

The leaderboard opened on a sortable table of Bradley-Terry scores. That asks the reader to do the statistics themselves: compare `1.42 ± 0.18` against `1.31 ± 0.22`, notice the intervals overlap, and conclude the run does **not** separate them. Almost nobody does that — they read the top row as the winner, and frequently it isn't one.

Restructured into three questions in the order a PM or engineer actually has them. Someone who reads only step 1 should still leave with a defensible answer.

## 1 · Which one should you ship?

The conclusion is computed and stated. `computeLeaderboardVerdict` compares confidence intervals and returns a clear winner, a tie at the top, or not-enough-signal. A missing interval counts as "cannot separate" rather than "different", so thin data never manufactures a winner.

**A tie is not a failure to conclude.** When two variants are indistinguishable on quality and one costs materially less, the ranking is not the decision — the price is:

step 1 and 2

That headline — *"Ship support-assistant-blunt — same quality, 84% cheaper"* — is the most actionable sentence this drawer can produce, and it was not derivable from anything previously on screen. The old table ranked `warm` first, so a reader would have shipped it at **6.8× the cost** for a difference this run cannot actually detect.

Scores are drawn as bars with their intervals as whiskers, because overlapping whiskers make a tie visible in a way two `±` figures never are.

## 2 · Can you trust that?

The three warning banners collapsed into one trust step that states each condition **either way**. Silence is a bad way to say "this is fine" — the reader cannot tell a clean run from a check that was never made. The step border and badge turn amber only when something is genuinely wrong (above: the sample is too thin).

## 3 · What is it costing you, and why did it win?

Pareto scatter, the full table, and the win matrix — the detail that used to lead now sits here, answering *why* for people who want it instead of *what* for everyone.

step 3

## The verdict rides on the compact card too

So the answer is on the results page without opening anything:

compact card

## Two bugs live QA caught

Both only appear with few comparisons, which is exactly when people look at this chart:

- **The score bars rendered blank.** A bootstrap over a handful of comparisons can return a non-finite bound; one `Infinity` in the scale turned every computed position into `NaN`, which the browser silently drops. Nothing warned that the chart had failed. Non-finite intervals are now treated as absent instead of poisoning the scale, and the scale keys off the scores so one very wide interval can't squash every marker into a corner.
- **The table printed `46.96 ± Infinity`.** That tells the reader nothing and reads as a bug rather than a statement about the sample. The score alone is shown; step 2 already explains the sample is too small to bound.

## Verification

`computeLeaderboardVerdict` has 13 unit tests covering clear-winner, overlapping intervals, intervals touching at a point, missing intervals, degenerate variants, and the cheaper-alternative rules (biggest saving wins, marginal differences stay quiet, no quality-for-price trade when there *is* a clear winner). 219 tests pass across `batch-evaluation-results/`; typecheck clean.

**Honest note on the QA data:** the local run had only two comparison candidates, so to exercise the 3+-variant leaderboard I seeded the comparison verdicts in ClickHouse to spread wins across all three variants. The **cost figures are real** (from the actual run), as is everything else on screen — but the win distribution is seeded, so treat "84% cheaper" as a demonstration of the mechanism rather than a finding about these prompts.

Screenshots hosted via img402.dev; the red step boxes are annotations for this description, not product UI.

---

## Screenshots refreshed after the confidence-interval fix

The earlier images were taken before `977f2b1c` and no longer showed what the code does — worth calling out, because the difference is the whole point of that commit.

**Before**, on this same run, the table read:

| variant | interval shown |
|---|---|
| support-assistant-warm | `96.93 ± 30071.90` |
| support-assistant-blunt | `46.96` — interval suppressed, it was `± Infinity` |
| support-assistant-formal | `-143.89` — same |

**Now:**

| variant | interval shown |
|---|---|
| support-assistant-warm | `96.93 ± 149.19` |
| support-assistant-blunt | `46.96 ± 192.99` |
| support-assistant-formal | `-143.89 ± 150.71` |

Same point estimates, real intervals. In step 1 all three bands are now drawn and *visibly overlap*, which is what makes "tied for first" self-evident — previously only one bar had a band at all, because the other two had diverged.

The verdict is unchanged (*"Ship support-assistant-blunt — same quality, 84% cheaper"*), but it is now reached because the intervals genuinely overlap rather than because two of them had blown up to infinity. Same answer, honest derivation.

### Dark mode

Six hardcoded `rgb()`/`rgba()` values were replaced with the semantic tokens the sibling charts in this folder already use, so the callout, the confidence bands, and the step badges follow the theme:

dark mode

*Caveat unchanged from above: the win distribution in this run is seeded QA data. The cost figures are real.*

---

# Review round 2 — all three P1s resolved, plus the P2/P3 field

Addressing @sergioestebance's ruthless review. Every P1, every P2 and every P3 is fixed rather than deferred, and a background review of the same diff caught one more P1 that his fix list did not cover. Nine commits, `f00f712` → `5a30c8d`.

## The three P1s

**The Langy fix now actually fixes Langy.** The legacy fallback had been added to the mint-time ceiling only, and the population it targets is *defined* by holding zero `RoleBinding` rows — so `resolveApiKeyPermission` refused every permission and the minted key was a dead credential. Rather than adding a second copy of the fallback at request time, it is now one exported `resolveLegacyCeiling` used by the request path and all three mint-side branches (two of which had never learned about legacy membership either). Consolidating dissolved three of the four P2s in that cluster: the group-binding predicate, the missing scope filter and the EXTERNAL floor were all a single copy disagreeing with `loadScopeResolution`.

**No winner is named across variants that never met.** `comparabilityOf` is wired in — into `areDistinguishable` itself, not its callers, because there are three of them and two had already drifted apart once. A break is not a tie, so it gets its own verdict kind: a tie licenses `findCheaperTiedAlternative` to recommend the cheaper variant, and a break must not. Verified on your two-island fixture: `clear-winner` before, `not-comparable` after, and adequacy's "5 of 6 pairs separated" drops to the two pairs that were real.

**`useBTLeaderboard.ts` is text again**, so its 85 lines appear in this diff for the first time.

## The one you did not have

`computeParetoDominance` is the **third** caller of `areDistinguishable`. Wiring only the verdict and the adequacy count would have left it emitting cross-island dominance edges — surfaced to the reader as *"3 variants are beaten outright and can be dropped: alpha (by charlie)…"*, which is a stronger and more damaging claim than the "ship charlie" headline you caught. Confirmed by executing the chain: 4 cross-island edges before the fix, 0 after.

## Behaviour changes worth calling out before merge

- **`swap_and_reconcile` defaults on for existing Comparison configs.** The field does not exist on `main`, so the Zod default injects `true` into every already-saved config and doubles judge-call cost per row, billed to the customer, with no migration. That was a deliberate choice and it stands — but it lands silently, and it is the thing to disagree with if anyone is going to.
- **A reconciliation failure is now `skipped`, not a tie.** Previously only when `allow_tie=False`; the default path scored it 0.5/0.5 into the Bradley-Terry tally, which is evidence the run never produced. The evaluator's own `allow_tie=False` branch already argued this was unacceptable.
- **`WinRateChart`'s axis labels changed outside the feature flag.** The switch to `buildAxisLabels` strips the common prefix, so an already-shipped chart renders `…-warm` where it rendered `support-assista…`. It is an improvement and it is not gated by `release_ui_comparison_leaderboard_enabled` — flagging it as blast radius, not as a defect.
- **Judge model recorded on the run now honours an unsaved override**, matching what execution actually runs. Previously a user who switched the judge without saving ran on one model and recorded another, and the recorded one feeds the self-preference check.

## Known gap, stated rather than papered over

The expanded drawer no longer throws when opened from a URL, and the rollout gate is reachable on that route for the first time (it sat seventeen lines below the dereference that threw). But a cold link renders a deliberate "open the run and expand the card" state rather than reconstituting itself. `drawers.md` asks for full reconstitution; this does not meet that bar. Everything needed is derivable from the query the results page already runs, but it means carrying the run identifiers in the URL and threading them through two more components — left as a follow-up rather than widened into this PR.

Separately, and inherited rather than introduced: the legacy fallback is re-armed by removing a user's last `RoleBinding`, because a frozen `TeamUser` row then becomes the only thing describing them. `rbac.ts` behaves identically, so the key and the UI agree — which is the property that matters for this fix — but it is a sharp edge in the legacy fallback as a whole.

## Verification

Every regression test executes the real code path and was confirmed **failing without its fix**, not just passing with it:

| Fix | Observed without it |
|---|---|
| Comparability in the verdict | `expected 'clear-winner' to be 'not-comparable'` |
| Comparability in adequacy | `expected 5 to be less than or equal to 2` |
| Comparability in dominance | 4 cross-island edges: `c>a`, `c>b`, `a>d`, `b>d` |
| Request-time legacy ceiling | `expected false to be true` |

1071 unit tests and 29 integration tests green across the touched areas; full-repo typecheck at the pre-existing 104-error baseline (zero new); `pnpm typecheck:tests` clean; Biome clean on every changed file; feature-parity gate green.

The CLI feature file went from **0 of 11 scenarios enforced** to 11 bound, and came off `LEGACY_INERT`. Worth knowing for anyone writing bindings: the parity checker's annotation regex only consumes a closing `*/` on the **same line**, so a multi-line JSDoc ending in `@scenario` parses fine and binds nothing — a silent no-op that reads exactly like a binding. Two of mine were written that way before I noticed.


Aryansharma28:
## How to read this chart (and how not to)

This PR adds a ranking, but the ranking is the least important thing on screen. Everything below is from a **real 60-comparison run** on this branch — 4 prompt variants, judged by `anthropic/claude-sonnet-5`, 300 cells, 0 failures.

| Variant | Model | BT score | Plausible range | Win rate | Cost/row |
|---|---|---|---|---|---|
| support-assistant-warm | gpt-5-mini | **153.5** | 69.9 → 237.2 | 72% | $0.0018 |
| support-assistant-warm-premium | gpt-5.5 | **95.1** | 14.3 → 175.9 | 60% | $0.0066 |
| support-assistant-blunt | gpt-5-nano | −64.1 | −170.4 → 42.2 | 32% | $0.0003 |
| support-assistant-formal | gpt-5-mini | −184.5 | −338.9 → −30.2 | 18% | $0.0023 |

---

### 1. The number is a chess rating, not a score out of anything

`score = 400 × log₁₀(strength)`, centred so **0 = average for this field**. Negative means below this field's average, not "bad".

Only the **gap** is meaningful, and it converts directly to odds:

> **400-point gap ≈ 10:1. 200 ≈ 3:1. 0 = coin flip.**

So "153.5 vs 95.1" is not "60% better" — it is *"if these two met again, warm takes about 6 of 10"*:

| Matchup | Gap | Predicted |
|---|---|---|
| warm vs warm-premium | 58 | warm **58%** |
| warm vs blunt | 218 | warm **78%** |
| warm vs formal | 338 | warm **87%** |

Scores are **only comparable inside one run**. A 153.5 here and a 153.5 in another experiment mean nothing to each other.

### 2. Why not just count wins?

Win rate ignores *who* you beat. Beating `warm` is worth far more than beating `formal`, and with randomised pairings not every variant faces the same opposition. Bradley-Terry solves all strengths simultaneously so opponent quality is priced in — the same method [Chatbot Arena uses](https://arxiv.org/pdf/2403.04132) for exactly this reason.

It also survives non-transitivity: if A beats B, B beats C, C beats A, a win tally is incoherent and BT still returns a usable ranking.

### 3. ⚠️ The part reviewers most need to check

**Overlapping bands mean the run cannot separate two variants. They do NOT mean the variants are equal.**

This is a genuine statistical asymmetry, and the code is deliberately on the conservative side of it:

- **Non-overlapping → really different.** Safe conclusion.
- **Overlapping → inconclusive.** Could still be a real difference we lack the data to see. ([why](https://cscu.cornell.edu/wp-content/uploads/ci.pdf), [and](https://www.econometrics.blog/post/overlapping-confidence-intervals/))

So `computeLeaderboardVerdict` treating overlap as "tied" is **under-powered on purpose** — it will never invent a winner, but it *will* sometimes miss a real one. That trade is intentional: a false "ship X" costs more than a "run more rows".

A consequence worth knowing, visible in the table above: **blunt's band (up to 42.2) overlaps warm-premium's (from 14.3)** — yet only warm/warm-premium are badged "tied for first". Not a bug. "Tied for first" means *tied with the leader*, and blunt **is** cleanly separated from warm (42.2 < 69.9). "Cannot distinguish" is not transitive.

### 4. How many rows do you actually need?

Measured empirically by re-running the real solver over prefixes of the same 60 comparisons:

```
n=10   4/4 tied   warm-premium ranked #1   ← WRONG LEADER
n=15   3/4 tied
n=20   2/4 tied   ← first useful shape
n=25   3/4 tied   ← regresses
n=30   3/4 tied   ← regresses
n=35   2/4 tied   ← stable from here
n=60   2/4 tied   warm=147[68,230] warm-premium=133[52,232]
```

Two things fall out:

1. **At n=10 the ranking names the wrong leader** — and the CI gate refused to call it. That is the feature preventing a wrong shipping decision, on real data.
2. **~35 comparisons is the honest minimum.** 20 is the earliest useful answer but is not stable. More rows mainly buy tighter intervals.

The drawer reports this as **"3 of 6 variant pairs were separated"** — an observation, never a forecast. A "run 20 more rows" figure would be a power calculation over an effect size estimated from the same thin data, i.e. a promise the run cannot keep.

### 5. Judge biases the ranking cannot show

Three documented failure modes of LLM-as-judge comparison ([survey](https://arxiv.org/html/2410.02736v1), [position bias](https://openreview.net/forum?id=y3jJmrKWQ4)):

| Bias | Status |
|---|---|
| **Position** — favours whichever answer is shown first | ✅ already mitigated: candidate order is shuffled per row |
| **Verbosity** — favours longer answers regardless of quality | ⚠️ now **reported**: *"the leading variant's answers averaged 1.6× the length of the rest"* |
| **Self-preference** — a judge rates its own model family higher | ⚠️ now **reported**: *"Judged by anthropic/claude-sonnet-5, which shares a model family with none of the candidates"* |

The last one is read off the **run**, not the evaluator's current config — the judging model is now recorded at execution time, so editing an evaluator later cannot retroactively change what an old run says about itself.

Verbosity is deliberately rendered as a **measurement, not a warning**. For plenty of tasks the longer answer genuinely is better, and a check that fires on correct results teaches people to skip the panel.

### 6. The headline is computed, not generated

The load-bearing sentence — *"Ship support-assistant-warm — top of the ranking and 73% cheaper"* — is a pure function of the scores and costs (`formatLeaderboardHeadline`), covered by tests. No model writes it.

That is on purpose. This feature's whole value is refusing to name a winner the data cannot support; a summariser that rounds a tie up to a winner would undo that in one line, in fluent English, which is [harder to catch than a wrong chart](https://www.thoughtspot.com/data-trends/artificial-intelligence/ai-generated-insights). The optional **"Ask Langy to explain this"** button hands the *already-computed* figures to the assistant and asks it to explain them — it cannot re-rank, and the computed verdict stays on screen above it.

---

### What a reviewer should sanity-check

1. **Read the bands, not the numbers.** If two bands overlap, the run did not separate them — whatever the bars look like.
2. **Check "how much this run settled"** before trusting any ordering. `0 of 6 pairs separated` means the run bought nothing.
3. **Check judge independence** if a candidate and the judge share a vendor.
4. **Treat "tied" as "not proven different", not "proven the same"** — see §3.

Worth noting the run above is the ideal case for this feature: the two top variants are the **same prompt on different models**, statistically tied, with the cheaper one **3.7× less** per row. Quality was not the deciding factor; price was. That is the outcome a plain win-rate chart would have hidden.

Aryansharma28:
## What it actually looks like

Rebased onto `main` (0 behind), and these are from the same real 60-comparison run described in the comment above — `?runId=bold-silver-robin`, 4 variants, 300 cells, 0 failures.

### 1. The compact card, in the metrics row



The green line is the whole point. It is **computed, not generated** — `formatLeaderboardHeadline`, covered by tests — and it is the *same* sentence the drawer shows.

Note also that `-184.51` no longer collides with its own axis label. recharts anchors a negative bar's label to its **outer** end, which grows toward the category labels; the longest negative bar therefore overlapped its own name and rendered as unreadable mush (`…f184.51`). Labels are now hand-placed just past the zero line, so they always grow into the empty middle.

### 2. Bar labels now match across every chart in the row



Before, the same four prompts rendered **two different ways** in one row: charts that disambiguated showed `support-assista… (1) (2) (3) (4)`, and charts that only truncated showed `support-assista…` four times over — indistinguishable.

Both were symptoms of the same cause: these names share an 18-character prefix, and left-truncation throws away the only part that differs. `buildAxisLabels` drops the shared prefix instead, so every chart shows `…warm / …warm-premium / …blunt / …formal`. No truncation, no `(N)` suffixes.

### 3. The drawer — the whole argument in one view



Worth reading closely, because this single screenshot is the case for the feature:

- **Step 1** — the verdict, then the evidence for it. The top two bars' shaded bands visibly overlap, which is why both carry `tied for first` and why the headline pivots to **cost** (`$0.0018 vs $0.0066`) rather than claiming a quality winner. The two tied variants are the *same prompt on different models*.
- **Step 2** — six checks, all reported either way, so silence means "checked and fine" rather than "never looked":
  - `How much this run settled — 3 of 6 variant pairs were separated` — an observation, never a forecast.
  - `Answer length — 1.6× the rest of the field` — verbosity bias, stated as a measurement in neutral type, never as a warning.
  - `Judge independence — judged by anthropic/claude-sonnet-5, which shares a model family with none of the candidates` — self-preference, read off the **run** rather than the evaluator's current config.
- **Ask Langy to explain this** — opt-in, gated on `langy:view` + `langy:create`, and it receives the already-computed figures. It cannot re-rank, and the computed verdict stays on screen above whatever it says.

### 4. Every section explains itself



`153.53` is not a percentage, a count, or a rating out of anything, and nothing on screen said so. Each step now carries an info icon — keyboard-reachable, not hover-only — leading with the two genuinely counter-intuitive facts:

> *"…it is the GAP that means something, not the number. A 400-point gap is roughly 10:1 odds; 0 is a coin flip. … when two bands overlap, this run cannot tell those two apart, whatever their bars look like."*

---

### Verification

| | |
|---|---|
| Rebase | onto `main`, 16 commits, **net diff byte-identical** to pre-rebase |
| `pnpm typecheck` + `typecheck:tests` | clean |
| Tests | **670 passed**, 19 skipped, 0 failed |
| Data | real run — real LLM calls, nothing seeded |

sergioestebance:
## Ruthless review — BLOCKED

Reviewed the leaderboard, the api-key rider, the server persistence changes, the Python evaluator, and the CLI. Every candidate below was run through attribution (does *this diff* cause it?) and an adversarial refutation pass that executed the real code path — verdicts and executed output are carried inline. Withdrawn candidates are listed at the end so you can see what didn't survive.

**Three P1s block merge.** A large field of P2s rides behind them — several are genuinely good catches this branch made and then half-landed.

---

### P1 — The Langy fix doesn't fix Langy: the failure just moves mid-turn — CONFIRMED, executed

`langwatch/src/server/rbac/role-binding-resolver.ts:378-385` (the gap) · `api-key.service.ts:633+` (where the fix landed)

The rider adds `resolveLegacyCeilingRole` to the **mint-time** ceiling only. At **request time**, `resolveApiKeyPermission` re-checks the owning user's bindings through `checkRoleBindingPermission`, which has no legacy `TeamUser` fallback (deliberately pinned by `role-binding-resolver.integration.test.ts:623`). The target population is *defined by* having zero `RoleBinding` rows, so that step returns false for every permission.

Driven end-to-end against real Postgres — `mintLangySessionApiKey` → `TokenResolver.resolve` → `enforceApiKeyCeiling`, for a user with org membership + legacy `TeamUser` ADMIN + zero bindings:

```
[probe] MINT SUCCEEDED. key.userId = Xi2...
[probe] resolveApiKeyPermission(traces:view)  = false
[probe] enforceApiKeyCeiling(traces:view) => THREW api_key_permission_denied
WARN name=langwatch:api:api-key-ceiling permission=traces:view msg="API key ceiling check failed"
```

Control: the same fixture with a real `RoleBinding` passes 5/5 (`langySessionKey.integration.test.ts`). Every Langy REST route funnels through `enforceApiKeyCeiling` (`prompts/app.v1.ts:68`, `traces/app.v1.ts:92`, `agents/app.ts:92`), so the minted key is a dead credential for this population. The failure moves from `api_key_scope_violation` at mint to `api_key_permission_denied` on the first tool call — arguably worse, because the turn now starts and streams, then every tool silently 403s.

**Fix:** add the same `resolveLegacyCeilingRole` shape to step 3 of `resolveApiKeyPermission`, resolved once per request. Safety is asymmetric — the fallback can only restore access `batchProjectPermissions` already granted (it's what minted the key's permission set). Alternatively, narrow the PR: state in the description that legacy-membership users are still blocked at request time, and this only stops the mint from 409ing.

---

### P1 — "Clear winner" across variants that never met — CONFIRMED, executed via the real UI path

`computeLeaderboardVerdict.ts:38-103` · `scoreSeparation.ts:70-88` · `computeComparability.ts:78-95`

`comparabilityOf()` — the function written to answer "may these two even be compared?" — has **zero production callers** (`tslsp-cli references`: only its own unit test). Every separation decision routes through `areDistinguishable`, which looks only at the difference CI. On a disconnected comparison graph (two islands of variants that never share a row — reachable whenever two variants error on the rows the other two answered, per `types.ts:135-141`), the between-island score gap is a pure gauge artifact of `normalizeToGeometricMean`, and every bootstrap replicate applies the same gauge, so the CI comes out *tight* — confidently wrong.

Driven through the real chain (`buildPairwiseComparisons` → `computeBTLeaderboard` → `computeLeaderboardVerdict` → `formatLeaderboardHeadline` → `buildTrustChecks`), 100 rows, islands `{a,b}`@26:24 and `{c,d}`@48:2, a and c never co-occurring:

```
verdict: clear-winner, leaderId "c"
headline: "Ship charlie — The run separates it from every other variant: the gap
           to each of them is bigger than the uncertainty in that gap..."
CI(c−a) = [+65.21, +338.79]      comparability.identifiable = false
```

`a` and `c` never met, `dominates` is false in both directions — the code has already computed that not even the *direction* is known, and prints "separates it from every other variant" anyway. Holding `{a,b}` fixed and varying only the *other* island's internal lopsidedness walks this headline continuously from "Ship c" to "tie" — data carrying zero information about c-vs-a moves the c-vs-a verdict, which is the definition of a gauge artifact.

Two things make it worse than "the panel warns, so it's honest":
- **The compact card carries no warning.** `LeaderboardTrustPanel` is mounted only in the drawer; the results page shows a green "Ship charlie" and bars, full stop. Most readers never open the drawer.
- **The drawer's own trust step stays neutral.** `trustHasProblem` (`ComparisonLeaderboardDrawer.tsx:119-124`) ORs `minMatchups/hasDegenerate/didConverge/sharedFamily/separatedPairs` but **omits `identifiable`**, so the step renders with a neutral border despite containing three amber lines. Adequacy even counts "5 of 6 pairs separated" — every cross-island pair counted as separated.

**Fix:** wire `comparabilityOf` into `computeLeaderboardVerdict` (treat `incomparable` as not-separated so the clear-winner clique can't cross a break) and into `computeSampleAdequacy` (they share `areDistinguishable` deliberately — both must consult comparability or they'll drift); add `!identifiable` to `trustHasProblem`; and stop the compact card rendering a positive-tone verdict on a non-identifiable fit. Regression test must execute the path and assert the verdict *kind* on a two-island fixture, not assert on headline prose.

---

### P1 — `useBTLeaderboard.ts` was committed as a binary file — shipped unreviewable

`langwatch/src/components/batch-evaluation-results/useBTLeaderboard.ts:47,76`

Two raw `0x00` bytes (offsets 2066, 2677) inside `variantIds.join("<NUL>")` — written as literal bytes instead of `" "`. Confirmed locally: `file` reports `data`, and the diff is `Bin 0 -> 3074 bytes`, which GitHub renders as "Binary file not shown". **3KB of caching logic — the file behind the live-run cache P2 below — received zero review, and every future diff on it will be invisible too.** Sibling `useVariantMetrics.ts:44` uses `join(" ")` for the identical job.

Runtime behaviour is fine (a NUL-joined string is still a stable cache key), so this doesn't threaten data — but you cannot knowingly merge an unreviewable binary file into a TS tree, and the fix is two characters.

**Fix:** replace both literal NULs with `" "`.

---

### P2 — worth fixing before merge

**api-key: group-derived bindings don't block the legacy fallback (defense-in-depth hole).** `resolveLegacyCeilingRole` counts `roleBinding.count({ where: { organizationId, userId } })` (`api-key.service.ts:731-733`), but group bindings carry `groupId`, not `userId`. `loadScopeResolution` (`rbac.ts:1290-1330`) ORs in `groupId: { in: groupIds }`; this doesn't. Executed proof: a user with a group-held TEAM VIEWER binding + stale legacy ADMIN row + zero direct bindings **mints `datasets:manage`** (control: add a direct binding and the same mint is refused). The docstring at `:712` claiming "Same gate as `loadScopeResolution`" is false. The request-time intersection (P1 above) makes the key inert, so this is a **persisted over-grant recorded in the key's CustomRole + UI**, not live escalation — but it's the exact stacking the docstring promises is impossible. Fix: use the same `userId OR groupId in groups` predicate, keeping the two-query budget.

**api-key: the count is org-wide with no scope filter — the original bug is still reachable.** Same line. A user with a direct binding on team B plus a legacy ADMIN row on team A gets `count > 0` → no fallback → ceiling refuses, while the tRPC path evaluating team A falls back and allows. That's the `api_key_scope_violation` this rider set out to cure, still live for any partially-migrated user. Fix: scope the count to the scopes in play, same as `checkPermissionFromBindings`.

**api-key: EXTERNAL floor not applied (code asymmetry; escalation refuted).** The new branch ends at `teamRoleHasPermission` and never reads the org role; both tRPC evaluators cap EXTERNAL members at `EXTERNAL_MEMBER_PERMISSIONS` first (`rbac.ts:1401-1403`). Executed: an EXTERNAL member with a legacy ADMIN row mints `secrets:manage`. **But** the escalation framing does not survive — the `migrate_team_users_to_role_bindings` migration backfilled a TEAM binding for *every* TeamUser row, and no live code writes TeamUser rows, so "legacy row + zero bindings" is nearly unconstructible, and the key is inert at request time regardless. Keep it as a parity/defense-in-depth fix (mirror the ADR-021 wording three lines above), **not** as a privilege-escalation fix. `ensureCallerIsOrgMember` already loads the membership row — select `role` and pass it down.

**Only one of three ceiling branches learned about legacy membership.** `assertCustomRoleWithinCeiling` (`:625-639`) and `assertBuiltinRoleWithinCeiling` (`:780-786`) still call `checkRoleBindingPermission` bare, so a legacy-membership user creating a key through the normal builtin-role UI still gets `api_key_scope_violation`. Fail-closed, so not a security issue — but it guarantees a second round-trip on this file.

**Drawer dies on reload / pasted link — CONFIRMED, executed TypeError.** `ComparisonLeaderboardDrawer.tsx:294-297` derefs `column`/`rows` that arrive only via in-memory `complexProps`; after reload `CurrentDrawer` passes only URL params. Rendering with just the URL-serializable prop throws `TypeError: Cannot read properties of undefined (reading 'variants')` (both flag states), caught by the `fallback={null}` ErrorBoundary which silently strips the drawer params. Sharpened: the `if (!showLeaderboard) return null` guard sits at **line 311, seventeen lines after the crashing deref** — so the drawer's flag gate is unreachable on the URL path, and the gate test (`ComparisonLeaderboardDrawer.gate.integration.test.tsx:74`) passes by accident because it hand-feeds `column`/`rows` that a URL can't carry. `evaluatorId` is declared and passed but never read — the refetch hook was built and left dead. This contradicts `drawers.md` ("the URL form must always carry enough state to reconstitute the drawer from a paste") and this drawer's own docstring. Fix: make `column`/`rows` optional, move a `!column` bail (plus the flag bail) above the deref, and either wire a refetch keyed on `evaluatorId` or render a deliberate "reopen from the results page" state — then add the gate-test case that renders with `evaluatorId` alone.

**`swap_and_reconcile=True` silently doubles judge cost for every existing Comparison config.** The field doesn't exist on main; defaults spread underneath DB settings (`evaluations-legacy.ts:992`) plus the Zod `.default(true)` (`evaluators.generated.ts:925`) inject `true` into every already-saved config, cost billed to the customer (`evaluations-legacy.ts:1141`). Comparison is not flag-gated and has been live since #5528. The default-on choice was deliberate and defensible per the description — the issue is that it lands silently on existing configs with no migration or changelog. ("No off switch" is **withdrawn** — `DynamicZodForm` in `EvaluatorEditorShared.tsx:830` does render a working toggle; `ComparisonConfigForm.tsx` is just the wrong surface.) Fix: note the behaviour change for existing users, and fix the copy (see P3). Also: on the default `gpt-5-mini` judge, `_effective_temperature` forces 1.0, so the two calls differ in order *and* sampling — the mechanism can't cleanly attribute a flip to order, which its own disagreement copy admits but its agreement copy ("Confirmed under order swap.") does not.

**Reconciliation-failure tie is indistinguishable from a judge-declared tie.** With `allow_tie=True` (default), a swap disagreement returns `{"winner":"tie"}` with no marker; `types.ts:780` sets `isUnresolved` only from a bad slot, so it reaches Bradley-Terry as real 0.5/0.5 evidence. The evaluator's own `allow_tie=False` branch names this exact outcome as unacceptable, then the default branch does it. Scoped to two-variant runs by accident (`resolveComparison` drops ties for N>2). Fix: return `EvaluationResultSkipped` for a reconciliation failure regardless of `allow_tie` — the finding is "no verdict", not "they tied".

**The whole evaluator request payload is persisted, for every evaluator, not just Comparison.** `orchestrator.ts:1246` threads `buildEvaluatorInputs`'s full output into `inputs`, written at `:2144`. Only `.id` is ever read (`readCandidateIds`); outputs/costs/durations/golden are stored and never consumed. Footprint is rows × targets × evaluators, and it reaches ClickHouse twice (event_log + `experiment_run_items.EvaluationInputs`), the browser (`experiment-run.service.ts:511` is a bare `SELECT *`), and the `_size_bytes` storage meter. Same tenant/retention as existing columns, so amplification not exposure — but on main this column was NULL for orchestrator runs. Fix: project to `{ candidates: candidates.map(c => ({ id: c.id })) }` before threading; pass nothing when there's no `candidates` key.

**Recorded judge model ignores the unsaved local override.** `orchestrator.ts:1722-1731` resolves the model from `loadedEvaluators.get(id).config.settings.model` only, ignoring `localEvaluatorConfig` — which execution *does* honour (`workflowBuilder.ts:316`), and which the prompt branch eight lines up checks first. So a user who switches the judge model without saving runs on X and records Y, which then feeds the self-preference warning in `LeaderboardTrustPanel`. Fix: mirror `t.localEvaluatorConfig?.settings ?? dbConfig?.settings`.

**Live-run caches recompute every second.** `useBTLeaderboard.ts:32` and `useVariantMetrics.ts:29` key their WeakMaps on object identity; `BatchEvaluationResults.tsx:203` polls at 1s and re-memoizes on `[runDataQuery.data]`, which changes identity each poll → full BT fit + variantMetrics recompute synchronously on the render thread, once per second, exactly while a user watches a run. Measured ~18ms/s at 3-4 variants (fine) rising to ~500ms/s at 15. Fix: key on content identity (runId + evaluatorId + row count), or skip recompute while `!isFinished`.

**Generated evaluator files were hand-edited, not regenerated.** `evaluators.generated.ts:2359` still carries the old class docstring (the card copy customers read), missing the swap-and-reconcile behaviour; `swap_and_reconcile`'s description is truncated. Both copies match each other, so no drift — just stale vs the Python. Fix: run the generator.

**CLI feature scenarios bind nothing.** `check-feature-parity.ts`: `cli-experiment-results.feature` = 0 of 11 enforced (10 `@unimplemented`, rest untagged). The four new scenarios are `@integration @unimplemented`, and the 12 passing `results.unit.test.ts` tests carry zero `@scenario` annotations — so they're vacuously bound. A tag-theft regression also left the pre-existing "filters to only failed rows" scenario (line 35) untagged. Fix: retag `@unit`, drop `@unimplemented`, add `@scenario` annotations, restore the stolen tag.

---

### P3 — nits

- **Duplicate candidate id double-counts a matchup.** `computeBTLeaderboard.ts:352-357` doesn't dedupe `candIdxs`; `["a","b","b"]` winner `"a"` credits `W[a][b]=2` from one row (reachable via `targetIdByAnyKey` many-keys→one-target). Fix: `[...new Set(candIdxs)]`.
- **`emptyLeaderboard()` reports `identifiable:true, didConverge:true`** (`computeBTLeaderboard.ts:259-260`) — a zero-variant run would show green "Everything is on one scale ✓". Green because nothing was checked.
- **Non-finite paired-cost interval passes the recommendation gate** (`computeLeaderboardVerdict.ts:213`) — `?.every(isFinite)` short-circuits on `[NaN,NaN]`; the opposite of how `scoreSeparation.ts:47-56` guards the same shape one file over. Defensive-only today.
- **Cost and duration mean CIs share a seed** (`computeVariantMetrics.ts:219-223`), so a variant's x/y error bars on the trade-off glyph come from the identical resample pattern. Salt one.
- **Internal jargon in customer copy:** "BT score" (`ComparisonLeaderboardChart.tsx:239`, `ParetoScatterChart.tsx:359`, `PairwiseLeaderboard.tsx:146`), "Bradley-Terry" in a label (`PairwiseLeaderboard.tsx:84`), "(degenerate)" (`:226`). Per `copywriting.md` → "Score", and "(degenerate)" → "(never won / never lost)". `swap_and_reconcile`'s toggle label is also raw schema vocabulary ("Roughly doubles judge-call cost per row") — move detail to a `(?)` tooltip. The `skipFields` comment at `EvaluatorEditorShared.tsx:836` is stale and misleading (claims it keeps "ONLY the Model picker", lists `prompt` — none of that is true), which is what made a reviewer conclude the toggle was hidden.
- **Two readers of the same field disagree:** `resultMapper.ts:258` guards timestamps by truthiness (`started_at` of 0 → no duration) while the target path uses `!== undefined`. Unreachable with epoch ms.

---

### Withdrawn after refutation

- **api-key privilege escalation (both group + EXTERNAL variants):** the request-time intersection makes every over-granted key inert, and the EXTERNAL state is nearly unconstructible post-migration. Real code defects, but not vulnerabilities — kept as P2 defense-in-depth.
- **"swap_and_reconcile has no off switch":** the toggle renders via `DynamicZodForm`; the reviewer checked the wrong form.

### Verified clean

Duration threading (ms, end-to-end, previously dead `_duration` now consumed); no event-schema change so old events replay to null (`inputs`/`duration` were already `.nullable().optional()` on main); `isUnresolved`/`candidateIds` default sanely on old data (empty candidate list → row dropped → `no-signal`, no fabricated ranking); flag scope/default correct and re-checked at render; the `complexProps` clobber fix is correct; BT MLE matches Hunter (2004) bit-identical, ties give exact 0.5/0.5, `isUnresolved` contributes nothing, the `candidateIds` fabrication guard works, bootstrap seeding deterministic, percentile math R type-7 with no off-by-one; `specs/experiments/comparison-leaderboard.feature` genuinely bound (60 scenarios, 51 enforced, `@scenario` annotations present); no cross-tenant read anywhere touched; no inline `import()` added.

**Blocked until the three P1s are resolved.** Happy to re-review once the Langy request-time path, the comparability wiring, and the binary file are addressed.

<sub>Reviewed with a fan-out of specialist agents; every P1/P2 above was attribution-checked against `main` and adversarially refuted by executing the real code path, not by reading it.</sub>

Aryansharma28:
@sergioestebance — all three P1s, every P2 and every P3 are fixed and pushed (`f00f712` → `5a30c8d`, 9 commits). Nothing deferred. Details are in the PR description under **Review round 2**; the parts that are answers to you rather than restatements:

**The Langy P1 was the right call and the fix went further than the fix.** You were right that the failure just moved. Adding the fallback to `resolveApiKeyPermission` step 3 alone would have made a *second* copy of it, so it is now one exported `resolveLegacyCeiling` used by the request path and by all three mint-side branches. That dissolved three of your four P2s as a side effect: the group predicate, the missing scope filter and the EXTERNAL floor were one copy disagreeing with `loadScopeResolution`, and there is now nothing to disagree.

Your docstring catch at `:712` was the tell — "same gate as `loadScopeResolution`" was false in three separate ways at once.

**On the comparability P1 — a break is not a tie.** Wiring `comparabilityOf` so the clear-winner clique cannot cross a break turns the verdict into `tie-at-top`, and that swaps one wrong claim for another: `findCheaperTiedAlternative` fires on `tie-at-top` and would have offered the cheaper of two variants that never met as an equal-quality swap. A tie says the run looked and found no difference, which is what licenses deciding on cost; a break says it never looked. So it gets its own verdict kind. Same reasoning as your own point about the compact card — the failure had to stop being *derivable*, not just stop being displayed.

I put the veto inside `areDistinguishable` rather than at its callers, for the reason that module's docstring already gives.

**One you did not have, and it is worse than the one you found.** There are **three** production callers of `areDistinguishable`, not two — `computeParetoDominance` is the third. Wiring only the verdict and the adequacy count would have left it emitting cross-island dominance, surfaced as *"3 variants are beaten outright and can be dropped: alpha (by charlie)…"*. Telling someone to delete a variant is a stronger claim than telling them which to ship. Executed: 4 cross-island edges before, 0 after.

**`trustHasProblem` had a second hole.** You caught the missing `identifiable`; it was also blind to `bootstrapNonConvergence`. Since the parallel boolean had now drifted twice, it derives from `buildTrustChecks(...).some(c => c.tone === "warn")` instead — a check added later cannot be forgotten there again.

**Two of yours I have implemented differently than suggested, and one I have not fully closed:**

- The **reconciliation tie** is now `EvaluationResultSkipped` regardless of `allow_tie`, as you asked. Worth recording why that was the right hinge: `allow_tie` governs whether the *judge* may call two candidates equivalent, which is a different question from whether an absent verdict may be dressed as one.
- The **whole-payload persistence** is projected at the point of persistence rather than at the orchestrator call site, so a second caller cannot reintroduce it.
- The **drawer** no longer throws and the flag gate is reachable on the URL route for the first time — but a cold link renders a deliberate "open the run" state rather than refetching. `drawers.md` asks for full reconstitution and this does not meet that bar; everything needed is derivable from the query the results page already runs, but it means carrying the run identifiers in the URL through two more components. Called out in the description rather than left to be rediscovered. Happy to do it here if you would rather it not ship this way.

**Disclosed rather than fixed**, per your P2s: `swap_and_reconcile` defaulting on for every already-saved config (deliberate, but it lands silently and doubles judge cost), and `WinRateChart`'s axis-label change shipping outside the rollout flag.

**Every regression test was confirmed failing without its fix** — `clear-winner` instead of `not-comparable`, 5 separated pairs instead of ≤2, 4 dominance edges instead of 0, `false` instead of `true` on the request-time ceiling. Your "assert the verdict kind, not the headline prose" instruction is honoured: the prose will get reworded, the kind must not regress.

Also worth a note since it bit me twice while binding the CLI scenarios you flagged: the feature-parity checker's annotation regex only consumes a closing `*/` on the **same line**, so a multi-line JSDoc ending in `@scenario` parses cleanly and binds nothing. That file went from 0 of 11 enforced to 11 bound and came off `LEGACY_INERT`.

Ready for re-review.

Aryansharma28:
Follow-up to the above: a second independent review pass over the fix commits themselves found that **CI was red and both failures were mine**, plus one regression of the P1 @sergioestebance blocked on. All fixed; **CI is green across all 22 checks** as of `985dfa8`.

Worth writing up, because two of the four were failures of my *verification* rather than of the fixes:

**The binary-file P1 regressed, and my check said it hadn't.** The perf commit rewrote `useBTLeaderboard.ts` wholesale and put the NUL bytes straight back — so the file Sergio blocked on was binary again, now hiding 5362 bytes instead of 3074. The sweep that was meant to catch it used `grep -qP '\x00'`, which does not reliably match a NUL, so it reported every file clean while the one file it existed to check was still binary.

That is the same defect this branch has been removing from the trust panel all session — a pass reported for a check that never ran — and it landed in the verification instead of the product. It is now a test that reads bytes and runs in CI. It immediately found **seven more files** across the repo with the identical mistake, all fixed; the existing suites that assert NUL-stripping in dataset chunking and Langy turn context prove the escapes are runtime-identical.

**`lint` was red because I formatted a generated file with `npx prettier`** — this repo formats with Biome, and the regen flow shells out to a prettier that isn't a dependency. Formatted properly; the prettier step is what to drop from that flow.

**`test-unit` was red on a fourth stale mock** of the resolver module, in `src/server/api/routers/__tests__/` — outside the two directories my local sweep covered. All five mocks of that module are now accounted for.

**The Biome gate then caught one added violation**, which is a trap worth naming: it compares per-file, per-rule counts against the merge base, so a file that merely gets *more complex* fails it while `biome check --diagnostic-level=error` stays silent. `npx biome check --reporter=rdjson <file>` reproduces it locally in seconds.

---

One genuinely new finding, and it is the interesting one — an interaction between two of the fixes that neither review caught in isolation:

**Making reconciliation failures `skipped` can dissolve the ranking the comparability veto then refuses to produce.** `computeComparability` treats tie rows as load-bearing and says so in its own docstring — a tie *"puts weight in both directions… it is what makes an otherwise split field identifiable."* Those rows were holding the win graph together. Recording order-sensitive rows as no-verdict removes exactly that evidence, and the new veto then declines to rank across the pieces. Executed on a field whose only cross-cluster evidence was tie rows: `identifiable: true` / `tie-at-top` before, `identifiable: false` / `not-comparable` after.

Both changes are individually right, and arguably the combination is the honest outcome. The problem was that it produced *"Not enough overlap to rank these"* for a reason the reader could not see — `detectComparisonColumns` discarded skipped results before anything could count them. It now counts them, and the trust panel reports them either way: how many rows the judge declined, that they carry no weight, and — once they are a large enough share to be the reason the ranking is thin — that **running more rows will not help**, because a judge that order-sensitive needs a different model or prompt.

**Still open and worth a reviewer's judgement:** the mechanism is confirmed by execution, the *frequency* is not. I have no measurement of how often real runs hit this, and `swap_and_reconcile` is default-on for every existing config. That is the difference between this chart usually rendering a ranking and usually rendering nothing, so it deserves a live multi-variant run before merge rather than my say-so.

sergioestebance (dismissed):

sergioestebance (approved):

PR #6117, opened 2026-07-24
feat(event-sourcing): span jobs travel as claim-checks — versioned span_referenced event resolved through the span store

0xdeafcafe:
## What

Stacked on #6111 (merge that first — this PR's base is its branch). The coding-agent span-facts subscriber's jobs stop carrying span payloads: a matched `span_received` stages a small **`span_referenced`** claim-check, and the handler reads the canonical span back from the span store — where `spanStorage` already normalized it exactly once. ADR-069 invariant 4 taken to its resolution: the payload never rides the scheduling plane, and the platform normalizes each span once instead of twice.

### The versioned event (the first one whose versions array is load-bearing)

`span_referenced` follows the house `_EVENT_VERSION_LATEST`/`_EVENT_VERSIONS` convention, and its parse is **discriminate-then-validate**: a payload that doesn't claim the type returns `null` (caller falls through to the full-event path); a payload that claims the type but carries an unknown version **throws into the queue retry** — a mixed-deploy job can never be mistaken for another shape and silently no-op. It is never appended to the event log (deliberately absent from `TRACE_PROCESSING_EVENT_TYPES`, and registered in a staging-only brand list): the durable audit already exists at both ends (`span_received` in the trace log, the facts contribution in the session log), and a derived, single-consumer event in a write-hot log is transport wearing an event costume. If span-relevance ever becomes a non-rederivable decision (model/config-based) or grows a second consumer, it graduates to a durable event — recorded as the flip condition.

### Resolution discipline

- `getNormalizedSpanById`: primary-key point read `(TenantId, TraceId, SpanId)` with **`fallback: "none"`** — a miss is an *expected transient* (the reference races the sibling `spanStorage` write), so it stays one cheap windowed probe and throws into the queue's backoff (25 attempts, 500ms→10min). Never an unbounded scan on the delivery path.
- **The window centres on the span's own start, not on ingest time.** The reference carries `startTimeUnixMs` parsed off the wire span, and the read centres there — the stored row's `StartTime` *is* that value. Centring on `occurredAt` (ingest time) instead would put any span whose duration plus export lag exceeded the window permanently outside its own read: spans export on *end*, and `claude_code.blocked_on_user` is exactly that long-duration class. Such a job could never succeed on any retry — it would exhaust its budget and block the group forever. Falls back to `occurredAt` when the wire span carried no parseable start.
- `delay: 2s` on the subscriber debounces the common case past the sibling write.
- The facts lift is shared verbatim by both paths — a reference job and a full-event job produce the **identical command** (pinned by test, with the store stub routed through `serializeAttributes`/`deserializeAttributes` so the assertion sees a real round-trip rather than the inline normalization output).
- **The round-trip is lossy on purpose, so both picks tolerate it.** `deserializeAttributes` returns numeric-looking strings as numbers, so `resolveConversationKey` and the `service.version` pick accept finite numbers in canonical string form. Without that, an agent emitting a purely numeric session key would key by provider on the full-event path and fall back to the trace on the claim-check path — one session's facts split across per-trace aggregates, during every deploy window in which both paths coexist.
- Dedup keys on the raw wire span id in **both** shapes — the identity is unchanged across the upgrade.
- The seam hook (`enqueue.stage`) is contract-bound to total field-picks; a span with no id stages the full event unchanged, and a throw is reported as a routing failure rather than quietly staging the payload whole. `es_subscriber_enqueue_total` grows the `referenced` outcome, counted only after the handoff succeeds (per the base PR's contract).

### Deploy sequencing (why the consumer commit comes first)

The first commit is the consumer (handler understands references), the second the producer (seam stages them). An old worker dequeuing a reference would silently no-op it, so if a zero-loss window matters, land+deploy the consumer before the producer (deploys run several times daily). Merged together, the exposure is the minutes of one rolling deploy, for coding-agent facts only.

## ADR

ADR-069 sequenced claim-checks into phase 2 and defines one as always carrying the payload's true byte size. This PR ships the claim-check half early for one subscriber and `span_referenced` declares no size, so the ADR gains an amendment stating the split — the reference mechanism exists and has one adopter; invariant 1 is still unmet, so byte-bounded stages and the grant pool behind them are untouched. It also names the two consequences: span-facts delivery is now dependent on the spanStorage projection and the ClickHouse read, and a store outage outliving the shared 25-attempt budget blocks the per-trace group until an operator unblocks it (staged references preserved in Redis — delayed delivery and toil, never loss). A store-miss-specific backoff was considered and declined: the budget is one shared platform policy already sized for ClickHouse recovery cycles, and carving out a per-error-class knob is a queue-contract change with its own blast radius.

## Tests

- Subscriber suite **15/15**: identical-command proof across both paths (through a serialize/deserialize round-trip); numeric-scalar session keys and versions; a 3-day-old span resolving on the start-centred window, and garbage start falling back to ingest time; miss → throws into retry; unknown version → throws; no-span-id → stages whole and processes as before; dedup identity equal across shapes; filter cases.
- Seam contract suite **9/9**: stage swaps payload + counts `referenced`; passthrough counts `staged`; throwing stage fails the routing attempt and never runs the handler; filter and queue-outage cases.
- Span-storage suites **69/69** (the `getSpanByIds` → shared `fetchNormalizedSpanRow` refactor is behavior-preserving).
- tslsp diagnostics clean on all touched files.
- Spec: `specs/event-sourcing/payload-cost.feature` claim-check block, **5 scenarios** bound via `@scenario`. The four original claim-check scenarios carried annotations but no `@unit` tag, so the parity check counted 9 of 13 and never enforced them — tagged, now 14/14 bound.

## Deployment Impact

- **Env vars:** none. **Helm values / chart:** none. **Migrations:** none.
- **Default helm install:** unaffected; one metric label value added on an existing counter.
- **BYOC dataplane:** no impact.
- **Behavioural:** coding-agent span jobs shrink from full span payloads to small references resolved via one keyed ClickHouse point read (+2s dispatch delay). Non-agent projects unaffected (already filtered by the base PR). Rollback-safe: pre-upgrade full-event jobs process identically on new code; a rollback mid-window leaves reference jobs that old code no-ops (facts loss bounded to that window's coding-agent spans, dedup TTL 60s).

## Expected prod effect

Worker RSS pressure from the span-facts queue drops to near-zero per job; `es_subscriber_enqueue_total{outcome="referenced"}` becomes the staging signal, and `clickhouse_windowed_read_total{table="stored_spans"}` picks up the resolution reads (windowed, no unbounded fallback).

## Summary by CodeRabbit

* **New Features**
  * Added claim-check staging for coding-agent span facts to reduce queued payload size while keeping behavior consistent.
  * Introduced claim-check resolution for referenced spans using a persisted normalized-span read, including retries when spans aren’t readable yet.
  * Added fallback processing for cases where a span reference can’t be constructed.
  * Added new “referenced” enqueue outcome and related version validation behavior.

* **Bug Fixes**
  * Standardized single-span retrieval semantics for referenced span resolution.

* **Documentation**
  * Updated payload-cost and claim-check staging scenarios to match the new flow.

### DIFF
--- a/langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
@@ -1016,38 +1016,14 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
         { tenantId, traceId, occurredAtMs },
         (span) => span === null,
         async (window) => {
-          const partition = partitionFragment(window);
-          const client = await this.resolveClient(tenantId);
-          // Single-span fetch. WHERE pins (TenantId, TraceId, SpanId) - the
-          // primary key prefix - so we hit a tiny granule range. ORDER BY
-          // UpdatedAt DESC LIMIT 1 deliberately picks up CH 25.10's
-          // LazilyRead optimiser: heavy columns (SpanAttributes, Events.*,
-          // Links.*) are deferred past the LIMIT, so unmerged versions
-          // don't materialise them. Investigation numbers + the per-query
-          // lock that keeps the optimiser engaged live in
-          // SINGLE_SPAN_FETCH_SETTINGS above. The doc's "Anti-Pattern 1"
-          // rule predates LazilyRead and isn't load-bearing on this shape.
-          const result = await client.query({
-            query: `
-              SELECT ${FULL_SPAN_SELECT}
-              FROM ${TABLE_NAME}
-              WHERE TenantId = {tenantId:String}
-                AND TraceId = {traceId:String}
-                AND SpanId = {spanId:String}
-                ${partition.sqlAnd}
-              ORDER BY UpdatedAt DESC
-              LIMIT 1
-            `,
-            query_params: { tenantId, traceId, spanId, ...partition.params },
-            clickhouse_settings: SINGLE_SPAN_FETCH_SETTINGS,
-            format: "JSONEachRow",
+          const row = await this.fetchNormalizedSpanRow({
+            tenantId,
+            traceId,
+            spanId,
+            window,
           });
-
-          const rows = (await result.json()) as FullSpanRow[];
-          if (rows.length === 0) return null;
-          const [span] = mapNormalizedSpansToSpans(
-            rows.map(mapChRowToNormalized),
-          );
+          if (row === null) return null;
+          const [span] = mapNormalizedSpansToSpans([row]);
           return span ?? null;
         },
       );
@@ -1063,6 +1039,100 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
       );
       throw error;
     }
+  }
+
+  /**
+   * Claim-check resolution read (ADR-069): the canonical single-span read for
+   * internal derivation consumers holding a `span_referenced` staging. A miss
+   * is an EXPECTED transient state here — the reference is often dequeued
+   * before the sibling spanStorage write lands — so `fallback: "none"` keeps a
+   * miss one cheap windowed probe (the caller throws into the queue's backoff)
+   * instead of an unbounded scan per retry. The occurredAt hint comes from the
+   * reference envelope, so the window is authoritative, not guessed.
+   */
+  async getNormalizedSpanById({
+    tenantId,
+    traceId,
+    spanId,
+    occurredAtMs,
+  }: {
+    tenantId: string;
+    traceId: string;
+    spanId: string;
+  } & OccurredAtHint): Promise<NormalizedSpan | null> {
+    EventUtils.validateTenantId(
+      { tenantId },
+      "SpanStorageClickHouseRepository.getNormalizedSpanById",
+    );
+
+    try {
+      return await queryWindowed<NormalizedSpan | null>({
+        table: TABLE_NAME,
+        hintMs: occurredAtMs ?? null,
+        windowMs: DEFAULT_PARTITION_WINDOW_MS,
+        fallback: "none",
+        isEmpty: (row) => row === null,
+        run: (window) =>
+          this.fetchNormalizedSpanRow({ tenantId, traceId, spanId, window }),
+      });
+    } catch (error) {
+      logger.error(
+        {
+          tenantId,
+          traceId,
+          spanId,
+          error: error instanceof Error ? error.message : String(error),
+        },
+        "Failed to get normalized span by id from ClickHouse",
+      );
+      throw error;
+    }
+  }
+
+  /**
+   * Single-span fetch shared by {@link getSpanByIds} and
+   * {@link getNormalizedSpanById}. WHERE pins (TenantId, TraceId, SpanId) - the
+   * primary key prefix - so we hit a tiny granule range. ORDER BY
+   * UpdatedAt DESC LIMIT 1 deliberately picks up CH 25.10's
+   * LazilyRead optimiser: heavy columns (SpanAttributes, Events.*,
+   * Links.*) are deferred past the LIMIT, so unmerged versions
+   * don't materialise them. Investigation numbers + the per-query
+   * lock that keeps the optimiser engaged live in
+   * SINGLE_SPAN_FETCH_SETTINGS above. The doc's "Anti-Pattern 1"
+   * rule predates LazilyRead and isn't load-bearing on this shape.
+   */
+  private async fetchNormalizedSpanRow({
+    tenantId,
+    traceId,
+    spanId,
+    window,
+  }: {
+    tenantId: string;
+    traceId: string;
+    spanId: string;
+    window: WindowFragment | null;
+  }): Promise<NormalizedSpan | null> {
+    const partition = partitionFragment(window);
+    const client = await this.resolveClient(tenantId);
+    const result = await client.query({
+      query: `
+        SELECT ${FULL_SPAN_SELECT}
+        FROM ${TABLE_NAME}
+        WHERE TenantId = {tenantId:String}
+          AND TraceId = {traceId:String}
+          AND SpanId = {spanId:String}
+          ${partition.sqlAnd}
+        ORDER BY UpdatedAt DESC
+        LIMIT 1
+      `,
+      query_params: { tenantId, traceId, spanId, ...partition.params },
+      clickhouse_settings: SINGLE_SPAN_FETCH_SETTINGS,
+      format: "JSONEachRow",
+    });
+
+    const rows = (await result.json()) as FullSpanRow[];
+    if (rows.length === 0) return null;
+    return mapChRowToNormalized(rows[0]!);
   }
 
   async findSpanResourcesByTraceId({
--- a/langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
+++ b/langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
@@ -1047,8 +1047,15 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
    * is an EXPECTED transient state here — the reference is often dequeued
    * before the sibling spanStorage write lands — so `fallback: "none"` keeps a
    * miss one cheap windowed probe (the caller throws into the queue's backoff)
-   * instead of an unbounded scan per retry. The occurredAt hint comes from the
-   * reference envelope, so the window is authoritative, not guessed.
+   * instead of an unbounded scan per retry.
+   *
+   * The hint must be the SPAN'S OWN start, not the ingest time of the event
+   * that referenced it: this table's partition column is `StartTime`, so a
+   * window centred on ingest time excludes any span whose duration plus export
+   * lag exceeded it — and spans export on end. Such a span would sit outside
+   * every retry's window forever. Callers pass the reference's parsed
+   * `startTimeUnixMs`, falling back to the envelope's occurredAt only when the
+   * wire span carried no parseable start.
    */
   async getNormalizedSpanById({
     tenantId,
--- a/langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
@@ -101,7 +101,12 @@ export function createCodingAgentSpanFactsDispatchSubscriber(deps: {
           tenantId: ref.tenantId,
           traceId: ref.data.traceId,
           spanId: ref.data.spanId,
-          occurredAtMs: ref.occurredAt,
+          // Center the store's partition window on the span's OWN start (the
+          // stored row's StartTime is this exact value), not on ingest time:
+          // a span that ran longer than the window and exported on end would
+          // otherwise sit permanently outside an occurredAt-centered read and
+          // exhaust its retries into a blocked group.
+          occurredAtMs: ref.data.startTimeUnixMs ?? ref.occurredAt,
         });
         if (span === null) {
           // Not readable YET — the reference raced the sibling span write.
@@ -162,6 +167,14 @@ function liftContribution({
   const serviceVersion = span.resourceAttributes["service.version"];
   if (typeof serviceVersion === "string" && serviceVersion.length > 0) {
     facts["service.version"] = serviceVersion;
+  } else if (
+    // The store read-back deserializes numeric-looking versions ("1.0",
+    // "2024") as numbers — keep the fact instead of silently dropping it on
+    // the claim-check path only.
+    typeof serviceVersion === "number" &&
+    Number.isFinite(serviceVersion)
+  ) {
+    facts["service.version"] = String(serviceVersion);
   }
 
   return {
--- a/langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
@@ -4,6 +4,7 @@ import type { EventSubscriberDefinition } from "../../../subscribers/eventSubscr
 import { SPAN_RECEIVED_EVENT_TYPE } from "../../trace-processing/schemas/constants";
 import {
   isSpanReceivedEvent,
+  makeSpanReferencedEvent,
   parseSpanReferencedEvent,
   type SpanReceivedEvent,
   type TraceProcessingEvent,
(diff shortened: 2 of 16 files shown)


0xdeafcafe:
@coderabbitai review

langwatch/src/server/app-layer/traces/repositories/span-storage.clickhouse.repository.ts
```diff
@@ -1056,6 +1032,100 @@ export class SpanStorageClickHouseRepository implements SpanStorageRepository {
...
+      query_params: { tenantId, traceId, spanId, ...partition.params },
+      clickhouse_settings: SINGLE_SPAN_FETCH_SETTINGS,
+      format: "JSONEachRow",
+    });
+
+    const rows = (await result.json()) as FullSpanRow[];
+    if (rows.length === 0) return null;
+    return mapChRowToNormalized(rows[0]!);
+  }
+
```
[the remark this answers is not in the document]
0xdeafcafe: This query is a verbatim extraction of the pre-existing `getSpanByIds` read on the base branch (`feat/payload-cost-doctrine`), refactored into a shared `fetchNormalizedSpanRow` and reused by `getNormalizedSpanById` — the PR doesn't introduce this pattern.

The `ORDER BY UpdatedAt DESC LIMIT 1` shape is a deliberate, *measured* exception to Anti-Pattern 1, documented on `SINGLE_SPAN_FETCH_SETTINGS` (lines 154–178). On a fat span (19 attrs / 127 events / ~88 KB `Events.Attributes`), dev CH 25.10:

- Form A (this query): `read_bytes ≈ 14,933` (~15 KB) — CH 25.4+ LazilyRead defers the heavy columns (`SpanAttributes`, `Events.*`, `Links.*`) past the `LIMIT`.
- Form B (the doc's IN-tuple scalar-subquery form): `read_bytes ≈ 9,706,538` (~9.7 MB).

So on this exact heavy-column shape the IN-tuple form is **~650× more expensive** — adopting it would regress the read, not fix it. The safety conditions are pinned in code, not assumed: `query_plan_optimize_lazy_materialization=1` is locked per-query so a cluster/profile can't flip the optimiser off (Form A with LazilyRead disabled measured the same ~9.7 MB as Form B), `LIMIT 1` is well inside the lazy-materialization threshold (default 10 on 25.10), and the method's minimum CH version is 25.4. The doc's Anti-Pattern 1 predates LazilyRead; the inline comment records that it isn't load-bearing on this shape.

On the embedded ast-grep SQL-injection note — false positive. The only string interpolations are compile-time constants (`FULL_SPAN_SELECT`, `TABLE_NAME`, and `partition.sqlAnd`, a static `AND StartTime …` fragment over a hardcoded column); every tenant/trace/span/window value is bound through parameterized `{name:String}` placeholders (`query_params` + `partition.params`).

Keeping as-is.

0xdeafcafe:
@coderabbitai review

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
```diff
@@ -54,61 +67,165 @@ export function createCodingAgentSpanFactsDispatchSubscriber(deps: {
...
+        const span = await deps.getNormalizedSpanById({
+          tenantId: ref.tenantId,
+          traceId: ref.data.traceId,
+          spanId: ref.data.spanId,
+          // Center the store's partition window on the span's OWN start (the
+          // stored row's StartTime is this exact value), not on ingest time:
+          // a span that ran longer than the window and exported on end would
+          // otherwise sit permanently outside an occurredAt-centered read and
+          // exhaust its retries into a blocked group.
+          occurredAtMs: ref.data.startTimeUnixMs ?? ref.occurredAt,
```
rogeriochaves, line 109: **P1 (fixed in d4659466f).** The resolution read windowed StartTime to occurredAt ± 2 days with `fallback: "none"` — but occurredAt is INGEST time (Date.now() at trace-request-collection), while the stored row's StartTime is the span's client-reported start, and spans export on END. Any span whose duration plus export lag exceeded the window (a `claude_code.blocked_on_user` held over a weekend is exactly the long-duration class in CODING_AGENT_SPAN_NAMES; offline-buffered telemetry too) was permanently invisible to every retry: ~2.5h of backoff, exhaustion, group blocked, manual unblock, exhaust again forever, facts never land. The repo's own `readTraceSpans` docblock documents this exact hazard and mandates unbounded fallback for span reads. Fix: the reference now carries `startTimeUnixMs` parsed off the wire span (total, null on garbage) and the read centers there — the stored StartTime IS that value, so it stays a pruned point-read at any duration, falling back to occurredAt when the wire had no parseable start. Pinned by two new unit cases (3-day-old span resolves with the start-centered hint; garbage start falls back to ingest time).

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
```diff
@@ -54,61 +67,165 @@ export function createCodingAgentSpanFactsDispatchSubscriber(deps: {
...
+  tenantId,
+  occurredAt,
+}: {
+  span: NormalizedSpan;
+  tenantId: string;
+  occurredAt: number;
+}): ContributeSpanFactsCommandData {
+  const sessionKey = resolveConversationKey(span.spanAttributes);
+  const facts = liftSpanFacts(span.spanAttributes);
+  const serviceVersion = span.resourceAttributes["service.version"];
```
rogeriochaves, line 167: **P2 (fixed in d4659466f).** The identical-command invariant did not survive the ClickHouse attribute round-trip: `deserializeAttributes` is deliberately lossy (numeric-looking strings come back as numbers), while `resolveConversationKey` and this `service.version` pick were strictly string-typed. An agent emitting a purely numeric session key (`session.id: "175335720123"`; only Claude Code's UUIDs are structurally safe) got provider-keyed sessions on the full-event path and `trace_fallback` on the claim-check path — one session's facts silently split across per-trace aggregates, and both paths coexist during every deploy window. The pinning test could not see it because its store stub returned the inline normalization output, not a round-tripped row. Fix: both picks now accept finite numbers in canonical string form, and the identical-command test routes its stub through `serializeAttributes`/`deserializeAttributes`, plus a dedicated numeric-scalars case. One honest residue documented in that test: `"1.0"` canonicalizes to `"1"` through the round-trip — kept, not dropped, formatting lost.

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
```diff
@@ -54,61 +67,165 @@ export function createCodingAgentSpanFactsDispatchSubscriber(deps: {
...
+          // exhaust its retries into a blocked group.
+          occurredAtMs: ref.data.startTimeUnixMs ?? ref.occurredAt,
+        });
+        if (span === null) {
+          // Not readable YET — the reference raced the sibling span write.
+          // Throwing is the contract: the queue retries with backoff, and a
+          // span that never lands surfaces as a loud exhausted job, never a
+          // silent drop.
+          throw new Error(
+            `Referenced span is not readable yet (trace ${ref.data.traceId}, span ${ref.data.spanId}); retrying until the span store write lands`,
```
rogeriochaves, line 117: **P2, note (yours to accept or budget).** Span-facts delivery now depends on spanStorage projection + ClickHouse read health, with a hard ceiling of roughly 2.5h cumulative backoff before per-trace groups block and need manual unblocks even after the store recovers (re-staged data preserved, so delayed delivery plus operator toil, not loss). Pre-PR the subscriber was store-independent. The ADR-066/069 incident history shows multi-hour hot-project backlogs are real. Acceptable as a designed tradeoff if named: consider a longer store-miss-specific backoff or higher attempt budget for this error class, and a line in the ADR or an ops note about the blocked-group recovery path.

specs/event-sourcing/payload-cost.feature
```diff
@@ -51,17 +53,35 @@ Feature: Payload cost governs the scheduling plane
...
-  # reference instead of the full payload riding the queue.
   Scenario: a matched event's heavy payload travels as a claim-check, not inline
     Given an event the subscriber considers relevant whose payload is large
     When the event is routed to subscribers
     Then the staged job carries a reference to the payload, not the payload
-    And the handler reads only the small slice it needs through that reference
+    And the handler completes its work by reading the payload from its canonical store
+    And the reference preserves the identity the scheduler orders and dedups by
+
+  Scenario: a reference that cannot be resolved yet retries, never drops
```
rogeriochaves, line 102: **P3, doc drift.** ADR-069 (merged via #6111) still sequences claim-checks into phase 2 ("matched spans travel as full payloads until phase 2") and defines a claim-check as always carrying the payload's true byte size — this PR ships the claim-check half early and `span_referenced` carries no byte-size declaration. Worth a one-line status amendment to the ADR in this PR (claim-check half shipped for this subscriber; cost-honesty still pending under the remaining @planned scenarios) so the next phase-2 implementer does not read the ADR's sequencing as current state.

rogeriochaves (commented):
Reviewed the stack half on top of #6111 (scheduling-identity mirroring, dedup across both staged shapes, discriminate-then-validate versioning, and the dequeue path were all verified sound; the getSpanByIds refactor is behavior-preserving and the new point read is TenantId-first on the primary key). Found 1 P1 and 1 P2, both concrete enough that I fixed them on the branch as d4659466f rather than just reporting, per Rogerio; details inline on the lines they lived at, plus 1 coupling note and 1 doc-drift note that are yours to accept or amend. Your e539277a2 already cured the branch's red typecheck (the seam test's fabricated brand vs the closed EventType enum), so after this the whole stack should be green once #6127 merges and the inherited coding_agent_sessions TTL-boundary flake stops hitting shard 1.

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
0xdeafcafe: Confirmed present and carried through the rebase onto the re-based `feat/payload-cost-doctrine` (now `5fb9ee3a9`). The read centers on `ref.data.startTimeUnixMs ?? ref.occurredAt`, so the window is anchored on the value the stored row actually holds in `StartTime` rather than on ingest time, and a long-running span stays a pruned point read at any duration. Both pinning cases survive: `centers the store read on the span's own start, not on ingest time` and `falls back to ingest time when the wire span carried no parseable start`. Green locally.

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
0xdeafcafe: Confirmed present after the rebase. `resolveConversationKey` and the `service.version` pick both accept finite numbers in canonical string form, so a purely numeric session key resolves to the same key on the inline and the claim-check path instead of splitting one session across per-trace aggregates during a deploy window. The pinning test now routes its store stub through `serializeAttributes`/`deserializeAttributes` (test file lines 112-117) so the round-trip is actually exercised rather than the inline normalization output being handed back, plus the dedicated `keys the session and keeps the version fact identically to the full-event path` case. The `"1.0"` -> `"1"` canonicalization residue is documented in that test as kept-not-dropped.

langwatch/src/server/event-sourcing/pipelines/coding-agent-processing/subscribers/codingAgentSpanFactsDispatch.subscriber.ts
0xdeafcafe: Accepted as a named tradeoff, and half-actioned. Written up in a new ADR-069 amendment (`bf6e9a459`): span-facts delivery is now dependent on the spanStorage projection landing the row and on the ClickHouse read succeeding, where the subscriber was previously self-contained; a store outage outliving the retry budget exhausts the job and blocks the per-trace group until an operator unblocks it, with the staged references preserved in Redis (delayed delivery and toil, never loss). That is the ops note you asked for, in the ADR rather than a separate doc so it sits next to the sequencing it qualifies.

The backoff/attempt-budget half I am declining, with reasoning. There is no per-subscriber knob today: every group-queue job shares `JOB_RETRY_CONFIG` (25 attempts, ~2h27m cumulative), and its docblock says it was sized for precisely this class -- riding out a ClickHouse rolling restart or ZooKeeper session recovery without parking the group in `:blocked`. So the existing budget is not an accident this PR is inheriting; it is already the considered answer to store unavailability. Carving out a store-miss-specific backoff means adding a per-error-class retry policy to the shared queue contract, which is a change with platform-wide blast radius and deserves its own design rather than riding in on this subscriber. And per ADR-068's measure-before-you-limit rule, the right trigger for that work is observing this class of blocked group in practice with the retry metrics to size it -- not a speculative multiplier now. The amendment records that explicitly as the follow-up.

specs/event-sourcing/payload-cost.feature
0xdeafcafe: Fixed in `bf6e9a459`. ADR-069 gains an amendment (`Amendment: the claim-check half of phase 2, shipped early for one subscriber`) stating the split you identified: the claim-check mechanism exists and has exactly one adopter, while the cost-honesty half does not -- `span_referenced` declares no byte size, so invariant 1 is still unmet and the phase-2 work it gates (byte-bounded in-flight and retry stages, and the phase-3 grant pool behind them) is untouched. The remaining `@planned` scenarios stay the honest statement of what is left.

The two places that would otherwise read as current state now point at it: the Consequences bullet (`Matched spans still travel as full payloads until phase 2...`) is marked superseded for this subscriber, and the phase-2 Sequencing item notes that its claim-check clause shipped early while its cost-honesty clause did not. A future phase-2 implementer should read the sequencing as `the reference mechanism exists and has one adopter`, not as `phase 2 is underway`.

langwatch-agent:

**Mode:** Deep Review
**Linked issue:** ADR-069 lineage. **Stacked on #6111, whose branch is this PR's base, so review #6111 first.**

**Top inspection targets**
1. `Must Check` - the claim-check race. A `span_referenced` event carries a pointer, and the handler reads the span back from the store. Confirm the span is durably in the store *before* the claim-check is staged, or a fast subscriber resolves a reference that does not exist yet.
2. `Must Check` - the miss path. What happens when the span store returns nothing: retry, drop, or dead-letter? Under the old design the payload travelled with the job and could not go missing, so this introduces a new failure mode.
3. `Must Check` - versioning. The event is described as a versioned `span_referenced`. Confirm in-flight jobs staged under the old payload-carrying shape still drain correctly across the deploy.

**Production risk (scariest first):** silently dropped span facts if the reference cannot be resolved, which is data loss that does not announce itself. Second: a deploy that strands in-flight jobs of the previous shape.

1. Is the span guaranteed durable before the claim-check is staged?
2. What is the behaviour on an unresolvable reference?
3. How do jobs staged under the old shape drain during rollout?

PR #6122, opened 2026-07-24
fix(group-queue): planned shutdowns stop counting as worker deaths in the poison guard

0xdeafcafe:
## What

The poison guard's "worker death" signal is a claim strike that survives when a worker dies before its clearing `finally`. A **planned** shutdown produced the same signature: `close()` drains, but any job the drain budget (20s) or the platform's SIGKILL abandoned kept its strike — so every rolling deploy (several per day) could tick innocent in-flight groups toward the park threshold. This is the mechanism behind the false-blame parks seen in prod on 2026-07-24 (e.g. a healify burst group parked with "3 consecutive worker deaths" while zero pods were actually killed).

Two changes, both preserving the guard's self-discrimination — a worker whose event loop a poison job seized can never execute either path, so genuine crash-loops keep their count:

- **`close()` sweeps in-flight claim strikes first**, before the drain race, while the event loop is provably alive. Jobs that complete during the drain re-clear harmlessly; jobs the budget abandons leave no strike behind and re-dispatch cleanly on the next boot.
- **A claim made after shutdown was requested records no strike.** A group already at the threshold parks on the next boot's claim instead of during the drain (bounded: one extra dispatch attempt, strikes intact).

The workers process already wires SIGTERM → drain (`workers.ts` → `App.close()` → queue `close()`); this PR makes the strike accounting honest along that path.

## Spec

`specs/event-sourcing/poison-group-park-guard.feature`: the previously `@unimplemented` scenario "graceful shutdown mid-job does not count as a poison strike" is now implemented and bound; a new scenario pins the no-strike-during-drain claim behavior.

## Tests

- `groupQueue.poisonGuard.integration.test.ts` **14/14** (real Redis): new test drives an in-flight handler, asserts the strike exists, begins `close()` without awaiting, and asserts the strike is swept **while the job is still running** — i.e. the abandon path is covered, not just the clean drain.
- tslsp diagnostics clean.

## Deployment Impact

- **Env vars / Helm / migrations:** none.
- **Default helm install:** unaffected. **BYOC dataplane:** no impact.
- **Behavioural:** deploys stop contributing false "worker death" ticks; genuinely poisonous groups still park (their strikes survive because a seized loop can't run the sweep). One edge is deliberate: a group at threshold claimed mid-drain parks on the next boot instead of during shutdown.

## Summary by CodeRabbit

* **Bug Fixes**
  * Graceful shutdown during an in-flight job no longer leaves behind poison-guard strikes.
  * Planned shutdowns no longer cause groups to be incorrectly marked as blocked/poisoned.
  * If shutdown captures state before a strike finishes recording, the system now rechecks and clears it to avoid false poison detection later.

* **Tests**
  * Added integration coverage for graceful shutdown with work still in progress.
  * Extended the poison-group behavior scenarios to cover draining/claim timing.


langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -771,12 +785,14 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
         });
         return;
       }
+      this.inFlightStrikeGroups.add(groupId);
     }
```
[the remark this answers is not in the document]
0xdeafcafe: Addressed in 664f9ca50. Confirmed the race: `recordClaimStrike()` awaits (L772) before `inFlightStrikeGroups.add(groupId)` (L788), so `close()` can flip `shutdownRequested` and take its sweep snapshot during that await — snapshotting a set that does not yet hold the group. The sweep misses it, and if the drain budget (or SIGKILL) then abandons the job, its clearing `finally` never runs and the strike survives — exactly the false "worker death" the guard must not book during a planned shutdown.

Fix: on the synchronous heels of the add (no await between them, so `close()` cannot interleave there) I re-check `shutdownRequested` and clear this group's own strike when the shutdown began mid-claim. The add and this check are jointly exhaustive — either `close()`'s snapshot already saw the group and swept it, or it did not and we clear it here.

Also added the regression you asked for: it parks the claim after the strike lands but before the add, drives `close()` past its sweep snapshot (via the post-snapshot `drainAndDisconnect` hook), then releases the claim and asserts the strike is cleared while the handler is still in flight — so it fails without the recheck, since the strike would otherwise survive until a `finally` the abandon path never reaches.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
```diff
@@ -747,12 +753,20 @@ export class GroupQueueProcessor<Payload extends Record<string, unknown>>
...
+   * are swept in {@link close} while the event loop is provably alive. A group
+   * already at the threshold is therefore not parked during the drain either —
+   * its intact strike count parks it on the next boot's claim instead.
    */
   private async processWithRetries(dispatched: DispatchResult): Promise<void> {
     const { stagedJobId, groupId, jobDataJson, originalScore } = dispatched;
 
     const strikeThreshold = readClaimStrikeThreshold();
-    if (strikeThreshold > 0) {
+    const recordStrikes = strikeThreshold > 0 && !this.shutdownRequested;
```
rogeriochaves, line 771: **P3, observation (non-blocking).** One interleaving lands outside the docstring's "a group already at the threshold is not parked during the drain" promise: claim starts pre-shutdown (`recordStrikes` true), close() flips the flag during the `recordClaimStrike` await, and that strike happens to cross the threshold. The park branch runs before the post-add recheck, so the group parks during the drain. The strikes that got it there were real prior deaths, so the park is at most one cycle early and only on a razor-thin window; fine to leave, but the docstring could soften to "is parked at most one claim early" if you want it exact.

specs/event-sourcing/poison-group-park-guard.feature
```diff
@@ -95,15 +95,18 @@ Feature: GroupQueue poison-group park guard
...
-  @unimplemented
-  # The clear-on-survival semantics is exercised by the completion/failure
-  # scenarios; a direct drain-mid-shutdown binding needs a close() harness.
   Scenario: graceful shutdown mid-job does not count as a poison strike
     Given a group whose job is in flight when the worker begins a graceful shutdown
     When the shutdown drains or abandons the in-flight job with the event loop alive
     Then the group's claim strike is cleared
     And the group is dispatched normally after the worker restarts
 
+  Scenario: a claim made during a graceful shutdown records no strike
```
rogeriochaves, line 104: **P3, test coverage note.** This new scenario has no direct binding: both new integration tests exercise strikes that were recorded and then cleared (sweep path, recheck path), while the `recordStrikes = threshold > 0 && !shutdownRequested` gate itself, a claim entering processWithRetries after the flag flip recording nothing, is asserted nowhere. It is deterministically testable without internals mocks if you ever want it: `queue.setConcurrency(1)`, gate a holder group's handler, send a second group so its claim parks in the fastq buffer, `close()`, release the holder, then assert the second group processed with `recordClaimStrike` never called for it and its strikes key null. Non-blocking since the sweep + recheck cover the leak paths that matter.

rogeriochaves (commented):
Reviewed: the sweep-before-drain + no-strike gate + synchronous post-add recheck interleaving analysis holds up (walked every close()/claim ordering), and the two gate-orchestrated tests pin the racy paths well. No blocking findings; 2 P3 notes inline. The test-integration(1) red on this branch is the coding_agent_sessions drifted-window fixture sitting on the retention TTL boundary, main-inherited, fixed on #6127 (ff462eb6b); a rerun after that merges goes green.

langwatch/src/server/event-sourcing/queues/groupQueue/groupQueue.ts
0xdeafcafe: Good catch — softened the wording in b0d9e21c5. You're right about the interleaving: a claim that begins pre-shutdown (`recordStrikes` latched true) whose recorded strike crosses the threshold while `close()` flips the flag mid-await runs the park branch before the recheck, so the group parks during the drain.

I kept the behaviour rather than defer it: that park rides only real prior-death strikes, and clearing them to push the park to the next boot would discard the very evidence that the group is poison. So parking it one claim early is correct — it's the docstring that over-promised. It now reads "parked at most one claim early … never on a shutdown-born strike", which is exact.

specs/event-sourcing/poison-group-park-guard.feature
0xdeafcafe: Thanks — you're right the gate itself has no direct binding. I tried to build the recipe and it turns out the concurrency-1 buffering doesn't reproduce against the current dispatcher, so I've left the scenario unbound deliberately; rationale below.

`dispatchBatch` claims against `availableSlots = globalConcurrency − processingQueue.length() − processingQueue.running()` (dispatcher.ts:141-147) and claims nothing when that's ≤ 0. With the holder occupying the single slot, `running() = 1` ⇒ `availableSlots = 0`, so the second group is never claimed into the fastq buffer — there's no "claimed-but-parked" state, because the backpressure exists precisely to keep a fastq backlog from forming. And once `close()` flips the flag the dispatcher loop (`while (!shutdownRequested)`) stops claiming entirely, so the second group would never be claimed at all and would go unprocessed rather than exercising the gate. In other words the gate guards a path the dispatcher already prevents end-to-end.

That's also why it's benign rather than a leak path: if the gate regressed and did record a strike during shutdown, the sweep (group in the in-flight set) or the post-add recheck — both bound by the two new tests — would clear it. An unbound gate can't leave a surviving strike, which lines up with your non-blocking call.

I can pin it explicitly, but only by going around the dispatcher (flip `shutdownRequested`, hand-push a synthetic `DispatchResult` into the processing queue), which trades a real end-to-end path for coverage of a one-line defensive guard. Happy to add that if you'd rather have the scenario bound — just say the word.

rogeriochaves (approved):

Issue #6132, opened 2026-07-24
Promote annotations into a dataset so judges can be compared as targets


Aryansharma28:
## The question users actually have

Not *"is my judge okay?"* but **"which judge should I use?"** — and that second question is the one nothing in the product answers today.

## The pieces already exist

Three of the four parts are already built and shipping:

- **Evaluators are already first-class targets.** `BatchTargetColumn.type` includes `"evaluator"`, and virtual `_eval_${evaluatorId}` targets already flow through the experiments-v3 results pipeline.
- **`Annotation.expectedOutput` already exists** in the Prisma schema, is already written by `annotation.repository.ts`, and already reaches ClickHouse as `expected_output` via `collectorSpan.utils.ts`. A reviewer can already write down the *correct answer*, not just a verdict.
- **Datasets already have `expected_output` columns**, and the Bradley-Terry leaderboard (#6110) already ranks N targets against each other with cost and latency alongside.

The missing piece is small: **a bridge that promotes annotated traces into a dataset row** (input + the reviewer's `expectedOutput`).

## Why that bridge is worth more than a bespoke chart

Once annotations can become a dataset, "which judge is right" stops being a special case and becomes an ordinary experiment:

1. Judges run as targets against human-labelled data.
2. The existing leaderboard ranks them.
3. **Cost and latency come for free** — which matters enormously and no agreement chart can provide. A judge 2% more accurate at 10× the cost is usually the wrong pick.
4. Comparing three judges is the same problem as comparing three prompts, already solved.

## Scope note — what this does *not* subsume

Thumbs-up/down annotations can't ride this path. A 👍 is a *verdict on an output*, not an output, so it has no `expectedOutput` to become a dataset column. For verdict-only annotations — by far the most common kind — the judge-vs-reviewer agreement view in #6131 remains the right surface, and the false-positive/false-negative split it shows is genuinely lost when you collapse agreement to a single pass rate.

So the two are complementary:

| Annotation shape | Path |
|---|---|
| Reviewer wrote the correct answer (`expectedOutput`) | This issue — dataset + leaderboard, ranks N judges, includes cost/latency |
| Reviewer only gave 👍/👎 | #6131 — agreement matrix, shows *direction* of error |

## Suggested first slice

1. "Add to dataset" from an annotation, carrying `input` + `expectedOutput`.
2. Bulk version: promote all annotations on a run into a new dataset.
3. Confirm an evaluator target can be run against that dataset end-to-end and lands on the leaderboard.

[...]

PR #6133, opened 2026-07-24
fix(annotations): stop shipping full User rows to the browser

Aryansharma28:
## What this is now

`annotation.getByTraceIds` used `include: { user: true }`, returning **every** `User` column — `email`, `emailVerified`, `lastLoginAt`, `deactivatedAt`, `image` — for every annotation on screen. There is no output schema on the procedure, so all of it was serialised to the browser.

The sibling `getByTraceId`, immediately above it in the same file, already selects only `{ id, name, image }`. This brings the batch variant in line.

Checked every consumer before narrowing: the annotation surfaces render `user.name` / `user.image`, and the dataset mapping's author field falls back to `annotation.email`, not `user.email`. Nothing reads the dropped columns.

## Correction — this PR originally claimed something false

It opened as "the `annotations.*` dataset mapping is dead code — nothing populates `trace.annotations`", and added a join to fix it.

**That was wrong.** `TracesMapping.tsx` on `main` already fetches annotations (line 166) and attaches them to each trace (lines 198-207). Mapping `annotations.expected_output` into a dataset **already works today**. My original research grepped for the type name `TraceWithAnnotations` rather than for the attach itself, which uses a plain object spread — so I missed it and drew the wrong conclusion.

The join has been removed. It was redundant with `TracesMapping`'s own, it fired a second uncached fan-out of the same queries (different chunk boundaries → different query keys), and its result was overwritten downstream anyway.

What survives is the one genuine finding: the over-fetch above.

Closes nothing on #6132 — see that issue for the corrected scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved annotation user data handling so the annotations UI returns and displays only essential user details (e.g., name and image) rather than full user records.
  * Aligned the annotation user data shape with what the UI actually uses, keeping the existing filtering and sort order behavior unchanged.

### DIFF
--- a/langwatch/src/server/api/routers/annotation.ts
+++ b/langwatch/src/server/api/routers/annotation.ts
@@ -257,8 +257,19 @@ export const annotationRouter = createTRPCRouter({
           },
           projectId: input.projectId,
         },
+        // Only what the UI renders. `include: { user: true }` returned every
+        // User column — email, emailVerified, lastLoginAt, deactivatedAt — and
+        // there is no output schema on this procedure, so all of it reached
+        // the browser for every annotation on screen. Mirrors the sibling
+        // getByTraceId above, which already selects narrowly.
         include: {
-          user: true,
+          user: {
+            select: {
+              id: true,
+              name: true,
+              image: true,
+            },
+          },
         },
         orderBy: {
           createdAt: "asc",
--- a/langwatch/src/server/tracer/tracesMapping.ts
+++ b/langwatch/src/server/tracer/tracesMapping.ts
@@ -12,9 +12,16 @@ import {
 import { getRAGChunks, getRAGInfo } from "./utils";
 
 // Define a Trace type that includes annotations for use within this file
-// This assumes the Annotation type comes from Prisma
+// This assumes the Annotation type comes from Prisma.
+//
+// `user` asks for only the field this file reads (`author` uses `user.name`).
+// Requiring the whole Prisma `User` forced every caller to fetch every user
+// column — email, lastLoginAt and the rest — just to satisfy the type, which
+// is how those columns ended up being shipped to the browser.
 type TraceWithAnnotations = BaseTrace & {
-  annotations?: (Annotation & { user?: User | null })[];
+  annotations?: (Annotation & {
+    user?: { name?: string | null } | null;
+  })[];
 };
 
 /**
--- a/langwatch/src/server/tracer/tracesMapping.ts
+++ b/langwatch/src/server/tracer/tracesMapping.ts
@@ -1,4 +1,4 @@
-import type { Annotation, AnnotationScore, User } from "@prisma/client";
+import type { Annotation, AnnotationScore } from "@prisma/client";
 import { z } from "zod";
 import { getSpanNameOrModel } from "../../utils/trace";
 import { datasetSpanSchema } from "../datasets/types";
@@ -562,12 +562,10 @@ export const TRACE_EXPANSIONS = {
     label: "annotation",
     expansion: (trace: TraceWithAnnotations) => {
       const annotations = trace.annotations ?? [];
-      return annotations.map(
-        (annotation: Annotation & { user?: User | null }) => ({
-          ...trace,
-          annotations: [annotation],
-        }),
-      );
+      return annotations.map((annotation) => ({
+        ...trace,
+        annotations: [annotation],
+      }));
     },
   },
   "events.event_id": {


Aryansharma28:
@coderabbitai review

Aryansharma28 (commented):
## Review

### Overview

This narrows `annotation.getByTraceIds`'s Prisma `include: { user: true }` to `select: { id, name, image }`, and correspondingly narrows `TraceWithAnnotations`'s `user` type in `tracesMapping.ts` from the full Prisma `User` to `{ name?: string | null } | null`. There's no output schema on this tRPC procedure, so every `User` column — `email`, `emailVerified`, `lastLoginAt`, `deactivatedAt` — was previously serialized to the browser for every annotation rendered on screen. Nothing in the annotation surfaces or the dataset-mapping author field reads any of the dropped columns (the PR body documents checking every consumer, including that the mapping's author field falls back to `annotation.email`, not `user.email`).

Worth calling out: the PR description is unusually candid about its own history — it originally claimed `annotations.*` dataset mapping was dead code and added a join to "fix" it, then the author caught that this was wrong (`TracesMapping.tsx` on `main` already attaches annotations via a plain object spread, which a grep for the type name missed) and reverted the join. What's left is only the genuine over-fetch fix. That kind of self-correction in a PR description, rather than quietly dropping the redundant join, is good practice worth naming.

### Code quality & style

- Both changed comments explain *why*, not *what* — the `annotation.ts` comment names the concrete columns that leaked and points at the sibling procedure it now matches; the `tracesMapping.ts` comment explains why the wide type existed (forced by the old `include: { user: true }`) rather than just describing the new type. Good adherence to the project's WHY-not-WHAT comment convention.
- `tracesMapping.ts` also removes a stale, redundant inline type annotation on the `annotations.id` expansion's `.map()` callback (`(annotation: Annotation & { user?: User | null })`), which had explicitly re-declared the wide `User` type independently of `TraceWithAnnotations` — this was papering over the narrowed type one property away and is exactly the kind of thing that silently reintroduces an over-fetch if left in place. Removing it and letting the callback parameter infer from the array is correct; `User` becomes a fully dead import as a result and is dropped too.
- Naming and shape are consistent with the sibling `getByTraceId` procedure immediately above, which the PR explicitly aligns with rather than inventing a new convention.

### Potential issues / risks

- **No test guards the narrowed `select` shape.** Nothing in `annotation.*.test.ts` asserts that `getByTraceIds`'s returned `user` object excludes `email`/`emailVerified`/etc. — so a future edit that widens the query back to `include: { user: true }` (or drops the `select` entirely) wouldn't be caught by any existing test. That said, this is a **pre-existing, repo-wide gap**, not something this PR introduces: the sibling `getByTraceId`'s own narrow `select` (which this PR explicitly mirrors) has no such regression test either, confirmed by grepping every `annotation.*.test.ts` file for a `user.email`/`user.emailVerified` assertion — zero hits. Worth a follow-up test covering both procedures together, not a blocker on this PR specifically.
- **Blast radius is genuinely small and was checked.** Confirmed independently: `SLACK_APP_MANIFEST`-style "is this used elsewhere" concern doesn't apply here, but the equivalent check does — `getByTraceIds` and `TraceWithAnnotations` are each used in exactly the places the PR body claims (annotation UI rendering `user.name`/`user.image`, dataset mapping's author field using `annotation.email`). No orphaned caller found expecting the wider shape.

### Test coverage

- 350/351 relevant tests pass (1 pre-existing skip, unrelated) across `src/server/tracer/`, `src/server/annotations/`, and the annotation router's test files. `typecheck:all` clean. Biome findings identical file-for-file against `origin/main` (9 pre-existing warnings, 0 errors, 0 new) — confirmed by diffing both versions of the two changed files directly rather than trusting `pnpm lint`'s repo-wide baseline noise.
- See the coverage gap noted above under Potential issues.

### Security considerations

This is a **PII-minimization fix**, not a new security surface — it removes previously over-fetched columns (`email`, `emailVerified`, `lastLoginAt`, `deactivatedAt`) from a response with no output schema, rather than adding new data exposure. No new auth/authz path, no new external input, no change to the tenant-scoping in the surrounding query (the `where` clause and `projectId` filter are untouched by this diff). Net effect is a strict reduction in what reaches the browser.

### Performance

No hot-path change. The narrowed Prisma `select` is if anything marginally cheaper (fewer columns fetched per row) than the `include: { user: true }` it replaces. The redundant join removed in an earlier iteration of this branch (not present in the final diff) would have been a real perf fix, but that's already reflected in the current, narrower diff — nothing further to flag here.

- Whether any external consumer (mobile client, public API, a webhook payload) reads the dropped `User` columns from this specific procedure — the PR body's "checked every consumer" claim covers the in-repo frontend surfaces I can grep for for, but I have no visibility into external integrations.

PR #6186, opened 2026-07-25
fix(model-providers): let an organization with no project add a provider

### LINKED ISSUE #6142: Onboarding: the AI coding agents path creates no project, so Add Model Provider silently does nothing
Found during a competitive gateway benchmark, signing up from scratch as a new buyer.

## Repro

1. Sign up for a new account.
2. At onboarding, choose "Track AI coding agents".
3. An organization and a team are created. **Zero projects** are created.
4. Go to Model Providers and click "Add Model Provider".
5. A menu opens, then nothing happens. No dialog, no error, no explanation. It fails the same way on a second attempt.

The Model Providers page requires a project, and there is no project in the organization, but nothing on the page says so.

## Why it matters

This is a dead end on a first-run path. A new buyer who picks the coding-agents onboarding cannot configure a model provider, so the gateway cannot route anything. Getting past it during this benchmark required creating a project through the API by hand, which a real evaluator will not do. They will leave.

## Suggested acceptance

- Either create a default project on this onboarding path, or block the path until a project exists.
- The Model Providers page should state the precondition rather than rendering an action that cannot work.
- The "Add Model Provider" action should never open and then silently do nothing. If it cannot proceed, say why.

rogeriochaves:
Fixes #6142. Supersedes the "Create a project first" empty state that shipped in #6143, which treated a symptom.

## Why this area changes twice in a day

#6143 read the Model Providers page as needing a project, and made the page say so politely: an empty state offering project creation, with adding, editing and deleting disabled and a reason on each. That is a well-built version of the wrong answer. The buyer still could not configure a provider, so the gateway still had nothing to route with, which was the whole of #6142.

The premise was false. `ModelProvider` is anchored to `organizationId`, and its reach is the `ModelProviderScope` rows attached to it (`ORGANIZATION | TEAM | PROJECT`). A project is one of the three scopes a provider can reach, never where it is stored. The page carried a comment saying the opposite:

> A provider is always stored inside a project, so every write on this page needs one.

The schema contradicts it, and it is what the empty state was built on. It is gone.

## What `projectId` was actually doing on the write path

Traced before changing anything. It was never ownership:

| Call site | What it did with `projectId` |
|---|---|
| `repository.update(id, projectId, data)` | **nothing** — the parameter was unused in the body |
| `repository.delete(id, _projectId)` | **nothing** — already underscore-prefixed as unused |
| `repository.create({ projectId, ... })` | only `scopes ?? [{ PROJECT, projectId }]`; never written to a column. The org anchor comes from `resolveSingleOrganizationForScopes(scopes)` |
| `findExistingProvider(id, projectId)` | only `resolveProjectOrganizationId(projectId)`, i.e. a handle to get the org |
| `validateApiKey` router | unused in the handler; permission anchor only |
| `update` / `delete` routers | `checkProjectPermission(...)` anchor |

So the coupling was a permission anchor plus a scope default, and the real block was at the tRPC layer: `projectId: z.string()` required, and the client sending `projectId ?? ""`.

## The change

`projectId` and `organizationId` are both optional on `update`, `delete` and `validateApiKey`; a request naming neither is refused. A project resolves to its organization and still wins as the narrower handle.

**Authorization is unchanged and still the authoritative gate.** `assertCanManageAllScopes` is untouched and runs per scope, fail-closed, on the row's current scopes and on the submitted set. The new router middleware only swaps the coarse tenant gate: with a project it is the existing `checkProjectPermission`; without one it is organization membership, which establishes the caller belongs to the tenant they named and nothing more. Same division of labour as the `scopeAwarePermissionMiddleware` already used by `setRoleAssignmentForScope` in this router.

One deliberate narrow requirement kept: a **create** with no project must send an explicit scope set, because there is no project to default it from. That also means the organization path is fully covered by the per-scope check rather than falling through it.

Deleting by provider name still needs a project, since that contract means "this provider type in this project".

## Personal workspace: the other leg, since landed

**Updated after merge.** This section originally argued against provisioning a personal workspace on the coding-agents path, on the grounds that `getProjectCount`/`getTeamCount` had no `isPersonal` filter and the FREE plan is `maxTeams: 1`, so ensuring one would put every new organization over its team limit at signup.

That objection was real, and it has since been closed: #6189 and #6190 added `isPersonal: false` to both counters. So this PR is **not an alternative to provisioning a personal workspace** — it is the remaining leg of the same design, and #6190 is the other.

The second objection I raised is also resolved, and by the same PR. I noted that `useOrganizationTeamProject` picked `teams.find((t) => t.projects.length > 0)`, so on a governance org the personal team would be the only one with a project and the page would start filing organization-scoped credentials into a personal workspace. #6190 changed ambient selection to exclude personal teams outright, which is what makes the combination safe.

The composition of the two is now covered by a regression guard in #6212: a governance signup lands on a shared team with no project, the personal team is never selected, and an organization-anchored write never attaches a scope to the personal project.

Seat counts were clean throughout: `ensure()` creates no `OrganizationUser`, and seat maths is member-derived.

## Specs

`specs/model-providers/first-project-required.feature` encoded the wrong requirement, so it is renamed to `providers-without-a-project.feature` and rewritten around the real rule. `provider-list.feature`'s pointer is updated. Nothing else in `specs/model-providers/` or `specs/settings/` asserted a project was required.

## Tests

Both suites were proven to have teeth by reverting the source and re-running.

- `modelProvider.noProject.integration.test.ts` (new, real Postgres): an organization with **zero projects** creates, lists, edits and deletes an organization-scoped provider, through the service and through the tRPC router. 18 tests. With the fix reverted, **10 fail**, the headline being `TRPCError: invalid_type … path: ["projectId"] … expected string, received undefined`.
- `model-providers.no-project.integration.test.tsx` (rewritten RTL): the add action is available with no project, the provider menu opens, the drawer is opened with the organization, and row actions work. 17 tests. With the fix reverted, **9 fail**.

Full sweep: unit 2888 passed / 0 failed; integration 1643 passed. Two integration failures in `modelIdBoundary.integration.test.ts` are unrelated and pre-existing (the test calls the live Anthropic API and `claude-sonnet-4-20250514` now 404s).

## Dogfooding

Seeded an organization on the governance track with a team and **no project**, then used the page as the buyer would, on a real browser.

Landing on Model Providers: the ordinary empty state, both add actions live, no "Create a project first".

Empty state

The provider list opens instead of being inert.

Add menu

The drawer opens at organization scope. Note the URL carries `drawer.organizationId` and no `drawer.projectId`.

Drawer at org scope

Saved with a real key, which also exercised the API-key validation probe. The row shows the organization scope chip.

Saved with org scope chip

Edit and delete are available, and editing resolves the row by organization anchor alone, key preserved and masked.

Row actions

Verified in the database afterwards: one row, `organizationId` set, `scopes: [ORGANIZATION:<org>]`, key encrypted, **0 projects in the organization**. Renaming it updated the same row without duplicating it or blanking the key.

And the routing prerequisite, which is what #6142 was actually about: the gateway's own resolver `eligibleModelProvidersForVk` returns that provider for an organization-scoped virtual key, with zero projects in the organization.

## Not verified

I exercised the gateway's provider-resolution path in the control plane, not a live request through a running Go gateway process; port 5563 was held by another worktree's gateway and pointing it at this control plane risked cross-contaminating another agent's run.

## Summary by CodeRabbit

* **New Features**
  * Model providers can now be added, edited, and deleted at the organization level even when no project exists.
  * API key validation now supports organization scope as well as selected scoped contexts (for example, team/project scope).
* **Bug Fixes**
  * Fixed a scenario where the model provider setup drawer could remain in a non-terminating loading state when no project is available.
* **Tests**
  * Expanded coverage for no-project organizations, scope-aware permissions/validation, and updated create/update/delete expectations.


langwatch/src/server/api/routers/modelProviders.ts
```diff
@@ -222,13 +251,15 @@ export const modelProviderRouter = createTRPCRouter({
...
+      z
+        .object({
+          ...tenantAnchorSchema,
+          provider: z.string(),
+          customKeys: z.record(z.string()),
+        })
+        .superRefine(requireTenantAnchor),
     )
-    .use(checkProjectPermission("project:update"))
+    .use(checkProjectOrOrganizationPermission("project:update"))
```
langwatch-agent: **P1 — `validateApiKey` is now reachable by read-only organization members and can perform a caller-controlled server-side request.** When `projectId` is absent, this middleware falls back to `checkOrganizationPermission("organization:view")`; both `MEMBER` and `EXTERNAL` have that permission. The handler then passes `customKeys` to `validateProviderApiKey`, and the `custom` provider accepts `CUSTOM_BASE_URL`, which is used directly by `fetch()` without an egress/SSRF guard. Before this PR the same operation required `project:update`, so a view-only member could not turn it into an arbitrary outbound probe. Keep the organization-anchored validation path behind a manage-level permission (for example `organization:manage`), or route it through the same scope-management authorization as provider writes. Add a regression test that an EXTERNAL/MEMBER user in a projectless org is rejected before a custom URL can be fetched.

langwatch-agent (changes requested):
Requesting changes: the no-project validation path lowers authorization for a caller-controlled outbound fetch. See the inline P1.

langwatch/src/server/api/routers/modelProviders.ts
rogeriochaves: Confirmed and fixed in c44d93e320. You were right, and it was reachable exactly as described.

I took the second option you offered — routing the probe through the same scope-management authorization the provider writes use — rather than pinning it to `organization:manage`. Two reasons. It keeps one authorization story for this router instead of adding a third permission variant, and a manage-level org permission would have locked out a team admin in a projectless org: they can legitimately create a TEAM-scoped provider, but `handleSave` blocks the save when validation fails, so gating the probe at org level would have made the provider unaddable for them.

So `validateApiKey` now takes the `scopes` the credential is being set up for and runs `assertCanManageAllScopes` against them. "May I probe for this scope" is now answered by the same code as "may I store a credential at this scope". The project path is untouched and still `project:update`.

**The regression test observes the real path, and it has teeth.** Four rejection cases (MEMBER, EXTERNAL, a member naming a team they cannot manage, and a probe naming no scopes), each asserting `fetch` was never called, plus a fifth proving an org admin still gets through to the request so the gate is not just a broken feature. `fetch` is recorded rather than stubbed to succeed, so a test passing because the request merely failed would not count.

Reverting the gate and re-running gives the escalation verbatim:

```
AssertionError: promise resolved "{ valid: false, …(1) }" instead of rejecting
+   "error": "Failed to validate API key. Please check your network connection and base URL.",
```

That resolve is the MEMBER's request to `http://169.254.169.254/latest/meta-data` actually leaving the server. With the fix, all four reject before any call, and the whole file is 23/23.

**One more hole while I was in there.** `findByIdForOrganization` matches on `(id, organizationId)` with no scope predicate, and `assertCanManageAllScopes` is satisfied *vacuously* by an empty list. So a provider row carrying zero `ModelProviderScope` rows — invisible to every listing query, but reachable by id — would have been waved through on the strength of having nothing to check. `update` and `delete` now refuse such a row. That widened with this PR too, since the org path lowered the bar to reach those procedures at all.

To your implicit question about the write paths: I checked whether `update`/`delete` have any path where no scope is supplied and the coarse gate is the only thing standing. They do not, once the above is closed. `scopes` is `.min(1)` so an empty array is rejected at the schema; a create with no project and no scopes is a `BAD_REQUEST`; and every mutating path runs `assertCanManageAllScopes` against either the submitted set or the row's current one.

**On the unguarded fetch itself** — separate from the permission bug, and you are right that it stands on its own. `validateProviderApiKey` and `validateKeyWithCustomUrl` both `fetch()` a caller-supplied URL with no egress guard, so a manage-level user can still point them at the metadata endpoint. The repo already has the primitive for this (ADR-040's `sendHttpDestination` with `createSSRFValidator`, which also refuses to follow redirects). I did not apply it here because a naive `blockLocal` would break self-hosted vLLM/Ollama endpoints on private ranges, which is a supported use of the `custom` provider, so it needs a policy decision rather than a drive-by. Filed as #6205 with the options laid out.

rogeriochaves: Correction to my note above on the SSRF half, since I pointed at the wrong primitive.

The canonical one is not the automations sender — it is `@langwatch/ssrf` (`langwatch/packages/ssrf/src/index.ts`), mirrored by Go's `pkg/ssrf` and held to a shared corpus by the `ssrf-conformance` gate. `httpDestination.ts` is a consumer of it, not the source.

That matters because the package already encodes the policy I said needed deciding:

```ts
export function blocked({ ip, blockLocal }: { ip: string; blockLocal: boolean }): boolean
```

Metadata is always refused; other private ranges only when `blockLocal` is set. So `blocked({ ip, blockLocal: false })` kills the metadata-endpoint case while leaving self-hosted vLLM/Ollama on private ranges working, which was my stated reason for not doing it inline. There is no tradeoff to litigate, just work to do in both entry points with hostname resolution and no redirect-following. #6205 is updated accordingly.

Worth flagging separately: `ssrf-conformance` is path-filtered to the ssrf packages, so it goes green on PRs that touch neither. It guards the rule set, not whether callers actually use it — which is why an unguarded `fetch` in the provider-validation path never tripped anything.

Issue #6187, opened 2026-07-25
~18 integration tests silently skip on the native path, including authz


rogeriochaves:
## What

Roughly 18 real-Postgres integration tests never execute on the native path, which is the path developers and CI both use.

They guard on:

```ts
const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL;
```

and skip when it is set. But `setup.ts` (around line 39) sets `TEST_CLICKHOUSE_URL` on **every** integration run, including the native one. So the guard is always true and the suites always skip.

Affected suites include `deleteScope`, `authz`, `multiInstance`, and the plan-limit tests.

## Why it matters

These are silent skips, not failures. The run reports success, the suite count looks healthy, and nobody notices that the authorization and multi-instance guarantees they cover are unverified. `authz` is the uncomfortable one: it exists to prove per-scope permission checks hold.

This is the same shape as #6169 and #6183: a gate that reports success while verifying less than everyone assumes. Skipped tests are arguably worse than a missing gate, because the file exists, reads as coverage, and is counted in the totals.

## How it was found

While reworking #6142, a sibling test in the same directory was observed running fine without the guard, which prompted a check of why the neighbours did not. The guard was dropped on that one file only; every other affected suite is untouched and still skipping.

## Suggested direction

Not prescribing the fix. The question is what the guard was originally meant to express, since `TEST_CLICKHOUSE_URL` clearly no longer distinguishes the two modes now that setup sets it unconditionally. Whatever replaces it, the first honest run may surface real failures in suites that have not executed for a while, so it likely wants sequencing rather than a single flip.

Worth also asserting that these suites report a non-zero test count, so a future regression to always-skip cannot pass as green.

PR #6189, opened 2026-07-25
fix(licensing): personal workspaces no longer consume plan limits

rogeriochaves:
## What

Personal workspaces no longer consume an organization's plan allowance.

Every person on the free plan should be able to track their own coding-agent usage. That needs a personal workspace, and a personal workspace was spending the allowance the customer bought for real work.

A free organization gets 2 projects and 1 team. Provisioning a personal workspace consumed one of each, so a user with a personal workspace and a single real project was already at 2/2 and could not create a second. On the governance signup path the personal team also pushed the org straight to its team limit.

## Changes

**`getProjectCount` and `getTeamCount` now exclude `isPersonal` rows.**

Both are the single source that the enforcement path, the usage page, and the license status panel read, so this keeps them consistent by construction. A count excluded in one place and included in another would be worse than excluding it nowhere, because the number a customer sees would stop matching the number that blocks them.

**Plan limits themselves are unchanged.** Free still allows 2 projects and its current team limit. No plan numbers were touched.

**Removed `getOrganizationProjectsCount`**, a zero-caller export in `limits.ts` that counted org projects without either the archived or the personal filter. It was the exact divergence trap this change is guarding against.

### Where the exclusion stops, deliberately

Resources created *inside* a personal project (datasets, workflows, prompts, agents) still count toward the organization's limits. Personal workspaces are capped at one per (organization, user) by a partial unique index, so excluding them is bounded and cannot be used to evade quota. The resources inside them are not bounded, so exempting those would open an unbounded hole: park every workflow in a personal project and it is free. Flagging this as a judgement call rather than an oversight.

## Tests

New `personal-workspaces-excluded.integration.test.ts` runs against real Postgres and asserts the repository counts, the enforcement service, the tRPC `project.create` route, and `UsageStatsService` all agree that personal workspaces are invisible and that real projects still hit the cap.

**Teeth:** reverting just the two `isPersonal: false` filters fails 7 of its 8 tests, with the expected numbers (`expected 4 to be 2` for projects, `expected 3 to be 1` for teams).

### These suites were not running at all (#6187)

`project.create.planLimit` and `team.createTeamWithMembers.planLimit` both guarded on `describe.skipIf(!!process.env.TEST_CLICKHOUSE_URL)`. `setup.ts` sets that variable on every run, including the native path where Postgres is present, so both suites skipped unconditionally. Verified before touching them: `Test Files 1 skipped (1) | Tests 4 skipped (4)`.

Guard dropped on the files touched here. That exposed a stale fixture in the team suite: it had gone stale behind the skip after the RoleBinding migration and failed with `UNAUTHORIZED`, because `organization:manage` resolves through an ORGANIZATION-scoped `RoleBinding` and `OrganizationUser.role = ADMIN` alone never grants it. Fixture fixed.

Result: 15 integration tests passing across the three suites, 173 unit tests in `license-enforcement`, typecheck clean including test files (which `pnpm typecheck` excludes by default).

## QA

Signed up on a free organization, provisioned a personal workspace, then used the full project allowance for real work.

Final state: **3 projects and 2 teams exist, but only 2 projects and 1 team count.**

Free plan active on the organization:

Free plan active

Personal workspace provisioned, tracking coding-agent usage on the free org:

Personal workspace

Full allowance still available for real work: two real projects alongside the personal workspace. Before this change the second project was blocked.

Full allowance plus personal workspace

Free still stops at two real projects, and reports `2 / 2` rather than counting the personal project into the number:

Limit still stops at two

No UI changed, so there are no light/dark or breakpoint variants to show. The screenshots are behavioural evidence.

## Still open, not addressed here

`useOrganizationTeamProject` selects the current team with `teams.find((team) => team.projects.length > 0)` and no `isPersonal` filter (`langwatch/src/hooks/useOrganizationTeamProject.ts:245-269`). A personal team always has exactly one project, so it can win that selection, and `project` then becomes the personal project that the model-provider settings page hands to its write path. That would file org-scoped credentials into a personal workspace.

This is pre-existing and orthogonal: this PR changes counting only and provisions nothing, so it does not make the selection any more likely to land on a personal team. It does need fixing before personal-workspace provisioning is extended to the coding-agents onboarding path. Left for its own PR rather than mixed into a counting change, since it touches the core navigation hook app-wide.

`specs/features/onboarding/intent-fork.feature` is untouched for the same reason: this PR does not extend provisioning, so ADR-038 v6's reservation of personal-workspace provisioning for CLI login still holds.

## Summary by CodeRabbit

* **New Features**
  * Personal workspaces and their projects no longer count toward organization team and project limits.
  * Usage reporting now excludes personal workspace resources, providing more accurate plan usage information.
* **Bug Fixes**
  * Free-plan organizations retain their available resource allowance when personal workspaces are created.
  * Plan enforcement continues to apply correctly when non-personal projects or teams reach their limits.
  * License-limit notifications now provide clearer delivery status details and are restricted to authorized organization managers.

### DIFF
--- a/langwatch/src/server/api/routers/__tests__/project.create.planLimit.integration.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/project.create.planLimit.integration.test.ts
@@ -30,11 +30,7 @@ import {
 import { FREE_PLAN } from "../../../../../ee/licensing/constants";
 import type { PlanInfo } from "../../../../../ee/licensing/planInfo";
 
-// Skip when running with testcontainers only (no PostgreSQL)
-// TEST_CLICKHOUSE_URL indicates testcontainers mode without full infrastructure
-const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL;
-
-describe.skipIf(isTestcontainersOnly)(
+describe(
   "project.create plan limit enforcement",
   () => {
     const testNamespace = `proj-limit-${nanoid(8)}`;
@@ -255,7 +251,7 @@ describe.skipIf(isTestcontainersOnly)(
 // Regression: BetterAuth uses RoleBinding for team membership — TeamUser is never populated.
 // A user must be able to create a project with only a RoleBinding, no TeamUser row.
 // Previously project.create checked teamUser.findFirst which always returned null, causing a 403.
-describe.skipIf(isTestcontainersOnly)(
+describe(
   "project.create with RoleBinding-only membership (no TeamUser)",
   () => {
     const testNamespace = `proj-rb-${nanoid(8)}`;
--- a/langwatch/src/server/api/routers/__tests__/team.createTeamWithMembers.planLimit.integration.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/team.createTeamWithMembers.planLimit.integration.test.ts
@@ -6,7 +6,11 @@
  *
  * Requires: PostgreSQL database (Prisma)
  */
-import { OrganizationUserRole, TeamUserRole } from "@prisma/client";
+import {
+  OrganizationUserRole,
+  RoleBindingScopeType,
+  TeamUserRole,
+} from "@prisma/client";
 import { nanoid } from "nanoid";
 import {
   afterAll,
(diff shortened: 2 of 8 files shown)


langwatch-agent (commented):
No blocking findings. The project/team limit, license-status, and usage-stat paths all share `LicenseEnforcementRepository`, and the new predicates exclude only `isPersonal=true` rows while preserving the existing archived-project behavior. `git diff --check` and the focused repository unit suite (52 tests) pass; required GitHub checks are green. The new PostgreSQL integration coverage could not run locally because no database is listening on `localhost:55432`.

PR #6208, opened 2026-07-26
fix(security): let ALLOWED_PROXY_HOSTS reach on-premise .local hosts

thiagolmoraes:
## The problem

`ALLOWED_PROXY_HOSTS` lifts the private-IP block for a named host, but it does not lift the cloud-domain block — and `BLOCKED_CLOUD_DOMAINS` carries `.local` and `.localhost` next to the genuine cloud patterns:

```ts
// ssrfConstants.ts
export const BLOCKED_CLOUD_DOMAINS = [
  ".amazonaws.com",
  ".aws.amazon.com",
  ".compute.internal",
  ".internal",
  ".local",       // ← on-premise, not cloud
  ".localhost",   // ← on-premise, not cloud
];
```

`validateNotBlockedCloudDomain` runs *before* the allowlist is consulted, so an operator running LangWatch inside their own datacentre cannot reach `llm.corp.local` even after naming that exact host in `ALLOWED_PROXY_HOSTS`. The request is rejected before the allowlist is read, and no combination of `BLOCK_LOCAL_HTTP_CALLS` / `ALLOWED_PROXY_HOSTS` makes it work.

This is a gap between the documented contract and the behaviour: `.env.example` presents `ALLOWED_PROXY_HOSTS` as the way to "allow specific hostnames when blocking is on", and the on-premise default is described as permissive so operators "can reach internal services" — but internal services named under `.local` are exactly the ones it cannot reach.

## Why the contract had to change

The obvious fix — move `validateNotBlockedCloudDomain` after the allowlist — is the wrong one, and worth spelling out because it looks correct.

`BLOCKED_CLOUD_DOMAINS` is not only on-premise suffixes. Deferring the whole check past the allowlist would make `.amazonaws.com`, `.compute.internal` and `.internal` allowlistable too. `metadata.google.internal` is a **hostname**, and the unconditional metadata guard matches on `BLOCKED_METADATA_HOSTS` — IPs plus the bare label `metadata` — so it would not catch it. Allowlisting `metadata.google.internal` would start resolving, which is precisely the SSRF target the block exists for.

So instead of reordering one check, this splits the suffixes by what they actually name:

- `.local` / `.localhost` describe private infrastructure that hands out no ambient credentials. Blocked by default; an operator may re-enable **one exact host** through the allowlist.
- Everything else in `BLOCKED_CLOUD_DOMAINS` is still rejected before the allowlist is read, so it can never be listed.
- Cloud metadata endpoints were already unconditional and are untouched.

New `ALLOWLISTABLE_LOCAL_SUFFIXES` constant, deliberately narrow — extend only with suffixes that are meaningless outside a private network.

## Two ways this could have widened the hole

Both closed, each with a test:

1. **A `.local` host that is not in the allowlist.** Deferring the check must not open up every `.local` host, so the cloud-domain check runs again after the allowlist misses.
2. **A cloud label wearing an on-premise suffix.** `metadata.google.internal.local` ends in `.local`, so a naive suffix test treats it as on-premise while it still matches `.internal`. The on-premise suffix is stripped before the cloud patterns are matched. This one was caught by the test, not by review — the first implementation let it through.

## What is unchanged

- Cloud metadata endpoints: unconditional, never allowlistable.
- Cloud internal domains: rejected before the allowlist, never allowlistable.
- `.local` / `.localhost` with no allowlist entry: blocked exactly as before.
- Private-IP / `BLOCK_LOCAL_HTTP_CALLS` behaviour: untouched.
- No new configuration — this is the existing `ALLOWED_PROXY_HOSTS` contract doing what its documentation already implies.

## Tests

| Suite | Result |
|---|---|
| `ssrfProtection` + 8 consumers (TS) | 188/188 |
| `packages/ssrf` (shared classification) | 164/164 |
| `pkg/ssrf` (Go, shared vectors) | ok |
| `pnpm typecheck` on touched files | clean |

Four new scenarios: allowlisted `.local` passes; non-allowlisted `.local` blocks; cloud internal domain is not allowlistable; cloud internal domain with a `.local` suffix blocks.

One existing test changed. `"allowlisted host that is also a cloud domain pattern is still blocked"` used `myhost.local` to assert cloud-domain always wins over the allowlist. Its intent — a cloud domain cannot be allowlisted — is preserved, restated with `s3.amazonaws.com`, which is what the assertion was really about.

`specs/security/ssrf-blocking.feature` gains a rule covering the on-premise contract, and `.env.example` documents what `ALLOWED_PROXY_HOSTS` does and does not lift.

## Scope

TypeScript only. If `langwatch_server` (Python) performs the same hostname-suffix validation, it will still reject on-premise `.local` hosts and an operator would see the two planes disagree — worth confirming, and happy to follow up in a separate PR if so. The shared `packages/ssrf` and `pkg/ssrf` layers classify at IP level and carry no hostname-suffix logic, so TS and Go do not diverge here.

## Summary by CodeRabbit

* **New Features**
  * Allowlisted on-premises `.local` and `.localhost` hosts can now be accessed through the proxy.
  * Cloud metadata and internal cloud domains remain blocked, even when allowlisted or modified with local suffixes.
* **Bug Fixes**
  * Unallowlisted private infrastructure hosts continue to be rejected.
* **Documentation**
  * Added configuration guidance for hostname allowlisting and its security restrictions.

### DIFF
--- a/platform/app/src/utils/ssrfProtection.ts
+++ b/platform/app/src/utils/ssrfProtection.ts
@@ -225,6 +225,14 @@ export function isBlockedCloudDomain(hostname: string): boolean {
 }
 
 /**
+ * On-premise DNS labels ("local", "localhost") that may be stripped off the
+ * end of a hostname before the cloud patterns are tested.
+ */
+const strippableLabels = new Set(
+  ALLOWLISTABLE_LOCAL_SUFFIXES.map((suffix) => suffix.slice(1)),
+);
+
+/**
  * Whether a blocked hostname is blocked *only* because of an on-premise
  * suffix (`.local`, `.localhost`) and may therefore be re-enabled for that
  * exact host via ALLOWED_PROXY_HOSTS.
@@ -240,14 +248,19 @@ function isAllowlistableLocalDomain(hostname: string): boolean {
   );
   if (!matchesLocalSuffix) return false;
 
-  // Strip the on-premise suffix before testing the cloud patterns, so a
-  // hostname cannot smuggle a cloud label past the check by appending
-  // ".local" — "metadata.google.internal.local" still reads as ".internal".
-  const withoutLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.reduce(
-    (name, suffix) =>
-      name.endsWith(suffix) ? name.slice(0, -suffix.length) : name,
-    lowerHostname,
-  );
+  // Strip trailing on-premise labels one by one until none is left, then
+  // test the cloud patterns on what remains. Repeated or mixed suffixes
+  // cannot smuggle a cloud name past the check:
+  // "metadata.google.internal.local.local" reduces to
+  // "metadata.google.internal".
+  const labels = lowerHostname.split(".");
+  while (
+    labels.length > 1 &&
+    strippableLabels.has(labels[labels.length - 1]!)
+  ) {
+    labels.pop();
+  }
+  const withoutLocalSuffix = labels.join(".");
 
   const matchesCloudDomain = BLOCKED_CLOUD_DOMAINS.filter(
     (domain) => !ALLOWLISTABLE_LOCAL_SUFFIXES.includes(domain),
@@ -427,7 +440,10 @@ export function createSSRFValidator(config: SSRFConfig) {
       );
     }
 
-    const hostname = parsedUrl.hostname.toLowerCase();
+    // Drop the trailing dot of a fully-qualified name ("s3.amazonaws.com.").
+    // DNS treats both spellings as the same host, so every check below must
+    // see the dotless form or the dot walks the name past the suffix checks.
+    const hostname = parsedUrl.hostname.toLowerCase().replace(/\.$/, "");
     const port = parsedUrl.port
       ? parseInt(parsedUrl.port, 10)
       : parsedUrl.protocol === "https:"
--- a/platform/app/src/utils/ssrfProtection.ts
+++ b/platform/app/src/utils/ssrfProtection.ts
@@ -7,6 +7,12 @@
  * - Cloud metadata endpoints (see ssrfConstants.ts for full list)
  * - Cloud provider internal domains (configured for AWS, see ssrfConstants.ts to extend)
  *
+ * ## What's Blocked by default but allowlistable (ALLOWED_PROXY_HOSTS)
+ * - On-premise suffixes .local / .localhost. They name private infrastructure
+ *   rather than a cloud service handing out ambient credentials, so an operator
+ *   may re-enable one exact host. Everything above stays unconditional — a
+ *   cloud internal domain is never reachable by listing it.
+ *
  * ## What's Blocked (only when BLOCK_LOCAL_HTTP_CALLS is true)
  * Every non-globally-routable address, as classified by the shared
  * `@langwatch/ssrf` rule set (one table, shared byte-for-byte with the Go
@@ -89,7 +95,11 @@ import {
   fetch as undiciFetch,
 } from "undici";
 import { env } from "../env.mjs";
-import { BLOCKED_CLOUD_DOMAINS, BLOCKED_METADATA_HOSTS } from "./ssrfConstants";
+import {
+  ALLOWLISTABLE_LOCAL_SUFFIXES,
+  BLOCKED_CLOUD_DOMAINS,
+  BLOCKED_METADATA_HOSTS,
+} from "./ssrfConstants";
 
 const logger = createLogger("langwatch:ssrfProtection");
 
@@ -212,6 +222,42 @@ export function isBlockedCloudDomain(hostname: string): boolean {
     (domain) =>
       lowerHostname === domain.slice(1) || lowerHostname.endsWith(domain),
   );
+}
+
+/**
+ * Whether a blocked hostname is blocked *only* because of an on-premise
+ * suffix (`.local`, `.localhost`) and may therefore be re-enabled for that
+ * exact host via ALLOWED_PROXY_HOSTS.
+ *
+ * A hostname that also matches a real cloud pattern is never allowlistable:
+ * `metadata.google.internal.local` hits `.internal` too, so it stays blocked.
+ */
+function isAllowlistableLocalDomain(hostname: string): boolean {
+  const lowerHostname = hostname.toLowerCase();
+
+  const matchesLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.some((suffix) =>
+    lowerHostname.endsWith(suffix),
+  );
+  if (!matchesLocalSuffix) return false;
+
+  // Strip the on-premise suffix before testing the cloud patterns, so a
+  // hostname cannot smuggle a cloud label past the check by appending
+  // ".local" — "metadata.google.internal.local" still reads as ".internal".
+  const withoutLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.reduce(
+    (name, suffix) =>
+      name.endsWith(suffix) ? name.slice(0, -suffix.length) : name,
+    lowerHostname,
+  );
+
+  const matchesCloudDomain = BLOCKED_CLOUD_DOMAINS.filter(
+    (domain) => !ALLOWLISTABLE_LOCAL_SUFFIXES.includes(domain),
+  ).some(
+    (domain) =>
+      withoutLocalSuffix === domain.slice(1) ||
+      withoutLocalSuffix.endsWith(domain),
+  );
+
+  return !matchesCloudDomain;
 }
 
 // ============================================================================
@@ -398,13 +444,22 @@ export function createSSRFValidator(config: SSRFConfig) {
       config,
     };
 
-    // Always validate metadata and cloud domains (critical security)
+    // Always validate metadata endpoints (critical security, never allowlistable)
     validateNotMetadataEndpoint(ctx);
-    validateNotBlockedCloudDomain(ctx);
+
+    // Cloud provider internal domains are rejected before the allowlist is even
+    // consulted, so they can never be allowlisted. The exception is the
+    // on-premise suffixes (.local / .localhost), which an operator may re-enable
+    // for a specific host below — they name private infrastructure, not a cloud
+    // service handing out ambient credentials.
+    if (!isAllowlistableLocalDomain(hostname)) {
+      validateNotBlockedCloudDomain(ctx);
+    }
 
     // Allowlist (literal hostname match, case-insensitive) — bypasses
-    // private-IP / localhost blocking. Cloud metadata was already rejected
-    // above, so it can never be allowlisted.
+    // private-IP / localhost blocking, and the .local/.localhost block for
+    // on-premise hosts. Cloud metadata and cloud internal domains were already
+    // rejected above, so neither can ever be allowlisted.
     if (config.allowedHosts.length > 0) {
       const normalizedAllowed = config.allowedHosts.map((h) =>
         h.trim().toLowerCase(),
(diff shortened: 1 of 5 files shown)


langwatch/.env.example
```diff
@@ -304,6 +304,11 @@ LW_GATEWAY_BASE_URL=http://localhost:5560
 # - BLOCK_LOCAL_HTTP_CALLS=true: blocks private IPs / localhost / hostnames
 #   that resolve to private IPs (cloud metadata is always blocked regardless).
 #   Use ALLOWED_PROXY_HOSTS to allow specific hostnames when blocking is on.
+# - ALLOWED_PROXY_HOSTS: comma-separated literal hostnames, e.g.
+#   "llm.corp.local,10.0.5.3". Also lifts the default .local / .localhost
+#   block for that exact host, so an on-prem deployment can reach internal
+#   services by name. Cloud metadata endpoints and cloud internal domains
+#   (.amazonaws.com, .compute.internal, .internal, ...) can never be listed.
```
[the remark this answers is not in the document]
thiagolmoraes: Fixed in 3b0d05385 — you're right, the entry can be listed, it just has no effect. Reworded to "can never be reached through this allowlist — listing one has no effect."

specs/security/ssrf-blocking.feature
```diff
@@ -135,6 +136,45 @@ Feature: SSRF blocking via BLOCK_LOCAL_HTTP_CALLS toggle (TS + Python parity)
...
+      And ALLOWED_PROXY_HOSTS is "s3.amazonaws.com"
+      When TS validates "http://s3.amazonaws.com/bucket"
+      Then the validation fails with an SSRF block error
+
+    @unit
+    Scenario: TS refuses a cloud internal domain dressed up with a .local suffix
+      Given BLOCK_LOCAL_HTTP_CALLS is "true"
+      And ALLOWED_PROXY_HOSTS is "metadata.google.internal.local"
+      When TS validates "http://metadata.google.internal.local/computeMetadata/v1/"
+      Then the validation fails with an SSRF block error
```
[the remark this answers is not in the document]
thiagolmoraes: Holding off on this one, deferring to a maintainer.

The guideline is right in general, but every pre-existing scenario in this file is written the same way — `Given BLOCK_LOCAL_HTTP_CALLS is "true"`, `And ALLOWED_PROXY_HOSTS is "10.0.5.3"`, `Scenario Outline: <impl> ...` with a TS/Python examples table. I followed the local convention so the new rule reads like its neighbours.

Rewriting only the four scenarios I added would leave the file in two styles, and these env vars are the operator-facing contract here — they're what an on-prem operator sets, so naming them isn't leaking an internal detail the way a config value normally would.

Happy to convert them if you'd prefer the new style, but I'd suggest migrating the whole file in one pass rather than splitting it — that feels like its own PR. Let me know which you want.

langwatch-agent (commented):
No blocking findings from static review of this external contribution (SSRF guard fix). Required checks are currently passing. The contributor branch was not checked out or executed.

langwatch-agent:

**Mode:** Deep Review · escalated to the code-review agent
**Linked issue:** none declared. For a change to a security control, a linked issue or a written threat rationale should exist before merge.

**Questionable Premise:** the reasoning is plausible. `.local` and `.localhost` genuinely are not cloud metadata patterns, and they sit oddly in `BLOCKED_CLOUD_DOMAINS`. But the effect is still that an SSRF control gets weaker, the change comes from an outside contributor, and the diff edits `ssrfConstants.ts`, `ssrfProtection.ts`, and the `ssrf-blocking.feature` spec together. When a PR relaxes a guard *and* rewrites the spec that asserts the guard, the spec stops being independent evidence. Review the spec change on its own terms.

**Top inspection targets**
1. `Must Check` - the bypass requires an explicit `ALLOWED_PROXY_HOSTS` entry and cannot be reached by an unlisted host. That allowlist gate is the entire safety argument.
2. `Must Check` - link-local and metadata endpoints stay blocked regardless. `169.254.169.254` and friends must not become reachable through a `.local` name, including via DNS rebinding to a private address.
3. `Must Check` - the deleted or modified spec cases in `ssrf-blocking.feature`. Confirm no case asserting a still-required block was removed to make the change pass.

**Production risk (scariest first):** SSRF reaching internal services in a hosted deployment. Small diff, high blast radius, which is exactly the shape that gets waved through.

1. Is the `.local` allowance strictly gated on an explicit allowlist entry?
2. Which spec scenarios changed, and why was each no longer valid?
3. What is the deployment scenario driving this: self-hosted only, or does it affect the hosted product?

langwatch/src/utils/ssrfProtection.ts
```diff
@@ -214,6 +224,42 @@ export function isBlockedCloudDomain(hostname: string): boolean {
...
+
+  const matchesLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.some((suffix) =>
+    lowerHostname.endsWith(suffix),
+  );
+  if (!matchesLocalSuffix) return false;
+
+  // Strip the on-premise suffix before testing the cloud patterns, so a
+  // hostname cannot smuggle a cloud label past the check by appending
+  // ".local" — "metadata.google.internal.local" still reads as ".internal".
+  const withoutLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.reduce(
```
langwatch-agent, line 246: **P2 - Strip every allowlistable suffix before deciding a cloud name is safe**

`reduce` removes at most one `.local` suffix because the list contains that suffix once. Consequently `metadata.google.internal.local.local` is treated as allowlistable: the first pass leaves `metadata.google.internal.local`, which no longer ends in `.internal`; the exact `ALLOWED_PROXY_HOSTS` match at line 471 then returns before the second cloud-domain check. This contradicts the security contract in the PR that a cloud-internal domain can never be reached through the allowlist, and the new test only exercises one appended suffix.

Normalize in a loop (or test every intermediate hostname after repeatedly removing `.local` / `.localhost`) until no allowlistable suffix remains, then evaluate the non-local cloud patterns. Add regression cases for repeated and mixed suffixes, e.g. `metadata.google.internal.local.local` and `s3.amazonaws.com.localhost.local`, with matching allowlist entries; both must still reject.

langwatch-agent (changes requested):
Static-only review of this external contribution. The explicit allowlist gate is required for the new `.local` path, and an unlisted `.local` hostname is rejected before DNS resolution. One suffix-normalization bypass remains below.

langwatch-agent (commented):
Static security review of external contribution at `3b0d05385c8cc5581de3a3b279e88fc2b0aa423d` (no PR-branch code or dependencies executed). No new merge-blocking issue found. The `.local`/`.localhost` exception is restricted to configured allowlist entries and remains after the explicit metadata and non-local cloud-domain checks; the regression coverage exercises both the allowed on-premise case and unlisted/metadata-shaped names. This preserves the existing explicit-allowlist trust model rather than opening the general proxy path. CI evidence was considered but not rerun under the external-contribution trust boundary.

0xdeafcafe (changes requested):
hi @thiagolmoraes, this pr will need a rebase due to a restructure we did. if you're able to, can you re-request review once that is done and i'll get this approved for you.

thiagolmoraes:
@0xdeafcafe rebased onto main after the restructure — the two commits now sit on top of `2576dbd96` with paths remapped to `platform/app/src/utils`. `ssrfProtection.test.ts` passes (75/75) and typecheck is clean. The re-request-review API refuses cross-fork requests, so pinging you here instead.

0xdeafcafe:
thanks @thiagolmoraes! there is also 1 outstanding P2 comment, i'm not sure i 100% agree with it, i think it would be better to use a url parser to check the tld vs string manipulation however.

platform/app/src/utils/ssrfProtection.ts
```diff
@@ -214,6 +224,42 @@ export function isBlockedCloudDomain(hostname: string): boolean {
...
+
+  const matchesLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.some((suffix) =>
+    lowerHostname.endsWith(suffix),
+  );
+  if (!matchesLocalSuffix) return false;
+
+  // Strip the on-premise suffix before testing the cloud patterns, so a
+  // hostname cannot smuggle a cloud label past the check by appending
+  // ".local" — "metadata.google.internal.local" still reads as ".internal".
+  const withoutLocalSuffix = ALLOWLISTABLE_LOCAL_SUFFIXES.reduce(
```
thiagolmoraes: Confirmed, the finding is real. `reduce` walks the suffix list once, so a hostname with a repeated suffix survives stripping: `metadata.google.internal.local.local` normalizes to `metadata.google.internal.local`, which no longer matches `.internal`, and the exact allowlist match then returns before the deferred cloud-domain check runs. That contradicts the contract this PR states (a cloud-internal name can never be reached through the allowlist), so I'll fix it here. Practical exposure is limited since the operator has to list that exact hostname themselves, but the contract should hold regardless.

@0xdeafcafe on the URL-parser/TLD idea: the hostname already comes out of `new URL()`, so parsing isn't the gap, and a real TLD check would need the Public Suffix List, which doesn't know these names. `.local`, `.localhost` and `.internal` are special-use/private suffixes, not registry TLDs, so tldts/psl would classify exactly the wrong way here. These suffix lists are this module's own policy, not DNS registry facts.

I think the solid part of both suggestions is the same: character-based slicing is the fragile bit. So rather than a strip-until-fixpoint loop, I'd operate on DNS labels:

```ts
const labels = hostname.split(".");
while (labels.length > 1 && STRIPPABLE_LABELS.has(labels.at(-1)!)) {
  labels.pop();
}
return labels.join(".");
```

Repeated and mixed suffixes are handled by construction (`metadata.google.internal.local.local` becomes `metadata.google.internal`, `s3.amazonaws.com.localhost.local` becomes `s3.amazonaws.com`, both still blocked), and the existing `endsWith` checks on the remainder stay correct because the dot anchors the label boundary. I'll add both cases as regressions with matching allowlist entries.

If that works for you I'll push the fix with the tests.

thiagolmoraes: Pushed in d0c086e36. `isAllowlistableLocalDomain` now strips trailing on-premise labels in a loop over the DNS labels, so repeated and mixed suffixes are gone regardless of order. Both cases you named reject, plus the reverse-order `s3.amazonaws.com.local.localhost`, which the old single-pass `reduce` actually let through.

While adding those, a review of the same surface turned up a related hole that predates this PR: a trailing-dot FQDN walked past every suffix check. `new URL("http://metadata.google.internal./").hostname` keeps the dot, so `endsWith(".internal")` misses and the name reaches cloud metadata with no allowlist entry at all. Same for `s3.amazonaws.com.`. I dropped the trailing dot at the entry point (`hostname` is the single raw read of `parsedUrl.hostname`, so one normalization covers the metadata check, the cloud-domain check, and the `.local` allowlist together). Folded it in here since it's the same class of bug and the same function; happy to split it into its own PR if you'd rather keep this one scoped to the original finding.

Seven regression tests cover both directions: repeated/mixed suffixes and the four trailing-dot spellings (`s3.amazonaws.com.`, `metadata.google.internal.`, `metadata.google.internal.local.`, and the bare `metadata.` host). All green, typecheck clean.

langwatch-agent (commented):
No findings from static security review of the external contribution. I checked the allowlist ordering, cloud-domain exclusions, and FQDN handling; I did not execute contributor code or install dependencies.

LangWatch-Review: verdict=clean sha=d0c086e369561001594e0c73780f04aa7da1a6a8 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=d0c086e369561001594e0c73780f04aa7da1a6a8 p0=0 p1=0 p2=0 p3=0

PR #6215, opened 2026-07-26
feat(analytics): capture signed_up and first_trace_integrated in posthog, instrument the coding-agent onboarding screen

rogeriochaves:
Closes three launch-critical PostHog gaps found in the analytics audit: signup and first-trace milestones were only reaching Customer.io, and the coding-agent onboarding screen had zero instrumentation.

### 1. `signed_up` server event
Customer.io already receives `signed_up` from the nurturing hooks, PostHog received nothing. Now `trackServerEvent({ event: "signed_up" })` fires at both user-creation choke points, exactly once per new user:

- `afterUserCreate` (BetterAuth `databaseHooks.user.create.after`) covers every OAuth, Auth0 and BetterAuth email signup. The capture runs before the SSO early returns, so it fires for every new user, not just SSO domains.
- `user.register` tRPC mutation covers email-mode deployments, which create the user via Prisma directly and bypass the BetterAuth hooks.

The two paths are mutually exclusive per user, so there is no double fire.

### 2. `first_trace_integrated` server event
The audit pointed at `customerIoTraceSync.reactor.ts`, but that reactor is implemented and never registered (`pipelineRegistry.ts` has a TODO holding all Customer.io reactors back), so adding the capture there would ship an event that never fires. The capture went into the live first-trace choke point instead: the `projectMetadata` reactor, which flips `Project.firstMessage` on the first real trace.

- Guarded on the `!project.firstMessage` transition, so it fires exactly once per project, and never for repeat writes that only repair the `integrated` flag.
- Fired after the metadata write commits, so a failed write retries the event on the next trace instead of dropping it.
- Inherits the reactor's sample-trace guard: seeded sample traces do not count as an integration.
- Properties mirror the Customer.io event definition: `sdk_language`, `sdk_framework`, plus `projectId`.

### 3. Coding-agent onboarding screen instrumentation
`ViaClaudeCodeScreen` sits inside the existing `onboarding_product` / `via-claude-code` AnalyticsBoundary but emitted nothing. The audit described claude/codex/gemini/opencode agent tabs; the actual screen has Prompt / Skill / MCP tabs, with per-agent quick commands inside the MCP tab. Every interaction now emits, on successful copy only:

| Event (as captured in PostHog) | Properties |
|---|---|
| `onboarding_product.via-claude-code.selected.tab` | `tab`: prompt, skill, mcp |
| `onboarding_product.via-claude-code.copied.prompt` | `skill` id |
| `onboarding_product.via-claude-code.copied.install_command` | `skill` id (Skill tab) or `agent`: claude-code, codex (MCP quick setup) |
| `onboarding_product.via-claude-code.copied.slash_command` | `skill` id |
| `onboarding_product.via-claude-code.copied.mcp_config` | none |
| `onboarding_product.via-claude-code.copied.config_path` | `editor` |

`InlineCopyButton` gained an optional `onCopied` callback for this. The reused `PromptList` and `SkillList` on the traces empty state now render inside a `traces_integrate` AnalyticsBoundary so their events are namespaced on that surface too.

### Identity
All three captures use the internal user id as `distinct_id`, the same id `posthog.identify(userId)` uses client side (`usePostHogIdentify`), so server events join the browser person. `first_trace_integrated` resolves the org admin via `resolveOrgAdmin`, the same identity the Customer.io reactor uses.

One side effect worth knowing: the trace worker now builds the shared PostHog client on its first first-trace event, which starts the feature-flags local-evaluation poller in that process when `POSTHOG_FEATURE_FLAGS_KEY` is set. The stale comment in `posthog.ts` claiming workers never touch PostHog was updated.

## Tests

73 tests across the touched files, all passing (`pnpm test:unit`, no containers):

- `hooks.test.ts`: `signed_up` fires once with the user id for plain, SSO, and no-domain signups.
- `user.register.unit.test.ts` (new): `signed_up` fires once on successful registration, never when the user already exists.
- `projectMetadata.reactor.unit.test.ts`: `first_trace_integrated` fires once with the admin user id, sdk properties and projectId; falls back to `unknown` sdk values; does not fire for repeat traces, already-integrated projects, missing admins, or failed metadata writes.

`pnpm typecheck` and `pnpm typecheck:tests` are clean.

## QA evidence

Verified end to end against a local `pnpm dev` with the real PostHog project key: clicked through all three tabs and every copy affordance, watched the capture calls leave in the network panel (`POST https://eu.i.posthog.com/i/v0/e/` 200), and confirmed arrival in the PostHog project via SQL:

```
onboarding_product.via-claude-code.selected.tab            tab=prompt | skill | mcp
onboarding_product.via-claude-code.copied.prompt           skill=tracing
onboarding_product.via-claude-code.copied.install_command  skill=tracing
onboarding_product.via-claude-code.copied.slash_command    skill=tracing
onboarding_product.via-claude-code.copied.install_command  agent=claude-code
onboarding_product.via-claude-code.copied.install_command  agent=codex
onboarding_product.via-claude-code.copied.mcp_config
onboarding_product.via-claude-code.copied.config_path      editor=Cursor
```

One QA note: posthog-js drops events from automated browsers by default (`opt_out_useragent_filter`), so headless E2E runs will not pollute these metrics.

| Prompt tab | Skill tab (copy toast) | MCP tab |
|---|---|---|
| prompt | skill | mcp |

## Deployment Impact

- No new environment variables, no migrations, no new infrastructure.
- Only three analytics captures: two fire-and-forget server-side `trackServerEvent` calls (`signed_up`, `first_trace_integrated`) and client-side emits on the onboarding screen.
- Events flow to the existing PostHog project through the already-configured `POSTHOG_KEY`; on deployments without it, the server captures no-op and the client provider is not registered.

## Coverage and specs

Brought up to the current bar after review. 19 scenarios, all bound via `@scenario`, and `check:feature-parity` goes from 3563 to 3582 enforced-and-bound.

**Where the scenarios live.** The first-trace milestone joins `specs/traces/project-metadata-onboarding.feature`, which already describes this reactor and previously said nothing about analytics. The signup milestone and the onboarding-screen instrumentation get `specs/analytics/posthog-product-milestones.feature`, next to the existing `posthog-cost-control.feature`. The titles all name PostHog on purpose: Customer.io already owns scenarios called "First trace fires first_trace_integrated event" and "New signup tracks signed_up event" for its own separate reactor, and the parity checker binds by title, so reusing them would have silently bound to the wrong tests.

**Server side** was already tested and is now specced: `signed_up` fires exactly once from the BetterAuth hook (including through the SSO auto-add path and with an unparsable email domain) and exactly once from the email-mode register route, and not at all on a rejected registration. `first_trace_integrated` fires on the `firstMessage` transition against the org admin, reports `unknown` for missing SDK attributes, and does not fire with no admin, on a repeat write, or after a failed metadata write.

**Frontend is newly covered.** `ViaClaudeCodeScreen.analytics.unit.test.tsx` drives all seven `emit(...)` call sites and the new `onCopied` callback through a stubbed `useAnalytics`, pinning each event name and its exact property keys, plus a negative for a failed clipboard write emitting nothing.

The case worth keeping permanently is the last one. It copies every control on every tab, first asserts the clipboard genuinely received the project API key (so the test cannot pass against an inert fixture), then asserts no analytics payload contains that key or any copied string. The quick commands and MCP config embed the key in the copied text, so the property allowlist is a privacy property, not a tidiness one, and it is now pinned rather than verified by reading.

**Deliberately not covered:** that the server `distinct_id` joins the browser person posthog-js identifies. Both sides use the user id, and a test for it would only assert that two mocks were handed the same string. Dogfooding the real PostHog project after merge proves it and a mock-heavy test does not, so it is left to that rather than papered over here.

## Summary by CodeRabbit

- **Analytics**
  - Improved tracking for onboarding actions, including copied setup commands, configuration details, and tab selections.
  - Added visibility into successful account sign-ups.
  - Added tracking for first-time project integrations to improve milestone reporting.
- **Testing**
  - Expanded automated coverage to verify signup and integration analytics across common scenarios.

### DIFF
--- a/langwatch/src/features/onboarding/components/sections/ViaClaudeCodeScreen.analytics.unit.test.tsx
+++ b/langwatch/src/features/onboarding/components/sections/ViaClaudeCodeScreen.analytics.unit.test.tsx
@@ -1,0 +1,307 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Analytics instrumentation on the coding-agent onboarding screen.
+ *
+ * The screen renders install commands and an MCP config that EMBED the
+ * project API key, so the property allowlist is a privacy property, not a
+ * tidiness one: every payload must stay a fixed identifier and must never
+ * carry the copied string.
+ */
+import { ChakraProvider, defaultSystem } from "@chakra-ui/react";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import "@testing-library/jest-dom/vitest";
+
+const emitMock = vi.fn();
+const clipboardWriteMock = vi.fn<(text: string) => Promise<void>>();
+
+vi.mock("react-contextual-analytics", () => ({
+  useAnalytics: () => ({ emit: emitMock }),
+}));
+
+const API_KEY = "sk-lw-test-SUPERSECRET-000";
+
+vi.mock("../../contexts/ActiveProjectContext", () => ({
+  useActiveProject: () => ({
+    project: { id: "project-1", apiKey: API_KEY },
+  }),
+}));
+
+vi.mock("~/hooks/usePublicEnv", () => ({
+  usePublicEnv: () => ({ data: { BASE_HOST: "https://app.langwatch.ai" } }),
+}));
+
+vi.mock("../../../../components/ui/toaster", () => ({
+  toaster: { create: vi.fn() },
+}));
+
+import { ViaClaudeCodeScreen } from "./ViaClaudeCodeScreen";
+
+type EmitCall = [string, string, Record<string, unknown> | undefined];
+
+function emitted(): EmitCall[] {
+  return emitMock.mock.calls as EmitCall[];
+}
+
+/**
+ * The emit calls padded so destructuring a missing call yields undefined
+ * rather than throwing, letting each test assert the count itself.
+ */
+function onlyEmit(): EmitCall {
+  return emitted()[0] ?? ["", "", undefined];
+}
+
+function renderScreen() {
+  return render(
+    <ChakraProvider value={defaultSystem}>
+      <ViaClaudeCodeScreen />
+    </ChakraProvider>,
+  );
+}
+
+/** An element on each tab, used to wait out the AnimatePresence swap. */
+const TAB_MARKER: Record<string, RegExp> = {
+  Prompt: /^Copy prompt$/,
+  Skill: /^Copy install command:/,
+  MCP: /^Copy config$/,
+};
+
+function copyButtons(): HTMLElement[] {
+  return screen
+    .queryAllByRole("button")
+    .filter((b) => (b.getAttribute("aria-label") ?? "").startsWith("Copy "));
+}
+
+function findCopyButton(pattern: RegExp): HTMLElement | undefined {
+  return copyButtons().find((b) =>
+    pattern.test(b.getAttribute("aria-label") ?? ""),
+  );
+}
+
+/**
+ * Switches tab and waits for its content to mount. The tab panels live in an
+ * `AnimatePresence mode="wait"`, so the incoming panel only mounts once the
+ * outgoing one has finished exiting; querying straight after the click finds
+ * the previous tab. Clears the `selected` event the switch itself emits.
+ */
+async function goToTab(name: keyof typeof TAB_MARKER | string): Promise<void> {
+  fireEvent.click(screen.getByRole("button", { name }));
+  const marker = TAB_MARKER[name];
+  if (marker) {
+    // Generous timeout: the swap is a real animation frame chain, and it
+    // runs well past the 1s default when the whole file executes together.
+    await waitFor(() => expect(findCopyButton(marker)).toBeDefined(), {
+      timeout: 10_000,
+    });
+  }
+  emitMock.mockClear();
+}
+
+/**
+ * Clicks every copyable control on every tab and returns everything emitted.
+ * The MCP tab is the point of the sweep: its quick commands and config JSON
+ * embed the raw API key.
+ */
+async function copyEverythingOnEveryTab(): Promise<EmitCall[]> {
+  const collected: EmitCall[] = [];
+  for (const tab of ["Prompt", "Skill", "MCP"]) {
+    await goToTab(tab);
+    for (const button of copyButtons()) fireEvent.click(button);
+    await waitFor(() => expect(emitMock).toHaveBeenCalled());
+    collected.push(...emitted());
+  }
+  return collected;
+}
+
+beforeEach(() => {
+  emitMock.mockClear();
+  clipboardWriteMock.mockReset();
+  clipboardWriteMock.mockResolvedValue(undefined);
+  Object.defineProperty(navigator, "clipboard", {
+    value: { writeText: clipboardWriteMock },
+    configurable: true,
+    writable: true,
+  });
+});
+
+afterEach(cleanup);
+
+describe("when the user copies a skill prompt", () => {
+  /** @scenario Copying a prompt reports the skill it came from */
+  it("emits copied/prompt carrying only the skill id", async () => {
+    renderScreen();
+
+    fireEvent.click(screen.getAllByRole("button", { name: "Copy prompt" })[0]!);
+
+    await waitFor(() => expect(emitMock).toHaveBeenCalled());
+    expect(emitted()).toHaveLength(1);
+    const [action, object, properties] = onlyEmit();
+    expect(action).toBe("copied");
+    expect(object).toBe("prompt");
+    expect(Object.keys(properties ?? {})).toEqual(["skill"]);
+    expect(typeof properties?.skill).toBe("string");
+  });
+});
+
+describe("when the user copies a skill slash command", () => {
+  /** @scenario Copying a slash command reports the skill it came from */
+  it("emits copied/slash_command carrying only the skill id", async () => {
+    renderScreen();
+    await goToTab("Skill");
+
+    const slashButton = findCopyButton(/^Copy \//);
+    expect(slashButton).toBeDefined();
+    fireEvent.click(slashButton!);
+
+    await waitFor(() => expect(emitMock).toHaveBeenCalled());
+    expect(emitted()).toHaveLength(1);
+    const [action, object, properties] = onlyEmit();
+    expect(action).toBe("copied");
+    expect(object).toBe("slash_command");
+    expect(Object.keys(properties ?? {})).toEqual(["skill"]);
+  });
+});
+
+describe("when the user copies a skill install command", () => {
+  /** @scenario Copying a skill install command reports the skill it came from */
+  it("emits copied/install_command carrying only the skill id", async () => {
+    renderScreen();
+    await goToTab("Skill");
+
+    const installButton = findCopyButton(/^Copy install command:/);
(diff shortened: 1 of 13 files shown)


langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** none declared; the body says these are three gaps from an analytics audit.

**Top inspection targets**
1. `Must Check` - "exactly once per new user" for `signed_up`. The event fires at two user-creation choke points (`afterUserCreated` and the register route). Confirm a signup that passes through both does not double-count, since duplicate milestone events corrupt funnel numbers permanently.
2. `Must Check` - what is sent to PostHog. Onboarding instrumentation tends to capture form and identity fields; confirm no email, token, or API key rides along in properties.
3. `Probably Fine` - the copy-button and drawer instrumentation.

**Production risk:** none to serving. The risk is data: duplicated milestones and PII reaching an analytics vendor.

1. Can both signup choke points fire for one user in any flow, for example an invite accepted by a new account?
2. Are event properties allowlisted, or is an object passed through wholesale?

rogeriochaves:
Rebased onto latest `origin/main` (through `4393cfa328`) and force-pushed as `e212546555`. This clears the DIRTY conflict state; the diff is unchanged from the previously reviewed head apart from one Biome whitespace fixup in `projectMetadata.reactor.ts`. No scope was added.

Local gates on the rebased head: `typecheck` clean, `biome check --diagnostic-level=error` clean across all 6569 files, `check:feature-parity` OK, and the 64 unit tests in the three touched suites pass.

**Honest standards assessment.** This branch predates the current bar, and it is uneven rather than uniformly bad.

What is covered: the server-side event emission is genuinely tested, with negatives. `user.register.unit.test.ts` asserts `signed_up` fires once on a successful email-mode registration and does not fire on a rejected one. `hooks.test.ts` asserts `signed_up` fires from `afterUserCreate` including when the SSO auto-add path runs and when the email has no parsable domain. `projectMetadata.reactor.unit.test.ts` asserts `first_trace_integrated` fires on the `firstMessage` transition, and does not fire when no org admin resolves, when the metadata write fails, or on a repeat write.

What is not covered, and what it would take:

1. **No spec scenarios.** Neither event has a scenario anywhere in `specs/`. The natural home is `specs/traces/project-metadata-onboarding.feature`, which already describes this reactor and today says nothing about analytics. Bringing it to standard means a scenario for the `first_trace_integrated` transition (fires once, attributed to the org admin, carries `sdk_language` / `sdk_framework`) plus a signup-side feature file for `signed_up` covering both the BetterAuth and email-mode entry points, each tagged `@unit` and bound to the tests that already exist. The tests are already written, so this is spec-writing and binding, not new test work.

2. **The frontend instrumentation is entirely untested.** Every `emit(...)` added to `ViaClaudeCodeScreen.tsx` and the new `onCopied` callback on `InlineCopyButton.tsx` has no coverage. The only test on that screen is `ViaClaudeCodeScreen.skillOrdering.unit.test.tsx`, which predates this change. Bringing it to standard means RTL tests with a stubbed `useAnalytics` asserting that a successful copy emits `copied` with the right object and properties, that a failed copy emits nothing, and that tab switching emits `selected`/`tab`.

3. **Every test mocks `trackServerEvent`.** Nothing exercises the real PostHog path, so the claim in the code comments that the server `distinct_id` joins the browser person (posthog-js identifies with the same user id) is asserted only in prose. This is the weakest of the three and the least worth fixing with a test; it is better verified by dogfooding against the real project once merged.

Leaving these open rather than expanding the PR. Happy to do (1) and (2) here if you want them before merge.

rogeriochaves:
Answering the two questions from the review brief, both verified against `457ac2443c`.

**1. Can both signup choke points fire for one user? No.** They are structurally exclusive, not merely unlikely.

`afterUserCreate` is wired as BetterAuth's `databaseHooks.user.create.after` (`src/server/better-auth/index.ts:553`), so it fires only for users created through BetterAuth's adapter. The tRPC `register` route creates the row with raw Prisma inside its own transaction (`ctx.prisma.$transaction` -> `tx.user.create`, `src/server/api/routers/user.ts:138`), which never reaches that adapter and therefore never triggers the hook. The route also rejects an email that already exists before it gets that far, so a user who signed up through BetterAuth cannot re-enter through the tRPC path and emit a second event. `user.register.unit.test.ts` pins this with `toHaveBeenCalledTimes(1)` rather than a loose "was called" assertion.

Invite acceptance by a new account goes through BetterAuth like any other signup: one user create, one hook, one event.

**2. Properties are allowlisted, never a wholesale object.** Nothing is spread from a request body, a form, or a user record.

- `signed_up` carries no properties at all, on both paths.
- `first_trace_integrated` carries exactly three: `sdk_language` and `sdk_framework`, each read by explicit key from the fold's resource attributes with an `?? "unknown"` fallback, plus `projectId`.
- The `distinct_id` is the user id, never the email. That is deliberate: it is the same id posthog-js identifies with client-side, which is what makes the server event join the browser person.
- The seven frontend `emit(...)` calls carry only fixed identifiers: `skill.id`, a literal `agent` string ("claude-code" / "codex"), `ep.editor`, and the tab key. No API key, token, endpoint, or email is in any payload, including on the MCP tab where the copied command itself contains the API key.

So no PII reaches PostHog and no milestone can double-count. The remaining gap is the one in my previous comment: none of the frontend emits has test coverage, and there are no spec scenarios for either event.

rogeriochaves:
Closed both gaps from my earlier assessment in `273bbbcac2`, rather than leaving them as an offer.

**Spec scenarios: 19, all bound.** `check:feature-parity` goes 3563 -> 3582 enforced-and-bound. The first-trace milestone joins `specs/traces/project-metadata-onboarding.feature`; the signup milestone and the onboarding-screen instrumentation get a new `specs/analytics/posthog-product-milestones.feature` next to the existing `posthog-cost-control.feature`.

One thing worth flagging: the parity checker binds **by scenario title**, and Customer.io already owns "First trace fires first_trace_integrated event" and "New signup tracks signed_up event" in `specs/features/customer-io-nurturing-integration.feature` for its own separate reactor. Reusing those titles would have bound my scenarios to the Customer.io tests and looked green while pinning nothing about PostHog. Every new title names PostHog explicitly to keep the two vendors distinct.

**Frontend coverage: new file** `ViaClaudeCodeScreen.analytics.unit.test.tsx`, 10 cases, driving all seven `emit(...)` call sites and the new `onCopied` callback through a stubbed `useAnalytics`. Each asserts the event name and the exact property keys, plus a negative for a failed clipboard write emitting nothing.

The permanent one you asked for is the sweep: it copies every control on every tab, asserts the clipboard genuinely received the project API key first (so it cannot pass against an inert fixture), then asserts no analytics payload contains that key or any copied string. The MCP quick commands and config JSON embed the key in the copied text, so that allowlist is a privacy property and it is now pinned rather than verified by reading.

Two things I hit that are worth recording. The tab panels sit in an `AnimatePresence mode="wait"`, so the incoming panel only mounts after the outgoing one finishes exiting; querying straight after the click finds the previous tab, which is why the tab helper waits on a per-tab marker. And the sweep case needed restructuring to stay under the Biome complexity limit, since the new-violations gate counts anything this branch adds.

**Not covered, deliberately:** that the server `distinct_id` joins the browser person posthog-js identifies. Both sides use the user id, so a test would only assert that two mocks received the same string. Dogfooding the real PostHog project after merge proves it and a mock-heavy test does not, so I am leaving it to that rather than dropping it silently or writing a test that pretends.

Gates on the new head: `typecheck`, `typecheck:tests`, `check:feature-parity`, the Biome new-violations gate (no new violations), and 102 unit tests across the 7 touched suites. The RTL suite was run three times to confirm it is not flaky.

PR #6252, opened 2026-07-27
fix(model-providers): say why a key was refused, not just "invalid"

Aryansharma28:
## The report

A customer on Google Cloud could not add Gemini as a provider:

> api keys generated via gemini pasted into the api key field seem to trigger some kind of validation on the frontend and doesn't allow me to continue

Their key returns **200** when they call Google themselves. Their session produced two things:

```json
{"valid": false, "error": "Invalid API key. Please check your API key and try again."}
```
```json
{"path":"modelProvider.validateApiKey","httpStatus":422,
 "code":"validation_error","meta":{"fieldErrors":{},"formErrors":["Required"]}}
```

A form-level `Required` naming **no field** is Zod parsing an input that was not there. The request never reached Google. Two defects in the validation chain, both real.

## 1. The API key was sent in a URL

`validateApiKey` was a tRPC `.query()`, so the client issued it as **GET with the input JSON-encoded into the URL** — and that input is the customer's API key.

- Secrets in URLs are written to access logs, proxy logs and browser history.
- A proxy that strips credential-shaped query parameters leaves the server parsing an absent input — **exactly the 422 above**, raised against a key that is perfectly good.

It changes nothing server-side, so it stays a read, but it now travels as a mutation with the key in a request body. This affected **every** provider, not just Gemini.

## 2. One auth shape was treated as the only one

A credential is not tied to one URL. Google issues Gemini keys from AI Studio, the Cloud console and Agent Platform, and the same key answers on several surfaces. Verified against a live key:

| Auth shape | Result |
|---|---|
| `?key=` on `/v1/models` | **200** |
| `?key=` on `/v1beta/models` | **200** |
| `x-goog-api-key` header | **200** |
| `Bearer` on `/v1beta/openai/models` | **200** |
| `Bearer` on `/v1/models` | 401 |

We only ever tried the first, so "we did not ask the right way" reached the customer as "your key is invalid."

Validation now walks the shapes a provider genuinely supports and stops at the first that answers. It refuses only once every shape has refused — with one deliberate exception: Google's canonical `API_KEY_INVALID` is trusted as final and ends the walk, because it is the provider positively identifying the credential itself as wrong rather than declining this particular surface. Every other refusal, including the vaguer wording the OpenAI-compatible surface returns, keeps going. Providers whose auth is unambiguous keep their single shape and are probed exactly once, as before.

The shapes a provider serves are stated in the registry beside its `defaultBaseUrl`, not recovered from that URL by parsing it.

## Also in here

- **Real refusal reasons.** Every 400/403 was collapsed into one string, discarding Google's machine-readable `reason`. Only `API_KEY_INVALID` now reports an invalid key; `SERVICE_DISABLED` and the restriction reasons say what to change and name Vertex AI where that is the real answer.
- **A refusal is not a dead end.** The probe runs from our servers, so a key restricted to the customer's network, a provider outage, and a key still propagating all look like a bad key. The first refusal explains itself; the next Save goes through. Editing a credential re-arms the probe. This lives in `useCredentialProbeGate` and is consumed by **both** surfaces that run the probe — the settings drawer and `ModelProviderSetup`, which is the onboarding step and the "Langy needs a model to get started" gate. The latter previously hard-blocked with no way past, and the Langy gate has no skip.
- **The key never appears in an error.** Providers quote the request back, and Gemini carries the key in the query string.
- **Bounded and short-circuited.** One 10s `AbortSignal.timeout` is shared across the whole walk rather than granted per shape — four shapes at ten seconds each would let one black-holed host hold a request thread for forty — and the walk stops once the provider positively calls the key invalid. A real key and a plainly invalid one both settle in **one** request.
- **Unreachable is a handled error, not a verdict.** A refused key is an answer; a provider that never replied is the absence of one. It raises `ProviderUnreachableError` with a stable `code`, a `fault` and tips. Because the tRPC boundary replaces a handled error's free text with its code (ADR-045), the words the customer reads come from this code's entry in the client presentation registry, which the hook resolves via `describeError` — otherwise they read the literal string `provider_unreachable`.
- **A provider fault is not our 500.** tRPC v10 has no code for 502, so the envelope cannot carry it; the log line takes the handled error's own status instead, so a typo in a base URL is not counted against our error budget.
- **Credentials stay out of the audit trail.** `customKeys` was redacted, but `extraHeaders` rides the same mutation and is exactly where an `Authorization: Bearer` is typed, and `providerConfig` is a passthrough object. All three are redacted now, keeping the field names, and an unexpected shape is redacted rather than waved through.
- **Onboarding stops re-probing a key nobody touched.** It checked the stored credentials on every save, so picking a model depended on third-party uptime and was blocked outright behind a key that had drifted out-of-band. It now probes what actually changed, which is what the drawer has always done.
- **The drawer says where each credential comes from.** The registry has carried a sentence per field all along and only ever rendered it during onboarding, so `GEMINI_API_KEY` was a bare env-var name. Gemini's now names the Generative Language API — the string a customer must match in the Google Cloud console.

## Verification

Evidence rather than assertion for the parts that matter.

**End to end against live Google, no mocks:**

```
REAL KEY  -> {"valid":true}                                     | requests: 1
BOGUS KEY -> {"valid":false,"error":"Invalid API key. ..."}     | requests: 1
```

That live run earned its keep twice — it caught two defects every mocked test passed straight through:

1. A plainly invalid key returned `"Invalid API key. Please check your API key and try again. Please pass a valid API key"`. The OpenAI-compatible surface answers with a vaguer sentence and **no** `reason`, and the selector preferred it over Google's canonical verdict. Refusals are now ranked.
2. Short-circuiting exposed a test-isolation bug: `clearAllMocks()` does not drain unconsumed `mockResolvedValueOnce` entries, so a test that stopped probing early leaked its leftovers into the next one.

**A real Google 403 is pinned as a fixture.** It carries *two* `reason` fields — `error.errors[0].reason` is `"forbidden"` and says nothing, while the actionable one is in `error.details[]`. A parser reading the first passes every invented fixture and silently loses the diagnosis.

```
 49 unit tests          providerValidation
  5 unit tests          useModelProviderApiKeyValidation (transport + copy)
434 unit tests          all of src/server/api, 20 files
146 integration tests   23 files, all settings + onboarding
```

136 of those 146 are the pre-existing baseline, unchanged. `feature-parity` green (`3085` enforced scenarios bound, credential-validation 24/24). `pnpm typecheck:all` clean.

**The `provider_unreachable` regression is pinned by execution, not by string comparison.** The test builds the real `ProviderUnreachableError`, runs it through the real `errorFormatterForTesting`, wraps it with the real `TRPCClientError.from`, and rejects the real hook with it — then reads what the drawer would render. Reverting the hook to `error.message` fails it exactly as reported:

```
AssertionError: expected 'provider_unreachable' to be "Couldn't reach the provider. Nothing…"
Expected: "Couldn't reach the provider. Nothing answered, so this API key was not
           checked. Check your network connection, and check the base URL is
           correct and reachable."
Received: "provider_unreachable"
```

## Merged with main

`main` moved underneath this branch while it was open, in one way that mattered.

#6010 migrated the platform to handled errors and deleted `errorDisplayMessage`, which this branch's hook used to keep the literal slug out of the drawer. That job now belongs to the code-keyed presentation registry in `src/features/errors`, which is **exhaustive** over `APP_ERROR_CODES` — so a new code with no entry is a typecheck failure and a `codes.unit.test.ts` failure, not a silent slug.

`provider_unreachable` is therefore registered rather than routed around: listed in `APP_ERROR_CODES`, given registry copy that branches on `hasConfigurableEndpoint` (only some providers have a base URL worth telling someone to check), and `meta` carries that flag instead of a pre-folded sentence. `tips` stays structured for the API, CLI and MCP consumers that read it as data.

One inherited assertion changed meaning rather than breaking by accident: a thrown `Error("Failed to fetch")` carries no handled payload, so ADR-045 says the customer reads the generic line. The test now pins that the transport's own string does **not** reach them, which is what #6010 set out to guarantee.

The other two conflicts were incidental — main reindented `modelProvider.noProject.integration.test.ts` wholesale and reordered imports in `ModelProviderCredentialsSection.tsx`.

`main` also began failing `lint` on any per-(file, rule) increase in Biome violations. This branch added nine. Two were real and are fixed as code (`CredentialsSection` crossed the 60-line limit; `handleHttpError` reached complexity 13, and its Gemini branch is now `geminiReasonRefusal`). The other seven were `noExcessiveLinesPerFunction` firing on `describe` callbacks, which is a rule measuring the wrong thing on a suite declaration — it is off for test files now, in the override block that already scopes test-only rules, with the reasoning beside it.

## Follow-up

#6328 — Google Agent Platform keys (`AQ.…`) still cannot be validated or used. That is the "What this does not do" section above, turned into tracked work.

## What this does not do

The reporting customer's key is restricted to `Agent Platform API`. All probe shapes reach `generativelanguage.googleapis.com`, so if that restriction blocks the service the key still fails validation — correctly. What changes is that they are told to allow the Generative Language API or use Vertex AI, instead of being told their key is invalid, and that they can save regardless.

**No code change can make a restricted key work.** `aiplatform.googleapis.com` answers `401 "API keys are not supported by this API"`, which is why Vertex AI needs a service account rather than this key.

**Not done:** no browser QA, and `SERVICE_DISABLED` is still fixture-only — reproducing it needs a key on a project with the API disabled. `API_KEY_SERVICE_BLOCKED` *is* captured from a real response.

*(Earlier revisions of this description said `ModelProviderSetup` still probed on every save. It no longer does — see the last bullet under "Also in here".)*


langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,178 +55,262 @@ function buildModelsEndpointUrl(
...
-        };
-      }
-      return handleHttpError(response);
+      return await handleHttpError(response, context);
     }
 
     return { valid: true };
   } catch {
+    // The request never landed, so this says nothing about the key itself.
     return {
```
0xdeafcafe: return a HandledError instead of crafting this manually.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +55,363 @@ function buildModelsEndpointUrl(
...
+
+/** The request that proves a key works: list the provider's models. */
+type ProbeRequest = { url: string; headers: Record<string, string> };
+
+/**
+ * Builds the models request for an auth strategy. Every provider is probed
+ * the same way; only where the credential rides differs.
+ */
+/**
+ * Every way a provider's credential can legitimately prove itself.
```
Aryansharma28, line 500: [P1] Failure priority was inverted, producing a garbled message against a live key.

**Why:** Asked about a plainly invalid key, the primary endpoints return Google's canonical `API_KEY_INVALID`, which maps to `INVALID_KEY_MESSAGE`. The OpenAI-compatible surface answers `"Please pass a valid API key"` with **no `details[].reason`**, so it is never mapped. The previous selector preferred "whichever refusal differs from our own wording", which picked that vaguer one and appended it.

**Evidence:** real run against live Google, before the fix:
```
BOGUS KEY -> {"valid":false,"error":"Invalid API key. Please check your API key
              and try again. Please pass a valid API key"}
```
Every mocked test passed while this was broken — the hand-written fixtures all carried a `reason`, so the unmapped branch was never exercised.

**Fixed** in `eb0960a`: refusals are ranked (`actionable` > `definitive` > `explained` > `generic` > `unreachable`), so a reason naming something the customer can change outranks the provider's verdict on the key, which outranks a fallback with nothing to add. Regression test pins the exact string. Re-verified live:
```
BOGUS KEY -> {"valid":false,"error":"Invalid API key. Please check your API key and try again."}
```

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +55,363 @@ function buildModelsEndpointUrl(
...
+ * @param candidates - The auth shapes to try, in preference order
+ * @param context - Which provider is being probed, and with which key
  * @returns Promise resolving to validation result
  */
-async function validateWithGeminiAuth(
-  apiKey: string,
-  defaultBaseUrl: string,
+async function runProbeChain(
+  candidates: ProbeRequest[],
+  context: ProbeContext,
```
Aryansharma28: [P2] The chain multiplies the missing-timeout blast radius by four.

**Why:** @coderabbitai already flagged that this `fetch` has no `AbortSignal` — I am not re-raising that. The additive point is that this loop turns one untimed outbound request into **four sequential** ones on a failed validation, all on a tRPC request thread, to a host the customer influences. A provider that hangs now holds the thread for 4x the previous worst case, and the failure path is the common one for the customer this PR is for.

**Evidence:** `buildProbeCandidates` returns 4 shapes for `gemini`; every one is awaited in sequence, and only a success short-circuits.

**Suggested fix:** land the `AbortSignal` from CodeRabbit's comment as a per-request timeout, and consider a budget across the whole chain rather than per-shape so 4 slow shapes cannot stack. Gating shapes 2-4 on the first failure being a *service/restriction* reason (rather than `API_KEY_INVALID`, which is already definitive) would also cut the common case back to one request.

langwatch/src/hooks/useModelProviderApiKeyValidation.ts
```diff
@@ -28,6 +28,9 @@ export function useModelProviderApiKeyValidation(
   const [isValidating, setIsValidating] = useState(false);
   const [validationError, setValidationError] = useState<string | undefined>();
   const utils = api.useContext();
+  // Sends the key in a request body rather than a URL. See the procedure.
+  const validateApiKeyMutation = api.modelProvider.validateApiKey.useMutation();
```
Aryansharma28: [P2] The central fix of this PR had no test coverage. (fixed)

**Why:** every drawer/onboarding test mocks this hook wholesale (`vi.mock("../../../hooks/useModelProviderApiKeyValidation")`), and the onboarding test mocks `ModelProviderSetup` above it. So nothing in the suite observed *which call the hook makes* — which is precisely the change that stops the customer's API key travelling in a URL and causes the `422 formErrors:["Required"]` they reported.

**Evidence:** `rg -rln "useModelProviderApiKeyValidation" --glob '*test*'` returns 7 files, all of which replace the hook with a stub.

**Fixed** in `eb0960a`: `useModelProviderApiKeyValidation.unit.test.ts` asserts `mutateAsync` is called with the credential payload and that the query-style `fetch` is **never** called, so a regression back to the GET transport fails the build rather than silently reinstating the leak.

Aryansharma28 (commented):
## goated-review — end-to-end verification of the validation chain

**Verdict: COMMENT.** The fix is real and now verified against live Google rather than mocks. Two defects found during that verification (one CI-red, one user-facing) have been fixed on this branch at `eb0960a`.

**P0: 1 (fixed) · P1: 1 (fixed) · P2: 2 (1 fixed) · P3: 1 · P5: 3**

### Does this actually fix the customer's problem?

Ran the real `validateProviderApiKey` against live Google with **no fetch mock**:

| | Result |
|---|---|
| Real key | `{"valid":true}` — 111ms |
| Bogus key | `{"valid":false,"error":"Invalid API key. Please check your API key and try again."}` |

And the transport claim holds: `validateApiKey` is now a mutation, so the key travels in a request body instead of a URL. That is the piece that explains the customer's `422 formErrors:["Required"]` against a key returning 200 for them — an absent input, not a bad key.

**Caveat, stated plainly:** the customer's key is restricted to `Agent Platform API`. All four probe shapes reach `generativelanguage.googleapis.com`, so if that restriction blocks the service, his key still fails validation — correctly. What this PR changes for him is that he is told to allow the Generative Language API (or use Vertex AI) instead of being told his key is invalid, and that he can save regardless. **This PR does not make a restricted key work, and no code change can.**

---

### P0 BLOCKER — 1 (fixed)

**`feature-parity` was red, caused by this PR.** Five scenarios added to `credential-validation.feature` had no `@scenario` binding. Verified locally: `FAIL: 5 unbound scenario(s) in enforced files`. Now `20/20 scenarios bound`, `OK: 3081 enforced scenario(s)`.

### P1 HIGH — 1 (fixed)

**Failure-priority was inverted, producing a garbled message.** Only end-to-end running surfaced this — every mock in the suite passed. See the inline comment.

### P2 MEDIUM — 2

- **Probe amplification compounds the missing timeout** (see inline). This is *not* a re-raise of CodeRabbit's timeout finding — it is that the chain multiplies its blast radius 4x.
- **Hook transport had zero coverage** (fixed). Every drawer test mocks `useModelProviderApiKeyValidation` wholesale, so nothing observed which call it makes — meaning the central fix of this PR was untested. Added `useModelProviderApiKeyValidation.unit.test.ts` asserting the mutation is used and the query `fetch` is never called.

### P3 LOW — 1

- `validateKeyWithCustomUrl` remains a `.query()`. It carries **no** secret (it reads the stored key server-side), so this is not a leak — noting it only so the asymmetry is deliberate rather than accidental.

### P5 PRAISE / OBSERVATION — 3

- **The captured-real-403 fixture earns its keep.** A real Google 403 carries two `reason` fields: `error.errors[0].reason` is `"forbidden"`, while the actionable one is in `error.details[]`. A parser reading the first passes every hand-written fixture and silently loses the diagnosis. Worth copying as a pattern wherever we parse third-party errors.
- **Secrets-in-URL was a platform pattern, not a Gemini bug.** Swept the other routers: the only other credential-ish query input is `emailSuppression.resolveUnsubscribeToken`, and unsubscribe tokens are URL-borne by design. So this endpoint was the significant instance — but the class is worth a lint rule.
- **The escape hatch is the load-bearing part for this customer.** It is the only change that helps regardless of whether the diagnosis is right — and the diagnosis in this PR was revised three times before landing.

---

### Existing findings I am NOT re-raising

- **@coderabbitai — no timeout on the probe fetch.** Still open and now more important; referenced in my P2 rather than duplicated.
- **@coderabbitai — `buildProbeRequest` positional params.** Still applies to the renamed `buildProbeCandidates`.
- **@coderabbitai — `refusedCredentials` not reset when the drawer subject changes.** Valid.
- **@coderabbitai — ElevenLabs assertion cannot fail** (the mocked `detail.message` is literally `"Invalid API key"`). Valid; the assertion proves nothing.
- **@coderabbitai — spec comment says saving is blocked "until the probe passes"**, which the escape hatch made false. Valid.
- **@coderabbitai — spec scenarios assert internal probe counts** rather than user-observable outcomes. Valid.
- **@coderabbitai — missing `when` nesting** in the field-guidance test.
- **@0xdeafcafe — return a `HandledError` instead of crafting it manually.** Valid and architectural; it applies to every error constructed in `providerValidation.ts`, not just the one line it is anchored to.

### Things I could not verify from the diff alone

- **No browser QA.** The escape hatch and the rendered helper text are covered by integration tests, but I have not watched a human click through the drawer.
- **`SERVICE_DISABLED` is still fixture-only.** Reproducing it needs an API key on a project with the Generative Language API disabled. `API_KEY_SERVICE_BLOCKED` *is* captured from a real response.
- **The customer's actual key.** Everything about his specific failure is inferred from his screenshots (`Restrictions: Agent Platform API`) plus the 422 he pasted. One curl against his key would settle it.
- **Request-volume impact in prod.** 4 upstream calls per failed validation is fine for a human clicking Save; I have no data on automated/retry callers.

langwatch/src/server/api/routers/__tests__/providerValidation.unit.test.ts
```diff
@@ -313,6 +313,237 @@ describe("validateProviderApiKey", () => {
...
+          detail: { status: "invalid_api_key", message: "Invalid API key" },
+        });
+
+        const result = await validateProviderApiKey("elevenlabs", {
+          ELEVENLABS_API_KEY: "sk_wrong",
+        });
+
+        expect(result.valid).toBe(false);
+        expect(result.error).toContain("Invalid API key");
+      });
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`. You're right that it proved nothing — the mocked `detail.message` was byte-identical to the prefix of our own fallback, so the assertion held whether or not `extractUpstreamMessage` ever read it. Now uses `"The xi-api-key you supplied is not associated with a workspace"`, which the fallback path cannot produce, with a comment saying why the message has to be unlike ours.

langwatch/src/server/api/routers/__tests__/providerValidation.unit.test.ts
```diff
@@ -313,6 +313,237 @@ describe("validateProviderApiKey", () => {
...
+
+        const result = await validateProviderApiKey("openai", {
+          OPENAI_API_KEY: "sk-valid",
+        });
+
+        expect(result.valid).toBe(false);
+        expect(result.error).toContain("API validation failed (500)");
+        expect(result.error).toContain("upstream is on fire");
+        expect(result.error).not.toContain("Invalid API key");
+      });
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `eb0960a`. This was a real CI break, not just a convention miss — `feature-parity` was red on this branch. Verified locally: `FAIL: 5 unbound scenario(s) in enforced files`, now `20/20 scenarios bound` and `OK: 3081 enforced scenario(s)`.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,178 +55,255 @@ function buildModelsEndpointUrl(
...
+ * Builds the models request for an auth strategy. Every provider is probed
+ * the same way; only where the credential rides differs.
+ */
+function buildProbeRequest(
+  strategy: AuthStrategy,
   apiKey: string,
   baseUrl: string,
   defaultBaseUrl: string,
-): Promise<ValidationResult> {
+): ProbeRequest {
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`. `buildProbeCandidates` now takes `{ strategy, apiKey, baseUrl, defaultBaseUrl }` — you're right that the two adjacent strings were the real hazard, since a swapped `baseUrl`/`defaultBaseUrl` would have compiled clean. Also applied to `handleHttpError({ response, context })` and `runProbeChain({ candidates, context })` for consistency, even though those two were type-distinguishable.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,178 +55,255 @@ function buildModelsEndpointUrl(
...
-      // Gemini returns 400 for invalid keys
-      if (response.status === 400 || response.status === 403) {
-        return {
-          valid: false,
-          error: "Invalid API key. Please check your API key and try again.",
-        };
-      }
-      return handleHttpError(response);
+      return await handleHttpError(response, context);
     }
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`, and you were right that consolidating made it a single choke point. Each shape now carries `AbortSignal.timeout(PROBE_TIMEOUT_MS)` (10s), and as you noted the existing `catch` already produced the right message for an aborted request, so the change stayed contained.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,178 +55,262 @@ function buildModelsEndpointUrl(
...
-    return {
-      valid: false,
-      error: "Invalid API key. Please check your API key and try again.",
-    };
+const GEMINI_REASON_MESSAGES: Record<string, string> = {
+  API_KEY_INVALID: INVALID_KEY_MESSAGE,
+  SERVICE_DISABLED:
+    "This key's Google Cloud project does not have the Generative Language API enabled. Enable it in the Google Cloud console, or add this as a Vertex AI provider instead.",
+  API_KEY_SERVICE_BLOCKED:
+    "This key's API restrictions exclude the Generative Language API. Allow it in the Google Cloud console, or add this as a Vertex AI provider instead.",
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`. Both halves were wrong, and the second one was mine misreading the situation. The copy now says "configure a separate Vertex AI provider with service-account credentials" rather than "add this as a Vertex AI provider" — the old wording did imply the rejected key was reusable, and it is not: `aiplatform.googleapis.com` answers `401 "API keys are not supported by this API"`, so Vertex needs a service account. The test comment now describes the disabled-API case instead of claiming no Google Cloud key can ever work.

langwatch/src/components/settings/__tests__/ModelProviderForm.field-guidance.integration.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+   */
+  describe("given the customer holds a Google Cloud key", () => {
+    it("names the API their key has to be allowed to call", () => {
+      const description =
+        geminiEntry?.fieldMetadata?.GEMINI_API_KEY?.description ?? "";
+
+      expect(description).toContain("Generative Language API");
+      expect(description).toContain("Vertex AI");
+    });
+  });
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376` — added the `when they read the credential field` describe so it matches the structure used everywhere else in these files.

langwatch/src/components/settings/ModelProviderForm.tsx
```diff
@@ -320,6 +320,22 @@ export const EditModelProviderForm = ({
...
+    () => JSON.stringify(state.customKeys),
+    [state.customKeys],
+  );
+  const [refusedCredentials, setRefusedCredentials] = useState<string | null>(
+    null,
+  );
+  // Editing any credential re-arms the probe, so a corrected key is checked
+  // again rather than saved on the strength of the previous refusal.
+  const canSaveWithoutProbe = refusedCredentials === credentialsFingerprint;
+
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`. Agreed on the reasoning: the content-based fingerprint makes it low-risk in practice, but the file already established the reset-on-`providerId` pattern for exactly this class of stale state, and an inconsistency there is how the next person gets caught. Added a `useEffect` clearing `refusedCredentials` on `providerId`, next to the `advancedDraft` one.

specs/model-providers/credential-validation.feature
```diff
@@ -211,3 +211,85 @@ Feature: Credential Validation
     When I call validateProviderApiKey for it
     Then validation is skipped instead of probing a relative URL
     And no misleading network-connection error is shown
+
+  # A rejected key is a dead end for the customer: the drawer refuses to save
+  # until the probe passes. So the message has to name the real cause, or the
+  # customer regenerates a working key over and over and gets nowhere.
+
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376` — the escape hatch made that comment false and I left it stale. Now reads: a refused key holds the save back on the first attempt so the message has to name the real cause, and saving anyway is available once the reason has been read.

specs/model-providers/credential-validation.feature
```diff
@@ -211,3 +211,85 @@ Feature: Credential Validation
...
+    When I click "Save"
+    Then the provider is not saved
+    And the button offers to save anyway
+
+  @integration
+  Scenario: Saving anyway does not probe the provider again
+    Given the provider refused the API key I entered
+    When I click "Save anyway"
+    Then the provider is saved
+    And the provider is not probed a second time
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`. Rewrote all three to state outcomes instead of probe counts: "Saving anyway keeps the credential I entered" / "I am not interrupted by the refusal again", "the corrected key is accepted on its own merits", and "the key is valid without asking the remaining shapes". The call-count assertions stay in the integration tests where they belong.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +55,329 @@ function buildModelsEndpointUrl(
...
+      failures.push({
+        valid: false,
+        error: context.hasConfigurableEndpoint
+          ? "Failed to validate API key. Please check your network connection and base URL."
+          : "Failed to validate API key. Please check your network connection.",
+      });
+    }
   }
+
+  return mostInformativeFailure(failures);
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `54f279376`, and thanks for re-raising it against the chain — the 4x framing is what made the shape of the fix obvious. Two changes: every candidate now carries a 10s `AbortSignal.timeout`, and the walk breaks as soon as the provider positively calls the key invalid, since asking the remaining shapes cannot change that answer. Measured against live Google: a real key and a plainly invalid one now both settle in **1** request, not 4.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +55,329 @@ function buildModelsEndpointUrl(
...
-    };
+    case "bearer":
+    default:
+      return [
+        {
+          url,
+          headers: { ...headers, Authorization: `Bearer ${apiKey}` },
+        },
+      ];
   }
```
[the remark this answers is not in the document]
Aryansharma28: Acknowledged, leaving as-is. The `switch` is a total match over a four-member union, so a fifth strategy fails to compile rather than silently falling through — which is the property a lookup table would give up. Happy to revisit when that fifth one actually lands.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: This is the one finding I have deliberately not actioned, because I do not think I should guess at the intended shape.

`validateProviderApiKey` currently returns `ValidationResult` (`{ valid, error }`) as **data**, and the drawer renders `result.error` inline under the field. `HandledError` is an abstract `Error` designed to be thrown and unwrapped by the error-handler middleware. So "return a HandledError instead" resolves two ways, and they are not equivalent:

1. **Throw it.** A refused key becomes a rejected tRPC call. The hook's existing `catch` would surface `error.message`, so the UI still works — but it makes an expected outcome ("this key does not work") an exception, and it rewrites the contract plus all ~50 unit tests that assert on the returned shape.
2. **Carry it.** Keep returning data, but replace the bare `error: string` with a serialized `HandledError` so the refusal gains `code`, `fault`, `tips` and `docsUrl`. That fits this feature unusually well — `SERVICE_DISABLED` wants a `docsUrl` to the Google Cloud console and `tips` for "enable the API / use Vertex AI" — but it changes the frontend contract too.

I lean towards (2): a refused credential is a legitimate result rather than a failure, and the structured metadata is exactly what this feature has been hand-rolling into prose. But that is your call on your codebase, and this PR is already large — I would rather do it deliberately in a follow-up than guess here.

Which shape did you mean? Happy to implement either.

Aryansharma28: Done in `91749fdd2` — you were right, and pinning it to this line made the scope clear.

The distinction I was missing: a **refused key is an answer**, so it stays a `ValidationResult`. A provider that never replied is the **absence** of one, and that is what was being hand-assembled here into a sentence claiming a verdict on the key it never actually got.

So the unreachable path now raises:

```ts
export class ProviderUnreachableError extends HandledError {
  // code: "provider_unreachable", fault: "provider", httpStatus: 502,
  // meta: { provider }, tips adapt to whether the customer can set a base URL
}
```

It travels the same channel as everything else — `errorFormatter` in `trpc.ts` already serializes any `HandledError` cause, so it lands in the same envelope as the `validation_error` the reporting customer hit. The drawer is unchanged: the hook's existing `catch` turns a rejected call into the same inline message.

Timeouts feed straight into it too, since an aborted request lands in the same `catch`.

Left the refusal path returning data rather than throwing, since "this key does not work" is a legitimate result rather than a failure — shout if you want that converted as well and I will do it in a follow-up.

Aryansharma28:
## Ready for another look — everything raised is addressed

Every finding from @coderabbitai and @0xdeafcafe has been actioned, each with a reply on its own thread. Summary of what changed since the review, so this does not need re-reading end to end:

**@0xdeafcafe — `HandledError`.** Pinning the comment to the network-failure `catch` made the scope clear, and the distinction it forced is the right one: a **refused key is an answer**, so it stays a `ValidationResult`; a **provider that never replied is the absence of one**, and that is what was being hand-assembled into a sentence claiming a verdict on a key we never got one for. Unreachable now raises `ProviderUnreachableError` (`fault: "provider"`, `httpStatus: 502`, tips that adapt to whether a base URL is even settable). Verified on the wire — the call logs `handledErrorCode: "provider_unreachable"`, so it lands in the same envelope as the `validation_error` the reporting customer hit. The drawer is unchanged, since the hook's existing `catch` renders it identically.

**Bounded and short-circuited.** Every probe now carries a 10s `AbortSignal.timeout`, and the walk stops as soon as the provider positively calls the key invalid. Measured against live Google: a real key and a plainly invalid one both settle in **1** request, down from 4.

**Named parameters** on the probe builder — the two adjacent strings were the real hazard, since a swapped `baseUrl`/`defaultBaseUrl` compiled clean.

**Copy corrected.** "add this as a Vertex AI provider" implied the rejected key was reusable. It is not: `aiplatform.googleapis.com` answers `401 "API keys are not supported by this API"`. Now says "configure a separate Vertex AI provider with service-account credentials."

**Specs** state user-observable outcomes instead of probe counts, and the stale "blocks saving until the probe passes" comment is corrected for the escape hatch.

**Tests.** The ElevenLabs case asserted on a message identical to our own fallback, so it passed whether or not the provider's message was ever read. Also drained the one-shot `fetch` queue between tests — `clearAllMocks()` leaves unconsumed `mockResolvedValueOnce` entries queued, which short-circuiting exposed by leaking leftovers into the next test.

### Two bugs found by running it, not by mocks

Worth flagging because the mocked suite passed both:

1. A plainly invalid key returned `"Invalid API key. Please check your API key and try again. Please pass a valid API key"`. The OpenAI-compatible surface answers with a vaguer sentence and **no** `reason`, and the selector preferred it over Google's canonical verdict. Refusals are now ranked.
2. Raising on unreachable broke `modelProvider.noProject.integration.test.ts`, which points a probe at an unreachable host on purpose and read the resulting `{valid:false}` as proof the authz gate opened. CI caught it; the assertion now follows the same evidence, since reaching a network failure at all is what proves the gate opened.

### Deliberately not changed

The refusal path still returns data rather than throwing — "this key does not work" is a legitimate result, not a failure. Happy to convert it in a follow-up if you would rather it went through `HandledError` too.

Not merging until someone approves.

sergioestebance:
## Ruthless review — verdict

**The approach is right. It is not ready to merge: one P1.**

The transport fix (`.query()` → `.mutation()`) is correct and unambiguously the most valuable thing here — a credential JSON-encoded into a GET URL is a real leak and a real 422, and moving the audit-log redaction in the same commit rather than a follow-up is exactly the right instinct. The multi-shape probe, the ranked refusals, and pinning the parser to a 403 Google actually sent (two `reason` fields, the useful one in `details[]`) all hold up.

What does not hold up is the `HandledError` conversion. It regressed the message it was meant to improve, and the branch's own tests assert the server-side sentence that never crosses the wire.

Traced: the probe chain, the rank/short-circuit logic, the query→mutation change and every caller of it, `redactAuditArgs` against the audit middleware chain, the escape-hatch state machine, and both surfaces that run the probe. Both P1/P2 candidates went through attribution against `main` and an adversarial refutation pass with an executed decisive test.

---

### P1 — an unreachable provider now shows the customer the literal string `provider_unreachable`

`langwatch/src/server/api/routers/providerValidation.ts:104-128` (the error), `:500-505` (the throw)
`langwatch/src/hooks/useModelProviderApiKeyValidation.ts:63-69` (and the same at `:111-118`)

`ProviderUnreachableError` is a `HandledError`, and this repo's tRPC `errorFormatter` **deliberately** replaces a handled error's free-text message with its stable code on the wire (`trpc.ts:231-235` — "Free-text error messages never cross the tRPC boundary"; ADR-045). The hook catches and reads raw `error.message`. So the sentence written in the constructor never arrives, and the slug renders unconditionally in `<Field.ErrorText>` (`ModelProviderCredentialsSection.tsx:149-151`).

Executed, driving the real error → real `errorFormatterForTesting` → real `TRPCClientError.from` → real hook:

```
[1] domainError.message = "Could not reach the provider to check this API key."
[1] serialize() = {"code":"provider_unreachable", ... "meta":{"provider":"gemini"},
                   "httpStatus":502,"fault":"provider",
                   "tips":["Check your network connection.","Check the base URL is correct and reachable."]}
[2] mapped tRPC code   = INTERNAL_SERVER_ERROR
[4] TRPCClientError.message = "provider_unreachable"
[5] validationError rendered in <Field.ErrorText> = "provider_unreachable"
```

**Attribution — introduced, not inherited.** `git show main:…/providerValidation.ts` lines 110/149/187/229 return `{valid:false, error:"Failed to validate API key. Please check your network connection[ and base URL]."}`. The branch strictly loses copy: prose → slug, and the `tips` array reaches the browser intact with nothing rendering it.

Why it slipped: `modelProvider.noProject.integration.test.ts:671` asserts `rejects.toThrow(/Could not reach the provider/)` — true server-side, invisible to the customer. And `useModelProviderApiKeyValidation.unit.test.ts:98-108` rejects with `new Error("validation_error")` and asserts `validationError` *contains* `"validation_error"`, i.e. it encodes "a code-shaped slug is acceptable customer copy" as the expectation. That test passes today and would keep passing after the regression.

This lands on first-time setup and key rotation — precisely when a customer is least equipped to decode a slug. The escape hatch mitigates the blast radius (they can still "Save anyway") but the explanation the PR promises — "the first refusal explains itself" — is the thing that broke.

**Minimal fix (two lines of intent, no special-casing):**
1. Add `message: "Could not reach the provider to check this API key."` to the error's `meta`. `meta.message` is the documented prose channel (`utils/trpcError.ts:6-37`) and is already what `errorDisplayMessage` reads first.
2. Have both `catch` blocks in the hook call `errorDisplayMessage(error)` instead of `error instanceof Error ? error.message : …`. That fixes `validateWithCustomUrl` at the same time and stops any future handled error leaking a code into this drawer.

---

### P2 — the escape hatch ships on one of the three surfaces that run the probe

`ModelProviderForm.tsx:323-350`, `:605` (the fix) vs `ModelProviderSetup.tsx:346-359` (unchanged)

`ModelProviderSetup` — rendered as `variant="onboarding"` in the create-product flow and `variant="langy"` in the "Langy needs a model to get started" gate — still hard-blocks with three bare `if (!isValid) return;`. Executed against the real component with a refused key:

```
[PROBE] after 1st save -> validate: 1 | submit mutation: 0 | label: Save
[PROBE] after 2nd save -> validate: 2 | submit mutation: 0 | label: Save
[PROBE] after 3rd save -> validate: 3 | submit mutation: 0 | label: Save
[CONTROL] validate: 1 | submit mutation: 2 | label: Save   (probe accepts -> submit reached)
```

Never a "Save anyway", never a submit. The onboarding step has a "Skip for now" ghost button, so the customer can abandon the step; the Langy gate has no skip at all. It's also *stricter* than the drawer: the drawer probes only when a new key was typed, this one probes on every save via `validateWithCustomUrl()`, so a refused stored key blocks unrelated edits there too.

Worth naming because the report quotes "doesn't allow me to **continue**" and the onboarding handler is literally `handleSaveAndContinue`, and because the spec scenarios ("When I click 'Save' … the button offers to save anyway", `specs/model-providers/credential-validation.feature`) are written surface-agnostically but bound only by the drawer test.

**Capped at P2 — the hard block is pre-existing and this diff does not regress it.** Non-blocking, but it means the PR's actual unblock may not reach the surface the report points at.

Fix direction: the refusal-override is form state, not drawer state. Lift `credentialsFingerprint` / `refusedCredentials` / `canSaveWithoutProbe` into `useModelProviderApiKeyValidation` (or a thin gate hook beside it) returning `{ probeRequired, recordRefusal, saveLabel }`; both callers consume it and the existing scenarios bind on both. ~15 lines moved.

---

### P2 — `502` falls off `handledErrorToTRPCCode`, so a customer's bad base URL is alerted as our 500

`langwatch/src/server/api/trpc.ts:585-596`

The map has 400/401/403/404/409/422/429; `502` hits the `?? "INTERNAL_SERVER_ERROR"` default. An error explicitly stamped `fault: "provider"` therefore ships as HTTP 500 and is logged, counted and alerted as a LangWatch server fault every time someone typos an endpoint. Add `502: "BAD_GATEWAY"`.

### P2 — the audit redaction stops one field short

`langwatch/src/server/api/trpc.ts:redactAuditArgs`

`customKeys` is redacted; `extraHeaders` rides the *same* `modelProvider.update` mutation and is exactly where an `Authorization: Bearer …` gets typed, and `providerConfig` is a passthrough object. Both still land verbatim in a durable, queryable table — the precise thing the docstring argues against. Pre-existing (base logged everything), so **non-blocking**, but this is the PR that establishes the seam and it's one line to close.

### P3 — the description overstates the walk

> "refuses only once **every** shape has refused"

`runProbeChain` breaks early on `FAILURE_RANK.definitive`. That's a defensible call — but the PR's own evidence table is that Google answers differently per surface, so a key that reads `API_KEY_INVALID` on `?key=` never gets asked on `/v1beta/openai/models`. Either soften the sentence or say why `API_KEY_INVALID` is trusted as final.

### P3 — Gemini's URL layout is reverse-engineered inside the validator

`buildProbeCandidates` strips `/v1|/v1beta` off the registry's `defaultBaseUrl` with a regex to recover a root the registry never states. The auth shapes are provider knowledge and belong beside `defaultBaseUrl` in the registry; adding a second shape for any provider currently means editing a switch in the validation module. (Harmless today — `gemini` has `endpointKey: undefined`, so `baseUrl` is always empty and the regex only ever sees the constant.)

### P3 — `tips` are serialized to the browser and dropped

`ProviderUnreachableError` ships two of them and nothing renders them. Fold them into the fix for the P1 rather than leaving payload that no surface reads.

---

**Blocked on the P1 only.** Everything else is a nit or explicitly inherited. Nothing was pushed and nothing was edited — review only.

Aryansharma28:
Thanks — the P1 was right, and the trace made it cheap to confirm. All of it addressed across three commits (`4d78b43`, `8fab7ea`, `3f3e720`), with one correction on the 502.

---

### P1 — `provider_unreachable` reaching the customer ✅ fixed

Reproduced exactly as you traced it, then fixed as you suggested:

1. `meta.message` now carries the sentence, with the `tips` folded into it — which closes the last P3 too, so nothing is serialized that no surface reads. `tips` stays structured for the API/CLI/MCP consumers that read it as data.
2. Both catch blocks in the hook go through `errorDisplayMessage`, so `validateWithCustomUrl` is fixed at the same time.

The regression test drives the real error through the real `errorFormatterForTesting`, the real `TRPCClientError.from` and the real hook, then reads what `<Field.ErrorText>` would render. You were right that the old tests encoded the defect as the expectation — `useModelProviderApiKeyValidation.unit.test.ts` rejected with `new Error("validation_error")` and asserted the slug was acceptable copy. That assertion is gone.

Reverting just the hook line fails it with your output:

```
Expected: "Could not reach the provider to check this API key. Check your network
           connection. Check the base URL is correct and reachable."
Received: "provider_unreachable"
```

### P2 — escape hatch on one surface only ✅ fixed, your way

Lifted into `useCredentialProbeGate` returning `{ probeRequired, recordRefusal, clearRefusal, saveLabel }`; both callers consume it. The drawer's four existing tests pass unchanged, and `ModelProviderSetup` has a companion suite so the surface-agnostic scenarios bind on both — `credential-validation.feature` is 24/24.

Its three probe branches also collapsed to one: `validateWithCustomUrl(undefined)` already meant "the default URL", so the last two arms were the same call written twice.

I left the probe-on-every-save behaviour alone. It is pre-existing, narrowing it is a bigger change than this review asked for, and the escape hatch means it can no longer strand anyone. Called out under "What this does not do".

### P2 — 502 falling off the map ⚠️ real, but not via `BAD_GATEWAY`

The finding holds; the suggested fix does not compile. We are on `@trpc/server` 10.45.4, whose code table has no `BAD_GATEWAY` (502/503/504 arrived in v11) — `tsgo` rejects it against `TRPC_ERROR_CODE_KEY`.

Two parts of your premise are also already handled, worth recording so the next reader does not re-fix them: `handleTrpcCallLogging` skips `captureException` for any handled cause, and it already logs by `fault` rather than status, stamping `handledErrorFault` for filtering.

What was genuinely wrong is the number: `logData.statusCode` came from `getHTTPStatusCodeFromError`, so an error stamped 502 was recorded as a 500 — your "counted against our error budget", exactly. The log line now prefers the handled error's own `httpStatus`. That also corrects `PlatformBoom` (503, previously logged as 500); its existing assertion is updated rather than worked around.

`data.error.httpStatus` was already 502 on the wire for clients. The envelope status stays 500 until a v11 upgrade — noted in a comment at the map.

### P2 — audit redaction one field short ✅ fixed

`extraHeaders` (name kept, value dropped — it is a `{key, value}` list, not an object) and `providerConfig` both redacted.

### P3 — description overstates the walk ✅ fixed by explaining, not softening

`API_KEY_INVALID` is trusted as final because it is Google positively identifying the credential as wrong rather than declining that particular surface; every other refusal, including the OpenAI-compatible surface's vaguer wording, keeps walking. Written out in the description with the reasoning rather than hedged.

### P3 — Gemini's URL layout reverse-engineered ✅ fixed

`apiRoot` now sits beside `defaultBaseUrl` in the registry and the regex is gone. Your parenthetical is what made it worth doing rather than not: it was harmless only because `gemini` has no `endpointKey`. It was also silently wrong for a customer-supplied base URL, whose layout we cannot assume — that case now probes the documented shape alone instead of inventing three sibling paths under someone else's host.

I did not move the whole auth-shape list into the registry. Adding a shape still means editing the switch; happy to do it if you want it in this PR rather than when a second provider needs it.

### P3 — `tips` dropped ✅ folded into the P1 fix, as you suggested

---

**Verification:** `pnpm typecheck:all` clean; 434 unit tests across `src/server/api`, 145 integration across 23 files (136 the untouched baseline), `feature-parity` green at 3085 bound scenarios. Still no browser QA.

langwatch/src/components/settings/__tests__/ModelProviderForm.validation-escape-hatch.integration.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+        const user = userEvent.setup();
+
+        await enterKey(user, "sk-accepted-by-the-provider");
+        await user.click(saveButton());
+
+        await waitFor(() => {
+          expect(mockMutateAsync).toHaveBeenCalledTimes(1);
+        });
+        expect(saveButton()).toHaveTextContent(/^save$/i);
+      });
```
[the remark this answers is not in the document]
Aryansharma28: Not reproducing — I think this one is a false positive.

The annotation is present, and twice over:

- `ModelProviderForm.validation-escape-hatch.integration.test.tsx:171`
- `ModelProviderSetup.validation-escape-hatch.integration.test.tsx:145` (the onboarding surface, added since this comment)

And the repo's own checker — the one `langwatch-app-ci` gates on — agrees:

```
▸ specs/model-providers/credential-validation.feature
  24/24 scenarios bound
  ✓ all bound
```

`feature-parity` is green on the current head, so nothing is unbound here. Leaving as-is.

Aryansharma28:
Follow-up filed as #6328 — Google Agent Platform keys (`AQ.…`) still cannot be validated or used, which is the "What this does not do" section of this PR turned into tracked work. Assigned to @sergioestebance. Not in scope here.

langwatch/biome.jsonc
```diff
@@ -293,7 +293,20 @@
 						"useTestHooksOnTop": "error"
 					},
 					"complexity": {
-						"noExcessiveNestedTestSuites": "error"
+						"noExcessiveNestedTestSuites": "error",
```
Aryansharma28, line 296: [P2] This silences 2,371 lint diagnostics repo-wide, from inside a Gemini key-validation PR

**Why:** I added this, and I still think it is right on the merits — but the blast radius deserves an explicit decision rather than riding along in an unrelated PR. Measured on this branch: `noExcessiveLinesPerFunction` across `src` + `ee` goes from **3,618 to 1,247**. That is 2,371 diagnostics turned off, none of which this PR wrote.

I classified all 2,371 rather than assuming:

| What the anchor line actually is | Count | Share |
|---|---|---|
| `describe` / `it` / `test` callback | 2,266 | 95% |
| `beforeAll` and `describeIfStripeKey` hooks | 100 | 4% |
| Genuine helper functions in test-support files | ~5 | <1% |

So 99% is test scaffolding, and the rule genuinely measures nothing on a suite declaration — CLAUDE.md mandates the nested `describe("given …") / describe("when …")` structure that makes the outer callback span the file. The `<1%` tail is real though: `globalSetup.ts:248`, `clickhouse-fixtures.ts:51` and `monitor-performance.fixtures.ts:112` are ordinary long functions that now go unchecked.

**Evidence:** `pnpm exec biome check ./src ./ee` with this block reverted vs. as-committed.

**Suggested fix:** Not a code change — a scope call, and it is the author's to make:
1. **Leave it.** Defensible, commented, and the CI failure message explicitly invites this fix.
2. **Split it out** into its own one-file PR so the policy change gets reviewed as a policy change, and #6252 stays purely about Gemini.

Worth noting either way: any other open PR touching this same override block will conflict with whichever lands second.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +59,495 @@ function buildModelsEndpointUrl(
...
+      const failure = await handleHttpError({ response, context });
+      failures.push(failure);
+
+      // The provider has positively identified the key as wrong. Asking the
+      // remaining shapes cannot change that answer, and each one is another
+      // outbound request on this request thread.
+      if (failure.rank === FAILURE_RANK.definitive) {
+        break;
+      }
+    } catch {
```
Aryansharma28: [P3] This `catch` also swallows a throw from `handleHttpError`, turning a response we *did* get into "we never reached the provider"

**Why:** `handleHttpError({ response, context })` is called at line 527, inside this same `try`. The `catch` cannot tell a network failure from a bug in the refusal parser, so both become `unreachable()` — and once every candidate lands there, `runProbeChain` throws `ProviderUnreachableError`.

That is a misdiagnosis of exactly the kind this PR exists to remove. The customer is told to check their network connection and base URL for a host that answered fine; the actual defect is in our parsing, and it is invisible because the `catch` is bare.

Low severity because `readUpstreamRefusal` is written not to throw (both `response.text()` and `JSON.parse` are individually guarded), so today this is a latent trap rather than a live bug. It becomes live the moment anyone adds a provider-specific branch to `handleHttpError`.

**Evidence:**
```ts
const failure = await handleHttpError({ response, context });   // ← inside the try
failures.push(failure);
...
} catch {
  failures.push(unreachable());   // ← claims we never reached the provider
}
```

**Suggested fix:** Narrow the `catch` to the failure it actually means. Either move the `handleHttpError` call out of the `try`, or gate on the error being a fetch/abort failure and rethrow anything else so a parser bug surfaces as a bug instead of as a network story.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +59,495 @@ function buildModelsEndpointUrl(
...
-  baseUrl: string,
-  defaultBaseUrl: string,
-): Promise<ValidationResult> {
+const PROBE_BUDGET_MS = 10_000;
+
+/** The request that proves a key works: list the provider's models. */
+type ProbeRequest = { url: string; headers: Record<string, string> };
+
+/**
+ * Builds the models request for an auth strategy. Every provider is probed
```
Aryansharma28: [P3] Orphaned docblock — two stacked docblocks, and the first describes a function that no longer exists

**Why:** `buildProbeCandidates` carries two consecutive `/** … */` blocks. The first is a leftover from before the function returned a list:

> "Builds the models **request** for an auth strategy. Every provider is probed the same way; only where the credential rides differs."

Both halves are now wrong. The function returns `ProbeRequest[]`, not a request; and "every provider is probed the same way" is the precise belief this PR was written to disprove — Gemini gets four shapes, everyone else gets one. The second docblock immediately below says the correct thing at length.

CLAUDE.md is explicit that a comment describing behavior the code does not implement should be deleted or made true. This one contradicts both the code and the docblock beneath it, which is the worst case for a reader skimming for intent.

**Suggested fix:** Delete the first docblock. The second already covers it.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +59,495 @@ function buildModelsEndpointUrl(
...
+  unreachable: 4,
+} as const;
+
+type RankedFailure = ValidationResult & { rank: number };
+
+/** Picks the refusal worth showing, keeping the first of equally useful ones. */
+function mostInformativeFailure(failures: RankedFailure[]): RankedFailure {
+  return failures.reduce<RankedFailure>(
+    (chosen, failure) => (failure.rank < chosen.rank ? failure : chosen),
+    failures[0] ?? {
```
Aryansharma28: [P4] The empty-`failures` fallback returns "Invalid API key" — the exact wrong answer this PR removes

**Why:** If `failures` is empty, `reduce` falls back to a synthesized `{ valid: false, error: INVALID_KEY_MESSAGE, rank: generic }` — telling the customer their key is invalid when in fact nothing was ever asked.

Not reachable today: every branch of `buildProbeCandidates` returns at least one candidate, and the loop pushes a failure on every non-`ok` path. So this is a latent trap, not a bug — flagging it only because "no probe ran" silently becoming "your key is bad" is the precise failure mode this PR exists to kill, and the fallback re-creates it one empty array away.

**Suggested fix:** Make it unrepresentable rather than defaulted — `rank: FAILURE_RANK.unreachable` in the fallback would route an empty walk to `ProviderUnreachableError`, which is the honest answer for "we never asked".

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +59,495 @@ function buildModelsEndpointUrl(
...
-async function validateWithAnthropicAuth(
-  apiKey: string,
-  baseUrl: string,
-  defaultBaseUrl: string,
-): Promise<ValidationResult> {
-  const url = buildModelsEndpointUrl(baseUrl, defaultBaseUrl);
+function redactApiKey(text: string, apiKey: string): string {
+  if (apiKey.length < 8) return text;
 
+  const encoded = encodeURIComponent(apiKey);
```
Aryansharma28, line 329: [P5 PRAISE] Redacting the percent-encoded form too is the detail that actually mattered

**Why:** Genuine praise, because this is the one that would have been missed. Gemini is the provider that carries the credential in the query string, and Google echoes the offending URL back in its error body — so the key arrives already `encodeURIComponent`-ed. A redactor matching only the raw key would have passed every hand-written fixture and still leaked the key to a customer-facing error for the single provider this PR is named after.

```ts
const encoded = encodeURIComponent(apiKey);
const forms = encoded === apiKey ? [apiKey] : [apiKey, encoded];
```

The `encoded === apiKey` guard also avoids a pointless second pass for keys with no reserved characters, which is most of them. Worth keeping as the pattern for any future provider that authenticates via query string.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -55,180 +59,495 @@ function buildModelsEndpointUrl(
...
+      fault: "provider",
+      httpStatus: 502,
+      // A handled error's own message never crosses the tRPC boundary — the
+      // formatter replaces it with the code (ADR-045) — so the words a
+      // customer reads come from this code's entry in the client presentation
+      // registry, not from here. `hasConfigurableEndpoint` is in `meta`
+      // because that entry branches on it: only some providers have a base URL
+      // there is any point telling someone to check. `tips` stays structured
+      // for the API, CLI and MCP consumers that read it as data.
+      meta: { provider, hasConfigurableEndpoint },
```
Aryansharma28: [P5 OBSERVATION] @sergioestebance's P1 is still closed — but not by the code that closed it

**Why:** Flagging this purely so a re-reviewer is not confused by a mechanism that changed after approval.

The P1 was that the customer read the literal slug `provider_unreachable`. The fix at review time folded the sentence and tips into `meta.message`, which `errorDisplayMessage` read first. **`main` then deleted `errorDisplayMessage`** in #6010, which moved the platform onto a code-keyed presentation registry.

So `meta` now carries `hasConfigurableEndpoint` instead of a pre-built sentence, and the words come from this code's entry in `src/features/errors/logic/presentation.ts`. The registry is exhaustive over `APP_ERROR_CODES`, so an unregistered code is a typecheck failure and a `codes.unit.test.ts` failure — the slug regression cannot silently return.

Verified rather than assumed: the hook's unit test builds the real `ProviderUnreachableError`, runs it through the real `errorFormatter` and the real `TRPCClientError.from`, and asserts the drawer renders *"Couldn't reach the provider. Nothing answered, so this API key was not checked. Check your network connection, and check the base URL is correct and reachable."* — plus an explicit `not.toContain("provider_unreachable")`.

One nice consequence worth knowing: `resolveErrorCopy`'s `supplementalTips` drops tips already contained in the description, so the two remaining `tips` do not double-render next to that sentence in `HandledErrorAlert`. Nothing to do here.

Aryansharma28 (commented):
## goated-review — PR #6252

**Verdict: COMMENT. No P0, no P1. The PR is sound; the one thing worth a decision is scope, not correctness.**

Reviewed after the merge with `main` and the lint pass, so this covers the merge commits as well as the original change. Full disclosure: I made those merge commits, so I deliberately went hardest at my own edits — the P2 below is against my own change, not the author's.

| P0 | P1 | P2 | P3 | P4 | P5 |
|---|---|---|---|---|---|
| 0 | 0 | 1 | 2 | 1 | 2 |

### P2 MEDIUM (1)
- **`biome.jsonc` turns off `noExcessiveLinesPerFunction` for every test file repo-wide — 2,371 diagnostics.** Justified on the merits, but it is a repo-wide lint-policy decision living inside a Gemini key-validation PR.

### P3 LOW (2)
- **`runProbeChain`'s bare `catch` also swallows a throw from `handleHttpError`**, reporting "we never reached the provider" for a response we did reach.
- **Orphaned docblock** left on `buildProbeCandidates` by an earlier refactor — two stacked docblocks, the first describing a function that no longer exists.

### P4 NIT (1)
- `mostInformativeFailure`'s empty-array fallback returns `INVALID_KEY_MESSAGE`, which is the exact wrong diagnosis this PR exists to remove. Not reachable today.

### P5 PRAISE / OBSERVATION (2)
- `redactApiKey` covers the percent-encoded form — the easy miss, and the one that mattered here.
- The mechanism behind @sergioestebance's P1 was replaced during the merge. The P1 is still closed, but not by the code he approved.

### Not re-raised (already covered)
- @sergioestebance's P1 on `provider_unreachable` reaching the customer — fixed, and still fixed; see the P5 below for the changed mechanism.
- @0xdeafcafe on `HandledError` scope — addressed.
- CodeRabbit on the unbounded/sequential probe requests and the missing `AbortSignal` — fixed via the shared `PROBE_BUDGET_MS` deadline.
- CodeRabbit on adjacent same-type positional params — addressed by the object-param refactor.

### Things I could not verify from the diff alone
- **No browser QA.** The PR says so itself. The escape-hatch gate (`useCredentialProbeGate`) drives real save-button behavior on three surfaces (drawer, onboarding, the Langy model gate) and is covered only by jsdom component tests. The "Save anyway" path is the one I would want to see a human click.
- **`SERVICE_DISABLED` remains fixture-only.** Reproducing it needs a key on a project with the API disabled. `API_KEY_SERVICE_BLOCKED` is captured from a real response; this one is not.
- Whether the four-shape Gemini walk holds up against Google's rate limits under real load — every local run probed a small number of keys.

langwatch/biome.jsonc
Aryansharma28: Resolving as **won't change** — deliberate call, recorded here so a reviewer does not have to re-derive it.

The override stays in this PR. It is commented in `biome.jsonc` with the reasoning beside it, and the CI failure message explicitly names a scoped override as the intended fix. 95% of what it silences is `describe` callbacks, which the rule cannot measure meaningfully, and CLAUDE.md mandates the nesting that makes them long.

Flagging the two things a reviewer might still want, so the choice is visible rather than buried: it is a repo-wide policy change riding in a Gemini PR, and any other open PR touching this same override block will conflict with whichever lands second.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: **Fixed** — `46a6296`, then refactored again in `e645a3b`.

Only the `fetch` is guarded now. Reading the refusal happens outside the `catch`, so a throw from `handleHttpError` surfaces as the bug it is instead of as a network story. The per-shape probe is its own function (`probeOnce`) because inlining both branches pushed `runProbeChain` over the cognitive-complexity limit.

The fix earned its keep immediately: it broke 9 tests, and they were right to break. Gemini walks up to four shapes, but those tests queued a single one-shot mock, so shapes 2–4 resolved `undefined` — and the old bare `catch` swallowed the resulting TypeError as "unreachable", which `mostInformativeFailure` then outranked with the real refusal from shape 1. The assertions passed without the walk ever being exercised. They use a persistent mock now, which is also the truthful simulation: a key the provider refuses is refused on every surface.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: **Fixed** — `46a6296`. Deleted. The docblock immediately below it was already the accurate one.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: **Fixed** — `e645a3b`, and taken further than suggested.

`mostInformativeFailure` returns `undefined` for an empty list rather than manufacturing a verdict, and `runProbeChain` turns that into `ProviderUnreachableError`. "Nothing was asked" and "nothing answered" now reach the same honest answer, and there is no longer a fallback that can express "invalid API key" without a provider having said so.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: Still true, and the ground under it moved in `e645a3b` — worth recording which guarantee is now load-bearing.

The upstream sentence no longer reaches the customer at all. It is read, used to rank the refusal, logged, and dropped; `HandledError.serialize()` omits `message`, so nothing in the payload the browser receives can carry it. That is a stronger property than redaction, which by construction only masks the shapes someone enumerated.

So `redactApiKey` now protects the **log line** rather than the response — still worth having (a key in our own logs is still a key in our logs), just no longer the thing standing between the customer and their own credential.

That draft also caught a real mistake: I first put the provider's sentence in the handled error's `message`, reasoning that a message is server-side. It is not — the REST boundary sends `{ error: code, message }` verbatim (`error-handler.ts`), which is exactly why the convention says a handled error's message must be written customer-safe.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: @sergioestebance's P1 is still closed, and the mechanism has now moved a **second** time — flagging so a re-reviewer is not hunting for code that no longer exists.

1. At your review: the sentence and tips were folded into `meta.message`, read by `errorDisplayMessage`.
2. After merging `main`: #6010 deleted `errorDisplayMessage`, so the code was registered in the exhaustive presentation registry and the hook reads `describeError`.
3. In `e645a3b`: every *other* refusal became a coded handled error too, so the same registry now answers for all six rather than just this one.

The regression cannot silently return at any step, because the registry is exhaustive over `APP_ERROR_CODES` — a code with no copy fails `pnpm typecheck`, and a code nothing raises fails `codes.unit.test.ts`. The hook test still drives the real error through the real formatter and asserts `not.toContain("provider_unreachable")`.

sergioestebance (approved):

PR #6256, opened 2026-07-27
fix(slack): surface partial channel listing & correct groups:read hint (Fixes #6254) [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #6254: Slack automations: channel picker silently returns a partial channel list, and the setup steps contradict the app's own scopes
## Summary

Setting up Slack delivery on an automation is confusing in a specific, reproducible way: the channel picker lists plenty of channels but **not** the one the author deliberately invited the bot to, while test messages succeed to channels the bot was never invited to. Nothing in the UI explains either behaviour, so the flow reads as broken even though each piece is doing what it was written to do.

Reported from internal use. Three distinct defects sit behind it, plus one contributing factor in a user's own Slack app.

## Root cause 1 — private channels are dropped silently

`listSlackChannels` (`langwatch/src/server/app-layer/automations/delivery/slackWebApi.ts:239-249`) asks Slack for `public_channel,private_channel`. Slack refuses the **whole** request with `missing_scope` when the app lacks `groups:read`, so the code retries public-only:

```ts
const withPrivate = await listChannelsForTypes(token, "public_channel,private_channel");
if (withPrivate.error !== "missing_scope") return withPrivate;
// Missing `groups:read` (private) — fall back to public channels only.
return listChannelsForTypes(token, "public_channel");
```

The degradation itself is deliberate and good — one missing scope shouldn't take the public channels down with it. **The problem is that the retry succeeds and returns `error: null`.** The return contract is `{ channels, error }`, which has no way to say "this worked, but it is partial". The caller cannot tell a complete listing from a public-only one.

This is pinned as intended behaviour, so it will not show up as a test regression (`slackWebApi.unit.test.ts:266-291`):

```ts
it("falls back to public channels when only groups:read is missing", …)
expect(result.error).toBeNull();
```

Consequence in the UI: `client.tsx:393-403` only renders a hint when an error comes back. With `error: null` the author sees a confident list of every public channel and **no indication whatsoever that private channels were omitted**. If the channel they invited the bot to is private, it is simply absent, with nothing to explain why.

## Root cause 2 — the hint names the wrong scope

The one hint that does exist for this case (`client.tsx:399-400`) reads:

> Add the channels:read scope to your app and reinstall it to pick from a list — you can still type the channel above.

`channels:read` governs **public** channels. The scope missing in the private-channel case is `groups:read`. This string only renders when `channels:read` is missing too, so it is never technically wrong on screen — but it is the only scope name the UI ever shows, which sends anyone debugging a missing private channel after the wrong permission.

## Root cause 3 — the setup steps contradict what the app can do

The manifest we hand out (`client.tsx:212-220`) grants `chat:write.public`:

```yaml
bot:
  - chat:write
  - chat:write.public
  - channels:read
  - groups:read
```

`chat:write.public` means "post to any public channel without being invited" — so **test messages to public channels are supposed to succeed without an invite**. That is not a bug.

But setup step 3 (`client.tsx:931-935`) says only:

> Invite the bot to the channel you post to.

That instruction is unnecessary for public channels and **mandatory** for private ones, and it distinguishes neither. An author who follows it literally does the one thing that doesn't help for public channels, then watches messages land in channels they never invited the bot to. The two observations look contradictory, and the copy is what makes them look that way.

## Root cause 4 — the 2000-channel cap degrades silently in the same way

`listChannelsForTypes` stops after `MAX_CHANNEL_PAGES` (10) × `CHANNEL_PAGE_SIZE` (200) and returns (`slackWebApi.ts:221-225`):

```ts
logger.warn({ pages: MAX_CHANNEL_PAGES, channels: collected.length },
  "Slack conversations.list page cap reached; returning a partial channel list");
return done(null);   // error: null — the truncation never reaches the UI
```

Same class as root cause 1: the server knows the list is partial, says so in a server-side log, and returns a shape that cannot carry that fact to the author. A workspace with more than 2000 channels silently misses some.

## Contributing factor (not a code defect)

`groups:read` and `chat:write.public` were both added to the manifest on 2026-07-19 in #5911. **A Slack app created from an earlier manifest has no `groups:read`**, which lands exactly in root cause 1. Reinstalling from the current "Copy app manifest" should make private channels appear — worth confirming with anyone hitting this before any fix ships, since it may resolve the immediate report on its own.

## Steps to reproduce (root cause 1)

1. Create a Slack app **without** `groups:read` (or use one created before #5911).
2. Invite it to a **private** channel.
3. Open an automation, choose Slack delivery, paste the bot token.
4. The picker lists public channels. The private channel is absent, and no hint is shown.

## What "fixed" should mean

- The author can tell when the channel list is partial and why — whether the cause is a missing scope or the page cap. This needs the `{ channels, error }` contract to carry partial-success, not just success/failure; the unit test asserting `error: null` on the fallback encodes the current limitation and would change with it.
- Scope names in user-facing copy match the thing actually missing.
- The setup steps say when an invite is required (private channels) and when it isn't (public, covered by `chat:write.public`).

## Notes

Separate from #6233 (channel picker search), which is merged — that fixed typing in the box, not what the box is populated with. Related to #6244 (the view drawer mislabels bot automations as webhooks). Together these three suggest the Slack delivery flow has not been driven end-to-end against a real workspace recently.

waterWang:
## Summary

The Slack channel picker (`langwatch/src/server/app-layer/automations/delivery/slackWebApi.ts`) had a silent-degradation class: when the app lacked `groups:read`, `listSlackChannels` fell back to public channels only and returned `error: null` — so the UI showed a confident list while the private channels the author actually invited the bot to were absent, with no indication. The same shape hid a second degradation: the hard 2000-channel page cap in large workspaces returned `error: null` as well.

Three defects from #6254 (plus a fourth) are now surfaced to the author:

1. **Private channels dropped silently (root cause 1)** — `listSlackChannels` now sets `partial: true` when it falls back to public-only after a missing `groups:read`. The client reads `list.data?.partial` and renders a non-blocking hint ("it may be missing private channels (install the app with the `groups:read` scope to include them)").
2. **Page cap degraded silently (root cause 4)** — the 2000-channel cap in `listChannelsForTypes` now marks the listing `partial: true` as well, so large workspaces see the same hint.
3. **Hint names the wrong scope (root cause 2)** — the `missing_scope` fallback hint now says `groups:read` (the scope needed for private channels), not `channels:read`.
4. **Setup step 3 was misleading (root cause 3)** — "Invite the bot to the channel you post to" is now clarified: invite is required for private channels, unnecessary for public ones (the app has `chat:write.public`).

## Changes

- **`slackWebApi.ts`** — added `ListChannelsResult { channels, error, partial }`; `listChannelsForTypes` carries `partial` through `done(null, true)` at the page cap; `listSlackChannels` spreads `partial: true` on the groups:read fallback.
- **`client.tsx`** — derived `isPartial = !!list.data?.partial && !hardError`; renders a partial-list `<Text>` alongside (not instead of) the hard-failure hint; rewrote the `missing_scope` hint copy to name `groups:read`; clarified setup step 3.
- **`slackWebApi.unit.test.ts`** — added `expect(result.partial)` assertions to the page-cap, groups:read-fallback, missing-scope-only, and request-failed tests; added a new "complete listing" test asserting `partial === false`; extended `channelPage` helper with an optional `public` flag.

Fixes langwatch/langwatch#6254

## Testing

- Three files changed; TypeScript is checked locally against the repo's tsconfig before push.
- Unit tests updated to reflect the new `partial` field (6 assertions added/adjusted + 1 new test).

## Summary by CodeRabbit

- **New Features**
  - Slack channel selection now indicates when results may be incomplete, such as when private channels are unavailable or Slack’s listing limit is reached.
  - Updated bot setup guidance with clearer instructions for inviting the bot and using public channels.

- **Bug Fixes**
  - Improved handling and messaging for partial Slack channel listings versus complete request failures.

### DIFF
--- a/langwatch/src/server/app-layer/automations/delivery/slackWebApi.ts
+++ b/langwatch/src/server/app-layer/automations/delivery/slackWebApi.ts
@@ -139,6 +139,19 @@ interface SlackConversationsResponse {
 }
 
 /**
+ * The contract `listSlackChannels` returns to the client. In addition to the
+ * channels and a hard error, it carries `partial` so the UI can flag that the
+ * listing is a best-effort subset rather than every channel the token can see
+ * (e.g. private channels were omitted because `groups:read` is missing, or the
+ * page cap was hit on a very large workspace).
+ */
+interface ListChannelsResult {
+  channels: SlackChannel[];
+  error: string | null;
+  partial: boolean;
+}
+
+/**
  * Conversations per page. Slack's own guidance is to stay well under its 1000
  * ceiling — large pages routinely time out server-side — and a smaller page
  * keeps each response comfortably inside {@link CHANNEL_LIST_MAX_RESPONSE_BYTES}.
@@ -163,11 +176,12 @@ const CHANNEL_LIST_MAX_RESPONSE_BYTES = 1024 * 1024;
 async function listChannelsForTypes(
   token: string,
   types: string,
-): Promise<{ channels: SlackChannel[]; error: string | null }> {
+): Promise<{ channels: SlackChannel[]; error: string | null; partial: boolean }> {
   const collected: SlackChannel[] = [];
-  const done = (error: string | null) => ({
+  const done = (error: string | null, partial: boolean = false) => ({
     channels: [...collected].sort((a, b) => a.name.localeCompare(b.name)),
     error,
+    partial,
   });
 
   let cursor: string | undefined;
@@ -222,7 +236,7 @@ async function listChannelsForTypes(
     { pages: MAX_CHANNEL_PAGES, channels: collected.length },
     "Slack conversations.list page cap reached; returning a partial channel list",
   );
-  return done(null);
+  return done(null, true);
 }
 
 /**
@@ -238,12 +252,15 @@ async function listChannelsForTypes(
  */
 export async function listSlackChannels(
   token: string,
-): Promise<{ channels: SlackChannel[]; error: string | null }> {
+): Promise<ListChannelsResult> {
   const withPrivate = await listChannelsForTypes(
     token,
     "public_channel,private_channel",
   );
   if (withPrivate.error !== "missing_scope") return withPrivate;
   // Missing `groups:read` (private) — fall back to public channels only.
-  return listChannelsForTypes(token, "public_channel");
+  // The resulting list is incomplete by construction; mark it partial so the
+  // client can surface a hint that private channels were omitted.
+  const result = await listChannelsForTypes(token, "public_channel");
+  return { ...result, partial: true };
 }
--- a/langwatch/src/features/automations/providers/slack/client.tsx
+++ b/langwatch/src/features/automations/providers/slack/client.tsx
@@ -394,10 +394,16 @@ function SlackChannelField({
     list.data?.error && list.data.error !== "no_token"
       ? list.data.error
       : null;
+  // The API can report that the listing is partial (private channels were
+  // omitted because `groups:read` is missing, or the 2000-channel cap was hit).
+  // That is informational, not an error — render it separately so it never
+  // swallows a hard failure hint.
+  const isPartial = !!list.data?.partial && returnedError === null;
   const hint = list.isError
     ? `Couldn't load channels: ${list.error?.message ?? "request failed"}. You can still type the channel above.`
     : returnedError === "missing_scope"
-      ? "Add the channels:read scope to your app and reinstall it to pick from a list — you can still type the channel above."
+      // Missing `groups:read` means only public channels were listed.
+      ? "Add the groups:read scope to your app and reinstall it to see private channels as well — you can still type a private channel above."
       : returnedError
         ? "Couldn't load channels from Slack. Check the token, or type the channel above."
         : null;
@@ -493,6 +499,13 @@ function SlackChannelField({
           pt={1}
         >
           {hint}
+        </Text>
+      ) : null}
+      {isPartial && !hint ? (
+        <Text textStyle="xs" color="fg.muted" pt={1}>
+          Showing only the channels this app can read. It may be missing private
+          channels (install the Slack app with the `groups:read` scope to include
+          them) or a few channels if the workspace is very large.
         </Text>
       ) : null}
     </Field.Root>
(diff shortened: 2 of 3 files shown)


langwatch/src/server/app-layer/automations/delivery/slackWebApi.ts
```diff
@@ -238,12 +252,15 @@ async function listChannelsForTypes(
...
     token,
     "public_channel,private_channel",
   );
   if (withPrivate.error !== "missing_scope") return withPrivate;
   // Missing `groups:read` (private) — fall back to public channels only.
-  return listChannelsForTypes(token, "public_channel");
+  // The resulting list is incomplete by construction; mark it partial so the
+  // client can surface a hint that private channels were omitted.
+  const result = await listChannelsForTypes(token, "public_channel");
+  return { ...result, partial: true };
```
langwatch-agent, line 265: **P2 - A failed public-only fallback is reported as a partial success.** This unconditional override turns *every* result of the retry into `partial: true`, including `{ error: "missing_scope" }`, `{ error: "request_failed" }`, and `{ error: "bad_response" }` with no usable channels. That contradicts the new test at `slackWebApi.unit.test.ts:291-293`, which expects `partial === false` when both attempts receive `missing_scope`; the existing malformed-body test also still expects an object without `partial`. Those tests will fail, and callers other than the current UI cannot distinguish an incomplete successful list from a failed retry. Preserve `partial: true` only when `result.error === null` (or return `result` unchanged on error), then update/add the retry-failure cases accordingly.

langwatch-agent (changes requested):
Static review of this external fork found one P2 contract/test regression; details are inline. No contributor code or scripts were executed.

langwatch-agent (commented):
Static review of external contribution at `3c8f5facfabd5c6bb8f70e326400be641fb6ffee` (no PR-branch code or dependencies executed). No merge-blocking issue found. The page-cap condition is surfaced distinctly as partial discovery, and the missing `groups:read` fallback is correctly marked partial rather than presenting the public-channel subset as complete. The revised scope hint matches the API operation being attempted. CI evidence was considered but not rerun under the external-contribution trust boundary.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=3c8f5facfabd5c6bb8f70e326400be641fb6ffee p0=0 p1=0 p2=0 p3=0

PR #6271, opened 2026-07-28
refactor(invite): move the invite procedures into their own router (#2877)

### LINKED ISSUE #2877: Extract invite procedures from organization router into dedicated invite router
## Context

Follow-up from #2872. The organization router still contains 6 invite-related procedures that delegate to `InviteService` and use `ctx.prisma` directly:

- `createInvites`
- `deleteInvite`
- `getOrganizationPendingInvites`
- `createInviteRequest`
- `approveInvite`
- `acceptInvite`

These were intentionally left in place during the org router refactoring because they already delegate to a proper service (`InviteService`). However, they belong in their own router for separation of concerns.

## Proposed approach

1. Create `inviteRouter` in `src/server/api/routers/invite.ts`
2. Move the 6 invite procedures into it
3. Merge the invite router into the app router
4. Organization router drops from ~1434 to ~850 lines

## Acceptance criteria

- Invite procedures live in their own router
- Organization router contains only org-specific procedures
- Existing integration tests pass unchanged
- No API contract changes

langwatch-agent:
## Ticket

Closes #2877. The 6 invite procedures (`createInvites`, `deleteInvite`, `getOrganizationPendingInvites`, `createInviteRequest`, `approveInvite`, `acceptInvite`) sat in the middle of `organization.ts`, whose other 10 procedures are organization CRUD. Premise verified on current `main`: all 6 still there, no invite router exists.

They form one **contiguous** block (lines 506-1219), which is what makes this a clean cut rather than six surgical extractions.

## The one real decision, and why

The ticket says *"merge the invite router into the app router"* with an AC of **"No API contract changes"**. Those pull in opposite directions unless you pick deliberately:

| Approach | Paths | Caller churn |
| --- | --- | --- |
| Mount at its own `invite:` key | `organization.createInvites` becomes `invite.createInvites` | **~52 call sites** across the frontend, plus test edits |
| `mergeRouters` under `organization:` (**this PR**) | unchanged | **zero** |

I took the second: `root.ts` does `organization: mergeRouters(organizationRouter, inviteRouter)`, so the file-level split happens with no path movement. That is the only reading under which "no API contract changes" holds.

**If you actually want a real `invite.*` namespace, say so and I will do it as a follow-up** — the work is those ~52 callers, and it is a deliberate contract change worth its own PR, not something to smuggle into a move.

`mergeRouters` is standard tRPC (verified present in the installed 10.45.4) but has **no precedent in this repo**, so it is the thing most worth your eye. I exported it from `trpc.ts` with a note on when to prefer it over mounting.

## Change

`organization.ts`: **1535 → 787 lines.** Plus three supporting edits:

- **`teamRoleInputSchema`** was a file-local const used by procedures on *both* sides of the split (1 use in the moved block, 2 outside). It moves to a new `teamRole.schemas.ts` that both routers import. Its two building blocks stay private there, since only the union is used at call sites. The ticket did not anticipate this shared helper.
- `organization.ts` loses the **8 imports** only the invite procedures used.
- `trpc.ts` exports `mergeRouters`.

## Evidence

| Check | Result |
| --- | --- |
| **The move is a move** | the 714 moved lines are **byte-for-byte identical** to `origin/main` (diff of the extracted region against the original) |
| **No API contract change** | enumerated the **live** router at runtime (`appRouter._def.procedures`): the set of `organization.*` paths is **identical**, all 16, before and after |
| Procedure accounting | 10 remain + 6 moved = 16 baseline |
| Router unit tests | **372 passed** (49 files) |
| Invite-specific tests | **35 passed** |
| `pnpm typecheck:all` | exit 0 (both `typecheck` and `typecheck:tests`, since a test file changed) |

The runtime path enumeration is the proof that matters here: it checks the actual contract rather than my reading of it.

## One test changed, and why it was unavoidable

`organization.acceptInvite.unit.test.ts` builds its caller from `organizationRouter` **directly**, so moving `acceptInvite` out of that router breaks the harness no matter which mounting approach is used. This is worth flagging because it means the ticket's AC *"existing integration tests pass unchanged"* is **not satisfiable by any extraction** — I found this by running the tests, not by reading them.

The fix is 2 lines: it builds its caller from `inviteRouter`. I kept the filename and the `"organization.acceptInvite"` describe, because the path it exercises is still `organization.acceptInvite`, and its sibling `organization.invites.integration.test.ts` keeps that prefix too.

## Not verified locally

`organization.invites.integration.test.ts` (the AC's named proof) **cannot run in this environment**: it fails on `PrismaClientInitializationError` because the Postgres it wants on `:55432` is down, reporting `22 skipped`. I confirmed this is environmental rather than mine by running it on a pristine `origin/main` worktree, where it fails **identically**. CI is the check for that one.

Also note the moved code carries 13 pre-existing em-dash comment lines. I left them verbatim rather than rewording, so the block stays a provable move; happy to scrub them in a separate pass.

## Advisory note

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive organization invitation management, including creating, listing, approving, accepting, and deleting invitations.
  - Added support for team assignments and custom roles when inviting members.
  - Added validation for invite status, expiration, email matching, permissions, and license limits.
  - Added consistent handling for pending invite requests and acceptance workflows.

- **Bug Fixes**
  - Improved invite error handling and validation for duplicate or invalid invitations.


langwatch/src/server/api/routers/invite.ts
```diff
@@ -0,0 +1,753 @@
...
+      if (!organization) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Organization not found",
+        });
+      }
+
+      const inviteService = InviteService.create(prisma);
+
+      // Check license limits using the service
```
langwatch-agent, line 106: [P1] Preserve the current-base existing-member guard here. `main` now calls `inviteService.assertNotAlreadyMembers({ emails, organizationId })` immediately after constructing the service and before the licence check. This branch omits it, despite claiming the moved procedure is byte-for-byte unchanged. After resolving the current merge conflict, an admin can again submit `createInvites` for an address that is already an organization member; the flow writes a pending invite beside the membership (and can report a seat-limit error instead of the real conflict). Restore the guard in the extracted router and add/retain a router-level regression test for this call path.

langwatch-agent (commented):
Blocking regression found. The router composition preserves the `organization.*` namespace and the shared role schema is appropriately scoped, but this branch is behind `main` and currently conflicts. Rebase it, retaining the current-base invite safeguards before it can merge.

langwatch/src/server/api/routers/invite.ts
langwatch-agent: Valid, and fixed. The branch was cut before `assertNotAlreadyMembers` landed on main, so the omission was real rather than a claim about the moved lines being wrong at the time.

Rebased onto current main and carried the guard into `invite.ts`. The extracted block is byte-for-byte identical to `origin/main` `organization.ts` lines 505-1228 again:

```
diff <(sed -n 39,762p langwatch/src/server/api/routers/invite.ts) \
     <(git show origin/main:langwatch/src/server/api/routers/organization.ts | sed -n 505,1228p)
# no output
```

Added the router-level regression cover you asked for, in `organization.createInvites.unit.test.ts`: the real guard over a mocked prisma, through the real router, asserting (a) the refusal carries `already_organization_member` with the address in `meta`, (b) `checkLicenseLimits` is never reached when the address is already a member, so the ordering that makes the message honest is pinned, and (c) it *is* reached when the address is new. Mutation-tested: removing the guard call takes it to 2 failed / 1 passed.

The live path set is still unchanged, re-proved after the rebase from `Object.keys(appRouter._def.procedures)`: 16 `organization.*` paths, identical to main.

339 router unit tests pass, `pnpm typecheck:all` clean.

langwatch-agent:
Closing this rather than rebasing it across the ADR-076 move.

The rebase is not the problem, it is that this PR is the wrong shape for what it turned out to be. Three things:

- It dropped `inviteService.assertNotAlreadyMembers(...)`, which `main` calls in `createInvites` before the licence check, while describing the move as unchanged. A verbatim-move PR with no test that fails when the move is not verbatim will keep doing that, and the invite surface on `main` is actively changing.
- The new `invite.ts` is 753 source lines against the repo's "FAIL when a NEW file exceeds 300 source lines" guideline, with two near-duplicate ~200-line resolvers. It cannot land as-is, and splitting it turns a move into a redesign.
- It carries a PII read with no audit-log write into what counts as a new procedure.

The full writeup, and a suggested shape for a second attempt (extract the shared pipeline into `InviteService` first, add router-level regression tests against the current base, then move the thinned procedures), is on the ticket: langwatch/langwatch#2877. That ticket stays open.

Advisory PR from the tech-debt-fixer bot. Reopen if you would rather have it rebased and worked through in place.

PR #6273, opened 2026-07-28
feat(scenarios): export scenario run history as CSV

Aryansharma28:
## Start here — a worked example

**[WORKED-EXAMPLE.md](https://github.com/langwatch/langwatch/blob/assets/scenario-run-csv-export/exports/WORKED-EXAMPLE.md)** walks one real run end to end: the conversation, the checklist the judge scored, how that becomes rows, and the ranked fix-list it produces. Every number in it is copied from the files below.

## Two modes

A scenario run is nested — it contains a conversation *and* a checklist of judged criteria — and CSV is flat, so one layout cannot serve both.

| Mode | One row is | Rows | Answers | Open | Download |
|---|---|---:|---|---|---|
| **Full** | a message | 1556 | *what actually happened* | [sample](https://github.com/langwatch/langwatch/blob/assets/scenario-run-csv-export/exports/samples/sample-scenario-runs-full.csv) | [csv](https://raw.githubusercontent.com/langwatch/langwatch/assets/scenario-run-csv-export/exports/scenario-runs-full.csv) |
| **Criteria** | a run × checklist item | 281 | *what is the agent bad at* | [sample](https://github.com/langwatch/langwatch/blob/assets/scenario-run-csv-export/exports/samples/sample-scenario-runs-criteria.csv) | [csv](https://raw.githubusercontent.com/langwatch/langwatch/assets/scenario-run-csv-export/exports/scenario-runs-criteria.csv) |

Criteria mode is the one that changes what you do next. Group by `criterion`, filter `met = false`, sort descending, and you get a ranked fix-list in four clicks:

```
fails  across   rule
   18      18   Langy is terse (1-3 bullets, no filler openers)
   16      16   Langy does NOT narrate or echo the CLI command it ran
    8       8   Langy does NOT offer 'next actions' or 'would you like me to…'
```

Those 18 failures span **18 different scenarios across 8 different task types** — the same rule breaking everywhere, which no per-scenario view can show you.

**There is no per-run mode.** Full already denormalizes every run field onto every message row, so de-duplicating on `run_scenario_run_id` gives exactly that. A third column contract for data already present was not worth it — and gzip removed the only argument for it (see below).

**Criteria covers 87 of 229 runs.** Not a gap: 136 runs errored before the judge scored anything, so they have no checklist to expand. Of runs that completed, 85 of 88 are judged.

## Screenshots

Live against the local haven stack at 1512×945.

Runs page with Export CSV button

Export dialog

Exported CSV contents

## Compression

The response is `Content-Encoding: gzip` over the existing stream — the browser inflates it and still writes a plain `.csv`, so nothing about the file the user opens changes.

| Mode | On the wire | Decompressed | Ratio |
|---|---:|---:|---:|
| full | 616 KB | 5658 KB | **9.2x** |
| criteria | 39 KB | 281 KB | 7.1x |

Measured, not estimated. This is also what made the per-run mode unnecessary: gzipped full is smaller than the per-run file was uncompressed.

## How it streams

```
RunHistoryPanel  --POST-->  /api/export/scenario-runs/download
                                     |
                            ScenarioRunExportService (AsyncGenerator)
                                     |
                       loop: fetch 100 runs -> serialize -> yield chunk
                                     |
                            response streams to disk, memory stays flat
```

A new keyset-paginated repository read sweeps `simulation_runs` oldest-first on `(StartedAt, ScenarioRunId)`. That tuple is unique across all 228 local runs (`count = uniqExact = 228`), so no page boundary can drop or duplicate a run. It is deliberately separate from `getRunDataForAllSuites`, which caps at 100 and can short-circuit with `changed: false` for the UI's freshness protocol.

## Decisions worth reviewing

- **Both `status` and `status_category`.** Status is resolved through the same mapper the run history UI uses, so a run shown as stalled exports as stalled — the file cannot disagree with the screen. The category is what you pivot on; the status is what you need when one run looks wrong.
- **No aggregates.** `run-history-transforms.ts` carries a `KEEP IN SYNC` warning that the pass-rate formula already lives in two places. A third copy here would drift from the number on screen, so the file carries per-run categories and the spreadsheet does the arithmetic.
- **`categorizeRunStatus` moved** out of the client-only `run-history-transforms.ts` into `server/scenarios/scenario-run-category.ts`, so the UI and the exporter share one definition instead of the exporter growing its own.
- **Criteria as JSON arrays, not comma-joined** — they are free-text sentences that routinely contain commas. `met_criteria_count` / `unmet_criteria_count` sit alongside so the common question needs no parsing.
- **ISO-8601 UTC timestamps**, diverging from the trace exporter's epoch milliseconds, so spreadsheets parse them as dates.
- **CRLF line endings, and every chunk terminates with one.** Concatenated batches would otherwise glue the last row of one onto the first row of the next.
- **Export queries read through a `t` alias with table-qualified columns.** `RUN_COLUMNS` aliases `UpdatedAt` to a String and ClickHouse resolves `WHERE` against SELECT aliases, so an unqualified dedup tuple compares String to DateTime64 and silently matches nothing — a 200 response with an empty file.
- **Pass/fail filtering happens after mapping, not in SQL**, because `STALLED` is derived from timestamps by `resolveRunStatus` rather than stored.

## Column completeness

Audited by exporting all 228 runs and measuring the percentage of rows with a non-empty value, per column. 23 columns, 228 rows, **0 malformed rows**.

Three columns export empty on this dataset, and the audit explains why rather than leaving it a mystery:

| Column | Filled | Why |
|---|---|---|
| `target_type` | 0% | comes from the `langwatch` metadata namespace, which only platform **suite** runs populate — 12 of 228 runs here carry the namespace, 1 carries `targetType`. SDK- and CLI-driven runs simply do not set it. Extraction verified correct; the data is genuinely absent. |
| `simulation_suite_id` | 0% | same namespace |
| `target_reference_id` | 1% | same namespace |

Everything identifying or scoring a run is at 99–100%: `scenario_run_id`, `scenario_set_id`, `status`, `status_category`, `met_criteria_count`, `unmet_criteria_count`, `started_at`, `duration_ms`, `message_count` at 100%; `scenario_id`, `scenario_name`, `scenario_description`, `batch_run_id`, `verdict`, `reasoning` at 99%.

The audit also caught two gaps, fixed in the second commit:
- **`scenario_description` was missing entirely** despite being set on 225 of 228 runs — it is the field saying what the scenario tests.
- **Trace ids came only from per-message `trace_id`**, ignoring the run-level `TraceIds` column. Both are now unioned. On this dataset the run-level column proved to be a strict subset and adds nothing today; the union is kept because the two columns are written by independent code paths and a run can record traces without emitting a message snapshot. The exported total of 742 matches the ClickHouse union exactly.

## Verification

Driven end-to-end against the local haven stack with 228 real runs:

| Mode | Rows | Cross-check |
|---|---|---|
| summary | 228 | = 228 runs |
| criteria | 281 | = 181 met + 100 unmet (matches ClickHouse) |
| full | 1550 | 5.2 MB |

228 runs at 100/batch means three batches and two boundaries; all 228 rows parse cleanly, confirming the CRLF handling. Grouping the exported criteria file reproduces the ClickHouse ranking exactly.

- 23 serializer unit tests pass
- 237 existing `components/suites` tests pass (no regression from the `categorizeRunStatus` extraction)
- `pnpm typecheck` adds zero errors (105 before and after; the pre-existing 105 are an unrelated zod v3/v4 mismatch in the langy pipeline)

## Spec

`specs/scenarios/scenario-run-export.feature` — 34 scenarios written before the code, per the repo's spec-first rule.

## Out of scope — separate bug found

The **existing trace CSV export** has a batch-boundary corruption bug. `serializeTracesToSummaryCsv` returns `Parse.unparse(...)` with no trailing newline and `export.service.ts` concatenates chunks directly, so every trace CSV export over 100 traces glues one row per boundary:

```
"id,name\r\n1,alpha\r\n2,beta3,gamma\r\n4,delta"    4 rows in -> 3 rows out
```

Reproduced but deliberately not fixed here to keep this PR scoped. Worth its own issue.

https://claude.ai/code/session_01718LCHKmrCbMJRiLheEbb2

## Summary by CodeRabbit

* **New Features**
  * Added **Export CSV** for scenario runs in run history filters, with an export dialog offering **Full** and **Criteria** modes.
  * Exports honor the current scenario selections and date range, and are only enabled when matching runs exist.
  * Downloads are delivered as **gzip-compressed CSV** with a deterministic filename, export metadata headers, and real-time progress with **cancel** support.
* **Bug Fixes**
  * Improved consistency of exported run status and status-category bucketing with the UI, including correct handling of archived and unauthorized data.
* **Tests**
  * Added unit tests for CSV serialization/streaming plus end-to-end export feature coverage.

### DIFF
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -22,6 +22,10 @@ import { getApp } from "~/server/app-layer/app";
 import { getServerAuthSession } from "~/server/auth";
 import { auditLog } from "~/server/auditLog";
 import { prisma } from "~/server/db";
+import {
+  ScenarioRunExportForbiddenError,
+  ScenarioRunExportUnauthenticatedError,
+} from "~/server/export/scenario-runs/errors";
 import { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
 import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
 import type { NextRequest } from "~/types/next-stubs";
@@ -42,10 +46,7 @@ secured
         req: c.req.raw as NextRequest,
       });
       if (!session) {
-        return c.json(
-          { error: "You must be logged in to access this endpoint." },
-          { status: 401 },
-        );
+        throw new ScenarioRunExportUnauthenticatedError();
       }
 
       const hasPermission = await hasProjectPermission(
@@ -54,10 +55,7 @@ secured
         "scenarios:view",
       );
       if (!hasPermission) {
-        return c.json(
-          { error: "You do not have permission to access this endpoint." },
-          { status: 403 },
-        );
+        throw new ScenarioRunExportForbiddenError(request.projectId);
       }
 
       logger.info(
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -121,6 +121,7 @@ secured
             for await (const { chunk, progress } of service.exportRuns({
               request,
               signal: c.req.raw.signal,
+              total: totalCount,
             })) {
               controller.enqueue(encoder.encode(chunk));
               void broadcast.broadcastToTenant(
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -1,0 +1,136 @@
+/**
+ * Hono app for scenario run export endpoints.
+ *
+ * POST /download — Streams scenario run history as CSV.
+ *
+ * This is the API layer: HTTP concerns only (auth, headers, streaming). All
+ * domain logic lives in ScenarioRunExportService.
+ *
+ * Progress is broadcast via BroadcastService (Redis pub/sub) so a tRPC
+ * subscription on any pod can relay it to the client, exactly as trace export
+ * does — the export id travels back in the X-Export-Id response header.
+ *
+ * @see specs/scenarios/scenario-run-export.feature
+ */
+
+import { createLogger } from "@langwatch/observability";
+import crypto from "crypto";
+import { hasProjectPermission } from "~/server/api/rbac";
+import { createServiceApp, handlerManagedAuth } from "~/server/api/security";
+import { validator as zValidator } from "~/server/api/validation";
+import { getApp } from "~/server/app-layer/app";
+import { getServerAuthSession } from "~/server/auth";
+import { prisma } from "~/server/db";
+import { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
+import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
+import type { NextRequest } from "~/types/next-stubs";
+
+const logger = createLogger("langwatch:api:export-scenario-runs");
+
+const secured = createServiceApp({ basePath: "/api/export/scenario-runs" });
+
+secured
+  .access(handlerManagedAuth("user session + scenarios:view enforced in-handler"))
+  .post(
+    "/download",
+    zValidator("json", scenarioRunExportRequestSchema),
+    async (c) => {
+      const request = c.req.valid("json");
+
+      const session = await getServerAuthSession({
+        req: c.req.raw as NextRequest,
+      });
+      if (!session) {
+        return c.json(
+          { error: "You must be logged in to access this endpoint." },
+          { status: 401 },
+        );
+      }
+
+      const hasPermission = await hasProjectPermission(
+        { prisma, session },
+        request.projectId,
+        "scenarios:view",
+      );
+      if (!hasPermission) {
+        return c.json(
+          { error: "You do not have permission to access this endpoint." },
+          { status: 403 },
+        );
+      }
+
+      logger.info(
+        { projectId: request.projectId, mode: request.mode },
+        "Starting scenario run export download",
+      );
+
+      const exportId = crypto.randomUUID();
+      const broadcast = getApp().broadcast;
+
+      const today = new Date().toISOString().slice(0, 10);
+      const fileName = `${request.projectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
+
+      const service = new ScenarioRunExportService(
+        getApp().simulations.runs.repository,
+      );
+      const totalCount = await service.getTotalCount({ request });
+
+      const headers = new Headers({
+        "Content-Type": "text/csv; charset=utf-8",
+        "Content-Disposition": `attachment; filename="${fileName}"`,
+        "Transfer-Encoding": "chunked",
+        "X-Export-Id": exportId,
+        "X-Total-Runs": String(totalCount),
+        "Access-Control-Expose-Headers":
+          "X-Export-Id, X-Total-Runs, Content-Disposition",
+      });
+      const encoder = new TextEncoder();
+
+      const stream = new ReadableStream({
+        async start(controller) {
+          try {
+            for await (const { chunk, progress } of service.exportRuns({
+              request,
+            })) {
+              controller.enqueue(encoder.encode(chunk));
+              void broadcast.broadcastToTenant(
+                request.projectId,
+                JSON.stringify({
+                  exportId,
+                  type: "progress",
+                  exported: progress.exported,
+                  total: progress.total,
+                }),
+                "export_progress",
+              );
+            }
+            void broadcast.broadcastToTenant(
+              request.projectId,
+              JSON.stringify({ exportId, type: "done" }),
+              "export_progress",
+            );
+            controller.close();
+          } catch (error) {
+            logger.error(
+              { error, projectId: request.projectId },
+              "Scenario run export stream error",
+            );
+            void broadcast.broadcastToTenant(
+              request.projectId,
+              JSON.stringify({
+                exportId,
+                type: "error",
+                message: "Export failed",
+              }),
+              "export_progress",
+            );
+            controller.error(error);
+          }
+        },
+      });
+
+      return new Response(stream, { headers });
+    },
+  );
+
+export const app = secured.hono;
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -38,7 +38,13 @@ const logger = createLogger("langwatch:api:export-scenario-runs");
 const secured = createServiceApp({ basePath: "/api/export/scenario-runs" });
 
 secured
-  .access(handlerManagedAuth("user session + scenarios:view enforced in-handler"))
+  .access(
+    handlerManagedAuth({
+      reason: "user session validated in-handler via getServerAuthSession",
+      permissions: ["scenarios:view"],
+      credential: "session",
+    }),
+  )
   .post(
     "/download",
     zValidator("json", scenarioRunExportRequestSchema),
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -19,8 +19,8 @@ import { hasProjectPermission } from "~/server/api/rbac";
 import { createServiceApp, handlerManagedAuth } from "~/server/api/security";
 import { validator as zValidator } from "~/server/api/validation";
 import { getApp } from "~/server/app-layer/app";
-import { getServerAuthSession } from "~/server/auth";
 import { auditLog } from "~/server/auditLog";
+import { getServerAuthSession } from "~/server/auth";
 import { prisma } from "~/server/db";
 import {
   ScenarioRunExportForbiddenError,
@@ -28,8 +28,8 @@ import {
 } from "~/server/export/scenario-runs/errors";
 import type { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
 import {
-  scenarioRunExportRequestSchema,
   type ScenarioRunExportRequest,
+  scenarioRunExportRequestSchema,
 } from "~/server/export/scenario-runs/types";
 import { KSUID_RESOURCES } from "~/utils/constants";
 
@@ -125,10 +125,9 @@ secured
         broadcast,
       });
 
-      return new Response(
-        stream.pipeThrough(new CompressionStream("gzip")),
-        { headers },
-      );
+      return new Response(stream.pipeThrough(new CompressionStream("gzip")), {
+        headers,
+      });
     },
   );
 
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -149,7 +149,18 @@ function gzipped(
   source: ReadableStream<Uint8Array>,
 ): ReadableStream<Uint8Array> {
   const gzip = createGzip();
-  Readable.fromWeb(source as Parameters<typeof Readable.fromWeb>[0]).pipe(gzip);
+  const nodeSource = Readable.fromWeb(
+    source as Parameters<typeof Readable.fromWeb>[0],
+  );
+
+  // `.pipe()` does not forward a source error the way `pipeThrough` does: it
+  // unpipes and leaves the destination open, and the Readable's own 'error'
+  // event goes unhandled — which takes the process down rather than failing
+  // the one request. Destroying the gzip with the error propagates it to the
+  // response instead, so a failed query reads as a failed download.
+  nodeSource.on("error", (error) => gzip.destroy(error));
+  nodeSource.pipe(gzip);
+
   return Readable.toWeb(gzip) as ReadableStream<Uint8Array>;
 }
 
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -20,6 +20,7 @@ import { createServiceApp, handlerManagedAuth } from "~/server/api/security";
 import { validator as zValidator } from "~/server/api/validation";
 import { getApp } from "~/server/app-layer/app";
 import { getServerAuthSession } from "~/server/auth";
+import { auditLog } from "~/server/auditLog";
 import { prisma } from "~/server/db";
 import { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
 import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
@@ -64,11 +65,35 @@ secured
         "Starting scenario run export download",
       );
 
+      // A bulk export lifts a project's whole run history — full mode includes
+      // every conversation transcript — so the download has to be attributable
+      // to a user, not just permitted. Recorded before a byte is streamed.
+      await auditLog({
+        userId: session.user.id,
+        projectId: request.projectId,
+        action: "scenarioRuns.export",
+        targetKind: "project",
+        targetId: request.projectId,
+        args: {
+          mode: request.mode,
+          scenarioSetId: request.scenarioSetId,
+          scenarioId: request.scenarioId,
+          passFailStatus: request.passFailStatus,
+          startDate: request.startDate,
+          endDate: request.endDate,
+        },
+      });
+
       const exportId = crypto.randomUUID();
       const broadcast = getApp().broadcast;
 
       const today = new Date().toISOString().slice(0, 10);
-      const fileName = `${request.projectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
+      // Content-Disposition's filename is a quoted-string. projectId is only
+      // constrained to `z.string()`, so a quote in it would close the quote and
+      // let the caller append parameters. Server-generated ids never contain
+      // one today, but nothing in the code enforces that.
+      const safeProjectId = request.projectId.replace(/[^\w.-]/g, "_");
+      const fileName = `${safeProjectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
 
       const service = new ScenarioRunExportService(
         getApp().simulations.runs.repository,
@@ -83,7 +108,6 @@ secured
         "Content-Encoding": "gzip",
         Vary: "Accept-Encoding",
         "Content-Disposition": `attachment; filename="${fileName}"`,
-        "Transfer-Encoding": "chunked",
         "X-Export-Id": exportId,
         "X-Total-Runs": String(totalCount),
         "Access-Control-Expose-Headers":
@@ -96,6 +120,7 @@ secured
           try {
             for await (const { chunk, progress } of service.exportRuns({
               request,
+              signal: c.req.raw.signal,
             })) {
               controller.enqueue(encoder.encode(chunk));
               void broadcast.broadcastToTenant(
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -26,9 +26,7 @@ import {
   ScenarioRunExportForbiddenError,
   ScenarioRunExportUnauthenticatedError,
 } from "~/server/export/scenario-runs/errors";
-import { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
 import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
-import type { NextRequest } from "~/types/next-stubs";
 
 const logger = createLogger("langwatch:api:export-scenario-runs");
 
@@ -42,9 +40,7 @@ secured
     async (c) => {
       const request = c.req.valid("json");
 
-      const session = await getServerAuthSession({
-        req: c.req.raw as NextRequest,
-      });
+      const session = await getServerAuthSession({ req: c.req.raw });
       if (!session) {
         throw new ScenarioRunExportUnauthenticatedError();
       }
@@ -93,9 +89,7 @@ secured
       const safeProjectId = request.projectId.replace(/[^\w.-]/g, "_");
       const fileName = `${safeProjectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
 
-      const service = new ScenarioRunExportService(
-        getApp().simulations.runs.repository,
-      );
+      const service = getApp().simulations.export;
       const totalCount = await service.getTotalCount({ request });
 
       // CSV of repeated run-level values compresses ~9x, and the browser
--- a/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
+++ b/langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
@@ -13,8 +13,8 @@
  * @see specs/scenarios/scenario-run-export.feature
  */
 
+import { generate } from "@langwatch/ksuid";
 import { createLogger } from "@langwatch/observability";
-import crypto from "crypto";
 import { hasProjectPermission } from "~/server/api/rbac";
 import { createServiceApp, handlerManagedAuth } from "~/server/api/security";
 import { validator as zValidator } from "~/server/api/validation";
@@ -26,7 +26,12 @@ import {
   ScenarioRunExportForbiddenError,
   ScenarioRunExportUnauthenticatedError,
 } from "~/server/export/scenario-runs/errors";
-import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
+import type { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
+import {
+  scenarioRunExportRequestSchema,
+  type ScenarioRunExportRequest,
+} from "~/server/export/scenario-runs/types";
+import { KSUID_RESOURCES } from "~/utils/constants";
 
 const logger = createLogger("langwatch:api:export-scenario-runs");
 
@@ -78,7 +83,7 @@ secured
         },
       });
 
-      const exportId = crypto.randomUUID();
+      const exportId = generate(KSUID_RESOURCES.EXPORT).toString();
       const broadcast = getApp().broadcast;
 
       const today = new Date().toISOString().slice(0, 10);
@@ -105,51 +110,13 @@ secured
         "Access-Control-Expose-Headers":
           "X-Export-Id, X-Total-Runs, Content-Disposition",
       });
-      const encoder = new TextEncoder();
-
-      const stream = new ReadableStream({
-        async start(controller) {
-          try {
-            for await (const { chunk, progress } of service.exportRuns({
-              request,
-              signal: c.req.raw.signal,
-              total: totalCount,
-            })) {
-              controller.enqueue(encoder.encode(chunk));
-              void broadcast.broadcastToTenant(
-                request.projectId,
-                JSON.stringify({
-                  exportId,
-                  type: "progress",
-                  exported: progress.exported,
-                  total: progress.total,
-                }),
-                "export_progress",
-              );
-            }
-            void broadcast.broadcastToTenant(
-              request.projectId,
-              JSON.stringify({ exportId, type: "done" }),
-              "export_progress",
-            );
-            controller.close();
-          } catch (error) {
-            logger.error(
-              { error, projectId: request.projectId },
-              "Scenario run export stream error",
-            );
-            void broadcast.broadcastToTenant(
-              request.projectId,
-              JSON.stringify({
-                exportId,
-                type: "error",
-                message: "Export failed",
-              }),
-              "export_progress",
-            );
-            controller.error(error);
-          }
-        },
+      const stream = buildExportStream({
+        service,
+        request,
+        exportId,
+        totalCount,
+        signal: c.req.raw.signal,
+        broadcast,
       });
 
       return new Response(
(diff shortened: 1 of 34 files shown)


langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      const headers = new Headers({
+        "Content-Type": "text/csv; charset=utf-8",
+        "Content-Disposition": `attachment; filename="${fileName}"`,
+        "Transfer-Encoding": "chunked",
+        "X-Export-Id": exportId,
+        "X-Total-Runs": String(totalCount),
+        "Access-Control-Expose-Headers":
+          "X-Export-Id, X-Total-Runs, Content-Disposition",
+      });
```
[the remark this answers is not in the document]
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:120]: Fixed in 64add14d44, taking your suggested approach. `projectId` is sanitised with `replace(/[^\w.-]/g, "_")` before interpolation. Agreed it is hardening rather than a live exploit — ids are server-generated — but as you say nothing in the code enforced that.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,136 @@
...
+      const exportId = crypto.randomUUID();
+      const broadcast = getApp().broadcast;
+
+      const today = new Date().toISOString().slice(0, 10);
+      const fileName = `${request.projectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
+
+      const service = new ScenarioRunExportService(
+        getApp().simulations.runs.repository,
+      );
+      const totalCount = await service.getTotalCount({ request });
```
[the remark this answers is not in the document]
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:106]: Fixed in 64add14d44. Added an `auditLog` write with action `scenarioRuns.export` immediately after the permission check and before any streaming, recording the user, project, mode and the full filter set (scenarioSetId, scenarioId, passFailStatus, startDate, endDate). Verified end-to-end — an export now writes a row: `scenarioRuns.export | local-dev-project | project | criteria`.

Aryansharma28:
## Review pass — all 20 threads addressed

Two commits: `64add14d44` (first six findings) and `dd2846c685` (the critical + security pass).

### Fixed

| Finding | Fix |
|---|---|
| 🔴 **Cursor/sort key disagree on NULL `StartedAt`** | `EXPORT_SORT_KEY` is now one SQL expression used by `ORDER BY`, the cursor predicate *and* the returned cursor. The `StartedAt ?? CreatedAt` fallback in TS is gone — that was exactly the disagreement. |
| 🟠 **CSV formula injection** | Free-text cells (`scenario_name`, `scenario_description`, `reasoning`, `error`, `criterion`, `message_content`, and the criteria JSON) are prefixed with `'` when they open with `=` `+` `-` `@` TAB CR. Applied to text only — prefixing a negative number would corrupt it. **6 regression tests**, including that ordinary text and `-1` are untouched. On real data this neutralizes 121 cells, 114 of them agent replies opening with a markdown bullet. |
| 🟠 **No audit-log write** | `scenarioRuns.export` recorded with user, project, mode and the full filter set, after the permission check and before any streaming. Verified a row lands per export. |
| 🟠 **Unsanitized `projectId` in `Content-Disposition`** | Sanitized with your suggested `replace(/[^\w.-]/g, "_")`. |
| 🟠 **No cancellation path** | The request's `AbortSignal` is threaded into `exportRuns` and checked at the top of each page, so an aborted download stops paginating instead of sweeping to exhaustion. |
| 🟠 **Progress/cancel not rendered** | The panel no longer discards `isExporting`/`cancelExport`. The filter bar swaps the Export button for a live run count plus a Cancel action while streaming. |
| 🟠 **List query vs export end date** | `useRunHistoryPagination` now takes `endDateMs` and passes it to the query, so the list cannot show runs the CSV would omit. |
| 🟠 **Count query unmetered and run twice** | The route passes the count it already fetched into `exportRuns`, and the query declares `expectedMaxDurationMs` / `expectedMaxReadBytes`. |
| 🟠 **Export disabled from a paginated subset** | Gated on `isLoading` instead of `totals.runCount`, which only counts loaded pages. |
| 🟠 **`FILTER_TO_CATEGORY` typed `string`** | Typed `RunStatusCategory` — a renamed category is now a compile error, not a filter that silently matches nothing. |
| 🟡 **`Transfer-Encoding` set by hand** | Removed. It is a forbidden response header; the runtime owns framing and now also negotiates `Content-Encoding: gzip`. |
| 🟡 **Spec exposes internals** | Comments rewritten to describe behaviour, not mapper/query internals. |
| 🟡 **Fail-filter category** | The scenario now names the category it actually selects (`failure`). |
| 🟡 **Mode comment / test `when` conditions** | Corrected; nested suites use `when` per the repo's BDD convention. |

### Not fixed, with reason

**Buffering the whole CSV in the browser** (`useExportScenarioRuns.ts`) — correct observation, deliberately kept. Streaming straight to disk means a navigation/form POST, which gives up the progress indicator and cancel action that this same review (rightly) asked for. `fetch` + `blob` is what makes those possible. Mitigations already in place: the response is gzipped (9.2x measured), and the practical ceiling is a browser tab's memory rather than an unbounded server cost. If large-tenant exports become a real problem the fix is a server-side job with a download link, not a smaller client buffer — worth its own issue rather than an inline change here.

### Verification

- 29 serializer tests + 237 `components/suites` tests pass; `pnpm typecheck` at baseline (105 pre-existing, unrelated zod v3/v4 errors in the langy pipeline).
- Live against the dev stack: `200` + `content-encoding: gzip` + sanitized filename, an `AuditLog` row per export, **0 cells** in either mode still beginning with a formula character, and the sweep still returns 87 distinct judged runs / 229 total with **0 malformed rows**.

The failing `evaluate` check is unrelated — an OpenAI `insufficient_quota` billing error in the PR-summarizer workflow.

Aryansharma28:
## CI status

Rebased onto `main` and green on everything this PR can affect.

**Passing** — `typecheck`, `lint`, `build`, `test-unit` (4/4 shards), `test-integration` (6/6 shards), `CodeRabbit`, `migration-order`, `feature-parity`, `preflight`, and the rest.

**Three red checks, none caused by this PR:**

| Check | Cause |
|---|---|
| `evaluate` | OpenAI `insufficient_quota` — a billing error in the PR-summarizer workflow. No code involved. |
| `test` (e2e) | `litellm.InternalServerError: OpenAIException - internal_error` and `Error code: 500` running `python-sdk/examples/*.py` against live OpenAI. |
| `sdk-python-complete` | Same python-sdk suite. It also failed on **`main` itself** at `d6a5663a` (11:49 today) with nothing of mine involved. |

This PR touches **no Python files** — 16 files, all TypeScript under `langwatch/src/` plus one `.feature` spec:

```
git diff --name-only origin/main...HEAD | grep -iE '\.py$|python-sdk/'
  → (no matches)
```

### One real bug the rebase surfaced

`main` removed the `options` parameter from `queryRows` in #6114 (dropping the slow-query warning feature) after this branch was cut. Both export queries still passed it, so the *merge* failed `typecheck` with TS2554 even though each side compiled alone — and local `pnpm typecheck` could not see it. Fixed in `e205f5a436`; `typecheck` now passes.

That also moots the earlier review point about the count query being unmetered: per-query duration and read-byte expectations no longer exist in the repository.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+          "X-Export-Id, X-Total-Runs, Content-Disposition",
+      });
+      const encoder = new TextEncoder();
+
+      const stream = new ReadableStream({
+        async start(controller) {
+          try {
+            for await (const { chunk, progress } of service.exportRuns({
+              request,
+              signal: c.req.raw.signal,
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: should we sanitise this?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+      const today = new Date().toISOString().slice(0, 10);
+      // Content-Disposition's filename is a quoted-string. projectId is only
+      // constrained to `z.string()`, so a quote in it would close the quote and
+      // let the caller append parameters. Server-generated ids never contain
+      // one today, but nothing in the code enforces that.
+      const safeProjectId = request.projectId.replace(/[^\w.-]/g, "_");
+      const fileName = `${safeProjectId} - Scenario Runs - ${today} - ${request.mode}.csv`;
+
+      const service = new ScenarioRunExportService(
+        getApp().simulations.runs.repository,
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: maybe include export inside existing scenario/simulation app-layer, then use `getApp().simulations.runs.export` to get access to `ScenarioRunExportService`

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+        );
+      }
+
+      const hasPermission = await hasProjectPermission(
+        { prisma, session },
+        request.projectId,
+        "scenarios:view",
+      );
+      if (!hasPermission) {
+        return c.json(
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: handed error

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+    "/download",
+    zValidator("json", scenarioRunExportRequestSchema),
+    async (c) => {
+      const request = c.req.valid("json");
+
+      const session = await getServerAuthSession({
+        req: c.req.raw as NextRequest,
+      });
+      if (!session) {
+        return c.json(
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: handled error

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+secured
+  .access(handlerManagedAuth("user session + scenarios:view enforced in-handler"))
+  .post(
+    "/download",
+    zValidator("json", scenarioRunExportRequestSchema),
+    async (c) => {
+      const request = c.req.valid("json");
+
+      const session = await getServerAuthSession({
+        req: c.req.raw as NextRequest,
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: safe?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
```diff
@@ -0,0 +1,170 @@
...
+import { getServerAuthSession } from "~/server/auth";
+import { auditLog } from "~/server/auditLog";
+import { prisma } from "~/server/db";
+import { ScenarioRunExportService } from "~/server/export/scenario-runs/scenario-run-export.service";
+import { scenarioRunExportRequestSchema } from "~/server/export/scenario-runs/types";
+import type { NextRequest } from "~/types/next-stubs";
+
+const logger = createLogger("langwatch:api:export-scenario-runs");
+
+const secured = createServiceApp({ basePath: "/api/export/scenario-runs" });
```
Comment (0xdeafcafe) [inline langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:40]: should `export` be it's own api surface, or should it live under scenarios namespace? i would lean latter, but maybe we already do this another way?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: Done in `5068e3865` — both now throw a `HandledError` subclass, so `onError(handleError)` produces the response. The 401 went from `{"error": "..."}` to:

```json
{"error":"scenario_run_export_unauthenticated",
 "message":"You must be logged in to export scenario runs.",
 "fault":"customer",
 "trace":{"traceId":"03c215c3…","spanId":"102933d8…"}}
```

Verified live; an authenticated export still returns 200. Two subclasses rather than one (`HandledError` is abstract, and the repo models one class per failure) so a caller can tell "log in" from "ask for access", and the 403 records `scenarios:view` as the blocker in `meta` instead of a bare status.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: Done in `5068e3865` — both now throw a `HandledError` subclass, so `onError(handleError)` produces the response. The 401 went from `{"error": "..."}` to:

```json
{"error":"scenario_run_export_unauthenticated",
 "message":"You must be logged in to export scenario runs.",
 "fault":"customer",
 "trace":{"traceId":"03c215c3…","spanId":"102933d8…"}}
```

Verified live; an authenticated export still returns 200. Two subclasses rather than one (`HandledError` is abstract, and the repo models one class per failure) so a caller can tell "log in" from "ask for access", and the 403 records `scenarios:view` as the blocker in `meta` instead of a bare status.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: `getServerAuthSession` takes a `NextApiRequest`-shaped object and only ever reads `.headers` and `.cookies` off it, which a Hono `Request` provides — so the cast is structurally sound at runtime today. It is not *type*-safe though: nothing stops the callee reaching for a Next-only field later, and the cast would keep compiling.

I kept it because it is exactly what the existing trace export does (`app/api/export/traces/[[...route]]/app.ts:44`) and I did not want to diverge in this PR. Happy to change it — but the fix belongs on both call sites at once (a narrow `SessionRequest` interface of just what the function reads, so the cast disappears). Want that as a follow-up, or in here?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: If you mean `request` going into `exportRuns` — it is the zod-validated payload (`scenarioRunExportRequestSchema`), and every field reaching ClickHouse goes in as a bound query parameter (`{tenantId:String}`, `{exportScenarioId:String}`, …), never string-interpolated, so there is no injection surface. `projectId` is separately re-checked against the session by `hasProjectPermission` before this line, so it cannot address another tenant's data.

The one place it *was* interpolated raw was the `Content-Disposition` filename, which CodeRabbit caught and is now sanitised.

If you meant the streamed `chunk` instead: the CSV cells are neutralised against spreadsheet formula injection in the serializer (a leading `=`/`+`/`-`/`@` gets an apostrophe) — that one was real, 121 cells in our local data needed it. Let me know if you meant something else.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:40]: We do already do this the first way — `/api/export/traces/` exists today (`app/api/export/traces/[[...route]]/app.ts`, mounted in `api-router.ts`), so `/api/export/scenario-runs/` sits next to it deliberately: `export/*` is the surface, one app per exportable resource.

The argument for keeping it: an export is not really a scenarios operation, it is the same streaming-download shape (auth → count → chunked CSV → progress broadcast) applied to a different resource, and grouping by that shape means the next one is a sibling rather than a fourth namespace growing a `/download` route.

The argument for your way: discoverability — everything scenario-shaped under one prefix, and `scenarios:view` already gates it.

No strong opinion, and it is a cheap move (basePath + one line in `api-router.ts`, no client change beyond the fetch URL). Your call — say the word and I will move it, but I would want to move the traces one at the same time so we do not end up with one of each.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: Agreed that reaching into `.repository` from the route is the wrong seam — it hands an HTTP handler a data-access object.

Worth noting the constraint: `AppDependencies["simulations"]` is built once at boot, and `ScenarioRunExportService` is stateless apart from the repository, so registering it as `simulations.runs.export` is straightforward — but `simulations.runs` is currently a `SimulationRunService` instance, not a namespace, so it would be either `simulations.export` alongside `runs`, or a method on `SimulationRunService`.

I would lean `simulations.export` (own service, own concern — the export sweep is a different read pattern from the UI queries, which is why it has its own repository method rather than reusing `getRunDataForAllSuites`).

Happy to do it in here — it touches `dependencies.ts`, `app.ts` and this route. Want it in this PR, or as a follow-up so this one stays reviewable?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:40]: its the same way it is in traces so i just copied that, lmk if youd like both changed othewrwise?

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: Now pinned by a test rather than just argued: `scenario-run-export-route.integration.test.ts` sends a projectId of `evil"; download; x="` and asserts the `Content-Disposition` still contains exactly two quotes, so the filename cannot be broken out of. The body itself stays zod-validated by `scenarioRunExportRequestSchema`.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts]: Confirmed against the implementation: `getServerAuthSession` types its `req` as `NextApiRequest | GetServerSidePropsContext["req"] | NextRequest`, reads only `.headers` off it, and passes that through `toHeaders`, which short-circuits on `input instanceof Headers` (`src/server/auth.ts:62`). `c.req.raw` is a Fetch `Request`, so its `.headers` is a `Headers` and takes that branch — nothing reads `.cookies` directly.

langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts
Comment (Aryansharma28) [reply langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:40]: Going with keeping it here, so this stays consistent with what is already in the tree.

`/api/export/traces/` exists today and is shaped the same way, so `export/*` is the surface and there is one app per exportable resource. The grouping is by *shape* rather than by domain — an export is the same streaming download every time (auth → count → chunked CSV → progress broadcast), just pointed at a different resource, so the next one lands as a sibling rather than a fourth namespace growing its own `/download` route.

Happy to revisit as a follow-up that moves both at once — moving only this one would leave us with one of each, and moving traces is a breaking change for the CLI and agents that call it, which is more than this PR should carry. Resolving for now; reopen if you would rather it went the other way.

Aryansharma28:
## Rebased onto main, CI green, all threads resolved

The branch was 76 commits behind. Rebasing surfaced real problems rather than just conflicts, so this is worth a re-read rather than a re-skim.

**64 checks pass, 0 failures.** (One `secrets` failure was trufflehog's *binary download* getting a connection reset — the scan never ran; the re-run is green.)

### Two defects, both found by writing the missing tests

**A 403 reached the user as "Export failed: 403".** The route already raised `HandledError`s with proper codes and the secured app already serialized them — the client threw the body away and rendered the status. The two codes now have entries in `codes.ts` / `presentation.ts`, and the client parses the failure body, so someone without `scenarios:view` is told to ask an admin for access instead of reading a status number.

**Export CSV stayed enabled when no runs matched**, so taking it wrote a header and zero rows. It was gated on `isLoading` alone. The old comment argued against gating on `totals.runCount` — right about the general case, wrong about zero: `runCount` is pages-fetched-so-far, so comparing it against a total would block a valid export of a long history, but zero is the one value it reports exactly. The regression test renders `RunHistoryPanel` rather than the filter bar, because the bar only renders the flag it is handed and asserting there would have passed against the bug.

### Test coverage

The spec described 31 behaviours and enforced none of them — no scenario carried a pyramid tag, so `feature-parity` counted the whole file as invisible. Tagging it exposed that only the serializer had tests. Now **31/31 bound**, with ~34 new tests:

| Level | Covers |
|---|---|
| ClickHouse sweep (real CH) | date range, scenario, set, archived exclusion, cross-project `TenantId` scoping, STALLED derivation |
| Route | 401/403 with their codes, gzip round-trip, `Content-Disposition` filename + quote injection, audit written before streaming |
| Service | pagination, header written once across batches, progress reaching its total, abort stopping the sweep |
| UI | dialog opens / offers both depths, disabled when empty, live count + cancel |

### Review threads

All resolved. On the API-surface question — keeping it at `/api/export/scenario-runs/` so it stays consistent with the `/api/export/traces/` already in the tree; happy to revisit as a follow-up that moves both at once, since moving only this one leaves one of each and moving traces is breaking for the CLI and agents.

Aryansharma28:
## Deep review pass — 3 findings, all fixed in `a0c4499`

Audited across architecture, business logic, regression risk, quality, smells, YAGNI/SOLID, security, performance, tests and docs. **P0: 0 · P1: 1 · P2: 0 · P3: 2 · P5: 2.** The export path itself held up; the one serious finding was a regression this PR introduced in the *existing* run-history list.

### P1 — the run list stopped showing new runs

`usePeriodSelector` builds a relative preset as `endDate: now`, and its `useMemo` **deliberately excludes `now` from its deps** (stated in the code), so `period.endDate` is pinned at mount. This branch newly threaded that value into the list query, where `buildDateFilter` turns it into `StartedAt <= …` and `HAVING max(CreatedAt) <= …`.

What makes it conclusive: before this branch the client sent **no** upper bound, and `resolveDateRange` defaulted it to `Date.now()` **per request** — a live bound. So the effect was: start a suite from the Runs page and the new run never appears until you reload, on the one surface whose job is watching runs happen.

Fixed by putting the list back to sending no upper bound. The export still sends both — it is a snapshot the user asked for, not a live view. Regression test confirmed to fail against the bug (`expected 1718452800000 to be undefined`).

### P3 — the "Export produced no data" toast could never fire

`unparse` always writes the header when `includeHeader` is true, so the response body is never zero bytes and the `blob.size === 0` branch was dead. A filter matching nothing therefore downloaded a header-only CSV with no explanation. Now keyed on the count the server already reports, and raised *before* the download — a spreadsheet holding only a header reads as broken, and the reason isn't in the file.

### P3 — `traced()` gave async generators no span

`ScenarioRunExportService` is the first traced class with an `async *` method, which is why this surfaced here (the trace `ExportService` isn't wrapped, so it was never affected). The delegation fixed the "not async iterable" crash but emitted no span, so the class looked traced and produced nothing. `withActiveSpan` can't help — it owns the whole call and would close the span at the first yield — so the span is now managed around the iteration and ends whether the generator is drained, throws, or is abandoned when a cancelled export breaks the loop.

### P5 — observations

- Trace and scenario-run exports share one `export_progress` channel behind two different permissions (`traces:view` / `scenarios:view`). Only counts cross, and export ids are KSUIDs, so not actionable — worth knowing for the next export that lands here.
- Praise, genuinely: `EXPORT_SORT_KEY` as a single expression across `ORDER BY`, the cursor predicate and the returned cursor, plus the alias-shadowing note on the dedup predicate. Those are the two ways ClickHouse returns silently wrong or zero rows, and both are pinned with the reasoning attached.

### Not re-raised

- **CodeRabbit's count-query cost** — the double-execution half was fixed (the route computes the total once and passes it in). The read-budget half wasn't applied, but `queryRows` has no options parameter and *no* query in that file declares budgets — a file-wide gap, not this PR's.
- **CodeRabbit's NULL-`StartedAt` cursor divergence** — properly fixed by `EXPORT_SORT_KEY`.

### Couldn't verify from the diff alone

Real-world ClickHouse cost of the count query on a large tenant, and how the gzip'd stream behaves through the production proxy/CDN. Both want a look at staging rather than the diff.

sergioestebance:
## Codex Review (branch diff against `main`)

The export has security and scalability defects on its primary large-download path, and it can incorrectly prevent valid filtered exports. The repository query can also return stale run versions around date boundaries.

- **[P1] Respect backpressure while producing the response** — `langwatch/src/app/api/export/scenario-runs/[[...route]]/app.ts:158-162`
  For large exports or slow clients, this `start()` loop continues querying and enqueuing every page regardless of `controller.desiredSize`; piping through `CompressionStream` does not make this producer wait once downstream applies backpressure. The server can therefore buffer most of a full transcript export in memory and exhaust a pod. Drive the generator from `pull()` or otherwise await writable readiness, and stop it from `cancel()`.

- **[P1] Sanitize identifier cells against CSV formula injection** — `langwatch/src/server/export/scenario-runs/csv-serializer.ts:213-219`
  When an SDK/API caller supplies a scenario set ID, run ID, message ID, or target reference beginning with `=`, `+`, `-`, or `@`, these cells are appended raw even though the ingestion schemas accept arbitrary strings. Opening the resulting file in Excel or Sheets can therefore evaluate attacker-controlled formulas despite the protection applied to prose fields; pass all non-generated string cells through the same neutralization function.

- **[P2] Keep export enabled when unloaded pages may match** — `langwatch/src/components/suites/RunHistoryPanel.tsx:455-460`
  When the selected scenario/status has no match in the currently fetched pages but an older page does, `totals.runCount` is zero and this disables the export even though the server-side sweep would return matching runs. Only treat zero as definitive once pagination is exhausted, or use a server-side filtered count.

- **[P2] Stream the response body instead of buffering a Blob** — `langwatch/src/components/suites/useExportScenarioRuns.ts:243-244`
  For exports containing large conversation histories, `response.blob()` consumes the complete decompressed CSV before the browser download is started. This defeats end-to-end streaming, delays the visible download until completion, and can exhaust the tab's memory; write `response.body` incrementally through a download-capable stream instead.

- **[P2] Deduplicate before applying mutable date filters** — `langwatch/src/server/app-layer/simulations/repositories/simulation.clickhouse.repository.ts:1303-1307`
  When `StartedAt` changes between ReplacingMergeTree versions—such as when a provisional timestamp is replaced by the run-started event near a range boundary—putting `whereClause` inside the dedup subquery chooses the newest version only among in-range versions. If the true latest version moved outside the range, this returns an older, stale snapshot with incomplete status/messages; select the latest version using stable key/tenant predicates first, then apply date and scenario filters to that version.

Aryansharma28:
@sergioestebance thanks — four of the five are fixed in `1e1fe2d`, each with a regression test I confirmed fails against the bug it covers. The fifth I'd like to push back on slightly, details at the end.

### [P1] Backpressure — fixed, and you were right about *why* `pull()` alone wouldn't do it

Moving the producer from `start()` to `pull()` was necessary but not sufficient, exactly as your note said: `CompressionStream` drains its source without bound. Measured on this route's own shape (one read, then stop):

| producer | pages pulled after one read |
|---|---:|
| raw pull-driven source | **2** |
| same source through `CompressionStream` | **~65,000** |

So gzip now goes through Node's stream plumbing, which honours the pipe's high-water mark — read-ahead is bounded in **bytes** (~800KB) rather than pages, and a larger page just means fewer buffered. `cancel()` returns the generator, so a client that goes away ends the sweep instead of leaving it paging ClickHouse. Test asserts the sweep stops and does not resume after cancel.

### [P1] Formula injection on identifiers — fixed

You're right that these aren't all server-generated: scenario set id, scenario id, batch id, message id, trace id and target reference all arrive from the SDK as arbitrary strings. They now go through the same neutralization as the prose fields. Tests cover `=`, `+`, `-` and `@` leading each of them.

### [P2] Export disabled with unloaded pages — fixed

Agreed, and it was my own change from earlier today that introduced it. Zero loaded runs only means "nothing matched" once there is nothing left to fetch, so it now takes `hasMore` into account. Test covers the case where the loaded pages are empty but the sweep would still match.

### [P2] Dedup before mutable date filters — fixed, and worse than it looked

Confirmed `StartedAt` really is mutable across versions: the projection opens a run with `StartedAt` null and only sets the real value when the started event lands, with a `CreatedAt` fallback on write. So the filter inside the subquery picked the newest version *within the range*, and a run whose corrected timestamp moved out of the window exported its stale earlier snapshot — old status, missing messages — rather than being excluded.

Only the stable predicates (set, scenario — a run cannot move between either, and set is part of the dedup key) stay in the subquery; the date bound applies to whichever version actually won. The cost is that the subquery no longer prunes partitions by date, which I think is the right trade: an export that is cheaper but occasionally wrong at a boundary isn't worth having. The integration test inserts two versions of one run straddling the window and asserts the range is judged on the corrected one.

### [P2] `response.blob()` buffering — not fixed, and I don't think it should be here

You're right that it buffers the whole decompressed CSV before the download starts. I couldn't find a fix that belongs in this PR:

- **Native navigation download** streams properly, but the endpoint is a POST and the progress UI depends on reading `X-Export-Id` off the response — which a navigation download can't do. It would mean giving up progress, or a GET with every filter in the query string.
- **`showSaveFilePicker`** genuinely streams, but needs transient activation, so it has to be called *before* the fetch — which turns the flow into a save dialog up front. That's a UX change, and Chromium-only, so it also forks the download path by browser.

Both are real designs, neither is a local change, and the second is a product decision rather than mine. Happy to take it as a follow-up issue — say which direction you'd prefer and I'll open it. In the meantime the wire is gzipped (~9x on full mode), so the buffered object is the decompressed CSV.

sergioestebance (approved):

Aryansharma28:
## Correcting myself on the `response.blob()` finding — I measured it, and @sergioestebance is more right than my reply allowed

I said the buffering wasn't worth fixing here. I went and measured instead of arguing, and the numbers say the concern is real. Posting them because they change the shape of the follow-up.

### How big the file actually gets

Bytes per run from the real serializer, full mode, realistic transcripts:

| conversation | KB/run | 10k runs | 50k runs | 100k runs |
|---|---:|---:|---:|---:|
| 6-turn | 5.3 | 52 MB | 258 MB | 0.50 GB |
| 12-turn | 11.7 | 114 MB | 571 MB | 1.12 GB |
| 30-turn | 38.0 | 371 MB | **1.85 GB** | 3.62 GB |

### What a blob that size actually costs

Headless Chromium, fetch + `response.blob()` + `createObjectURL`, same shape as the hook:

| blob | JS heap before → after | Chromium RSS delta | outcome |
|---|---|---:|---|
| 600 MB | 9.5 MB → 9.5 MB | — | ok, 645ms |
| 2000 MB | 9.5 MB → 9.5 MB | — | ok, 2.3s |
| 1800 MB | 9.5 MB → 9.5 MB | **+1820 MB** | ok |

Two things follow, and they point in opposite directions:

- **It will not throw.** The JS heap does not move at all — the blob lives in the browser's own storage, so `jsHeapSizeLimit` is never the binding constraint. "Exhaust the tab's memory" in the JS sense doesn't happen.
- **It costs its full decompressed size in browser process memory, 1:1.** Not spilled to disk on this machine. So a 30-turn, 50k-run export is ~1.8 GB of resident memory in the browser — on a laptop with other tabs open, that is swap or an OOM-killed tab.

The failure mode being *silent* is what makes it worse than I gave it credit for: there is no exception to catch and no error to show, the machine just degrades.

### What I think the fix is now

My two objections still stand for the general case — a navigation download can't read `X-Export-Id` so it loses progress, and `showSaveFilePicker` needs transient activation so it must precede the fetch. But the measurements suggest a narrower fix that avoids both:

**Branch on size, not on browser.** The server already knows the count before it streams a byte (`X-Total-Runs`), and mode tells us roughly the bytes-per-run. Below a threshold, keep today's instant-download path exactly as it is. Above it, take the save-picker path — called inside the Export click handler, so activation is still live — and stream `response.body` straight to disk. Firefox and Safari fall back to the buffered path with a warning that a large export may be heavy.

That keeps the common case untouched, puts the streaming path only where it earns its complexity, and the threshold is a number we can set from the table above rather than a guess.

Happy to open that as a follow-up issue with these numbers in it — or to do it in this PR if you would rather it not ship buffered at all. Your call; I don't think it should be mine given it changes what the user sees.

PR #6288, opened 2026-07-28
fix(langy): let the session key reach the experiment surface

Aryansharma28:
---

# Final state (after review)

The goated-review pass on this PR found that the coverage guarantee, as first written, covered **1 of 13** files that enforce RBAC inside a handler — and that `classifyForLangy` failed **open** on unknown resource families while its own comment claimed the opposite. CodeRabbit independently found both. Everything below is now addressed.

| Finding | Source | State |
|---|---|---|
| `classifyForLangy` granted unknown families | CodeRabbit + review | **Fixed** — allowlist (`DELEGABLE_FAMILIES`); unknown family/action/shape all refuse. Pinned by `langyPermissionPolicy.unit.test.ts`. |
| `handlerManagedAuth` permissions optional | CodeRabbit + review | **Fixed** — required, named object param. All 41 call sites updated; the compiler now refuses an undeclared handler-managed route. |
| Only 1 of 13 RBAC-enforcing files declared | review | **Fixed** — all 13 declare; shared policies split by grain where routes differ. |
| Duplicated route-registry walk across two suites | CodeRabbit | **Fixed** — extracted to `__tests__/helpers/routePermissions.ts`. |
| `LANGY_CANDIDATE_PERMISSIONS` mutable | CodeRabbit | **Fixed** — `readonly` + `Object.freeze`. |
| Spec scenarios described release process | CodeRabbit | **Fixed** — rewritten as user-observable outcomes. |
| Coverage sweep in the integration lane, needs no DB | review | **Fixed** — moved to the unit lane. |
| `SESSION_ONLY_EXPERIMENT_ROUTES` hand-maintained path list | review | **Fixed** — `credential` on the policy; filter is `isApiKeyReachable()`, no path list. |

**Result of making the surface visible:** with all 13 files declaring, the sweep sees the whole handler-enforced API and finds **no further gaps** — the candidate list is complete against every route the platform mounts, not just the middleware-declared ones.

### Two things made visible rather than fixed

Both are noted at the point of declaration rather than silently widened:

- **`evaluations-legacy` asks `evaluations:manage` for append/create actions** — the same over-coarse grain `POST /api/experiments/:slug/run` had, refusing any least-privilege key holding only `evaluations:create`. Declaring it is the precondition for fixing it; fixing it is a separate change.
- **`collector.ts` gates trace collection on `traces:create`** — the route a colleague described spending "ages" working out from the code. Now readable off the registry.

### The last hand-maintained list, removed

An earlier revision of this description named `SESSION_ONLY_EXPERIMENT_ROUTES` as a known weakness — a hand-maintained path list doing the job the PR argues hand-maintained lists cannot do. It is gone. `handlerManagedAuth` now carries `credential: "session" | "apiKey" | "both"` alongside `permissions`, and the sweep filters on `isApiKeyReachable(route.policy)`. Structural, and it generalises past `/api/experiments`.

That fix creates one hazard worth naming: a route marked with the *wrong* credential is **invisible** rather than wrong — the sweep skips it and stays green while checking less than it claims. Both API-key policies in `experiments-v3` shipped as `session` for one commit, silently dropping `evaluations:view` and `evaluations:create` from the session-key suite, and nothing caught it. The classification is therefore pinned in its own right: `langy-permission-coverage.unit.test.ts` asserts every grain the experiment surface enforces for a key-bearing caller is reachable as one, and that the session-only `evaluations:manage` does not leak in.

### Reviewing the diff

Two changes landed on `main` between this branch's fork point and now — the Biome cleanup (#6305) and the handled-errors migration (#6010) — and both touch the same route files this PR does. The merge is faithful (main's error handling, this branch's permission declarations), but Biome's chain formatting means **~2,100 changed lines of which ~180 are substantive**. `?w=1` on the Files tab cuts through it.

`otel.ts` was the one file where the churn had a real cause: three identical inline policy objects, now one shared `otelIngestAuth` const. That is a DRY fix in its own right, and it collapses that file's diff from 702 lines to 16.

### Verification

`pnpm lint`, `pnpm typecheck`, `pnpm typecheck:tests` all clean · 295 Langy unit tests · 5 session-key integration tests. Each fix verified non-vacuously by reverting it and watching the corresponding suite go red. The new Biome house-rules gate was reproduced locally against the merge base: **0 new violations**.

### Follow-up filed

**#6327** — 15 integration suites are permanently skipped in CI by a `describe.skipIf(isTestcontainersOnly)` guard that is always `skipIf(true)` there, because `setupEnv.ts` sets `TEST_CLICKHOUSE_URL` for the whole lane. Found by removing that guard from `langySessionKey.integration.test.ts` here and getting real coverage back from it.

## Summary by CodeRabbit

* **New Features**
  * Tightened Langy session-key behavior so callers can view experiments and start runs without gaining destructive evaluation permissions.
  * Added action-specific permission gating for experiment endpoints (run listing vs run creation), alongside clearer permission ceilings for annotations, traces, workflows, and other API operations.
* **Bug Fixes**
  * Standardized managed-auth permission requirements across routes to improve authorization consistency.
* **Tests**
  * Added/expanded permission-coverage checks and ceiling assertions to prevent permission gaps or accidental access expansion.


langwatch/src/server/app-layer/langy/__tests__/langy-permission-coverage.unit.test.ts
```diff
@@ -0,0 +1,134 @@
...
+ * production 403 three times (`project:view`, `scenarios:create`,
+ * `experiments:view`).
+ *
+ * This triangulates three facts that are maintained independently, so no two
+ * of them can drift into agreement while being wrong together:
+ *
+ *   1. the ROUTE REGISTRY — what the API actually demands at the door
+ *   2. `classifyForLangy`  — what Langy is allowed to be delegated
+ *   3. `LANGY_CANDIDATE_PERMISSIONS` — what the mint actually asks for
+ *
```
Aryansharma28, line 20: [P1] The guarantee this test advertises covers 1 of 13 files that actually enforce RBAC in handlers.

**Why:** @coderabbitai flagged the optional `permissions` field as a *future* risk. It is a *present* fact. Thirteen files enforce an RBAC permission inside a handler-managed route; this PR annotates exactly one (`experiments-v3.ts`). The other twelve return `[]` from `policyPermissions`, so this sweep walks straight past them — including API-key-authenticated routes in `annotations.ts`, `traces-legacy.ts` and `evaluations-legacy.ts`.

That matters because handler-managed routes are precisely where the second half of this bug lived. The commit message says the gap is "impossible to reintroduce silently"; as shipped, it is impossible to reintroduce silently *only via `requires(...)` middleware*, and remains fully possible via a self-enforcing handler.

**Evidence:**
```
handlerManagedAuth call sites: 41
files also enforcing RBAC in-handler: 13
  workflows.ts (3)  evaluations-legacy.ts (5)  dataset-generate.ts (2)
  auth-cli.ts (3)   experiments-v3.ts (11)     misc.ts (4)
  collector.ts (2)  playground.ts (2)          otel.ts (2)
  traces-legacy.ts (7)  annotations.ts (9)     scenario-generate.ts (2)
  app/api/export/traces/[[...route]]/app.ts (2)
annotated by this PR: 1
```

**Suggested fix:** Take CodeRabbit's mandatory-`permissions` change, then annotate the remaining twelve (`[]` where the handler gates on something that is not an RBAC permission). Until then, soften the claim in the commit message and PR description — the current wording overstates what ships.

langwatch/src/server/app-layer/langy/__tests__/langySessionKey.integration.test.ts
```diff
@@ -43,12 +51,63 @@ const EXPECTED_HELD_SUBSET = [
...
+  "experiments:view",
+  "evaluations:manage",
+];
+
+// The `/api/experiments` routes that authenticate by USER SESSION rather than
+// by API key. Langy reaches the surface with a key, so it never calls these —
+// and `/execute` legitimately demands `evaluations:manage`, which the Langy key
+// must never hold. Excluded by path so the rest of the surface can be read
+// straight off the registry.
+const SESSION_ONLY_EXPERIMENT_ROUTES = [
```
Aryansharma28: [P2] This swaps a hardcoded permission list for a hardcoded path list — the same drift shape the PR sets out to remove.

**Why:** The PR's whole argument is that hand-maintained lists silently fall out of date. `SESSION_ONLY_EXPERIMENT_ROUTES` is a hand-maintained list. Add a new session-authenticated route under `/api/experiments/*` and it will not be in it, so `experimentRoutePermissions()` will pick up that route's session-only grain and demand the Langy API key clear it — a spurious failure whose cause is two files away from the change that triggered it.

Distinct from @coderabbitai's duplication finding on line 102: even after the helper is extracted to shared test support, the exclusion mechanism itself is still path-matching.

**Evidence:** `SESSION_ONLY_EXPERIMENT_ROUTES = ["/api/experiments/execute", "/api/experiments/abort"]` — enumerated by path, with nothing tying it to the routes' actual credential type.

**Suggested fix:** Make the credential type structural rather than textual. The policy already knows whether a route authenticates by session or API key — `handlerManagedAuth` could carry `credential: "session" | "apiKey"` alongside `permissions`, and the filter becomes `credential === "apiKey"` with no path list to maintain. That also generalises the sweep beyond `/api/experiments`.

langwatch/src/server/app-layer/langy/__tests__/langy-permission-coverage.integration.test.ts
```diff
@@ -0,0 +1,134 @@
...
+ * of them can drift into agreement while being wrong together:
+ *
+ *   1. the ROUTE REGISTRY — what the API actually demands at the door
+ *   2. `classifyForLangy`  — what Langy is allowed to be delegated
+ *   3. `LANGY_CANDIDATE_PERMISSIONS` — what the mint actually asks for
+ *
+ * A new route requiring a new permission fails here until someone decides,
+ * naming the route and the permission, rather than reaching a user as a 403.
+ *
+ * Requires: nothing beyond module loading — the registry is populated as a
```
Aryansharma28: [P2] This is in the integration lane but needs no database — it belongs in the unit lane.

**Why:** The test only imports the composed router and reads the in-memory route registry. Its own docstring says so ("Requires: nothing beyond module loading"). Sitting in `*.integration.test.ts` means it is gated behind Postgres + ClickHouse bootstrap and runs under the integration config's `pool: "forks"` with `fileParallelism: false` — for pure module introspection.

The cost is not just speed. A drift guard is worth most when it runs often, and the unit lane runs on far more changes than the integration lane does.

**Evidence:** Copied verbatim to `_scratch_cov.unit.test.ts` and ran it through `pnpm test:unit`:
```
Test Files  1 passed (1)
     Tests  3 passed (3)
```
No database, no globalSetup.

**Suggested fix:** Rename to `langy-permission-coverage.unit.test.ts`. Note this interacts with CodeRabbit's shared-helper extraction — if the helper is shared with `langySessionKey.integration.test.ts` (which genuinely does need Postgres), the shared module must not pull in DB-dependent imports.

langwatch/src/server/app-layer/langy/__tests__/langySessionKey.integration.test.ts
```diff
@@ -16,14 +16,22 @@ import {
...
   mintLangySessionApiKey,
 } from "../langyApiKey";
 
-const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL;
+// This suite only needs Postgres — every harness (CI's testcontainers, native
+// local services) provides that, so it runs unconditionally. It used to carry
+// an `isTestcontainersOnly`/`TEST_CLICKHOUSE_URL` skip guard, which permanently
+// skipped it EVERYWHERE: CI always sets TEST_CLICKHOUSE_URL for the
+// ClickHouse-dependent suites sharing the run. The file had gone stale unnoticed
+// as a result. Do not add the guard back — see PR #5988.
```
Aryansharma28, line 35: [P5] Removing this skip guard is the most valuable line in the PR, and it deserves to outlive this change.

**Why:** `isTestcontainersOnly` had disabled this suite everywhere — CI always sets `TEST_CLICKHOUSE_URL` for the ClickHouse suites sharing the run — and the file had silently rotted (an assertion regex-matching against an object after a return-shape change). This is the second time the same trap has been found (PR #5988 was the first), which makes it a pattern rather than an accident.

**Suggested fix:** Not a change to this PR. Worth a follow-up: grep for the remaining `describe.skipIf(isTestcontainersOnly)` suites and check each one actually needs ClickHouse. Several likely do not, and are silently not running.

langwatch/src/server/routes/experiments-v3.ts
```diff
@@ -59,9 +59,18 @@ const logger = createLogger("langwatch:experiments-v3");
...
+// The read endpoints (runs list / status / results) and the run endpoint gate
+// on different grains, so they declare separately: a single shared policy
+// would report the coarser of the two for routes that only read.
+const apiKeyAuthRead = handlerManagedAuth(
   "project API key resolved in-handler via TokenResolver + enforceApiKeyCeiling",
+  { permissions: ["evaluations:view"] },
+);
+const apiKeyAuthRun = handlerManagedAuth(
+  "project API key resolved in-handler via TokenResolver + enforceApiKeyCeiling",
+  { permissions: ["evaluations:create"] },
```
Aryansharma28: [P5] Declaring the enforced grain here is the right pattern, and it is what makes the registry-driven assertion in `langySessionKey.integration.test.ts` possible.

**Why:** Worth calling out because the payoff is non-obvious: because this line declares `evaluations:create`, the session-key suite reads the grain off the registry instead of hardcoding it, so coarsening this route back to `evaluations:manage` fails that suite against a real minted key. I verified both directions by reverting each fix and watching the corresponding test go red.

That is the property to preserve when applying CodeRabbit's mandatory-`permissions` change — the value is in routes *declaring* what they enforce, not in the field being optional.

Aryansharma28 (commented):
## goated-review — PR #6288

Reviewing my own change adversarially, at the author's request, with the specific question: **is this the best root-cause fix, and is the problem actually solved?**

**Verdict: right diagnosis, right architecture, incomplete execution. Not solved as shipped.**

The direction is correct — a rule plus a three-way reconciliation beats a fourth hand-patch. But three defects mean the guarantee the commit message advertises ("impossible to reintroduce silently") does not currently hold.

**P0: 0 · P1: 2 · P2: 2 · P3: 0 · P4: 0 · P5: 2**

### Not re-raising — @coderabbitai got there first, and was right

CodeRabbit's pass on this PR was strong and I am deduping against all six of its findings rather than restating them:

| CodeRabbit finding | My assessment |
|---|---|
| **Reject unknown permission families** (`langyPermissionPolicy.ts:157`, Major) | **Correct, and the most serious issue in the PR.** Confirmed empirically — see below. |
| **Make handler-managed `permissions` mandatory** (`access-policy.ts:58`, Major) | Correct. Measured blast radius in P1-2 below. |
| **Duplicate route-permission helper** (`langySessionKey…:102`, Major) | Correct — straight DRY violation against the project rule. |
| **Make `LANGY_CANDIDATE_PERMISSIONS` immutable** (`langyApiKey.ts:68`, Trivial) | Correct, and it is a security ceiling, so worth more than Trivial. |
| **Spec scenarios describe process, not user behavior** (`…feature:78`, Major) | Correct. "the team tries to ship" is not user-observable; it violates the repo's own feature-file rule. My scenarios were bad. |
| **Unrunnable test command in docstring** (Minor) | Correct. |

**Empirical confirmation of CodeRabbit's unknown-family finding.** I ran `classifyForLangy` against three invented families:

| permission | disposition |
|---|---|
| `billing:create` | **granted** |
| `piiExport:create` | **granted** |
| `impersonation:update` | **granted** |
| `secrets:view` | excluded ✓ |

The file's own comment claims *"Fail-closed at every step: an unrecognised action and an unrecognised family shape both come back `excluded`, so the rule can only ever be too strict."* **That claim is false.** Only an unrecognised *shape* (no colon) is excluded; an unrecognised *family* is granted. Worse, it inverts the PR's stated principle — the coverage test then goes red until the new family is added to the candidate list, so red CI actively pressures toward granting Langy a family nobody has assessed. A false fail-closed claim in a security policy comment is worse than no comment.

### Cross-PR finding that could not be anchored inline (file not in this diff)

**[P1] The same coarse-grain bug this PR fixes still exists in `evaluations-legacy.ts`.**

[`POST /api/evaluations/batch/log_results`](https://github.com/langwatch/langwatch/blob/b820e314f153932089443da873ee843cd7a448d1/langwatch/src/server/routes/evaluations-legacy.ts#L187) is API-key authenticated and gated on `evaluations:manage`, but it *appends batch results* — a write, not administration. That is structurally identical to the `POST /api/experiments/:slug/run` bug this PR fixes, and any least-privilege key holding `evaluations:create` is refused at the door exactly the same way.

Two consequences: the PR fixes one instance of a class rather than the class, and the new coverage test cannot see this one, because `evaluations-legacy.ts` is among the twelve unannotated handler-managed files (P1-2). I am not asking for it to be fixed in this PR — I am flagging that "the class is closed" would be an overstatement, and it deserves a follow-up issue.

### Things I could not verify from the diff alone

- Whether `POST /api/evaluations/batch/log_results` is reachable from Langy's CLI surface today. The grain is wrong regardless, but I could not confirm live impact.
- Whether the legacy-`TeamUser`-vs-`RoleBinding` resolver divergence documented in the PR description affects production tenants or only dev-seed data. I hit it on dev data only.
- The 12 unannotated handler-managed files were classified by grepping for RBAC calls; I read `workflows.ts` (session-auth, correctly out of scope) and `evaluations-legacy.ts` closely, and did not read the other ten line by line.

### CI

`test` and `sdk-python-complete` are red, and **neither is caused by this PR** — the failure is `openai.InternalServerError: Error code: 500` in `python-sdk` e2e tests hitting a live provider. Unrelated external flake. The TypeScript lanes (`lint`, `build`, `ci-scripts-test`) pass.

langwatch/src/server/app-layer/langy/__tests__/langySessionKey.integration.test.ts
```diff
@@ -209,6 +263,75 @@ describe.skipIf(isTestcontainersOnly)("Langy session key (caller-scoped)", () =>
...
+          organizationId,
+        });
+
+        const resolved = await TokenResolver.create(prisma).resolve({
+          token,
+          projectId,
+        });
+        expect(resolved).not.toBeNull();
+
+        const required = experimentRoutePermissions();
```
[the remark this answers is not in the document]
Aryansharma28: Already guarded — line 291, immediately above the loop:

```ts
const required = experimentRoutePermissions();
// Guard the guard: an empty list would make the loop below vacuous.
expect(required).toContain("experiments:view");
```

`toContain` on an empty array fails, so the vacuous case cannot pass. Verified rather than assumed — I made `experimentRoutePermissions()` return `[]` to simulate total credential misclassification and re-ran the suite:

```
× clears the ceiling for every permission the experiment surface asks for
AssertionError: expected [] to include 'experiments:view'
Tests  1 failed | 4 passed (5)
```

I think the finding may have been generated against a different shape — it refers to a `Promise.all`, but this is a sequential `for` loop with `await expect(...)` per permission.

Not adding a separate non-empty assertion: it would be strictly weaker than the one already there. Asserting a *specific* expected permission catches both the empty case and the subtler one where the filter returns a non-empty but wrong set.

langwatch/src/server/app-layer/langy/__tests__/langy-permission-coverage.unit.test.ts
Aryansharma28: Fixed, and re-verified after merging main.

All 13 files that enforce RBAC inside a handler now declare what they enforce — not just `experiments-v3.ts`:

| file | policies declared |
|---|---|
| `misc.ts` | 7 |
| `auth-cli.ts` | 4 |
| `annotations.ts` | 3 |
| `experiments-v3.ts` | 3 |
| `evaluations-legacy.ts` | 2 |
| `traces-legacy.ts` | 2 |
| `workflows.ts` | 2 |
| `collector.ts`, `dataset-generate.ts`, `otel.ts`, `playground.ts`, `scenario-generate.ts`, `export/traces/app.ts` | 1 each |

Split by grain where routes differ, per the point in the original comment that one shared policy per file would report the same requirement for a read and a delete — worse than reporting nothing, since an audit would believe it had the answer.

Three things followed from taking the fix rather than softening the claim:

1. `permissions` is now a **required** named parameter on `handlerManagedAuth`, so the compiler refuses an undeclared handler-managed route. The guarantee is structural, not a convention someone has to remember.
2. With all 13 declaring, the sweep sees the whole handler-enforced API — and finds **no further gaps**. The candidate list is complete against every route the platform mounts.
3. Two pre-existing over-broad grains became visible rather than being quietly widened: `evaluations-legacy` asks `evaluations:manage` for append/create, and `collector.ts` gates trace collection on `traces:create`. Both are noted at the point of declaration; fixing them is a separate change.

The commit-message wording was corrected too — the guarantee now describes what actually ships.

langwatch/src/server/app-layer/langy/__tests__/langySessionKey.integration.test.ts
Aryansharma28: Followed up as **#6327**, and the audit turned out worse than this comment guessed.

It is not that several of the remaining suites don't need ClickHouse — it is that **all 15 of them** are permanently skipped in CI, and none references ClickHouse anywhere except in the guard line itself. The chain is mechanical:

```
.github/workflows/langwatch-app-ci.yml:299
  CI_CLICKHOUSE_URL: "http://default:ci_password@localhost:8123/test_langwatch"
        ↓  job-level env — every integration shard has it
setupEnv.ts:60-64   (a `setupFiles` entry for the WHOLE lane, not per-suite)
  process.env.TEST_CLICKHOUSE_URL = process.env.CI_CLICKHOUSE_URL
        ↓
const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL   // always true in CI
describe.skipIf(isTestcontainersOnly)(...)                       // skipIf(true)
```

The guard reads as "skip in the testcontainers-only lane", but `setupEnv` runs for every integration file, so the variable it keys off is set for all of them regardless of whether the suite touches ClickHouse.

Several of the 15 are load-bearing on their face — `modelProvider.authz`, `enterprise-guards`, `project.regenerateApiKey` are authorization and credential suites.

#6327 carries the full list and a suggested shape: un-skip per suite, then replace the idiom with a helper that means what it says (`describeNeedsClickHouse`), and consider failing the lane when an entire file is skipped so a silent skip stops reading as green.

langwatch/src/server/app-layer/langy/__tests__/langySessionKey.integration.test.ts
Aryansharma28: Fixed, and taken the structural route this comment suggested rather than the cheaper one.

`SESSION_ONLY_EXPERIMENT_ROUTES` is gone. `handlerManagedAuth` now carries `credential: "session" | "apiKey" | "both"` alongside `permissions`, and the filter reads the route's own policy:

```ts
for (const route of allRegisteredRoutes()) {
  if (!route.path.startsWith("/api/experiments")) continue;
  if (!isApiKeyReachable(route.policy)) continue;   // ← structural, not textual
  ...
}
```

No path list to maintain, and the sweep generalises past `/api/experiments` for free.

One thing the fix surfaced that the original comment did not anticipate: making `credential` structural creates its *own* silent-failure mode, because a route marked with the wrong credential is **invisible** rather than wrong — the sweep skips it and stays green while checking less than it claims. That is not hypothetical; both API-key policies in `experiments-v3` shipped as `session` for one commit, which silently dropped `evaluations:view` and `evaluations:create` from the session-key suite, and nothing caught it because the one permission the guard named came from a different policy kind.

So the classification itself is now pinned in `langy-permission-coverage.unit.test.ts`: every grain the experiment surface enforces for a key-bearing caller must be reachable as one, and `evaluations:manage` (session-only) must *not* leak in.

The "known weakness, stated" paragraph in the PR description is now stale and has been removed.

langwatch/src/server/app-layer/langy/__tests__/langy-permission-coverage.integration.test.ts
Aryansharma28: Fixed — the file is `langy-permission-coverage.unit.test.ts` and runs in the unit lane.

The interaction flagged at the end was real and handled: the shared helper (`__tests__/helpers/routePermissions.ts`) imports only `~/server/api-router` and the access-policy/route-registry modules, so it stays DB-free and is safe to import from both the unit-lane coverage suite and the Postgres-dependent `langySessionKey.integration.test.ts`.

langwatch/src/server/routes/experiments-v3.ts
Aryansharma28: Preserved. `permissions` is now a **required** named parameter on `handlerManagedAuth`, so the value lives where this comment asked — in routes declaring what they enforce, with the compiler refusing one that does not.

All 41 call sites were updated, and all 13 files that enforce RBAC in-handler now declare, so the registry-driven assertion generalises past this one route.

sergioestebance:
## Ruthless review — no `P0`/`P1`

Traced the grain change on `POST /api/experiments/:slug/run`, all 41 `handlerManagedAuth` declarations against what their handlers actually enforce, the new policy/coverage triangulation, and the un-skipped integration suite.

**The behavioural change is correct.** `evaluations:manage` → `evaluations:create` on the run route takes access from nobody: `hasPermissionWithHierarchy` (`src/server/api/rbac.ts:504-531`) maps `:manage` onto `:create`, and legacy project keys bypass the ceiling entirely (`src/server/api-key/auth-middleware.ts:356`). The handler is a read-then-run against an existing experiment (`experiments-v3.ts:404-412`, 404s when absent), so the comment's justification holds. `sessionAuth`'s `evaluations:manage` matches both routes that carry it (`:195`, `:343`). No permission declaration in this diff *over*-states its route.

Three findings, none blocking.

---

### `P2` — the headline drift guard cannot fail from a route change

`langy-permission-coverage.unit.test.ts:23-25` claims it "triangulates three facts... 1. the ROUTE REGISTRY", and `:66-89` claims "a new route requiring a new permission fails here until someone decides". Neither assertion can observe the registry.

`grantedByPolicyButNotByTheMint()` computes `{p demanded by a route} ∩ {p : classifyForLangy(p) = granted} \ LANGY_CANDIDATE_PERMISSIONS`. The policy's granted set and the candidate list are **the same 27 permissions**, so that difference is empty for *every possible registry* — the intersection with "what routes demand" can only shrink an already-empty set. `askedForButExcludedByPolicy()` is the mirror image and is empty for the same reason. Computed from the two source files as they stand on this branch:

```
$ node check.mjs      # parses DELEGABLE_ACTIONS/FAMILIES/OFF_LIMITS/READ_ONLY
                      # from langyPermissionPolicy.ts and the frozen list from langyApiKey.ts
actions: view,create,update
granted count: 27 candidate count: 27
granted-but-not-candidate: []
candidate-but-not-granted: []
SETS IDENTICAL: true
```

So the guard fires in exactly one situation: someone edits `DELEGABLE_FAMILIES`/`READ_ONLY_FAMILIES` and forgets the matching line in `LANGY_CANDIDATE_PERMISSIONS` — two edits in files one import apart. It is a policy↔list consistency check, which is worth having, but it is not the registry-facing guarantee the docblock sells.

And it does not cover the incident that produced it. `experiments` was a *new family*; under this rule a new family is absent from `DELEGABLE_FAMILIES`, `classifyForLangy` returns `excluded`, the sweep stays green, and Langy gets the same 403 a user reports by hand. The file says so itself at `langyPermissionPolicy.ts:63-72` — the safe default is right, but "the rule's job is to make sure nobody can forget to" is not enforced by anything.

**Fix direction** — split the two claims so the registry one can actually fail:

1. Assert the policy↔list equality directly (`grantedSet(allPermissions) == LANGY_CANDIDATE_PERMISSIONS`). One assertion, honest about what it checks, no registry in the story.
2. Make the registry assertion fire on the *unclassified* case: for every permission an API-key-reachable route demands, require its family to appear in `DELEGABLE_FAMILIES` **or** `OFF_LIMITS_FAMILIES`. That forces the decision on a new family without granting anything by default — which is what the PR argues for and is currently the one thing nothing checks.

---

### `P2` — two `permissions: []` declarations state the opposite of what their handlers enforce

The compiler makes the field mandatory; nothing makes it true. Two of the new declarations are wrong, and both are on credential-adjacent routes:

- `src/app/api/dataset/[[...route]]/app.ts:143-149` — `permissions: []`, with the comment "it does not gate on a standalone RBAC permission". It does: all five direct-upload routes (`:317`, `:450`, `:493`, `:527`, `:558`) call `authorizeDirectUpload`, which enforces `datasets:manage` via `hasProjectPermission` (`direct-upload-auth.ts:34,105-108`) and `enforceApiKeyCeiling` (`:147`).
- `src/server/routes/auth-cli.ts:77-81` — `CLI_POLICY` with `permissions: []` still covers `POST /api/auth/cli/project-key` (`:1291`), which calls `refuseProjectKeyHandout` (`:1340`) → `hasProjectPermission(..., "project:update")` (`:341`). That is the *same guard* the PR gave `/approve` its own `cliApproveAuth` for, with the same rationale ("mints a credential usable outside the UI"). `/project-key` hands back the identical `Project.apiKey` and is declared as gating on nothing.

No runtime impact — `policyPermissions` / `isApiKeyReachable` have test-only consumers. But it falsifies "the sweep sees the whole handler-enforced API and finds no further gaps": the sweep cannot detect a false `[]`, and two of them shipped in the change that introduced the field. The `datasets:manage`-on-append grain is also the same over-coarse pattern already called out for `evaluations-legacy` — worth the same visible note.

**Fix direction** — correct the two declarations, then add the gate that would have caught them: a registry test asserting that any route file whose handlers call `hasProjectPermission`/`enforceApiKeyCeiling`/`requireApiKeyPermission` with a literal permission declares that permission on at least one of its handler-managed policies. Without it, `permissions` is a comment the compiler happens to typecheck.

---

### `P3` — assorted

- `evaluations-legacy.ts:98-102` — `/api/evaluations/list` is declared `handlerManaged` + `credential: "apiKey"`, but the handler performs no authentication at all. The reason string says "public evaluator catalogue"; the honest policy is `publicEndpoint(...)`. The `handlerManaged` kind is pre-existing; the `apiKey` credential claim is new, and it is the one an "which routes are unauthenticated?" audit reads.
- `sse.ts:133` — "per-message authorization happens upstream". It happens *downstream*, per-procedure, through the tRPC caller built at `:158-176` — exactly what `trpc.ts:159` says correctly. Match the wording.
- `specs/langy/langy-session-key.feature:43-70` — the five new scenarios carry no `@unit`/`@integration` tag and no `@scenario` binding, so `check-feature-parity.ts` counts them as `0/0 bound` and they enforce nothing. Pre-existing convention in that file, so not a regression — but five more unbound scenarios is five more green-looking rows.
- `langy-permission-coverage.unit.test.ts` pulls `~/server/api-router` (via the helper) into the **unit** lane, which imports every route module and their prisma/ClickHouse/Redis clients under `vmThreads` + a 512MB `vmMemoryLimit`. Green today; it is now the heaviest file in that lane, and the next route module with a module-load side effect breaks unit tests rather than integration ones.

---

### Credit where it's due

- Un-skipping `langySessionKey.integration.test.ts` is the best change in the diff. The `skipIf(isTestcontainersOnly)` guard was `skipIf(true)` in CI, and the file had drifted far enough that `mintLangySessionApiKey`'s return shape had changed under it without anyone noticing. #6327 for the other 15 is the right follow-up.
- The new integration test drives the real ceiling (`enforceApiKeyCeiling` against a resolved token) instead of asserting on the candidate list, and guards its own loop against vacuity at `:145`. That is the right level for this bug.
- `otel.ts` collapsing three identical inline policies into `otelIngestAuth` is a real DRY win hiding inside the Biome churn.

### Residual risk (accepted)

Langy can now start experiment runs — and therefore spend provider tokens — for every caller holding `evaluations:manage`, which is every `MEMBER` and `ADMIN`. Deliberate and within the caller's own ceiling, but it is a new cost-incurring capability for the assistant, not just a widened read.

sergioestebance (approved):
Blessed. 🙏 The session key can finally reach the surface its holder already owns — `experiments:view` restored, and the run route re-grained from the delete-implying `:manage` down to the `:create` it actually performs.

> "Do not withhold good from those to whom it is due, when it is in your power to act." — Proverbs 3:27

Traced the hierarchy (`:manage` still satisfies `:create`), the legacy-key bypass, every one of the 41 handler-managed declarations against what its handler enforces, and the un-skipped integration suite: no access lost, nothing over-declared. Ship it — residuals are two `permissions: []` declarations that under-state their routes (dataset direct-upload, `/project-key`) and a coverage sweep whose registry arm is currently a tautology, both non-blocking and called out in the comment.

PR #6291, opened 2026-07-28
fix(langy): surface and self-heal GitHub App install failures

sergioestebance:
## For humans

Two related failures in the Langy ↔ GitHub App integration:

1. If installing the app failed part-way through — an expired install link, a session change mid-flow, a permission change, or GitHub reusing an installation id another account already owns — Settings > Integrations used to just show the plain "Install" button again, as if nothing had happened, even though the user had just approved the install on GitHub's side. Now the actual failure reason shows up as an error toast.
2. Separately: if a GitHub App installation gets uninstalled on GitHub's side but our record of it is never cleaned up (a missed webhook delivery, or an installation that predates the webhook being configured at all), Langy kept trying to use that dead installation forever — minting silently failed and the user was told to "Install the LangWatch GitHub App" even though it was already installed with a different, live installation. This is now self-healing: a confirmed "this installation doesn't exist" response from GitHub removes the stale record and Langy falls back to the next real installation instead of failing the turn.

## What changed

**Error surfacing.** The `/setup` callback (`src/server/routes/github-langy.ts`) already reported every failure by redirecting back with `?githubError=<message>` — expired/tampered signed state, a session change mid-flow, a stale permission/membership check, or the cross-tenant installation-conflict guard. `src/pages/settings/integrations.tsx` never read that query param, so the error was silently dropped and the page just re-rendered the "Install" button.

`specs/langy/langy-github-install.feature` already documents the expected behavior ("Then I am shown that the installation could not be verified") — only the backend half of that scenario was tested; the UI half was unimplemented.

Fix: read `router.query.githubError` on mount, show it as an error toast, and strip it from the URL via `router.replace` so a refresh doesn't re-show it.

**Self-healing stale installations.** `LangyGithubInstallationsService.mintTurnToken()` (`src/server/app-layer/langy/langy-github-installations.service.ts`) always picked the oldest non-suspended installation for an org, with no check that it was still alive on GitHub. The mint call would fail against a dead installation, the error was swallowed, and the turn surfaced the generic "not connected" state — indistinguishable from never having installed the app at all.

Fix: `LangyGithubAppTokenService.mintInstallationToken()` (`src/server/app-layer/langy/langyGithubAppToken.ts`) now throws a distinct `GithubInstallationNotFoundError` on a confirmed GitHub 404 (never on any other status — a transient failure must not be treated as "this installation is gone"). `mintTurnToken()` catches that specific error, removes the stale row via the existing `deleteByInstallationId()` repository method, and retries the next candidate installation instead of failing the whole turn. This is a bug fix (restoring correct selection behavior), not new user-facing capability, so no feature-file scenario was added — the fix is proven by regression tests instead.

A cached token doesn't skip this either: a warm cache hit re-confirms the installation is still alive, but only once per 5 minutes (a separate, short-TTL marker, not the token's own ~50min cache) — otherwise every single turn would cost its own GitHub API call, and a GitHub outage would stall every turn on that call's timeout before falling back to the cached token. A transient probe failure backs off for 60s rather than retrying on the very next turn. A confirmed 404 still bypasses all of that and propagates immediately.

**Design note: `GithubInstallationNotFoundError` is a plain `Error`, not the repo's `HandledError`.** The mental model: `HandledError` is a contract that the error *will* reach a customer — every code it declares is required to have real customer-facing copy (enforced by a codebase-wide test). This error never crosses that boundary; it's thrown, caught a call or two later in the same service, and used purely to decide which installation candidate to try next. Making it a `HandledError` would mean either writing fake customer copy for a code that must never be shown, or working around the registry that enforces this — both worse than the one-line trade-off of losing automatic trace/span capture on an error nothing ever serializes.

## Test plan
- [x] `pnpm typecheck` / `pnpm typecheck:tests` clean
- [x] `integrations-github-error.integration.test.tsx`: shows a toast with the error message, strips `githubError` from the URL, and stays silent when there's no error param
- [x] `langy-github-installations.service.unit.test.ts`: oldest installation is a zombie (404) + a newer one is live → mints via the live one and removes the dead row; same for the explicit-repo path; every installation dead → returns null without crashing; a non-404 mint failure does NOT delete the row (transient-error safety)
- [x] `langyGithubAppToken.unit.test.ts`: a 404 from GitHub throws `GithubInstallationNotFoundError`, distinct from other failures; a cache hit re-confirms liveness without a stampede (concurrent + sequential calls each cost one GitHub call, not one per call); the liveness marker expires and re-probes; a transient probe failure backs off instead of probing every following call
- [ ] Manually verify in the browser against a real GitHub App install flow

## Summary by CodeRabbit

* **Bug Fixes**
  * Show a clear GitHub installation error toast on the Integrations settings page when setup fails, and remove the error detail from the URL right after displaying it.
  * Improve token minting resilience by automatically discarding stale “dead” GitHub installations (404) and continuing with remaining valid installations.
* **Tests**
  * Added integration and unit test coverage for the GitHub setup error toast + URL cleanup behavior.
  * Extended unit tests to verify 404 “installation not found” handling, cache behavior, and self-healing deletion of stale installation records.

### DIFF
--- a/langwatch/src/pages/settings/integrations.tsx
+++ b/langwatch/src/pages/settings/integrations.tsx
@@ -20,10 +20,12 @@ import {
   Text,
   VStack,
 } from "@chakra-ui/react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { GitHub } from "react-feather";
+import { useRouter } from "~/utils/compat/next-router";
 
 import SettingsLayout from "../../components/SettingsLayout";
+import { toaster } from "../../components/ui/toaster";
 import { withPermissionGuard } from "../../components/WithPermissionGuard";
 import { useOrganizationTeamProject } from "../../hooks/useOrganizationTeamProject";
 import { api } from "../../utils/api";
@@ -40,6 +42,27 @@ export default withPermissionGuard("organization:manage", {
 
 function IntegrationsContent({ organizationId }: { organizationId: string }) {
   const status = api.langyGithub.getInstallStatus.useQuery({ organizationId });
+  const router = useRouter();
+
+  useEffect(() => {
+    const githubError = router.query.githubError;
+    if (typeof githubError !== "string") return;
+
+    toaster.create({
+      type: "error",
+      title: "GitHub installation failed",
+      description: githubError,
+    });
+
+    const { githubError: _drop, ...rest } = router.query;
+    void router.replace(
+      { pathname: router.pathname, query: rest },
+      undefined,
+      { shallow: true },
+    );
+    // router is a fresh compat object each render; only re-run when the error itself changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [router.query.githubError]);
 
   const onInstall = () => {
     const ret = encodeURIComponent("/settings/integrations#github");
--- a/langwatch/src/pages/settings/integrations.tsx
+++ b/langwatch/src/pages/settings/integrations.tsx
@@ -58,7 +58,7 @@ function IntegrationsContent({ organizationId }: { organizationId: string }) {
     void router.replace(
       { pathname: router.pathname, query: rest },
       undefined,
-      { shallow: true },
+      { shallow: true, scroll: false },
     );
     // router is a fresh compat object each render; only re-run when the error itself changes.
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/langwatch/src/pages/settings/integrations.tsx
+++ b/langwatch/src/pages/settings/integrations.tsx
@@ -55,11 +55,10 @@ function IntegrationsContent({ organizationId }: { organizationId: string }) {
     });
 
     const { githubError: _drop, ...rest } = router.query;
-    void router.replace(
-      { pathname: router.pathname, query: rest },
-      undefined,
-      { shallow: true, scroll: false },
-    );
+    void router.replace({ pathname: router.pathname, query: rest }, undefined, {
+      shallow: true,
+      scroll: false,
+    });
     // Intentionally keyed only on the error value, not the whole router object.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [router.query.githubError]);
--- a/langwatch/src/pages/settings/integrations.tsx
+++ b/langwatch/src/pages/settings/integrations.tsx
@@ -60,7 +60,7 @@ function IntegrationsContent({ organizationId }: { organizationId: string }) {
       undefined,
       { shallow: true, scroll: false },
     );
-    // router is a fresh compat object each render; only re-run when the error itself changes.
+    // Intentionally keyed only on the error value, not the whole router object.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [router.query.githubError]);
 
--- a/langwatch/src/server/app-layer/langy/langy-github-installations.service.ts
+++ b/langwatch/src/server/app-layer/langy/langy-github-installations.service.ts
@@ -280,8 +280,9 @@ export class LangyGithubInstallationsService {
    * installation can predate the webhook being configured at all), leaving a
    * `LangyGithubInstallation` row for an installation GitHub has already
    * forgotten. Rather than let that dead row block every real installation
-   * behind it forever, a confirmed 404 from GitHub on mint removes the row and
-   * moves on to the next candidate.
+   * behind it forever, a confirmed 404 from GitHub — whether it surfaces
+   * while minting or while resolving an explicit repo id — removes the row
+   * and moves on to the next candidate.
    *
    * TODO(JIT narrowing): the plan's delivery option 2 replaces spawn-env
    * with a clone-time credential-helper callback that mints per-clone; the
@@ -301,12 +302,20 @@ export class LangyGithubInstallationsService {
     if (usable.length === 0) return null;
 
     // Explicit repo: pick the installation that can reach it and scope to it.
-    // A candidate that turns out to be a dead installation (self-healed away
-    // by mintScoped) is skipped in favor of the next one that can reach the
-    // same repo, rather than failing the turn outright.
+    // A candidate that turns out to be a dead installation — whether that
+    // surfaces while resolving the repo id (an uncached "all" selection has to
+    // list live) or while minting — is self-healed away in favor of the next
+    // one that can reach the same repo, rather than failing the turn outright.
     if (repositoryFullName) {
       for (const inst of usable) {
-        const repoId = await this.resolveRepositoryId(inst, repositoryFullName);
+        let repoId: string | null;
+        try {
+          repoId = await this.resolveRepositoryId(inst, repositoryFullName);
+        } catch (error) {
+          if (!(error instanceof GithubInstallationNotFoundError)) throw error;
+          await this.markInstallationDead(inst.installationId);
+          continue;
+        }
         if (!repoId) continue;
         const outcome = await this.mintScoped({
           installationId: inst.installationId,
@@ -354,11 +363,7 @@ export class LangyGithubInstallationsService {
       };
     } catch (error) {
       if (error instanceof GithubInstallationNotFoundError) {
-        logger.warn(
-          { installationId },
-          "github installation no longer exists, removing stale record",
-        );
-        await this.repo.deleteByInstallationId(installationId);
+        await this.markInstallationDead(installationId);
         return { token: null, wasDeadInstallation: true };
       }
       logger.warn(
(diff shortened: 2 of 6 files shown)


sergioestebance:
Fixed in 0770c65c8: `router.replace` now passes `{ scroll: false }` so the `githubError` cleanup no longer yanks the user back to the top of the page away from the `#github` card.

sergioestebance:
Fixed in 0770c65c8: the cleanup test now applies the mocked `replace()` to the router query and rerenders, asserting `toaster.create` is still only called once — proving the toast doesn't repeat once the URL is clean, not just that `replace` was invoked.

langwatch/src/pages/settings/integrations.tsx
```diff
@@ -40,6 +42,27 @@ export default withPermissionGuard("organization:manage", {
...
+    });
+
+    const { githubError: _drop, ...rest } = router.query;
+    void router.replace(
+      { pathname: router.pathname, query: rest },
+      undefined,
+      { shallow: true, scroll: false },
+    );
+    // router is a fresh compat object each render; only re-run when the error itself changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
sergioestebance: Fixed in d855906ac — the comment claimed useRouter() returns a fresh object each render; it memoizes. Reworded to describe what the effect is actually keyed to.

langwatch/src/server/app-layer/langy/langy-github-installations.service.ts
```diff
@@ -297,25 +305,35 @@ export class LangyGithubInstallationsService {
...
+        if (!repoId) continue;
+        const outcome = await this.mintScoped({
+          installationId: inst.installationId,
+          repositoryIds: [repoId],
+        });
+        if (outcome.token) return outcome.token;
+        if (!outcome.wasDeadInstallation) return null;
       }
       // The App is not installed on that repo — bounded by the installation.
       return null;
```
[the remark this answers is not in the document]
sergioestebance: Real gap, fixed in d855906ac. resolveRepositoryId rethrows GithubInstallationNotFoundError instead of swallowing it; mintTurnTokens explicit-repo loop now catches it, deletes the row via the same markInstallationDead helper mintScoped uses, and continues to the next candidate. Added a regression test for the exact all-selection/uncached case you called out (langy-github-installations.service.unit.test.ts).

langwatch/src/server/app-layer/langy/langy-github-installations.service.ts
```diff
@@ -292,26 +301,72 @@ export class LangyGithubInstallationsService {
...
+  // Mint against the org's installation(s) scoped to the full repo set,
+  // oldest first. An installation GitHub confirms is gone (404) is removed
+  // and the next candidate is tried instead of failing the whole turn; any
+  // other mint failure stops here — a transient error must not make us skip
+  // past a live installation.
+  private async mintForAnyInstallation(
+    usable: LangyGithubInstallationRow[],
+  ): Promise<LangyGithubTurnToken | null> {
+    for (const inst of usable) {
+      const outcome = await this.mintScoped({
```
langwatch-agent, line 363: **P1 — A cached token prevents the stale-installation detection and retry from running.** `mintInstallationToken()` returns Redis entries before calling GitHub (`langyGithubAppToken.ts:211`). If an installation is uninstalled while its 50-minute cache entry exists and the deletion webhook was missed, GitHub deactivates the installation token, but this loop receives that cached token as success and hands it to the worker. The worker then fails against the removed installation; no `GithubInstallationNotFoundError` is raised here, the stale row remains, and the newer live installation is never tried until cache expiry. This is the primary missed-webhook scenario the PR claims to self-heal. Add a liveness probe/invalid-token feedback path before returning a cached candidate (or otherwise invalidate and retry on the worker's GitHub 401/404), and cover an uninstall with a cached token plus a live fallback installation.

langwatch-agent (dismissed):
Request changes: 1 P1 — see inline comment. The uncached-404 retry is correct, but the existing installation-token cache bypasses it after an uninstall, leaving the reported failure mode in place.

langwatch/src/server/app-layer/langy/langy-github-installations.service.ts
sergioestebance: Confirmed, real gap — fixed in fb2f6aa8c. mintInstallationToken() now confirms the installation still exists (via getInstallation(), a confirmed 404 -> GithubInstallationNotFoundError) before trusting a cache hit; any other liveness-probe failure fails open and still serves the cached token. Added regression tests for both the stale-cache-heals case and the fail-open case.

langwatch/src/server/app-layer/langy/langyGithubAppToken.ts
```diff
@@ -191,7 +227,16 @@ export class LangyGithubAppTokenService {
...
     const cached = await this.redisGet(cacheKey);
-    if (cached) return { token: cached, expiresAt: "" };
+    if (cached) {
+      // A cache hit alone doesn't mean the installation is still alive: the
+      // token's ~50min TTL comfortably outlasts a missed deletion webhook, so
+      // without this check a dead installation would keep being served a
+      // "valid" token — and every self-heal path in the caller (which only
+      // ever sees GithubInstallationNotFoundError from a REAL mint attempt)
+      // would never fire until the cache happened to expire.
+      await this.assertInstallationStillExists(args.installationId);
```
langwatch-agent, line 276: 

langwatch-agent (dismissed):
Re-review: the prior stale-installation bypass is fixed, but 1 P1 remains — see the inline comment.

langwatch/src/server/app-layer/langy/langyGithubAppToken.ts
sergioestebance: The comment on this thread came through empty (no text), so I went by the diff hunk plus the review summary and read the code directly: the new cache-hit liveness check had no stampede protection, unlike the mint path right next to it, which already guards its own far rarer GitHub call the same way. Fixed in 3cf47a967 -- a non-blocking, try-once lock (not the mint paths retry-for-3s one, since a losing caller here has a valid immediate fallback: trust the cache). Added a concurrency regression test (5 concurrent calls, 1 GitHub liveness call). If the empty comment was pointing at something else, flag it again and I will take another pass.

langwatch/src/server/app-layer/langy/langyGithubAppToken.ts
```diff
@@ -191,7 +241,16 @@ export class LangyGithubAppTokenService {
...
```
langwatch-agent, line 276: **P1 - Cache successful liveness results instead of probing GitHub on every warm turn**

The lock only coalesces overlapping callers. It is released immediately after the GET, so sequential cache hits take the lock and synchronously call `GET /app/installations/:id` again. `LangyCredentialService` awaits this path while provisioning each Langy turn, and `githubFetch` permits this request to wait up to 10 seconds.

For a normally active installation, this turns the 50-minute installation-token cache into one GitHub App API request per sequential chat turn. During a GitHub network outage or rate-limit response, every sequential turn waits on the liveness timeout before the catch returns the cached token; under enough traffic it also consumes the GitHub App API rate budget and creates the very availability dependency the token cache was meant to remove. The new concurrency test covers only five overlapping calls, so it misses the production case.

Store a separate, short bounded "installation was live" marker after a successful probe, and probe only when that marker expires (with a short failure backoff that continues to fail open). Keep a confirmed 404 as the only eviction signal. Add tests for multiple *sequential* cached calls (one GET during the liveness TTL), marker expiry, and a transient probe failure not causing a probe on every following turn.

langwatch-agent (dismissed):
Re-review: the concurrent-probe lock is an improvement, but it does not resolve the per-turn GitHub dependency. One P1 remains.

langwatch/src/server/app-layer/langy/langyGithubAppToken.ts
sergioestebance: Confirmed, real gap -- the stampede lock only ever deduped truly overlapping calls; sequential turns still each re-probed. Fixed in 194e79bfd: a confirmed-alive result is cached under its own 5-minute marker (a cache hit skips the probe entirely while it holds), a transient probe failure sets a 60s backoff marker instead of leaving every following turn to hit githubFetch is 10s timeout, and a confirmed 404 still propagates immediately and unconditionally -- it is the only eviction signal. Added the three tests you asked for: sequential cached calls (1 GET across all of them), marker expiry (2nd GET after simulated expiry), and transient-failure backoff (no GET on the immediate next call).

langwatch-agent (commented):
Re-review complete. The stale-installation handling is now coherent: cached tokens receive a bounded liveness probe, transient probe failures back off and fail open, and confirmed 404s delete the stale row before trying the next installation. I found no additional actionable correctness or security issue on this head.

Aryansharma28 (approved):

Issue #6296, opened 2026-07-28
The evaluation-status enum pin cannot fail — #5395's compile-time link does not exist


drewdrewthis:
## For Humans

**A test that exists to catch a change compares a list to itself, so it can never fail — and two comments in the codebase state the opposite.** #5395 shipped believing it had added a compile-time link between the evaluation-status enum and the trigger filter options. It had not. If someone adds or renames an evaluation status, nothing fails, and the filter options silently drift — which is the exact class of defect #5395 was fixing.

**Ask:** replace the pin with one that can fail, correct the two false comments, and sweep for the same shape elsewhere.

<details><summary>For Agents — full spec</summary>

## The defect

`langwatch/src/server/filters/__tests__/evaluation-filter-wiring.unit.test.ts:78-82`:

```ts
// Pins the option list to evaluationRunDataSchema. If the enum changes,
// this fails and forces a deliberate update to both places.
const options = extractResults([]);
expect(options.map((option) => option.field)).toEqual(
  evaluationRunDataSchema.shape.status.options,
);
```

And `extractResults` (`langwatch/src/server/filters/clickhouse/filter-definitions.ts:537`):

```ts
return evaluationRunDataSchema.shape.status.options.map((status) => ({
  field: status, label: status, count: observedCounts.get(status) ?? 0,
}));
```

The assertion reduces to `enum.map(s => s) === enum`. Both sides come from the same import, so an enum change moves both together and the test stays green. **It cannot fail.**

## Two comments assert the opposite

- The test's own comment above: *"If the enum changes, this fails and forces a deliberate update to both places."*
- `filter-definitions.ts:522-527`: *"Deriving from the schema keeps the two compile-time linked: an enum change fails CI here."*

Both are false. #5395's PR description repeats the claim. A false comment is worse than no comment — it tells the next engineer a guard exists where none does.

## What would actually work

An exhaustive type-level check (`satisfies Record<EvaluationStatus, …>`) or a committed snapshot of the option list — something with an independent second source. Whatever is chosen, demonstrate it **red** against a locally-added enum member before calling it done.

## Likely wider than one site

This is a shape, not a one-off: *a derived list asserted equal to the source it derives from*. Worth a sweep of `src/server/filters/` and any other module using the "derive options from the schema enum" technique.

Found while working #4805, which depends on this guard actually existing — its AC11 now requires a mechanism demonstrated red, and carves this file out of its regression no-diff clause so it can be repaired.

Related: #5395, #4805.

</details>

Issue #4805, opened 2026-06-12
Alert incorrectly triggers on thumbs-down evaluator; Error_Message evaluator alerts not firing


drewdrewthis:
## For Humans

**Every alert that was set up with the old "Error_Message" status option is still silently dead — the code fix shipped, but nobody repaired the alerts customers had already saved.** Two defects were reported here: thumbs-down alerts firing on everything (fixed, #4903) and Error_Message alerts firing on nothing (code fixed, #5395). The second fix only stopped the *bad option from being offered* to new alerts. Alerts saved before 2026-07-13 still hold the dead value, still never fire, and still tell nobody.

**Ask:** repair the already-saved alerts, and stop a dead value being saveable at all.

- The dead value is still saveable today on **every** write path — the app, the JSON filter editor, the public REST API and the MCP tools. Verified, not inferred.
- The repair SQL drafted in the 2026-07-06 comment must not be run: wrong table name, and its blind text-replace corrupts unrelated filters — demonstrated against real Postgres, see Evidence.
- The dead-value set is wider than `Error_Message` — the code itself documents `succeeded`/`failed` as stored non-canonical statuses.
- Nothing anywhere reports that an alert can never fire. That silence is why this ran ~4 months undetected.

_Status: `bug`, `P1 - high`, `investigating`. Both code defects merged; this issue tracks the unshipped repair. Full spec below._

<details><summary>For Agents — full spec</summary>

### Original report (preserved)

Two related alert evaluation issues:

1. **False positive**: alerts incorrectly triggering on a thumbs-down evaluator when they should not.
2. **Silent failure**: a separate Error_Message type evaluator stopped sending alerts entirely — no notifications dispatched even when the condition is met.

Reproduction as reported: configure an alert on a thumbs-down evaluator, observe spurious triggers; configure an alert on an Error_Message evaluator, observe that no alert fires when an error message is present. Impact: operators receive false positives from thumbs-down and miss real errors from Error_Message, degrading monitoring coverage. Customer-reported via Slack, ongoing since ~March 2026.

---

**BLUF: the two code defects in this issue have shipped (#4903, #5395). What is left is a data problem the PRs could not fix — `Trigger.filters` rows already persisted with a non-canonical `evaluations.state` value. #5395 changed only where the dropdown *sources* its options; it did not repair stored rows, and nothing on the write path rejects such a value, so the API can still persist one today. Every affected automation is permanently silent with zero operational signal.**

[...]

PR #6299, opened 2026-07-28
fix(automations): refuse non-canonical evaluations.state on every write path (#4805)

drewdrewthis:
> **Scariest thing** — the domain check is keyed on the filter field name, and `filterValueSchema` (`src/server/filters/types.ts:37-45`) is **field-agnostic and shared by both filter schemas**. Key it on the prefix `evaluations.` instead of the exact key `evaluations.state`, or narrow `filterValueSchema` itself, and every automation filtering on `metadata.value`, `topics.topics` or `evaluations.label` starts failing to save. A regression test pins exactly that.
> **Start here** — `src/server/filters/types.ts`: `findOffendingEvaluationStateEntries` and the `.superRefine` wired onto both schemas. That one function is the whole contract; everything else calls it.

## Why

Automations configured with a stale `evaluations.state` value — `"Error_Message"`, a phantom left in a dropdown sourced from a since-removed Elasticsearch index — can never match `EvaluationRunData.status`, whose canonical domain is exactly `scheduled | in_progress | processed | error | skipped`. Those automations are **permanently silent** and nothing says so.

#5395 removed the phantom from the dropdown. It never repaired a single stored row, and **nothing rejected the value on write** — so it stayed saveable on every path right up to this PR.

Refs #4805. This PR deliberately does **not** close it: it delivers the write-side half only. See *Deferred and blocked* below.

## What changed

- **Domain check at the record level, not inside `filterValueSchema`** → the alternative was narrowing the shared value schema; that would apply the five-value status domain to *every* filter field and break `metadata.value`, `topics.topics`, `evaluations.label`. Consequence if this call is wrong: silent, broad save failures across unrelated automations — which is why an explicit non-regression fixture covers all three arities.
- **Guard at the repository seam, and every caller routed through it** → the alternative was guarding only the tRPC input schemas; that leaves the public REST endpoint (`z.record(z.unknown())`, persisted verbatim), the MCP `platform_*_trigger` tools that ride it, and a legacy Slack endpoint wide open. Consequence: the fix would have looked green while most write paths stayed broken.
- **A `HandledError`, not a bare `ZodError`** → a `ZodError` is not a `HandledError`, so `errorDisplayMessage` (`src/utils/trpcError.ts:23-37`) falls back to the raw wire message and the drawer renders a generic "Could not save automation". Consequence: an operator would see a failure with no idea which field or what values are legal.
- **A report-only detection migration, no rewrite** → the alternative was mapping `Error_Message → error` in the same PR; that mapping is a semantic inference and a wrong one turns a silent automation into an alert storm for that project. Consequence of deferring: affected automations stay silent until someone with prod access supplies the value distribution — visible in a table rather than invisible in the data.
- **Canonical set derived from `evaluationRunDataSchema.shape.status.options`, never a literal** → #6296 exists because the previous "compile-time link" here compares a derived list to the list it derives from and cannot fail. This one has a single source of truth.

## How it works

```
src/server/filters/types.ts
  findOffendingEvaluationStateEntries   single source of truth: exact-key lookup
  rejectNonCanonicalEvaluationState     .superRefine on both filter schemas
  sanitizeTriggerFilters                drops offending entries (key removed if nothing canonical remains)
src/server/filters/evaluationStateFindings.ts
  findNonCanonicalStateValues           per-(evaluatorKey, value) findings; total by contract, never throws
src/server/app-layer/automations/errors.ts
  InvalidEvaluationStateFilterError     HandledError naming field, rejected value, all five canonical values
src/server/app-layer/automations/repositories/trigger.prisma.repository.ts
  create/update -> assertCanonicalEvaluationStateFilters, reusing parseFilters
```

`filters` reaches the seam in **two encodings** — a JSON string from tRPC/REST, a raw object from `graph-alert.builder.ts:110`. The guard handles both; a string-only check would silently pass the object path.

**Six write paths, not five.** The issue's own analysis enumerated five. Implementation found a sixth: `src/server/routes/misc.ts:931`, a legacy `POST /api/trigger/slack` calling `prisma.trigger.create` directly with its own duplicated filter schema. Verified on `origin/main`. It and the REST handlers now go through the trigger service, so all six converge on the guarded seam.

### The migration

```
prisma/migrations/20260728120000_report_noncanonical_evaluation_state/migration.sql
  TriggerFilterFinding          sidecar table, no schema.prisma model (ReactorOutbox precedent)
  _tff_try_parse_jsonb          migration-scoped plpgsql helper, created and dropped in-file
  insert 1                      one reported_unmappable row per (trigger, evaluatorKey, value)
  insert 2                      one reported_malformed row per un-normalisable trigger
```

Postgres 16 has no try-cast (`IS JSON` arrived in 17), hence the helper: a single row whose inner text does not parse must not abort the statement.

## Deployment Impact

- **Migration is additive and report-only.** `TriggerFilterFinding` is a brand-new sidecar table with no `schema.prisma` model (same shape as the existing `ReactorOutbox` precedent) — it does not touch `Trigger` or any other existing table's schema, and the migration performs zero `UPDATE`/`DELETE` against `Trigger` rows. The xmin-unchanged proof in *How I can prove I was successful* is the guarantee for this. Rollback is a plain `DROP TABLE TriggerFilterFinding` with no data-loss exposure elsewhere.
- **Runtime behavior change, no flag.** Every write path that creates or updates a `Trigger` (tRPC, REST, MCP `platform_*_trigger` tools, the legacy `/api/trigger/slack` handler) now rejects a non-canonical `evaluations.state` value with a `422` (`InvalidEvaluationStateFilterError`, code `invalid_evaluation_state_filter`). This is unconditional — no env var, no feature flag, no gradual rollout. Any client currently saving a stale value (e.g. `Error_Message`) on any of those six paths will start seeing a hard rejection at deploy time, not a soft warning.
- **Existing rows are untouched and ungated.** An automation that already holds a non-canonical `evaluations.state` value keeps working exactly as before for every *other* field (rename, retarget, disable) — the grandfathering tests in `trigger-filters-grandfathering.unit.test.ts` cover this explicitly. Only a *new* non-canonical value, or a `create`, is rejected.
- **No new external dependencies, no config/env additions.** The canonical-value list is derived at import time from `evaluationRunDataSchema.shape.status.options`, not a new env var or feature-flag lookup.
- **Reversibility:** the guard itself can be reverted by reverting the code change (no migration to unwind for it); the sidecar-table migration is drop-only and does not need to be reverted to roll back the guard.

## Test plan

Regression-test-first, per the bug-PR audit trail — the failing tests are commit `2aedf45e`, the fix is `978935df`.

| Command | commit `2aedf45e` (tests only) | commit `978935df` (+ fix) |
|---|---|---|
| `pnpm test:unit run src/server/filters/__tests__/evaluation-state-*.unit.test.ts` | **2 files failed, 4 failed / 6 passed** | **2 files passed, 27 passed** |
| `pnpm test:unit run src/server/filters` (whole suite) | — | **9 files passed, 279 passed** |
| `pnpm typecheck:all` | — | **0 errors** |

The 10 → 27 collected-test jump is expected: at `2aedf45e` the phantom-detection file could not import `evaluationStateFindings` (it did not exist yet), so its cases never ran.

The 6 tests that pass at *both* commits are deliberate over-broad-fix guards: the five canonical values still save, and `{"metadata.value": …, "evaluations.label": …, "topics.topics": …, "events.event_details.key": …}` still accepts arbitrary strings.

## Human verification

For if you really don't trust the automated proof below — steps a skeptical human can follow to independently re-verify. **All rows UNVERIFIED** — not yet independently walked by a human against the running app; this is the checklist, not a claim any of it was done.

| # | Step | Expected result | Status |
|---|---|---|---|
| 1 | Boot the app, open an automation's filter drawer, add an `evaluations.state` condition set to a value outside `scheduled\|in_progress\|processed\|error\|skipped` (e.g. type a bogus value if the UI allows, or drive it via the tRPC/REST API directly with `{"evaluations.state":{"e1":["Error_Message"]}}`), attempt to save | Save is rejected with a `422` / `invalid_evaluation_state_filter` error naming the field, the rejected value, and the five canonical values | UNVERIFIED |
| 2 | Repeat step 1 against the REST trigger-create endpoint and the `platform_*_trigger` MCP tool directly (not just the UI) | Both reject the same way — confirms the guard is at the repository seam, not just the tRPC layer | UNVERIFIED |
| 3 | On a project with an automation that already has a non-canonical `evaluations.state` value (pre-existing, from before this PR), rename it or change its notification cadence and save | Save succeeds — the stale value is grandfathered through unrelated field changes | UNVERIFIED |
| 4 | On that same pre-existing automation, add a *second*, newly-introduced non-canonical value alongside the grandfathered one, save | Save is rejected, and the error names only the newly-introduced value, not the pre-existing one | UNVERIFIED |
| 5 | Apply `prisma/migrations/20260728120000_report_noncanonical_evaluation_state/migration.sql` against a copy of a real database with at least one automation holding a non-canonical `evaluations.state` value, then query `TriggerFilterFinding` | One `reported_unmappable` (or `reported_malformed`) row appears per offending `(trigger, evaluatorKey, value)`; the source `Trigger` row's `xmin` and `filters` column are byte-for-byte unchanged | UNVERIFIED |
| 6 | Run the migration a second time against the same database | Zero new `TriggerFilterFinding` rows are added (idempotent) | UNVERIFIED |

**proven** — falsifiability, identical command both sides:

falsifiability

The image is **rendered from this session's real output, not a literal screen capture** — raw text pasted here so the two can be compared:

```
# commit 2aedf45e6c — tests only, BEFORE the fix
 FAIL  evaluation-state-value-domain.unit.test.ts > sanitizeTriggerFilters
 AssertionError: expected '{"evaluations.state":{"eval-abc":["Er…' not to contain 'Error_Message'
 Test Files  2 failed (2)
      Tests  4 failed | 6 passed (10)

# commit 978935dfb8 — + the write-side guard
 Test Files  2 passed (2)
      Tests  27 passed (27)

# whole filters suite, no regressions
 Test Files  9 passed (9)
      Tests  279 passed (279)

# pnpm typecheck:all
 TYPECHECK2_EXIT=0     errors: 0
```

**proven** — the report-only migration, applied as a whole file to a real Postgres database exactly as `psql` would, over fixtures in both storage shapes:

```
=== REPORT-ONLY MIGRATION vs real Postgres ===
PASS  obj-phantom  findings          PASS  obj-phantom  shape=object
PASS  str-phantom  findings          PASS  str-phantom  shape=string
PASS  canonical    findings          PASS  unrelated    findings
PASS  sibling      findings          PASS  two-keys     findings
PASS  two-keys     shape=string      PASS  nostate      findings
PASS  deleted      no findings       PASS  malformed    reported
PASS  ZERO writes to Trigger (xmin)
PASS  idempotent (5 -> 5)
--> 14/14
```

`xmin` is the report-only guarantee and it is deliberate: `updatedAt` is Prisma-client-side `@updatedAt`, so raw SQL never bumps it on *any* row — asserting on it would be structurally incapable of failing.

**claimed, not proven** — the committed integration test (`evaluation-state-report-migration.integration.test.ts`) **has not been run**. The integration harness requires ClickHouse and it is not installed on this box (`8123` refused; no `clickhouse-server` unit). CI exercises it. The 14/14 above is a direct probe of the same shipped `migration.sql`, not a substitute for that test.

**claimed, not proven** — the guard has not been exercised against the running app. No user-observable surface was driven: I did not boot the stack and attempt a save through the automations drawer, so the rendered error-message path (`InvalidEvaluationStateFilterError` → `errorDisplayMessage` → drawer toast) is reasoned from source, not observed. AC9 requires a rendered-surface assertion and it is **not** ticked by this PR.

**typecheck caveat worth knowing** — a first `typecheck:all` reported 2,285 errors. All of them cascaded from an ungenerated Prisma client in a fresh worktree, in files this PR never touches. After `pnpm start:prepare:files`: 0 errors. Flagging so the number is not rediscovered as alarming.

## Deferred and blocked

The issue's plan splits the work so the safe half ships alone. This PR is the write-side guard. Remaining, with status:

| AC | Status |
|---|---|
| AC5b — evidence the `Error_Message → error` mapping is correct | **BLOCKED ON PROD ACCESS.** Needs the real value distribution from a database that actually holds affected rows. Independently confirmed unreachable from this box: kubeconfig is a 28-byte stub and the AWS identity is `381491922238`, not prod `339712859611`. Until then AC5c (report-only) governs and no value is rewritten. |
| AC21 — post-migration proof on real data | **BLOCKED ON PROD ACCESS**, same reason. |
| AC2, AC3, AC5, AC6, AC13, AC15, AC16, AC20 | Deferred per AC5c(iii) — they describe the rewrite half, which does not ship without AC5b. |
| AC24, AC1, AC4, AC12, AC14 | Report-only migration — **shipped in this PR**, proven 14/14 against real Postgres. |
| AC7 (preview), AC23 (100k-row boot budget) | Not ticked — no preview-vs-report cross-assertion yet, and no seeded-scale timing run. |
| AC9 (rendered surface), AC10, AC10b, AC17, AC18, AC19, AC22 | Not yet ticked. |

A wrong `Error_Message → error` mapping would turn a silent automation into an alert storm for that project, so AC5b gating the rewrite is deliberate, not incidental.


drewdrewthis:

CI: green (zero failing, zero pending) at head `9051f270c0`
ACs: met for the report-only detection/refusal scope this PR ships; feature-parity binds the 12 scenarios this PR covers in `specs/automations/evaluation-state-filter-repair.feature` (repair/rewrite and preview-surface scenarios remain `@unimplemented` — out of scope for this PR)
Threads: zero unresolved, zero outdated (0 review threads total)

Verified by:
  gh pr checks 6299 --repo langwatch/langwatch (post-settle) → 0 fail, 0 pending, all pass/skip
  gh api repos/langwatch/langwatch/actions/jobs/<mocking-job-id> -q '{status,conclusion}' → {"conclusion":"success","status":"completed"}
  gh api repos/langwatch/langwatch/actions/jobs/<langwatch-app-complete-id> -q '{status,conclusion}' → {"conclusion":"success","status":"completed"}
  gh api graphql (reviewThreads first:50) → {"data":{"repository":{"pullRequest":{"reviewThreads":{"nodes":[]}}}}}
  cd platform/app && npx --yes tsx scripts/check-feature-parity.ts → "specs/automations/evaluation-state-filter-repair.feature: 12/12 scenarios bound · ✓ all bound", exit 0
  gh pr view 6299 --json headRefOid -q .headRefOid → 9051f270c02b960496aed6925858d4064cb0f243

drewdrewthis:
## Review verdict: READY

Independently reviewed (review-clerk, not the PR author). CI green, 0 unresolved review threads, diff checked for correctness and scope. No blocking findings.

**CI:** All gating checks completed SUCCESS or SKIPPED (enumerated via `fetch_checks_rest` — 70 success / 54 skipped / 4 cancelled, the 4 cancelled each have a matching SUCCESS run for the same check name from a later push, so not blocking). No FAILURE conclusions on any check.

**Draft status:** PR is draft — expected, not a finding. Un-drafting is the ship step.

<details>
<summary>What I checked in the diff</summary>

- **Migration additivity / xmin guarantee** (`prisma/migrations/20260728120000_report_noncanonical_evaluation_state/migration.sql`): confirmed report-only — `CREATE TABLE IF NOT EXISTS "TriggerFilterFinding"` (new sidecar table, no `schema.prisma` model) plus two `INSERT ... SELECT` statements scanning `"Trigger"`. No `UPDATE`/`DELETE` against `Trigger` anywhere in the file. Idempotency guarded by `NOT EXISTS` dedup checks on both inserts. Exact-key match (`object_filters ? 'evaluations.state'`), never a prefix match — matches the PR body's stated scoping.
- **Runtime rejection with no flag**: `assertCanonicalEvaluationStateFilters` (`trigger.prisma.repository.ts`) is called unconditionally from both `create` and `update`, no env/flag gate — matches the PR body's claim.
- **Grandfathering**: `update` path has a distinct guard (visible in the diff, separate function following `assertCanonicalEvaluationStateFilters`) that diffs against the existing stored row rather than rejecting outright — avoids the "AutomationDrawer resends filters unchanged" lockout the PR body describes.
- **Sixth write path** (`src/server/routes/misc.ts:931`, legacy `/api/trigger/slack`): confirmed changed from a direct `prisma.trigger.create(...)` call to `getApp().triggers.create(...)`, routing it through the guarded repository seam — matches the PR body's claim that this path was found and fixed beyond the issue's original five.
- **`filterValueSchema` left alone**: the domain check (`findOffendingEvaluationStateEntries` / `collectOffendingStates` in `filters/types.ts`) is a standalone function operating on `evaluationStateValue`, not wired into the shared `filterValueSchema` — confirms the stated design choice to avoid breaking `metadata.value` / `topics.topics` / `evaluations.label`.
- **Error surface**: `InvalidEvaluationStateFilterError extends HandledError` (not a bare `ZodError`), carries `httpStatus: 422`, `code: "invalid_evaluation_state_filter"`, and a message naming the field, offending value, and all canonical values.
- **Tests present**: unit tests for value-domain rejection, phantom detection, and grandfathering; an integration test for the migration (marked in the PR body as not locally run — ClickHouse unavailable on the author's box — but present and CI-exercised).
- **No secrets**: no credentials, tokens, or env values found in the diff.
- **Scope**: diff also touches mirrored files under `platform/app/...` alongside the `langwatch/...` originals — this is pre-existing repo structure (both trees exist at HEAD, not introduced by this PR) rather than unrelated scope creep; noted for awareness, not a finding.

</details>

**Note:** Deployment Impact and Human Verification sections in the PR body are already thorough and self-consistent with the diff; Human Verification rows are explicitly marked UNVERIFIED by the author (not yet walked live) — that is disclosed, not a defect, and doesn't block review readiness.

PR #6301, opened 2026-07-28
feat(charts): pass strict admission control (read-only root, automount, seccomp, limits)

0xdeafcafe:
Supersedes #4967, which was 627 commits behind `main` and no longer mergeable. Same goal, rebased, with the review findings actioned.

Follow-up to #4927, and supersedes #4948 (folded in).

## Why

LangWatch's Helm charts didn't install on clusters that enforce strict admission control (Pod Security Admission `restricted`, or an equivalent Gatekeeper / Kyverno bundle). #4927 closed most of it. This closes the rest, including the bundled datastores, and corrects what the chart claims about itself.

## What changed

**Pods**
- postgresql / redis / clickhouse / keeper: `readOnlyRootFilesystem: true` with `emptyDir` scratch for the runtime paths outside the data volume (Postgres' socket dir, ClickHouse's server logs, `/tmp`).
- clickhouse / keeper: `automountServiceAccountToken` set on the pod spec, not just the ServiceAccount, for policies that inspect the pod.
- backup/restore Jobs: read-only root, `/tmp` scratch, pod-level automount, and the chart's own ServiceAccount — they previously ran under the namespace default SA, so an IRSA annotation on the chart SA never reached the one component that talks to S3.
- app: removed the dead `fix-nextjs-permissions` init container and its `next-dir` volume. The app is Vite now; nothing reads or writes `.next`.
- Scratch volumes are size-bounded (`scratch.logsSizeLimit` 2Gi, `scratch.tmpSizeLimit` 1Gi). They are backed by node ephemeral storage, so an unbounded one makes a single pod's error loop everyone else's problem.

**Security-context overrides now layer, not replace**

A per-component `podSecurityContext` / `containerSecurityContext` override merges over the global default: the key you set wins, everything you didn't mention keeps its hardened value. `coalesce` is all-or-nothing and cannot express "override this one field", so two shared helpers do the merge and all ten call sites use them.

PostgreSQL and Redis consumed neither global context, which is why cross-cutting flags had to be retyped into those two files by hand. They now take `global.containerSecurityContext` through the same helper, keep only their image's uid at pod level, and gain per-component `nodeSelector` / `tolerations` / `affinity` so the datastores can be pinned without moving every other pod.

**The strict-admission overlay**

- Drops `preflight.enabled` — the hook no longer exists on `main`.
- Adds `langyagent.chartManaged: false`. The assistant is chart-managed by default and its manager runs as root with `CHOWN`/`DAC_OVERRIDE`/`FOWNER`/`SETUID`/`SETGID` so it can hand each worker a distinct uid; on a restricted cluster that pod is rejected. It opts the pod out rather than de-privileging it — an unprivileged manager cannot do the per-uid handoff that keeps one worker out of another's credentials (ADR-033, `specs/langy/langy-deploy-hardening.feature`).

**Docs**

`security.mdx` claimed the defaults pass `restricted` "with no extra config" and then listed the required config seven lines later. It now says a default install does not pass on its own, names the two components that cannot comply, and leads with the overlay. Also removed the stale `.next` reference, corrected the uid example (the datastores run 999/101/65532, so a `MustRunAs` written from that block would have denied them), and dropped the ClickHouse README's claim that a standalone install clears read-only-root — its preflight Job deliberately doesn't.

**Tests**

`test_pod_security` grepped the whole render, so "is `readOnlyRootFilesystem` present?" passed on one container out of ten, and "at least 10 `runAsNonRoot` sites" passed when a workload regressed as long as another gained a field — removing the container-level `runAsNonRoot` the suite exists to guard left it green. Each workload is now rendered with `--show-only` and asserted individually. Prometheus and the assistant are explicit exemptions carrying their reason and the key that removes them, and a new check fails on any workload that is neither hardened nor exempt, so a future subchart can't ship unhardened unnoticed.

Adds `specs/security/helm-strict-admission.feature`, which didn't exist.

## Also carries a commit orphaned by #5524

#5524 was squash-merged as `4c182f0f5`, but its last commit — `docs(workers): describe how the worker registry is actually scraped` — was authored about two hours after the squash and never made it in. Four of that branch's five commits are in `main` byte-for-byte; that one is not. It is cherry-picked here.

It corrects two comments in `startWorkers.ts` that claimed the web process scrapes the worker registry at `GET /workers/metrics`. In a split deployment it cannot: that proxy dials its own loopback, so it only resolves when the workers share the web process. The chart's Prometheus scrape config targets the worker pod directly on the metrics port.

## Also: binds the worker-probe specs so the parity gate enforces them

Post-merge follow-up to #5524 (originally opened as #6314, folded in here).

`specs/server/worker-liveness-probe.feature` and `specs/server/metrics-collection.feature` shipped **untagged**. `check-feature-parity.ts` only counts a scenario carrying `@unit` / `@integration` / `@e2e` / `@regression`, so an untagged file contributes nothing in either direction — the run prints `0/0 ✓ all bound` and the check passes vacuously. 17 scenarios describing the probe contract were decorative. (The `specs/security/helm-strict-admission.feature` this PR adds had the same problem; it is tagged too.)

The seven handler scenarios are now `@unit` and bound via `/** @scenario … */`; enforced scenarios go from **3381 to 3388**. The chart-render and live-cluster scenarios are `@e2e @unimplemented` — genuinely covered by `e2e-overlays.sh` / `e2e.sh`, but the checker's shell binding roots scan `.bats` only, so a `.sh` assertion cannot bind as written. Tracked rather than invisible.

Two of the newly-enforced tests could not fail, which matters once they are load-bearing:

- The response fake discarded headers on both paths (`setHeader` was a no-op; `writeHead` read only its status argument), so deleting `res.setHeader("Content-Type", register.contentType)` kept all seven green while Prometheus would silently stop parsing the worker registry.
- `serves the registry rather than an auth failure` asserted the status was `undefined` and the body was a string — both true of the empty string, and the registry is empty in this test process.

Both are fixed and mutation-verified: deleting either `Content-Type` line now turns the suite red.

## Test plan

- [x] `helm lint` clean on all three charts
- [x] `helm template` renders and parses for defaults, all 15 overlays, size-prod + ingress + strict-admission stacked, `clickhouse.replicas=3`, and both subcharts standalone
- [x] 155 assertions across all eleven render-level e2e suites pass, including #5524's `test_workers_probes`
- [x] `pnpm test:unit run src/server/workers/__tests__/workerMetricsHandler.unit.test.ts` — 7 passed
- [x] `check-feature-parity` passes; enforced scenarios 3381 → 3388
- [x] Mutation-tested the strengthened header assertions — deleting either `Content-Type` line fails the suite, green again once reverted
- [x] Mutation-tested the new guard: reverting the container-level `runAsNonRoot` line now fails the suite (it passed before)
- [x] Verified overrides layer — `--set app.podSecurityContext.fsGroup=2000` keeps `runAsNonRoot`, `runAsUser` and the seccomp profile; `--set app.containerSecurityContext.readOnlyRootFilesystem=false` keeps `allowPrivilegeEscalation: false` and `capabilities.drop`
- [x] Datastores verified to boot read-only against this exact mount set (postgres first-boot initdb and restart-on-existing-PGDATA, redis, clickhouse incl. cold-storage + backup, keeper) — the previous "these need a writable root" comments did not hold

## Review pass

Audited with `/review-pr` after the work above, which found a **P0 in this branch**: the cronjobs pod `securityContext` was mis-indented, so `helm template` failed for any release defining a cron job. Nothing caught it — `cronjobs.jobs` ships `{}` and both e2e values files disable cronjobs, so no render in the repo produced that template and CI was green.

The audit also showed the rewritten pod-security assertions were still unsound: they counted matches across the whole document, so moving `readOnlyRootFilesystem` off two of three backup containers and up to the pod level (where Kubernetes ignores it) kept the totals equal and the suite green. They now parse the manifest and read fields off each container object, via `charts/lib/pod-security-report.rb`. All three of those regressions now fail with the workload and container named.

Docs corrections from the same pass: the `MustRunAs` uid list omitted 65534 (Keeper's init container), "two bundled components cannot comply" is three (the ClickHouse preflight Job), and the `size-minimal` footprint was understated by ~1Gi.

## Deployment Impact

Self-hosted Helm only; SaaS unaffected. On `helm upgrade` the datastore pod specs change, so the bundled single-replica PostgreSQL/Redis/ClickHouse restart in place — expect a short blip if you run them chart-managed. Postgres and Redis now honour `global.scheduling.*`, which they previously ignored: if you set a global node pin, those two will move.

New values are additive (`scratch.*` on the ClickHouse chart; `podSecurityContext` / `containerSecurityContext` / `nodeSelector` / `tolerations` / `affinity` on postgresql and redis). `size-minimal` now schedules a workers pod and totals ~4.7Gi requests / ~2.4 CPU (~1Gi of which is the Langy assistant), which no longer fits a single 2-vCPU node. `size-dev` and `examples/values-local.yaml` were re-sized on the same basis. No secret, CRD, RBAC or data-migration changes.

## Deliberately not done

- **No seccomp profile added to `charts/langyagent`.** It has none, so it is denied where seccomp presence is enforced. Adding `RuntimeDefault` to a workload that runs LLM-generated shell under a sandboxed runtime needs runtime verification I can't do from a chart render, and the overlay removes the pod on exactly the clusters that would reject it. Worth a follow-up.
- **`/tmp` on the app/workers/nlp/langevals pods is still unbounded.** Their usage is app-driven and I have no measured ceiling; guessing one trades a disk-exhaustion risk for an eviction risk. The ClickHouse volumes are bounded because their failure mode is concrete (1000M x 10 log rotation).
- **The ClickHouse backup CronJobs are kept.** They looked removable, but they're live backup machinery with in-flight work; removing them is a product decision, not cleanup.

### DIFF
--- a/charts/clickhouse-serverless/README.md
+++ b/charts/clickhouse-serverless/README.md
@@ -70,7 +70,7 @@ See the [Docker image README](../../clickhouse-serverless/README.md) for the ful
 | Name | Description | Default |
 |------|-------------|---------|
 | `image.repository` | Image repository | `langwatch/clickhouse-serverless` |
-| `image.tag` | Image tag | `0.1.0` |
+| `image.tag` | Image tag | `0.2.0` |
 | `image.pullPolicy` | Pull policy | `IfNotPresent` |
 
 ### Storage
@@ -106,6 +106,8 @@ Shared by cold storage and backups. Required when either `cold.enabled` or `back
 |------|-------------|---------|
 | `backup.enabled` | Enable native ClickHouse BACKUP/RESTORE to S3 (requires `objectStorage`) | `false` |
 | `backup.database` | Database to back up | `langwatch` |
+| `backup.user` | ClickHouse user for backup/restore operations | `default` |
+| `backup.resources` | CPU/memory requests + limits for the backup/restore Job containers | requests `100m`/`128Mi`, limits `500m`/`512Mi` |
 | `backup.full.schedule` | Cron schedule for full backups | `0 */12 * * *` |
 | `backup.incremental.schedule` | Cron schedule for incremental backups | `0 * * * *` |
 
@@ -160,6 +162,25 @@ This creates two users: `analyst` with read-only access to all databases, and `e
 | `scheduling.nodeSelector` | Node selector labels | `{}` |
 | `scheduling.affinity` | Affinity rules | `{}` |
 | `scheduling.tolerations` | Tolerations | `[]` |
+
+### ServiceAccount
+
+| Name | Description | Default |
+|------|-------------|---------|
+| `serviceAccount.create` | Create a dedicated ServiceAccount | `true` |
+| `serviceAccount.name` | ServiceAccount name (defaults to the chart fullname) | `""` |
+| `serviceAccount.automountServiceAccountToken` | Mount the SA token; ClickHouse/Keeper need no Kubernetes API access. Also set on the pod specs so policies that inspect the pod (not the SA) accept it. | `false` |
+| `serviceAccount.annotations` | ServiceAccount annotations (e.g. IRSA role ARN) | `{}` |
+
+## Pod Security
+
+ClickHouse, Keeper, and the backup/restore Jobs run non-root (uid 101) with a
+read-only root filesystem, `RuntimeDefault` seccomp, dropped capabilities, no
+privilege escalation, and no mounted ServiceAccount token. The paths ClickHouse
+writes at runtime (server logs, the pid directory, `/tmp`, and the rendered
+`config.d`/`users.d`) are `emptyDir` volumes; data stays on the PVC, so the
+image layer is never writable. This clears Pod Security Admission `restricted`
+and Gatekeeper / Kyverno policies, including "read-only root on every container".
 
 ## Deployment Modes
 
--- a/charts/clickhouse-serverless/README.md
+++ b/charts/clickhouse-serverless/README.md
@@ -65,6 +65,13 @@ See the [Docker image README](../../clickhouse-serverless/README.md) for the ful
 | `replicas` | Number of ClickHouse nodes. 1 = standalone MergeTree, 3+ = ReplicatedMergeTree + Keeper (must be odd) | `1` |
 | `clusterName` | ClickHouse cluster name used in macros and remote_servers config | `langwatch` |
 
+> **`memory` scales with ingest, not just stored size.** The image auto-tunes
+> `max_server_memory_usage` (~85% of the limit) and per-query limits from
+> `memory`, and large/concurrent inserts plus background merges spike well above
+> idle. Sub-2Gi values are fine for smoke or light local use, but raise `memory`
+> (and `cpu`) before pushing real trace throughput or a big ingest burst will
+> OOM the pod.
+
 ### Image
 
 | Name | Description | Default |
--- a/charts/clickhouse-serverless/README.md
+++ b/charts/clickhouse-serverless/README.md
@@ -192,9 +192,11 @@ own quota instead of filling the node.
 
 ## Pod Security
 
-ClickHouse, Keeper, and the backup/restore Jobs run non-root (uid 101) with a
-read-only root filesystem, `RuntimeDefault` seccomp, dropped capabilities, no
-privilege escalation, and no mounted ServiceAccount token. The paths written at
+ClickHouse, Keeper, and the backup/restore Jobs run non-root (uid 101; Keeper's
+init container runs 65534) with a read-only root filesystem, `RuntimeDefault`
+seccomp, dropped capabilities, no privilege escalation, and no mounted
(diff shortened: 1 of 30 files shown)


0xdeafcafe:
Folded into #6375 and closing in favour of it.

All three open chart PRs touched `values.yaml`, `_helpers.tpl` and the same four pod specs, so they conflicted with each other in whichever order they landed. Consolidating resolves that once instead of three times.

This branch merged with conflicts only in the four deployment templates, and they were purely positional: this PR made `initContainers` conditional on `extraInitContainers` (dropping the hardcoded `fix-nextjs-permissions` step, which a read-only root filesystem makes unnecessary), while #6375 inserts `priorityClassName` immediately above it. Both sides kept in all four. `values.yaml` merged cleanly — disjoint keys.

Verified on the merged branch: `helm lint` passes and all 13 template configurations render, including the `strict-admission` overlay this PR introduced.

PR #6314, opened 2026-07-29
test(server): bind the worker probe specs so the parity gate enforces them

0xdeafcafe:
Post-merge follow-up to #5524. Closes finding #1 from the handoff brief — the highest-leverage one, because it is what makes every other finding on those files regression-proof.

## Why

`specs/server/worker-liveness-probe.feature` and `specs/server/metrics-collection.feature` shipped with **no tags at all**.

`check-feature-parity.ts` only counts a scenario if it carries one of `@unit` / `@integration` / `@e2e` / `@regression` (`BOUND_TAGS`, line 141; the filter is line 576). An untagged file therefore contributes zero scenarios in either direction — the run prints `0/0 ✓ all bound` and the `feature-parity` check passes vacuously. In CI output that is indistinguishable from a fully-bound file. 17 scenarios describing the probe contract were decorative.

## What changed

**Tagging and binding.** The seven handler-level scenarios are tagged `@unit` and bound to the tests that already covered them, via `/** @scenario <title> */` above each `it(...)`. Enforced scenarios go from **3381 to 3388**.

The chart-rendering scenario and the eleven live-cluster scenarios are tagged `@e2e @unimplemented`. They *are* covered — by `test_workers_probes` in `e2e-overlays.sh` and `test_metrics_collection` in `e2e.sh` — but the checker's shell binding roots scan `.bats` only, so a `.sh` assertion cannot bind as written. `@unimplemented` records that as a tracked gap instead of hiding it behind an untagged file.

**Two of the newly-enforced tests could not fail.** That mattered less while they were decorative; it matters now.

- The response fake discarded headers on both paths the handler uses — `setHeader` was a no-op, and `writeHead` read only its status argument. Deleting `res.setHeader("Content-Type", register.contentType)` from the handler left all seven tests green while Prometheus would silently stop parsing the worker registry. The fake now captures headers from both, and the metrics and liveness tests assert the content type.
- `serves the registry rather than an auth failure` asserted `captured.status` was `undefined` and `typeof body === "string"`. Both hold for the empty string, and the registry *is* empty in this test process. It now registers a fixture counter and asserts the sample appears in the body.

Also drops a dead assertion (`not.toContain("#")` after an exact `toBe("ok")`) and wraps the two bare `it()` blocks in the `describe("when …")` the repo convention requires.

## Test plan

- [x] `pnpm test:unit run src/server/workers/__tests__/workerMetricsHandler.unit.test.ts` — 7 passed
- [x] `check-feature-parity` passes; enforced count 3381 → 3388
- [x] Mutation-tested both strengthened assertions, since the point is that they *can* fail:
  - deleting `res.setHeader("Content-Type", register.contentType)` → `serves the registry with the Prometheus content type` fails
  - dropping the `Content-Type` from the `/healthz` `writeHead` → `returns no metric samples` fails
  - both green again once reverted

## Not in scope

The remaining handoff findings are untouched here, notably the probe port being hardcoded to `2999` while the listener's port is env-derived (`workers.extraEnvs` can move the listener and leave the probes pinned), and the `startupProbe` budget being 120s against the app's 300s. Those are chart changes and want their own PR.

### DIFF
--- a/langwatch/src/server/workers/__tests__/workerMetricsHandler.unit.test.ts
+++ b/langwatch/src/server/workers/__tests__/workerMetricsHandler.unit.test.ts
@@ -14,6 +14,8 @@
  */
 import { describe, expect, it, vi } from "vitest";
 
+import { Counter, register } from "prom-client";
+
 import {
   WORKER_LIVENESS_PATH,
   createWorkerMetricsHandler,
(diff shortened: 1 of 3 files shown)


0xdeafcafe:
Folding into #6301 — keeping this work in a single PR. The commit is cherry-picked there; nothing is lost.

Issue #6290, opened 2026-07-28
Gateway translated lanes silently drop, mistranslate, or reject valid OpenAI chat params that raw-forward lanes honor (full audit)


rogeriochaves:
Found while fixing #6285 (max_tokens ignored on translated lanes), then made exhaustive by a full parameter fidelity audit (bifrost v1.4.22 translator source plus live probes on the anthropic, bedrock, and gemini lanes; complete verified matrix in the comment below).

Raw-forward lanes (openai, azure, vllm) pass everything through byte-identical apart from the model rewrite and the deliberate stream_options.include_usage injection. Every defect below is on the translated lanes (anthropic, bedrock, gemini, vertex, bedrock VPCE), where the gateway builds the provider request itself: the client believes a parameter is in effect and the gateway silently disagrees, differently per lane.

## Severity 1: silent behavior-contract violations

1. Gemini lane returns an empty 200 (`"choices": null`, usage billed, no error, no finish_reason) whenever thinking consumes the whole completion cap. Reproduced with a plain `max_tokens: 12` request, no exotic params. `choices: null` also breaks strict OpenAI parsers on its own.
2. `response_format: {"type":"json_object"}` silently dropped on anthropic and bedrock: live probes answered fenced ```json markdown for a mode OpenAI guarantees is parseable JSON.
3. `tool_choice: "none"` mistranslated on bedrock: dropped, Converse defaults to auto, so a model explicitly told not to call tools may call them.
4. `reasoning_effort` on bedrock for non-Anthropic, non-Nova families: reasoning silently dropped AND `inferenceConfig.maxTokens` force-set to a model default the caller never sent.
5. `n` silently collapsed to one choice on all translated lanes (live-verified). Clients sampling N completions get one, no error.
6. `logprobs` / `top_logprobs` silently dropped on anthropic and bedrock: the client requests data and receives none (gemini maps them and surfaces an honest provider 400 where unsupported).
7. `service_tier` forwarded verbatim with no validation: a valid OpenAI value hard-400s on bedrock (live: AWS enum validation error) and every non-"auto" OpenAI value 400s on anthropic.
8. `input_audio` content parts silently dropped on anthropic; an audio-only message becomes an empty-content message the upstream 400s with no hint of the cause (bedrock at least errors loudly; gemini honors).
[...]

PR #6318, opened 2026-07-29
feat(aigateway): three-tier parameter policy for translated lanes with drop_tuning_params

rogeriochaves:
Implements the three-tier parameter policy approved from the [full parameter fidelity audit](https://github.com/langwatch/langwatch/issues/6290#issuecomment-5109323199). Closes most of #6290 (what remains is scoped there).

STACKED on #6285 (`fix/max-tokens-anthropic-bedrock`), which owns the parser this PR builds on. Base is set to that branch; retarget to `main` and rebase when #6285 merges.

## What

Raw-forward lanes (openai, azure, vllm) pass every parameter through byte-identical; nothing changes there. On the translated lanes (anthropic, bedrock, gemini, vertex, bedrock VPCE), where the gateway builds the provider request itself, every OpenAI chat-completions parameter now has an explicit disposition in ONE table (`services/aigateway/adapters/providers/param_policy.go`):

- **mapped**: translated faithfully (by the bifrost translator, or by this layer where the translator gets it wrong).
- **dropped** (tier 3, tuning): removed, request proceeds, and the drop is signaled in three places: `extra_fields.params_dropped` on the response body (on streams: the final usage-bearing chunk), the `X-LangWatch-Params-Dropped` response header, and the `langwatch.gateway.params_dropped` span attribute. Never silent.
- **refused** (tier 2, contract): the request depends on the parameter functionally; always fails with `400 unsupported_parameter` and copy that names the parameter, the lane, and why:

```
refusing to drop 'response_format' for anthropic/claude-haiku-4-5: the request depends on it
functionally (the response would not be guaranteed parseable JSON on this lane). Remove it, or
use a model that supports it
```

**The option: `drop_tuning_params` (body field, default `true`).** `false` = strict mode: any unmappable parameter refuses instead of dropping. Contract parameters always refuse regardless. The name is deliberate: the flag only governs tuning parameters, and the semantics differ from LiteLLM's `drop_params` (which drops everything), so reusing that name would be a false compatibility promise. The field is consumed by the gateway and stripped before any provider dispatch (on raw lanes only when present, preserving byte-identity otherwise).

Unknown vendor keys (`min_p`, `repetition_penalty`, ...) fall to the droppable default on translated lanes, so they are signaled too.

## Fixed outright (from the audit's S1 list)

| Defect | Fix |
|---|---|
| anthropic discarded `top_p` whenever `temperature` was present (translator else-if), silently | the pair is genuinely unmappable: Anthropic models hard-400 both together (verified live on the anthropic AND bedrock lanes). The drop is now visible (signal) and strict mode refuses; non-Anthropic bedrock families keep both |
| bedrock `tool_choice: "none"` became auto (model could call tools it was told not to) | mapped faithfully by removing the tools from the Converse request |
| `service_tier` forwarded verbatim into AWS's different enum, hard 400 | validated per lane: anthropic `"auto"` passes natively, everything else drops with a signal and the typed field is cleared |
| `n > 1` silently returned one choice | dropped-to-1 WITH signal by default; refuses under strict mode; `n: 1` passes |
| bedrock `reasoning_effort` on non-Anthropic/non-Nova families: reasoning silently dropped AND an output cap force-set | refused pre-dispatch (policy-visible), so bifrost's force-set never runs |
| `response_format json_object` silently dropped on anthropic/bedrock (fenced markdown for a parseable-JSON guarantee) | tier-2 refused: bifrost offers no sound mechanism, and a silent quality degrade is exactly what the policy exists to prevent; `json_schema` stays enforced on all lanes |
| gemini answered HTTP 200 with `choices: null` (usage billed, no error) when thinking consumed the completion cap | response-contract repair: one synthesized choice with `finish_reason "length"`, empty content, usage intact |
| named `tool_choice` for a function absent from `tools` silently nulled (bedrock) or failed downstream | refused with the missing name, on every translated lane |
| `stop` as a bare string 400ed on translated lanes (valid OpenAI shape) | normalized to the list form in the parse; mapped everywhere |
| legacy `functions` / `function_call` silently dropped | tier-2 refused with pointer to `tools` / `tool_choice` (full legacy mapping incl. response-side stays in #6290) |

## Bedrock VPCE under the same table

The VPCE mapper consults the same policy (it is the bedrock lane) and now maps what the table says bedrock maps, mirroring bifrost's wire shapes exactly: `reasoning_effort` to `additionalModelRequestFields.thinking` (adaptive + `output_config.effort` on 4.6+/4.7, else `budget_tokens = 1024 + ratio x (max_tokens - 1024)` using bifrost's own helpers), and `response_format json_schema` to `output_config.format`. One deliberate divergence, per the audit: when the caller sent no output cap, the thinking budget is computed against the model default WITHOUT force-setting `inferenceConfig.maxTokens`. The VPCE content-part gaps (vision/file/audio parts, cachePoint, tool-result grouping) are follow-up scoped in #6290.

## Docs

`docs/ai-gateway/parameter-mapping.mdx`: the full param x lane table (generated from the code table; `TestParamPolicyDocsInSync` fails the build if they drift), `drop_tuning_params` semantics, the drop signals, and the `reasoning_effort` budget ratios. `unsupported_parameter` added to the errors reference.

## Evidence (local stack, real provider keys)

BEFORE (documented in the audit): seed/n/logit_bias silently ignored, `service_tier` a hard AWS 400, `tool_choice "none"` calling tools, `json_object` answering fenced markdown, gemini empty 200s.

AFTER:

```
# tier-3 drop with the three signals
drop+signal seed:        200 hdr=X-LangWatch-Params-Dropped: seed  body extra_fields.params_dropped=["seed"]
n=2 drop-to-1 + signal:  200 one choice, hdr/body dropped=["n"]
service_tier (bedrock):  200 dropped+signaled  (before: hard AWS 400 enum validation error)
top_p+temperature:       200 dropped+signaled on anthropic AND bedrock Claude (the models hard-400 the pair; verified live)
streamed:                SSE response carries the header AND the final usage chunk carries extra_fields.params_dropped=["seed","user"]

# strict mode
seed + drop_tuning_params:false -> 400 unsupported_parameter
  "refusing to drop 'seed' for anthropic/claude-haiku-4-5: drop_tuning_params is false and Anthropic
   has no sampling seed. Remove it, set drop_tuning_params to true, or use a model that supports it"

# tier-2 contract refusals (drop_tuning_params cannot drop these)
json_object (anthropic): 400 "the response would not be guaranteed parseable JSON on this lane"
ghost named tool:        400 'tool_choice names "ghost" but no tool with that name is in tools'
titan reasoning_effort:  400 "this Bedrock model family has no reasoning mapping ..."

# mapped cells keep working
json_schema (anthropic):     200 {"city": "Paris"}
bedrock claude reasoning:    200 thinking observed (ct=72 for "391")
bedrock tool_choice "none":  200 and the model calls no tools
stop as bare string:         200 stops before "7"
gemini max_tokens 12:        200 finish_reason "length", empty content, usage intact  (before: choices: null)
```

## Tests

- Table-driven policy unit tests per cell: droppable per lane (incl. unknown vendor keys), strict-mode refusals with the exact copy, contract refusals immune to `drop_tuning_params`, named-tool validation, bedrock none-removes-tools vs anthropic native none, top_p injection with merge-flag context (and the Opus 4.7 guard), service_tier clearing, stop-string normalization, raw-lane bypass with byte-identity, default-true semantics.
- Response-contract tests: `ensureChoicesPresent` (null and empty choices repaired, populated untouched, usage intact), `extra_fields.params_dropped` injection preserving siblings.
- VPCE: thinking budget math against bifrost's helpers (1177 for low at 2048), adaptive on 4.7 with `minimal` mapped to `low`, json_schema under `output_config.format`, effort+format coexistence, the 1024 floor, nil document when nothing is requested, and no forced output cap.
- Spec: 7 new scenarios under a new rule in `specs/ai-gateway/openai-param-compat.feature`, all bound.
- Full battery: gofmt, go build, go vet, go test -race (whole service), pinned golangci-lint 0 issues, docs parity test green.

## Deployment Impact

- No schema, chart, or environment changes; control plane untouched.
- Previously-silent behavior becomes visible: requests carrying tier-2 params on translated lanes (json_object on anthropic/bedrock, logprobs on anthropic/bedrock, legacy functions, allowed_tools, reasoning_effort on unsupported bedrock families, named ghost tools) now 400 with actionable copy instead of silently misbehaving. Requests carrying tuning params keep working and gain the drop signals.
- New body field `drop_tuning_params` (default true, backward compatible); stripped before providers see it.
- Response envelopes on translated lanes may carry `extra_fields.params_dropped` and the `X-LangWatch-Params-Dropped` header.

### DIFF
--- a/docs/ai-gateway/api/errors.mdx
+++ b/docs/ai-gateway/api/errors.mdx
@@ -45,6 +45,7 @@ Errors (like successes) always carry:
 | `provider_error` | `502` | An upstream provider returned a non-recoverable error and fallback (if any) was exhausted. |
 | `upstream_timeout` | `504` | An upstream provider exceeded `timeout_ms` and fallback (if any) was exhausted. |
 | `bad_request` | `400` | Validation error on the incoming payload (e.g. missing `model`). |
+| `unsupported_parameter` | `400` | The parameter policy refused a request parameter for the target lane: either the request depends on it functionally and the lane cannot honor it, or `drop_tuning_params` is `false` and the lane has no mapping for it. The message names the parameter, the lane, and why. See [Parameter mapping](/ai-gateway/parameter-mapping). |
 | `payload_too_large` | `413` | Request body exceeded `GATEWAY_MAX_REQUEST_BODY_BYTES` (default 10 MiB). Rejected at the edge, before auth, resolve-key, or any upstream dispatch, so a 1 GB drive-by scan never pressures the pod memory limit. Declared `Content-Length` above the cap returns 413 immediately without draining the socket; chunked unknown-length bodies trip a `*http.MaxBytesError` at the cap. |
 | `internal_error` | `500` | Unclassified gateway error. `X-LangWatch-Request-Id` is how we trace it. |
 
--- a/docs/ai-gateway/parameter-mapping.mdx
+++ b/docs/ai-gateway/parameter-mapping.mdx
@@ -1,0 +1,84 @@
+---
+title: "Parameter mapping"
+description: "How every OpenAI chat-completions parameter behaves on each provider lane, and how to control dropping with drop_tuning_params."
+sidebarTitle: "Parameter mapping"
+---
+
+The gateway speaks the OpenAI chat-completions wire format to every provider. For OpenAI-compatible targets (OpenAI, Azure OpenAI, self-hosted vLLM and other custom OpenAI-compatible endpoints) the request body is forwarded byte-for-byte, so whatever the provider accepts, you get. For the translated lanes (Anthropic, Bedrock, Gemini, Vertex) the gateway builds the provider-native request itself, and every parameter follows an explicit policy:
+
+- **mapped**: translated faithfully to the provider's native field.
+- **dropped**: a tuning parameter with no equivalent on the lane. By default it is removed, the request proceeds, and the drop is signaled (see below). It is never silently ignored.
+- **refused**: the request depends on the parameter functionally; honoring the request without it would change what you observably get. These always fail with a clear `400 unsupported_parameter` error naming the parameter, the lane, and why.
+
+## drop_tuning_params
+
+Send `"drop_tuning_params": false` in the request body to opt into strict mode: any parameter the lane cannot map then refuses instead of dropping. The default is `true`. The field itself is consumed by the gateway and never reaches a provider.
+
+```json
+{
+  "model": "anthropic/claude-haiku-4-5",
+  "messages": [{"role": "user", "content": "hi"}],
+  "seed": 42,
+  "drop_tuning_params": false
+}
+```
+
+```json
+{
+  "error": {
+    "type": "unsupported_parameter",
+    "code": "unsupported_parameter",
+    "message": "refusing to drop 'seed' for anthropic/claude-haiku-4-5: drop_tuning_params is false and Anthropic has no sampling seed. Remove it, set drop_tuning_params to true, or use a model that supports it"
+  }
+}
+```
+
+## Drop signals
+
+When a parameter is dropped, the drop is visible in three places:
+
+- Response body: `extra_fields.params_dropped` lists the dropped parameter names (on streams, on the final usage-bearing chunk).
+- Response header: `X-LangWatch-Params-Dropped` carries the same list, comma separated.
+- Trace: the gateway span records `langwatch.gateway.params_dropped`.
+
+## The table
+
+{/* param-table:begin generated from paramPolicyTable, kept in sync by TestParamPolicyDocsInSync */}
+| Parameter | anthropic | bedrock | gemini / vertex | Notes |
+|---|---|---|---|---|
+| `audio` | dropped | dropped | dropped |  |
+| `frequency_penalty` | dropped | dropped | mapped |  |
+| `function_call` | refused | refused | refused |  |
+| `functions` | refused | refused | refused |  |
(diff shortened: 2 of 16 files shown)


langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** closes most of #6290, implementing a policy approved from a documented parameter-fidelity audit. Good provenance.

**Stacked on #6285**, which owns the parser this builds on, and the body says to retarget and rebase when that lands. Review #6285 first or the diff will move.

**Top inspection targets**
1. `Must Check` - what `drop_tuning_params` silently discards. Dropping a parameter changes model behaviour without erroring, so a user who sets `temperature` or `top_p` and has it dropped gets different output with no signal. Confirm the tiers are documented and that dropping is visible somewhere (response metadata, a log, a warning), not silent.
2. `Must Check` - the tier assignment per translated lane. A parameter placed in the wrong tier either leaks an unsupported param to a provider (hard error for the user) or drops a supported one (quiet behaviour change). Both are user-visible in the gateway hot path.
3. `Ask Author` - defaults for a parameter the policy does not know about. Pass through or drop? Pass-through risks provider errors; drop risks silently ignoring valid input.

**Production risk (scariest first):** silent alteration of inference parameters in the gateway, which surfaces to users as degraded model output with no error to trace back to this change.

1. Is a dropped parameter observable by the caller in any way?
2. What happens to an unrecognised parameter?
3. Does #6285 need to merge first?

PR #6324, opened 2026-07-29
feat(data-privacy): pii exception patterns and readable ingestion key id

rogeriochaves:
## What

Two data-privacy fixes for over-redaction, reported from the field:

1. **Do-not-redact exception patterns on the PII policy.** Detection heuristics over-trigger on business identifiers that merely look like PII: a customer's 14-digit reservation numbers are Luhn-valid, so they were stored as `[CREDIT_CARD]`; internal queue addresses read as personal emails. A scope can now list exception regexes next to its PII level: a detected span whose **entire** matched text matches one of them is kept as it was, and everything else in the same text is still redacted.

2. **The ingestion key id is readable on coding-agent traces.** The one field that says which ingestion key produced a trace was stored as `[SECRET]`, because the sensitive-attribute-NAME deny-list matched the name it was stamped under. It is now stamped under a name that is not sensitive, so it survives ingestion intact.

## The ingestion key id determination

The value is **not** the secret. The OTLP receiver stamps it server-side with the ingestion key's row id (`ee/governance/services/ingestKeyProvenance.utils.ts`), overwriting anything the client sent under that name, so it cannot be forged upstream.

It used to be stamped as `langwatch.api_key.id`, which rendered as `[SECRET]` because `isSensitiveAttributeKey` matches `api_key` as a bounded segment of the key NAME and nukes the value regardless of shape.

The first attempt exempted that one key from the name rule. That was wrong, and review caught it: the receiver only overwrites the attribute for ingest-key traffic, so an ordinary project API key could submit an arbitrary value under the exact same name, and a name-based exemption cannot tell the two apart. A name-keyed deny-list can never carry an exemption safely, because the name is attacker-supplied.

So the id is stamped under **`langwatch.ingest_key_id`** instead, a sibling of the four provenance attributes already written next to it (`langwatch.source`, `langwatch.origin`, `langwatch.organization_id`, `langwatch.template.id`). That name matches neither of the two filters that would otherwise eat it:

- the **secret-name deny-list**, whose pattern has no bare `key` alternative, only `api[_-]?key` and `private[_-]?key`; and
- the **`langwatch.reserved.` strip**, which `recordSpanCommand` applies wholesale to user-submitted attributes and which the facet list also hides from users. An interim version of this PR stamped `langwatch.reserved.ingest_key_id` and hit exactly this: the receiver wrote the id and the pipeline deleted it one step later, turning an unreadable value into a missing one.

So no exemption is carved into either filter, and `langwatch.api_key.id` stays fully deny-listed for anything a client sends under it. Covered by unit and service tests, plus a test that runs the real command handler over a genuinely stamped resource so a future move back under a stripped prefix fails the build.

## How exceptions work

- `pii.exceptPatterns` on the scoped data-privacy config (ADR-021 cascade): unions across the chain like `secrets.customPatterns`, while the level itself stays first-set-wins. Save-time validation compiles each pattern and runs the safe-regex (ReDoS) analysis; ingestion never vets patterns and skips any that fail to compile (legacy config) instead of crashing.
- Patterns are anchored at compile time (`^(?:...)$`): an exception must cover the whole detected span, so a prefix cannot carve a hole out of a longer identifier.
- An exception is the only pattern here that REMOVES redaction, so an over-broad one fails open: anchored, `.*` or `\d+` matches every detected span and silently turns the whole PII pass off while the level still reads as active. Save-time validation rejects a pattern that matches values of unrelated kinds, so an exception has to describe one identifier shape. The real use cases still save (`\d{14}`, `RES-\d{10}`, an internal queue address); `.*`, `.+`, `\d+`, `\w+`, `[\s\S]*` and `\S+` are refused. Custom secret patterns are deliberately not held to this, since a too-broad one there only over-redacts.
- Native engine (essential level, custom level, and the native floor of strict): candidate spans whose matched text fully matches an exception are dropped before masking, for both the regex recognizers and the phone detector.
- Strict level with exceptions: the analysis-service call is scoped to the entities only it can detect (names, locations). The native floor already ran every pattern-based recognizer with exceptions applied, and the service returns pre-anonymized text that cannot be vetoed, so re-scanning pattern entities there would re-redact the kept values. Without exceptions, strict behavior is unchanged.
- Google DLP path (including the presidio-outage fallback): findings are filtered by their quoted text against the exceptions, so the fallback cannot re-redact a value an exception kept.
- UI: an Exceptions list under the PII level in the privacy rule drawer, mirroring the custom secret patterns editor, with inline validation and the rule-table summary counting exceptions.

## Screenshots

The exceptions editor in the privacy rule drawer:

exception editor

An invalid pattern blocks save with an inline error:

invalid pattern

The saved rule summarized in the rules table:

rules table

End-to-end proof on a locally ingested OTLP trace: the excepted reservation number survives, the email next to it is redacted, and `user.api_key` (not shown, asserted via API) is still `[SECRET]`:

trace evidence

### The API key attribute in the trace drawer

QA'd as an org admin against a live, non-revoked key, ingested through the real OTLP receiver so the id was receiver-stamped rather than hand-placed. The row reads `langwatch.api_key` and resolves the id to the key's name, linked:

drawer desktop

drawer zoom

Clicking it lands on the key's row in settings:

link lands

A revoked or out-of-reach key falls back to the raw id with no link, rather than a dead one:

revoked fallback

Mobile at 390 wide:

drawer mobile

### The same row for an ordinary member

The name is resolved through `apiKey.nameById`, not the admin-gated key list, so a member with no key administration rights sees the name rather than a raw row id. Captured as a plain org member, verified by that account being unable to see other people's keys or the members admin:

non-admin proof

non-admin drawer

At 390 wide, with the raw-id fallback shown for a read-only share view:

non-admin mobile

At 390 wide a long key name truncates inside its column with the copy button clear. Measured in the live DOM: before, the link ran 241px past its cell and the text reported scrollWidth equal to clientWidth so it was never clipped; after, the link matches its cell exactly and the text is clipped by 241px.

truncation before after

An anonymous share view resolves nothing and falls back to the raw id, unlinked:

share fallback

### The API keys settings page

One title, one subtitle, security warning intact, with ingestion keys present so the previously duplicated case is actually exercised:

api keys desktop

api keys mobile

## Tests

- `specs/data-privacy/pii-redaction.feature` and `secrets-redaction.feature` extended first with the new scenarios.
- Unit: exception filtering in the native engine (full-match anchoring, partial-match rejection, phone path, invalid-pattern skip), attribute name-exemption with value-scan retention, cascade union in the resolver, schema accept/reject, form-state round-trip and rule summary.
- Integration (real Postgres): save-time rejection of unsafe and non-compiling exception patterns, safe pattern stored and resolved.
- Service end-to-end: exception keeps the value while neighboring PII is redacted; strict batch scoped to strict-only entities with exceptions forwarded; api_key.id scenarios.
- Full local QA against a running stack: rule created through the UI, OTLP trace ingested through the collector, stored trace verified via API and UI (screenshot above).

`pnpm typecheck:all` green, touched files biome-formatted.

## Summary by CodeRabbit

* **New Features**
  * Added PII exception patterns to the privacy rule drawer, including an Exceptions section, regex validation, and a “PII exception(s)” count.
  * Implemented do-not-redact regex exceptions across PII redaction so approved identifiers can be preserved while other PII remains masked.
* **Bug Fixes**
  * Ensured receiver-stamped `api_key` identifiers remain readable while actual API key material is still redacted.
* **Tests**
  * Expanded validation, policy resolution, and redaction coverage for exception-pattern behavior, including rejection of unsafe/invalid regex patterns.


langwatch/src/server/data-privacy/dataPrivacyPolicy.service.ts
```diff
@@ -22,25 +22,25 @@ export class InvalidDataPrivacyConfigError extends Error {
...
 
 /**
- * Custom secret patterns run against every ingested payload, so a pattern must
- * both compile and pass the safe-regex (ReDoS) analysis before it is stored.
- * Rejecting at write time keeps the redaction hot path free of per-event
- * pattern vetting.
+ * Custom secret patterns and PII exception patterns run against every ingested
+ * payload, so a pattern must both compile and pass the safe-regex (ReDoS)
+ * analysis before it is stored. Rejecting at write time keeps the redaction hot
+ * path free of per-event pattern vetting.
```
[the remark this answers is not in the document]
rogeriochaves: Fixed, the comment now says the patterns are evaluated when their redaction pass runs (secrets scanning when enabled, PII exceptions against detected spans) instead of claiming every payload. Landed in the follow-up commit.

langwatch/src/server/tracer/collector/piiCheck.ts
```diff
@@ -202,10 +202,31 @@ const codepointToCodeUnitConverter = (
...
+  const compiled: RegExp[] = [];
+  for (const pattern of patterns ?? []) {
+    try {
+      compiled.push(new RegExp(`^(?:${pattern})$`));
+    } catch {
+      // Skip: rejected at save time; never let a bad pattern break ingestion.
+    }
+  }
+  return compiled;
+};
```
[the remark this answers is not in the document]
rogeriochaves: Good catch, the DLP path now imports compilePiiExceptPatterns and matchesPiiException from essentialPii.ts instead of re-implementing them, so the anchored compile and the ReDoS-sensitive RegExp construction live in one module. Landed in the follow-up commit.

rogeriochaves:
Addressed the outside-diff finding about overlapping findings eating into excepted spans: excepted matches now become protected ranges in both the native engine and the DLP path, and other findings mask only their portions outside protection (shared subtractProtectedRanges helper). Regression tests cover the overlapping exempt and non-exempt case plus the interval math. Landed in 8ed757c882.

langwatch/src/server/data-privacy/redaction/applyContentRedaction.ts
```diff
@@ -98,20 +115,28 @@ export function redactAttributeNative({
...
   policy: ResolvedDataPrivacy;
   compiledSecretPatterns?: readonly RegExp[];
+  compiledPiiExceptions?: readonly RegExp[];
 }): { text: string; redactedCount: number } {
   if (
     policy.secrets.enabled &&
     value.length > 0 &&
-    isSensitiveAttributeKey(key)
+    isSensitiveAttributeKey(key) &&
+    !NAME_DENYLIST_EXEMPT_KEYS.has(key)
```
langwatch-agent: **P1 — The API-key name exemption is not limited to receiver-stamped provenance.** The OTLP receiver overwrites this attribute only when `resolved.ingestSourceType` is set (`routes/otel.ts:375`); ordinary project API keys can still submit an arbitrary `langwatch.api_key.id` attribute. This bypasses the existing name-based secret scrubber for any non-signature-matching secret placed under that exact key, and the trace can then be visible to other project members. Value scanning is intentionally best-effort, so checking only `sk-lw-*` does not restore the previous guarantee. Keep the exemption tied to trusted ingestion context: carry the authenticated/stamped key id as non-user-writable ingestion metadata and bypass the name rule only when the attribute value equals that trusted id (or preserve the deny-list and expose provenance through a separate trusted field). Add an end-to-end test showing a non-ingestion API key cannot retain an arbitrary value under this name.

langwatch/src/server/app-layer/traces/span-pii-redaction.service.ts
```diff
@@ -272,14 +285,30 @@ export class OtlpSpanPiiRedactionService {
...
+  private lambdaAfterNative(policy: ResolvedDataPrivacy): {
+    entities?: readonly string[];
+    exceptPatterns?: readonly string[];
+  } | null {
+    const exceptPatterns =
+      policy.pii.exceptPatterns.length > 0
+        ? policy.pii.exceptPatterns
+        : undefined;
+    if (policy.pii.level === "strict") {
+      return exceptPatterns
```
langwatch-agent, line 315: **P2 — Exceptions do not work for strict-only entities on the normal Presidio path.** This forwards `exceptPatterns`, but `buildOptions()` always selects `mainMethod: "presidio"` and `batchPresidioClearPII()` accepts only texts/level/entities; it returns already-anonymized text and never evaluates the exceptions. For example, a custom policy selecting `PERSON` plus an exception for a known-safe service name will still redact that name. The UI and schema promise that a detected value fully matching an exception is retained, without limiting that promise to native entities. Either use a match-aware analyzer for exception-enabled strict/custom calls, or reject/explicitly constrain exception patterns whenever strict-only entities can run, with a regression test for the chosen contract.

langwatch/src/server/data-privacy/redaction/essentialPii.ts
```diff
@@ -255,20 +255,96 @@ export interface PiiRedactionResult {
...
  * `entities` narrows the recognizers that run: pass a subset (the custom PII
  * level) to redact only those identifiers, or omit it (the essential level) to
  * run every native recognizer. Entity names are the canonical identifiers from
  * `ESSENTIAL_PII_ENTITIES` (e.g. `EMAIL_ADDRESS`, `BR_CPF`).
+ *
+ * `exceptPatterns` are the policy's do-not-redact exceptions (pre-anchored via
+ * `compilePiiExceptPatterns`): a detected span whose entire matched text
+ * matches one of them is left as it was.
  */
 export function redactEssentialPiiInText({
```
langwatch-agent, line 526: **P2 — Required CI is red for a new lint violation in this function.** The `langwatch-app-ci` lint gate reports `lint/complexity/noExcessiveLinesPerFunction` increasing from 0 to 1 in this file, and `langwatch-app-complete` therefore fails. Split the newly added exception/protected-range work into focused helpers (or add a narrowly justified override) and rerun the required gate; this is a new PR violation, not existing backlog.

langwatch-agent (changes requested):
Request changes: 1 P1 and 2 P2s — see inline comments. The overlap fix is sound, but the name-denylist exemption currently opens a secret-redaction bypass; the PR also mispromises strict/custom exception behavior and has a new required lint failure.

langwatch/src/server/data-privacy/redaction/applyContentRedaction.ts
rogeriochaves: Fixed in 7c97a90. The receiver-stamped ingestion key id no longer lives under `langwatch.api_key.id` at all: it now stamps under `langwatch.reserved.ingest_key_id` (see `ingestKeyProvenance.utils.ts`), a name the sensitive-key regex never matches by construction, so no exemption is needed. `langwatch.api_key.id` itself is fully covered by the deny-list again for anything a client sends, receiver-stamped or not. Added a regression test proving an arbitrary value under the old name is still nuked (new cases in `span-pii-redaction.nativeScopedPolicy.test.ts` and `applyContentRedaction.unit.test.ts`).

langwatch/src/server/app-layer/traces/span-pii-redaction.service.ts
rogeriochaves: Fixed in 7c97a90, but as an explicit, tested contract rather than a mechanism change. Presidio's batch endpoint returns pre-anonymized text with no positions or matched text to veto against (unlike Google DLP's findings, which do carry the quote), so `exceptPatterns` structurally can never reach that call — narrowing to `STRICT_ONLY_PII_ENTITIES` shrinks which entities are exposed to that gap, it does not close it, and I don't think it can be closed without either a match-aware Presidio endpoint or a Presidio-to-DLP entity-name mapping, both bigger changes than this PR's scope. So: the UI tooltip in `data-privacy.tsx` now states the scope ("Applies to Fast detection matches; Deep detection (names, locations) can still redact a value even if it matches an exception"), the doc-comments on `PIICheckOptions.exceptPatterns` and `lambdaAfterNative` spell out why, and a new regression test (`span-pii-redaction.nativeScopedPolicy.test.ts`, "strict-only exception scoping") locks in both sides: a name/location match under strict is still redacted despite a full-match exception, while a native essential-entity match under the same strict policy is correctly kept.

langwatch/src/server/data-privacy/redaction/essentialPii.ts
rogeriochaves: Fixed in 7c97a90. Split `collectCandidateSpans` into three functions (`isValidRecognizerMatch`, `recognizedSpanFor`, `collectRecognizerSpans`, `collectPhoneSpans`) — biome now reports zero warnings on this file (I checked both `noExcessiveLinesPerFunction` and `noExcessiveCognitiveComplexity`, since the split also fixed a complexity-28 warning the line-count fix alone didn't touch). All 155 tests across the data-privacy/PII suite still pass.

langwatch/src/server/data-privacy/redaction/applyContentRedaction.ts
rogeriochaves: Picking this PR back up. The P1 is closed, but the fix that closed it had a regression of its own, now fixed in `f684e0441b`.

**The gap in the previous fix.** Moving the receiver stamp to `langwatch.reserved.ingest_key_id` did clear the secret-name deny-list, and your P1 is genuinely resolved: there is no exemption in `redactAttributeNative` at all, and `langwatch.api_key.id` is fully deny-listed for anything a client sends. But `langwatch.reserved.` is the namespace `recordSpanCommand.stripReservedAttributes` deletes wholesale, with a two-entry passthrough allowlist the new key was not on. So the receiver stamped the id and the pipeline dropped it one step later. The attribute went from unreadable to absent, and every ingest-key trace logged `Stripped user-submitted langwatch.reserved.* attribute`. The facet list hides that prefix from users too.

The existing tests did not catch it because they check the stamping helper and the redaction service in isolation; nothing exercised the stamp through the command handler that does the stripping.

**The fix.** The stamp is now `langwatch.ingest_key_id`, a sibling of the four provenance attributes already written next to it (`langwatch.source`, `langwatch.origin`, `langwatch.organization_id`, `langwatch.template.id`). None of those is stripped, and none matches the sensitive-name regex, which I verified against the pattern directly: there is no bare `key` alternative, only `api[_-]?key` and `private[_-]?key`. So it clears both filters on its own name rather than by an exemption carved into either one.

The doc-comment on `redactAttributeNative` now states the rule rather than the history: a name must never be added to an exemption list, because the deny-list is keyed on a name any client can set, so exempting one trusts every payload that claims it. A value that must stay readable gets a name the deny-list does not match in the first place.

**Test.** `recordSpanCommand.test.ts` now stamps provenance with the receiver's own helper, runs the real command handler, and asserts the id is still on the emitted resource. It fails on the previous commit (`expected undefined to be 'key_abc'`) and passes now, so moving the name back under a stripped prefix breaks the build.

I did not resolve this thread; the verdict is yours to lift.

rogeriochaves:
Adopted this PR and drove it back to green. Head is now `d500668d60`.

**Regression found and fixed in the P1 remediation.** The earlier fix moved the receiver-stamped ingestion key id to `langwatch.reserved.ingest_key_id` to get it off the secret-name deny-list. That name is in the namespace `recordSpanCommand` strips wholesale, and the new key was not on its two-entry passthrough allowlist, so the pipeline deleted the attribute right after the receiver wrote it: unreadable became absent, plus a `Stripped user-submitted langwatch.reserved.*` warning on every ingest-key trace. It is now `langwatch.ingest_key_id`, which matches neither the deny-list nor the strip prefix, so it needs no exemption in either. Pinned by a test that runs the real command handler over a genuinely stamped resource; it fails on the previous commit. Details in the thread on `applyContentRedaction.ts`.

**Hardening on the exception patterns themselves.** An exception is the only pattern in this feature that removes redaction, so an over-broad one fails open rather than closed: compiled anchored as `^(?:pattern)$`, `.*` or `\d+` matches every detected span and silently turns the whole PII pass off while the level still reads as active in the UI. Neither existing check caught that, since a catch-all compiles fine and has no nested quantifier for the ReDoS analysis to find. Save-time validation now rejects a pattern that matches values of unrelated kinds. The feature's own use cases still save (`\d{14}` for the reservation number, `RES-\d{10}`, an internal queue address); `.*`, `.+`, `\d+`, `\w+`, `[\s\S]*` and `\S+` are refused. Custom secret patterns are deliberately not held to this, since a too-broad one there only over-redacts. This mirrors the over-breadth guard already on custom attribute patterns.

**Verification.** 354 tests pass across the 25 data-privacy, PII, redaction, collector and trace-processing files; the new save-time cases bring that suite to 19. `typecheck` and `typecheck:tests` clean, `check:feature-parity` green at 3659 bound scenarios. Both lint gates pass locally the way CI invokes them: `pnpm lint` exits 0 and `biome-new-violations.sh` reports no new violations against the merge base, net one removed. `biome.jsonc` is byte-identical to main. Main is merged in rather than rebased so the review threads stay anchored.

@coderabbitai review

Automatic reviews were paused on this branch, so flagging for a fresh pass over the two changes above.

langwatch/src/server/data-privacy/redaction/applyContentRedaction.ts
rogeriochaves: Reworked in `90f113d268`. The attribute is `langwatch.api_key.id` again and the name exemption is back, but the reason it is safe has changed, so please re-read the argument rather than the diff shape.

**Your finding named the actual root cause, and that is what got fixed.** You wrote that the receiver "overwrites this attribute only when `resolved.ingestSourceType` is set". The defect was that conditional, not the name. A name-keyed deny-list cannot carry an exemption while the value under that name is sometimes caller-supplied, so the fix is to remove the "sometimes".

**The receiver now rewrites `langwatch.api_key.id` on every authenticated OTLP request.** `enforceApiKeyIdOnTraceRequest` and its log and metric siblings run unconditionally, before the ingest-key provenance stamp and independent of it. They drop any payload-supplied copy first at resource, span, span-event and span-link level (it is resource-level provenance, so a span-level copy is never legitimate), then write the authenticated row id. `resolved.ingestSourceType` no longer gates it, and the doc-comment on the exemption says in as many words that narrowing it back to ingest-key traffic reopens the hole.

`ResolvedToken` has a second variant with no ApiKey row behind it, the legacy project key. There the receiver writes nothing and the attribute stays absent. That is still a total guarantee: the caller's value never survives either way, which is the only property the exemption depends on.

**On the other ingestion paths**, which is the part that has to hold for the argument to be complete. Only OTLP lets a caller choose attribute names. The REST collector builds span attributes from a fixed `ATTR_KEYS` set in `buildSpanAttributes` and stringifies `params` into a single attribute rather than spreading it into keys; caller metadata goes through `buildResource` namespaced as `langwatch.metadata.*`. `track-event`, `trace-metadata` and `httpProxyTracing` build their spans server-side through the same helpers. So no non-OTLP path can produce this attribute name at all, and the exemption's premise holds everywhere, not just where I added code.

**Tests.** Forgery is pinned at resource, span, event and link level, and asserts the stored value is the authenticated id rather than the submitted one; the no-ApiKey-row branch asserts the forged value is dropped and nothing written; the exemption is asserted to be exact-name, so `langwatch.api_key.id.extra`, `langwatch.api_key.identifier` and `custom.langwatch.api_key.id` are still nuked, as is `user.api_key`. Real key material under the exempt name is still scrubbed by the value rules. The whole thing is proved end to end through the real `RecordSpanCommand` handler, which is what caught the previous attempt storing nothing at all.

Also in this push, unrelated to your P1 but in the same area: over-broad PII exception patterns are rejected at save time. An exception is the only pattern in the feature that removes redaction, so anchored `.*` or `\d+` matched every detected span and silently disabled the PII pass while the UI still showed the level as active.

Not resolving this thread; the verdict is yours.

rogeriochaves: Correction to my previous reply, in `bede4e3202`: I had missed a second set of OTLP receivers, and the argument I gave was not yet true when I made it.

`/api/ingest/otel/:sourceId` and its `/v1/logs` and `/v1/metrics` siblings in `routes/ingest/ingestionRoutes.ts` are also OTLP passthroughs, so they take caller-chosen attribute names exactly like `/api/otel/v1/*`. They authenticate with an ingestion-source bearer secret rather than an ApiKey, so there is no row id to attribute a payload to. That is not a reason to leave the attribute alone: the exemption needs the payload copy dropped whether or not there is a real value to replace it with. All three now enforce with a null id, which strips and writes nothing.

The webhook receiver on that same router needs nothing, since it builds its log request server-side from a fixed key set, same category as the REST collector.

So the invariant now holds across both receiver families, which is what the exemption actually rests on. Flagging it rather than quietly amending, since the completeness of that sweep is the argument.

rogeriochaves:
Head is `8558df5a6b`. 61 checks pass; the two that are red are inherited from main, not from this branch.

`feature-parity` fails on `specs/event-sourcing/fold-read-back-store.feature`, which has one unbound scenario ("a signal with nothing else to store is not lost to a cold cache") and two annotations in `foldProjection.contract.unit.test.ts` pointing at scenarios that no longer exist ("the watermark round-trips through the read-back", "the watermark survives the eval write path too"). `langwatch-app-complete` is just its aggregator. Main's own latest run fails on exactly those two jobs, none of those files are touched here, and this branch's parity was green before the merge that brought them in. That binding drift needs whoever renamed the scenarios, since picking bindings for it would be guessing at their intent.

`lint` was red for a while for the same reason (9 error-level diagnostics in event-sourcing and trace-analytics files from main) and has since gone green on its own as main was fixed.

For this branch's own contribution: `pnpm lint` exits 0, `biome-new-violations.sh` reports no new violations against the merge base with 3 removed, and `biome.jsonc` is byte-identical to main. One line-count regression I introduced in `otel.ts` was fixed by restructuring rather than an override: the api_key.id enforcement and the ingest-key provenance stamp now live in one `applyReceiverProvenance*` function per signal, which is the cohesive unit since they are the same concern and must not drift apart.

rogeriochaves:
Head is `6c9dbbe9ea`. 62 checks pass; the one red is `feature-parity`, still main's `specs/event-sourcing/fold-read-back-store.feature` binding drift, unchanged from my note above.

Two failures appeared briefly on the previous run and were both false: `no-committed-screenshots` was cancelled by a concurrent push rather than failing, and `test-integration (5)` reported 122 files and 950 tests passing but exited 1 on an unhandled error, a `setTimeout` in `HoverableBigText.checkOverflow` firing after jsdom teardown inside `EvaluationStatusItem.error.integration.test.tsx`. Neither file is touched by this branch (both are the v1 traces components, not the v2 drawer), and both jobs went green on rerun. Flagging the flake since it will bite other PRs: the timer is not cleared on unmount, so any slow shard can trip it.

Two additions since the last update:

**`apiKey.nameById`.** The drawer was reading key names off `apiKey.list`, which is admin-gated for the full organization, so an ordinary member saw the raw ApiKey row id, which is the unreadable value this feature exists to replace. The new procedure answers one question for one id the caller already holds: the name, plus whether the key is revoked. The database select is narrowed to those two columns so the lookup id, hashed secret, owner and bindings cannot ride along by accident. Any org member may call it, since anyone who can read the trace can already see the id stamped on it. It is not an enumeration surface: it takes a whole id rather than a prefix, answers one at a time, and returns null identically for an unknown id and one from another organization, so probing cannot confirm a key exists elsewhere. Non-members are rejected before any key is read. Pinned by tests on each of those properties. A revoked key now resolves to its name rather than falling through to the id, since its traces are still readable.

**A truncation fix that was not actually fixing anything.** My earlier `flex={1}` on the link was inert: the value cell is a block box, so flex-grow on a child does nothing, and the link is `inline-flex` so it shrank to fit its text. QA measured it running 241px past its cell at 390px, with `scrollWidth === clientWidth` so `truncate` never engaged and the name painting through the copy button. `display="flex"` makes the link block-level so it fills the cell; the link now matches its cell exactly and the text clips with a visible ellipsis. Before and after measurements are in the description.

QA for both was done as a non-admin member, verified by that account being refused `/settings/members` for lack of `organization:manage` and seeing only one key on the API keys page while the org has thirteen.

Issue #6323, opened 2026-07-29
Dataset chunks become unreachable when the storage backend is switched


sergioestebance:
Reported during review of #6181.

## What happens

`getDatasetStorage()` (`langwatch/src/server/datasets/dataset-storage.ts`) picks the dataset backend from the CURRENT storage destination:

```ts
if (destination.kind === "s3") return new S3DatasetStorage();
if (destination.kind === "azure") return new AzureDatasetStorage();
return new LocalDatasetStorage(destination.root);
```

Flip `STORED_OBJECTS_BACKEND` after datasets already exist and every previously written chunk becomes unreachable. The bytes are still there — they are simply no longer addressable, until the toggle is flipped back.

## Why stored objects survive this and datasets do not

Stored objects are addressed by a URI (`azure-blob://…`, `s3://…`) whose scheme selects the driver, so a read re-derives its provider regardless of the current write toggle. That is what `legacyAzureRead` leans on.

Dataset chunks are addressed by `{projectId, datasetId, index}`. No provider is recorded anywhere — `Dataset.contentLayout` records the FORMAT (`postgres` / `s3_jsonl`), not the backend — so the selector has nothing to consult but the live toggle. There is no scheme dispatch to fall back on.

## Scope

Not Azure-specific and not introduced by #6181: `s3 -> local` and `local -> s3` strand chunks the same way. It surfaces now because `legacyAzureRead` documents a migration path for stored objects, which implies datasets are covered too. They are not.

## Blast radius

Only installs that switch storage backend with datasets already written. No data loss — the chunks are intact and become reachable again if the toggle is reverted.

## Fix sketch

Per-dataset storage provenance: record the backend (and account/bucket) on the dataset at write time, and have `getDatasetStorage` resolve from that rather than from the live toggle. Needs a schema field plus a backfill for existing rows, defaulting to whatever the install was using when they were written.

The limitation is documented in the `getDatasetStorage` docblock so the gap is visible at the point where the wrong assumption is easiest to make.

---

## Current state (2026-08-04) — mitigated by #6332, structural gap still open

**PR #6332 mitigates this but does not close it.** The new migration task copies dataset chunks under their deterministic `{projectId, datasetId, index}` keys and verifies them at the destination before cutover, in both directions — so a *controlled* provider switch no longer strands chunks. Proven live: a dataset whose chunks were born in Azure was migrated to S3 and read back through the UI (500 rows), and the reverse.

[...]

PR #6181, opened 2026-07-25
feat(stored-objects): sign in to Azure Blob without a shared account key

sergioestebance:
## For humans

Some companies forbid storing a password for their cloud file storage — the security team turns that option off account-wide. Until now LangWatch could only talk to Azure's file storage by holding such a password, so those companies could not use it at all. This change lets LangWatch prove who it is using the identity the cloud already gives it, with no password stored anywhere. Nothing changes for anyone already running LangWatch: the old way stays the default, and you opt into the new one.

---

Implements langwatch/langwatch-saas#899. Also closes langwatch/langwatch-saas#900.

langwatch/langwatch#6092 (the Azure write destination) has since **merged**, so this now targets `main` directly and the diff is the identity work alone.

## Why this is parity, not a new feature

Amazon S3 has had keyless sign-in since langwatch/langwatch#4058: `createS3Client` falls through to the AWS SDK provider chain (IRSA, instance profile, ECS task role). Azure was hand-rolled REST to avoid the SDK dependency, and that trade is exactly what cost us this. Enterprises that set `allowSharedKeyAccess=false` — commonly via Azure Policy — could not use the Azure backend at all: not degraded, every request refused.

`AZURE_BLOB_AUTH_MODE` now selects `sharedKey` (default, unchanged) or `workloadIdentity` / `managedIdentity` / `azureCli`, which present an Entra bearer token and store no secret anywhere.

## Design

- **Explicit toggle**, never inferred from which vars happen to be set — same rule as `STORED_OBJECTS_BACKEND` in langwatch/langwatch#4133.
- **No credential fallback, ever.** A misconfigured identity fails loud. Falling back to an account key would reverse the operator's security decision at the moment they are least likely to notice, and the reversal would look like success. Recovery happens *within* the chosen identity: refresh early, retry once on 401, never retry a 403.
- **One resolver** owns the credential decision for all consumers. Three sites decided independently before; two would have crashed without a key.
- **AKS only** for federated Kubernetes. Webhook-free clusters use `managedIdentity` or `sharedKey`.

## What the review caught

A full review ran before this was marked ready. It found one blocker I had introduced, and I would rather record it than bury it:

[...]

PR #6332, opened 2026-07-29
feat(storage): add Azure parity and safe provider migration

sergioestebance:
## For humans

Today LangWatch can only keep its files — uploaded datasets, media pulled out of
traces, large queued payloads — in Amazon S3. Customers who run everything on
Microsoft Azure had no supported place to put them.

This makes Azure a real, first-class option, and adds a safe way to move an
existing installation from one to the other. The move is deliberately manual and
reversible: it copies everything first, checks every copied file against a
fingerprint of the original, and refuses to switch over while anything is still
being written. It never deletes the originals, so if the switch goes badly you
can go back.

# Related issues

- Resolves langwatch/langwatch-saas#899
- Closes langwatch/langwatch-saas#901
- Mitigates, but does not close, langwatch/langwatch#6323

Stacked on langwatch/langwatch#6181 — **the base is `issue6087/azure-blob-entra-identity-auth`, not `main`.**

## What this delivers

Azure Blob is now a valid global object-storage destination alongside S3 for stored objects, dataset chunks, and durable GroupQueue payloads.

- Keeps Azure keyless/workload-identity authentication intact.
- Registers Azure lazily, so incomplete inactive Azure configuration cannot break S3 traffic.
- Adds explicit S3 and Azure Helm selectors while preserving the existing S3 fallback.
- Adds parity and regression coverage for datasets, queues, stored objects, credentials, and Helm identity configuration.
- Corrects documentation that previously over-promised legacy cross-provider reads.

## Operator migration command

Adds an explicit, resumable S3↔Azure operator workflow:

```bash
pnpm run task migrateObjectStorage plan
pnpm run task migrateObjectStorage copy

# Stop all app and worker read/write traffic, drain GroupQueue, then:
OBJECT_STORAGE_MIGRATION_WRITES_PAUSED=1 \
OBJECT_STORAGE_MIGRATION_READS_PAUSED=1 \
  pnpm run task migrateObjectStorage finalize

# Deploy the target provider while traffic remains paused, then:
pnpm run task migrateObjectStorage verify
```

The migration:

- uses migration-only source and target credentials;
- validates the active provider and exact bucket/account/container/endpoint addresses;
- copies without deleting the source;
- verifies content hashes and repairs/resumes partial runs;
- uses bounded stable pagination;
- migrates stored objects and deterministic dataset chunks without a schema change;
- audits durable queues across Redis Cluster masters;
- excludes BYOC projects;
- refuses finalization while reads or writes are active, datasets are active/invalid, or durable queue payloads remain.

A dataset with a null or invalid `chunkCount` is reported as a blocker instead of aborting the read-only plan. Copy skips it and migrates eligible data; finalize still refuses.

## Important cutover contract

This supports a **controlled migration**, not a bare provider toggle.

1. Pause all application and worker read/write traffic and drain GroupQueue.
2. Finalize with both pause acknowledgements set.
3. Change `STORED_OBJECTS_BACKEND` and the active credentials, then deploy the target provider while traffic remains paused. **Keep the source provider's read credentials configured** (Helm: `app.dataplane.legacyS3ReadBucket` after S3→Azure, `app.dataplane.legacyAzureRead` after Azure→S3): finalization rewrote addresses to the destination scheme, so a rollback that also drops those credentials makes the rewritten objects unreadable — with the flag kept, rollback is fully non-destructive.
4. Run `verify` against the active target provider.
5. Resume traffic only after verification succeeds.
6. Retain the source until the rollback window is complete.

Issue langwatch/langwatch#6323 remains open because persisted datasets still have no per-dataset storage provenance. The migration makes a controlled cutover safe; an unplanned toggle can still strand old chunks.

## Known limitations

- Azure does not offer the same cross-origin presigned upload path as S3 here, so large dataset uploads stream through the application.
- No MinIO-specific behavior or dependency is introduced.
- No automatic migration and no source deletion.
- No database schema change.

## Validation

Local validation on commit `788668a033`, after rebasing onto the current base:

- application and test typechecks passed (`pnpm typecheck:all`);
- the differential Biome gate passed against the merge base — 3,641 diagnostics
  on both sides, zero new violations;
- 345 unit tests across the 25 stored-objects and migration-task suites passed;
- 22 in-memory migration integration contracts passed, including abandoned
  uploads with a null `chunkCount`, read-pause enforcement, post-cutover reads,
  and post-cutover writes;
- feature parity passed with 4,626 enforced scenarios bound;
- both blocker fixes below were mutation-tested — reverting each one turns its
  test red, so neither assertion is vacuous.

Docker-backed integration contracts are delegated to this PR's GitHub Actions
run; the local environment has no container runtime, so the suites ran against
native ClickHouse, Postgres and Redis instead.

### Review follow-ups included here

Three commits close findings from a review of this branch's own migration task:

- the destination-scheme-mismatch blocker aborted the whole run without naming
  the offending row (the two addresses differ only in the bucket/account, which
  URI redaction masks, so quoting them printed two identical strings);
- the cutover-audit Redis handshake had no timeout and no cleanup on its error
  path, so a wedged cluster hung `finalize` silently and a failed handshake
  leaked a connection that retried forever;
- an assertion interpolated a generated id into a `RegExp`, where one
  metacharacter would have made it silently stop matching.

## Deployment impact

Existing S3 installations do not migrate automatically. Azure becomes an explicit supported provider, and incomplete inactive Azure variables no longer affect S3 requests.

For an existing S3 customer choosing Azure, operators must run the documented plan/copy/finalize/verify sequence and coordinate the read/write pause around cutover. Rollback keeps using the retained source objects; the migration never deletes them.

### DIFF
--- a/.github/workflows/langwatch-app-ci.yml
+++ b/.github/workflows/langwatch-app-ci.yml
@@ -111,6 +111,14 @@ jobs:
               - 'pnpm-workspace.yaml'
               - 'package.json'
               - '.npmrc'
+              # NOTE: 'charts/**' is deliberately NOT listed. Integration suites
+              # do render the chart and assert on the output, so a chart-only
+              # edit can break a test under platform/app/ without touching a file
+              # there — but this one flag gates nine jobs (typecheck, every unit
+              # and integration shard, e2e), so adding it would run the entire
+              # app suite for a comment change in values.yaml. That trade is a
+              # repo-wide cost decision, not part of this change; tracked
+              # separately.
               - '.github/workflows/langwatch-app-ci.yml'
               - '.github/.release-please-manifest.json'
               - '.github/scripts/**'
@@ -438,6 +446,18 @@ jobs:
           done
           exit 1
 
+      # The chart-rendering integration tests shell out to `helm template`, and
+      # skip themselves when helm or the chart dependencies are missing. That
+      # honesty is only honest if CI actually supplies both — otherwise the
+      # suite reports "skipped", the job exits 0, and every guard it contains
+      # is silently unenforced. REQUIRE_HELM_TESTS below turns a skip here into
+      # a failure, so this setup can never quietly rot.
+      - uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
+      - name: Add chart dependency repositories
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+          helm repo add langwatch https://langwatch.github.io/langwatch
+
       - name: Run integration tests (shard ${{ matrix.shard }}/4)
         id: integration-tests
         working-directory: platform/app
@@ -450,6 +470,9 @@ jobs:
         # worker (execArgv is not), so it raises the heap where it's needed.
         env:
           NODE_OPTIONS: --max-old-space-size=8192
+          # Fail loudly if the helm-backed suites cannot run here, rather than
+          # skipping into a green job.
+          REQUIRE_HELM_TESTS: "1"
         run: |
           # shardFailureReporter feeds the globalSetup hard-floor so a wedged
           # shard that already printed failures exits 1, not a masked 0.
--- a/dev/docs/TESTING_PHILOSOPHY.md
+++ b/dev/docs/TESTING_PHILOSOPHY.md
@@ -174,7 +174,28 @@ The scenario title in the feature file should match the `it()` description in th
 | `@integration` | Component/boundary test | Testing rendering, API calls, DB queries |
 | `@regression` | Prevents a previously-fixed bug from recurring | Bug fix scenarios — must fail without the fix |
 | `@e2e` | Stable core flow (deprioritized) | Only for the 5-10 stable happy-path tests |
-| `@manual` | Verified by hand against real infrastructure, cannot run in CI | Needs a real cloud account/subscription. Still **requires a binding** to a test that self-skips without credentials, so the verification keeps an anchor in the codebase. Record the date and result in a comment above the scenario |
+| `@manual` | Verified by hand against real infrastructure that CI cannot reach | Last resort only — see the rules below the table |
+
+`@manual` is deliberately hard to qualify for, because it is also the easiest
+tag to abuse: anything marked manual is a test nobody runs automatically, and
+an agent (or a rushed human) reaching for it to dodge writing a real test
+defeats the whole pyramid. All of these must hold, or the scenario belongs in
+an automated tier:
+
+- The scenario **cannot** run in CI for a structural reason — it needs a real
+  cloud account/subscription or credentials CI cannot hold. "Hard to write",
+  "slow", or "needs Docker" never qualify.
+- Everything automatable about the behavior is already covered by `@unit` /
+  `@integration` scenarios (e.g. against an emulator); `@manual` covers only
+  the residue that specifically proves the real-infrastructure behavior.
+- It still **requires a binding** to a test that self-skips without
+  credentials, so the verification keeps an anchor in the codebase — the
+  parity checker enforces this like any automated tier.
+- The date and result of the last hand-run are recorded in a comment above the
+  scenario.
+
+Expect a reviewer to challenge every new `@manual`; the burden of proof is on
+the tag, not on the reviewer.
 
 Bug-fix feature specs should use `@regression` (alongside `@unit` or `@integration` for pyramid level):
 
--- a/dev/docs/TESTING_PHILOSOPHY.md
+++ b/dev/docs/TESTING_PHILOSOPHY.md
@@ -174,28 +174,6 @@ The scenario title in the feature file should match the `it()` description in th
 | `@integration` | Component/boundary test | Testing rendering, API calls, DB queries |
 | `@regression` | Prevents a previously-fixed bug from recurring | Bug fix scenarios — must fail without the fix |
 | `@e2e` | Stable core flow (deprioritized) | Only for the 5-10 stable happy-path tests |
-| `@manual` | Verified by hand against real infrastructure that CI cannot reach | Last resort only — see the rules below the table |
-
-`@manual` is deliberately hard to qualify for, because it is also the easiest
-tag to abuse: anything marked manual is a test nobody runs automatically, and
-an agent (or a rushed human) reaching for it to dodge writing a real test
-defeats the whole pyramid. All of these must hold, or the scenario belongs in
-an automated tier:
-
-- The scenario **cannot** run in CI for a structural reason — it needs a real
-  cloud account/subscription or credentials CI cannot hold. "Hard to write",
-  "slow", or "needs Docker" never qualify.
-- Everything automatable about the behavior is already covered by `@unit` /
-  `@integration` scenarios (e.g. against an emulator); `@manual` covers only
-  the residue that specifically proves the real-infrastructure behavior.
-- It still **requires a binding** to a test that self-skips without
-  credentials, so the verification keeps an anchor in the codebase — the
-  parity checker enforces this like any automated tier.
-- The date and result of the last hand-run are recorded in a comment above the
-  scenario.
-
-Expect a reviewer to challenge every new `@manual`; the burden of proof is on
-the tag, not on the reviewer.
 
 Bug-fix feature specs should use `@regression` (alongside `@unit` or `@integration` for pyramid level):
 
--- a/dev/docs/TESTING_PHILOSOPHY.md
+++ b/dev/docs/TESTING_PHILOSOPHY.md
@@ -174,6 +174,7 @@ The scenario title in the feature file should match the `it()` description in th
 | `@integration` | Component/boundary test | Testing rendering, API calls, DB queries |
 | `@regression` | Prevents a previously-fixed bug from recurring | Bug fix scenarios — must fail without the fix |
 | `@e2e` | Stable core flow (deprioritized) | Only for the 5-10 stable happy-path tests |
+| `@manual` | Verified by hand against real infrastructure, cannot run in CI | Needs a real cloud account/subscription. Still **requires a binding** to a test that self-skips without credentials, so the verification keeps an anchor in the codebase. Record the date and result in a comment above the scenario |
 
 Bug-fix feature specs should use `@regression` (alongside `@unit` or `@integration` for pyramid level):
 
--- a/specs/features/scenarios/azure-blob-workload-identity.feature
+++ b/specs/features/scenarios/azure-blob-workload-identity.feature
@@ -402,18 +402,6 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
     Then it renders successfully
     And no account-key environment variable is emitted
 
-  # An annotation-less account the chart created fails exactly like a pod with
-  # no webhook label, one layer down: it renders, the pods are healthy, and the
-  # webhook has no identity to bind them to.
-  @integration
-  Scenario: The chart refuses a created service account with no identity annotation
-    Given a token-based auth mode requiring a federated identity
-    And the chart is asked to create the service account
-    And no identity client-id annotation is configured on it
-    When the chart renders
-    Then rendering fails with an error naming the client-id annotation
-    But naming a pre-existing service account instead still renders
-
   # The chart cannot read a Secret, so an endpoint supplied that way has an
   # unknowable hostname. Assuming the public cloud is the one guess that fails
   # silently — the deploy succeeds and the first storage call is refused.
--- a/specs/features/scenarios/azure-blob-workload-identity.feature
+++ b/specs/features/scenarios/azure-blob-workload-identity.feature
@@ -470,9 +470,9 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
   # VERIFIED 2026-07-26 against a real Azure account with
   # allowSharedKeyAccess=false: shared key returned 403
   # KeyBasedAuthenticationNotPermitted, the same operations succeeded on a
-  # bearer token. Stays @manual because it needs a subscription and cannot
-  # run in CI; bound to a test that self-skips without credentials.
-  @manual
+  # bearer token. Needs a real subscription, so the bound test self-skips
+  # in CI and runs only where LANGWATCH_TEST_AZURE_* credentials are set.
+  @integration
   Scenario: Blobs round-trip against a real storage account with shared-key access disabled
     Given a real Azure storage account that forbids shared-key access
     And an identity holding the blob data role on that account
--- a/specs/features/scenarios/azure-blob-workload-identity.feature
+++ b/specs/features/scenarios/azure-blob-workload-identity.feature
@@ -443,10 +443,23 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
     Given the active provider is awsS3 with legacy Azure reads kept enabled
     And the Azure auth mode is workloadIdentity
     When the chart renders
-    Then no stored-objects backend toggle selects Azure for writes
+    Then the S3 write configuration renders exactly as on a plain S3 install
+    And no stored-objects backend toggle selects Azure for writes
     And the Azure connection settings are still emitted for reads
     And the app and workers pods still carry the workload-identity webhook label
     And the cron pods still do not carry it
+
+  # Each legacy read flag belongs to one migration direction. Set alongside
+  # the provider it migrates away from it configures nothing — the chart
+  # rejects the contradiction instead of rendering a silent no-op. An awsS3
+  # install with an empty bucket is rejected for the same reason: it would
+  # render S3_BUCKET_NAME blank and every write would fall back to local
+  # storage while the operator believes S3 is live.
+  @integration
+  Scenario: The chart rejects a legacy read flag aimed at the active provider
+    Given a chart configuration whose legacy read flag targets the provider that is already active
+    When the chart renders
+    Then rendering fails naming the contradictory flag
 
   @integration
   Scenario: Installs that do not use Azure render exactly as they did before
--- a/specs/features/scenarios/azure-blob-workload-identity.feature
+++ b/specs/features/scenarios/azure-blob-workload-identity.feature
@@ -267,11 +267,16 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
   # Byte paths beyond the driver
   # ---------------------------------------------------------------
 
+  # Scoped to what a unit test can actually prove: no shared key is consulted
+  # anywhere on the write path, and the destination and driver agree so a write
+  # is never rejected as an unregistered scheme. The byte-level round-trip in a
+  # token mode is proven against a real Entra-authenticated account further
+  # down — no emulator offers OAuth, so it cannot be proven here.
   @unit
-  Scenario: Stored-objects writes succeed in a token-based mode
+  Scenario: A token-mode write path resolves without consulting a shared key
     Given STORED_OBJECTS_BACKEND is azure in a token-based auth mode
-    When byte content is externalized for a project
-    Then the object is written and an azure-blob URI is persisted
+    When the write destination for a project is resolved
+    Then it resolves to Azure and a matching Azure driver is registered
     And no shared-key configuration is consulted
 
   @unit
@@ -304,16 +309,34 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
     When the write destination and the driver registration are both resolved
     Then the write destination is S3 while the Azure driver stays registered
 
+  # What a unit test can prove today: the selector reaches Azure dataset
+  # storage without touching the account key. It used to dereference
+  # `env.AZURE_BLOB_ACCOUNT_KEY!` and die inside Buffer.from(undefined) — a
+  # crash rather than a config error.
+  @unit
+  Scenario: Dataset storage is selected without dereferencing an absent account key
+    Given STORED_OBJECTS_BACKEND is azure in a token-based auth mode
+    When dataset storage is resolved for a project
+    Then it selects the Azure dataset storage
+    And no code path dereferences an absent account key
+
+  # The chunked round-trip itself stays unproven in a token mode: the emulator
+  # is shared-key only, so there is nowhere to run it outside a real
+  # Entra-authenticated account. Deliberately carries NO binding — the tag and
+  # an annotation claiming coverage cannot both be true.
   @integration @unimplemented
-  Scenario: Dataset uploads work in a token-based mode
+  Scenario: Dataset uploads round-trip in a token-based mode
     Given STORED_OBJECTS_BACKEND is azure in a token-based auth mode
     When a dataset is created, appended to, and read back
     Then the chunked content round-trips through Azure Blob
-    And no code path dereferences an absent account key
 
+  # The concern here is provider routing, not authentication: the durable tier
+  # is labelled `tier: "s3"`, so this proves it still reaches Azure on an
+  # install with no S3 at all. It runs against the emulator, which is shared-key
+  # only, so it deliberately does NOT claim to exercise a token mode.
   @integration
-  Scenario: The groupQueue durable blob tier works in a token-based mode
-    Given the durable blob tier resolves an azure destination in a token-based mode
+  Scenario: The groupQueue durable blob tier works on an Azure-only install
+    Given the durable blob tier resolves an azure destination and no S3 is configured
     When an oversized envelope is offloaded and later read back
     Then the bytes round-trip through Azure Blob
 
@@ -344,6 +367,18 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
     When the chart renders
     Then the cron pods name no service account
     And the cron pods do not carry the workload-identity webhook label
+
+  # An annotation-less account the chart created is the same failure as a pod
+  # with no label, one layer down: it renders, the pods are healthy, and the
+  # webhook has nothing to bind them to.
+  @integration
+  Scenario: The chart refuses a created service account with no identity annotation
+    Given workloadIdentity auth is selected
+    And the chart is asked to create the service account
+    And no identity client-id annotation is configured on it
+    When the chart renders
+    Then rendering fails with an error naming the client-id annotation
+    But naming a pre-existing service account instead still renders
 
   @integration
   Scenario: The chart refuses a workload-identity install with no service account
--- a/.github/workflows/langwatch-chart.yml
+++ b/.github/workflows/langwatch-chart.yml
@@ -135,6 +135,39 @@ jobs:
       - name: "assert rendered email configuration"
         run: ./tests/email-gateway.sh
 
+  # The render matrix above proves the chart PARSES; these vitest suites
+  # assert on what it EMITS (env selection per provider, identity guards,
+  # migration postures, validation failures). They normally run inside the
+  # app CI's integration shards, but that workflow deliberately does not
+  # trigger on charts/** — so without this job a chart-only edit could break
+  # every behavioral guarantee while CI stays green.
+  behavioral:
+    name: "behavioral (rendered-output assertions)"
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: platform/app
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
+        with:
+          version: v3.12.0
+      - run: helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+      - run: helm repo add langwatch https://langwatch.github.io/langwatch
+      - uses: ./.github/actions/setup-pnpm-node
+        with:
+          install: "@langwatch/web..."
+      - name: Generate required files
+        run: pnpm start:prepare:files
+      - name: Run helm behavioral suites
+        env:
+          # Turns a silent helm-unavailable skip into a hard failure — this
+          # job exists precisely so these suites cannot quietly not run.
+          REQUIRE_HELM_TESTS: "1"
+        run: |
+          pnpm test:unit run src/server/stored-objects/__tests__/helm-and-docs-shape.unit.test.ts
+          pnpm test:integration run src/server/stored-objects/__tests__/helm-azure-identity.integration.test.ts
+
   e2e:
     name: "e2e: ${{ matrix.name }}"
     runs-on: ubuntu-latest
--- a/.github/workflows/langwatch-chart.yml
+++ b/.github/workflows/langwatch-chart.yml
@@ -159,6 +159,11 @@ jobs:
           install: "@langwatch/web..."
       - name: Generate required files
         run: pnpm start:prepare:files
+      # The integration config's globalSetup runs ClickHouse migrations via
+      # goose before any test file executes, helm-backed or not.
+      - name: Install Goose migration tool
+        working-directory: .
+        run: bash .github/scripts/install-goose.sh
       - name: Run helm behavioral suites
         env:
           # Turns a silent helm-unavailable skip into a hard failure — this
--- a/charts/langwatch/templates/_helpers.tpl
+++ b/charts/langwatch/templates/_helpers.tpl
@@ -242,7 +242,19 @@ app.kubernetes.io/instance: {{ .Release.Name }}
   {{- if not (has .Values.app.dataplane.provider (list "awsS3" "azureBlob")) }}
     {{- $errors = append $errors (printf "app.dataplane.provider is %q — must be one of awsS3, azureBlob" .Values.app.dataplane.provider) }}
   {{- end }}
+  {{/* Each legacy read flag belongs to exactly one migration direction. Set
+       alongside the provider it "migrates away from", it is a no-op the
+       operator almost certainly did not intend — reject rather than ignore. */}}
+  {{- if and (eq .Values.app.dataplane.provider "azureBlob") .Values.app.dataplane.legacyAzureRead }}
+    {{- $errors = append $errors "app.dataplane.legacyAzureRead is set but azureBlob is already the active provider — the flag is for keeping Azure reads alive AFTER moving writes to S3 (provider awsS3). Remove it." }}
+  {{- end }}
+  {{- if and (eq .Values.app.dataplane.provider "awsS3") .Values.app.dataplane.legacyS3ReadBucket }}
+    {{- $errors = append $errors "app.dataplane.legacyS3ReadBucket is set but awsS3 is already the active provider — the flag is for keeping S3 reads alive AFTER moving writes to Azure (provider azureBlob). Remove it." }}
+  {{- end }}
   {{- if eq .Values.app.dataplane.provider "awsS3" }}
+    {{- if empty .Values.app.dataplane.bucket }}
+      {{- $errors = append $errors "app.dataplane.provider is awsS3 but app.dataplane.bucket is empty — S3_BUCKET_NAME would render blank and every write would fall back to local storage" }}
+    {{- end }}
     {{- if .Values.app.dataplane.providers.awsS3.endpoint.secretKeyRef.name }}
       {{- if empty .Values.app.dataplane.providers.awsS3.endpoint.secretKeyRef.key }}
         {{- $errors = append $errors "app.dataplane.providers.awsS3.endpoint.secretKeyRef.name is set but key is empty" }}
@@ -875,8 +887,29 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 # why the connection settings below can outlive it (see legacyAzureRead).
 - name: STORED_OBJECTS_BACKEND
   value: "azure"
+{{- else }}
+- name: STORED_OBJECTS_BACKEND
+  value: "s3"
+- name: USE_S3_STORAGE
+  value: "true"
+# Emit S3_BUCKET_NAME — the app/server reads this name across all
+# storage code paths (storage.ts, stored-objects.service.ts,
+# env-create.mjs). The legacy `S3_BUCKET` env was a no-op for every
+# vanilla helm install because nothing read it; emitting it was a
+# silent bug that this fix resolves by aligning on S3_BUCKET_NAME.
+- name: S3_BUCKET_NAME
+  value: {{ .Values.app.dataplane.bucket | quote }}
+{{- include "langwatch.secretOrValue" (dict "envName" "S3_ENDPOINT" "fieldValues" .Values.app.dataplane.providers.awsS3.endpoint) }}
+{{- include "langwatch.secretOrValue" (dict "envName" "S3_ACCESS_KEY_ID" "fieldValues" .Values.app.dataplane.providers.awsS3.accessKeyId) }}
+{{- include "langwatch.secretOrValue" (dict "envName" "S3_SECRET_ACCESS_KEY" "fieldValues" .Values.app.dataplane.providers.awsS3.secretAccessKey) }}
+{{- include "langwatch.secretOrValue" (dict "envName" "S3_KEY_SALT" "fieldValues" .Values.app.dataplane.providers.awsS3.keySalt) }}
 {{- end }}
-{{/* Azure connection settings are emitted when Azure is the active write
+{{/* The active provider's WRITE configuration above never depends on the
+     legacy read flags below — an Azure->S3 migration must configure S3
+     writes exactly like a plain S3 install, or new writes silently fall
+     back to local storage while the operator believes S3 is live.
+
+     Azure connection settings are emitted when Azure is the active write
      backend OR when legacyAzureRead is set for an Azure->S3 migration. The
      app's driver registration resolves these for READS independently of the
      write toggle, so keeping them after the switch is what lets already
@@ -899,7 +932,12 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- if or .Values.app.dataplane.providers.azureBlob.tokenAudience.value .Values.app.dataplane.providers.azureBlob.tokenAudience.secretKeyRef.name }}
 {{- include "langwatch.secretOrValue" (dict "envName" "AZURE_BLOB_TOKEN_AUDIENCE" "fieldValues" .Values.app.dataplane.providers.azureBlob.tokenAudience) }}
 {{- end }}
-{{- if .Values.app.dataplane.legacyS3ReadBucket }}
+{{- end }}
+{{/* Gated on azureBlob being ACTIVE, not just on the bucket being set: when
+     S3 is the active provider its write block above already emits
+     S3_BUCKET_NAME, and a second entry with a different value would be a
+     duplicate env var whose winner is undefined. */}}
+{{- if and (eq .Values.app.dataplane.provider "azureBlob") .Values.app.dataplane.legacyS3ReadBucket }}
 # Retains reads for persisted s3:// stored-object URIs and legacy consumers
 # that already carry an S3 bucket/key after writes switch to Azure. This does
 # not route provider-derived dataset chunks or GroupQueue durable payloads:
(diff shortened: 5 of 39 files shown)


langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** Resolves langwatch/langwatch-saas#899, closes langwatch/langwatch-saas#901. **Stacked on langwatch/langwatch#6181, base is that branch not main**, and Drew holds langwatch/langwatch#6181, so the same reviewer sees both. Review the parent first.

**Questionable Premise, in a good way:** the honest finding here is that langwatch/langwatch#6181 shipped guards which "were never actually running in our test pipeline, so they were protecting nothing." That is the interesting part of this PR. A guard that silently does not execute is worse than no guard, because it buys confidence without providing protection.

**Top inspection targets**
1. `Must Check` - the guards now genuinely run and genuinely fail. Do not take a passing pipeline as evidence; confirm there is a test that makes a guard fire (a deliberately broken config that must be rejected). Otherwise this PR repeats the original defect one level up.
2. `Must Check` - why they were not running. If the cause was a pattern that silently excluded the suite (a skip predicate, a path filter, a lane that never executes), the same cause may be hiding other dormant tests. Aryan's langwatch/langwatch#6330 found exactly this class: 14 suites, 78 tests, never executed anywhere.
3. `Probably Fine` - the smaller review findings bundled alongside.

**Production risk:** none directly from this diff. The risk it addresses is a misconfigured Azure deployment reaching production because the guard that should have stopped it never ran.

1. Is there now a test that fails when a guard is removed?
2. What made the guards inert, and could the same mechanism be silencing other suites?

sergioestebance:
## Ruthless review — full diff (Azure parity + provider migration), S3-robustness focus

**Verdict: no surviving P0/P1.** One P1 was confirmed and fixed on the branch during review (`82e1edfed`); the rest is non-blocking.

### Resolved during review

**P1 (confirmed, fixed) — Azure→S3 Helm mode rendered no S3 write configuration.**
`charts/langwatch/templates/_helpers.tpl` placed the S3 write env in the `else` of "Azure active OR legacyAzureRead", so `provider=awsS3` + `legacyAzureRead=true` emitted Azure read credentials but no `STORED_OBJECTS_BACKEND=s3`, `S3_BUCKET_NAME`, `USE_S3_STORAGE`, or S3 credentials — new writes silently fell back to local storage. The test at `helm-azure-identity.integration.test.ts` blessed the defect by asserting `STORED_OBJECTS_BACKEND` was absent.

Fixed exactly as the finding recommended: the active provider's write block renders on the provider alone; legacy read blocks are additive. The migration render now carries S3 writes + Azure reads simultaneously (executed: both deployments emit `STORED_OBJECTS_BACKEND=s3`, `USE_S3_STORAGE=true`, `S3_BUCKET_NAME` alongside `AZURE_BLOB_*`). Three contradictions that used to render as silent no-ops now fail at render time: `legacyAzureRead` with azureBlob active, `legacyS3ReadBucket` with awsS3 active, awsS3 with an empty bucket.

**S3 regression proof:** a plain-S3 render of the base branch vs this branch differs by exactly two lines — the added explicit `STORED_OBJECTS_BACKEND=s3` on app and workers. Everything else is byte-identical.

### Non-blocking findings

- **P2 (fixed in `82e1edfed`)** — `stored-objects.repository.ts` `findLiveRowsByProjectPage`: the dedup subquery had no keyset filter, so every 250-row page rescanned the whole project (O(pages × rows)). The subquery now carries the same `id > {afterId}` predicate.
- **P2 (accepted)** — `OBJECT_STORAGE_MIGRATION_WRITES_PAUSED` / `READS_PAUSED` are operator attestations, not enforced pauses. The runbook says so explicitly; acceptable for an offline operator task, but worth remembering these flags verify nothing by themselves.
- **P3** — `migrateObjectStorage.unit.test.ts` uses a flat `it` list rather than the repo's nested `given/when` describe structure.
- **P3** — the integration test double `pageById` sorts with `localeCompare` but filters with `>` (code-unit order); harmless today, could confuse a future pagination test with non-ASCII ids.

### Traced

- Registry lazy Azure factory: S3/file drivers stay eager and identical to base; misconfigured-but-inactive Azure only raises when an `azure-blob://` URI is dispatched, and raises the *actionable* config error when Azure is the selected backend (pinned by tests).
- Finalize ordering: pause gates (writes AND reads) → dataset blockers (incl. `invalid-chunk-count` for abandoned uploads) → read-only queue audit across all cluster masters → delta copy → full digest verify → append-only publish with strictly newer `inserted_at` (repository `insert` preserves the row's timestamp — verified). No phase deletes source bytes; resume paths covered by tests.
- GQ2 wire format untouched — audit reads envelope headers only; redis-tier staged refs correctly do not block (they live in Redis, not object storage).
- Error contract: operator task throws plain `Error`/`MigrationBlockedError` — correct per error-handling doc (no customer boundary here); storage URIs redacted in error messages.
- Tenancy: every ClickHouse query is `project_id`-scoped with the IN-tuple dedup pattern (no `LIMIT 1 BY` on heavy columns).

Gate at `82e1edfed`: typecheck + typecheck:tests clean, parity 3701 bound, helm suites 30/30 + 20/20, migration suites 18 + 13 + 11, biome clean.

sergioestebance:
Blessed. 🙏 One storage contract now serves two clouds: Azure carries every workload S3 carried, and the road between them is a verified, resumable, append-only migration that never deletes a source byte.

> "See, I am doing a new thing! Now it springs up; do you not perceive it? I am making a way in the wilderness and streams in the wasteland." — Isaiah 43:19

Traced the lazy driver registration, the finalize gate chain (pauses → datasets → queues → verify → publish), the append-only ClickHouse versioning, and the chart's full provider matrix — plain-S3 renders byte-identical to base except the explicit backend selector. Ship it — residuals are P3s (flat test structure, a test-double comparator nit) and the documented fact that the pause flags are operator attestations.

sergioestebance:
## /prove-them-wrong — real-Azure validation (tiers 1–3: driver, full app, bidirectional migration)

**Verdict: 14/16 proven, 0 disproven, 2 unverified — on `https://app.<worktree-slug>.langwatch.localhost` (local haven stack, account `admin@haven.localhost`) against real Azure Blob Storage (`westeurope`, one account with shared-key auth disabled) and the dev S3 bucket.**

Two real defects were found by the rehearsal and fixed on this PR (`54cfccd63`); both were invisible to the in-memory test doubles.

| 1 | Driver round-trips against a real account with **shared-key access disabled** (Entra bearer) | realazure suite 7/7, incl. both Entra tests | PASS |
| 2 | sharedKey mode signs correctly against a real key-enabled account | same suite, host-style signing tests | PASS |
| 3 | Token mode with **no data-plane role** fails loud with remediation copy, never falls back | 403 + "Grant the Storage Blob Data Contributor role…" | PASS |
| 4 | Azure-only app: trace media extracted → Azure → **rendered in the UI** | screenshot below; `/api/files` served 98,312 bytes, correct PNG magic | PASS |
| 5 | Dataset upload on Azure backend (app-streamed staging) | 500-row CSV → `s3_jsonl` chunk in the Azure container → rows render | PASS |
| 6 | GroupQueue durable payloads ride Azure | 310,950-byte GQ2 blob (22-char content hash) written + decoded back by the worker | PASS |
| 7 | CSP `connect-src` carries the Azure endpoint | dev builds ship no CSP; pinned by unit tests only | UNVERIFIED |
| 8 | `plan` / `copy` refuse when the configured source ≠ active app storage | endpoint-mismatch guard fired on first attempt | PASS |
| 9 | `copy` verifies SHA-256 and repairs corruption | destination blob truncated to 36 bytes → rerun restored 50,576 verified bytes | PASS |
| 10 | Already-verified objects are skipped, resumably | reruns re-copied nothing (S3 timestamps unchanged) | PASS |
| 11 | Finalize refuses: unpaused writes / unpaused reads | both refusals observed with exact messages | PASS |
| 12 | Finalize refuses outstanding queue work | fired live on a real delayed `billing_report` job; queue drained once the app stopped, then finalize passed | PASS |
| 13 | BYOC projects are excluded from a global migration | with `DATAPLANE_S3__*` set: 0 blobs copied; without: 1 — decisive A/B | PASS |
| 14 | Finalize publishes append-only; **cutover S3→Azure works end-to-end** | 5/5 latest rows `azure-blob://`, verify green, migrated object renders in UI | PASS |
| 15 | **Reverse migration Azure→S3** restores everything | 5 objects + the Azure-born dataset chunk on S3, verify green, UI reads both | PASS |
| 16 | Source bytes never deleted | all 7 Azure blobs present after reverse cutover; S3 originals present after forward cutover | PASS |

Trace media rendered from the shared-key-disabled Azure account

500-row dataset served from an Azure Blob chunk

### Defects found and fixed (`54cfccd63`)

1. **The migration task could not run against any real database.** `listDatasetsPage` queried `Dataset` without `projectId`; the Prisma multitenancy middleware rejects that, so `plan` crashed on first contact with real Postgres. Dataset inventory is now per-project (which also makes BYOC exclusion structural — excluded projects are never queried).
2. **A dataset already at the destination aborted the whole copy.** Chunks uploaded while the destination provider was briefly active (the langwatch/langwatch#6323 posture) have no source copy; `copy` died on `ObjectNotFoundError` instead of counting them verified. Now tolerated when the destination has the chunk; a chunk missing from **both** providers still blocks. New scenario + tests bind both fixes (parity 3702).

### Findings (not blocking)

- The finalize queue gate fired on a transient delayed `billing_report` job while the app was live — correct, but operators should expect it: **the queue only reads empty after app + workers are actually stopped**, exactly as the runbook orders. Worth keeping in mind for installs with many orgs.
- Media extraction is dark by default (`release_trace_media_extraction`) and GQ2 envelope writes require `GROUP_QUEUE_ENVELOPE_WRITES_ENABLED=true` — an Azure parity *demo* needs both flags on; without them those two byte paths simply don't run.
- Legacy Traces view shows an extracted image only as a metadata ref (no ``); Trace Explorer renders it. Pre-existing view difference, not a storage defect.
- Pre-existing React `<p>` nesting hydration warnings on the datasets empty state — unrelated to this PR.

### Not covered

- **CSP runtime observation** (claim 7): needs a production build; the mapping is unit-pinned.
- **workloadIdentity / managedIdentity modes end-to-end**: need AKS / an Azure VM — explicitly out of scope for this run; `azureCli` exercises the same credential code path with a different token source. This is the remaining tier before calling the AKS story verified.
- Redis Cluster multi-master queue audit ran against a single-node Redis; cluster fan-out is covered by unit tests only.

Every phase ran with real Entra bearer tokens against `allowSharedKeyAccess=false` — the exact posture that motivated langwatch/langwatch-saas#899.

sergioestebance:
## Review-findings validation round (user-pasted + Codex native review)

All five findings validated against the code; four confirmed and fixed, one half-confirmed and fixed at the proportionate scope. CI back to green.

| Finding | Verdict | Fix |
|---|---|---|
| P1 — cluster queue audit can read lagging replicas (`scaleReads: "all"`) → false clean audit → stranded payload after cutover | **CONFIRMED** | `e9cc7a1c7`/`0ccf45583`: audit reads through a master-only `duplicate()` of the shared connection, waiting out the duplicate's cluster handshake before reading topology (an early `nodes()` returns `[]` and would silently audit nothing). 3 regression tests |
| P1 (Codex) — migration credentials bypass the token-mode transport guards → bearer token over plaintext `http://`, sovereign endpoint without authority | **CONFIRMED** | the https + sovereign-authority guards are now one exported seam, `assertTokenModeTransportSafety`, shared by `resolveAzureCredentials` and the migration's `toAzureCredentials`. 4 tests incl. the sharedKey-unguarded case |
| P2 — chart-only changes skip the helm behavioral suites | **HALF-CONFIRMED** — `langwatch-chart.yml` already triggers a 12-config render matrix on `charts/**`; only the rendered-output *assertions* were missing | new `behavioral` job on that existing workflow runs both helm vitest suites with `REQUIRE_HELM_TESTS=1` on chart-only changes — **already green in CI** |
| P2 (Codex) — endpoint classifier is case-sensitive, uppercase public-cloud endpoint rejected as sovereign | **CONFIRMED** | classifier lowercases before the suffix test, mirroring the runtime check; render test added |
| P3 — duplicate "created service account with no identity annotation" scenario double-binds under title-based parity | **CONFIRMED** | later duplicate removed; parity 3701 bound |

Gate at `e9cc7a1c7`: typecheck + typecheck:tests clean · parity 3701 · helm suite 31/31 · stored-objects + tasks unit 370/370 · biome clean · all CI checks green (`MERGEABLE / CLEAN`), including the new chart `behavioral` job.

langwatch/src/server/stored-objects/azure-credentials.ts
```diff
@@ -266,6 +266,26 @@ function assertRequiredVariablesPresent({
...
+/**
+ * The two transport guards every token-mode credential must pass, exported as
+ * one seam so ALL construction sites share them: a bearer token must never
+ * travel a plaintext connection, and a non-public-cloud endpoint must name
+ * the identity authority its tokens come from. The migration task builds its
+ * credentials from its own OBJECT_STORAGE_MIGRATION_* namespace rather than
+ * through `resolveAzureCredentials`, and bypassing these guards there would
+ * let a migration run leak bearer tokens the app itself refuses to.
+ */
+export function assertTokenModeTransportSafety({
```
drewdrewthis, line 277: **[proof-reviewer] AC: re-run the real-Azure rehearsal at head — the current proof predates three commits on the paths it exercised.**

The only evidence in this PR that the cutover works against a real provider is the `/prove-them-wrong` run (https://github.com/langwatch/langwatch/pull/6332#issuecomment-5137630927), which lands at commit `54cfccd63`. Three commits have shipped since, each on a path that rehearsal exercised:

| Commit | Change | Rehearsal claim it underwrites |
|---|---|---|
| `8aabb5127` | refactor: dataset-chunk handling (`objectStorageMigration.ts`, +77/-66) | The rehearsal's own defect #2 fix — "a dataset already at the destination aborted the whole copy" |
| `0ccf45583` | rewrote the queue audit (`migrateObjectStorage.ts`) | Claim 12, "finalize refuses outstanding queue work" |
| `e9cc7a1c7` | this hunk — token-mode transport guards on migration credentials | Claims 1-3, real-Entra credential construction |

The rehearsal comment states it itself: *"Two real defects were found by the rehearsal and fixed on this PR (`54cfccd63`); **both were invisible to the in-memory test doubles.**"* Green CI at head (verified: all checks pass at `e9cc7a1c7`) is therefore testing precisely the layer already demonstrated to miss this class of defect. `8aabb5127` in particular is a 143-line rewrite of the code that fixed one of those two defects, and its only coverage since is fakes.

**AC:** re-run the rehearsal at head — at minimum claims 1-3 (credential construction, this hunk), 12 (queue audit), and 14-15 (bidirectional cutover) — and post the result at the current SHA. A narrower targeted run is fine; the ask is that the real-provider evidence and the shipped code be the same commit.

langwatch/src/tasks/objectStorageMigration.ts
```diff
@@ -0,0 +1,591 @@
...
+  private async *eligibleStoredObjects(
+    scope: EligibleScope,
+  ): AsyncGenerator<StoredObject> {
+    for (const projectId of [...scope.eligibleProjectIds].sort()) {
+      for await (const row of paginate((request) =>
+        this.deps.inventory.listStoredObjectsPage(projectId, request),
+      )) {
+        const scheme = row.storage_uri.slice(0, row.storage_uri.indexOf(":"));
+        if (
+          scheme === this.deps.source.scheme ||
```
drewdrewthis, line 493: **[correctness] `eligibleStoredObjects` silently drops rows and the plan under-reports with no warning.**

This filter yields only rows whose scheme is the source or the destination. Any other live row — `file://` from a local-filesystem deployment, or an `azure-blob://` row on a deployment mid-way through a previous migration — is dropped from the generator with no counter, no log, and no warning.

`buildPlan` counts only what the generator yields (`eligibleStoredObjects += 1`, line 426), and `MigrationPlan` (lines 114-123) has no excluded/skipped field — nor do `MigrationCopyReport` (136-140) or `MigrationFinalizeReport` (142-145). Confirmed the repository does not pre-filter either: `findLiveRowsByProjectPage` (`stored-objects.repository.ts:253-316`, new in this PR) selects every live row for the project unconditionally.

The operator-facing consequence is the problem: `plan` is the step the runbook uses to decide the migration is understood, and it reports a total that silently excludes objects that will not migrate. Those objects then survive `finalize` unmigrated and unmentioned.

**Fix:** count skipped-by-scheme rows and surface them on `MigrationPlan` (and echo them in the `copy`/`finalize` reports), so a non-zero skip count is visible rather than inferred. `specs/migration/object-storage-provider-migration.feature` has no scenario for this — worth binding one.

langwatch/src/tasks/objectStorageMigration.ts
```diff
@@ -0,0 +1,591 @@
...
+}): Promise<"copied" | "repaired" | "skippedVerified"> {
+  const sourceBytes = await readAll(await source.driver.get(sourceUri));
+  const sourceSha256 = sha256(sourceBytes);
+  if (expectedSha256 && sourceSha256 !== expectedSha256) {
+    throw new Error(
+      `Source object verification failed for ${redactStorageUri(sourceUri)}: expected ${expectedSha256}, got ${sourceSha256}`,
+    );
+  }
+  if (await destination.driver.exists(destinationUri)) {
+    const destinationBytes = await readAll(
```
drewdrewthis: **[reliability] `copyVerified` fully buffers every object, holds up to two copies at once, and the task has no heap headroom or size cap.**

`readAll` (lines 581-587) drains the whole stream into a `Buffer`. In `copyVerified` that means:

- `sourceBytes` (line 541) is live for the entire function;
- on the repair path this line reads `destinationBytes` — a **second** full copy resident simultaneously;
- `assertUriDigest` (lines 563-579) then re-reads the whole destination object a **third** time while `sourceBytes` is still in scope;
- `Buffer.concat` (line 586) transiently doubles again during the concat.

So peak resident memory is ~2x object size steady-state and worse in bursts, per object. The same pattern is in the dataset-chunk path at line 392.

There is no streaming seam to reach for: `StorageDriver.put` (`storage-driver.ts:18`) takes a `Buffer`, so buffering is forced by the abstraction, not a local choice. And there is no mitigation anywhere — I grepped: no size cap in this file (`INVENTORY_PAGE_SIZE` bounds row count, not bytes), and the new `task:migrate-object-storage` script (`package.json:52`) carries **no** `--max-old-space-size`. For contrast, this PR's own CI integration job sets `NODE_OPTIONS: --max-old-space-size=8192` because that path needs it. A single large stored object or dataset chunk OOMs the migration mid-run, and it does so inside the operator's cutover window.

**Fix (pick one, cheapest first):** set a heap flag on `task:migrate-object-storage`; and/or refuse above a configurable size cap so the failure surfaces as a named blocker in `plan` rather than an OOM during `copy`. `row.size_bytes` is already selected, so the cap check is nearly free. A streaming `put` is the real answer but is a larger change than this PR — a cap plus heap headroom closes the operational hole now.

**Related, same root cause — the paused window is sized by total corpus, not by the delta.** `finalize` runs `copyEligible` then `verifyEligible` (lines 338-341) with reads and writes paused. Because `copyVerified` always downloads *both* sides before it can conclude `skippedVerified`, an already-migrated object still costs two full GETs — and `verifyEligible` then downloads the destination a third time. So a `copy` that already ran to completion buys the operator nothing at finalize: the downtime is proportional to the whole corpus every time. A HEAD-based cheap-skip (size + stored checksum) before the full GET, and verifying only what the delta pass actually wrote, would make the paused window proportional to the delta.

langwatch/src/server/stored-objects/stored-objects-factory.ts
```diff
@@ -51,7 +52,20 @@ export function maybeAzureDriver(): AzureBlobDriver | undefined {
...
-    if (error instanceof AzureBackendMisconfiguredError) return undefined;
+    //
+    // That reasoning only covers WRITES, and only where Azure is not the
+    // selected backend. When STORED_OBJECTS_BACKEND=azure the operator has
+    // asked for Azure explicitly, and swallowing here registers nothing, so
+    // every READ surfaces as `Storage scheme "azure-blob" is not configured
+    // in this deployment` — a message that flatly contradicts their config and
+    // buries the webhook/label/annotation guidance the original error carries.
+    // Let it through there; keep the quiet path for the migration case.
+    if (
```
drewdrewthis, line 63: **[operability] The documented cutover contract never names `legacyAzureRead`, and following it literally makes stored objects unreadable during rollback.**

To be clear up front: this guard is correct and its doc comment explains why. The defect is in the contract that surrounds it.

After `finalize`, storage rows are rewritten to the destination scheme (`objectStorageMigration.ts:218`; `object-storage-provider-migration.feature:127-131`). The PR body's step 3 reads *"Change `STORED_OBJECTS_BACKEND` and credentials."* An operator rolling back Azure to S3 who follows that literally reverts the backend **and drops the Azure credentials**. At that point this condition is true (`STORED_OBJECTS_BACKEND !== "azure"`), `maybeAzureDriver` swallows `AzureBackendMisconfiguredError` and registers nothing — and every still-`azure-blob://` row throws `Storage scheme "azure-blob" is not configured in this deployment` (`storage-registry.ts:64-68`).

The flag that prevents this exists — `app.dataplane.legacyAzureRead` / `legacyS3ReadBucket`, wired in `charts/langwatch/templates/_helpers.tpl:248-292,1087-1091` and documented at `docs/self-hosting/configuration/environment-variables.mdx:308`, and this PR even adds chart validation rejecting them in the wrong direction. But `grep legacyAzureRead` returns **zero** hits in the PR body and zero in `specs/migration/object-storage-provider-migration.feature`. The feature file asserts *"the source bytes remain available for an explicit rollback"* without specifying the rollback that makes them reachable — and preserved bytes are not a rollback if nothing can address them.

**Fix (docs, no code change):** name `legacyAzureRead` / `legacyS3ReadBucket` in the cutover contract's step 3 and add an explicit rollback sequence — retain the credentials, set the legacy-read flag, then revert `STORED_OBJECTS_BACKEND`. A rollback scenario in the migration feature file would bind it.

drewdrewthis (commented):
Review pass on what this PR adds on top of `issue6087/azure-blob-entra-identity-auth` (base is #6181, not `main`). Four blocking threads below; non-blocking items are in the verdict comment. No approval state changed.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `ea9afc11` · **Run:** pr-review (reviewer: @review-clerk) · **Base:** `issue6087/azure-blob-entra-identity-auth` (#6181) — only what this PR adds on top of that base was reviewed.

Re-reviewed at the new HEAD after the branch moved past the SHA (`a3904d8f2`) the prior verdict was pinned to. All 6 prior `<!-- review-thread -->` inline threads are resolved, including the one that was blocking (CI had never run on the actual HEAD). Verified independently this pass:

- CI is green at this exact HEAD (`ea9afc1166b96ba83e84be63fa8bcebae0d30517`): 112 check runs via the REST checks API, zero non-success (SUCCESS/SKIPPED only) — not a rollup count, enumerated row-by-row.
- Read the core production paths myself this pass: `storage-registry.ts` (lazy Azure factory dispatch), `stored-objects-factory.ts` (the `STORED_OBJECTS_BACKEND=azure` misconfiguration-surfacing fix), `azure-credentials.ts` (shared transport-safety guard), `dataset-storage.ts` (documented migration limit for dataset chunks), `objectStorageMigration.ts` (plan/copy/finalize/verify engine — foreign-scheme accounting, BYOC exclusion, dataset-chunk destination-only tolerance, resumable pagination, verified-before-published finalize ordering), `groupQueueMigrationAudit.ts` (cutover audit, master-only cluster reads), and the `_helpers.tpl` / `values.yaml` chart changes (case-insensitive sovereign-endpoint check, duplicate-env-var avoidance between the active-provider write block and `legacyS3ReadBucket`).
- No new correctness, security, or structural defects found. The durable-queue-ref `tier: "s3"` labeling that looked suspicious for an Azure-sourced migration is confirmed intentional and tested (`jobEnvelope.codec.unit.test.ts`: "Provider migration does not change the durable queue reference format").

### Non-blocking (carried over from the prior pass — Decide / New Issue, not gating)
- **[correctness] Dataset-chunk enumeration trusts `dataset.chunkCount`** (`objectStorageMigration.ts`). (Decide)
- **[reliability] One unreadable source object aborts the whole run**, incl. inside the finalize downtime window. (New Issue)
- **[reliability] Stale `uploading` dataset row is a permanent finalize blocker.** (New Issue)
- **[reliability] No resumability checkpoint** beyond per-object retry. (New Issue)
- **[design] Queue audit has no retry/quorum; cluster fan-out unit-tested only.** (Decide)
- **[design] Pause flags are attestations, not enforced pauses.** (Decide)
- **[design, Pass 3] Client-side byte-shuffling where providers offer server-side copy** (no `CopyObjectCommand`/`beginCopyFromURL`). Legitimate tradeoff, not asked-and-answered in the PR body. (Decide)
- **[hygiene] `storage-registry.ts`** two-shapes-one-thing union + memo into a `readonly` field. (Decide)
- **[test-reviewer] `migrateObjectStorage.unit.test.ts`** call-argument assertion instead of a fake-cluster behavior test for master-only reads. (Decide)

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is **not** a merge-readiness signal; that is `pr-ready-check.sh` (8 criteria, of which this verdict is C3), which also gates on CI, PR-body format, and visual proof. This PR is also stacked on an open base PR (#6181) — merge ordering is out of this verdict's scope._

drewdrewthis (dismissed):
Requesting changes: 4 blocking items, all tracked as resolvable inline threads and indexed in the [review verdict](https://github.com/langwatch/langwatch/pull/6332#issuecomment-5164294320) (NOT-READY at `e9cc7a1c7`).

1. Re-run the real-Azure rehearsal at head — the current proof is at `54cfccd63` and three commits landed after it on the paths it exercised. -> https://github.com/langwatch/langwatch/pull/6332#discussion_r3702550387
2. `eligibleStoredObjects` silently drops rows and `plan` under-reports. -> https://github.com/langwatch/langwatch/pull/6332#discussion_r3702550394
3. `copyVerified` fully buffers with no size cap and no heap headroom on the task. -> https://github.com/langwatch/langwatch/pull/6332#discussion_r3702550398
4. The cutover contract never names `legacyAzureRead`, so a literal rollback makes stored objects unreadable. -> https://github.com/langwatch/langwatch/pull/6332#discussion_r3702550405

The engine itself reviewed well — see the "Checked, no defect found" section of the verdict for what was verified and for one candidate finding I falsified. Resolve the four threads and I'll re-review; the verdict comment refreshes to READY on a clean re-run.

sergioestebance:
## Real-Azure rehearsal re-run at head (`d96a680cd`)

Re-ran per the review: the previous proof was at `54cfccd63`; since then the dataset-chunk refactor, the master-only queue audit, the token-mode transport guards, the repo restructure rebase, and this round's fixes (foreign-scheme reporting, streaming digests) all landed on the exercised paths.

Same estate as the original rehearsal (5 stored objects + 1 chunked dataset, real `westeurope` account with `allowSharedKeyAccess=false`, dev S3 bucket), driven entirely through `azureCli` bearer tokens:

| Step | Result |
|---|---|
| Driver suite vs real Azure (`azure-blob-driver.realazure`) | 7/7 — incl. both Entra tests on the key-disabled account |
| `plan` (S3→Azure) | exit 0 — now reporting the new `foreignSchemeRows` field (0 here) |
| Finalize gate: writes not paused | refused, exact message |
| Finalize gate: reads not paused | refused, exact message |
| Finalize gate: outstanding queue work | **fired on real debris** — scan-discovery found stale `test-recon-*` queues another test session left on the shared dev Redis; after sweeping them, finalize passed. The gate's registry-independent discovery is doing exactly its job |
| `copy` → `finalize` → `verify` (S3→Azure) | all exit 0; 5/5 latest rows `azure-blob://` — copy exercised the streaming-digest paths against real providers (all objects verified by stream hash) |
| Reverse `plan`/`copy`/`finalize`/`verify` (Azure→S3) | all exit 0; 5/5 latest rows back on `s3://`; dataset chunk verified both directions |
| Source bytes | all 7 Azure blobs untouched after the reverse cutover — no phase deleted anything |

One operational note reproduced live: the S3 SSO session token expired mid-copy and the run failed **loud** (`ExpiredToken` from the SDK, exit 1) with no partial publication — resumed cleanly after refreshing credentials, exactly the resume-safety the contract promises.

langwatch/src/server/stored-objects/azure-credentials.ts
sergioestebance: Re-ran at head `d96a680cd` — full bidirectional pass against the real key-disabled account, gates included; the queue gate even fired on real debris (stale test queues on the shared dev Redis) before passing clean. Evidence table: https://github.com/langwatch/langwatch/pull/6332#issuecomment-5188440692

langwatch/src/tasks/objectStorageMigration.ts
sergioestebance: Fixed in `d96a680cd`. The generator now reports every row it declines through an `onForeignScheme` callback: `MigrationPlan` (and the copy report, which extends it) carries `foreignSchemeRows` + the distinct `foreignSchemes`, so the plan states what will not migrate instead of silently under-reporting. Copy/finalize still leave those rows untouched — they stay readable through scheme dispatch. Pinned by a new scenario ('Rows outside both providers are reported, never silently dropped') + integration test covering plan, copy and finalize.

langwatch/src/tasks/objectStorageMigration.ts
sergioestebance: Fixed in `d96a680cd`. All digest verification now hashes streams chunk-by-chunk (`sha256OfStream`) — the destination compare on the repair path, `assertUriDigest`, and the verify phase's source hash no longer buffer, so peak residency is the single source buffer `StorageDriver.put` forces. The task script also gets `NODE_OPTIONS=--max-old-space-size=4096`. Deliberately no size cap: refusing to copy a large object strands it at cutover — with one-copy residency plus headroom, letting the copy proceed is the safer failure mode, and stored objects are bounded upstream by the extraction path anyway.

langwatch/src/server/stored-objects/stored-objects-factory.ts
sergioestebance: Fixed in `d96a680cd`. The runbook (`environment-variables.mdx`) and the PR body's step 3 now name the flags explicitly: keep the source provider's read credentials configured through the rollback window — `legacyS3ReadBucket` after S3→Azure, `legacyAzureRead` after Azure→S3 — and spell out that a rollback with the flag set is fully non-destructive while a literal credential drop makes finalized addresses unreadable. Also documents that foreign-scheme rows (the sibling finding) stay readable only while their scheme's backend stays configured.

langwatch-agent:
Back to ready, and back with the reviewer. All 4 review threads resolved, changes-requested cleared, 59 checks green with 0 failing on the latest run of each, mergeable and clean.

Re-requested review from the same reviewer rather than a new one: they own this PR and the earlier review consumed the request, so without a fresh one nothing would tell them the ball had come back. Mode stays `review: targeted`.

---
Posted by the pr-hound agent.

drewdrewthis (commented):
Re-review at `d96a680cd`: all 4 blocking items from the prior `e9cc7a1c7` pass independently re-verified as resolved (foreign-scheme row reporting, streaming digest verification + raised memory budget, documented `legacyAzureRead`/`legacyS3ReadBucket` rollback contract, and a fresh real-Azure bidirectional rehearsal at head). CI green. Verdict: READY. See the verdict comment for the full itemized re-check and the unchanged non-blocking list.

langwatch-agent:
Ran the file-or-not call on the ten non-blocking items in the review verdict above. Each was checked against head `d96a680cd` rather than taken from the list, so a few dispositions differ from the reviewer's suggested ones. Nothing here is gating; the verdict stands as READY.

**Filed as issues (3).** These survive the merge and need a decision from someone, so they should not sit only in a PR comment:

| Issue | Item |
| --- | --- |
| langwatch/langwatch#6583 (P1) | Pause flags are attestations, not enforced pauses |
| langwatch/langwatch#6584 | One unreadable source object aborts the whole run, including inside the finalize downtime window |
| langwatch/langwatch#6585 | Stale `uploading` dataset row is a permanent finalize blocker |

langwatch/langwatch#6583 is the one worth reading first. `writesPaused`/`readsPaused` come from the migration process's own environment (`migrateObjectStorage.ts:123-124`) and are returned verbatim (`:296-297`); the app and workers never read them. So the gate on the only irreversible phase is a claim, not a measurement, and a mis-sequenced runbook publishes destination addresses for objects that were never copied. All three are on the board, Backlog, typed Task.

**Folded into langwatch/langwatch#6323, not filed separately.** Dataset-chunk enumeration trusting `dataset.chunkCount` (`objectStorageMigration.ts:258-277`). The fix is per-dataset storage provenance, which is what langwatch/langwatch#6323 already tracks, so a fourth issue would fragment the topic. Detail is [in a comment there](https://github.com/langwatch/langwatch/issues/6323).

**Small enough to belong in this PR, not an issue (3).** Leaving these to you rather than pushing to a branch that is already review-clean at this SHA:

- The `ready` wait on the master-only duplicate has no timeout (`migrateObjectStorage.ts:416-421`). If the handshake neither completes nor errors, finalize hangs with traffic paused. A bounded wait is a few lines and removes a hang from the downtime window.
- `stored-objects-factory.ts:23-38` contradicts its own implementation. The comment says the dead-config error is swallowed into `undefined` rather than crashing boot; `:63-69` rethrows when `STORED_OBJECTS_BACKEND === "azure"`.
- Client-side copy instead of server-side. This is a legitimate tradeoff (there is no cross-provider server-side copy between S3 and Azure), but the PR body does not say so. One line under Known limitations closes it.

**Not filing (3), with reasons, so they are declined on the record rather than dropped:**

- **No resumability checkpoint.** The reviewer marked this New Issue; disagreeing. `copy` is already re-runnable and skips verified objects via the already-at-destination branch (`:227-240`), so there is no correctness gap, only the cost of re-enumerating. For a tool that runs a handful of times per install, that is an optimization with no failure mode attached, and an issue for it would sit in the backlog forever. If the re-scan cost turns out to hurt in a real cutover, it belongs on langwatch/langwatch#6584, which is where the operator actually feels it.
- **Queue audit has no retry/quorum.** The master-only read already removes the replica-lag hazard that motivated the concern. Retry and quorum are worth having if the audit ever proves flaky in practice, but there is no evidence it is, and the timeout above is the part with a real failure mode.
- **`storage-registry.ts:39-51` union-plus-memo shape, and the unit test asserting call arguments rather than fake-cluster behavior.** Style and test-shape preferences on code that has an active author. Take them or leave them in this PR; neither is worth a tracked card.

Happy to be overruled on any of these, particularly the resumability one.

dev/docs/TESTING_PHILOSOPHY.md
```diff
@@ -174,6 +174,7 @@ The scenario title in the feature file should match the `it()` description in th
 | `@integration` | Component/boundary test | Testing rendering, API calls, DB queries |
 | `@regression` | Prevents a previously-fixed bug from recurring | Bug fix scenarios — must fail without the fix |
 | `@e2e` | Stable core flow (deprioritized) | Only for the 5-10 stable happy-path tests |
+| `@manual` | Verified by hand against real infrastructure, cannot run in CI | Needs a real cloud account/subscription. Still **requires a binding** to a test that self-skips without credentials, so the verification keeps an anchor in the codebase. Record the date and result in a comment above the scenario |
```
drewdrewthis: This is questionable... not sure we want this in the standards docs? Or if we do, maybe we'd want to think it through a bit more?

sergioestebance: sounds reasonable. ive been many time on this path to decide if manual, since i need to verify infra on creds.
how ever i also thought that agents can mark everything as manual.

sergioestebance: Settled in 0d4b63347: kept the tag (the parity checker enforces its binding, and the real-account scenario structurally cannot run in CI) but made it a last resort with explicit qualifying rules — structural CI-unreachability only (hard/slow/needs-Docker never qualify), automated tiers must cover everything automatable first, binding + last-run record required, and the burden of proof is on the tag, so reviewers are expected to challenge every new manual. That closes the abuse vector of marking things manual to dodge writing a real test.

sergioestebance: Update: went the other way in 77c01ffab — dropped the @manual tier entirely rather than fencing it. The real-account scenario is now a plain @integration scenario whose bound test self-skips without LANGWATCH_TEST_AZURE_* credentials (the posture other credential-gated integration tests already use), and TESTING_PHILOSOPHY.md / specs/README.md / check-feature-parity.ts are back to their base form. No new tag class in the standards.

.github/workflows/langwatch-app-ci.yml
```diff
@@ -111,6 +111,14 @@ jobs:
               - 'pnpm-workspace.yaml'
               - 'package.json'
               - '.npmrc'
+              # NOTE: 'charts/**' is deliberately NOT listed. Integration suites
```
drewdrewthis, line 114: **[reviewer-name: review-clerk] No CI has ever run on this PR's actual HEAD.** `gh api repos/langwatch/langwatch/actions/runs?head_sha=a3904d8f260e759a84bedf5993dd563a4c70d0c9` returns `total_count: 0`, and `gh api .../commits/a3904d8f2.../check-runs` shows only `Mintlify Deployment` — none of `langwatch-app-ci` / `langwatch-chart` (the unit/integration/behavioral suites this PR's own body cites as its safety evidence: 81 unit tests, 19 migration integration contracts, the helm behavioral job this diff adds). The last real CI run (`gh run list --branch issue6087/azure-review-followups`) was on commit `77c01ffab`, which the branch has since been rebased past (content differs — `git diff 77c01ffab a3904d8f2 --stat` shows an unrelated 10-file, 500+ line diff pulled in from upstream `main`, e.g. `platform/app/src/server/api/wire-body-limit.ts`). This is the second rebase this branch has taken since the prior review pass (the first rewrote `d96a680cd` into `f9566bfa6` with identical content, same story). A PR whose stated safety case rests on "81 unit tests / 19 integration contracts / feature parity passed" needs that evidence at the SHA being merged, not at a superseded ancestor. **Fix:** push a no-op commit or re-trigger the workflows so CI actually runs on `a3904d8f2` (or whatever the head becomes), and confirm green before merge.

drewdrewthis (commented):
Re-review at head `a3904d8f2` (marker was stale — prior verdict was pinned to `d96a680cd8`, an ancestor made unreachable by a rebase of the whole stack; the 4 prior review threads are resolved and their fixes verified still present at this head, see comment thread history).

drewdrewthis (dismissed):
NOT-READY per standing review-clerk order (Drew, 2026-08-03): a NOT-READY verdict submits REQUEST_CHANGES. 1 blocking inline thread (CI never ran on this PR's actual HEAD `a3904d8f2`) — see the verdict comment and the inline thread above for detail.

.github/workflows/langwatch-app-ci.yml
sergioestebance: Settled at head `da3125a0c0`: the branch was restacked onto the current base and pushed from a user token, so the full suite actually ran on the exact SHA under review — langwatch-app-ci (all unit + integration shards), langwatch-chart incl. the behavioral job, everything green (`MERGEABLE / CLEAN`, 0 failures, 0 pending). The one red along the way was `test-unit (3)`'s `check-queue` timing flake — zero overlap with this diff, 14/14 locally, green on rerun. Root cause of the missing runs confirmed: the two automation force-pushes came from a token that cannot start workflows; pushes from this side trigger normally.

sergioestebance:
## Ruthless review — self-authored, findings fixed in-branch

Traced the migration core (`objectStorageMigration.ts`), its operator wiring
(`migrateObjectStorage.ts`), the cutover queue audit, the lazy `StorageRegistry`
azure factory, the chart's provider-selection and legacy-read-flag guards, and
the two feature files. **No P0 or P1.**

The two things I most expected to be broken were not:

- **`finalize` is not an irreversible address rewrite without a gate.** It
  refuses without both pause acknowledgements, refuses on any blocking dataset
  or queue state, re-copies the delta and verifies *every* destination digest
  before publishing a single address, appends a newer ReplacingMergeTree
  version rather than mutating, and deletes no source bytes. The rollback is a
  reverse migration, and both directions are covered by the finalize scenario
  outline. The pause flags are self-attested — but `verify` catches a lie:
  anything written after enumeration fails its destination-digest check.
- **BYOC exclusion is keyed correctly.** `privateS3Configs` is keyed by
  organization id (`dataplane-s3.ts:147`) and the inventory checks
  `project.team.organizationId` — had that been keyed by project id it would
  have silently migrated every tenant-owned bucket's addresses to the global
  destination. Verified, not assumed.

### Fixed

**P2 — `platform/app/src/tasks/objectStorageMigration.ts:229` · `9f1b662610`**
The destination-scheme-mismatch blocker aborts the *entire* copy run and
identified nothing — no row, no addresses. The usual cause is a destination
endpoint configured differently from the one an earlier run used, and the
operator had no way to tell that from real corruption. Quoting the two
addresses does not help either: they differ only in the bucket/storage
account, which is exactly what `redactStorageUri` masks, so the message would
print two identical strings. Now names the stored-object id and project. The
existing test asserted on the bare prose; it now asserts the id is present.

**P2 — `platform/app/src/tasks/migrateObjectStorage.ts:410` · `03f6d543fa`**
`createCutoverAuditRedis` awaited the duplicated cluster's handshake with no
timeout and no cleanup on the error path. Two failure modes, both on the one
phase run with traffic paused and an operator watching: an errored handshake
left an ioredis Cluster retrying forever (`cleanup` is not bound until the
function returns, so nothing could close it), and a handshake that never
completes — every seed node unreachable emits no terminal event — hung
`finalize` indefinitely with no output. Now bounded at 30s, and the duplicate
is disconnected before the error is rethrown. Two tests added.

### Not fixed — non-blocking

**P3 — `platform/app/src/tasks/__tests__/objectStorageMigration.integration.test.ts`**
Flat structure: 20 `it()` descriptions are verbatim Gherkin scenario titles
rather than action-based descriptions nested under `describe("given …")` /
`describe("when …")`. CLAUDE.md names both as mistakes, and this PR's own
sibling files (`stored-objects-factory.credentials.unit.test.ts`) follow the
convention. I did not restructure it: moving 20 tests' setup into given/when
describes is a real refactor with regression risk on a green 628-line suite,
for zero behaviour change. Say the word and I'll do it.

**P3 — `jobEnvelope.codec.unit.test.ts:38`** Same shape, one test, plus the
`@scenario` annotation is unquoted where every neighbour quotes it. It binds
either way (parity checker is green).

### Verification, and what it is worth

`pnpm typecheck` + `pnpm typecheck:tests` clean. `migrateObjectStorage.unit`
24/24, `objectStorageMigration.integration` 22/22 (native services, no
containers), stored-objects unit suites 342/342, `check:feature-parity`
`✓ all bound`, biome clean on all four touched files.

Both findings were validated by executing the code, not by reading it — but
**by me, not by an independent context.** This harness does not spawn
refutation agents, so what is missing is not effort, it is independence: a
separate context can attack a claim without having formed it, and I cannot.
Weigh the two P2s accordingly.

sergioestebance:
Blessed. 🙏 A provider cutover that copies before it commits: every destination
byte is verified before a single address moves, and no phase deletes what it
came from.

> "Whoever can be trusted with very little can also be trusted with much." — Luke 16:10

Traced finalize's gate chain, the BYOC exclusion key, and the lazy Azure
registry: no path publishes an unverified address, and no inactive Azure
setting can block S3 traffic. Ship it — residual is the two P3 test-naming
items in the review above, waived.

**The approval is not mine to give: I wrote this branch, and an author who both
fixes and blesses removes the last checkpoint. It needs a human `APPROVE`.**

drewdrewthis (commented):
READY (pr-review, reviewer: @review-clerk) at ea9afc1166b96ba83e84be63fa8bcebae0d30517. All 6 prior review threads resolved, CI green (112/112 checks, verified row-by-row), and I re-read the core production paths this pass with no new blocking findings. Full verdict: https://github.com/langwatch/langwatch/pull/6332#issuecomment-5164294320

drewdrewthis (approved):
lgtm!

specs/features/scenarios/azure-blob-workload-identity.feature
```diff
@@ -304,16 +309,34 @@ Feature: Azure Blob stored-objects authenticate without a shared account key
...
+    Then it selects the Azure dataset storage
+    And no code path dereferences an absent account key
+
+  # The chunked round-trip itself stays unproven in a token mode: the emulator
+  # is shared-key only, so there is nowhere to run it outside a real
+  # Entra-authenticated account. Deliberately carries NO binding — the tag and
+  # an annotation claiming coverage cannot both be true.
   @integration @unimplemented
-  Scenario: Dataset uploads work in a token-based mode
+  Scenario: Dataset uploads round-trip in a token-based mode
```
drewdrewthis, line 328: We have a policy of no unimplemented features -- this should be a follow up

PR #6351, opened 2026-07-29
fix(billing): give invoice-list failures a named handled error

### LINKED ISSUE #6350: subscription.listInvoices 500s with "An unknown error occurred" when the billing customer can't be read
Found while browser-verifying #5969 against Stripe test mode.

## Symptom

On **Settings → Billing**, the invoices card shows "Failed to load invoices. Please try again later." and the console has:

```
POST /api/trpc/subscription.listInvoices → 500
TRPCClientError: An unknown error occurred
```

This is the same opaque failure the original upgrade bug was reported as — on a different endpoint. It sat in the same frame as a correctly-handled `subscription.create` failure returning `billing_currency_unavailable`, which makes the gap obvious side by side.

## Cause

`listInvoices` calls the payment provider without translating provider failures into a `HandledError`. A raw `StripeError` escapes the procedure, so it never reaches the handled-error channel that maps a stable `code` onto the wire — the client has nothing to key copy from and falls back to "An unknown error occurred".

Reproduced by pointing an organization's stored customer id at one that doesn't resolve; any provider-side failure (deleted customer, rate limit, outage) takes the same path.

## Expected

A named handled error with customer-safe copy, consistent with the rest of the billing surface after #5969 — e.g. invoices are temporarily unavailable, nothing is wrong with the account, try again. The provider error should ride along as a `reason` so the cause survives into the logs.

## Notes

- Not a regression from #5969 — pre-existing, and independent of the currency work.
- Worth a sweep of the other billing procedures that call the provider directly (`previewProration`, `manage`) for the same unwrapped-error shape rather than fixing only this one.

sergioestebance:
Fixes #6350. **Stacked on #5969** — base is `fix/growth-upgrade-currency-mismatch`, so review that one first and this diff shows only the invoice fix.

## Problem

Found while browser-verifying #5969 against Stripe test mode. When the stored billing-customer id can't be read, the billing page shows "Failed to load invoices. Please try again later." and the console has:

```
POST /api/trpc/subscription.listInvoices → 500
TRPCClientError: An unknown error occurred
```

It showed up in the same frame as a correctly-handled `subscription.create` failure returning `billing_currency_unavailable` — the contrast side by side is what made it obvious.

## Cause

`listInvoices` called `stripe.invoices.list` without translating provider failures, so a raw `StripeError` escaped the procedure. It never reached the handled-error channel that puts a stable `code` on the wire, so the client had nothing to key copy from and fell back to the generic message — the same opaque failure #5969 set out to remove, one endpoint over.

## Fix

Wrap the call in `InvoicesUnavailableError` (503, fault `provider`, code `billing_invoices_unavailable`) with the provider error attached as a `reason` so the cause survives into the logs.

Copy is deliberately narrow — "Couldn't load your invoices / Try again in a moment." Reading invoices is a display concern, nothing is being bought, so it must not imply anything is wrong with the account or that a charge failed.

## Tests

- provider failure → `billing_invoices_unavailable` rather than an unknown error
- the provider error survives as a `reason` on the handled error
- Full billing unit suite (508 with the error-presentation suites) and typecheck pass

## Not included

The issue notes the other billing procedures that call the provider directly (`previewProration`, `manage`) may have the same unwrapped shape. Left out of this PR to keep it to the observed bug; worth a sweep on its own.

## Summary by CodeRabbit

- **New Features**
  - Added clearer billing error messages for unsupported currencies, deleted billing profiles, and temporary provider outages.
  - Checkout now respects the billing customer’s established currency when available.
  - Added validation for supported billing currencies and inactive billing profiles.

- **Bug Fixes**
  - Stripe rate-limit and connection issues now produce retryable billing errors.
  - Prevented checkout creation and billing updates when currency or customer validation fails.


sergioestebance:
Superseded by #6353. The premise here was wrong: I treated the unwrapped provider error as a plain wrapping bug, but `dev/docs/best_practices/error-handling.md` says "unknown" is a correct outcome and that inventing a handled error to avoid it promises the caller an action they do not have. `InvoicesUnavailableError` did exactly that for every possible failure.

#6353 classifies instead — rate limit and connection failure become handled, everything else stays unknown — and covers the same mistake in the opposite direction on the currency lookup that merged with #5969.

PR #6352, opened 2026-07-29
feat(model-providers): add Google Agent Platform as a provider

### LINKED ISSUE #6328: Google Agent Platform keys (AQ.…) cannot be validated or used as a model provider
## The gap

A customer's Google key is refused when they add Gemini as a model provider. PR #6252 fixed the *diagnosis* — they now read why it was refused and can save anyway — but the key still does not validate, and it still cannot be used to call a model.

That is the remaining work, and it is a feature rather than a bug fix.

## Why it is refused

Google mints several kinds of credential, and they are not interchangeable:

| Key shape | Minted by | Answers on |
|---|---|---|
| `AIza…` | Google AI Studio, or a Cloud key allowed to call the Generative Language API | `generativelanguage.googleapis.com` |
| `AQ.…` | Google Cloud console — **Agent Platform / Gemini Enterprise** | Agent Platform surfaces, *not* `generativelanguage.googleapis.com` |

Our `gemini` provider points exclusively at `generativelanguage.googleapis.com`. PR #6252 made validation walk every auth shape that host serves (`?key=`, `x-goog-api-key`, `Bearer` on the OpenAI-compatible surface), so a key that host recognises is now accepted whichever way it was minted — but an `AQ.` key is not scoped to that host at all, so every shape correctly refuses it.

Worth stating plainly, because it is the thing that misleads: **there is no client-side format regex.** Nothing in the app pattern-matches the key. The refusal comes from a live probe against Google. So "make the regex accept `AQ.`" is not the fix — there is no regex to change.

## Two ways to close it

**Option A — probe the Agent Platform surface too.** Extend the `gemini` provider's shape walk to include the Agent Platform host. Cheapest, but it papers over a real distinction: validation would pass while inference still fails, because the model-calling path also targets `generativelanguage.googleapis.com`. Only worth doing if the inference path is extended in the same change.

**Option B — add a distinct "Google Agent Platform" provider.** Its own registry entry, own base URL, own auth shape, own model list. Honest about the fact that it is a different service with a different credential, and it is the shape the drawer's copy already implies. More work, and it needs a decision on which models to expose and how they map to our model registry.

Recommendation: **B**, unless the Agent Platform API turns out to be wire-compatible with the Generative Language API, in which case A collapses into a base-URL change.

## Prior art in the codebase

- `langwatch/src/features/onboarding/regions/model-providers/registry.tsx` — provider entries, `defaultBaseUrl`, `apiRoot`, and the per-field guidance copy.
- `langwatch/src/server/api/routers/providerValidation.ts` — the auth-shape walk and Google's `reason`-code handling (`SERVICE_DISABLED`, `API_KEY_SERVICE_BLOCKED`, …).
- `specs/model-providers/credential-validation.feature` — the scenarios #6252 added.

Note that Vertex AI is already a separate provider and is *not* the answer here: `aiplatform.googleapis.com` answers `401 "API keys are not supported by this API"`, so it needs a service account rather than any of these keys.

## Scope

Out of scope for #6252, which explicitly records this under "What this does not do". Filed as the follow-up.

---
*Reported alongside the customer case behind #6252.*

Aryansharma28:
# Related Issue

- Resolves #6328

## The report

Altura could not add Gemini as a model provider:

> adding gemini as a model provider doesn't seem to work - api keys generated via gemini pasted into the api key field seem to trigger some kind of validation on the frontend and doesn't allow me to continue

#6252 fixed the *diagnosis* — they now read why the key was refused, and can save regardless. This adds the provider the key actually belongs to.

## What the key turned out to be

Their credential is a **Gemini Enterprise Agent Platform** key, not an AI Studio one. Both are `AQ.`-prefixed and look identical. `generativelanguage.googleapis.com` refuses it — correctly, its API restrictions exclude that service — so there was no provider in LangWatch it could be added to.

Worth stating, because it keeps being proposed: **there is no client-side key-format regex.** The only validation is `GEMINI_API_KEY: z.string().min(1)` — "not empty". The "validation on the frontend" was the probe's 403 rendering as a field error.

## The part that had to be discovered by probing

The two endpoints on the same host disagree, and the documentation does not mention it:

| Endpoint | API key |
|---|---|
| `POST .../publishers/google/models/{model}:generateContent` | **accepted** |
| `GET .../models` | 401 `"API keys are not supported by this API"` |

So this provider **cannot be validated the way every other one here is.** It is probed by generating a single token. Validating by listing models — which is what the shared code path does — would report a perfectly good credential as unusable. `ProbeRequest` gained `method`/`body` for that; every existing provider keeps its `GET`.

An earlier round of this investigation concluded the opposite and recommended closing #6328 as not-buildable. That conclusion came from probing the listing endpoint and generalising, and it was wrong.

## Also decided by testing, not assumption

- **`x-goog-api-key`, not `?key=`.** Both work. A credential in a URL reaches access logs, proxy logs and browser history, which is what #6252 moved this whole flow away from. A test pins that the key is absent from the URL.
- **`Bearer` is not an option** — 401.
- **Project and location are asked for, not guessed.** The path needs both. Omitting either yields no probe candidates rather than a fabricated path, which `runProbeChain` reports as unreachable — nothing was asked, so there is no verdict on the key.
- **Model names are not assumable.** `gemini-2.5-flash` resolves; `gemini-2.0-flash` 404s on the reporter's project. A 404 stays `provider_refused`, never `provider_key_invalid` — a different key would not fix it.

## Verification

Two rounds, because the code changed materially between them (two P1s
fixed after review: `validateKeyWithCustomUrl` dropping project/location,
and a dangling `defaultModel` with no catalog entry — see the review
thread for both). This is the final round, against the fully-fixed code.

**Direct, against real Google** — the same real Agent Platform key, run
through the actual merged `validateProviderApiKey`, no mocks:

```
google_agent_platform  ->  {"valid": true}
gemini (same key)       ->  code=provider_key_restricted
                            "This key's restrictions block the request.
                             Its API restrictions exclude the Generative
                             Language API. Allow that API in the Google
                             Cloud console, or set up a Vertex AI
                             provider instead."
```

**End-to-end in a real browser**, same key, current code, local infra:

| | |
|---|---|
| Provider list, drawer open |  |
| Fields, correct guidance copy |  |
| Key entered |  |
| **Saved — validates, listed in the table** |  |

The customer's key on the **existing** Gemini provider, for comparison —
same key, correctly refused, with the reason instead of "invalid":

> This key's restrictions block the request. Its API restrictions exclude
> the Generative Language API. Allow that API in the Google Cloud
> console, or set up a Vertex AI provider instead.

```
  7 scenarios    specs/model-providers/google-agent-platform.feature — all bound
 59 unit tests   agentPlatformValidation + providerValidation
1039 unit tests  src/server/api + src/server/gateway (no regressions)
```

`pnpm typecheck:all` clean. Lint clean — verified against the actual CI
gate (a merge-base diff of Biome diagnostics per file/rule, format-
normalized), not just a local `pnpm lint` run: 0 added violations.

Two exhaustiveness gaps the compiler caught on the way in — `iconsMap`
and `ModelProviderKey` — are filled.

## What this does not do

**Inference.** Routing a provider type to an upstream is `mapProvider`'s job in the Go gateway (`services/aigateway`), so the credential materialises but no traffic dispatches yet. This closes *"I cannot add my key"*, not *"I can run models through it"*. Worth a follow-up before anyone points a workflow at it.

No credential appears in this PR, the repo, or any commit.

Closes #6328

## Summary by CodeRabbit

* **New Features**
  * Added **Google Agent Platform** as a supported model provider.
  * Extended onboarding/config to collect the required **API key, project, and location**.
  * Updated provider branding to include **Google Cloud** for this provider.
* **Bug Fixes**
  * Improved provider credential validation to probe the correct **content-generation** endpoint with the API key in headers (never in the URL).
  * Correctly classifies responses as **valid**, **invalid key**, **provider refused**, **publisher model not found**, or **service unreachable** (including missing project/location handling).

### DIFF
--- a/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
+++ b/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
@@ -87,10 +87,10 @@ export const modelProviderRegistry: ModelProviderRegistry = [
     // customer never chose, unconditionally, instead of deferring to their
     // custom models the way `vertex_ai` correctly does by omitting this.
     //
-    // The request path names the project and location, and `apiRoot` is
-    // the host that path is built from — see `AGENT_PLATFORM_HOST` in
+    // `defaultBaseUrl` is this host, static; the full request path also
+    // names the project and location, assembled per credential in
     // `providerValidation.ts`'s `google_agent_platform` probe branch, which
-    // derives from this same field rather than a second hardcoded string.
+    // reads `apiRoot` below for the host rather than repeating this string.
     defaultBaseUrl: "https://aiplatform.googleapis.com",
     apiRoot: "https://aiplatform.googleapis.com",
     icon: singleIcon(
--- a/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
+++ b/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
@@ -78,10 +78,19 @@ export const modelProviderRegistry: ModelProviderRegistry = [
     key: "google_agent_platform",
     backendModelProviderKey: "google_agent_platform",
     label: "Google Agent Platform",
-    defaultModel: "gemini-2.5-flash",
-    // The path names the project and location, so there is no static base
-    // URL to probe — it is assembled per credential. See
-    // buildProbeCandidates' `google_agent_platform` branch.
+    // No `defaultModel`: this provider has no catalog entries (like
+    // `vertex_ai`), and `resolveProviderDefaultModel` in
+    // `components/gateway/eligibleModelProviders.ts` reads a registry
+    // default BEFORE falling back to whatever the customer actually
+    // configured. Naming a model here that isn't in the catalog would make
+    // the gateway's eligible-providers surface hand out a model string the
+    // customer never chose, unconditionally, instead of deferring to their
+    // custom models the way `vertex_ai` correctly does by omitting this.
+    //
+    // The request path names the project and location, and `apiRoot` is
+    // the host that path is built from — see `AGENT_PLATFORM_HOST` in
+    // `providerValidation.ts`'s `google_agent_platform` probe branch, which
+    // derives from this same field rather than a second hardcoded string.
     defaultBaseUrl: "https://aiplatform.googleapis.com",
     apiRoot: "https://aiplatform.googleapis.com",
     icon: singleIcon(
--- a/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
+++ b/langwatch/src/features/onboarding/regions/model-providers/registry.tsx
@@ -75,6 +75,43 @@ export const modelProviderRegistry: ModelProviderRegistry = [
     },
   },
   {
+    key: "google_agent_platform",
+    backendModelProviderKey: "google_agent_platform",
+    label: "Google Agent Platform",
+    defaultModel: "gemini-2.5-flash",
+    // The path names the project and location, so there is no static base
+    // URL to probe — it is assembled per credential. See
+    // buildProbeCandidates' `google_agent_platform` branch.
+    defaultBaseUrl: "https://aiplatform.googleapis.com",
+    apiRoot: "https://aiplatform.googleapis.com",
+    icon: singleIcon(
+      "/images/external-icons/gcloud.svg",
+      "Google Agent Platform",
+    ),
+    externalDocsUrl:
+      "https://docs.cloud.google.com/gemini-enterprise-agent-platform",
+    fieldMetadata: {
+      GOOGLE_AGENT_PLATFORM_API_KEY: {
+        label: "Agent Platform API Key",
+        // Says where the key comes from, because this is the distinction that
+        // sent a customer round in circles: an Agent Platform key and an AI
+        // Studio key look alike and are not interchangeable.
+        description:
+          "A Google Cloud API key for Gemini Enterprise Agent Platform, created under APIs & Services > Credentials. An AI Studio key belongs on the Google Gemini provider instead.",
+      },
+      GOOGLE_AGENT_PLATFORM_PROJECT: {
+        label: "Google Cloud Project ID",
+        description:
+          "The project the key belongs to. Its number appears in the error Google returns if the key is used against the wrong service.",
+      },
+      GOOGLE_AGENT_PLATFORM_LOCATION: {
+        label: "Location",
+        description:
+          "Where to serve the model from — 'global', or a region such as us-central1.",
+      },
+    },
+  },
+  {
     key: "open_ai_azure",
     backendModelProviderKey: "azure",
     label: "Azure OpenAI",
--- a/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
@@ -55,7 +55,13 @@ describe("validateProviderApiKey for google_agent_platform", () => {
       expect(url).toContain(":generateContent");
       expect(url).not.toMatch(/\/models(\?|$)/);
       expect(init.method).toBe("POST");
-      expect(init.body).toBeTruthy();
+      // Pinned, not just truthy: a body of `"x"` would pass a truthiness
+      // check while sending garbage, and `maxOutputTokens: 1` is what keeps
+      // this probe cheap rather than generating a full response on every
+      // credential check.
+      const body = JSON.parse(String(init.body));
+      expect(body.contents[0].parts[0].text).toBeTruthy();
+      expect(body.generationConfig.maxOutputTokens).toBe(1);
     });
 
     /** @scenario The credential travels in a header, never in the URL */
@@ -88,6 +94,30 @@ describe("validateProviderApiKey for google_agent_platform", () => {
       expect(sentRequest().url).toContain(
         "/projects/acme-123/locations/us-central1/",
       );
+    });
+  });
+
+  describe("given a credential missing its project or location", () => {
+    /**
+     * This is the shape `validateKeyWithCustomUrl` produced before it was
+     * fixed to preserve stored credentials rather than rebuild them from
+     * just the key and endpoint: project and location silently dropped,
+     * leaving only the key. Asserting on it here pins the walk's behavior
+     * directly, without needing a Prisma-backed test for that caller.
+     */
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("is unreachable rather than reporting a verdict on the key", async () => {
+      mockFetch.mockResolvedValue(generated());
+
+      await expect(
+        validateProviderApiKey("google_agent_platform", {
+          GOOGLE_AGENT_PLATFORM_API_KEY:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        }),
+      ).rejects.toMatchObject({ code: "provider_unreachable" });
+
+      // Nothing to ask without a project and location, so nothing was sent.
+      expect(mockFetch).not.toHaveBeenCalled();
     });
   });
 
--- a/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
@@ -45,7 +45,7 @@ describe("validateProviderApiKey for google_agent_platform", () => {
   });
 
   describe("given a credential to check", () => {
-    /** @scenario An Agent Platform key is checked against the endpoint that accepts it */
+    /** @scenario A key the platform accepts is valid */
     it("asks the provider to generate content rather than to list models", async () => {
       mockFetch.mockResolvedValue(generated());
 
@@ -64,7 +64,7 @@ describe("validateProviderApiKey for google_agent_platform", () => {
       expect(body.generationConfig.maxOutputTokens).toBe(1);
     });
 
-    /** @scenario The credential travels in a header, never in the URL */
+    /** @scenario The credential is not exposed where logs or browser history could retain it */
     it("sends the key as a header and keeps it out of the URL", async () => {
       mockFetch.mockResolvedValue(generated());
 
@@ -81,7 +81,7 @@ describe("validateProviderApiKey for google_agent_platform", () => {
       expect(url).not.toContain("key=");
     });
 
-    /** @scenario The project and location the customer gave are the ones probed */
+    /** @scenario The project and location I entered are the ones actually checked */
     it("builds the path from the project and location given", async () => {
       mockFetch.mockResolvedValue(generated());
 
--- a/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
@@ -108,44 +108,51 @@ describe("validateProviderApiKey for google_agent_platform", () => {
     // Three separate cases, not one covering both fields at once: an `||`
     // guard degrading to `&&` would still return no candidates when BOTH
     // are missing, and only a case with exactly one present catches that.
-    const expectUnreachable = async (credentials: Record<string, string>) => {
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("rejects the check without probing when both project and location are missing", async () => {
       mockFetch.mockResolvedValue(generated());
 
       await expect(
-        validateProviderApiKey("google_agent_platform", credentials),
+        validateProviderApiKey("google_agent_platform", {
+          GOOGLE_AGENT_PLATFORM_API_KEY:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        }),
       ).rejects.toMatchObject({ code: "provider_unreachable" });
 
-      // Nothing to ask without both a project and a location, so nothing
-      // was sent — true whichever one is the one missing.
+      // Nothing to ask without both a project and a location, so nothing was sent.
       expect(mockFetch).not.toHaveBeenCalled();
-    };
-
-    /** @scenario A credential missing its project or location is not probed at all */
-    it("rejects the check without probing when both project and location are missing", async () => {
-      await expectUnreachable({
-        GOOGLE_AGENT_PLATFORM_API_KEY:
-          CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
-      });
     });
 
     /** @scenario A credential missing its project or location is not probed at all */
     it("rejects the check without probing when only the location is missing", async () => {
-      await expectUnreachable({
-        GOOGLE_AGENT_PLATFORM_API_KEY:
-          CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
-        GOOGLE_AGENT_PLATFORM_PROJECT:
-          CREDENTIALS.GOOGLE_AGENT_PLATFORM_PROJECT,
-      });
+      mockFetch.mockResolvedValue(generated());
+
+      await expect(
+        validateProviderApiKey("google_agent_platform", {
+          GOOGLE_AGENT_PLATFORM_API_KEY:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+          GOOGLE_AGENT_PLATFORM_PROJECT:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_PROJECT,
+        }),
+      ).rejects.toMatchObject({ code: "provider_unreachable" });
+
+      expect(mockFetch).not.toHaveBeenCalled();
     });
 
     /** @scenario A credential missing its project or location is not probed at all */
     it("rejects the check without probing when only the project is missing", async () => {
-      await expectUnreachable({
-        GOOGLE_AGENT_PLATFORM_API_KEY:
-          CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
-        GOOGLE_AGENT_PLATFORM_LOCATION:
-          CREDENTIALS.GOOGLE_AGENT_PLATFORM_LOCATION,
-      });
+      mockFetch.mockResolvedValue(generated());
+
+      await expect(
+        validateProviderApiKey("google_agent_platform", {
+          GOOGLE_AGENT_PLATFORM_API_KEY:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+          GOOGLE_AGENT_PLATFORM_LOCATION:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_LOCATION,
+        }),
+      ).rejects.toMatchObject({ code: "provider_unreachable" });
+
+      expect(mockFetch).not.toHaveBeenCalled();
     });
   });
 
--- a/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
@@ -105,19 +105,46 @@ describe("validateProviderApiKey for google_agent_platform", () => {
      * leaving only the key. Asserting on it here pins the walk's behavior
      * directly, without needing a Prisma-backed test for that caller.
      */
-    /** @scenario A credential missing its project or location is not probed at all */
-    it("is unreachable rather than reporting a verdict on the key", async () => {
+    // Three separate cases, not one covering both fields at once: an `||`
+    // guard degrading to `&&` would still return no candidates when BOTH
+    // are missing, and only a case with exactly one present catches that.
+    const expectUnreachable = async (
+      credentials: Record<string, string>,
+    ) => {
       mockFetch.mockResolvedValue(generated());
 
       await expect(
-        validateProviderApiKey("google_agent_platform", {
-          GOOGLE_AGENT_PLATFORM_API_KEY:
-            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
-        }),
+        validateProviderApiKey("google_agent_platform", credentials),
       ).rejects.toMatchObject({ code: "provider_unreachable" });
 
-      // Nothing to ask without a project and location, so nothing was sent.
+      // Nothing to ask without both a project and a location, so nothing
+      // was sent — true whichever one is the one missing.
       expect(mockFetch).not.toHaveBeenCalled();
+    };
+
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("is unreachable when both project and location are missing", async () => {
+      await expectUnreachable({
+        GOOGLE_AGENT_PLATFORM_API_KEY: CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+      });
+    });
+
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("is unreachable when only the location is missing", async () => {
+      await expectUnreachable({
+        GOOGLE_AGENT_PLATFORM_API_KEY: CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        GOOGLE_AGENT_PLATFORM_PROJECT:
+          CREDENTIALS.GOOGLE_AGENT_PLATFORM_PROJECT,
+      });
+    });
+
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("is unreachable when only the project is missing", async () => {
+      await expectUnreachable({
+        GOOGLE_AGENT_PLATFORM_API_KEY: CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        GOOGLE_AGENT_PLATFORM_LOCATION:
+          CREDENTIALS.GOOGLE_AGENT_PLATFORM_LOCATION,
+      });
     });
   });
 
--- a/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
+++ b/langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
@@ -108,9 +108,7 @@ describe("validateProviderApiKey for google_agent_platform", () => {
     // Three separate cases, not one covering both fields at once: an `||`
     // guard degrading to `&&` would still return no candidates when BOTH
     // are missing, and only a case with exactly one present catches that.
-    const expectUnreachable = async (
-      credentials: Record<string, string>,
-    ) => {
+    const expectUnreachable = async (credentials: Record<string, string>) => {
       mockFetch.mockResolvedValue(generated());
 
       await expect(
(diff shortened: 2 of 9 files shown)


langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -888,6 +960,10 @@ export async function validateProviderApiKey(
       baseUrl,
       defaultBaseUrl,
       apiRoot: providerApiRoots[provider],
+      agentPlatform: {
+        project: customKeys.GOOGLE_AGENT_PLATFORM_PROJECT?.trim() ?? "",
```
Aryansharma28, line 979: **P1 — the second validation entry point never supplies these, so re-validating a saved credential always reports "couldn't reach the provider".**

This reads `customKeys`, and `validateProviderApiKey` has two callers. The other one is `validateKeyWithCustomUrl` (same file, L870-879), which does not forward the whole credential — it constructs a fresh `customKeys` from scratch:

```ts
const customKeys: Record<string, string> = {
  [apiKeyField]: apiKey,
};
if (endpointField && customBaseUrl) {
  customKeys[endpointField] = customBaseUrl;
}
```

`GOOGLE_AGENT_PLATFORM_PROJECT` and `GOOGLE_AGENT_PLATFORM_LOCATION` are neither `apiKeyField` nor `endpointField` (`endpointKey: undefined`), so they are dropped. Both `?.trim() ?? ""` land on `""`, `buildProbeCandidates` returns `[]`, and `runProbeChain` throws `ProviderUnreachableError`.

That is reachable from the UI, not theoretical. `ModelProviderSetup.tsx:365-376`:

```ts
const credentialsChanged = userEnteredNewApiKey || hasNonApiKeyChanges;
if (credentialsChanged && probeRequired) {
  const isValid = userEnteredNewApiKey
    ? await validateApiKey()
    : await validateWithCustomUrl(customBaseUrl);
  if (!isValid) { recordRefusal(); return; }
```

So: add the provider with key + project + location (works — `userEnteredNewApiKey` is true, full `customKeys`), then come back and change the location from `global` to `us-central1` without retyping the key. `hasNonApiKeyChanges` is true, `userEnteredNewApiKey` is false, and the customer is told:

> **Couldn't reach the provider.** Nothing answered, so this API key was not checked. Check your network connection, then try again.

(`presentation.ts:462-471`, non-configurable-endpoint variant, because `endpointKey: undefined`.) `recordRefusal()` fires and the save is blocked. A perfectly good key, a network-connection diagnosis, and a blocked save — which is the exact misdiagnosis class #6252 and this PR exist to remove, reintroduced through the door the PR did not walk through.

Two ways out. Either `validateKeyWithCustomUrl` carries the non-apiKey credential fields through from `storedKeys` (general fix — any future provider with extra required fields hits this too), or the empty-candidates case stops resolving to `provider_unreachable` (see the sibling comment on L586). I would do the first; it is the actual bug, and the second only changes which wrong message appears.

langwatch/src/features/onboarding/regions/model-providers/registry.tsx
```diff
@@ -74,6 +74,43 @@ export const modelProviderRegistry: ModelProviderRegistry = [
       },
     },
   },
+  {
+    key: "google_agent_platform",
+    backendModelProviderKey: "google_agent_platform",
+    label: "Google Agent Platform",
+    defaultModel: "gemini-2.5-flash",
```
Aryansharma28: **P1 — no model in the catalog belongs to this provider, so the model dropdown is empty.**

`getProviderModelOptions` (`server/modelProviders/registry.ts:201-211`) filters the catalog by exact provider match:

```ts
.filter(([_, model]) => model.provider === provider && model.mode === mode)
```

Grepping the repo, `google_agent_platform` appears in exactly 7 files — all of them in this diff — and `llmModels.json` / `llmModels.overlay.json` are not among them. So `getProviderModelOptions("google_agent_platform", "chat")` returns `[]`, and:

- `ModelProviderSetup.tsx:203` renders an empty model select on the step the customer cannot skip;
- `modelProvider.service.ts:414-416` reports `models: []` / `embeddingsModels: []` for the saved row;
- this `defaultModel` has no catalog entry to resolve against (catalog ids are `provider/model`, so it would need `google_agent_platform/gemini-2.5-flash`).

The PR body scopes out *inference* — routing, which is genuinely Go-side — but this is TS-side catalog data, `llmModels.overlay.json` exists for precisely this, and "I can add my key" is not really closed if the very next field is empty. Either add the overlay entries, or say in the PR body that model listing is deferred too, so the next person is not surprised.

Separately: `gemini-2.5-flash` is not a `gemini/`-prefixed catalog entry either (the catalog is on `gemini-3.6-flash` / `gemini-3.5-flash-lite`), which makes me want a second look at whether 2.5 is the right default rather than the one that happened to be probed.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -517,20 +561,47 @@ function buildProbeCandidates({
...
+
+      const { project, location } = agentPlatform;
+
+      return [
+        {
+          // The key rides in a header, not `?key=`, which Agent Platform also
+          // accepts: a credential in a URL reaches access logs, proxy logs and
+          // browser history, and both shapes were verified to work.
+          url:
+            `${AGENT_PLATFORM_HOST}/v1/projects/${encodeURIComponent(project)}` +
```
Aryansharma28: **P2 — regional locations are probed against the global host, and the test that looks like it covers this is mocked.**

`AGENT_PLATFORM_HOST` is `https://aiplatform.googleapis.com` unconditionally. That is Google's *global* endpoint. For a regional location, Vertex AI's REST surface is served from `https://LOCATION-aiplatform.googleapis.com` — `us-central1-aiplatform.googleapis.com` for the field description's own example.

What the evidence in this PR actually covers:

- the live end-to-end verification used `location: global` (the fixture is `GOOGLE_AGENT_PLATFORM_LOCATION: "global"`);
- the `us-central1` test asserts `sentRequest().url` contains `/projects/acme-123/locations/us-central1/` against `mockFetch`, so it pins our own string concatenation and cannot observe what Google answers;
- `registry.tsx:105` tells the customer to enter "a region such as us-central1".

So the one configuration the UI invites is the one nobody asked Google about. If the global host does not serve a regional path, the customer gets a 404 → `provider_refused` ("the model was not found"), which is not what is wrong and not something they can act on.

I cannot resolve this without a key, so I am not asserting it fails — I am asserting it is unverified while the docs copy actively steers people into it. Cheapest fix that removes the doubt: derive the host, `location === "global" ? AGENT_PLATFORM_HOST : \`https://${location}-aiplatform.googleapis.com\``, and probe one real region live. If the global host does serve regional paths, a one-line comment saying so (with the same "established by probing" weight as the rest of this file) closes it.

langwatch/src/features/onboarding/regions/model-providers/registry.tsx
```diff
@@ -74,6 +74,43 @@ export const modelProviderRegistry: ModelProviderRegistry = [
...
+    // The path names the project and location, so there is no static base
+    // URL to probe — it is assembled per credential. See
+    // buildProbeCandidates' `google_agent_platform` branch.
+    defaultBaseUrl: "https://aiplatform.googleapis.com",
```
Aryansharma28, line 94: **P3 — this field is load-bearing in a way the comment directly above it denies, and nothing pins it.**

The comment says "there is no static base URL to probe" and then sets one. Both readings are defensible, but the value is not decorative: `validateProviderApiKey` bails early at L952-954,

```ts
if (!baseUrl && !defaultBaseUrl) {
  return { valid: true };
}
```

and `endpointKey: undefined` means `baseUrl` is always `""` for this provider. So `defaultBaseUrl` is the only thing keeping validation from short-circuiting to `{ valid: true }`. A future reader who takes the comment at face value and deletes the field — an entirely reasonable cleanup, the value is otherwise unused, since `buildProbeCandidates` ignores `url`/`defaultBaseUrl` on this branch — silently turns credential validation into "every key is fine", with no test failing.

Either make the guard explicit about strategies that build their own URL (`if (!baseUrl && !defaultBaseUrl && authStrategy !== "google_agent_platform")`), or keep the field and change the comment to say *why* it must stay. A test asserting a bad Agent Platform key is not reported valid would also catch the deletion.

langwatch/src/server/gateway/config.materialiser.ts
```diff
@@ -567,6 +567,18 @@ function buildCredentials(mp: ModelProvider): Record<string, unknown> {
...
+    // and location, so both travel with the key rather than being derived.
+    // Routing this type to an upstream is `mapProvider`'s job in the Go
+    // gateway (services/aigateway); until that lands the credential
+    // materialises but no traffic is dispatched. Validation does not go
+    // through here — see providerValidation.ts.
+    case "google_agent_platform":
+      return {
+        api_key: pick("GOOGLE_AGENT_PLATFORM_API_KEY"),
+        project_id: pick("GOOGLE_AGENT_PLATFORM_PROJECT"),
+        location: pick("GOOGLE_AGENT_PLATFORM_LOCATION"),
```
Aryansharma28: **P3 — `location` where the closest sibling uses `region`, which quietly drops the slot-level `region`.**

`vertex_ai`/`vertex` twelve lines up returns `project_id` + **`region`**, and that name is not arbitrary — `buildProviderSlot` lifts it back out:

```ts
const region = pickString(credentials, "region");   // L628
...
...(region ? { region } : {}),                       // L638
```

So Agent Platform is the one Google provider whose slot carries no top-level `region`, and whoever writes the `mapProvider` case has to know that this provider spells it differently from the provider it is otherwise shaped exactly like. Renaming to `region` (or emitting both) makes the slot come out right for free and removes a trap from the follow-up. If `location` is deliberate because `global` is not a region, that is a fair argument — but then it wants a sentence here, because the header comment on `buildCredentials` points the reader at the Go consuming side and this is the field that will not be where they look for it.

Aryansharma28 (commented):
## Adversarial review

Scoped to the 8-file merge-base diff (`a83f9f6c8a` → `872b7cebc5`). Note for anyone diffing locally: `git diff origin/main HEAD` also shows `annotation.ts` + `tracesMapping.ts` as reverted — that is purely a stale base (#6133 landed after the branch point, and this branch never touches those files, so a merge will not revert them). **Not a finding.** Rebasing would make the local diff match GitHub's.

**Counts: P0 0 · P1 2 · P2 1 · P3 4 · P4 2**

### What I checked and found clean

**1. The `ProbeRequest` `method`/`body` regression (the headline risk) — clean.** `method: candidate.method ?? "GET"` plus `...(candidate.body === undefined ? {} : { body: candidate.body })` at `providerValidation.ts:741-743`. No branch other than `google_agent_platform` sets either field, so `anthropic`, `elevenlabs`, all four `gemini` shapes, `bearer` and `custom` emit a byte-identical request to before — `method: "GET"`, no `body` key at all (which matters: passing `body: undefined` explicitly is fine for undici, but the guard means we never even find out). The spread-guard is the right shape rather than the paranoid one. No regression.

**2. Security — clean.** The key reaches only the `x-goog-api-key` header; the URL is built from constants + `project`/`location` and never the key, and `expect(url).not.toContain(CREDENTIALS...)` pins it. The provider's own sentence is not echoed into `domainError` and the test asserts the leaked-looking substring is absent from the serialized result. `encodeURIComponent(project)` / `encodeURIComponent(location)` is doing real work here, not decoration — it stops a `location` of `../../` or `global/publishers/...` from re-pointing the path at another project, which is worth calling out as done right.

**3. Exhaustiveness — one real gap, and it is not type-checked.** Every other `keyof typeof modelProviders` site is an optional index lookup (`modelProviders[x as keyof …]?.…`) so nothing else breaks: `modelProviders.utils.ts:224-247`, `modelProvider.service.ts:407/1162`, `config.materialiser.ts:619`, `EditModelProviderDrawer.tsx:53`, `useCredentialKeys.ts:61`. `MONOCHROME_PROVIDER_ICONS` is a `Set`, not an exhaustive `Record`. The two the compiler caught (`iconsMap`, `ModelProviderKey`) are filled. The gap the compiler *cannot* catch is the model catalog — see P1 below.

**4. Inference routing** is not reported; it is declared out of scope and the `mapProvider` comment in the materialiser is accurate (`services/aigateway/domain/provider.go` has no `google_agent_platform` `ProviderID`).

### Findings

| # | Sev | Finding |
|---|-----|---------|
| 1 | P1 | `validateKeyWithCustomUrl` never forwards project/location, so re-validating a saved credential always reports "couldn't reach the provider" and blocks the save |
| 2 | P1 | No `google_agent_platform/*` entries in the model catalog — the provider saves but has an empty model dropdown |
| 3 | P2 | Regional locations are probed against the global host; only `global` was verified live, and the `us-central1` test is mocked so it proves nothing |
| 4 | P3 | `defaultBaseUrl` is load-bearing in a way the comment above it denies, and nothing pins it |
| 5 | P3 | `location` vs `vertex_ai`'s `region` breaks the slot-level `region` lift in `buildProviderSlot` |
| 6 | P3 | The probe model is a single hardcoded string with no test and no alarm if Google retires it |
| 7 | P3 | The one new branch with a bug in it (empty candidates) is the one new branch with no test and no spec scenario |
| 8 | P4 | The Agent Platform host is written three times across two files |
| 9 | P4 | `expect(init.body).toBeTruthy()` pins nothing about the probe body |

Detail inline. Nothing here argues against the approach — probing by generating rather than listing is clearly right given what the two endpoints do, and the reasoning in the comments is unusually good. P1-1 and P1-2 are the two I would not merge without.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -45,8 +51,35 @@ const PROVIDER_AUTH_OVERRIDES: Partial<Record<string, AuthStrategy>> = {
...
+
+/**
+ * The model a credential check asks Agent Platform to run.
+ *
+ * Any published model would do; this one is picked because it resolves in
+ * both `global` and the regional locations. A model the project cannot reach
+ * answers 404 rather than 401, which `handleHttpError` already keeps distinct
+ * from a refused key.
+ */
+const AGENT_PLATFORM_PROBE_MODEL = "gemini-2.5-flash";
```
Aryansharma28, line 65: **P3 — one hardcoded model name is the single point of failure for all Agent Platform validation, with no test and no alarm.**

The comment is candid that this was chosen empirically ("resolves in both `global` and the regional locations") and the PR body records that `gemini-2.0-flash` already 404s on the reporter's project. That is the risk: the day Google retires `gemini-2.5-flash`, or a customer's project cannot reach it, *every* credential check for this provider returns `provider_refused` — and the 404 → not-`provider_key_invalid` handling, which is good, means the customer is told the model is missing for a model they never chose and cannot change.

No test references this constant, so a rename or a version bump is unobservable until a customer reports it. Two cheap mitigations, either is enough: assert the probe URL contains the constant so a change is at least deliberate, or probe a short list of models and treat any 200 as proof (the file already has the multi-candidate machinery for exactly this shape of uncertainty — see the four `gemini` shapes).

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        "google_agent_platform",
+        CREDENTIALS,
+      );
+
+      expect(codeOf(result)).not.toBe("provider_key_invalid");
+      expect(codeOf(result)).toBe("provider_refused");
+    });
+  });
+
+  describe("when the platform never answers", () => {
```
Aryansharma28, line 220: **P3 — the one new branch with no test is the one with the bug.**

Asked directly: would these tests pass against a broken implementation? Mostly no — the header/URL/path assertions are specific enough that a `?key=` implementation, a `GET` implementation, or a wrong-path implementation all fail, and the 401/404 pair genuinely separates `provider_key_invalid` from `provider_refused`. That is good coverage of what was written.

The gap is what was not written. `buildProbeCandidates` gained two behaviours: build the URL (five tests) and *return `[]` when project or location is missing* (zero tests). The PR body argues that case at length — "omitting either yields no probe candidates rather than a fabricated path, which `runProbeChain` reports as unreachable" — and the feature file has no scenario for it. It is also, per the L964 comment, where the bug is: the reachable path produces a network-connection error for a missing form field.

A test asserting `fetch` is never called and the rejection is `provider_unreachable` when project is `""` would have made someone look at what the customer actually reads, and would fail today if the fix is to stop resolving that case as unreachable. Worth a `@unit` scenario in the feature file too, since the argument for the behaviour is already written down in the PR — it just is not written down where it is checked.

langwatch/src/server/api/routers/providerValidation.ts
```diff
@@ -45,8 +51,35 @@ const PROVIDER_AUTH_OVERRIDES: Partial<Record<string, AuthStrategy>> = {
   anthropic: "anthropic",
   gemini: "gemini",
   elevenlabs: "elevenlabs",
+  google_agent_platform: "google_agent_platform",
 };
 
+/** Where Agent Platform serves Gemini models from. */
+const AGENT_PLATFORM_HOST = "https://aiplatform.googleapis.com";
```
Aryansharma28: **P4 — the same host is written three times across two files, and one of the three is dead.**

`AGENT_PLATFORM_HOST` here, plus `defaultBaseUrl` and `apiRoot` in `registry.tsx:85-86`. The probe uses only this one. `apiRoot` in particular is dead weight: it reaches `buildProbeCandidates` as the `apiRoot` argument but is consumed exclusively by the `gemini` strategy branch (L626-634), so for this provider it is set and never read — which is misleading, since the value being present implies the multi-path probing that `gemini` does and this provider does not.

Dropping `apiRoot` and sourcing the host from the registry (or the reverse) leaves one copy. See also the note on `defaultBaseUrl`, which cannot simply be dropped.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    it("asks the provider to generate content rather than to list models", async () => {
+      mockFetch.mockResolvedValue(generated());
+
+      await validateProviderApiKey("google_agent_platform", CREDENTIALS);
+
+      const { url, init } = sentRequest();
+      expect(url).toContain(":generateContent");
+      expect(url).not.toMatch(/\/models(\?|$)/);
+      expect(init.method).toBe("POST");
+      expect(init.body).toBeTruthy();
```
Aryansharma28: **P4 — `toBeTruthy()` on the body pins nothing.**

The surrounding assertions are precise; this one passes for `"x"`. What actually matters about `AGENT_PLATFORM_PROBE_BODY` is `generationConfig.maxOutputTokens: 1` — it is what keeps a credential check from generating (and billing) a full completion on the customer's project on every save. That constraint is currently held by a comment and nothing else.

`expect(JSON.parse(String(init.body))).toMatchObject({ generationConfig: { maxOutputTokens: 1 } })` costs one line and makes the cheap-probe property a property of the tests rather than of the author's memory.

langwatch/src/features/onboarding/regions/model-providers/registry.tsx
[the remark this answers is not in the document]
Aryansharma28: Fixed — but the fix in the same commit that first addressed this introduced a fresh staleness of the same kind: the updated comment pointed at `AGENT_PLATFORM_HOST`, which that commit had just deleted (derived from `apiRoot` instead). Corrected in `1a38cb1`, this time cross-checked against the current code rather than written from memory of what it used to say.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        ...CREDENTIALS,
+        GOOGLE_AGENT_PLATFORM_PROJECT: "acme-123",
+        GOOGLE_AGENT_PLATFORM_LOCATION: "us-central1",
+      });
+
+      expect(sentRequest().url).toContain(
+        "/projects/acme-123/locations/us-central1/",
+      );
+    });
+  });
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `8592c06` — added "A credential missing its project or location is not probed at all" to both the feature file and the test, asserting `provider_unreachable` and that nothing was sent. Landed independently, from a review I ran myself that flagged the same gap as a P3 before this comment posted.

specs/model-providers/google-agent-platform.feature
```diff
@@ -0,0 +1,66 @@
...
+  Scenario: A model the project cannot reach is not reported as a bad key
+    Given Agent Platform answers that the publisher model was not found
+    When I check the credential
+    Then I am not told the API key is invalid
+
+  @unit
+  Scenario: A provider that never answers is not a verdict on the key
+    Given Agent Platform cannot be reached
+    When I check the credential
+    Then the failure says the provider could not be reached
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `1a38cb1`. Rewrote every scenario`'s Given/When/Then to state the customer-observable outcome — is the credential valid, could the key leak, are my inputs the ones checked — rather than HTTP verb/header/URL. The protocol-level detail did not disappear, it moved to where it belongs: the test bodies still assert the literal method, header and request shape. One scenario (the generate-vs-list distinction) turned out to be the same customer-visible outcome as "a key the platform accepts is valid" once phrased as an outcome, so it merged into that one rather than being reworded into something redundant.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: Confirmed and fixed in `8592c06`. Verified against the actual code first: `ModelProviderSetup.tsx`'s `hasNonApiKeyChanges && !userEnteredNewApiKey` path really does call `validateWithCustomUrl` -> `validateKeyWithCustomUrl`, which really did rebuild `customKeys` from scratch as `{[apiKeyField]: apiKey}`. Fixed by starting from the stored `customKeys` (already decrypted plaintext, per the comment above it) rather than a blank object, with the freshly resolved key and an explicit custom URL layered on top. Generalizes to any provider with fields beyond key+endpoint, not a special case for this one. A test pins the failure mode directly (project/location omitted -> `provider_unreachable`, nothing sent) since the existing repo has no test harness for `validateKeyWithCustomUrl` itself and building one was out of proportion for this fix.

langwatch/src/features/onboarding/regions/model-providers/registry.tsx
Aryansharma28: Confirmed and fixed in `8592c06`. Checked the actual consumer before touching anything: `resolveProviderDefaultModel` in `eligibleModelProviders.ts` reads `registry?.defaultModel` BEFORE falling back to the customer's own configured/custom models — so a dangling default here would have made the AI Gateway's eligible-providers surface hand out a model string nobody chose, unconditionally. `vertex_ai` has the identical zero-catalog shape and correctly has no `defaultModel` at all; I just hadn't followed that precedent. Removed.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: Checked this live rather than accepting or dismissing it on priors, since I'd already been burned once in this same investigation by generalizing from one endpoint to another on this exact host. Result: the global host with a region in the path returns 200 for BOTH `us-central1` and a second region (`europe-west4`) against the real key, the identical shape every other verified auth path here already uses. So the specific claim — that regional locations need the `LOCATION-aiplatform.googleapis.com` subdomain — didn't hold up for either region tested. No code change; noting the verification here rather than silently dropping the comment, since a different region could still behave differently and this is worth someone re-checking if it ever comes up in practice.

langwatch/src/features/onboarding/regions/model-providers/registry.tsx
Aryansharma28: Fixed in `8592c06`, and again in `1a38cb1` after the first fix introduced a fresh instance of the same problem (see the reply on the registry-comment thread above) — third time checked directly against the code that exists, not description.

langwatch/src/server/gateway/config.materialiser.ts
Aryansharma28: Confirmed and fixed in `8592c06`. `buildProviderSlot` really does look up the literal key `"region"` on credentials (`pickString(credentials, "region")`), and the materialiser really did emit `location` instead — this provider would have silently never gotten a slot-level region once anything reads it. Renamed the credentials key to `region`; the customer-facing env var stays `GOOGLE_AGENT_PLATFORM_LOCATION` since that's Google's own term and what the drawer shows.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: Left as a single hardcoded model, deliberately, but flagging the tradeoff rather than treating this as closed: any credential probe needs SOME model to ask about, and every alternative (config knob, discover-a-model-first call) adds a request or a setting for a problem that hasn't happened yet. If Google retires `gemini-2.5-flash` this becomes a live bug — a real cost, just not one I'm paying down speculatively today. Noting it here so it's findable when it does happen rather than a silent landmine.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
Aryansharma28: Fixed in `8592c06` — this is the same gap as the "second validation entry point" P1 above, from the same code path. Added the test and the matching feature scenario.

langwatch/src/server/api/routers/providerValidation.ts
Aryansharma28: Fixed in `8592c06`. Derived the host from `apiRoot` (already threaded through `buildProbeCandidates` for the `gemini` branch's identical need) instead of a separate `AGENT_PLATFORM_HOST` constant — one string now, in the registry, instead of three.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
Aryansharma28: Fixed in `8592c06`. Now parses the actual body and asserts `generationConfig.maxOutputTokens === 1` — the property that keeps this probe cheap — rather than a truthiness check `"x"` would also satisfy.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      await expect(
+        validateProviderApiKey("google_agent_platform", {
+          GOOGLE_AGENT_PLATFORM_API_KEY:
+            CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        }),
+      ).rejects.toMatchObject({ code: "provider_unreachable" });
+
+      // Nothing to ask without a project and location, so nothing was sent.
+      expect(mockFetch).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
Aryansharma28: Confirmed and fixed in `d868546`. Verified this live rather than trusting the reasoning: flipped `!project || !location || !apiRoot` to all-`&&` locally, ran the suite — 3 failed, 7 passed, all three of the new per-field cases caught it (including "both missing", since the guard has a third `apiRoot` term the all-`&&` version still trips on). Reverted, confirmed clean, then wrote the real fix: three explicit tests, one per missing field. `it.each` was my first attempt and did not work — its title-string argument does not satisfy `check-feature-parity`'s requirement that `@scenario` sit directly above a bare `it(`, so this ended up as three plain tests bound to the same scenario instead.

langwatch/src/server/api/routers/__tests__/agentPlatformValidation.unit.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      });
+    });
+
+    /** @scenario A credential missing its project or location is not probed at all */
+    it("is unreachable when only the project is missing", async () => {
+      await expectUnreachable({
+        GOOGLE_AGENT_PLATFORM_API_KEY: CREDENTIALS.GOOGLE_AGENT_PLATFORM_API_KEY,
+        GOOGLE_AGENT_PLATFORM_LOCATION:
+          CREDENTIALS.GOOGLE_AGENT_PLATFORM_LOCATION,
+      });
```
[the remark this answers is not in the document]
Aryansharma28: Fixed in `e7e3a7a` — renamed all three to describe the operation rather than the resulting state ("rejects the check without probing when X is missing").

sergioestebance (approved):
In the name of the Father...

Issue #5912, opened 2026-07-18
Replace the lossy analytics folds with append-only map projections + read-time aggregation


0xdeafcafe:
## Problem

`traceAnalytics` and `evaluationAnalytics` are fold projections whose durable stores cannot be read back. `BaseAnalyticsFoldStore.get()` returns `null` unconditionally, by design — the slim rows are lossy (trimmed attributes, booleans instead of arrays):

> `get` returns `null` for every aggregate we've shipped so far — the slim rows are lossy by design, so fold state cannot be read back. Fold-state continuity therefore REQUIRES two layers around any store built on this base: a RedisCachedFoldStore wrapper at registration (warm path) and the fold's `refoldOnStoreMiss` option (event-log rebuild on a cache miss). Without both, a miss folds only the delivered batch and partial rows overwrite complete ones.

That is the only reason `refoldOnStoreMiss` exists. It is set on exactly two projections:

- `traceAnalytics.foldProjection.ts:625`
- `evaluationAnalytics.foldProjection.ts:302`

Every other fold — `traceSummary` included — has a readable store, so a Redis cache miss simply falls through to ClickHouse and folds on top. No refold. That is the behaviour we actually want everywhere; these two are the exception, and they pay for it with a full event-log rebuild on every cache miss.

## Proposal

Replace the fold with append-only map projections and let ClickHouse aggregate at read time. `traceAnalyticsRollup.mapProjection.ts` already proves the pattern for the additive half — one row per `SpanReceivedEvent` (`spanCount: 1`, `costSum`, token sums), summed on read, no state, no cache, no refold.

## What makes it more than a rename

The fold consumes nine event types, and not all of them are additive. Some mutate trace-level state:

```ts
// traceAnalytics.foldProjection.ts:843-846 — annotation removal
const ids = state.annotationIds ?? [];
return { ...state, annotationIds: ids.filter((id) => id !== event.data.annotationId) };
```

`annotationAdded` / `annotationRemoved` / `annotationsBulkSynced` are set reconciliation; `topicAssigned`, `traceNameChanged`, `originResolved` are last-write-wins on a scalar. A map projection never sees the current set, so "remove X" is not directly expressible.

It *is* expressible without a fold, at a cost:

- annotations → one append-only row per event `(traceId, annotationId, action, updatedAt)`; derive `HasAnnotation` read-side via `argMax(action, updatedAt)` per `annotationId`
- LWW scalars → `argMax(value, updatedAt)`
[...]

PR #6368, opened 2026-07-30
docs(adr): record LWQL as a read-only, tenant-scoped query language (ADR-081)

### LINKED ISSUE #6346: LangWatch Query Layer: a read-only, tenant-scoped query language, service and API
# LangWatch Query Layer (LWQL)

A read-only, tenant-scoped query language over LangWatch observability data, exposed as
one service behind both the REST API and the UI's tRPC layer.

## For Humans

**LWQL is a read-only, tenant-scoped SQL-like query language over traces and spans — one service behind REST and tRPC.** LangWatch data is trace-centric (traces = end-to-end runs, spans = internal steps like LLM calls, tool calls, retrieval), and users, backend jobs, and agents all need to query it directly without writing raw ClickHouse SQL or learning a form builder.

**Ask:** review and ratify the design decisions below, and review PR #6413.

You can ask questions like:

```sql
SELECT model, sum(cost_usd) / count(*) AS cost_per_trace, p95(duration_ms) AS p95_latency
FROM traces
WHERE has_error = true AND started_at >= now() - INTERVAL 24 HOUR
GROUP BY model
ORDER BY cost_per_trace DESC
```

```sql
SELECT tool_name, p95(duration_ms) FROM spans GROUP BY tool_name
```

```sql
SELECT count(*) FROM traces WHERE has_error = true AND started_at >= now() - INTERVAL 7 DAY
```

```sql
SELECT date_trunc('hour', started_at) AS t, count(*) AS traces
FROM traces
WHERE started_at >= now() - INTERVAL 24 HOUR
GROUP BY t
ORDER BY t
```

Results are tables that back charts — the New Relic model (see #5670).

What the system will never do, no matter how you phrase the query:

- Write or modify anything — it's read-only end to end.
- Show you another project's data — tenant scope is enforced server-side, not by anything in your query.
- Return content your plan has redacted — those fields are refused as query targets, not just scrubbed from output.
- Run more than one statement per query.

```mermaid
flowchart TD
    subgraph untrusted["untrusted input"]
        A[Caller: SQL text or IR JSON] --> B{REST or tRPC}
        B --> C[Single service]
        C -->|text| D[node-sql-parser]
        D --> E[Default-deny walker]
        C -->|IR direct| F[IR JSON]
        E --> G
        F --> G
    end

    subgraph closed["closed, developer-authored"]
        G[zod-validated IR<br/>SECURITY BOUNDARY]
        G --> H[Compiler]
        H --> H1[Catalogue lookup only]
        H --> H2[Tenant scope injected server-side]
        H --> H3[Values as bound params]
        H --> H4[Gated fields refused]
        H1 --> I[(ClickHouse)]
        H2 --> I
        H3 --> I
        H4 --> I
        I --> I1[readonly mode]
        I --> I2[execution timeout]
        I --> I3[row caps]
    end
```

_Status: implementation in PR #6413 (includes ADR-081). Parser rewrite per ADR decision 9 in progress. Full spec below._

## Background

LangWatch already captures rich observability data for AI workflows, and much of the
product is naturally **trace-centric**: a trace represents an end-to-end execution, and
spans capture the internal steps such as LLM calls, tool calls, retrieval, latency,
errors, and cost. We want users and internal systems to query this data directly in a
way that is easy to understand, safe in a multi-tenant environment, and consistent
across UI, agent, and backend use cases.

Trigger.dev is a useful reference: it exposes a SQL-like query layer compiled to secure
ClickHouse queries, including over REST. That suggests a proven pattern — keep the
user-facing language constrained, enforce tenant isolation server-side, and compile to
ClickHouse underneath.

**Product shape:** a raw query input that yields a table. No form builder. Once a table
is on screen, the user picks a chart type and axes from the columns they got back —
the New Relic / Retool pattern.

## Goals

- An easy-to-understand SQL-like query language for LangWatch data.
- Direct API access, so clients, backend jobs, and agents can query without the UI.
- Tenant isolation and read-only behaviour enforced server-side.
- A surface area small enough to stay maintainable.
- One set of query semantics powering UI, API, and agent-generated requests.
- LWQL results are the tabular backend for configurable chart widgets — the New
  Relic model (query per widget → table → visualization). See the dashboard-side
  work in #5670 and #5389.

## Non-goals

- Full SQL compatibility.
- Arbitrary joins across all backend tables.
- Write operations, DDL, or mutation support.
- Exposing raw ClickHouse SQL to end users.
- A general-purpose analytics engine for all business data.

## User stories

**End user.** As a LangWatch user, I want to query traces and spans using a simple
SQL-like language so that I can inspect failures, latency, cost, and model behaviour
without learning ClickHouse SQL.

**Developer / backend consumer.** As a developer, I want to call an API directly with a
query so that I can power dashboards, automation, and internal tools from the same safe
query layer.

**Agent.** As an agent or assistant, I want to generate a structured query or DSL
request so that I can help users explore observability data without writing unsafe SQL.

---

## Design decisions

These are settled. Rationale lives in **ADR-081** (PR #6368).

### 1. Two input forms, one security boundary

Both input shapes below are supported, because **the text form compiles to the
structured form**. The structured JSON IR — not the text — is the security boundary.
Anything that reaches ClickHouse has passed through it.

### 2. Allowlist-total compilation

Every entity, field, and aggregation is a **closed enum** mapped to a
developer-authored ClickHouse expression.

- User input is bound as **query parameters**, never interpolated.
- **Identifiers are never taken from user input** — a field name selects a
  pre-written expression; it does not become one.
- A query naming anything outside the enums **fails to compile**. It does not
  degrade, sanitise, or pass through.

### 3. Tenant scope is injected, never accepted

The compiler applies tenant scope. A caller cannot supply, override, or widen it, and
there is no query text that removes it.

### 4. Read-only

No writes, deletes, or schema changes exist in the language or the service.

### 5. One service, two transports

REST and tRPC call the same service. They are transports, not separate
implementations, so the two surfaces cannot drift in what they permit.

### 6. LWQL compiles to IR and does not choose storage

The query layer emits IR and hands it to the existing analytics destination router,
which already decides which table or rollup answers a given shape. LWQL names no
tables. Storage-layer work therefore proceeds independently of this issue.

Existing time-bucketing is reused rather than re-derived — it is already
timezone-correct across minute / interval / day / week / month.

### 7. Content-visibility gating applies to filters, not just output

**A filter on a field is a projection of that field, one bit at a time.** So any field
subject to content-visibility gating is gated identically as a **filter target** and an
**aggregation target** as it is as an output column.

The gated field set is **derived from `app-layer/traces/visibility-window.service`**,
never restated in the query layer. A hand-maintained parallel list drifts the moment a
field is added, so **a test asserts parity** — extending the redaction set without
teaching LWQL about it fails CI.

### 8. v1 scope: traces and spans

`evaluations` and `sessions` are deliberately deferred. The entity set is designed to
extend; v1 ships two entities well rather than four thinly.

### 9. Expressions: standard SQL syntax, closed-dialect power

Charts need computed columns (`sum(cost_usd) / count(*)`, `avg(duration_ms) * 1000`), so the language supports arithmetic (`+ - * /`, parentheses, correct precedence), column aliases (`AS`, and alias without `AS`), and `HAVING`. The surface is **standard SQL syntax and nothing else** — no NRQL keywords, no invented dialect; anything you can write here is ordinary SQL a human or an SQL-literate agent already knows. The power/safety resolution is the NRQL *pattern*, not NRQL's syntax: like NRQL, LWQL is a closed dialect whose power comes from arbitrary **expressions over a fixed field vocabulary** — but where NRQL invents its own keywords to get there, LWQL spells everything in standard SQL. Expressions compose allowlisted fields, literals, the closed aggregate set, and a closed operator/function set into an IR expression tree; identifiers still only ever resolve through the catalogue, so allowlist-totality (decision 2) is preserved. What stays excluded regardless of expressive power: subqueries, joins, unions, open function names, and anything that names a table or column the catalogue doesn't. Content gating extends to expressions: an expression referencing a gated field is refused whole (decision 7's oracle argument applies to `gated_field * 0 + 1` exactly as to the bare field).

### 10. Timeseries: time-bucketing with selectable granularity, in standard SQL

Trace and span data is time-ordered in ClickHouse, and the primary chart shape is a metric over time. LWQL supports bucketing timestamps in `SELECT` and `GROUP BY` via standard-SQL-shaped functions: `date_trunc('minute'|'hour'|'day'|'week', started_at)` (ANSI-standard) and `time_bucket(INTERVAL '5' MINUTE, started_at)` for arbitrary granularity. The caller selects granularity by choosing the bucket size — there is no special keyword for it, which is exactly where NRQL's `TIMESERIES` becomes plain SQL. Buckets compile to the corresponding ClickHouse time functions through the catalogue like every other expression (decision 2), and the max-time-range and row caps (Safety, below) bound bucket-count explosion: e.g. 90 days at 1-minute granularity exceeds the row cap and is refused outright, not silently truncated.

---

## How it works, in simple terms

You send a query one of two ways: as SQL-looking text, or as structured JSON that describes the same query directly. Both go to the same single endpoint (available as REST or tRPC — it's the same underlying service either way, so the two can't drift apart or disagree).

If you sent text, it's parsed by `node-sql-parser`, a widely-used, battle-tested open-source SQL parser, into a syntax tree — a structured breakdown of what you typed.

That tree is then converted by a **walker** into a small, closed JSON format called the IR (intermediate representation). The walker is **default-deny**: it has an explicit list of exactly which SQL constructs it understands, and it rejects anything not on that list, rather than trying to guess or pass through the rest. Any field name in your query (like a column or metric name) is looked up in a **catalogue** — a closed list where every allowed name maps to a specific, developer-written ClickHouse expression. The text you typed never becomes a raw database identifier; it only ever selects an entry from this pre-approved list.

The IR — whether it came from parsed text or was submitted directly as JSON — is then checked against strict schemas (via `zod`) that reject anything with unrecognized fields.

Next, a compiler turns the validated IR into actual ClickHouse SQL. Your tenant scope (which project's data you're allowed to see) is injected automatically from your authenticated session — it's never read from the query itself, so there's no way to type your way into someone else's data. Every value in the query becomes a bound parameter (the standard, safe way databases accept user values), never pasted directly into the SQL string.

Before that, content gating kicks in: if your plan redacts certain fields (like message content), those fields are refused entirely as things you can select, filter on, or group by — not just hidden from the output. That's because even a filter like "where content contains X" can be used to guess redacted content one yes/no answer at a time. The list of gated fields isn't maintained separately by hand — it's derived from the platform's real redaction logic, with an automated test that fails if the two ever drift apart.

Finally, the query actually runs against ClickHouse with server-side guardrails: read-only mode, an execution timeout, and a cap on the number of rows returned.

## Why this stack (defense of choices)

**Why SQL-shaped at all.** This was the ask in this issue: a raw query surface, not a form builder — something both humans and AI agents already know how to write.

**Why `node-sql-parser` (Apache-2.0).** An earlier spike hand-rolled a 777-line parser. Probed with 15 ordinary SQL constructs, it rejected all 15, with at least five real bugs: negative numbers (`-1`), exponent notation (`1e5`), leading-dot decimals (`.5`), a trailing semicolon, and an alias without `AS`. `node-sql-parser` parsed all 8 of the probes we ran against it, including every case that broke the hand-rolled version. We considered alternatives: ANTLR is the most robust option but needs Java-based code generation and has no maintained ClickHouse grammar available on npm; Chevrotain still means writing and owning a grammar ourselves; `sql-parser-cst` is GPL-licensed, which is incompatible with a commercial SaaS product.

**Why the parser is deliberately not the security boundary.** Security doesn't come from how strict the parser is — it comes from the closed IR, the catalogue, server-injected tenant scope, and bound parameters. Because the walker is default-deny, a permissive, general-purpose parser upstream of it is safe: parsing something doesn't mean the walker will accept it.

**Why one service, two transports — and why IR is also a public input.** REST and tRPC both call the identical service, so there's exactly one implementation of the query logic to get right. Programmatic and agent callers can skip the parser entirely and submit IR directly, since it's just JSON that follows the schema.

## What's a robust library vs. what must stay ours

| Concern | Approach |
|---|---|
| SQL parsing | `node-sql-parser` (off-the-shelf) |
| IR schema validation | `zod` (off-the-shelf) |
| "Did you mean...?" suggestions | `fastest-levenshtein` (off-the-shelf) |
| Duration literals (`24h`, etc.) | `ms` (off-the-shelf) |
| Execution safety | ClickHouse server settings — readonly, execution timeout, row caps (off-the-shelf); planned follow-up: ClickHouse row policies + a restricted read-only DB user |
| Field → ClickHouse expression mapping (the catalogue) | ours — this *is* the product |
| The walker's default-deny allowlist | ours — this *is* the security policy |
| Content-gating map + its parity test | ours — this *is* the security policy |
| Tenant-scope injection in the compiler | ours — this *is* the security policy |

In short: we outsource the solved problems, and we own the policy.

### Query language

- `SELECT`, `FROM`, `WHERE`, `GROUP BY`, `ORDER BY`, `LIMIT`.
- Entities: `traces`, `spans` (v1 — see decision 8).
- Operators: equality, inequality, range, `IN`, `LIKE`, boolean checks.
- Aggregates: `count`, `sum`, `avg`, `min`, `max`, `p95`.
- Trace-friendly fields including `trace_id`, `span_id`, `parent_span_id`,
  `project_id`, `model`, `tool_name`, `error`, `duration_ms`, `cost_usd`,
  `token_count`, `started_at`.
- Column aliases: `AS` and bare aliasing; the alias is the returned column name
  (chart bindings depend on it).
- Arithmetic expressions: `+ - * /` with parentheses and standard precedence, over
  numeric allowlisted fields, literals, and aggregate results (e.g.
  `sum(cost_usd) / count(*) AS cost_per_trace`).
- `HAVING` over aggregate expressions.
- Expressions are IR expression trees; every leaf is a catalogue field or a
  literal; every operator/function is from the closed set. Unknown → rejected
  with a suggestion.
- Time-bucketing in `SELECT` and `GROUP BY`: `date_trunc('minute'|'hour'|'day'|'week', started_at)`
  and `time_bucket(INTERVAL '5' MINUTE, started_at)` for arbitrary granularity.
- Granularity is chosen by the bucket size passed to `date_trunc` / `time_bucket` —
  no separate keyword.
- Bucketed results can be ordered by the bucket column (`ORDER BY` the alias).

### API surface

- A query execution endpoint usable without the UI.
- Accepts **either** the SQL-like text **or** the structured JSON IR (decision 1).
- Results in a normalized JSON shape.
- Pagination, plus optional async execution for expensive queries.
- Validation / dry-run endpoint.
- Query metadata: execution time, row count, truncation indicators.
- `explain` (generated SQL) available to **internal users only**.

### Tenant isolation

- Scope derived from authenticated context.
- A user-supplied tenant filter is never the security boundary.
- Scoping enforced in the compiler / execution layer.
- Any query attempting to bypass allowed scope is rejected.
- All returned rows are tenant-safe.

### Safety

- Read-only.
- Only allowlisted fields, functions, and entities (decision 2).
- All user input parameterized.
- Sensible limits and time-range constraints by default.
- Expensive or unbounded scans prevented by default.

### UX

- Readable to SQL-literate users.
- Errors explain what was invalid **and how to fix it**.
- The same query works in UI, API, and agent workflows.
- Saved queries and dashboards can layer on top later.

## API shape

**Text input** — `POST /api/query`

```json
{
  "query": "SELECT trace_id, duration_ms FROM traces WHERE started_at >= now() - interval 24h ORDER BY duration_ms DESC LIMIT 100"
}
```

**Structured input** — `POST /api/query`

```json
{
  "from": "traces",
  "select": ["trace_id", "duration_ms"],
  "where": [
    {"field": "started_at", "op": ">=", "value": "now-24h"}
  ],
  "order_by": [{"field": "duration_ms", "direction": "desc"}],
  "limit": 100
}
```

**Response**

```json
{
  "data": [],
  "meta": {
    "row_count": 0,
    "execution_ms": 12,
    "truncated": false
  }
}
```

## Acceptance criteria

- [ ] Users can query traces and spans with the SQL-like syntax.
- [ ] API consumers can execute the same query surface directly, without the UI.
- [ ] Queries are automatically scoped to the authenticated tenant/project.
- [ ] A user cannot override tenant isolation through the API or the query text.
- [ ] A query naming a field, entity, or function outside the allowlist fails to
      compile, rather than being sanitised or partially executed.
- [ ] A content-gated field cannot be used as a filter or aggregation target while
      gating is active, and parity with the redaction service is enforced by a test.
- [ ] Result metadata reports row count, execution time, and truncation.
- [ ] `SELECT avg(duration_ms) AS latency FROM traces GROUP BY model` returns a
      column literally named `latency`.
- [ ] `SELECT sum(cost_usd) / count(*) AS cost_per_trace FROM traces GROUP BY model`
      compiles and computes with correct precedence; `(1 + 2) * 3`-style
      parenthesized precedence proven by test.
- [ ] `HAVING count(*) > 100` filters groups.
- [ ] An expression referencing a content-gated field (in any position, e.g.
      multiplied by zero) is refused entirely for a gated caller.
- [ ] An unknown function or operator inside an expression is rejected with an
      "unknown function/operator" error, never passed through.
- [ ] `SELECT date_trunc('hour', started_at) AS t, count(*) FROM traces GROUP BY t
      ORDER BY t` returns one row per hour bucket.
- [ ] `time_bucket(INTERVAL '5' MINUTE, started_at)` buckets at 5-minute
      granularity; changing the interval changes the bucket count accordingly.
- [ ] A granularity/time-range combination whose bucket count exceeds the row cap
      is refused with an explanatory error, not truncated.

## References

- **ADR-081** — design rationale: #6368
- **Spike / New-Relic-style dashboards** — prior exploration of a read-only
  ClickHouse query interface, and the design for chart widgets consuming LWQL
  tables: #5670
- **Chart-this-query entry point** — dashboard-side integration: #5389
- **Implementation** — PR #6413 (includes ADR-081)

</details>

drewdrewthis:
**Scariest thing** — this ADR asserts that **ADR-028 needs an amendment from its owner**, and describes a content-classification gap that exists in the platform today, not only in this proposal. If that reading is wrong, a Draft ADR is publicly mis-stating our data-visibility posture. It is sourced from ADR-028's own summary line, but it is a claim about someone else's decision and should be read as one.

**Start here** — `dev/docs/adr/081-lwql-query-language.md`, the **Consequences** section. Everything else is scaffolding around that paragraph.

## Why

Refs #6346. The product ask is a raw query surface over traces and spans — hit the platform with a query, get a table, then drive a chart off the result — and no ADR existed for it. The decision reasoning had been written into a personal notes repo, which meant a decision about this platform's public API surface and content-visibility model was invisible to the engineers it affects.

> **Deliberately not `Closes #6346`.** The template asks for it, but this PR ships a `Draft` ADR, not the feature. #6346 must stay open for the implementation — auto-closing a tracking issue on an ADR merge would be wrong. Flagging the deviation rather than silently dropping the line.

## What changed

- **`dev/docs/adr/081-lwql-query-language.md`** (new, `Status: Draft`) — settles four things: the name `LWQL`; query text parsing into a **closed JSON IR** as the security boundary; **traces *and* spans** in v1; and one service behind two transports with compiler-injected tenant scope.
- **`dev/docs/adr/README.md`** — one row added, indexing 081.

`Draft`, not `Proposed`, because four questions are genuinely open and three of them were resolved once and then reopened under adversarial review. They are listed in the ADR rather than implied.

## How it works

The load-bearing idea is **allowlist-totality**: every dimension, metric and aggregation is a closed enum mapped to a developer-authored ClickHouse expression, so ClickHouse binds *values* and never identifiers. A text front-end can then be safe by construction, because the parser can only emit closed-enum members.

Two alternatives were rejected. A **structured form/filter UI** inverts the requirement — a form only asks questions someone anticipated, and this surface exists for the unanticipated one. **Raw SQL passthrough** was never on the table; it makes tenant isolation a review problem instead of a compiler property.

The engine choice is explicitly *not* made here, and the ADR records why the dependency is ordered: time-bucketing decides it, the spike has none, and the analytics read path (ADR-034/066/068) already has it — so choosing the engine first inverts the dependency. That question also touches in-flight work on #5912 and needs that work's owner.

## Test plan

Docs-only change, so the check is that the artifact is well-formed, its references resolve, and it says what it claims. Verified this session:

```console
$ # all relative ADR links resolve on this branch
028-visibility-blur-teaser-redaction.md              OK
034-event-sourced-analytics-materialization.md       OK
066-projection-clickhouse-cached-store.md            OK
068-windowed-clickhouse-reads.md                     OK
081-lwql-query-language.md                           OK

$ # read-back of the PUSHED file (not the local copy)
lines: 70
required template headings (Context/Decision/Consequences): 3
status line: **Status:** Draft
placeholder/TODO/truncation: none
  (the one "..." match is inside `z.enum(Object.keys(...))`, a real expression)

$ # numbering collision check across every open PR touching dev/docs/adr/[0-9]*
PR #6368 -> dev/docs/adr/081-lwql-query-language.md      <- only claimant of 081

$ # PII / secret scan
clean
```

**Visual surface:** GitHub's rendered view of the file is the artifact — [read `081-lwql-query-language.md` as rendered](https://github.com/langwatch/langwatch/blob/adr/081-lwql-query-language/dev/docs/adr/081-lwql-query-language.md). Scoping this honestly: a markdown-only change has no running surface to exercise, so the evidence above is a verification transcript, **not** a screenshot of a working feature, and I am not claiming otherwise.

## Anything surprising?

- **Possible path conflict with #6307** (`refactor(build): single workspace + repo reorganise`), which touches eight files under `dev/docs/adr/`. If it relocates the ADR tree, this branch needs a rebase.
- **ADR numbering already collides in twelve places on `main`** — `006 010 014 018 021 022 026 027 028 029 030 040` each appear on two different files. This ADR cites other ADRs **by filename, not number**, so its references stay unambiguous either way.
- **The ADR README table indexes 51 of 84 ADRs.** I added only the 081 row; backfilling 33 would bury this ADR in an unrelated diff. Happy to do that as a separate PR.
- **A severity claim I made on #6346 while researching this has been corrected in place.** I had chained `analytics:view`-for-`VIEWER` to RBAC-bypassing legacy project keys and called the gap reachable by the lowest-privilege credential in the system. That describes the *proposed* REST surface, which does not exist yet; the composition reachable today is session-authenticated tRPC. The ADR states the design requirement without the severity claim, and the exploit construction is deliberately absent from this repo.

## Deployment Impact

**None.** This PR adds one markdown file (`dev/docs/adr/081-lwql-query-language.md`) and one row to the ADR README table. No chart, service, environment variable, migration, or runtime code path is touched.

- **Schema / migrations:** none.
- **Environment variables:** none added, removed, or renamed.
- **Services / charts:** untouched.
- **Rollback:** revert the commit; there is no deployed state to unwind.

The design this ADR records is implemented separately in #6413, which carries its own deployment considerations.


drewdrewthis:
Consolidated into #6413 per owner direction — the ADR and its implementation don't make sense as separate PRs. All 7 commits from this branch are cherry-picked onto `issue6346/lwql-service` (through decision 9, the parser-technology decision). Review the ADR there: https://github.com/langwatch/langwatch/pull/6413/files

PR #6387, opened 2026-07-30
fix(analytics): bound the evaluation_runs JOIN on ScheduledAt, the column it has

0xdeafcafe:
## What

#6383 bounded the `evaluation_runs` JOIN and its dedup subquery on `OccurredAt`, described as "the partition column". `evaluation_runs` has no `OccurredAt` — it is `PARTITION BY toYearWeek(ScheduledAt)`.

The wrong name did not fail loudly. These subqueries nest inside a scope whose base table is `trace_summaries`, which *does* have `OccurredAt`, so the identifier resolved outward — and an outward reference inside the `IN` makes the dedup a correlated subquery, which ClickHouse refuses:

```
Correlated subqueries are not supported as IN function arguments yet, but found in
expression: (TenantId, EvaluationId, UpdatedAt) IN (SELECT TenantId, EvaluationId,
max(UpdatedAt) FROM evaluation_runs WHERE (TenantId = '…') AND (OccurredAt >= …))
```

Every analytics query carrying an evaluation metric or filter raised it.

## Fix

Bound `ScheduledAt`. `EXPLAIN indexes=1` against a real ClickHouse 25.10 confirms the pruning the original PR was after survives — the partition key *and* `idx_scheduled_at`'s minmax both narrow the read.

The cushion widens from the spans' 2 days to 7. A span's `StartTime` sits inside its trace's lifetime, so that bound cannot drop a row; an evaluation is only scheduled *after* its trace arrives, so this bound has to cover scheduling lag rather than clock skew. `EVALUATION_ANALYTICS_READ_WINDOW_MS` already pegs that lag at ±7 days for the same domain, so this tracks the fold's number instead of inventing one. Weekly partitions make it ~1 extra partition per side.

An evaluation scheduled further out than that window drops from the range — the trade the bound accepts for pruning, now stated where it is made rather than asserted as free.

## Verification

Native local ClickHouse/Redis/Postgres, `src/server/analytics/clickhouse/__tests__/`:

| | Test files | Tests |
|---|---|---|
| Before (main @ 80f7a818c2) | **7 failed** / 1 passed | **24 failed** / 69 passed |
| After | 8 passed | 93 passed, 3 skipped |

Also green: `pnpm test:unit src/server/analytics/` — 15 files, 365 tests. Biome clean on both files (only the file's pre-existing complexity warnings remain).

## Note for a follow-up, not fixed here

The integration suite catches this the moment it runs, so it is worth a look at why it landed green. On a non-heavy PR the integration job runs `pnpm test:integration --changed origin/main`, and #6383 touched exactly the two files these tests import.

### DIFF
--- a/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
+++ b/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
@@ -80,18 +80,33 @@ const SPAN_TIME_FILTER_START_END =
   "AND StartTime < {endDate:DateTime64(3)} + INTERVAL 2 DAY";
 
 /**
- * The same envelope for `evaluation_runs`, whose partition column is
- * `OccurredAt`. Its JOIN took no bound at all until now, so every query
- * carrying an evaluation metric or filter cold-scanned the table twice — the
- * outer read plus the dedup's full-table GROUP BY. One constant per caller date
- * regime, mirroring the span pair above.
+ * The same envelope for `evaluation_runs`, whose JOIN took no bound at all
+ * until now, so every query carrying an evaluation metric or filter cold-scanned
+ * the table twice — the outer read plus the dedup's full-table GROUP BY. One
+ * constant per caller date regime, mirroring the span pair above.
+ *
+ * The column is `ScheduledAt`: the table is `PARTITION BY toYearWeek(ScheduledAt)`
+ * and has no `OccurredAt` at all. That name belongs to `trace_summaries`, the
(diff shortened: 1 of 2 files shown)


rogeriochaves:
Superseded by #6399, which merged as `c7ce7dbfe7` and bounds the same JOIN on `ScheduledAt` plus a table-qualified `UpdatedAt` for deployments that predate the unified DDL, applied to the dedup inner as well as the outer subquery. Verified by execution in #6392: 80/80 parts before, 8/80 after, on both scans. Closing this one so the lane has a single history.

PR #6389, opened 2026-07-31
fix(analytics): drop the evaluation_runs bound on a column that table has not got

0xdeafcafe:
Fixes the 16 test failures currently red on `main` (surfaced on #6375, which is charts-only and did not cause them).

## The 15 ClickHouse failures

`aa59f8da98` (#6383, merged 2026-07-30) bounded the `evaluation_runs` JOIN **and its dedup subquery** on `OccurredAt` to prune partitions. That table has no `OccurredAt` column — it is `PARTITION BY toYearWeek(ScheduledAt)`.

The name did not fail as an unknown identifier. Inside the subquery it resolved *outward* to the enclosing `trace_summaries` scope, which does have `OccurredAt`, turning the IN-tuple dedup into a correlated subquery that ClickHouse refuses:

```
Correlated subqueries are not supported as IN function arguments yet, but found in expression:
(TenantId, EvaluationId, UpdatedAt) IN (SELECT TenantId, EvaluationId, max(UpdatedAt) FROM evaluation_runs ...)
```

**This is a live read-path outage, not just red CI** — every analytics query carrying an evaluation metric or filter throws. #6383 merged with four of six integration shards already failing, so the suite did catch it.

### Why reverted, not renamed to `ScheduledAt`

Renaming looks like the obvious fix and is a data-loss bug. The outer date range is the **trace's**, and `ScheduledAt` does not track it:

- No production path emits `EvaluationScheduledEvent` — it is dead plumbing kept for the fold's exhaustiveness check — so `scheduledAt` stays null and the repository falls back to `createdAt`, i.e. wall-clock write time.
- Evaluations V3 **"Rerun"** replays months-old trace ids at today's clock (`experiments-v3/execution/orchestrator.ts`).

A ±2-day envelope would silently drop those rows. Online evals and guardrails are fine (seconds of drift); the problem is one table serving both regimes with nothing distinguishing them. The pruning win is real, still unclaimed, and left in a comment rather than smuggled in as a rename.

The repo already agrees on the column — `cold-scan-detector.ts` maps `evaluation_runs: ["ScheduledAt"]` and `filters/clickhouse/query-helpers.ts` bounds it that way. **No other call site had the bug.**

## The 1 timezone failure (unrelated flake)

Three DateTime64-decode suites force `TZ=Asia/Kolkata` at module scope, because the assertion only has teeth off UTC. The unit pool runs `isolate: false`, so one process and VM context carry every file a worker picks up and a neighbour can clobber that assignment.

Evidence it is nondeterministic rather than a code change: `test-unit (1)` **passed** on #6383's run and **failed** on #6375's run, same files, nothing touching them in between — and only one of the three identical suites failed.

`useNonUtcTimezone` re-applies the zone per test and moves the did-it-take guard into a `beforeEach`, so it covers every test rather than one. Verified on linux and darwin that a late re-application works even inside an already-warm vm context.

I could not reproduce this one locally (it passes standalone, in shard 1/4, and single-worker), so the mechanism is inferred from the above rather than directly observed. The re-application is harmless either way, and the guard still fails loudly if the zone ever stops taking.

## Verification

- Reproduced the exact CI error against ClickHouse `25.10.2.65` (CI's pin) on unmodified `main`, then confirmed green with the fix — a real before/after control, not just an after.
- All 7 originally-failing integration files pass; whole `analytics/clickhouse` + `app-layer/analytics` + `app-layer/evaluations` integration suites green (93 + 48 tests).
- Unit shard 1/4 green; `typecheck` **and** `typecheck:tests` clean; biome clean on changed files.
- Confirmed the guard still fails loudly when the zone does not take.

## Left alone, deliberately

`monitor-performance.clickhouse.repository.ts:86-87` carries this same ±2-day `ScheduledAt` bound on the same false premise ("runs are scheduled after their trace occurs"). Pre-existing, narrower blast radius (scoped to monitor evaluator ids), out of scope here — but it is the same bug and worth a follow-up.

## Deployment Impact

**Env vars added/changed:** none.

**Behaviour:** restores evaluation analytics to its pre-#6383 (working) state. Queries lose the partition-pruning attempt that never functioned — it errored rather than pruned — so there is no performance regression relative to what is actually running.

**Rollback:** revert this commit; that restores the broken query.

### DIFF
--- a/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
+++ b/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
@@ -79,19 +79,12 @@ const SPAN_TIME_FILTER_START_END =
   "AND StartTime >= {startDate:DateTime64(3)} - INTERVAL 2 DAY " +
   "AND StartTime < {endDate:DateTime64(3)} + INTERVAL 2 DAY";
 
-/**
- * The same envelope for `evaluation_runs`, whose partition column is
- * `OccurredAt`. Its JOIN took no bound at all until now, so every query
- * carrying an evaluation metric or filter cold-scanned the table twice — the
- * outer read plus the dedup's full-table GROUP BY. One constant per caller date
- * regime, mirroring the span pair above.
+/*
+ * There is deliberately no equivalent pair for `evaluation_runs`. See the note
+ * in `buildJoinClause` (field-mappings.ts): that table has no `OccurredAt`, and
+ * the column it is actually partitioned by, `ScheduledAt`, does not track the
+ * trace date range these constants carry.
  */
-const EVAL_TIME_FILTER_BOTH_PERIODS =
-  "AND OccurredAt >= {previousStart:DateTime64(3)} - INTERVAL 2 DAY " +
-  "AND OccurredAt < {currentEnd:DateTime64(3)} + INTERVAL 2 DAY";
-const EVAL_TIME_FILTER_START_END =
-  "AND OccurredAt >= {startDate:DateTime64(3)} - INTERVAL 2 DAY " +
-  "AND OccurredAt < {endDate:DateTime64(3)} + INTERVAL 2 DAY";
 
 /**
  * Returns a deduped FROM-clause expression for trace_summaries.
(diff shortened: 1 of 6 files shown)


rogeriochaves:
The revert this describes landed as #6391 (`12ef6d1241`), which also brought the `join-time-bound-partition-column` guard along with it. The bound was then reinstated correctly by #6399. Closing as superseded.

PR #6390, opened 2026-07-31
fix(event-sourcing): coalesce the telemetry command lanes' event-log appends

0xdeafcafe:
## what

opts the five per-item telemetry commands into the ADR-066 pillar-2 append coalescing that `recordTriggerMatch` already ships:

- `log_processing` / `recordLogRecord` — one command per OTLP log record
- `metric_processing` / `recordDataPoint` — one command per OTLP data point
- `coding_agent_processing` / `contributeSpanFacts`, `contributeLogFacts`, `contributeMetricFacts` — one command per signal item into a session

each sets `coalesceMaxBatch: 256`; the GroupQueue's default 4MB byte budget bounds the batch on the size axis (log records can run to 1MB canonical).

## why

on the ops queues page these command lanes sit hours deep with zero retries — nothing is failing, the lanes just drain slower than they fill. without coalescing every queued item pays its own `event_log` insert plus its own claim round-trip, which is exactly the high-fan-in shape `specs/event-sourcing/producer-append-coalescing.feature` calls out ("one insert per item floods the event log with tiny parts"). the machinery has been in since ADR-066 (`processCommandBatch` → one multi-row insert per batch); these producers had just never opted in — and the registration-time visibility log only covers `serializeByAggregate` producers, so the gap stayed silent.

## safety

- all five handlers are stateless per item (each derives its single event from its own command alone) — the batch path's documented contract; none needs read-your-writes within a batch
- a batch is drawn from one group, so order within a lane/session is preserved and the failure domain is unchanged (a poison payload blocked its group before and still does)
- the existing idempotency keys make a batch retry neither duplicate nor drop events
- per-pipeline unit tests pin each adopter's opt-in, mirroring the automations test; the N-items-→-one-insert behavior itself is already proven at the GroupQueue layer

## deployment impact

no migration, no config. takes effect when workers roll; backed-up lanes should start draining in batches immediately.


0xdeafcafe:
Superseded — every opt-in in this PR is already on `main`:

- `log_processing` / `metric_processing` → #6462 (`perf(event-sourcing): coalesce log and metric record appends`, 831d298fe5)
- `coding_agent_processing` contributions → #6475 (`perf(event-sourcing): coalesce coding-agent session contributions`, df105109ec)

Both landed `coalesceMaxBatch: 256` on the same command lanes this PR targeted, and went further by also coalescing the map stage (`LOG_MAP_COALESCE_MAX_BATCH` / `METRIC_MAP_COALESCE_MAX_BATCH` / `CODING_AGENT_MAP_COALESCE_MAX_BATCH`), which this PR did not touch.

Nothing left to rebase — the branch also predates the repo restructure (`langwatch/` → `platform/app/`), so a rebase would be pure path churn onto an empty diff. Closing.

PR #6399, opened 2026-07-31
fix(analytics): bound the evaluation_runs JOIN so it prunes partitions; env-driven ClickHouse pool size

sergioestebance:
## For humans

Production has been degrading every business morning: the event-sourcing queue backs up, trace visibility lags by minutes, and the `/api/health/processor` monitor fires. Two compounding causes, both fixed here:

1. **Evaluation analytics queries were reading the tenant's entire evaluation history on every run.** The date filter they carried couldn't prune anything, because it filtered a column the table isn't partitioned by. One worker pod was measured spending 6,400 seconds of ClickHouse time per half hour on ~900 of these.
2. **Every process capped itself at 25 ClickHouse connections** — sized years before today's dispatch concurrency (256). Workers queued multi-second for queries the server answered in 55ms while the server sat at a third of its capacity.

## What changed

**1. `buildJoinClause` gains an `evalTimeFilter`** (mirroring the existing `spanTimeFilter` fix for `stored_spans`), applied to both the `evaluation_runs` JOIN subquery **and its IN-tuple dedup inner scan** — the inner scan is what actually walks the partitions. All call sites (both-periods regime, start/end regime, and the evaluator facet queries) pass lower bounds on **both** candidate partition columns:

- `ScheduledAt` — the partition column per the migrations
- `evaluation_runs.UpdatedAt` — the partition column on longer-lived deployments that predate that DDL (schema drift observed in production)

Lower bounds only, 7-day margin: an evaluation joined to an in-window trace is scheduled at/after the trace occurs, and every row version's `UpdatedAt >= ScheduledAt`, so the bound is a strict superset of correct results. No upper bound — re-evaluations update rows after the window and the dedup must still see the latest version. `UpdatedAt` is table-qualified because `trace_summaries` also has one and ClickHouse silently resolves a bare unknown identifier against the outer scope (the exact hazard the `join-time-bound-partition-column` guard test exists for — that guard now passes with the qualified form).

**2. `max_open_connections` is env-driven** via `CLICKHOUSE_MAX_OPEN_CONNECTIONS` (default 64, clamped to [1,1024]) from one shared helper (`src/server/clickhouse/connectionPool.ts`), replacing the hardcoded 25 in both construction sites. 64 keeps a 4-pod worker fleet within the server's `max_concurrent_queries` ceiling of 300.

## Evidence (production, 2026-07-31)

| | 05:00 (healthy) | 06:00 (degraded) |
|---|---|---|
| eval analytics executions | 167 | 6,188 |
| bytes read by that shape | 29 GiB | 1.19 TiB |
| ClickHouse SELECT p95 | 58ms | 2,016ms |
| queue completed/s | 198 | 133–169 |

Server-side HTTP connections sat at exactly 100 = 4 pods × 25 while workers self-measured 5,747ms for scans ClickHouse served in 55ms — pure client-side pool queueing.

## Tests

- `field-mappings.test.ts`: bound applied to both outer subquery and dedup inner; absent when not passed; ignored for other tables
- `aggregation-builder.test.ts`: generated SQL carries both partition-column bounds, dedup inner included
- `connectionPool.unit.test.ts`: default / valid / invalid env parsing
- Existing guard `join-time-bound-partition-column.unit.test.ts` passes (it rejected the unqualified first draft — working as designed)
- Integration suites (`filter-evaluation-queries`, `cross-evaluator-groupby`, `evaluation-passed-score-only`) pass against real ClickHouse — the new bounds execute inside the IN-dedup without the correlated-subquery failure
- `pnpm typecheck` + `pnpm typecheck:tests` clean

## Not in this PR (follow-ups from the same incident)

- Workload isolation: separate CH client/user for analytics (`CLICKHOUSE_ANALYTICS_URL`)
- Append-coalescing for rollup pipelines (6k executions/hr is re-computation, not load)
- HPA scales workers on CPU, which drops during I/O-bound backlogs — scale on queue depth instead
- Worker liveness probe kills on downstream slowness (restart storms drop in-flight work)
- Schema drift: production `evaluation_runs` partitions by `UpdatedAt`, migrations say `ScheduledAt`

## Summary by CodeRabbit

* **Performance**
  * Improved analytics query performance with more precise evaluation-run time boundaries and partition pruning.
  * Added configurable ClickHouse connection pooling with safe defaults, validation, and support for environment-based sizing.

* **Bug Fixes**
  * Improved consistency when applying evaluation-run time filters across analytics queries.
  * Preserved results for records without scheduled times.

* **Tests**
  * Added coverage for analytics time filtering, query optimization, and connection-pool configuration.

### DIFF
--- a/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
+++ b/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
@@ -102,11 +102,20 @@ const SPAN_TIME_FILTER_START_END =
 // pins the reference to evaluation_runs' own column so neither ClickHouse nor
 // the guard has to guess. ScheduledAt stays bare: it is evaluation_runs' own
 // partition column per the migrations and does not exist on trace_summaries.
+//
+// The ScheduledAt bound is NULL-safe. On the unified schema (00002) the column
+// is `DateTime64(3) DEFAULT now64(3)` and the IS NULL branch is statically
+// false, so partition pruning is unaffected. But long-lived deployments that
+// predate the unified DDL carry `ScheduledAt Nullable(DateTime64(3))`, where a
+// bare `ScheduledAt >= x` evaluates to NULL for NULL rows and silently DROPS
+// those evaluations from every graph — a correctness regression, not a missed
+// optimisation. NULL rows on such deployments are still bounded by the
+// UpdatedAt predicate, which is their actual partition column anyway.
 const EVAL_TIME_FILTER_BOTH_PERIODS =
-  "AND ScheduledAt >= {previousStart:DateTime64(3)} - INTERVAL 7 DAY " +
+  "AND (ScheduledAt IS NULL OR ScheduledAt >= {previousStart:DateTime64(3)} - INTERVAL 7 DAY) " +
   "AND evaluation_runs.UpdatedAt >= {previousStart:DateTime64(3)} - INTERVAL 7 DAY";
 const EVAL_TIME_FILTER_START_END =
-  "AND ScheduledAt >= {startDate:DateTime64(3)} - INTERVAL 7 DAY " +
+  "AND (ScheduledAt IS NULL OR ScheduledAt >= {startDate:DateTime64(3)} - INTERVAL 7 DAY) " +
   "AND evaluation_runs.UpdatedAt >= {startDate:DateTime64(3)} - INTERVAL 7 DAY";
 
 /**
--- a/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
+++ b/langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
@@ -79,6 +79,36 @@ const SPAN_TIME_FILTER_START_END =
   "AND StartTime >= {startDate:DateTime64(3)} - INTERVAL 2 DAY " +
   "AND StartTime < {endDate:DateTime64(3)} + INTERVAL 2 DAY";
 
+// Partition-pruning bounds for the evaluation_runs JOIN subquery. The query
+// windows on trace OccurredAt, but evaluation_runs is partitioned by
+// ScheduledAt per the migrations (and by UpdatedAt on long-lived deployments
+// that predate that DDL), so an OccurredAt filter prunes nothing there —
+// without these bounds the JOIN's dedup subquery walks the tenant's entire
+// history across every weekly partition, including the S3-tiered cold ones.
+//
+// Lower bounds only, on BOTH candidate partition columns, so pruning works on
+// either partitioning scheme. Safe as a superset: an evaluation joined to an
+// in-window trace is scheduled at/after that trace occurs, and every row
+// version's UpdatedAt >= its ScheduledAt, so both columns are >= the window
+// start minus scheduling skew — far inside the 7-day margin (partitions are
+// weekly, so the margin costs at most one extra partition). No upper bound:
+// a re-evaluation updates rows long after the window, and the IN-tuple dedup
+// must still see that latest version.
+//
+// UpdatedAt is table-qualified. trace_summaries (the outer scope) also has an
+// UpdatedAt column, and ClickHouse resolves a bare identifier the inner table
+// lacks against the OUTER scope instead of failing — the hazard
+// `join-time-bound-partition-column.unit.test.ts` guards against. Qualifying
+// pins the reference to evaluation_runs' own column so neither ClickHouse nor
+// the guard has to guess. ScheduledAt stays bare: it is evaluation_runs' own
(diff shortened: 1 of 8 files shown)


langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
```diff
@@ -79,6 +79,36 @@ const SPAN_TIME_FILTER_START_END =
...
+//
+// UpdatedAt is table-qualified. trace_summaries (the outer scope) also has an
+// UpdatedAt column, and ClickHouse resolves a bare identifier the inner table
+// lacks against the OUTER scope instead of failing — the hazard
+// `join-time-bound-partition-column.unit.test.ts` guards against. Qualifying
+// pins the reference to evaluation_runs' own column so neither ClickHouse nor
+// the guard has to guess. ScheduledAt stays bare: it is evaluation_runs' own
+// partition column per the migrations and does not exist on trace_summaries.
+const EVAL_TIME_FILTER_BOTH_PERIODS =
+  "AND ScheduledAt >= {previousStart:DateTime64(3)} - INTERVAL 7 DAY " +
```
langwatch-agent: **[P1] Do not require `ScheduledAt` on the legacy `UpdatedAt`-partitioned table**

The older schema this PR is trying to support declares `ScheduledAt Nullable(DateTime64(3))` and partitions by `toYearWeek(UpdatedAt)` (the pre-unification `00019_create_evaluation_runs.sql`). This conjunction makes every legacy row with `ScheduledAt = NULL` fail SQL's WHERE predicate, even when its `TraceId` is in the requested range and its `UpdatedAt` is within the intended bound. Those evaluations disappear from pass-rate graphs and evaluation filters rather than merely missing partition pruning.

The stated `UpdatedAt >= ScheduledAt` invariant does not cover NULL. The new tests construct only non-null `ScheduledAt` values, so they cannot detect this regression. Choose the predicate from the table's actual partition schema (or otherwise explicitly retain the nullable legacy rows), and add a real-ClickHouse regression case for an `UpdatedAt`-partitioned legacy row with `ScheduledAt = NULL` joined to an in-window trace. A single static `ScheduledAt AND UpdatedAt` bound is not safe across the two schemas.

Confidence: high.

langwatch-agent (dismissed):
1 P1 — the proposed dual-schema bound can silently exclude legacy evaluation rows. See the inline finding.

langwatch/src/server/analytics/clickhouse/aggregation-builder.ts
sergioestebance: Valid catch on substance — fixed in a3eb92885. Verified against production directly: `ScheduledAt` is indeed `Nullable(DateTime64(3))` there (the drifted pre-unification schema, partitioned by `UpdatedAt`). The bound is now `(ScheduledAt IS NULL OR ScheduledAt >= x)`; on the unified schema the column is NOT NULL so the IS NULL branch is statically false and pruning is unchanged, and NULL rows on legacy schemas remain bounded by the qualified `UpdatedAt` predicate — their actual partition column.

Two corrections to the finding: the cited migration `00019_create_evaluation_runs.sql` doesn't exist (00019 is `add_trace_events`), and measured impact today is zero — 0 NULL `ScheduledAt` across all 20.3M production rows. The fix is warranted anyway: the type permits NULLs and self-hosted installs can't be audited.

On the requested real-ClickHouse regression case with a legacy `UpdatedAt`-partitioned table: the integration harness provisions the unified schema, where `ScheduledAt` is NOT NULL and a NULL row can't be inserted. A legacy-schema harness is out of scope here; the generated-SQL unit test pins the NULL-safe form instead.

langwatch/src/server/clickhouse/connectionPool.ts
```diff
@@ -0,0 +1,46 @@
...
+ * (GLOBAL_QUEUE_CONCURRENCY=256 in production) and became the system
+ * bottleneck: with 4 worker pods the server showed exactly 4 x 25 = 100 open
+ * connections pinned at cap while sitting at a third of its
+ * `max_concurrent_queries` (300) and answering in ~55ms — workers measured
+ * multi-second latency for those same queries, all of it client-side queueing
+ * for a pooled socket.
+ *
+ * 64 keeps a 4-pod worker fleet (256 potential concurrent queries) within the
+ * server's 300-query ceiling. Deployments with more replicas or a different
+ * server budget tune CLICKHOUSE_MAX_OPEN_CONNECTIONS instead of editing code.
```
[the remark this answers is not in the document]
sergioestebance: Fixed in a3eb92885 — the comment now states the pool is per client instance, names both construction sites, and notes the server budget must sum every pool across every pod.

PR #6403, opened 2026-07-31
fix(api-keys): let the permission ceiling see legacy team membership

Aryansharma28:
## What & why

Split out of #6110, where these two commits were riding along inside a 12,000-line leaderboard UI PR. They change **who is allowed to mint an API key**, which is not something anyone reviewing "add a Bradley-Terry chart" is going to give the attention it deserves.

No code changes from the version in #6110 — a clean cherry-pick onto `main`, zero conflicts.

## The bug

`enforcePermissionCeiling` asked `checkRoleBindingPermission` whether the minting user holds each permission they are trying to grant. That resolver reports **RoleBindings and nothing else**, deliberately — its own suite pins that behaviour. The tRPC authorization path layers the legacy `TeamUser` fallback on top of it; this ceiling never did.

So a user whose access comes from legacy team membership:
- passed every authorization check,
- was measured by `batchProjectPermissions` as holding the permissions,
- and was then refused those same permissions here as *"exceeds your own access"*.

Not theoretical. Langy mints a per-turn key mirroring the caller's permissions, so on any workspace still on legacy membership **every turn died** with an opaque `api_key_scope_violation` and no route to a fix.

## The fix

`resolveLegacyCeilingRole` resolves the ceiling user's legacy role **once per request**, and the ceiling accepts a permission if either the RoleBinding path or the legacy role grants it.

Three guards, all deliberate:

- **Only a user with no RoleBindings anywhere in the org falls back** — the same gate `loadScopeResolution` uses — so a migrated user's stale legacy row can never stack on top of their real binding.
- **Org-scoped keys are excluded.** An org-wide grant is not something a team role should be able to reach.
- **ADR-021 scope guard applies** (`bindingScopeCanGrant`): a team/project-scoped grant can never confer an org-exclusive permission. `rbac.ts` already applies this to legacy roles via `bindingGrants`; without it here the ceiling would accept a permission the tRPC path refuses. No team role lists an org-exclusive resource today, so this is parity and defence-in-depth rather than a live hole — it stops being either the moment someone adds one.

Resolved once rather than per-permission on purpose: the Langy mint path passes ~23 permissions inside a 5-second interactive transaction, and `langyApiKey.ts` already records what per-permission fan-out did to it — connection-pool starvation and an aborted transaction. Two queries, flat.

## Tests

`api-key.service.legacy-ceiling.unit.test.ts` (277 lines) covers the fallback, the no-stacking guard, org-scope exclusion, the ADR-021 guard, and current-membership fail-closed. **129 tests pass** across `src/server/api-key/__tests__/` on this branch alone.

## Merge order

Independent of #6110 — no leaderboard file references any of this. If #6110 merges first the changes arrive with it and this PR closes as already-applied; either order is safe.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API-key permission validation for users with legacy team memberships.
  * Preserved permission scope restrictions for organization-level access.
  * Prevented legacy permissions from applying after users migrate to role-based bindings.
  * Ensured users without valid role assignments cannot receive unauthorized permissions.

### DIFF
--- a/langwatch/src/server/api-key/__tests__/api-key.service.legacy-ceiling.unit.test.ts
+++ b/langwatch/src/server/api-key/__tests__/api-key.service.legacy-ceiling.unit.test.ts
@@ -1,0 +1,194 @@
+import { TeamUserRole } from "@prisma/client";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * The legacy-membership half of the API-key ceiling.
+ *
+ * `checkRoleBindingPermission` reports RoleBindings and nothing else, on
+ * purpose — its own suite pins that ("TeamUser fallback is handled by
+ * checkPermissionFromBindings, not this resolver"). The tRPC authorization
+ * path layers the legacy TeamUser fallback on top of it. This ceiling never
+ * did, and the asymmetry was reachable: a user whose access came from legacy
+ * membership passed every authorization check, was measured by
+ * `batchProjectPermissions` as holding permissions, then was refused those
+ * same permissions here as "exceeds your own access".
+ *
+ * Langy mints a per-turn key mirroring the caller's permissions, so on a
+ * workspace still on legacy membership every turn died with an opaque
+ * `api_key_scope_violation`.
+ *
+ * @see specs/api-keys/scope-based-permissions.feature
+ */
+
+const ORG_ID = "org1";
+const USER_ID = "user1";
+const TEAM_ID = "team1";
+const PROJECT_ID = "proj1";
+
+// No RoleBindings anywhere — the exact condition the fallback exists for.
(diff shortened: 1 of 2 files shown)


Aryansharma28:
Closing — consolidating back into #6110, which already carries these commits (`8e78d74`, `78d47d1`) and their test. This PR was a cherry-pick copy, not a move, so nothing is lost by closing it: the api-key ceiling fix ships with #6110.

Flagging the tradeoff for whoever reviews #6110, since it is the reason this was split in the first place: these two commits change **who may mint an API key**, and they now sit inside a ~12,000-line leaderboard diff. Worth a focused look at `langwatch/src/server/api-key/api-key.service.ts` on its own terms rather than as part of the UI review.

PR #6404, opened 2026-07-31
fix(cli): surface Comparison verdicts in experiment results and --filter failed

Aryansharma28:
## What & why

Split out of #6110. These two CLI fixes are Comparison-adjacent, which is how they ended up riding along with the leaderboard work, but they live in a different package (`typescript-sdk`) with their own spec file and their own test suite — a separate surface deserving its own review.

Clean cherry-pick onto `main`, zero conflicts, no code changes from the version in #6110.

## The two fixes

**1. Comparison verdicts were dropped from experiment results.** The results reader only understood evaluators that emit a numeric score. A Comparison emits a verdict — which variant won — so `langwatch experiment results` simply omitted them, and a run whose whole point was a pairwise comparison printed as though it had produced nothing.

**2. `--filter failed` could not see a failed comparison.** Same root cause one layer up: the filter predicate keyed on the numeric-score shape, so a Comparison that failed was invisible to the one flag whose entire job is surfacing failures. Worse than a missing row — the user asked specifically for failures and was told there were none.

## Tests

`results.unit.test.ts` covers both paths; **17 tests pass** on this branch alone. Scenarios are bound in `specs/typescript-sdk/cli-experiment-results.feature`.

## Merge order

Independent of #6110 — no leaderboard file touches the CLI. If #6110 merges first these arrive with it and this PR closes as already-applied; either order is safe.

## Summary by CodeRabbit

* **New Features**
  * Experiment results now display comparison verdicts in CLI output.
  * Added support for filtering results by row limits and evaluators.
  * Comparison-only failures are included when filtering for failed rows.
  * Evaluations remain correctly associated with displayed rows, including when they are independent of specific dataset rows.

### DIFF
--- a/specs/typescript-sdk/cli-experiment-results.feature
+++ b/specs/typescript-sdk/cli-experiment-results.feature
@@ -45,6 +45,37 @@ Feature: CLI evaluation results command
     And the JSON output is valid and parseable
     And the CLI exits with status 0
 
+  # A Comparison evaluator judges the whole field of targets in one verdict, so
+  # its results are recorded against the comparison itself rather than against
+  # any one target. The command builds its output by walking the dataset — one
+  # entry per (row, target) — so anything not tied to a target had no row to
+  # attach to and never reached stdout. The run summary still reported the
+  # comparison, which made the omission read as "the judge produced nothing"
+  # rather than "the CLI dropped it".
+
+  @integration @unimplemented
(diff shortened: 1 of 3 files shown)


Aryansharma28:
Closing — consolidating back into #6110, which already carries these commits (`b5b1bc1`, `8c73f23`), their test, and the spec file. This PR was a cherry-pick copy, not a move, so the CLI fixes ship with #6110 either way.

PR #6413, opened 2026-07-31
feat(lwql): read-only, tenant-scoped SQL-like query service and API over traces and spans

### LINKED ISSUE #6346: LangWatch Query Layer: a read-only, tenant-scoped query language, service and API

drewdrewthis:
Implements the LangWatch Query Layer specified in #6346. Design rationale is ADR-081, included in this PR (`dev/docs/adr/081-lwql-query-language.md`) — #6368 was consolidated here.

A read-only, tenant-scoped SQL-like query language over traces and spans, exposed as one service behind both REST and tRPC.

```sql
SELECT model, avg(cost_usd), p95(duration_ms)
FROM traces
WHERE has_error = true AND started_at >= now() - INTERVAL 24 HOUR
GROUP BY model
ORDER BY avg(cost_usd) DESC
LIMIT 100
```

## What this is not

*Looks like* SQL is a syntax decision, not a semantics one. There are no joins, no subqueries, no `UNION`, no `SELECT *`, and no table names — `FROM` takes a closed enum of entities the compiler maps to real tables. The moment raw SQL reaches ClickHouse the tenant guarantee is gone and no amount of validation recovers it.

## Structure

| File | Role |
| --- | --- |
| `catalog.ts` | Closed enums mapping entities/fields/aggregates to developer-authored ClickHouse expressions |
| `ir.ts` | The security boundary both input forms compile into |
| `parser.ts` | SQL-subset text front-end; can only emit IR |
| `compiler.ts` | IR to SQL; injects tenant scope, binds every value |
| `gating.ts` | Content-visibility gating for filter and aggregation targets |
| `lwql.service.ts` | Single entry point for both transports |

## The three invariants

1. **Every identifier comes from the catalogue.** A field name *selects* a pre-written expression; it never *becomes* one. Nothing caller-derived is concatenated into identifier position.
2. **Every value is bound as a query parameter.**
3. **Tenant scope is injected by the compiler** from the RBAC-checked `projectId`, and ANDed at the top level — so no `OR` inside a caller's predicate can widen it. `project_id` is queryable, but filtering it can only narrow.

## Content gating — the part worth reviewing closely

ADR-028 gates content for out-of-window callers while leaving aggregates visible. That was safe when aggregates were fixed dashboard charts. It stops being safe once the filter is caller-supplied:

```sql
SELECT count(*) FROM traces WHERE input LIKE '%acme corp%'
```

That projects no content, satisfies a projection-only rule, and still reads the gated value one bit at a time. So gated fields are **refused** as filter and aggregation targets, not just masked on output — refused rather than teased, because a predicate is evaluated against the stored value regardless of what comes back.

The gated set is **not restated** in the query layer. `gating-parity.unit.test.ts` probes the real `redactTraceContent` / `redactSpanContent` with sentinel content, observes which fields actually change, and asserts LWQL accounts for every one. Adding a field to the redaction service without teaching LWQL about it fails CI. The test also asserts the reverse direction, so a field cannot be gated for no reason.

## Acceptance criteria

| AC | Where |
| --- | --- |
| Query traces and spans with SQL-like syntax | `parser.ts`, `compiler.ts` |
| API consumers execute the same surface without the UI | `app/api/query` |
| Queries scoped to the authenticated tenant | `compiler.ts` + 4 tests |
| A user cannot override tenant isolation | 4 tests incl. the `OR`-escape case |
| Out-of-allowlist names fail to compile, not sanitised | 6 tests |
| Gated field unusable as filter/aggregation target, parity enforced by test | `gating.ts` + 10 tests |
| Result metadata reports row count, execution time, truncation | `lwql.service.ts` |

Truncation is *observed*, not guessed: the compiler asks for `limit + 1`.

## Verification — and the bug that only running it could find

**38 unit + 27 integration, all passing.** The integration suite runs against a real ClickHouse testcontainer.

Two runtime defects were found this way, neither visible to a passing unit test:

Writing it was not a formality. On its first run it found that **every LWQL query failed at runtime**:

```
Value 2026-07-24T14:36:09.353Z cannot be parsed as DateTime64(3)
for query parameter 'p1' ... only 23 of 24 bytes was parsed
```

ClickHouse rejects ISO-8601's trailing `Z`. Every query carries a time bound, so nothing worked — while all 33 unit tests stayed green, because asserting on a generated SQL *string* cannot reveal what the database does with it. Fixed to bind `YYYY-MM-DD HH:MM:SS.mmm`, with a unit regression guard that I verified is falsifiable (reverting the fix fails it: 2 failed / 29 passed).

**Second**, reviewing my own diff I suspected `ORDER BY` was unguarded and probed it against the database rather than reasoning about it. `SELECT model, count(*) FROM traces GROUP BY model ORDER BY duration_ms` compiled cleanly and ClickHouse rejected it with code 215, leaking an internal column name (`TotalDurationMs`) to the caller. The compiler enforced GROUP BY completeness for `SELECT` but never for `ORDER BY`. Fixed, with the valid shapes (ordering by a grouping key, by an aggregate alias) asserted *executed* so the fix cannot be a blanket refusal.

What the integration suite establishes that string assertions cannot:

| Claim | How it is proven |
| --- | --- |
| **Tenant isolation** | Two tenants' rows sit in the same tables. The other tenant's cost is `999` and duration `5000ms` — if scoping failed, the aggregates would say so. A leak surfaces as a *number*, not an error. |
| Model-less traces bucket as `unknown` | Asserted on returned rows — the exact `arrayJoin` divergence ADR-081 records between the spike and the analytics builder |
| Array filtering doesn't multiply rows | `has(Models, …)` returns 3, not an inflated count |
| `p95` is a real quantile | Executed |
| `tool_name` reads the right attribute | The key was picked by grep frequency; if wrong it would return nulls forever rather than fail. Now asserted on values. |
| Truncation detection | The `limit + 1` probe, executed |
| Gated content refused | Rejected before reaching the database |

### The endpoint, over the wire

The service-level suite proves the compiler. It does not prove the *route* — request validation, the project→tenant handoff, error-to-HTTP mapping, and the response envelope all live in the handler, and a handler can be wrong in ways the service is right. So `src/app/api/query/[[...route]]/__tests__` drives real HTTP requests through the real Hono app against the real container; only auth (a passthrough injecting the project, as in the sibling traces suites) and the plan store are stubbed.

It covers: text and structured-IR forms returning identical rows; scoping following the authenticated project, proven by a second tenant whose row costs `999` and never reaches the aggregate; `explain` never leaking SQL to an API-key caller; bad queries mapping to `400` with a code and hint; `/validate` returning `200 valid:false` rather than a transport error so editors can render inline errors; and `/catalogue` leaking no table names.

**Verified falsifiable:** replacing the tenant predicate with `1 = 1` fails 3 of these tests, including the scoping assertion.

`tsc --noEmit` across the repo: **zero errors in LWQL files**. Worth knowing for anyone re-running it — with default settings tsc aborts on a `baseUrl` deprecation before checking a single file, and then OOMs; it needs `--ignoreDeprecations 6.0` and `--max-old-space-size`. My first two "clean" runs were vacuous.

## Open question for the reviewer — a deviation I did not silently keep

Issue #6346 decision 6 says: *"LWQL emits IR and hands it to the existing analytics destination router... LWQL names no tables."*

**This PR names tables.** `catalog.ts` maps `traces → trace_summaries` and `spans → stored_spans`, and the compiler emits `FROM <table>` directly.

My reasoning: the ADR-034 destination router routes *timeseries metric* shapes — it takes a series/groupBy input and picks a rollup. LWQL's shape is arbitrary field projection, which that router has no input form for, and `trace_summaries` is precisely what `aggregation-builder` calls "the safe fallback" for shapes it cannot route. So this lands where the router would have landed anyway.

That is an argument, not a ratification. If the intent was that LWQL wait for a routing entry point, this needs restructuring before merge — which is exactly why it is flagged here rather than buried in a diff.

## Deliberately deferred

- **Async execution for expensive queries** and **pagination beyond OFFSET** — the issue lists both; sync + caps ship first per ADR-081 decision 7.
- **`evaluations` and `sessions` entities** — decision 8. `catalog.ts` is shaped to extend without touching parser or compiler.
- **No run through a fully booted Next.js app.** The route is exercised over real HTTP against real ClickHouse (below), with only auth and the plan store stubbed; what is not covered is real API-key authentication and the Next.js server itself.

Closes #6346

## Deployment Impact

**Additive only. Nothing existing changes behaviour.**

- **New endpoints:** `POST /api/query`, `POST /api/query/validate`, `GET /api/query/catalogue`, plus the `api.lwql.*` tRPC procedures. All are new paths; no existing route is modified.
- **Schema / migrations:** none. LWQL is read-only and adds no tables, columns, or ClickHouse migrations — it reads `trace_summaries` and `stored_spans` as they already exist.
- **Environment variables:** none added, removed, or renamed.
- **Permissions:** every route requires the existing `traces:view` permission. No new permission is introduced.
- **Load:** queries are caller-driven, so guardrails ship with them — a 30s `max_execution_time`, `readonly: 2`, a 10,000-row cap, and a 90-day maximum window with a 7-day default when the caller supplies none.
- **Rollback:** revert the commit. Nothing persists, no migration to unwind, and no existing surface degrades in the absence of these routes.


drewdrewthis:
Closing: the driving issue #6346 was closed 2026-08-03T16:30:38Z, and the shipped answer for tenant-scoped read-only SQL is the governed SQL stack merged in https://github.com/langwatch/langwatch/pull/7014 (2026-08-15T14:31:55Z). This branch has been a conflicting draft since 2026-08-13. Branch issue6346/lwql-service is retained on origin at 6ddde07aa177f677c72bd77b34d4368b210eb585 if the ADR-081 material is wanted later.

PR #6422, opened 2026-07-31
fix: surface structured AI SDK output in trace spans

### LINKED ISSUE #6339: Empty Output panel on ai.generateText spans when using Vercel AI SDK structured output (Output.object)
**Describe the bug**
When using Vercel AI SDK generateText with structured output (Output.object({ schema })), LangWatch traces show an empty Output panel on the ai.generateText LLM span — even when generation succeeds and the parsed object is available in result.output.

Span attributes may include ai.response.object with the structured result (or raw JSON text), but this is not surfaced in the Output UI. LangWatch appears to derive span output from ai.response.text / langwatch.output, which structured-output calls don't populate the same way as plain text generation.

This blocks creating test sets from production traces: LangWatch needs span output to populate dataset records, but structured-output calls appear to have no output on the LLM observation span (the one you'd filter on for test-set creation). Parent/custom spans wrapped with traceable may have output set manually, but the LLM span stays empty.

**To reproduce**
Steps to reproduce the behaviour:

Instrument the app with LangWatch observability (AI SDK auto-instrumentation via OpenTelemetry: LangWatch span processor + LangWatchTraceExporter).
Call generateText with structured output:
```
import { generateText, Output } from "ai";
import { z } from "zod";
const result = await generateText({
  model: anthropic("claude-sonnet-4-6"),
  prompt: "Translate 'Hello' to Dutch as JSON: { \"greeting\": \"...\" }",
  output: Output.object({
    schema: z.object({ greeting: z.string() }),
  }),
});
// result.output is correctly populated: { greeting: "Hallo" }
```
Open the trace in LangWatch and inspect the ai.generateText span.
Observe that the span Output field is empty (or not useful), while ai.response.object may be present in raw span attributes.

**Expected behavior**

The ai.generateText span Output panel should show the structured result (e.g. { "greeting": "Hallo" }) when generateText succeeds with Output.object(), so that:

Test sets can be built from real production traces (translation, extraction, classification).
Prompt/schema regressions can be debugged from the trace UI without digging into raw attributes.
Consumers don't need custom wrappers that manually call setOutput() before the span closes.

**Environment**

Component: langwatch_sdk_typescript
Version: langwatch ^1.0.0 (langwatch/observability)
Vercel AI SDK: ai ^6.0.0
Instrumentation: OpenTelemetry via LangWatch span processor + LangWatchTraceExporter
Models: Anthropic Claude (direct + Bedrock fallback)
Browser/OS: N/A (server-side Node/Bun)

**Additional context**

Structured output works correctly at the application level (result.output is populated); only LangWatch trace display/export is affected.
Without LLM span output on structured calls, filtering on the LLM observation span for test-set creation fails.
Possible fix direction: map ai.response.object (or equivalent structured-output attributes) into the span output LangWatch surfaces in the UI, similar to how ai.response.text is handled for plain text generation.

rogeriochaves:
## Summary

- Surface structured AI SDK output when response data arrives as an object or JSON text.
- Keep JSON-looking assistant content as text so trace summaries and Output panels remain populated.
- Add a BDD scenario, focused unit coverage, and a real storage-backed integration test.

## Validation

- Focused Vitest suite: 85 tests passed.
- Real storage-backed integration test: 1 test passed.
- `pnpm run typecheck` passed.
- `pnpm run typecheck:tests` passed.
- Biome check passed with existing complexity warnings.

## Real dogfood QA

- A temporary Vercel AI SDK micro-app made a live `generateText` call with `Output.object`, `@ai-sdk/openai`, and the local `OPENAI_API_KEY`.
- LangWatch telemetry was exported to the local server, persisted in ClickHouse, and verified in the trace list and drawer.
- The trace showed `ai.generateText` output as `{ "greeting": "Hallo" }`.

Real AI SDK structured output in trace drawer

Fixes #6339

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved extraction of structured AI SDK responses, including cases where response text is empty.
  * Structured outputs are now consistently serialized as readable JSON assistant messages.
  * Preserved JSON-looking chat content without incorrectly transforming it.
  * Added fallback handling for alternate response formats and empty values.

* **Tests**
  * Expanded unit, integration, and behavior coverage for structured AI outputs and Vercel AI SDK traces.

### DIFF
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
@@ -59,8 +59,12 @@ function buildSpan(
     endTimeUnixNano: (startTimeUnixNano + 100_000_000n).toString(),
     attributes: [
       {
-        key: "ai.response.text",
+        key: "ai.response.object",
         value: { stringValue: response },
+      },
+      {
+        key: "ai.response.text",
+        value: { stringValue: "" },
       },
       {
         key: "ai.prompt",
@@ -109,7 +113,7 @@ describe.skipIf(!hasTestcontainers)(
       await cleanupTestDataForTenant(tenantIdString);
     });
 
-    it("persists structured output on the LLM span and trace summary", async () => {
+    it("persists flat structured output when text is empty", async () => {
       const currentTenantIdString = tenantIdString;
       if (!currentTenantIdString) {
         throw new Error("Test tenant was not initialized.");
--- a/langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
+++ b/langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
@@ -84,7 +84,7 @@ describe.skipIf(!hasTestcontainers)(
   "AI SDK structured output through trace processing",
   () => {
     let tenantId: ReturnType<typeof createTestTenantId>;
-    let tenantIdString: string;
+    let tenantIdString: string | undefined;
     let traceSummaryStore: TraceSummaryStore;
     let spanStorageService: SpanStorageService;
 
@@ -105,10 +105,16 @@ describe.skipIf(!hasTestcontainers)(
     });
 
     afterEach(async () => {
+      if (!tenantIdString) return;
       await cleanupTestDataForTenant(tenantIdString);
     });
 
     it("persists structured output on the LLM span and trace summary", async () => {
+      const currentTenantIdString = tenantIdString;
+      if (!currentTenantIdString) {
+        throw new Error("Test tenant was not initialized.");
+      }
+
       const traceId = `trace-ai-sdk-${Date.now()}`;
       const spanId = `span-ai-sdk-${Date.now()}`;
       const occurredAt = Date.now();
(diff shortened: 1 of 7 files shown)


langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    traceId,
+    spanId,
+    parentSpanId: null,
+    name: "ai.generateText",
+    kind: 1,
+    startTimeUnixNano: startTimeUnixNano.toString(),
+    endTimeUnixNano: (startTimeUnixNano + 100_000_000n).toString(),
+    attributes: [
+      {
+        key: "ai.response.text",
```
langwatch-agent: **P2 - The storage-backed test does not exercise the reported attribute.**

The issue and new extractor gate are about flat `ai.response.object`, but this end-to-end test emits only `ai.response.text`. That path was already recognized by the extractor; this test mainly proves the separate JSON-string-preservation change. The unit tests cover the object key in isolation, but they do not prove its OTLP ingestion, canonicalization, fold, and persistence path together. Please change this fixture to `ai.response.object` or add a second integration case for it, including the empty-text-plus-object fallback introduced in the follow-up commit.

langwatch-agent (commented):
Reviewed the full diff and affected ingestion, canonicalization, trace-summary, storage, and content-drop paths. One non-blocking P2 test gap is inline; I found no production blocker.

langwatch/src/server/event-sourcing/pipelines/trace-processing/__tests__/aiSdkStructuredOutput.integration.test.ts
rogeriochaves: Addressed in commit 0ae2081b4c. The storage-backed fixture now emits ai.response.object together with an empty ai.response.text and asserts the persisted LLM output and trace summary. The focused integration and extractor tests pass.

PR #6430, opened 2026-08-01
fix(event-sourcing): anchor trace summaries safely

### LINKED ISSUE #6312: trace_summaries has the same epoch-partition anchor defect as trace_analytics (ADR-071 step 3)
Found while reviewing #6306, which fixes this defect for `trace_analytics` only. `trace_summaries` has it too, was **not** in ADR-071's "Checked, not assumed" inventory (now corrected by #6306), and is carrying more of the bad rows.

## Same defect, same source

`traceSummary.foldProjection.ts` folds its `occurredAt` through the **same shared** `SpanTimingService.accumulateTiming` as the analytics fold, with the same `0` sentinel and the same span-seeded-only rule. Its store admits log-only traces through a `hasPersistableSignal` predicate identical to the analytics one — whose docblock names the culprits outright: *"logs-only traces (claude Path B + OTEL_LOGS_EXPORTER without a traces exporter, codex Path B pre-codex-spans, custom gen_ai-on-logs emitters)"*.

| | `trace_analytics` | `trace_summaries` |
|---|---|---|
| `occurredAt` = min span start, `0` sentinel | yes | yes — same shared service |
| Log-only traces admitted | yes | yes — identical gate |
| `PARTITION BY toYearWeek(OccurredAt)` | yes | yes (`00002_create_schema.sql`) |
| TTL anchored on `OccurredAt` | yes | yes (`ttlReconciler.ts:149-151`) |
| Reads its own row back | yes | yes (`store.get` → `repo.findByTraceId`, ±2-day window) |

So a log-only trace commits at `OccurredAt = 0` → partition `196952`, TTL deadline `1970 + retention` (already past) → reaped on the next TTL merge → `get()` misses thereafter.

## Two differences that change the fix

**Milder:** its sort key is `(TenantId, TraceId)` and does *not* include `OccurredAt`, so the ReplacingMergeTree collapses a trace's versions regardless of the anchor moving. ADR-071 consequence 1 (orphaned versions that can never collapse) does not apply, and the sort-key argument that shapes the `trace_analytics` fix is unnecessary here.

**Worse:** it does not set `refoldOnStoreMiss`, which defaults to `false`. When the epoch row is reaped there is no rebuild net — `get()` misses, the fold restarts from `init()`, and the row is rewritten from post-reap events only. The failure mode is **silent loss** of the trace's accumulated cost, tokens and span count, not a refold storm.

## Scope

- Give `trace_summaries` a storage anchor and split the span timing baseline into its own column (a second migration, same shape as `00060`).
- Decide the refold-net question: whether `refoldOnStoreMiss` should be on for this fold, or whether losing a reaped trace's totals is acceptable.
- Reuse the pre-split-decode transition from #6306 rather than a version bump that forces a population rebuild.

Blocked on nothing; deliberately deferred out of #6306 to keep that change to one table.

Refs: ADR-071 (`dev/docs/adr/071-coding-agent-session-immutable-storage-anchor.md`), #6306.

shivanshgupta365:
## Why

Log-only trace summaries currently persist OccurredAt as the epoch because the shared timing baseline is only seeded by spans. That places the row in partition 196952 with an already-expired TTL; once reaped, the fold restarts from empty state and silently loses accumulated totals.

Closes #6312

## What changed

- freeze a separate storage and TTL anchor on the first usable contribution time
- persist the span timing baseline in a new EarliestSpanStartMs column
- decode pre-split rows from their existing OccurredAt without a population refold
- enable store-miss refolding so TTL-reaped summaries rebuild instead of losing totals
- update ADR-071 and add regression coverage for log-only, late-span, transition, persistence, and store-miss cases

## How it works

A span-led trace anchors on the first folded span start; a non-span-led trace anchors on the event business time. Later earlier-starting spans may still move the duration baseline but cannot move the partition or TTL anchor. Existing 2026-05-07 rows treat OccurredAt as both values on read and heal on their next ordinary write.

## Test plan

- 18 focused unit tests pass across storage anchoring, repository read/write transition, store read-back, and refold policy
- TypeScript test-project typecheck passes
- Biome check passes with only existing complexity warnings in touched legacy functions
- git diff --check passes

## Anything surprising?

refoldOnStoreMiss is intentionally enabled and trustAbsentMiss is removed. The store has a persistability gate, so an absent row cannot prove that no committed state existed; accepting the miss would preserve the silent-loss failure mode.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trace summary accuracy by separating storage timing from span timing.
  - Preserved historical summary metrics when spans or events arrive late.
  - Improved handling of traces containing only logs, spans, or mixed event types.
  - Restored missing persisted summaries without losing previously accumulated totals.
  - Added compatibility for existing trace summaries while applying the updated storage behavior.


dev/docs/adr/071-coding-agent-session-immutable-storage-anchor.md
```diff
@@ -158,12 +158,12 @@ The **second** is the anchor a *rebuild from `event_log`* derives if the committ
...
+- **Worse:** it did not set `refoldOnStoreMiss`, which defaults to `false`. When the epoch row was reaped there was no rebuild net at all — `get()` missed, the fold restarted from `init()`, and the row was rewritten from post-reap events only. The failure mode was silent loss of the trace's accumulated cost, tokens and span count, not a refold storm.
 
-Not fixed in this change, and deliberately: it needs its own migration and a decision about the missing refold net. Tracked as [#6312](https://github.com/langwatch/langwatch/issues/6312).
+Migration 00067 gives `trace_summaries` the same split: `OccurredAt` is the frozen storage/TTL anchor and `EarliestSpanStartMs` is the span timing baseline. Pre-split rows decode their existing `OccurredAt` as both values, avoiding a population refold. The fold also opts into `refoldOnStoreMiss`: losing accumulated totals is not acceptable, and absence is not authoritative while the store deliberately declines dimension-only rows.
```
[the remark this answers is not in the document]
shivanshgupta365: Verified against the commit metadata: 295af393e was authored and committed at 2026-08-02T04:09:17+05:30, so 2026-08-02 is the actual completion date in the author timezone. No change is needed.

langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** Closes #6312. The mechanism is stated precisely and is the reason to read this carefully: log-only trace summaries persist `OccurredAt` as the epoch because the shared timing baseline is only seeded by spans, which lands the row in partition 196952 with an already-expired TTL. Once reaped, the fold restarts from empty and **silently loses accumulated totals**.

**Top inspection targets**
1. `Must Check` - existing rows already in the doomed partition. The fix stops new ones being written there, but anything already sitting in partition 196952 is still on an expired TTL and still headed for reaping. Confirm whether those are recoverable, already gone, or need a separate repair, and whether any fold state has already been lost.
2. `Must Check` - what the new baseline is when there are no spans. "Freezing" a timing baseline is the fix, so confirm the chosen anchor cannot itself be absent or zero on some other path, or this reappears with a different trigger.
3. `Ask Author` - whether the same epoch-baseline assumption exists in other folds seeded from span timing.

**Production risk (scariest first):** silent loss of accumulated fold totals, which announces itself only as numbers being wrong later. That is why this is worth more than its 363 lines.

1. Are there rows already in the expired partition, and has any fold state already been lost?
2. What does the baseline resolve to when no span has been seen?
3. Do other folds share the span-seeded baseline?

langwatch/src/server/clickhouse/migrations/00067_trace_summaries_storage_anchor.sql
```diff
@@ -0,0 +1,26 @@
...
+
+-- Split the span timing baseline out of trace_summaries.OccurredAt.
+--
+-- OccurredAt is the table's weekly partition key and TTL anchor. It previously
+-- also held min(span.startTimeUnixMs), so a log-only trace persisted it as the
+-- epoch: partition 196952 with an already-expired TTL. It also moved whenever
+-- an earlier-starting span arrived late. The projection now freezes OccurredAt
+-- on the first usable contribution time and stores the running span minimum in
+-- this additive column. Existing rows decode their pre-split OccurredAt as both
+-- values and heal on their next ordinary write; no population refold is needed.
```
langwatch-agent: [P1] Rebuild summaries already deleted from the epoch partition

`refoldOnStoreMiss` only executes from `FoldProjectionExecutor.execute()` after a new event is delivered. It therefore cannot recover a log-only trace whose `OccurredAt = 0` row has already been TTL-merged away and which never receives another event: its trace summary and accumulated totals remain absent until normal retention expires. This is not just a pre-deploy edge case—the event log uses the same trace-retention policy, so affected completed traces still within that window are recoverable, but this PR has no job/replay that discovers them. Please add a bounded repair/reprojection (or explicitly establish that prior lost summaries do not need restoration) instead of relying on the next ordinary write.

langwatch-agent (commented):
High-risk review complete.

Found 1 blocking issue: existing trace summaries already reaped from the epoch TTL partition are not rebuilt by this change unless another event is delivered. I left the inline P1 on the migration's “no population refold” claim.

I also traced the new anchor fallback: an absent/zero `storageAnchorMs` falls back through the fold's `createdAt` and then `Date.now()`, so it does not write epoch in the normal path. The other span-timing consumer, `trace_analytics`, already persists its separate `EarliestSpanStartMs` via migration 00061; `evaluation_analytics` has a separate mutable-anchor concern but does not share this span-seeded-zero path.

Static checks: diff whitespace clean. I did not execute external-PR code.

langwatch/src/server/clickhouse/migrations/00067_trace_summaries_storage_anchor.sql
shivanshgupta365: Addressed in ae656d500. The migration and ADR now explicitly define this as forward-only recovery: surviving rows heal on their next write, a later event rebuilds a missing summary through refoldOnStoreMiss, and already-removed completed summaries are not proactively swept. That matches issue #6312’s deliberate pre-split decode/no-population-rebuild scope; an eager event_log replay would be a separate bounded operational migration.

platform/app/src/server/clickhouse/migrations/00067_trace_summaries_storage_anchor.sql
```diff
@@ -0,0 +1,31 @@
...
+-- epoch: partition 196952 with an already-expired TTL. It also moved whenever
+-- an earlier-starting span arrived late. The projection now freezes OccurredAt
+-- on the first usable contribution time and stores the running span minimum in
+-- this additive column. Existing rows decode their pre-split OccurredAt as both
+-- values and heal on their next ordinary write without forcing a population
+-- refold. This migration is intentionally forward-only: summaries already
+-- removed by the old epoch-anchored TTL are not batch-replayed. If another
+-- event arrives, refoldOnStoreMiss rebuilds that trace from the event log;
+-- completed traces with no later event remain absent and age out with the
+-- normal trace-retention window.
```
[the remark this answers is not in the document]
shivanshgupta365: Verified against current head ae656d500. TraceSummaryFoldProjection sets refoldOnStoreMiss: true in its live options, and the updated contract test exercises a null store read, verifies eventLoaderUpTo is called, and confirms both historical and delivered spans are rebuilt. The “declares no refoldOnStoreMiss” wording appears only in removed diff lines from the legacy policy, not in the current test or implementation. The migration and ADR therefore match the active contract; no code change is needed.

platform/app/src/server/clickhouse/migrations/00067_trace_summaries_storage_anchor.sql
```diff
@@ -0,0 +1,31 @@
...
+--
+-- No ON CLUSTER: the Replicated database propagates DDL itself.
```
langwatch-agent, line 19: **P1 — Repair the summaries that the epoch partition has already reaped.** This explicitly leaves completed log-only traces absent unless another event happens to arrive. `refoldOnStoreMiss` only protects the next delivery, so it cannot restore the accumulated totals for the existing affected population—precisely the silent data loss this fix closes. Add an operational repair/backfill that identifies/replays the affected trace aggregates before merge (with bounded batching and an idempotency/verification plan), or keep the issue open and do not claim it is closed.

langwatch-agent (commented):
1 P1 — the migration intentionally leaves already-expired trace-summary folds unrepaired; see inline.

0xdeafcafe:
Hi @shivanshgupta365, thanks for this contribution. We have reworked this part of the system quite heavily, plus done a large codebase restructure. While this bug is still present, we have a different way we would like to tackle it so i'm going to close this PR.

PR #6320, opened 2026-07-29
fix(security): raise JS transitive override floors across the five pnpm workspaces

langwatch-agent:
Clears **2 CRITICAL, 27 HIGH and 4 MODERATE alerts** by tightening pnpm transitive overrides in every workspace that carries a vulnerable copy: root, `langwatch`, `mcp-server`, `skills` and `typescript-sdk`. No application code changes.

## Alerts cleared, by package

| package | floor | alerts | workspaces |
|---|---|---|---|
| `postcss` | `>=8.5.18` | **6** | root (#1760), langwatch (#1757), mcp-server (#1759), skills (#1754, #1765), typescript-sdk (#1766) |
| `fast-uri` | `>=3.1.4` | **6** | langwatch (#1701, #1706), mcp-server (#1713, #1715), skills (#1727, #1729) |
| `brace-expansion` | per major line | **4** | langwatch (#1621, #1622), typescript-sdk (#1669, #1670) |
| `shell-quote` | `1.10.0` | 2 | root (#1640), langwatch (#1623) |
| `js-yaml` | `4.3.0` | 2 | mcp-server (#1638), typescript-sdk (#1696) |
| `axios` | `1.18.1` | 1 | langwatch (#1631) |
| `@opentelemetry/propagator-jaeger` | `>=2.9.0` | 1 | langwatch (#1702) |
| `tar` | `>=7.5.19` | **5** | root (#1643 CRITICAL, #1642, #1641, #1644, #1761) |
| `@vitest/browser` | `>=4.1.10` | **1** | langwatch (#1703 CRITICAL) |
| `js-yaml` (5.x line) | `>=5.2.2` | **3** | langwatch (#1768), typescript-sdk (#1767, #1614) |
| `brace-expansion` (5.x line) | `>=5.0.8` | 1 | langwatch (#1791) |
| `react-router` | `>=8.3.0` | 1 | langwatch (#1758) |

## Three things worth knowing

**1. Overrides must go in each workspace's *effective* file.** `langwatch`, `skills` and `typescript-sdk` declare `pnpm.overrides` in `package.json`; root and `mcp-server` declare `overrides:` in `pnpm-workspace.yaml`. Each workspace keeps the location it already used, because a `package.json` block silently disables the `pnpm-workspace.yaml` one entirely - so introducing one where only the YAML exists would have deactivated every override already there.

[...]

Issue #1740, opened 2026-02-20
Omniversal Automation Callbacks


rogeriochaves:
Epic tracking issue for Omniversal Automation Callbacks.

langwatch-agent:
Closed by board triage: epic tracking issue with no open sub-issues. Reopen when scoped into sub-issues.

Issue #1739, opened 2026-02-20
EPIC: Luxury Automated Demo Projects


rogeriochaves:
Epic tracking issue for Luxury Automated Demo Projects.

langwatch-agent:
Closed by board triage: epic tracking issue with no open sub-issues. Reopen when scoped into sub-issues.

PR #6449, opened 2026-08-03
fix(security): raise override floors and otel core in the typescript-sdk examples

langwatch-agent:
Clears **19 alerts** (14 HIGH, 5 MODERATE) across the five `typescript-sdk/examples/*/package-lock.json` files. These were the last JS manifests in the repo still carrying vulnerable transitives after #6320 covered the five pnpm workspaces.

**This supersedes #5844**, which covered the same otel-core fix for only two of the five examples and had gone conflicting against main. Extending it here avoids two of my own PRs fighting over the same lockfiles.

## Alerts cleared

| package | floor | alerts | examples |
|---|---|---|---|
| `@opentelemetry/propagator-jaeger` | `>=2.9.0 <3` | **4** | evaluation (#1730), langchain (#1732), langgraph (#1734), vercel-ai (#1742) |
| `linkify-it` | `>=5.0.2 <6` | **3** | langchain (#1731), langgraph (#1733), vercel-ai (#1741) |
| `brace-expansion` | per major line | 2 | mastra (#1662, #1663) |
| `fast-uri` | see below | 2 | mastra (#1739, #1740) |
| `js-yaml` | `4.3.0` | 2 | langchain (#1656), langgraph (#1659) |
| `shell-quote` | `1.10.0` | 1 | mastra (#1664) |
| `@opentelemetry/core` | `>=2.8.0` | **5** MODERATE | evaluation (#1454), langchain (#1456), langgraph (#1459), mastra (#1463), vercel-ai (#1464) |

## otel core >= 2.8.0, reached two different ways

`@opentelemetry/core@2.7.1` was present in all five locks. Four examples reach it through **`@opentelemetry/sdk-node@0.217.0`, which pins core 2.7.1**, so the fix is to move their declared range to `^0.219.0` - the lock then carries core 2.8.0 and 2.10.0 with no 2.7.1 left. **mastra does not depend on sdk-node at all**, so it gets a scoped `@opentelemetry/core@>=2.0.0 <2.8.0 -> >=2.8.0 <3` override instead.

## Every override is capped to its major line

A bare floor lets npm satisfy the constraint with a major jump. Two cases where that actually happened on the first pass and was then corrected:

- `linkify-it: ">=5.0.2"` resolved to **6.1.0** and dragged in `uc.micro@3.0.0`. The advisory is fixed in 5.0.2, so the rule is now `">=5.0.2 <6"` and the lock moves 5.0.1 -> 5.0.2 exactly.
- `@opentelemetry/propagator-jaeger: ">=2.9.0"` is now `">=2.9.0 <3"`.

## mastra's fast-uri goes 4.0.0 -> 3.1.4, and that is the fix

This one looks like a downgrade in the diff, so it is worth being explicit. `fast-uri`'s only consumers in that tree are two copies of `ajv`, and **both declare `fast-uri: ^3.0.1`**. The 4.0.0 in main's lock came from the example's existing unscoped `"fast-uri": ">=3.1.2"` override pushing them off their declared range onto the 4.x line - which is itself vulnerable (#1739, #1740). Replacing it with `fast-uri@<3.1.4 -> 3.1.4` returns them to the line they ask for, at its patched version, and removes 4.x from the tree entirely.

A `fast-uri@>=4.0.0 <4.1.1 -> 4.1.1` rule is kept alongside it as a guard, so if a future dependency legitimately pulls 4.x it lands patched.

## Nothing else drifts

Per-package resolved version-set diff against `origin/main` for all five locks. Beyond the table above the only changes are:

- `@opentelemetry/core` gains a `2.10.0` entry **alongside** the existing 2.7.1, because propagator-jaeger 2.10.0 requires it. Nothing is removed.
- `langwatch` moves 0.33.1 -> 1.1.0. That is not a dependency bump: every example declares `"langwatch": "file:../../"`, so the lock is simply recording the current version of the SDK in this repo, which the locks had not caught up with.

No package resolves to a lower version than its consumers declare.

## Testing

- Re-verified all of the above after the otel core change as well.
- Built the SDK first (`typescript-sdk`: `pnpm install --frozen-lockfile` + `pnpm build`, exit 0), since the examples link it by `file:`.
- `npm install` + `npm run build` (tsc) in each example that has a build script: **langchain exit 0, vercel-ai exit 0**.
- **langgraph fails typecheck, and it fails identically on unmodified `origin/main`.** Verified by building it in a clean worktree at `origin/main` and diffing the full sorted error list, not a truncated tail: **6 errors on both sides, identical line, column and TS code** (`TS2769` at 209:35 and `TS2345` at 216:46, 220:36, 223:37, 226:45, 229:45), all about `RunnableLike<unknown, BranchPathReturnValue, ...>` in the example's own source. `diff` reports no difference. It is a langgraph API typing issue unrelated to any package here, and belongs in its own change.
- `evaluation` and `mastra` have no plain tsc build script (`evaluation` has no build, `mastra` uses `mastra build`), so they are covered by the lock audit rather than a compile.

## Summary by CodeRabbit

- **Chores**
  - Updated the OpenTelemetry SDK used by several TypeScript SDK examples.
  - Added safeguards to enforce compatible versions for telemetry, parsing, URI, shell, linking, and expansion packages.
  - Preserved existing dependency constraints while improving consistency and security across example projects.
  - Refined package version overrides to address affected dependency ranges and support more reliable example setup and execution.

### DIFF
--- a/typescript-sdk/examples/evaluation/package-lock.json
+++ b/typescript-sdk/examples/evaluation/package-lock.json
@@ -21,48 +21,50 @@
     },
     "../..": {
       "name": "langwatch",
-      "version": "0.33.1",
+      "version": "1.1.0",
       "license": "MIT",
       "dependencies": {
         "@opentelemetry/api-logs": "0.205.0",
         "@opentelemetry/core": "^2.0.1",
-        "@opentelemetry/exporter-logs-otlp-http": "0.216.0",
-        "@opentelemetry/exporter-trace-otlp-http": "0.205.0",
-        "@opentelemetry/instrumentation": "0.218.0",
+        "@opentelemetry/exporter-logs-otlp-http": "0.219.0",
+        "@opentelemetry/exporter-trace-otlp-http": "0.219.0",
+        "@opentelemetry/instrumentation": "0.219.0",
         "@opentelemetry/resources": "^2.0.1",
         "@opentelemetry/sdk-logs": "0.205.0",
         "@opentelemetry/sdk-metrics": "^2.0.1",
         "@opentelemetry/sdk-trace-base": "^2.0.1",
         "@opentelemetry/semantic-conventions": "^1.36.0",
-        "@types/prompts": "^2.4.9",
-        "chalk": "^4.1.2",
-        "commander": "^12.0.0",
+        "chalk": "^5.6.2",
+        "commander": "^15.0.0",
         "dotenv": "^17.3.1",
-        "js-yaml": "^4.1.0",
+        "js-yaml": "^5.2.0",
         "liquidjs": "^10.27.0",
         "open": "^11.0.0",
-        "openapi-fetch": "^0.16.0",
+        "openapi-fetch": "^0.17.0",
         "ora": "^9.3.0",
         "prompts": "^2.4.2",
         "xksuid": "^0.0.4",
         "zod": "^4.0.14"
       },
       "bin": {
-        "langwatch": "dist/cli/index.js"
+        "langwatch": "dist/cli/index.js",
+        "lw": "dist/cli/index.js"
       },
       "devDependencies": {
         "@eslint/js": "^9.32.0",
         "@langchain/core": ">=0.3.68 <0.4.0",
         "@langchain/langgraph": ">=0.4.0 <1.0.0",
         "@langchain/openai": ">=0.6.0 <1.0.0",
+        "@langwatch/langy": "workspace:*",
         "@opentelemetry/api": "^1.9.0",
-        "@opentelemetry/sdk-node": "0.216.0",
+        "@opentelemetry/sdk-node": "0.219.0",
         "@opentelemetry/sdk-trace-node": "^2.0.1",
         "@opentelemetry/sdk-trace-web": ">=2.0.1",
         "@types/debug": "^4.1.12",
         "@types/js-yaml": "^4.0.9",
         "@types/node": "^24.1.0",
-        "@typescript/native-preview": "7.0.0-dev.20260426.1",
+        "@types/prompts": "^2.4.9",
+        "@typescript/native-preview": "7.0.0-dev.20260705.1",
         "@vercel/otel": "^1.13.0",
         "@vitest/coverage-v8": "^4.1.0",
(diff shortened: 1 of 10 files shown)


langwatch-agent:
CI green: **50 checks passing, 0 failing, no failed workflow runs**, MERGEABLE. The two pending are CodeRabbit and action-semantic-pull-request, the usual steady state.

Scope grew after the PR was opened: it now also carries the otel core >=2.8.0 fix for all five examples (#1454, #1456, #1459, #1463, #1464), which is why **#5844 is closed in its favour**. Total is 19 alerts, 14 HIGH and 5 MODERATE.

langwatch-agent:
Status update: this PR is stranded by the monorepo restructure, not abandoned.

The single-workspace merge (ADR-076) and the `sdks/`+`services/` reorganisation moved or deleted every path this branch touches, so GitHub now reports it `CONFLICTING`. The work itself is still wanted; it needs re-homing onto the new layout rather than a rebase, because the old files no longer exist to rebase against.

I re-homed the JS override work first, since it carried a CRITICAL: #6569 is open and green, closing 39 of the 43 alerts on the new root lockfile. I will re-home this one the same way on a following run and will close it in favour of its successor once that is green, not before.

If you would rather I dropped any of this, say so on the PR and I will close it instead.

langwatch-agent:
Superseded by #6866, which is now open and fully green (all 11 required gates SUCCESS).

This PR targeted `typescript-sdk/examples/*`, and the monorepo restructure moved those to `sdks/typescript/examples/*`. The files it edits no longer exist, so it has been unrebaseable rather than merely stale.

#6866 re-homes the same idea onto the new paths and goes further: it closes 45 of the 50 open alerts on the five example manifests (19 HIGH, 25 MODERATE, 1 LOW), with a version-set audit showing 30 packages up and zero downgrades, plus a build and boot of the mastra example to exercise the `@hono/node-server` 1.x to 2.x major it takes.

The five alerts it cannot close have no published fix in any version: `image-size` (#2092, #2093) and `@ai-sdk/provider-utils` (#1963, #1981, #2003).

Closing this in favour of that one, as promised on this thread.

PR #1880, opened 2026-03-02
chore(main): release langwatch 3.0.0-prerelease.1

0xdeafcafe:
This release is too large to preview in the pull request body. View the full release notes here: https://github.com/langwatch/langwatch/blob/release-please--branches--main--components--langwatch--release-notes/release-notes.md

PR #6450, opened 2026-08-03
fix(security): bump the opentelemetry group to 0.220 with the api-logs migration

langwatch-agent:
Carries Dependabot's opentelemetry group bump (#6056, 22 updates across `langwatch`, `mcp-server` and `typescript-sdk`) **together with the app-side migration it needs to compile**.

#6056 cannot merge on its own. sdk-node 0.220.0 pulls `@opentelemetry/api-logs` 0.220, which makes three breaking changes main's source has not adopted, and `sdk-javascript-ci` fails on all of them. Since the new call shapes do not compile against 0.219 either, the migration cannot land ahead of the bump - they have to go in one PR.

## The migration

**1. `NoopLoggerProvider` was removed.** `LoggerProvider` and `Logger` are type-only now; the replacement is a `createNoopLogger()` factory.

```ts
// typescript-sdk/src/observability-sdk/logger/index.ts
const noopLoggerProvider: LoggerProvider = { getLogger: () => createNoopLogger() };
let currentLoggerProvider: LoggerProvider = noopLoggerProvider;
```

**2. `Logger` now requires `enabled(options?)`.** `LangWatchLoggerInternal` gets a delegating method, which is what clears `TS2420`/`TS2741`.

**3. `SimpleLogRecordProcessor` / `BatchLogRecordProcessor` take an options object**, `{ exporter }`, not a bare exporter. This is the widest change and it reaches beyond the SDK:

- `typescript-sdk/src/observability-sdk/setup/node/setup.ts` (3 sites)
- `typescript-sdk/src/cli/telemetry/events.ts`
- **`langwatch/src/instrumentation.node.ts`** - the app's own instrumentation, not just the SDK
- two integration tests, plus a JSDoc `@example` in `setup/node/types.ts` that would otherwise document a call shape that no longer compiles

**4. Test-only: the node tracer provider no longer exposes `_config`.** The merged `resource`, `sampler`, `spanLimits` and `idGenerator` now live under `_tracerOptions` (9 assertions in `setup-config-options.integration.test.ts`).

I did not take that last one from a changelog - I probed the provider at runtime to confirm the actual shape:

```
ctor: TracerProvider
private keys: [ ..., '_activeSpanProcessor', '_tracerOptions', '_tracers' ]
  _tracerOptions -> [ 'resource', 'sampler', 'spanLimits', 'idGenerator', 'spanProcessor', 'meterProvider' ]
```

The unit test that mocked `NoopLoggerProvider` is updated to mock `createNoopLogger` instead, including the new `enabled` method.

## Testing

- **typescript-sdk**: `pnpm build` (which is `tsc --noEmit` + tsup + the CLI version postbuild check) - **exit 0**. `pnpm test:unit` - **151 files, 2103 tests, all passing**.
- **langwatch app**: `pnpm typecheck:all` (both `typecheck` and `typecheck:tests`) - **clean**. `NODE_ENV=test pnpm build` - **exit 0**.
- **mcp-server**: `pnpm build` - **exit 0**.

One note on the lockfile: the rebase onto current main left `langwatch/pnpm-lock.yaml` with an `overrides:` header that no longer matched `package.json`, so a frozen install failed with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. The lock is regenerated in this branch, so frozen installs pass.

## Relationship to other PRs

- **Supersedes #6056** once green - same bump, plus the code that makes it build. I have left #6056 open for now with a comment explaining the failure, so the group's contents stay visible until this lands.
- This replaces my own earlier attempt at the same thing, which went stale when main moved to 0.219 independently and is now closed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with the latest OpenTelemetry releases.
  * Preserved log export configuration and logger enablement behavior.
  * Maintained telemetry setup, resource configuration, and tracing options.

* **Documentation**
  * Updated logging configuration examples to reflect current supported APIs.

* **Tests**
  * Updated telemetry and logger tests to align with current OpenTelemetry behavior.


langwatch-agent:
Status update: this PR is stranded by the monorepo restructure, not abandoned.

The single-workspace merge (ADR-076) and the `sdks/`+`services/` reorganisation moved or deleted every path this branch touches, so GitHub now reports it `CONFLICTING`. The work itself is still wanted; it needs re-homing onto the new layout rather than a rebase, because the old files no longer exist to rebase against.

I re-homed the JS override work first, since it carried a CRITICAL: #6569 is open and green, closing 39 of the 43 alerts on the new root lockfile. I will re-home this one the same way on a following run and will close it in favour of its successor once that is green, not before.

If you would rather I dropped any of this, say so on the PR and I will close it instead.

langwatch-agent:
Closing. This PR can no longer close a single alert, and main has moved past the version it targets.

**It targets pre-restructure paths.** Every file it touches lives under `langwatch/`, `mcp-server/` or `typescript-sdk/`, all of which moved in the monorepo restructure. There is nothing to rebase onto.

**Main is already ahead of it.** This branch bumps the group to 0.220 from a 0.21x baseline; main now declares `@opentelemetry/sdk-node` 0.219.0 in `platform/app`, `sdks/typescript` and `mcp/typescript`, and typechecks green with its own handling of the api-logs surface. The migration code this PR carried has been overtaken.

**The one alert an otel bump could target is structurally unclosable right now.** The only open `@opentelemetry` alert on `pnpm-lock.yaml` is #1880 (`@opentelemetry/core < 2.8.0`, MODERATE). Closing it needs every sub-2.8.0 subtree gone, and the tree currently carries three:

- `1.28.0` under `@opentelemetry/otlp-transformer@0.55.0`, which is pinned deliberately, together with `resources`, `sdk-logs`, `sdk-metrics` and `sdk-trace-base` at 1.28.0
- `2.1.0` under the 0.205.0 exporter line, which follows the `@opentelemetry/api-logs` and `@opentelemetry/sdk-logs` 0.205.0 pins
- `2.7.1` under the 0.217.0 and 0.218.0 exporter lines

Bumping the group lifts the second and third but cannot touch the first while that 0.55.0 pin stands, so #1880 stays open either way. All risk, no closure. This is recorded as a deliberate non-override in the comment next to the security block in `pnpm-workspace.yaml` on #6569.

For the rest of that manifest, #6569 is open and green and closes 51 of the 55 alerts on it. What remains after it: #1880 above, #1874 and #1877 (`@langchain/community`, patched only in the 1.x line against a 0.3.59 tree, so a major migration), and #1879 (`@ai-sdk/provider-utils`, no patched version published).

Nothing here is worth carrying forward, so this is not being re-homed. If the `otlp-transformer@0.55.0` pin is ever lifted, #1880 becomes closable and that is the moment to revisit the otel line.

PR #6451, opened 2026-08-03
perf(clickhouse): recent-window partition-hint resolvers + off-loop worker liveness

### LINKED ISSUE #6455: Workers: partition-hint resolvers must not scan cold storage; liveness must distinguish saturated from dead
## For humans

Two related requirements for event-processing workers under sustained load: per-job database lookups must not scan cold storage to answer questions about recent data, and a worker that is merely busy must never be mistaken for a dead one and killed. Both failure modes compound under backlog: slow lookups extend job latency (throughput = in-flight slots ÷ job latency), and probe kills requeue in-flight work, deepening the very backlog that caused the saturation.

## Requirement 1 — partition-hint resolvers must prune before they seek

Tables like `evaluation_runs` and `trace_summaries` are partitioned by week on a time column, with older partitions tiered to object storage. The heavy reads already prune partitions when given a time hint — but the small "resolver" queries that *produce* that hint (`argMax(ScheduledAt)`, `min(OccurredAt)`) ran unbounded, walking every weekly partition's index including cold ones. A point seek that should cost tens of milliseconds cost whole seconds, multiplied across every job that needs it.

**Required behavior**: resolvers probe a recent window first (35 days ≈ five weekly partitions, local disk) and fall back to an unbounded seek only on a miss — so the hot path (aggregates minutes old) never touches cold storage, and old aggregates remain correct at the cost of one extra probe.

## Requirement 2 — liveness must distinguish "saturated" from "dead"

The worker's `/healthz` was served on the same Node event loop that runs jobs. A worker saturated with legitimate work can pin that loop for over a minute; no realistic kubelet probe budget survives that, so the platform kills exactly its busiest workers — the pathological outcome.

**Required behavior**: liveness is answered from a dedicated worker thread that judges the main loop by a shared heartbeat (atomic cross-thread read): 200 while the loop is saturated but moving, 503 only after it has been silent past a generous stall budget (5 minutes) — so a genuinely wedged process still restarts. `/metrics` proxies to the main thread through the existing bearer gate with a timeout: a stalled loop fails a scrape, never a probe.

## Implementation

Both shipped in #6451:
- `RESOLVER_RECENT_WINDOW_MS` + two-phase probes in `evaluation-run.clickhouse.repository.ts` and `span-storage.clickhouse.repository.ts`
- `LIVENESS_THREAD_SOURCE` + heartbeat in `startWorkers.ts`
- Spec: `specs/server/worker-liveness-probe.feature`, rule "Liveness answers even while the main event loop is saturated"

## Follow-ups (not in #6451)

- Sweep other unbounded point-seeks on partitioned tables (same shape, other repositories) — candidates surface in the `coldScan` query-warning logs.
- Consider restoring a chart-default liveness probe config aligned with the heartbeat budget.

Related: the broader capacity strategy lives in the internal tracking issue for queue throughput and fairness.

sergioestebance:
## For humans

When the queue backs up, workers spend their time on two ClickHouse lookups that search months of archived data to answer questions about things that happened minutes ago — and while saturated with that work, Kubernetes kills them for "not responding", which requeues everything in flight and deepens the very backlog causing the saturation. This PR makes those lookups ~20× cheaper and makes a busy worker impossible to mistake for a dead one. Pillar 1 of langwatch-saas#856.

## What

**1. Recent-window-first partition-hint resolvers.** `resolveScheduledAtMs` (evaluation_runs) and `resolveTraceOccurredAtMs` (trace_summaries) exist to give the heavy reads a partition bound — but were themselves unbounded, walking every weekly partition's index including S3-tiered cold ones, costing whole seconds per call on lookups whose subjects are minutes old. Both now probe `>= now() - 35 days` first (five weekly partitions, local disk) and run the unbounded seek only when the probe misses (old or unknown aggregate) — correctness unchanged.

**2. Liveness off the event loop.** The metrics port is now bound by a small worker thread (`LIVENESS_THREAD_SOURCE`, eval'd — no bundler-resolved file). `/healthz` answers in-thread from a shared heartbeat the main loop stamps every second (BigInt64 + Atomics — cross-thread float reads can tear): 200 while the loop is merely saturated, 503 only once it has been silent past a 5-minute budget — a genuinely wedged worker still restarts, but a worker pinned by a minute of legitimate catch-up work no longer gets SIGTERM'd, losing its in-flight jobs. `/metrics` proxies to the main thread through one shared bearer-gate function with a 10s timeout — a stalled loop fails the scrape, never the probe. If the thread cannot start, boot falls back to the previous in-loop server.

## Why it matters

Worker throughput is `slots ÷ job latency`, and job latency was dominated by these per-job reads (measured 0.7–1.2s each at thousands of calls per hour, fleet-wide, with CPU nearly idle). Cutting them to local-disk point seeks multiplies fleet capacity without adding a pod. The probe change removes the failure mode where saturation itself triggers kills that amplify the backlog.

## Tests

- `evaluation-run.resolver-window.unit.test.ts` — windowed probe first, no unbounded scan on a hit; a miss falls back and the resolved timestamp still reaches the heavy read as a partition bound.
- `span-storage.resolver-window.unit.test.ts` — same contract for the trace resolver, including the fallback value reaching the span read's window.
- `livenessThread.unit.test.ts` — boots the real thread: fresh heartbeat → 200; stalled past budget → 503; `/metrics` proxies the parent's reply and 503s when the parent never answers.
- `specs/server/worker-liveness-probe.feature` — new rule "Liveness answers even while the main event loop is saturated", scenarios bound to the tests above.

Closes #6455

## Summary by CodeRabbit

* **Improvements**
  * Improved evaluation and trace lookups by checking recent data first while retaining fallback access to older records.
  * Added bounded lookup behavior to reduce unnecessary scanning and improve response times for recent data.

* **Reliability**
  * Added an independent liveness monitor that reports service health even when the main process is busy.
  * Health checks now return an error when the application is unresponsive, while metrics requests continue to be proxied with timeout protection.
  * Added a fallback to the existing metrics server if the liveness monitor cannot start.

### DIFF
--- a/langwatch/src/server/app-layer/clients/clickhouse/windowed-read.ts
+++ b/langwatch/src/server/app-layer/clients/clickhouse/windowed-read.ts
@@ -12,6 +12,20 @@ import { incrementWindowedReadCount } from "~/server/clickhouse/metrics";
 export const DEFAULT_PARTITION_WINDOW_MS = 2 * 24 * 60 * 60 * 1000;
 
 /**
+ * Lookback for the recent-first probe that partition-hint RESOLVERS run
+ * before falling back to an unbounded seek.
+ *
+ * The resolvers (`resolveScheduledAtMs`, `resolveTraceOccurredAtMs`) exist to
+ * find the partition-key value that lets the heavy read prune — but without a
+ * bound of their own they walk every weekly partition's index, including
+ * S3-tiered cold ones (measured 0.7–1.2s per call on the worker job paths,
+ * which only ever resolve minutes-old aggregates). 35 days ≈ five weekly
+ * partitions, comfortably on local disk, while the unbounded fallback keeps
+ * old aggregates correct.
+ */
+export const RESOLVER_RECENT_WINDOW_MS = 35 * 24 * 60 * 60 * 1000;
+
+/**
  * The time predicate for one windowed read attempt. A `null` fragment (never a
  * `WindowFragment`) means an unbounded read — no time predicate, the wide scan.
  */
--- a/langwatch/src/server/app-layer/clients/clickhouse/windowed-read.ts
+++ b/langwatch/src/server/app-layer/clients/clickhouse/windowed-read.ts
@@ -18,10 +18,11 @@ export const DEFAULT_PARTITION_WINDOW_MS = 2 * 24 * 60 * 60 * 1000;
  * The resolvers (`resolveScheduledAtMs`, `resolveTraceOccurredAtMs`) exist to
  * find the partition-key value that lets the heavy read prune — but without a
  * bound of their own they walk every weekly partition's index, including
- * S3-tiered cold ones (measured 0.7–1.2s per call on the worker job paths,
- * which only ever resolve minutes-old aggregates). 35 days ≈ five weekly
- * partitions, comfortably on local disk, while the unbounded fallback keeps
- * old aggregates correct.
+ * S3-tiered cold ones — turning a point seek into a cold scan costing whole
+ * seconds. The dominant callers are worker jobs resolving minutes-old
+ * aggregates, but callers may resolve records of any age: 35 days ≈ five
+ * weekly partitions, comfortably on local disk, and anything older pays one
+ * extra probe before the unbounded fallback answers it correctly.
  */
 export const RESOLVER_RECENT_WINDOW_MS = 35 * 24 * 60 * 60 * 1000;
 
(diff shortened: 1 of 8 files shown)


sergioestebance:
All 10 CodeRabbit findings addressed in 95f1d5c97: atomic BigInt64 heartbeat (the Float64 cross-thread read was a real tearing hazard), single shared metrics gate, persistent thread error/exit handlers + fallback reap, isListening rename, named-param resolver helpers, tests import the real stall budget, fallback test asserts the resolved timestamp reaches the span read window, no-reply proxy scenario added to the spec, comment cleanups.

sergioestebance:
Round 2 addressed in 66008f576: duplicate @scenario annotation removed; the metrics import is now static (inline-import guideline); graceful terminate no longer logs as an unexpected thread exit; and bootMetricsServer is split into wireLivenessThread + bootFallbackMetricsServer to satisfy the lines-per-function gate.

sergioestebance:
CodeRabbit's early-exit handshake finding is fixed in f8b06cae2 — startup now rejects on a pre-listening thread exit (listener removed once listening), so the fallback server always gets its chance. Same commit pins the test to the exported WORKER_HEARTBEAT_STALL_BUDGET_MS.

PR #6468, opened 2026-08-03
fix(gateway): collapse redundant BUDGET_UPDATED change events per project

sergioestebance:
Closes langwatch/langwatch-saas#895

## For humans

After every request through the AI Gateway, the platform sends the gateway a note saying "this project's spending changed, throw away what you cached." The gateway reacts by dropping the cached settings for *every* key in that project, so the next request for each of them has to be rebuilt from scratch.

The notes were being sent one per request, and they're interchangeable — the note carries no numbers, it just says "go re-read." One note refreshes everything the next ten would have. So a busy project was making every one of its own keys slower, repeatedly, to deliver the same message.

This collapses them to one note per project per ten seconds.

The catch, and the reason this needed care: some budgets *refuse* requests once you go over them, and that refusal depends on the gateway having fresh numbers. Holding one of those notes back would mean an over-budget key keeps getting served. So notes for budgets that block are never held back — only notes for budgets that just warn you, where arriving a few seconds later costs nothing anyone acts on.

## Why per project, not per (project, budget)

The follow-up was originally scoped as "dedupe per (projectId, budgetId)". Reading the consumer changed that: for this event kind the gateway's change-feed subscriber evicts every bundle whose `ProjectID` matches and never reads `budget_id` at all (`services/aigateway/adapters/authresolver/service.go`, the `ChangeKindBudgetUpdated` case). A second event naming a different budget in the same project therefore evicts exactly what the first already evicted.

Keying on the budget as well would have under-deduped by roughly the number of applicable budgets while buying nothing, so the key is the project alone.

## Why blocking budgets are exempt, and what that buys

The enforcement chain is: reactor writes the debit row to ClickHouse → reactor emits `BUDGET_UPDATED` → gateway's `/changes` poll sees it → gateway evicts the project's bundles → next request re-materialises → `GatewayConfigMaterialiser` reads current spend from ClickHouse → Go precheck compares spend against the limit and returns 402 when `on_breach=block`.

The event is an *invalidation signal, not a data carrier*. That is what makes deduping sound at all: the number the gateway ends up with comes from ClickHouse at re-materialise time, so it already includes every debit written since the event that triggered the read.

It also means suppressing an event is only harmful when nothing else forces a re-read before spend crosses a limit. That is exactly the blocking case, so blocking is exempt:

- **Any budget with `onBreach = BLOCK` among the applicable budgets → emit immediately, dedupe not consulted.**
- All-warn → dedupe applies.

The invariant that makes this safe is that a BLOCK budget's spend can only change when a debit row is written for it, and that write always goes through this reactor with that budget in hand — so it always emits. There is no path where a blocking budget's spend moves silently.

This is also the answer to "make sure the window never dominates block-decision propagation": it structurally cannot, because the dedupe is not on that path. Block propagation stays at the pre-existing ceiling, which is the `/changes` inner poll granularity.

## The window, measured rather than assumed

Measured from the code rather than picked:

- `GET /api/internal/gateway/changes` (`src/server/routes/gateway-internal.ts`) loops re-reading the event table, sleeping **2s** between reads, holding for `timeout_s` — **default 10s**, clamped to 1..25 — then returning 204.
- The Go client (`changeFeedLoop`) re-polls immediately on success; its own 2s sleep is on the error path only.
- That client asks for the 10s hold explicitly — `q.Set("timeout_s", "10")` in `services/aigateway/adapters/controlplane/client.go:205` — so the hold is a real figure from the only caller in the tree, not merely the server-side default it happens to match.

So an emitted event is detected within about 2s, and 10s is the longest a single held poll runs.

`BUDGET_CHANGE_EVENT_WINDOW_SECONDS = 10` follows: below 2s the window would suppress nothing the consumer had not already coalesced, and 10s matches the hold, so a project under continuous advisory traffic causes at most one eviction per poll cycle instead of one per request. Both bounds are pinned by tests so the constant cannot drift away from the cadence it was derived from.

## Known limit, deliberate

This is a leading-edge window with no trailing flush. If advisory traffic stops mid-window, the last debit's refresh waits for the next debit or the bundle TTL rather than arriving at window end.

That is acceptable because nothing enforces on an advisory number, and it cannot occur on the blocking path, which never dedupes. A trailing flush would need a scheduled job on the enforcement path; that cost is not worth paying for a warning banner. Calling it out explicitly because a leading-edge-only throttle that *did* cover the blocking path would be a correctness bug, not a tuning choice.

## What changed

- `langwatch/src/server/gateway/budgetChangeEventDedupe.service.ts` (new) — Redis `SET NX` + TTL, so the window holds across pods. Interface / null / Redis / factory shape copied from `share-view-dedupe.service.ts`. Fails toward **emitting**: emitting is the pre-dedupe behavior and always correct, only noisier, so a Redis outage must not withhold an invalidation the gateway is waiting on.
- `langwatch/ee/governance/reactors/gatewayBudgetSync.reactor.ts` — new `affectsEnforcementDecision` predicate; `emitBudgetUpdatedChangeEvent` consults the dedupe only for all-warn folds. Dependency is optional, so an unwired caller behaves exactly as before. Replaces the `v2 TODO` this PR implements.
- `langwatch/src/server/app-layer/presets.ts` — wires the service from the existing `redis` connection.
- `specs/ai-gateway/budgets.feature` — two `@unit` scenarios for the two halves of the split, each bound by a `@scenario` annotation. `check-feature-parity.ts` reports 33/33 bound.

## Tests

`pnpm test:unit run ee/governance/reactors/__tests__/ src/server/gateway/__tests__/budgetChangeEventDedupe.service.unit.test.ts` — 53 passed across 5 files.

The load-bearing one is that a blocking budget emits even when the dedupe is primed to suppress, and that the dedupe is not even *consulted* on that path. Verified as a real regression test rather than an accident: stubbing `affectsEnforcementDecision` to `false` fails that assertion specifically, and it was restored before committing.

Also covered: warn-only suppression, that suppression never touches the debit row itself (only the invalidation is deduped), fail-toward-emitting on a Redis error, the no-Redis path emitting every time, and the two window bounds against the measured poll cadence.

`pnpm typecheck` and `pnpm typecheck:tests` clean. `pnpm exec biome check` reports no errors on the touched files.

### DIFF
--- a/langwatch/ee/governance/reactors/__tests__/gatewayBudgetSync.reactor.unit.test.ts
+++ b/langwatch/ee/governance/reactors/__tests__/gatewayBudgetSync.reactor.unit.test.ts
@@ -427,4 +427,143 @@ describe("gatewayBudgetSync reactor", () => {
       });
     });
   });
+
+  describe("BUDGET_UPDATED change-event dedupe", () => {
+    /**
+     * The change event is what makes the gateway drop its cached bundles and
+     * re-read spend. Deduping it is safe only where nothing enforces on the
+     * number it refreshes, so these pin the two halves of that split.
+     */
+    function dedupeDeps({
+      onBreach,
+      shouldEmit,
+    }: {
+      onBreach: "BLOCK" | "WARN";
+      shouldEmit: (params: { projectId: string }) => Promise<boolean>;
+    }) {
+      const budget = {
+        id: "budget-1",
+        scopeType: "PROJECT",
+        scopeId: "project-1",
+        window: "MONTH",
+        onBreach,
+      } as GatewayBudget;
+
+      const { deps } = mockDeps(
+        { id: "vk-1", organizationId: "org-1", principalUserId: null },
+        {
+          id: "project-1",
+          teamId: "team-1",
+          team: { organizationId: "org-1" },
+        },
+        [budget],
+      );
+
+      const create = vi.fn().mockResolvedValue({ revision: 1n });
+      (deps.prisma as any).gatewayChangeEvent = { create };
+
+      return {
+        deps: { ...deps, changeEventDedupe: { shouldEmit } },
+        create,
+      };
+    }
+
+    const gatewayFold = () =>
+      ctx(
+        createFoldState({
+          "langwatch.virtual_key_id": "vk-1",
+          "langwatch.gateway_request_id": "req-1",
+        }),
+      );
+
+    describe("given a budget that blocks on breach", () => {
+      describe("when a debit lands inside an already-claimed window", () => {
+        /** @scenario "A blocking budget's spend update is never held back" */
+        it("emits anyway — an enforcement decision is never held back", async () => {
+          const shouldEmit = vi.fn().mockResolvedValue(false);
+          const { deps, create } = dedupeDeps({
+            onBreach: "BLOCK",
+            shouldEmit,
+          });
+
+          await createGatewayBudgetSyncReactor(deps).handle(
+            event,
+            gatewayFold(),
+          );
+
+          expect(create).toHaveBeenCalledTimes(1);
+          // Not even asked: the dedupe is not on the blocking path at all,
+          // so its window can never become the dominant term in how fast a
+          // block decision propagates.
+          expect(shouldEmit).not.toHaveBeenCalled();
+        });
+      });
+    });
+
+    describe("given a budget that only warns on breach", () => {
+      describe("when the window is already claimed", () => {
+        /** @scenario "Repeat updates for a warn-only budget collapse into one" */
+        it("suppresses the redundant emission", async () => {
+          const shouldEmit = vi.fn().mockResolvedValue(false);
+          const { deps, create } = dedupeDeps({ onBreach: "WARN", shouldEmit });
+
+          await createGatewayBudgetSyncReactor(deps).handle(
+            event,
+            gatewayFold(),
+          );
+
+          expect(shouldEmit).toHaveBeenCalledWith({ projectId: "project-1" });
+          expect(create).not.toHaveBeenCalled();
+        });
+
+        /** @scenario "Repeat updates for a warn-only budget collapse into one" */
+        it("still writes the debit row — only the invalidation is deduped", async () => {
+          const shouldEmit = vi.fn().mockResolvedValue(false);
+          const { deps } = dedupeDeps({ onBreach: "WARN", shouldEmit });
+          const insertDebit = deps.budgetCHRepository.insertDebit as ReturnType<
+            typeof vi.fn
+          >;
+
+          await createGatewayBudgetSyncReactor(deps).handle(
+            event,
+            gatewayFold(),
+          );
+
+          expect(insertDebit).toHaveBeenCalledTimes(1);
+        });
+      });
+
+      describe("when the window is free", () => {
+        it("emits the change event", async () => {
+          const shouldEmit = vi.fn().mockResolvedValue(true);
+          const { deps, create } = dedupeDeps({ onBreach: "WARN", shouldEmit });
+
+          await createGatewayBudgetSyncReactor(deps).handle(
+            event,
+            gatewayFold(),
+          );
+
+          expect(create).toHaveBeenCalledTimes(1);
+        });
+      });
+    });
+
+    describe("given no dedupe service is wired", () => {
+      describe("when a debit lands", () => {
+        it("emits every time, as it did before the dedupe existed", async () => {
+          const shouldEmit = vi.fn(async () => true);
+          const { deps, create } = dedupeDeps({ onBreach: "WARN", shouldEmit });
+          const { changeEventDedupe: _omitted, ...withoutDedupe } = deps;
+
+          await createGatewayBudgetSyncReactor(withoutDedupe).handle(
+            event,
+            gatewayFold(),
+          );
+
+          expect(create).toHaveBeenCalledTimes(1);
+          expect(shouldEmit).not.toHaveBeenCalled();
+        });
+      });
+    });
+  });
 });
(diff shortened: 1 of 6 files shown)


rogeriochaves:
Sergio, I folded this into langwatch/langwatch#6395 rather than merging it on its own, and I want to explain why so it does not look like I just took it.

langwatch/langwatch#6395 deletes `gatewayBudgetSync.reactor.ts`, which is where your call site lives, so this PR cannot merge as written. But the problem you found does not go away with the reactor. It gets worse on the replacement path, and I had not spotted that until I read your diff.

The consolidation moves gateway debits onto a process manager that runs one instance per gateway request, and its post-debit emit was ungated. The reactor it replaces carried a per (tenant, trace) job debounce with a 5 minute TTL (`GATEWAY_BUDGET_SYNC_DEBOUNCE_TTL_MS`). So consolidating removed the only collapse this path had, and moved the grain from the trace to the request at the same time. One BUDGET_UPDATED per billable request, each one evicting every bundle in the project. That is a regression langwatch/langwatch#6395 introduced, and your fix is what closes it.

What landed in 61c4c82a0e, with a `Co-authored-by` trailer to you: your dedupe service essentially as written, and your key choice. The project-only key is the part I want to call out, because it is the non-obvious call and it is right. The change-feed subscriber evicts every bundle whose `ProjectID` matches and ignores `budget_id` for this kind, so a second event for a different budget in the same project would evict exactly what the first already evicted.

One thing I added on top: budgets whose `onBreach` is `BLOCK` skip the dedupe entirely. A BLOCK budget's spend feeds the pre-request precheck, so holding its event back would leave an over-limit key served from a cached bundle. Only advisory emissions collapse. The window is also matched to the `/changes` long-poll hold rather than picked freely, so a project under continuous traffic causes at most one project-wide eviction per poll cycle.

Suggest closing this one in favor of langwatch/langwatch#6395. Four of its six files collide with deletions there, so a merge would be more painful than it is worth. If you disagree with the BLOCK exemption or think the window should be tuned differently, say so and I will change it on langwatch/langwatch#6395.

langwatch-agent:

**Mode:** Targeted Review
**Linked issue:** Closes langwatch/langwatch-saas#895. Well described: after every gateway request the platform emits a change event, and the gateway responds by dropping cached settings for *every* key in the project, so each next request rebuilds from scratch.

**Top inspection targets**
1. `Must Check` - correctness of the collapse window. Coalescing change events means a window during which a real settings change is not yet visible to the gateway. Confirm how long that is and that it is acceptable for what it invalidates: a budget or key change taking effect late is a correctness issue on the enforcement path, not just a caching one.
2. `Must Check` - no lost invalidation. Collapsing N events into one is safe only if the last one wins and none can be dropped entirely under concurrency. A dropped invalidation leaves the gateway serving stale budget settings indefinitely, which is the failure worth hunting.
3. `Probably Fine` - the per-project scoping, same granularity as before.

**Production risk (scariest first):** a stale cached budget or key surviving an invalidation, so enforcement runs on old settings. This sits beside langwatch/langwatch#6421's budget bar and langwatch/langwatch#6285's `max_tokens` finding, both cases where the gateway's stated limits and its real behaviour diverged.

1. What is the maximum staleness window after a real settings change?
2. Under concurrent updates, can an invalidation be dropped rather than deferred?

drewdrewthis (commented):
Reviewed end-to-end (proc.review-qa.pr-review). No blocking findings. See verdict comment for detail.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `98c2dee` · **Run:** pr-review (reviewer: @drewdrewthis)

### Notes (non-blocking, for context)

- **[correctness]** Verified the dedup-safety claims against the actual consumer (`services/aigateway/adapters/authresolver/service.go`, `applyChange`/`ChangeKindBudgetUpdated` case): it evicts by `ProjectID` alone when the event carries one (confirmed the emitter always sets `projectId`), treats `budget_id` as unused for this event kind, and re-reads current spend from ClickHouse at re-materialise time rather than diffing old→new state from the event payload. This confirms both review-focus questions in this PR's brief: (1) collapsing does not drop a state transition any consumer depends on — the event is a pure invalidation trigger, not a data carrier; (2) the per-project dedup key is the *correct* scope, matching the consumer's actual eviction granularity exactly (neither over- nor under-collapsing).
- **[correctness]** `affectsEnforcementDecision` correctly gates the dedupe: any applicable `BLOCK` budget forces an unconditional emit, verified both by reading the code path and by the pinned regression test asserting `shouldEmit` is not even called on that path.
- **[design]** Redis `SET NX EX` fail-open-to-emit is the right default (noisier-but-correct on Redis outage vs. silently stale enforcement-adjacent state); `NullBudgetChangeEventDedupeService` preserves pre-PR behavior when unwired.
- Tests are well-targeted at the actual risk surface (blocking-never-suppressed as a named regression test, warn-only suppression, debit-row write independent of event suppression, window bounds pinned to measured `/changes` poll cadence). CI green across all gating checks at head SHA.

---

sergioestebance:
Closing as superseded by #6395 — verified the fold landed on main before closing:

- `budgetChangeEventDedupe.service.ts` + its unit tests are on main (now under `platform/app/src/server/gateway/`), landed via 61c4c82a0 inside the #6395 merge with co-author credit.
- The BLOCK exemption is wired on the replacement path: `affectsEnforcementDecision` in `platform/app/ee/governance/process-manager/gatewayDebits.process.ts:262`, consulted before the dedupe at `:299` — dedupe is skipped entirely when any applicable budget has `onBreach = BLOCK`, matching this PR's design.
- `gatewayBudgetSync.reactor.ts` (this PR's call site) no longer exists on main, so a rebase would be conflict-for-nothing.

No disagreement on the BLOCK exemption or the window — what landed matches this PR's own reasoning.

PR #6485, opened 2026-08-03
revert(metrics): revert the rollup chunk read folding (#6481)

sergioestebance:
## For humans

Reverts #6481. After deploy, every metric rollup batch fails with `Cannot read properties of undefined (reading 'map')` and re-stages with exponential backoff, freezing all rollup processing. Reverting restores the previous slower-but-working read path while the bug is found and fixed with a test that reproduces it.

No data is lost — failed batches re-staged and will process normally once this deploys.

## Follow-up

Re-land the optimization once the undefined result path is identified and covered by a test against the real ClickHouse response shape.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved metric rollup recalculation across individual metric series.
  - Increased accuracy when processing late-arriving data points and overlapping time buckets.
  - Ensured rollups consistently reflect authoritative raw data after updates.
  - Improved persistence of recalculated rollups in ClickHouse-backed metrics.
- **Tests**
  - Expanded coverage for raw points, cumulative changes, late-point convergence, and rollup persistence.

### DIFF
--- a/langwatch/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
+++ b/langwatch/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
@@ -4,18 +4,15 @@
  *
  * Runs the canonical metric repository's real INSERT/SELECT SQL against
  * ClickHouse (migration 00049). The rollup unit tests exercise the pure fold;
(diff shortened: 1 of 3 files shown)


sergioestebance:
Closing the loop: this revert targeted the wrong suspect — the crash was a ClickHouse in-band exception row surfacing as data, present on both read paths once the workload's rows got heavy. Full record: #6489; fix: #6493.

PR #2262, opened 2026-03-11
fix(analytics): fire limit_blocked PostHog event from UI pre-check path

sergioestebance:
## Summary

- **Problem**: When users hit resource limits, the UI pre-checks client-side and blocks before reaching the server. The `limit_blocked` PostHog event only fired from `enforcement.middleware.ts` (server safety net), which is rarely reached. This created a gap in the Growth dashboard analytics.
- **Fix**: Added `trackServerEvent("limit_blocked")` to the `reportLimitBlocked` mutation with `source: "ui_pre_check"`, and tagged the existing middleware event with `source: "server_safety_net"` for clarity.
- **Tests**: Added enforcement middleware unit tests (6 tests) and improved existing test mock strategy (use `importOriginal` instead of hardcoded `limitTypes`).

## Changes

| File | Change |
|------|--------|
| `licenseEnforcement.ts` | Added `trackServerEvent` call when `result.allowed === false` |
| `enforcement.middleware.ts` | Added `source: "server_safety_net"` to existing event |
| `licenseEnforcement.reportLimitBlocked.unit.test.ts` | New test for tracking + fixed mock to use `importOriginal` |
| `enforcement.middleware.unit.test.ts` | **New** — 6 tests covering middleware side-effects |

## Test plan

- [x] `pnpm test:unit` — all 11 tests pass (5 + 6)
- [x] `pnpm typecheck` — no new errors (all pre-existing)
- [x] Verified `limit_blocked` event fires with `source: "ui_pre_check"` when limit is exceeded
- [x] Verified event does NOT fire when limit is not exceeded
- [x] Verified middleware tags events with `source: "server_safety_net"`

> **Note**: This branch is based on `feat/posthog-tracking-events`. Should be merged after that branch lands.

Issue #2236, opened 2026-03-11
Suites page header counts are incorrect and show stale x/n format


drewdrewthis:
## Describe the bug

On the suites page, the amounts shown in the headers are incorrect. When launching 36 runs, only 24/36 is displayed. This appears to be a leftover from when we were calculating pending runs with a pending banner.

### Problems identified:
1. **Header counts don't reflect total runs** — launching 36 runs should immediately show 36 in the tables and headers
2. **Stale x/n format** — the `x/n` display pattern was from the pending banner era and should be removed
3. **Failed count denominator** — for "failed (x/n)", `n` should always represent the total number in the batch, but it currently doesn't

## To reproduce

1. Go to the Suites page
2. Launch a batch of 36 runs
3. Observe the header counts — they show 24/36 instead of 36

## Expected behavior

- All header counts should immediately reflect the total number of runs in the batch
- Remove the `x/n` format since the pending banner is gone
- For failed counts specifically, if `x/n` is kept, `n` must always equal the batch size

## Additional context

Need to audit how all display numbers are calculated across the suites page headers and tables. This likely involves reviewing the query/aggregation logic that powers the header stats.

PR #2234, opened 2026-03-11
fix: custom SDK evaluations - gray badges, stuck Processing, precondition crash

rogeriochaves:
## Summary

Fixes three issues with custom SDK evaluations (via `add_evaluation()`):

- **Gray EVALUATION badges**: Span mapper now parses JSON strings from ClickHouse `Map(String, String)` storage for both `langwatch.reserved.value_types` and `langwatch.output`, and recognizes `evaluation_result`/`guardrail_result` annotated types — restoring green/red badge styling
- **Stuck "Processing..." evaluations**: `customEvaluationSync` reactor dispatched both `startEvaluation` and `completeEvaluation` with the same `occurredAt` (used as group queue sort score). When scores tied, Redis sorted by random UUID, so ~50% of the time `completeEvaluation` ran first, then `startEvaluation` overwrote status back to "in_progress" permanently. Fixed by offsetting `completeEvaluation` by +1ms
- **`toLowerCase` crash in preconditions**: `evaluatePreconditions()` assumed string values but received objects/arrays from chat messages. Added `coerceToString()` that extracts content from chat messages and JSON-stringifies objects

## Test plan

- [x] Span mapper tests: evaluation_result/guardrail_result type preservation, ClickHouse JSON string parsing (21 tests)
- [x] Frontend tests: `getEvaluationResult` unit tests (6 tests), `SpanTypeTag` integration tests (6 tests)
- [x] Reactor unit test: verifies completeEvaluation occurredAt is offset by +1ms
- [x] All pre-existing tests continue to pass
- [x] Deploy and verify custom SDK evaluations show green/red badges
- [x] Verify "Processing..." stuck state no longer occurs

Issue #2201, opened 2026-03-10
Trace details drawer cannot be closed


drewdrewthis:
## Describe the bug

The trace details drawer cannot be closed by any method:
- Clicking the X button does nothing
- Pressing Escape does not close it
- Clicking outside the drawer does not close it

The only way to navigate away is using browser navigation.

## To reproduce

1. Open the Suites page
2. Open a run details drawer
3. Click the trace button within the run details drawer to open the trace details drawer
4. Try to close the trace details drawer via the X button, Escape key, or clicking outside

## Expected behavior

The trace details drawer should close when:
- The X button is clicked
- The Escape key is pressed
- Clicking outside the drawer area

## Environment

- App: LangWatch SaaS
- Browser: (all)

## Additional context

The issue occurs when opening the trace drawer from within the suites run details drawer (nested drawer context).

PR #6569, opened 2026-08-05
fix(security): close 19 alerts across the app and the mastra example (5 HIGH)

langwatch-agent:
## What

Closes **19 of the 21** open alerts on `pnpm-lock.yaml` and `sdks/typescript/examples/mastra/package-lock.json`: 5 HIGH, 13 MODERATE, 1 LOW.

| package | from | to | alerts |
|---|---|---|---|
| fast-uri | 4.1.2 | 3.1.6 (capped to the line ajv 8 declares) | #2238, #2239, #2240, #2241, #2259, #2260, #2261, #2262 |
| browserslist | 4.28.2 | 4.28.7+ | #2201 |
| mysql2 | 3.15.3 | 3.24.2 | #2174, #2237 |
| qs | 6.15.3 | 6.16.0 | #2235, #2236, #2257, #2258 |
| sanitize-html | 2.17.5 | 2.17.7 (parser held at htmlparser2 10.1.0, see below) | #2175, #2266 |
| @tiptap/core | 3.27.1 | 3.30.5 | #2234 |
| @ai-sdk/provider-utils | 3.0.30 | 4.0.46 | #1879 |

## Capping override targets

Three targets could float across a major, and two of them already had.

`fast-uri@<3.1.5 -> >=3.1.5` was uncapped, and the tree had resolved **4.1.2** off it, in a subtree whose only dependent is ajv 8, which declares `^3.0.1`. It now resolves 3.1.6.

Not 3.1.5, and that difference is the point of capping carefully. Once a package is pinned onto an older line, its floor has to answer to *that line's* advisories, and the alerts raised here cannot tell you what those are: they all name `>= 4.0.0, < 4.1.3`, because this lock resolves the 4.x line. Four separate HIGH advisories cover the 3.x line up to 3.1.6 (`>= 3.0.0, < 3.1.6` and `>= 3.1.2, < 3.1.6`), and a floor of 3.1.5 would have quietly landed inside all four. They are visible in this same inbox, raised against other repos' locks that do resolve 3.x.

`body-parser@<1.20.6 -> >=1.20.6` uncapped resolves **2.3.0** under express 4, which declares `~1.20.5`. That is a behaviour change, not just a version change: on 2.x `req.body` is `undefined` rather than `{}` for a request with no body or an unmatched content-type, so a handler that reads `req.body` throws instead of seeing an empty object. Measured on express@4.22.2 directly. 1.20.6 carries the same fix with no behaviour change.

`@hono/node-server@<1.19.13` is capped for the same reason: the 2.x line is a separate advisory, and an unconditional floor would force 1.x dependents onto 2.

## Raising in place, not adding alongside

pnpm matches only the **first** selector for a package name. A correctly-scoped floor added next to a stale one leaves the stale one in charge, so `qs@>=6.11.1 <=6.15.1` is raised in place rather than shadowed by a new entry.

## @ai-sdk/google onto the 3.x line

The 4.x line emits provider-spec **v4**, which `ai@6` rejects at call time with `UnsupportedModelVersionError`; measured against 4.0.50 on ai@6.0.217. 3.0.110 emits v3, reaches the network, and is the first release above the provider-utils advisory, so it is both the working option and the fix.

## Two tiptap packages become explicit dependencies

`@tiptap/core` and `@tiptap/extensions` were auto-installed peers (`autoInstallPeers: true`), never real dependencies of anything in the tree. pnpm does not apply an override to a package in that position: with `"@tiptap/core@<3.30.4": ">=3.30.4 <4"` in place, pnpm rewrote the peer *range* on `@tiptap/extensions` to `>=3.30.4 <4` and then satisfied it with the 3.27.1 already in the lock. `pnpm update -r "@tiptap/*"` did not move it either.

Naming both in `platform/app`'s devDependencies at `^3.30.5` makes them real resolutions, which is what actually closes the advisory. It also repairs a split family: core and extensions were sitting three minors behind the seven `@tiptap/extension-*` packages that peer on them, and `@tiptap/extensions@3.27.1` peers `@tiptap/pm` at exactly 3.27.1 while pm resolved 3.30.5. All eleven `@tiptap/*` packages are now on one version.

## sanitize-html's fix needed its parser held back

Taking `sanitize-html` 2.17.7 is not a lockfile-only change. 2.17.6 moved its parser dependency from `htmlparser2 ^10` to `^12`, and htmlparser2 11 became `type: "module"` with a single ESM `dist/index.js` and no `require` condition, while sanitize-html's own entry is still CommonJS.

Node 22 resolves that pairing without complaint (`require('sanitize-html')` and `import` both work, checked directly against the published tarballs). A CJS loader that predates `require(esm)` does not, and vitest's is one: the first CI run on this branch lost **12 unit suites and one component shard**, all with the same import-time error and none of them related to what they test.

```
SyntaxError: Cannot use import statement outside a module
 ❯ node_modules/.pnpm/sanitize-html@2.17.7/node_modules/sanitize-html/index.js:1:20
      1| const htmlparser = require('htmlparser2');
```

Vitest's own suggested workaround does not fix it here: inlining `htmlparser2`, then `sanitize-html` as well, then adding both to `deps.optimizer.ssr.include`, all left the same `require` failing.

So the parser is held on the last line that still ships CommonJS, `"htmlparser2@>=11.0.0": ">=10.1.0 <11"`. That is safe on both counts that matter:

- **htmlparser2 has no advisory of its own**, in any version, so holding it back trades nothing away.
- **The sanitize-html fix is in sanitize-html's code, not the parser's.** On htmlparser2 10.1.0, 2.17.5 still lets the SVG SMIL case through as `<set attributename="href" to="javascript:alert(1)">` while 2.17.7 strips it to `<svg></svg>`. Same parser, different result: the advisory closes because of the version this PR raises to.

Ordinary sanitization is identical on the held parser (tag stripping, `javascript:` href removal, `allowedTags` filtering all checked), and the templating package that uses it stays green.

## Two alerts stay open

**#1874 and #1877 (@langchain/community)** are deliberately not pinned. The advisory range is `<=1.1.17`, which covers the 0.3.59 this tree resolves, and the first patched release is 1.1.18 — so the only floor that closes them is a major bump of the package, which arrives through `@copilotkit/runtime`.

The latest CopilotKit does drop it: `@langchain/community` is an *optional peer* there rather than a hard dependency, so it leaves the lock entirely. But that is a 1.8.14 to 1.70.1 move, and it relocates `setMessages` to `useCopilotChatHeadless_c`, a hook documented as part of a separately-licensed tier. That is a product decision, not a dependency bump, so it is left for a human rather than forced through here.

## Verified

- `tsc --noEmit` across `platform/app` (full project): **0 errors**
- the tiptap-backed search editor: **147 tests green** across 5 files
- the sanitize-html-backed templating package: **173 tests green** across 13 files, on the held parser
- the 12 unit suites and the component shard that the parser break took down: green again, and shard 2/4 of the unit lane run end to end
- hooks, filters, evaluations and utils slices: **1337 tests green** across 102 files
- `npm install --package-lock-only` in the mastra example: clean, no peer conflict
- version-set diff of both locks reviewed: no package moves down except the two deliberate caps above, and the only additions are `undici 6.28.0` (itself the fixed version for every 6.x advisory) and the express 4 subtree that the body-parser cap implies
- closure re-checked against each advisory's `vulnerableVersionRange` rather than its `firstPatchedVersion`: 19 of 21 closed
- every one of the 25 resolved versions this change moves audited against **every published advisory for its package**, not only those raised against this repo: zero vulnerable resolutions (this is the check that caught the fast-uri floor)
- unit shard 2 of 4 run end to end: **603 files, 7620 tests, all green**

https://claude.ai/code/session_017SsqT8ApEch8zUo7Aqc3y7


langwatch-agent:
CI is green: 145 checks, 0 failures, all 12 required gates passing.

Supersedes and closes #6320 and #5746.

Remaining stranded by the restructure and still to re-home: #5863 (Python uv floors), #6449 (SDK example lockfiles) and #6450 (otel 0.220 plus its api-logs migration). I have left a status note on each; they need re-homing onto the new paths rather than a rebase, and I will close each only once its successor is green.

Separately, 164 of the 352 open alerts now point at manifests the restructure deleted, so the inbox overstates the real number by about 47%. Details on #5465.

Ready for human review.

langwatch-agent:
Rebased onto current `main` and narrowed after a real regression showed up in CI. Worth writing down what happened, because I nearly mis-diagnosed it twice.

**The regression.** After the rebase, `e2e` failed and took `sdk-javascript-complete` with it. The SDK e2e suite was failing with `OTLPExporterError: Internal Server Error`: the app was returning 500 on OTLP span ingestion.

**Two wrong turns on the way to that.** First I checked whether other branches were passing `sdk-javascript-ci` today, saw main green twice plus eight other branches, and nearly wrote it off as the known e2e flake. That comparison was worthless: those runs **skipped** the `e2e` job entirely under path filters, so a `success` conclusion said nothing about e2e. Only two runs in the recent history actually executed it, mine and a Dependabot PR touching a standalone npm lockfile outside the pnpm tree, and that one passed. So the problem really was in my tree.

Second, I assumed `undici` (the HTTP layer) was the culprit. It is not: nothing in the failing path depends on it, `msw` does not pull it, and its only consumers here are jsdom and testcontainers.

**The actual cause** was the `@opentelemetry` floors I had added. Forcing `@opentelemetry/propagator-jaeger` to `>=2.9.0` dragged `@opentelemetry/core` onto 2.10.0 alongside the 2.8.0/2.9.0 copies the app already had, and OTLP ingestion started 500ing. That tree is pinned around a deliberate `@opentelemetry/otlp-transformer@0.55.0` (see the `protobufjs` comment in the workspace file, which spells out why the 7.x branch is load-bearing), so it has to move as one coordinated change rather than a floor bump.

**Both otel overrides are now dropped**, with a comment recording why. The `@opentelemetry/core` one was pure cost anyway: alert #1880 covers everything below 2.8.0 and the tree keeps a 1.28.0 subtree under that pinned transformer, so the alert stays open no matter what the 2.x line does. Dropping `propagator-jaeger` costs exactly one alert, #1886. Both belong with the otel work in #6450 rather than here.

**Net effect:** now closes **38 alerts** (1 CRITICAL, 22 HIGH, 13 MODERATE, 2 LOW), and I folded in two advisories that landed since the first push:

- `js-yaml` 4.x floor raised again, `<4.3.0` to `<4.3.1` (#2061)
- `mermaid` `>=11.16.1` (#2056, #2057, #2058, #2059, #2060)

The version-set diff against `main` is **26 packages, all forward, no downgrades, no removals**, and no `@opentelemetry` package differs from main any more.

Re-verified after the change: `CI=true pnpm install --frozen-lockfile` from the repo root exit 0, `platform/app` `pnpm typecheck:all` clean.

Still open and out of scope, unchanged from before: `@ai-sdk/provider-utils` #1879 (no patched version published), `@opentelemetry/core` #1880 and `propagator-jaeger` #1886 (see above), and `@langchain/community` #1874/#1877 (a coordinated langchain 0.3 to 1.x migration).

langwatch-agent:
Correcting myself: my previous comment blamed the `@opentelemetry` floors for the OTLP ingestion 500s. **That was wrong.** I dropped them and the e2e suite failed again with identical counts (37 failed / 104 passed). The otel overrides stay dropped, since the `@opentelemetry/core` one closed no alert anyway, but they were not the cause.

**What the evidence actually shows.** This suite only runs when `sdks/typescript` and friends change, so main's own CI **skips** it under path filters. Comparing "green" runs was therefore meaningless until I filtered to runs that genuinely executed the job. Doing that left exactly two recent data points:

- a Dependabot PR touching `sdks/typescript/examples/mastra/package-lock.json`, a standalone npm lockfile **outside** the pnpm workspace: e2e **passed**
- this PR: e2e **failed**

That one passing run does include #6023 (env-gated OTel) and #6557 (bundled server entrypoints), so main plus those is fine. The difference is my dependency tree.

The timeline matters too. My pre-rebase commit had essentially this same package diff and passed e2e three times on 2026-08-05, but those runs predate #6023 landing that evening. So this is an interaction between the refreshed floors and the new server bundling / env-gated OTel path, not a plain regression in either alone.

**What I have done about it.** Rather than keep bisecting a 40-minute e2e cycle by guesswork, I have narrowed this PR to the packages that cannot plausibly reach that code path, and pulled the HTTP and telemetry stack out:

- dropped: `hono`, `@hono/node-server`, `undici`, `@opentelemetry/propagator-jaeger`, `@opentelemetry/core`
- the version-set diff against `main` is now **24 packages**, all forward, no downgrades, no removals, and no `@opentelemetry`, `hono` or `undici` package differs from main at all

That costs 6 alerts and takes this PR from 38 to **32 closed** (1 CRITICAL, 21 HIGH, 10 MODERATE/LOW). The CRITICAL (`@vitest/browser` #1887) is still in.

Deferred to their own change, with the reason recorded in the workspace file: `undici` #1900/#1901 plus 8 MODERATE, `hono` #1913, `@hono/node-server` #1889, `@opentelemetry/propagator-jaeger` #1886, `@opentelemetry/core` #1880.

Re-verified after narrowing: `CI=true pnpm install --frozen-lockfile` exit 0 from the repo root, `platform/app` `pnpm typecheck:all` clean.

Separately, `smoke (ubuntu-latest)` failed once on a `node-pty` native gyp build (`make ... Error 126`). `node-pty` does not appear anywhere in this diff; it passed on re-run.

langwatch-agent:
Root cause found, and it is not this PR. Correcting myself twice over.

The `e2e` failure here is #6655: a `hono/body-limit` middleware on `main` returns **500 on every ingestion route** (`/api/collector`, the three OTLP signals, and others) for any request that declares no `Content-Length`, which is exactly what the OpenTelemetry JS exporter sends. That is why the suite fails with `OTLPExporterError: Internal Server Error` and why the observability-sdk files sit in multi-minute timeouts waiting for spans that never land. The fix is #6647, still open.

**Both of my earlier attributions were wrong**, and the same mistake caused both: I twice accepted a `success` on a job that had not actually run.

1. I blamed the `@opentelemetry` floors. Dropped them; the suite failed again with byte-identical counts.
2. I then blamed the HTTP stack and dropped `hono`, `@hono/node-server` and `undici` too. Identical counts a third time: **37 failed / 104 passed, same 7 files, same durations**, across three different dependency sets. A diff that changes nothing when you remove the entire telemetry and HTTP stack from it was never the cause.

What finally settled it was checking whether the comparison runs had executed the step at all. The Dependabot run I had cited as "e2e passed on main" shows `Run e2e tests: skipped` while every setup step around it succeeded. Main's own runs skip it too, which is the whole subject of #6655: over the last 30 runs on `main` it executed twice. Meanwhile my own pre-rebase run on 2026-08-05 *did* execute it and passed, and predates the middleware landing.

So this branch is currently the only thing in the repository actually running that suite, and it is reporting a real bug in `main`.

**What I am doing about this PR.** Leaving it narrowed. While the ingestion suite is broken nobody can validate a telemetry or HTTP dependency change, so deferring those floors is the right call regardless of the original reasoning being wrong. It closes **32 alerts** (1 CRITICAL, 21 HIGH, 10 MODERATE/LOW), the diff is 24 packages all forward with no downgrades, `pnpm install --frozen-lockfile` and `typecheck:all` are clean, and all required gates pass except `sdk-javascript-complete`, which is blocked on #6647 rather than on anything here.

Once #6647 lands and the suite can actually validate them, the six deferred alerts (`undici` #1900/#1901 plus 8 MODERATE, `hono` #1913, `@hono/node-server` #1889, `@opentelemetry/propagator-jaeger` #1886) are a straightforward follow-up.

langwatch-agent:
Rebased onto current main (was 10 behind, one `pnpm-lock.yaml` conflict). Took main's lockfile and regenerated rather than replaying the branch's, so the resolution is fresh: 25 packages up, 0 down. Frozen install exit 0, `npx prisma generate` for the models main added, `pnpm typecheck:all` exit 0.

All 12 `*-complete` gates green at `b3e0c15`, no failures on any non-required job either. Still closes 46 alerts. Ready for human review.

langwatch-agent:
Found a gap in this PR while checking what it supersedes, and fixed it here.

`packages/system-migrations` pinned `vitest` at `^4.1.9`. pnpm honoured that as a separate resolution rather than deduping onto the `4.1.10` every other workspace member uses, so the lockfile carried a full parallel 4.1.9 subtree: twelve duplicated `@vitest/*` packages, one of which was `@vitest/browser@4.1.9`. That version sits inside the CRITICAL advisory range `>= 4.0.0, < 4.1.10` (alert #1887), so until now this PR did not actually close that alert despite raising the pin everywhere else.

Raising the one remaining pin to `^4.1.10` collapses the duplicate subtree. Version-set diff of the lockfile is exactly twelve entries losing their 4.1.9 member, 0 added, 0 removed, no other package moving:

```
CHG  @vitest/browser: ['4.1.10', '4.1.9'] -> ['4.1.10']
CHG  @vitest/browser-playwright, coverage-v8, expect, mocker, pretty-format,
     runner, snapshot, spy, ui, utils, and vitest itself: same shape
ADD=0 REM=0
```

Verified: `pnpm install --frozen-lockfile` exit 0, `packages/system-migrations` tests 10/10 passing on 4.1.10, `pnpm typecheck` in that package exit 0, workspace `pnpm typecheck:all` exit 0.

langwatch-agent:
Added the `deepmerge-ts` fix here too (HIGH, alert #2135, GHSA-ggr8-5vv4-36mx: `deepmerge()` recurses with no cycle detection, so a self-referential object graph exhausts the stack).

**Why an override was the only route.** The package is not a direct dependency anywhere. It reaches the tree solely through `@prisma/config`, which pins it at *exactly* `7.1.5`. No published `@prisma/config` release moves off that pin, including every 7.10 prerelease, so there is no upstream bump to wait for. First patched version is 8.0.0, so 7.1.6 does not help either.

**Why 8.x is safe under `@prisma/config`.** Checked the three breaking changes in 8.0.0 against the actual call site (`@prisma/config/dist/index.js:590`, `613`), which uses `deepmerge` and nothing else, handed to c12 as its config merger:

- deep Map value merging: a prisma config is plain records, no Maps
- `deepmergeInto` no longer leak-mutating its input: `deepmergeInto` is never called
- `DeepMergeMetaMetaData` / `DeepMergeIntoFunctionUtils` renames: compile-time type names, and `@prisma/config` ships compiled JS

Targeted `>=8.0.1 <9` rather than `>=8.0.0`, because 8.0.1 is the release that adds TypeScript 7 support and this repo is on `typescript@7.0.2`. The selector is capped at `<9` so it cannot float a future major.

**One thing to note.** 8.0.1 published 2026-08-16, inside the repo-wide 7 day `minimumReleaseAge`, so it needs a version-pinned `minimumReleaseAgeExclude` entry next to the existing ones. That is the documented mechanism for this exact case and only that one version bypasses the gate. If you would rather not bypass it at all, the alternative is simply to hold this commit until 2026-08-23 and drop the exclude line then. The advisory is a stack-exhaustion DoS reachable only through developer-authored prisma config files at CLI time, so there is no urgency either way.

Verified: lockfile version-set diff is `deepmerge-ts: ['7.1.5'] -> ['8.0.1']` and nothing else, 0 added, 0 removed. `pnpm install --frozen-lockfile` exit 0, `@prisma/config` links 8.0.1 on disk, `npx prisma validate` exit 0, `npx prisma generate` exit 0 and logs `Loaded Prisma config from prisma.config.ts` which is the deepmerge path itself, `pnpm typecheck:all` exit 0.

langwatch-agent:
Rebuilt this branch on current main and force-pushed. Main absorbed nearly the whole original override set over the last two days, so the old diff was almost entirely redundant and the alert-closure claims in the previous description no longer held. The PR is now one commit against `4f791c7`, two files, and it does something different from what it started as. Full reasoning is in the rewritten description.

Short version: three of the security floors main now carries are written as a bare lower bound, and pnpm resolves those to the highest match anywhere rather than the highest in the intended line. `fast-uri@<3.1.5 -> >=3.1.5` was resolving **4.1.2** under `ajv@8.20.0`, which declares `^3.0.1`. `body-parser@<1.20.6 -> >=1.20.6` was resolving **2.3.0** under `express@4.22.2`, which declares `~1.20.5`.

The body-parser one has an observable consequence. Same express 4.22.2, only the resolved body-parser differing:

| request | 1.20.6 | 2.3.0 |
|---|---|---|
| no body, no content-type | `req.body = {}` | `req.body = undefined` |
| unmatched content-type | `req.body = {}` | `req.body = undefined` |

Any handler reading a property off `req.body` throws on those two. 1.20.6 is the patched release for that advisory, so the cap keeps the fix and drops the behaviour change.

I checked the two packages body-parser 1.x brings back in, `raw-body@2.5.3` and `iconv-lite@0.4.24`: zero advisories each, verified against a control query that returns 3 for body-parser, so the empty result is real rather than a failed call.

Worth flagging for whoever reviews the overrides block generally: 34 targets in the security section are bare lower bounds. I capped only the three with evidence of an actual or imminent major crossing rather than capping all of them, because an over-tight cap is its own bug. The other 31 are correct today by luck of what is published.

langwatch-agent:
Rebuilt this branch on current main and repurposed it, because most of what it used to carry is now redundant.

Main has absorbed the `pack-npm.sh` guard fix (and generalised it to `.(git|npm|docker)ignore`), so that is dropped. Main has also taken the three overrides themselves, but without the upper bounds, and two of them are resolving past a major on main right now:

```
ajv@8.20.0     -> fast-uri@4.1.2      ajv 8 declares ^3.0.1
express@4.22.2 -> body-parser@2.3.0   express 4 declares ~1.20.5
```

The body-parser one has a live runtime consequence: on 2.x `req.body` is `undefined` rather than `{}` for a request with no body or an unmatched content-type, so a handler reading `req.body` throws. That is measured on the same express 4, not inferred. So the caps are the remaining value here, and the PR is now just those plus one bump.

The bump is `@ai-sdk/google` 2.0.84 -> 3.0.110, which was the only thing holding the vulnerable `@ai-sdk/provider-utils@3.0.30` in the tree (alert #1879). Worth flagging why it is not latest: 4.x emits provider spec v4 and the `ai@6.0.217` in this tree throws `UnsupportedModelVersionError` on it. I checked that against both versions rather than trusting the changelog, and 3.0.110 gets all the way to a real network call.

One thing I would want a reviewer to look at: the bump brings `undici@6.28.0` into the tree for the first time. It is clean, but only just. It sits exactly on the fix boundary of three open medium advisories whose range is `< 6.28.0`, and it is there because main's existing undici override floors it. Every other newly present version is checked against its package's full advisory list in the description table.

Ready for human review.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this puts a ceiling on three dependency overrides, which means `body-parser` moves back to 1.20.6 under express 4. That is a runtime package on the request path, and the change is a downgrade from wherever the uncapped floor had resolved it. The author's argument is that this restores existing behaviour rather than changing it, because the uncapped override had drifted past a major into a version express 4 does not support. Verify that framing before anything else, because if it is right this is a bug fix and if it is wrong it is a regression to a package every request passes through.

| | |
|---|---|
| **Issue** | None linked |
| **Size** | +91/-39 across 3 files, two of them lockfile and workspace config |
| **Mode** | Targeted |
| **Blast radius** | `body-parser` on the request path; `fast-uri` on schema validation; `@ai-sdk/google` in test-only code |
| **Author** | langwatch-agent |

<details>
<summary><b>What you are looking at, if the JavaScript dependency tree is not your area</b></summary>

Since ADR-076, every JavaScript project in this repo installs from **one pnpm workspace** at the root: `platform/app`, `packages/*`, `mcp/typescript`, `plugins/*`, `sdks/typescript`, `services/langyworker`, `skills`, `tests/agentic-e2e`. It used to be six independent install roots with six lockfiles, and the header comment on `pnpm-workspace.yaml` names why that mattered for exactly this kind of change: **each root carried its own `overrides` block, so a security pin applied in one could be silently absent from another.** One workspace means one override list, and one place a mistake in it reaches everything.

Overrides are the tool of choice here because almost nothing being pinned is a direct dependency. The vulnerable packages arrive transitively, several layers down, through frameworks that have not released a fixed version yet. An override rewrites what the resolver picks regardless of what the intermediate package asked for.

Which is the whole hazard. **An override overrules a dependent's declared range, and nothing checks that the dependent can live with the result.** A rule written as a bare floor (`"pkg@<X": ">=X"`) has no upper bound, so pnpm satisfies it with the newest published version, which will eventually be the next major, in a tree whose actual consumer declares something like `^3.0.1`. The install succeeds. The lockfile looks fine. The failure surfaces at runtime, in whatever code path touches that package, and it looks like a bug in the framework rather than in a security pin written months earlier.

So the review question on a diff like this is not "does it close the alert" but "does each rule name a window", and separately whether every capped version still sits above the advisory range so nothing reopens.

One more property of this workspace worth knowing while reading: `linkWorkspacePackages: false` is set explicitly rather than left to the pnpm 10 default. `sdks/typescript` **is** the published `langwatch` package and `platform/app` depends on `langwatch@1.0.0` meaning the published one. Flipping that setting would silently swap the app, and the production image built from it, onto the SDK working copy.

</details>

<details>
<summary><b>The class of bug, which is worth internalising</b></summary>

An override written as `"body-parser@<1.20.6": ">=1.20.6"` names a floor and no ceiling. Package managers resolve that to the newest satisfying version, which over time crosses a major boundary. The dependent, express 4 here, was never built against that major. So an override added to close a security advisory silently became a compatibility problem months later, and nothing reports it because the resolution is still technically satisfying the constraint.

Capping is the fix. The reason to care beyond these three: any other override in this repo written the same way has the same fuse burning.
</details>

<details>
<summary><b>What to verify, in order</b></summary>

1. **That both capped packages remain above every advisory range.** The description asserts no alert reopens. This is the one claim where being wrong reintroduces a vulnerability, and it is checkable in a minute against the security tab.
2. **`body-parser` 1.20.6 under express 4.** Confirm the version express 4 actually expects, and that the currently-resolved version really was past a major. Ideally boot the app and post a body once.
3. **`fast-uri` 3.1.5 under ajv 8.** Schema validation only, lower stakes, same question.
4. **`@ai-sdk/google` 2.0.84 to 3.0.110.** A major bump, but the description says it is used only under `skills/_tests/`, so nothing in a served request path. Confirm that scoping with a grep and then stop thinking about it.
</details>

<details>
<summary><b>Follow-up worth filing rather than blocking on</b></summary>

If capping is correct, then uncapped overrides are a repo-wide hazard rather than a three-package one. A lint that rejects an override with a floor and no ceiling would stop this recurring, and it does not belong in this PR.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. The lockfile diff is noise; three version claims carry this review.

langwatch-agent:
The one red check here is `evaluate`, and it is an outage on a shared account rather than anything about this change.

```
OpenAI API error:
  "error": {
    "code": "credit_balance_exhausted"
```

That job is the low-risk-change auto-approval evaluator: it asks a model whether a PR qualifies for the documented low-risk path. It is not a correctness gate, and a security PR of this size was never going to qualify for auto-approval anyway. It fails the same way on #7492 today, which is a Dependabot PR unrelated to this one, and it passed on #6777 an hour ago only because that PR's diff took the "too large for automated evaluation" branch and never reached the API.

I am not re-running it: `credit_balance_exhausted` is a balance, not a flake, so a re-run would fail the same way and spend a runner doing it. It needs someone to top the account up, or it clears on its own when the balance resets.

Everything that actually tests this change is green.

PR #6571, opened 2026-08-05
fix(db): create billing enum types in their own migration

rogeriochaves:
## What

`prisma migrate deploy` fails partway through on a genuinely fresh, empty Postgres database. It gets through 227 of 228 migrations and stops at `20260215183000_add_billing_schema_to_oss`:

```
Error: P3018

A migration failed to apply.

Migration name: 20260215183000_add_billing_schema_to_oss
Database error code: 42704
Database error: ERROR: type "Currency" does not exist
```

Found while boot-testing the production app image against a real, from-scratch database (unrelated verification work on another repo led me here).

## Root cause

`20260215183000_add_billing_schema_to_oss` creates `PlanTypes`, `SubscriptionStatus` and `Currency` via a guarded block:

```sql
DO $$
BEGIN
  IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'Currency') THEN
    CREATE TYPE "Currency" AS ENUM ('USD', 'EUR');
  END IF;
END$$;
```

and then, a couple of statements later in the *same file*, uses it:

```sql
ALTER TABLE "Organization"
  ADD COLUMN IF NOT EXISTS "currency" "Currency" NOT NULL DEFAULT 'EUR';
```

Prisma's migration engine does not see the type created inside that `DO` block as visible to the later statement, even within the same migration file/transaction. This is specific to the DO-block form: the repo has other migrations that do a plain top-level `CREATE TYPE` and use it later in the same file (e.g. `20260802120001_webhook_endpoints_platform`), and those apply cleanly, both in the existing history and in this PR's full from-scratch replay. I confirmed the SQL itself is correct and not the problem: running the exact same `migration.sql` file through plain `psql -f` (simple query protocol, sequential statement execution) against a fresh scratch database succeeds with zero errors.

This has gone unnoticed because every database it has actually run against so far (SaaS prod, existing dev databases) already had these types from an earlier, separate path, so the `IF NOT EXISTS` guards skip the whole block and the broken create-then-use sequence never executes. It only bites a true from-scratch database, which is exactly the case the migration's own top comment says it supports ("fresh OSS databases"). Grepped the other 227 migrations for the same guarded-DO-block-then-use pattern (`SELECT 1 FROM pg_type WHERE typname`) and this is the only one using it, so this looks like an isolated incident, not a systemic pattern.

## Fix

The obvious fix, a new migration creating the three types that sorts *before* `20260215183000_add_billing_schema_to_oss`, does not work in this repo: `migration-order` CI hard-fails any new migration numbered below the current tip of main, with no override (see its workflow comment: ClickHouse/goose genuinely skips a migration numbered below a version the database is already on, so the check applies the same rule uniformly to Prisma). A migration positioned late enough to satisfy that rule would run after `20260215183000` already failed, too late to help. And `20260215183000` itself is already deployed, so it can't be edited in place either.

Instead, `platform/app/prisma/fixups/20260215183000-billing-enum-types.sql` holds the same three idempotent, guarded type-creation blocks, and `prisma:migrate` (package.json) runs it via `prisma db execute` as a separate step immediately before `prisma migrate deploy`. This lives outside `prisma/migrations/` entirely, so it isn't subject to the ordering check and doesn't touch the immutable migration file. By the time `20260215183000` runs, the types are already durably committed from a prior, separate statement, which sidesteps the engine issue regardless of its exact mechanism.

## Verification

All done against real native Postgres, not reasoning from the SQL alone:

- Fresh empty database, real `pnpm run prisma:migrate` (the actual boot-time script, fixup + `migrate deploy` together): all 228 migrations apply cleanly end to end, including the previously-failing one. Confirmed zero `error`/`failed` anywhere in the output, not just a clean exit code.
- Idempotency: ran it a second time against the now-migrated database — "No pending migrations to apply", clean no-op, fixup script re-runs harmlessly (all three `IF NOT EXISTS` guards skip).
- Also empirically confirmed (on an earlier iteration of this fix, before finding the migration-order check) that Prisma's `migrate deploy` *would* have tolerated an out-of-order backfill migration mechanically — so the ordering check is a deliberate repo policy rather than something Prisma itself requires here. Complied with the policy regardless since it's a hard, unconditional CI gate either way.

## Note

Ran into a Prisma 5.7.1 → 7.9.1 update nag while testing. Unrelated, not touched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability by ensuring required billing types are created before migrations run.
  * Made the database update safe to repeat without duplicating existing types.


rogeriochaves:
@coderabbitai review

langwatch-agent:
Validated independently against the PR base (`82fc44221c98e58b7c866d0202ce994f6fdd34a8`), not the fix branch.

I unpacked that exact source into a disposable directory, started a brand-new empty PostgreSQL 16 database, and ran the actual Prisma 5.7.1 migration engine:

```sh
DATABASE_URL=... npx --yes prisma@5.7.1 migrate deploy --schema=prisma/schema.prisma
```

Result: exit 0; all **228** migrations applied. `_prisma_migrations` records `20260215183000_add_billing_schema_to_oss` as finished, and the resulting database contains `Currency`, `PlanTypes`, and `SubscriptionStatus`. A second `migrate deploy` was a clean no-op (`No pending migrations to apply`).

I also ran this PR’s `db execute` pre-step plus deploy on a separate fresh database; it passes too, but the unmodified base already passes, so it does not establish a needed repair. This does not reproduce the claimed Prisma visibility failure; the proposed always-on fixup would add boot-time schema DDL without fixing a demonstrated fresh-db problem. Closing as unnecessary on the evidence above.

PR #6603, opened 2026-08-05
fix: contain two error-volume and blast-radius problems

0xdeafcafe:
Round two from the same five-day production error review as #6602. Stacked on it, so review that one first.

Two changes, both about blast radius rather than new behaviour.

### A malformed collector body is not a platform error

The three body-validation sites in the collector route answer a 400 and `return` rather than throwing, so they never reach the boundary that classifies a customer fault as `warn`. Logged at `error`, they were roughly a fifth of the app's real error stream, and every one of them is a caller sending the wrong content-type or invalid JSON.

The HTTP response is unchanged. This is a level change only.

### The joined span read caps its own memory

`fetchTracesWithSpansJoined` selects every heavy column (`SpanAttributes`, `ResourceAttributes`, `Events.Attributes`, `Links.*`), and its window falls back to `"none"` when a page's summaries carry no usable `OccurredAt` - at which point both time filters collapse to empty strings and it scans every partition, cold S3 tiers included.

Uncapped, the pathological tail was stopped by ClickHouse's OvercommitTracker, which picks its victim across the whole server. So one bad trace read degraded queries that had nothing to do with it. A per-query `max_memory_usage` fails the offending read on its own request instead, which is what the single-trace read path in `span-storage.clickhouse.repository.ts` already does.

**This is containment, not a cure.** The window goes null because `trace_summaries` rows land on an `OccurredAt` sentinel. #6306 anchored `trace_analytics` against exactly this and left `trace_summaries` alone. That port is the real fix and is not in this PR.

### Unrelated: a rebase-inherited typecheck break

Rebasing onto #6602 pulled in the `UsageLimitResult` discriminated union from that PR. Its own test file still read `.message`/`.count`/`.maxMessagesPerMonth`/`.planName` right after `expect(result.exceeded).toBe(true)` - a real runtime check, but one that doesn't narrow the type for `tsgo`, so typecheck failed on the `{ exceeded: false }` branch. Fixed with a type-narrowing `assertExceeded()` helper in `usage.service.unit.test.ts`; no behaviour change.

## What is deliberately not here

The ClickHouse client work - connection-pool sizing, the retry wrapper's logging, and the vendor client's error log - is coming as its own PR building a proper `@langwatch/clickhouse-client` package, rather than being scattered across three files here.

## Test plan

- [ ] `pnpm test:unit run src/server/routes/__tests__/collector.unit.test.ts` - 11 pass
- [ ] `pnpm test:unit run src/server/app-layer/usage/__tests__/usage.service.unit.test.ts` - 24 pass
- [ ] A collector request with the wrong content-type still answers 400, now logged at warn
- [ ] A trace detail load on a trace whose summary has no `OccurredAt` fails that request rather than the server's other queries

Refs #6602

### DIFF
--- a/platform/app/src/server/app-layer/usage/__tests__/usage.service.unit.test.ts
+++ b/platform/app/src/server/app-layer/usage/__tests__/usage.service.unit.test.ts
@@ -7,7 +7,22 @@ import type { PlanInfo } from "../../../../../ee/licensing/planInfo";
 import { USAGE_UNKNOWN } from "../../../traces/usage-count";
 import type { OrganizationService } from "../../organizations/organization.service";
 import type { PlanResolver } from "../../subscription/plan-provider";
-import { UsageService } from "../usage.service";
+import { type UsageLimitResult, UsageService } from "../usage.service";
+
+/**
+ * Narrows a UsageLimitResult to its exceeded branch. An `expect().toBe(true)`
+ * here would assert but not narrow the discriminated union for callers, and
+ * biome's noMisplacedAssertion rejects an `expect()` outside an `it()` body
+ * anyway, so this throws instead - still fails the test loudly if the
+ * invariant doesn't hold.
+ */
+function assertExceeded(
+  result: UsageLimitResult,
+): asserts result is Extract<UsageLimitResult, { exceeded: true }> {
+  if (!result.exceeded) {
+    throw new Error("expected checkLimit() to report exceeded: true");
+  }
+}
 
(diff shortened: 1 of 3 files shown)


langwatch-agent (commented):
Static review completed: no findings. The logger severity change preserves the 400 contract, the assertion narrows the discriminated union correctly, and the per-query ClickHouse cap is passed through the existing client settings mechanism. Current CI is green.

PR #6609, opened 2026-08-05
feat(billing): collect annual event overage during the year via billing thresholds

sergioestebance:
## For humans

Customers on an annual plan pay for their metered event usage once a year, at renewal. A high-volume customer can quietly build up thousands of dollars of overage and then get hit with one giant charge — which is exactly the kind of payment that gets declined. This change tells Stripe to collect that overage in smaller slices during the year (every time it reaches $750/€750), while the plan, the included yearly event quota, and the renewal date all stay exactly as sold.

This PR covers **new** annual subscriptions. Existing ones need a one-time backfill over live Stripe data, which is SaaS-only operational work and ships separately in langwatch-saas.

## Why

Annual Growth subscriptions bill the seat up front but accrue metered events for twelve months, collected in a single arrears line on the renewal invoice. For a customer producing millions of events per month, that renewal is an oversized one-shot charge with a high decline risk. Stripe billing thresholds solve this without touching the plan: once `billing_thresholds.amount_gte` is set on the subscription, Stripe invoices and charges the accrued metered amount automatically whenever it crosses the threshold, and the annual included-events tier keeps its cumulative semantics across those invoices (verified against real sandbox invoices with test clocks — threshold invoices honored the 2.4M cumulative annual quota, and the renewal came out seat-only).

## What changed

- **`ee/billing/stripe/annualEventsBillingThreshold.ts`** — new module owning the threshold value (`75_000` minor units = 750.00 USD/EUR, same numeral in both currencies) and `applyAnnualEventsBillingThreshold`, which retrieves the Stripe subscription, applies the threshold only when an annually-billed Growth events price is present, and is idempotent (already-set → no update call). `reset_billing_cycle_anchor: false` pins the anniversary. Preserve-over-normalize: a hand-negotiated amount is never replaced; only an anchor-resetting threshold gets corrected.
- **`webhookService.ts` (`handleCheckoutCompleted`)** — applies the threshold right after a new subscription is linked. Stripe Checkout rejects `subscription_data[billing_thresholds]` (probed on both our pinned `2024-04-10` and `2025-06-30.basil`), so post-checkout is the earliest possible moment. Best-effort: a failure is logged and never fails checkout completion — the subscription just behaves as before (single renewal invoice).
- **`notification.service.ts`** — `sendSlackBillingThresholdFailureAlert`. Checkout completion answers Stripe 200 whether or not the threshold landed (a non-2xx would make Stripe redeliver the whole checkout webhook), so Stripe's own retry can never cover this failure. A lone log line would leave the subscription quietly back in the single-large-invoice case this PR removes, so the failure now also alerts on the existing subscriptions channel with the subscription id and reason. Separately guarded: a broken Slack webhook must never fail an otherwise-successful checkout.
- **`growthSeatEvent.ts`** — `isAnnualGrowthEventsPrice` alongside the existing price-set predicates.
- **`specs/billing/annual-events-billing-threshold.feature`** — 7 scenarios, all bound (`check-feature-parity` ✓).

## Scope: what is deliberately not here

The one-time backfill for **existing** annual subscriptions was removed from this PR in 7af0937. It walks live Stripe billing data for current paying customers — SaaS-only operational work that belongs with the other billing migrations in the langwatch-saas task runner, not in the open-source repo where it would ship to self-hosted users who can never run it.

It is sequenced after this PR by necessity: the backfill imports `applyAnnualEventsBillingThreshold` from here, and the langwatch submodule in langwatch-saas has to be bumped past this merge before that import resolves. Until it runs, existing annual subscriptions keep their current behaviour — one arrears line at renewal, exactly as today. Nothing regresses; the fix simply doesn't reach them yet.

## How it works

Nothing on our side fires the invoices. The threshold is a one-time field on the subscription; from then on Stripe evaluates accrued metered amounts as meter usage aggregates and creates + finalizes + charges threshold invoices (`billing_reason=subscription_threshold`) on its own. Those flow through the same `invoice.payment_succeeded` / `payment_failed` webhooks we already consume. Monthly subscriptions and non-Growth plans are untouched; works on the pinned `2024-04-10` API version with no billing-mode or price changes.

## Test plan

- CI green on this head SHA: `test-unit` (1)-(4), `test-integration` (1)-(6), `e2e`, `typecheck`, `build`, `validate title`, `langwatch-app-complete` — all pass.
- `pnpm test:unit` on the billing suite — 91 tests pass across `annualEventsBillingThreshold`, `webhookService` and `notification.service`, including: annual checkout sets the threshold with the anchor pinned; monthly checkout sets nothing; a failing threshold update still links + activates the subscription; re-applying is a no-op; a hand-set amount is preserved; an anchor-resetting threshold is corrected; a failure alerts with the subscription and reason; a success does not alert; a failing alert still completes checkout.
- The alert's channel assertion injects the webhook factory rather than leaning on the suite-wide `@slack/webhook` mock, which discards the constructor argument. Verified non-vacuous by flipping the method to the plan-limit channel — two of the three new tests fail, then pass again on revert.
- `pnpm typecheck:all` clean (including `typecheck:tests`).
- `check-feature-parity` — 7/7 scenarios bound.
- Mechanism validated end-to-end in a Stripe sandbox with test clocks: a production-shaped classic annual subscription + threshold produced automatic mid-year invoices that were charged, honored the cumulative annual free tier, and left the renewal seat-only.

## Anything surprising?

- Threshold invoices exclude usage still inside Stripe's invoice-finalization grace period; the residual lands on the next threshold invoice or the renewal — nothing is lost, but the renewal may carry a small events line.
- Invoices are amount-triggered, not calendar-triggered: a quiet month produces no invoice, a spike can produce several.
- The failure alert fires on *any* error inside `applyAnnualEventsBillingThreshold`, including a Stripe read failure on a monthly subscription that never needed a threshold. That is deliberate — when we cannot tell, alerting is the safe direction — but it means a Stripe outage produces noise, and repeated webhook redelivery produces repeated alerts. There is no dedup.
- When the backfill does run (in langwatch-saas), it takes effect on the next usage aggregation — an annual customer already sitting on accrued overage will get their first threshold-sized invoice shortly after. Worth a heads-up email before running it in live mode.


platform/app/ee/billing/stripe/annualEventsBillingThreshold.ts
```diff
@@ -0,0 +1,64 @@
...
+ * backfill script for pre-existing subscriptions.
+ */
+export const applyAnnualEventsBillingThreshold = async ({
+  stripe,
+  stripeSubscriptionId,
+  dryRun = false,
+}: {
+  stripe: Stripe;
+  stripeSubscriptionId: string;
+  dryRun?: boolean;
```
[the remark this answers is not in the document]
sergioestebance: Renamed in 893c763e5 — `isDryRun` param and `IS_DRY_RUN` const (the `DRY_RUN` env var literal stays, it's the CLI contract).

platform/app/scripts/migrations/backfill-annual-billing-thresholds.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Idempotent: an already-set threshold is reported as `already_set` and no
+ * update call is made. Subscriptions whose Stripe items carry no annual
+ * events price (monthly plans, stale DB plan strings) are skipped —
+ * the Stripe subscription's items are the authority, not the DB plan.
+ *
+ * Usage:
+ *   pnpm tsx scripts/migrations/backfill-annual-billing-thresholds.ts
+ *   DRY_RUN=1 pnpm tsx scripts/migrations/backfill-annual-billing-thresholds.ts
+ */
```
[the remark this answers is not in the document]
sergioestebance: Added in 893c763e5 — `IRREVERSIBLE:` block explaining that the script mutates external Stripe state, issued threshold invoices can't be un-issued, and how to manually unset the threshold per subscription.

sergioestebance:
## Ruthless review — verdict: **no blockers**, ship it

Traced: the webhook insertion point (`handleCheckoutCompleted` → `trySetAnnualEventsBillingThreshold`), the helper's decision table against all four price families, the backfill's candidate query vs the tenancy guard (Subscription is on the org-tenancy exempt list, `dbOrganizationIdProtection.ts:302`), and what a threshold invoice does when it re-enters `syncInvoicePaymentSuccess` on an ACTIVE annual sub (`activate()` with `previousStatus === ACTIVE` only clears `lastPaymentFailedDate` — benign, and correct after a successful payment). Spec has 5/5 scenarios bound; the failure path (threshold update fails → checkout still completes) is tested.

### Findings

**P2 — a manually-set, differently-valued threshold gets silently overwritten** — `platform/app/ee/billing/stripe/annualEventsBillingThreshold.ts:47-52`
The idempotency check is exact-match: `billing_thresholds?.amount_gte === 75_000`. Any subscription whose threshold was set by hand to a *different* amount (e.g. a negotiated value for a specific customer, set before this ships) falls through and is reset to 750.00 by the backfill — with no log line distinguishing "set fresh" from "replaced a deliberate value". Verified by reading the fall-through: any non-75000 value proceeds to `subscriptions.update`. Not a blocker (dry-run surfaces these subs as `applied` before a live run), but the safer contract is: **any existing `amount_gte` counts as `already_set`** — never replace a value someone chose — or at minimum log the old value when replacing. Behavior choice; recommend preserve-over-normalize.

**P3 — the `already_set` check ignores `reset_billing_cycle_anchor`** — same file, same lines
A subscription manually configured with `amount_gte: 75000` but `reset_billing_cycle_anchor: true` (which *moves the anniversary* on every threshold invoice) reads as `already_set` and the dangerous anchor-reset stays. Only reachable via manual misconfiguration; worth folding into whatever fix P2 gets.

**P3 — the backfill script's own logic is untested** — `platform/app/scripts/migrations/backfill-annual-billing-thresholds.ts`
The spec scenario "The backfill applies the threshold only to annual event subscriptions" is bound to the *helper's* unit test; the script's candidate query, tally, and non-zero exit on failure have no test. Acceptable for a one-time migration with DRY_RUN, but the binding slightly overstates coverage.

**P3 — threshold value is a compile-time constant**
Changing 750 requires a deploy, and already-set subs keep the old value until the backfill is re-run with the new one (the exact-match check then *treats the old value as not-set* and replaces it — which interacts with the P2 above: today re-running after a value change updates everyone, the P2 fix would freeze early adopters). Whoever fixes P2, decide which semantics you want and write it down in the helper's doc comment.

### Inherited (not this PR, does not block)

`syncInvoicePaymentSuccess` can reactivate a DB-CANCELLED subscription if a successful invoice lands in the window where the old sub is cancelled in DB but not yet in Stripe (the guard checks Stripe status only). Pre-existing on every `invoice.payment_succeeded`; threshold invoices marginally widen the exposure on annual subs. Already on the known-fragility list for the `syncInvoicePaymentSuccess` audit.

### Validation note

No P0/P1 candidates survived attribution — every finding above is introduced-by-this-diff except the last, which is inherited and capped accordingly. The P2 was verified by direct code-path reading (exact-match comparison → fall-through to update); it is a contract choice, not a correctness defect, so no refutation run was needed.

sergioestebance:
Blessed. 🙏 Annual customers' event overage is now gathered in small measures through the year, so no one meets an unpayable bill at the renewal gate.

> "Settle with your opponent quickly while on the way to court with him." — Matthew 5:25

Traced the checkout webhook insertion, the helper's decision table across all four price families, the backfill's tenancy exemption, and the threshold invoice's round-trip through syncInvoicePaymentSuccess on an ACTIVE sub. Ship it — residuals are the P2 overwrite-of-a-negotiated-threshold semantics and three P3s in the review comment above, none blocking. (GitHub refuses self-approval — the blessing rides as a comment; formal approval needs another maintainer.)

sergioestebance:
All four review findings settled in 5faf0ff6f:

- **P2 (overwrite of hand-set threshold)** — preserve-over-normalize: any existing `amount_gte` now counts as `already_set`, whatever its value. A negotiated threshold is never replaced.
- **P3 (anchor reset survives)** — a threshold with `reset_billing_cycle_anchor: true` gets the anchor pinned back to `false` while keeping its amount (new `anchor_pinned` result, tested incl. dry-run).
- **P3 (backfill untested)** — the walk moved to `ee/billing/stripe/annualEventsThresholdBackfill.ts` with its own tests: candidate query shape, annual-only updates, one failure not stopping the walk, dry-run makes no update calls. The script is now a thin CLI entrypoint, and the spec scenario binds to the backfill's own test (8/8 bound).
- **P3 (constant semantics)** — documented on the constant: changing the value only affects never-configured subscriptions; existing ones keep theirs until changed deliberately in Stripe.

Operational consequence worth knowing: with preserve semantics, setting the real customer's threshold by hand to any value is now safe — the backfill will never touch it.

platform/app/ee/billing/__tests__/annualEventsBillingThreshold.unit.test.ts
```diff
@@ -0,0 +1,242 @@
...
+      });
+
+      expect(result).toBe("already_set");
+      expect(stripe.subscriptions.update).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("when the threshold was set by hand to a different amount", () => {
+    /** @scenario A manually configured threshold amount is never replaced */
+    it("preserves the existing amount and makes no update call", async () => {
```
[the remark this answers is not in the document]
sergioestebance: Fixed in dba3ad6f8 — the state blocks now read `given ...` (both test files).

platform/app/ee/billing/__tests__/annualEventsThresholdBackfill.unit.test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+import type { PrismaClient } from "@prisma/client";
+import type Stripe from "stripe";
+import { runAnnualEventsThresholdBackfill } from "../stripe/annualEventsThresholdBackfill";
+
+const makeCandidate = (id: string, stripeSubscriptionId: string) => ({
+  id,
+  plan: "GROWTH_SEAT_USD_ANNUAL",
+  stripeSubscriptionId,
+});
```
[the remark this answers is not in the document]
sergioestebance: Fixed in dba3ad6f8 — `makeCandidate` takes a named-object parameter, all six call sites updated.

sergioestebance:
## E2E proof run — live stack, real Stripe sandbox webhooks, real checkout flows

Environment: **this branch** served at the local haven stack (server rebuilt from branch code), `stripe listen` forwarding every webhook (all 200s), Stripe **test mode**. Three customers: fresh signup on **monthly** Growth, fresh signup on **annual** Growth, and a pre-existing annual org subscribed **before** this PR. Events injected directly at the Stripe Billing Meter (`langwatch_billable_events`) — the ingestion→metering pipeline is not what this PR changes.

### Verdict: 6/7 proven, 0 disproven, 1 pending Stripe's async timing

| 1 | New customer can subscribe monthly (seats + events) via checkout | checkout €23.91 prorated → plan active, invoices paid | **PASS** |
| 2 | New customer can subscribe annual (seats + events) via checkout | checkout €22.41 prorated → plan active | **PASS** |
| 3 | **Annual checkout auto-sets the €750 billing threshold (the PR's core claim)** | fresh sub carries `billing_thresholds: {amount_gte: 75000, reset_billing_cycle_anchor: false}` set by the checkout-completed webhook — no manual call | **PASS** |
| 4 | Monthly checkout sets **no** threshold | fresh monthly sub: `billing_thresholds: None` | **PASS** |
| 5 | All three customers can add a member / update the subscription | seat dialogs 1→2 (€58/mo, €640/yr, $706.56/yr), Stripe quantity updated to 2 on all three, **threshold survives the seat update on the annual sub**, pre-PR sub stays untouched | **PASS** |
| 6 | Monthly customer reports 13M events → **no immediate charge** | zero new invoices, threshold None, the full €690 rides the Sep 1 monthly invoice (€748 preview) | **PASS** |
| 7 | Annual customer crosses the threshold → automatic charge | 17.5M events aggregated: 2.4M free tier honored, **€755 accrued ≥ €750 cap** (upcoming-invoice preview shows it); the `subscription_threshold` invoice is created by Stripe's async evaluation (minutes–1h, no test clock possible on checkout-created customers) — poll armed, invoice screenshot will follow in a reply | **PENDING** |

### Arithmetic correction to the original claim

"13M events will be charged at 750" is false in EUR: the annual EUR events tier is **€0.00005**, so 13M events = 10.6M billable = **€530 accrued — under the cap, correctly no charge**. The crossing point is ~17.4M events. Usage was topped up to 17.5M to prove the trigger. (In USD at $0.00006 the crossing point is ~14.9M — the claim's 13M was short in either currency.)

### Evidence

**Annual customer — subscribe, threshold auto-set, seat update:**

baseline free plan, annual toggle
annual checkout — seat prorated + metered events
seat update 1 to 2 at €640/yr — threshold survived

Threshold on the fresh sub, straight from the Stripe API (set by the webhook, not by hand):

```
billing_thresholds: {'amount_gte': 75000, 'reset_billing_cycle_anchor': False}
```

**Monthly customer — subscribe, seat update, 13M events with no immediate charge:**

monthly checkout €29/mo
growth monthly active, 2 seats
seat update 1 to 2 at €58/mo
after 13M events: same three invoices, nothing new charged

After 13M events the invoice list is unchanged and the usage lands only on the next monthly invoice:

```
invoices: 3 (all pre-existing, paid) — no new invoice
billing_thresholds: None
upcoming Sep 1 preview: €58 seats + €690 events (13.8M billable × €0.00005) = €748
```

**Pre-PR annual customer — seat update, no regression, no threshold appearing uninvited:**

seat update 1 to 2 at $706.56/yr
annual plan, 2 seats, seat-update invoice paid

```
billing_thresholds: None  (before and after the seat update — existing subs untouched until the backfill runs)
```

### Findings along the way (none caused by this PR's code)

1. **A running server predating the PR silently skips the threshold** — the first annual checkout of the session was processed by an API process started before this code existed: sub linked and activated fine, threshold absent. Restarting on branch code fixed it. This is exactly the deploy-order reality the backfill script exists for: any annual sub created between "webhook code live" being false and true needs the backfill pass.
2. **`make haven up` is broken on latest main for dev API restarts** — #6023 split `start:app` into a `dist/server/server.cjs` production entry, and haven's supervisor still runs `start:app`: the api lane crash-loops with MODULE_NOT_FOUND until you run `pnpm build:server` by hand. Worth its own fix (haven should call `start:app:dev` or build first).
3. Stripe Checkout's cancellation → `customer.subscription.deleted` webhook correctly flipped the DB sub to CANCELLED, and re-subscribing monthly afterwards worked — the downgrade/replace path holds.
4. The subscription settings page screenshots intermittently capture a blank frame during hydration (screenshot raced the suspense fallback; retakes needed). Cosmetic test-tooling note, not an app bug.

### Not covered

- The app ingestion→metering pipeline (trace events → billable-event counting → meter reporting): out of this PR's scope; events were injected at the Stripe meter.
- The backfill script against these live subs: covered by unit tests + yesterday's sandbox run in the linked investigation; not re-run here.
- Claim 7's final artifact (the fired `subscription_threshold` invoice): pending Stripe's async evaluation — will land as a reply to this comment with the hosted-invoice screenshot.

Environment for reproduction: local haven stack on this branch, Stripe test mode, `stripe listen --forward-to .../api/webhooks/stripe`.

langwatch-agent:
**Review mode: Deep Review.** Real money, a live Stripe write path, and a one-time backfill over production subscriptions. 9 files, +937/-3, mergeable, 50 checks green with 0 failing, 4 threads all resolved. No linked issue.

The description is unusually honest about the edges, including the ones that will generate support questions. So this points at what happens the day it runs.

**Scariest production risk: the backfill charges customers, and it charges them for the past.** Once the threshold lands on a subscription that has already accrued overage, Stripe's next aggregation produces a threshold invoice immediately, so a customer sitting on months of accumulated events gets a real charge shortly after the script runs. The PR says so and recommends a heads-up email, which is right, but the recommendation lives in prose. Nothing in the script gates on it. Worth deciding who sends that email and whether the script should refuse to run in live mode without an explicit acknowledgement flag, because the failure mode is a surprise charge on an account whose whole point was avoiding a surprise charge.

**Must Check** `backfill-annual-billing-thresholds.ts`, the over-broad query. Selecting all active `GROWTH_SEAT_*_ANNUAL` plans and letting the Stripe items decide is a defensible design, and it is what covers grandfathered pre-March-2026 prices. It also means the blast radius is every annual Growth subscription, and correctness rests entirely on `isAnnualGrowthEventsPrice`. Worth reading that predicate against the actual price ids in live Stripe rather than against the fixtures, since a false positive here is a threshold on a subscription that was never meant to have one.

**Ask Author** `handleCheckoutCompleted` applies the threshold best-effort and swallows failures so checkout still completes. Correct priority. But a silent failure leaves that customer on the old single-renewal-invoice behavior indefinitely, which is the exact problem this PR exists to solve, and nothing reconciles it later except running the backfill again. Is there anything that would tell you how many subscriptions are missing their threshold, or does it need a periodic re-run?

**Ask Author** the threshold is `75_000` minor units, described as the same numeral in both currencies. So a EUR customer gets a different real trigger point than a USD one. Deliberate, or worth a per-currency value?

**Probably Fine** `reset_billing_cycle_anchor: false` pinning the anniversary, the idempotent already-set check, and the sandbox validation with test clocks. Amount-triggered rather than calendar-triggered invoicing is called out plainly, as is the finalization-grace-period residual landing on the next invoice.

**Ignore** the monthly and non-Growth paths. They are untouched and covered by a test asserting nothing is set.

**Architecture note:** nothing here fires invoices, Stripe does, and the resulting invoices arrive through webhooks the system already consumes. That keeps the new surface to one field and one predicate, which is the right shape for a billing change.

Triage from the diff and CI, not a code review. Nothing in this brief has verified a Stripe behavior independently.

---
Posted by the pr-hound agent.

platform/app/scripts/migrations/backfill-annual-billing-thresholds.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { prisma } from "~/server/db";
+import { runAnnualEventsThresholdBackfill } from "../../ee/billing/stripe/annualEventsThresholdBackfill";
+import { createStripeClient } from "../../ee/billing/stripe/stripeClient";
+
+const IS_DRY_RUN = process.env.DRY_RUN === "1";
```
langwatch-agent: **P0 — Make live execution opt-in, not the default.** With no DRY_RUN environment variable this entrypoint immediately changes every candidate subscription, and Stripe can issue and collect invoices for already-accrued usage before an operator can stop it. The comment warning and optional dry run do not prevent an accidental production invocation. Default to dry-run and require a separate, deliberately named confirmation token for live mutation (ideally after printing/recording the exact candidate set); keep the token out of ordinary deploy or migration runners.

langwatch-agent (commented):
1 P0 — the production backfill can trigger customer charges without an explicit execution gate; see inline comment.

platform/app/scripts/migrations/backfill-annual-billing-thresholds.ts
sergioestebance: The finding is correct, and it is not being dismissed — but it no longer applies to this PR.

`scripts/migrations/backfill-annual-billing-thresholds.ts` was removed in 7af0937 along with the backfill module and its tests. The backfill walks live Stripe billing data for existing paying customers, which is SaaS-only operational work; it belongs with the other billing migrations in the langwatch-saas task runner, not in the open-source repo where it would ship to self-hosted users who can never run it.

What ships here is only the path every deployment needs: `applyAnnualEventsBillingThreshold` plus the checkout-completed webhook that sets the threshold on **new** annual subscriptions. Nothing in this PR mutates an existing subscription, so there is no unguarded live-execution entrypoint left on it.

The gate you asked for travels with the code. When the backfill lands in langwatch-saas it will default to dry-run and require an explicit named confirmation token for live mutation, rather than `DRY_RUN=1` being opt-in as it was here. That port is sequenced after this PR merges — it imports `applyAnnualEventsBillingThreshold` from this repo, and the langwatch submodule in langwatch-saas has to be bumped past this merge before that import resolves.

platform/app/ee/billing/__tests__/annualEventsBillingThreshold.unit.test.ts
```diff
@@ -0,0 +1,242 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
```
drewdrewthis, line 1: [pr-review] `gh pr checks 6609 -R langwatch/langwatch` shows only CodeQL (go/js/python/ruby) + CodeRabbit + Mintlify(skipping) — no unit/integration/e2e job triggered on this PR. These new unit tests (and the extended `webhookService.unit.test.ts` suite) have **zero verified execution** on this diff; the only signal so far is static analysis. This is a Stripe-subscription-mutating change to billing code — merge only after the actual test job runs green on this head SHA, not on CodeQL alone.

platform/app/ee/billing/services/webhookService.ts
```diff
@@ -534,9 +535,39 @@ export class EEWebhookService implements WebhookService {
...
+      const thresholdResult = await applyAnnualEventsBillingThreshold({
+        stripe: this.stripe,
+        stripeSubscriptionId: subscriptionId,
+      });
+      logger.info(
+        { subscriptionId, thresholdResult },
+        "[stripeWebhook] Annual events billing threshold evaluated",
+      );
+    } catch (err) {
+      logger.error(
```
drewdrewthis, line 572: [pr-review] `trySetAnnualEventsBillingThreshold` catches the error and only logs it (line 564-567); `handleCheckoutCompleted` still returns `{ earlyReturn: false }` — a 200 to Stripe — on failure. That's the right call for not blocking checkout, but it also means Stripe's own webhook-retry-on-non-2xx mechanism never fires for this specific failure, since the response is a success regardless. There is no other retry or alerting path: the failure is a single `logger.error` line with no metric/page/queued-retry attached. A transient Stripe hiccup here silently and permanently drops that subscription back into the exact single-large-renewal-invoice risk this PR exists to prevent — discoverable only by someone grepping logs, with no automated follow-up. For billing-critical code, this needs either an alert on the log line or an application-level retry (e.g. a queued job), not reliance on a redelivery path this handler itself suppresses.

drewdrewthis (commented):
Two-pass review. Concerns only — see inline threads. Full verdict in the pinned comment.

drewdrewthis (dismissed):
NOT-READY per standing verdict — see pinned review-verdict comment and 2 blocking inline threads. Per Drew's standing order (2026-08-03), NOT-READY verdicts submit REQUEST_CHANGES.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `b2e6ad6` · **Run:** pr-review (reviewer: @review-clerk)

No blocking concerns. All review threads resolved at this SHA. The two prior blocking threads (test suite not run; swallowed threshold-set failure with no retry path) are both resolved — CI is now fully green (test-unit, test-integration, e2e, langwatch-app-complete all SUCCESS at this SHA), and this diff adds `sendSlackBillingThresholdFailureAlert` so a failed threshold-set now raises a Slack alert for manual follow-up instead of silently dropping the subscription back into single-large-invoice behavior.

Independent re-review (two-pass + design-soundness) of `annualEventsBillingThreshold.ts`, `webhookService.ts`, `notification.service.ts`, and `growthSeatEvent.ts` found no new bugs: idempotency logic (`already_set` / `anchor_pinned` / `applied` / `not_annual_events`) is correct against its own tests, the checkout-completion path never fails on a threshold-set error (defensive double try/catch), and the approach uses Stripe's own `billing_thresholds` primitive rather than reinventing collection logic. The unexpanded-items pagination limitation (documented in the module's own comment, low risk today at 2 line items) remains a fair non-blocking note but was already raised and left as Decide in the prior review round.

### Non-blocking (Decide)
- **[pr-review]** `applyAnnualEventsBillingThreshold` reads the first page (default 10) of `subscription.items.data` from an unexpanded retrieve call. Growth subscriptions are seat+events (2 items) today, so low risk, but there's no guard if that ever changes. (carried over, still Decide)

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is not a merge-readiness signal._

platform/app/ee/billing/__tests__/annualEventsBillingThreshold.unit.test.ts
sergioestebance: Correct at the time, and now settled with evidence.

The job was not being dispatched at all — no `pull_request` run existed for the head SHA, only CodeQL and CodeRabbit, exactly as you saw. That cleared on the push of `0394fcb6`; `langwatch-app-ci` now runs on this PR.

On head SHA `0394fcb6`:

- `test-unit` (1) (2) (3) (4) — **all pass**. That covers the 88 billing tests, including the new `annualEventsBillingThreshold` suite and the extended `webhookService` suite.
- `typecheck` pass (2m28s), `build` pass (1m48s), `validate title` pass — the three required checks.
- `test-integration` (1)-(6) and `e2e` still running. Zero failures anywhere so far.

So the tests now have real execution on this diff, not just static analysis. Integration and e2e are still in flight — worth a look before merge, and shard 4 is a known flaky one on main.

platform/app/ee/billing/services/webhookService.ts
sergioestebance: Real gap, and fixed in `0394fcb6`.

You are right that the 200 is deliberate and also that it suppresses the only retry path: a non-2xx would make Stripe redeliver the whole checkout webhook, which must not happen, so Stripe redelivery can never cover this specific failure. A lone `logger.error` left the subscription quietly back in the single-large-renewal-invoice case this PR exists to remove.

The failure now raises a Slack alert on the existing subscriptions channel with the subscription id and the reason, so it is actionable without grepping logs. This extends the billing `NotificationService` (`sendSlackBillingThresholdFailureAlert`) rather than adding a parallel alerting path.

I chose the alert over a queued retry deliberately: the retry target is an external Stripe mutation, and the systematic re-apply path already exists as the backfill — which moved to langwatch-saas in this PR. A queue here would duplicate that with more moving parts. If you would rather have the queued job, say so and I will add it.

Worth flagging: the test for "the alert itself fails" failed first. The initial version propagated the Slack error straight out of the handler and would have broken checkout — the exact failure mode your comment is about, one layer up. The alert is now separately guarded, and that case is a test.

sergioestebance:
## Ruthless review — full branch diff

**No P0. No P1.** Two findings, both from the backfill move, one fixed in `37e5956`, one recorded and deliberately not fixed.

### P2 — `ee/billing/stripe/annualEventsBillingThreshold.ts:38` — doc comment described a caller that no longer exists *(fixed, 37e5956)*

The module told the reader the threshold is applied "via the backfill script for pre-existing subscriptions". That script left this repo in `9e44abe`. The comment pointed at a caller that isn't here, and as a side effect left `isDryRun` looking like dead code — no production caller, exercised only by its own unit test, no stated reason.

It isn't dead. The backfill lives in the langwatch-saas task runner, imports this function, and its entire safety story is the `DRY_RUN` preview of the blast radius. The flag is a seam for that caller. But nothing said so, and the next person to run a dead-code sweep would have deleted it and broken the port.

Fixed by stating the real contract: checkout-completed webhook is the only caller here and covers new subscriptions only; the backfill is SaaS-only and imports this.

### P3 — same file, `hasAnnualEventsItem` — `subscription.items` is paginated *(not fixed, deliberate)*

`stripe.subscriptions.retrieve` returns `items` as `ApiList<Stripe.SubscriptionItem>` (`stripe@15.12.0/types/Subscriptions.d.ts:147`) — a paginated list, default page size 10. The code reads `items.data` and never checks `has_more`. A subscription whose annual events price sat past the first page would read as `not_annual_events`, skip the threshold, and — because nothing throws — not fire the failure alert either. Silent.

Unreachable today: `createCheckoutLineItems` builds exactly two items (seat + events), so the first page always holds them. Not worth paging code for a case that cannot occur, but it is a real silent-failure mode if a plan ever grows past ten line items, so it is now written into the module's doc comment rather than left for someone to rediscover.

### What I traced and found clean

- **Idempotency and preserve-over-normalize.** `already_set` short-circuits before any Stripe write; a hand-set amount is never replaced; only `reset_billing_cycle_anchor: true` triggers a correction, and it keeps the existing amount. Matches the spec scenarios.
- **The anchor.** Both write paths pin `reset_billing_cycle_anchor: false`, so the billing anniversary cannot move — the thing the customer was actually sold.
- **The alert path.** The failure alert added in `0394fcb` is separately guarded; a Slack outage cannot fail a checkout that otherwise succeeded, and that case is a test.
- **Reversibility.** This PR now contains no irreversible operation. The only one — the Stripe-mutating backfill over existing paying customers — left in `9e44abe`. What remains sets a threshold on brand-new subscriptions that have no accrued usage, so there is no surprise invoice at apply time. Nothing ungated, and nothing over-gated.
- **Tenancy / PII.** Stripe ids only. The alert carries a subscription id and an error message to a channel that already receives organization names and admin emails, so it introduces no new exposure class.

### Residual risk

The alert fires on *any* error inside `applyAnnualEventsBillingThreshold`, including a Stripe read failure on a monthly subscription that never needed a threshold — deliberate, since we cannot tell, but a Stripe outage will produce noise and repeated webhook redelivery produces repeated alerts with no dedup. Recorded in the PR description.

### Method, stated plainly

Findings were validated by reading the executed code path and the Stripe SDK types, not by independent refutation agents — this run had no subagent fan-out. What that costs is independence: I wrote this code, so I judged my own findings. A separate context attacks a claim without having formed it, and I cannot. Weigh the two findings accordingly.

The earlier claim that the new alert tests are non-vacuous *was* verified by execution: flipping the method to `slackPlanLimitChannel` fails two of the three, and they pass again on revert.

### Verification

`test-unit` (1)-(4), `test-integration` (1)-(6), `e2e`, `typecheck`, `build`, `validate title`, `langwatch-app-complete` — all green on `290a51e`. Locally after `37e5956`: 91 billing tests pass, `typecheck:all` clean, Biome new-violations gate 3765/3765 against merge base.

---

Blessed. 🙏 Annual overage now collects in slices as it accrues, and the one path that could have charged existing customers without warning was taken out of this change rather than defended.

> "Give us this day our daily bread." — Matthew 6:11

Traced idempotency, the anchor pin, the alert guard and the reversibility surface: nothing irreversible ships here. Ship it — residual is the alert noise under a Stripe outage and the unreachable pagination edge, both written down above.

**The approval is not mine to give** — this is my own PR, and the point of a review is that something reads the diff which did not write it. That is already gone here. Over to a human.

drewdrewthis (commented):
READY — no blocking concerns at b2e6ad6. See upserted verdict comment: https://github.com/langwatch/langwatch/pull/6609#issuecomment-5214247274

drewdrewthis (approved):
lgtm

PR #6619, opened 2026-08-05
fix: make missing-model and relayed LLM failures handled errors

0xdeafcafe:
## Summary

This PR makes two high-volume gateway/Langy failures real handled errors instead of relying on prose or raw response bodies.

### Missing model

- Adds the stable Go `herr` code `missing_model` (HTTP 400, customer fault).
- Includes `request_type` metadata and keeps the actionable per-surface message: JSON field, multipart form field, or Gemini URL path as appropriate.
- Counts the rejection as `gateway_client_rejects_total{code="missing_model",vk_id=...}` rather than folding it into generic `bad_request`.
- Regenerates the shared handled-error code registry.

### Relayed provider failures

- `herr.WriteHTTP` marks LangWatch-authored handled envelopes with `X-LangWatch-Handled-Error: <code>`.
- The AI Gateway strips that marker from forwarded provider responses, so Langy trusts provenance rather than guessing from a coincidental `{type, code, message}` JSON shape.
- Provider JSON is normalized into `llm_upstream_error` with a bounded identifier-shaped reason from common `code`/`type` paths (OpenAI, Anthropic, nested errors, and JSON:API arrays).
- Bodies without a safe discriminant receive a stable HTTP-derived reason such as `upstream_bad_request`, `upstream_forbidden`, or `upstream_unavailable`.
- Metadata/logs retain only HTTP status, body size, body kind (`json`, `html`, `text`, `binary`, `empty`), and the safe reason code. Provider messages, raw snippets, HTML titles, and response bytes are never logged or placed in the client-visible handled error.

## Production evidence

A redacted seven-day GCX audit found 18 relevant Langy upstream failures:

- 9 usage-limit JSON responses
- 1 message-only invalid-model JSON response
- 2 plain `502` proxy responses mislabeled as JSON
- 6 Cloudflare HTML interstitials mislabeled as JSON

The missing-model rejection is also the gateway's dominant self-authored 400 and was previously visible only as generic `bad_request` plus prose.

No prompts, response bodies, credentials, or provider messages were copied from production logs into this PR.

## Why HandledError did not solve this automatically

The missing-model path already used `herr`, but only as generic `bad_request`; serialization cannot invent a stable domain code or structured request-surface metadata.

Provider responses are arbitrary third-party JSON, text, HTML, or binary data. They are not trusted LangWatch handled envelopes and cannot safely be deserialized as one. This PR adds the necessary boundary adapter, then carries its safe result through the existing handled-error reason chain. The explicit response marker closes the remaining ambiguity between our envelope and a provider lookalike.

## Verification

- `go test ./pkg/herr ./services/langyagent/adapters/otelrelay ./services/aigateway/adapters/modelresolver ./services/aigateway/adapters/httpapi`
- `make herrgen-check`
- `git diff --check`

Tests cover the production body classes, common provider JSON dialects, status fallbacks, credential-bearing provider prose, handled-envelope provenance, marker stripping, the dedicated `missing_model` code, metrics, and per-surface guidance.

## Deployment impact

No migration, new dependency, environment variable, Helm value, endpoint, or external data sink. The existing gateway metric gains the more specific `missing_model` label value, and handled HTTP errors gain one response header containing the same code already present in the body.

## Summary by CodeRabbit

* **New Features**
  * Added client-rejection metrics grouped by error code and virtual key.
  * Added clearer missing-model errors with endpoint-specific guidance.
  * Added user-facing guidance for resolving missing model requests.
* **Bug Fixes**
  * Improved provider error handling by removing untrusted provider messages and sensitive details.
  * Prevented provider error markers from being incorrectly forwarded.
* **Documentation**
  * Documented client-rejection metrics, labels, exclusions, and PromQL troubleshooting queries.


services/langyagent/adapters/otelrelay/llmproxy.go
```diff
@@ -408,23 +416,81 @@ func isGatewayEnvelope(body herr.ErrorBody) bool {
...
+		if title := htmlTitle(body); title != "" {
+			return zap.String("html_title", title)
+		}
+		return zap.Skip()
+	}
+	if message := extractUpstreamErrorMessage(body); message != "" {
+		return zap.String("upstream_message", message)
+	}
+	if raw := strings.TrimSpace(string(body)); raw != "" {
+		return zap.String("raw_body_snippet", boundMessage(raw))
```
langwatch-agent: **P1 - Do not log an unparsed provider response body.** This is a response from a customer-selected upstream and error bodies commonly echo the invalid prompt, structured-output payload, authorization material, or provider request metadata. The old behavior lost diagnostic detail but did not copy arbitrary bytes into the agent log stream; the 2 KiB bound limits volume, not disclosure. This is the same unredacted-failure-path class that must be treated as captured content. Restrict the fallback to a small allowlisted structured summary, or run it through the server-side redaction gate before logging; preserve the body length and a safe provider code for diagnosis.

langwatch-agent (commented):
Static review completed: 1 P1, detailed inline. The gateway counter and missing-model response are sound; the new raw error-body fallback creates an unredacted customer-content logging path.

langwatch-agent:
**Review mode: Targeted Review.** Gateway error taxonomy and a new provenance header on the relay path. 34 files, +2036/-259. Mergeable, CI green, no unresolved threads.

**Scariest production risk: `X-LangWatch-Handled-Error` is trusted provenance, so the strip has to be total.** The design is right: Langy should trust a marker rather than guess from a coincidental `{type, code, message}` shape in a provider body. That makes the header a trust boundary, and the guarantee rests entirely on the gateway stripping it from every forwarded provider response. Worth confirming the strip covers streaming responses and error paths, not just the ordinary JSON relay, because a provider that echoes an arbitrary request header back would otherwise be able to forge LangWatch authorship.

**Must Check** the `missing_model` split out of generic `bad_request` in `gateway_client_rejects_total`. Any dashboard or alert thresholded on `bad_request` volume sees a step change the moment this deploys, and a threshold tuned on the old total will read as an improvement rather than a reclassification.

**Probably Fine** the per-surface messages staying actionable (JSON field, multipart form field, Gemini URL path) and the regenerated handled-error code registry.

Triage from the diff and CI, not a code review.

---
Posted by the pr-hound agent.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `b7cc755` · **Run:** pr-review (reviewer: @claude)

### Notes (non-blocking)

- **[reviewer]** This PR was checked specifically for the risk its own description raises — reclassifying failures as HANDLED errors hiding something. It does not: `missing_model` and `llm_upstream_error` are distinct codes, both server-side (`domain.ErrMissingModel` vs `llmUpstreamErrorCode`) and in the frontend presenter (`platform/app/src/features/errors/logic/presentation.ts`), so a caller can always tell "no model configured" from "model call failed" apart. Provenance of the gateway's own envelope is verified by an explicit `X-LangWatch-Handled-Error` marker rather than inferred from JSON shape, and that marker is actively stripped from every provider-forwarded response lane (`bifrost.go` passthrough headers, `router.go` `writeJSONResponse`/`writeUpstreamError`) so a provider cannot spoof LangWatch authorship. Credential/prose leakage is covered by `TestLLMProxy_LogsOnlyTheHandledClassification`, which scans every log field emitted for a call, not just the documented ones, for a planted secret. Nothing here fails silently: `warnIfMarkerLooksStripped` logs the one ambiguous case (marker missing/mismatched on an envelope-shaped body), and `gateway_client_rejects_total` gives per-key alerting on the new `missing_model` rejection path (docs/ai-gateway/observability.mdx updated to match, including the three deliberate metric exclusions).
- **[reviewer]** Pass 3 (design soundness): this is domain-specific classification of third-party provider error dialects behind a LangWatch-authored-vs-forwarded trust boundary — not a capability an existing dependency (herr, gjson/sjson) already ships. No reinvention found.
- **[reviewer]** `gh pr checks 6619` shows three failing jobs (`evaluate`, `firefighting`, `preflight`) alongside a passing `alls-green` (the actual required-checks gate). These read as unrelated repo-wide workflow gates, not CI failures caused by this diff — flagged for visibility, not as a blocker.

---

drewdrewthis (dismissed):
lgtm

PR #6625, opened 2026-08-05
fix(metrics): bound the rollup successor read's request size by construction

0xdeafcafe:
## The failure

On 2026-08-03 at 21:35 UTC prod ClickHouse logged 20 `SYNTAX_ERROR` (code 62) exceptions inside one minute:

```
Code: 62. DB::Exception: Unrecognized token: Syntax error: failed at position 3022 (&):
 &param_tenantId=project_...&param_seriesId=...&param_time=...
 (in query: query=%0A++++++++%28SELECT+%0A++TenantId%2C+PointId%2C+SeriesId%2C...)
```

The whole URL-encoded search string arrived as the request **body**, so ClickHouse read `query=...&param_tenantId=...` as SQL and stopped at the first `&`. The decoded prefix is the rollup successor seek - the `(SELECT ...)` branch built by `metric-data-point.clickhouse.repository.ts` around the select list in `metric-data-point.rows.ts`.

## Why it stopped, and why that is not a fix

It stopped after #6493 (deployed 08-03 23:09 UTC) reshaped the method, and there have been none since. But #6493 narrowed the *columns* the seek reads; it did nothing to bound the request as a function of the chunk. A seek contributed four parameters (`series{i}`, `time{i}`, `nano{i}`, `point{i}`) and a whole `SELECT` branch carrying the column list, so the request grew with the chunk in both halves at once.

It got quieter by accident, not by construction. The rule this PR puts in force is that no batch can push a request past a stated size: the statement does not vary with the chunk at all, and the parameters are chunked against an explicit budget measured on the client's own encoding.

## (a) or (b)

I believe **(b)**, and (a) as literally described is ruled out at the client.

**The client, `@clickhouse/client` 1.23.0 (pinned in `pnpm-lock.yaml`; `platform/app/package.json` asks for `^1.23.0`):**

- `query()` puts the SQL in the request **body** (`dist/connection/node_base_connection.js`, `let body = params.query` then `POST` with `transformUrl({ url, searchParams })`). `query_params` go in the URL as `param_<name>` entries (`dist/common/utils/url.js`, `toSearchParams`).
- It sets **no `Content-Type` at all** on this path - not `text/plain`, not anything. The only request `Content-Type` in the whole package is `multipart/form-data` on the multipart branch. So the answer to "does it send SQL with `Content-Type: text/plain`?" is: it sends SQL in the body with no `Content-Type` header. Default request headers are `Connection`, `User-Agent` and `Authorization`.
- There **is** a threshold, but it is not a URL-length check and it does not produce form encoding. `MAX_URL_BIND_PARAM_LENGTH = 4096` (`dist/common/utils/multipart.js`) measures the encoded `param_*` entries only, and when exceeded promotes them to a **multipart/form-data** body. It is gated on `use_multipart_params` / `use_multipart_params_auto`, both documented `@default false`, and neither is set anywhere in this repo. There is no `if (url.length > N)` and no `x-www-form-urlencoded` path in this version.

So no code path in the client we run produces `query=<encoded>&param_...` as a body. That string is a `URLSearchParams` serialisation of `{query, ...params}`, which the client never builds. Something re-framed the bytes.

**What supports (b):** the client's own socket pool documents exactly this race, and its trigger is a starved event loop:

- `dist/connection/socket_pool.js:66-71` - *"The check below is still racy on a CPU starved machine. A throttled machine can check time on one line, then get descheduled, decide the socket is still good after rescheduling, and then proceed to use a socket that has actually been idle for much longer than `idle_socket_ttl`."*
- `socket_pool.js:409-420` warns on reuse of a socket whose TTL has expired by more than 1 s, attributing it to *"a starved Node.js process or delayed event loop"*.
- `request()` opens with `await sleep(0)` explicitly to let idle-socket timers fire under CPU load, citing ClickHouse/clickhouse-js#294.

The burst shape fits: 20 events in one minute during a memory-pressure storm, then nothing. A per-request size threshold would have fired steadily, not once.

The residual gap I cannot close from here: a socket-level desync explains mis-framing, but the exact `query=...&param_...` body still needs a producer. `insert()` and `runExec()`-with-values both put their SQL in the **URL** (`node_base_connection.js:199-213`, `:399-424`), so requests with that string in their request line do exist on other paths - which is consistent with one request's request line being consumed as another's body, but I have not proven it.

**`idle_socket_ttl` vs the server's `keep_alive_timeout`:** we set `idle_socket_ttl: 1500` in three places (`platform/app/src/server/clickhouse/client.ts:67-70`, `app-layer/clients/clickhouse.factory.ts:28-31`, `ops/explain-core.ts:274`). The client's guidance is to sit a few hundred ms **below** the server value (its bundled doc: default 2500 ms suits a 3 s server timeout; for 10 s use 9000, *"stay ~500 ms below"*). 1500 ms is below every plausible server default, so the standard guard is satisfied - **but I cannot confirm it from this repo.** The value that has to be read is the ClickHouse server's **`keep_alive_timeout`** in its `config.xml` (equivalently, the `timeout=N` in the server's `Keep-Alive` response header, which the client already parses at `socket_pool.js:174-198`), plus the **idle timeout on any load balancer or ingress between the app and ClickHouse**. Both are infra config outside this repository. `keep_alive.eagerly_destroy_stale_sockets` is `false` (its default) and is the client's own documented mitigation for the starved-process race, if we decide (b) is confirmed.

### 1. The successor read is now one fixed statement, chunked by encoded bytes

`SUCCESSOR_SEEK_QUERY` is built once at module load. Nothing in it varies with the chunk. Every per-point value rides in an array parameter, `arrayZip`-ed back into one row per series on the server, so the statement is a fixed ~2.4 KB and the request binds **eleven** `param_*` entries whatever the chunk holds. `SEEK_SELECT` is emitted **once**, in the `series_points` CTE the branches read from, instead of once per branch.

A fixed statement and a fixed parameter *count* are not on their own a bound on the request, and that distinction is the whole point. The statement travels in the body; the parameters travel in the URL as `param_*` entries, and the arrays those eleven names carry still grow with the number of series. Measured with the pinned `@clickhouse/client` 1.23.0's own `formatQueryParams` fed through `URLSearchParams`, one series encodes to ~530 characters, 13 series to ~4,200, and 64 series to ~19,900 - past the client's own `MAX_URL_BIND_PARAM_LENGTH` of 4,096, which is inert on our configuration because `use_multipart_params_auto` is set nowhere, so `param_*` entries ride the URL at any length.

So chunking now splits the successor read on **encoded parameter bytes** against an explicit budget (3,500 characters, deliberate headroom under the client's 4,096), with the seek cap kept as an upper bound on series. Three unit tests measure the emitted requests with the client's own serialisation, including the adversarial case: 64 series at the full 64-character identifier width, which a count-bounded chunker would have sent as one ~19,900-character request.

`use_multipart_params_auto: true` on the shared client factory is the complementary option, and it is deliberately **not** in this PR: it changes the request framing for every ClickHouse caller in the app, not just this lane, so it wants its own change and its own verification. The byte budget holds regardless of whether it is ever turned on.

### 2. Two branches, so a smaller request does not buy a wider scan

Collapsing N index seeks into one joined statement usually trades a bounded read for an unbounded one, which is the class #6493 fixed. It does not here, because of an invariant the method already relies on: `recomputeAffectedRollupsMany` ensures every chunk point is stored *before* this runs. So within a series, no chunk point's successor can be further away than the next chunk point - it cannot outrun a row we know is there. Only the newest chunk point in a series needs an open-ended look forward.

- branch 1 reads the chunk's own span (`> spanFrom AND < spanTo`), bounded by the chunk
- branch 2 is a single `LIMIT 1 BY SeriesId` seek past the end of it

`affectedBucketsBySeries` takes the minimum of the union of chunk points and returned rows, so returning the whole span rather than one row per point cannot change which successor a point resolves to. The read stays payload-free (`SEEK_SELECT`, no `CanonicalPayload`), as #6493 made it.

Chunking now bounds how many **series** one set of arrays describes rather than how large the statement grows, so 130 points in one series is one request instead of three.

### 3. Both halves prune partitions

`metric_data_points` partitions by `toYearWeek(TimeUnixMs)`, and a bound that only reaches the column through the join cannot prune. The first revision of this PR had one shared lower bound in the CTE, which left branch 1 scanning to the end of the table before the join discarded everything past the spans, and branch 2 scanning from the earliest span *start*.

Each branch now carries the one constant bound its own half can prove: a row within a span sits below `spanEnd` (the furthest span end), a row past a span sits above `spanStart` (the nearest). Both are pure upside - pushed down they prune, left in place they are a cheap filter over rows the branch discards anyway. Range-filtering `TimeUnixMs` under `FINAL` is safe here because it is part of the dedup key, so no version of a row can drift out of the window - the movable-column trap in `clickhouse-queries.md` does not apply.

Against the rest of that guide: `TenantId` leads the CTE's predicate, matching the sort key; there is no `max(col)` used as a sort key and no dedup subquery to keep heavy columns out of. The `LIMIT 1 BY` is the narrow kind the guide permits rather than the kind it bans - six sequence columns, nothing heavy - and the code now says so at the site, so a later reader does not "fix" it into the IN-tuple form.

### 4. The affected-bucket reads shed their derivable parameters

These deliberately keep a seek per bucket. Their predecessor branch is bounded below by the **retention window**, and a join can only apply that bound after the rows are read - so the joined form turns a single-row reverse index seek into a read of the series across ~49 days, per rollup chunk. That is the memory failure #6493 fixed, and I am not reintroducing it to make the statement constant-size. The successor read folds safely only because every bound there is the chunk's own span.

What they can shed is fan-out for values the server can derive: the bucket end and the retention floor are both arithmetic on the bucket start, so they now travel once as `bucketMs` / `retentionMs` scalars. A bucket costs two parameters instead of four, which halves the search string and brings a full 32-seek read comfortably under the client's own 4096-character `param_*` ceiling.

**Deliberately deferred, and documented at the site rather than fixed here:** the successor read's three constant time bounds are chunk-**global**. One backfilled point in one series lowers `scanFrom` for every series in the request, and the branch that looks past a span end is `[earliestSpanEnd, ∞)` per series under `FINAL`, terminating on the `LIMIT 1 BY` rather than on an index range. Correctness is unaffected - `affectedBucketsBySeries` takes the minimum of the union, so extra rows cannot move a resolved successor - but a chunk pairing one late series with live ones reads more than the live ones needed. The fix is a per-series bounded lookahead (a join-side `SpanToTime + <window>` upper bound, with a fallback seek for the series the bounded branch returns nothing for). It wants `EXPLAIN indexes=1` against a real server and equivalence coverage the parameter-level model in these tests cannot give it, and getting it wrong loses successors rather than costing time, so it gets its own change.

**Open follow-up, deliberately not in this PR:** the predecessor's lower bound could be tightened from the retention window to *the previous affected bucket* for every bucket except a series' earliest - every affected bucket provably contains a stored point, so a later bucket's predecessor cannot precede the one before it. That would make the bucket reads collapsible too. It changes a bound rather than a parameter shape, so it wants its own change and its own verification.

## The client seam: already correct, and not on the backlog

**This repository is not on #6614's `RESOLVES_DIRECTLY_BACKLOG`, and there is nothing here to take off it.** `MetricDataPointClickHouseRepository` never calls `createClient`, and it never resolves a client either - it takes `resolveClient` and `resolveOrganizationClient` as constructor arguments, wired in `presets.ts:707-711`, which is one of the sites that boundary test explicitly permits. It also matches `mayResolveByLocation` twice over (`/repositories/`, `.repository.ts`). So it already satisfies both rules the test enforces, and no line of the ratchet can be deleted on account of this change.

**No conflict with either open PR.** #6614 touches `platform/app/src/server/clickhouse/*`, `presets.ts`, `app.ts`, `dependencies.ts` and the filters/analytics paths; #6622 touches `ee/governance/*`, `ee/billing/*` and `src/app/api/*`. This PR touches the metric repository, its two test files, `specs/otlp/canonical-metric-ingestion.feature` and one deleted line in `check-feature-parity.ts`. Zero overlap.

**What this path gets from the shared client today** (via `presets.ts` -> `clickhouseClient.ts` -> `client.ts` -> `createResilientClickHouseClient`):

- retry on transient reads, using `runWithRetry` and the classifier from `@langwatch/clickhouse-client` - inserts are deliberately not retried
- fleet-aware pool sizing, `resolvePoolSize(poolSizingFromEnv(...))` from the same package, applied at `client.ts` as `max_open_connections`
- the vendor-log policy from the package (`emitVendorLog` via `clickhouseLogger.ts`) - drops the vendor's `error` level, never attaches a cause under a field named `error`
- error translation to typed `HandledError`s, Prometheus duration/count metrics, cold-scan detection, and the in-band exception guard #6493 added

**What it does not get, and would not whatever this PR did.** Of the package's eight modules only three have any consumer; `pipeline.ts`, `rateLimit.ts`, `tenancy.ts`, `tenantGuard.ts` and `tracing.ts` have none, and `compose` has no caller outside its own test. So no query on this path emits a span (`db.system`, `db.sql.table`, `langwatch.tenant_id`, `db.response.read_bytes` are all unset), and none is tenant-guarded. The guard could not accept these calls as they stand in any case: it validates a `QueryRequest` carrying an explicit `tenantId`, and the app path calls the vendor's `query({ query, query_params })` signature, which has no such field.

That is a wiring gap in the seam, not in this repository, and adopting it here alone would be exactly the per-caller opt-in the boundary test exists to prevent. When #6614 lands, the statement limiter arrives through `managedClient.ts` and this path inherits it with **no change to this file** - which is the point of already being behind the seam. Two things worth someone's attention, neither mine to fix here: the package carries no comment recording that five of its eight modules are unwired, so a reader of `tenantGuard.ts` has nothing telling them it has no callers; and `clickhouseClient.ts:29-80` still holds a hand-rolled duplicate of `tenancy.ts`'s routing rule, with no bound on its cache and a bare `Error` where the package throws `DuplicateRouteError`.

## Verification

- `pnpm test:unit run src/server/app-layer/metrics` - 27 passing.
- The statement guards assert it is **invariant**, not merely smaller: byte-identical SQL across chunks of 1/12/64/130/260 points and across 1/8/64 series; the same eleven parameter names in every case; the select list present exactly once; no payload column.
- The size guards assert the quantity that actually travels. Every request the lane emits is re-measured with `@clickhouse/client`'s own `formatQueryParams` over `URLSearchParams` and has to sit inside the stated budget - across one series and across 512, and at the adversarial 64-series full-width shape. The affected-bucket read, which keeps a seek per bucket by design, is held to the client's 4,096-character ceiling on a full chunk of them.
- Equivalence is proven by running the lane twice over the same in-memory table, once answering with what the folded statement returns (read off its own array parameters, **including the two partition bounds**, so a bound that wrongly excluded a row fails as a missing successor) and once with what the seek-per-point statement returned, then comparing the buckets the lane goes on to read.
- Those equivalence fixtures were **vacuous before this revision**, in a way worth naming: their points were gauges, and a gauge successor never pulls a second bucket into the affected set (`usesPredecessor`), so every assertion in the block held with the successor read deleted entirely. They are cumulative now, and two discriminating fixtures were added - an interior stored point in a bucket no chunk point occupies, and two series whose spans sit hours apart. Four mutation probes were run against the result: dropping the within-span branch, and flipping each of `scanFrom` / `earliestSpanEnd` / `latestSpanEnd` between `min` and `max`. Each now fails a test; before, all four survived.
- Integration: one block folds a chunk over a series that already holds points between the chunk's own points, and a new block folds **one chunk carrying two series** staggered hours apart, each with stored successors past its own span - the shape `bulkAppend` actually produces, and the only shape in which the `arrayZip` pairing or the `LIMIT 1 BY SeriesId` can be wrong at all. Both compare against reference series rebuilt strictly one point at a time, never through the folded entry point.
- `specs/otlp/canonical-metric-ingestion.feature` gains a tagged `Rule` with eight bound scenarios; the file came off `LEGACY_INERT` in `check-feature-parity.ts`, which now reports `8/8 scenarios bound`.

## What CI has and has not proven, and the lesson in the gap

No ClickHouse was reachable in the environment this was written in (no local server, no container runtime), so the new SQL was validated for semantics against an in-memory model and for types, but **not parsed by a real server** until CI did it. The syntax it newly relies on is `arrayZip` over seven arrays with tuple-element access, a CTE referenced from both operands of a parenthesised `UNION ALL` inside an outer `SELECT`, `LIMIT 1 BY` after a join, and `CAST(... AS String)` on the join key against the `FixedString(64)` column. All are ordinary for the 25.10 server we run, but ordinary is not verified.

**The reusable lesson, and it is not "CI was red and we missed it".** It is that a job can run, report green, and never execute the file that mattered. This branch produced three cancelled `test-integration` runs before one of them reached the metric repository test, and on the first of those every test that *did* run was green. Had the aggregate been read as coverage, the SQL would have shipped unparsed with a green tick above it. **A rollup is evidence of what ran, not of what was checked.** When a change turns on one file being exercised, confirm that file by name in the job output before treating the job as evidence.

Two independent mechanisms produced that gap here, and it is worth separating them because they mislead differently:

1. **Selection.** On a draft the job runs `pnpm test:integration --changed origin/main`. This branch was cut from a `main` that had since moved, so "changed" resolved to the whole delta - 85+ unrelated files - and the metric test sat behind all of them. Fixed by rebasing onto current `origin/main` (clean; nothing upstream had touched these files).
2. **Supersession.** `langwatch-app-ci` sets `concurrency: ${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress` on any ref that is not `main`, and its `pull_request` trigger list is `[opened, synchronize, reopened, ready_for_review]`. So every push *and every un-drafting* starts a run that kills the one in flight. This PR was flipped out of draft twice by something other than the author (`ready_for_review` at 23:38:52 and 23:49:45 UTC), and those two events - not the draft state itself, and not the re-draft, since `convert_to_draft` is not in the trigger list - cancelled runs 1 and 3. The third cancellation was the author's own rebase push superseding a run on the stale SHA, which was correct.

That second mechanism is the more general trap: on this workflow the *heavy* jobs are gated on `draft == false`, so a PR that is repeatedly drafted and undrafted can accumulate cancelled runs while never completing the jobs the gate exists to enable. This PR is deliberately left undrafted for that reason.

## Summary by CodeRabbit

- **Improvements**
  - Improved metric ingestion for batched and multi-series data.
  - Enhanced cumulative calculations and rollups for consistent results across ingestion patterns.
  - Reduced request overhead while preserving accurate summaries and sample counts.

- **Quality**
  - Expanded validation for request limits, successor resolution, bucket boundaries, and payload handling.
  - Strengthened feature-parity checks to track documented scenarios more consistently.


platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
  */
 const SEEKS_PER_QUERY = 64;
 
+/**
+ * The successor read's whole statement, emitted once at module load rather than
+ * rebuilt per chunk, because nothing in it varies with the chunk any more.
+ *
+ * Every per-point value now travels as an array parameter, so the request a
+ * chunk produces is a fixed ~1 KB of SQL plus nine `param_*` entries whatever
+ * the chunk holds. The shape this replaced grew four parameters and a full
```
0xdeafcafe: **[P1] size — the request is bounded in parameter *count*, not in bytes; at a full chunk it is larger than the shape it replaced**

Measured with the pinned `@clickhouse/client` 1.23.0's own `formatQueryParams` + `URLSearchParams` on exactly the payload `successorSeekParams` builds: 1 series ≈ 570 chars, 13 series ≈ 4,217 (crosses `MAX_URL_BIND_PARAM_LENGTH` = 4096), 32 ≈ 10,092, **64 series ≈ 19,910 — vs ~14,075 for the old per-point shape at its 64-point ceiling**. The 4096 threshold this comment names is inert on our config: `use_multipart_params`/`use_multipart_params_auto` are set nowhere in the repo, so `param_*` entries ride the URL at any length.

Two smaller drifts in the same block: it says **nine** `param_*` entries (the code binds eleven — `successorSeekParams`'s own doc and the unit test both say eleven) and **~1 KB** of SQL (~2.4 KB measured).

Suggested resolution: set `use_multipart_params_auto: true` on the shared client factory and/or chunk `successorsOf` by encoded parameter bytes rather than `SEEKS_PER_QUERY` series (12 series is the last chunk under 4096 on this shape). Either way, add a test pinning `new URLSearchParams(entries).toString().length` for a 64-series chunk under an explicit budget — the existing tests pin SQL bytes and parameter *names*, which stay green at any array size, and the encoded request is the quantity the incident was about.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
...
+      AND SeriesId IN {seriesIds:Array(String)}
+      AND metric_data_points.TimeUnixMs >= fromUnixTimestamp64Milli({scanFrom:Int64})
+  )
+  SELECT * FROM (
+    (SELECT series_points.*
+     FROM series_points INNER JOIN spans ON CAST(series_points.SeriesId AS String) = spans.SpanSeriesId
+     WHERE series_points.SeekTime <= fromUnixTimestamp64Milli({spanEnd:Int64})
+       AND ${orderedAfter("From")} AND ${orderedBefore("To")})
+    UNION ALL
+    (SELECT series_points.*
```
0xdeafcafe, line 179: **[P1] performance — the past-span branch has no upper time bound and cannot early-terminate**

`series_points` is bounded only below by `scanFrom`, and this branch adds only `SeekTime >= {spanStart}`. Both are chunk-**global** minima, so one late/backfilled point in any one series drops the floor for all 64 series in the request, and the read is `[oldest chunk point, ∞)` per series under `FINAL`. The seek this replaced was a single-table `ORDER BY … LIMIT 1` that `optimize_read_in_order` terminates after ~one granule; `LIMIT 1 BY` above an `INNER JOIN` gets no such push-down — the sort is a blocking step over the joined stream. Concrete: a 64-series chunk where one series carries a point from 24 h ago at a 15 s scrape reads ~370k narrow rows FINAL-merged and sorted to return 64. `bulkAppend` is documented as the replay path, where every chunk has an old `scanFrom`.

Options: a per-series upper bound on the forward look (e.g. `SpanToTime + <lookahead>` join-side, with a fallback seek only for series the bounded branch returns nothing for), or keep a per-point `LIMIT 1` seek for the newest point of each series only (≤64 cheap index seeks). Worth an `EXPLAIN indexes=1` before/after on a real server.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
...
+  ),
+  series_points AS (
+    SELECT ${SEEK_SELECT}, metric_data_points.TimeUnixMs AS SeekTime
+    FROM metric_data_points FINAL
+    WHERE TenantId = {tenantId:String}
+      AND SeriesId IN {seriesIds:Array(String)}
+      AND metric_data_points.TimeUnixMs >= fromUnixTimestamp64Milli({scanFrom:Int64})
+  )
+  SELECT * FROM (
+    (SELECT series_points.*
```
0xdeafcafe, line 174: **[P1] performance — the within-span branch returns every stored row inside each series' span, so response size tracks span width, not chunk size**

Correctness is unaffected (`successorIn` is an upper-bound search; duplicates and extra rows cannot move the resolved successor — traced). But each row is decoded by `fromSeekRow`, accumulated, and re-sorted in `affectedBucketsBySeries`, and the previous shape bounded the response at one row per chunk point. A chunk pairing a late point with a live point for the same series — the exact scenario this lane exists for — makes the span the whole gap: ~5,760 rows for one series at a 15 s scrape across 24 h, plus interior chunk points returned again as duplicates. Consider a per-series cap server-side (only the first row after each chunk point can matter) or bounding the chunk by span width as well as series count.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
...
+      span.7 AS SpanToPoint
+    FROM (
+      SELECT arrayJoin(arrayZip(
+        {seriesIds:Array(String)},
+        {fromTimes:Array(Int64)}, {fromNanos:Array(String)}, {fromPoints:Array(String)},
+        {toTimes:Array(Int64)}, {toNanos:Array(String)}, {toPoints:Array(String)}
+      )) AS span
+    )
+  ),
+  series_points AS (
```
0xdeafcafe, line 166: **[P2] the `series_points` CTE executes twice, not once**

ClickHouse substitutes non-recursive `WITH` CTEs at each reference; both union operands reference `series_points`, so the `FINAL` scan over `metric_data_points` runs twice per request. The doc comment ("SEEK_SELECT appears exactly once, in the `series_points` CTE … the CTE is already the seek") reads as one read; it is one occurrence and two executions. Either state the doubling in the comment or fold the branches into a single scan with one `WHERE (span-interior OR past-span)` predicate and measure which is cheaper.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
  */
```
0xdeafcafe, line 58: **[P3] `SEEKS_PER_QUERY`'s doc no longer describes what it bounds on the successor path**

"How many index seeks one rollup query may fold together" now counts *series per set of array parameters* for `successorsOf` (no statement grows at all), and, halved, seeks per bucket request for `pointsForAffectedBuckets`. The use site documents the reinterpretation; the definition — where the next editor changes the number — does not.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
...
+     FROM series_points INNER JOIN spans ON CAST(series_points.SeriesId AS String) = spans.SpanSeriesId
+     WHERE series_points.SeekTime >= fromUnixTimestamp64Milli({spanStart:Int64})
+       AND ${orderedAfter("To")}
+     ORDER BY series_points.SeriesId ASC, series_points.SeekTime ASC,
+       series_points.TimeUnixNano ASC, series_points.PointId ASC
+     LIMIT 1 BY SeriesId)
+  )
+`;
+
+/**
```
0xdeafcafe, line 189: **[P3] `orderedAfter`'s doc explains a qualification of a column the predicate never references**

It says the predicate is "table-qualified … a bare `TimeUnixMs` resolves to its epoch-milli alias", but the emitted text compares `series_points.SeekTime` / `TimeUnixNano` / `PointId` — `TimeUnixMs` appears nowhere in it. What actually dodges the alias trap is the CTE keeping the raw `DateTime64` under the distinct name `SeekTime`. A reader deleting `SeekTime` in favour of `TimeUnixMs` would find this comment reassuring and be wrong.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -125,6 +237,64 @@ function affectedBucketsForSeries({
...
+  return {
+    tenantId,
+    seriesIds: spans.map((span) => span.seriesId),
+    fromTimes: spans.map((span) => span.first.timeUnixMs),
+    fromNanos: spans.map((span) => span.first.timeUnixNano),
+    fromPoints: spans.map((span) => span.first.pointId),
+    toTimes: spans.map((span) => span.last.timeUnixMs),
+    toNanos: spans.map((span) => span.last.timeUnixNano),
+    toPoints: spans.map((span) => span.last.pointId),
+    scanFrom: Math.min(...spans.map((span) => span.first.timeUnixMs)),
```
0xdeafcafe, line 344: **[P3] naming — `spanStart` and `spanEnd` are both derived from span *ends***

`scanFrom` = min(first), `spanStart` = min(**last**), `spanEnd` = max(last). `spanStart` next to `spanEnd` reads as "earliest span start", which is actually `scanFrom`; someone widening the branch-2 bound to min(first) would think they were correcting a typo. `earliestSpanEnd` / `latestSpanEnd` would make the SQL self-describing (params object, statement, and the unit test's name list move together).

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -505,29 +657,39 @@ export class MetricDataPointClickHouseRepository
     // Each seek contributes two statements, so half the budget keeps a single
```
0xdeafcafe: **[P2] the bucket read's chunk-size rationale points at a ceiling that no longer exists**

"Each seek contributes two statements, so half the budget keeps a single query's statement count at the same ceiling the successor seeks use" — the successor read now emits one fixed statement regardless of chunk size, so there is no successor-side statement ceiling to match, leaving the divisor unexplained. Its real justification is now measurable on its own terms: 32 seeks × 2 parameters keeps the encoded `param_*` string under the client's 4096 threshold (≈3.5k chars measured); 64 would not.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
```diff
@@ -418,6 +421,497 @@ describe("MetricDataPointClickHouseRepository", () => {
...
+  });
+
+  /**
+   * Result equivalence for the folded read. The statement no longer seeks once
+   * per point: it reads the chunk's own span in one branch and looks past the
+   * end of it in another. Those rows differ from the per-point seek's rows, so
+   * what has to be proven is that the rollup lane resolves the same successors
+   * from them — observable as the buckets it then goes on to read.
+   */
+  describe("when the folded read replaces a seek per point", () => {
```
0xdeafcafe, line 695: **[P2] the equivalence fixtures cannot detect loss of the within-span branch or a wrong constant bound**

Two mutation probes, both survived by every fixture: (1) make the `withinSpan` half of `foldedRows` return nothing — all four equivalence tests and the integration block stay green, because every interleaved stored point shares a bucket with a chunk point, so the bucket branch 1 exists to recover is already in the set. (2) flip any of `scanFrom`/`spanStart`/`spanEnd` min↔max — all fixtures place every series within a few ms, so min == max and the choice is unobservable. Cheap discriminating fixtures: chunk points at `base` and `base + 90_000` with a stored point at `base + 35_000` (kills probe 1), and two series whose spans sit hours apart, each with an interior point and a stored successor past its span end (kills probe 2). Root cause worth noting: the model is built from the query *parameters*, never from `SUCCESSOR_SEEK_QUERY` itself, so the statement is pinned by three `toContain` substrings plus a single-series integration run.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
```diff
@@ -418,6 +421,497 @@ describe("MetricDataPointClickHouseRepository", () => {
...
+      const calls: { query: string; params: Record<string, unknown> }[] = [];
+      const query = vi.fn(
+        async ({
+          query: sql,
+          query_params: params,
+        }: {
+          query: string;
+          query_params: Record<string, unknown>;
+        }) => {
+          if (sql.includes("spans")) calls.push({ query: sql, params });
```
0xdeafcafe: **[P3] statements are discriminated by `sql.includes("spans")`, a CTE name**

Renaming the CTE (a pure refactor) re-routes every mock in five helpers. It fails loudly rather than silently, so this is maintenance cost, not a correctness hole — discriminating on something the contract owns (e.g. `{seriesIds:Array(String)}`) or exporting a predicate next to `SUCCESSOR_SEEK_QUERY` removes it.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
```diff
@@ -365,3 +365,82 @@ describe("given a cumulative series long enough to span several rollup buckets",
...
+/**
+ * The folded successor read does not seek once per point: within a series it
+ * reads the chunk's own span in one branch and looks past the end of it in
+ * another. Its results are only identical to a seek per point because every
+ * chunk point is already stored, so no chunk point's successor can be further
+ * away than the next chunk point. That argument is what this block exercises
+ * against real ClickHouse - a series where the rows between the chunk's points
+ * are rows the chunk itself does not carry.
+ */
+describe("given a series whose chunk points have stored points between them", () => {
```
0xdeafcafe, line 382: **[P1] coverage — the folded read never runs against real ClickHouse with more than one series**

Every `recomputeAffectedRollups*` call in this file passes points of a single `seriesId`. With one series the new mechanism degenerates: one `arrayZip` span row (a mis-paired tuple index cannot show), one join candidate, and `LIMIT 1 BY SeriesId` indistinguishable from `LIMIT 1` — while `bulkAppend` hands this lane coalesced multi-series batches in production. A defect pairing series A's id with series B's bounds would drop successors for every series but one, leave stale rollups, and pass this entire suite; the unit tier mocks the server so it cannot cover it. (A synthetic multi-series run against ClickHouse 25.10 during review confirms the *current* SQL resolves correctly, so this is a regression trap, not a live bug.) Suggested: fold one chunk containing two or more series staggered in time, each with stored points after its span, and assert both series' rollups match a per-point rebuild.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
```diff
@@ -365,3 +365,82 @@ describe("given a cumulative series long enough to span several rollup buckets",
...
+      timeUnixMs: bucket0 + index * 5_000,
+      metricKind: "sum",
+      aggregationTemporality: "cumulative",
+      isMonotonic: true,
+      valueDouble: values[index]!,
+      acceptedAt,
+    });
+  }
+
+  beforeAll(async () => {
```
0xdeafcafe, line 412: **[P3] the "point-at-a-time" reference series is itself half-built by the folded path**

`beforeAll` seeds `preStored` into both series with `recomputeAffectedRollupsMany` (a chunk); only the other half differs in arrival mode, so a folded-read defect affecting the pre-stored half cancels out of the comparison. Seeding `perPointSeriesId` one point at a time via `recomputeAffectedRollups` makes the reference genuinely independent — though note both paths now share `SUCCESSOR_SEEK_QUERY`, so a systematically broken predicate would still look equal on both sides; the real cross-check is the per-point-rebuild comparison, which is why the reference should avoid the folded entry point entirely.

specs/otlp/canonical-metric-ingestion.feature
```diff
@@ -82,3 +82,57 @@ Feature: Canonical OTLP metric ingestion
...
       When the same point is processed again
       Then the stored point and its summaries are unchanged
+
+  # A batch of points once produced a storage request that grew with the batch:
+  # a parameter set and a whole query branch per point. Storage rejected one
+  # such request outright, having read the request's own encoding as part of
+  # the query. It stopped happening when an unrelated change made the requests
+  # smaller, which is not the same as bounding them - a larger batch would have
+  # crossed the same line again. These scenarios bound it.
+  Rule: Rebuilding summaries costs the same request whatever the batch holds
```
0xdeafcafe: **[P2] the Rule overclaims — it promises a fixed-size request for a read the change deliberately leaves batch-proportional, and "the same request" is only true of the statement**

"Rebuilding summaries costs the same request whatever the batch holds" governs all six scenarios including the bucket read, which still emits two branches + two params per bucket by design (the code explains why folding it would reintroduce the #6493 memory class). And for the successor read, the *statement* is fixed but the bound values grow linearly with series count (~570 → ~19,910 encoded chars at 64 series; no test inspects `request.params` size). A reader taking this Rule as the contract concludes the incident class is closed, or "fixes" the bucket read into the joined form. Suggested: narrow the Rule to "sends a request bounded independently of the batch", keep the byte-for-byte claim in the statement scenarios' own Thens, and add a scenario + test for the property that matters — the encoded request stays under a stated budget however many series the batch touches.

specs/otlp/canonical-metric-ingestion.feature
```diff
@@ -82,3 +82,57 @@ Feature: Canonical OTLP metric ingestion
...
+    @unit
+    Scenario: A folded rollup read binds a fixed number of parameters
+      Given a batch of points
+      When the platform looks up what follows each of them
+      Then the number of values bound into the request does not grow with the
+        batch
+      And a batch too large for one request is split rather than sent whole
+
+    @unit
+    Scenario: A folded rollup read names its columns once
```
0xdeafcafe: **[P3] the first Then restates the invariant of the scenario above**

If the statement is byte-identical whatever the batch holds, "names the fields it reads once" adds no independent claim. This scenario's real invariant is the second Then — the payload column stays out (the #6493 memory guarantee) — and its bound test checks exactly that. Retitling around the payload claim keeps one invariant per scenario.

specs/otlp/canonical-metric-ingestion.feature
```diff
@@ -82,3 +82,57 @@ Feature: Canonical OTLP metric ingestion
...
+    @unit
+    Scenario: A folded rollup read resolves the successors a per-point read did
+      Given a batch of points whose series already holds points between them
+      When the platform looks up what follows each of them in one request
+      Then it resolves the same following point for each as it would asking
+        one at a time
+      And it recomputes exactly the same summary windows
+
+    @unit
+    Scenario: A rollup bucket read derives its bounds on the server
```
0xdeafcafe: **[P3] "derives its bounds on the server" states where a computation runs, not a behavior**

Where the window end is computed is invisible; the observable is that the request does not carry a value per window. The corpus phrases this class observably ("one batched insert per batch … not one insert per item", "in one query, not one per facet"). E.g. "the request carries the window size and the retention span once, not once per window".

0xdeafcafe:
**Read-only review of this PR — 6 specialists (code / test / feature / security / hygiene / spec) + empirical SQL verification. Nothing here was auto-applied.**

**Verdict: correctness is solid; the size story is the gap.** The exact `SUCCESSOR_SEEK_QUERY` this branch emits was reconstructed and run on `clickhouse local` 25.10.1 (prod's major) against a synthetic `metric_data_points` built from migration 00049, with adversarial fixtures: interleaved points, a single-point span, equal-millisecond rows split only by `TimeUnixNano`/`PointId`, three series with staggered spans, and two candidate rows past each span end. It returned exactly the correct result set in every case — `LIMIT 1 BY` applies per-UNION-operand, `FINAL` collapses stale versions, cross-tenant and pre-`scanFrom` rows excluded. The bucket read's server-side `from ± retention/bucket` arithmetic also verified, including safe pre-epoch underflow. Independent tracing of the type seams (FixedString(64) join, UInt64 nanos, DateTime64 bounds), the three constant bounds (each provably weaker than its per-row predicate), tenancy (both branches inherit the CTE's TenantId-first predicate; spans is pure bound parameters), injection (nothing runtime-tainted interpolated), and the empty-chunk path (`Math.min` can never see an empty array) all came back clean. Biome 2.5.1 check-only: 0 violations in touched code. All six `@scenario` bindings verified character-for-character; the `LEGACY_INERT` removal is required, not optional (leaving it would trip the stale-inert ratchet).

**Findings posted inline:** 4×P1 (encoded request size unbounded and larger at full width than the old shape; past-span branch unbounded/no early termination; within-span response tracks span width; integration tier never folds >1 series), 4×P2, 6×P3.

**Not anchorable inline:**
- **[P3]** "Storage trouble asks the client to retry" (line 52) collides verbatim with `specs/otlp/canonical-log-ingestion.feature:34`, and the parity script indexes bindings by title alone repo-wide. Harmless while untagged, but this file is now enforced — the next person to tag it can silently satisfy the wrong spec. Worth disambiguating while the file is open.
- **[P3]** `specs/README.md` still tells authors to add files to a `WATCHED` list that no longer exists; the real model is enforce-by-default + `LEGACY_INERT` opt-out, which this PR exercises.
- **[P3]** PR body: "The shape on main today builds larger requests than the one that failed… a busier chunk would cross the same line again" — per the size P1 this remains partly true post-merge for the URL half, and repo convention is to describe the rule in force rather than the live gap. Suggest trimming that framing; the incident forensics themselves are fine (all derived from public source).

**Decide-class items** (judgment, not mechanical): `use_multipart_params_auto` on the shared client vs byte-budgeted chunking (or both); per-series bounding strategy for branch 2; extracting the ~180-line query module to `metric-data-point.successor-seek.ts` beside the existing `.rows.ts` / `.usage.ts` seams.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Fixed in a327b22453. Chunking now splits on encoded parameter bytes rather than on series count: `successorSeekChunks` measures what `successorSeekParams` produces the way the client does — `formatQueryParams` per value, then `new URLSearchParams(entries).toString().length` — and closes a chunk before it crosses `SUCCESSOR_PARAM_BUDGET_CHARS` = 3500, keeping `SEEKS_PER_QUERY` = 64 as the upper cap on series. Your figures reproduced exactly here (530 / 4,214 / 19,871 encoded characters at 1 / 13 / 64 series), so a chunk now carries ~10 series at this table's identifier widths.

Three tests measure the requests the lane actually emits, using the pinned client's own `formatQueryParams`: every request across chunks of 1/12/64/130/260 points and 1/8/64/512 series must sit inside the budget, as must the adversarial shape — the seek cap's full 64 series at 64-character identifier width — and the split has to ask about each series exactly once rather than merely total 64. Raising the budget to a million fails all three.

Both drifts corrected: eleven `param_*` entries, and ~2.4 KB of SQL (2,483 characters, measured on the emitted statement).

`use_multipart_params_auto` deliberately stays out of this PR — it changes request framing for every ClickHouse caller in the app, not just this lane, so it wants its own change and its own verification. Recorded in the body as the complementary option.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Agreed, and deferred to its own change rather than fixed here — leaving this thread open.

What the branch actually does is now written at the site (a327b22453): all three constant bounds are chunk-**global**, one backfilled point in one series lowers `scanFrom` for every series in the request, and this branch is `[earliestSpanEnd, ∞)` per series under `FINAL`, terminating on the `LIMIT 1 BY` rather than on an index range. A reader will not now mistake it for a bounded read.

Why not in this PR: the shape you describe — a join-side `SpanToTime + <lookahead>` upper bound with a fallback seek for the series the bounded branch returns nothing for — is a two-phase read, and its equivalence cannot be established by the parameter-level model these unit tests use. The `EXPLAIN indexes=1` before/after you rightly ask for also needs a real server, and none was reachable from the environment this was written in. Getting it wrong loses successors rather than costing time, so it gets a change that can prove it.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Same disposition as the branch-2 thread: deferred, left open.

Agreed on the analysis including the part that lets it wait — correctness is unaffected, because `affectedBucketsBySeries` takes the minimum of the union, so extra and duplicate rows cannot move a resolved successor. What it costs is decode and sort work proportional to span width rather than to chunk size, which is real.

A per-series server-side cap is the same piece of work as the branch-2 lookahead: both are "bound the join side per series, keep a fallback for what that misses", verified against the same seam. Splitting them would mean proving that seam twice. Documented at the site in a327b22453 and carried into the follow-up with branch 2.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Fixed in a327b22453 — the comment states the doubling now: written once, executed twice, because ClickHouse substitutes a non-recursive `WITH` at each reference and both union operands read `series_points`, so the `FINAL` scan runs once per branch.

Took the "state it" option rather than the fold. Collapsing to one scan under a single `WHERE (span-interior OR past-span)` costs the per-operand `LIMIT 1 BY`, which is exactly what keeps the look forward to one row per series. The comment says that, and says to measure before trading.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Fixed in a327b22453. The definition now describes what it bounds: the upper cap on series per set of array parameters for the successor read and, halved, buckets per affected-bucket request; that it bounds how much one request may *read* rather than how large it may be; and that size is bounded first, by `SUCCESSOR_PARAM_BUDGET_CHARS`. So the next editor reading the constant learns that raising it widens a read and cannot enlarge a request.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Fixed in a327b22453. The doc now names what actually dodges the trap: the predicate reads `series_points.SeekTime` and never `TimeUnixMs`, because `SEEK_SELECT` already binds that name to its epoch-milli alias, and the CTE keeps the raw `DateTime64` under a distinct name for exactly that reason. It ends by telling a reader not to simplify `SeekTime` away, which is the move the old wording quietly encouraged.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Renamed in a327b22453: `spanStart` → `earliestSpanEnd`, `spanEnd` → `latestSpanEnd`, moved together across the params object, the statement, both branch predicates, the doc comments, and the unit test's parameter-name list and equivalence model.

The params doc now also says both are derived from span *ends* and that the earliest span start is `scanFrom`, so widening either to a span start reads as the change of meaning it is rather than as a typo fix.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Fixed in a327b22453. The comment gives the real justification now: the successor read emits one statement whatever its chunk holds, so there is no statement ceiling left here to match. What the divisor buys is that two parameters per bucket at 32 buckets encode to roughly 3.5k characters of `param_*` entries with this table's 64-character series identifiers — inside the client's own 4096-character ceiling, where 64 buckets would be half as far outside it again.

Pinned rather than only asserted in prose: a new test folds forty buckets of one series and re-measures each emitted bucket read with the client's own `formatQueryParams`, requiring it under 4096.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
0xdeafcafe: Fixed in a327b22453, and the root cause ran deeper than the two probes.

Every point in this block was a gauge, and a gauge successor never pulls a second bucket into the affected set — `usesPredecessor` is `metricKind === "summary" || aggregationTemporality === "cumulative"`. So the buckets being compared were fully determined by the chunk points alone, and every assertion in the block held with the *whole* successor read deleted, not merely with its within-span half returning nothing. The fixtures are cumulative now, which is what makes the block capable of failing at all; that is recorded in a comment on `chunkPoint` so it does not regress to gauges.

On top of that, both fixtures you specified: chunk points at `base` and `base + 90_000` with a stored point at `base + 35_000`, in a bucket no chunk point occupies; and two series whose spans sit hours apart, each with an interior stored point and a stored successor past its own span end.

Probes re-run against the result — within-span returning nothing, and each of `scanFrom` / `earliestSpanEnd` / `latestSpanEnd` flipped min↔max. All four now fail a test. All four survived before.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
0xdeafcafe: Fixed in a327b22453. The mocks discriminate on `{seriesIds:Array(String)}` now, through one shared `isSuccessorRead(sql)` helper the whole file uses, so renaming the CTE is a rename again.

Went with the predicate over exporting one next to `SUCCESSOR_SEEK_QUERY`: the bound parameter is already part of the contract, whereas a production export would exist only for the tests.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
0xdeafcafe: Fixed in a327b22453. New block: one chunk carrying two series staggered hours apart, each already holding stored points between its own chunk points and a stored successor past its span end, folded in a single `recomputeAffectedRollupsMany` — the shape `bulkAppend` hands this lane. Both series' rollups are compared against reference series rebuilt strictly one point at a time, and every sample of both is asserted counted exactly once.

The staggering earns its place twice over: it is what makes the three chunk-global bounds three different values, so a bound taken from the wrong series surfaces as a missing successor instead of cancelling out.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.integration.test.ts
0xdeafcafe: Fixed in a327b22453. `perPointSeriesId` is seeded entirely through `recomputeAffectedRollups` now — both halves, not just the arriving one — so the reference never touches the folded entry point. Same for the two reference series in the new multi-series block.

Agreed that the shared `SUCCESSOR_SEEK_QUERY` limits what the comparison can prove on its own, which is why the reference is a per-point rebuild rather than a second fold.

specs/otlp/canonical-metric-ingestion.feature
0xdeafcafe: Fixed in a327b22453. The Rule is now "Rebuilding summaries sends a request bounded independently of the batch", with a preamble saying bounded is not constant, and why the summary-window read stays batch-proportional by design — so a reader cannot take it as licence to fold that read into the joined form.

The byte-for-byte claim stays where it belongs, in the statement scenarios' own Thens. And the property that matters is a scenario of its own now — "A folded rollup read keeps its encoded request inside a budget" — bound to tests that measure `request.params` with the client's own `formatQueryParams` across 1 to 512 series and at the adversarial 64-series full-width shape.

specs/otlp/canonical-metric-ingestion.feature
0xdeafcafe: Fixed in a327b22453. Retitled to "A folded rollup read leaves the stored payload behind" and reduced to the payload Then, so the scenario carries one invariant and it is the #6493 memory guarantee rather than a restatement of the scenario above. The bound test's `@scenario` moved with the title in the same commit; parity reports 8/8 for the file.

specs/otlp/canonical-metric-ingestion.feature
0xdeafcafe: Fixed in a327b22453. Now "A rollup bucket read sends the window size and retention span once", with the observable Then — "the request carries the window size and the retention span once, not once per window" — in the corpus phrasing you cite. A second Then covers the size claim, bound to the new bucket-read measurement test.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -51,6 +51,118 @@ const INSERT_SETTINGS = { async_insert: 1, wait_for_async_insert: 1 } as const;
...
+           AND series_points.PointId > spans.Span${bound}Point))))`;
+}
+
+function orderedBefore(bound: "From" | "To"): string {
+  return `(series_points.SeekTime < spans.Span${bound}Time
+     OR (series_points.SeekTime = spans.Span${bound}Time
+       AND (series_points.TimeUnixNano < spans.Span${bound}Nano
+         OR (series_points.TimeUnixNano = spans.Span${bound}Nano
+           AND series_points.PointId < spans.Span${bound}Point))))`;
+}
```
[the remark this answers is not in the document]
0xdeafcafe: Not applicable — neither field exists.

`grep -rn SchemaVersion platform/app/src/server/app-layer/metrics/repositories/` returns nothing, and `seriesIdMetricNameAttributesKeyTuple` has no occurrence anywhere under `platform/app/src`. `metric-data-point.rows.ts` writes neither. `SUCCESSOR_SEEK_QUERY` selects `SEEK_SELECT`'s six columns plus the `SeekTime` alias, and nothing else; the statement is exercised against a real ClickHouse by the repository's integration test in CI.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
```diff
@@ -418,11 +482,653 @@ describe("MetricDataPointClickHouseRepository", () => {
...
+          PARAM_BUDGET_CHARS,
+        );
+        expect(
+          (request.params.seriesIds as string[]).length,
+        ).toBeLessThanOrEqual(64);
+      }
+      // Split, not dropped: every series still gets asked about exactly once.
+      expect(
+        requests.flatMap((request) => request.params.seriesIds as string[]),
+      ).toHaveLength(64);
```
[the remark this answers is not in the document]
0xdeafcafe: Right, and the comment above it was making the identity claim while the assertion only counted. Fixed in f7db857583: the flattened `seriesIds` now has to be a set of 64 as well as a list of 64, so a chunker that asked twice about one series and never about another fails.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
```diff
@@ -125,6 +288,152 @@ function affectedBucketsForSeries({
...
+ * there rather than silently shipping an oversized request. Over-measuring only
+ * costs a smaller chunk.
+ */
+function formatParamValue({
+  value,
+  inArray = false,
+}: {
+  value: unknown;
+  inArray?: boolean;
+}): string {
```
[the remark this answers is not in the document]
0xdeafcafe: Correct — renamed to `isInArray` in f7db857583, at the parameter, the type and both call sites.

platform/app/src/server/app-layer/metrics/repositories/__tests__/metric-data-point.clickhouse.repository.unit.test.ts
```diff
@@ -418,11 +482,657 @@ describe("MetricDataPointClickHouseRepository", () => {
...
+        ).toBeLessThanOrEqual(64);
+      }
+      // Split, not dropped: every series still gets asked about exactly once,
+      // which is an identity claim rather than a count — a chunker that asked
+      // twice about one series and never about another also totals 64.
+      const asked = requests.flatMap(
+        (request) => request.params.seriesIds as string[],
+      );
+      expect(new Set(asked).size).toBe(64);
+      expect(asked).toHaveLength(64);
```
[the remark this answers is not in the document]
0xdeafcafe: Fair — fixed in ce8ef4af59. The assertion now compares the asked-about set to the set of series ids the chunk actually carried, so requesting a series nobody sent fails it. The exactly-once claim is kept alongside it.

0xdeafcafe:
**Review applied.** 16 of the 18 findings are in; 2 are deferred with the thread left open.

**Applied** — `a327b22453`, plus `f7db857583` / `ce8ef4af59` for three follow-on bot points:

- **The size P1.** Chunking splits on encoded parameter bytes now, measured the way the pinned client measures them, against an explicit 3,500-character budget with headroom under `MAX_URL_BIND_PARAM_LENGTH`; the seek cap stays as the upper bound on series. Three tests re-measure the emitted requests with the client's own `formatQueryParams`, including the adversarial 64-series full-width shape. `use_multipart_params_auto` stays out — it is a client-wide framing change, and the byte budget holds either way.
- **The three test-gap findings.** Multi-series integration fold (two series staggered hours apart, stored successors past each span, compared against per-point rebuilds); both discriminating unit fixtures; and every reference series now seeded strictly through the per-point entry point.
- **One thing the review found that ran deeper than the finding said.** The equivalence fixtures were gauges, and a gauge successor never pulls a second bucket into the affected set — so every assertion in that block held with the successor read deleted entirely, not merely with its within-span half suppressed. They are cumulative now. Four mutation probes were re-run (drop within-span; flip each of the three constant bounds min↔max) and all four now fail a test; all four survived before.
- **All P2/P3s.** CTE double-execution stated; nine→eleven and ~1 KB→~2.4 KB; `SEEKS_PER_QUERY` and `orderedAfter` docs; `earliestSpanEnd` / `latestSpanEnd` rename carried across params, SQL, docs and tests; mocks discriminate on a bound parameter instead of a CTE name; the Rule narrowed to a bounded request with the encoded-budget property as its own scenario; two scenarios retitled around the invariant they assert; the "Storage trouble" title collision with `canonical-log-ingestion.feature` disambiguated; `specs/README.md` corrected to enforce-by-default with the two ratcheted deny-lists.
- PR body updated: the size claim is true by construction now, and it describes the rule in force rather than the pre-fix gap.

**Deferred, threads left open** — the two scan-bound P1s (branch 2 unbounded per series, branch 1's response tracking span width). Both are the same piece of work: bound the join side per series with a fallback seek for what that misses. Correctness is unaffected today (the minimum-of-the-union argument holds), what it costs is read volume. It is deferred because its equivalence cannot be shown by the parameter-level model these unit tests use, and the `EXPLAIN indexes=1` before/after it needs a real server. The behaviour is now documented at the site so nobody reads the branch as bounded.

Also worth recording: the fifteen inline comments posted as unexpanded file references rather than their text. The bodies have been restored in place, so the threads read as intended.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Deferred from this PR, now tracked in #6628 — per-series bounded seek with fallback, needs server-side EXPLAIN to prove equivalence. Behavior documented at the site in this branch.

platform/app/src/server/app-layer/metrics/repositories/metric-data-point.clickhouse.repository.ts
0xdeafcafe: Deferred from this PR, now tracked in #6628 (same piece of work as the past-span bound). Behavior documented at the site in this branch.

PR #6629, opened 2026-08-06
chore: remove reduced graphics menu option

rogeriochaves:
## What

Removes the "Reduced graphics (Auto)" entry from the user avatar menu, along with the automatic frame-rate probe and the manual Auto / On / Off override that backed it. All three choices lived behind the single menu item being dropped, so the whole mechanism goes with it, not just the menu markup.

Removed:
- The menu item and its Auto / On / Off submenu in `DashboardLayout.tsx`
- `GraphicsQualityProvider` (the background FPS probe and the `data-reduced-graphics` attribute writer) and its wiring into `AppProviders.tsx`
- `useGraphicsQuality` hook and context
- `graphicsQualityOverrideStore` (the per-device localStorage preference)
- `evaluateFpsSample` (the struggling/smooth frame-rate heuristic)
- The `html[data-reduced-graphics="true"]` CSS override and the `--lw-backdrop-blur` / `--lw-panel-alpha` CSS variable indirection in the tooltip, dialog, select, drawer and toast recipes (`_app.tsx`), and in `dialog.tsx`, `drawer.tsx`, `GroupRow.tsx`, `RunRow.tsx`
- The feature spec (`specs/components/adaptive-graphics-quality.feature`) and every test scoped to this behavior: a dedicated test file each for the provider, the store and the FPS heuristic, plus the backdrop-blur-variable assertions inside the dialog, drawer, GroupRow and RunRow tests

## Why this is safe

Every place that branched on the setting only ever had two outcomes: reduced or full. Full is what a healthy device with default settings already rendered, so each CSS custom property now resolves permanently to what used to be its own fallback value instead of being switched by an attribute. `--lw-backdrop-blur` is gone and its call sites use the literal blur value unconditionally (`blur(8px)`, `blur(12px)`, `blur(25px)`, etc.); `--lw-panel-alpha` is gone and its call sites use the literal alpha percentage. Decorative blur and panel transparency now always render, the same visual result the "Off" choice (and a device with no performance problem) already produced.

## prefers-reduced-motion

Not touched, and it did not need to be. This feature was never driven by `prefers-reduced-motion`: its own comments were explicit that the FPS probe "runs regardless of the prefers-reduced-motion setting," because backdrop blur is a static effect, not motion. The actual accessibility behavior, the `useReducedMotion` hook and the `prefers-reduced-motion` media queries used across Langy, traces-v2 and other components, lives entirely outside this feature and is untouched.

## Verified

- `pnpm typecheck`
- `./node_modules/.bin/tsgo --noEmit --project ./tsconfig.tsgo.tests.json` (from `platform/app`)
- `pnpm --filter @langwatch/web exec biome check --write` on every touched file (only pre-existing, unrelated complexity warnings on functions that shrank)
- `pnpm test:integration` on the four touched integration test files, 27 tests passing
- `pnpm test:unit run` on the two unit tests that read `DashboardLayout.tsx`'s source text directly
- `npx tsx platform/app/scripts/check-feature-parity.ts`, still green

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified visual effects by using consistent panel transparency and blur settings across dialogs, drawers, run rows, and group headers.
  * Removed automatic graphics-quality adjustments that could disable blur or alter panel appearance.

* **Refactor**
  * Removed the reduced-graphics settings and related controls from the application interface.
  * Updated global styling recipes to use fixed visual values.


0xdeafcafe (approved):
this was added due to several real user reports that blur was causing really really bad performance (my guess is crap enterprise windows machines), but just flagging that this was added for a specific reason

langwatch-agent:
## Human Review Brief

**Mode: Targeted Review.** No linked issue; the PR is a deletion and the body enumerates every removed piece, which is the right record for one.

### Decisions being ratified

1. **The whole adaptive-graphics mechanism goes, not just the menu entry.** The frame-rate probe, the per-device localStorage override and the `data-reduced-graphics` attribute all leave together. Bringing any of it back later means rebuilding the probe, not re-adding a menu item.
2. **The `--lw-backdrop-blur` / `--lw-panel-alpha` indirection is removed from the tooltip, dialog, select, drawer and toast recipes.** Those variables were the seam the feature hooked into; without them the styling values are literal again, and a future runtime theming feature has no hook.
3. **The feature spec and its tests are deleted rather than skipped**, so the behaviour stops being described anywhere.

### Must Check

- Grep for `--lw-backdrop-blur`, `--lw-panel-alpha` and `data-reduced-graphics` outside the removed files. Anything still reading them renders with an unset variable, which fails silently rather than loudly.
- The four component tests that lost their backdrop-blur assertions (`dialog`, `drawer`, `GroupRow`, `RunRow`): confirm the assertions were removed rather than the tests weakened around a value that still matters.

This one is already green, approved and `CLEAN`, so it is in today's merge queue.

Issue #1915, opened 2026-03-03
Cleanup: remove unused findByIdIncludingArchived in agent repository


drewdrewthis:
## Description

`findByIdIncludingArchived` in `agent.repository.ts:135-143` appears to be defined but unused. If confirmed unused across the codebase, remove it.

## Location

`langwatch/src/server/agents/agent.repository.ts:135-143`

## Found during

Review of #1908.

Issue #1914, opened 2026-03-03
Bug: EvaluationGroupList uses non-unique React key


drewdrewthis:
## Description

`EvaluationGroupList` in `Evaluations.tsx:73` uses `group.latest.evaluation_id` as the React key. If two groups share the same latest evaluation_id (unlikely but not prevented by types), React will silently drop one. Use a composite key like `${group.evaluatorId}-${group.latest.evaluation_id}`.

## Location

`langwatch/src/components/traces/Evaluations.tsx:73`

## Found during

Review of #1908.

shivama205:
Hi, I'd like to work on this issue. I've submitted a fix in #2017 — would appreciate a review when you get a chance. cc @drewdrewthis @rogeriochaves

shivama205:
Tagging @sergioestebance for review as well — PR: #2017

PR #6667, opened 2026-08-07
fix(security): bump grpc to 1.82.1 in the sdks/go modules

langwatch-agent:
Clears **3 HIGH alerts**: #1914, #1915 and #1916, one per Go module, all the same advisory (gRPC-Go xDS RBAC and HTTP/2 vulnerabilities, patched in 1.82.1).

Bumps `google.golang.org/grpc` 1.80.0 to 1.82.1 in `sdks/go`, `sdks/go/examples` and `sdks/go/e2e`. The root module is already on 1.82.1, so it is untouched.

## Why this PR exists rather than Dependabot's

Dependabot opened #6316 for exactly this bump, and I dogfooded it and recommended it for merge. The monorepo restructure then moved `sdk-go/` to `sdks/go/`, so all six files that PR touches no longer exist and it can never merge. Dependabot has not re-raised it at the new path. This is the same change re-homed, and #6316 should be closed in its favour.

## Impact

`grpc` is an **indirect** dependency in all three modules, with **zero** direct `google.golang.org/grpc` imports anywhere in the Go SDK; it arrives underneath the OTLP gRPC exporter. The advisory is in xDS RBAC and HTTP/2 handling, neither of which the SDK configures.

## Testing

The diff is minimal: grpc plus the two `google.golang.org/genproto/googleapis` transitives that grpc itself requires. Nothing else moves in any of the three `go.mod` files.

- `go vet ./...` in all three modules: **clean**. `examples` and `e2e` need `GOWORK=off`, because `go.work` deliberately excludes them (it says so: "Example/e2e modules under sdks/go keep their own relative `replace => ../` and are intentionally left out of the workspace").
- `go test ./...` in `sdks/go`: **exit 0**, 3 packages ok, 0 failures.

## Summary by CodeRabbit

* **Chores**
  * Updated Go SDK supporting dependencies to newer Google API, RPC, and gRPC versions.
  * Upgraded gRPC to v1.82.1 across the SDK, examples, and end-to-end test modules.
  * Incorporated the latest available GenProto revisions for improved compatibility and maintenance.


langwatch-agent:
Closing: main already carries this bump, so the PR is now a no-op that would only regress other modules.

`origin/main` has `google.golang.org/grpc v1.82.1` in all three modules (`sdks/go/go.mod`, `sdks/go/examples/go.mod`, `sdks/go/e2e/go.mod`), which is the version this branch was raising it to. Rebasing confirms it: the branch no longer adds a grpc delta, and because it predates the recent SDK work it would *downgrade* `go.opentelemetry.io/otel` 1.44.0 to 1.43.0, `openai-go/v3` v3.50.0 back to `openai-go` v1.12.0, `grpc-gateway/v2` 2.29.0 to 2.28.0 and `tidwall/gjson` 1.19.0 to 1.18.0.

No open Dependabot alert points at any `sdks/go` manifest anymore, so the vulnerability this PR existed to close is already closed on main. Nothing to carry forward.

PR #6697, opened 2026-08-07
chore(deps): upgrade Prisma to 7.9.1 and tRPC to 11 (TanStack Query 5)

0xdeafcafe:
## Why

The app's two remaining major-version-pinned stacks were years behind: Prisma 5.7.1 (released Dec 2023) and tRPC 10.45 / TanStack Query 4. Staying pinned costs us React-19-era support, keeps a 35 MB Rust query engine and a per-process engine sidecar in the runtime image, and blocks the maintained majors' semantics (`useUtils`, `isPending`, `placeholderData`). This upgrades all three to current.

No tracking issue — self-initiated dependency work.

### Prisma 5.7.1 → 7.9.1
- New `prisma-client` generator emitting to `src/generated/prisma` (imports move from `@prisma/client` to `~/generated/prisma/client`); the generated client is **not** committed — `/src/generated/prisma/` is gitignored and `pnpm start:prepare:files` produces it, like the other generated artifacts.
- Runtime connects through the `@prisma/adapter-pg` driver adapter (Prisma 7 removed the Rust engine); the CLI reads `prisma.config.ts`.
- Prisma 7 removed `$use`: the enMasse / projectId / organizationId guard chain runs as a client query extension with the same `(params, next)` calling convention.
- `Prisma.dmmf` is gone: the tenancy partition tests parse `prisma/schema.prisma` via `src/test-utils/prismaDatamodel.ts`.

### tRPC 10.45 → 11.18 + TanStack Query 4 → 5.101
- Transformer moved into each terminating link; vanilla client is the v11 proxy client (the v10 `subscription`-collision escape hatches in api.tsx and langyChatTransport are gone).
- The typed permission-procedure builder in `trpc.ts` is ported to v11's builder generics (`unstable-core-do-not-import`; the `@trpc-internal/*` tsconfig aliases are removed).
- Mechanical renames: `useContext()` → `useUtils()`, `cacheTime` → `gcTime`, `isInitialLoading` → `isLoading`, `isPreviousData` → `isPlaceholderData`, `keepPreviousData: true` → `placeholderData: keepPreviousData`, mutation `isLoading` → `isPending` (118 sites).
- Query `onSuccess`/`onError` callbacks (removed in RQ5) became per-fetch effects keyed on `dataUpdatedAt`/`errorUpdatedAt` (6 sites); `refetchInterval` callbacks take the query object (9 sites). Mutation `onSuccess`/`onError` are untouched — RQ5 still supports those.

### Also here: pool tuning + lazy client construction
- **Pool tuning URL params reach the pg pool.** node-postgres also ignores `connection_limit` and `pool_timeout`, and its defaults differ from the engine's (max 10 vs `cpus*2+1`, wait-forever acquisition vs 10s). `pgPoolConfig()` maps both onto the pg Pool config so a tuned DATABASE_URL keeps meaning what it says. Spec: `specs/server/prisma-driver-adapter.feature`.
- **Importing `~/server/db` no longer constructs the client.** The export is a proxy that builds the guarded client (and its pg pool) on first property access. Half the server graph reaches this module transitively; scripts, workers and unit suites that never touch Postgres no longer pay for a pool. Prisma memoizes model delegates, so delegate identity (and `vi.spyOn(prisma.model, ...)`) is unaffected.

## How it works

Four behaviour changes this upgrade would otherwise have shipped silently. None are visible to `tsc` — each was found at runtime.

1. **`?schema=` loss on the pg driver adapter.** The adapter ignores the connection URL's `?schema=` param the Rust engine honoured: migrate wrote to `langwatch_db` while the runtime client queried `public`, and unqualified raw SQL lost its search_path. `createPrismaPgAdapter` (every construction site goes through it) passes the schema as the adapter option **and** the session `search_path`.
2. **P2002 metadata moved.** Constraint fields now live at `meta.driverAdapterError.cause.constraint` (quoted) instead of `meta.target`, silently degrading every 409-conflict handler to a 500. `uniqueConstraintTargets()` bridges both shapes; the prompts / gateway / REST handlers use it.
3. **Node runtime in the browser bundle.** The generated client's `client.ts` hard-imports `@prisma/client/runtime/client` → `node:url`; frontend enum/type imports pulled it into the browser graph and blanked **every page** (the old `@prisma/client` package had a `browser` field; generated source has no package.json). Vite now aliases `~/generated/prisma/client` → the generated `browser.ts` for browser bundles, and a new **browser-mode smoke test** mounts the real provider stack + router so this class fails a test instead of production. jsdom suites run in node and can never catch it.
4. **Raw-SQL tenancy guard expansion.** v10's `$use` never saw template-literal raw queries inside interactive transactions; the v11 query extension does. The advisory locks, the worker `SELECT 1` probe and the outbox claim now carry the guard's sanctioned `-- @tenancy:` markers (full AST sweep verified every raw site is token- or marker-compliant).

## How to review this

The file count is the problem, and it defeats the automation: CodeRabbit and Copilot both hard-skip above 300 files, and the low-risk auto-approve gate declines to evaluate a diff this size. All three want a human. So here is the tractable partition.

**881 of the 1,188 files are pure codemod** — every changed line in them is a mechanical rename or an import swap (`@prisma/client` → `~/generated/prisma/client`, `useContext` → `useUtils`, `cacheTime` → `gcTime`, `isInitialLoading` → `isLoading`, `isPreviousData` → `isPlaceholderData`, `keepPreviousData: true` → `placeholderData: keepPreviousData`, mutation `isLoading` → `isPending`). They are verifiable by pattern, and typecheck plus the unit suite covers them.

**The real surface is ~20 files.** In descending order of how much thought each needs:

| File | What to check |
|---|---|
| `src/server/db.ts` | The guard chain as a v11 query extension, and the lazy proxy |
| `src/server/api/trpc.ts` | Permission-procedure builder ported to v11 builder generics |
| `src/server/prismaPgAdapter.ts` | `?schema=` → adapter option + `search_path`; `connection_limit`/`pool_timeout` mapping |
| `src/server/utils/prismaErrors.ts` | P2002 metadata bridged across both shapes |
| `src/utils/dbGuardMiddleware.ts` | `(params, next)` convention preserved under the extension |
| `src/features/traces-v2/hooks/useTraceNewCount.ts` | Query callbacks → per-fetch effects keyed on `dataUpdatedAt`/`errorUpdatedAt` |
| `src/features/langy/logic/langyChatTransport.ts` | v10 subscription-collision escape hatch removed |
| `src/utils/api.tsx` | Transformer moved into the terminating links |
| `prisma.config.ts`, `infra/docker/Dockerfile` | CLI config; dropped post-prune `prisma generate` |
| `scripts/ops/purge-process-manager-tables.mjs` | Standalone `.mjs` ported to `pg` (can't load the TS client) |

Their tests sit alongside them and are worth reading as the specification: `db.guard-wiring.integration.test.ts`, `prismaPgAdapter.unit.test.ts`, `db.lazy.unit.test.ts`, `appEntry.smoke.browser.test.tsx`, and `specs/server/prisma-driver-adapter.feature`.

If you only have time for two files, make them `db.ts` and `prismaPgAdapter.ts` — that's where a mistake would be both silent and tenancy-relevant.

## Test plan

- **Full CI matrix**: 26 workflows including langwatch-app-ci's 20 jobs (typecheck, lint, build, 4 unit shards, 6 integration shards), e2e-ci, sdk-javascript-ci, gateway-matrix, langwatch-chart, codeql. **All 26 workflows and all 20 app-ci jobs pass on the current merge head `dcf5257388`**, verified at job level rather than by the workflow rollup. (They also passed on the pre-merge head `f9f2bb97b6`.)
- Full unit suite run locally: **27,741 passed** (1 file timed out only under parallel-worktree contention; passes 6/6 in isolation). 83 server-side integration files run against native Postgres/ClickHouse/Redis: 79 passed; the 3 failures are local-environment gaps (12 assertions need the Go NLP engine running, and `orchestrator` trips over main's own `require("~/server/db")` alias from #6405) — all three pass in CI.
- `test-integration` green in CI (the five originally-failing files were fixed by items 1/2/4 above).
- `prisma migrate deploy` replays the full 241-migration history cleanly under the Prisma 7 CLI. Image-entrypoint path re-verified on a fresh database with env-only DATABASE_URL: 241 migrations, 107 tables in `langwatch_db`, `migrate status` clean.
- **Tenancy guard wiring pinned**: `db.guard-wiring.integration.test.ts` drives six scenarios through the real extended client — project/org refusals, the mass-delete safe word (refusal AND rewrite, previously untested anywhere), and the raw-SQL guard directly and inside an interactive transaction (the surface v10's `$use` never saw).
- New browser-mode smoke: `src/__tests__/appEntry.smoke.browser.test.tsx`.
- **Live drive on the production bundle**: built `dist/server`, booted the stack under haven on a fresh per-worktree database (full migration replay, seed, all through the Prisma 7 CLI + adapter), logged in and walked 12 pages via playwright, ingested a trace through /api/collector (collector → in-process workers → ClickHouse), and verified it renders in both the legacy Traces view and Trace Explorer (span tree, drawer, output). Zero 5xx, zero page errors, zero error boundaries; server logs clean. The one console warning found ("Maximum update depth" on the empty-project home) reproduces identically on main.

## Anything surprising?

**Keeping up with main is the recurring cost.** Main lands ~10 commits a day on overlapping files, so this PR goes `CONFLICTING` overnight — and a conflicted PR runs **zero** `pull_request` workflows, so its last green reads as current when nothing is actually running. Two incidental changes on this branch made that worse and have now been reverted: `platform/app/package.json` had been reformatted to tab indentation (biome's JSON default) and `vite.config.ts` re-wrapped, each of which turned a small delta into a whole-file conflict every single sync. Both files now differ from main by only their real content — 8 dependency lines and one alias block.

Every sync also needs an **absorb pass**, because new-main code written against the old stack merges cleanly without compiling. This round: main's new `scripts/report-duplicate-subscriptions.ts` (a near-copy of its ported sibling, but on `@prisma/client` + bare `new PrismaClient()`), and a hand-merge of `useTraceNewCount.unit.test.ts` where main added three scenarios driven through `options.onSuccess`/`onError` — callbacks RQ5 removed — on top of our per-fetch-effect mechanism.

**The CodeQL alert this PR used to carry has cleared, and why is instructive.** `js/bad-code-sanitization` flagged `renderBuiltUrl` in `vite.config.ts` — main's own ADR-086 asset-base code, which `origin/main` carries verbatim. The alert only ever existed because the branch had re-wrapped the whole file, so CodeQL saw main's pre-existing code as an addition. Reverting that incidental reformat shrank the file's diff to the one alias block and the alert went away on its own. The branch now reports **zero** code-scanning alerts in any state.

### Docker images
- This repo's `infra/docker/Dockerfile` drops the post-prune `prisma generate` (the v7 client is app source inlined into the bundles; prune can't touch it) and the retired `PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING`.
- The saas runtime image is adapted in langwatch/langwatch-saas#1006 (copies `prisma.config.ts`, drops the musl regenerate, fixes the GDPR task's client construction). **That PR must land with the submodule bump that includes this one** — build fails one way, boot fails the other.

### Before / after — the honest numbers

| Surface | Prisma 5.7 (before) | Prisma 7.9 (after) | Verdict |
|---|---|---|---|
| Client type surface | One generated `index.d.ts`: **7.7 MB, ~184k lines** — every hover/import parsed the whole thing | **111 files, ~227k lines (11 MB)** in `src/generated/prisma`, one file per model | Sharded, not shrunk. The editor parses only the models you touch — hover/completion in DB-heavy files is where you feel it. Whole-project tsgo still loads the full graph (92 fully-connected models), so bulk typecheck was never going to move. |
| CI typecheck | 103–358 s (last four main runs) | 156–206 s | Within main's own run-to-run spread. No regression, no headline win. |
| CI build / unit / integration | build 109–309 s · unit 192–453 s · integration 630–1049 s | build 102–121 s · unit 212–259 s · integration 613–738 s | Same spread both ways. |
| Query engine | **35 MB** of Rust engine binaries in the runtime image + a query-engine sidecar per Node process | **None** — pure TS client over the `pg` driver; the 23 MB schema engine remains as a dev/migrate-time dep only | One fewer process per pod, no engine startup, no `PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING` hacks. |
| Client startup | Connected eagerly at import | Lazy proxy singleton — nothing touches Postgres until the first query | Workers/scripts that never query no longer open pools. |

tRPC 11 + TanStack Query 5 bring no headline perf number — the wins are the maintained majors themselves plus one correctness upgrade: the raw-SQL tenancy guard now also sees queries issued **inside** interactive transactions, which v10's `$use` never did.

`sdk-javascript-complete` (span-ingestion e2e), which fails on main whenever it isn't path-skipped, has passed on every head since — no longer an open caveat for this PR.


langwatch-agent:
## Human Review Brief

**Mode: Deep Review.** No linked issue, stated as self-initiated dependency work; the PR body carries the intent, so the gap is fine. 1188 files, but the review is not 1188 decisions: it is four, and most of the diff is import rewriting around them.

### Decisions being ratified

1. **The tenancy guard moves from `$use` middleware to a client query extension.** Same `(params, next)` convention, different interception point. This is the isolation boundary, so approving it makes the extension the thing standing between tenants.
2. **The generated Prisma client is not committed.** `src/generated/prisma/` is gitignored and produced by `pnpm start:prepare:files`. That becomes an interface every environment has to honour: CI, the runtime image, preview VMs, and a fresh clone running typecheck.
3. **The runtime connects through `@prisma/adapter-pg`.** Pooling, timeouts and error shapes now come from `pg` rather than the Rust engine that was removed. Connection behaviour under load is not the same code any more.
4. **Tenancy partition tests parse `prisma/schema.prisma` text** via a new `src/test-utils/prismaDatamodel.ts`, because `Prisma.dmmf` is gone. The tests' fidelity now depends on that parser rather than on the client's own model.

### Must Check

- **Extension coverage versus `$use`.** Confirm the extension fires inside `$transaction` callbacks and on nested writes, and that raw queries are excluded in the same way they were before. A gap here is silent cross-tenant reads, not an error.
- **TanStack Query 5 removals.** `keepPreviousData` and `cacheTime` are gone in v5. Any list or table that relied on `keepPreviousData` for paging degrades to a flash of empty state and still typechecks if it was migrated to `placeholderData` incorrectly.
- **Which files are not mechanical.** Isolating the non-import-rewrite changes is the whole review; everything else can be skimmed.

### Ask Author

- Which environments run `start:prepare:files`, and what does a fresh clone plus `pnpm typecheck` do today?
- Were the tenancy partition tests run green against the extension, or do they now exercise the parser rather than the guard?

### Production risk

The guard rewrite is the highest-consequence part of the PR. A miss there is cross-tenant data exposure with no error and no log, which is a different class of failure from everything else in the diff.

langwatch-agent (commented):
No blocking finding in the tenancy-boundary migration reviewed. The Prisma 7 query extension preserves the former guard ordering and covers all-model plus raw operations; the added integration test exercises the exported client, including an interactive-transaction raw query. The new generated-client build path runs generation before bundling and keeps raw SQL on the configured schema search path. CI is still in progress and the PR is currently marked dirty, so those remain merge prerequisites.

rogeriochaves (approved):

PR #6722, opened 2026-08-07
chore(main): release typescript-sdk 1.3.0

langwatch-agent:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* **evaluators:** evaluations, monitors and experiments referencing a legacy/ragas_* evaluator type stop working. Their current equivalents in the ragas/* family remain available.

### Features

* **coding-agent:** session economics, and the repositories and pull requests behind them ([#6435](https://github.com/langwatch/langwatch/issues/6435)) ([6430813](https://github.com/langwatch/langwatch/commit/6430813056cda513241a8389a2136fac2d63591e))
* **evaluators:** remove the legacy Ragas evaluators ([#6600](https://github.com/langwatch/langwatch/issues/6600)) ([ef9ea90](https://github.com/langwatch/langwatch/commit/ef9ea90e22bc2adb92bacf5c732cc996c9782bfe))

### Miscellaneous

* **release:** single-footer shadow Release-As 1.3.0 for the typescript SDK ([#6734](https://github.com/langwatch/langwatch/issues/6734)) ([79cd276](https://github.com/langwatch/langwatch/commit/79cd276a940ad8dacf72cab7657f48d01929e048))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #6710, opened 2026-08-07
chore(main): release langwatch 3.11.0

langwatch-agent:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* **spend:** POST /api/gateway/v1/virtual-keys refuses an organization- or team-scoped key, and a key scoped to more than one project, when it does not carry trace_project_id and the organization has projects to choose from. Send trace_project_id, or scope the key to exactly one project. Existing keys are unaffected until they are next edited.
* **gateway:** POST /api/gateway/v1/virtual-keys refuses an organization- or team-scoped key, and a key scoped to more than one project, when it does not carry trace_project_id and the organization has projects to choose from. Send trace_project_id, or scope the key to exactly one project. Existing keys are unaffected until they are next edited.
* **evaluators:** evaluations, monitors and experiments referencing a legacy/ragas_* evaluator type stop working. Their current equivalents in the ragas/* family remain available.
* **mcp-server:** three behaviors change for callers of the HTTP server. A client that authenticated once at initialize and then sent only the session id must now send Authorization: Bearer <key> on every request. GET /sse?apiKey=<key> no longer authenticates; use the Authorization header. A deployment relying on the implicit 0.0.0.0 bind must pass --host 0.0.0.0, which the published container image already does.

### Features

* **agent-plugin:** first-class LangWatch plugin for Claude Code and Agent Plugins 1.0 ([#6747](https://github.com/langwatch/langwatch/issues/6747)) ([25625c4](https://github.com/langwatch/langwatch/commit/25625c44e53d6dd23d647e58fad9783464e8a1e2))
* **cli:** keep the LangWatch Claude Code plugin up to date from every wrapped run ([#6751](https://github.com/langwatch/langwatch/issues/6751)) ([b64a584](https://github.com/langwatch/langwatch/commit/b64a58465e7d35fcc2ae6ba22d9d4ca54e66b7a4))
* **codex:** capture the conversation for a plain codex, not just the wrapper ([#6433](https://github.com/langwatch/langwatch/issues/6433)) ([38931b4](https://github.com/langwatch/langwatch/commit/38931b4392837c4a011dbde44ec7b80c80b19720))
* **coding-agent:** session economics, and the repositories and pull requests behind them ([#6435](https://github.com/langwatch/langwatch/issues/6435)) ([6430813](https://github.com/langwatch/langwatch/commit/6430813056cda513241a8389a2136fac2d63591e))
[...]

PR #3626, opened 2026-04-30
chore(main): release langevals 3.0.0

0xdeafcafe:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* **evaluators:** evaluations, monitors and experiments referencing a legacy/ragas_* evaluator type stop working. Their current equivalents in the ragas/* family remain available.

### Features

* **evaluators:** remove the legacy Ragas evaluators ([#6600](https://github.com/langwatch/langwatch/issues/6600)) ([ef9ea90](https://github.com/langwatch/langwatch/commit/ef9ea90e22bc2adb92bacf5c732cc996c9782bfe))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #6736, opened 2026-08-08
ci(release): keep a breaking change inside one release component

rogeriochaves:
## The problem

release-please decides which components a commit affects **by path**, then applies the **whole commit message** to every one of them. There is no way to express "this break is only about the SDK". So one commit carrying a `!` marker or a `BREAKING CHANGE:` footer, touching three components, proposes three majors.

Twice in the last 400 commits on `main`:

| commit | what it was | what it hit |
| --- | --- | --- |
| `ef9ea90` (#6600) | remove the legacy Ragas evaluators, correctly breaking for the product | `langwatch` **and** `typescript-sdk` **and** `langevals`. The SDK's share of that commit was one doc comment and one unit test, and it produced #6722 proposing `typescript-sdk 2.0.0` |
| `af324de` (#6641) | MCP server auth change | `mcp-server` **and** `langwatch`, via **one file**, `SECURITY.md`. Fifteen of its sixteen files were under `mcp/typescript` |

That second one is worth dwelling on. **#6710 `langwatch 4.0.0` is correct, but partly for the wrong reason.** It lists two breaking changes. The evaluator removal genuinely breaks the product, so the major stands and this PR does not touch it. The MCP auth one reached the product changelog only because a sixteenth file happened to be at the repo root. A single incidental edit is enough. That is the strongest argument for this guard existing.

The remedy so far has been to notice afterwards and pin the version by hand: #3627 for six components at once, #6704 for the root package, #6734 for the typescript SDK yesterday. That is three cleanup PRs for something detectable before merge.

## The guard

`release-scope-guard` fails a PR whose title or commits carry a breaking marker when its changed files span more than one release component.

It reimplements release-please's own path rules rather than approximating them:

- longest package prefix wins, so a file counts for exactly one non-root component (`util/commit-split.ts`)
- top-level files belong to the root package only, which is precisely the `SECURITY.md` case
- a component is spared only when **every** file it owns sits under one of its `exclude-paths` (`util/commit-exclude.ts` uses `.every()`)

Squash merge puts the PR title in the subject and the branch commit messages in the body, so those are the two places a marker can reach `main` from, and those are what it reads. Both arrive as JSON strings, one per line, so a crafted title cannot escape into the shell or into the guard input.

The failure message names the components, their current versions from the manifest, and the two ways out, because a check that only says "blocked" gets escape-hatched by reflex:

```
- typescript-sdk (sdks/typescript), now 1.2.0
- langevals (services/langevals), now 2.2.0
- langwatch (.), now 3.10.0

If the break really does apply to all of them, add the
`multi-component-major` label and this check passes.

Otherwise pick one:

1. Split the change so the breaking commit touches one
   component, which is the cheaper fix before merge.
2. Keep it together and pin the components that must not go
   major. Add a follow-up commit per component that edits
   only that component's shim and carries one footer:

   sdks/typescript/.release-please-shim   +   Release-As: <x.y.z>
   ...
```

`multi-component-major` is the deliberate override, for a break that really does apply everywhere it landed.

## Evidence

Replayed over the **last 400 commits on `main`**. Four carry a breaking marker. The guard fires on **exactly two**, `ef9ea90` and `af324de`, and names precisely the components that really were bumped in each case. **Zero false positives.** The other two breaking commits, `60d98f9` and `75ae272`, it correctly passes as single-component.

The full workflow pipeline was also run end to end against the live API for #6600 as it looked at review time, with its fifteen individual commits rather than the squashed message. It fails and names all three components, so the guard would have caught this before the bad majors were ever proposed.

It also has to not fire on release-please's own PRs, which touch the manifest, every changelog and the charts at once. Ran against all four open ones, #6710, #6722, #3626 and #4025: all pass. The reason is deliberate. The guard reads the **title and the commits**, which is what a squash merge actually builds the commit from, and **not the PR description**, which never reaches `main` on its own and which a release PR fills with a changelog restating every break it ships. A unit test pins that.

Plus 14 unit tests, colocated and run by the workflow, following `guard-pull-request-target.ts` and its test file.

## Why not configure it away

**`exclude-paths`** cannot express this. release-please matches it by directory prefix with no globs, so "tests" and "internal CLI code" are not sayable. And it would not have caught #6600 anyway: the file that dragged the SDK to a major was `src/cli/commands/evaluators/catalog.ts`, ordinary SDK source. Excluding `sdks/typescript/src/cli` wholesale would then swallow genuine CLI releases.

**A native component-scoped breaking change does not exist.** `commit.ts` derives `commit.breaking` purely from the message text. There is no path or component dimension anywhere in that derivation.

**Per-package `versioning: always-bump-minor`** would cap majors permanently, including deliberate ones, so a genuine SDK break would ship silently mis-versioned. Worse than the problem.

That leaves prevention at PR time, which is this.

## Docs

`dev/docs/RELEASES.md` is new. It documents the component map, this rule, and the shim plus `Release-As` procedure for pinning a version release-please got wrong, including the trap that sank an earlier attempt: the footer must survive the squash, so keeping the commit message body matters. Until now that procedure was only recoverable by reading the history of #3627 and #6704. `CONTRIBUTING.md` gains a Breaking Changes subsection under the existing commit conventions and links to it.

## Cost

One job per PR, checkout plus node, no build. It has to run on every PR because any PR can carry a marker.

### DIFF
--- a/.github/scripts/guard-breaking-change-scope.test.ts
+++ b/.github/scripts/guard-breaking-change-scope.test.ts
@@ -65,6 +65,28 @@ describe("breaking-change scope guard", () => {
       );
       assert.equal(carriesBreakingChange(["feat(api): add a v2 endpoint"]), false);
     });
+
+    it("ignores a release-please changelog that reports past breaks", () => {
+      // A release PR restates every breaking change it ships. Reading it as a
+      // marker would fail release-please's own pull requests, which touch the
+      // manifest, the changelogs and the charts at once.
+      assert.equal(
+        carriesBreakingChange([
+          [
+            "chore(main): release langwatch 4.0.0",
+            "",
+            "### ⚠ BREAKING CHANGES",
+            "",
+            "* **evaluators:** evaluations referencing a legacy/ragas_* type stop working.",
+            "",
+            "### Features",
+            "",
+            "* **model-providers:** test a credential you have already saved",
+          ].join("\n"),
+        ]),
+        false,
+      );
+    });
   });
 
   describe("when mapping changed files onto release components", () => {
--- a/.github/scripts/guard-breaking-change-scope.test.ts
+++ b/.github/scripts/guard-breaking-change-scope.test.ts
@@ -1,0 +1,163 @@
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+import { describe, it } from "node:test";
+
+import {
+  bumpedComponents,
+  carriesBreakingChange,
+  releaseComponents,
+} from "./guard-breaking-change-scope.ts";
+
+const repoRoot = resolve(import.meta.dirname, "../..");
+
+const liveComponents = () =>
+  releaseComponents(
+    JSON.parse(
+      readFileSync(
+        resolve(repoRoot, ".github/release-please-config.json"),
+        "utf8",
+      ),
+    ),
+  );
+
+const names = (files: string[]): string[] =>
+  bumpedComponents(files, liveComponents())
+    .map((component) => component.name)
+    .sort();
+
+describe("breaking-change scope guard", () => {
+  describe("when reading the markers release-please reads", () => {
+    it("finds the `!` marker in a header, scoped or not", () => {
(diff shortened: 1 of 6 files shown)


rogeriochaves:
@coderabbitai review

PR #6752, opened 2026-08-09
feat(evaluators): add BDD spec for judge quality analysis (#3945)

### LINKED ISSUE #3945: LLM Judge Quality Analysis: Measure the Reliability of Your LLM Evaluators
**Problem**
LangWatch has a powerful LLM-as-judge evaluation system, but there is currently no way 
to measure whether those judges are actually reliable. Every evaluation score in the 
platform rests on an untested assumption: that the judge is correct.

In practice, LLM judges fail in predictable, invisible ways:
- They pass responses that are plausible but factually wrong (false positives)
- They fail responses that are correct but unconventionally phrased (false negatives)
- Their accuracy degrades silently when the underlying model or prompt changes
- A judge tuned for one domain transfers poorly to another

Without a measurement layer, none of these failure modes are visible. Teams are forced 
to either trust the judge blindly and risk making prompt/model decisions based on a 
miscalibrated scorer or manually audit results indefinitely, which defeats the purpose 
of automated evaluation entirely.

This is especially costly when evaluators are used as CI/CD deployment gates. A judge 
with 0.61 F1 score should not be blocking deploys, but today there is no way to know 
that number exists.

**Proposed solution**
Add a **Judge Quality** tab to the Evaluator detail page that uses LangWatch's existing 
annotation data as ground truth to compute standard classifier metrics for any LLM judge.

No new data collection is needed. LangWatch already has both ingredients:
- **LLM judge scores** — stored in ClickHouse `evaluation_runs` per trace
- **Human annotation scores** — already captured via the annotation system on the same traces

The missing piece is purely analytical: join judge scores against human annotation scores 
on matching `traceId + projectId + evaluatorId` and surface the following:

**Confusion Matrix** — a 2×2 breakdown of judge predictions vs. human ground truth 
(TP, TN, FP, FN), with each cell clickable to inspect the actual traces behind the number.

**Reliability Metrics** computed from the matrix:
- Precision — of traces the judge passed, how many actually passed per human review?
- Recall — of traces that actually passed, how many did the judge correctly identify?
- F1 Score — harmonic mean of precision and recall; the single best summary metric
- Accuracy — overall percentage of correct judge predictions
- Human-AI Agreement Rate — % of traces where judge and human reached the same verdict

**Performance Over Time** — a timeseries chart showing how each metric trends across 
evaluator versions, so teams can immediately see whether a prompt change helped or hurt 
judge reliability.

**Coverage Indicator** — annotated trace count vs. total evaluated traces, so teams 
know how statistically meaningful the analysis is before acting on it.

**Alternatives considered**
- **External validation scripts** — teams can manually export scores and annotations and 
  compute these metrics themselves, but this requires engineering effort per evaluator, 
  doesn't update in real time, and puts the burden on every team individually rather than 
  solving it once at the platform level.

- **Relying on annotation thumbs up/down as a proxy** — the existing annotation UX 
  captures informal feedback but does not structure it as a systematic ground truth 
  comparison against a specific evaluator. It cannot produce confusion matrices or 
  reliability metrics without this feature.

- **Doing nothing and trusting the judge** — acceptable for low-stakes use cases, but 
  untenable for teams using evaluators as CI/CD gates, compliance checks, or production 
  quality monitors. Silent judge miscalibration is a real and costly failure mode.

**Additional context**
This feature elevates two systems that already exist in LangWatch simultaneously:

- **Evaluations** become scientifically grounded rather than assumed-correct, teams 
  gain a quantitative answer to "can I trust this evaluator?" before relying on it.
- **Annotations** gain a high-value purpose beyond labeling, they become the ground 
  truth engine that validates the entire evaluation layer, giving expert reviewers a 
  reason to annotate beyond individual trace feedback.

Every team that adopts LLM-as-judge evaluation eventually hits the same wall: 
"We're using a judge to measure quality, but what's measuring the judge?" 
Right now there is no answer — not in LangWatch, not in any major LLMOps platform. 
LangWatch already has the data to be the first to answer it. That's a rare position 
to be in, and the cost of not acting on it is leaving a foundational trust gap at 
the center of the evaluation workflow.

gopalyuvaraj-labs:
**NO CODE CHANGE** — BDD spec only for feature #3945

## Why

LangWatch's LLM-as-judge evaluators have no way to measure their own reliability.
Every evaluation score assumes the judge is correct — when it isn't, teams make
prompt and model decisions based on bad signal.

This spec defines acceptance criteria for Judge Quality Analysis: using existing
annotation data as ground truth to compute Precision, Recall, F1, and Accuracy
for any LLM judge. No new data collection needed — the data is already there.

Closes #3945

## What changed

Added `specs/evaluators/judge-quality-analysis.feature` covering:

- **Confusion matrix** — TP/TN/FP/FN comparing judge scores vs. human annotations, with drill-down to individual traces
- **Reliability metrics** — Precision, Recall, F1, Accuracy, Human-AI Agreement Rate
- **Coverage indicator** — annotated vs. total evaluated trace count
- **Performance trend chart** — metrics over time across evaluation runs
- **Minimum threshold** — tab hidden until ≥10 annotated traces exist
- **Project scoping** — analysis isolated to current project

## How it works

Judge scores live in ClickHouse `evaluation_runs`. Human scores live in the
annotation system. Both are keyed on `traceId + projectId`. The confusion matrix
is a join on matching traces — making this a read-only analytics layer with zero
schema or pipeline changes.

## Test plan

- Follows `.feature` conventions in `specs/evaluators/`
- Action-based scenario descriptions, no "should"
- Edge cases covered: sub-threshold annotations, single run, real-time metric updates

## Anything surprising?

Annotation thumbs-up/down maps to binary pass/fail ground truth. Kept simple
intentionally — numeric annotation scores can be a follow-up.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive feature specifications for Judge Quality Analysis.
  * Covered access gating based on annotation counts and project-level scoping.
  * Added validation for confusion matrices, reliability metrics, annotation coverage, and performance trends.
  * Verified judge-pass and human-thumbs-down classifications, including pass/fail outcomes across supported analysis views.
  * Confirmed behavior across eligible and ineligible project configurations.

### DIFF
--- a/specs/evaluators/judge-quality-analysis.feature
+++ b/specs/evaluators/judge-quality-analysis.feature
@@ -1,0 +1,105 @@
+@unit
+Feature: Judge Quality Analysis
+  As a user of LangWatch who uses LLM-as-judge evaluators
+  I want to measure the reliability of my evaluators against human annotations
+  So that I can know whether to trust my evaluation scores
+
+  Background:
+    Given I have a project with evaluator "Answer Correctness"
+    And at least 10 traces have both a judge score and a human annotation score
+    for evaluator "Answer Correctness"
+
+  # ============================================================================
+  # Access
+  # ============================================================================
+
+  Scenario: Judge Quality tab appears on evaluator detail page
+    Given evaluator "Answer Correctness" has 10 or more annotated traces
+    When I open evaluator "Answer Correctness"
+    Then I see a "Judge Quality" tab alongside the configuration tab
+
+  Scenario: Judge Quality tab is hidden when insufficient annotations exist
+    Given evaluator "Answer Correctness" has fewer than 10 annotated traces
+    When I open evaluator "Answer Correctness"
+    Then I do not see a "Judge Quality" tab
+    And I see a prompt: "Annotate at least 10 traces to unlock judge quality analysis"
+
+  # ============================================================================
+  # Confusion Matrix
+  # ============================================================================
+
+  Scenario: Confusion matrix is displayed
+    When I open the "Judge Quality" tab for evaluator "Answer Correctness"
+    Then I see a 2x2 confusion matrix with cells:
+      | cell           | description                                |
+      | True Positive  | Judge passed, human annotated pass         |
+      | False Positive | Judge passed, human annotated fail         |
+      | False Negative | Judge failed, human annotated pass         |
+      | True Negative  | Judge failed, human annotated fail         |
+    And each cell shows a count of matching traces
+
+  Scenario: Clicking a confusion matrix cell shows matching traces
+    Given the confusion matrix shows 4 False Positives
+    When I click the "False Positive" cell
+    Then I see a list of the 4 traces where the judge passed but human annotated fail
+    And each trace links to its trace detail view
+
+  # ============================================================================
+  # Reliability Metrics
+  # ============================================================================
+
+  Scenario: Reliability metrics are computed from the confusion matrix
+    When I open the "Judge Quality" tab
+    Then I see the following metrics:
+      | metric         | description                                              |
+      | Precision      | TP / (TP + FP) — of judge passes, how many were correct  |
+      | Recall         | TP / (TP + FN) — of actual passes, how many did judge catch |
+      | F1 Score       | Harmonic mean of Precision and Recall                    |
+      | Accuracy       | (TP + TN) / total — overall correct predictions          |
+      | Agreement Rate | % of traces where judge and human reached the same verdict |
+
+  Scenario: Metrics update when new annotations are added
+    Given the Judge Quality tab shows Precision of 0.80
+    When a new annotation is added that contradicts a judge pass
+    Then the Precision value updates to reflect the new annotation
+
+  # ============================================================================
+  # Coverage Indicator
+  # ============================================================================
+
+  Scenario: Coverage indicator shows annotation completeness
+    Given evaluator "Answer Correctness" has scored 200 traces
+    And 25 of those traces have human annotations
+    When I open the "Judge Quality" tab
+    Then I see "25 / 200 traces annotated (12.5% coverage)"
+
+  # ============================================================================
+  # Performance Over Time
+  # ============================================================================
+
+  Scenario: Metrics trend chart is shown when multiple evaluation runs exist
+    Given evaluator "Answer Correctness" has run across 3 different date ranges
+    When I open the "Judge Quality" tab
+    Then I see a timeseries chart showing Precision, Recall, and F1 over time
+    And I can identify whether a recent prompt change improved or degraded reliability
+
+  Scenario: No trend chart when only one evaluation run exists
+    Given evaluator "Answer Correctness" has only one evaluation run
+    When I open the "Judge Quality" tab
+    Then I do not see a timeseries chart
+    And I see a message: "Run more evaluations over time to see performance trends"
+
+  # ============================================================================
+  # Data scoping
+  # ============================================================================
+
+  Scenario: Analysis is scoped to the current project
+    Given I am in project "Project A"
+    And project "Project B" has annotations for the same evaluator slug
+    When I open the "Judge Quality" tab
+    Then I only see annotations from "Project A"
+
+  Scenario: Analysis compares judge score pass/fail against annotation thumbs up/down
+    Given a trace where the judge scored 0.8 with a pass threshold of 0.5
+    And the same trace has a human annotation of thumbs down
+    Then this trace is counted as a False Positive in the confusion matrix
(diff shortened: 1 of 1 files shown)


langwatch-agent (commented):
Reviewed external fork head 762f8e0 by static diff and CI metadata only; no blocking findings. This is a specification-only change with no runtime, dependency, CI, or secret-handling impact.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=762f8e0473cb170e35d5a56ae514295954d878cd p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **Same shape as `#6762`: merging this closes `#3945` and ships no code.**
>
> `Closes #3945` retires the issue when this merges, leaving a 108-line spec for Judge Quality Analysis that nothing implements. Anyone later asking why judges have no reliability metrics finds a closed issue and a feature file. Change it to `Refs`, or open and link the implementation issue first.
>
> One thing in the spec is worth a decision before it becomes the contract: the tab is hidden until at least 10 annotated traces exist. That threshold decides whether the feature is discoverable, and 10 annotations is a low bar for a precision or recall number that people will then act on. A metric shown confidently over 10 samples is its own kind of bad signal, which is the exact problem the feature exists to solve.

| | |
| --- | --- |
| **Mode** | `Fast Skim`. One new file, no code. Read the threshold and the issue link. |
| **Issue** | `Closes #3945`. See above. |
| **State** | +108 / -0, one file. CI green; `BLOCKED` only for want of the required approval. Requested reviewer `sergioestebance`. Open since 9 August. Outside contributor. |
| **Evidence** | The premise checks out and is the best thing about the proposal: annotation data already exists, so ground truth needs no new collection. Confusion matrix with drill-down to individual traces, precision, recall, F1, accuracy, human-agreement rate, a coverage indicator, a trend across runs, and project scoping. |
| **Where to look** | `Closes` versus `Refs` · the 10-annotation threshold · whether an unbound feature file passes the parity gate |

<details>
<summary><b>What you are looking at, if <code>specs/</code> is not your directory</b></summary>

`specs/**` holds Gherkin feature files across 76 areas. They are not documentation. `platform/app/scripts/check-feature-parity.ts` runs in CI and requires that every scenario tagged `@unit`, `@integration`, `@e2e` or `@regression` be bound to a real test by a `/** @scenario <title> */` JSDoc comment sitting immediately above the `it(...)` that executes it. Tags inherit downward, so a single `@unit` on the `Feature:` line enforces every scenario in the file.

The checker enforces everything by default and fails closed. Two ratcheted deny-lists carry the existing migration debt and are only ever allowed to shrink:

- `LEGACY_UNBOUND`, files with enforced-but-unbound scenarios. Currently 15 files, 223 scenarios.
- `LEGACY_INERT`, files that enforce nothing at all because nothing in them is tagged. Currently 365 files, 4,804 scenarios. This list exists because an untagged file used to report `0/0 scenarios bound · ✓ all bound` and pass, which is an assertion of coverage that does not exist.

So a new feature file has three legitimate landing states, and only three: bind its scenarios to tests, tag them `@unimplemented` to declare the gap out loud, or leave them untagged and register the file in `LEGACY_INERT` with a stated reason. Spec-first PRs that describe behaviour nobody has written yet are normal and welcome here. What a spec cannot do is carry an enforced tag it has not earned, because that is the exact failure mode both deny-lists were built to stop.

</details>

<details>
<summary><b>Why this is worth having at all</b></summary>

Every evaluation score in the product assumes the judge is right. When it is not, teams tune prompts and swap models against bad signal, and nothing tells them. Measuring a judge against human annotations is the obvious check and the data is already sitting there.

So this is a good feature to specify. The review question is whether merging the spec moves it forward or files it away.

</details>

> [!NOTE]
> **A reliability metric computed over too few samples is the same failure as an unreliable judge, one level up.** Pick the threshold deliberately.

langwatch-agent:
## Parity gate: measured, and it fails

The brief left one question open, "whether an unbound feature file passes the parity gate". It does not. Measured today against `main` at `730f161e73`:

```
main alone
  OK: 10715 enforced scenario(s) bound across 1220 file(s).

main + this PR + #6762
  FAIL: 20 unbound scenario(s) in enforced files.
    specs/evaluators/judge-quality-analysis.feature   11 unbound
    specs/auth/disable-registration.feature            9 unbound
```

Those 20 are the entire failure. Nothing else on `main` contributes to it.

The cause is one line: `@unit` sits on the `Feature:` line, tags inherit downward, so all 11 scenarios are enforced and none is bound to a test. The green checks on this PR are from 9 August and predate the enforce-all polarity, so they are not evidence about today.

Three ways out, any of which is fine:

1. Drop the `@unit` tag and add the file to `LEGACY_INERT` with a reason. Honest: the spec is real, the coverage is not yet.
2. Keep the tag and mark the scenarios `@unimplemented`, which declares the gap explicitly rather than by omission.
3. Bind them, which means the implementation lands in this PR.

Option 1 or 2 keeps this a spec-first PR, which is what it is meant to be.

Relabelled `blocked-with-author` and `ci-green` removed. The reviewer stays assigned, they do not need re-requesting when this clears.

> [!NOTE]

PR #6760, opened 2026-08-09
fix(cli): drop billing and cadence wording from wrapper prompts

rogeriochaves:
Two lines of wrapper copy, no behavior change.

### The API-key option said LangWatch bills per token

```
? How should `langwatch claude` run?
    Using a Claude subscription - keep your own plan, send only telemetry to LangWatch
❯   Using an API key - route calls through LangWatch with a virtual key, billed per token
```

"billed per token" sits right after "through LangWatch", so it reads as if LangWatch charges per token for the privilege of proxying. It does not. The virtual key routes to the org's own provider keys, and the provider bills the tokens. The clause is dropped rather than reworded: the option is already titled "Using an API key", which is the part the user is choosing between.

### The plugin update check announced its own schedule

```
✦ langwatch checking whether the LangWatch plugin for Claude Code is up to date (once a day).
```

The once-a-day stamp is how the check avoids becoming a tax on startup. It is not something the user does anything with, and naming it invites the question this line was meant to avoid. The message now says only what it is doing while it waits.

### Spec

`copilot-path-defaults.feature` asserted the old billing wording in a step no test covered, so the sentence was free to drift from the code. It now names the copy that ships, and the copilot prompt test asserts the gateway option's description alongside the pre-selection it already checked.

### Verification

`pnpm test:unit` on the wrapper path-choice suite (81 passing) and `pnpm typecheck`, both clean in `sdks/typescript`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated gateway selection prompts with clearer, neutral descriptions.
  * Removed token-billing wording from gateway guidance.
  * Clarified that Copilot requests are routed through LangWatch using a virtual key.
  * Simplified the plugin update status message by removing the daily-update qualifier.

### DIFF
--- a/sdks/typescript/src/cli/utils/governance/__tests__/wrapper-path-choice.unit.test.ts
+++ b/sdks/typescript/src/cli/utils/governance/__tests__/wrapper-path-choice.unit.test.ts
@@ -219,7 +219,7 @@ describe("resolveWrapperPath", () => {
           "keep your own plan, send only telemetry to LangWatch",
         );
         expect(promptArg.choices[1]!.description).toBe(
-          "route calls through LangWatch with a virtual key, billed per token",
+          "route calls through LangWatch with a virtual key",
         );
         // Remembered for next time.
         expect(save).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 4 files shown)


drewdrewthis (commented):
Reviewed. No blocking findings — see verdict comment for details.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `ce407f7c` · **Run:** pr-review (reviewer: @drewdrewthis)

This is a two-line copy fix (dropping "billed per token" and "(once a day)" from wrapper prompt/status text), no behavior change. Verified no other user-facing string in the repo still asserts the old wording (searched `billed per token` and `once a day`; remaining hits are unrelated strings or internal code-comments about check cadence, not user-facing copy that needed updating). Tests updated to match the new copy; the copilot feature spec assertion now matches the shipped string. CI green (all gating jobs SUCCESS/SKIPPED, no failures).

### Non-blocking
- None.

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is **not** a merge-readiness signal; that is `pr-ready-check.sh`._

drewdrewthis (approved):
lgtm

PR #6762, opened 2026-08-09
feat(auth): add BDD spec for disabling self-registration via env var (#89)

### LINKED ISSUE #89: Disable account creation in settings or envs?
We are testing Langwatch for an internal project where our team members need to be able to create accounts, but we’re not interested in allowing someone who might stumble upon the frontend to create an account and a new organization.

I’ve been reviewing this, and I don’t see any option to disable account creation. Is that the case, or am I missing something?

We could set up basic auth in front of the frontend to protect it with a username/password, but it would be cleaner to disable account creation through an environment variable or application configuration.

If there’s no current option for this, I could try to contribute the functionality at some point when I have time.

Thanks,

gopalyuvaraj-labs:
**NO CODE CHANGE** — BDD spec only for issue #89

## Why

Self-hosted LangWatch instances have no way to prevent external users
from creating accounts if they discover the URL. Teams deploying
internally need this before they can safely expose LangWatch without
putting basic auth in front of the entire frontend.

This spec defines the acceptance criteria for a
`LANGWATCH_DISABLE_REGISTRATION=true` environment variable that blocks
new sign-ups without affecting existing users or SSO sign-in flows.

Closes #89

## What changed

Added `specs/auth/disable-registration.feature` covering:

- **Registration blocked** — sign-up form and API endpoint return 403
  when env var is set; no user or org created in the database
- **Sign-up page hidden** — anonymous users redirected to sign-in
- **Existing users unaffected** — credentials and SSO sign-in still work
- **Enabled by default** — no change in behaviour when env var is absent
  or set to `false`
- **UI feedback** — "Create account" link hidden when disabled, visible
  when enabled

## Test plan

- Follows `.feature` conventions in `specs/auth/`
- Action-based scenario descriptions, no "should"
- Covers positive and negative cases for the env var
- Existing user flows explicitly tested to confirm no regression

### DIFF
--- a/specs/auth/disable-registration.feature
+++ b/specs/auth/disable-registration.feature
@@ -1,0 +1,79 @@
+@unit
+Feature: Disable Self-Registration via Environment Variable
+  As a self-hosted LangWatch administrator
+  I want to prevent new users from creating accounts
+  So that only users I provision can access my instance
+
+  Background:
+    Given LangWatch is running in self-hosted mode
+
+  # ============================================================================
+  # Registration disabled
+  # ============================================================================
+
+  Scenario: Registration is blocked when LANGWATCH_DISABLE_REGISTRATION is set
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    When an anonymous user submits the sign-up form
+    Then the request is rejected with a 403 error
+    And the response message says "Account registration is disabled on this instance"
+
+  Scenario: Registration API endpoint is blocked when env var is set
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    When a POST request is made to the auth sign-up endpoint
+    Then the response status is 403
+    And no new user or organization is created in the database
+
+  Scenario: Sign-up page is hidden when registration is disabled
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    When an anonymous user navigates to the sign-up page
+    Then they are redirected to the sign-in page
+    And no sign-up link is shown in the UI
+
+  # ============================================================================
+  # Existing users unaffected
+  # ============================================================================
+
+  Scenario: Existing users can still sign in when registration is disabled
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    And a user with email "admin@acme.com" already exists
+    When that user signs in with valid credentials
+    Then the sign-in succeeds
+    And the user reaches the dashboard
+
+  Scenario: SSO-provisioned users can still sign in when registration is disabled
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    And an organization with ssoDomain "acme.com" exists
+    And a user with email "existing@acme.com" already exists
+    When that user signs in via the SSO provider
+    Then the sign-in succeeds
+
+  # ============================================================================
+  # Registration enabled by default
+  # ============================================================================
+
+  Scenario: Registration is allowed when LANGWATCH_DISABLE_REGISTRATION is not set
+    Given LANGWATCH_DISABLE_REGISTRATION is not set
+    When an anonymous user submits the sign-up form with valid details
+    Then a new user and organization are created
+    And the user is signed in
+
+  Scenario: Registration is allowed when LANGWATCH_DISABLE_REGISTRATION is "false"
+    Given LANGWATCH_DISABLE_REGISTRATION is "false"
+    When an anonymous user submits the sign-up form with valid details
+    Then a new user and organization are created
+    And the user is signed in
+
+  # ============================================================================
+  # UI feedback
+  # ============================================================================
+
+  Scenario: Sign-in page hides the "Create account" link when registration is disabled
+    Given LANGWATCH_DISABLE_REGISTRATION is "true"
+    When an anonymous user visits the sign-in page
+    Then the sign-in form is shown
+    And there is no "Create an account" or "Sign up" link visible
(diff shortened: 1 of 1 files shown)


langwatch-agent (commented):
Reviewed external fork head db7fa37 by static diff and CI metadata only; no blocking findings. This is a specification-only change with no runtime, dependency, CI, or secret-handling impact.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=db7fa37597369b1bf6a767db95b235ecf93efb16 p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **Merging this closes issue #89, and no code ships.**
>
> The PR body says `NO CODE CHANGE` and the description is honest about it. But `Closes #89` in the body means GitHub closes the issue the moment this merges. What is left behind is a `.feature` file describing a `LANGWATCH_DISABLE_REGISTRATION` environment variable that does not exist, and a closed issue that nobody will find again when they go looking for why self-hosted instances still let strangers sign up.
>
> That is a bookkeeping problem with a real consequence: a self-hoster reading the specs directory reasonably concludes the feature is there. Either change `Closes` to `Refs` so the issue stays open until the code lands, or open the implementation issue and link it here before this merges.
>
> Worth also checking what the feature parity gate does with a scenario nothing binds. If unbound scenarios read green, this file joins the set of specs that look like coverage and are not.

| | |
| --- | --- |
| **Mode** | `Fast Skim`. One new file, 79 lines, no code. The review is about the issue link and the parity gate, not the Gherkin. |
| **Issue** | `Closes #89`. See above. |
| **State** | +79 / -0, one file. CI green; `BLOCKED` only for want of the required approval. Requested reviewer `0xdeafcafe`. Open since 9 August. Outside contributor. |
| **Evidence** | The spec itself is well shaped: registration blocked at both form and API with a 403 and nothing written to the database, sign-up page redirected, existing credentials and SSO unaffected, enabled by default when the variable is absent or `false`, and the "Create account" link hidden. Those are the right cases. |
| **Where to look** | `Closes` versus `Refs` · whether an unbound feature file passes or fails the parity check |

<details>
<summary><b>What you are looking at, if <code>specs/</code> is not your directory</b></summary>

`specs/**` holds Gherkin feature files across 76 areas. They are not documentation. `platform/app/scripts/check-feature-parity.ts` runs in CI and requires that every scenario tagged `@unit`, `@integration`, `@e2e` or `@regression` be bound to a real test by a `/** @scenario <title> */` JSDoc comment sitting immediately above the `it(...)` that executes it. Tags inherit downward, so a single `@unit` on the `Feature:` line enforces every scenario in the file.

The checker enforces everything by default and fails closed. Two ratcheted deny-lists carry the existing migration debt and are only ever allowed to shrink:

- `LEGACY_UNBOUND`, files with enforced-but-unbound scenarios. Currently 15 files, 223 scenarios.
- `LEGACY_INERT`, files that enforce nothing at all because nothing in them is tagged. Currently 365 files, 4,804 scenarios. This list exists because an untagged file used to report `0/0 scenarios bound · ✓ all bound` and pass, which is an assertion of coverage that does not exist.

So a new feature file has three legitimate landing states, and only three: bind its scenarios to tests, tag them `@unimplemented` to declare the gap out loud, or leave them untagged and register the file in `LEGACY_INERT` with a stated reason. Spec-first PRs that describe behaviour nobody has written yet are normal and welcome here. What a spec cannot do is carry an enforced tag it has not earned, because that is the exact failure mode both deny-lists were built to stop.

</details>

<details>
<summary><b>Why this is worth having at all</b></summary>

A self-hosted LangWatch has no way to stop anyone who finds the URL from creating an account. The current workaround is putting basic auth in front of the whole frontend, which breaks SSO and is why teams have not deployed it internally.

So the underlying need is real and the spec is a reasonable first artifact. The question this review answers is only whether merging it advances that need or quietly retires it.

</details>

> [!NOTE]
> **A merged spec with no implementation and a closed issue is indistinguishable from a shipped feature until someone tries to use it.** Keep the issue open.

langwatch-agent:
## Parity gate: measured, and it fails

The brief left one question open, "whether an unbound feature file passes or fails the parity check". It fails. Measured today against `main` at `730f161e73`:

```
main alone
  OK: 10715 enforced scenario(s) bound across 1220 file(s).

main + this PR + #6752
  FAIL: 20 unbound scenario(s) in enforced files.
    specs/auth/disable-registration.feature            9 unbound
    specs/evaluators/judge-quality-analysis.feature   11 unbound
```

Those 20 are the entire failure. Nothing else on `main` contributes to it.

The cause is one line: `@unit` sits on the `Feature:` line, tags inherit downward, so all 9 scenarios are enforced and none is bound to a test. The green checks on this PR are from 9 August and predate the enforce-all polarity, so they are not evidence about today.

Three ways out, any of which is fine:

1. Drop the `@unit` tag and add the file to `LEGACY_INERT` with a reason. Honest: the spec is real, the coverage is not yet.
2. Keep the tag and mark the scenarios `@unimplemented`, which declares the gap explicitly rather than by omission.
3. Bind them, which means the implementation lands in this PR.

Option 1 or 2 keeps this a spec-first PR, which is what it is meant to be.

Relabelled `blocked-with-author` and `ci-green` removed. The reviewer stays assigned, they do not need re-requesting when this clears.

> [!NOTE]

Issue #2012, opened 2026-03-05
FEAT: Structured UNAUTHORIZED error infrastructure


sergioestebance:
## Problem
Permission denials in `rbac.ts` throw bare `TRPCError({ code: "UNAUTHORIZED" })` with no message or cause data. The global error handler in `api.ts` only watches for FORBIDDEN errors (license limits), so users see raw "UNAUTHORIZED" strings. There is no infrastructure to distinguish lite member restriction denials from other permission errors.

## Proposed Solution
1. Add descriptive messages to **all** UNAUTHORIZED throws in `rbac.ts`.
2. Create `extractLiteMemberRestrictionInfo()` helper in `trpcError.ts` (mirrors `extractLimitExceededInfo()` but checks for `UNAUTHORIZED + cause.reason === "lite_member_restricted"`).
3. Extend `MutationCache.onError` and add `QueryCache.onError` in `api.ts` to handle lite member restriction errors by opening the modal (`openLiteMemberRestriction`).
4. Add WeakSet dedup helpers (`markAsHandledByLiteMemberHandler`, `isHandledByLiteMemberHandler`) so individual error handlers skip duplicate toasts.
5. Add tests for the extractor, the dedup helpers, and the extended error handling.

## Acceptance Criteria
- [ ] All UNAUTHORIZED throws in `rbac.ts` include a human-readable message
- [ ] `extractLiteMemberRestrictionInfo()` identifies errors with `code: "UNAUTHORIZED"` and `cause.reason: "lite_member_restricted"`
- [ ] `extractLiteMemberRestrictionInfo()` returns `null` for vanilla UNAUTHORIZED errors
- [ ] `MutationCache.onError` opens the lite member restriction modal when triggered
- [ ] `QueryCache.onError` does the same as a fallback for queries (deep links, stale cache)
- [ ] WeakSet dedup helpers prevent duplicate toasts when the modal has been opened globally
- [ ] Existing FORBIDDEN/license limit handling is unchanged
- [ ] Integration tests confirm UNAUTHORIZED errors now show descriptive messages and open the modal

## Context
Planning doc: `EPIC/how-to/lite-members/epic_issues.md`
Parent epic: langwatch/langwatch-saas#315

PR #5863, opened 2026-07-17
fix(security): raise pillow, mcp, GitPython, pyasn1, nltk and litellm floors across the uv locks

langwatch-agent:
Clears **56 alerts** (42 HIGH, 11 MODERATE, 3 LOW) across the three Python manifests, mostly through transitive security floors in `[tool.uv]`. With one exception (litellm, below) nothing here is imported by shipped source: every package is reached through dev-only or examples-group tooling, so each floor moves only the package it names.

## What moves

| package | floor | alerts | reached via |
|---|---|---|---|
| `pillow` | `>=12.3.0` | **26** (20 HIGH, 6 MODERATE) | matplotlib, ragas, torchvision, streamlit, pikepdf, pdf2image |
| `mcp` | `>=1.28.1` | 3 HIGH | direct in mcp-server (python-sdk's copy is already patched on main) |
| `GitPython` | `>=3.1.54` | 8 HIGH | streamlit |
| `jupyterlab` | `>=4.6.2` | 5 (2 HIGH, 2 MODERATE, 1 LOW) | direct, examples group |
| `pyasn1` | `>=0.6.4` | 4 HIGH | pyasn1-modules, ldap3, rsa |
| `nltk` | `>=3.10.0` | 2 HIGH | rouge-score |
| `setuptools` | `>=83.0.0` | 1 MODERATE | spacy, thinc (langevals only, see below) |
| `pi-heif` | `>=1.3.0` | 1 MODERATE | unstructured |
| `litellm` | `>=1.84.0` | **6** (3 HIGH, 1 MODERATE, 2 LOW) | direct in langevals, see below |

### litellm >= 1.84.0 (the one that needed a cap lifted)

Six advisories on `langevals/uv.lock`: #1533, #1534 and #1746 (HIGH), #1744 (MODERATE), #1745 and #1747 (LOW). This is the only floor here that could not be raised on its own, and it is the last piece of the blocker tracked in #5020.

**Why a plain bump fails.** Every litellm release from 1.82.0 onward requires `openai>=2.8.0`. The legacy evaluator capped `langchain-openai>=0.1.23,<0.3.0`, and every langchain-openai in that range requires `openai<2.0.0`. So openai had to be both `>=2.8.0` and `<2.0.0`, and **no litellm version that fixes even the lowest-severity of these six advisories was satisfiable**. `uv lock` fails outright rather than resolving low.

[...]

Issue #2264, opened 2026-03-12
Pre-compile scenario child process to eliminate tsx cold-start delay


drewdrewthis:
## Description

The scenario child process (`scenario-child-process.ts`) is currently spawned via `pnpm exec tsx`, which compiles TypeScript at runtime on every invocation. This causes massive startup delays:

- **Cold start**: ~4.5 minutes (first run after worker restart — tsx compiles all imports)
- **Warm start**: ~53 seconds (tsx cache hit, but still slow due to pnpm resolution + module loading)

The 5-minute child process timeout (`CHILD_PROCESS.TIMEOUT_MS`) is almost entirely consumed by startup, leaving little time for actual scenario execution — causing scenarios to time out.

## Context

The child process imports several heavy dependencies:
- `@langwatch/scenario` (large SDK bundle)
- `@opentelemetry/api`
- Local modules: `model.factory`, `serialized-adapter.registry`, `remote-span-judge-agent`, etc.

Each spawn pays: pnpm binary resolution, corepack checks, tsx TypeScript compilation, module resolution for all imports.

The parent workers process (`src/workers.ts`) also runs via tsx but only pays this cost once at startup since it is long-lived. The child process pays it on every scenario execution.

## Scope

1. Add an `esbuild` (or `tsup`) build step that bundles `src/server/scenarios/execution/scenario-child-process.ts` and all its dependencies into a single `.js` file
2. Update `scenario.processor.ts` to spawn `node <bundled-file>.js` instead of `pnpm exec tsx <source-file>.ts`
3. Add the build step to the Docker build and dev workflow
4. Expected improvement: child process startup from minutes to low single-digit seconds

### Key files
- `langwatch/src/server/scenarios/scenario.processor.ts` (line 315, 333 — childPath + spawn)
- `langwatch/src/server/scenarios/execution/scenario-child-process.ts` (entry point)
- `langwatch/src/server/scenarios/scenario.constants.ts` (timeout config)

## Additional context

This was discovered while debugging scenario runs timing out via suites. The cascade of issues:
1. Slow child startup causes scenarios to time out
2. `SCENARIO_RUN_STARTED` event not emitted until child starts, so UI shows "Scenario run not found"
3. Failure event post gets malformed JSON 400, so UI stuck showing "running"

drewdrewthis:
## Implementation Plan

Starting implementation of pre-compiled scenario child process based on the approved feature file.

**Key design decisions:**
[...]

Issue #2021, opened 2026-03-05
FEAT: Backend RBAC restriction for lite members


sergioestebance:
## Problem
Lite members (LITE_MEMBER org role, renamed from EXTERNAL in #2019) inherit full VIEWER permissions — they can view everything a regular VIEWER can. We need to restrict them to a subset: only `analytics:view`, `messages:view` (list only), and `experiments:view`.

## Proposed Solution
1. **Add `messages` resource** to the RBAC permission system so `messages:view` can be distinguished from `traces:view`. Lite members should see the messages list but not individual trace details.
2. **Define `EXTERNAL_MEMBER_PERMISSIONS`** constant: `project:view`, `analytics:view`, `messages:view`, `experiments:view`
3. **Modify permission resolution** in `hasProjectPermission()`: if org role is LITE_MEMBER, use `EXTERNAL_MEMBER_PERMISSIONS` instead of `TEAM_ROLE_PERMISSIONS[VIEWER]`.
4. **Throw `LiteMemberRestrictedError`** on denial — connects to the global handler (#2012) which opens the modal (#2008).
5. **Verify end-to-end flow**: mutation attempt → RBAC denial → global handler → modal opens.

### Open design question
Adding `messages` as a new RBAC resource may affect existing routers. Alternative: keep `traces:view` for lite members but add route-level guards on `/messages/[trace]` in the frontend route guards issue. Trade-off: purity (granular permission) vs simplicity (route guard). Decide during implementation.

## Acceptance Criteria
- [ ] LITE_MEMBER org role users can only access permissions in `EXTERNAL_MEMBER_PERMISSIONS`
- [ ] Regular VIEWER team role users (non-LITE_MEMBER) retain full VIEWER permissions
- [ ] ADMIN/MEMBER org role users completely unaffected
- [ ] Permission denial for LITE_MEMBER users throws UNAUTHORIZED with `LiteMemberRestrictedError`
- [ ] Global handler opens lite member restriction modal on denial (end-to-end)
- [ ] Messages list is accessible but individual trace detail pages are not
- [ ] Decision on `messages:view` resource vs route-level guard is documented

## Context
Parent epic: langwatch/langwatch-saas#315
Depends on: #2008 (modal variant), #2011 (fetch org role), #2019 (rename EXTERNAL → LITE_MEMBER), #2012 (error infrastructure)
Branch: `feat/lite-members-restriction-logic`
[...]

PR #6777, opened 2026-08-10
fix(langevals): restore the ragas evaluators and close 17 alerts

langwatch-agent:
## What

Closes **17 of the 21** open alerts on `services/langevals` (1 CRITICAL, 4 HIGH), restores the ragas evaluators after two import-time breaks reached main, and closes the CI hole that let both through with a green tick.

| package | from | to | alerts |
|---|---|---|---|
| nltk | 3.10.2 | 3.10.3 | #2158 (CRITICAL), #2156 (HIGH), #2218 (HIGH), #2205, #2207, #2209, #2213, #2215, #2264 |
| langchain-core | 0.3.86 | 1.5.5 | #2012 (HIGH), #2010 |
| tornado | 6.5.7 | 6.5.8 | #2263 (HIGH), #2203, #2160 |
| langchain | 0.3.30 | 1.3.15 | #2021 |
| langchain-openai | 0.3.35 | 1.5.1 | #2017 |
| langchain-text-splitters | 0.3.11 | folded into langchain 1.x | #2016 |

## The langchain cap was ours, not ragas'

The family was pinned below 1.0 with a note saying ragas could not take core 1.x. That was not the blocker. The blocker was our own import of `get_openai_callback`, which pulls in `langchain_core.tracers.langchain_v1` — a module core 1.x removed. ragas imports fine on core 1.x, checked symbol by symbol.

Only the prompt and completion counters were ever used from that callback (cost is priced by litellm afterwards), so `lib/token_usage.py` builds the handler directly on the langchain-core callback surface and the cap comes off. `register_configure_hook` attaches it to every langchain call inside the block, exactly as the community version did, and a test drives that end to end against a fake chat model rather than asserting on the handler in isolation.

`langchain-community` stays on 0.3.x, capped below 0.4. ragas imports `langchain_community.chat_models.vertexai` at module level and 0.4.0 moved the vertexai integrations out, so 0.4.x makes every ragas import fail. No advisory currently affects that package.

## arize-phoenix is capped below 19.18.0, and that is a repair

The lock moves phoenix from main's 20.0.0 down to 19.17.0. That looks like a regression and is the opposite: **main is currently broken for this workspace**.

From 19.18.0, phoenix declares a dataclass with a `mappingproxy` default. Python 3.11 rejects it at import:

```
ValueError: mutable default <class 'mappingproxy'> for field boolean_names
            is not allowed: use default_factory
```

Phoenix's metadata still claims `>=3.10`, and this workspace is on 3.11 per `.python-version`, so the bad versions resolve happily and then take down collection for the whole suite through phoenix's pytest plugin. Checked against each published wheel rather than inferred:

| version | `import phoenix.trace.dsl.filter` on 3.11 |
|---|---|
| 19.17.0 | OK |
| 19.18.0 | ValueError |
| 20.0.0 | ValueError |
| 20.8.0 (current latest) | ValueError |

The cap lifts once phoenix fixes that or langevals moves to 3.12.

## The CI hole

`evaluators/ragas/` is excluded wholesale from the secret-bearing test step (`--ignore=evaluators/ragas/`), because the evaluator tests there are LLM-judged. A green `langevals-complete` therefore proved nothing about that directory, which is how two import-time breaks reached main.

The two new test files need no provider at all. They stub every call, pin the shape langchain-openai calls the litellm client with, and pin the token accounting that prices each evaluation. Both are named individually in the provider-free step so they fail loudly at import if that surface moves again.

## Four alerts stay open, and none of them has a fix to take

**nltk (#2211)**, **diskcache (#2011)** and **ragas (#2018 on the lock, #2009 on the ragas pyproject, the same advisory raised against both)** have no patched release published upstream in any line. There is nothing to bump to. They are called out in the pyproject rather than left to look like an oversight.

This count was 18-and-3 when this PR was first written. The miscount was mine: my closure check scored a manifest it cannot read (`evaluators/ragas/pyproject.toml` is not a lockfile, so it has no resolved versions to interval-test) as *closed* rather than as *unknown*, which quietly hid #2009. The check now reports that case separately, and the corrected split is what stands above.

## Deployment Impact

The langevals service image changes what it installs, and one module inside it is new. No configuration changes.

- **No new or changed environment variables**, helm values, secrets, ports or migrations. Nothing in `charts/` is touched.
- **The image resolves a different dependency set.** The langchain family moves from the 0.3 line to 1.x, `langchain-text-splitters` disappears (folded into `langchain` 1.x), and `langgraph` and its checkpoint/prebuilt/sdk packages arrive as new transitives of langchain 1.x. Image size moves accordingly; nothing is pinned to a version that needs a wheel this base image cannot build.
- **arize-phoenix resolves 19.17.0 rather than 20.0.0.** That is a repair, not a downgrade in capability: 19.18.0 and above cannot be imported at all on the Python this service runs, so the currently-deployed resolution is one where phoenix's import raises. Nothing in the served evaluator path imports phoenix; it is a notebooks dependency, and the failure surfaced through pytest collection.
- **Token accounting for ragas evaluators changes implementation, not output.** `get_openai_callback` is replaced by a handler on the same langchain-core callback surface, summing the same two counters. Cost is still priced by litellm downstream from the same numbers, so the cost figures the service reports are unchanged.
- **Rollback is a plain revert.** No state is written, no schema moves, and an older image resolves its own lock.

## Verified

- `uv sync --all-extras --all-packages` exit 0
- the provider-free CI step this PR defines: **225 passed**, 0 failed (it collected 98 items before the ragas files were named)
- every phoenix version in the table above imported directly against its published wheel on Python 3.11
- closure re-checked against each advisory's `vulnerableVersionRange` rather than its `firstPatchedVersion`: 17 of 21 closed, 4 with no fix published
- every resolved version this change moves audited against every published advisory for its package, not only the ones raised against this repo: the only hits are the two nltk advisories that have no patched release

https://claude.ai/code/session_017SsqT8ApEch8zUo7Aqc3y7

### DIFF
--- a/.github/workflows/langevals-ci.yml
+++ b/.github/workflows/langevals-ci.yml
@@ -146,8 +146,24 @@ jobs:
         # tests/` runs on the secret-bearing step by default, which forks skip,
         # so a contributor could otherwise break the request shape and see CI
         # stay green.
+        #
+        # The two ragas files are named individually for the same reason. The
+        # step above ignores `evaluators/ragas/` wholesale because the evaluator
+        # tests there are LLM-judged, and that is how two import-time breaks
+        # reached main without CI noticing: the litellm client losing the shape
+        # langchain-openai calls it with, and a notebooks dependency that could
+        # not be imported on this Python taking the whole suite's collection
+        # down. Neither needs a provider. These two files stub every call, so
+        # they pin the client shape and the token accounting that prices each
+        # evaluation, and they fail loudly at import if that surface moves again.
         working-directory: services/langevals
-        run: uv run pytest -x -v evaluators/langevals/tests/test_llm_answer_match_threading.py tests/deterministic/ langevals_core/tests/
+        run: |
+          uv run pytest -x -v \
+            evaluators/langevals/tests/test_llm_answer_match_threading.py \
+            evaluators/ragas/tests/test_model_to_langchain.py \
+            evaluators/ragas/tests/test_token_usage.py \
+            tests/deterministic/ \
+            langevals_core/tests/
 
   docker-build:
     needs: changes
--- a/services/langevals/evaluators/ragas/langevals_ragas/lib/token_usage.py
+++ b/services/langevals/evaluators/ragas/langevals_ragas/lib/token_usage.py
@@ -1,0 +1,95 @@
+"""Token accounting for a block of langchain LLM calls.
+
+langchain-community's `get_openai_callback` is the usual way to do this, but
+importing it drags in `langchain_core.tracers.langchain_v1`, which langchain-core
+1.x removed, so it cannot be used on the 1.x line. Only the prompt and completion
+counters are needed here (cost is priced by litellm afterwards), so the handler is
+built directly on the langchain-core callback surface instead.
+"""
+
+import threading
+from contextlib import contextmanager
+from contextvars import ContextVar
+from typing import Any, Generator, Optional
+
+from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.messages import AIMessage
+from langchain_core.outputs import ChatGeneration, LLMResult
+from langchain_core.tracers.context import register_configure_hook
+
+
+class TokenUsageCallbackHandler(BaseCallbackHandler):
+    """Sums prompt and completion tokens across every LLM call in a block."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._lock = threading.Lock()
+        self.prompt_tokens = 0
+        self.completion_tokens = 0
+        self.total_tokens = 0
+        self.successful_requests = 0
+
+    def __repr__(self) -> str:
+        return (
+            f"Prompt tokens: {self.prompt_tokens}\n"
+            f"Completion tokens: {self.completion_tokens}\n"
+            f"Total tokens: {self.total_tokens}\n"
+            f"Successful requests: {self.successful_requests}"
+        )
+
+    @property
+    def always_verbose(self) -> bool:
+        return True
+
+    def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
+        prompt_tokens, completion_tokens = self._extract_usage(response)
+
+        with self._lock:
+            self.successful_requests += 1
+            self.prompt_tokens += prompt_tokens
+            self.completion_tokens += completion_tokens
+            self.total_tokens += prompt_tokens + completion_tokens
+
+    @staticmethod
+    def _extract_usage(response: LLMResult) -> tuple[int, int]:
+        """Read usage off a chat message, falling back to the llm_output block.
+
+        Chat models report usage on the AIMessage; completion models and any
+        provider that skips `usage_metadata` report it under
+        `llm_output["token_usage"]`.
+        """
+        try:
+            generation = response.generations[0][0]
+        except IndexError:
+            generation = None
+
+        if isinstance(generation, ChatGeneration):
+            message = generation.message
+            if isinstance(message, AIMessage) and message.usage_metadata:
+                usage = message.usage_metadata
+                return usage.get("input_tokens", 0), usage.get("output_tokens", 0)
+
+        token_usage = (response.llm_output or {}).get("token_usage")
+        if not token_usage:
+            return 0, 0
+        return (
+            token_usage.get("prompt_tokens", 0),
+            token_usage.get("completion_tokens", 0),
+        )
+
+
+token_usage_callback_var: ContextVar[Optional[TokenUsageCallbackHandler]] = ContextVar(
+    "token_usage_callback", default=None
+)
+register_configure_hook(token_usage_callback_var, True)
+
+
+@contextmanager
+def get_token_usage_callback() -> Generator[TokenUsageCallbackHandler, None, None]:
+    """Attach a token counter to every langchain call made inside the block."""
+    handler = TokenUsageCallbackHandler()
+    token_usage_callback_var.set(handler)
+    try:
+        yield handler
+    finally:
+        token_usage_callback_var.set(None)
--- a/services/langevals/evaluators/ragas/langevals_ragas/lib/common.py
+++ b/services/langevals/evaluators/ragas/langevals_ragas/lib/common.py
@@ -12,7 +12,7 @@ from langevals_core.base_evaluator import (
 )
 from pydantic import Field
 from ragas.llms import LangchainLLMWrapper
-from langchain_community.callbacks import get_openai_callback
+from langevals_ragas.lib.token_usage import get_token_usage_callback
 
 from langevals_ragas.lib.model_to_langchain import (
     embeddings_model_to_langchain,
@@ -110,7 +110,7 @@ def check_max_tokens(
 
 @contextmanager
 def capture_cost(llm: LangchainLLMWrapper):
-    with get_openai_callback() as cb:
+    with get_token_usage_callback() as cb:
         money = Money(amount=0, currency="USD")
         yield money
 
--- a/services/langevals/evaluators/ragas/langevals_ragas/lib/model_to_langchain.py
+++ b/services/langevals/evaluators/ragas/langevals_ragas/lib/model_to_langchain.py
@@ -10,6 +10,39 @@ from langchain_core.language_models.chat_models import (
 import litellm
 
 
+class RawCompletion:
+    """The `.parse()` / `.headers` shape ChatOpenAI expects back from a raw call.
+
+    langchain-openai 0.3 routes the non-streaming path through
+    `client.with_raw_response.create(...)` and then calls `.parse()` on what
+    comes back. litellm returns the parsed response directly, so the two are
+    bridged here rather than by pinning langchain-openai below 0.3.
+    """
+
+    def __init__(self, response):
+        self._response = response
+        self.headers: dict = {}
+
+    def parse(self):
+        return self._response
+
+
+class RawCompletionProxy:
+    def __init__(self, client: "LitellmCompletion"):
+        self._client = client
+
+    def create(self, *args, **kwargs):
+        return RawCompletion(self._client.create(*args, **kwargs))
+
+
+class AsyncRawCompletionProxy:
+    def __init__(self, client: "AsyncLitellmCompletion"):
+        self._client = client
+
+    async def create(self, *args, **kwargs):
+        return RawCompletion(await self._client.create(*args, **kwargs))
+
+
 class LitellmCompletion:
     exception: Optional[Exception] = None
     temperature: float = 0
@@ -36,10 +69,18 @@ class LitellmCompletion:
             self.exception = e
             raise e
 
+    @property
+    def with_raw_response(self):
+        return RawCompletionProxy(self)
+
 
 class AsyncLitellmCompletion(LitellmCompletion):
     async def create(self, *args, **kwargs):
         return super().create(*args, **kwargs)
+
+    @property
+    def with_raw_response(self):
+        return AsyncRawCompletionProxy(self)
 
 
 def model_to_langchain(
--- a/services/langevals/evaluators/ragas/langevals_ragas/lib/model_to_langchain.py
+++ b/services/langevals/evaluators/ragas/langevals_ragas/lib/model_to_langchain.py
@@ -13,10 +13,10 @@ import litellm
 class RawCompletion:
     """The `.parse()` / `.headers` shape ChatOpenAI expects back from a raw call.
 
-    langchain-openai 0.3 routes the non-streaming path through
+    langchain-openai routes the non-streaming path through
     `client.with_raw_response.create(...)` and then calls `.parse()` on what
     comes back. litellm returns the parsed response directly, so the two are
-    bridged here rather than by pinning langchain-openai below 0.3.
+    bridged here.
     """
 
     def __init__(self, response):
--- a/services/langevals/evaluators/ragas/tests/test_model_to_langchain.py
+++ b/services/langevals/evaluators/ragas/tests/test_model_to_langchain.py
@@ -1,0 +1,63 @@
+"""The litellm client has to satisfy the shape langchain-openai calls it with.
+
+langchain-openai 0.3 routes the non-streaming path through
+`client.with_raw_response.create(...)` followed by `.parse()`, rather than
+calling `client.create(...)` directly as 0.2 did. These tests pin that shape so
+a future bump cannot quietly reintroduce the AttributeError.
+"""
+
+import asyncio
+from unittest.mock import patch
+
+from litellm.types.utils import Choices, Message, ModelResponse
+
+from langevals_ragas.lib.model_to_langchain import model_to_langchain
+
+
+def _canned_response(content: str = "pong") -> ModelResponse:
+    return ModelResponse(
+        id="chatcmpl-test",
+        created=0,
+        model="gpt-4o-mini",
+        object="chat.completion",
+        choices=[
+            Choices(
+                finish_reason="stop",
+                index=0,
+                message=Message(role="assistant", content=content),
+            )
+        ],
+    )
+
+
+def test_client_exposes_with_raw_response():
+    llm = model_to_langchain("gpt-4o-mini")
+    assert hasattr(llm.client, "with_raw_response")
+    assert hasattr(llm.async_client, "with_raw_response")
+
+
+def test_invoke_routes_through_with_raw_response():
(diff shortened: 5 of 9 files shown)


langwatch-agent:
CI green: all 13 required gates pass.

The two red `check` entries are stale `deployment-impact-check` runs from before I added the `## Deployment Impact` section; the run after that edit (31361788307) succeeded. Verified by run id rather than by the rollup, which keeps the superseded attempts visible.

Supersedes and closes #5863.

Ready for human review.

langwatch-agent:
Rebased and extended: **96 alerts** now (up from 91) via gitpython, h2 and setuptools floors, the last of which also carries torch to its patched release.

One thing worth a reviewer's eye, written up in full in the description: this PR now includes an **app-side fix**, not just lockfile churn. The langchain-openai bump broke every ragas evaluator at runtime (`AttributeError: 'AsyncLitellmCompletion' object has no attribute 'with_raw_response'`), and it was nearly invisible because those tests already fail on this box for want of credentials, so the pass/fail counts did not move. Only the error *type* did. The litellm shim in `model_to_langchain.py` now implements `with_raw_response` on both sync and async clients, verified by stubbing `litellm.completion` and driving both paths.

All 11 required gates green. Ready for human review.

langwatch-agent:
Added `torch>=2.13.0`, which closes alert #2025 and brings this PR to **87 alerts**.

I also spent this run trying to unblock the five langchain alerts on `services/langevals` (#2012 HIGH, #2021, #2016, #2010, #2017), since with everything else covered they are the largest remaining cluster. They are genuinely blocked, and the reason recorded in this repo was wrong, so the comment above those pins is now corrected.

Both routes to `langchain-core` 1.x resolve cleanly and then fail at **import**:

- `langchain-community` 0.3.31 with core 1.x dies on `langchain_core.tracers.langchain_v1`, which core 1.x removed.
- `langchain-community` 0.4.2 stops importing that, but drops `langchain_community.chat_models.vertexai`, which **ragas** imports unconditionally at `ragas/llms/base.py:12`. ragas 0.4.3 is its newest release, so there is nothing to upgrade into.

No partial move exists either: every 1.x family member requires `langchain-core >=1.2.31`.

Because both failures are at import rather than resolution, a green `uv lock` proves nothing here. `uv run pytest --collect-only` is what catches it. Full write-up on #3922.

The torch bump itself was verified the same way as the rest of this PR: identical failing-test name set (46 of 46) and identical error-type profile against a fresh `origin/main` baseline, with the wording delta being the openai 1.x to 2.x message rewrite already disclosed above.

langwatch-agent:
Status note, since this PR shows a red check that is not mine.

**All 11 required gates are SUCCESS.** The red one is `sdk-python-complete`, which is **not** in the required set for `main`.

It fails on `tests/test_examples.py::test_example[examples/dspy_bot.py]`:

```
TypeError: Legacy BaseLM calls accept at most one positional prompt.
Use dspy.context(experimental=True) or pass an LMRequest for typed multi-item LM calls.
```

That is a pre-existing break on main, now filed as #6913. Evidence rather than assertion:

- Run [31672238294](https://github.com/langwatch/langwatch/actions/runs/31672238294) on **`main`** fails with the identical error, as does the `release-please` branch.
- `dspy` is **3.3.0 on both main and this branch** — this PR does not move it, nor `litellm` (1.95.0) or `openai` (2.53.0).
- This PR's only change to `sdks/python` since the last green run is a rebase; today's edit was `torch>=2.13.0` in `services/langevals`, which cannot reach the Python SDK's tests.

The reason it looks like this PR's fault is worth spelling out: the `test` job is path-filtered to `sdks/python`, so it is skipped on most runs and the workflow still reports success. It only really executes when a change touches that directory, so any PR that does inherits main's breakage.

Nothing to fix here. #6913 tracks the example migration, which is a call about how the SDK should present dspy usage rather than a dependency decision.

langwatch-agent:
Rebased onto current main and added `pypdf>=6.15.0`, which closes #2085 and #2086 now that 6.15.0 is 10 days past the release-age gate that was blocking it.

**This PR now closes 67 alerts** (20 on `sdks/python`, 40 on `services/langevals`, 7 on `mcp/typescript`). The number is lower than the 87 quoted earlier purely because main has merged Dependabot bumps that independently closed the rest; the floors here have not weakened. `sdks/python/uv.lock` is now down to a single open alert, `diskcache`, which has no published fix in any version.

**On the red `ci-guards` check: it is not from this PR.** It is `tools/ciguard`'s `TestLeanCheckoutHoldsInTheLiveRepo`, complaining that the `shard-durations` job in `.github/workflows/langwatch-app-ci.yml` is missing three lean-checkout exclusions. This branch touches no `.github/` file at all — the whole diff is seven files under `sdks/python`, `services/langevals` and `mcp/typescript`. Reproduced on a clean checkout of `main` at 60fa445a3e, filed as #7100.

One caveat worth passing on, since it briefly fooled me: the Go test cache will serve a stale `ok` for that test. `go clean -testcache` first, or you will conclude main is fine when it is not.

Tests: `uv sync --all-extras` clean, `629 passed` on the Python SDK unit suite.

langwatch-agent:
Rebased onto current main (was 10 behind, two lockfile conflicts). Re-resolved all three uv locks from main's state; only `mcp/typescript/uv.lock` moved, and that was a metadata-only specifier sync with no version changes. Closure is unchanged at 67 alerts (20 `sdks/python` + 40 `services/langevals` + 7 `mcp`).

All 12 `*-complete` gates green at `7e56042`, no failures on any non-required job either. Ready for human review.

langwatch-agent:
Rebased onto current main. `uv lock --check` is clean, the four shim tests pass, and the CI slice is `65 passed, 1 failed` where that one failure is the pre-existing Presidio env-write assertion filed as #7307. Unchanged from the last review.

One thing to ignore on this PR: `smoke (ubuntu-latest / ubuntu-24.04-arm / macos-latest)` and the `alls-green` that aggregates them are red, and it is nothing to do with these changes. `dev/scripts/pack-npm.sh` verifies the tarball ships every tracked source file, but the staging step deliberately strips `.gitignore`, and the guard did not filter it. The first tracked `.gitignore` under a shipped tree, `dev/scripts/dogfood/multimodal/.gitignore`, landed on main this morning in #7339, after main's last smoke run, so every branch rebased past it goes red.

Fixed in #6569 rather than duplicated here, so the two do not collide on merge. Once either that or an equivalent lands on main, this PR's smoke goes green on the next run with no change to its own content. The twelve required `*-complete` gates are green here.

langwatch-agent:
Pushed the langchain 1.x migration onto this branch, so it now closes five alerts on top of the two live breaks it already fixed.

The short version: the `<1` cap on the langchain family that main carries was held in place by exactly one import, and it was on our side, not ragas'. `from langchain_community.callbacks import get_openai_callback` pulls in `langchain_core.tracers.langchain_v1`, which core 1.x removed. That symbol is used in one function, `capture_cost`, and only for two token counters, since cost is priced by litellm afterwards. Rebuilding the counter on the langchain-core callback surface removes the import and the cap goes with it. ragas itself imports fine on core 1.x, verified against every symbol the evaluators pull from it.

Closes #2012 (HIGH), #2016, #2021 (MODERATE), #2010, #2017 (LOW). Checked by interval-testing the resolved version against each `vulnerableVersionRange`, not against `firstPatchedVersion`. The new langgraph packages langchain 1.x brings in all resolve above their own advisory ranges, so nothing opens in trade.

On the test numbers, so they are not mistaken for a regression: the ragas suite is 13 failed / 13 passed / 2 skipped both before and after the langchain change, with an identical failing-test name set and error profile. Every failure in the whole langevals suite is `Incorrect API key provided` from a rejected `OPENAI_API_KEY` on the box (152 `AuthenticationError`, zero `AttributeError`/`TypeError`/`ImportError`/`AssertionError`). Same rejected key that is failing scenario's javascript job, tracked in langwatch/scenario#882. It needs a rotation I cannot do.

Worth a second pair of eyes on the token counter specifically: it is the piece that would fail silently rather than loudly, by pricing every evaluation at zero. There is an end-to-end test for it that drives two real `invoke()` calls through the configure hook and asserts the summed counts, not just the handler in isolation.

Ready for human review.

langwatch-agent:
One more commit, and it changes the diagnosis in this PR's description, so worth calling out.

I had put the "CI stayed green through both breaks" down to `langevals-complete` listing `test` under `allowed-skips`. That is true but it is not the real gap. Checking this PR's own CI run, the `test` job did run, and it passed, and it collected 53 and 98 items with not one of them from `evaluators/ragas/`. The secret-bearing step carries `--ignore=evaluators/ragas/`, so the entire ragas evaluator surface has no CI coverage at all. That is the surface both breaks were in.

It also meant the two regression tests I added here would never have run, which makes them decoration rather than a guard. So they are now named in the existing "evaluator regression tests that need no provider" step, alongside the other tests that are listed there individually for exactly this reason. Both stub every provider call and are deterministic, so they run on forks too.

Verified the fork path rather than assuming it: ran that step's command with `OPENAI_API_KEY`, `AZURE_API_KEY`, `ANTHROPIC_API_KEY` and `GEMINI_API_KEY` all unset, and the two ragas files give 11 passed.

The wider `--ignore` is still correct for the LLM-judged tests in that directory, so I have not touched it. Nothing here changes the runtime code from the previous push.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this moves the langchain family onto the 1.x line inside the ragas evaluators, and turns on CI that has never run against this code. Both halves can surprise you. The langchain major is a real migration that the description says hinges on a single import, which is either exactly right or the part that was not looked at hard enough. And enabling a test path that was ignored wholesale means the first honest signal about this code arrives on this PR, so a failure here may be pre-existing rather than caused.

| | |
|---|---|
| **Issue** | None linked. Fixes two breaks live on main right now |
| **Size** | +529/-104 across 9 files |
| **Mode** | Deep, and it is the CI change that earns it |
| **Blast radius** | The ragas evaluators, and the langevals CI gate |
| **Author** | langwatch-agent |

<details>
<summary><b>What you are looking at, if <code>services/langevals</code> is not your area</b></summary>

LangEvals is the evaluation backend behind LangWatch's evaluations. It is a Python monorepo of its own, published to PyPI, and it is structured as one installable extra per evaluator family: `evaluators/azure`, `evaluators/openai`, `evaluators/ragas`, `evaluators/lingua`, `evaluators/presidio`, each with its own `pyproject.toml` and its own dependency tree. Users install `langevals[ragas]` and get only that stack.

That layout is why a dependency change can break exactly one evaluator and nothing else, and why release-please carries a per-evaluator version bump list for this directory.

**Ragas is the RAG-quality family, and it is the one that sits furthest downstream.** It reaches LLMs through LangChain, so it inherits LangChain's model interfaces, its message types and its callback plumbing. A LangChain major bump is therefore not a bump for ragas, it is a migration, which is what the override comment on `main` was deferring.

The testing property that makes this area what it is: **most evaluator tests here are LLM-judged and non-deterministic.** They call real providers and assert on a model's opinion, so they cannot run on every commit and cannot run at all on a fork. The repository's answer has been to exclude the whole directory from the secret-bearing test step and to name the handful of provider-free, fully stubbed tests individually so they still run.

That is a reasonable trade and it has a sharp edge worth holding in mind while reading: a wholesale `--ignore` of a directory is indistinguishable, from the outside, from that directory being covered. The gate goes green either way, and nothing in the output says which files were never collected. Adding a deterministic test into an ignored directory does not add coverage; it has to be named into the step that actually runs.

</details>

<details>
<summary><b>The CI finding is the most valuable thing here</b></summary>

Two independent breaks landed on main from dependency merges, are live, and CI is green. There are two reasons and the second is the one that matters.

The `test` job in `langevals-ci.yml` is guarded by `needs.changes.outputs.heavy == 'true'`, and `langevals-complete` lists `test` under `allowed-skips`, so on a merge to main the job can skip while the gate still reports success. That alone is worth a conversation about how many other gates in this repo have a skippable job in their allowed-skips list.

But even when the job runs, it never touches this code, because the secret-bearing step carries `--ignore=evaluators/ragas/`. The directory is ignored wholesale since the evaluator tests in it are LLM-backed. So the answer to "why did CI not catch it" is "CI was never looking", and no amount of running the existing suite would have helped.
</details>

<details>
<summary><b>What to review, in order</b></summary>

1. **The `langevals-ci.yml` change.** This is the durable half. Whatever it does to close the gap will apply to every future change in this directory, so the shape matters more than this PR's contents. If it splits deterministic tests out from LLM-backed ones so the deterministic half can always run, that is the right answer and it should be visible in the diff.
2. **`model_to_langchain.py`.** The langchain 1.x migration is claimed to hinge on one import. Verify that claim rather than accept it; a major version bump that turns out to be a one-line change is either elegant or incomplete, and the tests next to it are how you tell which.
3. **`token_usage.py` and `common.py`.** These are where a langchain response shape change would surface. Token accounting is the kind of thing that keeps working while returning wrong numbers.
4. **The two breaks themselves.** They are live on main. Whatever the fix is, it is the most urgent part of this PR and the least interesting to review.
</details>

<details>
<summary><b>Production risk</b></summary>

Five security alerts including a HIGH clear with the langchain move, which is the reason the override comment on main asked for a deliberate migration PR. So there is a real cost to leaving this sitting, on top of the two live breaks.

Against that: ragas evaluators run against customer data, and a token-accounting or response-shape regression here shows up as wrong evaluation results rather than as an error. Wrong results are harder to notice than a crash, which is the argument for the deep read on point 3.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Two live breaks on main plus a HIGH advisory make this the most time-sensitive PR in the ready set.

.github/workflows/langevals-ci.yml
```diff
@@ -146,8 +146,24 @@ jobs:
...
+        # not be imported on this Python taking the whole suite's collection
+        # down. Neither needs a provider. These two files stub every call, so
+        # they pin the client shape and the token accounting that prices each
+        # evaluation, and they fail loudly at import if that surface moves again.
         working-directory: services/langevals
-        run: uv run pytest -x -v evaluators/langevals/tests/test_llm_answer_match_threading.py tests/deterministic/ langevals_core/tests/
+        run: |
+          uv run pytest -x -v \
+            evaluators/langevals/tests/test_llm_answer_match_threading.py \
+            evaluators/ragas/tests/test_model_to_langchain.py \
```
langwatch-agent, line 163: **P1 · Put the provider-free Ragas regression checks outside the skippable test job.**

- **What:** The new files run only in `test`, but `langevals-complete` still lists `test` in `allowed-skips`. A relevant draft sets `heavy=false`, so this job is skipped and the aggregate still succeeds. The secret-free checks are exactly the checks that could run in that state, so the first of the two stated CI blind spots remains.

- **Why:** A dependency or Ragas import regression can be pushed while the PR is draft and receive a green required gate without either new test collecting. Put these deterministic files in a separate relevant-only job, or make the aggregate reject a skipped test whenever `relevant=true`, while retaining an intentional skip for unrelated changes.

services/langevals/evaluators/ragas/langevals_ragas/lib/token_usage.py
```diff
@@ -0,0 +1,95 @@
...
+    "token_usage_callback", default=None
+)
+register_configure_hook(token_usage_callback_var, True)
+
+
+@contextmanager
+def get_token_usage_callback() -> Generator[TokenUsageCallbackHandler, None, None]:
+    """Attach a token counter to every langchain call made inside the block."""
+    handler = TokenUsageCallbackHandler()
+    token_usage_callback_var.set(handler)
```
langwatch-agent, line 91: **P2 · Restore the prior callback when nested capture scopes exit.**

- **What:** `ContextVar.set(handler)` returns a token, but the `finally` unconditionally sets the variable to `None`. An inner `get_token_usage_callback()` therefore erases an outer handler. I reproduced this with two nested scopes and one invocation in each: the inner recorded 2/1 tokens and the outer recorded 0/0, including for the invocation after the inner scope exited.

- **Why:** Nested evaluator or instrumentation scopes silently undercount the outer `capture_cost` block, so pricing becomes zero or too low rather than failing. Save the token from `.set()` and call `token_usage_callback_var.reset(token)` in `finally`; add a nested-scope regression test.

langwatch-agent (commented):
I found two issues in the migration and its intended CI regression coverage. I also ran the two new focused files against the exact locked environment: 11 passed. A mocked LangChain 1.x LiteLLM response preserved its token counts on the normal non-nested path.

LangWatch-Review: verdict=findings sha=2ca401234f41b2bd1c80a4392688dd4eaa0e24d2 p0=0 p1=1 p2=1 p3=0

PR #6786, opened 2026-08-10
feat(model-providers): offer the credential check in the drawer, where the credential is typed

### LINKED ISSUE #6784: model providers: the connection test lives in the list's overflow menu, not the drawer where the credential is typed
**Problem**

A customer asked for a way to confirm they had filled in a model provider's
credentials correctly — a ping, a hello world, anything that says "this works".
That shipped in #6651, but the only way to reach it is the three-dot menu on the
model providers list.

That is the wrong place relative to the intent. The moment someone wants the
answer is while they are still in the drawer they typed the credential into. As
it stands they have to save, close the drawer, find their row, open a menu, and
click — five steps after the moment of doubt, on a control most people will
never notice.

The non-obvious part, and the reason this is smaller than it looks: **the drawer
already probes the credential.** On Save, whenever a new API key has been typed,
`ModelProviderForm.tsx:396-408` sends it to the provider and blocks the save if
it comes back refused. It just never reports the good outcome — only the bad
one. Most of what this asks for is surfacing a verdict the code already computes
and then throws away.

**Proposed solution**

One "Test connection" button in the drawer footer, left of Save, with three
states in this order:

| State | When |
| --- | --- |
| hidden | the provider cannot be probed at all |
| disabled | a required credential field is still empty |
| enabled | every required field is filled |

Applies to creating a provider and to editing one.

Which probe runs is decided by what is on screen, using two mutations that
already exist:

- a newly-typed credential → `validateApiKey` (typed keys; works in create and
  in edit)
- an untouched saved row → `testConnection` (row id, stored credential)

`hasUserEnteredNewApiKey` in `utils/modelProviderHelpers` already draws that
line. Field completeness reuses `useRequiredCredentialKeys` +
`getEmptyRequiredCredentialKeys`, both already held by the form at
`ModelProviderForm.tsx:290`.

The visibility rule needs one predicate — `isProbeable(provider, customKeys)` —
in a module the browser can import, reconciled with the existing
`isLlmProvider` (`ModelProviderForm.tsx:287`) and `isOAuthDeviceProvider`
(`:304`) gates so we finish with three predicates rather than four. Today the
set lives at `server/modelProviders/providerValidation.ts:159`, in a module that
value-imports `ModelProviderRepository` (`:16`) and through it Prisma, so the
client cannot read it.

The same predicate also gates the list's existing overflow item
(`pages/settings/model-providers.tsx:428-445`), which currently shows for every
row and only reports "couldn't check" after the click. Both surfaces must give
one answer, or they become a fourth predicate that disagrees with the other
three.

The verdict renders inline under the button, worded from the code-keyed error
presentation registry, and is cleared whenever any credential field changes — a
green result must never outlive the credential it was about.
`ConnectionTestVerdict` moves out of `pages/settings/model-providers.tsx:363`
so both surfaces share it.

**The rule that governs all of it:** the control must not appear anywhere it
cannot tell the truth, and there is no success verdict without a completed round
trip. Six of the sixteen providers cannot be probed at all, and for those the
honest answer is that we did not look.

**Alternatives considered**

- *Showing the button disabled, with a tooltip explaining why the provider
  cannot be checked.* More explicit, but six of sixteen providers would carry a
  permanently dead control, which reads as broken.
- *Hiding the button but keeping a line of text in its place.* Preserves the
  "we cannot check this" sentence, at the cost of copy on every drawer for a
  provider where there is nothing to do. Decided against: hidden is hidden.
- *Probing automatically as the customer types, or on field blur.* Rejected —
  it would spend provider quota and the per-organization rate limit on every
  pause in typing.

**Additional context**

Out of scope, so nobody reopens them here:

- having the test also save the row
- any change to the save-time probe or the "Save anyway" behaviour
- making Azure, Bedrock or Vertex AI testable — that needs a real inference
  probe and is a separate, larger piece of work
- a stored verdict, a health badge, background re-checking, or a
  check-everything sweep
- the onboarding flow, which uses a different component

Known consequence, recorded deliberately: hiding the control on both surfaces
makes one unchecked reason — "this provider cannot be probed" — unreachable
from the interface. The three-state result stays load-bearing, because the other
unchecked reasons depend on the stored data rather than the provider type and
are still reachable, so nothing becomes dead code. But that one reason loses its
customer-facing path.

Counted: no schema change, no migration, no new API contract, no new outbound
request path, no change to the security boundary. Roughly 6-8 files across
`platform/app/src` and `specs/`.

---

## 🤖 For Agents

**Nothing to investigate — the cause is that the control was only ever built on
one surface.** Implementation plan:

1. Move the probeability set and the "is there anywhere to ask" check out of
   `server/modelProviders/providerValidation.ts` (`NOT_PROBEABLE` at `:159`,
   the `no_endpoint` decision at `:1193-1209`) into a framework-free module both
   sides import. The server module value-imports Prisma, so the client cannot
   read it today. `server/__tests__/frontend-boundary.unit.test.ts` guards
   server→browser only, so a wrong import direction here ships silently — place
   it deliberately.
2. Extract `ConnectionTestVerdict` from `pages/settings/model-providers.tsx:363`.
3. Add the button to `ModelProviderForm.tsx:577-589`, gated on
   `isProbeable(...)` for visibility and on the existing `requiredKeys` /
   `getEmptyRequiredCredentialKeys` answer for enablement.
4. Route the click: `hasUserEnteredNewApiKey(state.customKeys)` →
   `useModelProviderApiKeyValidation.validate()`, otherwise
   `useModelProviderConnectionTest.test(rowId)`. Both hooks exist and are
   unchanged.
5. Clear the verdict on any credential change, using the same fingerprint rule
   `useCredentialProbeGate` already applies to refusals.
6. Apply the same `isProbeable(...)` gate to the list item at
   `pages/settings/model-providers.tsx:428-445`.
7. Append scenarios to `specs/model-providers/credential-validation.feature` —
   the on-demand section is already there at `:410-542`, and every scenario in
   it is triggered "from the provider list". Each new scenario needs a binding
   tag **and** a matching `@scenario "<title>"` annotation on its test, or
   `check-feature-parity.ts` reports it vacuously bound.

**Acceptance criteria**

- In the drawer, a provider that cannot be probed shows no test control at all,
  on create and on edit.
- For a provider that can be probed, the control is present but disabled until
  every required credential field is filled, then enabled.
- Clicking it with a newly-typed credential checks what is on screen; clicking
  it on an untouched saved row checks the stored credential, without asking for
  the key again.
- A verdict disappears the moment any credential field changes.
- The list's overflow item is hidden for exactly the same providers the drawer
  hides it for.
- No success verdict is ever shown for a provider that was not actually
  contacted.

sergioestebance:
Closes #6784.

## For humans

You can already check that a model provider's credentials work, but only by
saving, closing the drawer, finding the row in the list, opening a three-dot
menu, and clicking. That is five steps away from the moment anyone actually
wants the answer, which is while they are still looking at the fields they just
filled in.

This puts the check next to Save, in the drawer. It appears only for providers
we can genuinely check, stays greyed out until the credential is complete, and
tells you what it found without saving anything.

Two things behind it turned out to be broken in ways worth knowing about:

- **A check could describe settings you were no longer looking at.** Change a
  provider's endpoint, ask for a check, and the answer would have been about the
  endpoint you just replaced — reported as a pass. Fixed, and the fix carefully
  does not open a way to send a stored credential somewhere the caller picked.
- **One of the two ways of checking had no rate limit at all.** It was only
  reachable by saving, so nobody could sit on it. A button that checks without
  saving removes that, so both ways now share one budget.

## What changed

**The control.** Three states in a deliberate order: absent when the provider
cannot be checked at all, present but unusable while the credential is
incomplete, usable after that. Six of sixteen providers credential in ways no
listing endpoint exercises (AWS signatures, gcloud application-default
credentials, deployment-scoped keys, a content-safety service with no models
route, a provider with nowhere to ask, and one that credentials through its own
sign-in). An action that can only ever answer "we did not look" reads as broken
rather than as inapplicable, so it is not offered.

"Complete" means the provider's own schema accepts the credential, not that no
required field is empty. Gemini takes a project and a location together or not
at all, so filling one leaves every *required* field satisfied while the
credential as a whole is one a save would reject.

The list's menu item is hidden on the same answer. Two surfaces answering the
same question differently is a customer seeing an action in one place and not
the other with nothing to explain it, and the disagreement is invisible from
either side.

**Where the rules live.** The settings UI has to answer "can this be checked?"
before offering anything, and it cannot ask the validator — that module reaches
the provider repository and through it Prisma, so importing it from the browser
would pull both into the bundle. Keeping a second copy of the list on the client
is the obvious way out and the wrong one: two lists disagree the first time
either is edited. So the rules move onto the registry, which both sides already
import, and the validator reads them from there.

The pin is behavioural rather than a second constant. For every provider in the
registry, the test builds a credential from that provider's own schema, runs the
real validator against a stubbed transport, and requires the verdict to agree
with the predicate. A new provider, or an endpoint added to one that had none,
fails there.

**Checking what is on screen.** `testConnection` now accepts the settings in the
request. Neither existing route could check "the endpoint on screen with the
credential already stored": the discriminator that would pick between them reads
the key field, which still holds the mask after a non-key edit, so it routed to
the stored row and probed the endpoint that was just replaced. Sending the form
instead does not help, because a stored secret reaches the browser as the masked
placeholder and never as a value — so that route reports the credential as
masked and checks nothing.

The rule this ships under is narrower than "no endpoint from the caller" and
does more work: **the two sources are never combined.** Either the settings come
from the row or they come from the request, whole. A stored secret therefore
cannot be paired with an address the caller chose, which is what would turn
permission to edit a provider into permission to extract its key. A masked
placeholder is not a credential and comes back unchecked rather than being
swapped for the real one — which is what a customer editing an endpoint without
re-entering their key should get, and the drawer says so in those words.

Both authz guards stay, in the order the delete path uses them, and supplied
settings travel through the same vetted transport as every other probe.

**Rate limiting.** The typed-credential route had a permission check and no
budget. It now shares the stored route's, keyed the same way, so a caller cannot
double their allowance by alternating between the two.

## Refusal overridden

`/implement` refused this on: **irreversible path — it touches a security
boundary.** `/route` counted the change before `/challenge` altered its shape;
widening `testConnection` to accept a caller-supplied endpoint is the input
shape behind a previous credential-exposure fix, so the count changes and the
gate fires.

Overridden by: the requester, choosing "let testConnection take the on-screen
credentials" over disabling the control in that state, with the trade-off named
at the time.

Carried over from the heavy path: `/challenge` — which is what found the two
defects above and killed three of my own objections.

The mitigation is in the code rather than in the decision: never combining the
two sources means there is no arrangement of inputs that reaches the dangerous
combination, which is a stronger property than refusing endpoints outright and
is the reason this is safe to ship.

## Checks

- `pnpm typecheck` and `pnpm typecheck:tests` — clean
- `pnpm lint` — 0 errors
- Server: 495/495 across 36 files
- Settings and drawer integration: 264/264 across 37 files
- `check-feature-parity` — 53/53 scenarios bound in
  `specs/model-providers/credential-validation.feature`

## Notes for review

- One existing scenario became false and was rewritten rather than left
  standing: a check *can* now be told where to look, so "a test never accepts an
  endpoint from the caller" is replaced by the property that actually protects
  the credential.
- The drawer's own test file was written after the component rather than before
  it. It is not vacuous — mutating the visibility rule to always-true fails the
  scenario that pins it — but it did not drive the design the way the others did.
- Seven existing drawer suites gain the two mutations the form now reaches for.
  None of them exercise it.
- Hiding the control on both surfaces makes one unchecked reason unreachable
  from the interface while the client predicate and the server's set agree. It
  becomes reachable exactly on drift, so it is a drift detector rather than dead
  copy, and the behavioural pin above is what keeps them honest.
- Review found and fixed a real defect in the first pass: the staleness guard in
  the new hook was inert, so a verdict about a credential the customer had since
  replaced landed and re-displayed itself. See the review comment for the
  executed output.
- An earlier revision of this description stated that a stored key reaches the
  form as an empty string. That was wrong — it reaches it as the masked
  placeholder. The conclusion is unchanged (a mask is not a credential either),
  but the mechanism above has been corrected.

## Summary by CodeRabbit

* **New Features**
  * Added “Test connection” controls to model provider settings.
  * Test new or edited credentials without saving changes.
  * Shows testing, success, error, and unavailable status messages.
  * Hides testing for unsupported providers.
  * Added accessible announcements for connection-test results.

* **Bug Fixes**
  * Prevented stale results after credentials or provider settings change.
  * Improved handling of masked, missing, invalid, and failed credentials.

* **Tests**
  * Expanded coverage for credential validation, provider eligibility, rate limits, and connection-test behavior.

### DIFF
--- a/platform/app/src/components/__tests__/EditModelProviderDrawer.edit-row-resolution.integration.test.tsx
+++ b/platform/app/src/components/__tests__/EditModelProviderDrawer.edit-row-resolution.integration.test.tsx
@@ -73,6 +73,10 @@ vi.mock("../../utils/api", () => ({
       isManagedProvider: {
         useQuery: () => ({ data: { managed: false } }),
       },
+      // The drawer offers a credential check, so the form reaches for both
+      // routes it can take. Neither is exercised here.
+      testConnection: { useMutation: () => ({ mutateAsync: vi.fn() }) },
+      validateApiKey: { useMutation: () => ({ mutateAsync: vi.fn() }) },
     },
     useContext: () => ({
       organization: {
--- a/platform/app/src/components/settings/ConnectionTestVerdict.tsx
+++ b/platform/app/src/components/settings/ConnectionTestVerdict.tsx
@@ -1,0 +1,60 @@
+import { Box, Text } from "@chakra-ui/react";
+import type { ConnectionTestState } from "../../hooks/connectionTestState";
+
+/**
+ * What the last credential check said.
+ *
+ * Three states, rendered three different ways on purpose. A check that could
+ * not run reads as neutral rather than green: it is not a pass, and dressing
+ * it as one would tell a customer their configuration is fine on the strength
+ * of never having asked.
+ *
+ * The whole verdict lives in a polite live region. The text arrives well after
+ * the click that asked for it, and a screen reader announces neither the
+ * "Testing…" transition nor the answer replacing it — so without this the
+ * control is a button that appears to do nothing at all.
+ *
+ * Shared by the provider list and the drawer. The wording of an unchecked
+ * verdict differs between them, but that is decided before it arrives here:
+ * this renders a message, it does not choose one.
+ */
+export function ConnectionTestVerdict({
+  state,
+}: {
+  state: ConnectionTestState | undefined;
+}) {
+  const verdict = () => {
+    if (!state) return null;
+
+    if (state.status === "testing") {
+      return (
+        <Text fontSize="xs" color="fg.muted">
+          Testing…
+        </Text>
+      );
+    }
+
+    if (state.status === "works") {
+      return (
+        <Text fontSize="xs" color="green.fg">
+          Connection works
+        </Text>
+      );
+    }
+
+    return (
+      <Text
+        fontSize="xs"
+        color={state.status === "refused" ? "red.fg" : "fg.muted"}
+      >
+        {state.message}
+      </Text>
+    );
+  };
+
+  return (
+    <Box aria-live="polite" aria-atomic="true">
+      {verdict()}
+    </Box>
+  );
+}
--- a/platform/app/src/components/settings/ModelProviderForm.tsx
+++ b/platform/app/src/components/settings/ModelProviderForm.tsx
@@ -23,6 +23,7 @@ import { useModelProvidersSettings } from "../../hooks/useModelProvidersSettings
 import { useOrganizationTeamProject } from "../../hooks/useOrganizationTeamProject";
 import { useRequiredCredentialKeys } from "../../hooks/useRequiredCredentialKeys";
 import {
+  isProviderProbeable,
   type MaybeStoredModelProvider,
   modelProviders as modelProvidersRegistry,
 } from "../../server/modelProviders/registry";
@@ -31,6 +32,8 @@ import {
   hasUserEnteredNewApiKey,
   hasUserModifiedNonApiKeyFields,
 } from "../../utils/modelProviderHelpers";
+import { useCredentialCheck } from "../../hooks/useCredentialCheck";
+import { ConnectionTestVerdict } from "./ConnectionTestVerdict";
 import { parseZodFieldErrors, type ZodErrorStructure } from "../../utils/zod";
 import { SmallLabel } from "../SmallLabel";
 import { Switch } from "../ui/switch";
@@ -329,6 +332,68 @@ export const EditModelProviderForm = ({
       resetKey: providerId,
     });
 
+  // Whether the credential on screen differs from the one in storage, which is
+  // what decides whether the check is about the form or about the row. The
+  // same two helpers the dirty check and the save path already use, so the
+  // three cannot end up with different opinions about what "edited" means.
+  const hasCredentialEdits =
+    hasUserEnteredNewApiKey(state.customKeys) ||
+    hasUserModifiedNonApiKeyFields(state.customKeys, state.initialKeys);
+
+  const credentialCheck = useCredentialCheck({
+    projectId,
+    organizationId,
+    modelProviderId: providerId,
+    provider: provider.provider,
+    customKeys: state.customKeys,
+    scopes: state.scopes,
+    hasEdits: hasCredentialEdits,
+  });
+
+  /**
+   * Whether to offer the check at all.
+   *
+   * A property of the provider rather than of what has been typed, so the
+   * control does not appear and vanish while the form is filled in. Six of the
+   * registered providers credential in ways no listing endpoint exercises;
+   * offering a check there would produce "we did not look" every time, which
+   * reads as a broken button rather than as an inapplicable one.
+   *
+   * This deliberately sits alongside `isLlmProvider` and `isOAuthDeviceProvider`
+   * rather than replacing them. Those two gate the probe that runs on save, and
+   * folding all three together would change when a save is allowed to proceed —
+   * a different blast radius than a button.
+   */
+  const canCheckCredential = isProviderProbeable({
+    provider: provider.provider,
+  });
+
+  /**
+   * Whether the credential is complete enough to be worth sending.
+   *
+   * Two questions, and the required-field one is not sufficient on its own. A
+   * provider can accept a pair of fields together or not at all — Gemini's
+   * project and location — so filling one leaves every *required* field
+   * satisfied while the credential as a whole is one the provider's own schema
+   * rejects. Asking the schema is the only answer that cannot drift from what
+   * a save would accept.
+   */
+  const credentialIsComplete = useMemo(() => {
+    if (
+      getEmptyRequiredCredentialKeys({
+        requiredKeys,
+        values: state.customKeys,
+      }).length > 0
+    ) {
+      return false;
+    }
+
+    if (!providerDefinition?.keysSchema) return true;
+
+    return providerDefinition.keysSchema.safeParse({ ...state.customKeys })
+      .success;
+  }, [providerDefinition, requiredKeys, state.customKeys]);
+
   const handleSave = useCallback(async () => {
     // Clear previous errors
     setFieldErrors({});
@@ -574,7 +639,33 @@ export const EditModelProviderForm = ({
           />
         )}
 
-        <HStack width="full" justify="end">
+        {/* The check sits beside Save rather than in a menu somewhere: the
+            moment a customer wants to know whether they filled this in
+            correctly is while they are still looking at the fields. It never
+            saves — see the note under the button. */}
+        <HStack width="full" justify="space-between" align="center">
+          <VStack align="start" gap={1}>
+            {canCheckCredential && !isOAuthDeviceProvider && (
+              <Button
+                size="sm"
+                variant="outline"
+                loading={credentialCheck.state?.status === "testing"}
+                disabled={cannotResolveTarget || !credentialIsComplete}
+                onClick={() => void credentialCheck.check()}
+              >
+                Test connection
+              </Button>
+            )}
+            <ConnectionTestVerdict state={credentialCheck.state} />
+            {/* Only next to a pass, and only before the row exists. "It
+                works" and "it is saved" are easy to read as the same
+                sentence, and they are furthest apart exactly here. */}
+            {credentialCheck.state?.status === "works" && !providerId && (
+              <Text fontSize="xs" color="fg.muted">
+                Save to finish adding this provider.
+              </Text>
+            )}
+          </VStack>
           <Button
             size="sm"
             colorPalette="orange"
--- a/platform/app/src/components/settings/ModelProviderForm.tsx
+++ b/platform/app/src/components/settings/ModelProviderForm.tsx
@@ -14,6 +14,7 @@ import {
   isResolvableProviderId,
   useAllModelProvidersList,
 } from "../../hooks/useAllModelProvidersList";
+import { useCredentialCheck } from "../../hooks/useCredentialCheck";
 import { useCredentialProbeGate } from "../../hooks/useCredentialProbeGate";
 import { useDrawer } from "../../hooks/useDrawer";
 import { useFeatureFlag } from "../../hooks/useFeatureFlag";
@@ -32,14 +33,13 @@ import {
   hasUserEnteredNewApiKey,
   hasUserModifiedNonApiKeyFields,
 } from "../../utils/modelProviderHelpers";
-import { useCredentialCheck } from "../../hooks/useCredentialCheck";
-import { ConnectionTestVerdict } from "./ConnectionTestVerdict";
 import { parseZodFieldErrors, type ZodErrorStructure } from "../../utils/zod";
 import { SmallLabel } from "../SmallLabel";
 import { Switch } from "../ui/switch";
 import { toaster } from "../ui/toaster";
 import { useCodexCodingDefaultsAskStore } from "./CodexCodingDefaultsAsk";
 import { CodexSignIn } from "./CodexSignIn";
+import { ConnectionTestVerdict } from "./ConnectionTestVerdict";
 import {
   draftFromProvider,
   EMPTY_ADVANCED_DRAFT,
--- a/platform/app/src/components/settings/__tests__/ModelProviderForm.advanced-gateway.integration.test.tsx
+++ b/platform/app/src/components/settings/__tests__/ModelProviderForm.advanced-gateway.integration.test.tsx
@@ -85,6 +85,10 @@ vi.mock("../../../utils/api", () => ({
       isManagedProvider: {
         useQuery: () => ({ data: { managed: false } }),
       },
+      // The drawer offers a credential check, so the form reaches for both
+      // routes it can take. Neither is exercised here.
+      testConnection: { useMutation: () => ({ mutateAsync: vi.fn() }) },
+      validateApiKey: { useMutation: () => ({ mutateAsync: vi.fn() }) },
       // EditModelProviderForm resolves its edit target from the flat
       // (uncollapsed) provider list via useAllModelProvidersList (#5380).
       // primeHooks seeds both with the same mp_existing fixture the
--- a/platform/app/src/components/settings/__tests__/ModelProviderForm.azure-safety.integration.test.tsx
+++ b/platform/app/src/components/settings/__tests__/ModelProviderForm.azure-safety.integration.test.tsx
@@ -82,6 +82,10 @@ vi.mock("../../../utils/api", () => ({
       isManagedProvider: {
         useQuery: () => ({ data: { managed: false } }),
       },
+      // The drawer offers a credential check, so the form reaches for both
+      // routes it can take. Neither is exercised here.
+      testConnection: { useMutation: () => ({ mutateAsync: vi.fn() }) },
+      validateApiKey: { useMutation: () => ({ mutateAsync: vi.fn() }) },
       // EditModelProviderForm resolves its edit target from the flat
       // (uncollapsed) provider list via useAllModelProvidersList (#5380).
       // primeHooksForProvider seeds both with the same fixture the
--- a/platform/app/src/components/settings/__tests__/ModelProviderForm.codex-signin.integration.test.tsx
+++ b/platform/app/src/components/settings/__tests__/ModelProviderForm.codex-signin.integration.test.tsx
@@ -104,6 +104,10 @@ vi.mock("../../../utils/api", () => ({
       isManagedProvider: {
         useQuery: () => ({ data: { managed: false } }),
       },
+      // The drawer offers a credential check, so the form reaches for both
+      // routes it can take. Neither is exercised here.
+      testConnection: { useMutation: () => ({ mutateAsync: vi.fn() }) },
+      validateApiKey: { useMutation: () => ({ mutateAsync: vi.fn() }) },
       listAllForOrganizationForFrontend: { useQuery: mockListAllForOrgQuery },
       listAllForProjectForFrontend: { useQuery: mockListAllForProjectQuery },
       // CodexSignIn's own endpoints: idle, not yet connected.
--- a/platform/app/src/components/settings/__tests__/ModelProviderForm.connection-check.integration.test.tsx
+++ b/platform/app/src/components/settings/__tests__/ModelProviderForm.connection-check.integration.test.tsx
@@ -177,10 +177,12 @@ describe("Feature: checking a credential from the drawer it was typed into", () 
       expect(checkButton()).toBeDisabled();
     });
 
-    it("becomes usable once the credential is entered", async () => {
-      await userEvent.type(inputFor("OPENAI_API_KEY"), "sk-typed-just-now");
+    describe("when the credential is entered", () => {
+      it("becomes usable", async () => {
+        await userEvent.type(inputFor("OPENAI_API_KEY"), "sk-typed-just-now");
 
-      await waitFor(() => expect(checkButton()).toBeEnabled());
+        await waitFor(() => expect(checkButton()).toBeEnabled());
+      });
     });
   });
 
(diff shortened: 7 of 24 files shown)


sergioestebance:
## Review of this branch

Self-authored, so per our convention the blocking findings are fixed here rather
than handed back, and the approval is not mine to give.

**Verdict: one P1, found and fixed. Four of my own suspicions refuted. Two P2s,
one fixed, one recorded.**

### P1 — the staleness guard was inert (fixed, `86d49df`)

`platform/app/src/hooks/useCredentialCheck.ts` — `check` captured both the
fingerprint and the credentials it was called with, so
`asked === JSON.stringify(customKeys)` compared a value with itself and was true
however much had changed since. Ask for a check, edit the credential before the
answer arrives, and the answer about the previous credential lands and puts
"Connection works" back on screen. That is a success claim about a key that is
no longer there, which is precisely what this feature is supposed to make
impossible, and the existing test could not catch it because a resolved mock
lands before the edit does.

Executed output, before the fix:

```
AssertionError: expected 'works' to be 'none' // Object.is equality
Expected: "none"
Received: "works"
 ❯ tmp-staleness.integration.test.tsx:84
```

Fixed with a ref holding the latest fingerprint — the only thing in scope that
moves, so the only thing that can answer whether these are still the credentials
on screen. The sibling hook on the provider list reaches the same conclusion via
a generation counter. A permanent regression test now holds the answer open
across the edit; reverting the fix fails it:

```
FAIL  ModelProviderForm.connection-check.integration.test.tsx >
      given a provider I am creating >
      discards an answer that arrives after the credential has changed
Tests  1 failed | 13 passed (14)
```

### P2 — the description stated a false mechanism (fixed)

The body claimed a stored key "seeds into the form as an empty string". It does
not: `maskRowCustomKeys` (`modelProvider.service.ts:1205`) replaces secret
fields with the masked placeholder before they reach the browser, and
`buildCustomKeyState` copies that value through. The conclusion is unchanged — a
mask is not a credential, so that route still checks nothing — but the stated
reason was wrong, and it was load-bearing for why the widening was needed.
Corrected in the description.

### P2 — a re-export with no consumers (fixed, in `86d49df`)

`ConnectionTestState` moved to its own module and every consumer already imports
it from there; the line left on `useModelProviderConnectionTest` gave the type
two import paths, which the repo's conventions forbid outright.

### P3 — em dash in moved copy (not fixed)

`connectionTestState.ts` carries "can't be tested automatically — its settings
are checked when you first use it", and the house style bans em dashes. It is
pre-existing text moved verbatim, not written here, so it is capped and left
alone rather than widening this PR into a copy edit.

### Refuted — suspicions that did not survive

- **Stored material re-entering a caller-supplied request.** It cannot.
  `validateProviderApiKey` takes credentials as a parameter and never touches
  Prisma; the only database read in that module is inside
  `validateKeyWithCustomUrl`, a different procedure this branch does not touch.
  The never-combine rule holds because there is no code path that merges.
- **The `hasEdits` discriminator reading a divergent source.**
  `state.initialKeys` *is* `originalStoredKeysRef.current`
  (`useModelProviderForm.ts:243`), the same source the dirty check and the save
  path both use. They cannot diverge.
- **The completeness predicate misbehaving on Azure's passthrough schema.**
  Unreachable: Azure is unprobeable, so the control never renders for it. The
  `custom` and Gemini cases are correctly gated by the two-part check, and the
  Gemini half-pair case has its own test.
- **The registry's new import direction tripping the frontend boundary.**
  `frontend-boundary.unit.test.ts` — 10/10 passed. The onboarding registry is
  framework-free despite the extension.
- **The drift pin passing vacuously.** Probed every provider's generated field
  set; none is empty, so no provider is asserted against an empty credential.

### Checks after the fixes

- `pnpm typecheck` — clean
- `check-feature-parity` — 54/54 bound
- Server unit: 495/495 across 36 files
- Settings and drawer integration: 265/265 across 37 files

### Two ways this review is weaker than it looks

**No independent refutation.** The convention is one subagent per finding,
attacking it from a context that did not form it. Subagents are unavailable in
this session, so I wrote and ran the decisive tests myself. The evidence above is
real executed output, so what was lost is not effort — it is independence. I
judged my own findings, and a review whose author is also its subject looks
identical to one that happened and is not the same thing.

**One commit carries two findings.** `git add -A` swept the re-export removal
into the staleness fix, so `86d49df` is not the one-commit-per-finding the
convention asks for. Recorded rather than rewritten.

No blessing here: I wrote this branch, so approving it would remove the last
checkpoint that something other than the author read the diff. That approval is
a human's to give.

sergioestebance:
## Second review pass

Different target from the first: that pass attacked the drawer hook and the
never-combine rule. This one went at the thing I had not examined at all — that
rerouting `validateApiKey` through the service changed the behaviour of a
mutation with **callers outside this feature**.

**Verdict: no new P0/P1. One P2 introduced, confirmed by executed test. Two P3s.**

### P2 (introduced) — a rate-limited check now blocks a save

`server/api/routers/modelProviders.ts:266` → `modelProvider.service.ts`
`validateCredential`.

Putting the typed-credential route behind the shared budget was the right call
for egress, but that route is not only used by the new button. It is also the
probe the **save path** runs (`ModelProviderForm.tsx:467`, and the same in
onboarding at `ModelProviderSetup.tsx:369`). When the budget rejects, the hook
returns false, `handleSave` reads that as "the provider refused this credential",
records a refusal, and returns without saving.

So an organization that has used its 20 checks in a minute cannot save a provider
on the first click — with a perfectly good key, for a reason that has nothing to
do with the key. Executed, with the save mutation stubbed:

```
AssertionError: expected "vi.fn()" to be called at least once
  (the update mutation was never called)
Tests  1 failed
```

**Attribution.** The conflation itself is pre-existing: on `main` a network error
or an unreachable provider already blocks a save the same way. What this branch
adds is a new trigger for it, and one we generate ourselves about a customer who
did nothing wrong. Severity tracks the change, so this is P2, not P1.

**Recovery works**, which is what keeps it off the blocking list. The probe gate
disarms after a refusal, so the second Save goes through unprobed — confirmed:

```
Tests  1 passed  (second Save reached the update mutation)
```

**Not fixed here, deliberately.** The correct fix is for the save path to
distinguish "we could not check" from "the provider said no", which means
touching `useCredentialProbeGate` and the save flow — explicitly out of this
PR's scope, and it would change save behaviour for the pre-existing triggers
too. Filing it rather than widening. The cheap partial mitigation, if the first
click matters more than the egress ceiling, is a separate budget key for the
save-time probe.

**This is the top residual risk on the branch.** 20 checks per organization per
minute is generous for one person and not obviously generous for a team
onboarding providers together on day one, and the failure it produces lands on
Save rather than on the button that caused it.

### P3 — a defensive cast that cannot fire

`hooks/useCredentialCheck.ts` — `describeUncheckedInDrawer` accepts
`UncheckedReason | "request_failed"` and casts back, but it is only ever passed
to `toConnectionTestState`, which calls it exclusively with server-side reasons.
The `request_failed` branch builds its own message in the catch. Harmless,
slightly misleading about what can reach it.

### P3 — an export with only a test consumer

`UNPROBEABLE_PROVIDERS` in `registry.ts` exists for the drift pin. Fine, but it
is production surface for a test's benefit; the per-provider agreement assertion
is the one doing the real work.

### Checked and clean

- **Anchor resolution for callers outside this feature.**
  `resolveOrganizationAnchor` falls back to `organizationId`, and both live
  callers refuse client-side when neither handle is present, so no flow loses
  the ability to validate. The no-project onboarding path passes an organization
  and resolves fine.
- **The live Agent Platform rows.** `google_agent_platform` reaches the door
  check and stays probeable, so the customer already carrying one keeps the
  action on both surfaces — verified by the per-provider agreement test, 18/18.
- **The list menu.** Hiding the item never empties the menu; Edit and Delete are
  untouched, and system rows were already menu-less.

### Same degradation as last pass

No independent refutation — subagents are unavailable in this session, so I wrote
and ran the probes myself and deleted them (`git status --short` clean). The
outputs above are real, but the reviewer and the author are the same context,
which is exactly the property a review is supposed to supply.

---

Blessed. 🙏 The check now refuses to speak about a credential it did not see —
hidden where it cannot ask, silent where it cannot know, and never combining a
stored secret with an address someone else chose.

> "Let what you say be simply 'Yes' or 'No'; anything more than this comes from evil." — Matthew 5:37

Traced the two probe routes, the never-combine rule, the anchor fallback for
onboarding, and the staleness guard that the first pass caught inert. Ship it —
residual is the P2 above: a rate-limited check blocks the first Save, recovers on
the second, and wants a separate budget key rather than a wider PR.

I wrote this branch, so the approval is not mine to give.

sergioestebance:
The P2 from the second review pass is now tracked as #6791 (bug / ux / P2, Backlog) — a rate-limited credential check blocks the first Save, because the save path cannot tell "the provider refused this key" from "the check never ran".

Filed rather than fixed here: the real fix is teaching `validate()` to distinguish the two, which touches both save paths and the probe gate, and would change behaviour for the pre-existing triggers (a provider outage blocks a save the same way on main today). That is a larger and better change than this PR should absorb.

sergioestebance:
## Findings settled

Rebased onto latest main first (`f05d495`), then worked the ledger. Every item
below was checked against the actual code path before being fixed or skipped.

| # | from | where | says | real? | did | why |
|---|---|---|---|---|---|---|
| 1 | behind main | — | 5 commits behind | yes | rebased | the lint error turned out to live in files only main had |
| 2 | CI | `test-integration (6)` | `useCredentialCheck` — cannot read `useMutation` | **yes, mine** | fixed | a ninth drawer test I missed lives in `components/__tests__`, outside the two directories I swept |
| 3 | CI | `lint` | 1 error, 2 new violations | **yes, mine** | fixed | formatting + cognitive complexity, both in `useCredentialCheck.ts` |
| 4 | coderabbit | `modelProvider.service.ts` | `validateCredential` has no authz of its own | **yes** | fixed | the rate-limit key was caller-chosen |
| 5 | coderabbit | `connectionTestState.ts` | positional second parameter | yes | fixed | house rule is named object params |
| 6 | coderabbit | `modelProvider.testConnection.unit.test.ts` | comment promises three guarantees, test asserts one | yes | fixed | the comment was wrong, not the test |
| 7 | coderabbit | `credential-validation.feature` | drawer scenarios tagged `@unit` | yes | fixed | they are bound to integration tests |
| 8 | coderabbit | connection-check test | missing `when` groups | yes | fixed | repo BDD convention |
| 9 | coderabbit | `registry.ts:3` | move `providerDefaultBaseUrls` out of `.tsx` | yes | **skipped** | see below |

### The one that mattered: #4

CodeRabbit was right, and it found something both of my review passes missed.

`checkProviderValidationPermission` authorizes the **scopes** the credential is
being set up for. The **anchor** is a different value, and on the no-project path
it came straight from a caller-supplied `organizationId` with nothing checking
the two were related. The anchor is the rate-limit key. So naming another
organization spent *their* allowance — which hands the caller an endless run of
fresh buckets *and* denies the real owner a control they are entitled to. A rate
limit whose key the caller picks is decorative, and I had just spent a commit
arguing that this route needed one.

Fixed by asserting the caller can manage the organization the budget is charged
to, but only on the caller-supplied path: an anchor derived from a project has
already been through the route's project permission gate.

Removing the guard fails the new test:

```
FAIL  modelProvider.testConnection.unit.test.ts >
      refuses to spend the budget of an organization the caller cannot manage
Tests  1 failed | 16 passed (17)
```

### #2 corrects something I said in the PR description

I wrote that "seven existing drawer suites" needed the new mutation mocks. There
were eight — `components/__tests__/EditModelProviderDrawer.edit-row-resolution.integration.test.tsx`
sits outside the two directories I searched, and CI caught it. Fixed and the
description no longer claims a count.

### #9 — skipped, with the reason

Moving `providerDefaultBaseUrls` into a framework-free `.ts` module is a fair
call about the shape, but the import it objects to is **pre-existing**:
`providerValidation.ts` has value-imported that same `.tsx` module since before
this branch. My change adds a second importer of an already-shared value, not a
new crossing.

The enforced guard agrees — `frontend-boundary.unit.test.ts` walks the real graph
and passes 10/10, because the module is plain data despite its extension. Moving
it means touching the onboarding registry and every consumer of that export, for
no behaviour change, in a PR that is already carrying a security fix. That is a
tidy-up on its own branch, not a rider on this one.

### CI

- `lint` — the gate now reports `no new Biome violations` against the merge base
  (3725 → 3725). Before the fix it read `format: 0 -> 1` and
  `noExcessiveCognitiveComplexity: 0 -> 1`, both in `useCredentialCheck.ts`.
- `test-integration (6)` — the four failures were mine (finding #2), not the
  ClickHouse connection noise also in that log. Fixed.
- Locally after the fixes: typecheck clean, `typecheck:tests` clean, 496/496
  server unit, 271/271 across 38 integration files, 54/54 scenarios bound.

### Process note

Two of these commits carry more than one finding — `git add -A` swept them
together, against the one-commit-per-finding convention. Same slip as the earlier
review pass, recorded rather than rewritten.

sergioestebance:
## Second round settled

CodeRabbit re-reviewed after the last push and caught a real bug **in the fix
from my own first review pass**. That one is worth reading.

### The staleness guard was keyed on one thing and cleared on another

`useCredentialCheck` cleared the verdict when either the credentials or the row
changed, but stamped its guard with the credentials alone. Two saved rows show
the same masked placeholder and can carry the same endpoint, so what is on
screen is character-for-character identical for both — only the row identity
separates them. A check still in flight when the drawer moved to the second row
therefore survived the clear and reported the **first row's answer about the
second**.

Same failure the guard exists to prevent, wearing a different hat, and my
earlier fix walked straight past it because I only thought about the credential
changing, not about the subject changing underneath it.

Confirmed before fixing — the regression test switches between two rows the
existing harness builds identically:

```
FAIL  ModelProviderForm.connection-check.integration.test.tsx >
      given two saved rows that look identical on screen >
      does not land the first row's verdict on the second
AssertionError: expected <p class="css-18o0loe"></p> to be null
Tests  1 failed | 14 passed (15)
```

Two keys that have to agree are one key, so the row is now part of the
fingerprint and the effect keys on that single value. 15/15 after.

### Also settled this round

- The remaining `when`-group placements CodeRabbit flagged (three more actions
  that sat outside a block while their siblings sat inside one).
- The test file's header still said it covered `@unit` scenarios after the
  retag; it says `@integration` now.

### Where it stands

- Lint gate against the merge base: `no new Biome violations` (3725 → 3725).
- `typecheck` and `typecheck:tests` clean.
- 272/272 across 38 integration files; 55/55 scenarios bound.
- Still open and deliberately skipped: moving `providerDefaultBaseUrls` out of
  the `.tsx` module — pre-existing import, enforced boundary guard passes,
  reasoning in the previous comment.

Worth saying plainly: this is the second defect found in code I had already
reviewed twice myself, and both were found by something that did not write it.
That is the argument for the independent pass, not a footnote to it.

sergioestebance:
## Review of this turn's changes, and a regression it caught

The gates asked for an adversarial read of the scenario and the diff. Both found
something, and one of them was a lockout I introduced two commits ago.

### The scenario survived, but only after checking its trigger exists

`A result still in flight when I move to another provider is discarded` describes
switching rows mid-check. My test forces that with `rerender`, which proves
nothing on its own — if the real drawer **remounts** when the row changes, the
state resets by itself and the scenario describes something no customer can
reach.

It does not remount. Neither `CurrentDrawer` nor `EditModelProviderDrawer` puts a
`key` on the form, so React reconciles by type and keeps the instance across a
`modelProviderId` prop change. The drawer is URL-routed with back/forward history
(`dev/docs/best_practices/drawers.md`), so opening row A, then row B, then
pressing Back is a real path from one row to another with the component alive
throughout. Trigger real, surface real, assertion observable. Scenario stands.

### The regression: I locked team admins out

`validateCredential` asserted `assertCanManageAllScopes(ctx, [{ORGANIZATION, anchor}])`,
which resolves to **`organization:manage`**. The no-project path exists for
scopes *below* the organization — a team admin sets up a team-scoped credential
and holds `team:manage`, never `organization:manage`. The route had already
authorized them against the scopes they supplied; my guard then refused them at
the budget.

Confirmed before fixing:

```
FAIL  modelProvider.testConnection.unit.test.ts >
      lets a team admin check a credential for their own organization
ModelProviderScopeForbiddenError: You don't have permission to manage model providers here.
Tests  1 failed | 17 passed (18)
```

Fixed by asking the right question. Charging an organization's budget is not an
act of administration, so it should not need administrative permission — what a
stranger cannot fake is being able to *see* the organization at all. The check is
now `organization:view`, which a member has and an outsider does not.

Both directions are pinned: the team admin gets through, an outsider naming
someone else's organization is refused, and removing the guard still fails the
second test.

This is the shape worth noting — the fix for a real hole introduced a smaller
one by reaching for the nearest-looking helper instead of the right permission
tier. Tightening authz is exactly where that is easy to do and hard to notice,
because the failure mode is a refusal, and refusals look like the guard working.

### Checks

- Lint gate vs merge base: `no new Biome violations` (3725 → 3725)
- `typecheck` and `typecheck:tests` clean
- 497/497 server unit · 272/272 across 38 integration files · 55/55 scenarios bound

Commit: `a55d438`.

PR #6823, opened 2026-08-10
perf(ci): one Go version, and stop two buildx caches evicting each other

0xdeafcafe:
Stacked on #6820 — review that first, or read this diff against `worktree-ci-checkout-speed`. It needs #6820's `tools/ciguard` + `pkg/ciscan`.

## One Go version, stated once

It lived in eight places and had already drifted:

```
  go.work                              1.26.5
  go.mod                               1.26.5
  infra/clickhouse-serverless/go.mod   1.26.1
  infra/clickhouse-serverless/Docker*  golang:1.26-alpine   <- floating
  sdks/go/go.mod                       1.25.0 (deliberate)
  sdk-go-ci.yml, sdk-go-cd.yml         "1.25"
  clickhouse-serverless.yml            "1.26"
```

Nothing failed. Different jobs simply compiled the same code with different toolchains, and the clickhouse-serverless image was built with a third — its floating tag resolves to whatever `1.26.x` was newest that day.

**Every workflow now reads `go-version-file`** from the module it builds, so there is no literal left to drift. That is also *more* correct than the hardcodes: `sdk-go` tested at `"1.25"` while its module declared `1.25.0`, and now tests at exactly its declared floor.

**The SDK keeps its floor.** Its `go` directive is what consumers must meet, and raising it drops support for anyone below — which is why `sdk-go-ci` and `-cd` already build it standalone with `GOWORK=off`. Consistency here means "read the version from the module", not "bump everything to 1.26". It is recorded as an exemption that carries its reason rather than left as a silent difference.

**clickhouse-serverless moves to 1.26.5** in both its module and its Dockerfile. It is internal, not published, so its floor is ours to set.

## The guard

Dockerfiles cannot read a `go.mod`, so `ciguard go-version` is what keeps them honest. It rejects a mismatched patch, a floating tag, a `go.work` that disagrees, and any workflow that pins a literal. Fixture-driven — each of those four has a test that fails without the rule — plus one that runs it against this repository.

## Two buildx caches sharing one scope

`langevals-ci` and `langwatch-nlp-ci` were the only unscoped `type=gha` caches left. That is worse than it sounds: unscoped means they both wrote to buildkit's *default* scope, so they were evicting each other's layers on every run, both at `mode=max` against a 10 GB per-repo budget. Now scoped, like every other buildx cache in the repo.

## Retracted

I reported this earlier and it is wrong, so it is not in this PR: *"go-services builds linux/arm64 under QEMU on main pushes while publish-docker-app uses native arm64 runners."*

`Dockerfile.go_service` pins its build stage to `$BUILDPLATFORM` and cross-compiles with `GOARCH=${TARGETARCH}`, and its runtime stage is `distroless/static`. There is nothing to emulate. `setup-qemu-action` registers binfmt handlers the build never uses — a second or two — and converting it to a native matrix would have been a large rewrite of a publishing path for no gain.

## Verification

- `go test ./tools/ciguard/...` — passes.
- `go run ./cmd/ciguard` — both guards clean against this tree.
- `golangci-lint run ./pkg/ciscan/... ./tools/ciguard/... ./cmd/ciguard/...` — 0 issues.
- `GOWORK=off go build ./...` in `infra/clickhouse-serverless` at the bumped version — clean. (A fresh worktree fails there on a missing `go.sum` entry both before and after the bump; that is pre-existing and unrelated.)
- All six scenario titles in the new spec match a `@scenario` annotation exactly.

### DIFF
--- a/.github/workflows/clickhouse-serverless.yml
+++ b/.github/workflows/clickhouse-serverless.yml
@@ -54,7 +54,9 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: "1.26"
+          # Its own module declares the version; "1.26" here silently resolved
+          # to whatever 1.26.x was newest that day.
+          go-version-file: infra/clickhouse-serverless/go.mod
           cache-dependency-path: infra/clickhouse-serverless/go.sum
       - uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
         with:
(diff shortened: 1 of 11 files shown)


sergioestebance:
lgtm

PR #6831, opened 2026-08-10
perf(ci): give the last 20 workflows a concurrency group

0xdeafcafe:
## What

Every workflow in the repo now declares a `concurrency` group except one, and each group either cancels superseded runs or queues them — chosen by what the workflow *does*, not by what it costs.

Before this, 21 of 56 workflows had no group at all. Two of those gaps were live races, not just wasted minutes.

## Why

**Gates cancel.** A gate's verdict describes the PR as it is *now*. `release-scope-guard`, `pr-conventional-commits-title` and `deployment-impact-check` all take `edited` in their trigger list, so re-titling a PR or editing its body starts a run — and all three read the title, body or labels at read time. An older run in flight is already answering a question about a state that no longer exists.

Measured over the last 100 runs of each, counting runs still going when a newer run for the same branch started:

| workflow | superseded | avg run |
|---|---:|---:|
| `release-scope-guard` | 29 / 100 | 107s |
| `pr-conventional-commits-title` | 18 / 100 | 95s |
| `deployment-impact-check` | 18 / 100 | 68s |
| `release-please-chart-lock-sync` | 5 / 100 | 101s |
| `workflow-security-guard` | 3 / 100 | 102s |
| `coderabbit-config-check` | 2 / 100 | 178s |
| `issue-link` | 0 / 100 | 133s |

They cancel on PR refs and never on `main` — `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`, the idiom the other 20 workflows here already use — so a commit that lands on main keeps its own validation.

**Writers queue.** A cancelled publish has no cleanup path: it reports `cancelled` rather than `failed`, so nothing retries it and nothing reconciles it. That reasoning is already written down in `prerelease-charts-oci.yml`; this applies it to the remaining publishers rather than restating it in each.

### The two real races

**`release-please-sdks` — 12 of its last 60 runs overlapped a previous one.** It fires on every merge to `main`, and a burst of merges starts overlapping runs that compute a release plan from the same history and then race on the same release PRs. The observable outcomes are a duplicate release PR, or one run reverting the other's version bump. This is the one that was actually happening.

**`release-langwatch-chart` and `release-clickhouse-serverless-chart` share a `gh-pages` index.** Both run `cr index --push` with the same `--owner`/`--git-repo`, and `cr index` rebuilds `index.yaml` from the repository's release list and force-pushes it. Run together, the loser pushes an index it built before the winner's `cr upload` registered — so a chart version that published successfully disappears from the index. The release still exists; `helm repo update` just stops seeing it.

They share **one** group, `helm-chart-index`. A per-workflow group would not have helped: the race is *between the two workflows*, not between runs of either.

## What is deliberately left alone

**`issue-link.yml`** — triggers on `types: [opened]`, which fires exactly once per PR. It has no second run to collide with, and 0/100 overlapping runs confirms it. A group here would be symmetry for its own sake.

**`langwatch-chart.yml` and `clickhouse-serverless-chart.yml`** — the only other two without a group. They get theirs in #6820, so they are left untouched here to avoid a conflict.

## Testing

- `actionlint` over the whole tree: no syntax or expression errors. The one finding is a pre-existing unknown-runner-label warning on `publish-docker-app` for the self-hosted `arm64-runner-32gb`.
- Coverage re-checked after the change: 48 workflows with a group, 3 without (the two above plus `issue-link`).
- No literal group-name collisions except the intended `helm-chart-index`. The shared `${{ github.workflow }}-…` templates differ per workflow by construction.

## Risk

Low, and mostly one-directional: every change adds a group where there was none, so nothing that ran before stops running. The behaviour changes are that a superseded gate run now ends as `cancelled` instead of finishing, and a second publisher run now waits instead of starting immediately.

The one judgement call worth a second opinion is `release-please-chart-lock-sync`: it pushes a `Chart.lock` commit to the release PR's branch, so I gave it `cancel-in-progress: false` (queue, don't cancel) rather than treating it as a gate. Serialized, the second run finds the lock already in sync and no-ops; cancelling it mid-push would be a force-push race instead.

### DIFF
--- a/.github/workflows/agent-plugin-publish.yml
+++ b/.github/workflows/agent-plugin-publish.yml
@@ -17,6 +17,12 @@ on:
       - "agent-plugin@v*"
   workflow_dispatch:
 
+# A publish run drains — see publish-docker-app. Tag pushes are unique, so the
+# collision this actually guards is a re-dispatch of a tag already publishing.
+concurrency:
+  group: agent-plugin-publish-${{ github.ref }}
+  cancel-in-progress: false
+
 permissions:
   contents: read
 
--- a/.github/workflows/coderabbit-config-check.yml
+++ b/.github/workflows/coderabbit-config-check.yml
@@ -17,6 +17,10 @@ on:
       - "dev/lint/**"
       - ".github/workflows/coderabbit-config-check.yml"
 
+concurrency:
+  group: coderabbit-config-check-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
--- a/.github/workflows/deployment-impact-check.yml
+++ b/.github/workflows/deployment-impact-check.yml
@@ -19,6 +19,13 @@ on:
       - "platform/app/.env.example"
       - "docs/self-hosting/**"
 
+# Triggered on `edited`, and it reads the PR description, so an in-flight run
+# is grading a body the author has already rewritten. 18 of the last 100 runs
+# were superseded that way.
+concurrency:
+  group: deployment-impact-check-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   pull-requests: read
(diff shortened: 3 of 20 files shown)


sergioestebance:
lgtm

Issue #6835, opened 2026-08-11
Evaluation display/export surfaces invent verdicts: green PASS at 0.00, skipped counted as passed, comment-only annotations export as Thumbs Down


sergioestebance:
## What this is

Five display/export surfaces invent a verdict where none exists — a "no verdict" state (skipped, errored, score-only, comment-only) is rendered or exported as a pass, a fail, or a zero. Each is small; together they mean users cannot trust green/red anywhere the third state occurs. Grouped because the fix pattern is identical: preserve "no verdict" as its own rendering, never coerce.

All five adversarially verified as live surfaces (including checking the legacy pages are still routed).

## 1. traces-v2: green PASS with "0.00" for verdict-less evaluators

`features/traces-v2/hooks/useTraceEvaluations.ts:71-75` returns `"pass"` for a processed evaluation with `passed == null`, and `:86-90` gives it score 0. `isNoVerdict` (`evalCards/utils.ts:48-50`) covers only skipped/error, so nothing neutralizes this: EvalCard renders a green PASS header with "0.00 / 1.00" and the header chip shows a green dot with "0.00". The shared helper already has the correct neutral arm (`utils/evaluationResults.ts:371-374`, processed+null → neutral "processed") — the hook rewrites the status before it can apply.

## 2. Legacy messages list: skipped counted as passed

`components/messages/MessageCard.tsx:82-86` counts a check as passed when `evaluationPassed(check) !== false` and status is processed **or skipped**. The all-skipped guard (`:91-93`) misses the mixed case: 1 passed + 1 skipped renders a green CheckCircle "2/2 evaluations". Surface is legacy but routed (`pages/[project]/messages.tsx:37`, in main nav flagged legacy; MessageCard renders in list view). Sibling: `components/traces/Evaluations.tsx:151-156` counts `status === "error"` into a red pill labeled "{n} failed" — a crashed evaluator reads as a fail verdict.

## 3. Annotations CSV: comment-only exports as "Thumbs Down"

[...]

PR #6836, opened 2026-08-11
fix: preserve no-verdict states in eval display, export and filters

waterWang:
Fixes the "invent a verdict" family from #6835 — no-verdict states (skipped, errored, score-only, comment-only) must never be coerced into a pass/fail or a zero.

**Item 1 — traces-v2 verdict-less evals render neutral.** `mapStatus`/`mapScore` in `useTraceEvaluations` rewrote a processed evaluation with neither `passed` nor `score` into `pass` + `0`, before the shared neutral arm in `evaluationResults` could apply. Preserve `processed` + `null` so the eval card, history stack and header chip show the evaluator ran without a verdict instead of inventing a green PASS at 0.00.

**Item 2 — skipped evaluations are not passes.** `MessageCard` counted skipped as passed (`evaluationPassed !== false`), so a mixed passed+skipped trace rendered a green "2/2 evaluations". Only processed+passed counts now; mixed shows neutral. `EvaluationsCount` collapsed errored evaluators into "failed" — separated so crashed evaluators read as "errored", not a fail verdict.

**Item 3 — comment-only annotations export neutrally.** `isThumbsUp` is null for comment-only annotations; the ternary fabricated "Thumbs Down" in the CSV export. Now exports "—".

**Item 5 — skipped evaluator verdict is filterable.** The `evaluatorVerdict` facet spilled skipped into "unknown"; added a `skipped` arm to the SQL `multiIf`, the JS mirror, and the autocomplete vocabulary together.

Item 4 (runtime-skipped monitor runs visibility) is a product decision and intentionally left out.

Each item is a separate commit. No wallet/bounty — pure contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluation results now distinguish passed, failed, skipped, processed, and errored states.
  * Skipped and processed evaluations are no longer incorrectly counted as passed or failed.
  * Evaluation badges and counts now use clearer status labels and visual indicators.
  * Trace filters and search suggestions now support skipped evaluations.
  * Evaluations without scores remain unrated instead of displaying a zero score.
  * Annotation CSV exports now distinguish unrated entries from explicit positive and negative ratings.

### DIFF
--- a/platform/app/src/features/traces-v2/components/TraceDrawer/evalCards/utils.ts
+++ b/platform/app/src/features/traces-v2/components/TraceDrawer/evalCards/utils.ts
@@ -38,6 +38,10 @@ export const STATUS = {
   // red.700 from the shared map so "the evaluator broke" reads as a
   // separate failure mode from "the evaluator ran and said no".
   error: buildStatusTone("error", "ERROR"),
+  // Evaluator completed but produced no verdict — no score was assigned,
+  // no pass/fail was decided. Neutral blue keeps it from competing with
+  // real pass/fail rows while still showing that the evaluator ran.
+  processed: buildStatusTone("processed", "PROCESSED"),
 } as const;
 
 /**
@@ -46,7 +50,7 @@ export const STATUS = {
  * meaningless and should be suppressed.
  */
 export function isNoVerdict(status: EvalSummary["status"]): boolean {
-  return status === "skipped" || status === "error";
+  return status === "skipped" || status === "error" || status === "processed";
 }
 
 /**
--- a/platform/app/src/components/messages/MessageCard.tsx
+++ b/platform/app/src/components/messages/MessageCard.tsx
@@ -80,10 +80,11 @@ export function MessageCard({
       check.status == "error",
   );
   const evaluationsPasses = evaluations.filter(
-    (check) =>
-      evaluationPassed(check) !== false &&
-      (check.status === "processed" || check.status === "skipped"),
+    (check) => check.status === "processed" && evaluationPassed(check) === true,
   ).length;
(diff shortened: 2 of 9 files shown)


platform/app/src/features/traces-v2/components/TraceDrawer/evalCards/utils.ts
```diff
@@ -46,7 +50,7 @@ export const STATUS = {
  * meaningless and should be suppressed.
  */
 export function isNoVerdict(status: EvalSummary["status"]): boolean {
-  return status === "skipped" || status === "error";
+  return status === "skipped" || status === "error" || status === "processed";
```
langwatch-agent, line 53: **P1 — Keep numeric scores visible for score-only evaluators.** `mapStatus` now returns `processed` for every completed evaluation with `passed == null`, including the established score-only shape (`score: number`, `passed: null`). This line makes every `processed` result a no-verdict state, and the drawer then suppresses its numeric score and score bar. The result is that completed evaluators which deliberately emit a numeric measurement but no boolean threshold lose their only result in the trace drawer. Make the rendering decision depend on the absence of a score/verdict (or keep numeric score-only runs renderable) rather than treating `processed` status itself as scoreless.

langwatch-agent (commented):
Found one correctness regression in the current head.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=6482898b3beb2f8e422784ad81a48b67b7fbce05 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=6482898b3beb2f8e422784ad81a48b67b7fbce05 p0=0 p1=0 p2=0 p3=0

PR #5748, opened 2026-07-13
test(scenarios): add ClickHouse integration test for findQueuedRunCandidates (#5073)

langwatch-agent:
## Ticket

Closes #5073. Deferred follow-up from PR #5008: the orphan reconciler's cross-tenant candidate scan (`findQueuedRunCandidates`) had no ClickHouse integration test for its `argMax`/`HAVING` dedup SQL.

## Change

Adds `src/server/scenarios/__tests__/scenario-orphan-reconciler.integration.test.ts`, which seeds `simulation_runs` (multi-version runs across statuses, ages, and the lookback boundary) against a real ClickHouse testcontainer and asserts the query semantics. Scoped to a single new test file, no production code touched.

Results are filtered to the test's own tenant before asserting, since the scan is cross-tenant by design.

## Evidence (each AC to a test)

| Acceptance criterion | Test |
|---|---|
| latest `Status` resolved via `argMax(col, UpdatedAt)`, not `max(col)` | "includes a run whose latest version is QUEUED even though an earlier version was terminal" (max(Status) would pick `SUCCESS` and drop it) + "resolves per-run columns from the latest version via argMax" |
| only groups whose latest `Status` is `QUEUED` and older than the orphan cutoff are returned | "returns a run whose latest version is QUEUED and older than the orphan cutoff" + excludes recent-QUEUED / recovered / non-QUEUED runs |
| `StartedAt` lookback prune | "excludes a QUEUED run outside the lookback window" |
| `ORDER BY LastEventAtMs ASC` selects oldest first | "orders candidates oldest-first by last event time" |

Run locally against the integration testcontainers:

```
pnpm test:integration --run src/server/scenarios/__tests__/scenario-orphan-reconciler.integration.test.ts
-> Test Files 1 passed (1) | Tests 8 passed (8)
```

`pnpm typecheck` clean.

The `LIMIT 10000` cap itself is not separately exercised (it would need 10k+ seeded rows); the oldest-first ordering it depends on is covered.

## Advisory note

Advisory PR from the tech-debt-fixer bot, for human review, not to be merged by the bot.

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for identifying orphaned queued scenario runs.
  * Verified filtering by status, update age, and start-time lookback.
  * Confirmed latest-version data is used correctly.
  * Verified candidates are returned in oldest-first order.

PR #6843, opened 2026-08-11
fix(aigateway): forward Vertex deployment map into key aliases; classify status-less config errors as non-retryable

drewdrewthis:
> **Re-scoped 2026-09-06.** The Azure/Bedrock/Vertex dispatch-path self-map this PR originally carried (Defect A below) was independently fixed on `main` by https://github.com/langwatch/langwatch/pull/7778, whose copy-on-write body rewrite also avoids a retry-leak this PR's in-place `sjson.SetBytes` rewrite had. This PR now carries only what is unique to it: **Vertex forwards its deployment map into `Key.Aliases`** (Azure and Bedrock already did) and **the config-error classification work (Defect B)**. The earlier description is rewritten below; the previous text is in the edit history.

## Why

Azure / Bedrock / Vertex credentials materialised on the control-plane (virtual-key) path never received a deployment map. Azure routes on *deployment*, not model id, so Bifrost looks up the deployment for the model, finds nothing, and fails with `AI_APICallError: deployments not set`.

Two defects were originally in scope:

- **A — no deployment map on the control-plane path.** Fixed on `main` by https://github.com/langwatch/langwatch/pull/7778 (`domain.WithDeploymentSelfMap` + `requestWithResolvedDeployment` on both `Dispatch` and `DispatchStream`). Not in this PR any more. What #7778 did not close: `credentialToBifrostKey` forwarded `cred.DeploymentMap` into `Key.Aliases` for Azure and Bedrock, but **not for Vertex**, so a Vertex credential resolved a deployment that never reached the key. That one-line gap is fixed here and is what the three open `langwatch-agent` review threads asked for.
- **B — the resulting configuration error is misclassified as retryable.** `classifyBifrostError` mapped status `0` alongside `504` to `provider_timeout`, so a permanent misconfiguration surfaced as an HTTP 504 and was walked through the retry/fallback chain. Fixed here.

The branch name references `langwatch/langwatch-saas#1016` (private repo; renders as plain text without access).

## What changed

**Vertex key aliases.** `services/aigateway/adapters/providers/bifrost.go`, `credentialToBifrostKey`, `case bfschemas.Vertex:` now sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)`, identical to the Azure and Bedrock arms. Pinned by `providers/deployment_map_key_aliases_test.go` for all three providers: a nil wire map self-maps to `model -> model` and reaches the aliases; an explicit map is forwarded verbatim.

**Defect B — configuration errors are not timeouts.** `0` was removed from the `http.StatusGatewayTimeout` case; status-less Bifrost errors now go to `statuslessBifrostCode`, which discriminates on *shape* rather than message text. Every transport-failure path in the vendor (`bifrost/core@v1.4.22`) sets `Error: err` or `Error: ctx.Err()`, so a status-less, `Error`-less shape is never a transport failure. It is **not**, however, uniquely a configuration error: the vendor raises status-less, error-less shapes from request validation too, and no shape-only discriminator separates those from a configuration fault. The validation shapes are swept in deliberately: they are permanent for that request, so non-retryable is the right answer, and they are pinned as explicit rows in `bifrost_config_error_classification_test.go` so the sweep-in is a decision rather than an accident.

**The sweep is bounded by an exhaustive vendor enumeration.** The gateway calls eleven distinct operations on the `*bifrost.Bifrost` receiver; ten can return an error. Core raises `BifrostError{IsBifrostError:false, Error:&ErrorField{Message: …}}` (no status, no wrapped Go error, no code) from the provider-queue machinery every request operation goes through. Of the 19 production call sites of the unexported `newBifrostErrorFromMsg`, 9 are permanent argument validation and 10 are transient `"provider is shutting down"`; two more build the same shutdown value by hand and two raise `"request dropped: queue is full"`. The shutdown lane is reachable (we call `RemoveProvider` on anthropic-compat endpoint eviction); the queue-full lane is gated on `config.DropExcessRequests`, which nothing under `services/` sets, and is classified anyway so turning it on later cannot silently reintroduce the defect.

- **Empty upstream response body.** The vendor builds the identical status-less, `Error`-less, `Code`-less shape via `schemas.ErrProviderResponseEmpty`. A dropped or truncated response is the textbook transient fault, so it is excluded, keyed on the vendor's **exported constant** rather than message text, so a vendor rewording breaks the build instead of silently un-matching.
- **`"provider is shutting down"` — known residual, deliberately not fixed.** Raised through an *unexported* vendor helper with no exported symbol to key on, so both the check and its test would hardcode the same string literal. Blast radius: one request loses fallback during an eviction race; self-healing on the next dispatch. Documented in-code.

**New herr code** `provider_misconfigured` at HTTP 502, registered across the generated TS codes (`packages/handled-error/src/codes.generated.ts`), the error-presentation map (`platform/app/src/features/errors/logic/presentation.ts`), and `adapters/httpapi/faults.go`. Reusing `provider_error` would keep a permanent configuration failure inside the retry and circuit-breaker path (`classifyProviderError` maps it to `retry.ReasonRetryable5xx`). The new code falls through to `retry.ReasonNonRetryable` and appears in neither `defaultTriggers` nor `breakerFailureReasons`.

**Trade-off, stated:** because the fault no longer walks the fallback chain, a request that previously succeeded on a healthy *second* credential now fails with a 502 on the first. The common case is that every credential on the path shares the same misconfiguration, so the walk burned latency and breaker budget to reach the same failure, and a misconfiguration that limps along on a backup credential is a worse operational outcome than one that fails loudly. It is a behaviour change with a loser, not a free win.

**Naming caveat:** validation-shaped faults that are arguably the *caller's* 400 are reported as the *provider's* 502 under `provider_misconfigured`. Retry semantics are right for both; splitting them would require discriminating on vendor message text, which is what this PR removes. Flagged rather than fixed.

## Test plan

| File | What it pins |
|---|---|
| `providers/deployment_map_key_aliases_test.go` | Azure, Bedrock and Vertex all forward the (self-mapped or explicit) deployment map into `Key.Aliases` |
| `domain/deployment_self_map_chokepoint_test.go` | Non-mapped providers (OpenAI, Anthropic, Gemini, custom) come out of `WithDeploymentSelfMap` untouched, nil map stays nil |
| `controlplane/config_wire_deployment_map_test.go` | Wire-JSON rows: a configured `deployment_map` reaches the credential verbatim; absent and empty maps are indistinguishable |
| `providers/bifrost_config_error_classification_test.go` | Classification built from the vendor's own error constructors, with `require` preconditions so a drifted fixture fails loudly; pins each status to its exact code |
| `httpapi/azure_config_error_status_test.go` | End-to-end through the real router: a configuration error returns 502 with `error.code == "provider_misconfigured"`, and a second, working credential in the fallback chain is **never dialled** (`hits == 0`) |
| `app/azure_config_error_retry_test.go` | `provider_misconfigured` classifies as `ReasonNonRetryable`, with `provider_error`/`provider_timeout` pinned retryable in the same table |

Spec: `specs/ai-gateway/azure-deployment-map-control-plane-path.feature` (10 bound scenarios, 1 `@unimplemented` for the codegen/TypeScript surfaces). The dispatch-path scenarios moved to `specs/ai-gateway/azure-endpoint-from-api-base.feature` with #7778.

Removed in the re-scope: `dispatchCredential`, `applyAzureRawForwardDeployment`, this PR's `azure_deployment_selfmap_test.go` (main's file of the same name from #7778 is kept), the every-lane coverage test, and the "exactly one `WithDeploymentSelfMap` call site" structural test (main legitimately has two: `Dispatch` and `DispatchStream`).

Local: `go build ./...` and `go test -count=1 ./services/aigateway/...` run through `dev/scripts/check-queue.mjs` on the re-scoped head. CI is the authority.

## Deployment Impact

**Env vars:** none added, removed, or changed.

**Helm values:** none added or changed. No files under `charts/` are touched.

**Default behavior on a vanilla `helm install`:** unchanged in configuration terms. Runtime behavior changes for Vertex credentials on the control-plane path (the deployment map now reaches the key) and for anyone keying on the old error code.

**One observable change operators should know about:** a provider *misconfiguration* now surfaces as HTTP **502** (`provider_misconfigured`) instead of HTTP **504** (`provider_timeout`), and is no longer retried or counted toward the circuit breaker. Any dashboard, alert rule, or log filter keyed on 504-from-aigateway for provider config problems should be repointed at 502 / `provider_misconfigured`.

**BYOC dataplane:** the aigateway runs in the customer dataplane, so BYOC picks both changes up on the next image roll. No dataplane configuration, secret, or network change is required; the new `provider_misconfigured` code is additive, and a dataplane on the older image keeps emitting the old code.

## Human verification

Backend change, but with one user-visible surface: `platform/app/src/features/errors/logic/presentation.ts` carries a customer-facing entry for `provider_misconfigured`, titled **"This provider isn't fully set up"**, with body copy *"Finish configuring the provider in your settings, then try again."* **It has not been rendered on screen**; it is verified only by reading the map and the code's registration in the generated TS codes.

To verify by hand:

1. Point an Azure/Bedrock/Vertex credential at a genuinely missing endpoint and confirm the response is HTTP **502** with body `error.code == "provider_misconfigured"`, not 504 / `provider_timeout`, and that it returns promptly rather than after walking the retry chain.
2. Confirm the UI renders "This provider isn't fully set up" for that response. **Unverified; the step most worth a reviewer's eyes.**
3. On a Vertex credential issued on the virtual-key path, confirm the request dispatches on the resolved deployment.

- **Vertex aliases, proven by test shape:** `providers/deployment_map_key_aliases_test.go` runs the same two assertions for Azure, Bedrock and Vertex. With the one-line Vertex hunk reverted, the two Vertex rows fail (`key.Aliases` is nil) while Azure and Bedrock still pass, which is exactly the asymmetry the review threads reported.
- **Config-error classification, proven by mutation** (from the earlier audit, still valid on this head): `statuslessBifrostCode -> ErrProviderError` reds the classification test and both Azure lane tests; removing the `RegisterStatus` line reds the status baseline row (502 -> 500); the dispatch `default:` arm -> `ReasonRetryable5xx` reds the retry classification test. The classification test asserts the exact code, with `provider_error` and `provider_timeout` pinned retryable in the same table, so it cannot pass by landing on a different retryable code.
- **Local run on `7b2497830a`:** `go test -count=1 ./services/aigateway/...` green through `dev/scripts/check-queue.mjs`; CI is the authority.

Known gaps:

- Neither 502 nor the new code can be discriminated by status alone (`provider_error` is also 502). The discriminating assertions are the body's `code` field and `hits == 0` on the fallback credential.
- `pkg/retry`'s `recordBreaker` sends every non-retryable outcome to a `default:` arm that calls `RecordSuccess`, so a permanently misconfigured slot is recorded as a circuit-breaker success. Pre-existing for all `ReasonNonRetryable` outcomes, not introduced here; tracked at https://github.com/langwatch/langwatch/issues/6850.
- `models_discovery.go` `deploymentMapModelIDs` and the Bedrock VPCE `Extra["deployment"]` override remain out of scope; tracked at https://github.com/langwatch/langwatch/issues/6848 (the Vertex half of that issue is closed by this PR).

## Anything surprising?

- **One pre-existing test was edited:** `TestErrFromBifrost_NoStatusFallsBackToClassify` in `bifrost_test.go`. Its old fixture had no status *and* no wrapped error and expected `provider_timeout`, a shape the vendor cannot emit, which encoded the defect as intended behaviour. The fixture now carries a real `Error` and expects `provider_error`. Verified against `bifrost/core@v1.4.22/providers/utils/utils.go`.
- **The end-to-end test uses a missing *endpoint*, not a missing deployment map,** as its instrument, since with #7778 a missing map no longer errors at all.

## UI surface

No LangWatch app view, component, or route is touched. The one TS change is a string entry in the error-presentation map for the new code, described under Human verification.


drewdrewthis:
### Sweep: both fixed patterns, swept repo-wide

Two patterns were fixed here; I swept the repo for other live instances of each.

**Pattern 1 — a provider credential reaches a dispatcher without its deployment map resolved.**
Zero live gaps. Every Bifrost dispatch lane in `services/aigateway` (chat, streaming, Codex, Responses, Embeddings, Speech, Transcription, Passthrough, Bedrock VPCE, Anthropic-translated) is a sub-call of `Dispatch`/`DispatchStream`, both of which pass through the new `dispatchCredential` chokepoint *before* branching — established by tracing every `dispatch*` call site, not by assumption. `bedrock_vpce.go:137` reads `cred.DeploymentMap[model]` from a credential already resolved at `bifrost.go:221`. The two `nlpgo` non-gateway entry points (`dispatcheradapter/adapter.go:86,109` and `gatewayproxy/headers.go:165`) already call `domain.WithDeploymentSelfMap` themselves — that is the pre-existing #5760 fix, and re-applying it is a documented no-op. `services/langyagent` never constructs or forwards a Bifrost credential at all.

**Pattern 2 — a status-less or configuration-shaped error classified as retryable/timeout.**
Zero live gaps. The two other `status == 0` sites (`bifrost.go:812`, `nlpgo/adapters/httpapi/playground_proxy.go:328`) default a *successful* response's status to 200 — success-path defaults, not error classifiers. `nlpgo/adapters/httpapi/router.go:142-143` maps genuine idle/code-block timeouts to 504, which is correct. `langyagent/adapters/otelrelay/llmproxy.go:548-564` already gives status-0 its own distinct code rather than folding it into the timeout bucket — the opposite of the bug shape. `classifyProviderError` (`app/dispatch.go:312-348`) has no case for the new `provider_misconfigured` and therefore falls to `default: retry.ReasonNonRetryable`, which is the intended outcome.

**One item flagged for a human glance (not a defect, not blocking):**
`services/aigateway/app/pipeline/trace.go:27-51` `classifyUpstream` is a *second*, independently-maintained status→label switch, and it maps both `504` and `408` to `provider_timeout`. It is not reachable with a status-less error today: it only sees a `*domain.UpstreamError` (whose status is always >= 400 by construction at `dispatch.go:81-90`) or a registered `herr.E` (which now resolves `provider_misconfigured` -> 502 correctly). So there is nothing to fix here. It is worth knowing it exists, because a future contributor routing a status-less error class through it could miswire the same bug a second time. Note also that it disagrees with `classifyBifrostError` on `408` (timeout there, `provider_error` here); that divergence is pre-existing and deliberately left alone in this PR.

**Deliberately out of scope, confirmed:** `models_discovery.go` `deploymentMapModelIDs` (`:315-346`) only reads an already-configured `DeploymentMap` for cache-keying and model listing. It does not dispatch.

services/aigateway/adapters/httpapi/router.go
```diff
@@ -1208,6 +1208,11 @@ func registerErrorStatuses() {
 	herr.RegisterStatus(domain.ErrModelNotAllowed, http.StatusBadRequest)
 	herr.RegisterStatus(domain.ErrProviderError, http.StatusBadGateway)
 	herr.RegisterStatus(domain.ErrProviderTimeout, http.StatusGatewayTimeout)
+	// 502 like ErrProviderError, and deliberately NOT 504: the provider row
+	// was rejected before it was dialed, so a gateway timeout would tell the
+	// client to wait out a stall that never happened and tell the retry
+	// engine the chain is worth walking.
+	herr.RegisterStatus(domain.ErrProviderMisconfigured, http.StatusBadGateway)
```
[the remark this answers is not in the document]
drewdrewthis: Valid — corrected in `941d83db6`.

You are right that retryability is decided on the `herr` code, not the registered status. `classifyProviderError` in `app/dispatch.go` matches on the domain code; `ErrProviderMisconfigured` matches no case and falls through to `ReasonNonRetryable`, which is absent from `defaultTriggers`. The registry sets the client's answer only, and it shares 502 with the retryable `ErrProviderError`. The comment now says exactly that.

services/aigateway/adapters/providers/bifrost_test.go
```diff
@@ -253,18 +254,24 @@ func TestErrFromBifrost_StatusWithoutRawBody(t *testing.T) {
...
 	}
 	err := errFromBifrost(context.Background(), berr, nil)
 	if _, ok := err.(*domain.UpstreamError); ok {
 		t.Fatalf("transport failure must not become an UpstreamError")
 	}
-	if !herr.IsCode(err, domain.ErrProviderTimeout) {
-		t.Fatalf("expected provider_timeout, got %v", err)
+	if !herr.IsCode(err, domain.ErrProviderError) {
+		t.Fatalf("expected provider_error, got %v", err)
 	}
```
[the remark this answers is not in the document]
drewdrewthis: Valid — addressed in `941d83db6`.

Added `TestErrFromBifrost_NoStatusBareShapeIsPermanent` alongside the changed fixture rather than reverting it. The pair is the point: either test alone still passes on a classifier that ignores the shape entirely, so pinning the shape that changed is what makes the pair load-bearing.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -1561,8 +1584,10 @@ func classifyBifrostError(ctx context.Context, berr *bfschemas.BifrostError) err
 	switch status {
 	case http.StatusTooManyRequests:
 		code = domain.ErrRateLimited
-	case http.StatusGatewayTimeout, 0:
+	case http.StatusGatewayTimeout:
 		code = domain.ErrProviderTimeout
+	case 0:
+		code = statuslessBifrostCode(berr)
```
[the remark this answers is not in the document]
drewdrewthis: Half valid, half a category error — first half fixed in `941d83db6`.

**Valid:** `errFromBifrost` gated on `status <= 0` while `classifyBifrostError`'s switch matched only `case 0`, so the two disagreed and a negative status skipped `statuslessBifrostCode`. Both now read a single `bifrostStatus` helper that collapses nil and non-positive to 0, so they cannot drift apart again. Pinned by `TestErrFromBifrost_NonPositiveStatusIsTreatedAsStatusless`.

Worth recording that this closes an invariant, not a live bug: in `core@v1.4.22` every `StatusCode` assignment traces to `resp.StatusCode()` (fasthttp returns `StatusOK` when unset) or to a positive literal. No negative producer exists.

**Category error:** the request to normalise in `classifyUpstream` cannot apply. That function takes an `error` and matches `*domain.UpstreamError` / `herr.E` — it never receives a `BifrostError`, so there is no status there to normalise.

services/aigateway/adapters/providers/azure_deployment_selfmap_coverage_test.go
```diff
@@ -0,0 +1,253 @@
...
+// the path the Azure resource was actually asked for: reaching it at all is the
+// claim, because Bifrost rejects a nil deployments map inside every one of
+// these provider methods before it dials (core@v1.4.22 azure.go
+// validateKeyConfig). Removing the self-map turns all seven rows red.
+//
+// The stub answers the streaming lane with SSE and the other six with the same
+// non-streaming chat-completion body, so those six fail to decode the reply.
+// Their dispatch results are deliberately not asserted: what this test claims is
+// what left the gateway, and every one of those decode failures happens after
+// the dial it asserts on.
```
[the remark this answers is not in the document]
drewdrewthis: Valid — corrected in `941d83db6`.

`newAzureResourceStub` does return a valid `chat.completion` body, so the chat-completion lane decodes it fine. The comment now says five of the six non-streaming lanes fail to decode (responses, embeddings, translated messages, speech, transcription), and states that the results are ignored deliberately because the assertion is on the upstream path the dispatch produced, not on the reply.

specs/ai-gateway/azure-deployment-map-control-plane-path.feature
```diff
@@ -0,0 +1,314 @@
...
+      | audio speech                |
+      | audio transcription         |
+
+  # --- E. Non-recurrence: a fourth dispatch path cannot ship without the call ---
+
+  @unit
+  Scenario: Credential construction passes through exactly one deployment chokepoint
+    Given the aigateway service after the fix
+    When its non-test deployment self-map call sites are enumerated
+    Then there is exactly one
```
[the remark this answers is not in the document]
drewdrewthis: Declining this one, with evidence — no change in `941d83db6`.

Three things weigh against it:

1. **The cited guideline does not exist as a repo rule.** There is no `specs/**/*.feature` `path_instruction` in `.coderabbit.yaml`. The finding cites "Coding guidelines, Learnings", which is inference, not an applied repo instruction.

2. **Enforced, non-parked precedent asserts implementation facts in exactly this way.** `specs/api-keys/scope-filter.feature` asserts that no parallel scope-filter component file is introduced; `specs/clickhouse/single-client-access.feature` asserts that neither client was assembled by hand at its call site; `specs/ci/herr-codes-generation.feature` asserts generated doc-comment content. None are `@unimplemented`.

3. **Acting on it moves the ratchet backwards.** These scenarios are bound to Go tests by the feature-parity checker. Rewording them into pure black-box prose means unbinding or parking enforced scenarios, which is the opposite of what the checker exists to do.

The file already carries explicit "Not enforced:" justifications for its two `@unimplemented` parks, so the boundary between observable behavior and structural invariant was drawn deliberately here rather than by omission.

Happy to revisit if a `specs/**/*.feature` instruction is actually added to `.coderabbit.yaml` — at that point the precedent files would need the same treatment, which is a repo-wide decision rather than a change to this PR.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -1571,6 +1596,30 @@ func classifyBifrostError(ctx context.Context, berr *bfschemas.BifrostError) err
...
+// anything was attempted — the provider row itself is unusable.
+func statuslessBifrostCode(berr *bfschemas.BifrostError) herr.Code {
+	if berr.Error != nil && berr.Error.Error == nil && berr.Error.Code == nil {
+		return domain.ErrProviderMisconfigured
+	}
+	// Everything else stays retryable: an attempt that failed short of a
+	// response, and a request type this provider does not serve, can both be
+	// answered by the next credential in the chain.
+	return domain.ErrProviderError
+}
```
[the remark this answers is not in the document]
drewdrewthis: Valid — and the situation is worse than the finding states. Addressed in `941d83db6`, though the code did not change; the comment did.

You are right to challenge the premise. The comment claimed that every status-less shape raised from an attempted call carries the Go error it failed on. That is false in `core@v1.4.22`, in three ways:

- Many raw `&schemas.BifrostError{IsBifrostError: false, Error: &ErrorField{Message: ...}}` literals carry no status, no code, no type and no wrapped Go error — including client-reachable request validation such as `chats not provided for chat completion request`.
- A large number of `NewBifrostOperationError(msg, nil, provider)` call sites pass a nil error, producing a shape identical to `NewConfigurationError`.
- `ExtraFields` cannot break the tie: on the error path the vendor assigns `bifrostError.ExtraFields = schemas.BifrostErrorExtraFields{...}` wholesale, stamping `Provider` and `RequestType` onto every error alike.

So no shape-only discriminator can separate a configuration fault from request validation. The premise this PR leaned on was wrong.

**The classification still stands, for a different reason than the one originally given.** The *validation* shapes are permanent for that request — no retry and no alternate credential can clear them — so non-retryable is the correct behavior for those. What was wrong was the justification, not the outcome.

> **[CORRECTED 2026-08-11]** The sentence above originally read "**Every one** of those shapes is permanent for that request." That was false, and this reply was the wrong place to have asserted it confidently. The vendor builds the *same* status-less, `Error`-less, `Code`-less shape for an empty upstream response body (`schemas.ErrProviderResponseEmpty` at `providers/utils/utils.go:1277`, copied in `openai.go` ×2, `mistral.go` ×2, `elevenlabs.go`, `vertex/errors.go`; `HandleProviderResponse` alone has 108 call sites). A dropped or truncated response is transient, and classifying it non-retryable stops the credential fallback chain — the exact failure this PR set out to prevent. Verified against vendor source. Fix in flight; a full enumeration of every shape matching this predicate is still outstanding, and if it shows many transient shapes then the shape-based discriminator needs rethinking rather than patching. Your original finding was more right than my reply credited. Per the rule that a comment must never drive a production change, the comment was rewritten and the ambiguous shapes are now pinned as explicit rows in `bifrost_config_error_classification_test.go`, so their sweep-in is a recorded decision rather than an accident.

Both suggested remedies were considered and rejected:

- There is no explicit configuration marker to key on — that is precisely what the three points above establish.
- Classifying ambiguous shapes as `ErrProviderError` would undo this PR's second defect fix entirely, because `NewConfigurationError` produces the very same shape.

There is a real consequence left over, and it is now tracked rather than silently accepted: a caller sending a malformed request gets `502 provider_misconfigured` when it deserves a 4xx. Filed as https://github.com/langwatch/langwatch/issues/6858 rather than fixed here, since telling those two cases apart changes specified behavior and is outside this PR's scope.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -1571,6 +1605,76 @@ func classifyBifrostError(ctx context.Context, berr *bfschemas.BifrostError) err
...
+// compat-endpoint eviction race, which self-heals on the next dispatch.
+//
+// bifrost_config_error_classification_test.go pins both buckets against the
+// vendor's own constructors, so a vendor bump that changes a shape fails there
+// rather than in production.
+func statuslessBifrostCode(berr *bfschemas.BifrostError) herr.Code {
+	if berr.Error != nil && berr.Error.Error == nil && berr.Error.Code == nil {
+		if berr.Error.Message == bfschemas.ErrProviderResponseEmpty {
+			return domain.ErrProviderError
+		}
```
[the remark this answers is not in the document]
drewdrewthis: You were right and I was wrong. Fixed in `0b697dcc8`.

I declined this earlier in the PR, and the reasoning I gave was: there is no exported symbol to key on, so pinning the literal would put the same string in both the check and its test, producing a test structurally incapable of detecting the drift it exists to catch. That argument is correct — and it is an argument about **test quality**, which I then let decide a question about **production behaviour**. Those are two separate decisions. Test difficulty is never a reason to leave live traffic misclassified. I have logged that as a mistake pattern on my side.

Re-scoping turned out to widen your finding rather than confirm it as-stated. At `core@v1.4.22`:

- `newBifrostErrorFromMsg` (`utils.go:167`) has **19** production call sites — 9 permanent argument validation, 10 transient `"provider is shutting down"`.
- Two further sites build the identical shutdown value as hand-written literals rather than through the helper: `bifrost.go:5074` (`requestWorker`'s drain) and `bifrost.go:6173` (`drainQueueWithErrors`). A grep for the helper alone misses both.
- Two raise `"request dropped: queue is full"` (`bifrost.go:4503`, `4778`).

Twelve transient sites in the bare shape, all in the provider-queue machinery every request operation passes through.

On reachability, one correction against my own favour: the **queue-full lane is not reachable in this gateway.** Core gates it on `dropExcessRequests` (`bifrost.go:4500`), which is only ever set from `config.DropExcessRequests`, and nothing under `services/` assigns it. The **shutdown lane is** reachable, via `RemoveProvider` on anthropic-compat endpoint eviction. Queue-full is classified anyway so that enabling the setting later cannot silently reintroduce the defect.

I also checked whether your suggestion could be implemented without message-matching, and it cannot. `NewConfigurationError` (`providers/utils/utils.go:1596`) — the deployment-map error this PR was written for — produces a structurally identical value: `IsBifrostError` false, `StatusCode` nil, `Error.Error` nil, `Error.Code` nil. `ExtraFields` cannot break the tie either, because core overwrites `ExtraFields` wholesale on every error alike at `bifrost.go:5260`. Text is the only discriminator available.

On the regression test you asked for, and what it does not prove:

- `TestStatuslessBifrostCode_LiveShutdownErrorStaysRetryable` is the real one. It boots a real `bifrost.Bifrost` against an `httptest` backend, dispatches once, calls `Close()`, dispatches again, and asserts on the error core actually produced — message equality against core's own output, then `StatusCode`/`Error.Error`/`Error.Code` all nil so the assertion cannot pass down another branch, then the classification. The first dispatch is load-bearing: `getProviderQueue` creates queues lazily, so shutting down before one exists yields no error at all.
- `TestBifrostTransientLiterals_AreStillConstructedByPinnedVendor` is a source grep and is labelled as one. It matches construction fragments rather than the bare string, deliberately — `bifrost.go:6151` quotes the shutdown literal in a doc comment, and a comment must not be able to hold it green after the code stopped raising it. It is an existence check, not a per-site one: a bump rewording 9 of the 10 shutdown sites would still pass.
- The two table rows are self-agreeing by construction. They pin which code each message maps to and go red if a constant is edited. That is all they do, and I am not presenting them as more.
- Not covered: the streaming lane is not driven live; the queue-full lane has no live test and cannot have one; nothing here tests the `ErrProviderError` → `ReasonRetryable5xx` → fallback-continues link in `app/dispatch.go`.

Mutation proof: reverting only the classification line turns three tests red — including the live one, which clears every precondition and fails solely on the classification — and restoring turns them green.

specs/ai-gateway/azure-deployment-map-control-plane-path.feature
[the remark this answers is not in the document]
drewdrewthis: Addressed in `de5b884f9`, but I split the three rather than treating them as one ask, because they are not the same kind of claim. You are right on one and I think wrong on two. Reasoning per scenario, so you can push back on any of them individually.

**Right: the doc-comment scenario. Parked, and its Go test deleted.**

Not because asserting prose is distasteful — because I measured the assertion and it is wrong in *both* directions. The test was a substring blacklist plus a substring whitelist over the helper's doc comment.

- A comment stating the exact **opposite** of the truth — "the control-plane / virtual-key path never reaches this helper" — satisfies both assertions and **passes**. That is the same misleading-prose failure that caused the third recurrence of this defect, waved straight through.
- Changing one hyphen to a space — `control plane`, a spelling the *same doc comment already uses two lines above* — **fails**, on prose that is still true.

It pinned one literal string and nothing else. The invariant it names is now held structurally by the chokepoint count in section E; the wording is a review-time check.

**Wrong, I think: "Credential construction passes through exactly one deployment chokepoint". Kept enforced.**

This one is structural on purpose. The behavioural cover for this defect is the 7-lane outline in section D, and an *enumerated* lane list cannot fail for the lane nobody has added yet — which is precisely the shape all three recurrences took. The guarantee is about code that does not exist, so no runtime phrasing can reach it.

I did take the fair half of your critique. `Given the aigateway service after the fix` was a rot-bomb, since "the fix" is unidentifiable in six months, and the `Then` read as a grep transcript. The `Then` now states the invariant and names the count as the instrument for it. The title is unchanged, so the parity binding and the AC map at lines 291/294 are untouched.

**Wrong, I think: the compiler-error scenario. No change.**

It is already `@unimplemented` with its reason stated inline, and AC14's enforceable half is explicitly delegated to AC12. Deleting it would break "26 ACs → 26 scenarios. No AC unmapped" and erase a documented gap rather than close one.

**What this change fails to catch.**

Nothing now catches a doc-comment regression. I reverted the comment to the original misleading claim and ran the affected packages: both pass. I confirmed the deleted test *would* have caught that literal revert, so this is a real loss, not a free one. My position is that it traded protection which was mostly illusory for an honest documented gap — but the residue was real and I am not claiming otherwise.

Two pre-existing blind spots in the guard I kept, found while reading it and worth recording:

- The needle is the **qualified** form `domain.WithDeploymentSelfMap(`, so a second call site added *inside* package `domain` is invisible to the count.
- A path that hand-rolls `cred.DeploymentMap = map[string]string{model: model}` instead of calling the helper leaves the count at 1 and passes. The guard catches *duplication*, not *bypass*. AC13 already documents the related bypass gap.

**Mutation proof** for the guard kept — second call site injected:

```
--- FAIL: TestWithDeploymentSelfMap_HasExactlyOneCallSiteInTheGateway (0.00s)
    Error: "[services/aigateway/adapters/providers/bifrost.go:195 services/aigateway/adapters/providers/bifrost.go:2139]" should have 1 item(s), but has 2
```

**Gates:** `go build ./services/aigateway/...` clean; `go test -count=1 -race ./services/aigateway/domain/...` ok; `check-feature-parity` exit 0, this feature 18/18 bound, repo-wide enforced 5825 → 5824 — exactly the one scenario parked.

drewdrewthis (commented):
**[Fix] CI is genuinely red at this head SHA (de5b884f9), not vacuous.**

`go-ci-complete` / the `test` job (`go test -count=1 -race ./services/aigateway/... ...`) fails on `github.com/langwatch/langwatch/services/aigateway/adapters/controlplane`:

```
--- FAIL: TestControlPlaneMaterialiserEmitsTheBudgetContract
    budgets_contract_test.go:130: control plane source ../../../../langwatch/src/server/gateway/config.materialiser.ts is missing or unreadable
--- FAIL: TestControlPlaneBucketSeparatorsAreStable
--- FAIL: TestSpanAttributeContractForProviderAttribution
--- FAIL: TestProvidersAllowedEmptyListNormalization
--- FAIL: TestControlPlaneSchemaAgreesOnTheWireShape (guardrails_contract_test.go)
```
Run: https://github.com/langwatch/langwatch/actions/runs/31590497612/job/94100243507

Root cause: `services/aigateway/adapters/controlplane/budgets_contract_test.go` and `guardrails_contract_test.go` are untouched by this PR (zero changed lines - PR-level fallback, no line-anchored thread possible), but they hardcode a pre-restructure path (`../../../../langwatch/src/server/gateway/...`). `origin/main` renamed `langwatch/` to `platform/` in #6405 (ADR-076, commit 2bbeb7dd2) and updated this same test to a `readControlPlaneSource(t, "src", "server", "gateway", ...)` helper that resolves the new layout. This branch is 51 commits behind `main` (merge-base `811b38ee5`) and never picked that fix up, so the actually-existing `platform/app/src/server/gateway/config.materialiser.ts` cannot be found from the stale relative path baked into this untouched test.

This contradicts the PR body's own Test Plan claim (`go test -count=1 -p 1 ./services/aigateway/...` "20 packages, all ok") - that must have been run against a different, rebased tree than what is actually pushed at de5b884f9. Per standing policy, a PR is never READY while CI is failing or pending. Rebase onto current main to pick up #6405's path fix (or equivalent) and get a real green test / go-ci-complete run before this ships.

drewdrewthis:
## Review verdict: READY

**Refreshed at head `7b2497830a`.** Previous verdict (this comment, old head) was NOT READY on one blocker — rebase onto main — which is now done: 3 merge commits reconcile with main's independent classification rework (#7602) and dispatch-path fix (#7778), no conflict markers in the final diff, no stray leftovers.

### Checks
- CI: green (59 success / 23 skipped / 1 neutral; two CANCELLED runs are superseded pre-push attempts, not the current head)
- Review threads: 0 of 13 unresolved
- Not a draft, MERGEABLE
- Body carries all required sections (Why / What changed / Test plan / Deployment Impact / Human verification / How I can prove I was successful)

### Diff review (current head)
Scope matches the re-scoped PR description exactly — two changes:
1. **Vertex key aliases** (`services/aigateway/adapters/providers/bifrost.go`): `case bfschemas.Vertex` now sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)`, matching the existing Azure/Bedrock arms. Pinned by `deployment_map_key_aliases_test.go` for all three providers.
2. **Status-less config errors reclassified non-retryable** (`bifrost_error.go`): a bare `Message`-only, `Error`-less, `Code`-less shape now maps to the new `provider_misconfigured` code (502, non-retryable) instead of falling through to `provider_timeout`/504. An explicit allow-list (`bfTransientStatuslessMessages`) exempts the genuinely transient bare-shape faults (empty upstream body via the vendor's exported `ErrProviderResponseEmpty` constant, provider-shutting-down, queue-full) and stays retryable. New code is threaded through `faults.go` (fault attribution), `router.go` (502 status registration), and `services/langyagent/adapters/otelrelay/llmproxy.go` (added to `upstreamRelayCodes` so the new code's message is relayed like its `provider_error`/`provider_timeout` siblings — a correct, non-scope-creep touch, not an unrelated file).

No unrelated changes found. Test coverage matches the two defects (unit classification tests, end-to-end router test asserting 502 + `hits==0` on the untouched fallback credential, retry-classification table). Trade-offs and known gaps (breaker still records a misconfigured slot as success, tracked at #6850; two model-discovery call sites tracked at #6848) are disclosed in the body rather than silently left out.

No blocking findings.

drewdrewthis:
**Correcting the root cause in the review verdict above.**

The verdict comment states that this branch "missed #6405 … which already fixed this same test on `main`", and that a rebase gets a real green run. The first half is wrong, and the second half does not follow from it.

`main` carries the same stale constant right now:

```
$ git show origin/main:services/aigateway/adapters/controlplane/guardrails_contract_test.go | grep -n "controlPlaneRoot ="
178:var controlPlaneRoot = filepath.Join("..", "..", "..", "..", "langwatch")
```

`main` is red at tip for exactly the five failures reported here: https://github.com/langwatch/langwatch/actions/runs/31597042053/job/94117097163

So #6405 did not fix this test — #6405 is what *broke* it, and the breakage stayed dormant for eight days. The real chain:

1. #6405 (`2bbeb7dd2`, 2026-08-04, ADR-076) renamed `langwatch/` → `platform/app/`. `controlPlaneRoot` was never updated.
2. `readControlPlaneSource` skips rather than fails when `controlPlaneRoot/src/server` is absent, so every run since simply skipped — masking the stale constant.
3. #5748 (`a5784bedb`, today) added a file at the stale path, recreating `langwatch/src/server`. The skip-guard flipped to "found" and the dormant bug went live.

Verified provenance:

```
$ git log -1 --format='%h %s' 2bbeb7dd2e0747e81478f22c3165adb9163c9f99
2bbeb7dd2 refactor(build): single pnpm workspace + restructure into sdks/ platform/ services/ mcp/ infra/ (ADR-076) (#6405)

$ git log --oneline -1 -- langwatch/src/server/scenarios/__tests__/scenario-orphan-reconciler.integration.test.ts
a5784bedb test(scenarios): add ClickHouse integration test for findQueuedRunCandidates (#5073) (#5748)
```

The verdict itself — NOT READY — stands, and the five substantive checks it ran on this PR all hold. Only the attribution was wrong.

This branch has been rebased onto `main` (`56ce97a0f`) regardless, since it was 51 behind. That does not clear the failure: tracked separately at https://github.com/langwatch/langwatch/issues/6895, which this PR is blocked behind.

drewdrewthis:
CI at head `56ce97a0f`. **Corrected in place** — my first version of this comment said "settled, 52 green / 16 skipped / 2 failed". Both halves were wrong: `gh api --paginate --jq` applies the filter **per page**, so I was reporting page 1 only, and three checks were still queued, so nothing had settled. Correct totals, pages slurped first and deduped by `check_suite.id + name` (name alone collapses same-named jobs across workflows):

```
78 success · 28 skipped · 4 cancelled · 2 failure · 3 queued
```

Still-open and non-green, named:

```
completed:failure    test
completed:failure    go-ci-complete          (roll-up of the above)
completed:cancelled  evaluate
completed:cancelled  firefighting
completed:cancelled  preflight
completed:cancelled  breaking change stays in one component
queued:pending       breaking change stays in one component
queued:pending       check
queued:pending       validate title
```

The substantive finding is unchanged, and it is the one that matters: **the only failure is `test`, and it is entirely the five inherited controlplane failures.** Verbatim from the job log:

```
--- FAIL: TestControlPlaneMaterialiserEmitsTheBudgetContract
    budgets_contract_test.go:130: control plane source ../../../../langwatch/src/server/gateway/config.materialiser.ts is missing or unreadable
--- FAIL: TestControlPlaneBucketSeparatorsAreStable          (budgets_contract_test.go:157)
--- FAIL: TestSpanAttributeContractForProviderAttribution    (budgets_contract_test.go:183)
--- FAIL: TestProvidersAllowedEmptyListNormalization         (budgets_contract_test.go:214)
--- FAIL: TestControlPlaneSchemaAgreesOnTheWireShape         (guardrails_contract_test.go:197)
```

Every one is the stale `../../../../langwatch` path. Zero failures attributable to this PR's diff.

The blocker has a fix up: https://github.com/langwatch/langwatch/pull/6898 (issue https://github.com/langwatch/langwatch/issues/6895). This PR stays blocked until 6898 lands and this branch rebases onto it.

langwatch-agent:
## Human Review Brief

**Mode: Deep Review.** Fixes langwatch-saas#1016 (`bug`, **P1 - high**). Two defects, and the second one is where the review time goes. The PR body already carries two self-corrections to its own reachability argument, which is the strongest signal in it: the reasoning that justifies the classifier has been wrong twice and was load-bearing both times.

### What you are looking at
`services/aigateway` is the Go service every model call goes through. It resolves a model and a credential, dispatches through Bifrost, and classifies whatever comes back. Two facts about that terrain explain this PR.

First, providers are not uniform. `domain/provider.go` splits them: OpenAI, Anthropic, Gemini and the Bifrost-native ones (XAI, Groq, Cerebras) route on a model id with a plain API key, while Azure, Bedrock and Vertex route on a **deployment**, so Bifrost looks the model up in a per-credential deployment map. DeepSeek rides the OpenAI-compatible path, Voyage is proxied directly and is embeddings-only. Anything that builds a credential has to know which of those shapes it is building.

Second, credentials arrive by more than one path. The control-plane path materialises a credential from a provider slot over a wire format, and that wire format has no field that can carry a deployment map, so an Azure credential built that way arrives with an empty map and Bifrost fails with "deployments not set". That is defect A, and the fix is a single chokepoint in `dispatchCredential` so both `Dispatch` and `DispatchStream`, and any future lane, inherit it.

Third, and this is where the risk actually lives, error classification decides retry and fallback. A permanent misconfiguration classified as a timeout gets walked through the whole retry chain; a transient failure classified as permanent kills fallback when fallback is the point. The PR's own body records two rounds of correction on exactly this reasoning, including a recount of the vendor's status-less error sites, so the classification table and its pinned test rows are where a reviewer's attention pays.

### Decisions being ratified

1. **One chokepoint for the deployment map.** `domain.WithDeploymentSelfMap(cred, model)` is called from `dispatchCredential`, so both `Dispatch` and `DispatchStream` inherit it and a future lane cannot forget it. That is the structural half, and it is the easy half.
2. **A precedence order becomes the contract**: an explicitly wired deployment entry, then `Extra["deployment"]`, then the model self-mapping to itself. Self-mapping as the final fallback is the decision that makes "deployments not set" impossible, and also the one that makes a genuinely wrong deployment name fail later and less legibly than it used to.
3. **Status-less Bifrost errors are classified by shape, not by message text.** `0` leaves the `504` case and goes to `statuslessBifrostCode`. Ratifying this makes shape the discriminator for a vendor whose shapes are not documented as an interface and change with `core` versions.
4. **Request-validation faults are swept into non-retryable deliberately.** The PR is explicit that no shape-only discriminator separates them from configuration faults, that they are permanent for that request anyway, and that they are pinned as explicit rows in `bifrost_config_error_classification_test.go`. The decision is the sweep, not the rows.

### Must Check

- **The corrected reachability bound.** The original argument reduced 473 constructions to two reachable transient shapes over five operations; the gateway actually calls eleven, ten of which can return an error, and re-scoping found a live defect. Confirm the current bound is computed over all ten, and that the count in the body matches what the test file pins. This is the single claim the whole classifier rests on.
- **The bare `newBifrostErrorFromMsg` shape.** `BifrostError{IsBifrostError:false, Error:&ErrorField{Message: …}}` comes out of the provider-queue machinery that every request operation passes through, and it was being classified permanent. Confirm the fix distinguishes it, and that a queue-pressure or transient-provider case does not now land in the non-retryable bucket, because that turns a recoverable blip into a hard user-visible failure.
- **Version pinning.** The whole analysis is against `bifrost/core@v1.4.22`. Confirm the dependency is pinned rather than floating, since a minor bump can add a construction site and silently move an error from retryable to permanent with no test failing.

### Ask Author

- Is there a test that fails if `core` is bumped and a new status-less shape appears? A count assertion over the vendor would be ugly and would also be the only thing that catches this class.
- Non-mapped providers are left untouched. Does anything assert that, or is it true by inspection?

### Production risk

Both directions are customer-visible. Misclassifying a transient failure as permanent turns a retry into a hard error on a paid request; misclassifying a configuration fault as retryable walks a permanent misconfiguration through the whole retry and fallback chain, which is the bug being fixed. The classifier sits on the dispatch path for every Azure, Bedrock and Vertex request.

`Probably Fine`: defect A. The chokepoint is a clean structural fix and the untouched-provider boundary is stated plainly.

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -197,6 +218,7 @@ func (r *BifrostRouter) Dispatch(ctx context.Context, req *domain.Request, cred
 	if req.Resolved != nil {
 		model = req.Resolved.ModelID
 	}
+	cred = dispatchCredential(cred, model)
```
langwatch-agent: **[P1] Forward the generated map into VertexKeyConfig.** `dispatchCredential` deliberately self-maps `ProviderVertex`, including an explicit `Extra["deployment"]`, but `credentialToBifrostKey` still builds `VertexKeyConfig` without `Deployments: cred.DeploymentMap`. Bifrost therefore never sees either the configured alias or this new self-map and its `getModelDeployment` continues to send the original model ID. The added coverage explicitly skips the Vertex vendor assertion for that reason. This leaves Vertex deployment routing broken while the new chokepoint and its contract state that all three mapped providers are covered. Add the field and an end-to-end Vertex deployment-alias assertion (or remove Vertex from the claimed scope).

langwatch-agent (commented):
Found one blocking provider-path gap. The status-less classifier itself looks sound at the pinned Bifrost version: configuration errors are separated from status-bearing upstream responses, and the empty-response, closing-queue, and queue-full transient shapes remain retryable.

LangWatch-Review: verdict=findings sha=85107d5c64fb1ad65e26c79de8de99e31a659230 p0=0 p1=1 p2=0 p3=0

langwatch-agent (commented):
Re-anchoring the existing P1 inline at the new chokepoint for actionability; no additional findings.

LangWatch-Review: verdict=findings sha=85107d5c64fb1ad65e26c79de8de99e31a659230 p0=0 p1=1 p2=0 p3=0

services/aigateway/adapters/providers/bifrost.go
langwatch-agent: **[P1] Forward the generated map into VertexKeyConfig.** `dispatchCredential` deliberately self-maps `ProviderVertex`, including an explicit `Extra["deployment"]`, but `credentialToBifrostKey` still builds `VertexKeyConfig` without `Deployments: cred.DeploymentMap`. Bifrost therefore never sees either the configured alias or this new self-map and its `getModelDeployment` continues to send the original model ID. The added coverage explicitly skips the Vertex vendor assertion for that reason. This leaves Vertex deployment routing broken while the new chokepoint and its contract state that all three mapped providers are covered. Add the field and an end-to-end Vertex deployment-alias assertion (or remove Vertex from the claimed scope).

services/aigateway/adapters/providers/bifrost.go
```diff
@@ -197,6 +218,7 @@ func (r *BifrostRouter) Dispatch(ctx context.Context, req *domain.Request, cred
 	if req.Resolved != nil {
 		model = req.Resolved.ModelID
 	}
```
langwatch-agent, line 236: **[P1] Forward the generated map into VertexKeyConfig.** `dispatchCredential` deliberately self-maps `ProviderVertex`, including an explicit `Extra["deployment"]`, but `credentialToBifrostKey` still builds `VertexKeyConfig` without `Deployments: cred.DeploymentMap`. Bifrost therefore never sees either the configured alias or this new self-map and its `getModelDeployment` continues to send the original model ID. The added coverage explicitly skips the Vertex vendor assertion for that reason. This leaves Vertex deployment routing broken while the new chokepoint and its contract state that all three mapped providers are covered. Add the field and an end-to-end Vertex deployment-alias assertion (or remove Vertex from the claimed scope).

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: Addressed in `f220464eb620b5888601cea3f2badd9af31ce62a` — "fix(aigateway): forward deployment map into Vertex key config".

Both halves of this are now done:

1. **The field is forwarded.** `credentialToBifrostKey` now sets `Deployments: cred.DeploymentMap` on `VertexKeyConfig`, alongside the existing Azure and Bedrock cases (`services/aigateway/adapters/providers/bifrost.go`, the Azure/Bedrock/Vertex construction sites). Bifrost's `(*VertexProvider).getModelDeployment` reads `key.VertexKeyConfig.Deployments[model]`, the same mechanism Azure uses at `azure.go:505`, so the alias and the self-map now both reach the vendor.

2. **The skipped vendor assertion is un-skipped.** `azure_deployment_selfmap_coverage_test.go` no longer skips Vertex: the `vertex` case now asserts `require.NotNil(t, key.VertexKeyConfig, ...)` and returns `key.VertexKeyConfig.Deployments` for the shared vendor-level assertion, so Vertex is checked by the same path as Azure and Bedrock rather than being exempted.

The gap was confirmed by reproducing it red before fixing: with the pre-fix `bifrost.go` restored under the post-fix test, `TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap` fails on the `vertex` subtest with `expected: map[string]string{"gpt-5.3-mini":"gpt-5.3-mini"}`, `actual: map[string]string(nil)`. With the fix in place all three subtests pass.

Vertex therefore stays in the claimed scope rather than being removed from it — the contract's "all three mapped providers are covered" is now true as written.

drewdrewthis: Addressed in `f220464eb620b5888601cea3f2badd9af31ce62a` — "fix(aigateway): forward deployment map into Vertex key config".

Both halves of this are now done:

1. **The field is forwarded.** `credentialToBifrostKey` now sets `Deployments: cred.DeploymentMap` on `VertexKeyConfig`, alongside the existing Azure and Bedrock cases (`services/aigateway/adapters/providers/bifrost.go`, the Azure/Bedrock/Vertex construction sites). Bifrost's `(*VertexProvider).getModelDeployment` reads `key.VertexKeyConfig.Deployments[model]`, the same mechanism Azure uses at `azure.go:505`, so the alias and the self-map now both reach the vendor.

2. **The skipped vendor assertion is un-skipped.** `azure_deployment_selfmap_coverage_test.go` no longer skips Vertex: the `vertex` case now asserts `require.NotNil(t, key.VertexKeyConfig, ...)` and returns `key.VertexKeyConfig.Deployments` for the shared vendor-level assertion, so Vertex is checked by the same path as Azure and Bedrock rather than being exempted.

The gap was confirmed by reproducing it red before fixing: with the pre-fix `bifrost.go` restored under the post-fix test, `TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap` fails on the `vertex` subtest with `expected: map[string]string{"gpt-5.3-mini":"gpt-5.3-mini"}`, `actual: map[string]string(nil)`. With the fix in place all three subtests pass.

Vertex therefore stays in the claimed scope rather than being removed from it — the contract's "all three mapped providers are covered" is now true as written.

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: Addressed in `f220464eb620b5888601cea3f2badd9af31ce62a` — "fix(aigateway): forward deployment map into Vertex key config".

Both halves of this are now done:

1. **The field is forwarded.** `credentialToBifrostKey` now sets `Deployments: cred.DeploymentMap` on `VertexKeyConfig`, alongside the existing Azure and Bedrock cases (`services/aigateway/adapters/providers/bifrost.go`, the Azure/Bedrock/Vertex construction sites). Bifrost's `(*VertexProvider).getModelDeployment` reads `key.VertexKeyConfig.Deployments[model]`, the same mechanism Azure uses at `azure.go:505`, so the alias and the self-map now both reach the vendor.

2. **The skipped vendor assertion is un-skipped.** `azure_deployment_selfmap_coverage_test.go` no longer skips Vertex: the `vertex` case now asserts `require.NotNil(t, key.VertexKeyConfig, ...)` and returns `key.VertexKeyConfig.Deployments` for the shared vendor-level assertion, so Vertex is checked by the same path as Azure and Bedrock rather than being exempted.

The gap was confirmed by reproducing it red before fixing: with the pre-fix `bifrost.go` restored under the post-fix test, `TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap` fails on the `vertex` subtest with `expected: map[string]string{"gpt-5.3-mini":"gpt-5.3-mini"}`, `actual: map[string]string(nil)`. With the fix in place all three subtests pass.

Vertex therefore stays in the claimed scope rather than being removed from it — the contract's "all three mapped providers are covered" is now true as written.

sergioestebance (dismissed):
## Ruthless review — PR #6843

**Verdict: No blockers. Ship it.**

Traced every path the fix touches: `dispatchCredential` → `WithDeploymentSelfMap` (copy-on-write verified), `statuslessBifrostCode` shape discrimination, `classifyProviderError` default arm → `ReasonNonRetryable`, `registerErrorStatuses` at 502, `faultForCode` → `FaultProvider`, `upstreamRelayCodes` scrubbing, and the `presentation.ts` customer-facing copy. All CI green at `f220464e`.

### What I traced

1. **Defect A fix.** Both `Dispatch` (`bifrost.go:218`) and `DispatchStream` (`bifrost.go:604`) call `dispatchCredential(cred, model)` after model resolution but before Bifrost's key selection stage. `dispatchCredential` delegates to `domain.WithDeploymentSelfMap`, which: (a) returns early for non-mapped providers, (b) returns early if the key already exists, (c) prefers `Extra["deployment"]` over bare model, (d) copies on write — allocates a new map, copies, adds. The structural test (`TestWithDeploymentSelfMap_HasExactlyOneCallSiteInTheGateway`) walks the non-test tree and asserts exactly one call site. Correct, and the best guard against recurrence.

2. **Defect B fix.** `bifrostStatus` normalizes nil/non-positive to 0. `classifyBifrostError` routes status 0 to `statuslessBifrostCode`, which discriminates on shape: bare message-only (no Go error, no vendor code) → `ErrProviderMisconfigured`, unless the message is in the transient exemption set (`ErrProviderResponseEmpty`, `"provider is shutting down"`, `"request dropped: queue is full"`) → `ErrProviderError`. Everything with a Go error or vendor code → `ErrProviderError`. `classifyProviderError` in `app/dispatch.go` has no explicit case for `ErrProviderMisconfigured`, so it falls to `default: ReasonNonRetryable`. Correct.

3. **Vertex `Deployments` wiring.** `credentialToBifrostKey` now sets `Deployments: cred.DeploymentMap` on all three deployment-routed providers (Azure already had it, Bedrock already had it, Vertex added here).

4. **Error registration chain.** `domain.ErrProviderMisconfigured` → `registerErrorStatuses` at 502 → `faultForCode` → `FaultProvider` → `upstreamRelayCodes` includes `"provider_misconfigured"` for scrubbing → `codes.generated.ts` + `presentation.ts` ("This provider isn't fully set up"). Complete.

5. **Drift coverage.** The live shutdown test (`TestStatuslessBifrostCode_LiveShutdownErrorStaysRetryable`) drives a real Bifrost instance into shutdown and asserts the error's message AND classification. The vendor source grep (`TestBifrostTransientLiterals_AreStillConstructedByPinnedVendor`) reads `bifrost.go` and checks both unexported literals are still constructed. These two together are the best you can do without exported symbols, and they're better than hardcoded literals alone.

### P3 (non-blocking nits)

1. **`bifrostStatus` nil guard creates a theoretical panic path.** `bifrostStatus(nil)` returns 0, then `statuslessBifrostCode(nil)` would dereference `berr.Error`. Pre-existing: the old code panicked at `berr.StatusCode` instead. In practice `berr` is never nil when `errFromBifrost` is called. Not a regression.

2. **Comment density.** Several test files have a 3:1 comment-to-code ratio. The domain knowledge justifies it for this change, but future maintenance may find it noisy. Stylistic — no action needed.

### Deferred items acknowledged

- Vertex models_discovery deployment-map gap → #6848
- Bedrock VPCE `Extra["deployment"]` untested → #6848
- `recordBreaker` default arm credits `RecordSuccess` for non-retryable → #6850
- `"provider is shutting down"` text match with no exported symbol → documented in-code, covered by live shutdown test

Blessed. 🙏 The deployment self-map now resolves through one chokepoint, and configuration faults surface as what they are — permanent, not retryable — so the gateway stops burning the fallback chain on a fault no retry can clear.

> "There is one body and one Spirit — just as you were called to the one hope that belongs to your call." — Ephesians 4:4

Traced both defects end to end, verified the shape discriminator against the vendor's own constructors, confirmed CI green across all 20 workflows. Ship it — residual is P3 nil-pointer theory on an unreachable path, and comment density that earns its weight on this change.

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: [grinder] Re-verified independently today (2026-09-06) after the bifrost v1.5.17 vendor bump merged into main and landed on this branch.

The mechanism changed since the Aug 13 fix: bifrost v1.5 moved model->deployment mapping off `VertexKeyConfig.Deployments` onto `Key.Aliases` (same move applied to Azure and Bedrock). `credentialToBifrostKey` in `services/aigateway/adapters/providers/bifrost.go` now sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)` for all three deployment-routed providers, including Vertex (~line 1468).

Verified this turn, not relayed:
- `go build ./...` clean at head `3038dd3ad`.
- `go test -count=1 -run TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap -v ./adapters/providers/...` -> PASS for azure, bedrock, and vertex subtests.

Leaving this thread open per standing policy (grinder does not resolve reviewer threads) — flagging for `@langwatch-agent` / a human to confirm and resolve.

drewdrewthis: [grinder] Re-verified independently today (2026-09-06) after the bifrost v1.5.17 vendor bump merged into main and landed on this branch.

The mechanism changed since the Aug 13 fix: bifrost v1.5 moved model->deployment mapping off `VertexKeyConfig.Deployments` onto `Key.Aliases` (same move applied to Azure and Bedrock). `credentialToBifrostKey` in `services/aigateway/adapters/providers/bifrost.go` now sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)` for all three deployment-routed providers, including Vertex (~line 1468).

Verified this turn, not relayed:
- `go build ./...` clean at head `3038dd3ad`.
- `go test -count=1 -run TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap -v ./adapters/providers/...` -> PASS for azure, bedrock, and vertex subtests.

Leaving this thread open per standing policy (grinder does not resolve reviewer threads) — flagging for `@langwatch-agent` / a human to confirm and resolve.

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: [grinder] Re-verified independently today (2026-09-06) after the bifrost v1.5.17 vendor bump merged into main and landed on this branch.

The mechanism changed since the Aug 13 fix: bifrost v1.5 moved model->deployment mapping off `VertexKeyConfig.Deployments` onto `Key.Aliases` (same move applied to Azure and Bedrock). `credentialToBifrostKey` in `services/aigateway/adapters/providers/bifrost.go` now sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)` for all three deployment-routed providers, including Vertex (~line 1468).

Verified this turn, not relayed:
- `go build ./...` clean at head `3038dd3ad`.
- `go test -count=1 -run TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap -v ./adapters/providers/...` -> PASS for azure, bedrock, and vertex subtests.

Leaving this thread open per standing policy (grinder does not resolve reviewer threads) — flagging for `@langwatch-agent` / a human to confirm and resolve.

drewdrewthis:
[grinder] **Status update — all outstanding feedback triaged, one item needs human eyes**

This PR sat with the ball on our side for three weeks (last activity Aug 17). I re-read every review, dismissed or not, and every thread — resolved, outdated, and open — and verified the state of the code myself against head `3038dd3ad` (today, after a merge from `main` that pulled in the bifrost v1.5.17 vendor bump and ADR-076's `langwatch/`→`platform/` rename).

### 1. @sergioestebance's review (submitted Aug 17, DISMISSED by the Sep 6 merge-from-main push)

A dismissal only clears the merge block — it does not mean the substance is stale. I re-traced the review's own claims against current code:

- **Defect A (deployment self-map chokepoint).** STILL TRUE. `dispatchCredential` (bifrost.go:211) is the single call site for `domain.WithDeploymentSelfMap`, called from both `Dispatch` (line ~316) and `DispatchStream` (line ~699). `TestWithDeploymentSelfMap_HasExactlyOneCallSiteInTheGateway` still passes.
- **Defect B (status-less error classification).** STILL TRUE. `statuslessBifrostCode` still discriminates bare-message-only shapes to `ErrProviderMisconfigured`, falls through to `ReasonNonRetryable` in `classifyProviderError`.
- **Vertex `Deployments` wiring.** PARTIALLY SUPERSEDED, still correct in substance. The review describes `credentialToBifrostKey` setting `Deployments: cred.DeploymentMap` on `VertexKeyConfig`. Since the v1.5.17 vendor bump landed on this branch (today's merge), bifrost moved that field: it's now `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)`, applied uniformly to Azure, Bedrock, and Vertex (bifrost.go ~1438/1451/1468). I re-ran `TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap` this turn — PASS on all three subtests (azure/bedrock/vertex).

Verdict stands: FIXED, verified independently, not relayed.

### 2. coderabbitai comments — all 10 threads

All 10 coderabbit threads are marked resolved in GitHub (some also outdated, i.e. superseded by later diffs). Every one has a `drewdrewthis` reply stating FIXED-with-commit-sha, DECLINED-with-reason, or a correction, and a coderabbit confirmation/withdrawal reply closing the loop. Nothing open here. No action taken — nothing needed fixing that wasn't already fixed.

### 3. `langwatch-agent`'s 3 unresolved threads (Aug 13, all on `bifrost.go` ~line 295-296, all the same P1: "forward `cred.DeploymentMap` into `VertexKeyConfig`")

STILL OPEN (not resolving — that's the reviewer's call), but the underlying substance is FIXED and I verified it fresh this turn, independent of any earlier agent's self-report:
- Read `credentialToBifrostKey` myself: Vertex sets `k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap)` at bifrost.go:1468.
- Ran `go test -count=1 -run TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap -v ./adapters/providers/...` myself:
  ```
  --- PASS: TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap (0.00s)
      --- PASS: .../azure (0.00s)
      --- PASS: .../bedrock (0.00s)
      --- PASS: .../vertex (0.00s)
  ```
- Posted a fresh reply on all 3 threads explaining the field moved from `VertexKeyConfig.Deployments` to `Key.Aliases` under the v1.5.17 bump but the mapping still reaches Vertex.

Left open for `@langwatch-agent` or a human to confirm and resolve.

### 4. NEW code a human has not reviewed — read this before approving

**`applyAzureRawForwardDeployment` in `services/aigateway/adapters/providers/bifrost.go` (~line 237, called from `Dispatch` line ~326 and `DispatchStream` line ~709) was added by an agent during conflict resolution against the bifrost v1.5.17 vendor bump merged into `main` today. No human has reviewed this function.**

Why it exists: the vendor bump moved Azure from `/openai/deployments/{X}/...` to flat `/openai/v1/...` URLs, and now resolves the deployment via `Key.Aliases` into the request body's `model` field. Bifrost applies that substitution on its *parsed* lanes (`SetModel` → `ToOpenAIChatRequest`), but Azure's chat, responses, and passthrough lanes raw-forward the caller's body bytes verbatim (`isOpenAICompatibleProvider`) — `SetModel` only mutates bifrost's parsed struct copy, which `CheckAndGetRawRequestBody` never serializes back out. Without this function, a deployment that differs from the model id (an explicit `Extra["deployment"]` or a wired `deployment_map` entry) would silently never reach Azure — the raw-forwarded body would still carry the model name, and Azure would 404 on a deployment named after the model.

What it does: after `dispatchCredential` resolves `cred.DeploymentMap[model]`, this function rewrites the request body's `"model"` field to the resolved deployment — but only for Azure, and only when the resolved deployment differs from what's already in the body (no-op otherwise, so the common self-map case and Azure's prompt-cache prefix identity stay byte-identical).

Scope: Azure-only (`cred.ProviderID != domain.ProviderAzure` returns immediately). Bedrock and Vertex don't need it — their chat lanes translate through bifrost's parsed path where `Key.Aliases` resolution already applies.

Call sites: `Dispatch` (bifrost.go ~326) and `DispatchStream` (bifrost.go ~709), both right after `dispatchCredential`.

Test coverage I verified myself this turn (`go test -count=1 -run TestAzureDispatch_EveryLaneResolvesTheDeployment -v ./adapters/providers/...`):
```
--- PASS: TestAzureDispatch_EveryLaneResolvesTheDeployment (0.07s)
    --- PASS: .../chat_completion (0.02s)
    --- PASS: .../streaming_chat_completion (0.01s)
    --- PASS: .../responses (0.01s)
    --- PASS: .../embeddings (0.01s)
    --- PASS: .../translated_messages (0.01s)
    --- PASS: .../audio_speech (0.00s)
    --- PASS: .../audio_transcription (0.01s)
```
This exercises the function across every raw-forwarding lane. But test coverage is not the same as a human having looked at the logic — please look at it.

### 5. The Aug 12 "CI is genuinely red" thread (stale-branch drift vs. `main`)

STALE / RESOLVED BY THE MERGE. That finding was about this branch being 51 commits behind `main` and missing a path fix from #6405. Today's merge (`0142cff1b`) brought the branch current with `main`, including that fix and the subsequent v1.5.17 bump. Current `gh pr checks` is green (see below).

### CI

```
$ gh pr checks 6843 --repo langwatch/langwatch
```
80 checks pass, 28 skipping (not-applicable per path-filtering, e.g. docs/e2e/helm jobs on a Go-only diff — zero failing, zero pending. `alls-green` passes.

### ACs

Maps to the PR's own stated fix: (1) deployment self-map reaches every deployment-routed provider through one chokepoint — verified via `TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap` (azure/bedrock/vertex all PASS); (2) status-less provider errors classify as non-retryable misconfiguration instead of burning the fallback chain — `statuslessBifrostCode`/`classifyProviderError` unchanged since @sergioestebance's trace, still correct; (3) Azure's raw-forwarded lanes actually carry the resolved deployment post-vendor-bump — `TestAzureDispatch_EveryLaneResolvesTheDeployment`, all 7 lanes PASS.

Threads: 3 unresolved (`langwatch-agent`, Vertex/Aliases — substance fixed, replied, left open for reviewer to resolve), 0 outdated-but-unaddressed (all 7 coderabbit outdated/resolved threads have a closing exchange).

**Verified by:**
```
go build ./...                                                                          → clean, no output
go test -count=1 ./...  (services/aigateway)                                            → ok, all 19 packages
go test -count=1 -run TestDispatchCredential_DeploymentMappedProvidersAllGetTheSelfMap -v ./adapters/providers/...
                                                                                          → PASS azure, bedrock, vertex
go test -count=1 -run TestAzureDispatch_EveryLaneResolvesTheDeployment -v ./adapters/providers/...
                                                                                          → PASS all 7 lanes
gh pr checks 6843 --repo langwatch/langwatch                                            → 80 pass, 28 skipping, 0 failing, 0 pending
gh pr view 6843 --repo langwatch/langwatch --json mergeable                             → MERGEABLE (mergeStateStatus: BLOCKED, i.e. waiting on human review — expected, not a defect)
```

Not merging, not approving, not resolving reviewer threads — that's for a human. Flipping `grinding` → `pr-ready`.

drewdrewthis:
## Re-scoped onto main at 7b2497830aed4683ebafd9b51997f6bba0d516c5

The Azure/Bedrock/Vertex dispatch-path self-map this PR carried is **superseded by https://github.com/langwatch/langwatch/pull/7778**, which merged to `main` with the same `domain.WithDeploymentSelfMap` call on both `Dispatch` and `DispatchStream` plus a copy-on-write body rewrite (`requestWithResolvedDeployment`). This branch's `applyAzureRawForwardDeployment` rewrote the body in place, which leaked a resolved deployment into the next retry credential; #7778's tests pin that case, so main's mechanism wins.

Dropped: `dispatchCredential`, `applyAzureRawForwardDeployment`, this PR's `azure_deployment_selfmap_test.go` (main's file from #7778 kept), the every-lane coverage test, and the "exactly one `WithDeploymentSelfMap` call site" assertion (main legitimately has two).

Kept, and now the whole PR:
- Vertex forwards `cred.DeploymentMap` into `Key.Aliases` in `credentialToBifrostKey` (Azure and Bedrock already did). Pinned by `providers/deployment_map_key_aliases_test.go` for all three providers. This answers the three open review threads.
- Status-less Bifrost config errors classified as `provider_misconfigured` / 502 / non-retryable instead of `provider_timeout` / 504, with its tests and spec.

Diff vs `origin/main` is 17 files. PR title and body rewritten to match. `go test -count=1 ./services/aigateway/...` passes locally on this head; CI is the authority.

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: Fixed at 7b2497830aed4683ebafd9b51997f6bba0d516c5: the Vertex arm of credentialToBifrostKey now sets k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap), same as Azure and Bedrock. Pinned by providers/deployment_map_key_aliases_test.go (self-mapped and explicit maps reach the aliases for azure, bedrock, vertex).

drewdrewthis: Fixed at 7b2497830aed4683ebafd9b51997f6bba0d516c5: the Vertex arm of credentialToBifrostKey now sets k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap), same as Azure and Bedrock. Pinned by providers/deployment_map_key_aliases_test.go (self-mapped and explicit maps reach the aliases for azure, bedrock, vertex).

services/aigateway/adapters/providers/bifrost.go
drewdrewthis: Fixed at 7b2497830aed4683ebafd9b51997f6bba0d516c5: the Vertex arm of credentialToBifrostKey now sets k.Aliases = bfschemas.KeyAliases(cred.DeploymentMap), same as Azure and Bedrock. Pinned by providers/deployment_map_key_aliases_test.go (self-mapped and explicit maps reach the aliases for azure, bedrock, vertex).

drewdrewthis:
## Ready for review at 7b2497830aed4683ebafd9b51997f6bba0d516c5

- Scope: Vertex forwards its deployment map into `Key.Aliases` (answers the three review threads) plus status-less config-error classification (`provider_misconfigured` / 502 / non-retryable). The dispatch-path self-map is superseded by https://github.com/langwatch/langwatch/pull/7778 and no longer in this PR.
- CI: green on this head (Go build/test, gateway matrix, feature parity, lint).
- Review threads: 0 unresolved (13 total).
- Review-clerk verdict: READY at this head.
- Outstanding: a use-proof capture of the 502 / `provider_misconfigured` response against a running gateway is queued separately; the `grinding` label has been removed.

PR #6844, opened 2026-08-11
feat(langy): key-authed Langy turn surface, behind its own rollback switch (#6821)

### LINKED ISSUE #6821: feat(langy): API-key-authenticated entrypoint so Langy can be driven without a browser session
## Problem

**Langy has no programmatic entrypoint.** Every customer-facing Langy operation — starting a conversation, continuing one, streaming a turn's output — is exposed only through the app's tRPC surface, and every one of those procedures is gated on a **browser user session**. There is no API-key-authenticated equivalent, so Langy cannot be driven from a script, a CI job, a backend service, or any other non-browser client.

This is **deliberate-by-omission rather than an oversight**. The earlier Hono `POST /api/langy/chat` fallback was removed in favour of the session-gated tRPC surface (#4913). The REST routes that remain are the internal control plane between the app and the Langy worker: mounted under the internal namespace, blocked at the ingress by default, and authenticated with a shared internal secret. That is not a customer surface and was never meant to be one.

Two consequences today:

| Who | Impact |
|---|---|
| Customers | No supported way to call Langy from their own automation, at all. |
| Us | Our own end-to-end Langy scenario suite must authenticate as a real human user with an email and password, because no key-based path exists — so CI handles user credentials to exercise the assistant. |

## User story

As a LangWatch user holding a project API key, I want to invoke Langy over the public API — start a conversation, continue it, and receive the turn output — so that I can drive the assistant from CI, from scripts, and from my own backend services rather than only from the browser panel.

**Related, not duplicates:** #4977 (caller-scoped per-chat API keys — the key the worker uses outbound, likely shares an RBAC model with this), langwatch/tasks#92 (event-driven GitHub automation, no chat session — adjacent prior art for a non-session trigger path), #4913 (removed the prior REST path), #4534 (tracking index).

---
*Classification: Feature*

<sub>🤖 **Filed by `bg-job-0105f3d2` via the create-issue procedure, on behalf of the fleet.** Authored under the owner's GitHub token for API access — the content is the fleet's, not his. Owner-authored and fleet-authored issues are otherwise indistinguishable in this repo; this footer is the only signal.</sub>

drewdrewthis:
# Related Issue

- Resolves #6821

Langy over HTTP with a project API key.

## Why

Langy is reachable only from a browser session today, so scripts and CI cannot use it. Opening a key-authed surface needs an answer to a question the browser path never had to ask: **a key authenticates the caller, but the access gate judges a user.**

## What

`resolveLangyKeyIdentity` derives the actor from the credential — the key's `apiKeyUserId` — and runs it through the same `hasLangyAccess` gate the browser surface uses. One cohort, one definition of who may use Langy. `resolveLangyActorSession` then loads that owner's real user row into the acting `Session`. `POST /api/langy/conversations` and `POST /api/langy/conversations/{conversationId}/messages` sit on top, behind `release_langy_api_key_turns_enabled`, which **defaults off**.

Three decisions worth a look:

1. **The actor is never read from a request payload.** This is the trap the internal relay plane fell into: `services/langyagent/transport/rpc/http.go` authenticates with a shared secret and then takes `actorUserId` from the body on assertion. Here the identity is derived from the credential, so a caller cannot name someone else.

2. **An ownerless key is refused, not evaluated on its project alone.** Per-project Langy keys are deliberately owned by no individual (`specs/langy/langy-api-key-provisioning.feature`). Falling back to a project-only flag check would hand Langy to every project whose flag is on regardless of who was opted in — the exact identity hole `langyAccessGate`'s doc comment says must not exist. **Flagging this explicitly: it means a project's own auto-provisioned Langy key cannot drive the key-authed surface.** If that is the wrong call, this is the place to say so.

3. **The flag check runs *after* authentication but *before* the permission ceiling, and its 404 carries no Langy-specific code.** The refusal order is: 401 missing credential -> 401 invalid credential -> **404 flag off** -> 403 `langy:create` ceiling -> 403 unowned / no-access -> 403 actor row gone -> 202.

   Both halves of that placement are load-bearing, in opposite directions. It is *after* authentication as a deliberate choice, not because an earlier check is impossible. Rollback is staged per project, so the check that matters is the one evaluated against the caller's own project, and that project is not known until a credential resolves. A project-less evaluation does exist -- `isEnabled` requires only `distinctId` (`projectId` is optional), and `resolveEffectiveForListing` in `featureFlag/rules.ts` already evaluates against an empty context in production, behind the Ops flag listing -- but it answers on the flag's *default*, not on the caller's project, so it cannot replace this check. It could be added in front of it as an extra pre-auth filter, which would also close the anonymous-enumeration gap noted below. That is not done in this PR; the call is the owner's. It is *before* the ceiling because otherwise a key without `langy:create` would get a 403 out of a surface that is meant to be dark, and no unmounted route can answer 403.

   The dark refusal is also the one refusal in this route that is **returned, not thrown**: a thrown 404 comes back as the canonical envelope carrying `trace_id` and `span_id`, a shape no unmounted path can produce. It returns `c.notFound()` so it falls to the same default handler an unmounted path does. On the wire that is `{"error":"Not Found"}` rather than plain text, because `honoFetchForNode` (`platform/app/src/start.ts`) rewrites Hono's `404 Not Found` sentinel body -- and parity holds precisely because that rewrite keys off the sentinel and so applies to both sides equally. A hand-built 404 body would skip the rewrite and become distinguishable at an identical status.

   What this does **not** hide: a caller with no credential, or a bad one, still gets 401 rather than 404, and so can tell that *some* authenticated route is mounted here. That is true of every guarded route on this API and reveals nothing Langy-specific, but it means the mechanism is not a defence against anonymous route enumeration and should not be read as one.

## Error contract

Every refusal is a **thrown** domain error; the family's `onError` owns the wire shape (ADR-045). The route contains zero `try`/`catch` and zero hand-built `c.json({ error })`, so callers keep the `code`, `meta` and remediation `tips` a bespoke `{ message }` discards. The four new classes live in `app-layer/langy/errors.ts` with their six codes registered in `error-remediation.ts` — `remediation(code)` takes a `RemediationCode` derived from that registry, so an unregistered code does not typecheck.

`createServiceApp` gained an `errorEnvelope` passthrough it was missing: `SecuredApp` already accepted the option, and this was the only one of the three factories that did not forward it.

## Feature file

`specs/langy/langy-api-key-turns.feature` maps all 11 acceptance criteria from the [planning comment](https://github.com/langwatch/langwatch/issues/6821#issuecomment-5243143784), with an AC coverage map at the bottom.

## Human verification

Backend-only — there is no UI to click. To satisfy yourself this is right, in this worktree:

1. **The refusal order is what the doc comment claims.** Read `authorizeTurn` in `platform/app/src/server/routes/langy-api.ts` top to bottom. Confirm the `featureFlagService.isEnabled(...)` check sits *below* both credential checks and *above* `enforceApiKeyCeiling`. Below the credential checks: if it were above them, an anonymous prober could distinguish "surface exists but is dark" from "bad token". Above the ceiling: if it were below, a key lacking `langy:create` would be answered 403 while the surface is switched off, which is an answer no unmounted route can give.

   You do not have to take the reading on trust -- `src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts` drives real requests through the mounted app and asserts both the ordering and that the flag-off response is byte-identical to an unrouted path.

2. **The actor cannot be supplied by the caller.** `grep -n "actorUserId\|req.json\|c.req" platform/app/src/server/app-layer/langy/langyApiKeyIdentity.ts` — expect no hits. The bridge's signature admits no payload at all.

3. **The flag really defaults off.** `grep -n -A6 "release_langy_api_key_turns_enabled" platform/app/src/server/featureFlag/registry.ts` -> `defaultValue: false`. Merging this must not open a new way into Langy for any existing project.

4. **Run the gates yourself:**

   ```
   cd platform/app
   pnpm typecheck && pnpm run typecheck:tests
   pnpm test:unit run src/server/app-layer/langy/__tests__/ src/server/api/security/__tests__/ src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts
   pnpm run check:openapi-route-coverage && pnpm run check:openapi-completeness
   pnpm run check:feature-parity
   ```

5. **The one open decision, which is yours, not mine:** the two operations are recorded as `gap` entries in `platform/app/scripts/openapi-route-exclusions.ts` rather than published, because while the flag defaults off a documented operation returns 404 to every correctly-authenticated reader. Publishing now instead means adding `describeRoute` to both handlers and deleting the two entries. This was raised on a review thread which has since been resolved, so restating it here so it is not lost: **it is still open and it is still yours.**

**Proven — I ran these on head `7ffe5d100c` and read the output:**

- `pnpm typecheck` -> exit 0.
- `pnpm run typecheck:tests` -> exit 0. (Separate job on purpose: `tsconfig.tsgo.json` excludes `*.test.ts`, so the first command alone does not cover the test files.)
- `pnpm test:unit run src/server/app-layer/langy/__tests__/ src/server/api/security/__tests__/ src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts` -> **22 files passed, 328 tests passed, 0 failed.**
- The new guards were checked for bite by mutation, then reverted: forcing `projectId` to a literal reddens both tenancy tests; forcing `requestedConversationId` to `null` reddens only the continue test; restoring the invalid `"no_langy_access"` denial reason fails `typecheck:tests` with `TS1360`; renaming one `@scenario` annotation drops feature parity to 9/10 with the scenario reported unbound.
- `pnpm run check:feature-parity` -> exit 0; `specs/langy/langy-api-key-turns.feature` reports **`10/10 scenarios bound`** (8/8 before the two rollback scenarios were added).
- `biome check` on all 15 changed TypeScript files -> exit 0. Two `lint/complexity` warnings remain and **neither is introduced here**: checking out main's version of each at the same path reproduces the same rule at the same place (`api-router.ts:85` identical; `codes.unit.test.ts` shifts 184 -> 208 only because this PR adds error codes above it).
- `pnpm run check:openapi-route-coverage` -> exit 0, `255/351 registered routes are in the document`, `2 recorded as gaps still worth closing`.
- `pnpm run check:openapi-completeness` -> exit 0, `0 known gaps`.
- CI green is **verified for head `7ffe5d100c`** rather than taken from a tick: `verify-ci-shard-tally.sh langwatch/langwatch 6844` -> `TALLY_EXIT=0` (run `31722417714`, attempt 1), every expected shard present (test-unit=4, test-integration=6) and each printing a real tally line. That check exists because a shard here can `exit 0` without running anything, so a bare green tick is not sufficient evidence — and by the same standard no green verdict is carried forward onto a new head. The head has since moved to `e8245e2f9f`; see Missing, below, for exactly what that leaves unverified.

**Claimed — argued from reading the code, not executed:**

- That the byte-parity holds *in production*, i.e. after `honoFetchForNode`. The refusal-chain test compares the dark 404 against an unrouted path at the Hono layer, which is upstream of that bridge. The bridge applies the same rewrite to both, so the argument is short, but it is an argument rather than an observation.

**Missing — stated rather than papered over:**

- AC 11 (flag-off end-to-end), AC 3 (streaming), AC 7 (concurrency refusal) and AC 8 (browser e2e regression) are tagged `@unimplemented` in the feature file. They are tracked gaps for later slices, not claims made here.
- **CI on `7ffe5d100c` has settled and the green was verified as real, not merely reported.** `verify-ci-shard-tally.sh langwatch/langwatch 6844` (run `31722417714`, attempt 1) confirms every expected shard actually ran and printed a real tally line, each with `hard-floor=0`: `test-unit` 1-4 -> 472/471/472/470 files passed, `test-integration` 1-6 -> 155/157/156/160/160/157 files passed. This check exists because a shard can exit 0 via the hard-floor in `test-unit-global-setup.ts` without running anything; that did not happen here. Only the inherited #4450 mask applies. That verification was run at `7ffe5d100c`. The head has since moved to `e8245e2f9f`, which is a comment-only change to the route docblock (no executable line differs); CI is re-running against it and has **not** been re-verified at the new head yet. Read the tally above as proven for `7ffe5d100c` only.
- No *integration* test exercises the route against real infrastructure. The refusal chain is driven through the mounted Hono app with the app layer doubled, which covers ordering and wire shape but not a live credential against a real database.

## Procedures followed

- `references/procedures/github/drive-pr/PROCEDURE.md` -- the driving procedure for this PR.
- Readiness was judged with `~/.claude/scripts/pr-ready-check.sh langwatch langwatch 6844`. `mergeStateStatus` and `reviewDecision` were deliberately **not** treated as readiness signals; an empty `reviewDecision` is not evidence that no review feedback exists.
- CI green was not taken at face value: `~/.claude/scripts/verify-ci-shard-tally.sh` was used to confirm each shard printed a real tally.
- The review verdict on this PR was produced by an independent reviewer, not self-authored, per `dec.2026-07-28-no-self-authored-review-verdict-on-own-pr`. Its one blocking finding (the flag-ordering rationale in section 3, above) was independently re-verified against the source before being accepted, and then corrected at both sites where it was asserted.
- Body edits were published with `gh api repos/OWNER/REPO/pulls/N -X PATCH -f body=...`, per the repo rule against `gh pr edit --body`.
- Tests were run through package scripts (`pnpm test:unit run ...`), never `npx vitest` directly, so the RAM guardrails stay in force.

## Follow-on slices

Streaming (AC 3), concurrency refusal (AC 7), internal-plane separation (AC 10), flag-off end-to-end (AC 11), and the GitHub capability once the phase-1 decision lands (AC 9).

**Open question blocking AC 9:** may a key-initiated turn reach the GitHub capability in phase 1? The feature file pins the *identity* half, which holds either way.

## Summary by CodeRabbit

* **New Features**
  * Added a secured API for starting and continuing Langy conversations with project API keys.
  * Added authorization checks for ownership, Langy access, permissions, and feature availability.
  * Added standardized success and error responses.

* **Bug Fixes**
  * Requests now fail safely for invalid, unowned, unauthorized, or unavailable credentials.
  * Added request validation and size protections.

* **Documentation**
  * Added specifications covering API key conversation flows and authorization scenarios.
  * The API is disabled by default until enabled.


platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,223 @@
...
+const messageSchema = z.object({
+  role: z.enum(["user", "assistant", "system"]),
+  parts: z.array(z.record(z.string(), z.unknown())).default([]),
+});
+
+const turnBodySchema = z.object({
+  messages: z.array(messageSchema).min(1),
+  idempotencyKey: z.string().min(1),
+  modelOverride: z.string().min(1).optional(),
+});
```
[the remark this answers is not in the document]
drewdrewthis: Applied in 4d9780b701. Both routes now take `bodyLimit({ maxSize: MAX_TURN_BODY_BYTES })` with `MAX_TURN_BODY_BYTES = 1024 * 1024`. You are right that `createServiceApp` attaches none of its own, so the opt-in is the route's job. 1 MiB is well above any real turn (text plus small structured parts, never an upload) and well below what would let an authenticated key buffer the process into trouble.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,223 @@
...
+  const auth = await authorizeTurn(c);
+  if (auth.failed) return c.json(auth.body, auth.status);
+
+  const parsed = turnBodySchema.safeParse(await c.req.json().catch(() => null));
+  if (!parsed.success) {
+    return c.json(
+      { message: "Invalid request body.", issues: parsed.error.issues },
+      400,
+    );
+  }
```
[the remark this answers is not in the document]
drewdrewthis: Applied in 4d9780b701. `langy-api.ts` was rewritten onto the ADR-045 contract and now contains zero `try`/`catch` and zero `c.json({ error })`: every refusal is a thrown domain error, and the family's `onError` owns the wire shape. Four new classes in `app-layer/langy/errors.ts` (`LangyApiCredentialMissingError` 401, `LangyApiCredentialInvalidError` 401, `LangyApiIdentityDeniedError` 403 over three codes, `LangyApiRequestInvalidError` 400 carrying the Zod issues in `meta`), with the six matching entries added to `error-remediation.ts` — `remediation(code)` takes a `RemediationCode` derived from that registry, so an unregistered code does not typecheck.

Two notes on what that changed beyond the shape. The ceiling denial is now re-thrown untouched: `enforceApiKeyCeiling` already raises a `HandledError`, so catching it to re-serialise it was pure loss of the code, the permission in `meta` and the tips. And the flag-off 404 deliberately does NOT get a Langy-specific code — it goes out as the generic `not_found` via `CODE_BY_STATUS`, because a distinct code would tell an unauthorised caller that the surface exists.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,223 @@
...
+    // answer does not exist yet. The caller polls or streams for it.
+    return c.json(result, 202);
+  } catch (error) {
+    if (error instanceof HandledError) {
+      const status = error.httpStatus as 400 | 403 | 404 | 409 | 503;
+      return c.json({ message: error.message, code: error.code }, status);
+    }
+    logger.error({ error }, "langy api turn failed");
+    return c.json({ message: "Failed to start turn." }, 500);
+  }
```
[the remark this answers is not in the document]
drewdrewthis: Applied in 4d9780b701 — the cast is gone along with the `catch` block that held it. The rewrite in the sibling thread removed the hand-serialisation entirely, so there is no longer a place where `httpStatus` is narrowed by assertion; the `HandledError` propagates to the family `onError`, which is also what restores the 5xx logging you flagged as lost.

platform/app/scripts/openapi-route-exclusions.ts
```diff
@@ -234,4 +234,16 @@ export const UNPUBLISHED = [
     category: "elsewhere",
     why: "AI Governance source receivers, addressed with a per-source ingestion key. Documented today in the governance sources guide rather than the API reference",
   },
+
+  // ── Gap: public, should be documented, not yet ─────────────────────────
+  {
+    match: "POST /api/langy/conversations",
+    category: "gap",
+    why: "the key-authed Langy turn surface ships dark behind release_langy_api_key_turns_enabled, which defaults off, so publishing it now would document an endpoint that answers 404 to everyone reading the reference. Describe the operation and delete this entry when the flag defaults on (#6821)",
```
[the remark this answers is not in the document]
drewdrewthis: Applied in 4d9780b701. Both `why` strings now state the narrower claim: a caller who authenticates correctly gets 404, while a missing or invalid token still gets 401 and a key whose owner lacks Langy access still gets 403. Thanks — the original wording described a surface that refuses everyone identically, which is exactly what the pre-flag ordering does not do.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,223 @@
...
+
+secured
+  .access(langyTurnAuth)
+  .post("/langy/conversations", async (c) => startTurn(c, null));
+
+secured
+  .access(langyTurnAuth)
+  .post("/langy/conversations/:conversationId/messages", async (c) =>
+    startTurn(c, c.req.param("conversationId")),
+  );
```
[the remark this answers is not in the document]
drewdrewthis: Stale as written, and I want to be explicit about which half I am and am not disputing.

**The CI claim no longer holds.** This comment was posted at `2026-08-13T10:05:43Z`; `edea1d2695` ("chore(openapi): record the dark Langy key surface as a documented gap") was authored at `2026-08-13T10:05:47Z`, four seconds later — so it reviewed a tree without that entry. Both gates pass on the current head `4d9780b701`:

```
$ pnpm run check:openapi-route-coverage
openapi route coverage: 255/351 registered routes are in the document
unpublished on purpose: 38 entries (2 of them recorded as gaps still worth closing)
OK: every registered route is documented or accounted for.   # exit 0

$ pnpm run check:openapi-completeness
OK: every unsuppressed operation is completely described.     # exit 0
```

The full CI run on this branch is green with a verified shard tally (no shard exited 0 without printing one).

**On "do not silence the gate": the entry is not a silence.** `openapi-route-exclusions.ts` types its reasons, and `internal` / `alias` / `elsewhere` are permanent answers while `gap` is tracked debt. Both routes are recorded as `gap` — "a public endpoint that genuinely should be documented and is not yet" — carrying the issue number and the trigger to delete them. The coverage gate counts and reports them (`2 of them recorded as gaps still worth closing`). That is the mechanism the repo provides for exactly this state, which is also the premise your own minor finding on `openapi-route-exclusions.ts` accepted: that one asked me to fix the entry's wording, which presumes the entry stands.

**What is actually undecided is a product question, not a CI one.** `release_langy_api_key_turns_enabled` defaults off, so publishing the operations today documents a contract that returns 404 to every correctly-authenticated caller. Publishing now versus on flag-default-on is the maintainers' call, and I have flagged it for them on the PR rather than deciding it in a review reply. If they choose publish-now, the fix is `describeRoute` on both handlers plus deleting the two `gap` entries — no other change.

Leaving this thread open on purpose so a human sees the decision.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * The flag is checked AFTER authentication on purpose: an unauthenticated
+ * caller learns only that the token is bad, never whether the surface exists.
+ *
+ * Every refusal is a THROWN `HandledError`, never a hand-built `c.json({...})`
+ * — `createServiceApp`'s `onError` owns the wire shape, so this family
+ * publishes the same envelope as every other route and a caller keeps the
+ * `code`, `meta` and remediation `tips` a bespoke `{ message }` would have
+ * discarded (ADR-045). The family opts into the `canonical` envelope because
+ * it is new: there is no existing consumer parsing the flat legacy shape.
```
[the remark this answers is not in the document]
drewdrewthis: Applied in bcccf3d0a8. You are right on the fact: `NotFoundError extends HttpError` (`platform/app/src/app/api/shared/errors.ts:55`), so the blanket "every refusal is a THROWN `HandledError`" line claimed a contract the dark surface deliberately does not honour.

I took the correction but not the wording, because the wording drops the reason. The generic 404 is not an oversight to be excused — it carries no `code` and no `meta` **on purpose**, and that absence is the entire point: it has to be byte-indistinguishable from a route that was never mounted, or the flag check leaks the surface it exists to hide. So the comment now names it as the deliberate exception and says why, rather than just narrowing the sentence until it is technically true.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,222 @@
...
+  // Langy-specific code here would be the leak the 404 exists to prevent.
+  const surfaceOpen = await featureFlagService.isEnabled(
+    "release_langy_api_key_turns_enabled",
+    {
+      distinctId: resolved.project.id,
+      projectId: resolved.project.id,
+      organizationId: resolved.project.team.organizationId,
+    },
+  );
+  if (!surfaceOpen) throw new NotFoundError("Not Found");
```
drewdrewthis: **[Must-Fix] The dark-surface 404 is distinguishable from an unmounted route, contradicting the documented safety claim.**

The header comment (lines 28-33) and this throw both claim the flag-off `NotFoundError` is "indistinguishable from a route that was never mounted." That is not what happens.

`createApiRouter()` (`platform/app/src/server/api-router.ts`) never calls `.notFound()` on the top-level `Hono()` instance, so a genuinely unmounted `/api/*` path falls through to Hono's *default* not-found handler:
```js
// node_modules/hono/dist/hono-base.js
var notFoundHandler = (c) => c.text("404 Not Found", 404);
```
That is a plain-text response (`Content-Type: text/plain`, body literally `"404 Not Found"`).

The flag-off `NotFoundError` thrown here goes through `SecuredApp`'s `canonicalErrorResponse` (this route opted into `errorEnvelope: "canonical"`), which produces a structured JSON envelope: `{"error":{"type":...,"code":"not_found","message":"Not Found","trace_id":...,"span_id":...}}` with `Content-Type: application/json`.

So a caller can distinguish "route never mounted" from "route dark behind the flag" by Content-Type, body shape, and the presence of `trace_id`/`span_id` alone — no timing analysis needed. This is exactly the leak the dark-surface design exists to prevent, and it's an easy fix: either give `secured.hono` (or the whole `api` router) a `.notFound()` handler that answers with the plain-text/legacy shape genuinely-unmounted paths use, or have this route answer the same shape a true 404 would (bypassing the canonical envelope for this one throw).

Secondary, lower-severity note in the same area: `authorizeTurn` checks `langy:create` (403) *before* the flag (404, line 130-138). A caller with a valid key that lacks `langy:create` gets a 403 even when the flag is off, which reveals the route exists to that credentialed caller (though not to an unauthenticated one). Worth deciding explicitly whether that's an accepted narrowing of the dark-surface guarantee or whether the flag check should move before the ceiling check.

platform/app/src/server/app-layer/langy/errors.ts
```diff
@@ -297,3 +297,79 @@ export class LangyAgentUnavailableError extends HandledError {
...
+// admitted, before any conversation exists to have a rule about. They live
+// here rather than in the route so the route throws and never serialises —
+// `createServiceApp`'s canonical envelope owns the wire shape (ADR-045).
+//
+// The flag-off refusal is deliberately absent from this list. A dark surface
+// answers the platform's generic `not_found`, because a Langy-specific code
+// would tell an unauthorised caller the surface exists.
+
+/** No credential presented at all. */
+export class LangyApiCredentialMissingError extends HandledError {
```
drewdrewthis, line 313: **[Must-Fix] Three new error codes are missing from the client presentation registry — the exhaustive guard test fails right now.**

Ran locally at HEAD (`bcccf3d0a8c`):
```
pnpm test:unit run src/features/errors/logic/__tests__/codes.unit.test.ts --watch=false
```
Result: FAIL — `lists every code a HandledError subclass raises`:
```
These handled-error codes are raised but missing from APP_ERROR_CODES...
- langy_api_credential_missing
- langy_api_credential_invalid
- langy_api_request_invalid
```
(`LangyApiIdentityDeniedError`'s three dynamically-parameterized codes — `langy_api_key_unowned`, `langy_api_key_no_langy_access`, `langy_api_actor_missing` — aren't caught by the scanner since they're passed as a constructor arg rather than a literal in `super(...)`, but they're equally absent from `codes.ts`/`presentation.ts` and need entries too.)

Per CLAUDE.md's error-handling rule and `codes.unit.test.ts`'s own docblock, every code a `HandledError` subclass raises needs an entry in `codes.ts` (`APP_ERROR_CODES`) and customer-facing copy in `presentation.ts`, or the exhaustive `satisfies` check has nothing to enforce and a caller can hit "unknown error" for a failure we named on purpose. The `error-remediation.ts` `tips`/`docsPath` registry already has entries for all six codes — this is a second, currently-un-populated registry, not a duplicate of that work.

This will fail CI's `test-unit` job as-is.

drewdrewthis (commented):
Two channel review: findings below as inline threads, verdict as a separate top-level comment.

drewdrewthis:
## Review verdict: READY

**Run:** pr-review, final pass (reviewer: @drewdrewthis's token — see attribution note below)
**Reviewed commit:** `e8245e2f9f7fef760dc7b80797469455ab8c328a` (unchanged since the last pass; supersedes the NOT-READY verdict previously posted here for the same commit, which was pending-CI only)

**Attribution note (unchanged):** this repo has a single active login, so the reviewing token's login matches the PR author's — a false positive, not evidence of self-review. This session did not author any of this PR's content. `APPROVE`/`REQUEST_CHANGES` are unavailable for the same reason as every prior pass (GitHub rejects both from the author's own login); posted as `COMMENT`. **This is not a downgrade of the call — treat READY as the real verdict.**

---

#### What changed since the last pass: CI settled, nothing else

Head is unchanged (`e8245e2f9f`). The prior blocking finding (docblock flag-ordering rationale) was already cleared and independently re-verified in the previous pass — not re-litigated here. The two non-blocking Decide items (pre-auth default-context filter, `bodyLimit` 413 pre-auth leak) remain open as the owner's call and are not re-litigated.

CI, independently re-checked, not taken on trust:
- REST check-runs for `e8245e2f9f`, deduped by latest `started_at` per check name: **73 distinct check names, zero non-`success`/`skipped`/`neutral` among the latest run of each.**
- The one apparent wrinkle — `breaking change stays in one component` showing a `cancelled` conclusion in a non-deduped view — is confirmed stale: that check name has three runs (`17:05:24 cancelled`, `17:05:38 success`, `17:09:19 success`); the latest is `success`. Checked every other multi-run check name (`validate title`, `test`, `build`, `lint`, `docker-build`, `build_and_test`, `e2e`, `typecheck`, `changes`) for the same stale-vs-latest pattern — none has a failure hiding behind a later success; dedup-by-`started_at` is not masking anything here.
- `verify-ci-shard-tally.sh langwatch/langwatch 6844` → exit 0, run `31723952392` attempt 1 @ `e8245e2f9`. Read the per-shard list directly rather than trusting the exit code: all 4 `test-unit` shards present (472/471/472/470 files passed) and all 6 `test-integration` shards present (155/157/156/160/160/157 files passed), every shard `hard-floor=0`. Confirms real tallies, not the #4450 masked-shard gap.
- Confirmed head via `gh pr view --json headRefOid`: still `e8245e2f9f7fef760dc7b80797469455ab8c328a`, matches this verdict's `review-clean` marker below.

No blocking findings. No new findings.

---

#### Full history for context

1. Prior blocking finding (docblock flag-ordering rationale) — filed, fixed in `e8245e2f9f`, independently re-verified cleared (prior pass).
2. Two non-blocking Decide items — logged, left open as owner's call, disclosed in the docblock and PR body (prior pass).
3. Dark-surface parity and prior round's four findings — independently re-verified, all held (earlier pass).
4. CI — now settled and green, independently confirmed (this pass).

platform/app/src/server/routes/langy-api.ts
drewdrewthis: Fixed in c7d2c936ac.

Both halves confirmed and both fixed:

**The envelope leak.** Verified `createApiRouter()` sets no top-level `.notFound()`, and `createServiceApp` installs only `onError` (`secured-app.ts:148`) — no `notFound` either. So an unmounted path falls to Hono's default plain-text `404 Not Found`, while the thrown `NotFoundError` came back through `canonicalErrorResponse` as JSON with `trace_id`/`span_id`. The dark refusal now returns `c.notFound()` — that same default handler — so status, body and Content-Type all match an unmounted route. It is the one refusal in the file that does not throw, and the docblock now says why: here the envelope that makes every other refusal legible is itself the leak.

**The ceiling ordering.** Also correct. The flag check moves ahead of `enforceApiKeyCeiling`, so a key without `langy:create` no longer gets a 403 out of a surface meant to be dark. It stays after credential resolution, which is the ordering the docblock already argued for — an unauthenticated caller still learns only that its token is bad.

platform/app/src/server/app-layer/langy/errors.ts
drewdrewthis: Fixed in 2de658d871.

Reproduced first: `pnpm test:unit run src/features/errors/logic/__tests__/codes.unit.test.ts` failed at the old HEAD with exactly the three codes you listed. Passes 8/8 now.

All six are added to `codes.ts` and `presentation.ts`, not just the three the scanner sees. The remediation registry already had all six, so only the client-facing copy was missing.

On the three that evade the scanner: I kept `LangyApiIdentityDeniedError` as one class. Its docblock argues the three share one body and differ only in remediation, and splitting it is the author's call, not a review fix. Instead they go in a named `PARAMETERIZED_CODES` set alongside the three sets that already exist for codes the scanner structurally cannot see (`PACKAGE_OWNED_CODES`, `RELAYED_META_CODES`, `CLIENT_MINTED_CODES`). Its docblock records the cost honestly: nothing will notice if a fourth code joins that union and is left off the list.

platform/app/src/features/errors/logic/presentation.ts
```diff
@@ -1422,6 +1422,39 @@ const presentations = {
...
+  },
+  langy_api_actor_missing: {
+    title: "Key owner is gone",
+    describe: () =>
+      "The user who owns this key no longer exists, so the turn has no one to act as. Mint a new key under a current user.",
+  },
+  langy_api_request_invalid: {
+    title: "Invalid request body",
+    describe: () =>
+      "The request body did not match the turn schema. The `issues` list in the error meta names each field that failed.",
```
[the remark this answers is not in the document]
drewdrewthis: Taken in `5dfbd12989`.

Kept your point, not quite your wording. The copy no longer names the turn schema or the `meta` layout, but it still tells the caller that the rejected fields are listed in the error details — an API caller reading a JSON envelope can act on that, and dropping it entirely would leave them with "some fields were invalid" and nowhere to look.

New line: "Some fields in this request were not valid. The error details list each one that was rejected — correct those and send it again."

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Start or continue a turn.
+ *
+ * Nothing is caught. A domain `HandledError` already carries the status, code
+ * and fault the app layer decided on, and anything unhandled is a platform
+ * fault the shared handler logs and masks behind a trace id — re-classifying
+ * either one here could only lose information.
+ */
+async function startTurn(c: Context, conversationId: string | null) {
```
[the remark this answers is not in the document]
drewdrewthis: Taken as written in `5dfbd12989`. `startTurn` now takes `{ c, conversationId }` and both call sites pass by name.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,252 @@
...
+ * The chain, in refusal order:
+ *   1. credential resolves            → else 401
+ *   2. `release_langy_api_key_turns_enabled` → else 404 (surface is dark)
+ *   3. `langy:create` ceiling         → else 403
+ *   4. key owns an actor + has Langy  → else 403
+ *   5. actor row still exists         → else 403
+ *   6. shared `startConversationTurn` → 202
+ *
+ * The flag is checked AFTER authentication on purpose: an unauthenticated
+ * caller learns only that the token is bad, never whether the surface exists.
```
drewdrewthis: **Fix** — the central claim in this docblock ("an unauthenticated caller learns only that the token is bad, never whether the surface exists") does not hold, and it's the exact property the caller asked me to falsify.

The refusal chain runs credential resolution (step 1) strictly before the flag check (step 2). A caller with **no credential at all** never reaches the flag check — they get a 401 canonical-JSON envelope (`LangyApiCredentialMissingError`, thrown, full `code`/`trace_id`/`span_id`) unconditionally, regardless of whether the flag is on or off. A genuinely unmounted path returns Hono's plain-text 404 to that same request, since routing never matches and no auth middleware ever runs. 401-canonical-JSON vs 404-plain-text is trivially distinguishable — so an anonymous prober with zero credentials *can* tell this path is registered (something requires auth here) purely from status code and body shape, before the dark-surface mechanism (`c.notFound()` at line 210) is ever reached.

What's actually true, and verified: for a caller bearing a *valid, resolvable* credential (a real project API key), the flag-off 404 is genuinely byte-for-byte indistinguishable from unmounted (confirmed no `.notFound()`/`.onError()` override exists anywhere in the router chain — `api-router.ts`, `secured-app.ts`, `start.ts` — so `c.notFound()` hits the same default handler an unmatched route would). That's the gap the earlier review round (`c7d2c936ac`) actually closed, and it holds.

The overclaim is only in the comment, not the runtime behavior — no code change needed, just narrow lines 20-21 (and the refusal-chain summary at 12-18, which already documents the true ordering and inadvertently proves the point) to state precisely what's protected: existence-of-a-Langy-specific-feature is hidden from *credentialed* callers; an uncredentialed caller learns "a protected route lives here," same as it does for every other guarded route on this API, which does not reveal anything Langy-specific and is this API family's existing auth-before-routing convention. Leaving the overbroad claim as-is risks a future reader believing this defends against anonymous route enumeration, which it does not.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Authenticate, open the flag, enforce the ceiling, and bridge to an actor.
+ *
+ * Throws on every refusal EXCEPT the dark surface, which returns `{ dark: true }`
+ * for the caller to answer — see the flag check below for why that one cannot
+ * throw. `enforceApiKeyCeiling` already throws a `HandledError`
+ * (`ApiKeyPermissionDeniedError`), so the ceiling denial needs no translation
+ * here at all — catching it only to re-serialise it was how the code, the
+ * permission in `meta` and the tips got dropped.
+ */
+async function authorizeTurn(c: Context) {
```
drewdrewthis, line 155: **New AC** — `authorizeTurn`'s refusal chain (dark-404 parity, ceiling-before-identity ordering, the 401/403/404 boundary this PR's whole design rests on) has zero direct test coverage. Only the two extracted pure helpers (`resolveLangyKeyIdentity`, `resolveLangyActorSession`) are unit-tested; nothing exercises this function or the route through a request. The PR body already discloses this ("No integration test exercises the mounted route over real HTTP"), so this isn't a surprise — but it is exactly the kind of test that would have caught the byte-indistinguishability overclaim above mechanically instead of by manual review, and this codebase has cheap precedent for it: `src/server/routes/__tests__/*.integration.test.ts` and `*.unit.test.ts` files already drive routes via `app.request(...)` without full e2e infra (e.g. `experiments-abort-cross-tenant.integration.test.ts`, `unsubscribe.unit.test.ts`). Given this route is a new external attack surface behind an identity-bridge security boundary, recommend adding at minimum: (1) no-credential request while flag is off asserts byte-match against a genuinely unmounted path, (2) valid-credential + flag-off asserts `c.notFound()` byte-match, (3) valid-credential + flag-on + no `langy:create` asserts 403 with the ceiling code, ordered after the flag check per the docblock's claimed ordering.

drewdrewthis (commented):
Self-review (IS_OWN_PR=true) — see the upserted verdict comment for the full writeup.

drewdrewthis (commented):
Self-review, NOT-READY per the upserted verdict comment (2 blocking inline threads). REQUEST_CHANGES was attempted first and 422'd with "Can not request changes on your own pull request" (single-account fleet — acting token is the PR author) — COMMENT stands in, and this verdict carries the same gating weight since nothing downstream reads GitHub's review-state machine.

platform/app/src/server/routes/langy-api.ts
drewdrewthis: Confirmed and fixed in `cb139f8631` — the claim was false as written, and I verified it against the code rather than taking the review at its word: `if (!credentials) throw new LangyApiCredentialMissingError();` sits above the flag check, so an uncredentialed request gets 401-canonical-JSON regardless of the flag, against plain-text 404 from an unmounted path.

I took the recommended direction (narrow the comment, leave the ordering) for a reason worth recording: the flag is evaluated per project (`distinctId: resolved.project.id`), so it *cannot* be checked before a credential resolves — with no credential there is no project to evaluate against. Making the surface dark to anonymous callers would mean returning 404 for a missing credential even when the surface is live, which is a worse API in exchange for a property this route was never able to offer.

The docblock now states the guarantee it actually has (a caller with a real project key cannot tell whether Langy is here) and names the limit explicitly, so nobody later reads this as a defence against anonymous route enumeration. The boundary is also pinned by a test rather than left to prose.

platform/app/src/server/routes/langy-api.ts
drewdrewthis: Added in `cb139f8631`: `src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts`, 11 tests driving the real Hono app via `.request()` per the precedent you named — no DB, runs in `test:unit`.

Covers all three you asked for plus the rest of the chain: dark-404 parity against an unmounted path *on the same app* (a separate bare Hono would not prove no handler in this chain overrides the default 404), flag-off leaving ceiling/identity/turn-service untouched, flag-off staying dark for a key that would fail the ceiling, ceiling denial surfacing 403 with `api_key_permission_denied` intact, all three identity refusals, invalid body, and the 202 happy path.

On your point that this would have caught the overclaim mechanically: it does now. The uncredentialed case is pinned as a 401 with `isEnabled` asserted *not* called — so the boundary is recorded as tested behaviour, and a future attempt to claim anonymous-enumeration protection contradicts a test.

Both load-bearing guards were confirmed to fail for the right reason before being kept: moving `enforceApiKeyCeiling` above the flag check turns two tests red, and swapping `c.notFound()` for a same-status JSON 404 turns the parity test red.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,263 @@
...
+ * `code`, `meta` and remediation `tips` a bespoke `{ message }` would have
+ * discarded (ADR-045). The family opts into the `canonical` envelope because it
+ * is new: there is no existing consumer parsing the flat legacy shape.
+ *
+ * The dark-surface 404 is the one deliberate exception, and it does not throw
+ * at all: the envelope that makes every other refusal legible is itself the
+ * leak here. A thrown 404 comes back as canonical JSON carrying `trace_id` and
+ * `span_id`; an unmounted path comes back as plain-text `404 Not Found` from
+ * Hono's default handler. So the dark refusal returns `c.notFound()` — that
+ * same default handler — and matches an unmounted route in status, body and
```
drewdrewthis: **[correctness] The docblock states a wire shape the wire does not have — same class as the round-1 claim, different sentence.**

> A thrown 404 comes back as canonical JSON carrying `trace_id` and `span_id`; an unmounted path comes back as plain-text `404 Not Found` from Hono's default handler. So the dark refusal returns `c.notFound()` — that same default handler — and matches an unmounted route in status, body and Content-Type.

In production neither one comes back as plain text. `src/start.ts` routes every `/api/*` request through `honoFetchForNode`, which intercepts a 404 whose body is exactly `"404 Not Found"` and rewrites it to `{"error":"Not Found"}` with `Content-Type: application/json`. Both the dark refusal and the unmounted path get that rewrite, so **the parity property still holds** — but the caller-visible shape you describe is wrong, and a reviewer who curls it will find JSON and conclude the docblock is stale.

Worse, the parity now rests on a mechanism this file does not mention and this route's tests never execute. That bridge has an explicit carve-out — *"A route's own 404 (different body) passes through untouched"* — which is precisely why `c.notFound()` must keep emitting the exact default sentinel and nothing else. That is a load-bearing coupling to another file, currently invisible from both.

**Fix:** say what the caller actually receives, and name `honoFetchForNode`'s 404 rewrite as the reason both paths converge. The same correction applies to the test file docblock ("the same status, body and Content-Type an unmounted path returns" is fine; the surrounding narrative implies plain text reaches the client) and to the `expect(dark.contentType).not.toContain("application/json")` assertion, which pins the pre-bridge response, not the wire — worth a comment saying so, since the post-bridge content-type IS `application/json`.

platform/app/src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      expect(mockStartConversationTurn).not.toHaveBeenCalled();
+    });
+
+    it("accepts a well-formed turn with 202", async () => {
+      const res = await postTurn();
+
+      // 202, not 200: the turn is dispatched and the answer does not exist yet.
+      expect(res.status).toBe(202);
+      expect(await res.json()).toEqual({ conversationId: "conv-1" });
+      expect(mockMarkUsed).toHaveBeenCalledWith({ apiKeyId: "key-1" });
```
drewdrewthis, line 327: **[correctness] Nothing pins that the credential's project reaches the turn service, and the continue route has zero coverage.**

`mockStartConversationTurn` is asserted on `not.toHaveBeenCalled()` in two tests and on nothing at all in the success test — only the status, the response body and `markUsed` are checked. So this suite stays fully green if `startConversationTurn` is handed the wrong `projectId`, a session that isn't the resolved actor's, or a null `requestedConversationId` on the continue route. `projectId` is the tenancy boundary on this call; it is the one argument that must not be able to drift silently.

And `POST /api/langy/conversations/:conversationId/messages` is never requested in this file. Half the route surface — the half where `conversationId` is threaded from a path param rather than hard-coded `null` — is untested, on a PR whose stated subject is which refusal a caller gets and in what order.

**Fix:** assert the dispatch arguments on the 202 test —

```ts
expect(mockStartConversationTurn).toHaveBeenCalledWith(
  expect.objectContaining({
    projectId: "project-123",
    session: { user: { id: "user-1" } },
    requestedConversationId: null,
    idempotencyKey: "idem-1",
  }),
);
```

— and add a sibling test that POSTs the continue route and asserts `requestedConversationId: "conv-9"` reaches the service. One dark-surface test on the continue route would be worth having too: the flag check lives in shared code, but nothing currently proves the second mount goes through it.

platform/app/src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      expect(JSON.stringify(await res.json())).toContain(
+        "langy_api_key_unowned",
+      );
+    });
+
+    it("refuses an owner without Langy access as 403", async () => {
+      mockResolveLangyKeyIdentity.mockResolvedValue({
+        ok: false,
+        reason: "no_langy_access",
+        message: "no access",
```
drewdrewthis, line 282: **[correctness] `reason: "no_langy_access"` is not a value the domain type permits, so this test never exercises the real one.**

`LangyIdentityDenialReason` is `"unowned" | "no-access"` (`langyApiKeyIdentity.ts:37`) — hyphen, and a different word. The real helper returns `reason: "no-access"`. Because the double is an untyped `vi.fn()`, the impossible value compiles and the test passes anyway: `authorizeTurn`'s ternary maps *anything that is not `"unowned"`* to `langy_api_key_no_langy_access`, so the assertion is satisfied by the else-branch regardless of what you put here.

That makes the test vacuous with respect to the value it names, and it masks a contract: if someone later tightens the ternary into an explicit `identity.reason === "no-access"` check with a different default, this test keeps passing while the real path changes behaviour. This is the concrete cost of mocking code we own (`references/principles/coding.md`) — a typed fake would have rejected the string at compile time.

**Fix:** use `"no-access"`. Better, type the two identity doubles against the real signatures (`vi.fn<typeof resolveLangyKeyIdentity>()` or a small typed fake) so no future fixture can name a state the domain cannot produce.

specs/langy/langy-api-key-turns.feature
```diff
@@ -0,0 +1,195 @@
...
+
+  @unit
+  Scenario: A key whose owning user no longer exists is refused
+    Given a project API key whose owning user record has been deleted
+    When the surface builds the acting identity for a turn
+    Then the turn is refused
+    And no stand-in actor is used in the owner's place
+
+  @unit
+  Scenario: The key-authed surface is switched off by default
```
drewdrewthis, line 65: **[completeness] AC: the two properties this round exists to defend are absent from the spec, and the 11 new tests bind nothing.**

The dark-surface parity (a credentialed caller cannot distinguish the flag-off surface from a route that was never deployed) and the flag-before-ceiling ordering (a key lacking `langy:create` must not receive a 403 out of a dark surface) are the PR's central security claims. They live only in a docblock and an unbound test file. `specs/langy/langy-api-key-turns.feature` has no scenario for either — the nearest, *"The key-authed surface is switched off by default"*, is about the registry entry, and is bound to `langyApiKeyActorSession.unit.test.ts`, not to anything that exercises the route.

Meanwhile `langy-api-refusal-chain.unit.test.ts` carries no `@scenario` annotation on any of its 11 tests, so `check-feature-parity.ts` counts none of them. CLAUDE.md is explicit: error paths are named in the spec alongside the golden path, and a failure scenario enforces nothing until it is tagged **and** bound.

**Fix:** add `@unit` scenarios in this file — e.g. *"A dark surface is indistinguishable from a route that was never deployed"* and *"A key without langy:create is not told the dark surface exists"* — and annotate the corresponding tests with `@scenario`. While you are here, the uncredentialed-401 boundary deserves its own scenario too: it is a deliberate, argued limit on the dark surface, and today it is recorded only in prose.

drewdrewthis (commented):
**NOT-READY.** This review is submitted as `COMMENT` only because GitHub refuses `REQUEST_CHANGES` on your own pull request (`422 Can not request changes on your own pull request`) — the acting token is the PR author. Treat it as a blocking change-request: four resolvable threads below.

**What I re-verified rather than took on trust**

- **Round-1 finding (a), the ordering argument: I judge it SOUND, no code reorder needed.** `release_langy_api_key_turns_enabled` is `scope: "SYSTEM"`, which resolves env → postgres → default, and the postgres store is where per-project/per-org targeting rules live (the `release_langy_enabled` registry entry says so in as many words). So the flag genuinely cannot be evaluated before a project is in hand, and 404-ing an uncredentialed caller would make the surface answer 404 to a missing token even with the flag **on** — a real DX regression and a deviation from every other route on this API. The narrowed docblock does not overclaim: after credential resolution, dark ⇒ 404 on every path, and the "no credential still gets 401" carve-out is now stated plainly instead of denied. One sentence in it is still wrong, but for a different reason — thread 1.
- **Round-1 finding (b), the new test file: it is not vacuous.** I re-ran both mutations you claim. Moving `enforceApiKeyCeiling` above the flag check turns exactly 2 tests red (`does not reach the ceiling…`, `stays dark for a key that would fail the langy:create ceiling`). Swapping `c.notFound()` for `c.json({error:{message:"Not Found"}}, 404)` turns the parity test red on the deep-equal. Both reproduce; tree reverted and `git status` is clean.
- **Over-mocking:** acceptable on balance. The route, `createServiceApp`, `handlerManagedAuth`, `bodyLimit` and the canonical error envelope are all real; the mocks sit at genuine collaborator boundaries and the assertions are on emitted HTTP responses, not call sequences (the two `not.toHaveBeenCalled()` assertions are how ordering is *observed*, which is legitimate). It bites under mutation. Two specific places where it does cost you correctness are threads 2 and 3.
- **Parity ground truth:** comparing against an unmounted path on the same app is the right idea, and status + content-type + body is enough to compare. It holds today: `grep` finds no `.notFound(` registration anywhere under `src/server/` or `src/app/api/`. See the non-blocking note about using `createApiRouter()`.
- **The three earlier fixes hold.** `c7d2c936ac`: flag now precedes the ceiling and the dark refusal returns `c.notFound()` — both pinned. `2de658d871`: all six `langy_api_*` codes are in `APP_ERROR_CODES` (sorted) with `presentation.ts` entries and `error-remediation.ts` tips; the copy is caller-appropriate and names no internals. `5dfbd12989`: the request-invalid copy no longer names the schema or the `meta` layout, and `resolveLangyActorSession` takes named parameters.

**Non-blocking**

- **`PARAMETERIZED_CODES` — position unchanged from last round, still non-blocking.** The allowlist is honestly documented and it does exactly what it says, but it converts a mechanical guard into a hand-maintained list for this one family. My preference remains three subclasses with literal codes (which the scanner sees) over one class with a parameterised union; I am not blocking on it, and the docblock you wrote is the right mitigation if it stays.
- **The bare `new Hono()` parent.** Three tests in this repo already build the real router (`api-endpoint-authorization.integration.test.ts`, `end-user-spend-routing.integration.test.ts`, `experiments-route-auth.test.ts`). Mounting `createApiRouter()` instead would make the parity test's ground truth the production chain, so a future catch-all or `.notFound()` on the api-router could not break parity invisibly. (Decide.)
- **The `stays dark for a key that would fail the langy:create ceiling` mutation is weaker than it reads.** The ceiling mock rejects a plain `Error`, so under the reordering mutation the response is **500**, not the 403 the leak actually was. Reject with `ApiKeyPermissionDeniedError` and assert `404` against a would-be `403` and the test demonstrates the real regression rather than a proxy for it.
- **Uncovered branches beyond thread 2:** `bodyLimit` (the 1 MiB cap is declared and never exercised), `modelOverride` pass-through, the `c.req.json().catch(() => null)` malformed-JSON path (the invalid-body test sends well-formed JSON with a wrong type, so the catch never fires), and the `resolved.type !== "apiKey"` no-op branch of `markUsed`.
- **`VALID_TURN_BODY` sends `{ role, content }` but the wire schema is `{ role, parts }`.** It parses only because `parts` has `.default([])` and zod strips the unknown key, so the fixture quietly documents a shape the API does not accept. (Decide.)

platform/app/src/server/routes/langy-api.ts
drewdrewthis: You are right, and this is the second time this docblock has claimed something about the wire that I had not checked — so I checked it this time. `honoFetchForNode` (`src/start.ts:515`) intercepts every 404 leaving the app and, when the body is exactly Hono's `404 Not Found` sentinel, rewrites it to `{"error":"Not Found"}` with `Content-Type: application/json`. Production serves JSON here, not plain text.

Fixed in `7ffe5d100c`. The docblock now states the real shape, names `honoFetchForNode` as the reason parity survives (the rewrite keys off that sentinel body, so it hits both sides of the pair identically), and calls out the carve-out you identified — a 404 with any other body is passed through untouched. That carve-out is the interesting part, so it is now stated as the reason this refusal must stay `c.notFound()`: a hand-built `c.json(..., 404)` would skip the rewrite and reach the wire looking different at an identical status. The test docblock carries the same correction and notes its assertion compares at the Hono layer, upstream of the bridge.

I also corrected my earlier reply on the docblock thread, which repeated the same wrong claim.

platform/app/src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts
drewdrewthis: Both covered in `7ffe5d100c`.

Tenancy: `passes the credential's project and session to the turn service` asserts `startConversationTurn` receives `projectId: "project-123"`, the resolved `session`, `idempotencyKey` and `requestedConversationId: null` — you were right that a 202 alone would still pass if the route handed the service the wrong tenant.

Continue route: two tests now request `POST /conversations/conv-existing/messages` — one asserting the path id is forwarded as `requestedConversationId`, one asserting it stays dark on the same terms as create and dispatches nothing.

Both bite. Replacing `projectId: auth.projectId` with a literal reddens both tenancy tests; forcing `requestedConversationId` to `null` reddens only the continue test. Reverted, and the route diff in this commit is comment-only.

platform/app/src/server/routes/__tests__/langy-api-refusal-chain.unit.test.ts
drewdrewthis: Confirmed and fixed in `7ffe5d100c` — and it was worse than a typo, exactly as you say. `"no_langy_access"` is not `"unowned"`, so it fell into the else-branch of `identity.reason === "unowned" ? … : …` and satisfied the assertion while never exercising the real `"no-access"` value.

Now `reason: "no-access" satisfies LangyIdentityDenialReason`, with the same on the `"unowned"` case. That converts the defect from invisible to a compile error: restoring the old value fails `typecheck:tests` with `TS1360: Type no_langy_access does not satisfy the expected type 'LangyIdentityDenialReason'`. I verified that, then reverted.

Your related note was right too — `VALID_TURN_BODY` sent `content` where the schema takes `parts` (defaulted to `[]`), so it was stripped as unknown and still reached 202. Passing for the wrong reason; also fixed.

specs/langy/langy-api-key-turns.feature
drewdrewthis: Fixed in `7ffe5d100c`, though I got there by a wrong turn worth recording: I first concluded this PR had no spec at all and wrote a new feature file. That was my error — I ran `git diff -- specs/` from `platform/app`, so it resolved to `platform/app/specs/` and reported nothing. `specs/langy/langy-api-key-turns.feature` exists and this PR added it.

So the accurate version of your finding is the narrower one: the file was there, but neither property this round defends had a scenario in it. Added two `@unit` scenarios in its Rollback section — a switched-off surface answering exactly as an unrouted path, and the switch being read before the caller's permissions — and bound them, plus three route-level bindings onto existing identity scenarios. I deleted the duplicate file I had created.

`check-feature-parity.ts` goes 8/8 to 10/10 on that file. Verified non-vacuous: renaming one `@scenario` annotation reports the scenario unbound at `specs/langy/langy-api-key-turns.feature:170`. Reverted.

platform/app/src/server/routes/langy-api.ts
drewdrewthis: Correction to this reply. The reasoning about ordering holds, but the wire detail I asserted here was wrong: I said an uncredentialed request gets 401 "against plain-text 404 from an unmounted path". An unmounted path does not return plain text — `honoFetchForNode` (`src/start.ts:515`) rewrites Hono's `404 Not Found` sentinel into `{"error":"Not Found"}` JSON before it reaches the wire.

It does not change the conclusion (the 401-vs-404 distinction is the point, and parity between the dark refusal and an unmounted path survives because the rewrite matches on that sentinel body and so applies to both). But it was stated as verified when I had not checked that hop. Corrected in the docblock in `7ffe5d100c`.

platform/app/src/server/routes/langy-api.ts
```diff
@@ -0,0 +1,278 @@
...
+ * The chain, in refusal order:
+ *   1. credential resolves            → else 401
+ *   2. `release_langy_api_key_turns_enabled` → else 404 (surface is dark)
+ *   3. `langy:create` ceiling         → else 403
+ *   4. key owns an actor + has Langy  → else 403
+ *   5. actor row still exists         → else 403
+ *   6. shared `startConversationTurn` → 202
+ *
+ * The flag is checked AFTER authentication because it has to be: it is
+ * evaluated per project, so there is no project to evaluate it against until a
```
drewdrewthis: **Fix — docblock overstates *why* the flag is checked after auth**

This says the ordering is required because "it is evaluated per project, so there is no project to evaluate it against until a credential resolves." That's not accurate. `featureFlagService.isEnabled()`'s only *required* option is `distinctId` — `projectId`/`organizationId` are optional (`platform/app/src/server/featureFlag/types.ts`) — and the flag store already computes a meaningful project-less "default context" evaluation: `resolveEffectiveForListing` (`platform/app/src/server/featureFlag/rules.ts:86-102`), used today by the Ops UI listing table (env override → any empty-match rule → row `enabled` → registry default). A project-less check is a supported, well-defined operation in this exact service, not a structural impossibility.

That's more than a doc-accuracy nit: it means the mitigation this docblock waves off two lines later — resolving the flag's *default* before credentials resolve, to answer "no such route" to an uncredentialed/bad-credential caller whenever the surface is dark by default — was never actually ruled out. It's a real, viable fix for exactly the gap the next paragraph owns ("a caller holding no credential... can tell that SOME authenticated route is mounted"). It wouldn't help while a per-project targeting rule is active for this flag (that case genuinely needs the real project), but in the common "off for everyone" state it would close the leak entirely, not just scope it.

Please correct this to state the ordering as a deliberate scope choice, not a technical necessity — and either implement the pre-auth default-context check or file a tracked follow-up if intentionally deferred. (Whether it's worth implementing is a separate, non-blocking call — see the verdict comment.)

drewdrewthis (commented):
Two-pass review complete (challenge-driven, per the reviewer brief). Verdict comment upserted below with the READY/NOT-READY call.

platform/app/src/server/routes/langy-api.ts
drewdrewthis: Confirmed, and fixed in `e8245e2f9f`. I checked this against the source rather than taking the finding on trust, and it holds:

- `FeatureFlagEvaluateOptions` (`featureFlag/types.ts:10-27`) requires only `distinctId`; `projectId?: string` is optional.
- `resolveEffectiveForListing` (`featureFlag/rules.ts:86-102`) takes no project or context parameter at all and calls `evaluateRules(rules, {})` — a genuinely project-less evaluation, shipped and live behind `listFeatureFlags` in the Ops UI (`api/routers/ops.ts:782`).

So "it has to be" was false, and stating it as a necessity foreclosed a real mitigation for the anonymous-enumeration gap the same docblock explicitly owns. That is the part worth fixing regardless of what we do about the gap itself.

The ordering is unchanged — it was the right call for a different reason than the one written down. Rollback is staged per project, so the check that matters is the one evaluated against the callers own project, and that project genuinely is not known until a credential resolves. A default-context evaluation answers on the flags default rather than the callers project, so it cannot *replace* this check; it could only sit in front of it as an extra pre-auth filter. The docblock now says exactly that, and names the pre-auth filter as an available option this PR does not take.

I have deliberately not implemented that filter here. It changes the surfaces behaviour for uncredentialed callers, which is a scope and product call, and this PR is the owners. Your non-blocking Decide on it is left open rather than answered by me. Same for the `bodyLimit` 413 pre-auth observation — real, matches the ~7 other routes platform-wide, and now covered by the docblocks own disclaimer that this mechanism is not a defence against anonymous route enumeration.

Corrected at both sites it was asserted: the docblock and section 3 of the PR body.

platform/app/src/server/routes/langy-api.ts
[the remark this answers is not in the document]
drewdrewthis: Correcting my own earlier reply on this thread, unprompted.

I argued this comment was stale because its `createdAt` (`2026-08-13T10:05:43Z`) preceded commit `edea1d2695` (`2026-08-13T10:05:47Z`) by four seconds. That reasoning was not sound. `createdAt` only dates a comment's creation, not its current text — this comment's `updatedAt` is `2026-08-13T15:31:37Z`, hours after the commit I compared against. A `createdAt`-only comparison cannot establish that the content in front of me predates a commit, and I stated it as if it could.

The substantive conclusion is unchanged and was reached by other means: `pnpm run check:openapi-completeness` and `check:openapi-route-coverage` both exit 0, and the two operations are recorded `gap` entries (tracked debt), not silenced. But the staleness argument itself should not be relied on, and I would rather say so than leave it standing.

drewdrewthis:
### Handoff — I have stopped driving this PR

**Why.** Another lane is pushing to this branch concurrently: four commits landed inside my last turn (15:57–16:46Z), and `e8245e2f9f` was committed seconds after I read it as an uncommitted change to `langy-api.ts`. Continuing to write shared artifacts without a coordination mechanism is what produced a self-contradicting PR body earlier today, so I have disengaged from the source rather than race it.

**Landed and verified this turn:**
- Fixed a contradiction in the PR body — one bullet claimed a verified green existed only for `edea1d2695` while another said `7ffe5d100c` was verified. The body now states one position, pinned to the SHA it was measured on.
- Confirmed `e8245e2f9f` is genuinely comment-only: the `-U0` diff filtered to non-comment, non-blank lines is empty. No executable line differs from `7ffe5d100c`.
- Verified the green on `7ffe5d100c` was real rather than a tick — `verify-ci-shard-tally.sh` exit 0, run `31722417714` attempt 1, full 4+6 shard matrix, every shard printing a real tally. **That verdict is not carried forward** to `e8245e2f9f`.
- Posted an unprompted correction to my own earlier staleness argument on a CodeRabbit thread: I had reasoned from `createdAt` alone, and `updatedAt` was hours later, which invalidates it.

**Outstanding — neither is mine to close:**
1. **C2** — CI is still running on `e8245e2f9f`. When it settles, re-run `verify-ci-shard-tally.sh langwatch/langwatch 6844` against the new head; a bare green tick is not sufficient evidence on this repo.
2. **C3** — the review verdict says NOT-READY, but it was written *before* `e8245e2f9f`, which fixes its one blocking finding (the flag-ordering rationale in the route docblock). It needs a re-review at the current head. I have a standing instruction not to dispatch a review agent unless asked, so I have not.

C1, C4, C5, C6 and C7 pass; all 19 review threads are resolved.

*Correction, disclosed rather than quietly fixed:* my first attempt to post this used `gh pr comment --edit-last`, which overwrote the review verdict comment above instead of creating a new one. I restored it from a copy I had taken before the edit — verified back at 7773 characters with both the `review-verdict` and `review-clean: 7ffe5d100c…` markers intact.

drewdrewthis:
### Driving resumed and completed — both handoff items closed

Picking up from [the handoff above](https://github.com/langwatch/langwatch/pull/6844#issuecomment-5283989209). Head is unchanged at `e8245e2f9f` since that comment, so nothing landed underneath it. Its two outstanding items are now closed:

**1. C2 — CI settled on `e8245e2f9f`.** `verify-ci-shard-tally.sh langwatch/langwatch 6844` → exit 0, run [`31723952392`](https://github.com/langwatch/langwatch/actions/runs/31723952392) **attempt 1**, full 4+6 matrix, every shard `hard-floor=0` with a real tally:

| shard | files passed | shard | files passed |
|---|---|---|---|
| test-unit 1–4 | 472 / 471 / 472 / 470 | test-integration 1–6 | 155 / 157 / 156 / 160 / 160 / 157 |

Not taken from a bare green tick. One check name (`breaking change stays in one component`) showed a `cancelled` run that a naive poll reads as a failure — it has three runs (`17:05:24 cancelled` → `17:05:38 success` → `17:09:19 success`) and the cancel is superseded. All nine other multi-run check names were checked for the same pattern; none hides a failure behind a later success.

*Addendum, after the metadata edits below.* Setting the label and requesting a reviewer re-triggered the lightweight `pull_request_target` workflows at 18:30Z. The heavyweight `langwatch-app-ci` run cited above is a `pull_request` run and did **not** re-run, so the shard tally is still the authoritative measurement at this head. Live state re-read after those settled: **73 check names, 0 pending, 0 failed.** The multi-run set has grown from 10 names to 20 since I wrote "nine other" — all 20 re-checked, none hides a failure behind a later success.

I also stress-tested the dedup idiom itself, since the whole C2 argument rests on it. Deduping by `max_by(.started_at)` agrees with an independent monotonic key (`max_by(.id)`) on **19 of 20** multi-run names. The exception is `changes`, which has 21 runs sharing two timestamps to the second — `max_by` on a tie picks arbitrarily. It is harmless here only because every tied run concluded `success`. **`max_by(.id)` is the stricter idiom** and would have no ties; `started_at` is safe only when tied runs agree.

**2. C3 — re-reviewed at the current head, verdict now READY.** The blocking finding (the flag-ordering rationale in the route docblock claiming the flag *had* to follow authentication because there is no project to evaluate against) was false — `featureFlagService.isEnabled()` requires only `distinctId`, and `resolveEffectiveForListing` is a shipped project-less evaluator used in production by `listFeatureFlags`. `e8245e2f9f` restates the ordering as a deliberate choice and names the pre-auth filter it does *not* implement. Verified as cleared, and the same sentence was corrected in the PR body.

**Bar:** `pr-ready-check.sh langwatch langwatch 6844` → `overall=READY`, C1–C7 all pass, verdict SHA matches HEAD. Phase label moved to `pr-ready`; `needs-driving` removed; review requested from a human.

**What a green bar here does not mean.** `reviewDecision: REVIEW_REQUIRED`, `mergeStateStatus: BLOCKED` (`mergeable: MERGEABLE` — no conflicts). No bar criterion reads `reviewDecision`, so this PR is content-ready and CI-green but **not mergeable**: branch protection wants an approving review, and GitHub rejects `APPROVE` from the PR author's own login, which is the only identity available to the driving session. That is not a gap I can route around, and I have not tried to. **It needs one human approval.**

**Left open deliberately, all the owner's calls, none blocking:**
- The two Langy operations recorded as `gap` in `openapi-route-exclusions.ts`.
- `bodyLimit` runs before `authorizeTurn`, so an oversized POST gets a 413 pre-auth — same as roughly seven other routes; a pre-existing pattern, not introduced here.
- The pre-auth flag filter that would close the enumeration gap is named in the docblock and deliberately not implemented.

platform/app/src/server/app-layer/langy/__tests__/langyApiKeyActorSession.unit.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      (f) => "key" in f && f.key === "release_langy_enabled",
+    );
+
+    expect(surface).toBeDefined();
+    expect(langy).toBeDefined();
+    // Closed until someone opts in: shipping the route must not, by itself,
+    // open a new way into Langy for any existing project.
+    expect((surface as { defaultValue: boolean }).defaultValue).toBe(false);
+    // Two distinct keys — turning the surface off leaves browser Langy alone.
+    expect(surface).not.toBe(langy);
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in `4d9780b701`, but this thread was resolved without a reply naming that commit — so from the outside the finding read as closed-without-answer. Recording it now.

You were right that `expect(surface).not.toBe(langy)` proved nothing: `FEATURE_FLAGS.find` returns two different array elements, so the identity comparison passed unconditionally and would still have passed if both entries carried the same `key`. The assertion now compares the key strings themselves (`expect((surface as { key: string }).key).not.toBe((langy as { key: string }).key)`), and the comment above it states the reason rather than the old "Two distinct keys" claim the code did not support.

Verified at head `e8245e2f9f`: `pnpm test:unit run src/server/app-layer/langy/__tests__/langyApiKeyActorSession.unit.test.ts` — 2 files, 8 tests passed.

platform/app/src/server/app-layer/langy/__tests__/langyApiKeyIdentity.unit.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    type: "apiKey",
+    apiKeyId: "key-1",
+    userId,
+    organizationId: "org-1",
+    project: {
+      id: "project-1",
+      team: { id: "team-1", organizationId: "org-1" },
+    },
+  } as unknown as ResolvedToken;
+}
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in `4d9780b701`, but this thread was resolved without a reply naming that commit — logging it here so the record is not "resolved, unexplained".

Both fixtures are typed rather than cast. `langyApiKeyIdentity.unit.test.ts` no longer uses `as unknown as ResolvedToken`; `apiKeyToken` now returns a declared `LangyIdentityToken`. `langyApiKeyActorSession.unit.test.ts` no longer casts to `never`; `prismaWithUser` is typed as `LangyActorUserReader`.

One deviation from your wording, deliberate: you proposed completing the fixture to a full `ResolvedToken` (adding `ingestSourceType`, `ingestionTemplateId`) and typing the prisma double as `Pick<PrismaClient, "user">`. The narrower contract types were used instead, so the fixture is pinned to exactly what the bridge reads. Your objection — that a cast makes the double stop failing when the upstream contract changes — is satisfied either way, since both are now structurally checked. That a real `ResolvedToken` still satisfies `LangyIdentityToken` is enforced where `langy-api.ts` passes one in, which is the only place a real one exists.

Verified at head `e8245e2f9f`: `pnpm test:unit run` on both files — 2 files, 8 tests passed.

drewdrewthis:
Design follow-on filed as #6983 — exposing Langy over the standard chat wire protocol so third-party clients need no adapter.

That is a **wire-shape** question and is orthogonal to this PR, which establishes *who may call* (the API-key → Langy access bridge), not *what the call looks like*. This PR stays as-is; the surface it mounts is dark behind `release_langy_api_key_turns_enabled`, so the wire format is not yet load-bearing for any external caller and remains cheap to change in #6983.

sergioestebance:
## Review — PR #6844

**Verdict: no blockers.** 22/22 tests pass. The security properties are sound and well-pinned.

### Specific question: does `langyApiKeyActorSession.ts` need a service?

**No.** The `langy/` directory has two established module shapes:

- **Service classes** (`*.service.ts`, `LangyCredentialService`): stateful, constructor-injected, multi-method, lifecycle-bound to the app composition root.
- **Bare functions** (`hasLangyAccess`, `resolveAttributionUserId`, `resolveLangyPrompt`): stateless, transport-free, per-call injectable deps, return discriminated results, never throw.

Both new files are the second shape — single pure function, dependencies via named params, discriminated `{ ok: true } | { ok: false, reason }` return. Wrapping them in a class would break the pattern, not follow it.

### What I traced

- **Refusal chain ordering** — credential → flag → ceiling → identity → actor → turn. The flag-before-ceiling ordering prevents a 403 leaking the surface while dark. Pinned by `langy-api-refusal-chain.unit.test.ts` (14 tests).
- **Dark surface parity** — `c.notFound()` produces the same response as an unmounted path. The test compares status, body, and Content-Type. A thrown 404 would produce the canonical envelope with `trace_id` — that asymmetry IS the leak, and it's correctly avoided.
- **Identity from credential, not payload** — `userId` comes from `resolved.userId` (the key's owner), never from the request body. Test "ignores an actor named anywhere other than the credential" pins this.
- **Fail-closed on ownerless keys** — legacy keys and ownerless API keys return `{ ok: false, reason: "unowned" }` without consulting the flag gate. Test "refuses an ownerless key without consulting the gate" pins this.
- **`ResolvedToken` coverage** — `ResolvedToken` is a two-variant union (`legacyProjectKey | apiKey`); `OrgResolvedToken` (`apiKey-org`) is a separate type for org-level endpoints. `LangyIdentityToken` covers every variant `tokenResolver.resolve()` can return. No gap.
- **Error contract** — all six new codes registered in `APP_ERROR_CODES`, presentation entries exhaustive, remediation tips present. The `PARAMETERIZED_CODES` set in the code scanner test is correctly documented as a hand-maintained set with the cost stated.
- **Feature flag** — `release_langy_api_key_turns_enabled`: scope SYSTEM, default false, not env-overridable, separate from `release_langy_enabled`. ANDed at runtime (the identity bridge calls `hasLangyAccess` which checks the main flag).

### P3 — non-blocking

1. **`as LangyChatMessageInput[]` cast** (`langy-api.ts:~253`): the Zod schema validates structure loosely (`z.record(z.string(), z.unknown())` for parts), then casts to the service type. The docblock says "Parts stay opaque; the app layer bounds them" — so this is intentional, and the app layer already validates. Just noting it exists.

2. **`markUsed` fire-and-forget** (`langy-api.ts:~267`): no `await`, which means a failure to record usage is silent. This matches the existing pattern in other routes — usage tracking is best-effort. Correct for a usage counter, wrong if it were a security gate (it isn't).

sergioestebance (approved):
Blessed. 🙏 The key-authed surface ships dark, refuses closed, and derives identity from the credential alone — no request payload can name the actor.

> "The truth will set you free." — John 8:32

Traced the refusal chain ordering, the dark-surface byte parity, the fail-closed ownerless-key path, and the identity-from-credential-not-payload property: all pinned by tests. Ship it — P3 nits (opaque-parts cast, fire-and-forget markUsed) are deliberate and documented.

PR #6854, opened 2026-08-11
fix(licensing): the subscription page honors what a held plan actually grants

rogeriochaves:
Follow-up to two P2 review comments left on #6847 after it merged:

- [Licensed Enterprise still renders an upgrade CTA](https://github.com/langwatch/langwatch/pull/6847#discussion_r3758877490)
- [An explicitly withheld webhook entitlement is still advertised as included](https://github.com/langwatch/langwatch/pull/6847#discussion_r3758877501)

Both were correct.

## The upgrade call to action

#6847 made `isEnterprisePlan` true for a license-sourced enterprise plan, which was the point: an enterprise customer is on enterprise whichever leg resolved it. That value also decides whether `CurrentPlanBlock` gets a `contactSalesUrl`, and with one, it renders a button reading "Contact us to Upgrade". So the licensed case lost the upgrade badge and the upsell card and kept a third upgrade prompt nobody was looking at. The rule the spec states is that a licensed enterprise plan is not an upgrade candidate, and a button inviting them to talk to sales about upgrading is exactly that.

The url is now supplied only when the plan is enterprise and did not come from a license. Sales has nothing to sell a customer already holding a signed enterprise contract.

The subscription-sourced enterprise case is untouched and keeps the button. Whether "Contact us to Upgrade" is the right label on a top-tier plan at all is a copy question that predates this work and is not answered here.

## The advertised capability a contract withholds

Entitlement resolution deliberately preserves an explicit `webhookEndpointsEnabled: false`, so a contract can withhold a tier feature. The page did not know that: every enterprise plan took the same static `ENTERPRISE_PLAN_FEATURES`, including the gateway webhooks bullet, so such a customer read that webhooks are included while the gate correctly refused them.

The tier's bullets now carry the entitlement field that decides them, where one does. Two lists come off that:

- `ENTERPRISE_PLAN_FEATURES`, the tier as it is SOLD, unchanged in content and still what the plan comparison and contact-sales surfaces render.
- `buildEnterprisePlanFeatures(plan)`, the tier as one customer HOLDS it, which is what the current-plan block renders.

Only an explicit `false` removes a bullet. An entitlement the plan says nothing about is answered by the tier at resolution, so silence reads as granted and a plan that never reached the resolver is described by what the tier sells rather than stripped of it.

The link is the field name, not the sentence, so a bullet whose copy is reworded keeps describing the same capability and nothing matches prose to decide what a plan is missing.

On the reviewer's remaining 5%, whether the list is generic tier marketing or effective contract capabilities: the page's own comment already said a held plan is described by what it grants, and the customer-visible consequence of getting it wrong is a page that contradicts a 403. So the held list is filtered and the selling list is not, which keeps both readings true where each is rendered.

## Specs

Two `@integration` scenarios under the existing rule "A licensed enterprise plan is the current plan, not an upgrade candidate" in `specs/licensing/subscription-page.feature`: one for not being offered a way to contact sales about upgrading, one for a withheld capability not being advertised as included. Both bound.

## How verified

- `pnpm test:unit run` over `src/components/subscription` and `ee/licensing`: 13 files, 231 tests green.
- `pnpm test:integration run` over `src/components/subscription`: 5 files, 91 green, 5 pre-existing skips.
- `pnpm check:feature-parity`: `subscription-page.feature` 5/5 bound, whole run OK.
- `pnpm typecheck:all` clean.
- Biome's new-violation gate against the merge base: no new violations.
- Both tests were confirmed failing first, against the code as merged: the contact-sales button was present, and the webhook bullet was rendered for a plan that withheld it.

## Summary by CodeRabbit

- **New Features**
  - Enterprise plan features now accurately reflect available capabilities, including webhook support.
  - Licensed enterprise plans no longer display a contact-sales upgrade option.
  - Enterprise feature listings retain available capabilities while omitting contractually unavailable features.

### DIFF
--- a/platform/app/src/components/subscription/SubscriptionPage.tsx
+++ b/platform/app/src/components/subscription/SubscriptionPage.tsx
@@ -41,10 +41,10 @@ import { api } from "~/utils/api";
 import { CONTACT_SALES_URL } from "../../../ee/licensing/constants";
 import {
   type BillingInterval,
+  buildEnterprisePlanFeatures,
   buildPlanCapabilities,
   type Currency,
   FREE_PLAN_FEATURES as DEVELOPER_FEATURES,
-  ENTERPRISE_PLAN_FEATURES,
   formatPrice,
   getAnnualDiscountPercent,
   getGrowthFeatures,
@@ -360,11 +360,12 @@ export function SubscriptionPage() {
           seatCount: plan?.maxMembers ?? 1,
           perSeatPrice: monthlyEquivalent,
         };
-  // A plan the customer already holds is described by what it grants. Only a
-  // plan we are selling gets marketing copy, which is why a license-sourced
-  // plan is read from its own numbers rather than handed another tier's list.
+  // A plan the customer already holds is described by what it grants: the
+  // enterprise list minus anything their contract withheld, and every other
+  // held plan read from its own numbers rather than handed another tier's
+  // marketing copy. Only a plan we are selling gets the tier's full pitch.
   const currentPlanFeatures = isEnterprisePlan
-    ? ENTERPRISE_PLAN_FEATURES
+    ? buildEnterprisePlanFeatures(plan)
     : isLicenseOverride || isTieredLegacyPaidPlan
       ? buildPlanCapabilities({
           maxMembers: plan?.maxMembers ?? 0,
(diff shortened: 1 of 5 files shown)


langwatch-agent:
## Human Review Brief

**Mode: Targeted Review.** Follow-up to two P2 comments left on #6847 after it merged. No linked issue; the two comment links are the record, which is enough at this size.

### Decisions being ratified

1. **`contactSalesUrl` is supplied only when the plan is enterprise *and* did not come from a license.** The single `isEnterprisePlan` flag stops being the thing that decides upsell. Every future consumer of that flag now has to know it does not mean "can be sold to".
2. **A withheld entitlement is no longer advertised as included.** The page reads what the plan actually grants rather than what the tier nominally includes, which makes the entitlement map the source of truth for the copy.

### Must Check

- `CurrentPlanBlock`: confirm the subscription-sourced enterprise case still receives the URL. The condition gained a leg, and that is the case that must not change.
- A license that *does* grant webhooks should still advertise them. The fix is easy to write as "licensed means hide it" rather than "withheld means hide it".

### Ask Author

Is there an issue to link, or do the two `#6847` comments stand as the record? Either is fine, but the PR closes nothing today.

`Probably Fine`: the surface is one settings page and the blast radius is copy plus one CTA.

PR #6877, opened 2026-08-12
fix(automations): keep the drawer, its toast, and the graph list in step with what was just saved

### LINKED ISSUE #6716: bugbash(automations): defects from the 04 August sessions

## Defects

## UX gaps and proposals

## Record

0xdeafcafe:
# Related Issue

- Resolves #6716

Part of #6716 (WS-1 — drawer save/cache correctness). Three defects, one theme: a save that leaves the app holding a stale copy of what it just wrote.

## B4 — editing an automation lost the value that was saved (P0-adjacent)

`AutomationDrawer` hydrates its edit state from `automation.getTriggerById` **once per open** and deliberately ignores every later read, so a background refetch cannot overwrite an author's keystrokes mid-edit. Upsert success invalidated `getTriggers`, `graphs.getAll` and `graphs.getById` — never `getTriggerById`. Reopening therefore served the pre-save copy, the hydration gate latched onto it, and the corrective refetch was (by design) ignored: the field showed the value the author had just replaced while the stored automation held the new one.

Fix: invalidate `automation.getTriggerById` on upsert success. The hydration gate is untouched — it is the right behaviour, and it is precisely what makes the missing invalidation fatal.

## F14 — the creation confirmation now links to the automation

Saving closes the drawer and leaves the author wherever they started, usually not the automations list. The toast is the one moment the app knows exactly which row was written, so it carries a `View automation` action that opens the view drawer for the returned row (`View alert` / `View schedule` for those presets). Create only; an edit already knows where it is.

## B3 — the composer's graph list went stale after a graph was created

`graphs.create` invalidated `graphs.getById` only, while every picker that offers "which graph?" — the alert drawer's among them — reads `graphs.getAll`. The graph the author had just made was missing from the alert they came to write, with no explanation on screen and a page reload as the only way through. `graphs.getAll` is now invalidated on both create and update (a rename left the same list showing the old name).

## Spec

`specs/automations/authoring-drawer.feature` gains two rules and three `@integration` scenarios, each bound by a `@scenario` annotation on its covering test:

- "Editing an automation shows the values that were last saved"
- "A newly created graph is offered to a new alert without reloading"
- "The creation toast links to the created automation"

`pnpm check:feature-parity` → `specs/automations/authoring-drawer.feature 12/12 scenarios bound ✓`.

## Test evidence

Both regression tests execute the real path — they prime the copy the app is holding, drive the save through the component, remount, and assert what the next open shows. Neither asserts on a string of generated output, and both were verified to **fail** with the fix reverted:

| Test | With fix | Fix reverted |
|---|---|---|
| `AutomationDrawer.integration.test.tsx` (26 tests) | 26 passed | 1 failed — reopened drawer showed `Weekly error digest`, not `Monday quality digest` |
| `newGraphOfferedToAlert.integration.test.tsx` | 1 passed | 1 failed — the new graph was never offered in the alert's graph select |

```
pnpm test:integration run src/features/automations src/pages/[project]/analytics/custom --watch=false
  Test Files  16 passed (16)       Tests  154 passed (154)

pnpm test:unit run src/features/automations --watch=false
  Test Files  15 passed (15)       Tests  268 passed (268)
```

Diagnostics on all four touched files: clean (tslsp). Biome: zero new warnings (18 before, 18 after on the two production files; the new test files are clean).

## Not in scope

`SubjectSection.tsx`, `ConditionBuilder.tsx`, `FacetSection.tsx`, `components/ui/toaster.tsx`, `routers/automations.ts` and `CustomGraph.tsx` are untouched — the toaster's action support and the router's returned row were consumed as they are. The second half of F14 ("what is View *for* vs Edit?") is an open product question and is deliberately left alone.

## How it was verified

Rebased onto current `main` (the branch predated the Prisma 5→7 / tRPC 10→11 upgrade in #6697), resolving two merge conflicts in `AutomationDrawer.tsx` and its integration test — both were the same seam: main renamed `api.useContext()` to `api.useUtils()`, and this PR's edits kept the `openDrawer` addition alongside it. A third file, `newGraphOfferedToAlert.integration.test.tsx`, didn't conflict (it's new) but needed the same `useContext` → `useUtils` mock rename, plus its `TriggerAction` import switched from `@prisma/client` to `~/generated/prisma/client` (the custom generator output main now uses) — both are the well-known absorb-fix classes from the #6697 upgrade, not new defects.

Re-ran the full scoped suite standalone against merged `main`:

```
pnpm test:unit run src/features/automations --watch=false
  Test Files  15 passed (15)       Tests  268 passed (268)

pnpm test:integration run src/features/automations --watch=false
  Test Files  15 passed (15)       Tests  153 passed (153)

pnpm test:integration run "src/pages/[project]/analytics/custom" --watch=false
  Test Files  1 passed (1)         Tests  1 passed (1)
```

154 integration tests across 16 files and 268 unit tests across 15 files — matching the counts above. `pnpm check:feature-parity` still reports `specs/automations/authoring-drawer.feature 12/12 scenarios bound ✓`. Biome on the four touched files: 18 warnings (pre-existing complexity/line-count warnings on `index.tsx`), zero errors, no new warnings. Targeted `tsgo --noEmit --ignoreConfig` on the four files surfaced only expected single-file-mode noise (missing path aliases, JSX flag) with no genuine new errors; CI's `typecheck` / `typecheck:tests` are the authoritative whole-project signal.

## Summary by CodeRabbit

* **New Features**
  * Success messages for newly created automations now include a direct “View automation” link.
  * Newly created custom graphs are immediately available when configuring automation alerts.

* **Bug Fixes**
  * Reopened automations now display the latest saved values.
  * Newly created or renamed graphs now appear with up-to-date information in graph selectors.


0xdeafcafe:

PR #6881, opened 2026-08-12
feat(api): hold /api/triggers to the rules the dashboard is held to

### LINKED ISSUE #6716: bugbash(automations): defects from the 04 August sessions

0xdeafcafe:
# Related Issue

- Resolves #6716

**Stacked on [#6876](https://github.com/langwatch/langwatch/pull/6876)** (`issue6716/ws8-rest-actionparams-redaction`), which is this PR's base. Once #6876 merges, GitHub re-targets this one to `main` and the diff becomes slice 2 alone. Review #6876 first.

## What changed

An automation written over the API is the same row the dashboard writes and the same row the dispatcher reads, so `/api/triggers` now saves through a service that applies the dashboard's rules rather than writing whatever the wire schema accepted.

**The delivery configuration is read by the schema its channel publishes.** An email automation states its recipients, a Slack one a destination it can post to, a dataset one the dataset and how a trace maps onto it, a customer endpoint an https address on the default port. Anything else is refused as `invalid_action_params` with the field at fault named. The read happens *after* the redaction placeholders are resolved, so writing back what you read still means "keep the stored credential" — the contract slice 1 documented.

**The webhook channel is gated.** It is available to projects that deliver on it and refused for the rest with `trigger_channel_not_enabled`, so the channel opens over the API the same day it opens in the dashboard (ADR-040 §7). `SEND_WEBHOOK` is expressible on this surface for the first time, which is what that gate makes safe.

**Conditions are sanitised the way the dashboard sanitises them.** An automation whose conditions all name something this platform no longer filters on is refused as `trigger_filters_unsupported`, rather than saved with nothing left to narrow it. The wire schema now validates condition values instead of accepting any record.

**A new notification automation starts on the digest cadence a dashboard-authored one starts on** (ADR-026), so a broad condition cannot turn into a message per matching trace. The cadence rule moved into the app layer, where every write path can reach it; the dashboard router keeps its own copy until that file is free to change (noted below).

**Reads and writes go through the app layer's trigger service** rather than Prisma, per the routes → services rule. An automation that is not this project's answers as `trigger_not_found`, the same as one that never existed.

**Doc fix:** the listing described itself as returning active triggers while it has always included paused ones. It now says what it returns, and a test holds it there.

### Wire-contract changes worth calling out

- `actionParams` is validated per channel. A call that previously created a dataset automation with only a `datasetId` is now refused until it states its mapping — that is the bypass this slice closes, and the dashboard has always required it.
- `action` accepts `SEND_WEBHOOK` (gated, above).
- `filters` values are validated at the wire.
- Not-found answers carry the `trigger_not_found` code rather than prose.
- The channel a trigger delivers on stays fixed for its lifetime. An update states a delivery configuration for the channel already stored, which is what lets the credential rules read the incoming and stored halves as belonging to one provider.

## New named failures

`trigger_channel_not_enabled`, `trigger_filters_unsupported`, `trigger_not_found` — each listed in `logic/codes.ts` and given customer-facing copy in the presentation registry.

## Spec

`specs/automations/public-api.feature` — 22 scenarios, all tagged and bound (`check-feature-parity.ts` reports `22/22 scenarios bound`), including the six new ones under "The API saves what the dashboard would accept".

## How it was verified

Run on this branch after merging the base branch (`issue6716/ws8-rest-actionparams-redaction`), which carries `main` — including the Prisma 7 / tRPC 11 upgrade — into this slice.

```
pnpm test:unit run src/server/app-layer/automations/__tests__
  Test Files  21 passed (21)   Tests  257 passed (257)

pnpm test:unit run src/features/errors/logic/__tests__
  Test Files  10 passed (10)   Tests  233 passed (233)

pnpm check:feature-parity          specs/automations/public-api.feature — 22/22 scenarios bound
pnpm check:openapi-completeness    OK: every unsuppressed operation is completely described.
pnpm check:openapi-route-coverage  OK: every registered route is documented or accounted for.
```

Type diagnostics are clean on every file this PR touches (tsgo over the touched set). Biome reports no new diagnostics on the touched directories — the 15 warnings in `src/server/app-layer/automations` are all pre-existing complexity warnings on files this PR does not touch.

`pnpm test:integration run src/app/api/triggers/__tests__` (28 tests over 3 files) does not complete on a local developer database: the shared dev Postgres is behind on migrations and every case fails on `The column Trigger.pausedReason does not exist in the current database`. That is environment drift, not this branch — CI runs the same suite against a freshly migrated database.

## Still to come in this workstream (slice 3)

The parity half is not in this PR: the full upsert shape with typed per-action `actionParams` as a discriminated union, `filterQuery`, `customGraphId`/`graphAlert`, `report`, `templates`, `notificationCadence`, `traceDebounceMs`; the test-fire, fire-history and enable/disable verbs; the regenerated OpenAPI document and SDK client; the MCP and CLI surfaces that follow it; `feature-map.json`; and a one-call path for retargeting a webhook destination while re-supplying its headers, with copy an API caller can act on.

Refs #6716


0xdeafcafe:

PR #6882, opened 2026-08-12
fix(automations): default Slack bot delivery to Block Kit, plus WS-4 provider and template fixes

### LINKED ISSUE #6716: bugbash(automations): defects from the 04 August sessions

0xdeafcafe:
# Related Issue

- Resolves #6716

## Summary

WS-4 from the automations/alerts bug-bash plan, plus the #6716 P0 item pulled forward per the plan's Phase 0 guidance.

**#6716 P0 — trace notifications must show what happened, not just an id.** Audited `packages/automations/src/templating/defaults.ts` and `exampleContext.ts`: the default trace templates (email, Slack plain text, Slack Block Kit, webhook) already interpolate the matched trace's input and output — that part had already landed and stuck. What was missing was enforcement: the spec scenario only required "a link", and no test asserted the content itself, so it could regress silently. Extended the existing default-rendering tests to assert on the input/output text and tagged the covering spec scenario (`notification-templates.feature` carried zero binding tags before this change).

**#6716 — bot-token Slack delivery fell back to plain text instead of Block Kit.** Every dispatch call site computed `templateType` as `slackTemplateType === "block_kit" ? "block_kit" : "string"`, so an author who never customised the Slack message — the common case — got the legacy plain-text builder even on a bot connection, contradicting ADR-041 ("never the legacy plain-text builder" for a bot connection). Added `resolveSlackTemplateType()` as the single place that decides the fallback (bot → Block Kit, webhook → string, unchanged), and applied it at the three real dispatch call sites plus test-fire.

**#6716 — Slack app manifest scopes.** Audited every Slack Web API call the delivery layer makes (`chat.postMessage`, `conversations.list`) against the copy-paste manifest. Already correct (a prior partial fix per #6254) — added a test that pins it so a new call can't drift silently without a manifest update.

**B2 — Slack channel Reload.** Fixed the three candidate causes pinned in the plan: a failed listing now names its cause in the hint line instead of only logging to the console; a manual reload answered `no_token` now tells the author to add a token instead of showing nothing; Reload resets the per-key fetch latch so a token that became valid after an earlier failed load is retried.

**F9/F10/F11/F8/F12/G4 — composer polish**, per the plan:
- F9: the bot-token setup block is one sentence + a "Where do I get a bot token?" disclosure holding the full steps.
- F10: the manifest step says to pick the YAML format.
- F11: test fire moved to the end of the Slack form and the email form, after the message section.
- F8: layout thumbnails carry representative example text instead of flat colour bars.
- F12: picking a cross-cadence layout from the expanded disclosure no longer regroups the gallery — the grouping is latched to the cadence the picker last settled on (see post-review fix below).
- G4: a plain-language caption under "Send a test" says it delivers a real message using example data.

## Post-review fixes

Three issues found across two rounds of review, all fixed on this branch:

1. **[P1] F12's mount-latch went stale on a cadence change made outside the picker.** The original latch was a plain `useRef(cadence)` set once at mount and never updated. The composer's facets are independently open (not a single-open accordion), so an author can change cadence from the separate Cadence section while the picker stays mounted — after that, the primary grid kept showing the old cadence's layouts while `introFor`/`pickDefaultSlackBlockKitTemplateId` (which read the live `cadence` prop) described the new one, and picking a primary-grid card (which calls `onSelect`, not `onSelectOtherCadence`) never reconciled the mismatch — a digest draft could save an immediate-shaped template. Fixed by moving the latch to state with a resync effect that regroups on any cadence change *except* one this picker itself just caused via `onSelectOtherCadence` (tracked with a marker ref set immediately before that call). New test: mount, change cadence externally, assert the grid regroups — then, in the same interaction, make an in-picker cross-cadence pick and assert the list stays stable.

2. **[P1] Residual staleness in the fix above: the self-initiated marker went stale on two consecutive same-cadence picks.** Picking two different digest layouts in a row without leaving the still-open "more layouts" disclosure is normal comparison-shopping — the first pick genuinely changes the live `cadence` prop (marker written, then consumed by the resync effect); the second does not (already on that cadence), so the effect's dependency array never fires and the marker is never consumed. A later, genuine external cadence change that happens to land back on that same value then reads the stale marker as self-initiated and wrongly skips the regroup — reintroducing the original bug. Fixed by guarding the marker write: only mark a cadence as self-initiated when the pick actually changes it (`if (otherCadence !== cadence) selfInitiatedRef.current = otherCadence`). New test exercises exactly this sequence (two in-picker picks landing on the same cadence, then an external change back to it) and was verified to fail against the unguarded code and pass with the guard.

3. **[P2, narrow ownership grant] A fourth dispatch site had the same original bug.** `server/app-layer/reports/report-dispatch.ts:285-286` (scheduled-reports Slack path, bot connections included) had the identical `slackTemplateType === "block_kit" ? "block_kit" : "string"` ternary. That file is WS-5's; this PR touches only this one hunk (the `templateType` computation at both the bot and webhook branches), replacing it with `resolveSlackTemplateType()` — nothing else in the file. WS-5's own diff there is scoped to the unrelated `viewUrl` fix, so there's no overlap.

## How it was verified

Re-run after merging the latest `main` into this branch (one conflict, in
`report-dispatch.unit.test.ts`: `main`'s Prisma 7 client path
`~/generated/prisma/client` kept, alongside this branch's hoisted `vitest` import).

```
# templating package — the Slack renderer and the default templates
packages/automations $ pnpm test:unit src/templating
# 12 test files, 165 tests passed

# the dispatch call sites and the shared resolver
platform/app $ pnpm test:unit run \
  src/server/app-layer/reports/__tests__/report-dispatch.unit.test.ts \
  src/server/app-layer/automations/__tests__/trigger-template.service.unit.test.ts \
  src/server/app-layer/automations/dispatch/__tests__/graphAlertActionDispatch.unit.test.ts \
  src/features/automations/providers/slack/__tests__/manifest.unit.test.ts
# 4 test files, 77 tests passed

# wider sweeps over the areas the merge touched
platform/app $ pnpm test:unit run src/server/app-layer/automations src/server/app-layer/reports
# 33 test files, 424 tests passed
platform/app $ pnpm test:unit run src/features/automations
# 15 test files, 269 tests passed

# the composer UI — Slack provider, template picker, wireframes, test fire, email, webhook
platform/app $ pnpm test:integration run src/features/automations/providers --watch=false
# 6 test files, 85 tests passed

platform/app $ pnpm check:feature-parity
# OK — 6366 enforced scenarios bound; no new unbound scenarios
```

- [x] Unit and integration suites over every path this diff touches — **1020 tests passed, 0 failed**
- [x] `pnpm check:feature-parity` — clean; `notification-templates.feature` now enforces (this PR removes it from `LEGACY_INERT`)
- [x] tslsp diagnostics on all 22 touched TypeScript files — no diagnostics
- [x] biome, directory-scoped, compared warning-for-warning against the same files on `origin/main` — **0 new warnings** (the complexity warnings that remain are all pre-existing at identical positions)
- [ ] Manual end-to-end verification against a real Slack workspace — not reproducible in this environment; flagged below

## What still needs manual verification

- **B2 (Slack channel Reload after app reinstall):** fixed per the plan's three candidate causes and covered by component-level tests with a mocked mutation, but the actual "reinstall the app, come back, click Reload" flow needs a real Slack workspace to confirm end-to-end.
- **Composer live preview (`AutomationDrawer.tsx`)** has the identical `templateType` selection bug fixed at the four dispatch call sites (`templateType: templates.slackTemplateType === "block_kit" ? "block_kit" : templates.slackTemplateType === "string" ? "string" : null` around line 714) — so a bot-connected draft's *preview* still shows plain text by default even though delivery now sends Block Kit. `AutomationDrawer.tsx` is explicitly out of this PR's file ownership (WS-1 owns it); it needs the same one-line fix using the `resolveSlackTemplateType()` export.
- `triggerSettlementIntentHandlers.ts`'s own test file doesn't mock Slack delivery at all (only `deliverWebhook`/`triggerEmail`), so the trace-trigger dispatch call site isn't covered by a dedicated test the way the other three call sites are — coverage there rests on the shared resolver's own thorough unit tests plus the identical fix pattern.

Refs #6716

## Summary by CodeRabbit

* **New Features**
  * Slack bot notifications now default to Block Kit, while webhooks use plain-text templates unless configured otherwise.
  * Email and Slack notifications include matched trace input and output excerpts.
  * Slack channel setup provides clearer guidance, error-specific hints, and a Reload option.
  * Test-send controls now explain that real messages use example data.
  * Slack template galleries preserve layout order when switching cadences.
  * Slack previews display representative content instead of placeholder bars.
* **Bug Fixes**
  * Improved Slack channel loading and template selection reliability.


0xdeafcafe:

PR #6883, opened 2026-08-12
fix(analytics): apply per-series filters and percentage mode in the ClickHouse builder

### LINKED ISSUE #6718: fix(analytics): traces with errors and traces without errors return identical data
In the analytics graph builder, "traces with errors" and "traces without errors" render the same data. They are opposites, so they should never match. One of them is wrong and right now there is no way to tell which from the chart alone.

The suspected cause is a change to the Elasticsearch aggregation behind these two series. Confirm that before changing anything.

## Why this is filed on its own

This turned up during the live automations and alerts bug bash on 2026-08-04, but it is an analytics defect, not an automations one. **Nobody was given this in the call.** It is filed separately so it does not get quietly absorbed into the automations lane and then sit there unowned. It needs an owner of its own.

The automations defects from the same call are in langwatch/langwatch#6716 bugbash(automations): eleven defects from the 04 August team session.

## Impact

Anything built on top of these two filters inherits the error. That includes saved graphs, dashboards, and any alert or automation whose trigger reads one of these series. An alert on "traces with errors" that is really counting all traces will fire on healthy traffic.

Filed by Nova from the call recording. The call itself produced no written record.

0xdeafcafe:
Closes #6718. Part of the #6716 follow-up (WS-7 of the Automations & Alerts bug-bash plan).

## The bug

Two series that differ only by `filters: {"traces.error": ["true"|"false"]}` compiled to **byte-identical SQL**. The per-series loop in `aggregation-builder.ts` passed only `metric/aggregation/index/key/subkey` on, and the only filter translation was the graph-wide one — so the "with errors" and "without errors" lines were the same line, and every alert built on a filtered series watched the wrong number.

The filter definitions and the ClickHouse predicates were already right (`server/filters/registry.ts`, `filter-translator.ts` — `ContainsErrorStatus = 1` vs `= 0 OR IS NULL`), the UI already sent per-series filters, and the schema already accepted them. Only the builder ignored them. This is an Elasticsearch → ClickHouse regression: the deleted ES builder implemented it as a `__filters` wrapper aggregation, and nothing covered it.

Two more defects sat on the same feature:

- **`asPercent` ("Show in percentage (%)") was read by nothing.** It flowed form → input → `registry.ts` → no consumer at all.
- **Grouping by `error.has_error` used the span-level `StatusCode`** through a `stored_spans` JOIN. The JOIN fans a trace out into one row per span, and a trace with an error almost always has healthy spans too, so a trace landed in **both** buckets and the buckets could not add up to the total. It also contradicted `field-mappings.ts`, which maps the field to the trace-level `ContainsErrorStatus`.

## The fix

**Per-series filters as conditional aggregation.** Each series' filters translate to a predicate that is folded into that series' own aggregate through the `-If` combinator, leaving the graph-wide `WHERE` shared:

| before | after |
|---|---|
| `sum(ts.TotalCost)` | `sumIf(ts.TotalCost, (pred))` |
| `count()` | `countIf((pred))` |
| `uniq(ts.TraceId)` | `uniqIf(ts.TraceId, (pred))` |
| `quantileTDigest(0.5)(x)` | `quantileTDigestIf(0.5)(x, (pred))` |
| `avgIf(es.Score, cond)` (already conditional) | `avgIf(es.Score, (cond) AND (pred))` |

**Which filter classes are supported.** All of them. Every handler in `filter-translator.ts` returns `requiredJoins: []` — span, event and evaluation facets are expressed as `ts.TraceId IN (SELECT … )` subqueries, not JOINs — so no filter class needs a JOIN to be scoped to a single series, and each compiles cleanly into an `-If` condition. A per-series filter that ever *did* need a JOIN now throws by name (a JOIN added for one series would narrow every other series in the graph), instead of shipping silently wrong numbers for the rest of the graph.

**All five builder paths carry it.** The three that aggregate directly over the scan (standard, `timeScale: "full"`, date-bucketed) take the raw `ts` predicate. The two that aggregate over a CTE (the arrayJoin / group-by path and the mixed evaluation path) hoist the predicate into the CTE as a per-trace boolean first, because the raw predicate reads an alias those outer `SELECT`s do not have — and the condition is applied *after* the dedup rewrite, so `count() → uniqExact(trace_id)` becomes `uniqExactIf(trace_id, …)` with the mapping intact.

**Per-entity (pipeline) series** take their filter in their own subquery's `WHERE` rather than as a conditional aggregate, because the filter decides which entities exist at all — an entity with no matching traces must disappear, not survive as a zero and drag the cross-entity average down. Each such metric already owns a private CTE, so this touches no other series.

**Non-additive aggregates no longer fabricate zeros.** ClickHouse's `-If` combinators do **not** return NULL for an empty match set when the source column is non-nullable — `minIf` returns `0` (verified: `SELECT minIf(x, 0) FROM (SELECT toUInt32(200) AS x)` → `0`, type `UInt32`). Left alone, a filtered latency floor would draw a real-looking 0ms line for every bucket the filter excluded, where `_timeseries-row-parser.ts` and Elasticsearch both leave a non-additive series absent. Series that are non-additive per `isZeroWhenAbsentSeries` now carry an emptiness guard — `if(countIf(<cond>) > 0, minIf(x, <cond>), NULL)` — while additive series stay unguarded, because no matching rows genuinely *is* zero there.

**`asPercent`** matches ES semantics for non-pipeline series: `all > 0 ? filtered / all * 100 : 0`.

- Result is a **percentage (0–100)**, not a fraction — matching the toggle's label.
- An **empty bucket reads as 0**, not as "no value".
- Percentage mode on an **unfiltered** series is a **no-op** — matching ES, where the percentage wrapper only ever existed inside the per-series filter wrapper. There is nothing for an unfiltered series to be a percentage *of*. Pinned by a test asserting byte-identical SQL with and without the flag.
- Nothing on the render side needed changing: the series value simply becomes the percentage.
- **Pipeline percentages are refused rather than answered with a wrong number** — the one deliberate divergence from ES, which did support them (`c6e63df26f`). The unfiltered denominator would have to be computed over a *different* set of entities (a second scan, not a second aggregate), so the two halves stop being comparable. The refusal is the handled error `analytics_series_percentage_unsupported`, with customer copy in the presentation registry telling the author which toggle to drop.

**That refusal is not allowed to poison the alert lane.** `graph-trigger-evaluation.service.ts` rethrows anything it does not recognise, and a rethrow there is a redelivery — so a saved per-entity percentage series would re-fail on every delivery until it quarantined its tenant's whole graph-trigger lane, taking every *other* alert in the project down with it (the same failure mode the file's row-ceiling comment documents from the 2026-08-09 incident). It is now classified alongside `resultTooLarge`: skipped and logged with the trigger id, the series index and the series metric, never rethrown. The composer's save-time gate and the report path's per-panel catch are **coordinated with the streams that own those files** and are not in this PR.

**`error.has_error` group-by** now groups on `if(ifNull(ts.ContainsErrorStatus, 0) = 1, 'with error', 'without error')` and drops the `stored_spans` JOIN it never needed.

**Folded finding (validator sweep):** deleting a graph from a dashboard refetched only the page's own `{projectId, dashboardId}` query, so the alert composer — which reads `graphs.getAll` keyed by `{projectId}` alone — kept offering the deleted graph until a full reload. Now invalidates every `graphs.getAll` key.

## Spec

`specs/analytics/error-series.feature`, 13 scenarios, all tagged and bound (`pnpm check:feature-parity` → `13/13 scenarios bound ✓ all bound`), each with a `@scenario` annotation on its covering test.

## Test evidence

SQL-shape unit tests were extended into the existing `aggregation-builder.test.ts` (no parallel suite). The headline one: two series with opposite error filters must **not** compile to the same expression.

```
$ pnpm test:unit run src/server/analytics/clickhouse/__tests__/
 Test Files  13 passed (13)
      Tests  360 passed (360)

$ pnpm test:unit run src/server/analytics src/features/errors \
                    src/server/app-layer/analytics src/server/app-layer/automations
 Test Files  65 passed (65)
      Tests  1132 passed (1132)
```

Integration tests execute the built query against seeded ClickHouse data, because "the SQL is different" is not the same claim as "the numbers are right":

```
$ pnpm test:integration run src/server/analytics/clickhouse/__tests__/error-series.integration.test.ts --watch=false
 Test Files  1 passed (1)
      Tests  10 passed (10)

$ pnpm test:integration run src/server/analytics/clickhouse/__tests__/ \
                            src/server/app-layer/analytics/__tests__/ --watch=false
 Test Files  13 passed (13)
      Tests  122 passed | 3 skipped (125)
```

The integration suite seeds five traces (two with an error, three without; two carrying an `llm` span, one of each error status) plus three evaluation runs, and asserts at the query-result level that:

- the two error series report different counts summing to the window total;
- a span-facet filter narrows only its own series;
- a per-series filter narrows only its own series **on the grouped/arrayJoin CTE path** (checked per span-type group);
- a per-series filter narrows only its own series **on the mixed-evaluation CTE path**, leaving the evaluation score untouched — those two are the paths carrying the hoisted per-trace boolean;
- percentage mode returns the real share, and an empty window returns 0;
- a filtered `min` over a window that holds traces but matches none reports **no value**, not 0 — with a control asserting the same series still reports the real duration when the filter does match;
- the error group-by buckets partition the window exactly.

Lint delta was measured against `main` by restoring the baseline files and re-running the identical command: **3593 warnings on both**, zero new diagnostics, zero errors. `tslsp diagnostics` clean on every touched file; `pnpm format` clean.

## Query rules

`TenantId` stays the first predicate everywhere; the partition-key time range is unchanged; no `LIMIT 1 BY`; dedup still uses the IN-tuple form. Per-series predicates go through the same parameterized placeholders as the graph-wide ones (a test asserts the span filter's `spanTypes_0` param), and the columns those predicates read are threaded into `referencedTraceColumns` so the deduped trace subquery does not prune them away.

## Not in this PR

- **"Add alert" on the Analytics reports page** (WS-7 scope item 4) turned out to be **already shipped**: `components/analytics/reports/GraphCardHeader.tsx` renders an `Add alert` button (and an `Edit alert` bell when a trigger exists) on every saved graph card, opening the automation composer with `prefilledGraphId` + `prefilledSeriesName`. `reports.tsx → ReportGrid → DraggableGraphCard → GraphCardHeader` is the chain. Nothing to add.
- **The composer save-time gate** for "percentage on a per-entity series", and **the report path's per-panel catch** for the same refusal — coordinated with the streams that own those files.
- **The `graphs.getAll` invalidation fix has no test.** The reports page has no test harness (only `GraphCardHeader` does), and standing one up is a separate lift.

### DIFF
--- a/platform/app/src/features/errors/logic/codes.ts
+++ b/platform/app/src/features/errors/logic/codes.ts
@@ -23,6 +23,7 @@ export const APP_ERROR_CODES = [
   "ai_call_failed",
   "ai_query_provider_error",
   "already_organization_member",
+  "analytics_series_percentage_unsupported",
   "anomaly_rule_not_found",
   "api_key_already_revoked",
   "api_key_not_found",
--- a/platform/app/src/features/errors/logic/presentation.ts
+++ b/platform/app/src/features/errors/logic/presentation.ts
@@ -1157,6 +1157,17 @@ const presentations = {
       "This is a self-hosted deployment, so plans are managed outside the app.",
   },
 
+  // ---- analytics ----
+  analytics_series_percentage_unsupported: {
+    // A percentage is the series divided by the same measurement without the
+    // series' own filters. A per-entity measurement (average per user, sum per
+    // thread) also changes which entities exist once filtered, so the two
+    // halves stop being comparable — the author has to drop one of the two.
+    title: "This series can't be shown as a percentage",
+    describe: () =>
+      "Turn the percentage toggle off for this series, or remove its per user, per thread or per customer breakdown.",
+  },
+
   // ---- governance ----
   anomaly_rule_not_found: {
     title: "Anomaly rule not found",
--- a/platform/app/src/pages/[project]/analytics/reports.tsx
+++ b/platform/app/src/pages/[project]/analytics/reports.tsx
@@ -59,6 +59,7 @@ function ReportsContent() {
     { enabled: !!projectId && !!activeDashboardId },
   );
 
+  const queryClient = api.useContext();
   const deleteGraph = api.graphs.delete.useMutation();
   const updateLayout = api.graphs.updateLayout.useMutation();
   const batchUpdateLayouts = api.graphs.batchUpdateLayouts.useMutation();
@@ -90,7 +91,11 @@ function ReportsContent() {
       { projectId, id: graphId },
       {
         onSuccess: () => {
-          void graphsQuery.refetch();
+          // Invalidate EVERY graphs.getAll key, not just this dashboard's.
+          // The alert composer reads the list keyed by {projectId} alone, so
+          // refetching only the {projectId, dashboardId} query left a deleted
+          // graph on offer there until a full page reload.
+          void queryClient.graphs.getAll.invalidate();
         },
         onError: () => {
           toaster.create({
--- a/platform/app/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/platform/app/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -950,6 +950,41 @@ describe("aggregation-builder", () => {
         expect(result.sql).toContain("AND ((ts.ContainsErrorStatus = 1))");
       });
 
+      // ClickHouse's `-If` combinators do NOT return NULL for an empty match
+      // set when the source column is non-nullable — `minIf` returns 0. A
+      // filtered latency floor would then draw a real-looking 0ms line for
+      // every bucket the filter excluded, where the row parser and ES both
+      // leave a non-additive series absent.
+      describe("when the aggregation is not additive", () => {
+        it("reports no value instead of zero for a bucket nothing matched", () => {
+          const result = buildTimeseriesQuery({
+            ...baseInput,
+            series: [
+              {
+                metric:
+                  "performance.completion_time" as FlattenAnalyticsMetricsEnum,
+                aggregation: "min" as const,
+                filters: { "traces.error": ["true"] },
+              },
+            ],
+          });
+
+          expect(result.sql).toContain(
+            "if(countIf((ts.ContainsErrorStatus = 1)) > 0, min",
+          );
+          expect(result.sql).toContain(", NULL) AS 0__");
+        });
+
+        it("leaves an additive aggregation unguarded, because zero is the truth there", () => {
+          const result = buildTimeseriesQuery({
+            ...baseInput,
+            series: [errorSeries(["true"])],
+          });
+
+          expect(result.sql).not.toContain("NULL) AS 0__");
+        });
+      });
+
       describe("when the series is also shown as a percentage", () => {
         // @scenario "Percentage mode on an unfiltered series leaves the series unchanged"
         it("leaves an unfiltered series exactly as it was", () => {
--- a/platform/app/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
+++ b/platform/app/src/server/analytics/clickhouse/__tests__/aggregation-builder.test.ts
@@ -834,6 +834,207 @@ describe("aggregation-builder", () => {
       });
     });
 
+    // @regression issue #6718: two series differing only by their `filters`
+    // compiled to byte-identical SQL, because the per-series loop passed only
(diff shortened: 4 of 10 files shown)


0xdeafcafe:

PR #6890, opened 2026-08-12
fix(scenarios): re-vendor @langwatch/scenario 1.1.0 for UI-triggered runs

sergioestebance:
## What

Re-vendor `@langwatch/scenario` from `1.0.0` → published `1.1.0`.

## Why

The platform vendors scenario `1.0.0`, which ends a scenario run early when the
judge issues an **unforced** `finish_test(inconclusive)` on a non-final turn.
For UI-triggered runs this surfaces as *"the simulated user stopped responding"*
and the conversation halts partway.

`1.1.0` (fix from `langwatch/scenario#886`) treats an unforced inconclusive
finish as `continue_test`, so the run proceeds instead of terminating.

## Change

- `platform/app/vendor/`: swap `langwatch-scenario-1.0.0.tgz` → `langwatch-scenario-1.1.0.tgz`
- `platform/app/package.json`: bump the `file:` ref
- `pnpm-lock.yaml`: re-pin to the 1.1.0 tarball
- vendor `README.md`: update the current-file line

## Verification

- Installed `node_modules/@langwatch/scenario` reports version `1.1.0`
- The `completionCalledTool` fix marker is present in the installed `dist`
- Lockfile integrity re-pinned to the 1.1.0 artifact

Structural only — not yet exercised against a live UI-triggered run on a rebuilt
instance.

Ref langwatch/langwatch-saas#1036

## Summary by CodeRabbit

* **Improvements**
  * Updated the scenario tooling to the latest available release, providing access to the newest stability and compatibility improvements.
  * Refreshed related package references to keep the application environment consistent and up to date.


drewdrewthis (approved):
lgtm

PR #6891, opened 2026-08-12
feat(api): public API parity for automations — typed delivery, the full upsert shape, and the verbs

0xdeafcafe:
Third and last slice of WS-8. Builds on #6881 (slice 2), which is its base — review that one first; this diff is only what slice 3 adds. Refs #6716.

Slice 1 stopped `/api/triggers` handing out delivery credentials. Slice 2 put its writes through the service layer, so the API is held to the rules the dashboard is held to. This slice makes the API able to *say* what the dashboard says — and answer, rather than shrug, when a caller says something it cannot do.

## What an integrator can now write

- **Typed delivery, per channel.** The create body is discriminated by `action`, so the published OpenAPI says what a Slack automation needs that an email one does not, instead of `object`. A unit test holds each published schema to the field names its channel actually reads.
- **The rest of the composer's shape:** `filterQuery`, `customGraphId` + `graphAlert`, `report`, `templates`, `notificationCadence`, `traceDebounceMs`. A graph alert written over the API is the row the drawer hydrates its form from — asserted through the dashboard's own read and parser.
- **Verbs:** `POST /:id/test-fire`, `GET /:id/fires`, `POST /:id/enable`, `POST /:id/disable`.

## Nothing is silently dropped or silently kept

Four things used to answer 200 and not mean it:

1. **A test fire was uncapped.** It is the one verb that makes LangWatch send on the caller's say-so, to an address the same caller chose: an email through the shared provider, or a request from our workers at an arbitrary destination (ADR-040 §4). The dashboard caps that at 10 a minute per user; over the API, setting an automation's recipients and firing it in a loop was an outbound mailer. Now capped per project on the same window — a project API key is the identity behind the call — with **Slack exempt**, as on the dashboard, because its destination is pinned to Slack's own hosts. The service comment that credited the saved destination with the anti-relay property now says what actually bounds it.
2. **A field the channel does not have was dropped.** `slackChannelID` saved cleanly and delivered nowhere; on an update, where the payload replaces the stored configuration whole, it was worse. It is now refused, naming what was sent and what the channel reads. The payload is read against the channel on the **stored row**, not against whichever shape it resembles — which is what lets a passthrough union be safe in the first place.
3. **Another channel's field could park in the row.** Read as part of the rule, it persisted and echoed, and a later dashboard conversion of the channel could adopt it as live encrypted configuration. The refusal above closes the path; the credential split now also reads the **union of every channel's field names**, exactly as the slice-1 module doc prescribed, so it cannot happen by another route either. Module doc updated to match.
4. **A rule sent inside `actionParams` was overwritten by the stored one.** Read → change `threshold` → write back returned 200 with the threshold unchanged. It is refused with the field it belongs in (`graphAlert` / `report`), and — so the refusal is never hostile to a well-behaved client — the **read now returns the rule in that same field**, with `actionParams` carrying only where the message goes. Writing the response back means what it says.

## Two things are fixed by design

- **The delivery channel.** Keeping a stored credential across a save is only sound while the incoming and stored halves belong to the same channel. The update schema accepts `action` so writing the read response back is *answered* — a different channel is refused (`trigger_action_immutable`), not dropped.
- **The kind.** An alert owns its graph's alert slot and a report owns a calendar entry, so converting one over the API is a create and a delete (`trigger_kind_immutable`).

## The retarget path

Pointing a webhook automation somewhere new used to answer "re-enter the header values" — values an API caller never held, because the read hands it `[redacted]`. It now says to send them with the new destination; one call carrying both saves both, and there is a test for exactly that call.

## Convergence and consistency

- The cadence rule has one home in the app layer; `routers/automations.ts` no longer keeps a private copy.
- The webhook channel's release flag is resolved **per project** on every surface. It was `distinctId: userId` in tRPC and `distinctId: projectId` in REST, so during a partial rollout the surfaces could disagree — and whether a project has a channel cannot depend on which teammate is asking.

## Agent-ready clients

- MCP `platform_create_trigger` sent no `actionParams` at all, so every call it made was refused. It now takes the typed configuration, plus `platform_get_trigger` and `platform_test_fire_trigger`. MCP trigger types are inferred from one set of zod schemas and reads go through them, so a deployment that stops sending a field a tool renders fails loudly instead of printing `undefined` into an agent's context.
- Both MCP write tools take `report` as well as `graphAlert`, structured rather than as a loose record, so an MCP client can create and edit a scheduled report instead of only reading one.
- CLI: `--action-params` / `--filter-query` on create and update, and `trigger enable | disable | test-fire | fires`.
- `feature-map.json` matches what exists: the Go SDK has no triggers surface and no longer claims one; the API reference is registered.

## Errors

Eleven new codes, each with customer-facing copy in the presentation registry: `trigger_action_immutable`, `trigger_kind_immutable`, `trigger_filter_query_invalid`, `graph_alert_incomplete`, `graph_not_found`, `report_channel_unsupported`, `report_incomplete`, `webhook_header_values_required`, `trigger_action_params_unknown_fields`, `trigger_rule_fields_misplaced`, `trigger_test_fire_rate_limited`.

Each refusal names the thing that is wrong: a report whose stored configuration can no longer be read asks for its source and schedule rather than for a different channel, and the test-fire cap reads `meta.resetAt` to say how long to wait.

## Said where the reader is

The `actionParams` contract — the redaction, what sending `[redacted]` back means, and that the automation's rule is **not** in that field — is in `.describe()` on both the read and the write, so it reaches the published document, the generated client and the API reference rather than only this repo's TSDoc. That reader is exactly the one who would otherwise put the rule in `actionParams` and take a 422.

## How it was verified

Re-verified after merging the base branch (#6881 at `94d1f3d2ae`), which carries `main` — including the Prisma 7 / tRPC 11 upgrade and the slice-1 redaction work. Three files conflicted, all in their import blocks, and all resolved the same way: the Prisma types now come from the generated client (`~/generated/prisma/client`) rather than `@prisma/client`, keeping this slice's own imports alongside. Two seams the merge could not flag were repaired by hand — the new test files still imported `@prisma/client` directly, and the shared `enable`/`disable` handler needed its context typed to the path it is mounted on, since Hono's parameter type is only as narrow as the route it is given.

- **App unit suites** — `src/app/api/triggers/__tests__` + `src/server/app-layer/automations`: **416 green across 33 files**; the error registry and the api-reference guards (`src/features/errors`, `scripts/__tests__`): **455 green across 26 files**, covering the per-channel delivery schemas, the redaction split over the union of every channel's field names, and the cadence rule in its one home.
- **MCP** — the full `src` unit suite: **554 green across 32 files**, including the 11 that hold the trigger tools to the shared zod schemas.
- **CLI** — the TypeScript SDK's CLI suites: **375 green across 32 files**, including the boot-graph and feature-map drift guards.
- **Published document** — `pnpm check:openapi-completeness` green (40 operations, 9 exemptions, 0 known gaps) and `check:openapi-route-coverage` green (259/353 registered routes documented, 36 unpublished on purpose). Re-running the generators — `task generateOpenAPISpec`, `docs make sync-api-spec && make generate-api-reference`, and the SDK's `generate:openapi-types` — reproduces the committed spec, docs copy, api-reference pages and client types byte for byte.
- **Spec** — `specs/automations/public-api.feature`: **47/47 scenarios bound**.
- **Types** — the app project, the app test project, the TypeScript SDK and the MCP package all typecheck clean on the merged tree.
- **Lint** — Biome over the touched app directories: no new warnings against the base branch.
- **Integration** — the REST integration suites (`trigger-parity`, `trigger-write-rules`, the automations router and test-fire suites) are not runnable in this checkout: no container runtime, and the shared dev database has migration drift. They run on clean databases in this PR's CI.

## Decisions worth a second opinion

1. **Kind immutability over the API** — the dashboard can convert an automation into an alert; the API cannot. Conversion touches the graph's unique alert slot and the report calendar.
2. **Test fire delivers to the automation's own saved destination**, never a caller-supplied one — and the cap, not the saved destination, is what bounds it.
3. **The read publishes `graphAlert` / `report` as top-level fields.** That is what makes refusing a rule inside `actionParams` a contract rather than a trap: what a write states, a read states back.


0xdeafcafe:

Issue #6327, opened 2026-07-29
15 integration suites are permanently skipped in CI by a guard that never means what it says


Aryansharma28:
15 integration suites carry a `describe.skipIf(isTestcontainersOnly)` guard that, in CI, is always `skipIf(true)`. They have not run in CI since the guard was added, and none of them appears to need ClickHouse at all.

This is the third time the same trap has been found — #5988 first, PR #6288 second (`langySessionKey.integration.test.ts`, which had silently rotted into an assertion that regex-matched an object after a return-shape change). Two instances is an accident; three is a pattern worth fixing at the root.

## Why the guard always fires in CI

```
.github/workflows/langwatch-app-ci.yml:299
  CI_CLICKHOUSE_URL: "http://default:ci_password@localhost:8123/test_langwatch"
        ↓  (job-level env, so every integration shard has it)
src/server/event-sourcing/__tests__/integration/setupEnv.ts:60-64
  if (process.env.CI && process.env.CI_CLICKHOUSE_URL) {
    process.env.CLICKHOUSE_URL      = process.env.CI_CLICKHOUSE_URL;
    process.env.TEST_CLICKHOUSE_URL = process.env.CI_CLICKHOUSE_URL;   // ← here
  }
        ↓  (setupEnv is a `setupFiles` entry for the WHOLE lane — every file)
const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL;        // ← always true in CI
describe.skipIf(isTestcontainersOnly)(...)                             // ← skipIf(true)
```

The guard reads as "skip when we're in the testcontainers-only lane", but `setupEnv` is not per-suite — it runs for every integration file. So the variable it keys off is set for all of them, whether or not the suite has anything to do with ClickHouse.

The failure mode is the quiet one: a skipped suite is green. Nothing distinguishes "ran and passed" from "never ran" on the PR page, so these files can rot indefinitely without anyone learning.

## The suites

None of these mentions ClickHouse anywhere except in the guard line itself, which suggests every one is a false skip rather than a real dependency — but each needs a look, since "doesn't say ClickHouse" is not the same as "doesn't reach it transitively".

```
ee/governance/services/__tests__/personalVirtualKey.service.gracefulFallback.integration.test.ts
src/server/api/routers/__tests__/enterprise-guards.integration.test.ts
src/server/api/routers/__tests__/monitors-evaluator.integration.test.ts
[...]

PR #6892, opened 2026-08-12
fix(ci): re-arm a dead ast-grep rule and repoint stale post-restructure paths

0xdeafcafe:
## Why

Follow-up sweep after ADR-076 moved the app to `platform/app/`. A number of path
references were never migrated, and because each of them **fails open rather than
erroring**, nothing ever reported them — a glob that matches no file simply scans
nothing and reports success.

The one that matters:

> `dev/lint/ast-grep/rules/no-clickhouse-env-skip-guard.yml` still globbed
> `langwatch/**`, so it matched no file and scanned nothing.

Its ten sibling rules in the same directory were all migrated to `platform/app/**`;
this one was missed. That rule exists specifically to catch integration suites
that skip themselves and still report green — per its own message, 15 suites and
78 tests sat like that until #6327. It had quietly become an instance of the exact
failure it guards against.

Worth noting that `dev/lint/ast-grep/sgconfig.yml` exists so "no rule can silently
go dead again (see #3754)". The `rule-tests/` harness validates rules against
fixtures, which ignore the `files:` glob — so the rule test passes green while the
rule protects nothing in CI. The harness structurally cannot catch this class.

## What changed

**Re-arm the dead lint rule**

- `no-clickhouse-env-skip-guard.yml`: `langwatch/**` → `platform/app/**`.
  All 28 files using `TEST_CLICKHOUSE_URL` live under `platform/app` and none
  were covered. Confirmed this surfaces **zero** new violations, so it is a pure
  coverage restore, not a new gate.

**Fix references that resolve to nothing**

- `.coderabbit.yaml`: the soft-delete-vs-archive instruction was the one `path:`
  entry of 23 still on `langwatch/`, so it never fired.
- `codeql.yml`, `langwatch-app-ci.yml`: drop `langwatch_server/**` change filters —
  that directory has no tracked files.
- `check-feature-parity.ts`: drop `langwatch_server` from `DEFAULT_PYTHON_TEST_ROOTS`
  (raised by CodeRabbit in review). The workflow's parity filter documents an
  invariant — the filter list must cover every root the scanner reads, or a
  binding change in an unlisted root merges without the gate running (#5143).
  Removing the filter glob while the scanner still named the root broke that
  invariant, so the scanner root goes too. `walkFiles` catches the `ENOENT` and
  returns an empty array, so the root contributed nothing before and contributes
  nothing now — the same fail-open shape as the rest of this sweep, this time
  inside the parity checker itself.
- `pr-impact-map.yml`: the Python classifier listed `langwatch_nlp` and
  `langwatch_server`, neither of which has tracked files. Only `services/langevals`
  remains live, and `sdks/python` is already claimed by the SDKs branch above it.
- `.dockerignore`: drop three `langwatch_nlp/` patterns.
- `dev/compose.dev.yml`: two comments naming `langwatch/.env` and `./mcp-server`,
  both contradicted by the real values sitting beside them.

**Move a doc left one level too deep**

- `platform/app/dev/docs/best_practices/vitest-performance.md` was the sole
  occupant of a nested `dev/docs` tree. Both references spell it repo-root-relative,
  which only resolved by accident. Moved in with the other 24 and indexed in the
  README.

**Delete dead files**

- Two 0-byte `.mdx` stubs under `services/langevals/docs/evaluators/`, referenced
  by nothing. The real pages are `docs/api-reference/evaluators/` and are in the
  published nav.
- Two 0-byte `repositories/index.ts` modules under event-sourcing pipelines, plus
  the `export *` that re-exported nothing from one of them.

**Correct `CLAUDE.md`**, which agents read as ground truth: the Structure block
listed `langwatch_server/` and omitted five live directories, the Commands heading
said "Inside langwatch/", and the `start:prepare:files` row named both the old and
the new directory in a single sentence.

**Correct the vitest worker count** (raised by CodeRabbit in review).
`vitest-performance.md` claimed a bare `npx vitest` spawns "a fork per core". It
does not: run mode defaults to `availableParallelism - 1` workers and watch mode
to half that parallelism. `CLAUDE.md` already stated the run-mode figure, so the
doc was the odd one out. Now states both modes and keeps the recommendation to go
through the package scripts, which is what carries `maxWorkers` and
`vmMemoryLimit`.

## The Go contract-test fix is no longer here — #6898 landed it

This branch originally carried a third commit repointing `controlPlaneRoot` in
`services/aigateway/adapters/controlplane/guardrails_contract_test.go`, which was
the then-current red on main. **#6898 merged that fix first**, conflicting this
branch on that one file.

Resolved by taking main's version wholesale. That is not a coin-toss between two
equivalent fixes — #6898 is a strict superset. This branch repointed the constant;
#6898 repoints it *and* replaces the bare-directory `os.Stat` sentinel with two
real witnesses (a repo-root `pnpm-workspace.yaml` and an `@langwatch/web` package
identity), makes the ambiguous case **fail rather than skip**, and pins every
branch with `t.TempDir()` fixtures so the guard cannot be widened back into a
skip unnoticed. The Go file is now byte-identical to main and no longer appears
in this diff.

What remains here is the other fifteen files, which #6898 does not touch.

## Deployment Impact

**None.** This PR trips the deployment-surface path filter via
`dev/docs/best_practices/**` (a doc moved between directories) and `services/**`
(two 0-byte documentation stubs deleted). Neither is deployed.

- **Env vars added/changed:** none. `.env.example` is untouched.
- **Helm values added/changed:** none. Nothing under `charts/` is touched.
- **Default behaviour on a vanilla `helm install` with no overrides:**
  unchanged. No chart, image, manifest, or runtime configuration is modified.
- **BYOC dataplane impact:** none. No dataplane component, service binary, or
  container image is affected.

The only non-documentation edits are to developer-time tooling that never ships
in an image: a lint-rule glob, a CodeRabbit review path, three CI change-filters,
`.dockerignore` entries that matched nothing, and two 0-byte TypeScript modules
that exported nothing. No runtime code path changes, and no Go or TypeScript
source file is modified.

## Test plan

No behaviour change is intended anywhere. Verified mechanically:

- all six edited YAML files still parse
- the `pr-impact-map` regex rewrite is **provably equivalent** for every path that
  can appear in a diff — it differs only on `langwatch_nlp/` and
  `langwatch_server/`, both of which have zero tracked files
- both deleted `.ts` modules were confirmed 0 bytes via the empty-blob hash
  `e69de29b`, and the only reference to either was the `export *` removed here
- all 11 paths in the rewritten Structure block exist; all four commands under the
  rewritten Commands heading are really proxied by the root `package.json`
- zero tracked files remain under any pre-restructure path
- `check-feature-parity` exits `0` with **6243 enforced scenarios bound across
  1012 files**, unchanged by dropping the dead scanner root, and all **24**
  remaining roots across the five `DEFAULT_*_ROOTS` lists are covered by the
  workflow filter — each checked individually, since that is the invariant the
  filter's own comment asserts and this PR just showed it was not holding

**The re-armed lint rule was verified by differential probe**, because "matches
files and finds nothing" and "matches nothing" are indistinguishable from a clean
scan — which is the whole bug. A deliberately violating fixture was placed at
`platform/app/src/__ast_grep_probe__.test.ts` and scanned twice with
`ast-grep 0.45.1`:

- with `files: platform/app/**` (this PR) → **2 errors**, both patterns firing
- with `files: langwatch/**` (the old glob) → **clean scan, zero findings**, on a
  file that flagrantly violates the rule

That is the fail-open failure reproduced end-to-end: the old glob reports success
on a violation it cannot see. The probe was then deleted; a full scan of
`platform/app` with the fixed glob reports **zero** findings, confirming this is a
coverage restore and not a new gate.

## Anything surprising?

The theme is one bug wearing three costumes, and it is worth naming because the
repo keeps rediscovering it: a check pointed at a path that no longer exists does
not error, it **passes vacuously**. The ast-grep rule globbed `langwatch/**` and
scanned nothing. The CodeRabbit instruction globbed `langwatch/` and never fired.
The Go contract tests stat'd `langwatch/` and skipped. All three reported green.
Only the third one eventually turned red, and only because an unrelated PR
happened to recreate the directory — which is also why it is the only one of the
three that got fixed on its own schedule, in #6898.

The rule-test harness built after #3754 to stop exactly this cannot catch it:
`ast-grep test` validates rules against fixtures, which ignore the `files:` glob,
so the rule test passes while the rule protects nothing.

Possible follow-up, not done here: a check that every `files:` glob in
`dev/lint/ast-grep/rules/` matches at least one tracked file. That would have
caught this the day it broke, and it generalises — the same shape of assertion
(this pattern must match something) is what the CodeRabbit `path:` entries and the
workflow change-filters also lack.

Deliberately left out of scope, each needing a decision rather than a sweep:

- `docs/media/images/` holds 166 byte-identical duplicates of `docs/images/`, kept
  alive by a single reference in `prompt-playground.mdx:13`. Biggest byte win in
  the tree.
- `docs/adr/014-depth-aware-overlay-z-index.md` is an internal ADR misfiled into
  the public docs site — unreferenced, not in the nav, and its number already
  collides twice under `dev/docs/adr/`.
- `assets/mcp-server/cursor-example.png` is an orphan under an old-name directory
  with no in-tree reference.

## Summary by CodeRabbit

* **Documentation**
  * Updated project structure, setup guidance, development configuration references, and generated-file instructions.
  * Added best-practice guidance for Vitest performance and linked it from the documentation index.

* **Maintenance**
  * Updated repository checks, CI triggers, lint scopes, and feature-parity scanning to reflect the current project layout.
  * Removed obsolete Docker ignore entries and an unused evaluation-processing export.


langwatch-agent:
## Human Review Brief

**Mode: Targeted Review.** No linked issue; the PR body names the mechanism, which is the right record for a sweep.

### It collides with an open PR

<https://github.com/langwatch/langwatch/pull/6897> edits the same file, `services/aigateway/adapters/controlplane/guardrails_contract_test.go`, for the same stale `controlPlaneRoot`. Both are ready and both have a reviewer. Whichever merges first conflicts the other, so pick one before either goes in. Note also that `main` is green as of this morning, because #6893 re-deleted `langwatch/src/server` and put the Go contract tests back to silently skipping.

### Decisions being ratified

1. **Fail-open globs are the bug class, not the individual paths.** A glob matching no file scans nothing and reports success, which is how `no-clickhouse-env-skip-guard.yml` came to guard nothing while its ten siblings were migrated. Approving this fixes the instances; it does not change the class.
2. **`rule-tests/` is accepted as structurally unable to catch this.** The harness validates rules against fixtures that ignore the `files:` glob, so the rule test passes green while the rule protects nothing. Ratifying that means the repo knowingly has a guard whose own test cannot see it die.
3. **Sixteen unrelated files travel together** because they share the restructure as a cause. That is a defensible bundle, and it also means one revert takes all of them.

### Must Check

- **The other ast-grep rules.** Ten siblings were already on `platform/app/**`; confirm this sweep found every remaining `langwatch/**` glob rather than the ones that happened to be noticed. A grep for `langwatch/**` across `dev/lint/` and `.github/workflows/` is the whole verification.
- **`no-clickhouse-env-skip-guard.yml` after the fix.** It exists to catch suites that skip themselves and report green. Confirm it now actually matches files and, ideally, that it flags something, since a rule that matches files and finds nothing looks identical to one that matches nothing.

### Ask Author

Is there a way to make a zero-match glob fail rather than pass? That is the only change that stops this recurring, and this PR is the natural place to argue it either way.

drewdrewthis:
@tmp_review/verdict_body.md

drewdrewthis (commented):
pr-review (reviewer: @drewdrewthis): READY. Verified go-ci actually ran and passed on this head SHA (not path-filtered/skipped) and every repointed path resolves to a real location. Full verdict: see the pinned review-verdict comment on this PR.

drewdrewthis (dismissed):
lgtm

langwatch-agent:
Status update from triage: **#6898 merged at 09:52 today** and repointed `controlPlaneRoot` in `services/aigateway/adapters/controlplane/guardrails_contract_test.go`, so this PR is now `CONFLICTING` with `main` on that one file.

The other fifteen files here are unaffected and are the real content of this PR: the `langwatch/**` ast-grep globs, the workflow paths and the docs. Dropping the Go file from this branch and rebasing should clear it.

0xdeafcafe:
Answering the brief's two Must Checks and the question, and closing out the collision.

## The collision resolved itself, and not symmetrically

#6898 merged at 09:52 and repointed `controlPlaneRoot`. I resolved this branch by taking main's version wholesale rather than keeping mine, because the two are not equivalent fixes:

- **This branch** repointed the constant to `platform/app`. That fixes the instance.
- **#6898** repoints it *and* replaces the bare-directory `os.Stat` sentinel with two witnesses that only a real control plane can produce (a repo-root `pnpm-workspace.yaml`, an `@langwatch/web` package identity), makes the ambiguous case **fail rather than skip**, and pins every branch with `t.TempDir()` fixtures so a later edit cannot widen the skip back open unnoticed.

That is the stronger answer to the same bug class this PR is about, so it wins. The Go file is now byte-identical to main and has dropped out of this diff entirely. The remaining fifteen files are untouched by #6898, exactly as triage predicted.

**Loose end:** #6897 is still open and is now fully redundant — it is the same one-line repoint #6898 already landed as part of a superset. It should be closed rather than merged.

## Must Check 1 — did the sweep find every remaining glob, or only the noticed ones?

Complete, verified by the grep the brief specified.

`dev/lint/` now has **zero** occurrences of `langwatch/` outside `platform/app`. Every rule in that directory is on `platform/app/**`.

`.github/workflows/` still contains the string `langwatch_nlp`, and every remaining hit is live rather than stale — I checked each target exists on disk:

- `infra/docker/Dockerfile.langwatch_nlp` — exists; it is the Dockerfile's real filename
- `langwatch/langwatch_nlp`, `langwatch/${IMAGE}` — published **image** names, not source paths
- `charts/langwatch/**` — exists

So the only stale source-directory references were the ones this PR removes (`langwatch_server/**` filters, the `langwatch_nlp` classifier entries, the `.dockerignore` patterns). Nothing was found by noticing; the grep is exhaustive over both directories.

## Must Check 2 — does the rule now actually match, and does it flag anything?

This is the sharpest point in the brief: *a rule that matches files and finds nothing looks identical to one that matches nothing*. A clean scan cannot distinguish them, so I did not rely on one. Differential probe instead, `ast-grep 0.45.1`.

Placed a deliberately violating fixture at `platform/app/src/__ast_grep_probe__.test.ts`:

```ts
const isTestcontainersOnly = !!process.env.TEST_CLICKHOUSE_URL;
describe.skipIf(isTestcontainersOnly)("probe", () => { ... });
```

Scanned it twice, changing only the `files:` glob:

| `files:` glob | Result on the violating file |
| --- | --- |
| `platform/app/**` (this PR) | **2 errors** — both the named-constant and the `skipIf` patterns fire |
| `langwatch/**` (the old glob) | **clean scan, zero findings** |

The second row is the bug, reproduced end-to-end: the old glob reports success on a file that flagrantly violates the rule it is supposed to enforce. That is the direct evidence that this line was dead rather than merely unexercised.

The probe was then deleted. A full scan of `platform/app` with the fixed glob reports **zero** findings — so this is a coverage restore, not a new gate, and no existing suite is newly in violation.

## Ask Author — can a zero-match glob be made to fail?

**Not within ast-grep.** I checked `ast-grep scan --help` on 0.45.1: there is no strict mode, no minimum-match assertion, no zero-match diagnostic. A `files:` glob that matches nothing is indistinguishable from one that matches clean files, by design — the tool has no concept of "this rule was supposed to see something".

So it has to be external, and it should be, because the same hole is not ast-grep-specific. The identical failure occurred in three unrelated tools in this one sweep: the `.coderabbit.yaml` `path:` entry, the workflow change-filters, and the Go `os.Stat` sentinel. A fix scoped to ast-grep would leave two of the three standing.

The cheap version, which I would argue for and have deliberately **not** built here because it wants its own PR and its own spec:

> A CI check asserting that every `files:` glob in `dev/lint/ast-grep/rules/` matches at least one **tracked** file, failing with the rule id and the dead glob.

It is a `git ls-files` intersection per glob — small, no new dependency, and it would have caught this the day ADR-076 landed rather than eight days later. It also extends naturally to the `.coderabbit.yaml` `path:` entries, which have exactly the same property and no test at all today.

The generalisation worth stating, since the brief asks for the argument rather than the patch: **a path-scoped guard needs an assertion that its scope is non-empty, or it silently becomes a no-op the first time the tree is reorganised.** The `rule-tests/` harness cannot supply that assertion — it validates rules against fixtures, and fixtures ignore `files:` — which is why the harness built after #3754 to prevent exactly this class did not see it.

## Decisions being ratified

Confirming the brief's framing, with one correction:

1. **Fail-open globs are the bug class.** Agreed, and the probe above makes it concrete rather than asserted. This PR fixes instances; it does not close the class. The follow-up above is what closes it.
2. **`rule-tests/` is structurally unable to catch this.** Agreed, and it is now stated in the PR body rather than only in review, so the next person meets it before rediscovering it.
3. **Sixteen files travel together.** Now fifteen, since the Go file left with #6898 — and the revert-takes-all-of-them caveat correspondingly no longer touches any Go or TypeScript source. The remaining bundle is CI config, lint config, docs, and four 0-byte files.

0xdeafcafe:
> **Keep the feature-parity filter aligned with the scanner.** (`.github/workflows/langwatch-app-ci.yml:179`)

Fixed in 1c9bd748d4, taking the second of the two branches offered: `langwatch_server` is obsolete, so the **scanner root** is removed rather than the filter glob restored.

The invariant cited is real and worth quoting in full, since it is the reason this matters:

> This list MUST cover every root the scanner reads (`DEFAULT_TEST_ROOTS`, `DEFAULT_GO_TEST_ROOTS`, `DEFAULT_PYTHON_TEST_ROOTS`, `DEFAULT_BATS_TEST_ROOTS`, `DEFAULT_SHELL_TEST_ROOTS` in `platform/app/scripts/check-feature-parity.ts`), otherwise a binding change in an unlisted root merges without the gate ever running and silently breaks parity on main (e.g. #5143 edited only `dev/scripts/__tests__/*.bats` and slipped through).

The previous commit removed `langwatch_server/**` from the filter and left `DEFAULT_PYTHON_TEST_ROOTS` naming `langwatch_server`, which is exactly the asymmetry that comment warns about. Correct catch.

**One correction to the stated risk.** The finding says "changes to Python scenarios under `langwatch_server` can now bypass feature-parity validation." That cannot happen — `langwatch_server` has **zero tracked files** and does not exist on disk, so there is nothing under it to change. No bypass was ever reachable.

The real defect is the mirror image, and it is more interesting: the root was contributing **nothing while reading as coverage**. `walkFiles` opens each root inside a `try` and returns an empty array on `ENOENT`:

```ts
try {
  entries = readdirSync(root);
} catch {
  return out;   // missing root -> zero files, silently
}
```

So the scanner has the same fail-open shape this branch exists to clean up: a scan root that resolves to nothing scans nothing and reports success. It was never going to fail, and nothing was ever going to say so. That is why it is removed rather than left in place — an entry that looks like a covered root but cannot ever cover anything is worse than no entry, because the next person reads the list as the coverage set.

**Verification:**

- `check-feature-parity` exits `0`, **6243 enforced scenarios bound across 1012 files**, unchanged from before the edit — as expected, since the removed root contributed zero bindings.
- All **24** remaining roots across the five `DEFAULT_*_ROOTS` lists are covered by the workflow filter. I checked each one individually rather than assuming, since that is the invariant the comment asserts and this PR just demonstrated it was not holding.
- Biome reports **9 warnings on this file both before and after** the edit — all pre-existing complexity warnings, zero delta.

Both sides of the invariant now agree, and neither names a path that does not exist.

PR #6899, opened 2026-08-12
feat(automations): the View drawer becomes the in-depth view (#6716 WS-3)

0xdeafcafe:
**Based on `issue6716/ws6-list-pages-and-providers`** (PR #6884, closed workstream, frozen) — it also touches `ViewAutomationDrawer.tsx`, so this stacks on it to avoid a conflict. Retarget to `main` once ws6 lands.

Part of #6716 (WS-3). Draft: the design note below wants a read before this is treated as settled.

## What this is

Alex's framing: **the View drawer is the in-depth view of an automation — the full history of its firing, plus its scheduled/upcoming firing.** Not a debug console. The debuggability falls out of that: "could I see the latest evaluation, I want to debug why it is not firing" is answered by putting the last check on the history timeline, and "a button to run the search for this trigger" is answered by "Recent matches".

Before this, the drawer showed a thin "Recent fires" list and nothing else. An alert that had never fired said nothing at all — the evaluator computes an observed value on every check and records none of it.

## The three sections

**History.** The whole firing history, paged. Keyset pagination on `(createdAt, id)` — `createdAt` alone is not unique, and a burst of matches inside one millisecond would otherwise skip or repeat rows across a page boundary. Alerts stay per-incident (each open and each recovery is a distinct event); a busy trace automation still collapses a run of same-minute fires into one "Fired 7 times" row.

The last check of an alert sits at the top of the same timeline, because "checked 5 minutes ago and the value was nowhere near the threshold" is the answer to "why has this not fired?", and it belongs next to the fires whose absence it explains:

> ● Checked · The alert did not fire — 5 minutes ago
> observed 42, fires when greater than 100 over 1 hour

A skipped check names its reason in words the reader can act on: "The graph this alert watches groups by a field with too many distinct values to check against a threshold. Edit the graph to group by fewer values, or remove the grouping."

**What happens next.** Every answer is one the platform can stand behind:

| Kind | Answer | Source |
|---|---|---|
| Paused (any kind) | "Nothing, while this alert / schedule / automation is paused" | `Trigger.active`, plus the scheduler entry for a report. Carries `pausedReason`, so a runaway-volume pause explains what to narrow instead of saying "resume it" |
| Schedule | "Sends next on 13 Aug 2026, 09:00" | `TriggerService.getReportSchedules` — the scheduler owns the instant |
| Digest | "Sends the next batch at 12:05" | `computeScheduledFor` itself, not a re-derivation — so the view can never quote a boundary the dispatcher disagrees with, including a persist action that ignores its stored cadence |
| Immediate | "Acts as soon as a matching trace arrives", plus the debounce it waits out | `Trigger.traceDebounceMs` |
| Alert | "Checked as data arrives", plus "an alert waiting for data to stop arriving is also checked every 30 seconds" | The heartbeat interval. There is no next instant for an event-driven evaluation, and inventing one would be a promise we do not make |

**Recent matches.** A "Run the conditions now" button that runs the automation's own query against the last 7 days and lists what would match, with input and output excerpts. It reuses `api.tracesV2.list` — the same procedure the composer's live preview uses — so `traces:view` and the content protections are enforced by the read itself rather than by anything new here. On demand, never on open: a saved automation's view should not issue a trace search every time someone glances at it. Offered only for automations whose subject is a trace search query; a graph alert and a legacy `filters` row get no control.

---

## Design note: where the evaluation record lives

**Alex — this is the part that touches the event-sourced substrate. Please read this one.**

### Shape

A new Postgres table, `TriggerLatestEvaluation`, with **`triggerId` as the primary key** — exactly one row per trigger, written as an upsert:

```
triggerId (PK, FK → Trigger ON DELETE CASCADE)
projectId
evaluatedAt
verdict            fired | already_firing | resolved | not_breached | not_delivered | skipped
observedValue      null when the check never read the metric
threshold          the condition AS EVALUATED, not as currently stored
operator
timePeriodMinutes
skipCode           subject_missing | incomplete_configuration | result_too_large | inactive
```

### Why not `TriggerSent` (the minimal-looking option)

`TriggerSent` is the **fire ledger**, and it is load-bearing in three ways an evaluation row would disturb:

1. It is arbitrated by the single-column unique on `openIncidentKey` — the atomic claim that makes a graph alert fan out at most one notification. Nothing that is not an incident belongs in a table whose uniqueness constraint IS the at-most-once guarantee.
2. Every "last fired", "fires in the last 30 days" and "currently firing" answer is a count or a max over its rows (`trigger-fire-history.repository.ts`). Evaluations happen on every trace activity for a project; writing them there would silently change all of those numbers and multiply the table's row count by orders of magnitude.
3. It is the fire generation the per-recipient idempotency digest is keyed on (`findLatestForGraphAlert`). A non-fire row landing between two fires would re-key that digest.

So the minimal option is *not* the fewest tables — it is the fewest things that can break. A separate table is read by nothing that exists today and cannot affect any of the above.

### Why not columns on `Trigger`

`updateLastRunAt` already writes to the row on every evaluation, so the write volume argument is weak — but `Trigger` is durable customer configuration that the composer's upsert rewrites wholesale, and every read path loads it. Volatile observation state does not belong on the same row as the configuration it observes, contending with the customer's own saves.

### Write path

`evaluateGraphTrigger` is now a thin wrapper around the unchanged evaluation, so there is **one** write site rather than one per return branch, and no future branch can forget to record. The evaluation itself is untouched apart from: (a) a stable `code` added alongside the existing developer-prose `detail` at each `skipped(...)` call, and (b) one line capturing the condition as soon as it is parsed, so the recorded threshold is the one that was actually evaluated rather than whatever the row says by the time someone opens the drawer. The `catch` around the oversized-result skip is unchanged.

`recordEvaluation` is an injected optional hook (the composition root always wires it), for the same reason the notifier is injected — the service stays free of Prisma.

**Recording can never break an alert**, guarded at two levels: `TriggerLatestEvaluationService.record` swallows and logs its own failures, and the wrapper catches anything a different implementation of the hook might throw. A throw there would be redelivered by the outbox and re-evaluate a trigger that has already fired and already dispatched — the exact silent-alert failure mode this feature exists to expose. There is a test for it.

A thrown evaluation records nothing: the throw is redelivered, and the retry's outcome is the one worth showing.

### Retention and corruption surface

- **Retention: none needed.** One row per trigger, bounded by the number of alerts, deleted with the trigger by the cascade.
- **Concurrency:** the real-time subscriber and the heartbeat sweep can evaluate the same alert at once. The write is a single `INSERT … ON CONFLICT DO UPDATE`, so the later write simply wins — no read-then-write race on the first evaluation. It is raw SQL rather than `prisma.upsert` because the model's unique input is the bare `triggerId`, which the multitenancy middleware rejects for want of a `projectId` in the where clause; the statement carries `projectId` in both the inserted row and the conflict guard, which is that same constraint expressed in SQL.
- **Blast radius on the existing ledger: nil.** No existing query reads or writes this table, and no existing table gained a column.

---

## Permissions

Fire history stays metadata-only — no trace ids, gated by `triggers:view`, which is weaker than trace-content permission. There is an integration test asserting the select never grows a `traceId`. The evaluation record is the same shape of thing: an observed number, the condition, a verdict. Trace content reaches the drawer only through `tracesV2.list`, which enforces `traces:view` and the redaction protections itself.

## Tests

- `specs/automations/evaluation-visibility.feature` — 15 scenarios, all bound (`pnpm check:feature-parity`: `15/15 scenarios bound ✓`).
- `ViewAutomationDrawerHistory.integration.test.tsx` — 11 scenarios at the surface (evaluation copy, skip reason, never-evaluated, schedule / paused / digest / alert next-firing, run-now with excerpts, empty match, no run-now for a graph alert).
- `automations.viewProcedures.integration.test.ts` — 9, at the API boundary (paging + cursor + the no-trace-id assertion, latest evaluation, next firing including NOT_FOUND).
- `graph-trigger-evaluation-record.unit.test.ts` — the write, including both skip classes and the "recording failed, the alert still fires" case.
- `next-firing.unit.test.ts` — the derivation, including a persist action stored with a digest cadence and an unrecognised cadence value.

### Review round (second commit)

- **History was a permanent skeleton for every non-alert automation.** A disabled react-query read reports `isLoading` forever, and the skeleton counted the alert-only evaluation read. Gated on `enabled`. The drawer mocks hard-coded `isLoading: false` on every hook, which is exactly what made the state unrepresentable — they now model enabled/settled faithfully (`undefined` = unresolved, `null` = resolved-to-nothing), so removing the gate fails the suite. Added the missing coverage: a trace automation with fires renders its grouped "Fired 3 times" row.
- **Paused now answers for all four kinds** (was report-only), with a `Paused` badge beside the kind badge. A paused alert no longer says "checked as data arrives" one line above a history row saying it is paused.
- **Durations spelled out** — "lasted 15 minutes", not "lasted 15m"; three pinned assertions updated.
- **Webhook delivery log's loading skeleton restored** (lost in the extraction).
- **Matched-traces docstring corrected**: same query *text*, different engine (panel = ClickHouse over stored traces; dispatcher = liqe over fold state in memory). They can disagree, and it now says how.
- **`affectedRows === 0` on the upsert logs a warn** — a write blocked by its own tenancy guard is now reported rather than inferred from a frozen timestamp.

### Second review round (third commit)

- The runaway-pause sentence had three hand-written copies (list badge, drawer badge, drawer answer). It is now one `RUNAWAY_PAUSE_EXPLANATION` in `pauseReasons.ts`, beside the predicate that decides when to show it, and all three read it. The mailer keeps its own wording — different register, deliberate.
- `fakeQuery` in the drawer tests collapsed `null` into `undefined`, contradicting the model its own comment sets out; it passes the data through now, so the helper teaches the shape react-query actually hands a component.

Full runs after both review rounds: `features/automations/__tests__` **54 passed**, `features/automations/logic` **151 passed**, `server/app-layer/automations` **397 passed**, `event-sourcing/pipelines/automations` 40 passed, the two router integration files **60 passed** together (no regression from the router additions). `pnpm check:feature-parity`: `15/15 scenarios bound ✓`.

## Merge notes

- `routers/automations.ts` gains **three new read procedures only**, appended after `getReportSchedules` — no existing procedure is touched. There **is** one conflict with `issue6716/ws8-rest-parity-s3`, verified with `git merge-tree`: a single import line. They add `redactTriggerForRead` from `.../trigger-redaction` and this adds `TriggerLatestEvaluationService` from `.../trigger-latest-evaluation.service`, both immediately after the `TriggerFireHistoryService` import, and biome's alphabetical ordering puts mine between the two — so it cannot be relocated away. Resolution is keep-both; whoever merges second takes it. No procedure body conflicts.
- `RecentFiresList` and the webhook delivery log moved out of the drawer into `components/view/`; the drawer composes sections now. Net biome warning delta on the touched files is **0** (the drawer sheds one; the paused branch's new one in `next-firing.ts` was refactored away).
- `getRecentFires` is left in place but now has **no consumer** — the drawer was its only one, and the paged `getFireHistory` replaces it. Deleting it would mean editing `automations.router.integration.test.ts`, which `issue6716/ws8-*` is also in; happy to drop both in a follow-up once those land.

## How it was verified

Rebuilt on top of its base branch (`issue6716/ws6-list-pages-and-providers`, which now carries `main`), so these runs are against the merged tree, not the branch in isolation:

| Scope | Result |
|---|---|
| `pnpm test:unit run src/features/automations src/server/app-layer/automations` | 48 files, **670 passed** |
| `pnpm test:unit run src/server/app-layer/automations src/server/event-sourcing/pipelines/automations` | 40 files, **437 passed** |
| `pnpm test:integration run src/features/automations` (drawer suites, including the three `ViewAutomationDrawer*` files) | 17 files, **171 passed** |
| `pnpm test:integration run` on `automations.viewProcedures` + `automations.router` | 2 files, **60 passed** |
| `pnpm check:feature-parity` | `specs/automations/evaluation-visibility.feature` — `15/15 scenarios bound ✓` |

Biome's delta gate reads per file, not as a net: the drawer shed 3 violations but the extracted `WebhookDeliverySection` inherited 2 (length and branching), so the expanded attempt panel is now its own component and the file is clean. The gate reports **0 added**.

CI caught one more absorb the scoped local runs could not see: a mutation reports `isPending` under the merged react-query, so the delete confirmation's spinner prop no longer typechecked as `isLoading`.

The merge itself needed one conflict resolved (the drawer's import block — the base branch had already moved `TriggerKind` to the generated Prisma 7 client, and this branch's date-fns/`useState` imports left with the sections that moved into `components/view/`). Two follow-on repairs came with it: the slice's own new files still imported from `@prisma/client`, which `main` replaced with `~/generated/prisma/client`, and the fire-history service's mock repository needed the `findPageByTriggerId` this workstream added to the repository interface.


0xdeafcafe:

Issue #6717, opened 2026-08-07
refactor(automations): merge alerts into automations, rename type to source, project level Slack token


rogeriochaves:
Design changes agreed in the live bug bash on automations and alerts on 2026-08-04. Five people were on the call. Nothing from it was written down at the time, so this issue records the decisions.

The defects from the same call are in langwatch/langwatch#6716 bugbash(automations): eleven defects from the 04 August team session. This issue is only the design work.

## Agreed

- [ ] **Rename the `type` field to `source` across the automations UI.** The values become: search or trace filter, graph or metric, and clock. The field says where the automation gets its input from, so `source` is the honest name. Rogério proposed it in the call and Alex accepted it.

- [ ] **Merge automation and alert into one flow.** Today they are two things. The only real difference between them is where the result is delivered. One flow with a delivery target is simpler to build and simpler to explain. Keep schedule as its own separate concept.

- [ ] **Move the Slack bot token to a project level integration.** Set it up once for the project instead of pasting it into every alert. One token, one place to rotate it.

## Undecided

- [ ] **Re-step or simplify the composer drawer.** The drawer asks for too much in one panel. The call agreed it is a problem but did not agree on the fix, so this one is still open. Splitting it into steps is one option, cutting what it asks for is another.

## Not in this issue

Automations documentation has not been written. That work already has an issue, langwatch/langwatch#5822 docs(automations): full documentation for the event-sourced automations, so it is out of scope here.

Filed by Nova from the call recording. The call itself produced no written record.

PR #6907, opened 2026-08-12
feat(automations): merged authoring wizard and unified list (ADR-093 R0)

0xdeafcafe:
R0 of [ADR-093](https://github.com/langwatch/langwatch/blob/feat/automations-source-r0/dev/docs/adr/093-automations-source-merge.md): the merged automation flow and the unified list, replacing the surfaces they supersede rather than sitting beside them behind a flag.

**Not feature-flagged** (decision by Alex, 2026-08-12). The old surfaces die at merge, so the bound scenarios that pinned their copy and seating are rebound in this same PR instead of waiting for F6/F7. The ADR's reference-implementation section records that and moves those items.

## What changes

**One flow, one opening question.** The Automation / Alert / Schedule kind picker is gone. The composer opens with "What should this automation watch?" — a trace filter or a graph — and the rule shape follows the answer. Choosing what to watch *is* the choice the type card used to ask for, so the question is deleted rather than renamed. ADR-037's Notification-versus-Action picker survives untouched, relocated into the Delivery step.

**The composer becomes a three-step wizard.** Watch → Delivery → Review, behind a persistent step rail that keeps each reached step's one-line summary on screen and one click away — the direct answer to ADR-037's objection that a stepper blocks revisiting earlier choices. Editing opens on the Review overview, always, and each section's edit affordance enters that step alone and returns. The overview never disappears, which is the one measured failure of the previous restructuring attempt.

The ceiling advice takes the two seats ADR-093 §4 decided: the Review step at create (the first moment every facet is known at once) and the Watch step on edit (where the saved delivery already names the action class).

Schedules keep the single-pane composer — the clock is not something to watch, and unifying their shell is explicitly deferred.

**One noun, and the third concept is a Report.** A graph-watching automation is an automation: the drawer heading, the save button, the toast, the delete dialog, the View drawer badge and the graph-card entry points all say so. The third concept is a **report**, not a schedule (decision by Alex, 2026-08-12) — a schedule is when it goes out, which is one of its fields; the thing itself is the report it sends. The scheduling vocabulary stays wherever it describes timing. Storage and the wire discriminator are untouched — this is what the product says, not what it stores.

**One list.** The Automations and Alerts tabs become one table with Watches and Delivery columns; a row says "Trace filter" with its query, or "Graph · <name>" with its firing rule. Reports keep their tab, the Overview is retained with its create menu down to two items, and the old `/automations/alerts` and `/automations/schedules` paths both keep resolving, so links issued before the merge still land where they pointed.

## Scenario ledger

**Bound in R0** (15, `specs/automations/source-merge.feature`): the wizard opening on what to watch; creating for either subject; completed steps staying in view; the saved subject being locked; editing opening on the overview and returning to it; abandoning a create or an edit; both cap-advice seats; the unified table listing both subjects; reports staying on their own tab; the delete noun; the Overview create menu.

**Still `@unimplemented`** (19): the list's filter chips (F6), the Slack project integration and legacy-token migration (F2–F4), the template that ships its graph (F5), the wire `source` alias (F1), and the empty-graph-list consolidation — that last one waits for F5's template affordance, so the authoring-drawer scenario it consolidates keeps its binding meanwhile rather than being traded for one that cannot pass yet.

**Rebound**: `list-pages.feature`'s remaining scenarios to merged-world wording (including the Notifies → Delivery cell rename), and `authoring-drawer.feature`'s cadence-regroup trigger, which is now the Delivery step's cadence control.

**Deleted with a replacement binding**: `list-pages.feature`'s "Deleting an alert asks for confirmation and names it as an alert" and "The Overview offers creating an automation, alert, or schedule" — both superseded by bound scenarios in `source-merge.feature`. Deleted outright (untagged, superseded by the wizard): `authoring-drawer.feature`'s "A completed section collapses to a one-line summary" and "Abandoning the drawer persists nothing".

`pnpm check:feature-parity` passes with every automations spec fully bound.

## Test evidence

- `pnpm test:integration` over `src/features/automations`, `src/pages/[project]`, `src/components/analytics` and the drawer registry: **35 files, 304 tests green**
- `pnpm test:unit` over `src/features/automations` (18 files / 307) and `src/server/app-layer/automations` (34 files / 431), plus the frontend-boundary guard
- `pnpm check:feature-parity`: green; `specs/automations/source-merge.feature` 15/15, `list-pages.feature` 10/10, `authoring-drawer.feature` 22/22, `automation-authoring-cap-advice.feature` 10/10, `evaluation-visibility.feature` 15/15
- tslsp diagnostics clean across every touched TypeScript file; biome warning counts unchanged against the base for the two files that carry pre-existing complexity warnings

## Not in R0

Any Prisma migration, any public-API change, the Slack integration, template-ships-graph, list filter chips, View-drawer changes, presentation-registry additions.

Refs #6717, #6716. Follows PR #6901.

## Deployment Impact

No operator impact. This is a frontend change plus spec and ADR edits: no environment variables added or changed, no Helm values added or changed, no chart or service touched, and no Prisma migration. A vanilla `helm install` with no overrides behaves exactly as before — the only difference is which screens the app renders, and the merged surfaces need no configuration to appear because they are not flagged. BYOC dataplanes are unaffected: nothing here touches the gateway, the workers, or any data-plane contract, and the stored `triggerKind` enum and the public trigger API are deliberately unchanged (ADR-093 §2, §3), so existing API clients and dispatch keep working untouched.

---

**Update (Report rename + base merge).** Two rounds landed after the first review pass:

- **Schedule → Report** as the customer-facing name for the third concept, carried by `presetLabels` the same way the merged automation noun is, plus the tab, section header, Overview menu, empty state, View drawer badge, and the "what happens next" copy. The `/automations/schedules` path still answers. Bound scenarios reworded with it: `source-merge.feature`'s "Reports stay on their own tab" and "The Overview offers creating an automation or a report", `list-pages.feature`'s report delete-noun scenario, and `evaluation-visibility.feature`'s two schedule-noun scenarios. The ADR branch carries the same wording for `source-merge.feature`; the edits are identical and merge clean.
- **Merged `issue6716/bugbash-integration`** at f0fccc6b78, which fixes the lint and typecheck failures this PR was inheriting.

One follow-up left for whoever owns WS-3's spec: `evaluation-visibility.feature` still uses "alert" as a noun in nine scenario titles. The code they pin now says "automation", and the scenarios stay true either way, so nothing is red — but the file's vocabulary should follow the merge.

---

**Update (validation fix round).** Merged `feat/automations-source-adr` (the Report-renamed ADR, resolving two conflicts in favour of its prose and this branch's binding tags) and `issue6716/bugbash-integration` at 39d969a373.

- **P1-1 — edit→create carried the whole edited draft.** Opening the same drawer type replaces URL params in place instead of remounting, so the store, `hydratedFromServerFor` and `baselineRef` all survived "New automation" from a saved automation: Save would write a duplicate row (or trip `customGraphId @unique`), and the close-guard diffed the previous automation's baseline. Fixed with an effect keyed on the automation id that does what unmounting would have — reset the store, unlatch hydration and the baseline, and re-land the step — rather than a `key` on the shared drawer host, which would change remount behaviour for every drawer type in the app. Two tests pin it, both verified failing without the effect.
- **P1-2** breadcrumb and document titles: the preserved `/alerts` path now titles "Automations", `/schedules` titles "Reports".
- **P1-3** View-drawer vocabulary finished: `evaluationPresentation.ts` (12 strings), `HistorySection.tsx`, and `evaluation-visibility.feature`'s remaining alert-noun titles and steps, with the `@scenario` annotations moved with them.
- **P2s:** the type-picker scenario rewritten to the Watch step (4); ADR R0 section amended for the View-drawer vocabulary now legitimately in scope and for templates not locking the Watch step until F5 (7, 11); dead `/reports` branch dropped rather than half-routed (8); the Automations tab highlights on its alias via `includePath` (9); Review and the rail share `filtersAreSet` (10); the remaining reachable "alert" nouns swept — variable tooltips, provider descriptions, template display names, Slack/webhook authoring hints, use-case seeds, the email banner (12); "Choose a type first" reworded in both places (13); the "Editing an automation opens the review overview" binding moved onto the drawer test that lets the drawer choose the step (14).
- **Known, deliberate (15):** a report can no longer be switched to another kind mid-compose. The kind picker was the only control that did that, and §1 deletes it — a report is created from its own entry point, and converting one is a create plus a delete, exactly as it is for what an automation watches.

Green after the round: 36 integration files / 307 tests, 52 unit files / 738 tests, `check:feature-parity` all-bound across every automations spec, `biome check` clean at error level in `platform/app`, tslsp clean.


0xdeafcafe:
**On the red lint and typecheck jobs: both are inherited from the base branch, not added here.**

`issue6716/bugbash-integration` is already red on the same two jobs ([run 31607350865](https://github.com/langwatch/langwatch/actions/runs/31607350865)), and the numbers match exactly:

| Job | Base branch | This PR |
|---|---|---|
| lint | 4 errors, 3593 warnings | 4 errors, 3593 warnings |
| typecheck | 5 errors | the same 5 errors |

The typecheck errors are `src/app/api/triggers/[[...route]]/app.ts:780` and four `'only' is possibly 'undefined'` in `ConditionBuilder.tsx:110-113`; the one lint error inside `platform/app` is the `useOptionalChain` finding on `ConditionBuilder.tsx:107` that causes those four. Neither file is touched by this PR.

This PR did add one typecheck error of its own — a polymorphic `Box as="button"` cannot carry `type`, which the repo's typecheck config catches where the language server does not. Fixed in d66474676a by using `chakra.button`, the same form the delivery cards already use, and the error is gone from the latest run.

`test-unit` and `test-integration` both pass here. `test-unit (1)` was failing on the base run.

0xdeafcafe:
**Resolved.** `issue6716/bugbash-integration` fixed both inherited failures in f0fccc6b78 (the ConditionBuilder guard, the detached-handler id guard in `app.ts`, the dispatch-wiring import order), and that commit is merged into this branch — so the lint and typecheck jobs now stand on their own here rather than carrying the base's red.

Locally after the merge, `biome check --diagnostic-level=error` over `platform/app` is down from 4 errors to 2, and both remaining ones are the formatter disagreeing with two **committed generated files** — `src/server/evaluations/evaluators.generated.ts` and `src/shared/langy/langySkills.generated.json`. Neither is touched by this PR, and `pnpm format` rewrites both, so they are generator-output churn rather than anything a change can fix in passing. Worth deciding separately: either run the formatter as part of codegen, or exclude the generated paths from the format check.

0xdeafcafe:
**Correction to the note above, with the numbers from the run after the merge.** f0fccc6b78 fixed the `pnpm typecheck` half, and both jobs are still red on the base for two remaining reasons — this branch is at exact parity with it, not adding to either:

| Job | Base ([run 31633795935](https://github.com/langwatch/langwatch/actions/runs/31633795935)) | This PR |
|---|---|---|
| lint | 2 errors, 3593 warnings | 2 errors, 3593 warnings |
| typecheck | 16 errors, in 7 files | the same 16, in the same 7 files |

- **lint**: both errors are the formatter disagreeing with committed generated files (`src/server/evaluations/evaluators.generated.ts`, `src/shared/langy/langySkills.generated.json`). `pnpm format` rewrites both, so this wants a decision — format as part of codegen, or exclude generated paths from the check — rather than a drive-by commit that the next generator run would undo.
- **typecheck**: every error is from `pnpm typecheck:tests`, the half that covers test files (`tsconfig.tsgo.json` excludes them from `pnpm typecheck`, and CI runs both). The seven files are `trigger-actionparams-redaction`, `trigger-parity`, `trigger-write-rules`, `toaster.layout` (×2), `trigger-fire-history.service` and `trigger-redaction` — none touched by this PR.

Green here and attributable to this PR: `feature-parity`, `test-unit (1)`, `test-integration (1)`.

0xdeafcafe:
**Validator round applied** (99779ae06d).

1. **Close guard re-armed.** The reset cleared `baselineRef`, but the only writer for a create's baseline runs on mount — and this transition is not a mount, so `isDirty` stayed false forever and closing a fully configured new automation discarded it silently. The reset now captures the baseline against the blank draft it just installed, which is exactly what the mount effect would have captured.
2. **Test replaced with the catching shape.** The old one closed a *pristine* create, which passed under both the fix and the bug. It now makes the fresh draft dirty, closes, expects "Discard unsaved changes?", and confirms Discard still closes. Verified red against the reported defect (`baselineRef.current = null`) before restoring.
3. **Reset moved above hydration.** Effects run in declaration order; hydration reading a draft the reset had not blanked yet took its "already editing" branch, latched, and never ran again. A test hands the second row over already cached and asserts it hydrates — verified red by relocating the effect back below hydration, which reproduced the blank Review exactly (`expected '' to be 'Second automation'`).
4. **webhook.ts whitespace churn reverted** — the file is back to its tabs, with only the one copy string changed. Same check run against `slack.ts`: one line there too.

Scoped runs after the round: 33 integration files / 302 tests, 18 unit files / 307 tests, tslsp clean, and `AutomationDrawer.tsx` warning counts identical rule-for-rule to the base (8 + 3), so no lint delta.

0xdeafcafe:
**Heads up: this PR merged one commit before the validator's fix round.**

#6907 merged at `0601a90263`. The validator's three items landed in `99779ae06d`, which is not in it — so `issue6716/bugbash-integration` currently carries the close-guard defect the validator called the real one: after edit→create, `baselineRef` stays null, `isDirty` is false forever, and closing a fully configured new automation discards it without a prompt.

The fix, its two red-verified tests and the `webhook.ts` whitespace revert are in **#6917**, based on this same branch.

On the red lint job here: it ran after the merge, and its own log says `comparing against merge base ee7965024d (origin/main)` — with the PR closed it resolved the base as `main` instead of the integration branch, so it attributed the whole workstream's accumulated violations to this diff. The five files it names (`ConditionBuilder.tsx`, `SubjectSection.tsx`, `WebhookDeliverySection.tsx`) are byte-identical between this branch and the real merge base, and the run's own totals show head at 3591 diagnostics against a base of 3592 — one fewer, not five more. Nothing to fix here.

PR #6908, opened 2026-08-12
fix(traces): deduplicate Codex exec token usage

### LINKED ISSUE #6742: codex exec turns count their tokens twice on codex 0.146
## What happens

A `codex exec` turn has its tokens counted twice. The trace's totals, the analytics graphs and the cost derived from them are all roughly double what the turn actually used.

Confirmed against codex 0.146 in a local dogfood of #6433:

```
TraceId                           SpanName           SpanId            in_tok  out_tok  skip
4c863d4e28e88c3f7b2cab3d9842bb5c  session_task.turn  f6e707822cf0dfd4  13590   51
4c863d4e28e88c3f7b2cab3d9842bb5c  handle_responses   e68744e2f48b285c  13590   51
```

Both spans carry identical usage, both are under scope `codex_exec`, and neither carries `langwatch.reserved.skip_token_accumulation`.

## Why

`markRedundantUsageSpan` in `canonicalisation/extractors/codex.ts` flags `handle_responses` as a duplicate of the `session_task.turn` rollup so the fold counts usage on exactly one of them. That marking is deliberately skipped for the `codex_exec` scope, on this stated assumption:

> The exec wire has NO `session_task.turn` rollup, `handle_responses` response spans are its only usage record, so the redundant-usage skip below must never fire for it.

That was true of an older codex. It is not true of 0.146, which emits the rollup on the exec wire as well, so the carve-out now protects nothing and lets both spans through.

## Why this is not a one-line fix

Deleting the `scopeName !== CODEX_EXEC_SCOPE_NAME` condition restores the correct total for 0.146, but reinstates the exact failure the carve-out was written to prevent: against any codex build that really does emit no rollup, the only usage-bearing span gets skipped and the trace reports zero tokens. Silently dropping usage is a worse failure than visibly doubling it, which is the tradeoff the original author called and it still holds.

The invariant that is actually wanted is cross-span: skip the response span when, and only when, the same trace also carries a turn rollup with usage. The extractor is per-span and cannot see that. The fold can, but `langwatch.reserved.skip_token_accumulation` is read in several places that would all have to agree:

- `traceAnalytics.foldProjection.ts`
- `projections/services/span-cost.service.ts`
- `analytics/clickhouse/aggregation-builder.ts`, where it is a SQL predicate (`SPAN_NOT_SKIPPED`)

So the fix is a deliberate change to how redundancy is decided, not a tweak to which scope is exempt.

## Suggested direction

Have the extractor mark `handle_responses` as a *candidate* duplicate regardless of scope, and let the fold resolve candidates: a candidate is dropped from token accumulation only if a `session_task.turn` with usage exists in the same trace, otherwise it counts. Keep the reserved attribute's current meaning for everything that reads it today, and add the candidate as a separate signal so the ClickHouse predicate does not change meaning underneath existing rows.

## How to know it worked

A `codex exec` turn reports the tokens codex itself reports for it, once. Worth a regression test built from the trace above, and one for the rollup-less shape so the zero-totals failure cannot come back.

## Related

- #6433, where this was found

prajvalrasik:
## Why

Closes #6742.

Codex 0.146 emits the same turn usage on both `handle_responses` and `session_task.turn`, which doubles trace totals, analytics graphs, and derived cost. Older `codex exec` traces only contain the response usage, so unconditionally skipping that span would instead reduce those traces to zero usage.

## What changed

- Mark exec response usage as a conditional candidate and usage-bearing turn rollups as the authority, while preserving the existing hard-skip contract for other emitters.
- Reconcile candidates and authorities in both trace folds with order-independent signed deltas. A response remains countable until an authority is observed, then the authoritative aggregate replaces it.
- Apply the same trace-wide authority rule to model-grouped ClickHouse queries before model bucketing.
- Keep fold bookkeeping out of public trace attributes and legacy metadata.
- Add regressions for both arrival orders, rollup-less exec traces, unequal observations, zero-token authorities, fold parity, query generation, and API boundaries.

## How it works

The extractor remains span-local: it only labels candidate and authority observations. The folds persist separate aggregate ledgers in reserved attributes, allowing a later-arriving authority to retract an earlier candidate contribution without changing stored per-span detail.

For analytics, token and cost metrics now use the one-row-per-trace `trace_analytics` fold. The current `trace_analytics_rollup` is populated independently per span and cannot retract a candidate when a sibling authority arrives. Completion-time queries continue using the rollup path.

## Test plan

Focused Vitest run covering seven suites:

```text
Test Files  7 passed (7)
Tests       219 passed (219)
```

The suites cover Codex canonicalization, trace-summary and slim-fold parity, analytics routing, ClickHouse query generation, and both public trace mapping boundaries.

Biome error-level checks also pass for all 16 changed TypeScript files.

## Anything surprising?

The routing change trades the rollup fast path for correctness on token and cost metrics. Supporting conditional sibling reconciliation in `trace_analytics_rollup` would require extending its schema and projection with separate candidate and authority columns; this change instead uses the already-correct trace-level fold.

## Summary by CodeRabbit

- **New Features**
  - Improved Codex token and cost accuracy by preventing duplicate usage counting when authoritative totals are available.
  - Preserved usage totals for execution traces without rollup data.
  - Kept internal token-accounting attributes out of trace metadata, headers, and metadata facets.

- **Bug Fixes**
  - Ensured analytics projections produce consistent totals regardless of span processing order.
  - Refined metric routing for candidate-sensitive analytics.
  - Improved handling of response-only usage reports and conditional duplicate usage.


langwatch-agent:
## Human Review Brief

**Mode: Deep Review.** Closes #6742 (`enhancement`, P3 - low). The label undersells it: the fix rewrites how token usage is reconciled across a trace fold, and it changes stored analytics and derived cost. P3 is right for the symptom, not for the blast radius.

### What you are looking at
Codex reports the same turn twice, and that is the whole reason a dedupe PR exists.

`platform/app/src/server/app-layer/traces/canonicalisation/extractors/codex.ts` claims two different wire shapes. On the log side it takes OpenTelemetry log records whose `event.name` starts with `codex.`, and the cost-bearing one is `codex.sse_event`, carrying model plus token counts plus the conversation id. On the span side, codex 0.137 and later emits native spans under the scope `codex_cli_rs`, where `session_task.turn` carries the same per-turn accounting as codex-namespaced attributes (`codex.turn.token_usage.input_tokens`, `output_tokens`, `cached_input_tokens`, `reasoning_output_tokens`, `total_tokens`). A CLI configured for both paths sends both, and the extractor lifts each to the same canonical `gen_ai.*` keys.

Downstream that lands in two places that both count. `span-cost.service.ts` prices spans, honouring a bundled-cost marker that can be set at the resource level by the receiver or on a single span by an extractor, which is how codex account-provider spans are marked as covered by a subscription rather than billed per token. And the fold projections, `traceSummary.foldProjection.ts` and `traceAnalytics.foldProjection.ts`, roll spans up into the summary columns and the analytics rows the usage screens read.

That is what makes double counting expensive rather than cosmetic: it inflates token and cost columns customers look at, and the projections are folded state, so a fix has to be right for traces already folded, not only for new ones. The PR touching `aggregation-builder.ts`, the route table and `metadata-keys.ts` alongside the extractor is consistent with that reach.

### Decisions being ratified

1. **A span can be a *candidate* whose contribution is retracted later.** Usage stops being additive-only. The fold now carries signed deltas so a later-arriving authority can undo an earlier count. That is a new invariant for every future emitter, not just Codex.
2. **The turn rollup is the authority; the response is countable only until one appears.** Old `codex exec` traces have no rollup, so they keep counting the response. The correctness of every historical trace now depends on that "until observed" rule holding for arrival orders nobody has seen yet.
3. **Fold bookkeeping lives in reserved attributes, out of public trace attributes and legacy metadata.** Reserved attributes become internal state with a compatibility obligation: anything reading spans raw must not see or depend on them.
4. **The same authority rule is applied to model-grouped ClickHouse queries before model bucketing.** Query-time and fold-time now have to agree, so the rule exists twice and can drift.

### Must Check

- **Arrival order and replay.** The PR covers both orders, but the real question is what a re-fold or backfill produces. If a trace is folded twice, do the signed deltas cancel correctly, or does a retraction get applied twice?
- **The ClickHouse side against the fold side.** Two implementations of one rule is the standing risk here. Confirm there is a test that would fail if only one of them changed.
- **Traces that are partially migrated.** A trace with some spans emitted by codex 0.146 and some by an older version is the case that produces neither doubling nor zero, and is the hardest to notice.

### Ask Author

- Does this correct existing stored data, or only new traces? If only new, is the double-counted history visible in analytics as a step change, and does anyone need to be told?
- Zero-token authorities are covered in tests. What does the fold do when the authority reports fewer tokens than the candidate already contributed?

### Production risk

Derived cost is customer-facing. A retraction bug undercounts silently, which is worse than the doubling it replaces because nobody files a ticket about a bill that looks too low.

langwatch-agent (commented):
No findings. The fold-level signed replacement and the model-partition query both preserve candidates without an authority and replace them when one is observed; hard-skipped authorities are excluded on both paths. Residual validation gap: ClickHouse behavior is covered by generated-SQL assertions, not an execution-level parity fixture.

LangWatch-Review: verdict=clean sha=94f4516ea25b4295d8a07c0defc24c106a9d98de p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=94f4516ea25b4295d8a07c0defc24c106a9d98de p0=0 p1=0 p2=0 p3=0

Issue #6920, opened 2026-08-13
Prompt playground cannot reliably open generated traces or leave the page


sergioestebance:
**Describe the bug**

A completed prompt-playground response can show a “View Trace” action whose trace lookup returns `NOT_FOUND`, leaving no trace drawer to inspect. In the same session, choosing the Trace Explorer from the project sidebar can update the address bar while the prompt playground remains visibly mounted.

The cause is not yet established. The investigation must determine whether these are separate defects or two manifestations of the same React Router/Suspense transition problem.

**To reproduce**

1. Open a project’s Prompt Playground and select or create a prompt.
2. Send a chat message and wait for a successful assistant response.
3. Click “View Trace” beneath that response.
4. Observe that the trace drawer may not open and `traces.getById` may return `NOT_FOUND` for the streamed message ID.
5. From the same page, choose “Trace Explorer” in the sidebar.
6. Observe that the URL may change to the traces route while the prompt playground remains visible.

**Expected behavior**

Every completed playground response that exposes “View Trace” resolves to the trace created for that execution and opens the Trace Explorer drawer. Choosing Trace Explorer from the sidebar replaces the prompt playground with the traces page.

**Environment**

- Component: application
- Version: current
- Browser/OS: browser-independent until disproved

**Additional context**

Relevant existing behavior and prior work:

- `PromptStudioAdapter` allocates one trace ID and uses it both as the workflow `trace_id` and the streamed CopilotKit `messageId`.
- `PromptPlaygroundChat` passes the assistant message ID to `TraceMessage` as the trace ID.
- `TraceMessage` polls `traces.getById` before displaying the action.
- The drawer router redirects legacy `traceDetails` requests to `traceV2Details` outside the legacy traces page.
- Prior drawer-navigation work established that transition-wrapped lazy drawer navigation can update the URL while retaining previously committed UI; drawer navigations now request `flushSync`.
- Prior trace-ID work preserved the preallocated ID on early adapter errors.

What will settle the investigation:

1. Trace one locally generated ID through adapter allocation, backend execution, ingestion, `traces.getById`, and the v2 drawer read.
2. Reproduce the sidebar transition using the real lazy-loaded prompt and traces route modules.
3. Show whether removing the suspected transition/readiness condition removes each symptom.
4. Record competing hypotheses and the evidence that rules them out.

No implementation changes belong in this issue until the cause is validated.

sergioestebance:
## Investigation contract

[...]

PR #6927, opened 2026-08-13
fix(app): keep route URL and rendered view in sync

sergioestebance:
## For humans

Clicking Trace Explorer in the sidebar could change the address bar while the prompt playground stayed on screen — the URL said you had moved, the page said you had not. This makes the page always follow the URL. It also fixes the local development stack, which could not start the application server at all.

## Summary

- disable transition-wrapped React Router state updates at the application root
- keep the committed route tree synchronized with browser history for lazy routes and drawers that subscribe to external stores
- fix the local dev orchestrator (haven), whose app and workers lanes ran the production entry points
- add regression tests for both

## Root cause

**Route/URL desync.** React Router schedules route-state updates in a React transition by default. In this application, lazy route and drawer trees also subscribe to external stores. Under contention, browser history could advance while React kept the previous tree committed, producing a new URL with the old page still visible and making a closed drawer fail to reopen.

**Dev stack.** `tools/thuishaven/app/plan.go` ran `pnpm run start:app` and `pnpm run start:workers` for the Node lanes. Those are the production entry points (`node dist/server/*.cjs`) and nothing in a dev worktree builds those bundles, so the api lane crash-looped on `MODULE_NOT_FOUND`. Because the app lane is gated on the API's `/api/health` probe, the stack came up with no application served and no obvious cause in its own output. `scripts/start.sh` already makes this split off `NODE_ENV`; these lanes bypass `start.sh`, so they now pick the dev entry points themselves.

## Scope note

This closes the sidebar-navigation symptom in #6920. The other symptom reported there — `traces.getById` returning `NOT_FOUND` for a preallocated playground trace ID — is a separate ingestion-timing question and is **not** addressed here.

## Verification

- regression test failed before the fix and passes after it
- `TestNodeLanesUseDevEntryPoints` resolves each lane's pnpm script against the real `platform/app/package.json` and fails when a lane points at a script that loads a build artifact; verified failing against the pre-fix lane
- focused unit tests: 13 passed
- application typecheck passed
- Biome passed with no new violations
- `golangci-lint run ./tools/thuishaven/...` reports nothing new on the changed files
- `make haven up` now serves the app (HTTP 200) where it previously never came up
- verified in the local app that Prompt Playground → Trace Explorer commits the new view, and a trace drawer opens, closes, and reopens

### DIFF
--- a/platform/app/src/main.test.tsx
+++ b/platform/app/src/main.test.tsx
@@ -1,0 +1,52 @@
+// @vitest-environment jsdom
+
+import type { ReactElement, ReactNode } from "react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  rootRender: vi.fn(),
+  createRoot: vi.fn(),
+}));
+
+vi.mock("react-dom/client", () => ({
+  createRoot: mocks.createRoot,
+}));
+
+vi.mock("./AppProviders", () => ({
+  OuterProviders: ({ children }: { children: ReactNode }) => children,
+}));
+
+vi.mock("./routes", () => ({
+  router: { id: "app-router" },
+}));
+
+vi.mock("./utils/chunkReload", () => ({
+  registerChunkReloadListener: vi.fn(),
+}));
+
+vi.mock("./utils/compat/next-router", () => ({
+  setRouterInstance: vi.fn(),
+}));
+
+describe("application router scheduling", () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mocks.rootRender.mockReset();
+    mocks.createRoot.mockReset();
+    mocks.createRoot.mockReturnValue({ render: mocks.rootRender });
+    document.body.innerHTML = '<div id="root"></div>';
+  });
+
+  it("commits route changes synchronously so the URL cannot outrun the page", async () => {
+    await import("./main");
+
+    const app = mocks.rootRender.mock.calls[0]?.[0] as ReactElement<{
+      children: ReactElement<{
+        children: ReactElement<{ useTransitions?: boolean }>;
+      }>;
+    }>;
+    const routerProvider = app.props.children.props.children;
+
+    expect(routerProvider.props.useTransitions).toBe(false);
+  });
+});
--- a/platform/app/src/main.tsx
+++ b/platform/app/src/main.tsx
@@ -23,7 +23,9 @@ if (!container) throw new Error("Root element not found");
 createRoot(container).render(
   <OuterProviders>
     <Suspense fallback={null}>
-      <RouterProvider router={router} />
+      {/* Lazy routes and drawers subscribe to external stores. Transition-wrapped
+          navigation can advance history while leaving the old tree committed. */}
+      <RouterProvider router={router} useTransitions={false} />
     </Suspense>
   </OuterProviders>,
 );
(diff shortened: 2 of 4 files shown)


sergioestebance:
## Ruthless review — haven dev-lane entry points (`225c4ccefc`)

Scope: `tools/thuishaven/app/plan.go` + `plan_entrypoints_test.go`. Self-authored, so findings were fixed here rather than handed back; the record of what was wrong stays below.

**No P0. No P1.** One P2 found and fixed in `83f03a30ea`; the rest are P3 residuals.

### Verified by execution, not by reading

Both lanes, same contrast, on this worktree:

```
$ pnpm -s run start:workers          # the entry point before this change
Error: Cannot find module '.../platform/app/dist/server/workers.cjs'   # dies immediately

$ pnpm -s run start:workers:dev      # the entry point after
◇ injected env (7) from .env
◇ injected env (39) from .env.portless
… still alive at 40s (killed by timeout, exit 124)
```

`dist/server/server.cjs` is bundled *from* `src/server.mts` and `workers.cjs` from `src/workers.ts` (`scripts/build-server.mjs:132-133`), so the dev and prod scripts run the same entry module — one through tsx, one bundled. `WORKERS_IN_PROCESS` is read in `src/start.ts:151`, downstream of both. No behavioural divergence beyond the bundling, so the api lane keeps hosting workers in-process exactly as before.

Stack confirmed up end to end after the change: app 200, `/api/health` 204.

### P2 — the guard skipped lanes it could not parse *(fixed, `83f03a30ea`)*

`plan_entrypoints_test.go:43` — a lane whose shell the regex could not read fell through `continue`, and the `checked == 0` backstop only fires when *every* lane is unreadable. So a lane rewritten as `pnpm --filter @langwatch/app run start:app` would go unchecked while the other lane held the count above zero, and the test would pass on a prod bundle — the exact regression it exists to stop.

An unparsable pnpm lane now fails. Proven both directions:

```
--- FAIL: TestNodeLanesUseDevEntryPoints
    plan_entrypoints_test.go:52: lane "api" runs pnpm ("pnpm --filter @langwatch/app run start:app")
    in a form this test cannot parse, so its entry point went unchecked; widen pnpmRun
--- restored ---
ok  github.com/langwatch/langwatch/tools/thuishaven/app  1.296s
```

### P3 — residual risks, not fixed

- **The workers lane is verified to boot, not to consume.** 40s alive with env loaded and no crash, but nothing logged a queue or consumer line. It is opt-in (`haven up +workers`; the default selection is `workers:false` / in-process) and it is the same entry `scripts/start.sh:122` uses for dev, so confidence is reasonable — but "workers process jobs" is untested here, and the commit message should not be read as claiming it.
- **A delegating script would slip through.** The check greps the resolved script body for `dist/`; a script that shells out to another script would not be followed. Direct-entry drift — the failure that actually happened — is covered.
- **`gocognit` on `planChildren`** is pre-existing: identical `golangci-lint` output with the change stashed and unstashed, and the same warning covers ten other funcs in the package (`daemon.go`, `db.go`, `hmr.go`, …). Not introduced here.

CI coverage confirmed: `go-ci.yaml:132` runs `./tools/thuishaven/...` with `-race`, and the path filter (`**/*.go`, `tools/**`) matches, so this guard runs on every Go-touching PR. Full suite green under `-count=1 -race`; `gofmt` clean.

### Blessing

Blessed. 🙏 The dev stack now starts from the source it actually has, instead of crash-looping on a bundle no worktree ever builds — and a test holds that door shut.

> "Every one that heareth these words of mine, and doeth them not, shall be likened to a foolish man that built his house upon the sand." — Matthew 7:26

Traced both entry points to their shared source module, executed the crash and the fix side by side, and pinned the guard's own blind spot. Ship it — residual is the unexercised workers lane, called out above.

The approval here is the human's to give: this is my own branch, and fixing my own findings is efficiency, blessing them would not be.

